PR #1321, opened 2023-04-10
[VL] package.sh support centos7 and centos8

zhejiangxiaomai:
1. support package thirdparty libs for centos7 and centos8.
2. use can run on centos7/centos8 with one command.
3. package.sh is just run on clean docker. user can change the path of lib for themselves. And package.sh is only work for velox now.
## How was this patch tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,19 @@ There are several key component in Gluten:
 
 # 3 Usage
 
-Gluten is still under active development now. There isn't a released binary yet. The only way to use Gluten is to build from source, copy the jar to your spark jars, then enable Gluten plugin when you start your spark context. Here is the simple example. Refer to Velox or Clickhouse backend below for more details
+Gluten is still under active development now. There are two ways to use Gluten. 
+One Way is use released binary jar. Here is the simple example.
+
+```
+spark-shell --name run_gluten \
+ --master yarn --deploy-mode client \
+ --conf spark.plugins=io.glutenproject.GlutenPlugin \
+ --conf spark.gluten.sql.columnar.backend.lib=velox \
+ --conf spark.gluten.loadLibFromJar=true \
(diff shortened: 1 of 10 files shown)


backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxInitializerApi.scala
```diff
@@ -29,10 +29,16 @@ class VeloxInitializerApi extends InitializerApi {
   def loadLibFromJar(load: JniLibLoader): Unit = {
       val system = "cat /etc/os-release".!!
       if (system.contains("Ubuntu") && system.contains("20.04")) {
```
jinchengchenghh, line 31: val loader = if () {
new VeloxSharedlibraryLoaderUbuntu2004
}

loader.loadLib(load)

zhejiangxiaomai: Thanks, just addressed it.

jinchengchenghh (dismissed):

README.md
```diff
@@ -69,7 +86,7 @@ spark-shell
   --conf spark.executor.extraClassPath=${gluten_jvm_jar} \
   --conf spark.shuffle.manager=org.apache.spark.shuffle.sort.ColumnarShuffleManager \
   ...
-  ```
+```
 
 ## 3.1 Build and Install Gluten with Velox backend
```
zhouyuan, line 91: the format is not correct here, it has two "3.1" bullet

zhejiangxiaomai: I will change 
"### 3.1 Build and Install Gluten with Velox backend" to “### 3.2.1 Build and Install Gluten with Velox backend”.

jinchengchenghh (approved):

README.md
weiting-chen, line 1: Current prebuilt jar can only support for velox, could you add a note to describe this.

dev/package.sh
weiting-chen, line 1: Current package location is in fixed location such as /usr/lib64 or /usr/local/lib, check if we can add a description to explain if the users want to change the location to their custom location just in case.

zhejiangxiaomai: Thanks, I added some details in this pr's description.

PR #1332, opened 2023-04-11
[GLUTEN-1331] [CORE] Enable some functions

jinchengchenghh:


rui-mo (approved):

rui-mo:
Is any unit test to verify these functions?

jinchengchenghh:
Array type does not support in ProjectNode, I just enable it, but it can not offload to gluten, I just enable it.
Other functions are tracked by Spark UT

PR #1345, opened 2023-04-13
[GLUTEN-1404][CH] Bug fix no result returned in max/min/sum/count while select from empty table/partition

KevinyhZou:
Bug fix no result returned in max/min/sum/count while select from empty table/partition

the related pr of clickhouse backend:  https://github.com/Kyligence/ClickHouse/pull/424
the related issue: https://github.com/Kyligence/ClickHouse/issues/425

fix issue #1404 

## How was this patch tested?
by unit test


liuneng1994 (approved):
LGTM

Issue #1348, opened 2023-04-13
Fallback if projections in Expand node have scalar function 


exmy:
**Describe the solution you'd like**

 Fallback if projections in Expand node have scalar function, which is not supported currently.

This will happen in such sql: 
```sql
select a, count(distinct b), count(distinct concat(b, c)) from tbl group by a
```

PR #1349, opened 2023-04-13
[GLUTEN-1348][CORE] Fallback if projections in Expand node have scalar function

exmy:

(Fixes: #1348)

## How was this patch tested?

manual test

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
@@ -227,10 +227,13 @@ case class ExpandExecTransformer(projections: Seq[Seq[Expression]],
     }
 
     // Fallback if projections have scalar function, which is not supported currently.
-    if (projections.exists(
-          _.exists(expr => !expr.isInstanceOf[Attribute] && !expr.isInstanceOf[Literal]))) {
-      logWarning("There are scalar functions in expand node, which is not supported currently.")
-      return false
+    if (BackendsApiManager.getBackendName.equalsIgnoreCase(
+      GlutenConfig.GLUTEN_CLICKHOUSE_BACKEND)) {
+      if (projections.exists(
+        _.exists(expr => !expr.isInstanceOf[Attribute] && !expr.isInstanceOf[Literal]))) {
+        logWarning("There are scalar functions in expand node, which is not supported currently.")
+        return false
+      }
     }
 
     // FIXME.
--- a/gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
@@ -185,10 +185,13 @@ case class ExpandExecTransformer(projections: Seq[Seq[Expression]],
 
       // Add groupID col index
       if (SQLConf.get.integerGroupingIdEnabled) {
+        // scalastyle:off
         // When 'integerGroupingIdEnabled' set, groupID is integer type but velox always gerenate long type value.
         // Convert result to fix test case 'SPARK-30279 Support 32 or more grouping attributes for GROUPING_ID()'.
+        // scalastyle:on
         val typeNode = ConverterUtils.getTypeNode(DataTypes.IntegerType, false)
-        selectNodes.add(ExpressionBuilder.makeCast(typeNode, ExpressionBuilder.makeSelection(projections(0).size - 1), SQLConf.get.ansiEnabled))
+        selectNodes.add(ExpressionBuilder.makeCast(typeNode,
+          ExpressionBuilder.makeSelection(projections(0).size - 1), SQLConf.get.ansiEnabled))
       } else {
         selectNodes.add(ExpressionBuilder.makeSelection(projections(0).size - 1))
       }
@@ -220,6 +223,13 @@ case class ExpandExecTransformer(projections: Seq[Seq[Expression]],
       return false
     }
     if (projections.isEmpty) {
+      return false
+    }
+
+    // Fallback if projections have scalar function, which is not supported currently.
+    if (projections.exists(
+          _.exists(expr => !expr.isInstanceOf[Attribute] && !expr.isInstanceOf[Literal]))) {
+      logWarning("There are scalar functions in expand node, which is not supported currently.")
       return false
     }
 


gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
```diff
@@ -223,6 +226,13 @@ case class ExpandExecTransformer(projections: Seq[Seq[Expression]],
       return false
     }
 
+    // Fallback if projections have scalar function, which is not supported currently.
+    if (projections.exists(
+          _.exists(expr => !expr.isInstanceOf[Attribute] && !expr.isInstanceOf[Literal]))) {
+      logWarning("There are scalar functions in expand node, which is not supported currently.")
+      return false
```
rui-mo, line 233: @JkSelf Do we have such validation in our native validator for Expand?

exmy:
@JkSelf By the way, I am a little confused about the `ExpandExecTransformer`.
```scala
projections.foreach { projection =>
  val groupExprNodes = new util.ArrayList[ExpressionNode]()
  for (i <- aggSize until (projection.size - 1)) {
    if (!(projection(i).isInstanceOf[Literal] &&
      projection(i).asInstanceOf[Literal].value == null)) {
      var groupExprNode = ExpressionConverter
        .replaceWithExpressionTransformer(
          projection(i),
          originalInputAttributes
        )
```
Here, projections that are not null literals are transformed, which may cause problems in the following case
```shell
ExpandExecTransformer [[uid#12L, dispatch_id#27, null, 1], [uid#12L, null, video_id#23L, 2]]
       ProjectExecTransformer [uid#12L, video_id#23L, dispatch_id#27]
```
in this case, groupsetExprNodes is transformed to something like  `[ [0, 2], [0, 1] ]`, and in the backend it may parsed as `[uid, null, dispatch_id, gid], [uid, video_id, null, gid]`, which is not consistent with original projections. 

Do I understand it right?

JkSelf:
@exmy I don't understand why the `groupsetExprNodes `is transformed to `[ [0, 2], [0, 1] ]`.

This code is to calculate the group cols and filter the gid col. Here the `groupsetExprNodes ` cols is [dispatch_id] and [video_id] for your example. And the child output is `[uid, dispatch_id, video_id]`. Velox backend will judge whether the group key exist in the output. The `groupsets `default value is null and will be set to the group key if exist. Then we will get the value is `[dispatch_id, null]` and `[null, video_id]`.

gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
rui-mo: Looks Velox backend does not need this, is that right?

FelixYBW:
If we solve the fallback, should we add a project before expand? Does Velox's expand support expression?

exmy:
> @exmy I don't understand why the `groupsetExprNodes `is transformed to `[ [0, 2], [0, 1] ]`.
> 
> This code is to calculate the group cols and filter the gid col. Here the `groupsetExprNodes ` cols is [dispatch_id] and [video_id] for your example. And the child output is `[uid, dispatch_id, video_id]`. Velox backend will judge whether the group key exist in the output. The `groupsets `default value is null and will be set to the group key if exist. Then we will get the value is `[dispatch_id, null]` and `[null, video_id]`.

@JkSelf  Sorry if I didn't express myself clearly. What I meant was that `[ [0, 2], [0, 1] ]` is the SelectionNode for the child output references, for example:
```json
"groupings": [{
    "groupSetsExpressions": [{
      "selection": {
        "directReference": {
          "structField": {
          }
        }
      }
    }, {
      "selection": {
        "directReference": {
          "structField": {
            "field": 2
          }
        }
      }
    }]
  }, {
    "groupSetsExpressions": [{
      "selection": {
        "directReference": {
          "structField": {
          }
        }
      }
    }, {
      "selection": {
        "directReference": {
          "structField": {
            "field": 1
          }
        }
      }
    }]
  }]
```
And in this case, the child output is actually `[uid, video_id, dispatch_id]`. The issue is caused by the difference in order between `video_id` and `dispatch_id`, which is also why there may be parsing problems at least in the clickhouse backend. I am not sure if velox also has this issue.

gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
exmy: > Looks Velox backend does not need this, is that right?

ok, have made the check only in CH backend

zhli1142015:
> > @exmy I don't understand why the `groupsetExprNodes `is transformed to `[ [0, 2], [0, 1] ]`.
> > This code is to calculate the group cols and filter the gid col. Here the `groupsetExprNodes ` cols is [dispatch_id] and [video_id] for your example. And the child output is `[uid, dispatch_id, video_id]`. Velox backend will judge whether the group key exist in the output. The `groupsets `default value is null and will be set to the group key if exist. Then we will get the value is `[dispatch_id, null]` and `[null, video_id]`.
> 
> @JkSelf Sorry if I didn't express myself clearly. What I meant was that `[ [0, 2], [0, 1] ]` is the SelectionNode for the child output references, for example:
> 
> ```json
> "groupings": [{
>     "groupSetsExpressions": [{
>       "selection": {
>         "directReference": {
>           "structField": {
>           }
>         }
>       }
>     }, {
>       "selection": {
>         "directReference": {
>           "structField": {
>             "field": 2
>           }
>         }
>       }
>     }]
>   }, {
>     "groupSetsExpressions": [{
>       "selection": {
>         "directReference": {
>           "structField": {
>           }
>         }
>       }
>     }, {
>       "selection": {
>         "directReference": {
>           "structField": {
>             "field": 1
>           }
>         }
>       }
>     }]
>   }]
> ```
> 
> And in this case, the child output is actually `[uid, video_id, dispatch_id]`. The issue is caused by the difference in order between `video_id` and `dispatch_id`, which is also why there may be parsing problems at least in the clickhouse backend. I am not sure if velox also has this issue.

Hello @exmy, I raised a PR to fix a similar problem for velox, the basic idea is to pass the projections to BE, no need to distinguish the agg cols and group cols. I think this behavior aligns more with spark ExpandExec. I'm verifying it with Velox. Please let me know if this can also help for CH, maybe we can corrdinate the both velox change and CH change.
Thanks.
https://github.com/oap-project/gluten/pull/1361

exmy:
> Hello @exmy, I raised a PR to fix a similar problem for velox, the basic idea is to pass the projections to BE, no need to distinguish the agg cols and group cols. I think this behavior aligns more with spark ExpandExec. I'm verifying it with Velox. Please let me know if this can also help for CH, maybe we can corrdinate the both velox change and CH change. Thanks. #1361

Looks great, it can help for CH. Looking forward to it very much~

PR #1358, opened 2023-04-14
[GLUTEN-1357][CORE] Change soft-affinity log level from INFO to DEBUG

jackylee-ch:
The job gets stuck in log printing for a long time while trying to read a large amount of files, thus we add a config to control the loglevel for soft-affinity.

## How was this patch tested?
unit tests.


zzcclp (approved):
LGTM

PR #1361, opened 2023-04-14
[GLUTEN-842][VL] convert expand op to expand exec in velox

zhli1142015:

Here are project expressions we observed in expand operation:

agg exprs + group by exprs + gid.
agg exprs + group by exprs + gid + _gen_grouping_pos. --> The last column is for handling duplicate grouping sets.
group by exprs + gid + agg exprs. --> gid is calculated by different way from above two cases. it's assigned with the sequence number of project set.

Original ExpandExecTrandofrmer can only handle the first case. I'm adding the expand exec in velox side by the PR: https://github.com/oap-project/velox/pull/199/files. This PR is for converting spark ExpandExec to Expand OP in Velox, We don't need to do the columns' mapping in Gluten.

The original ExpandExecTrandofrmer is renamed to GroupIdExecTrandofrmer to not break ClicKHouse.

## How was this patch tested?

Unit test.

### DIFF
--- a/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
+++ b/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
@@ -358,17 +358,12 @@ message ExchangeRel {
 message ExpandRel {
   RelCommon common = 1;
   Rel input = 2;
+.
+  repeated Projections projections = 3;
 
-  repeated Expression aggregate_expressions = 3;
-
-  // A list of expression grouping that the aggregation measured should be calculated for.
-  repeated GroupSets groupings = 4;
-
-  message GroupSets {
-    repeated Expression groupSets_expressions = 1;
+  message Projections {
+    repeated Expression project_expressions = 1;
   }
-
-  string group_name = 5;
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
 }
--- a/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
+++ b/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
@@ -355,14 +355,32 @@ message ExchangeRel {
   }
 }
 
+message GroupIdRel {
+  RelCommon common = 1;
+  Rel input = 2;
+
+  repeated Expression aggregate_expressions = 3;
+
+  // A list of expression grouping that the aggregation measured should be calculated for.
+  repeated GroupSets groupings = 4;
+
+  message GroupSets {
+    repeated Expression groupSets_expressions = 1;
+  }
+
+  string group_name = 5;
+
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
 message ExpandRel {
   RelCommon common = 1;
   Rel input = 2;
-.
-  repeated Projections projections = 3;
 
-  message Projections {
-    repeated Expression project_expressions = 1;
+  repeated ProjectionSets projections = 3;
+
+  message ProjectionSets {
+    repeated Expression projectionSets_expressions = 1;
   }
 
   substrait.extensions.AdvancedExtension advanced_extension = 10;
@@ -396,9 +414,10 @@ message Rel {
     //Physical relations
     HashJoinRel hash_join = 13;
     MergeJoinRel merge_join = 14;
-    ExpandRel expand = 15;
+    GroupIdRel group_id = 15;
     WindowRel window = 16;
     GenerateRel generate = 17;
+    ExpandRel expand = 18;
   }
 }
 
--- a/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
+++ b/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
@@ -355,6 +355,35 @@ message ExchangeRel {
   }
 }
 
+// Duplicates records, possibly switching output expressions between each duplicate.
+// Default output is all of the fields declared followed by one int64 field that contains the
+// duplicate_id which is a zero-index ordinal of which duplicate of the original record this
+// corresponds to.
+message ExpandRel {
+  RelCommon common = 1;
+  Rel input = 2;
+  repeated ExpandField fields = 4;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+  message ExpandField {
+    oneof field_type {
+      // Field that switches output based on which duplicate_id we're outputting
+      SwitchingField switching_field = 2;
+
+      // Field that outputs the same value no matter which duplicate_id we're on.
+      Expression consistent_field = 3;
+    }
+  }
+
+  message SwitchingField {
+    // Array that contains an expression to output per duplicate_id
+    // each `switching_field` must have the same number of expressions
+    // all expressions within a switching field be the same type class but can differ in nullability.
+    // this column will be nullable if any of the expressions are nullable.
+    repeated Expression duplicates = 1;
+  }
+}
+
 message GroupIdRel {
   RelCommon common = 1;
   Rel input = 2;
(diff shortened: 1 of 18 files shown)


JkSelf:
@baibaichen 

This PR is to implement ExpandTransformer similar with vanilla spark. And then we no need to distinguish the agg cols and group cols in Gluten. It seems the Expand [PR](https://github.com/substrait-io/substrait/pull/368) to substrait community is ready to merge except the doc update. How about directly use  [PR](https://github.com/substrait-io/substrait/pull/368) in Gluten to support Expand  in this PR?

FelixYBW:
> @baibaichen
> 
> This PR is to implement ExpandTransformer similar with vanilla spark. And then we no need to distinguish the agg cols and group cols in Gluten. It seems the Expand [PR](https://github.com/substrait-io/substrait/pull/368) to substrait community is ready to merge except the doc update. How about directly use [PR](https://github.com/substrait-io/substrait/pull/368) in Gluten to support Expand in this PR?


We should follow substrait's solution. Does the PR need change to substrait?

JkSelf:
> > @baibaichen
> > This PR is to implement ExpandTransformer similar with vanilla spark. And then we no need to distinguish the agg cols and group cols in Gluten. It seems the Expand [PR](https://github.com/substrait-io/substrait/pull/368) to substrait community is ready to merge except the doc update. How about directly use [PR](https://github.com/substrait-io/substrait/pull/368) in Gluten to support Expand in this PR?
> 
> We should follow substrait's solution. Does the PR need change to substrait?

Yes. This PR add new ExpandRel message in algebra.proto, which follow up the substrait community solution except the definition in [PR](https://github.com/substrait-io/substrait/pull/368).

zhli1142015:
> > > @baibaichen
> > > This PR is to implement ExpandTransformer similar with vanilla spark. And then we no need to distinguish the agg cols and group cols in Gluten. It seems the Expand [PR](https://github.com/substrait-io/substrait/pull/368) to substrait community is ready to merge except the doc update. How about directly use [PR](https://github.com/substrait-io/substrait/pull/368) in Gluten to support Expand in this PR?
> > 
> > 
> > We should follow substrait's solution. Does the PR need change to substrait?
> 
> Yes. This PR add new ExpandRel message in algebra.proto, which follow up the substrait community solution except the definition in [PR](https://github.com/substrait-io/substrait/pull/368).

Thanks, updated.

zhli1142015:
@JkSelf , @zhouyuan  and @lgbo-ustc, could you please help to take a look?

gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
```diff
@@ -84,160 +81,148 @@ case class ExpandExecTransformer(projections: Seq[Seq[Expression]],
...
-        .replaceWithExpressionTransformer(
-          projections.head(i),
-          originalInputAttributes)
-        .doTransform(args)
-      aggExprNodes.add(aggExprNode)
+    def needsPreProjection(projections: Seq[Seq[Expression]]): Boolean = {
+      projections
+        .exists(set => set.exists(p => !p.isInstanceOf[Attribute] && !p.isInstanceOf[Literal]))
     }
+    if (needsPreProjection(projections)) {
```
JkSelf, line 93: Can we add simple unit test for this case?

JkSelf (dismissed):
LGTM except small comments.

FelixYBW (approved):

gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
zhli1142015: Added, thanks.
image

lgbo-ustc:
It seems the `ExpandRel` has been changed, is it backward compatible with `ClickhHouse`?

zhli1142015:
> It seems the `ExpandRel` has been changed, is it backward compatible with `ClickhHouse`?

new 'ExpandRel' is different and can't be compatible with CH. The original Rel is renamed to 'GroupIdRel' and changes are made in CH also.
Thanks.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -355,9 +355,38 @@ message ExchangeRel {
...
   Rel input = 2;
+  repeated ExpandField fields = 4;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+
+  message ExpandField {
+    oneof field_type {
+      // Field that switches output based on which duplicate_id we're outputting
+      SwitchingField switching_field = 2;
+
+      // Field that outputs the same value no matter which duplicate_id we're on.
```
lgbo-ustc, line 373: Is there case that this will be used?

lgbo-ustc (commented):
updated

zhli1142015:
> It's greate. This implementation is simple. I think it could solve some problems we have meet.
> 
> LGTM.

Thanks for review @lgbo-ustc , will you work on CH side to consume new ExpandRel contract?

lgbo-ustc:
> > It's greate. This implementation is simple. I think it could solve some problems we have meet.
> > LGTM.
> 
> Thanks for review @lgbo-ustc , will you work on CH side to consume new ExpandRel contract?

We will do it soon

zhouyuan (approved):
👍

zhanglistar:
@zhli1142015 

```
agg exprs + group by exprs + gid + _gen_grouping_pos. --> The last column is for handling duplicate grouping sets.
group by exprs + gid + agg exprs. --> gid is calculated by different way from above two cases. it's assigned with the sequence number of project set.
```
I have a question, the two cases, could you give some sql? We don't know when spark will generate the two cases. 
Thanks.

zhli1142015:
> @zhli1142015
> 
> ```
> agg exprs + group by exprs + gid + _gen_grouping_pos. --> The last column is for handling duplicate grouping sets.
> group by exprs + gid + agg exprs. --> gid is calculated by different way from above two cases. it's assigned with the sequence number of project set.
> ```
> 
> I have a question, the two cases, could you give some sql? We don't know when spark will generate the two cases. Thanks.

Hello @zhanglistar ,
Please check below sample code:
```
case class TestData3(a: Int, b: Option[Int])
val df = spark.sparkContext.parallelize(
      TestData3(1, None) ::
      TestData3(2, Some(2)) :: Nil).toDF()
import org.apache.spark.sql.functions._
df.agg(count($"a"), count($"b"), count(lit(1)), count_distinct($"a"), count_distinct($"b")).collect // case 3
df.createOrReplaceTempView("df")
spark.sql("select count(a) from df group by grouping sets((a), (a), (b))").collect // case 2
```

zhanglistar:
> > @zhli1142015
> > ```
> > agg exprs + group by exprs + gid + _gen_grouping_pos. --> The last column is for handling duplicate grouping sets.
> > group by exprs + gid + agg exprs. --> gid is calculated by different way from above two cases. it's assigned with the sequence number of project set.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I have a question, the two cases, could you give some sql? We don't know when spark will generate the two cases. Thanks.
> 
> Hello @zhanglistar , Please check below sample code:
> 
> ```
> case class TestData3(a: Int, b: Option[Int])
> val df = spark.sparkContext.parallelize(
>       TestData3(1, None) ::
>       TestData3(2, Some(2)) :: Nil).toDF()
> import org.apache.spark.sql.functions._
> df.agg(count($"a"), count($"b"), count(lit(1)), count_distinct($"a"), count_distinct($"b")).collect // case 3
> df.createOrReplaceTempView("df")
> spark.sql("select count(a) from df group by grouping sets((a), (a), (b))").collect // case 2
> ```

@zhli1142015 Thanks! For the sql `spark.sql("select count(a) from df group by grouping sets((a), (a), (b))").collect // case 2`, why not just duplicate the two grouping sets (a) for optimization?

PR #1375, opened 2023-04-17
[GLUTEN-1374] [VL] RangePartitioning supports velox columnar batch

jinchengchenghh:


cpp/velox/memory/VeloxColumnarBatch.cc
```diff
@@ -1,8 +1,40 @@
...
+// destroyed input vector
+RowVectorPtr addColumnToVector(int32_t index, RowVectorPtr vector, RowVectorPtr col) {
+  auto names = asRowType(vector->type())->names();
+  auto newNames = std::move(names);
+  newNames.insert(newNames.begin() + index, asRowType(col->type())->nameOf(0));
+
+  auto childern = vector->children();
+  auto newChildren = std::move(childern);
+  newChildren.insert(newChildren.begin() + index, col->childAt(0));
+  return makeRowVector(newNames, newChildren, vector->pool());
```
FelixYBW, line 35: The children vector are used directly, not memcpy here, right?

jinchengchenghh: Yes, I will release the input 2 vectors after add column to generate the new Columnbatch

cpp/velox/memory/VeloxColumnarBatch.h
```diff
@@ -35,6 +35,8 @@ class VeloxColumnarBatch final : public ColumnarBatch {
 
   int64_t GetBytes() override;
 
+  std::shared_ptr<ColumnarBatch> addColumn(int32_t index, std::shared_ptr<ColumnarBatch> col) override;
+
```
zhztheplayer, line 39: Probably we can make some new designs to avoid losing immutability of the columnar batch.

This is what came to my mind (I'll omit the use of smart pointers since I'm just drafting):

1. Add function `gluten::ToVeloxColumnarBatch(gluten::ColumnarBatch batch)` in file `VeloxColumnarBatch.h`;
2. Add utility class `gluten::CompositeColumnarBatch` that extends `gluten::ColumnarBatch` in `ColumnarBatch.h`
   * The constructor of `gluten::CompositeColumnarBatch` can be `CompositeColumnarBatch(gluten::ColumnarBatch one, gluten::ColumnarBatch other)`, or `CompositeColumnarBatch(std::vector<gluten::ColumnarBatch> batches)`
 3. In the new function `gluten::ToVeloxColumnarBatch` mentioned in 1, use some kind of visitor to convert all kinds of columnar batches including `gluten::CompositeColumnarBatch` to `gluten::VeloxColumnarBatch` without extra copy-paste on the internal Velox columnar batch(es).
 4. Then we can add method `public native long compose(long one, long other)` or `public native long compose(long[] handles)` in the JNI API `ColumnarBatchJniWrapper.java`, to replace `addColumn(...)` from this patch

Would you think the above can be possible?

By the way if you don't have time to do this you can open an issue and leave some comments in the code so that I can take the work some time later.

jinchengchenghh: This is a middle stage, so we will have other batch to convert to VeloxColumnarBatch, after I remove all the Arrow conversion, we will just have VeloxColumnBatch, no ArrowBatch.
`gluten::CompositeColumnarBatch` is a good idea, I will open an issue , thanks!

zhztheplayer: > after I remove all the Arrow conversion, we will just have VeloxColumnBatch, no ArrowColumnarBatch.

Sometimes I think we can keep some minimum codes around `ArrowColumnarBatch`. With simply exporting to Spark built-in `ArrowColumnVector` we can easily adopt all Spark SQL's functionalities for fall-back processing, including the **Python stuffs**.

It would also benefit if we want to integrate with other Arrow-compatible libraries like datafusion.

cc @FelixYBW Any suggestions?

jinchengchenghh: https://github.com/oap-project/gluten/issues/1419

zhztheplayer: Can you add a single TODO comment line in code? Something like
 
```
// TODO https://github.com/oap-project/gluten/issues/1419
```

Thanks a lot.

PR #1378, opened 2023-04-17
[VL] Upgrade velox to 2023/4/20

zhejiangxiaomai:
Velox main change list:
* Refactor UnsafeRowDynamicSerializer.h/UnsafeRowSerializer.h. like remove deserializeComplex function and UnsafeRowDynamicVectorBatchDeserializer class in https://github.com/facebookincubator/velox/pull/4612 and https://github.com/facebookincubator/velox/pull/4610.
* Refactor memory pool related api.  replace `addChild` with `addAggregateChild` and `addLeafChild` in https://github.com/facebookincubator/velox/pull/4629.
* change a little for kPreferredOutputBatchSize config in  https://github.com/facebookincubator/velox/pull/4439.

## How was this patch tested?

### DIFF
--- a/cpp/velox/memory/VeloxMemoryPool.cc
+++ b/cpp/velox/memory/VeloxMemoryPool.cc
@@ -31,6 +31,22 @@ namespace {
       /* isRetriable */ true,                                       \
       "Exceeded memory manager cap of {} MB",                       \
       (cap) / 1024 / 1024);
+
+std::shared_ptr<velox::memory::MemoryUsageTracker> createMemoryUsageTracker(
+    velox::memory::MemoryPool* parent,
+    velox::memory::MemoryPool::Kind kind,
+    const velox::memory::MemoryPool::Options& options) {
+  if (parent == nullptr) {
+    return options.trackUsage ? velox::memory::MemoryUsageTracker::create(options.capacity, options.checkUsageLeak)
+                              : nullptr;
+  }
+  if (parent->getMemoryUsageTracker() == nullptr) {
+    return nullptr;
+  }
+  const bool isLeaf = kind == velox::memory::MemoryPool::Kind::kLeaf;
+  VELOX_CHECK(isLeaf || options.threadSafe);
+  return parent->getMemoryUsageTracker()->addChild(isLeaf, options.threadSafe);
+}
 } // namespace
 
 //  The code is originated from /velox/common/memory/Memory.h
@@ -50,9 +66,7 @@ class WrappedVeloxMemoryPool final : public velox::memory::MemoryPool {
       DestructionCallback destructionCb,
       const Options& options = Options{})
       : velox::memory::MemoryPool{name, kind, parent, options},
-        memoryUsageTracker_(
-            parent_ == nullptr ? velox::memory::MemoryUsageTracker::create(options.capacity)
-                               : parent->getMemoryUsageTracker()->addChild(kind == MemoryPool::Kind::kLeaf)),
+        memoryUsageTracker_(createMemoryUsageTracker(parent_.get(), kind, options)),
         memoryManager_{memoryManager},
         velox_alloc_{&memoryManager_->getAllocator()},
         gluten_alloc_{gluten_alloc},
@@ -299,8 +313,12 @@ class WrappedVeloxMemoryPool final : public velox::memory::MemoryPool {
     return alignment_;
   }
 
-  std::shared_ptr<velox::memory::MemoryPool>
-  genChild(std::shared_ptr<velox::memory::MemoryPool> parent, const std::string& name, Kind kind) override {
+  std::shared_ptr<MemoryPool> genChild(
+      std::shared_ptr<MemoryPool> parent,
+      const std::string& name,
+      MemoryPool::Kind kind,
+      bool /*unused*/,
+      std::shared_ptr<facebook::velox::memory::MemoryReclaimer> /*unused*/) override {
     return std::make_shared<WrappedVeloxMemoryPool>(
         memoryManager_, name, kind, parent, gluten_alloc_, nullptr, Options{.alignment = alignment_});
   }
@@ -363,6 +381,18 @@ class WrappedVeloxMemoryPool final : public velox::memory::MemoryPool {
         velox::memory::MemoryAllocator::kindString(velox_alloc_->kind()));
   }
 
+  uint64_t freeBytes() const override {
+    VELOX_NYI("{} unsupported", __FUNCTION__);
+  }
+
+  uint64_t shrink(uint64_t targetBytes = 0) override {
+    VELOX_NYI("{} unsupported", __FUNCTION__);
+  }
+
+  uint64_t grow(uint64_t bytes) override {
+    VELOX_NYI("{} unsupported", __FUNCTION__);
+  }
+
  private:
   VELOX_FRIEND_TEST(MemoryPoolTest, Ctor);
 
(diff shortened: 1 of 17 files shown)


FelixYBW:
@zhejiangxiaomai  Can you add a shot list what's changed in the rebase? Let's add this for all rebase in future

cpp/velox/memory/VeloxMemoryPool.cc
```diff
@@ -406,16 +436,20 @@ std::shared_ptr<velox::memory::MemoryPool> AsWrappedVeloxMemoryPool(
       options);
 }
 
-std::shared_ptr<velox::memory::MemoryPool> GetDefaultWrappedVeloxMemoryPool() {
+std::shared_ptr<WrappedVeloxMemoryPool> GetDefaultWrappedVeloxMemoryPool() {
```
jinchengchenghh, line 439: `WrappedVeloxMemoryPool` is the child of `WrappedVeloxMemoryPool`, we do not need to change return value type

zhejiangxiaomai: Thanks, will fix it.

zhejiangxiaomai:
> @zhejiangxiaomai Can you add a shot list what's changed in the rebase? Let's add this for all rebase in future

Thanks, I will add it later in this PR.

cpp/velox/tests/ColumnarToRowTest.cc
```diff
@@ -83,7 +83,7 @@ class ColumnarToRowTest : public ::testing::Test {
     return rb;
   }
 
-  std::shared_ptr<facebook::velox::memory::MemoryPool> veloxPool_ = GetDefaultWrappedVeloxMemoryPool();
+  std::shared_ptr<facebook::velox::memory::MemoryPool> veloxPool_ = GetDefaultLeafWrappedVeloxMemoryPool();
```
jinchengchenghh, line 86: Can you use addLeafChild to track if which test hash memory leak?

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -714,7 +714,7 @@ arrow::Status VeloxShuffleWriter::SplitListArray(const velox::RowVector& rv) {
 
     // TODO: rethink the cost of `exportToArrow+ImportArray`
     ArrowArray arrowArray;
-    velox::exportToArrow(column, arrowArray, GetDefaultWrappedVeloxMemoryPool().get());
+    velox::exportToArrow(column, arrowArray, GetDefaultLeafWrappedVeloxMemoryPool().get());
```
jinchengchenghh, line 717: Add the issue https://github.com/oap-project/gluten/issues/1194 to here, the memory is not tracked

jinchengchenghh (dismissed):

rui-mo (commented):
@FelixYBW Do we want to still control the batch size totally with Spark config? Looks with recent change in Velox, an estimation of row size, the preferred batch bytes and also the config of preferred rows from Spark co-work to decide the read batch size.
```
  uint64_t preferredOutputBatchBytes() const {
    static constexpr uint64_t kDefault = 10UL << 20;
    return get<uint64_t>(kPreferredOutputBatchBytes, kDefault);
  }

  uint32_t preferredOutputBatchRows() const {
    return get<uint32_t>(kPreferredOutputBatchRows, 1024);
  }

  uint32_t maxOutputBatchRows() const {
    return get<uint32_t>(kMaxOutputBatchRows, 10'000);
  }
```

cpp/velox/memory/VeloxMemoryPool.h
```diff
@@ -21,11 +21,14 @@
...
+class WrappedVeloxMemoryPool;
 
 std::shared_ptr<facebook::velox::memory::MemoryPool> AsWrappedVeloxMemoryPool(
     MemoryAllocator* allocator,
     const facebook::velox::memory::MemoryPool::Options& options);
 
-std::shared_ptr<facebook::velox::memory::MemoryPool> GetDefaultWrappedVeloxMemoryPool();
+std::shared_ptr<WrappedVeloxMemoryPool> GetDefaultWrappedVeloxMemoryPool();
+
+std::shared_ptr<facebook::velox::memory::MemoryPool> GetDefaultLeafWrappedVeloxMemoryPool();
```
zhztheplayer, line 32: I don't yet have much context but would you like to explain a bit more about this change? Why do we have to add this function?

Is this because "aggregate pool" is no longer accepted by some APIs in Velox?

zhztheplayer: If we much have two default pools, can we rename the functions using the same convention,

e.g.

```
std::shared_ptr<facebook::velox::memory::MemoryPool> GetDefaultWrappedVeloxAggregateMemoryPool();
std::shared_ptr<facebook::velox::memory::MemoryPool> GetDefaultWrappedVeloxLeafMemoryPool();
```

I put "Aggregate"/"Leaf" towards right since the attribute "Velox" is more important than "Aggregate"/"Leaf" for a "MemoryPool".

cpp/velox/tests/ColumnarToRowTest.cc
zhejiangxiaomai: Sorry, I do not understand here. because I think we already call `addLeafChild` in `GetDefaultLeafWrappedVeloxMemoryPool` [here](https://github.com/oap-project/gluten/pull/1378/files#diff-163aa1d4bcffd7b4dd3b270eab81f55748d7e44fe185cb9a8c736f8f20b2d3ccR452).

cpp/velox/memory/VeloxMemoryPool.cc
```diff
@@ -31,6 +31,22 @@ namespace {
       /* isRetriable */ true,                                       \
       "Exceeded memory manager cap of {} MB",                       \
       (cap) / 1024 / 1024);
+
+std::shared_ptr<velox::memory::MemoryUsageTracker> createMemoryUsageTracker(
+    velox::memory::MemoryPool* parent,
+    velox::memory::MemoryPool::Kind kind,
+    const velox::memory::MemoryPool::Options& options) {
+  if (parent == nullptr) {
+    return options.trackUsage ? velox::memory::MemoryUsageTracker::create(options.capacity, options.checkUsageLeak)
```
zhztheplayer, line 40: In Velox, is `options.trackUsage` true by default?

cpp/velox/memory/VeloxMemoryPool.h
zhejiangxiaomai: Yes, you are right.  Here is some context for you. In previous velox version. `GetDefaultWrappedVeloxMemoryPool` will create a memory pool which can allocate memory and add child. But now only aggregate pool can add child. So I think we need create a new function to create a wrapped leaf pool for allocate memory.

GetDefaultWrappedVeloxMemoryPool -> aggregate pool
GetDefaultLeafWrappedVeloxMemoryPool  -> leaf pool

cpp/velox/memory/VeloxMemoryPool.cc
zhejiangxiaomai: Yes, here is [link]( https://github.com/zhejiangxiaomai/velox/blob/main_rebase_417/velox/common/memory/MemoryPool.h#L122)

cpp/velox/memory/VeloxMemoryPool.h
zhejiangxiaomai: > std::shared_ptr<facebook::velox::memory::MemoryPool> GetDefaultWrappedVeloxAggregateMemoryPool();
> std::shared_ptr<facebook::velox::memory::MemoryPool> GetDefaultWrappedVeloxLeafMemoryPool();

I will rename it.

rui-mo (approved):

Issue #1386, opened 2023-04-18
Improve max(NULL)/min(NULL) execution efficiency


KevinyhZou:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
When try to exeucte a query, like select max(NULL) min(NULL) , we found that it will not use native engine, which may be inefficiency. And the execute physical plan show below
![1](https://user-images.githubusercontent.com/37431499/232659843-f22c5f4d-6aeb-4b09-9d9d-3d318bea3c42.png)

In this plan we find it will not use native engine to execute aggregate query, and contains c2r / r2c transformation 2 times, which will lead to cost more times.

**Describe the solution you'd like**
Change the query plan, make it execute by native engine, and descrease the c2r / r2c transformation,  the query plan should like

![2](https://user-images.githubusercontent.com/37431499/232660309-58d52adf-f8b9-40d8-b75b-fe55ad82bfb0.png)

**Describe alternatives you've considered**

**Additional context**

PR #1387, opened 2023-04-18
[GLUTEN-1386][CH] Improve max(NULL)/min(NULL) execution efficiency

KevinyhZou:
Improve max(NULL)/min(NULL) execution efficiency: The max(NULL)/min(NULL) aggreate function's return type is NullType, which make the transform support validation to be false, and lead this query can not be exeuecte on the native engine.

So we make the change, let the transform support validation be supported on NullType

(Fixes: \#1386)

## How was this patch tested?
manual tests

This is tested on clickhouse-backend.  The tested query ` select id, max(NULL), min(NULL) from test_tbl3 where id > 0  group by id limit 3` , the tested table has about 63000000 rows, and stored as parquet, which archived 60% improvement.


zhztheplayer:
Should we use the same naming convention e.g. `[CH]` on the lhs of the PR title? I see there are some `[Clickhouse-backend]`s and some `[CH]`s currently.

KevinyhZou:
> Should we use the same naming convention e.g. `[CH]` on the lhs of the PR title? I see there are some `[Clickhouse-backend]`s and some `[CH]`s currently.

done

zzcclp:
please add the test cases.

KevinyhZou:
The test case add done，and merge pr : https://github.com/oap-project/gluten/pull/1383 into this,  as it will cause core when query max(NULL) / min(NULL) . Please have a review at this pr. @zzcclp

zzcclp (approved):
LGTM

liuneng1994 (approved):
LGTM

PR #1399, opened 2023-04-18
[VL] Short decimal should memcpy int64 data to arrow int128

liujiayi771:

https://github.com/oap-project/gluten/pull/1370 fixed the core dump error for decimal data. However, we found that the correctness of shuffle write is wrong. We cannot directly use SplitFixedType<uint64_t> method to handle short decimal. In velox short decimal is stored in int64, arrow always use int128 to store decimal, so we should copy int64 data to int128 dst.

## How was this patch tested?
Run 1T tpcds, and compare results.

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -514,7 +514,7 @@ arrow::Status VeloxShuffleWriter::SplitFixedWidthValueBuffer(const velox::RowVec
             auto end = partition_2_row_offset_[pid + 1];
             for (; pos < end; ++pos) {
               auto row_id = row_offset_2_row_id_[pos];
-              const uint64_t value = reinterpret_cast<const uint64_t *>(src_addr)[row_id]; // copy
+              const uint64_t value = reinterpret_cast<const uint64_t*>(src_addr)[row_id]; // copy
               memcpy(dst_pid_base, &value, sizeof(uint64_t));
               dst_pid_base += 1;
             }
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -502,7 +502,23 @@ arrow::Status VeloxShuffleWriter::SplitFixedWidthValueBuffer(const velox::RowVec
 #elif defined(__aarch64__)
       case 128:
         if (column->type()->isShortDecimal()) {
-          RETURN_NOT_OK(SplitFixedType<uint64_t>(src_addr, dst_addrs));
+          std::transform(
+              dst_addrs.begin(),
+              dst_addrs.end(),
+              partition_buffer_idx_base_.begin(),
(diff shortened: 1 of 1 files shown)


liujiayi771:
@FelixYBW From the perspective of the open source community, ARM adaptation is also a very important part of gluten. Is it possible to add a simple CI test to arm platform? Similar to the test of CentOS8. We can try to add an ARM CI test, use github resources or provide an arm machine to run the test.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -502,7 +502,23 @@ arrow::Status VeloxShuffleWriter::SplitFixedWidthValueBuffer(const velox::RowVec
...
+            auto dst_pid_base = reinterpret_cast<uint32x4_t*>(partition_buffer_idx_offset_[pid]);
+            auto pos = partition_2_row_offset_[pid];
+            auto end = partition_2_row_offset_[pid + 1];
+            for (; pos < end; ++pos) {
+              auto row_id = row_offset_2_row_id_[pos];
+              const uint64_t value = reinterpret_cast<const uint64_t*>(src_addr)[row_id]; // copy
+              memcpy(dst_pid_base, &value, sizeof(uint64_t));
+              dst_pid_base += 1;
+            }
+          }
```
FelixYBW, line 521: How velox handles the x86 vs. arm code? I'm thinking if it's a good way to add if defined in codes

FelixYBW:
> @FelixYBW From the perspective of the open source community, ARM adaptation is also a very important part of gluten. Is it possible to add a simple CI test to ARM platform? Similar to the test of CentOS8. We can try to add an ARM CI test, use github resources or provide an ARM machine to run the test.

We used a powerful machine to do the CI tests. The github resources definitely aren't fast enough. The other big problem is most developers don't have a ARM development, once there is any bug we can hardly debug them, like this PR. What I'm suggested to AWS guys are that they run the CI tests from their side every night, once there is any issues, they should fix them on ARM by themselves. We are OK to merge the PRs

cpp/velox/shuffle/VeloxShuffleWriter.cc
liujiayi771: Currently, there is also if defined in the velox code to handle aarch64, but velox uses xsimd, which shields many differences between different platforms. In terms of int128, velox does not choose the vectorization type according to the platform. If shuffle write wants to use the vectorized type, it should be differentiated by platform, but maybe we can create a lower-level method to deal with the difference of int128.

FelixYBW (approved):

FelixYBW:
@zzcclp Looks CK CI doesn't work again. 
Also the PR should skip the CK backend CI test. Looks our filter failed

FelixYBW:
@liujiayi771 Ready to merge?

liujiayi771:
> @liujiayi771 Ready to merge?

Yes, there are no other modifications, the test is correct.

PR #1403, opened 2023-04-18
[GLUTEN-1402][ClickHouse-Backend] Fix CHHashAggregate output is error when child is instanceOf InputAdapter.

loneylee:

If `CHHashAggregateExec` has child is instanceOf InputAdapter. Will use wrong iuput and output


zzcclp:
> Why close this pr, is there any other issues?

I will fix this issue in an other pr, which includes some other similar issue.

loneylee:
This fix solution will takes an other problem in tpcds 14 with error 'Invalid number of columns in chunk pushed to OutputPort'

Issue #1405, opened 2023-04-19
[CH] Clickhouse continue to execute when JNI call throw exception


shuai-xu:
**Describe the bug**
Now after a JNI call, CH check the exception, and only throw it in JNIEnv, so native code can not get the exception, and will continue to execute, this may cause other problems.

**Expected behavior**
CH can get the exception, stop executing and return.

PR #1406, opened 2023-04-19
[GLUTEN-1405] [CH] fix CH continue executing when JNI call throw exception

shuai-xu:

This pr fix the problem that if don't throw CH exception, CH will continue to execute after JNI call failed. So it should throw a Clickhouse Exception with the JNI exception message so the root cause canbe shown in spark.

(Fixes: \#1405)

## How was this patch tested?

This patch was tested manually


liuneng1994 (approved):
LGTM

Issue #1410, opened 2023-04-19
[CH] can not recognize the partition when path contains capital letter


shuai-xu:
**Describe the bug**
When using gluten to read a table whose hdfs path contains capital letter in parititon, it will fail with the exception 'No such name in Block::erase(): 'DAY''. The hdfs path is like ***/DAY=2023-04-13/.

PR #1411, opened 2023-04-19
[GLUTEN-1410] [CH] convert partition name to lower case

shuai-xu:

This pr converts partition names get from hdfs path to lower case, as spark paritiions are in lower case.

(Fixes: \#1410)

## How was this patch tested?


liuneng1994 (approved):
LGTM

Issue #1441, opened 2023-04-21
[CH] parquet and orc reader use the wrong settings for insensitive case


shuai-xu:
**Describe the bug**
ORC and Parquet reader in gluten use the FormatSettings.use_lowercase_column_name to judge whether convert the column names to lower case, but this flag is not set and not used in clickhouse.

**Expected behavior**
Should use FormatSettings.parquet.case_insensitive_column_matching and FormatSettings.orc.case_insensitive_column_matching instead.

taiyang-li:
It is fixed now.

taiyang-li:
fixed by : 
- https://github.com/oap-project/gluten/pull/1390
- https://github.com/oap-project/gluten/pull/1578
- https://github.com/oap-project/gluten/pull/1991

PR #1390, opened 2023-04-18
[GLUTEN-1389][CH] Fix ORC/Parquet column name case-insensitive matching issue

taiyang-li:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#1389)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #1444, opened 2023-04-21
[GLUTEN-1441] [CH] fix using wrong settings for case insensitivity in orc and parquet reader

shuai-xu:

This pr fix the bug that parquet and orc read in gluten use an unset settings for case  insensitivity.

(Fixes: \#1441)

## How was this patch tested?

### DIFF
--- a/cpp-ch/local-engine/Storages/ArrowParquetBlockInputFormat.cpp
+++ b/cpp-ch/local-engine/Storages/ArrowParquetBlockInputFormat.cpp
@@ -75,7 +75,7 @@ DB::Chunk ArrowParquetBlockInputFormat::generate()
     if (*batch)
     {
         auto tmp_table = arrow::Table::FromRecordBatches({*batch});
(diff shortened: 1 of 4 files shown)


taiyang-li:
duplicated with previous pr: https://github.com/oap-project/gluten/pull/1444

lwz9103:
Run Gluten Clickhouse CI

shuai-xu:
> duplicated with previous pr: #1390

This problem still exists in 1390 when enable local parquet/orc reader.

liuneng1994:
the parameter format_settings.use_lowercase_column_name The source code has been modified before, whether it is no longer needed @taiyang-li @shuai-xu

shuai-xu:
> the parameter format_settings.use_lowercase_column_name The source code has been modified before, whether it is no longer needed @taiyang-li @shuai-xu

image
I think it's needed as in ArrowParquetBlockInputFormat it still uses this setting which will never be initialized. @liuneng1994

PR #1463, opened 2023-04-23
[GLUTEN-1500][VL] feat: Use 0.6 * task memory cap as spill threshold for all spillable operators

zhztheplayer:


zhztheplayer:
Dependency https://github.com/oap-project/velox/pull/210

zhztheplayer:
@marin-ma @zhouyuan

docs/Velox.md
```diff
@@ -336,26 +336,51 @@ At runtime, `MEMKIND_HBW_NODES` enviroment variable is detected for configuring
...
+
+Following are some tips to tune against spilling-related configuration options:
+
+* `spark.gluten.sql.columnar.backend.velox.memoryCapRatio`
+
+  This is to set the memory cap limit (represented as MEMORY_LIMIT) to each Velox task. The formula is: MEMORY_LIMIT = (offHeap.size / task per executor) * memoryCapRatio. OOM will be raised if the memory usage exceeds this limit.
+
+* `spark.gluten.sql.columnar.backend.velox.spillMemoryThresholdRatio`
+
+  This is to set the spill limit (represented as SPILL_LIMIT) to each operator in each Velox task. The formula is: SPILL_LIMIT = MEMORY_LIMIT * spillMemoryThresholdRatio. Once a single Velox operator has allocated memory larger than this size, a spill request will be sent to the operator to force spilling-to-disk procedure before it continues processing.
```
zhouyuan, line 373: so this param is to do spill() before reaching  `spark.gluten.sql.columnar.backend.velox.memoryCapRatio`, does this mean `spark.gluten.sql.columnar.backend.velox.spillMemoryThresholdRatio` has higher priority?

docs/Velox.md
```diff
@@ -336,26 +336,51 @@ At runtime, `MEMKIND_HBW_NODES` enviroment variable is detected for configuring
...
+| spark.gluten.sql.columnar.backend.velox.spillStartPartitionBit          | 29                                                                                  | The start partition bit which is used with 'spillPartitionBits' together to calculate the spilling partition number                                                     |
+| spark.gluten.sql.columnar.backend.velox.spillPartitionBits              | 2                                                                                   | The number of bits used to calculate the spilling partition number. The number of spilling partitions will be power of two                                              |
+| spark.gluten.sql.columnar.backend.velox.spillableReservationGrowthPct   | 25                                                                                  | The spillable memory reservation growth percentage of the previous memory reservation size                                                                              |
+
+## Guidance to tune against spilling
+
```
zhouyuan, line 362: thanks for adding this note!

docs/Velox.md
zhouyuan, line 373: is there a formular on "how many bytes to spill"?

zhztheplayer: Exactly.

zhztheplayer: I think usually `spark.gluten.sql.columnar.backend.velox.spillableReservationGrowthPct` which by default is 25% is used to do the calculation. 

But the option value is also used to grow memory pool reservation. I don't quite get the intention of using the same option to do these two things yet so I didn't document that here. Maybe we can just let the logic stay in box.

Issue #1466, opened 2023-04-23
[CH] columns may be different in type in union 


shuai-xu:
**Describe the bug**
In a sql contains union, the two columns with same name may be differnt in type, one is nullalbe while the other not. It will throw exception when merge them.

**Expected behavior**
I think there are two possiable way to fix it. 1: judge the column type when merging in shuffle splitter. 2: check and uniform the column types when override the plan.

taiyang-li:
fixed by https://github.com/oap-project/gluten/pull/1901

PR #1467, opened 2023-04-23
[GLUTEN-1466] [CH] support adding two columns with differnt nullable type for union

shuai-xu:

This pr supports adding two columns with differnt nullable type for union.

(Fixes: \#1466)

## How was this patch tested?


liuneng1994:
please add some unit tests

Issue #1416, opened 2023-04-19
The shaded arrow class should not be passed to Spark


liujiayi771:
**Describe the bug**
Some code in ArrowWritableColumnVector.java pass a shaded arrow class to Spark method, this will result in a no such method exception.
https://github.com/oap-project/gluten/blob/e6def13f50d86353e7e059b1bb881bf8c5ef66b5/gluten-data/src/main/java/io/glutenproject/vectorized/ArrowWritableColumnVector.java#L1159

```
java.lang.RuntimeException: Error during calling Java code from native code: java.lang.NoSuchMethodError: org.apache.spark.sql.vectorized.ArrowColumnVector.<init>(Lio/glutenproject/shaded/org/apache/arrow/vector/ValueVector;)V
	at io.glutenproject.vectorized.ArrowWritableColumnVector$ArrayAccessor.<init>(ArrowWritableColumnVector.java:1159)
	at io.glutenproject.vectorized.ArrowWritableColumnVector.createVectorAccessor(ArrowWritableColumnVector.java:258)
	at io.glutenproject.vectorized.ArrowWritableColumnVector.<init>(ArrowWritableColumnVector.java:191)
	at io.glutenproject.vectorized.ArrowWritableColumnVector.reserveNewColumn(ArrowWritableColumnVector.java:350)
	at io.glutenproject.vectorized.ArrowWritableColumnVector.reserveNewColumn(ArrowWritableColumnVector.java:77)
	at org.apache.spark.sql.execution.vectorized.WritableColumnVector.<init>(WritableColumnVector.java:838)
	at org.apache.spark.sql.execution.vectorized.WritableColumnVectorShim.<init>(WritableColumnVectorShim.java:39)
	at io.glutenproject.vectorized.ArrowWritableColumnVector.<init>(ArrowWritableColumnVector.java:160)
	at io.glutenproject.vectorized.ArrowWritableColumnVector.<init>(ArrowWritableColumnVector.java:155)
	at io.glutenproject.vectorized.ArrowWritableColumnVector.allocateColumns(ArrowWritableColumnVector.java:110)
	at io.glutenproject.execution.GlutenRowToArrowColumnarExec$$anon$1.javaConvert(GlutenRowToArrowColumnarExec.scala:399)
	at io.glutenproject.execution.GlutenRowToArrowColumnarExec$$anon$1.next(GlutenRowToArrowColumnarExec.scala:423)
[...]

PR #1487, opened 2023-04-24
[GLUTEN-1416][VL] Fix no such method exception of shaded arrow

kerwin-zk:
Fix no such method exception of shaded arrow
#1416

### DIFF
--- a/gluten-data/src/main/java/io/glutenproject/vectorized/ArrowColumnVector.java
+++ b/gluten-data/src/main/java/io/glutenproject/vectorized/ArrowColumnVector.java
@@ -1,0 +1,577 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


gluten-data/src/main/java/io/glutenproject/vectorized/ArrowColumnVector.java
```diff
@@ -0,0 +1,577 @@
...
+import org.apache.spark.sql.vectorized.ColumnVector;
+import org.apache.spark.sql.vectorized.ColumnarArray;
+import org.apache.spark.sql.vectorized.ColumnarMap;
+import org.apache.spark.unsafe.types.UTF8String;
+
+/**
+ * A column vector backed by Apache Arrow.
+ */
+@DeveloperApi
+public class ArrowColumnVector extends ColumnVector {
```
zhztheplayer, line 38: Would you like to add some notes here to let developer know from which Version of Spark this file is originated from?

zhztheplayer (commented):
Thanks @kerwin-zk !

gluten-data/src/main/java/io/glutenproject/vectorized/ArrowColumnVector.java
kerwin-zk: @zhztheplayer done

zhztheplayer (approved):

PR #1497, opened 2023-04-24
[GLUTEN-384][CH] support reading from S3 and local cache on S3

binmahone:

checkout https://github.com/Kyligence/ClickHouse/issues/384

## How was this patch tested?

this Patch will later be test covered by our integration test


liuneng1994 (approved):
LGTM

Issue #1199, opened 2023-03-24
[CORE] Handle the exception when task is interrupted.


jackylee-ch:
**Describe the bug**
Currentlly, we meet some problems, such as null pointer or memory leak, when spark speculation is enabled. It seems that we need recheck the code and handle the Exception, such as `ClosedByInterruptException`, and then return it to Java side, rather than executor core dump.

**To Reproduce**
Run TPCDS with `spark.speculation=true`

**Expected behavior**
Only Exception is thrown.

jinchengchenghh:
Can we close this issue? @jackylee-ch

jackylee-ch:
> Can we close this issue? @jackylee-ch

Sorry for late response, we are still working on this.

PR #1187, opened 2023-03-22
[CORE] Draft: Prevent executor core dump while tring to kill speculative tasks

jackylee-ch:
## What changes were proposed in this pull request?
When speculation is turned on, in some cases, the executor will core dump due to speculative task kill, which is very dangerous for those applications that do not have a shuffle service.

## How was this patch tested?
unit tests

PR #1505, opened 2023-04-25
[GLUTEN-1199] Avoid throwing exception from destructor of JavaInputStreamAdaptor

cambyzju:

`HINT: Throw an exception in destructor is dangerous and you should never let the exception leave destructors. If there are two exceptions propagating, the program will terminate or yield undefined behavior.`

inside function Java_io_glutenproject_vectorized_ShuffleReaderJniWrapper_make
1. we new object of JavaInputStreamAdaptor
2. we new object of Reader, and constructor of Reader throw an exception
3. inside JNI_METHOD_END, we will catch this exception
But between step2 and step3, destructor of JavaInputStreamAdaptor is called, then we got another exception, the process terminate.

(Fixes: #1199)

## How was this patch tested?

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -114,10 +114,10 @@ class JavaInputStreamAdaptor final : public arrow::io::InputStream {
         std::cout << __func__ << " call JavaInputStreamAdaptor::Close() failed, status:" << status.ToString()
                   << std::endl;
 #endif
-        }
+      }
     } catch (std::exception& e) {
 #ifdef GLUTEN_PRINT_DEBUG
-        std::cout << __func__ << " call JavaInputStreamAdaptor::Close() got exception:" << e.what() << std::endl;
+      std::cout << __func__ << " call JavaInputStreamAdaptor::Close() got exception:" << e.what() << std::endl;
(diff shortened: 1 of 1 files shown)


jackylee-ch:
cc @zhztheplayer @jinchengchenghh @zhouyuan @PHILO-HE

FelixYBW:
Thank you for catching this!

jinchengchenghh (dismissed):
Please fix the code style, thanks!

zhztheplayer (approved):
Thanks. 

And how does this relate to the mentioned memory leak issue? Would you like to explain a bit more about that?

cambyzju:
> Please fix the code style, thanks!

done

cambyzju:
> Thanks.
> 
> And how does this relate to the mentioned memory leak issue? Would you like to explain a bit more about that?

This pr do not relate to memory leak issue. I will try to found it out and fix it next week.

jackylee-ch:
@zhztheplayer This pr is related to #1187   and it is a better fix to deal with `checkException`.

PHILO-HE:
Will merge the patch soon, if there is no more comment.

PHILO-HE (approved):

Issue #1537, opened 2023-04-28
show source info in node name


shuai-xu:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
In spark, the ui clearly show the source table as bellow:
![image](https://user-images.githubusercontent.com/20960838/235051018-5202019a-c401-4f48-a4b3-39d66d7391eb.png)

In gluten, if source is execute in native engine, it will be put in the WholeStageCodegenTransformer, and in Spark UI, users can only see a WholeStageCodegenTransformer without the source info, this is inconvinent to debug.
![image](https://user-images.githubusercontent.com/20960838/235051084-09bdce28-ec3f-4dc9-bfa7-e9cef5485917.png)

**Describe the solution you'd like**
Add the source name to nodename, so users can get it easily as below.
![image](https://user-images.githubusercontent.com/20960838/235051153-32217a42-f0db-4e77-9118-03641d3ab5e3.png)

Yohahaha:
How do you test? I see Spark UI is make sense in my cluster.

![image](https://user-images.githubusercontent.com/34979747/235089726-bf0117f8-bce7-46a9-8bab-7c612a0f83fe.png)

PR #1538, opened 2023-04-28
[GLUTEN-1537] add source info to nodename

shuai-xu:

This pr add source info to nodename so they can be shown in SparkUI.

(Fixes: \#1537)

## How was this patch tested?

manual tests

This patch involves UI changes:
image

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformerExec.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformerExec.scala
@@ -170,7 +170,15 @@ case class WholeStageTransformerExec(child: SparkPlan)(val transformStageId: Int
   // It's misleading with "Codegen" used. But we have to keep "WholeStageCodegen" prefixed to
   // make whole stage transformer clearly plotted in UI, like spark's whole stage codegen.
   // See buildSparkPlanGraphNode in SparkPlanGraph.scala of Spark.
-  override def nodeName: String = s"WholeStageCodegenTransformer ($transformStageId)"
+  override def nodeName: String = {
+    var fullName = s"WholeStageCodegenTransformer ($transformStageId)"
+    child.foreach(c => {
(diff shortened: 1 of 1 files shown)


gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformerExec.scala
```diff
@@ -170,7 +170,15 @@ case class WholeStageTransformerExec(child: SparkPlan)(val transformStageId: Int
   // It's misleading with "Codegen" used. But we have to keep "WholeStageCodegen" prefixed to
   // make whole stage transformer clearly plotted in UI, like spark's whole stage codegen.
   // See buildSparkPlanGraphNode in SparkPlanGraph.scala of Spark.
-  override def nodeName: String = s"WholeStageCodegenTransformer ($transformStageId)"
+  override def nodeName: String = {
+    var fullName = s"WholeStageCodegenTransformer ($transformStageId)"
+    child.foreach(c => {
+      if (c.isInstanceOf[FileSourceScanExecTransformer]) {
```
zzcclp, line 176: why it needs to add file source scan node?

shuai-xu: If a sql select from serval tables, it's hard to know which table does a stage read from by spark ui, as all stages have the same name WholeStageCodegenTransformer.

PHILO-HE: If it is strongly needed, can we control the use of verbose name via a config? I think It should be disabled by default to make the UI brief for most scenarios. Right?

Yohahaha:
I remember Gluten will show scan info like this

image

Is any things changes for UI?

FelixYBW:
What's the UI before and after the PR? What's the vanilla Spark's behavior?

If it's used for debug more easier, we may add an undocumented config for it.

PR #1541, opened 2023-04-28
Revert "[GLUTEN-1386][CH] Improve max(NULL)/min(NULL) execution efficiency"

zzcclp:
There is a `ArrayIndexOutOfBoundsException` in the new test case `test 'max(NULL)/min(NULL) from table'`

Reverts oap-project/gluten#1387


zzcclp:
Fix issue in #1345

PR #1542, opened 2023-04-28
[VL] Rename kSparkBatchSize

zhejiangxiaomai:

Previous we use 3 parameters to control the batch size, all of them reuse existing Spark config. But none of them exactly describe the columnar size we used.
  --conf spark.sql.parquet.columnarReaderBatchSize=${batchsize} \
  --conf spark.sql.inMemoryColumnarStorage.batchSize=${batchsize} \
  --conf spark.sql.execution.arrow.maxRecordsPerBatch=${batchsize} \

we also one spark.gluten.sql.columnar.shuffleSplitDefaultSize to control the split batch size.

Now we use spark.sql.execution.arrow.maxRecordsPerBatch to control the batch size in Gluten. It's passed to Velox as the max row number per row vector.

But the real batch size may still be bigger than the number because CoalesceBatch Operator after shuffle may combine two batches which is less than maxRecordsPerBatch  into one which is bigger.


jinchengchenghh:
Please also update the document

zhejiangxiaomai:
> Please also update the document

Thanks, just updated.

docs/Configuration.md
```diff
@@ -27,22 +27,23 @@ You can add these configuration into spark-defaults.conf to enable or disable th
...
-| spark.gluten.sql.columnar.shuffle.codecBackend | Enable using hardware accelerators for shuffle de/compression. Valid options are QAT and IAA.                                                                                                                                                                                                           |                                                      |
+| spark.sql.execution.arrow.maxRecordsPerBatch | Set up the Max Records per Batch                                                                                                                                                                                                                                                                              | 10000                                                |
```
FelixYBW, line 38: It's not used in Gluten anymore, right? if so let's remove it from readme

FelixYBW (approved):

FelixYBW (approved):

jinchengchenghh (approved):

PR #1547, opened 2023-04-28
[VL] Shuffle reader is not tracked by Spark memory manager

zhztheplayer:


gluten-core/src/main/scala/org/apache/spark/util/memory/TaskMemoryResources.scala
```diff
@@ -155,4 +160,29 @@ class TaskMemoryResourceRegistry extends Logging {
   private[memory] def getSharedMetrics(): TaskMemoryMetrics = {
     sharedMetrics
   }
+
+  private[memory] def addRecycler(recycler: TaskMemoryResourceRecycler) = {
```
FelixYBW, line 164: Why can't we use the addTaskCompletionListener?

zhztheplayer: `addTaskCompletionListener` doesn't implement priority for listeners. In our case we should make sure the allocators /memorypools to be released after all dependent buffers are released.

cpp/core/jni/JniWrapper.cc
```diff
@@ -913,15 +912,25 @@ JNIEXPORT void JNICALL Java_io_glutenproject_vectorized_OnHeapJniByteInputStream
...
+    jlong alloc_id) {
   JNI_METHOD_START
+  auto* allocator = reinterpret_cast<MemoryAllocator*>(alloc_id);
+  if (allocator == nullptr) {
+    gluten::JniThrow("Memory pool does not exist or has been closed");
+  }
+  auto pool = AsWrappedArrowMemoryPool(allocator);
   std::shared_ptr<arrow::io::InputStream> in = std::make_shared<JavaInputStreamAdaptor>(env, jni_in);
   ReaderOptions options = ReaderOptions::Defaults();
+  options.ipc_read_options.memory_pool = pool.get();
```
jinchengchenghh, line 929: Is this option pool duplicated with Reader constructor pool?

gluten-core/src/main/scala/org/apache/spark/util/memory/TaskMemoryResources.scala
```diff
@@ -84,12 +87,12 @@ object TaskMemoryResources extends Logging {
     }
   }
 
-  def addLeakSafeTaskCompletionListener[U](f: TaskContext => U): TaskContext = {
+  def addRecycler(priority: Long)(f: TaskContext => Unit): Unit = {
```
jinchengchenghh, line 90: Can we have a new function `addRecycler()` with default value 100?

gluten-data/src/main/scala/io/glutenproject/vectorized/GlutenColumnarBatchSerializer.scala
```diff
@@ -67,8 +65,16 @@ private class GlutenColumnarBatchSerializerInstance(schema: StructType,
...
-          JniByteInputStreams.create(in), cSchema.memoryAddress())
+          JniByteInputStreams.create(in), cSchema.memoryAddress(),
+          NativeMemoryAllocators.contextInstance.getNativeInstanceId)
         cSchema.close()
+        // Close shuffle reader instance as lately as the end of task processing,
+        // since the native reader could hold a reference to memory pool that
+        // was used to create all buffers read from shuffle reader. The pool
+        // should keep alive before all buffers finish consuming.
+        TaskMemoryResources.addRecycler(50) { _ =>
+          ShuffleReaderJniWrapper.close(handle)
```
jinchengchenghh, line 76: If we just close the `ShuffleReaderJniWrapper` at the end of the task, we do not release this instance as soon as possible.

The `arrow::schema` may exist as a very long lifecycle.

Maybe we could release other members except pool as soon as possible?

cpp/core/jni/JniWrapper.cc
zhztheplayer: No, this would be passed to the underlying IPC reader.

The one in constructor is to keep the pool alive.

gluten-core/src/main/scala/org/apache/spark/util/memory/TaskMemoryResources.scala
zhztheplayer: Yes we can but personally I don't suggest doing that because once one calls this API, he/she should better know exactly what priority means. If we hide this detail from users they may not be able to manage the resources appropriately.

gluten-data/src/main/scala/io/glutenproject/vectorized/GlutenColumnarBatchSerializer.scala
zhztheplayer: > The arrow::schema may exist as a very long lifecycle.

I think this doesn't really matter since the schema just holds a little of memory space.

> Maybe we could release other members except pool as soon as possible?

I was thinking about it but that requires for some arch-level changes. Let's just solve the issues with a minimal change then start a new topic about that.

zhztheplayer: I'll add some comments to the API in https://github.com/oap-project/gluten/pull/1527

gluten-core/src/main/scala/org/apache/spark/util/memory/TaskMemoryResources.scala
jinchengchenghh: For most case, the listener has same priority, it is just some special case that it has different priority. Only the user who has special need should ko=now the meaning of priority. What do you think? @FelixYBW

jinchengchenghh (approved):

gluten-core/src/main/scala/org/apache/spark/util/memory/TaskMemoryResources.scala
zhztheplayer: Since patch is somehow important for spilling, let's move the discussion into https://github.com/oap-project/gluten/pull/1527

PR #1568, opened 2023-05-06
[GLUTEN-1434][VL] Shuffle serialize and deserialize row vector

jinchengchenghh:


cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1139,25 +1139,80 @@ int64_t get_batch_nbytes(const arrow::RecordBatch& rb) {
...
+  }
+
+  std::shared_ptr<arrow::ResizableBuffer> value_buffer;
+  GLUTEN_ASSIGN_OR_THROW(
+      value_buffer, arrow::AllocateResizableBuffer(rowVector->estimateFlatSize() * 1.1, options_.memory_pool.get()));
+  auto output = std::make_shared<arrow::io::BufferOutputStream>(value_buffer);
+
+  auto arena = std::make_unique<StreamArena>(veloxPool_);
+  auto rowType = asRowType(rowVector->type());
+  auto serializer = serde_->createSerializer(rowType, numRows, arena.get(), serdeOptions);
```
winningsix, line 1170: Question: Any consideration of not using ```UnsafeRowVectorSerializer```? It also can append new row vector? 

Did we compare the micro-benchmark e.g., ```ShuffleSplitBenchmark``` to compare ```ArrowShuffleWriter``` and ```VeloxShuffleWriter```? Also any space efficiency when adopting different serialization?

cpp/velox/shuffle/VeloxShuffleWriter.h
```diff
@@ -243,6 +256,8 @@ class VeloxShuffleWriter final : public ShuffleWriter {
   // store arrow column types
   std::vector<std::shared_ptr<arrow::DataType>> arrow_column_types_; // column_type_id_
 
+  std::unique_ptr<facebook::velox::serializer::presto::PrestoVectorSerde> serde_;
```
winningsix, line 259: Question: is it b/c ```PrestoVectorSerde``` better perf than ```UnsafeRowVectorSerde``` that we choose ```PrestoVectorSerde```?

cpp/core/shuffle/ShuffleWriter.h
```diff
@@ -44,6 +44,10 @@ class ShuffleWriter {
 
   virtual arrow::Status Stop() = 0;
 
+  virtual std::shared_ptr<arrow::Schema>& Schema() {
```
winningsix, line 47: Minor hint: we can leave some comments here that this is for write schema and for ```VeloxShuffleWriter``` is difference from ```ArrowShuffleWriter``` since it's serialized as a Arrow array.

cpp/velox/shuffle/VeloxShuffleWriter.cc
jinchengchenghh: Both of these 2 serializer will copy the data one by one to several big buffers, and then copy the big buffers to OutputStream.
I will implement new SparkSerializer.
Meituan will help to split velox to velox, and new SparkSerializer will serialize the velox RowVector buffer, and write to one big buffer with only once memory copy.

cpp/core/shuffle/ShuffleWriter.h
jinchengchenghh: We will delete `ArrowShuffleWriter` after we completely implement `VeloxShuffleWriter` (velox to velox).

zhejiangxiaomai (approved):

gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
```diff
@@ -218,10 +222,35 @@ private static ColumnarBatch offload(BufferAllocator allocator, ColumnarBatch in
     }
   }
 
+  public static Iterator<InternalRow> emptyRowIterator(ColumnarBatch batch) {
```
zhztheplayer, line 225: @jinchengchenghh 

Should we use signature like

```
Iterator<InternalRow> emptyRowIterator(int maxRows)
```
? Which seems to be more explainable.

jinchengchenghh: Yes, next PR.

jinchengchenghh: It is only used in ColumnartoRow, if change input to maxRows, which file should I put it? Do you have some suggestions?

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -691,15 +806,15 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
...
 
-          binaryBuf.value_ptr = valueBuffer->mutable_data();
-          binaryBuf.value_capacity = capacity;
-          dstValuePtr = binaryBuf.value_ptr + valueOffset - stringLen;
+          binaryBuf.valuePtr = valueBuffer->mutable_data();
+          binaryBuf.valueCapacity = capacity;
+          dstValuePtr = binaryBuf.valuePtr + valueOffset - stringLen;
 
-          VsPrintSplit("Split value buffer resized colIdx", binaryIdx);
+          std::cout << "Split value buffer resized colIdx" << binaryIdx << std::endl;
```
liujiayi771, line 817: @rui-mo Should we remove this log?

rui-mo: Maybe change to only dump the log when debugging?

PR #1569, opened 2023-05-06
[GLUTEN-1551][VL][Fix] Trim leading/tailing whitespace in string for casting to integral type

PHILO-HE:

Fix Gluten-1551.

## How was this patch tested?

UT.


backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -43,4 +44,28 @@ class VeloxSparkPlanExecApi extends GlutenSparkPlanExecApi {
...
+   * To align with spark in casting string type input to integral type,
+   * add trim node for trimming whitespace. See spark's toInt in UTF8String.java.
+   */
+  override def genCastWithNewChild(c: Cast): Cast = {
+    // Whitespace to be trimmed, including: ' ', '\n', '\r', '\f', etc.
+    // The space separator (except ' ', i.e., '\u0020'), line separator &
+    // paragraph separator are not supported to trim, assuming they are rarely used.
+    val trimStr = " \t\n\u000B\u000C\r\u001C\u001D\u001E\u001F"
+    c.dataType match {
+      case ByteType | ShortType | IntegerType | LongType =>
```
jackylee-ch, line 58: After check the `Cast` in Spark, I found the `Character.isWritespace` is also used while casting `String` to `TimeStamp, Date, YearMonthInterval, DayTimeInterval, CalendarInterval, Boolean, TimestampNTZ`. As for casting from `String` to `Double, Float, Decimal`, Spark use `String.trim`.
May be this code is better?
```
c.dataType match {
  case _: ArrayType | _: MapType | _: StructType | _: UserDefinedType[_] =>
     c
  case FloatType | DoubleType | DecimalType =>
    c.child.dataType match {
      case StringType =>
        val trim = StringTrim(c.child, Some(Literal(trimSpaceStr)))
          c.withNewChildren(Seq(trim)).asInstanceOf[Cast]
       case _ =>
          c
      }
    }
  case _: =>
    c.child.dataType match {
      case StringType =>
        val trim = StringTrim(c.child, Some(Literal(trimStr)))
          c.withNewChildren(Seq(trim)).asInstanceOf[Cast]
        case _ =>
          c
      }
    }
```

PHILO-HE: Thanks so much for your check! Yes, we should also consider all other types even though the potential cast issues have not been exposed yet.

PHILO-HE: For BinaryType, we should keep it not trimmed.

jackylee-ch: Ah, yes

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/GlutenDataFrameSuite.scala
```diff
@@ -297,6 +298,17 @@ class GlutenDataFrameSuite extends DataFrameSuite with GlutenSQLTestsTrait {
...
 
+  test(GLUTEN_TEST +
+      "Allow leading/trailing whitespace in string for casting to integral type") {
+    Seq(" 123", "123 ", " 123 ", "123\n\n\n", "123\r\r\r", "123\f\f\f", "123\u000C")
+        .toDF("col1").createOrReplaceTempView("t1")
+    val df = spark.sql("select cast(col1 as int) from t1")
+    checkAnswer(df, Row(123) :: Row(123) ::
+        Row(123) :: Row(123) :: Row(123) :: Row(123) :: Row(123) :: Nil)
+    assert(find(df.queryExecution.executedPlan)(
+      _.isInstanceOf[ProjectExecTransformer]).isDefined)
```
jackylee-ch, line 309: nit: may be add some test for binary and double type?

FelixYBW (approved):

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/GlutenDataFrameSuite.scala
PHILO-HE: Yes, just added some more tests. Thanks!

PHILO-HE:
Run Gluten Clickhouse CI

FelixYBW (approved):

PR #1573, opened 2023-05-08
[VL] Fix some compilation errors with Clang 12

zhztheplayer:
This would not mean adding support for clang-12, just to solve some issues that clang-12 has complained about.


zhztheplayer:
One of the way to switch to clang-12 on ubuntu 20
```
apt-get install clang-12
update-alternatives --install /usr/bin/cc cc /usr/bin/clang-12 60
update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++-12 60
```

zhanglistar:
Why not use clang-15?

zhztheplayer:
> Why not use clang-15?

@zhanglistar ubuntu 20.04 provides up to llvm 12 by apt, that's why I am using 12 for this PR.

Besides the PR is just to fix some errors from clang-12, but not to officially support clang-12. I assume the fixes should also benefit when using a higher version although I don't intend to test against them yet.

zhztheplayer:
@zuochunwei @rui-mo

rui-mo (approved):

PR #1580, opened 2023-05-09
[VL] C++: Rename identifiers based on the CPP code style

zhztheplayer:
Requires for https://github.com/oap-project/gluten/pull/1573


cpp/.clang-tidy
```diff
@@ -0,0 +1,31 @@
+Checks: >
```
winningsix, line 1: Minor suggestion: We can enable this in Github action to do some code style pre-check.

zhztheplayer: Sure. I'll open another PR for that.

zhztheplayer: Issue: https://github.com/oap-project/gluten/issues/1599

PR #1588, opened 2023-05-09
[VL] Add shortcut tools/gluten-it/sbin/gluten-it.sh to run gluten-it

zhztheplayer:


FelixYBW:
Any impact to current build script? if there is, then can you highlight what's it before and after this PR?

zhztheplayer:
> Any impact to current build script? if there is, then can you highlight what's it before and after this PR?


It doesn't impact any part of current build. It is just to add a shortcut .sh file to shorten the command used in CI.

FelixYBW (approved):

PR #1595, opened 2023-05-09
[GLUTEN-1594][CH] support native parquet writer for clickhouse backend

binmahone:

a basic working version is provided by this PR. (so that we're unblocked to  perform experiments)

however, as there're many TODO notes in the PR:

image

so, this ability is still incomplete.

experiments show that native parquet writer is about 20%-25% better than previous (Gluten used to convert columnar to row first, and use vanilla Spark's ParquetOutputFormat for writing):

image

## How was this patch tested?

I'll write a simple test case to ensure basic write routine works


liuneng1994 (commented):
LGTM

backends-clickhouse/src/main/java/org/apache/spark/sql/execution/datasources/CHDatasourceJniWrapper.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.datasources;
+
+public class CHDatasourceJniWrapper {
```
zzcclp, line 20: maybe `CHDatasourceWriterWrapper` is better ?

zzcclp:
@zhztheplayer @rui-mo please help to review, thanks

zzcclp:
please add the ut for this new feature.

binmahone:
> please add the ut for this new feature.

i'm working on it

backends-clickhouse/src/main/java/org/apache/spark/sql/execution/datasources/CHDatasourceJniWrapper.java
binmahone: it's named after this. https://github.com/oap-project/gluten/pull/1344/files#diff-379c35236c5f4cfe33d76e4ed3c9fc90ee447794dfc25b5674b05402c767af3f . I guess this class is also planed for future APIs to support a complete native spark-datasouce

JkSelf (dismissed):
LGTM. And also verified in 500gb TPC-DS parquet write test in velox backend. And performance data unchanged.

winningsix:
Just curious what benchmark we used in our measurements?

binmahone:
> Just curious what benchmark we used in our measurements?

customer table in tpch 100.
More comprehensive experiments are still on the way, this is an early PR just to avoid conflict resolve efforts

gluten-data/src/main/scala/io/glutenproject/execution/VeloxColumnarRules.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  override def output: Seq[Attribute] = {
+    child.output
+  }
+
+  override protected def withNewChildInternal(newChild: SparkPlan): SparkPlan =
+    copy(child = newChild)
+}
+
+object VeloxColumnarRules {
```
rui-mo, line 57: Verified on internal Jenkins. Just a tiny question, can this file be put in `backends-velox` instead of `gluten-data` if it only for Velox backend?

binmahone: Hi morui, I confirmed with @zzcclp , we think:

1. currently all classes in gluten-data seems to be "for velox" . Classes "for velox+ch" are all located in gluten-core.
2. Due to 1, some classes named "GlutenXXX" in gluten-data is misleading
3. This is an unreasonable state, but we don't have the bandwidth to correct it right now

zhztheplayer: > some classes named "GlutenXXX" in gluten-data is misleading

The classes `GlutenXXX` were added for backends that rely on Gluten to manage data conversions between Java and C++. So we put this layer (gluten-data) as an individual module out from backends-velox. But indeed we don't have another backend to use this part of code yet.

binmahone:
Run Gluten Clickhouse CI

gluten-data/src/main/scala/io/glutenproject/execution/VeloxColumnarRules.scala
rui-mo: Got it. In that case, gluten-data is designed to be a common module, right? Anyway, per previous discussion with Binwei, I will do a clean up for classes named as `GlutenXXX` and `VeloxXXX` because the prefixes are not necessary.

zhztheplayer: > gluten-data is designed to be a common module, right?

Yes that was the original design.

binmahone: thanks @zhztheplayer for the explanation!

@rui-mo , If a class is already backend-agnostic, then I'll agree with you that the Gluten prefix is unnecessary.

However, if a class is velox-only , it's perferrable to keep the Velox prefix, so that we can easily distinguish velox-specific classes from the common classes. 

Bedi

rui-mo: > I will do a clean up for classes named as GlutenXXX and VeloxXXX because the prefixes are not necessary.

@binmahone Got your point. The motivation for above action is we assume Velox related code is put in `backends-velox` module, so the `Velox` prefix becomes unnecessary. If some velox-only code exists in a common module, that's a different case which shall be rechecked. I will take a look further, thanks for pointing this out!

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenColumnarRules.scala
```diff
@@ -101,67 +104,50 @@ private case class ColumnarToFakeRowLogicAdaptor(child: LogicalPlan)
...
-  case class LoadBeforeColumnarToRow() extends Rule[SparkPlan] {
-    override def apply(plan: SparkPlan): SparkPlan = plan.transformUp {
-      case c2r @ ColumnarToRowExec(_: ColumnarShuffleExchangeExec) =>
-        c2r // AdaptiveSparkPlanExec.scala:536
-      case c2r @ ColumnarToRowExec(_: ColumnarBroadcastExchangeExec) =>
-        c2r // AdaptiveSparkPlanExec.scala:546
-      case ColumnarToRowExec(child) => ColumnarToRowExec(LoadArrowData(child))
+  case class NativeWritePostRule(session: SparkSession) extends Rule[SparkPlan] {
+    override def apply(p: SparkPlan): SparkPlan = p match {
+      case rc @ DataWritingCommandExec(cmd, child) if isGlutenInsertInto(cmd) =>
```
JkSelf, line 148: If child is a row-based operator, adding a FakeRowAdaptor here will report an error, right? Our current code does not remove the ColumnarToRow in front of AQE.
image

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenColumnarRules.scala
```diff
@@ -82,86 +83,90 @@ class FakeRow(val batch: ColumnarBatch) extends InternalRow {
...
+        child match {
+          // if the child is columnar, we can just wrap&transfer the columnar data
+          case c2r: GlutenColumnarToRowExecBase =>
+            rc.withNewChildren(Array(FakeRowAdaptor(c2r.child)))
+          // If the child is aqe, we make aqe "support columnar",
+          // then aqe itself will guarantee to generate columnar outputs.
+          // So FakeRowAdaptor will always consumes columnar data,
+          // thus taking no risk to downgrade performance
+          case aqe: AdaptiveSparkPlanExec =>
```
JkSelf, line 156: How to make aqe suport columnar here?

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenColumnarRules.scala
```diff
@@ -82,86 +83,90 @@ class FakeRow(val batch: ColumnarBatch) extends InternalRow {
...
+              Array(
+                FakeRowAdaptor(
+                  AdaptiveSparkPlanExec(
+                    aqe.inputPlan,
+                    aqe.context,
+                    aqe.preprocessingRules,
+                    aqe.isSubquery,
+                    supportsColumnar = true
+                  ))))
+          case other => rc.withNewChildren(Array(FakeRowAdaptor(other)))
```
JkSelf, line 167: We may need to add `GlutenRowToColumnar ` here.

JkSelf (approved):
LGTM.

PR #1601, opened 2023-05-10
[VL] Test OAP/main_squash branch

zhejiangxiaomai:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


cpp/core/CMakeLists.txt
```diff
@@ -141,7 +141,13 @@ if(USE_AVX512)
 endif()
 
 # Set up Proto
-set(PROTO_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/proto")
+if (NOT DEFINED VELOX_HOME)
+  set(VELOX_HOME ${CMAKE_SOURCE_DIR}/../ep/build-velox/build/velox_ep)
+  message(STATUS "Set VELOX_HOME to ${VELOX_HOME}")
+endif()
+
+set(PROTO_OUTPUT_DIR "${VELOX_HOME}/velox/substrait/proto/")
```
winningsix, line 149: Hmm, it's more suitable to maintain substrait proto files under Gluten since CH backend will use it also, instead of Velox ones.

One option is to maintain exact same proto files under OAP Velox with Gluten ones. And add one Github actions to verify proto contents or manually check during code reviews.

zhejiangxiaomai: Thanks, I will create a new PR to use substrait proto files under Gluten. At present, this method will encounter some problems. Because the proto file created by substrait has different MACRO. we will met `redefinition problem`. This will take a while to research.

cpp/core/CMakeLists.txt
```diff
@@ -141,7 +141,13 @@ if(USE_AVX512)
 endif()
 
 # Set up Proto
-set(PROTO_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/proto")
+if (NOT DEFINED VELOX_HOME)
```
jinchengchenghh, line 144: This code is duplicated with https://github.com/oap-project/gluten/blob/main/cpp/velox/CMakeLists.txt#L40

And core directory is used for common code, can we move it to velox CmakeLists file?

cpp/core/CMakeLists.txt
```diff
@@ -162,7 +168,7 @@ set(CODEGEN_HEADERS
     third_party/
     )
 set(THIRD_PARTY_INCLUDE ${CMAKE_CURRENT_SOURCE_DIR}/third_party)
-include_directories(${CMAKE_CURRENT_SOURCE_DIR} ${THIRD_PARTY_INCLUDE})
+include_directories(${CMAKE_CURRENT_SOURCE_DIR} ${THIRD_PARTY_INCLUDE} ${VELOX_HOME})
```
jinchengchenghh, line 171: If we support other backend in the future, this include will be not suitable

cpp/core/CMakeLists.txt
zhejiangxiaomai: Thanks, Just Updated.

cpp/core/CMakeLists.txt
zhejiangxiaomai: Thanks, I use `if(BUILD_VELOX_BACKEND)` to decide whether we need this.

jinchengchenghh: Can you also move it to velox CmakeLists.list

rui-mo (approved):

PR #1611, opened 2023-05-11
[GLUTEN-1434] [VL] Remove unused arrow code and add GLUTEN_CHECK and GLUTEN_DCHECK

jinchengchenghh:


cpp/core/shuffle/ArrowShuffleWriter.cc
```diff
@@ -34,6 +34,7 @@
 #include "utils/compression.h"
 #include "utils/macros.h"
 
+using namespace arrow;
```
zhztheplayer, line 37: Why this becomes required here?

jinchengchenghh: This file uses many arrow classes? Why not `using namespace arrow` to make code clean? I'm not sure when we could add `using namepsace`.

After all, this file will be deleted after I implement VeloxShuffleWriter.

zhztheplayer: I am fine to adding this but I didn't see corresponding changes remove the `arrow::` usages in this file.

zhztheplayer (approved):

Issue #1626, opened 2023-05-12
[CH] tpcds q78 's round func problem


lhuang09287750:
**Describe the bug**
tpcds q78's result is not correct for round func problem with ch backend
ch backend uses round to even method for float that is not the same with  which spark uses.
that is, round(2.625,2) gets 2.62 with ch backend ,while 2.63 with vanilla spark,mysql or pg

**To Reproduce**
just run tpcds q78 suit in GlutenClickHouseTPCDSParquetSuite

**Expected behavior**
 round func of ch backend  has the same result with the one of vanilla spark

PR #1627, opened 2023-05-12
[GLUTEN-1626][CH] make round func act in the same way as vanilla spark

lhuang09287750:

Add a new ch roundHalfUp func to implement rounding away from zero (towards +inf/-inf) and make gluten round func mapping to it.
No suiltable SSE instructions to realize a fast version func, so use std round func instead

(Fixes: \#1626)

## How was this patch tested?

GlutenMathFunctionsSuite round/bround
GlutenClickHouseTPCDSParquetSuite Q78


liuneng1994:
Run Gluten Clickhouse CI

PR #1639, opened 2023-05-15
[GLUTEN-1638][VL] feat: Add hdfs  support in parquet write

JkSelf:
This patch adds hdfs support for parquet write

(Fixes: #1638)

## How was this patch tested?

### DIFF
--- a/cpp/velox/compute/VeloxParquetDatasource.cc
+++ b/cpp/velox/compute/VeloxParquetDatasource.cc
@@ -49,29 +49,43 @@ void VeloxParquetDatasource::init(const std::unordered_map<std::string, std::str
   pool_ = veloxPool->addLeafChild("velox_parquet_write");
 
   // Construct the file path and writer
-  std::string localPath = "";
+  auto pos = filePath_.find("_temporary", 0);
+  std::string destinationPathWithSchame = filePath_.substr(0, pos - 1);
+
   if (strncmp(filePath_.c_str(), "file:", 5) == 0) {
-    localPath = filePath_.substr(5);
+    finalPath_ = (destinationPathWithSchame + "/" + fileName_).substr(5);
+    std::string command = "touch " + finalPath_;
+    auto ret = system(command.c_str());
+    if (ret != 0) {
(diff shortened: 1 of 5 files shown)


cpp/velox/compute/VeloxParquetDatasource.cc
```diff
@@ -49,29 +49,43 @@ void VeloxParquetDatasource::init(const std::unordered_map<std::string, std::str
...
+    }
 
-  auto pos = localPath.find("_temporary", 0);
-  std::string dirPath = localPath.substr(0, pos - 1);
+    sink_ = std::make_unique<velox::dwio::common::LocalFileSink>(finalPath_);
+  } else if (strncmp(filePath_.c_str(), "hdfs:", 5) == 0) {
+    finalPath_ = destinationPathWithSchame + "/" + fileName_; // with hdfs://vsr246:9000
+    auto destinationPathPos = destinationPathWithSchame.substr(7).find("/", 0);
+    std::string destinationPath = finalPath_.substr(destinationPathPos + 7);
+    std::string command = "hadoop fs -touch " + destinationPath;
```
zhouyuan, line 69: this seems a workaround

PHILO-HE: Can we let it be done by client API? I think we should not assume `hadoop` is available in user's env.

cpp/velox/compute/VeloxParquetDatasource.h
```diff
@@ -34,6 +34,8 @@
 #include "operators/writer/Datasource.h"
 
 #include "velox/common/file/FileSystems.h"
+#include "velox/connectors/hive/storage_adapters/hdfs/HdfsFileSink.h"
```
zhejiangxiaomai, line 37: I think we should use macro `#ifdef ENABLE_HDFS`. like  https://github.com/oap-project/gluten/blob/main/cpp/velox/compute/VeloxInitializer.cc#L28-L30

zhouyuan (approved):

PR #1655, opened 2023-05-16
[Minor] Auto set system arch library path

marin-ma:
Set SYSTEM_LIB_MULTIARCH_PATH to /usr/lib/aarch64-linux-gnu for ARM build. If it's not set, some libraries like re2 cannot be found.


marin-ma:
cc @liujiayi771

zhouyuan (approved):

Issue #1682, opened 2023-05-18
[CH] if input is empty, the hasNext may always be true


shuai-xu:
**Describe the bug**
If the input of a CH transform is empty, the hasNext in LocalExecutor will alwasy return true, and it will return a empty block. The task will run into dead loop.

PR #1683, opened 2023-05-18
[GLUTEN-1682] [CH] fix hasNext will always be true if input is empty

shuai-xu:

This pr fix that if the current block is empty, it should return false for hasNext.

(Fixes: \#1682)

## How was this patch tested?


baibaichen:
LGTM

baibaichen:
Run Gluten Clickhouse CI

liuneng1994 (approved):

PR #1696, opened 2023-05-19
[GLUTEN-1695][VL] Bug: Show the right writen row number of DataWritingCommandExec in spark UI

JkSelf:
This patch make the right written row number of parquet writer in spark UI.
Fixes: https://github.com/oap-project/gluten/issues/1695

## How was this patch tested?


zhouyuan:
Run Gluten Clickhouse CI

shims/write/src/main/scala/org/apache/spark/sql/execution/datasources/BasicWriteStatsTracker.scala
```diff
@@ -0,0 +1,252 @@
...
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.execution.SQLExecution
+import org.apache.spark.sql.execution.datasources.BasicWriteJobStatsTracker._
+import org.apache.spark.sql.execution.metric.{SQLMetric, SQLMetrics}
+import org.apache.spark.util.SerializableConfiguration
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
+
```
jinchengchenghh, line 29: Is this file override spark class file? Why add new module `write`, you can add it to module `common` if it is same for all the spark version.

Can we have other solution to fix this bug?

JkSelf: Thanks for your review. We add a write directory here to solve the dependency problem. Because we need to access `FakeRow `under the `gluten-core` package in the `common ` directory, but since `gluten-core` depends on `commom`, a circular dependency problem will be formed. And This fix is to replace the `BasicWriteTaskStat ` class in vanilla spark. I think it is more appropriate to put it in Shim layer. Or do you have any better suggestion?

jinchengchenghh: If it only use `FakeRow`, I think we can also put FakeRow to module `common`.

Other projects dependency maybe
-spark33
   -gluten-core
   -gluten-data

This will solve the problem, if this problems occur too many times, we can try to refactor code.

JkSelf: `FakeRow `is defined in the Gluten project and is not suitable for shim layer. How about adding a new `gluten-datasource` package and move all the classes related to native write to this package?

JkSelf: @binmahone @rui-mo Do you have any suggestion?

JkSelf: Offline discussed with @binmahone , prefer to move `FakeRow ` into `common`. @jinchengchenghh already updated, please help to review again. Thanks.

jinchengchenghh (approved):

Issue #1478, opened 2023-04-23
[VL] Unit tests with incorrect result


rui-mo:
- [x] "cast from float"   --feilong
- [x] "cast from double"  --feilong
- [x] "from decimal"   --rui
- [x] "cast string to date #&#8203;2"  --feilong
- [x] "casting to fixed-precision decimals" --rui
- [x] "cast from date"  --feilong
- [x] "SPARK-32828: cast from a derived user-defined type to a base type" --rui
- [x] "Fast fail for cast string type to decimal type"  --rui
- [x] "missing cases - from boolean" --feilong
- [x] "Fast fail for cast string type to decimal type in ansi mode"  --rui
- [x] "Fast fail for cast string type to decimal type in ansi mode" --rui
- [x] "Fast fail for cast string type to decimal type in ansi mode" --rui
- [x] "Fast fail for cast string type to decimal type in ansi mode" --rui
- [x] "default" --jian
- [x] "decimal" --jian
- [x] "seconds" -- rong
- [x] "ANSI: extract days, hours, minutes and seconds" -- rong
- [x] "SPARK-30633: xxHash with different type seeds" -- rong
- [x] "Map Concat"   --ke
- [x] "Shuffle"         -- ke
- [x] "DayOfYear"   --ke
- [x] "Year"             --ke
- [x] "Quarter"       --ke 
- [x] "Month"        --ke
- [x] "Day / DayOfMonth"  --ke
- [x] "DayOfWeek"     --ke
- [x] "MakeDecimal"  --ke
- [x] "BinaryComparison: lessThan" --yan
- [x] "BinaryComparison: LessThanOrEqual" --yan
- [x] "BinaryComparison: GreaterThan" --yan
- [x] "BinaryComparison: GreaterThanOrEqual" ---yan
- [x] "SPARK-32764: compare special double/float values" --yan
- [x] "cos" --hongze
- [x] "cosh" --hongze
- [x] "toDegrees" --hongze
- [x] "toRadians" --hongze
- [x] "cbrt" --hongze
- [x] "exp" --hongze
- [x] "log10" --hongze
- [x] "log2" --hongze
- [x] "pow" --hongze
- [x] "atan2" --hongze
- [x] "Substring"
- [x] "string for ascii" ---izchen
- [x] "replace" ---izchen
- [x] "MonotonicallyIncreasingID"
- [x] "SparkPartitionID"
- [x] "random"
- [x] "SPARK-9127 codegen with long seed"
[...]

PR #1697, opened 2023-05-19
[GLUTEN-1478]fix: decimal converter in check

Ma-Jian1:

(#1478) decimal check in "default" test case

## How was this patch tested?

ut


gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
```diff
@@ -174,7 +174,7 @@ trait GlutenTestsTrait extends GlutenTestsCommonTrait {
     }
     if (!(checkResult(result.head.get(0), expected, expression.dataType, expression.nullable)
       || checkResult(
-      CatalystTypeConverters.convertToCatalyst(result.head.get(0)),
+      CatalystTypeConverters.createToCatalystConverter(expression.dataType)(result.head.get(0)),
```
marin-ma, line 177: Does this change indicate the result schema could be wrong? If so, we should add some notes here.

Ma-Jian1:
close this with #1735 merged

PR #1715, opened 2023-05-22
[GLUTEN-1714][VL][Fix] Align the implementation for ascii func with latest spark

PHILO-HE:

Spark has a fix for ascii function. Since it is more like a bug fix in spark, we are proposing to align the imp. Spark 3.2.2 users should expect the behavior is consistent with spark 3.2.3 for characters whose ascii is above 127. 

## How was this patch tested?
UT.

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -2,8 +2,8 @@
 
 set -exu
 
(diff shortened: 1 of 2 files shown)


PHILO-HE:
Need to rebase on the code with chr function fixed.

rui-mo (approved):

PR #1730, opened 2023-05-23
[GLUTEN-1478][VL]Fix SPARK-30633: xxHash with different type seeds

marin-ma:
https://github.com/oap-project/gluten/issues/1478

Depends on https://github.com/oap-project/velox/pull/275


gluten-core/src/main/scala/io/glutenproject/expression/HashExpressionTransformer.scala
```diff
@@ -30,6 +30,15 @@ class HashExpressionTransformer(
   extends ExpressionTransformer with Logging {
 
   override def doTransform(args: java.lang.Object): ExpressionNode = {
+    original match {
+      case XxHash64(_, seed) if seed != 42 =>
```
PHILO-HE, line 34: @marin-ma, can we move the check logic to velox side? See `SubstraitToVeloxPlanValidator.cpp`.
I am afraid this check is not applicable to CH backend. And we prefer to letting each backend do their specific check. Thanks!

marin-ma: As of now, we can only get the `seed` value on Java side, because in this case the `seed` is not in the spark2substrait conversion.

gluten-data/src/main/scala/io/glutenproject/expression/GlutenExpressionTransformer.scala
```diff
@@ -54,3 +55,37 @@ class GlutenNamedStructTransformer(substraitExprName: String,
     ExpressionBuilder.makeScalarFunction(functionId, expressionNodes, typeNode)
   }
 }
+
+
+class GlutenHashExpressionTransformer(
```
PHILO-HE, line 60: nit: better to make it a case class and remove `new` at the instantiating place.

PHILO-HE (approved):
LGTM.

gluten-data/src/main/scala/io/glutenproject/expression/GlutenExpressionTransformer.scala
jinchengchenghh: may must make it case class, otherwise the equal() function will return wrong result and then operator such as ReuseExchange cannot make sense

jinchengchenghh (dismissed):

PHILO-HE (approved):

Issue #1725, opened 2023-05-22
[CH] Change temporary storage path to current working dir instead of '/tmp/libch'


exmy:
This will avoid exception `Not enough space on temporary disk: While executing AggregatingTransform`

PR #1737, opened 2023-05-23
[GLUTEN-1725][CH]  Change temporary storage path from '/tmp/libch' to current working dir

exmy:

(Fixes: #1725)

## How was this patch tested?

manual tests


baibaichen:
I think we should configure the temporary directory instead of chaning it in the code.

exmy:
cc @liuneng1994 @zzcclp

liuneng1994 (approved):
LGTM

PR #1739, opened 2023-05-23
[GLUTEN-1478] Support ordered result check for MapData

rui-mo:

Support ordered result check for MapData.

#1478

## How was this patch tested?

Testing on CI.


gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
```diff
@@ -145,6 +147,62 @@ trait GlutenTestsTrait extends GlutenTestsCommonTrait {
     }
   }
 
+  /**
+   * Sort map data by key and return the sorted key array and value array.
+   * TODO: Sort requires known type, so Any does not work here.
```
zhztheplayer, line 152: I had tried a little and the following can pass compilation, would you like to check if it actually works? Thanks.

```scala
case (_, _) =>
  val keyArray = input.keyArray().toArray[Any](kt)
  val valueArray = input.valueArray().toArray[Any](vt)
  val newMap = (keyArray zip valueArray).toMap
  val sortedMap = mutable.SortedMap(newMap.toSeq: _*)(TypeUtils.getInterpretedOrdering(kt))
  (new GenericArrayData(sortedMap.keys.toArray),
    new GenericArrayData(sortedMap.values.toArray))
```

zhztheplayer (dismissed):
the comment was non-blocking

gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
rui-mo: Thanks for your instructive comment. It works.

zhztheplayer (approved):

PR #1748, opened 2023-05-24
[GLUTEN-1747][VL] Change the profile for backends velox and rss

jackylee-ch:
Currently, we found mvn test run failed with `backends-velox-without-rss`, as `gluten-ut` module does not support `backends-velox-without-rss`, and would not add backends dependencies into it. 
Although it works while adding `backends-velox-without-rss` in spark-ut, I found it look awful as we have too many similar config in pom.xml. The `backends-velox` in pom.xml has nothring to do with rss in almost every module except for `backends-velox` module and parent `pom.xml`.
 Thus I recomanded we use `-Pbackends` to define the config for velox and use `-Prss` to define the config for rss module, such as `gluten-celeborn`.

## How was this patch tested?
unit tests

### DIFF
--- a/.github/workflows/velox_be.yml
+++ b/.github/workflows/velox_be.yml
@@ -75,17 +75,17 @@ jobs:
         run: |
           docker exec velox-backend-ubuntu2004-test-$GITHUB_RUN_ID bash -c '
           cd /opt/gluten && \
-          mvn clean install -Pspark-3.2 -Pbackends-velox -DskipTests -Dspark32.version=3.2.0'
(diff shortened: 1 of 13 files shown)


jackylee-ch:
cc @zhztheplayer @zhouyuan @kerwin-zk

kerwin-zk (approved):

zhztheplayer (approved):
Can you rebase? Thanks.

jackylee-ch:
@zhztheplayer @zhouyuan CI has been passed now.

zhztheplayer (approved):

PR #1757, opened 2023-05-25
[VL] Make git checkout script in get_arrow.sh use an unique local tag name

zhztheplayer:
Follow-up to https://github.com/oap-project/gluten/pull/1660

Why doing this is not over-engineering (correct me if I am wrong):
1. It makes the checkout code 100% same with the checkout code in get_velox.sh. 
2. It fixes a bug when changing to another tag (e.g `ARROW_BRANCH=apache-arrow-13.0.0`), say if we upgrade to `apache-arrow-13.0.0`, `apache-arrow-12.0.0` will be wrongly kept and used without any message reported, simple to reproduce.

### DIFF
--- a/ep/build-arrow/src/get_arrow.sh
+++ b/ep/build-arrow/src/get_arrow.sh
@@ -34,8 +34,12 @@ function checkout_code {
     echo "Arrow source folder $ARROW_SOURCE_DIR already exists..."
     cd $ARROW_SOURCE_DIR
     git init .
-    git fetch $ARROW_REPO tag $ARROW_BRANCH
+    EXISTS=$(git show-ref refs/tags/build_$TARGET_BUILD_COMMIT || true)
+    if [ -z "$EXISTS" ]; then
+      git fetch $ARROW_REPO $TARGET_BUILD_COMMIT:refs/tags/build_$TARGET_BUILD_COMMIT
+    fi
     git reset --hard HEAD
+    git checkout refs/tags/build_$TARGET_BUILD_COMMIT
   else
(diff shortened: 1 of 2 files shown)


ep/build-arrow/src/get_arrow.sh
```diff
@@ -29,16 +29,22 @@ for arg in "$@"; do
...
+    EXISTS=$(git show-ref refs/tags/build_$TARGET_BUILD_COMMIT || true)
+    if [ -z "$EXISTS" ]; then
+      git fetch $ARROW_REPO $TARGET_BUILD_COMMIT:refs/tags/build_$TARGET_BUILD_COMMIT
+    fi
     git reset --hard HEAD
+    git checkout refs/tags/build_$TARGET_BUILD_COMMIT
   else
     git clone $ARROW_REPO -b $ARROW_BRANCH $ARROW_SOURCE_DIR
     cd $ARROW_SOURCE_DIR
+    git checkout $TARGET_BUILD_COMMIT
```
jinchengchenghh, line 47: is it `git checkout refs/tags/build_$TARGET_BUILD_COMMIT`?

zhztheplayer: `refs/tags/build_$TARGET_BUILD_COMMIT` will only be created in the previous if branch, running into the second branch means it's a brand-new build, we can ensure the current branch contains the target commit, so `so git checkout $TARGET_BUILD_COMMIT` can be enough.

jinchengchenghh (approved):

PR #1762, opened 2023-05-25
[VL] Do not unload shared libraries

zhztheplayer:
Unloading could be problematical when user pre-installs the shared libraries on executor's OS. Removing it in this patch.


zhouyuan (approved):
👍

Issue #1767, opened 2023-05-25
[CH] Argument's type illegal when query posexplode(split(map['key'], ','))


KevinyhZou:
**Describe the bug**
While do the query like `select id from test_tbl lateral view posexplode(split(data['k'], ',')) tx as a, b`, the exception while throw like below

```
java.lang.RuntimeException: Second argument for function mapFromArrays must be Array or Map
0. Poco::Exception::Exception(String const&, int) @ 0x1454b5ba in /data6/zouyunhe/gluten-ck-backend/libch.so
1. DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0xde5bb75 in /data6/zouyunhe/gluten-ck-backend/libch.so
2. DB::Exception::Exception<String>(int, FormatStringHelperImpl<std::type_identity<String>::type>, String&&) @ 0x69b2d04 in
3. DB::(anonymous namespace)::FunctionMapFromArrays::getReturnTypeImpl(std::vector<std::shared_ptr<DB::IDataType const>, std::allocator<std::shared_ptr<DB::IDataType const>>> const&) const @ 0x9b33ba4 in /da
ta6/zouyunhe/gluten-ck-backend/libch.so
4. DB::IFunction::getReturnTypeImpl(std::vector<DB::ColumnWithTypeAndName, std::allocator<DB::ColumnWithTypeAndName>> const&) const @ 0x6d97c13 in /data6/zouyunhe/gluten-ck-backend/libch.so
5. DB::IFunctionOverloadResolver::getReturnTypeWithoutLowCardinality(std::vector<DB::ColumnWithTypeAndName, std::allocator<DB::ColumnWithTypeAndName>> const&) const @ 0xf17943b in /data6/zouyunhe/gluten-ck-b
ackend/libch.so
6. DB::IFunctionOverloadResolver::getReturnType(std::vector<DB::ColumnWithTypeAndName, std::allocator<DB::ColumnWithTypeAndName>> const&) const @ 0xf17917f in /data6/zouyunhe/gluten-ck-backend/libch.so
7. DB::IFunctionOverloadResolver::build(std::vector<DB::ColumnWithTypeAndName, std::allocator<DB::ColumnWithTypeAndName>> const&) const @ 0xf179a54 in /data6/zouyunhe/gluten-ck-backend/libch.so
[...]

PR #1768, opened 2023-05-25
[GLUTEN-1767][CH]Bug fix posexplode function argument check error

KevinyhZou:

(Fixes: \#1767)

posexplode in clickhouse backend will make the convert: 
posexplode(array_or_map) -> CH: arrayJoin(map), in which map = mapFromArrays(range(length(array_or_map)), array_or_map)

But somethimes in mapFromArrays the second argument type is nullable, which can not pass the arugment check.

And when use the posexplode function to transfrom value from a map, like posexplode(split(map['key'], ',')), in clickhosue backend will  use the arrayElement function to extract the map value, and it would force the convert the key argument to UInt32, and this also will be check as illegal argument .

## How was this patch tested?

unit tests

### DIFF
--- a/cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
+++ b/cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
@@ -1316,19 +1316,15 @@ void SerializedPlanParser::parseFunctionArguments(
         // could be positive or negative and have different effects. So we make a cast here.
         // In clickhosue, map element are also accessed by arrayElement, not make the cast.
         parseFunctionArgument(actions_dag, parsed_args, required_columns, function_name, args[0]);
-        auto element_type = actions_dag->getNodes().back().result_type;
-        const auto * first_arg_type = parsed_args.back()->result_type.get();
-        const auto * nested_type = element_type.get();
-        if (nested_type->isNullable())
-        {
-            nested_type = typeid_cast<const DB::DataTypeNullable *>(nested_type)->getNestedType().get();
-        }
+        auto element_type = parsed_args.back()->result_type;
+        const auto * first_arg_type = element_type.get();
         if (first_arg_type->isNullable())
         {
             first_arg_type = typeid_cast<const DB::DataTypeNullable *>(first_arg_type)->getNestedType().get();
         }
         const auto * index_node = parseFunctionArgument(actions_dag, required_columns, function_name, args[1]);
-        if (nested_type->getTypeId() == DB::TypeIndex::Array && first_arg_type->getTypeId() != DB::TypeIndex::Map)
+        WhichDataType which(first_arg_type);
+        if (which.isArray())
         {
             DB::DataTypeNullable target_type(std::make_shared<DB::DataTypeUInt32>());
             index_node = ActionsDAGUtil::convertNodeType(actions_dag, index_node, target_type.getName());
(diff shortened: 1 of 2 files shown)


cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
```diff
@@ -1317,13 +1317,18 @@ void SerializedPlanParser::parseFunctionArguments(
...
         {
             nested_type = typeid_cast<const DB::DataTypeNullable *>(nested_type)->getNestedType().get();
         }
+        if (first_arg_type->isNullable())
+        {
+            first_arg_type = typeid_cast<const DB::DataTypeNullable *>(first_arg_type)->getNestedType().get();
+        }
         const auto * index_node = parseFunctionArgument(actions_dag, required_columns, function_name, args[1]);
-        if (nested_type->getTypeId() == DB::TypeIndex::Array)
+        if (nested_type->getTypeId() == DB::TypeIndex::Array && first_arg_type->getTypeId() != DB::TypeIndex::Map)
```
taiyang-li, line 1331: it is more simple to use this function 
```
inline bool isArray(const DataTypePtr & data_type) { return WhichDataType(data_type).isArray(); }

```

cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
```diff
@@ -1317,13 +1317,18 @@ void SerializedPlanParser::parseFunctionArguments(
         // In clickhosue, map element are also accessed by arrayElement, not make the cast.
         parseFunctionArgument(actions_dag, parsed_args, required_columns, function_name, args[0]);
         auto element_type = actions_dag->getNodes().back().result_type;
```
taiyang-li, line 1319: auto element_type = parsed_args.back().result_type;

cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
```diff
@@ -1317,13 +1317,18 @@ void SerializedPlanParser::parseFunctionArguments(
...
+        if (first_arg_type->isNullable())
+        {
+            first_arg_type = typeid_cast<const DB::DataTypeNullable *>(first_arg_type)->getNestedType().get();
+        }
         const auto * index_node = parseFunctionArgument(actions_dag, required_columns, function_name, args[1]);
-        if (nested_type->getTypeId() == DB::TypeIndex::Array)
+        if (nested_type->getTypeId() == DB::TypeIndex::Array && first_arg_type->getTypeId() != DB::TypeIndex::Map)
         {
             DB::DataTypeNullable target_type(std::make_shared<DB::DataTypeUInt32>());
             index_node = ActionsDAGUtil::convertNodeType(actions_dag, index_node, target_type.getName());
```
taiyang-li, line 1334: Actually, function arrayElement's second argument accept all kind of native interger types, like (U)Int8/16/32/64, refer to src/Functions/array/arrayElement.cpp. There is not need to cast it to UInt32, it will cause performance degradation.

cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
taiyang-li, line 1331: `nested_type` and `first_arg_type` are both the non-nullable data type of first arg. The second condition is always true if the first conditions is true.

KevinyhZou: fix done.

cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
lgbo-ustc: sign and unsign integer in clickhouse has different result

baibaichen:
LGTM

liuneng1994 (approved):

Issue #1585, opened 2023-05-09
Crash on submit query to Yarn cluster with clickchouse backend


zhanglistar:
**Describe the bug**
Crash on submit query to Yarn cluster with clickchouse backend.
gluten version: 9768880a44399b41aa449e28aab54df58499f168
kyligence ch version: 52be833fd21d3289623ce4ae682f6537e438fb57
spark version: 3.2.2
The stack is:
```
* thread #1, name = 'bash', stop reason = signal SIGSEGV
  * frame #0: 0x00007fe7bc8a3389 libch.so`sallocx [inlined] atomic_load_p(a=0x0000000000010f58, mo=atomic_memory_order_relaxed) at atomic.h:83:1
    frame #1: 0x00007fe7bc8a3389 libch.so`sallocx [inlined] rtree_leaf_elm_bits_read(elm=0x0000000000010f58, dependent=true) at rtree.h:182
    frame #2: 0x00007fe7bc8a3389 libch.so`sallocx [inlined] rtree_leaf_elm_read(elm=0x0000000000010f58, dependent=true) at rtree.h:242
    frame #3: 0x00007fe7bc8a3389 libch.so`sallocx [inlined] rtree_metadata_read(rtree_ctx=0x00007fe7be5088c0, key=<unavailable>) at rtree.h:446
    frame #4: 0x00007fe7bc8a3357 libch.so`sallocx [inlined] emap_alloc_ctx_lookup(tsdn=<unavailable>, ptr=<unavailable>) at emap.h:238
    frame #5: 0x00007fe7bc8a313c libch.so`sallocx [inlined] arena_salloc(tsdn=<unavailable>, ptr=<unavailable>) at arena_inlines_b.h:176
    frame #6: 0x00007fe7bc8a313c libch.so`sallocx [inlined] isalloc(tsdn=<unavailable>, ptr=<unavailable>) at jemalloc_internal_inlines_c.h:40
    frame #7: 0x00007fe7bc8a313c libch.so`sallocx(ptr=<unavailable>, flags=<unavailable>) at jemalloc.c:3905
    frame #8: 0x00007fe7b45854f9 libch.so`operator delete(void*, unsigned long) [inlined] void Memory::untrackMemory<>(ptr=0x00000000021eb008, size=16) at memory.h:204:44
    frame #9: 0x00007fe7b45854ca libch.so`operator delete(ptr=0x00000000021eb008, size=16) at new_delete.cpp:157
[...]

PR #1770, opened 2023-05-25
Fix libch.so crash on ubuntu 16.04

zhanglistar:

(Fixes: \#1585)

## How was this patch tested?

Manually test.


zhanglistar:
There is no need to using `-fvisibility=hidden -fvisibility-inlines-hidden` since we have version script.

PR #1782, opened 2023-05-26
[VL] Enable preferSpill=false for LocalPartitionWriter

marin-ma:
In order to resolve the shuffle writer small file issue, all cached partitions will be spilled into one file per call.

During the split operation, we pre-allocate the memory for each reducer, size is controlled by batch size, default is 4K. Once the 4K batch is filled, preferSpill control the next step. 
- If preferSpill=True, Gluten creates a small file for each reducer and then serialize the 4K batch as Arrow IPC payload and spill to the small file immediately. At the end of task, the small files are merged into a large file and pass to driver along with reducer partition id and its offset in large file.
- If preferSpill=False, Gluten will create the Arrow IPC payloads which is compressed and then cached in memory. When memory is not enough the spill interface is called from Spark's memory management, then we spill all the cached batches to a single file. Release the memory. At the end of task, the spill files are merged into a large file, then follow the same steps as preferSpill=True. In this way one file is created for each OOM which is the same as Vanilla Spark's unsafeShuffleWritter.


FelixYBW:
@jinchengchenghh  is there any conflict with your PR? I'd expect some but not much.

FelixYBW:
@zzcclp The PR solved the small file issue. Perf gained a little.

jinchengchenghh:
Yes, very much conflict, I will fix it

PR #1791, opened 2023-05-29
[GLUTEN-1794][VL] support split preload

zhli1142015:
support split preload

Fixes: #1794

### DIFF
--- a/cpp/velox/compute/VeloxInitializer.cc
+++ b/cpp/velox/compute/VeloxInitializer.cc
@@ -62,7 +62,7 @@ const std::string kVeloxSsdCacheIOThreadsDefault = "1";
 const std::string kVeloxSsdODirectEnabled = "spark.gluten.sql.columnar.backend.velox.ssdODirect";
 
 const std::string kVeloxIOThreads = "spark.gluten.sql.columnar.backend.velox.IOThreads";
-const std::string kVeloxIOThreadsDefault = "1";
+const std::string kVeloxIOThreadsDefault = "0";
 
 const std::string kVeloxSplitPreloadPerDriver = "spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver";
 const std::string kVeloxSplitPreloadPerDriverDefault = "2";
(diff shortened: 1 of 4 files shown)


shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -756,6 +766,20 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(false)
 
+  val COLUMNAR_VELOX_CONNECTOR_IO_THREADS =
+    buildConf("spark.gluten.sql.columnar.backend.velox.IOThreads")
+      .internal()
+      .doc("The IO threads for cpnnector split preloading")
```
zhouyuan, line 772:
```suggestion
      .doc("The IO threads for connector split preloading")
```

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -756,6 +766,20 @@ object GlutenConfig {
...
+      .intConf
+      .createWithDefault(Runtime.getRuntime.availableProcessors)
```
zhouyuan, line 774: using all threads by default maybe too aggressive, how about set this to 1?

cpp/velox/compute/VeloxInitializer.cc
```diff
@@ -242,7 +238,27 @@ void VeloxInitializer::initCache(const std::unordered_map<std::string, std::stri
...
+  }
+  int32_t splitPreloadPerDriver = std::stoi(kVeloxSplitPreloadPerDriverDefault);
+  got = conf.find(kVeloxSplitPreloadPerDriver);
+  if (got != conf.end()) {
+    splitPreloadPerDriver = std::stoi(got->second);
+  }
+  if (ioThreads > 0) {
+    ioExecutor_ = std::make_unique<folly::IOThreadPoolExecutor>(ioThreads);
+  }
+  FLAGS_split_preload_per_driver = splitPreloadPerDriver;
```
zhouyuan, line 259: move to L259 to `if { }` ?

zhouyuan (commented):
works for me, many thanks for improving this!

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
zhli1142015: Thanks, updated.

cpp/velox/compute/VeloxInitializer.cc
zhli1142015: Moved line 259 into if block.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
zhli1142015: synced with yuan offline, set it as 0 to disable this feature by default.

zhouyuan (approved):
👍

Issue #1792, opened 2023-05-29
[VL] Make cache cleaned on Executor Exit


jackylee-ch:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
Currently, the cache files writing by velox cache won't be removed after Executor exit. Since the cache file can not be reused, it may cause disk full when restarting Executor or Application.

**Describe the solution you'd like**
Clean cache file on Executor Exit.

PR #1793, opened 2023-05-29
[GLUTEN-1792][VL] Remove local cache files on Executor exit

jackylee-ch:
Remove cache file on Executor exit to prevent disk full as Executor restarts.
Fixes: #1792 

## How was this patch tested?
unit tests


zhouyuan (approved):
👍

FelixYBW:
@zhouyuan how many files it needs to remove? If number is too large, let's move it to an async thread

jackylee-ch:
> @zhouyuan how many files it needs to remove? If number is too large, let's move it to an async thread

Basically, this is controlled by `spark.gluten.sql.columnar.backend.velox.ssdCacheShards`, we won't have too many files to delete.

PR #1821, opened 2023-06-01
[VL][BUILD] Fix use pre-build arrow/parquet/thrift lib

Yohahaha:
In Ubuntu, thrift will be installed manually in setup scripts, but Arrow still compile thrift, then Velox use system thrift and pre-build arrow/parquet.
In Centos, Velox could not found system thrift, so it will compile Arrow and thrift twice.
Since Arrow will compile thrift in all environments, lets use these pre-build shared libs to save time and keep toolchain consistency.

Support native dependency could be overrided by env variables.

### DIFF
--- a/cpp/CMake/BuildGTest.cmake
+++ b/cpp/CMake/BuildGTest.cmake
@@ -8,6 +8,10 @@ set(GLUTEN_GTEST_SOURCE_URL
     "https://github.com/google/googletest/archive/refs/tags/v${GLUTEN_GTEST_VERSION}.tar.gz"
     )
 
+if(DEFINED ENV{GLUTEN_GTEST_SOURCE_URL})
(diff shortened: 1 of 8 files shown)


Yohahaha:
@zhouyuan @ccat3z @zhejiangxiaomai

Yohahaha:
Why we need link thrift for gluten cpp module? It should be linked under Velox as static lib. @ccat3z

Yohahaha:
All CI has been passed, could someone help review this? @zhouyuan @zhejiangxiaomai @ccat3z

ccat3z (dismissed):

Yohahaha:
Thanks for approving! could we merge? @ccat3z @zhejiangxiaomai

Issue #1843, opened 2023-06-05
[CH] HiveTableScanTransformer infer dirty partitions would throw exception


KevinyhZou:
**Describe the bug**
When use hive txt read a txt format table, which partitions may be have illegal paths, and the HiveTableScanTransformer would use file index to inter partitions need to read, the exception would be throw
```
hdfs://testcluster/abc/day=2020-11-06/hour=11
hdfs://testcluster/abc/_temp_day=2020-11-07_20201116_120932_00738_ksdid
        at scala.Predef$.assert(Predef.scala:223)
        at org.apache.spark.sql.execution.datasources.PartitioningUtils$.resolvePartitions(PartitioningUtils.scala:380)
        at org.apache.spark.sql.execution.datasources.PartitioningUtils$.parsePartitions(PartitioningUtils.scala:181)
        at org.apache.spark.sql.execution.datasources.PartitioningUtils$.parsePartitions(PartitioningUtils.scala:110)
        at org.apache.spark.sql.execution.datasources.PartitioningAwareFileIndex.inferPartitioning(PartitioningAwareFileIndex.scala:158)
        at org.apache.spark.sql.execution.datasources.InMemoryFileIndex.partitionSpec(InMemoryFileIndex.scala:73)
        at org.apache.spark.sql.execution.datasources.PartitioningAwareFileIndex.listFiles(PartitioningAwareFileIndex.scala:74)
        at org.apache.spark.sql.execution.datasources.v2.FileScan.partitions(FileScan.scala:138)
        at org.apache.spark.sql.execution.datasources.v2.FileScan.partitions$(FileScan.scala:137)
        at org.apache.spark.sql.execution.datasources.v2.TextBasedFileScan.partitions(TextBasedFileScan.scala:28)
        at org.apache.spark.sql.execution.datasources.v2.FileScan.planInputPartitions(FileScan.scala:183)
        at org.apache.spark.sql.execution.datasources.v2.FileScan.planInputPartitions$(FileScan.scala:182)
        at org.apache.spark.sql.execution.datasources.v2.TextBasedFileScan.planInputPartitions(TextBasedFileScan.scala:28)
        at org.apache.spark.sql.hive.CHHiveTableScanExecTransformer.<init>(CHHiveTableScanExecTransformer.scala:122)
        at org.apache.spark.sql.hive.CHHiveTableScanExecTransformer$.transform(CHHiveTableScanExecTransformer.scala:212)
[...]

PR #1844, opened 2023-06-05
[GLUTEN-1843][CH]Bug fix for HiveScanExecTransformer infer dirty partitions

KevinyhZou:
When scan a hive txt table, HiveTableScanExecTransformer would get partitions use FileIndex, In the implements, the implementation is InMemoryFileIndex. If no partition spec defined, it would inter partitions by itself, and validate the partition columns, if the partition path is illegal, such as `hdfs://testcluster/abc/_temp_day=20221106abcwda11`, the exception will throws, and the query will be failed.

So here we defined the partition spec first in InMemoryIndex, which will fiter the dirty partitons like above

(Fixes: \#1843)

## How was this patch tested?
unit tests

### DIFF
--- a/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseHiveTableSuite.scala
+++ b/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseHiveTableSuite.scala
@@ -127,7 +127,7 @@ class GlutenClickHouseHiveTableSuite()
     "array_field array<int>," +
     "array_field_with_null array<int>," +
     "map_field map<int, long>," +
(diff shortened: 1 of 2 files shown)


zzcclp:
please modify the PR title and add some ut

KevinyhZou:
> please modify the PR title and add some ut

done @zzcclp

KevinyhZou:
cc @zzcclp @liuneng1994

loneylee:
LGTM

KevinyhZou:
This problem is sloved in https://github.com/oap-project/gluten/pull/2023, So only add UT.

baibaichen (approved):
LGTM

PR #1852, opened 2023-06-05
[GLUTEN-2104][CH] Fix size(null) throw exception when spark.sql.legacy.sizeOfNull is true

exmy:
close: #2104 
when `spark.sql.legacy.sizeOfNull` is true,  size(null) should return -1, but currently it will throw exception 
```
Cannot convert NULL value to non-Nullable type: while converting source column `length(_1#572)` to destination column `length(_1#572)`: while executing 'FUNCTION _CAST(length(_1#572) :: 2, Int32 :: 1) -> length(_1#572) Int32 : 0'
```
## How was this patch tested?

spark ut


zzcclp:
it's better add some test cases in ch bakcend.

rui-mo (dismissed):

exmy:
> it's better add some test cases in ch bakcend.

OK

zzcclp (approved):
LGTM

Issue #1871, opened 2023-06-07
[CH] There is a race for BroadcastJoinBuilder :: storage_join_map


baibaichen:
**Describe the bug**

Occasionally，CICD failed with following error.
 ```
java.lang.RuntimeException: unordered_map::at: key not found
[2023-06-01T05:45:14.931Z] 0. /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/cpp-ch/build/./contrib/llvm-project/libcxx/include/stdexcept:166: std::out_of_range::out_of_range[abi:v15000](char const*) @ 0x000000000603d4e9 in /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/cpp-ch/build/utils/extern-local-engine/libch.so
[2023-06-01T05:45:14.931Z] 1. /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/cpp-ch/build/./contrib/llvm-project/libcxx/include/stdexcept:268: std::__throw_out_of_range[abi:v15000](char const*) @ 0x000000000603d4a0 in /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/cpp-ch/build/utils/extern-local-engine/libch.so
[2023-06-01T05:45:14.931Z] 2. /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/cpp-ch/build/./contrib/llvm-project/libcxx/include/unordered_map:0: local_engine::SerializedPlanParser::parseJoin(substrait::JoinRel, std::unique_ptr<DB::QueryPlan, std::default_delete<DB::QueryPlan>>, std::unique_ptr<DB::QueryPlan, std::default_delete<DB::QueryPlan>>, std::vector<DB::IQueryPlanStep*, std::allocator<DB::IQueryPlanStep*>>&) @ 0x00000000060b6db4 in /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/cpp-ch/build/utils/extern-local-engine/libch.so
[...]

PR #1873, opened 2023-06-07
[GLUTEN-1871][CH]Fix: Remove BroadcastJoinBuilder :: storage_join_map and using guava cache to synchronize access

baibaichen:

In the old implementation, we have two caches for BHJ hashtable.
1. At JVM side, there is a `CHBroadcastBuildSideCache::buildSideRelationCache`, which is used to avoid accessing Broadcast value multiple times.
2. At C++ side there is a `BroadcastJoinBuilder :: storage_join_map`, which is used to avoid building hash table multiple times.

The `BroadcastJoinBuilder :: storage_join_map`(`std::unorder_map`) isn't thread safe, so we must have a lock to protect it. It turns out that `BroadcastJoinBuilder :: storage_join_map` is redundant, since `buildSideRelationCache` is guava cache which is thread safe.  We can use it at c++ backend to get a hash table without synchronization by ourselves.

### Warning

This implementation still has a flaw, if we set `GLUTEN_CLICKHOUSE_BROADCAST_CACHE_EXPIRED_TIME` quite short, then we probably can't get hash table, because it was invalidated by guaca cache.

However, we can‘t rebuild the hash table since we are not in the context of `BroadcastBuildSideRDD`, so we can't get `Broadcast[BuildSideRelation]`, even though it is still in the JVM.

(Fixes: \#1871)

## How was this patch tested?

Using existed UT.


baibaichen:
Run Gluten Clickhouse CI

baibaichen:
> Run Gluten Clickhouse CI

3

baibaichen:
Run Gluten Clickhouse CI

5

baibaichen:
Run Gluten Clickhouse CI

7

baibaichen:
> Run Gluten Clickhouse CI
> 
> 7


An irrelevant error

baibaichen:
Run Gluten Clickhouse CI

8

An irrelevant error

baibaichen:
Run Gluten Clickhouse CI

baibaichen:
Run Gluten Clickhouse CI

zzcclp (approved):
LGTM

PR #1886, opened 2023-06-08
[CH] Early throw exception if ColumnarBatch can't convert to CHNativeBlock

exmy:

We may encounter the failure of converting ColumnarBatch to CHNativeBlock in rare cases, but we are unable to obtain more meaningful error info when it occurs. I think we can early throw exception with more meaningful info in the situation.


exmy:
cc @zzcclp

zzcclp (approved):
LGTM

liuneng1994 (approved):
LGTM

PR #1888, opened 2023-06-08
[VL] Fix QAT/IAA build

marin-ma:
Macro GLUTEN_ENABLE_QAT and GLUTEN_ENABLE IAA were not ported to "cpp/velox" after separating the shuffle writer for backends.


marin-ma:
cc @weiting-chen

zhztheplayer (approved):

PR #1895, opened 2023-06-09
[GLUTEN-1871][CH] Refactor: StorageJoinFromReadBuffer doesn't inherit from StorageSetOrJoinBase any more.

baibaichen:

This PR has two purposes:

1. This is part of https://github.com/oap-project/gluten/pull/1873, that PR is quite complex, and needs some time to verify.
2. Also, we need this pr to avoid building error due to https://github.com/ClickHouse/ClickHouse/pull/50663, so that gluten can reduce dependency.

## How was this patch tested?
Using existed UT.


liuneng1994 (approved):
LGTM

PR #1904, opened 2023-06-09
[GLUTEN-1434] [VL] Shuffle velox to velox and broadcast velox RowVector

jinchengchenghh:


cpp/core/jni/JniWrapper.cc
```diff
@@ -673,23 +685,21 @@ Java_io_glutenproject_columnarbatch_ColumnarBatchJniWrapper_getNumRows(JNIEnv* e
...
   std::shared_ptr<ColumnarBatch> batch = glutenColumnarBatchHolder.lookup(handle);
-  std::shared_ptr<ColumnarBatch> col = glutenColumnarBatchHolder.lookup(colHandle);
-#ifdef DEBUG
-  if (col->GetNumColumns() != 1) {
-    throw GlutenException("Add column should add one col");
-  }
-#endif
-  auto newBatch = batch->addColumn(index, col);
+  int32_t size = env->GetArrayLength(colData);
+  jint* col = env->GetIntArrayElements(colData, JNI_FALSE);
```
zhztheplayer, line 698: We'd call `env->Release...` to release the column after used. Otherwise memory leak takes place.

gluten-data/src/main/scala/org/apache/spark/sql/execution/utils/ExecUtil.scala
```diff
@@ -130,63 +121,23 @@ object ExecUtil {
...
+                data(i) = pid
             }
-          } else {
-            val loaded = ArrowColumnarBatches.ensureLoaded(
-              ArrowBufferAllocators.contextInstance(), cb)
-            computeArrowAddPartitionId(loaded, partitionKeyExtractor)
-          }
+            val newHandle = ColumnarBatchJniWrapper.INSTANCE.addIntColumn(
+              GlutenColumnarBatches.getNativeHandle(cb), 0, "pid", data);
+            (0, GlutenColumnarBatches.create(newHandle))
```
zhztheplayer, line 140: This would usually conduct copy over JNI. Is it something intended?

cpp/core/jni/JniWrapper.cc
jinchengchenghh: Yes, I have called after addIntColumn

Issue #1918, opened 2023-06-12
Guava Cache has race issue (https://github.com/google/guava/issues/2131)


baibaichen:
**Describe the bug**
From https://github.com/google/guava/issues/2131, it recommends using caffine cache instead of guava cache, we meet the same issue  that `LoadingCache.getIfPresent(key) returns null` while `get` already return a vlaue.

PR #1919, opened 2023-06-12
[GLUTEN-1918][Core]Fix: Using Caffeine cache instead of guava cache

baibaichen:
This PR follow ups https://github.com/oap-project/gluten/pull/1873.  

From https://github.com/google/guava/issues/2131, it recommends using caffine cache instead of guava cache, we meet the same issue  that `LoadingCache.getIfPresent(key) returns null` while `get` already return a vlaue.

I also add `log4j(2). properties` for dump `CHBroadcastBuildSideCache`'s debug meessage.

(Fixes: \#1918)

## How was this patch tested?
Using existed UT.

### DIFF
--- a/backends-clickhouse/pom.xml
+++ b/backends-clickhouse/pom.xml
@@ -15,6 +15,10 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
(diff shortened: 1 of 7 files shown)


zhouyuan (approved):
+1

zzcclp (approved):
LGTM

PR #1931, opened 2023-06-13
[CORE] Support vectorize read in vanilla scan

jackylee-ch:
This pr is trying to support read with vectorize in vanilla scan, this is specially used in iceberg reading in our environment.

## How was this patch tested?
added tests.


PHILO-HE:
Isn't vanilla spark's row based reader workable for reading iceberg? We disabled vanilla spark's vectorized reader by intention to avoid perf. overhead, considering vanilla C2R + native R2C will be used for data transition. cc @zhouyuan

jackylee-ch:
> Isn't vanilla spark's row based reader workable for reading iceberg? We disabled vanilla spark's vectorized reader by intention to avoid perf. overhead, considering vanilla C2R + native R2C will be used for data transition. cc @zhouyuan

I think current PR does not conflict with it, we can still continue to recommend users to close the vectorize reader. This is only worked in some scenarios where the vectorize reader cannot be closed, and the job can still run through even if there is a performance regression.

jinchengchenghh (approved):

backends-velox/src/test/scala/io/glutenproject/execution/VeloxStringFunctionsSuite.scala
```diff
@@ -295,9 +295,10 @@ class VeloxStringFunctionsSuite extends WholeStageTransformerSuite {
       s"from $LINEITEM_TABLE limit $LENGTH") { checkOperatorMatch[ProjectExecTransformer] }
     runQueryAndCompare(s"select l_orderkey, like($NULL_STR_COL, '%a%') " +
       s"from $LINEITEM_TABLE limit $LENGTH") { checkOperatorMatch[ProjectExecTransformer] }
-    runQueryAndCompare(s"select l_orderkey, l_comment " +
+    runQueryAndCompare(s"select l_orderkey, like(l_comment, '%a%') " +
```
jackylee-ch, line 298: This test need to be changed as vanilla would remove extra project if datasource support columnar in `RemoveRedundantProjects`

gluten-core/src/main/scala/io/glutenproject/backendsapi/BackendSettingsApi.scala
```diff
@@ -54,7 +55,10 @@ trait BackendSettingsApi {
   // Whether to fallback aggregate at the same time if its child is fallbacked.
   def fallbackAggregateWithChild(): Boolean = false
 
-  def disableVanillaColumnarReaders(): Boolean = false
+  def disableVanillaColumnarReaders(conf: SparkConf): Boolean =
```
PHILO-HE, line 58: Can we remove conf arg?
It looks `GlutenConfig.getConf` can be used to get a conf instance inside the method. Thanks!

PHILO-HE (commented):
Looks great! Just a nit spotted. See comment.
In addition, there are some unit tests excluded with comment "...enableVectorizedReader=true not supported". Can those be included for tests? Thanks!

gluten-core/src/main/scala/io/glutenproject/backendsapi/BackendSettingsApi.scala
jackylee-ch: Hm, `disableVanillaColumnarReaders` is used in `DriverPlugin` while initliazing `SparkContext`, the SQLConf is not initlized at that time.

PHILO-HE: Got it. It makes sense.

PHILO-HE (approved):
Looks great!
@jinchengchenghh, let's merge this version. If you have other review comment, please let me know. I will do a follow-up fix. Thanks!

PR #1946, opened 2023-06-14
[GLUTEN-1945] Support summarizing the supported spark built-in functions

PHILO-HE:
This is a rough summary by testing example query for spark registered functions.
There are some false positive & false negative cases for some functions. For such situation, we need to use a suitable test sql to do the check. Will include this in a follow-up PR.

### DIFF
--- a/gluten-ut/spark32/src/test/scala/org/apache/spark/sql/statistics/SparkFunctionStatistics.scala
+++ b/gluten-ut/spark32/src/test/scala/org/apache/spark/sql/statistics/SparkFunctionStatistics.scala
@@ -84,7 +84,8 @@ class SparkFunctionStatistics extends QueryTest {
     // these functions are registered only for testing, not available for end users.
     // Other functions like current_database is NOT necessarily offloaded to native.
     val ignoreFunctions = FunctionRegistry.expressionsForTimestampNTZSupport.keySet ++
-        Seq("get_fake_app_name", "current_catalog", "current_database")
+        Seq("get_fake_app_name", "current_catalog", "current_database", "spark_partition_id",
+          "current_user", "current_timezone")
     val supportedFunctions = new java.util.ArrayList[String]()
     val unsupportedFunctions = new java.util.ArrayList[String]()
     val needInspectFunctions = new java.util.ArrayList[String]()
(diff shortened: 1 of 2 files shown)


gluten-ut/spark32/src/test/scala/org/apache/spark/sql/statistics/SparkFunctionStatistics.scala
```diff
@@ -0,0 +1,134 @@
...
+          }
+          }
+          if (isSupported && !needInspectFunctions.contains(exprInfo.getName)) {
+            supportedFunctions.add(exprInfo.getName)
+          } else if (!isSupported && !needInspectFunctions.contains(exprInfo.getName)) {
+            unsupportedFunctions.add(exprInfo.getName)
+          }
+      }
+    }
+    print("overall functions: " + (sparkBuiltInFunctions.size - ignoreFunctions.size) + "\n")
```
pan3793, line 129: why not `println`?

PHILO-HE: Fixed. Thanks!

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/statistics/SparkFunctionStatistics.scala
```diff
@@ -0,0 +1,189 @@
...
+              isSupported = false
+              break
+            }
+            break
+          }
+        }
+        if (isSupported && !needInspectFunctions.contains(exprInfo.getName)) {
+          supportedFunctions.add(exprInfo.getName)
+        } else if (!isSupported && !needInspectFunctions.contains(exprInfo.getName)) {
+          unsupportedFunctions.add(exprInfo.getName)
```
rui-mo, line 139: Is there any case meet this condition `!isSupported && needInspectFunctions.contains(exprInfo.getName)`? If not, this clause can just be like `else if (!isSupported)`.

PHILO-HE: Just removed the extra check condition that never goes into. Thanks!

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/statistics/SparkFunctionStatistics.scala
```diff
@@ -0,0 +1,188 @@
...
+              break
+            }
+            break
+          }
+        }
+        if (isSupported && !needInspectFunctions.contains(exprInfo.getName)) {
+          supportedFunctions.add(exprInfo.getName)
+        } else if (!isSupported) {
+          unsupportedFunctions.add(exprInfo.getName)
+        }
```
rui-mo, line 140: Does it make sense to add an unreachable check here for `else`?

PHILO-HE: The unchecked else is actually handled for `needInspectFunctions` in the below. Thanks!

PR #1986, opened 2023-06-16
[GLUTEN-1985] avoid forceShuffledHashJoin when the join condition does not supported by the backend

zheniantoushipashi:
Is your feature request related to a problem or challenge? Please describe what you are trying to do.

In our custom JoinSelectionOverrides rule, if forceShuffledHashJoin is set to true, SortmergeJoin will be converted to ShuffledHashJoin regardless of whether the backend supports it or whether it is fallback. If it is executed in spark in the case of falllback, SortmergeJoin will be converted For ShuffledHashJoin, it will take up more memory, resulting in OOM,

Describe the solution you'd like

avoid forceShuffledHashJoin when the join condition does not supported by the backend

### DIFF
--- a/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCDSParquetSuite.scala
+++ b/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCDSParquetSuite.scala
@@ -22,7 +22,7 @@ import org.apache.spark.SparkConf
 import org.apache.spark.sql.catalyst.expressions.{DynamicPruningExpression, Not}
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.exchange.ReusedExchangeExec
-import org.apache.spark.sql.execution.joins.BroadcastHashJoinExec
+import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, SortMergeJoinExec}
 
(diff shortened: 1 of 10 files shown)


backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCDSParquetSuite.scala
```diff
@@ -172,6 +172,33 @@ class GlutenClickHouseTPCDSParquetSuite extends GlutenClickHouseTPCDSAbstractSui
...
+        | LEFT JOIN store ON ss_store_sk = s_store_sk AND substr(ca_zip,1,5) <> substr(s_zip,1,5)
+        | WHERE d_moy = 11
+        |   AND d_year = 1999
+        | GROUP BY i_brand_id, i_brand, i_manufact_id, i_manufact
+        | ORDER BY ext_price DESC, i_brand, i_brand_id, i_manufact_id, i_manufact
+        | LIMIT 100;
+        |""".stripMargin
+
+    val df = spark.sql(testSql)
+    val sortMergeJoinExecTransFormer = df.queryExecution.executedPlan.collect {
```
zzcclp, line 196: SortMergeJoinExecTransformer is not supported, maybe skip transform SMJ in transform rules.

zheniantoushipashi: done

zzcclp (approved):
LGTM

PR #1987, opened 2023-06-16
[UT] Minor: add jvm xss configuration in gluten-ut to avoid stackoverflow

lwz9103:

add jvm args in compile scala test code to avoid stackoverflow


zhouyuan (approved):
👍

baibaichen (approved):
LGTM

PR #1988, opened 2023-06-16
[VL] Adopt Spark local-cluster run mode in gluten-it

zhztheplayer:
This is to make gluten-it capable to run jobs in Spark local cluster mode.

CI's run mode remains unchanged (local mode) and we may change to use this local cluster mode later some time.


zhztheplayer:
@zhouyuan

zhouyuan (approved):
👍

Issue #1975, opened 2023-06-15
case-insensitive column matching for hive text format


taiyang-li:

taiyang-li:
let's do it !

PR #1991, opened 2023-06-16
[GLUTEN-1975][CH] hive text format case-insensitive column matching  

taiyang-li:
code changes: 
- support column case-insensitive matching for hive text format file
- fix bug when select complex type columns from hive text table: it is expected to fallback, but actually not as expected. 
- fix bug: select count(1) from hive text table returns wrong results
- improve function adjustFileReadStartAndEndPos

(Fixes: \#1975)

## How was this patch tested?

``` sql 
create table if not exists test_text (string_field string,int_field int,long_field long,float_field float,double_field double,short_field short,byte_field byte, bool_field boolean, decimal_field decimal(23, 12), date_field date, array_field array<int>, array_field_with_null array<int>, map_field map<int, long>, map_field_with_null map<int, long>) stored as textfile; 

insert into test_text select
cast(id as String), 
cast(id as int), 
cast(id as long), 
cast(id as float), 
cast(id as double),
cast(id as short),
cast(id as byte),
id % 2 = 0,
cast((id + 0.56) as decimal(23, 12)), 
date(now()),
array(id+1, id+2, id+3),
array(id+1, null, id+3),
map(id+1, id+2, id+3, id+4),
map() from range(200); 

explain formatted 
 select string_field,
        sum(int_field),
        avg(long_field),
        min(float_field),
        max(double_field),
        sum(short_field),
        sum(decimal_field)
 from test_text
 group by string_field
 order by string_field



```

### DIFF
--- a/cpp-ch/local-engine/Storages/SubstraitSource/ReadBufferBuilder.cpp
+++ b/cpp-ch/local-engine/Storages/SubstraitSource/ReadBufferBuilder.cpp
@@ -43,7 +43,7 @@ namespace ErrorCodes
     extern const int CANNOT_OPEN_FILE;
     extern const int UNKNOWN_FILE_SIZE;
     extern const int CANNOT_SEEK_THROUGH_FILE;
-    extern const int CANNOT_CLOSE_FILE;
+    extern const int CANNOT_READ_FROM_FILE_DESCRIPTOR;
 }
 }
 
@@ -156,7 +156,7 @@ public:
                 auto n = hdfsRead(fs, fin, buf, buf_size);
                 if (n < 0)
                     throw DB::Exception(
-                        DB::ErrorCodes::CANNOT_SEEK_THROUGH_FILE,
+                        DB::ErrorCodes::CANNOT_READ_FROM_FILE_DESCRIPTOR,
                         "Fail to read HDFS file: {}, error: {}",
                         file_path,
                         std::string(hdfsGetLastError()));
--- a/cpp-ch/local-engine/Storages/SubstraitSource/ReadBufferBuilder.cpp
+++ b/cpp-ch/local-engine/Storages/SubstraitSource/ReadBufferBuilder.cpp
@@ -83,12 +83,12 @@ public:
     build(const substrait::ReadRel::LocalFiles::FileOrFiles & file_info, const bool & set_read_util_position) override
     {
         Poco::URI file_uri(file_info.uri_file());
-        std::unique_ptr<DB::ReadBuffer> read_buffer;
-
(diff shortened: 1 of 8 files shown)


taiyang-li:
@loneylee @KevinyhZou can you help review this pr, thanks!

KevinyhZou:
It seems that the index disorder problem still exists, it can reproduced by like `select count(*) from test_tbl where id > 0` @taiyang-li

taiyang-li:
> Run Gluten Clickhouse CI

That's another issue different from current one. I'll fix it later in another pr

taiyang-li:
> It seems that the index disorder problem still exists, it can reproduced by like `select count(*) from test_tbl where id > 0` @taiyang-li

fixed now.  see latest commit.

cpp-ch/local-engine/Storages/SubstraitSource/TextFormatFile.h
```diff
@@ -13,6 +13,8 @@ class TextFormatFile : public FormatFile
         DB::ContextPtr context_, const substrait::ReadRel::LocalFiles::FileOrFiles & file_info_, ReadBufferBuilderPtr read_buffer_builder_);
     ~TextFormatFile() override = default;
 
+    bool supportSplit() override { return true; }
+
```
KevinyhZou, line 17: This should also be enabled at JsonFormatFile.

cpp-ch/local-engine/Storages/SubstraitSource/ReadBufferBuilder.cpp
```diff
@@ -118,87 +119,100 @@ class HDFSFileReadBufferBuilder : public ReadBufferBuilder
...
+                        return buf[0] == '\r' ? pos + i + 2 : pos + i + 1;
                     }
-                }
-                if (i == row_delimiter_size)
-                {
-                    char r[1];
-                    // The end of row maybe '\n', '\r\n', or '\n\r'
-                    if (hdfsRead(fs, fin, r, 1) != 0 && r[0] == '\r')
+                    else if (buf[i] == '\r')
                     {
```
KevinyhZou, line 193: \r\n and \n can be handled by just judge whether the last character is \n

taiyang-li: This logic is from skipEndOfLine in src/Processors/Formats/Impl/CSVRowInputFormat.cpp

loneylee:
LGTM

baibaichen (approved):
LGTM

PR #1993, opened 2023-06-16
[GLUTEN-1938][CH] Fix hive reader read empty as null

loneylee:
* Fix the null behavior consistent with spark
* Fix date and datetime may read wrong day

## How was this patch tested?

test by ut


taiyang-li:
LGTM

baibaichen (approved):
LGTM

PR #2000, opened 2023-06-18
[GLUTEN-2103][VL] feat: allow to dump the fallback information for plan validation

yma11:
In customer workloads, fallbacks often happen but can't find the root cause directly with current fallback logs. 
So in this PR, we will propose 2 changes:
1) add more detailed log about fallback reasons with unified format like "Validation failed for *** due to ***" and "native validation failed due to ***" at native side(refer [PR](https://github.com/oap-project/velox/pull/319))
2) refactor it to make it controllable by “spark.gluten.sql.validate.failure.logLevel”. Log will be printed if it's "DEBUG" as default.

The PR is used to print the fallback reason in stderr/stdout when Gluten user disables AQE and calls df.explain(). In this way user needn't to run the whole query on Gluten, just explain() gluten's plan. Gluten will check each operator if it can offload to Velox, if not a formatted info is dumped to driver's log file. Then we can use a simple python script to summarize the fallback reasons.

Fixes: #2103 
## How was this patch tested?

CI

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -50,18 +50,31 @@ object BackendSettings extends BackendSettingsApi {
     // Validate if all types are supported.
     def validateTypes: Boolean = {
       // Collect unsupported types.
-      fields.map(_.dataType).collect {
-        case _: ArrayType =>
-        case mapType: MapType if mapType.keyType.isInstanceOf[StructType] =>
-        // Parquet scan of nested map with struct as key type is not supported in Velox.
+      val unsupportedDataTypes = fields.map(_.dataType).collect {
         case _: ByteType =>
-      }.isEmpty
+        case _: ArrayType =>
+        case _: MapType if mapType.keyType.isInstanceOf[StructType] => "StructType in MapType"
+        // Parquet scan of nested map with struct as key type is not supported in Velox.
+      }
+      for (unsupportedDataType <- unsupportedDataTypes) {
+        // scalastyle:off println
+        println(
+          s"Validation failed for ${this.getClass.toString}" +
+            s"due to Not supported: data type $unsupportedDataType in file schema. ")
+        // scalastyle:on println
+      }
+      unsupportedDataTypes.isEmpty
     }
 
     def validateFilePath: Boolean = {
       // Fallback to vanilla spark when the input path
       // does not contain the partition info.
       if (partTable && !paths.forall(_.contains("="))) {
+        // scalastyle:off println
+        println(
+          s"Validation failed for ${this.getClass.toString}" +
+            s"due to Not supported: input path doesn't contain split info. ")
+        // scalastyle:on println
         return false
       }
       true
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -60,7 +60,7 @@ object BackendSettings extends BackendSettingsApi {
         // scalastyle:off println
         println(
           s"Validation failed for ${this.getClass.toString}" +
-            s"due to Not supported: data type $unsupportedDataType. in file schema. ")
+            s" due to: data type $unsupportedDataType. in file schema. ")
         // scalastyle:on println
       }
       unsupportedDataTypes.isEmpty
@@ -73,7 +73,7 @@ object BackendSettings extends BackendSettingsApi {
         // scalastyle:off println
         println(
           s"Validation failed for ${this.getClass.toString}" +
-            s"due to Not supported: input path doesn't contain split info. ")
+            s"due to: input path doesn't contain split info. ")
         // scalastyle:on println
         return false
       }
@@ -91,7 +91,7 @@ object BackendSettings extends BackendSettingsApi {
           // scalastyle:off println
           println(
             s"Validation failed for ${this.getClass.toString}" +
-              s"due to Not supported: data type $unsupportedDataType. in file schema. ")
+              s" due to: data type $unsupportedDataType. in file schema. ")
           // scalastyle:on println
         }
         unsupportedDataTypes.isEmpty
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -51,8 +51,8 @@ object BackendSettings extends BackendSettingsApi {
     def validateTypes: Boolean = {
       // Collect unsupported types.
       val unsupportedDataTypes = fields.map(_.dataType).collect {
-        case _: ByteType =>
-        case _: ArrayType =>
+        case _: ByteType => "ByteType"
+        case _: ArrayType => "ArrayType"
         case _: MapType if mapType.keyType.isInstanceOf[StructType] => "StructType in MapType"
         // Parquet scan of nested map with struct as key type is not supported in Velox.
       }
@@ -60,7 +60,7 @@ object BackendSettings extends BackendSettingsApi {
         // scalastyle:off println
         println(
           s"Validation failed for ${this.getClass.toString}" +
-            s"due to Not supported: data type $unsupportedDataType in file schema. ")
+            s"due to Not supported: data type $unsupportedDataType. in file schema. ")
         // scalastyle:on println
       }
       unsupportedDataTypes.isEmpty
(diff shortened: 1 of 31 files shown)


yma11:
@zhouyuan @FelixYBW @PHILO-HE Please take a look at the change.

Yohahaha (commented):
Some questions here:
1, Could we use logxxx instead of println? println is uncontrollable.
2, Does this PR make validation log and code more clearly? I see lots of extra codes added.
3, Do we have a summary about which config will control all validation log level, maybe a clearly PR description or docs would make sense.
4, Consider add more context info to log, such as failure due to window/aggregate/scalar function not supported, or due to type not supported.

yma11:
Thanks for comments.
> Some questions here: 1, Could we use logxxx instead of println? println is uncontrollable. 

Updated.
> 2, Does this PR make validation log and code more clearly? I see lots of extra codes added.

I think so. Current we can control most of the log print and the fallback info is consolidated together for both native and java side at operator level.
> 3, Do we have a summary about which config will control all validation log level, maybe a clearly PR description or docs would make sense.

Description added.
> 4, Consider add more context info to log, such as failure due to window/aggregate/scalar function not supported, or due to type not supported.

This is handled at native side, refer [PR](https://github.com/oap-project/velox/pull/316/files). I looked through the call stack and make the log as precise as possible.

Yohahaha:
Thanks! Looks great!

FelixYBW:
@yma11  can we change log level to info? debug printed too much information.

FelixYBW:
@yma11 can you also create a document and show how it works using df.explain()?

yma11:
> @yma11 can we change log level to info? debug printed too much information.

This is a dedicated configuration for validation log and only fallback logs printed when it's set to `DEBUG`. Less logs will be printed when it's set to `INFO` but by default, it's set to `DEBUG`.

yma11:
> @yma11 can you also create a document and show how it works using df.explain()?

Add relate [info](https://github.com/oap-project/gluten/pull/2000/files#diff-93367912d5dc70f8cac4b20a134247e10f3764bfc77f836cc72d18e5ce1f80f3) in Velox.md

PHILO-HE:
Run Gluten Clickhouse CI

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -50,18 +50,31 @@ object BackendSettings extends BackendSettingsApi {
...
+        case _: ByteType => "ByteType"
+        case _: ArrayType => "ArrayType"
+        case mapType: MapType if mapType.keyType.isInstanceOf[StructType] => "StructType as Key in MapType"
         // Parquet scan of nested map with struct as key type is not supported in Velox.
-        case _: ByteType =>
-      }.isEmpty
+      }
+      for (unsupportedDataType <- unsupportedDataTypes) {
+        // scalastyle:off println
+        println(
```
PHILO-HE, line 61: It would be better to print with commonly used log.

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -166,23 +166,36 @@ case class FilterExecTransformer(condition: Expression, child: SparkPlan)
         substraitContext, condition, child.output, operatorId, null, validation = true)
     } catch {
       case e: Throwable =>
-        logValidateFailure(
-          s"Validation failed for ${this.getClass.toString} due to ${e.getMessage}", e)
+        this.appendValidateLog(s"Validation failed for ${this.getClass.toString}" +
```
PHILO-HE, line 169: It looks the constructing of arg for `appendValidateLog` can be simplified since many are in similar pattern,
e.g., 
```
def appendValidateLog(failureMsg) {
 log = "${this.getClass.toString}: validation failed, reason: failureMsg"
 validateLog = validateLog :+ log
}
```

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
```diff
@@ -54,7 +54,11 @@ trait TransformSupport extends SparkPlan with LogLevelUtil {
 
   lazy val validateFailureLogLevel = GlutenConfig.getConf.validateFailureLogLevel
   lazy val printStackOnValidateFailure = GlutenConfig.getConf.printStackOnValidateFailure
+  var validateLog: Vector[String] = Vector()
 
+  def appendValidateLog (log: String): Unit = {
```
PHILO-HE, line 59: I think this new method should be added in `GlutenPlan` to cover all gluten plans. `TransformSupport` only covers plans that can be put into wholestage transformer.

PHILO-HE (commented):
Looks great! Just few suggestions for improvement.

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
yma11: Not all of them are in this case and we use keyword `due to` for both java and native side.

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
yma11: Agree as I found there are nodes which are not inherited from `TransfromSupport`. It's good to add the interface in `GlutenPlan` as you said. Let me update.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
yma11: That will make this method unclear and I think it's okay to do print in this single case.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -50,18 +50,31 @@ object BackendSettings extends BackendSettingsApi {
...
+          s"Validation failed for ${this.getClass.toString}" +
+            s"due to Not supported: data type $unsupportedDataType. in file schema. ")
```
PHILO-HE, line 63: Should leave a space before "due to...", same to the below changes.

FelixYBW: @yma11 make sure all output have the same format, then we can easily extract info like fallback_reason = l.split(":")[-1]

yma11: > Should leave a space before "due to...", same to the below changes. Please do an overall check.

Already noticed this and updated. Now still need do a rebase.

yma11: > @yma11 make sure all output have the same format, then we can easily extract info like fallback_reason = l.split(":")[-1]

When we need to have a `split`? Each log is not long and I think we can just use the keyword `due to` to filter them out.

zhouyuan:
This is a big patch... we could try to merge the core framework and follow up with small fixes

-yuan

ulysses-you:
I'm wondering if we can leverage Spark event framework that post a custom event which contains fallback information, then we can do some analysis. In this way, we can even inject a tab into Spark ui to show why the query execution can not be offloaded.

ulysses-you:
If you are interested in, I can do some followup after this pr merged.

FelixYBW:
> If you are interested in, I can do some followup after this pr merged.

Good proposal! Can you follow up?

zhouyuan:
@ulysses-you  cool, we are also looking for better way to address this issue, can you please start one feature issue and write some quick design there?

-yuan

ulysses-you:
@FelixYBW @zhouyuan  sure, I'm going to do this in these days

yma11:
> I'm wondering if we can leverage Spark event framework that post a custom event which contains fallback information, then we can do some analysis. In this way, we can even inject a tab into Spark ui to show why the query execution can not be offloaded.

Once this PR merged, each gluten plan node will have a property `validateLog` . You can implement the custom event based on analysis of it. You may need better log format for easy analysis and we can sync on that late.

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
yma11: @PHILO-HE This is a API in `trait GlutenPlan` and it's not quite necessary to add another similar API in it. But I have consolidate the ones in `TransformHintRule` and seems cleaner. Please take a further look.

zhouyuan (approved):

zhouyuan:
@yma11  thank you for for your great work on implementing this!

-yuan

PR #2013, opened 2023-06-19
[VL] Maintain a list of overridden classes against Gluten's dependencies

zhztheplayer:
Use the list to track on the classes that were from dependencies and then overridden by Gluten.


zhztheplayer:
@JkSelf @rui-mo

backends-velox/src/main/scala/io/glutenproject/execution/VeloxFilterExecTransformer.scala
```diff
@@ -32,7 +32,7 @@ import org.apache.spark.sql.catalyst.expressions.{And, Attribute, Expression}
 import org.apache.spark.sql.execution.SparkPlan
 
 
-case class FilterExecTransformer(condition: Expression, child: SparkPlan)
+case class VeloxFilterExecTransformer(condition: Expression, child: SparkPlan)
```
rui-mo, line 35: Why do we need this change? Since this class in in `backends-velox` module, the scope could be apparant.

zhztheplayer: A class with same name is in `gluten-core` and accessed by `backends-clickhouse`.

rui-mo: Maybe we can merge that `FilterExecTransformer` into `FilterExecBaseTransformer`, and use `FilterExecBaseTransformer` in CH backend. Does that make sense? I can do this refactor if needed.

zhztheplayer: Making sense to me, so let's use a separate patch for that change?

This patch is simply to add extra restrictions for future development. It's more or less unrelated to specific functionalities.

rui-mo: OK, sure. Will open PR to do that after this change.

rui-mo (approved):

package/pom.xml
```diff
@@ -112,23 +134,94 @@
...
+            </configuration>
+          </execution>
+          <execution>
+            <id>enforce-ban-duplicate-classes</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <banDuplicateClasses>
```
JkSelf, line 185: We no need to ensure the gluten jar pre-load with this change, right?

zhztheplayer: Correct, this change only takes effect at Maven compile time.

JkSelf (approved):

JkSelf (approved):

package/pom.xml
```diff
@@ -128,6 +150,78 @@
...
+                    <ignoreClass>org.apache.commons.logging.*</ignoreClass>
+                    <!-- hive -->
+                    <ignoreClass>org.apache.hadoop.hive.*</ignoreClass>
+                    <!-- celeborn -->
+                    <ignoreClass>org.apache.spark.sql.execution.columnar.ByteBufferHelper</ignoreClass>
+                    <ignoreClass>org.apache.spark.sql.execution.columnar.ByteBufferHelper$</ignoreClass>
+                    <!-- The overridden class list by Gluten. Carefully add entries to this list only when you knew exactly what is going to happen -->
+                    <ignoreClass>org.apache.spark.rdd.EmptyRDD</ignoreClass>
+                    <ignoreClass>org.apache.spark.sql.hive.execution.HiveFileFormat</ignoreClass>
+                    <ignoreClass>org.apache.spark.sql.hive.execution.HiveFileFormat$$$$anon$1</ignoreClass>
```
Yohahaha, line 201: I download spark-3.2.2-bin-hadoop2.7.tgz find this class may seem different, could you give more notes on how to add entries? @zhztheplayer 

```
[root@host]# jar -tf spark-hive_2.12-3.2.2.jar | grep HiveFile
org/apache/spark/sql/hive/execution/SaveAsHiveFile.class
org/apache/spark/sql/hive/execution/HiveFileFormat.class
org/apache/spark/sql/hive/execution/HiveFileFormat$$anon$1.class
```

zhztheplayer: It was said that you don't have to do anything but just to use spark-submit / spark-shell `--jar` to let gluten jar to be loaded before other jars.

The list is just an exclusion list to let use know how many classes gluten overrides from vanilla Spark.

PR #2034, opened 2023-06-21
[VL] StdMemoryAllocator: reallocateAligned doesn't use aligned_alloc

zhztheplayer:


zhztheplayer:
Thanks @marin-ma for catching this.

@zuochunwei Not sure if this patch can possibly fix the crash?

zuochunwei:
> Thanks @marin-ma for catching this.
> 
> @zuochunwei Not sure if this patch can possibly fix the crash?

I‘m not sure!
but this is a reasonable modification for me, so this PR should be merged.

zuochunwei (approved):

Issue #2051, opened 2023-06-23
[VL] Make ColumnarBatchSerializer supports relocation


WangGuangxin:
Spark supports fetch the contiguous shuffle blocks in batch, which is enabled by default (by conf `spark.sql.adaptive.fetchShuffleBlocksInBatch`).  This feature has a big performance improvement in our production.

However, currently, since ColumnarBatchSerializer's `supportsRelocationOfSerializedObjects` return false, so that this feature cann't take effect. 
In fact, the arrow serialization does support reloation if we don't write schema (which is default to true) and don't write EOS (which is an optional in arrow rpc serialization format)

https://wesm.github.io/arrow-site-test/format/IPC.html#streaming-format
<img width="719" alt="image" src="https://github.com/oap-project/gluten/assets/1312321/9018b190-ccfc-407b-8e5d-f58540b43a12">

https://github.com/apache/arrow/blob/maint-12.0.1/cpp/src/arrow/ipc/message.cc#L496

PR #2052, opened 2023-06-23
[GLUTEN-2051] Make ColumnarBatchSerializer supports relocation so that continuous shuffle block fetching can be enabled

WangGuangxin:
Spark supports fetch the contiguous shuffle blocks in batch, which is enabled by default (by conf `spark.sql.adaptive.fetchShuffleBlocksInBatch`).  This feature has a big performance improvement in our production.

However, currently, since ColumnarBatchSerializer's `supportsRelocationOfSerializedObjects` return false, so that this feature cann't take effect. 
In fact, the arrow serialization does support reloation if we don't write schema (which is default to true) and don't write EOS (which is an optional in arrow rpc serialization format)

https://wesm.github.io/arrow-site-test/format/IPC.html#streaming-format
image

(Fixes: \#2051)

## How was this patch tested?
Manually test using our internal query

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -754,6 +754,7 @@ JNIEXPORT jlong JNICALL Java_io_glutenproject_vectorized_ShuffleWriterJniWrapper
     jboolean preferEvict,
     jlong allocatorId,
     jboolean writeSchema,
+    jboolean writeEOS,
     jlong firstBatchHandle,
     jlong taskAttemptId,
     jint pushBufferMaxSize,
@@ -816,6 +817,7 @@ JNIEXPORT jlong JNICALL Java_io_glutenproject_vectorized_ShuffleWriterJniWrapper
     }
 
     shuffleWriterOptions.write_schema = writeSchema;
+    shuffleWriterOptions.write_eos = writeEOS;
     shuffleWriterOptions.prefer_evict = preferEvict;
 
     if (numSubDirs > 0) {
--- a/cpp/core/shuffle/LocalPartitionWriter.cc
+++ b/cpp/core/shuffle/LocalPartitionWriter.cc
@@ -96,7 +96,9 @@ class PreferEvictPartitionWriter::LocalPartitionWriterInstance {
     }
 
     RETURN_NOT_OK(writeRecordBatchPayload(dataFileOs.get()));
-    RETURN_NOT_OK(writeEos(dataFileOs.get()));
+    if (shuffleWriter_->options().write_eos) {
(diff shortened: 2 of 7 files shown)


zhouyuan:
@WangGuangxin good finding! Curious how the performance gain look like? The marker is 4 bytes only if I i understand this correctly

-yuan

WangGuangxin:
> spark.sql.adaptive.fetchShuffleBlocksInBatch

Hi, the benefits are not comes from the 4bytes savings, but make the Spark AQE feature ["batch fetch shuffle blocks"](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L723) take effect.
You can refer the feature by the initial commit https://github.com/apache/spark/pull/26040 

In short, when Spark AQE do coalesce partitions, it can fetch continuous blocks in batch, instead of fetch blocks one by one.  But it has some preconditions, one is the serialization of shuffle must support the so called relocation https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/shuffle/BlockStoreShuffleReader.scala#L48C48-L48C85.

Currently, the `ColumnarBatchSerializer` doesn't support relocation.  In fact, We can make some simple modifications to  make it support relocation by not writing EOS after each shuffle blocks.

zhouyuan:
> > spark.sql.adaptive.fetchShuffleBlocksInBatch
> 
> @zhouyuan Hi, the benefits are not comes from the 4bytes savings, but make the Spark AQE feature ["batch fetch shuffle blocks"](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L723) take effect. You can refer the feature by the initial commit [apache/spark#26040](https://github.com/apache/spark/pull/26040)
> 
> In short, when Spark AQE do coalesce partitions, it can fetch continuous blocks in batch, instead of fetch blocks one by one. But it has some preconditions, one is the serialization of shuffle must support the so called relocation https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/shuffle/BlockStoreShuffleReader.scala#L48C48-L48C85.
> 
> Currently, the `ColumnarBatchSerializer` doesn't support relocation. In fact, We can make some simple modifications to make it support relocation by not writing EOS after each shuffle blocks.

@WangGuangxin thanks for detailed explanation, got the idea now

zhouyuan (approved):
👍

CLTFOREVER:
@WangGuangxin @zhouyuan we have meet the same problem，if the pr is ready，can we merge it？

zhouyuan (approved):
👍  works for me

PHILO-HE (approved):

PR #2069, opened 2023-06-26
[VL] Add CompositeColumnarBatch

zhztheplayer:
CompositeCoumnarBatch will be a columnar batch implementation to hold refs to multiple child columnar batches, and will be treated as a horizontal combination of the child batches. By doing this we can:

* In Velox backend, lazily convert child batches to Velox batches, when a batch is composed by child batches
* Gain immutability of the ColumnarBatch API

The patch also includes some minor code cleanups to the batch API.


cpp/velox/memory/VeloxColumnarBatch.cc
```diff
@@ -103,15 +72,35 @@ velox::RowVectorPtr VeloxColumnarBatch::getFlattenedRowVector() {
...
-  if (cb->getType() != "velox") {
-    auto vp = velox::importFromArrowAsOwner(*cb->exportArrowSchema(), *cb->exportArrowArray(), pool.get());
-    return std::dynamic_pointer_cast<velox::RowVector>(vp);
-  } else {
-    return std::dynamic_pointer_cast<VeloxColumnarBatch>(cb)->getRowVector();
+  if (cb->getType() == "velox") {
+    return std::dynamic_pointer_cast<VeloxColumnarBatch>(cb);
+  }
+  if (cb->getType() == "composite") {
+    auto composite = std::dynamic_pointer_cast<gluten::CompositeColumnarBatch>(cb);
```
jinchengchenghh, line 82: composite batch is only used in shuffle split range partition compute pid, after remove Arrow PR, this API will be deleted, we don't need to implement here

zhztheplayer: No need to worry about that, we can either remove or it or not once we think it's no longer used. The code is independent enough to other parts.

zhztheplayer: The new API 'addIntColumn' can be optimized out using `CompositeColumnarBatch` too. 

Composite can mean Velox+Velox as well as Velox+Arrow. After removing Arrow we can make PID column as a single column Velox batch.

jinchengchenghh (approved):

zhejiangxiaomai (dismissed):

jinchengchenghh (approved):

zhejiangxiaomai (approved):

PR #2081, opened 2023-06-26
[VL] Forbid unsupported write using velox

ulysses-you:

This pr does two things:
- Forbid velox write with ctas. It can be supported in Spark3.4
- Forbid velox dynamic partition and bucket write. As mentioned at https://github.com/oap-project/gluten/issues/2042, it would fail if we use dynamic partition write with velox. There is a lot of work need to be done. This pr just throw exception in advance when contains partition or bucket.

## How was this patch tested?

add test

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenColumnarRules.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenColumnarRules.scala
@@ -26,14 +26,10 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.Attribute
 import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, OrderPreservingUnaryNode}
 import org.apache.spark.sql.catalyst.rules.Rule
-import org.apache.spark.sql.catalyst.util.{ArrayData, MapData}
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec
-import org.apache.spark.sql.execution.command.{DataWritingCommand, DataWritingCommandExec}
+import org.apache.spark.sql.execution.command.{CreateDataSourceTableAsSelectCommand, DataWritingCommand, DataWritingCommandExec}
 import org.apache.spark.sql.hive.execution.InsertIntoHiveDirCommand
-import org.apache.spark.sql.types.{DataType, Decimal}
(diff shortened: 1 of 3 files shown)


ulysses-you:
cc @JkSelf @zhouyuan thank you

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenColumnarRules.scala
```diff
@@ -87,10 +83,26 @@ object GlutenColumnarRules {
...
+  // TODO: support ctas in Spark3.4, see https://github.com/apache/spark/pull/39220
+  // TODO: support dynamic partition and bucket write
+  //  1. pull out `Empty2Null` and required ordering to `WriteFilesExec`, see Spark3.4 `V1Writes`
+  //  2. support detect partition value, partition path, bucket value, bucket path at native side,
+  //     see `BaseDynamicPartitionDataWriter`
   def isGlutenInsertInto(cmd: DataWritingCommand): Boolean = {
     cmd match {
-      case command: InsertIntoHadoopFsRelationCommand =>
-        command.fileFormat.isInstanceOf[GlutenParquetFileFormat]
+      case command: CreateDataSourceTableAsSelectCommand =>
```
JkSelf, line 93: Can we fallback these cases to vanilla spark?

ulysses-you: I'm not sure it's a clear behavoir to user. If people specify format to `velox` then they want a velox format, so I think it should be better just fast fail if we do not support. And in future, we plan to add rule to convert parquet format to velox if we support, then things would go to complex if we support both velox -> parquet and parquet -> velox ...

JkSelf: Ok. Can you help to update the doc [here ](https://github.com/oap-project/gluten/blob/main/docs/VeloxNotSupport.md#parquet-write)to add the unsupported command?

ulysses-you: sure, added docs. I refactor the parquet write docs to `parquet write` and `velox parquet write`. The behavior of `velox parquet write` should no affect Spark user, it only affects who specify format to `velox`. So far, `parquet write` only contains hive file format which has been overwritten by velox parquet write.

JkSelf (approved):

PR #2087, opened 2023-06-27
[VL] skip the calculation of the size for NULL String values

yimin-yang:
When calculating `lengths_` in the function `VeloxColumnarToRowConverter::init()`, if there is a NULL value in a StringType column, `strViews[rowIdx].size()` might return an unexpected and potentially large value. This can lead to a "Negative buffer resize" error.
image

In the function `VeloxColumnarToRowConverter::write`, if there is a NULL value in a StringType column, no additional memory is required for it. Therefore, we can simply skip the calculation of the size for NULL String values.


zuochunwei:
LGTM

zuochunwei (approved):

PR #2116, opened 2023-06-28
[VL] Add API to provide different underlying memory allocator implementations at runtime

zhztheplayer:
This also relocates the Java side allocator codes from `gluten-core` to `backend-clickhouse` and `gluten-data`. @zzcclp


zhztheplayer:
@marin-ma As per discussion, it's straightforward to implement listenable Arrow pools by inheriting the pool from `gluten::ArrowMemoryPool`, please try doing that in another PR.

marin-ma (approved):

cpp/core/jni/JniWrapper.cc
```diff
@@ -159,7 +154,7 @@ class JavaInputStreamAdaptor final : public arrow::io::InputStream {
 
   arrow::Result<std::shared_ptr<arrow::Buffer>> Read(int64_t nbytes) override {
     GLUTEN_ASSIGN_OR_THROW(auto buffer, arrow::AllocateResizableBuffer(nbytes, defaultArrowMemoryPool().get()))
```
marin-ma, line 156: Should we use ArrowMemoryPool with listenable allocator here?

zhztheplayer: Agreed. The explicit calls to default Velox/Arrow pools should be finally reviewed and optimized from our code base.

zzcclp (approved):
LGTM

zzcclp (approved):
LGTM

Issue #2095, opened 2023-06-27
[CH] cast(string as binary) return incorrect result


exmy:
**Describe the bug**

`select cast("ABC" as binary)`

gluten + ch return: CBA

spark return: ABC

PR #2119, opened 2023-06-28
[GLUTEN-2095][CH] Fix the incorrect result of  cast(string as binary)

exmy:

`cast(xxx as binary)` is implemented to `reinterpretAsStringSpark` function in CH backend, `xxx` here is either StringType or IntegerType. 
When `xxx` is IntegerType, we need output reinterpreted bytes in big-endian order to align with spark. However, when it is  StringType, this step is not necessary.

(Fixes: \#2095)

## How was this patch tested?

unit tests


taiyang-li (approved):
LGTM

liuneng1994 (approved):
LGTM

Issue #2125, opened 2023-06-29
[VL] Decouple S3 endpoint/ssl.enabled/path-style-access config from AK/SK


leesf:
**Describe the bug**
endpoint/ssl.enabled/path-style-access should be a common config and should not be coupled with ak/sk config

**To Reproduce**

**Expected behavior**

**Additional context**

PR #2126, opened 2023-06-29
[GLUTEN-2125][VL] Decouple S3 endpoint/ssl.enabled/path-style-access config from AK/SK

leesf:

Decouple S3 endpoint/ssl.enabled/path-style-access config from AK/SK since they are common configs.

(Fixes: \#2125)

## How was this patch tested?

manual tests


zhouyuan (approved):
👍  verified locally

PR #2130, opened 2023-06-29
[CORE] Auto generate libhdfs3 conf file from spark hadoop configuration

wForget:

Automatically generate libhdfs3 configuration file from spark hadoop configuration.

After this pr, we only need to add the following configuration when using hdfs, and the libhdfs3 configuration file will be automatically generated in the executor work dir.

```
spark.executorEnv.LIBHDFS3_CONF=hadoop-test.xml;
spark.gluten.generate.libhdfs3.conf.auto=true
```
​
## How was this patch tested?

successfully read hive table of hdfs path and the hadoop-test.xml file is generated in the executor work dir

```
23/06/29 13:03:11 INFO GlutenExecutorPlugin: Writing Hadoop configuration to hadoop-test.xml
```

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/GlutenPlugin.scala
@@ -32,8 +32,9 @@ import org.apache.spark.rpc.{GlutenDriverEndpoint, GlutenExecutorEndpoint}
 import org.apache.spark.sql.SparkSessionExtensions
 import org.apache.spark.sql.internal.StaticSQLConf
 import org.apache.spark.sql.utils.ExpressionUtil
-import org.apache.spark.util.SparkResourcesUtil
+import org.apache.spark.util.{SparkResourcesUtil, SparkUtil}
 
+import java.io.{File, FileOutputStream}
 import java.util
 import java.util.{Collections, Objects}
 import scala.language.implicitConversions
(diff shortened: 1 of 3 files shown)


Yohahaha:
Could you describe more about how to use this new config and what case we should use?

Yohahaha:
> libhdfs3 configuration file will be automatically generated in the executor work dir.

Generated from what content?

wForget:
> Could you describe more about how to use this new config and what case we should use?

When running in yarn cluster mode, we need to use `--files hdfs-client.xml` to copy hadoop conf to each executor node. This PR allows automatic generation of hadoop configuration files from `SparkHadoopUtil#newConfiguration`, and also enables the hadoop configuration specified through  `spark.hadoop.*` to take effect for gluten.

PHILO-HE:
@wForget, thanks so much for your proposal!
Here, we still need a config file (specified by `LIBHDFS3_CONF`) to keep these config properties for libhdfs3 to use.
I am thinking whether it is better to directly pass these hdfs config properties to native side. It needs some velox code changes, i.e., calling `hdfsBuilderConfSetStr` on `hdfsBuilder` (see [here](https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp#L33)) to do the setting. How do you think?

wForget:
> @wForget, thanks so much for your proposal! Here, we still need a config file (specified by `LIBHDFS3_CONF`) to keep these config properties for libhdfs3 to use. I am thinking whether it is better to directly pass these hdfs config properties to native side. It needs some velox code changes, i.e., calling `hdfsBuilderConfSetStr` on `hdfsBuilder` (see [here](https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp#L33)) to do the setting. How do you think?

Yes, it is better to pass hdfs conf through the native method. I will close this PR first and try to implement it through native methods later.

PR #2151, opened 2023-06-30
[GLUTEN-2360][CORE] feat: adding more fallback policies

zhouyuan:

This patch adds more fallback policies to run the query if there's one fallback:

```
operator: fallback per operator if not supported
stage: fallback per stage if number of unsupported operators > threashold
query: fallback whole query if there are more than one operators not supported

```
Fixes: #2360 

Signed-off-by: Yuan Zhou <yuan.zhou@intel.com>
## How was this patch tested?

UT


baibaichen:
@zhouyuan 
 Why non-AQE？ I think it is better to apply to both.

zhouyuan:
> @zhouyuan Why non-AQE？ I think it is better to apply to both.

yes, let me try to combine the code on AQE and non-AQE code path

zuochunwei:
good

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -752,6 +756,27 @@ case class ColumnarOverrideRules(session: SparkSession)
...
+          fallbacks = fallbacks + 1
+        // Possible fallback for leaf node.
+        case leafPlan: LeafExecNode if !leafPlan.isInstanceOf[GlutenPlan] =>
+          fallbacks = fallbacks + 1
+        case _ =>
+      }
+      plan.children.map(p => countFallback(p))
+    }
+    countFallback(plan)
+    fallbacks > 1
```
zzcclp, line 777: can we consider to use `wholeStageFallbackThreshold` as threshold too for query policy ?

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -779,6 +781,16 @@ object GlutenConfig {
       .longConf
       .createWithDefault(100 * 1000 * 1000)
 
+  val COLUMNAR_FALLBACK_POLICY =
+    buildConf("spark.gluten.sql.columnar.fallback.policy")
+      .internal()
+      .doc("The fallback policy in gluten. By default, 'operator' means fallback each " +
+        "operator, 'stage' means fallback each stage, 'query' means fallback query")
```
zzcclp, line 788: stage fallback only for AQE on, right ?

zhouyuan: right, the current impl replies AQE to check if it's fully supported

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhouyuan: yes, let me try to add another configurable value for this

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -752,6 +758,30 @@ case class ColumnarOverrideRules(session: SparkSession)
...
+          fallbacks = fallbacks + 1
+        // Possible fallback for leaf node.
+        case leafPlan: LeafExecNode if !leafPlan.isInstanceOf[GlutenPlan] =>
+          fallbacks = fallbacks + 1
+        case _ =>
+      }
+      plan.children.map(p => countFallback(p))
+    }
+    countFallback(plan)
+    fallbacks > queryFallbackThreshold
```
zzcclp, line 782: use `>=`

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/execution/FallbackStrategiesSuite.scala
```diff
@@ -25,8 +25,25 @@ import org.apache.spark.sql.GlutenSQLTestsTrait
 
 class FallbackStrategiesSuite extends GlutenSQLTestsTrait {
 
+  test("Fall back the whole query if one unsupported") {
```
zzcclp, line 28: add a ut for query fallback threshold

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -793,7 +823,10 @@ case class ColumnarOverrideRules(session: SparkSession)
 
   override def postColumnarTransitions: Rule[SparkPlan] = plan => PhysicalPlanSelector.
     maybe(session, plan) {
-      if (isAdaptiveContext && fallbackWholeStage(plan)) {
+      if (fallbackPolicy == "query"&& !isAdaptiveContext  && fallbackWholeQuery(plan)) {
```
zzcclp, line 826: add a blank between `fallbackPolicy == "query"` and `&&`

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -752,6 +758,30 @@ case class ColumnarOverrideRules(session: SparkSession)
     fallbacks >= wholeStageFallbackThreshold
   }
 
+  def fallbackWholeQuery(plan: SparkPlan): Boolean = {
```
zzcclp, line 761: the logical of this function is the same as fallbackWholeStage, can it be removed ?

zhouyuan: indeed, will merged in to one function

zzcclp (approved):
LGTM

baibaichen:
@zhouyuan I don't understand the codes before trying it, please merge it.

Issue #2152, opened 2023-06-30
[CORE] Skip local sort if child output ordering satisfies topK


ulysses-you:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
For example, the local topK before shuffle does not need a extra local sort since the child output ordering already satisfies it.
```
SELECT * FROM t1 JOIN t2 ON t1.c1 = t2.c2 ORDER BY t1.c1 LIMIT 5;
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2153, opened 2023-06-30
[GLUTEN-2152][CORE] Skip local sort if child output ordering satisfies topk

ulysses-you:

(Fixes: \#2152)

## How was this patch tested?

Pass existed topk test

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/TakeOrderedAndProjectExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/TakeOrderedAndProjectExecTransformer.scala
@@ -59,7 +59,6 @@ case class TakeOrderedAndProjectExecTransformer(limit: Int,
     throw new UnsupportedOperationException(s"This operator doesn't support doExecute().")
   }
 
-
   protected override def doExecuteColumnar(): RDD[ColumnarBatch] = {
     // No real computation here
     val childRDD = child.executeColumnar()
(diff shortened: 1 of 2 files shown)


rui-mo:
@lgbo-ustc Would you like to take a review?

lgbo-ustc:
LGTM

lgbo-ustc:
LGTM
BTW, is there any unit tests to cover this ?

ulysses-you:
@lgbo-ustc due to it changes plan inside transformer, it's a bit hard to plan validation.  I have tested this locally:

```
val row = 10 * 1000
val df1 = spark.range(0, row, 1, 2).selectExpr("id % 3 as c1")
val df2 = spark.range(0, row, 1, 2).selectExpr("id % 3 as c2")
df1.join(df2, col("c1") === col("c2")).orderBy(col("c1")).limit(100).write.format("noop").mode("Overwrite").save()

// before
7.481942 s
// after 
9.447648 s
```

gluten-core/src/main/scala/io/glutenproject/execution/TakeOrderedAndProjectExecTransformer.scala
```diff
@@ -68,32 +67,40 @@ case class TakeOrderedAndProjectExecTransformer(limit: Int,
     if (childRDDPartsNum == 0) {
       sparkContext.parallelize(Seq.empty, 1)
     } else {
+      val orderingSatisfies = SortOrder.orderingSatisfies(child.outputOrdering, sortOrder)
+      def withLocalSort(sparkPlan: SparkPlan): SparkPlan = {
+        if (orderingSatisfies) {
+          sparkPlan
+        } else {
+          SortExecTransformer(sortOrder, false, sparkPlan)
```
zhztheplayer, line 75: Just out of curiosity (lack of some backgrounds), does the code mean we are not using optimized algorithms for the top-k calculation here? I noticed that vanilla Spark calls Guava's top-k which is ideally O(n).

@lgbo-ustc

ulysses-you: It seems velox support topN. I'm not sure why we did not support transform TakeOrderedAndProjectExec to TopN before. This pr itself just removes a local sort based on previous code.

liuneng1994:
The optimization on one partition also failed on tpcds q67, should we remove directly

rui-mo:
> The optimization on one partition also failed on tpcds q67, should we remove directly

@lgbo-ustc @liuneng1994 Can we merge this PR? Is there any remaining issue on CH backend?

zzcclp:
I will take a review ASAP, if ok I will merge this.
BTW, @ulysses-you please help to open tpcds q67 test by this way:
```
\\ in GlutenClickHouseTPCDSAbstractSuite.scala line 98:
remove "q67"
```

zzcclp:
> I will take a review ASAP, if ok I will merge this. BTW, @ulysses-you please help to open tpcds q67 test by this way:
> 
> ```
> \\ in GlutenClickHouseTPCDSAbstractSuite.scala line 98:
> remove "q67"
> ```

The TPCDS q67 issue in CH is not related, please revert q67 in excluding list, thanks.
Others LGTM.

rui-mo (approved):

PR #2183, opened 2023-07-03
[GLUTEN-1434][VL] Optimize shuffle get batch

jinchengchenghh:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -374,10 +374,11 @@ std::shared_ptr<arrow::Buffer> VeloxShuffleWriter::generateComplexTypeBuffers(ve
 }
 
 arrow::Status VeloxShuffleWriter::split(std::shared_ptr<ColumnarBatch> cb) {
-  auto veloxColumnBatch = VeloxColumnarBatch::from(veloxPool_.get(), cb);
-  auto& rv = *veloxColumnBatch->getFlattenedRowVector();
-  RETURN_NOT_OK(initFromRowVector(rv));
   if (options_.partitioning_name == "single") {
+    auto veloxColumnBatch = std::dynamic_pointer_cast<VeloxColumnarBatch>(cb);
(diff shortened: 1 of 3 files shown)


zhztheplayer (commented):
@jinchengchenghh Can you explain a little bit about the optimization that the patch actually does? Thanks a lot.

jinchengchenghh:
The function `from` will combine a 1 column ColumnarBatch(pid) and a several columns Columnarbatch(data), if the latter batch columns are too much, this will be a burden. The old function `addColumn` use `std::move(children)` to avoid generating all new children vector. @zhztheplayer

zhztheplayer:
> if the latter batch columns are too much, this will be a burden.

Ideally it just brings some meta-data level copy-and-pastes out of the data. Is there some real performance regressions observed from your side?

jinchengchenghh:
No, it may gain performance in user situation.

zhztheplayer (dismissed):
I would not be against this but as an out-of-box observer of this code, it turn's to me that the codes slightly get more complex than it used to be since another flag `removePid` is introduced to the routine. Since we already had a `hasPid()` API in partitioner.

`VeloxColumnarBatch::from(..)` was added to centralize all the batch converting work to reduce the code complexity. If no performance regression is guaranteed to take place I would suggest doing performance optimization work only when it's really needed.

zhztheplayer:
Would you like to rebase? Thanks.

jinchengchenghh:
I optimize the code logic to remove `removePid`, the old logic is too complicated.

jinchengchenghh:
For velox shuffle, we don't need Arrow schema_, do you think we should move to velox types? @zhztheplayer There may be a lot of code refactor.

zhztheplayer (approved):

zhztheplayer:
>  For velox shuffle, we don't need Arrow schema_, do you think we should move to velox types? @zhztheplayer There may be a lot of code refactor.

By this did you mean using the current approach that infers schema from the first columnar batch?

Issue #2139, opened 2023-06-29
[VL] Make fallback details visualize and analyzable


ulysses-you:
Based on the current validation framework, I propose to make the changes:

1. make `doValidate` return a structure which contains the fallback reason
2. put the reason into the vanilla spark node tag
3. add a rule to collect fallback reason for the whole plan, the text view whose style like the `explain formatted`. It shows the fallback reason with the corresponding vanilla spark node
5. post event to spark listener bus. note that, in AQE, each stage will post a new event. The behavior is similar with `SparkListenerSQLAdaptiveExecutionUpdate`
6. add a ui tab show the event, the style should follow the sql execution tab

An example of the fallback text view:

```
== Physical Plan ==
* Project (4)
+- * ColumnarToRow (3)
   +- LoadArrowData (2)
      +-  Scan parquet default.t1 (1)


(1) Scan parquet default.t1
Fallback reason:
Velox backend parquet scan does not support map type ...

(2) LoadArrowData

(3) ColumnarToRow [codegen id : 1]

(4) Project [codegen id : 1]
```

ulysses-you:
cc @FelixYBW @zhouyuan @yma11 do you have any thoughts about it ?

PR #2185, opened 2023-07-03
[GLUTEN-2139][CORE] Inject fallback reason to plan tag

ulysses-you:

This pr resolves the steps (1, 2, 2.5) in #2139 

The mian change is:
1. make `TransformSupport` inherit `GlutenPlan` and pull out validate related code to `GlutenPlan`.
2. change the `doValidateInternal` result type from `boolean` to `ValidationResult `, `ValidationResult` holds a string type reason.
3. centralize `Throwable` catching and logging to `doValidate`
4. add a new rule `GlutenFallbackReporter` to collect unsupport transform tag and print fallback reason. (post event will do in a seperate pr)

Note, This pr assumes that:
- if a vanilla SparkPlan can be offloaded, then Gluten make a new `TransformSupport`
- if a plan is produced by Gluten, then it should inherit `GlutenPlan`
- every vanilla SparkPlan should contains transform hint

(Fixes: \#2139)

## How was this patch tested?
Pass CI and add test

An example:
```
create table t (c1 int) using parquet;

set spark.gluten.sql.validate.failure.logLevel=warn;
set spark.gluten.sql.columnar.filescan=false;

select * from t;

23/07/03 20:22:54 WARN GlutenFallbackReporter: Validation failed for plan: Scan parquet default.t, due to: columnar FileScan is not enabled in FileSourceScanExec.

```

### DIFF
--- a/backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashJoinExecTransformer.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashJoinExecTransformer.scala
@@ -16,6 +16,7 @@
  */
 package io.glutenproject.execution
 
+import io.glutenproject.extension.ValidationResult
 import io.glutenproject.utils.CHJoinValidateUtil
 
 import org.apache.spark.sql.catalyst.expressions._
@@ -46,11 +47,11 @@ case class CHShuffledHashJoinExecTransformer(
       newRight: SparkPlan): CHShuffledHashJoinExecTransformer =
     copy(left = newLeft, right = newRight)
 
-  override def doValidateInternal(): Boolean = {
+  override def doValidateInternal(): ValidationResult = {
     val shouldFallback =
       CHJoinValidateUtil.shouldFallback(joinType, left.outputSet, right.outputSet, condition)
     if (shouldFallback) {
-      return false
+      return notOk("ch join validate fail")
     }
     super.doValidateInternal()
   }
@@ -80,17 +81,15 @@ case class CHBroadcastHashJoinExecTransformer(
       newRight: SparkPlan): CHBroadcastHashJoinExecTransformer =
     copy(left = newLeft, right = newRight)
 
-  /*
-
-   */
-  override def doValidateInternal(): Boolean = {
-    var shouldFallback =
+  override def doValidateInternal(): ValidationResult = {
+    val shouldFallback =
       CHJoinValidateUtil.shouldFallback(joinType, left.outputSet, right.outputSet, condition)
-    if (isNullAwareAntiJoin) {
-      shouldFallback = true
-    }
+
     if (shouldFallback) {
-      return false
+      return notOk("ch join validate fail")
+    }
+    if (isNullAwareAntiJoin) {
+      return notOk("ch does not support NAAJ")
     }
     super.doValidateInternal()
   }
--- a/backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashJoinExecTransformer.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashJoinExecTransformer.scala
@@ -81,7 +81,7 @@ case class CHBroadcastHashJoinExecTransformer(
       newRight: SparkPlan): CHBroadcastHashJoinExecTransformer =
     copy(left = newLeft, right = newRight)
 
-  override def doValidateInternal(): ValidationResult = {
+  override protected def doValidateInternal(): ValidationResult = {
     val shouldFallback =
       CHJoinValidateUtil.shouldFallback(joinType, left.outputSet, right.outputSet, condition)
 
--- a/backends-velox/src/main/scala/io/glutenproject/execution/FilterExecTransformer.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/execution/FilterExecTransformer.scala
@@ -23,6 +23,7 @@ import scala.collection.JavaConverters._
 
 import com.google.common.collect.Lists
 import io.glutenproject.GlutenConfig
+import io.glutenproject.extension.ValidationResult
 import io.glutenproject.substrait.SubstraitContext
 import io.glutenproject.substrait.plan.PlanBuilder
 import io.glutenproject.substrait.rel.RelBuilder
@@ -34,42 +35,26 @@ import org.apache.spark.sql.execution.SparkPlan
 case class FilterExecTransformer(condition: Expression, child: SparkPlan)
   extends FilterExecTransformerBase(condition, child) with TransformSupport {
 
-  override def doValidateInternal(): Boolean = {
+  override def doValidateInternal(): ValidationResult = {
     val leftCondition = getLeftCondition
     if (leftCondition == null) {
       // All the filters can be pushed down and the computing of this Filter
       // is not needed.
-      return true
+      return ok()
     }
     val substraitContext = new SubstraitContext
     val operatorId = substraitContext.nextOperatorId(this.nodeName)
     // Firstly, need to check if the Substrait plan for this operator can be successfully generated.
-    val relNode = try {
-      getRelNode(
-        substraitContext, leftCondition, child.output, operatorId, null, validation = true)
-    } catch {
-      case e: Throwable =>
-        this.appendValidateLog(
-          s"Validation failed for ${this.getClass.toString} due to: ${e.getMessage}")
-        return false
-    }
+    val relNode = getRelNode(
+      substraitContext, leftCondition, child.output, operatorId, null, validation = true)
     val planNode = PlanBuilder.makePlan(substraitContext, Lists.newArrayList(relNode))
     // Then, validate the generated plan in native engine.
     if (GlutenConfig.getConf.enableNativeValidation) {
       val validator = new NativePlanEvaluator()
       val validateInfo = validator.doValidateWithFallBackLog(planNode.toProtobuf.toByteArray)
-      if (!validateInfo.isSupported) {
-        val fallbackInfo = validateInfo.getFallbackInfo()
-        for (i <- 0 until fallbackInfo.size()) {
-          this.appendValidateLog(fallbackInfo.get(i))
-        }
-        this.appendValidateLog(s"Validation failed for ${this.getClass.toString}" +
-          s" due to: native check failure.")
-        return false
-      }
-      true
+      nativeValidationResult(validateInfo)
     } else {
-      true
+      ok()
     }
   }
 
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
@@ -141,7 +141,7 @@ abstract class FilterExecTransformerBase(val cond: Expression,
     }
   }
 
-  override def doValidateInternal(): ValidationResult = {
+  override protected def doValidateInternal(): ValidationResult = {
     if (cond == null) {
       // The computing of this Filter is not needed.
       return ok()
@@ -229,7 +229,7 @@ case class ProjectExecTransformer(projectList: Seq[NamedExpression],
 
   override def supportsColumnar: Boolean = GlutenConfig.getConf.enableColumnarIterator
 
-  override def doValidateInternal(): ValidationResult = {
+  override protected def doValidateInternal(): ValidationResult = {
     val substraitContext = new SubstraitContext
     // Firstly, need to check if the Substrait plan for this operator can be successfully generated.
     val operatorId = substraitContext.nextOperatorId(this.nodeName)
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
@@ -22,7 +22,7 @@ import com.google.protobuf.Any
 import io.glutenproject.GlutenConfig
 import io.glutenproject.backendsapi.BackendsApiManager
 import io.glutenproject.expression.{ConverterUtils, ExpressionConverter, ExpressionTransformer}
-import io.glutenproject.extension.GlutenPlan
+import io.glutenproject.extension.{GlutenPlan, ValidationResult}
 import io.glutenproject.extension.columnar.TransformHints
 import io.glutenproject.metrics.MetricsUpdater
 import io.glutenproject.substrait.SubstraitContext
@@ -47,7 +47,6 @@ import scala.collection.JavaConverters._
 abstract class FilterExecTransformerBase(val cond: Expression,
                                          val input: SparkPlan) extends UnaryExecNode
   with TransformSupport
-  with GlutenPlan
   with PredicateHelper
   with AliasAwareOutputPartitioning
   with Logging {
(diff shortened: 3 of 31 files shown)


backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashJoinExecTransformer.scala
```diff
@@ -82,17 +81,15 @@ case class CHBroadcastHashJoinExecTransformer(
...
-    if (isNullAwareAntiJoin == true) {
-      shouldFallback = true
-    }
+
     if (shouldFallback) {
-      return false
+      return notOk("ch join validate fail")
+    }
+    if (isNullAwareAntiJoin) {
+      return notOk("ch does not support NAAJ")
```
ulysses-you, line 92: these are missed fallback reason

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -525,13 +526,14 @@ case class TransformPreOverrides(
...
-      if (transformer.doValidate()) {
+      val validationResult = transformer.doValidate()
+      if (validationResult.validated) {
         logDebug(s"Columnar Processing for ${plan.getClass} is currently supported.")
         transformer
       } else {
         logDebug(s"Columnar Processing for ${plan.getClass} is currently unsupported.")
         val newSource = plan.copy(partitionFilters = newPartitionFilters)
-        TransformHints.tagNotTransformable(newSource)
+        TransformHints.tagNotTransformable(newSource, validationResult.reason.get)
```
ulysses-you, line 536: missed fallback reason before

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -545,13 +547,14 @@ case class TransformPreOverrides(
...
-      if (transformer.doValidate()) {
+      val validationResult = transformer.doValidate()
+      if (validationResult.validated) {
         logDebug(s"Columnar Processing for ${plan.getClass} is currently supported.")
         transformer
       } else {
         logDebug(s"Columnar Processing for ${plan.getClass} is currently unsupported.")
         val newSource = plan.copy(runtimeFilters = newPartitionFilters)
-        TransformHints.tagNotTransformable(newSource)
+        TransformHints.tagNotTransformable(newSource, validationResult.reason.get)
```
ulysses-you, line 557: missed fallback reason before

gluten-core/src/main/scala/io/glutenproject/extension/StrategyOverrides.scala
```diff
@@ -183,7 +184,7 @@ case class JoinSelectionOverrides(session: SparkSession) extends Strategy with
     // Ignore forceShuffledHashJoin if exist multi continuous joins
     if (GlutenConfig.getConf.enableLogicalJoinOptimize &&
       existsMultiJoins(plan) && existLeftOuterJoin(plan)) {
-      tagNotTransformableRecursive(plan)
+      tagNotTransformableRecursive(plan, "exist multi continuous joins")
```
ulysses-you, line 187: missed fallback reason before

backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashJoinExecTransformer.scala
ulysses-you, line 92: missed fallback reason before

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -559,7 +520,8 @@ object FilterHandler {
               .transformDynamicPruningExpr(
                 batchScan.runtimeFilters,
                 reuseSubquery))
-            TransformHints.tagNotTransformable(newSource)
+            TransformHints.tagNotTransformable(newSource,
+              "The scan in BatchScanExec is not a FileScan")
```
ulysses-you, line 524: missed fallback reason before

backends-velox/src/main/scala/io/glutenproject/execution/FilterExecTransformer.scala
```diff
@@ -34,42 +35,26 @@ import org.apache.spark.sql.execution.SparkPlan
...
     val substraitContext = new SubstraitContext
     val operatorId = substraitContext.nextOperatorId(this.nodeName)
     // Firstly, need to check if the Substrait plan for this operator can be successfully generated.
-    val relNode = try {
-      getRelNode(
-        substraitContext, leftCondition, child.output, operatorId, null, validation = true)
-    } catch {
-      case e: Throwable =>
-        this.appendValidateLog(
-          s"Validation failed for ${this.getClass.toString} due to: ${e.getMessage}")
```
ulysses-you, line 53: IIUC, all exception during validation should be print by `logValidateFailure`

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -115,7 +127,7 @@ object TagBeforeTransformHits {
 case class FallbackOnANSIMode(session: SparkSession) extends Rule[SparkPlan] {
   override def apply(plan: SparkPlan): SparkPlan = PhysicalPlanSelector.maybe(session, plan) {
     if (GlutenConfig.getConf.enableAnsiMode) {
-      plan.foreach(TransformHints.tagNotTransformable)
+      plan.foreach(TransformHints.tagNotTransformable(_, "does not support ansi mode"))
```
ulysses-you, line 130: missed fallback reason before

ulysses-you:
cc @FelixYBW @zhouyuan @yma11 I think it's ready to review, thank you

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -250,36 +229,20 @@ case class ProjectExecTransformer(projectList: Seq[NamedExpression],
...
-    val relNode = try {
-      getRelNode(
-        substraitContext, projectList, child.output, operatorId, null, validation = true)
-    } catch {
-      case e: Throwable =>
-        this.appendValidateLog(s"Validation failed for ${this.getClass.toString}" +
-          s" due to: ${e.getMessage}")
-        return false
-    }
+    val relNode = getRelNode(
```
yma11, line 236: I think this change is not correct. It will cause: 1) failure of translation of plan node is not caught. 2) should return directly as this step indicates the validation failed.

gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
```diff
@@ -33,7 +33,7 @@ import org.apache.spark.sql.execution.InSubqueryExec
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
-trait BasicScanExecTransformer extends TransformSupport with GlutenPlan {
+trait BasicScanExecTransformer extends TransformSupport {
```
yma11, line 36: Why remove the depends of `GlutenPlan`? I think it's not proper even nothing important implemented in it. This is a `trait` that all Gluten plan node inherited from, no matter it supports transform or not.

gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
```diff
@@ -207,51 +208,29 @@ case class ExpandExecTransformer(projections: Seq[Seq[Expression]],
...
   }
 
-  override def doValidateInternal(): Boolean = {
+  override protected def doValidateInternal(): ValidationResult = {
     if (!BackendsApiManager.getSettings.supportExpandExec()) {
-      this.appendValidateLog(
-        s"Validation failed for ${this.getClass.toString}" +
-          s" due to: {ExpandExec}. ")
-      return false
+      return notOk("backend does not support expand")
```
yma11, line 213: will this message like `backend does not support expand` be shown in Spark UI directly or you will further process it? I think we'd better consider both meaning and simplicity.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -132,50 +132,31 @@ abstract class HashAggregateExecBaseTransformer(
       case d: DecimalType => true
       case a: ArrayType => true
       case n: NullType => true
-      case other => this.appendValidateLog(
-        s"Validation failed for ${this.getClass.toString}" +
-          s" due to: {data type ${dataType}}");
-        false
+      case other => false
```
yma11, line 135: Then how do we know the name of this unsupported data type?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -593,18 +534,13 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
 
             val maybeExchange = buildSidePlan.find {
               case BroadcastExchangeExec(_, _) => true
-              case _ =>
-                printValidateLog(this.getClass.toString,
-                  "not BroadcastExchangeExec at join build side.")
-                false
+              case _ => false
```
yma11, line 537: So this failure info will be lost?

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
ulysses-you: > should return directly as this step indicates the validation failed.

If `getRelNode` throw exception, it would be caught by caller side (`doValidate`). Isn't it return directly for you ?

>  failure of translation of plan node is not caught

plan node is caught as we would inject unsupport tansformable hint to that plan, and before logging, it would add plan node name.

gluten-core/src/main/scala/io/glutenproject/extension/GlutenPlan.scala
```diff
@@ -17,13 +17,77 @@
...
+import io.glutenproject.validate.NativePlanValidatorInfo
+import org.apache.spark.sql.execution.SparkPlan
+
+case class ValidationResult(
+    validated: Boolean,
+    reason: Option[String])
+
+object ValidationResult {
+  def ok: ValidationResult = ValidationResult(validated = true, None)
+  def notOk(reason: String): ValidationResult = ValidationResult(validated = false, Some(reason))
```
yma11, line 35: `reason` is not always a `String` here. For the validate log from native side, it's a `Vector`.

gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
ulysses-you: See the pr description, this pr would print something like:

```
23/07/03 20:22:54 WARN GlutenFallbackReporter: Validation failed for plan: Scan parquet default.t, due to: Some(columnar FileScan is not enabled in FileSourceScanExec).
```

This logging format is based on your previous pr. We can improve it in UI in next pr if we want.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
ulysses-you: no, the `maybeExchange` is not the final validation result. The code here is used to find a broadcast exchange, and we would match both Some and None case to do the final validation, see line 576.

gluten-core/src/main/scala/io/glutenproject/extension/GlutenPlan.scala
ulysses-you: yea, it has been flattened by `.mkString("native check failure:", ", ", "")`, see `Gluten.nativeValidationResult`

gluten-core/src/main/scala/io/glutenproject/execution/ExpandExecTransformer.scala
yma11: Yes. My previous planned next step is to format the error message further. Eventually, we should only show some key words for failed operations like "data type: map, function: get_struct_feild(XXXX)".

gluten-core/src/main/scala/org/apache/spark/sql/execution/GlutenFallbackReporter.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  private def printFallbackReason(plan: SparkPlan): Unit = {
+    val validateFailureLogLevel = glutenConfig.validateFailureLogLevel
+    plan.foreach {
+      case _: GlutenPlan => // ignore
+      case plan: SparkPlan =>
+        // some SparkPlan do not have hint, e.g., `ColumnarAQEShuffleRead`
+        TransformHints.getHintOption(plan) match {
+          case Some(TRANSFORM_UNSUPPORTED(reason)) =>
+            logOnLevel(validateFailureLogLevel,
```
yma11, line 52: maybe you need to change default value `validateFailureLogLevel`  to `INFO` so that fallback reasons will be printed as default.

gluten-core/src/main/scala/org/apache/spark/sql/execution/GlutenFallbackReporter.scala
```diff
@@ -0,0 +1,58 @@
...
+ */
+case class GlutenFallbackReporter(
+    glutenConfig: GlutenConfig,
+    spark: SparkSession) extends Rule[SparkPlan] with LogLevelUtil {
+
+  override def apply(plan: SparkPlan): SparkPlan = {
+    if (!glutenConfig.enableFallbackReport) {
+      return plan
+    }
+    printFallbackReason(plan)
```
yma11, line 40: can this ` printFallbackReason(plan)` be put ahead of check `enableFallbackReport`?  We'd allow fallback reasons dump into log even not show in UI.

ulysses-you: I'm fine with this

ulysses-you: or, we can add one more config

yma11: No need to add one more config. Change `logOnLevel` to `logInfo` and do a check for `printFallbackReason` by `validateFailureLogLevel == DEBUG` and keep default of `validateFailureLogLevel` to DEBUG. It should work as previous print logic.

ulysses-you: I wonder why we need do `validateFailureLogLevel == DEBUG`, why not `logOnLevel(validateFailureLogLevel)` ?

yma11: We want not to print fallback log when validateFailureLogLevel is `INFO` but by default we print the logs so we set it to `validateFailureLogLevel ` to `DEBUG` by default. The logic is a little tricky so I am okay if you keep using `logOnLevel`. Just make sure by default it's dumped.

ulysses-you:
any more thought ? @yma11  this pr is quite big, and we can improve in followup..

yma11:
> any more thought ? @yma11 this pr is quite big, and we can improve in followup..

LGTM. @FelixYBW @zhouyuan can you also take a look?

zhouyuan:
@lgbo-ustc would you please kindly take a look on the changes on CK backend? 

-yuan

PR #2197, opened 2023-07-04
[CORE] Add QAT zstd codec

marin-ma:


backends-clickhouse/src/main/scala/org/apache/spark/shuffle/CHColumnarShuffleWriter.scala
```diff
@@ -49,7 +49,7 @@ class CHColumnarShuffleWriter[K, V](
   private val subDirsPerLocalDir = blockManager.diskBlockManager.subDirsPerLocalDir
   private val splitSize = GlutenConfig.getConf.maxBatchSize
   private val customizedCompressCodec =
-    GlutenShuffleUtils.getCompressionCodec(conf)
+    GlutenShuffleUtils.getCompressionCodec(conf).toUpperCase(Locale.ROOT)
```
jinchengchenghh, line 52: Do we need to support both uppercase and lowercase codec config? I would suggest to align with spark config `spark.io.compression.codec`  to only support lower case.

marin-ma: I agree. @zzcclp Can you take a look? Using lower case codec string will fail on CH native side.

gluten-core/src/main/scala/org/apache/spark/shuffle/GlutenShuffleUtils.scala
```diff
@@ -40,18 +41,25 @@ object GlutenShuffleUtils {
...
         }
+        codec
       case None =>
         val sparkCodecKey = IO_COMPRESSION_CODEC.key
         val codec =
-          conf.get(sparkCodecKey, IO_COMPRESSION_CODEC.defaultValueString).toUpperCase(Locale.ROOT)
-        checkCodecValues(sparkCodecKey, codec, GlutenConfig.GLUTEN_SHUFFLE_SUPPORTED_CODEC)
+          conf
+            .get(sparkCodecKey, IO_COMPRESSION_CODEC.defaultValueString)
+            .toLowerCase(Locale.ROOT)
```
jinchengchenghh, line 58: default is lower case, no need to convert.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -709,19 +709,19 @@ object GlutenConfig {
...
-          "When spark.gluten.sql.columnar.shuffle.codecBackend=iaa, the supported codec is gzip.")
+          "When spark.gluten.sql.columnar.shuffle.codecBackend=qat," +
+          "the supported codecs are gzip and zstd. " +
+          "When spark.gluten.sql.columnar.shuffle.codecBackend=iaa," +
+          "the supported codec is gzip.")
       .stringConf
-      .transform(_.toUpperCase(Locale.ROOT))
-      .checkValues(
-        GLUTEN_SHUFFLE_SUPPORTED_CODEC ++ GLUTEN_QAT_SUPPORTED_CODEC ++ GLUTEN_IAA_SUPPORTED_CODEC)
+      .transform(_.toLowerCase(Locale.ROOT))
```
jinchengchenghh, line 717: also here

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -709,19 +709,19 @@ object GlutenConfig {
...
       .createOptional
 
   val COLUMNAR_SHUFFLE_CODEC_BACKEND =
     buildConf(GlutenConfig.GLUTEN_SHUFFLE_CODEC_BACKEND)
       .internal()
       .stringConf
-      .transform(_.toUpperCase(Locale.ROOT))
+      .transform(_.toLowerCase(Locale.ROOT))
```
jinchengchenghh, line 724: and here

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
marin-ma: Here we want it to be converted when user provides uppercase config.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
marin-ma: Same as above

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
jinchengchenghh: Same as this one, https://github.com/oap-project/gluten/pull/2197#discussion_r1257735246 If we only accept lower case, we don't need to transform.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
jinchengchenghh: Most of config options only support lower case, consistence is good. But if you aim to support both, I don't against it.

marin-ma: Thanks. We need to support both for backward compatibility.

backends-clickhouse/src/main/scala/org/apache/spark/shuffle/CHColumnarShuffleWriter.scala
zzcclp: In CH, it only support uppercase codec.

FelixYBW: So now we are align with Vanilla Spark's config but convert to uppper in Gluten, right? if so it's OK

marin-ma: Yes, we're align with spark's config. Here we only convert to uppercase for CH backend.

FelixYBW (dismissed):

jinchengchenghh (approved):

jinchengchenghh (approved):

PR #2201, opened 2023-07-04
[GLUTEN-2198] [CH] handle different header between upstream and downstream

shuai-xu:

This pr handle the case that input block is different with the header and fix the problem that keys may be wrong in aggerator. This two problems are all caused by that columns may be in different order in spark plan betweent upsteam and its downstream.

(Fixes: \#2198)

## How was this patch tested?

### DIFF
--- a/cpp-ch/local-engine/Storages/SourceFromJavaIter.cpp
+++ b/cpp-ch/local-engine/Storages/SourceFromJavaIter.cpp
@@ -40,8 +40,7 @@ DB::Chunk SourceFromJavaIter::generate()
             size_t rows = data->rows();
             if (original_header.columns())
             {
-                result.setColumns(data->mutateColumns(), rows);
-                convertNullable(result);
+                buildChunk(data, result);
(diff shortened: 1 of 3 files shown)


cpp-ch/local-engine/Storages/SourceFromJavaIter.cpp
```diff
@@ -73,19 +72,50 @@ Int64 SourceFromJavaIter::byteArrayToLong(JNIEnv * env, jbyteArray arr)
...
+        }
+        else
+        {
+            out_columns[i] = columns.at(pos);
+        }
+    }
+    chunk.setColumns(out_columns, rows);
+}
+
+int SourceFromJavaIter::getPositionByNameForUnion(DB::Block * header, std::string & output_name)
```
liuneng1994, line 105: 这里可能会出现重名的列~需要进行错误处理

liuneng1994 (commented):
需要补充逻辑

PR #2205, opened 2023-07-04
[GLUTEN-1925][CH] allow all parquet write jobs

binmahone:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenParquetWriterInjectsBase.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.execution.ColumnarCollapseTransformStages.transformStageCounter
+import org.apache.spark.sql.execution.SparkPlan
+
+trait GlutenParquetWriterInjectsBase extends GlutenParquetWriterInjects {
+
+  /**
+   * FileFormatWriter wraps some Project & Sort on the top of the original output spark plan, we
+   * need to replace them with Columnar version
```
JkSelf, line 34: Why need Project and Sort?

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -80,48 +74,53 @@ case class FakeRowAdaptor(child: SparkPlan) extends UnaryExecNode with GlutenPla
...
-  // TODO: support Insert clause
-
+object GlutenWriterColumnarRules {
   // TODO: support ctas in Spark3.4, see https://github.com/apache/spark/pull/39220
   // TODO: support dynamic partition and bucket write
   //  1. pull out `Empty2Null` and required ordering to `WriteFilesExec`, see Spark3.4 `V1Writes`
   //  2. support detect partition value, partition path, bucket value, bucket path at native side,
   //     see `BaseDynamicPartitionDataWriter`
-  def isGlutenInsertInto(cmd: DataWritingCommand): Boolean = {
+  def isNativeParquetAppliable(cmd: DataWritingCommand): Boolean = {
```
JkSelf, line 83: isNativeParquetAvailable?

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -80,48 +74,53 @@ case class FakeRowAdaptor(child: SparkPlan) extends UnaryExecNode with GlutenPla
...
+    }
+
     cmd match {
       case command: CreateDataSourceTableAsSelectCommand =>
         if (command.table.provider.contains("velox")) {
           throw new UnsupportedOperationException(
             "Velox file format does not support create table as select.")
         }
-        false
+        true
```
JkSelf, line 95: Why change to true?

shims/common/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala
```diff
@@ -0,0 +1,181 @@
...
+import org.apache.hadoop.hive.serde2.Serializer
+import org.apache.hadoop.hive.serde2.objectinspector.{ObjectInspectorUtils, StructObjectInspector}
+import org.apache.hadoop.hive.serde2.objectinspector.ObjectInspectorUtils.ObjectInspectorCopyOption
+import org.apache.hadoop.io.Writable
+import org.apache.hadoop.mapred.{JobConf, Reporter}
+import org.apache.hadoop.mapreduce.{Job, TaskAttemptContext}
+
+import scala.collection.JavaConverters._
+
+/* -
```
JkSelf, line 46: we need to put HiveFileFormat to shim-spark32 and shim-spark33. The HiveFileFormat API is incompatiable in spark 3.2 and spark3.3.

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/v1/CHParquetWriterInjects.scala
```diff
@@ -14,39 +14,20 @@
...
-  }
-
-  override def prepareWrite(
-      sparkSession: SparkSession,
-      job: Job,
-      options: Map[String, String],
-      dataSchema: StructType): OutputWriterFactory = {
+class CHParquetWriterInjects extends GlutenParquetWriterInjectsBase {
 
+  def createOutputWriterFactory: OutputWriterFactory = {
```
JkSelf, line 30: `HiveFileFormat `and `ParquetFileformat `both call this method to create OutputWriterFactory? The filename of `HiveFileFormat  `and `ParquetFileformat  `is different. [PR#2328](https://github.com/oap-project/gluten/pull/2328) update the filename of `Hiveformat `to align with vanilla spark.

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenParquetWriterInjectsBase.scala
binmahone: it's a behavior of spark 3.2, spark wants to use a project and local sort to help output partitioning and bucketing data

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
binmahone: native parquet is always available , but in some cases it's not applicable, that's why we need to check it.

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
binmahone: it's a false for velox. otherwise (for ch case) it's good to be true

shims/common/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala
binmahone: ok. good to know. I thought 33 is compatible with 32

binmahone: will change it

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/v1/CHParquetWriterInjects.scala
binmahone: our implementation does not involve naming of the file.  the file name is passed in as a parameter. correct me again if I understand it wrong

JkSelf: VeloxParquetFileFormat generate the filename in [here](https://github.com/oap-project/gluten/blob/55103a3e113c597572c6c552fd090a73ac6d5da9/backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxParquetFileFormat.scala#L66). And the HiveFileFormat code is [here](https://github.com/oap-project/gluten/blob/55103a3e113c597572c6c552fd090a73ac6d5da9/backends-velox/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala#L138).

shims/common/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala
binmahone: done

binmahone:
Run Gluten Clickhouse CI

binmahone:
Run Gluten Clickhouse CI

zzcclp (approved):
LGTM

JkSelf (approved):
LGTM.

ulysses-you:
Personally, I really do not like this approach to implement columnar write. The hacking way that uses the same class name to overwrite the Spark's.

It's flaky, and make it hard to maintain in future. 
1. We can not easily troubleshooting and workaround the issue if there is class path conflicts in prod env. There is no such a switch can disable the behavior at runtime. It will be more complex in user env due to class path can be modified. I do not want to deal with this issue at every morning.
2. No one will know all the changes in the copied code. Every time we touch these code we need to compare the vanilla Spark. A bad example is the Hive code in Spark, it prevents the new contributors to improve that part.
3. It's also not easy to compatible with new Spark version in futurn. It's the kind of rebase.

But I see there is the requirements to support columnar write in Spark 3.2 and 3.3. I do not want to block this pr.

Can we at least put these code in new modules ? So we have a choose to not compile with it and Gluten should still work with vanilla Spark normally.

binmahone:
> Personally, I really do not like this approach to implement columnar write. The hacking way that uses the same class name to overwrite the Spark's.
> 
> It's flaky, and make it hard to maintain in future.
> 
> 1. We can not easily troubleshooting and workaround the issue if there is class path conflicts in prod env. There is no such a switch can disable the behavior at runtime. It will be more complex in user env due to class path can be modified. I do not want to deal with this issue at every morning.
> 2. No one will know all the changes in the copied code. Every time we touch these code we need to compare the vanilla Spark. A bad example is the Hive code in Spark, it prevents the new contributors to improve that part.
> 3. It's also not easy to compatible with new Spark version in futurn. It's the kind of rebase.
> 
> But I see there is the requirements to support columnar write in Spark 3.2 and 3.3. I do not want to block this pr.
> 
> Can we at least put these code in new modules ? So we have a choose to not compile with it and Gluten should still work with vanilla Spark normally.

hi,  @ulysses-you we do understand this approach has maintenance issues, however there're a few points I want to clarify:

1. The practice of overwriting spark class already exists in Gluten (https://github.com/oap-project/gluten/blob/main/docs/developers/NewToGluten.md#how-to-prioritize-loading-gluten-jars-in-spark), and there is a clean way to ensure right class get loaded in both PROD/DEV env.
2. "There is no such a switch can disable the behavior at runtime" -> This is not true, there's a toggle called "spark.gluten.sql.native.parquet.writer.enabled" (by default false now) to disable this feature.  
3. It's not that hard to know what's changed by comparing code changes with handy tools like Intellij IDEA's "Compare with clipboard"
4. It seems to me, no other cheap alternative approach is there for Spark 3.2 and 3.3. Do you have any practical suggestions? It's impossible for us to give up Spark 3.2 and Spark 3.3

zzcclp (approved):
LGTM

ulysses-you:
thank you @binmahone for the clarification!

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -80,60 +77,80 @@ case class FakeRowAdaptor(child: SparkPlan) extends UnaryExecNode with GlutenPla
...
+      case command: InsertIntoHiveTable =>
+        "org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat".equals(
+          command.table.storage.outputFormat.get)
+      case _: CreateHiveTableAsSelectCommand =>
+        throw new UnsupportedOperationException(
+          "native writer cannot recognize command: " + cmd.getClass.getName +
+            " please append `using parquet` in your CTAS query")
+      case _ =>
+        throw new UnsupportedOperationException(
+          "native writer cannot recognize command: " + cmd.getClass.getName)
```
jackylee-ch, line 121: Here, we are only doing an `isNativeParquetAppliable` analysis on the plan. For inapplicable situations, it should return false rather than throwing an exception.

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -80,60 +77,80 @@ case class FakeRowAdaptor(child: SparkPlan) extends UnaryExecNode with GlutenPla
...
+                Array(
+                  FakeRowAdaptor(
+                    AdaptiveSparkPlanExec(
+                      aqe.inputPlan,
+                      aqe.context,
+                      aqe.preprocessingRules,
+                      aqe.isSubquery,
+                      supportsColumnar = true
+                    ))))
+            case other => rc.withNewChildren(Array(FakeRowAdaptor(other)))
```
jackylee-ch, line 150: It is better to check `ColumnarToRowExec` then add `FakeRowAdaptor`, as the output of child may be `Row`, not `ColumnarBatch`

PR #2206, opened 2023-07-04
[CH] Fix tpcds q67

liuneng1994:

Fix tpcds q67 failed on aqe enabled

## How was this patch tested?

unit tests

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/TakeOrderedAndProjectExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/TakeOrderedAndProjectExecTransformer.scala
@@ -69,6 +69,7 @@ case class TakeOrderedAndProjectExecTransformer(limit: Int,
       sparkContext.parallelize(Seq.empty, 1)
     } else {
       // The child should have been replaced by ColumnarCollapseTransformStages.
+//      val localSort = childRDDPartsNum == 1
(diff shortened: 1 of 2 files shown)


liuneng1994:
Optimizations for partition 1 do not work

lgbo-ustc:
there is a pr about this in #2153

liuneng1994:
> there is a pr about this part in #2153

the pr cann't work on ds q67, maybe we can remove this optimization directly.

gluten-core/src/main/scala/io/glutenproject/execution/TakeOrderedAndProjectExecTransformer.scala
```diff
@@ -82,19 +83,16 @@ case class TakeOrderedAndProjectExecTransformer(limit: Int,
       }
       val transformStageCounter: AtomicInteger =
         ColumnarCollapseTransformStages.transformStageCounter
-      val finalLimitPlan = if (childRDDPartsNum == 1) {
-          limitExecPlan
```
ulysses-you, line 86: This is interesting, it seems ok to remove a shuffle if the input rdd only has one partition. Why does it not work ?

liuneng1994: Lack of knowledge, I don't know why.

PR #2230, opened 2023-07-06
[GLUTEN-2227][VL] Fix coverity Java scan issues

xieqi:

Fix coverity Java scan issues

## How was this patch tested?

Using exist UTs


zhztheplayer (approved):
I guess simply adding `final` modifier to the fields (e.g. `public long inputRows` -> `public final long inputRows`) would pass the code scan

PR #2235, opened 2023-07-06
[CORE] Skip path validate check in scan

jackylee-ch:
The path validation check would failed when reading a parquet external table with BatchScan, as the partition path is arbitrary for the external table. Here is a little example:
```
SET spark.sql.catalogImplementation=hive;
SET spark.sql.hive.convertMetastoreParquet=true;
CREATE EXTERNAL TABLE test (a STRING, b STRING)
PARTITIONED BY (day STRING)
STORED AS PARQUET
LOCATION '/path/to/test';
ALTER TABLE test ADD PARTITION(day='20230706') LOCATION '/tmp/20230706';
SELECT * FROM test WHERE day='20230706';
```


PHILO-HE:
Hi @jackylee-ch, there is a PR to support custom partition path. See https://github.com/oap-project/gluten/pull/2405. It looks your code changes are covered there. Please have a try. Thanks!

jackylee-ch:
> Hi @jackylee-ch, there is a PR to support custom partition path. See #2405. And it looks your code changes are covered there. Please have a try. Thanks!

Yeap, it can be covered in that pr. I would close this pr and moving to #2405. Thanks.

PR #2238, opened 2023-07-07
[GLUTEN-1632][CH]Daily Update Clickhouse Version (20230707)

lwz9103:
Auto commit by gluten daily build, please check the build status and merge it if it's green.


baibaichen:
see 
image

see following `getConfigRef` implementation
```
const Poco::Util::AbstractConfiguration & Context::getConfigRef() const
{
    auto lock = getLock();
    return shared->config ? *shared->config : Poco::Util::Application::instance().config();
}
```

gluten doesn't create applicaiton intance.

liuneng1994 (approved):
LGTM

PR #2247, opened 2023-07-07
[GLUTEN-2139][CORE] Add gluten event and UI

ulysses-you:

This pr adds a new module `gluten-ui` to holds the event and UI related code. So it would be more easy to put a simple gluten-ui jar instead of gluten-core jar into history server.

Adds two Gluten events:
- GlutenBuildInfoEvent, used to show build information in UI
- GlutenPlanFallbackEvent, used to show query execution fallback reason in UI

Introduce a new config `spark.gluten.ui.enabled` with enabled by default.

(Fixes: \#2139)

## How was this patch tested?
Add test and test locally

image


ulysses-you:
cc @FelixYBW @zhouyuan @yma11 can you help review this pr ? thank you

yaooqinn (approved):

zhouyuan:
@ulysses-you thank you for improving on this! I tried this patch locally and it seems working, just two questions:
-  if the backend is not providing the events, will the UI show empty results? I'm asking b/c CK backend is not tested and it may not provide these events/metrics
-  This also work on AQE enabled case, right? 

-yuan

ulysses-you:
thank you @zhouyuan for the thoughs, 

> if the backend is not providing the events, will the UI show empty results? I'm asking b/c CK backend is not tested and it may not provide these events/metrics

I think it works with ch backend. All backends use the same code path, since the main code chang is at GlutenDriverPlugin and columnar rule. It should have no difference between velox and ch backend.

> This also work on AQE enabled case, right?

yes, it's right. This pr is compatible with AQE as it follows the behavior of vanilla Spark event `SparkListenerSQLAdaptiveExecutionUpdate`.

zhouyuan (approved):
👍 
there are some TODOs but we could fix those in the next PRs.

PR #2249, opened 2023-07-07
[VL] Backport Delete shuffle spilled file directories

marin-ma:


zhouyuan:
@marin-ma it seems GHA is failing due to this change

PR #2254, opened 2023-07-10
[GLUTEN-2221][VL] Fix count(1) fallback check

rui-mo:

Fix count(1) fallback check. Fix getChildrenPlan for plan check.

## How was this patch tested?

Testing on CI.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -229,11 +229,12 @@ object BackendSettings extends BackendSettingsApi {
    */
   private def isCount1(plan: SparkPlan): Boolean = {
     plan match {
-      case exec: HashAggregateExec if exec.aggregateExpressions.forall(expression =>
-        expression.aggregateFunction.isInstanceOf[Count] &&
-          expression.aggregateFunction.asInstanceOf[Count].children.forall(child =>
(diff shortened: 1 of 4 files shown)


backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -229,7 +229,8 @@ object BackendSettings extends BackendSettingsApi {
    */
   private def isCount1(plan: SparkPlan): Boolean = {
     plan match {
-      case exec: HashAggregateExec if exec.aggregateExpressions.forall(expression =>
+      case exec: HashAggregateExec if exec.aggregateExpressions.nonEmpty &&
+        exec.aggregateExpressions.forall(expression =>
```
PHILO-HE, line 233: Can we use the blow code for this part? Assume we just expect one child for each aggregation expression and it should be Literal(1).  
```
  private def isCount1(plan: SparkPlan): Boolean = {
    plan match {
      case exec: HashAggregateExec if exec.aggregateExpressions.forall(expression =>
          expression match {
            case c: Count if c.children.size == 1 =>
              c.children.head.equals(Literal(1))
            case _ =>
              false
          }) =>
        true
      case _ =>
        false
    }
  }
```

rui-mo: Thanks! Updated.

PHILO-HE (approved):
Looks great!

PHILO-HE (approved):

PR #2286, opened 2023-07-11
[GLUTEN-2139][CORE] Add default fallback reason

ulysses-you:

The main change in this pr is:
1. Make all gluten plan inherit `GlutenPlan`, e.g., ColumnarToRowExec, RowToColumnarExec
2. Then we can add default fallback reason for all vanilla `SparkPlan`

(Fixes: \#2139)

## How was this patch tested?
add test

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/CoalesceExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/CoalesceExecTransformer.scala
@@ -17,6 +17,7 @@
 
 package io.glutenproject.execution
 
+import io.glutenproject.extension.ValidationResult
 import io.glutenproject.metrics.{MetricsUpdater, NoopMetricsUpdater}
 import io.glutenproject.substrait.SubstraitContext
 import org.apache.spark.{Partition, SparkContext, TaskContext}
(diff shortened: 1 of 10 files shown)


PHILO-HE:
Thanks so much for your efforts!
I just noted there are some fallback rules in `TransformHintRule.scala`, e.g., `FallbackOnANSIMode`. Is it possible to add fallback reason for them? I think it may be also helpful to users to understand why a query plan falls back. Thanks!

ulysses-you:
hi @PHILO-HE , it should has already supported. If you compile and rule current gluten main branch, then it will show:

image

PHILO-HE:
> hi @PHILO-HE , it should have already supported. If you compile and run current gluten main branch, then it will show:
> 
> image

Great! Thanks!

ulysses-you:
cc @zhouyuan this is for those SparkPlan which do not contain unsupport transformable tag, e.g., broadcast nested loop join

PHILO-HE:
Could you please update gluten doc to tell user how to check the fallback reason in a follow-up PR? Maybe, ending part of `Velox.md` is a good place to document it. Thanks!

ulysses-you:
@PHILO-HE sure, I will do a followup for docs

PR #2290, opened 2023-07-11
[VL] Use single parent dir for shuffle spill

marin-ma:
Port https://github.com/oap-project/gluten/pull/2249 to main. This will align the behavior of creating/deleting spill file dirs with Spark DiskBlockManager.


FelixYBW (approved):

zhouyuan (approved):
👍  works for me

PR #2302, opened 2023-07-11
[CORE] Correct and simplify stage prep rules

ulysses-you:

This pr corrects and simplify: 
- remove rule `FallbackOneRowRelation` which has been covered by `FallbackEmptySchemaRelation`
- only inject `FallbackBroadcastExchange` to stage prep rule so that we can remove the wrong `tagBeforeTransformHitsRules `

## How was this patch tested?

Pass CI

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/expression/ScalarSubqueryTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/expression/ScalarSubqueryTransformer.scala
@@ -35,7 +35,8 @@ class ScalarSubqueryTransformer(plan: BaseSubqueryExec, exprId: ExprId,
     // the first column in first row from `query`.
     val rows = query.plan.executeCollect()
     if (rows.length > 1) {
-      sys.error(s"more than one row returned by a subquery used as an expression:\n${query.plan}")
+      throw new IllegalStateException(
+        s"more than one row returned by a subquery used as an expression:\n${query.plan}")
     }
     val result: AnyRef = if (rows.length == 1) {
       assert(rows(0).numFields == 1,
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -663,9 +663,14 @@ case class ColumnarOverrideRules(session: SparkSession)
   // while creating the rules. At this time SQLConf may not be there yet.
 
   def preOverrides(): List[SparkSession => Rule[SparkPlan]] = {
+    val tagBeforeTransformHitsRules = if (this.isAdaptiveContext) {
+      // When AQE is supported, rules are applied in ColumnarQueryStagePrepOverrides
+      List.empty
+    } else {
+      TagBeforeTransformHits.ruleBuilders
+    }
+    tagBeforeTransformHitsRules :::
     List(
-      (spark: SparkSession) => FallbackOnANSIMode(spark),
-      (spark: SparkSession) => FallbackMultiCodegens(spark),
       (spark: SparkSession) => PlanOneRowRelation(spark),
       (_: SparkSession) => FallbackEmptySchemaRelation(),
       (_: SparkSession) => AddTransformHintRule(),
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -663,14 +663,9 @@ case class ColumnarOverrideRules(session: SparkSession)
   // while creating the rules. At this time SQLConf may not be there yet.
 
   def preOverrides(): List[SparkSession => Rule[SparkPlan]] = {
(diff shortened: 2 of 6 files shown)


gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -663,14 +663,9 @@ case class ColumnarOverrideRules(session: SparkSession)
   // while creating the rules. At this time SQLConf may not be there yet.
 
   def preOverrides(): List[SparkSession => Rule[SparkPlan]] = {
-    val tagBeforeTransformHitsRules = if (this.isAdaptiveContext) {
-      TagBeforeTransformHits.ruleBuilders
-    } else {
-      // When AQE is supported, rules are applied in ColumnarQueryStagePrepOverrides
-      List.empty
```
ulysses-you, line 670: this is wrong, if isAdaptiveContext is true that means we are in AQE...

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarQueryStagePrepOverrides.scala
```diff
@@ -89,7 +89,6 @@ case class FallbackBroadcastExchange(session: SparkSession) extends Rule[SparkPl
 
 object ColumnarQueryStagePrepOverrides extends GlutenSparkExtensionsInjector {
   override def inject(extensions: SparkSessionExtensions): Unit = {
-    val builders = TagBeforeTransformHits.ruleBuilders :+ FallbackBroadcastExchange
-    builders.foreach(extensions.injectQueryStagePrepRule)
+    extensions.injectQueryStagePrepRule(FallbackBroadcastExchange)
```
ulysses-you, line 92: I think the cost of stagePrepRule is bigger than columnarRule because AQE re-optimize would produce some ineffective results and fallback to original plan. So we can avoid applying unnecessary rules after moving to columnarRule.

gluten-core/src/main/scala/io/glutenproject/expression/ScalarSubqueryTransformer.scala
```diff
@@ -35,7 +35,8 @@ class ScalarSubqueryTransformer(plan: BaseSubqueryExec, exprId: ExprId,
     // the first column in first row from `query`.
     val rows = query.plan.executeCollect()
     if (rows.length > 1) {
-      sys.error(s"more than one row returned by a subquery used as an expression:\n${query.plan}")
+      throw new IllegalStateException(
+        s"more than one row returned by a subquery used as an expression:\n${query.plan}")
```
ulysses-you, line 39: Vanilla Spark throw IllegalStateException but `sys.error` throw RuntimeException

ulysses-you: This changes is due to this pr removes `FallbackOneRowRelation`, queries are offlaoding to Gluten in some tests.

ulysses-you:
cc @marin-ma @PHILO-HE @zhouyuan do you have time to take a look ? thank you

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
marin-ma: Oops but nice catch!

marin-ma:
When AQE is on, there's one change look suspicious to me:
Before: QueryStagePreparation will inject FallbackOnANSIMode, FallbackMultiCodegens, FallbackBroadcastExchange  => divide QueryStages => apply the rest columnar rules for each individual QueryStage
After: QueryStagePreparation will inject FallbackBroadcastExchange => AQE divide QueryStages => apply the rest columnar rules for each individual QueryStage(including FallbackOnANSIMode, FallbackMultiCodegens)

If we apply the two rules, which may let operators fallback at runtime, after dividing the QueryStages, below situation that contains BHJ might happen:

QueryStage0 finish execution, the last operation is ColumnarBroadcastExchange.
QueryStage1 start to apply columnar rules, but BroadcastJoinExec fallback due some reasons.
QueryStage1 start to execute, then it fails because BroadcastJoinExec cannot take the output from ColumnarBroadcastExchange.

ulysses-you:
Good point, I missed this part. Rule `FallbackMultiCodegens` could hit your case after moving to columnar rules.

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/extension/GlutenSessionExtensionSuite.scala
```diff
@@ -31,6 +32,7 @@ class GlutenSessionExtensionSuite extends GlutenSQLTestsTrait {
   }
 
   test("test gluten extensions") {
+    assert(spark.sessionState.queryStagePrepRules.contains(FallbackMultiCodegens(spark)))
```
marin-ma, line 35: Also check FallbackOnANSIMode?

ulysses-you: sure, added

marin-ma (approved):

PR #2303, opened 2023-07-11
[GLUTEN-1626][CH] make round func act in the same way as vanilla spark

lhuang09287750:

Add a new ch roundHalfUp func to implement rounding away from zero (towards +inf/-inf) for float and make gluten round func mapping to it.
No suiltable SSE instructions to realize a fast version func, so use std round func instead.
Test round and roundHalfUp with local ch server,
use
select roundHalfUp(x) as round_half_up from system.numbers_mt limit 10000;
select round(x) as round from system.numbers_mt limit 10000;
performance difference is very little

(Fixes: \#1626)

## How was this patch tested?

GlutenClickHouseTPCDSParquetSuite  Q78
GlutenClickHouseTPCDSParquetSuite  GLUTEN-1626: test 'roundHalfup'
clickhouse server response


liuneng1994 (approved):
LGTM

Issue #2217, opened 2023-07-05
[CH] Introduce Signal Handler to print stack info when get `SIGSEGV`


baibaichen:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

In Java, we have the `NullPointerException` to help us identify errors, but in C++, we lack a similar mechanism. While we have a code dump that can help us locate issues, it is too heavy to be of much use.

**Describe the solution you'd like**
A Signal handler is needed to catch `SEGSEGV`, and print stack

PR #2314, opened 2023-07-12
[GLUTEN-2217][CH] Introduce Signal Handler to print stack info when get `SIGSEGV`

baibaichen:

(Fixes: \#2217)

## How was this patch tested?

using exited UTs

# Design

Experimental feature.

> Codes are copied from `BaseDameon`, since I believe it's difficult to merge into the main branch.

The idea is very simple, register a single handler via [Signal Chaining](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/signals003.html), when jvm failed to handle signal, it will call ours. We create a pipe during initialization, and in our Signal handler, we collect stack info, and then write those info into pipe. ​We also create a Signal Listener to read from that pipe, and dump the stack info. 

In most cases, the method works as I tried out. However, if the original SIGSEGV issue results from a corrupted stack, such as a JNI [stack overflow](https://pangin.pro/posts/stack-overflow-handling#shadow-pages), our signal handler won't be able to work. I spent a lot of time figuring out why, but I made no progress.

For example, calling the following function in the JNI could easily reproduce the case.

```c
void deepNative64K(Poco::Logger * log, int depth = 10000) {
    char BUF[256*1024];
    memset(BUF, 0x5D5D5D5DL, sizeof(BUF));
    LOG_WARNING(log, "thread({}), stack grows {} kbytes", getThreadId(), (sizeof(BUF) * (10001-depth))/1024);
    if (depth > 0) {
        deepNative64K(log, depth - 1);
    }
}
```

# How to use 

```shell
export LD_PRELOAD=libjvm.so-directory/libjsig.so:libch.so
```

### DIFF
--- a/cpp-ch/local-engine/Common/GlutenSignalHandler.cpp
+++ b/cpp-ch/local-engine/Common/GlutenSignalHandler.cpp
@@ -371,7 +371,7 @@ struct SignalHandler::Impl
     /// initialize termination process and signal handlers
     void initializeTerminationAndSignalProcessing()
     {
-        addSignalHandler({SIGABRT, SIGSEGV, SIGBUS, SIGTSTP}, signalHandler, &handled_signals);
+        addSignalHandler({SIGABRT, SIGSEGV, SIGILL, SIGBUS, SIGSYS, SIGFPE, SIGPIPE, SIGTSTP, SIGTRAP}, signalHandler, &handled_signals);
 
         /// TODO:: Set up Poco ErrorHandler for Poco Threads.
         // static KillingErrorHandler killing_error_handler;
--- a/cpp-ch/local-engine/Common/GlutenSignalHandler.cpp
+++ b/cpp-ch/local-engine/Common/GlutenSignalHandler.cpp
@@ -1,0 +1,422 @@
+#include <cstring>
+#include <vector>
+#include <IO/ReadBufferFromFileDescriptor.h>
+#include <IO/ReadHelpers.h>
+#include <IO/WriteBufferFromFileDescriptorDiscardOnFailure.h>
+#include <IO/WriteHelpers.h>
+#include <Interpreters/Context.h>
+#include <base/getThreadId.h>
+#include <base/phdr_cache.h>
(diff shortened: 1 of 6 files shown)


cpp-ch/local-engine/Common/GlutenSignalHandler.cpp
```diff
@@ -0,0 +1,422 @@
...
+{
+    /// A thread that acts on HUP and USR1 signal (close logs).
+    Poco::Thread signal_listener_thread;
+    std::unique_ptr<Poco::Runnable> signal_listener;
+    std::vector<int> handled_signals;
+
+    /// initialize termination process and signal handlers
+    void initializeTerminationAndSignalProcessing()
+    {
+        addSignalHandler({SIGABRT, SIGSEGV, SIGBUS, SIGTSTP}, signalHandler, &handled_signals);
```
zhanglistar, line 374: According this, https://man7.org/linux/man-pages/man7/signal.7.html, some signals causing core should also be handled.

cpp-ch/local-engine/Common/GlutenSignalHandler.cpp
```diff
@@ -0,0 +1,422 @@
...
+                    readPODBinary(context, in);
+                }
+
+                readPODBinary(stack_trace, in);
+                readBinary(thread_num, in);
+                readPODBinary(thread_ptr, in);
+
+                /// This allows to receive more signals if failure happens inside onFault function.
+                /// Example: segfault while symbolizing stack trace.
+                std::thread([sig, info, context, stack_trace, thread_num, thread_ptr, this]
```
zhanglistar, line 206: 这里会不会起比较多的线程？

baibaichen: not a problem. those threads are deteached, once one signal is handled,  the process will be terninated.

zhanglistar:
LGTM

liuneng1994 (approved):
LGTM

PR #2321, opened 2023-07-13
[GLUTEN-2404][VL] Enable Velox memory reclaimer for auto disk-spilling

zhztheplayer:
Issue link #2404 

Rely on https://github.com/oap-project/velox/pull/362

Update:

Still not sure between the 2 different approaches:

1. Use `threshold` and shared arbitrator to do cap-based-spilling
2. Use `auto` and memory reclaimer to do auto-spilling

The PR is now updated with inclination to **approach 2**.

### DIFF
--- a/cpp/core/memory/MemoryAllocator.cc
+++ b/cpp/core/memory/MemoryAllocator.cc
@@ -100,18 +100,16 @@ bool ListenableMemoryAllocator::free(void* p, int64_t size) {
   return succeed;
 }
 
-bool ListenableMemoryAllocator::reserveBytes(int64_t size) {
-  listener_->allocationChanged(size);
-  return true;
-}
-
-bool ListenableMemoryAllocator::unreserveBytes(int64_t size) {
-  listener_->allocationChanged(-size);
-  return true;
-}
-
 int64_t ListenableMemoryAllocator::getBytes() const {
   return bytes_;
+}
+
+MemoryAllocator* ListenableMemoryAllocator::delegatedAllocator() {
+  return delegated_;
+}
+
+AllocationListener* ListenableMemoryAllocator::listener() {
+  return listener_.get();
 }
 
 bool StdMemoryAllocator::allocate(int64_t size, void** out) {
@@ -166,19 +164,21 @@ bool StdMemoryAllocator::free(void* p, int64_t size) {
   return true;
 }
 
-bool StdMemoryAllocator::reserveBytes(int64_t size) {
-  bytes_ += size;
-  return true;
-}
-
-bool StdMemoryAllocator::unreserveBytes(int64_t size) {
-  bytes_ -= size;
-  return true;
-}
-
 int64_t StdMemoryAllocator::getBytes() const {
   return bytes_;
 }
+
+class NoopAllocationListener : public gluten::AllocationListener {
+ public:
+  void allocationChanged(int64_t diff) override {
+    // no-op
+  }
+};
+
+AllocationListener* AllocationListener::noop() {
+  static AllocationListener* instance = new NoopAllocationListener();
+  return instance;
+};
 
 std::shared_ptr<MemoryAllocator> defaultMemoryAllocator() {
 #if defined(GLUTEN_ENABLE_HBM)
--- a/cpp/core/utils/TaskContext.cc
+++ b/cpp/core/utils/TaskContext.cc
@@ -40,18 +40,18 @@ class TaskContextStorage {
 };
 
 thread_local std::unique_ptr<TaskContextStorage> taskContextStorage = nullptr;
-static std::unique_ptr<TaskContextStorage> fallbackStorage = std::make_unique<TaskContextStorage>();
-static std::mutex fallbackStorageMutex;
+std::unique_ptr<TaskContextStorage> fallbackStorage = std::make_unique<TaskContextStorage>();
+std::mutex fallbackStorageMutex;
 } // namespace
 
 namespace gluten {
 
-bool onSparkTaskMainThread() {
+bool isOnSparkTaskMainThread() {
   return taskContextStorage != nullptr;
 }
 
 void bindToTask(std::shared_ptr<void> object) {
-  if (onSparkTaskMainThread()) {
+  if (isOnSparkTaskMainThread()) {
     taskContextStorage->bind(object);
     return;
   }
--- a/cpp/core/utils/TaskContext.cc
+++ b/cpp/core/utils/TaskContext.cc
@@ -19,20 +19,49 @@
 #include <folly/Likely.h>
 #include "utils/exception.h"
 
-namespace gluten {
+#include <iostream>
+#include <mutex>
+
+namespace {
+class TaskContextStorage {
+ public:
+  virtual ~TaskContextStorage() {
+    for (auto itr = objects_.rbegin(); itr != objects_.rend(); itr++) {
+      itr->reset();
+    }
+  }
+
+  void bind(std::shared_ptr<void> object) {
+    objects_.push_back(object);
+  }
+
+ private:
+  std::vector<std::shared_ptr<void>> objects_;
+};
 
 thread_local std::unique_ptr<TaskContextStorage> taskContextStorage = nullptr;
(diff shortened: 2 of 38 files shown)


cpp/velox/memory/VeloxMemoryPool.cc
```diff
@@ -122,91 +95,126 @@ class VeloxMemoryAllocator final : public velox::memory::MemoryAllocator {
...
-    defaultPoolRoot->setHighUsageCallback(
-        [=](velox::memory::MemoryPool& pool) { return pool.reservedBytes() >= spillThreshold; });
-    defaultPoolRoot->testingSetCapacity(facebook::velox::memory::kMaxMemory);
-    return true;
-  };
+  gluten::AllocationListener* listener_;
+};
 
-  static bool do_once = f();
-  (void)do_once; // suppress warning
```
zuochunwei, line 191: why not use `MemoryManager::getInstance()` instead

cpp/core/utils/TaskContext.cc
```diff
@@ -19,20 +19,49 @@
...
+  void bind(std::shared_ptr<void> object) {
+    objects_.push_back(object);
+  }
+
+ private:
+  std::vector<std::shared_ptr<void>> objects_;
+};
 
 thread_local std::unique_ptr<TaskContextStorage> taskContextStorage = nullptr;
+static std::unique_ptr<TaskContextStorage> fallbackStorage = std::make_unique<TaskContextStorage>();
```
zuochunwei, line 43: variables under the anonymous namespace need not be declared with `static`.
the anonymous namespace is used to replace `static` when defining a variable in cpp.

cpp/core/utils/TaskContext.cc
```diff
@@ -19,20 +19,49 @@
...
+static std::mutex fallbackStorageMutex;
+} // namespace
 
-void bindToTask(std::shared_ptr<void> object) {
-  GLUTEN_CHECK(taskContextStorage != nullptr, "Not in a Spark task");
-  taskContextStorage->bind(object);
+namespace gluten {
+
+bool onSparkTaskMainThread() {
```
zuochunwei, line 49: use `isXyz()` to represent checking meaning?

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -166,20 +164,22 @@ bool StdMemoryAllocator::free(void* p, int64_t size) {
...
-bool StdMemoryAllocator::unreserveBytes(int64_t size) {
-  bytes_ -= size;
-  return true;
-}
-
 int64_t StdMemoryAllocator::getBytes() const {
   return bytes_;
 }
 
+class NoopAllocationListener : public gluten::AllocationListener {
```
zuochunwei, line 171: There is a thinking that we should only derive from abstract classes in cpp programming.
You can get some information from google, and then make a decision. whatever you do, it will be ok for me.

cpp/core/utils/TaskContext.cc
```diff
@@ -19,20 +19,49 @@
...
 
-namespace gluten {
+#include <iostream>
+#include <mutex>
+
+namespace {
+class TaskContextStorage {
+ public:
+  virtual ~TaskContextStorage() {
+    for (auto itr = objects_.rbegin(); itr != objects_.rend(); itr++) {
```
zuochunwei, line 29: do we need this for the loop?
it will be destroyed auto after TaskContextStorage deconstructing.

cpp/velox/memory/VeloxMemoryPool.cc
```diff
@@ -122,91 +95,126 @@ class VeloxMemoryAllocator final : public velox::memory::MemoryAllocator {
...
-  auto pool = rootVeloxMemoryPool()->addAggregateChild(
-      "wrapped_root" + std::to_string(id++), facebook::velox::memory::MemoryReclaimer::create());
-  auto wrapped = std::dynamic_pointer_cast<VeloxMemoryPool>(pool);
-  VELOX_CHECK_NOT_NULL(wrapped);
-  velox::memory::MemoryAllocator* veloxAlloc = wrapped->getAllocator();
-  wrapped->setAllocatorShared(std::make_shared<VeloxMemoryAllocator>(allocator, veloxAlloc));
+  velox::memory::MemoryAllocator* veloxAlloc = velox::memory::MemoryAllocator::getInstance();
+  gluten::MemoryAllocator* glutenAlloc;
+  gluten::AllocationListener* listener;
+  if (dynamic_cast<gluten::ListenableMemoryAllocator*>(allocator)) {
```
zuochunwei, line 186: consider adding an interface `type()` for `gluten::MemoryAllocator` to tell `what is it`

cpp/velox/memory/VeloxMemoryPool.cc
zhztheplayer: I'm OK to either using this approach or `MemoryManager::getInstance()`.

cpp/core/utils/TaskContext.cc
zhztheplayer: Sure.

cpp/core/utils/TaskContext.cc
zhztheplayer: OK.

cpp/core/memory/MemoryAllocator.cc
zhztheplayer: Wasn't `gluten::AllocationListener` already an abstract class? The code is

```cpp
class AllocationListener {
 public:
  static AllocationListener* noop();

  virtual ~AllocationListener() = default;

  // Value of diff can be either positive or negative
  virtual void allocationChanged(int64_t diff) = 0;

 protected:
  AllocationListener() = default;
};
```

cpp/core/utils/TaskContext.cc
zhztheplayer: The loop has to be here because we expect the objects to be destroyed in reversed order. Seems to me that `std::vector`'s destructor doesn't guarantee that.

cpp/velox/memory/VeloxMemoryPool.cc
zhztheplayer: Did you mean a `enum` or `std::string` type?

zuochunwei: yes

cpp/core/utils/TaskContext.cc
zuochunwei: OK

cpp/core/memory/MemoryAllocator.cc
zuochunwei: OK

zuochunwei (dismissed):

marin-ma (approved):

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -101,9 +101,44 @@ std::shared_ptr<ColumnarBatch> WholeStageResultIterator::next() {
...
 
 int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
+  std::string poolName{pool_->root()->name() + "/" + pool_->name()};
+  std::string logPrefix{"Spill[" + poolName + "]: "};
+  LOG(INFO) << logPrefix << "Trying to reclaim " << size << " bytes of data...";
+  LOG(INFO) << logPrefix << "Pool has reserved " << pool_->currentBytes() << "/" << pool_->root()->reservedBytes()
+            << "/" << pool_->root()->capacity() << "/" << pool_->root()->maxCapacity() << " bytes.";
+  LOG(INFO) << logPrefix << "Shrinking...";
+  int64_t shrunk = pool_->shrinkManaged(pool_.get(), size);
+  LOG(INFO) << logPrefix << shrunk << " bytes released from shrinking.";
```
liujiayi771, line 111: @zhztheplayer Should we remove this log, or change log level.

zhztheplayer: Thanks for the suggestion.

I think we can at least use debug for shrinking log ?

Or for reclaiming (spilling) log as well?

PR #2345, opened 2023-07-14
[GLUTEN-2193][CH] Support functions array_max/array_min

exmy:
CH arrayMax is different from Spark array_max, since it cannot handle Array(String) type argument.

But we can map Spark array_max(arr) to CH arrayReduce("max", arr)

## How was this patch tested?

UT


baibaichen (approved):
LGTM

baibaichen:
没看到 spark 33 的UT

PR #2351, opened 2023-07-15
[GLUTEN-2198][CH] Fix wrong schema when there is no aggregate function

### LINKED ISSUE #2198: [CH] Fix wrong schema when there is no aggregate function
**Describe the bug**
In spark, the output of an operator may be different with its input. For example: 
image
The columns of HashAggregateTransformer are in different order with ProjectExecTransformer, this may cause two problems in gluten, 1: downsteam may can not read the input block as in gluten it suppose the schema should be same. 2: HashAggregatekeys may be wrong, as it use the children's output.

zzcclp:
Fix wrong schema when there is no aggregate function

Close #2198.

Fixes: #2198 

## How was this patch tested?


zzcclp:
@KevinyhZou @shuai-xu please help to review.

liuneng1994 (approved):
LGTM

baibaichen (approved):
LGTM

PR #2352, opened 2023-07-15
[CORE] Minor refactor for doValidate logic and related config

Yohahaha:

Unify spark.gluten.sql.enable.native.validation usage, it's only used by GlutenPlan#doNativeValidation now, and user could use it to decide whether validate substrait plan in native side or not.

Rename `spark.gluten.sql.validate.failure.logLevel` to `spark.gluten.sql.validation.logLevel`, `spark.gluten.sql.validate.failure.printStack` to `spark.gluten.sql.validation.printStackOnFailure`

Fix some related class typo and lots of log content.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxValidator.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxValidator.scala
@@ -27,7 +27,7 @@ import io.glutenproject.vectorized.NativePlanEvaluator
 import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.execution.SparkPlan
 
-class VeloxValidator extends ValidatorApi {
+class Validator extends ValidatorApi {
 
   override def doExprValidate(substraitExprName: String, expr: Expression): Boolean =
     doExprValidate(Map(), substraitExprName, expr)
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxValidator.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxValidator.scala
@@ -21,13 +21,13 @@ import io.glutenproject.backendsapi.ValidatorApi
 
 import org.apache.spark.sql.types.{ArrayType, BinaryType, BooleanType, ByteType, DataType, DateType, DecimalType, DoubleType, FloatType, IntegerType, LongType, MapType, ShortType, StringType, StructType, TimestampType}
 import io.glutenproject.substrait.plan.PlanNode
-import io.glutenproject.validate.NativePlanValidatorInfo
+import io.glutenproject.validate.NativePlanValidationInfo
 import io.glutenproject.vectorized.NativePlanEvaluator
 
 import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.execution.SparkPlan
 
-class Validator extends ValidatorApi {
+class VeloxValidator extends ValidatorApi {
 
   override def doExprValidate(substraitExprName: String, expr: Expression): Boolean =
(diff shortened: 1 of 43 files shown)


Yohahaha:
Hi @rui-mo @zhouyuan @PHILO-HE  would you help take a look? thanks!

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxValidator.scala
```diff
@@ -21,13 +21,13 @@ import io.glutenproject.backendsapi.ValidatorApi
...
 import io.glutenproject.substrait.plan.PlanNode
-import io.glutenproject.validate.NativePlanValidatorInfo
+import io.glutenproject.validate.NativePlanValidationInfo
 import io.glutenproject.vectorized.NativePlanEvaluator
 
 import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.execution.SparkPlan
 
-class Validator extends ValidatorApi {
+class VeloxValidator extends ValidatorApi {
```
rui-mo, line 31: Why do we need this change? This class exists in backends-velox folder, so the scope should be apparant.

Yohahaha: No obvious reason for this, could revert if you insisted.

rui-mo: Let's revert this change. We need to avoid duplicate prefix.

gluten-core/src/main/scala/io/glutenproject/extension/GlutenPlan.scala
```diff
@@ -52,41 +67,45 @@ trait GlutenPlan extends SparkPlan with LogLevelUtil {
...
 
-  protected def logValidateFailure(msg: => String, e: Throwable): Unit = {
-    if (printStackOnValidateFailure) {
-      logOnLevel(validateFailureLogLevel, msg, e)
+  protected def doNativeValidation(context: SubstraitContext, node: RelNode): ValidationResult = {
+    if (node != null && enableNativeValidation) {
+      val planNode = PlanBuilder.makePlan(context, Lists.newArrayList(node))
+      // TODO Unify doValidate interface, remove this branch
+      if (BackendsApiManager.chBackend &&
+        BackendsApiManager.getValidatorApiInstance.doValidate(planNode)) {
```
PHILO-HE, line 91: It may be better to move `BackendsApiManager.getValidatorApiInstance.doValidate(planNode)` into this `if` branch and return ok or notOk accordingly. Right?

gluten-core/src/main/scala/io/glutenproject/extension/GlutenPlan.scala
```diff
@@ -52,41 +67,45 @@ trait GlutenPlan extends SparkPlan with LogLevelUtil {
...
+        ValidationResult.ok
+      } else if (BackendsApiManager.veloxBackend) {
+        val info = BackendsApiManager.getValidatorApiInstance.doValidateWithFallBackLog(planNode)
+        ValidationResult.convertFromValidationInfo(info)
+      } else {
+        ValidationResult.notOk("Native backend check failed.")
```
PHILO-HE, line 97: I think we can make this else branch just cover exception cases, i.e, neither velox backend nor CH backend is used. Maybe, just throw an exception here. Right?

Yohahaha: > neither velox backend nor CH backend is used

It would be a fatal error I think, and should not be covered here.

PHILO-HE: OK to me.

Yohahaha:
Thanks for merging!

PR #2356, opened 2023-07-17
[CORE] Correct guava version

ulysses-you:

It seems the only place in Gluten to use guava is `substrait-spark`, which scope is `provided`.  There is no strong reason to shade it (otherwise it would make jar size bigger), so this pr just removes it.

gluten-it is a sub-project which depends on high guava version due to Trino tpcds/tpch. This pr shades them.

## How was this patch tested?

Pass CI

### DIFF
--- a/tools/gluten-it/common/pom.xml
+++ b/tools/gluten-it/common/pom.xml
@@ -16,18 +16,6 @@
 
   <dependencies>
     <dependency>
-      <groupId>io.glutenproject</groupId>
-      <artifactId>gluten-package</artifactId>
-      <version>${gluten.version}</version>
-      <scope>runtime</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>io.glutenproject</groupId>
-          <artifactId>gluten-core</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
       <groupId>org.scala-lang</groupId>
       <artifactId>scala-library</artifactId>
       <version>${scala.library.version}</version>
--- a/tools/gluten-it/common/pom.xml
+++ b/tools/gluten-it/common/pom.xml
@@ -14,72 +14,6 @@
   <version>1.1.0-SNAPSHOT</version>
   <packaging>jar</packaging>
 
-  <properties>
-    <maven.compiler.source>8</maven.compiler.source>
-    <maven.compiler.target>8</maven.compiler.target>
-    <scala.library.version>2.12.15</scala.library.version>
-    <spark32.version>3.2.2</spark32.version>
-    <spark33.version>3.3.1</spark33.version>
-    <scala.binary.version>2.12</scala.binary.version>
-    <gluten.version>1.1.0-SNAPSHOT</gluten.version>
-  </properties>
-
-  <profiles>
-    <profile>
-      <id>spark-3.2</id>
-      <activation>
-        <activeByDefault>true</activeByDefault>
-      </activation>
-      <properties>
-        <spark.version>${spark32.version}</spark.version>
-      </properties>
-      <dependencies>
-        <dependency>
-          <groupId>org.apache.spark</groupId>
-          <artifactId>spark-sql_${scala.binary.version}</artifactId>
-          <version>${spark.version}</version>
-          <exclusions>
-            <exclusion>
-              <groupId>com.google.protobuf</groupId>
-              <artifactId>protobuf-java</artifactId>
-            </exclusion>
-          </exclusions>
-        </dependency>
-        <dependency>
-          <groupId>org.apache.spark</groupId>
-          <artifactId>spark-sql_${scala.binary.version}</artifactId>
-          <version>${spark.version}</version>
-          <type>test-jar</type>
-        </dependency>
-      </dependencies>
-    </profile>
-    <profile>
-      <id>spark-3.3</id>
-      <properties>
-        <spark.version>${spark33.version}</spark.version>
-      </properties>
-      <dependencies>
-        <dependency>
-          <groupId>org.apache.spark</groupId>
-          <artifactId>spark-sql_${scala.binary.version}</artifactId>
-          <version>${spark.version}</version>
-          <exclusions>
-            <exclusion>
-              <groupId>com.google.protobuf</groupId>
-              <artifactId>protobuf-java</artifactId>
-            </exclusion>
-          </exclusions>
-        </dependency>
-        <dependency>
-          <groupId>org.apache.spark</groupId>
-          <artifactId>spark-sql_${scala.binary.version}</artifactId>
-          <version>${spark.version}</version>
-          <type>test-jar</type>
-        </dependency>
-      </dependencies>
-    </profile>
-  </profiles>
-
   <dependencies>
     <dependency>
       <groupId>io.glutenproject</groupId>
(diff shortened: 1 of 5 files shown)


ulysses-you:
cc @zhouyuan thank you

zhouyuan:
@ulysses-you 
The issue is due to several CVEs found in 14.0.1, so I bumped it to more recent release
https://mvnrepository.com/artifact/com.google.guava/guava/14.0.1

I just did a quick check, seems guava is not necessary now. Can you please help to check if this could completely removed?

```
diff --git a/pom.xml b/pom.xml
index 7b86ffa38..5aaaf785f 100644
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,6 @@
     <junit.version>4.13.1</junit.version>

     <substrait.version>0.5.0</substrait.version>
-    <guava.version>32.0.1-jre</guava.version>
     <protobuf.version>3.16.3</protobuf.version>
     <protobuf.substrait.version>3.21.7</protobuf.substrait.version>
     <!--spotless-->
@@ -328,12 +327,6 @@
         <version>${spark.version}</version>
         <scope>provided</scope>
       </dependency>
-      <dependency>
-        <groupId>com.google.guava</groupId>
-        <artifactId>guava</artifactId>
-        <version>${guava.version}</version>
-        <scope>provided</scope>
-      </dependency>
       <dependency>
         <groupId>io.substrait</groupId>
         <artifactId>core</artifactId>
diff --git a/substrait/substrait-spark/pom.xml b/substrait/substrait-spark/pom.xml
index e6d3ff53d..c6d32130d 100644
--- a/substrait/substrait-spark/pom.xml
+++ b/substrait/substrait-spark/pom.xml
@@ -35,10 +35,6 @@
       <artifactId>hadoop-client</artifactId>
     </dependency>

-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-    </dependency>
     <dependency>
       <groupId>io.substrait</groupId>
       <artifactId>core</artifactId>

```

ulysses-you:
I see, let me check if we really need guava

tools/gluten-it/common/pom.xml
```diff
@@ -186,6 +193,40 @@
...
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>3.4.1</version>
+        <configuration>
+          <shadedArtifactAttached>false</shadedArtifactAttached>
+          <artifactSet>
+            <includes>
+              <include>io.trino.tpcds:tpcds</include>
+              <include>io.trino.tpch:tpch</include>
+              <include>com.google.guava:guava</include>
+              <include>io.glutenproject:gluten-it-common</include>
```
ulysses-you, line 208: I'm not sure how it works before. IIUC, Trino tpcds/tpch depend on high version guava which can be conflict with Spark guava... This pr shade them to make sure it's right.

tools/gluten-it/common/pom.xml
```diff
@@ -14,16 +14,6 @@
...
 
-  <properties>
-    <maven.compiler.source>8</maven.compiler.source>
-    <maven.compiler.target>8</maven.compiler.target>
-    <scala.library.version>2.12.15</scala.library.version>
-    <spark32.version>3.2.2</spark32.version>
-    <spark33.version>3.3.1</spark33.version>
-    <scala.binary.version>2.12</scala.binary.version>
-    <gluten.version>1.1.0-SNAPSHOT</gluten.version>
-  </properties>
```
ulysses-you, line 25: it has been defined in parent.pom

tools/gluten-it/common/pom.xml
```diff
@@ -14,72 +14,6 @@
...
-
-  <profiles>
```
ulysses-you, line 27: moved to parent

tools/gluten-it/common/pom.xml
zhouyuan: @zhztheplayer could you help to take a look here?

zhztheplayer: will take a look asap

tools/gluten-it/package/pom.xml
```diff
@@ -17,8 +17,50 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-core_${scala.binary.version}</artifactId>
+      <type>test-jar</type>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-catalyst_${scala.binary.version}</artifactId>
+      <type>test-jar</type>
```
zhztheplayer, line 60: Any reason to add the dependencies?

zhztheplayer: In case of possible mismatching between our understandings on the module relationship of `gluten-it`, let's clarify it:

1. `gluten-it-common`: This module includes all the jars needed by gluten-it
2. `gluten-it-package`: This module only copies the jars to `target/libs`

Since `gluten-it` doesn't use a fat jar to do its jobs, is this PR going to put a shaded version of guava to `target/libs`?

If the answer is yes, the PR makes sense to me.

By the way, would you like to share the error log from `gluten-it` about this issue from your environment?

ulysses-you: thank you for the clarification! yes, that's what this pr did.

The gluten-it lib after this pr:
image

The error msg:
```
Successfully switched to SessionDesc(SessionToken(baseline),Data Gen) session. 
Generating table lineitem...
23/07/17 17:30:16 ERROR Executor: Exception in task 0.0 in stage 0.0 (TID 0)
java.lang.IncompatibleClassChangeError: Class com.google.common.base.Suppliers$NonSerializableMemoizingSupplier does not implement the requested interface java.util.function.Supplier
	at io.trino.tpch.Distributions.getDefaultDistributions(Distributions.java:48)
	at io.trino.tpch.LineItemGenerator.<init>(LineItemGenerator.java:62)
	at io.glutenproject.integration.tpc.h.TpchDataGen.$anonfun$lineItemGenerator$1(TpchDataGen.scala:48)
	at io.glutenproject.integration.tpc.h.TpchDataGen.$anonfun$lineItemGenerator$1$adapted(TpchDataGen.scala:47)
	at io.glutenproject.integration.tpc.h.TpchDataGen.$anonfun$generate$2(TpchDataGen.scala:310)
	at org.apache.spark.sql.execution.MapPartitionsExec.$anonfun$doExecute$3(objects.scala:201)
```

zhztheplayer: @ulysses-you Thanks for providing the information.

However can we still put all jars to `target/lib` rather than generating a single one? Since it's preferred not to touch the jar we are going to test against, which is `gluten-package-xxx.jar` in this case.

Will it work to exclude guava jar in trino jars from gluten-it directly?

ulysses-you: > However can we still put all jars to target/lib rather than generating a single one?

This pr just shades gluten-it-common.jar/trino-tpcds/trino-tpcd/guava into one jar. Others are same as before.
image

> Will it work to exclude guava jar in trino jars from gluten-it directly?

I'm not sure, but I think it's risky whatever it can work or not..

zhztheplayer: Thanks for clarifying, so the lib folder still includes all the jars separately. That makes sense to me.

In addition, if using this approve, could you please try moving dependency 'gluten-package' from `gluten-it-common` to `gluten-it-package` too? So that we have all the jars we are testing against made as provided/runtime dependency of `gluten-it-package`.

tools/gluten-it/pom.xml
```diff
@@ -17,9 +17,118 @@
...
+    <tpch.version>1.1</tpch.version>
+    <tpcds.version>1.4</tpcds.version>
   </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.spark</groupId>
+        <artifactId>spark-core_${scala.binary.version}</artifactId>
+        <version>${spark.version}</version>
```
zhztheplayer, line 35: Can we use `runtime` scope here? The same for other Spark jars.

tools/gluten-it/package/pom.xml
ulysses-you: sure, addressed

zhztheplayer (approved):
Thanks @ulysses-you !

PR #2364, opened 2023-07-17
[VL] Correct hive write table

ulysses-you:

This pr add some additional vanilla Spark config check if we can safely convert hive write to velox write.
- HiveUtils.CONVERT_METASTORE_PARQUET
- HiveUtils.CONVERT_INSERTING_PARTITIONED_TABLE

## How was this patch tested?

Add tests

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala
@@ -42,16 +42,13 @@ import org.apache.spark.sql.errors.QueryExecutionErrors
 import org.apache.spark.sql.execution.datasources.velox.VeloxParquetFileFormat
 import org.apache.spark.sql.execution.datasources._
 import org.apache.spark.sql.hive.HiveShim.{ShimFileSinkDesc => FileSinkDesc}
-import org.apache.spark.sql.hive.{HiveInspectors, HiveTableUtil}
+import org.apache.spark.sql.hive.{HiveInspectors, HiveTableUtil, HiveUtils}
 import org.apache.spark.sql.internal.SQLConf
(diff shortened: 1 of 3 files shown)


ulysses-you:
cc @JkSelf

backends-velox/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala
```diff
@@ -77,6 +74,19 @@ class HiveFileFormat(fileSinkConf: FileSinkDesc)
...
+  /**
+   * Gluten only supports convert [[InsertIntoHiveDirCommand]], see [[GlutenColumnarRules]].
+   */
+  private def isGlutenHiveWrite(sparkSession: SparkSession): Boolean = {
+    fileSinkConf.tableInfo
+      .getOutputFileFormatClassName()
+      .equals("org.apache.hadoop.hive.ql.io.parquet.MapredParquetOutputFormat") &&
+      sparkSession.sparkContext.conf.getOption("spark.plugins")
+        .contains("io.glutenproject.GlutenPlugin") &&
+      sparkSession.sessionState.conf.getConf(HiveUtils.CONVERT_METASTORE_PARQUET) &&
```
JkSelf, line 86: Do we need to add these configurations check [here](https://github.com/oap-project/gluten/blob/116d3dcdbec819d3fb06514bdd0a3ed03986652a/gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenColumnarRules.scala#L107)?

ulysses-you: The relation conversion is at post resolution hoc, which is always been applied before columnar rules. So we won't hit the right pattern if the config is unexpected.

JkSelf (approved):
LGTM

PR #2382, opened 2023-07-18
[VL] Fix insistent SMJ params

WangGuangxin:
The plan `SortMergeJoinExec` is converted to `MergeJoin` in velox only when the param `isSMJ=1` instead of `0`
https://github.com/oap-project/velox/blob/8cffafc01fbd8ea453022540fea6df97aebd2c32/velox/substrait/SubstraitToVeloxPlan.cpp#L319


WangGuangxin:
cc @JkSelf @zhouyuan

ulysses-you:
The changes looks good to me but seems it cause CI failed.. the SMJ cause timeout ?

zhouyuan:
@ulysses-you previously there are two UT failed due to some plan change, I'm rebasing this patch to verify

ulysses-you (approved):
cc @zhouyuan

PR #2408, opened 2023-07-20
[GLUTEN-2343] [VL] Copy buffers to a big buffer and then compress

jinchengchenghh:
Try to compress all the buffer to one buffer, and then compress the big buffer.
If this can gain performance improvement, will move to folly stream compression. It can accept several buffers to compress.

### DIFF
--- a/cpp/CMake/BuildQATZstd.cmake
+++ b/cpp/CMake/BuildQATZstd.cmake
@@ -19,7 +19,7 @@ include(ExternalProject)
 
 macro(build_qatzstd)
   # Find ZSTD
-  include(FindZstd)
+#   include(FindZstd)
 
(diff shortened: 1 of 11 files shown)


FelixYBW:
duplicated?

zhouyuan:
@Guhaiyan

jinchengchenghh:
Regretly, it has no performance improvement for both lz4 and zstd compression.
I will switch to zstd stream codec to find if there is potential improvement.

jinchengchenghh:
@Guhaiyan Do you test this PR in your environment? What's the performance compared to before?

jinchengchenghh:
stream codec performance is not good, so close this PR.

PR #2413, opened 2023-07-20
[CORE] Add substrait json plan to Spark UI

Yohahaha:


FelixYBW:
can we add to eventlog metrics instead of dump to stdout/err?

Yohahaha:
> can we add to eventlog metrics instead of dump to stdout/err?

Make sense for me.

Issue #2422, opened 2023-07-20
[CH] Parse NaN/Inf exception in RangeSelectorBuilder


exmy:
**Describe the bug**
`select a from values (cast('NaN' as double)), (1.0), (null) as data(a) order by a asc nulls last` 
will throw exception:

```
Caused by: java.lang.RuntimeException: Syntax error
0. ./build/../contrib/llvm-project/libcxx/include/exception:134: Poco::Exception::Exception(String const&, String const&, int) @ 0x0000000012fa7cbf  
1. ./build/../base/poco/Foundation/src/Exception.cpp:161: Poco::SyntaxException::SyntaxException(String const&, String const&, int) @ 0x0000000012facca9  
2. ./build/../base/poco/Foundation/src/NumberParser.cpp:0: Poco::NumberParser::parseFloat(String const&, char, char) @ 0x0000000012fdcf67  
3. ./build/../base/poco/Foundation/include/Poco/Dynamic/VarHolder.h:2287: Poco::Dynamic::VarHolderImpl<String>::convert(double&) const @ 0x000000000f1dd157  
4. ./build/../base/poco/Foundation/include/Poco/Dynamic/Var.h:171: double Poco::Dynamic::Var::convert<double>() const @ 0x0000000006dbd03d  
5. ./build/../src/Core/Field.h:687: local_engine::RangeSelectorBuilder::initRangeBlock(Poco::SharedPtr<Poco::JSON::Array, Poco::ReferenceCounter, Poco::ReleasePolicy<Poco::JSON::Array>>) @ 0x0000000006db9e21  
6. ./build/../base/poco/Foundation/include/Poco/SharedPtr.h:293: local_engine::RangeSelectorBuilder::RangeSelectorBuilder(String const&, unsigned long) @ 0x0000000006db86a6  
7. ./build/../contrib/llvm-project/libcxx/include/__memory/unique_ptr.h:0: local_engine::RangePartitionNativeSplitter::RangePartitionNativeSplitter(local_engine::NativeSplitter::Options, _jobject*) @ 0x0000000006daf54b  
8. ./build/../contrib/llvm-project/libcxx/include/__memory/unique_ptr.h:0: local_engine::NativeSplitter::create(String const&, local_engine::NativeSplitter::Options, _jobject*) @ 0x0000000006dae3cc  
[...]

PR #2423, opened 2023-07-20
[GLUTEN-2422][CH] Fix parse nan/inf exception in RangeSelectorBuilder

exmy:
`field_value.convert<DB::Float64>()` will fail when field_value is special value like NaN/Inf.

We check it and insert the corresponding value.

(Fixes: \#2422)

## How was this patch tested?
UT


liuneng1994 (approved):
LGTM

PR #2427, opened 2023-07-21
[GLUTEN-2182][VL] FEAT: Enable native Velox UDF

marin-ma:
This patch adds the framework for native UDF implementation in Gluten Velox backend. The general workflow is users need to implement the UDF with C++ templates, compile it into dynamic libs and then configure to use the libs for Spark applications. 

Main components are
- UDFResolver: register the functions into Spark
- UDFLoader: load the native libs and register the function into velox
- UdfExpression: transform the Scala func name to C++ in substrait

One small example on C++ UDF is added
image


ep/build-velox/src/build_velox.sh
```diff
@@ -63,9 +63,9 @@ function compile {
   fi
   # create libvelox_hive_connector.a for VeloxInitializer.cc
   sed -i 's/OBJECT//' velox/connectors/hive/CMakeLists.txt
-  COMPILE_OPTION="-DVELOX_ENABLE_PARQUET=ON"
+  COMPILE_OPTION="-DVELOX_ENABLE_PARQUET=ON -DVELOX_ENABLE_DUCKDB=ON"
```
zhouyuan, line 66: does this patch really requires duckdb?

zhouyuan (approved):

cpp/core/utils/StringUtil.cc
```diff
@@ -0,0 +1,57 @@
...
+
+#include <filesystem>
+#include <iostream>
+#include <string_view>
+#include <vector>
+
+#include "StringUtil.h"
+#include "exception.h"
+
+std::vector<std::string> gluten::splitByDelim(const std::string& s, const char delimiter) {
```
zhouyuan, line 26: folly has several helper function on this
https://github.com/facebook/folly/blob/main/folly/String.h#L498

zhouyuan:
👍  works for me

ep/build-velox/src/build_velox.sh
marin-ma: Yes. There's undefined symbol in libvelox.so
```
ldd -r libvelox.so
...
undefined symbol: _ZN8facebook5velox7parquet13duckdb_reader13ParquetReaderC1ESt10shared_ptrINS0_4dwio6common11InputStreamEERKNS6_13ReaderOptionsE       (cpp/build/releases/libvelox.so)
```
When trying to link udf library libmyudf.so it fails.

backends-velox/src/test/scala/io/glutenproject/expression/VeloxUdfSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+    super.sparkConf
+      .set("spark.plugins", "io.glutenproject.GlutenPlugin")
+      .set("spark.default.parallelism", "1")
+      .set("spark.memory.offHeap.enabled", "true")
+      .set("spark.memory.offHeap.size", "1024MB")
+
+  test("test udf") {
+    val df = spark.sql("""select myudf1(1), myudf2(100L)""")
+    df.collect().sameElements(Array(6, 105))
+  }
```
zhztheplayer, line 39: Have you tested whether the UDF works in Spark cluster mode?

gluten-core/src/main/scala/io/glutenproject/extension/OthersExtensionOverrides.scala
```diff
@@ -35,5 +35,8 @@ object OthersExtensionOverrides extends GlutenSparkExtensionsInjector {
     BackendsApiManager.getSparkPlanExecApiInstance
       .genExtendedStrategies()
       .foreach(extensions.injectPlannerStrategy)
+    BackendsApiManager.getSparkPlanExecApiInstance
+      .genInjectedFunctions()
+      .foreach(extensions.injectFunction)
```
zhztheplayer, line 40: Same issue here: is the extension able to be called from Spark executor? In the case of running a Spark cluster.

zhztheplayer: If not, probably we can try if C++ side VeloxInitializer is OK to do the initialization.

marin-ma: You are right. I assume it cannot.

marin-ma:
@zhztheplayer Pushed some fix and verified on yarn. The functions `myudf1` `myudf2` can successfully registered on executors now.

GlutenPerfBot (commented):
Velox TPCH SF2000 performance report: Error: No such file or directory @ rb_sysopen - /home/sparkuser/workspace/workspace/Gluten_Perf_Github_Action_Check/perf_script/log/native_gap_2427*csv

GlutenPerfBot (commented):
Velox TPCH SF2000 performance report: Error: No such file or directory @ rb_sysopen - /home/sparkuser/workspace/workspace/Gluten_Perf_Github_Action_Check/perf_script/log/native_gap_2427*csv

marin-ma:
@zhouyuan Some updates:
1. Fix functions not registered on executor side.
2. Separate UDF scala UT with other UTs. This is because running those tests together (in the same JVM) can cause libmyudf.so not being loaded by VeloxIntializer if other UTs run first.
3. Added documents.

zhouyuan:
Note: this patch replies on placing the udf libs into below positions:
`spark.gluten.sql.columnar.backend.velox.udfLibraryPaths`

zhztheplayer (approved):
Thanks! @zhouyuan If you have comments.

I recall that in Gazelle we used to adopt API `SparkContext.addFile` to distribute compiled binaries to worker nodes. I am not sure if something like that can be helpful for functionality like this one. Just FWIW.

marin-ma:
> I recall that in Gazelle we used to adopt API `SparkContext.addFile` to distribute compiled binaries to worker nodes. I am not sure if something like that can be helpful for functionality like this one. Just FWIW.

Since the udf library paths are already known, we can simply distribute with `--files`

Yohahaha:
Thanks for this great feature, two questions:
1. Why we need include more velox lib from ENABLE_BENCHMARK compile option?
2. Does centos not supported? I see velox_be.yml not added.

CC @marin-ma @zhouyuan

zhztheplayer:
> Since the udf library paths are already known, we can simply distribute with `--files`

Aha, sounds great if that works. Thank you for your explanation.

marin-ma:
> Thanks for this great feature, two questions:
> 
> 1. Why we need include more velox lib from ENABLE_BENCHMARK compile option?
> 2. Does centos not supported? I see velox_be.yml not added.
> 
> CC @marin-ma @zhouyuan

1. This is a fix. There are Undefined symbols in `libvelox.so` if not linking to these libs.
2. The compilation of the example udf library is included in centos CI.

zhouyuan:
Note: 
The Scala UDF can be converted to C++ with below LLM tools
https://www.codeconvert.ai/scala-to-c++-converter

PR #2435, opened 2023-07-21
[VL] Support empty dataframe write

ulysses-you:

If the write data is empty, we should write a valid parquet file rather than an empty file.

```shell
val df = spark.emptyDataFrame.select(lit(1).as("i"))
df.write.format("velox").save("test3")
```

before:
image

after:
image

## How was this patch tested?

Add test

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -2,8 +2,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=main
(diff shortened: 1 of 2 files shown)


ep/build-velox/src/get_velox.sh
```diff
@@ -2,8 +2,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=main
+VELOX_REPO=https://github.com/NetEase/gluten-velox
```
JkSelf, line 5: Please help to revert this change. Thanks.

PR #2480, opened 2023-07-27
[VL] Add shuffle split to generic benchmark

marin-ma:


zhouyuan:
Benchmark Velox

cpp/core/shuffle/LocalPartitionWriter.cc
```diff
@@ -35,7 +36,8 @@ arrow::Status LocalPartitionWriterBase::setLocalDirs() {
   // dir with prefix "columnar-shuffle"
   if (shuffleWriter_->options().data_file.length() == 0) {
     std::string dataFileTemp;
-    ARROW_ASSIGN_OR_RAISE(shuffleWriter_->options().data_file, createTempShuffleFile(configuredDirs_[0]));
+    size_t id = std::hash<std::thread::id>{}(std::this_thread::get_id()) % configuredDirs_.size();
```
jinchengchenghh, line 39: Why we need to change this?

marin-ma: This can have data files written to different disks, simulating multiple disks like SPARK_LOCAL_DIRS to spread the I/O pressure when running micro benchmark in multiple threads. And this code path is for tests/benchmarks only.

jinchengchenghh: Can you add a comment to mark it as tests only？Or set the code as #if defined test or benchmark?

cpp/core/shuffle/LocalPartitionWriter.cc
```diff
@@ -35,7 +36,8 @@ arrow::Status LocalPartitionWriterBase::setLocalDirs() {
   // dir with prefix "columnar-shuffle"
```
marin-ma, line 36: @jinchengchenghh Already commented here.

jinchengchenghh (approved):

FelixYBW (approved):

PR #2481, opened 2023-07-27
[VL] Add JolFileSystem (JNI-or-local, JVM-over-local), and spill-to-heap implementation

zhztheplayer:


zuochunwei (approved):

zuochunwei:
LGTM

backends-velox/pom.xml
```diff
@@ -179,6 +179,12 @@
       <artifactId>jackson-module-scala_${scala.binary.version}</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.jimfs</groupId>
+      <artifactId>jimfs</artifactId>
+      <version>1.3.0</version>
+      <scope>compile</scope>
```
ulysses-you, line 186: it depends on guava 32.1.1-jre... and seems we can not exclude the guava directly.

It would throw exception if set `spark.driver.userClassPathFirst=true;`:
```
Error operating ExecuteStatement: java.lang.NoSuchMethodError: com.google.common.util.concurrent.Futures.addCallback(Lcom/google/common/util/concurrent/ListenableFuture;Lcom/google/common/util/concurrent/FutureCallback;)V
at org.apache.hadoop.mapred.LocatedFileStatusFetcher.getFileStatuses(LocatedFileStatusFetcher.java:123)
	at org.apache.hadoop.mapred.FileInputFormat.listStatus(FileInputFormat.java:244)
	at org.apache.hadoop.mapred.FileInputFormat.getSplits(FileInputFormat.java:322)
	at org.apache.spark.rdd.HadoopRDD.getPartitions(HadoopRDD.scala:208)
```

I can not find a better way except shade and relocation it.

PR #2489, opened 2023-07-27
[GLUTEN-1141][FOLLOWUP][VL]Fix q72 duplicate join key

liujiayi771:

Fix q72 More than one key found from build side relation output error.


liujiayi771:
@zhouyuan I'm sorry that I didn't consider the previous revision clearly. We need to use the count method to count the columns with the same name, and the find method will only find the first matching column.

zhouyuan:
> @zhouyuan I'm sorry that I didn't consider the previous revision clearly. We need to use the count method to count the columns with the same name, and the find method will only find the first matching column.

Hi @liujiayi771 no worries, it seems to be a corner case, as it is not discovered by CI and some internal workloads. 

thanks, -yuan

zhouyuan (approved):
👍

PR #2499, opened 2023-07-28
[CORE] Fix a potential issue in hive table scan when DPP is used

PHILO-HE:
The filter pruning expression of hive table scan may come from DPP. We need to do the check and transform for the pruning expression, the corresponding subqueryBroadcast and the possible re-used broadcast exchange. Otherwise, a mismatched subqueryBroadcast can be used and cause hash relation cast issues.

## How was this patch tested?
No regression issue in spark UTs.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -284,6 +284,8 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
         applyScanTransformer(plan)
       case plan: FileSourceScanExec =>
         applyScanTransformer(plan)
+      case plan if HiveTableScanExecTransformer.isHiveTableScan(plan) =>
+        applyScanTransformer(plan)
       case plan: CoalesceExec =>
         logDebug(s"Columnar Processing for ${plan.getClass} is currently supported.")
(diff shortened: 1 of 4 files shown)


rui-mo (commented):
Can we desribe a bit in the PR description what does this PR trying to solve, and what is the potential issue?

PHILO-HE:
> Can we desribe a bit in the PR description what is this PR trying to solve, and what is the potential issue?

Done.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -572,6 +574,20 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
         TransformHints.tagNotTransformable(newSource, validationResult.reason.get)
         newSource
       }
+    case plan if HiveTableScanExecTransformer.isHiveTableScan(plan) =>
+      // TODO: Add DynamicPartitionPruningHiveScanSuite.scala
```
rui-mo, line 578: Is this suite disabled now?

rui-mo (approved):

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
PHILO-HE: It's just introduced in spark 3.3. And may already be in our 3.3 UT plan list.

PR #2500, opened 2023-07-28
[VL] Use value of spark.io.compression.codec as Velox's spill codec

jinchengchenghh:
Use Spark's existing option `spark.io.compression.codec` as Velox backend's spill codec.


zhztheplayer (approved):

zhztheplayer:
Two questions:
1. What happens if unsupported codec type passed to Velox?
2. Do spill files get compressed without this patch?

zhouyuan:
> Two questions:
> 
> 1. What happens if unsupported codec type passed to Velox?
> 2. Do spill files get compressed without this patch?

for q2: it's not compressed w/o this patch AFAIK. 

CC: @jackylee-ch 

thanks, -yuan

jinchengchenghh:
For Q1, it will throw exception.
https://github.com/facebookincubator/velox/blob/main/velox/common/config/SpillConfig.cpp#L46

PR #2519, opened 2023-07-30
[GLUTEN-2531][CORE][CH][VL] Introuce GlutenException

baibaichen:

`RuntimeException` is used when we need throw an expcetion, this PR introuces `GlutenException`. We can improve error meesage in the future.

(Fixes: \#2531)

## How was this patch tested?
Using Existed UTs


PHILO-HE (approved):
Looks great!

PR #2545, opened 2023-07-31
[WIP] Simplify state/resource management

Yohahaha:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


gluten-core/src/main/scala/org/apache/spark/util/GlutenTaskContext.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  def get(): TaskResourceRegistry = registry.get
+
+  private[this] val registry: ThreadLocal[TaskResourceRegistry] =
+    new ThreadLocal[TaskResourceRegistry]
+
+  protected[spark] def setTaskContext(tc: TaskResourceRegistry): Unit = registry.set(tc)
+
+  protected[spark] def unset(): Unit = registry.remove()
+}
```
zhztheplayer, line 38: What's the pros of adding this thread local storage? Seems to me that we should manage an extra task lifecycle state in Gluten's code. Previously `TaskContext.get()` is exposed in a stateless way (although it manages states internally)

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
```diff
@@ -60,27 +46,18 @@ object TaskResources extends TaskListener with Logging {
...
-            "is added to GlutenExecutorPlugin's task listener list")
-      }
-      return RESOURCE_REGISTRIES.get(tc)
+      throw new IllegalStateException("Not in a Spark task")
     }
+    GlutenTaskContext.get()
   }
 
-  def addRecycler(prio: Long)(f: => Unit): Unit = {
+  def addRecycler(name: String, prio: Long)(f: => Unit): Unit = {
```
zhztheplayer, line 54: 👍

cpp/core/utils/TaskContext.cc
```diff
@@ -61,12 +64,13 @@ void bindToTask(std::shared_ptr<void> object) {
   std::lock_guard<std::mutex> guard(fallbackStorageMutex);
   std::cout << "Binding a shared object to fallback storage. This should only happen on sub-thread of a Spark task. "
             << std::endl;
+  GLUTEN_CHECK(false, "Force throw");
```
zhztheplayer, line 67: Does this mean we no longer need the fallback storage at all?

gluten-core/src/main/scala/org/apache/spark/util/GlutenTaskContext.scala
Yohahaha: > What's the pros of adding this thread local storage? 

Avoid synchronized lock and thread check between across JNI.

> we should manage an extra task lifecycle state in Gluten's code

For what kinds of resources?

zhztheplayer: > For what kinds of resources?

"we should manage an extra task lifecycle state in Gluten's code" -> "within this change, we should manage an extra task lifecycle state in Gluten's code"

cpp/core/utils/TaskContext.cc
Yohahaha: It's an attempt, I have passed all tpc queries with this check.

Yohahaha:
Will add more description when PR is ready.

gluten-core/src/main/scala/org/apache/spark/util/GlutenTaskContext.scala
zhztheplayer: Is one of the main purpose to remove the "FALLBACK_REGISTRY"?

Yohahaha: > Is one of the main purpose to remove the "FALLBACK_REGISTRY"?

Yes, Gluten use single thread execution for now, thus all resources/states could be ''thread local''.

Yohahaha: > "within this change, we should manage an extra task lifecycle state in Gluten's code"

TaskResourceRegistry is bound with TaskContext, so they have same lifecycle.

zhztheplayer: > TaskResourceRegistry is bound with TaskContext, so they have same lifecycle.

OK. FWIW let's take it into consideration that Spark uses this task-thread model:

```
Thread 1 --> TaskContext 1
   |__ (Child) Thread  2   --> TaskContext 1

Thread 3 --> TaskContext 2

Thread 4 --> TaskContext 3

Driver Thread 5 --> No TaskContext (null)
```

To me here exists some tricky parts to optimize this part of code:

1. What happens if we are on a child thread of a task thread? e.g. Thread 2 
2. What happens if we are on the driver thread? e.g. Thread 5
3. What happens if we are not on any thread that are created by Spark, for example, a UT thread?

You can just refer to these questions when doing the optimization, I can get back to the PR when it's ready.

Yohahaha: Thanks for great examples!

Issue #2391, opened 2023-07-19
[CH] Improve date type read when using excel format


loneylee:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently, read csv use excel format when datatype is date32, mostly times will triger rollback. It takes more times on read csv.

PR #2558, opened 2023-08-01
[GLUTEN-2391][FOLLOWUP] Fix hive scan miss do validation

ulysses-you:

This pr is a followup of https://github.com/oap-project/gluten/pull/2392.

Add back the file format validation and native validation. Otherwise, such query will fail:

```sql
CREATE TABLE t AS SELECT 1 as c
SELECT * FROM t
```

```
Caused by: java.lang.RuntimeException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: ReaderFactory is not registered for format unknown
```

(Fixes: \#2391)

## How was this patch tested?

add test

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
@@ -175,6 +175,11 @@ class HiveTableScanExecTransformer(
   }
 
   override protected def doValidateInternal(): ValidationResult = {
+    val validationResult = super.doValidateInternal()
+    if (!validationResult.isValid) {
+      return validationResult
+    }
+
     val tableMeta = relation.tableMeta
     val planOutput = output.asInstanceOf[Seq[AttributeReference]]
     var hasComplexType = false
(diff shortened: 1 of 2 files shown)


ulysses-you:
cc @baibaichen  @zhouyuan @JkSelf thank you

gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
```diff
@@ -198,7 +203,7 @@ class HiveTableScanExecTransformer(
             if (!hasComplexType) {
               ValidationResult.ok
             } else {
-              ValidationResult.notOk("Hive scan is not defined")
+              ValidationResult.notOk("does not support complex type")
```
Yohahaha, line 206: nit: add more verbose validate failure reason would make sense.

PHILO-HE:
Hi @ulysses-you, thanks so much for your fix!
I am not sure whether there is any repeated (or even conflict) check in  `HiveTableScanExecTransformer's` `doValidateInternal` & `BasicScanExecTransformer's`. E.g., I note they all checked complex types, but have some differences. Could you please confirm it? In addition, I think it may be better to put `HiveTableScanExecTransformer's` validation logic into `BasicScanExecTransformer's` to let scala validation be executed first and then go into native validation. Not sure whether it is feasible technically. Just some quick observations. Thanks!

ulysses-you:
@Yohahaha @PHILO-HE  It's CH backend feature, I did not take a deep look. This pr itself just fixes the regression.

PHILO-HE:
> @Yohahaha @PHILO-HE It's CH backend feature, I did not take a deep look. This pr itself just fixes the regression.

NP. I will track the possible potential issue. Thanks!

PHILO-HE (approved):
Looks great! Thanks!

PR #2560, opened 2023-08-01
[CORE] Simplify Hadoop dependency management

pan3793:

The Gluten only calls a few Hadoop public API, with a quick glance, all of them are stable. And during runtime, Gluten always relies on Hadoop jars provided by Spark.

Since 3.2.0 SPARK-33212, Spark moves to use Hadoop shaded client, which simplifies the Hadoop dependencies management, as a downstream of Spark, Gluten should follow it too.

## How was this patch tested?

Pass GA.

### DIFF
--- a/backends-clickhouse/pom.xml
+++ b/backends-clickhouse/pom.xml
@@ -142,12 +142,6 @@
 	    </exclusions>
     </dependency>
     <dependency>
-      <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-client</artifactId>
-      <version>${hadoop.version}</version>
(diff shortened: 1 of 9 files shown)


pan3793:
cc @baibaichen as it touches the clickhouse module's pom

baibaichen:
looks great

baibaichen:
let's  re-run it

ulysses-you:
>  @ulysses-you I believe this change could address the guava incompatible issues.

This change does not address the guava issue. The guava issue is caused by https://github.com/oap-project/gluten/pull/2481#discussion_r1280535355

PR #2566, opened 2023-08-02
Remove QAT and IAA Arrow compress patch

jinchengchenghh:


jinchengchenghh:
@marin-ma Can you help verify this PR? Thanks!

cpp/core/utils/compression.h
```diff
@@ -33,11 +33,43 @@
...
+    arrow::Compression::LZ4_FRAME,
+    arrow::Compression::ZSTD,
+    arrow::Compression::GZIP};
+inline Result<std::unique_ptr<arrow::util::Codec>> createCodec(arrow::Compression::type compressedType) {
+  std::unique_ptr<arrow::util::Codec> codec;
+  switch (compressedType) {
+    case arrow::Compression::LZ4_FRAME:
+      return arrow::util::Codec::Create(compressedType);
+    case arrow::Compression::ZSTD:
+      codec = qat::makeDefaultQatZstdCodec();
```
marin-ma, line 47: This will ignore "codecBackend" when creating native writer. Here we want to create software codec but not qat codec if the "codecBackend" is null.

jinchengchenghh:
Is it possible to both defined QAT and IAA, and use codecBackend config to decide compress codec?

marin-ma:
> Is it possible to both defined QAT and IAA, and use codecBackend config to decide compress codec?

Yes. That's also what we do on main branch.

jinchengchenghh:
Can you help review again? @marin-ma

cpp/core/utils/Compression.cc
```diff
@@ -0,0 +1,75 @@
...
+    case arrow::Compression::LZ4_FRAME: {
+      GLUTEN_ASSIGN_OR_THROW(codec, arrow::util::Codec::Create(compressedType));
+    } break;
+    case arrow::Compression::ZSTD: {
+      if (codecBackend == CodecBackend::NONE) {
+        GLUTEN_ASSIGN_OR_THROW(codec, arrow::util::Codec::Create(compressedType));
+      } else if (codecBackend == CodecBackend::QAT) {
+#if defined(GLUTEN_ENABLE_QAT)
+        codec = qat::makeDefaultQatZstdCodec();
+#else
```
marin-ma, line 46: If the configured codec backend doesn't match cpp compile option, it will fallback to uncompressed without any hint. Can we print some log or just throw exception here?

marin-ma:
Verified. Thanks!

marin-ma (approved):

Issue #2430, opened 2023-07-21
[CH] Add pre-projection for hash shuffle with expressions failed


lgbo-ustc:
**Describe the bug**

Following query failed
```SQL
Comments
@[lgbo-ustc](https://github.com/lgbo-ustc)
lgbo-ustc commented [5 minutes ago](https://github.com/Kyligence/ClickHouse/issues/451#issue-1815240685)
(you don't have to strictly follow this form)

Describe the unexpected behaviour

Following query failed

select count(1) from ( select '2023-03-07' as sday
,abflag
,coalesce(country,'all') as country
,coalesce(usertype,'all') as usertype
,count(a.uid) as dau
,cast(null as bigint) as retention_1d_uv
,cast(null as bigint) as retention_2d_uv
,cast(null as bigint) as retention_3d_uv
,cast(null as bigint) as retention_4d_uv
,cast(null as bigint) as retention_5d_uv
,cast(null as bigint) as retention_6d_uv
,cast(null as bigint) as retention_7d_uv

,count(b.uid)/count(a.uid) as retention_1d_rate
,cast (null as float) as retention_2d_rate
,count(c.uid)/count(a.uid) as retention_3d_rate
,cast (null as float) as retention_4d_rate
,cast (null as float) as retention_5d_rate
,cast (null as float) as retention_6d_rate
,count(d.uid)/count(a.uid) as retention_7d_rate

from
    (select a.day
    ,a.abflag
    ,b.country
    ,if(c.uid is not null,'new','old') as usertype
    ,a.uid
    from
        (select day,abflag,app_uid as uid
        from swh_client_abflag
        where day = '2023-03-07'
        group by day,abflag,app_uid
        ) a
    join
        (select day,app_uid as uid,country
        from  com_dim_snapshot_user
        where day = '2023-03-07'
        and app_use_time_order_desc_rank=1
        ) b
    on a.day=b.day and a.uid=b.uid
    left join
        (select day,uid
        from account_registrations
        where day = '2023-03-07'
        and num_deleted=0
        group by day,uid
        ) c
    on a.day=c.day and a.uid=c.uid
    ) a
left join
    (
    select day,app_uid as uid
    from com_dim_snapshot_user
    where day = date_add('2023-03-07',1)
    and app_use_time_order_desc_rank=1
    ) b
on a.uid=b.uid
left join
    (
    select day,app_uid as uid
    from com_dim_snapshot_user
    where day = date_add('2023-03-07',3)
[...]

PR #2596, opened 2023-08-03
[GLUTEN-2430] [CH] Fix missing partition column when column has alias

liuneng1994:

Fix missing partition column when column has alias

(Fixes: \#2430)

## How was this patch tested?

unit tests


zzcclp (approved):
LGTM

PR #2597, opened 2023-08-03
[VL] Celeborn shuffle writer should use compressed byte size

ulysses-you:

This pr changes the partition lengths to use compressed byte size which follows the jvm client side behavior.

The partition lengths is very import that would affect AQE decision.

## How was this patch tested?

test manually

Before:
image

After:
image

### DIFF
--- a/cpp/core/jni/JniCommon.h
+++ b/cpp/core/jni/JniCommon.h
@@ -253,15 +253,17 @@ class CelebornClient : public RssClient {
     env->DeleteGlobalRef(javaCelebornShuffleWriter_);
   }
 
-  void pushPartitonData(int32_t partitionId, char* bytes, int64_t size) {
+  int32_t pushPartitonData(int32_t partitionId, char* bytes, int64_t size) {
     JNIEnv* env;
     if (vm_->GetEnv(reinterpret_cast<void**>(&env), jniVersion) != JNI_OK) {
       throw gluten::GlutenException("JNIEnv was not attached to current thread");
     }
     jbyteArray array = env->NewByteArray(size);
     env->SetByteArrayRegion(array, 0, size, reinterpret_cast<jbyte*>(bytes));
-    env->CallIntMethod(javaCelebornShuffleWriter_, javaCelebornPushPartitionData_, partitionId, array);
+    jint celebornBytesSize =
+        env->CallIntMethod(javaCelebornShuffleWriter_, javaCelebornPushPartitionData_, partitionId, array);
     checkException(env);
+    return static_cast<int32_t>(celebornBytesSize);
   }
 
   JavaVM* vm_;
--- a/cpp/core/shuffle/rss/CelebornPartitionWriter.cc
+++ b/cpp/core/shuffle/rss/CelebornPartitionWriter.cc
@@ -36,10 +36,10 @@ arrow::Status CelebornPartitionWriter::pushPartition(int32_t partitionId) {
   auto buffer = celebornBufferOs_->Finish();
   int32_t size = buffer->get()->size();
   char* dst = reinterpret_cast<char*>(buffer->get()->mutable_data());
-  celebornClient_->pushPartitonData(partitionId, dst, size);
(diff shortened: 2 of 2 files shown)


ulysses-you:
I'm wondering if we can add a new flag `compressed` to method `pushData`/`mergeData`, so that we can do compression at native side and pass a compressed buffer to celeborn client. cc @kerwin-zk

ulysses-you:
@kerwin-zk oh, I see, but that's the problem.. we then do compress buffer again at jvm side. Accroding to the compressed byte size, it seems the jvm side also takes affect.  We should at least follow the final compressed byte size.

kerwin-zk:
> @kerwin-zk oh, I see, but that's the problem.. we then do compress buffer again at jvm side. Accroding to the compressed byte size, it seems the jvm side also takes affect. We should at least follow the final compressed byte size.
> 
> It seems would cause slower than non-native case due to double-compression.

@ulysses-you You can turn off compression on the java side by configuring spark celeborn.client.shuffle.compression.codec=none.

ulysses-you:
@kerwin-zk  thank you for the config. Then shall we use the final compressed byte size if people enable compression at jvm side ?

kerwin-zk:
> @kerwin-zk thank you for the config. Then shall we use the final compressed byte size if people enable compression at jvm side ?



> @kerwin-zk thank you for the config. Then shall we use the final compressed byte size if people enable compression at jvm side ?

@ulysses-you sure.

cpp/core/shuffle/rss/CelebornPartitionWriter.cc
```diff
@@ -36,10 +36,11 @@ arrow::Status CelebornPartitionWriter::pushPartition(int32_t partitionId) {
   auto buffer = celebornBufferOs_->Finish();
   int32_t size = buffer->get()->size();
   char* dst = reinterpret_cast<char*>(buffer->get()->mutable_data());
-  celebornClient_->pushPartitonData(partitionId, dst, size);
+  int32_t commpressedBytesSize = celebornClient_->pushPartitonData(partitionId, dst, size);
```
kerwin-zk, line 39: ditto.

cpp/core/shuffle/rss/CelebornPartitionWriter.cc
```diff
@@ -36,10 +36,11 @@ arrow::Status CelebornPartitionWriter::pushPartition(int32_t partitionId) {
...
   shuffleWriter_->partitionCachedRecordbatch()[partitionId].clear();
   shuffleWriter_->setPartitionCachedRecordbatchSize(partitionId, 0);
-  shuffleWriter_->setPartitionLengths(partitionId, shuffleWriter_->partitionLengths()[partitionId] + size);
+  shuffleWriter_->setPartitionLengths(
+      partitionId, shuffleWriter_->partitionLengths()[partitionId] + commpressedBytesSize);
```
kerwin-zk, line 43: ditto.

cpp/core/jni/JniCommon.h
```diff
@@ -253,15 +253,17 @@ class CelebornClient : public RssClient {
...
-  void pushPartitonData(int32_t partitionId, char* bytes, int64_t size) {
+  int32_t pushPartitonData(int32_t partitionId, char* bytes, int64_t size) {
     JNIEnv* env;
     if (vm_->GetEnv(reinterpret_cast<void**>(&env), jniVersion) != JNI_OK) {
       throw gluten::GlutenException("JNIEnv was not attached to current thread");
     }
     jbyteArray array = env->NewByteArray(size);
     env->SetByteArrayRegion(array, 0, size, reinterpret_cast<jbyte*>(bytes));
-    env->CallIntMethod(javaCelebornShuffleWriter_, javaCelebornPushPartitionData_, partitionId, array);
+    jint commpressedBytesSize =
```
kerwin-zk, line 263: This name needs to be changed, such as ```celebornBytesSize```, because Celeborn may not compress.

kerwin-zk (approved):

yaooqinn (approved):
LGTM

PR #2599, opened 2023-08-03
[GLUTEN-2504][VL] Support upload UDF libraries via --files and --archives

marin-ma:
This patch improves the native UDF support by allowing users to upload UDF libs via two Spark methods:
- `--files`
```
```

- `--archives`
```
--archives /path/to/udf_archives.zip#udf_archives
--conf spark.gluten.sql.columnar.backend.velox.udfLibraryPaths=udf_archives
```

The documents are also updated with more detailed instrutions.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
@@ -117,7 +117,9 @@ object UDFResolver extends Logging {
 
   def resolveUdfConf(conf: java.util.Map[String, String]): Unit = {
     if (isDriver) {
-      conf.put(BackendSettings.GLUTEN_VELOX_UDF_LIB_PATHS, localLibraryPaths)
+      if (localLibraryPaths != null) {
+        conf.put(BackendSettings.GLUTEN_VELOX_UDF_LIB_PATHS, localLibraryPaths)
+      }
     } else {
       val sparkConf = SparkEnv.get.conf
       Option(conf.get(BackendSettings.GLUTEN_VELOX_UDF_LIB_PATHS)) match {
--- a/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
@@ -17,6 +17,7 @@
 package org.apache.spark.sql.expression
 
 import io.glutenproject.backendsapi.velox.BackendSettings
+import io.glutenproject.exception.GlutenException
 import io.glutenproject.expression.{ConverterUtils, ExpressionTransformer, Transformable}
 import io.glutenproject.expression.ConverterUtils.FunctionConfig
 import io.glutenproject.substrait.{ExpressionType, TypeConverter}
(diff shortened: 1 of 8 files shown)


zhouyuan:
CC @jackylee-ch

backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
```diff
@@ -80,17 +94,132 @@ object UDFResolver extends Logging {
...
+      case None =>
+        Seq.empty
+      case Some(udfFiles) =>
+        val master = sparkConf.getOption("spark.master")
+        val isYarn = master.isDefined && master.get.equals("yarn")
+        localLibraryPaths = if (isYarn) {
+          // For Yarn-client or Yarn-cluster mode,
+          // driver cannot get uploaded files via SparkFiles.get.
+          // So we need to copy and unpack files for driver first.
+          val hadoopConf = new YarnConfiguration(SparkHadoopUtil.newConfiguration(sparkConf))
```
jackylee-ch, line 163: AFAIK, for yarn-cluster mode, the archives can be unpacked on driver side, so we only need pass the unpaced directory, such as `./libs`, to `udfLibraryPaths`. As for yarn-client mode, why not let user to unpack the archives and pass the directories to `udfLibraryPaths`. 

BTW, if the `udfLibraryPaths` is different between driver side and executor side, we only need to add a `driver.udfLibraryPaths`, which is defauled with `udfLibraryPaths`, to solve this problem. Just like the `spark.pyspark.driver.python` and `spark.pyspark.python`.

marin-ma: Sounds great. Will do.

marin-ma:
@jackylee-ch Updated and lists 3 supported ways to configure in doc https://github.com/oap-project/gluten/blob/472d5fbbb6bfd50280ab8e64e1ba52453fa05ee9/docs/get-started/Velox.md#using-udf-in-gluten

Unpacking archives is still preserved. If the configured path is absolute path and is an archive, it will be fetched or copied to a local dir and unpacked.

marin-ma:
@jackylee-ch @zhouyuan Do you have any more comments?

backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
```diff
@@ -80,17 +96,129 @@ object UDFResolver extends Logging {
...
+    val sourceName = source.getName
+    val dest = new File(destDir, sourceName)
+    logInfo(
+      s"Unpacking an archive $sourceName from ${source.getAbsolutePath} to ${dest.getAbsolutePath}")
+    try {
+      Utils.deleteRecursively(dest)
+      Utils.unpack(source, dest)
+    } catch {
+      case e: Exception =>
+        throw new GlutenException(s"Unpack ${source.toString} failed.", e)
```
jackylee-ch, line 146: nit: `Unpack ${source.toString} failed, please check if it is a compressed file`

jackylee-ch:
> @jackylee-ch Updated based on your suggestion and lists 3 supported ways to configure in doc https://github.com/oap-project/gluten/blob/472d5fbbb6bfd50280ab8e64e1ba52453fa05ee9/docs/get-started/Velox.md#using-udf-in-gluten
> 
> Unpacking archives is still preserved. If the configured path is absolute path and is an archive, it will be fetched or copied to a local dir and unpacked.

Okey to me, only left some small comments.

marin-ma:
@zhouyuan Do you have further comments? Or can you help to merge? Thanks!

zhouyuan (approved):
👍

zhouyuan:
@marin-ma thanks for the great work!

PR #2613, opened 2023-08-04
[CORE][VL] Fix memory leak of fallback storage

Yohahaha:

Before this patch, the resources which not tracked by Spark executor task will be managed in FALLBACK_REGISTRY(java side) and fallbackStorage(native side), however, these will be released when process exit, and may bring memory problem when running lots of query in one Spark application.

E.g. DPP subquery execution will submitted to another thread pool in Spark driver, Gluten need do c2r conversion in the same time. We add explicit release after conversion.
Some computation happens on another thread, such as PythonUDF, since Velox backend has not support PythonUDF, it will insert a c2r, this c2r will do conversion in PythonRunner writer thread. We add a check in c2r to find this case and release early.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarToRowExec.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarToRowExec.scala
@@ -134,6 +134,8 @@ class ColumnarToRowRDD(
             if (!hasNext && !closed) {
               jniWrapper.nativeClose(c2rId)
               closed = true
+
+              TaskResources.releaseCurrentResources()
             }
             hasNext
           }
@@ -183,7 +185,7 @@ class ColumnarToRowRDD(
             }
           }
         }
-        res.flatten
+        res.flatten.map(_.copy())
       }
   }
 
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarToRowExec.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarToRowExec.scala
@@ -125,6 +125,7 @@ class ColumnarToRowRDD(
       }
 
       if (batches.isEmpty) {
+        TaskResources.releaseCurrentResources()
         Iterator.empty
       } else {
         val res: Iterator[Iterator[InternalRow]] = new Iterator[Iterator[InternalRow]] {
@@ -146,6 +147,7 @@ class ColumnarToRowRDD(
             numOutputRows += batch.numRows()
 
             if (batch.numRows == 0) {
+              batch.close()
               logInfo(s"Skip ColumnarBatch of ${batch.numRows} rows, ${batch.numCols} cols")
               Iterator.empty
             } else if (
@@ -159,20 +161,25 @@ class ColumnarToRowRDD(
             } else if (output.isEmpty) {
               numInputBatches += 1
               numOutputRows += batch.numRows()
-              ColumnarBatches.emptyRowIterator(batch).asScala
+              val rows = ColumnarBatches.emptyRowIterator(batch).asScala
+              batch.close()
+              rows
             } else {
+              val cols = batch.numCols()
+              val rows = batch.numRows()
               val beforeConvert = System.currentTimeMillis()
               val batchHandle = ColumnarBatches.getNativeHandle(batch)
               val info = jniWrapper.nativeColumnarToRowConvert(batchHandle, c2rId)
 
               convertTime += (System.currentTimeMillis() - beforeConvert)
+              batch.close()
 
               new Iterator[InternalRow] {
                 var rowId = 0
-                val row = new UnsafeRow(batch.numCols())
+                val row = new UnsafeRow(cols)
 
                 override def hasNext: Boolean = {
-                  rowId < batch.numRows()
+                  rowId < rows
                 }
 
                 override def next: UnsafeRow = {
--- a/cpp/core/utils/TaskContext.cc
+++ b/cpp/core/utils/TaskContext.cc
@@ -57,6 +57,7 @@ void bindToTask(std::shared_ptr<void> object) {
     taskContextStorage->bind(object);
     return;
   }
+  GLUTEN_CHECK(false, "Force throw");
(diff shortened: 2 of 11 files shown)


PHILO-HE:
@zhztheplayer, could you please take a review? Thanks.

zhztheplayer:
@Yohahaha 
Is this ready to review? Or let's see if the python test suites can pass.

Yohahaha:
> @Yohahaha Is this ready to review? Or let's see if the python test suites can pass.

Yes, you can start review, python suites has been verified in my machine.

Yohahaha:
> > @Yohahaha Is this ready to review? Or let's see if the python test suites can pass.
> 
> Yes, you can start review, python suites has been verified in my machine.

Not ready... Modification of TaskResources is not enough.

Yohahaha:
@zhztheplayer this PR is ready for review, I have run whole tests in local.

cpp/core/utils/TaskContext.cc
```diff
@@ -40,6 +40,8 @@ class TaskContextStorage {
 };
 
 thread_local std::unique_ptr<TaskContextStorage> taskContextStorage = nullptr;
+
+// TODO: Remove this and let related unmanaged resources be managed.
```
jackylee-ch, line 44: So we still have some memory unmanaged, right?

zhztheplayer:
> > @Yohahaha Is this ready to review? Or let's see if the python test suites can pass.
> 
> Yes, you can start review, python suites has been verified in my machine.

Thanks, I'll take a look ASAP.

cpp/core/utils/TaskContext.cc
Yohahaha: Please ignore this obsolete comments and will clean later, I add false check when using fallback storage, ideally all native resources in under managed.

backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarToRowExec.scala
```diff
@@ -134,6 +135,8 @@ class ColumnarToRowRDD(
             if (!hasNext && !closed) {
               jniWrapper.nativeClose(c2rId)
               closed = true
+
+              TaskResources.releaseCurrentResources()
```
zhztheplayer, line 139: Would you like to explain a bit about this? Thanks!

zhztheplayer: If there is a case: (G for Gluten, V for Vanilla Spark)

```
G_ShuffleRead -> C2R_1 -> V_Plan_1 -> R2C_1 -> G_Plan_1 -> C2R_2 -> V_ShuffleWrite
```

Once C2R_1 reached EoS, C2R_2 might still consume data. Is the case covered by the code?

Yohahaha: Sure, the key changes is `RESOURCE_REGISTRIES` which contains mapping of 
(TaskContext -> (ThreadID -> TaskResourceRegistry)) now, to handle different thread computation under same Spark TaskContext scope, let it could be released.

If all computations are offloaded to native, there should no multi computation under same Spark task scope. If we fallback some special operators, c2r will be inserted, and this parts of computation could happens in another thread, such as PythonRunner, so we need add extra mapping to release resources created in another thread.

Yohahaha: > Once C2R_1 reached EoS, C2R_2 might still consume data. Is the case covered by the code?

Yes, I add a UnsafeRow#copy at the end of ColumnarToRowRDD.

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
```diff
@@ -60,20 +51,97 @@ object TaskResources extends TaskListener with Logging {
...
+  }
+
+  def releaseCurrentResources(): Unit = {
+    RESOURCE_REGISTRIES.synchronized {
+      val currentTId = Thread.currentThread().getId
+      val registryMap = RESOURCE_REGISTRIES.get(getLocalTaskContext())
+      if (registryMap.containsKey(currentTId) && registryMap.get(currentTId).needExplicitRelease) {
+        val registry = registryMap.remove(currentTId)
+        logDebug(s"Release all resources in ${registry.name}")
+        registry.releaseAll()
```
zhztheplayer, line 143: So by the condition `needExplicitRelease`, `releaseCurrentResources` only releases the resources from off-task thread, does it?

Yohahaha: Yes.

cpp/core/utils/TaskContext.cc
```diff
@@ -55,13 +52,7 @@ void bindToTask(std::shared_ptr<void> object) {
...
     return;
   }
-  // The fallback storage is used. Spark sometimes creates sub-threads from a task thread. For example,
-  //   PythonRunner.scala:183 @ Spark3.2.2
-  //   GlutenSubqueryBroadcastExec.scala:73 @6bce2c33
-  std::lock_guard<std::mutex> guard(fallbackStorageMutex);
-  std::cout << "Binding a shared object to fallback storage. This should only happen on sub-thread of a Spark task. "
-            << std::endl;
-  fallbackStorage->bind(object);
+  GLUTEN_CHECK(false, "[BUG] Found untracked computation, please submit issue with query plan.");
```
zhztheplayer, line 55: This fix is not related to the Java side [TaskContext, [TID, Resource]] design?

Yohahaha: Just put check here to find corner cases we ignored.

zhztheplayer:
@Yohahaha Sorry for the slow response but let's find some time to have a dedicated sync for this change. Thanks.

Yohahaha:
> @Yohahaha Sorry for the slow response but let's find some time to have a dedicated sync for this change. Thanks.

Sure, we could have a phone call by wechat.

Yohahaha:
PR #2843 is better way to solve this issue.

Issue #2610, opened 2023-08-04
[CORE] Refactor some ExpressionTransformer


exmy:
There are some `ExpressionTransformer` like `UnaryArgumentCollectionOperationTransformer` and `UnaryExpressionTransformer` are the same expression transformer essentialy in current code base, we can unify them to make code more clean.

PR #2632, opened 2023-08-07
[GLUTEN-2610][CORE] Refactor expression transformer

exmy:

(Fixes: \#2610)
In the current codebase, there are many duplicate and redundant `doTransform` codes about ExpressionTransformer. This patch aims to:

* Remove redundant `ExpressionTransformer`, like `UnaryArgumentCollectionOperationTransformer` and `UnaryExpressionTransformer`

* Remove unnecessary `ExpressionTransformer` and use `GenericExpressionTransformer ` to uniformly transform expression that do not require special handling, making the code more concise and clear.

* Move some `ExpressionTransformer` to more appropriate locations according to their function group, e.g., move like/in to `PredicateTransformer`, if/caseWhen to `ConditionalTransformer`

## How was this patch tested?

existed ut


zhouyuan:
@PHILO-HE

exmy:
@rui-mo @PHILO-HE Could you help review this? Thanks~

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
```diff
@@ -222,21 +221,17 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
-        case _ => false
-      }
-    } else false
+  private def isDecimalArithmetic(expression: BinaryExpression): Boolean = {
+    expression match {
+      case b: BinaryArithmetic if b.children.forall(_.dataType.isInstanceOf[DecimalType]) =>
+        b match {
+          case _: Divide | _: Multiply => true
+          case _: Add | _: Subtract => true
+          case _: Remainder | _: Pmod => true
```
rui-mo, line 230: Why don't we put these possibilities into one case clause, and return true?

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
```diff
@@ -330,15 +286,15 @@ object ExpressionConverter extends SQLConfHelper with Logging {
 
     TestStats.addExpressionClassName(expr.getClass.getName)
     // Check whether Gluten supports this expression
-    val substraitExprName = ExpressionMappings.expressionsMap.get(expr.getClass)
+    val substraitExprName = ExpressionMappings.expressionsMap.getOrElse(expr.getClass, "")
```
rui-mo, line 289: Should we throw if name can not be got?

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionTransformer.scala
```diff
@@ -18,6 +18,75 @@ package io.glutenproject.expression
...
+}
+
+object TernaryExpressionTransformer {
+  def apply(
+      substraitExprName: String,
+      first: ExpressionTransformer,
+      second: ExpressionTransformer,
+      third: ExpressionTransformer,
+      original: Expression): ExpressionTransformer = {
+    new NaryExpressionTransformer(substraitExprName, Seq(first, second, third), original)
```
rui-mo, line 78: If terneary transformer just reuses nary transformer, why do we still need a ternary transformer?

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionTransformer.scala
```diff
@@ -18,6 +18,75 @@ package io.glutenproject.expression
...
+
+object QuaternaryExpressionTransformer {
+  def apply(
+      substraitExprName: String,
+      first: ExpressionTransformer,
+      second: ExpressionTransformer,
+      third: ExpressionTransformer,
+      forth: ExpressionTransformer,
+      original: Expression): ExpressionTransformer = {
+    new NaryExpressionTransformer(substraitExprName, Seq(first, second, third, forth), original)
```
rui-mo, line 90: Same here.

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionTransformer.scala
```diff
@@ -18,6 +18,75 @@ package io.glutenproject.expression
...
+import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.catalyst.optimizer.NormalizeNaNAndZero
+
 trait ExpressionTransformer {
   def doTransform(args: java.lang.Object): ExpressionNode
 }
+
+object LeafExpressionTransformer {
+  def apply(substraitExprName: String, original: Expression): ExpressionTransformer = {
+    new NaryExpressionTransformer(substraitExprName, Seq(), original)
```
rui-mo, line 30: Same.

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionTransformer.scala
exmy: I also want to remove ternary transformer completely, do you think it's necessary to retain `UnaryExpressionTransformer` and `BinaryExpressionTransformer`? They contain some case clauses.

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
exmy: It still can throw exception, see 
```
if (substraitExprName.isEmpty) {
  throw new UnsupportedOperationException(s"Not supported: $expr. ${expr.getClass}")
}
```

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionTransformer.scala
rui-mo: I see. Looks to me they can also be replaced with `NaryExpressionTransformer` but this requires more efforts. How do you think? Maybe that should belong to another PR.

exmy: So, we completely remove these Unary/Binary/Ternary/Quaternary transformers and one thing that's not certain is whether we should move the case clauses from `UnaryExpressionTransformer ` to `ExpressionConverter` or `NaryExpressionTransformer `. Maybe for the sake of consistency, we can move them to `ExpressionConverter` first?

gluten-core/src/main/scala/io/glutenproject/expression/NaryExpressionTransformer.scala
```diff
@@ -19,11 +19,11 @@ package io.glutenproject.expression
 import io.glutenproject.expression.ConverterUtils.FunctionConfig
 import io.glutenproject.substrait.expression.{ExpressionBuilder, ExpressionNode}
 
-import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.catalyst.expressions._
 
 import com.google.common.collect.Lists
 
-class BasicCollectionOperationTransfomer(
+class NaryExpressionTransformer(
```
PHILO-HE, line 26: Thanks for your efforts!
It looks we can just make it a case class. Then, there is no need to define `object NaryExpressionTransformer`. Right?

We may do the same change for many transformer classes, each of which extends `ExpressionTransformer` and is not extended by other class.

exmy: OK, I will remove `object NaryExpressionTransformer` and check other transformer classes later on.

exmy:
@PHILO-HE @rui-mo have updated, could you take a look again?

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionTransformer.scala
rui-mo: I agree. In ExpressionConverter, some special cases can be handled, and others are processed in NaryExpressionTransformer.

gluten-core/src/main/scala/io/glutenproject/expression/StringExpressionTransformer.scala
```diff
@@ -0,0 +1,110 @@
...
+    subject: ExpressionTransformer,
+    regexp: ExpressionTransformer,
+    rep: ExpressionTransformer,
+    pos: ExpressionTransformer,
+    original: RegExpReplace)
+  extends ExpressionTransformer {
+
+  override def doTransform(args: java.lang.Object): ExpressionNode = {
+    if (BackendsApiManager.isVeloxBackend) {
+      NaryExpressionTransformer(substraitExprName, Seq(subject, regexp, rep, pos), original)
```
PHILO-HE, line 93: Looks a historical issue. Shouldn't we add `return` for this statement?

BTW, I think `nary` indicates there is no child. In contrast, unary/binary respectively for one/two child/children may be derived from this word. The name `NaryExpressionTransformer` may not correctly reflect its usage. Right?
Maybe, `GenericExpressionTransformer` is better to indicate there is no special handling to transform. Any other proposal? cc @rui-mo

exmy: >  Shouldn't we add return for this statement?

Great catch!

And `Nary` what I'm trying to mean is `N-ary`, it seems a bit ambiguous, I agree with using `GenericExpressionTransformer `.

PHILO-HE (approved):
LGTM!
I note `GenericExpressionTransformer`  uses `OPT` as function config. Maybe, using which value doesn't matter now. I'm assuming `FunctionConfig` can be cleaned up. cc @rui-mo. Just filed an issue https://github.com/oap-project/gluten/issues/2807 for future fix by a separate PR.

zzcclp (approved):
LGTM

PR #2637, opened 2023-08-08
[GLUTEN-2343] Add support to copy buffers and then compress

jinchengchenghh:
Config spark.gluten.sql.columnar.shuffle.compressionMode to control the compression mode in shuffle


jinchengchenghh:
@Guhaiyan if you need to copy buffer for all the compression, you can change the function `getCompressionMode` in your repo.

cpp/core/utils/compression.h
```diff
@@ -23,8 +23,13 @@ namespace gluten {
 
 enum CodecBackend { NONE, QAT, IAA };
 
+// ONE_BY_ONE mode will preallocate max compressed buffer, and then compress each buffer to the max compressed buffer
+// Stream mode will coly the buffers to a big buffer and then compress the big buffer
```
marin-ma, line 27: typo: coly -> copy

cpp/core/utils/compression.h
```diff
@@ -23,8 +23,13 @@ namespace gluten {
...
+enum CompressionMode { ONE_BY_ONE, STREAM };
+
 std::unique_ptr<arrow::util::Codec> createArrowIpcCodec(
     arrow::Compression::type compressedType,
     CodecBackend codecBackend);
 
+CompressionMode getCompressionMode(CodecBackend codecBackend);
```
marin-ma, line 34: We should also pass compression codec here. QAT supports both gzip and zstd codec. We only need to copy buffers for gzip codec.

cpp/core/jni/JniCommon.h
```diff
@@ -98,6 +98,20 @@ static inline gluten::CodecBackend getCodecBackend(JNIEnv* env, jstring codecJst
   }
 }
 
+static inline gluten::CompressionMode getCompressionMode(JNIEnv* env, jstring compressionModeJstr) {
+  if (compressionModeJstr == nullptr) {
+    return gluten::CompressionMode::ONE_BY_ONE;
+  }
+  auto compressionMode = jStringToCString(env, compressionModeJstr);
+  if (compressionMode == "one_by_one") {
```
FelixYBW, line 106: let's change to "buffer" and "rowvector"

marin-ma:
Please also update document.

jinchengchenghh:
Can you help review this PR again? @marin-ma

docs/Configuration.md
```diff
@@ -37,6 +37,7 @@ You can add these configurations into spark-defaults.conf to enable or disable t
...
+| spark.gluten.sql.columnar.shuffle.compressionMode          | Setting different compression mode in shuffle, Valid options are buffer and rowvector, buffer means compress each buffer of RowVector to pre allocated big buffer, rowvector means to copy the buffers of RowVector to a big buffer, and then compress the buffer.                                                                                                                                                                                                                                                                                                                                                        | buffer                                                 |
```
marin-ma, line 40: Probably better explanation:

`buffer` option compress each buffer of RowVector individually into one pre-allocated large buffer. `rowvector` option first copies each buffer of RowVector to a large buffer and then compress the entire buffer in one go.

marin-ma:
@jinchengchenghh w/ QAT can pass compilation on the QAT env. But the performance validation may take some time. Have you verified spark.gluten.sql.columnar.shuffle.compressionMode=rowvector with lz4 or gzip codec on Spark? If yes maybe we can merge first.

marin-ma (approved):

khemchand-zetta:
I have tried by setting the shuffleCompressionMode to both buffer and rowvector so see the compression behavior. Like as it says the compression will be one by one in buffer and in case of rowvector its stream based like all buffers are in a large buffer  and compressed in one go. but its not behaving like that it only compressing one by one in both cases, anybody help me to get the piece of code where this mode login is implemented or like how this conf works.

marin-ma:
@khemchand-zetta This configuration is deprecated, and buffer mode is no longer supported, as we did not observe any significant changes in compressed size across different modes. Could you please provide a demonstration of your use case that requires testing different compression modes?

khemchand-zetta:
Do you mean by the row vector mode supported anymore as its compressing one by one for the buffers so buffer mode it , right.

khemchand-zetta:
And talking about the use case I was just trying to see the behavior of this configuration.

marin-ma:
@khemchand-zetta Apologies for the confusion earlier. Currently, we only support compressing buffers one by one, so it's buffer mode.

khemchand-zetta:
Thanks @marin-ma for the clarification.

PR #2642, opened 2023-08-08
[GLUTEN-2611][CH] Fix read field error throw exception

loneylee:

Using excel format, read error field will throw an exception. Fix it with bool instead.

## How was this patch tested?

Test by ut


baibaichen:
how is test？

baibaichen:
Existed  UT is OK .

baibaichen (approved):
LGTM

PR #2651, opened 2023-08-08
[VL] Add build script for QAT

marin-ma:


.github/workflows/velox_be.yml
```diff
@@ -92,6 +92,22 @@ jobs:
           cmake -DBUILD_TESTS=ON -DENABLE_HBM=ON .. && \
           cmake --build . --target hbw_allocator_test -- -j && \
           ctest -V -R TestHbw'
+      - name: Install QAT dependencies
```
jinchengchenghh, line 95: Can you add a file `dev/setup_qat.sh`? So we can easily build QAT locally.

jinchengchenghh: And we can also use this script in CI.

jinchengchenghh (approved):

PR #2654, opened 2023-08-08
[VL] gluten-it-parameterized: Change delimiter '->' to ':' to avoid escaping '>' in shell

zhztheplayer:
After the change, '->' should be ':' in `parameterized` integration testing.

For example: 

```sh
sbin/gluten-it.sh parameterized \
--local \
--threads=8 \
--benchmark-type=ds \
-s=100 \
--queries=q95 \
--iterations=1 \
--shuffle-partitions=8 \
--enable-ui \
--enable-history \
-m=OffHeapExecutionMemory \
-d=offheap:3g,spark.memory.offHeap.size=3g \ # note ":" is used here rather than '->'
-d=offheap:4g,spark.memory.offHeap.size=4g \
--extra-conf=spark.gluten.sql.columnar.backend.velox.spillStrategy=auto \
--extra-conf=spark.gluten.sql.columnar.backend.velox.spillFileSystem=local \
--extra-conf=spark.executor.userClassPathFirst=true
```


zhztheplayer:
@yma11

zhztheplayer:
@jinchengchenghh

jinchengchenghh (approved):

Issue #2658, opened 2023-08-08
[core] attribute bind error happens when Exchange rangepartitioning get attribute from its child's output


lhuang09287750:
### Backend

CH (ClickHouse)

### Bug description

add the following code into any testsuite(such as GlutenClickHouseFileFormatSuite, source csv does not matter much,give one row enough)
```
  test("fix issue KY2717 for attribute error") {
    val schema = StructType.apply(
      Seq(
        StructField.apply("dpqy", StringType, nullable = true)
      ))
    val df = spark.read
      .option("delimiter", ",")
      .schema(schema)
      .csv(csvDataPath + "/dpqy.csv")
      .toDF()
    df.createTempView("retail_stores_CN1")

    val schema1 = StructType.apply(
      Seq(
        StructField.apply("spfl", StringType, nullable = true)
      ))
    val df1 = spark.read
      .option("delimiter", ",")
      .schema(schema1)
      .csv(csvDataPath + "/spfl.csv")
      .toDF()
    df1.createTempView("retail_stores_by_transaction")

    val query =
      """
        |SELECT
        |  *
        |FROM
        |  (
        |    select
        |      *
        |    from
        |      (
        |        (
        |          select
        |            dpqy spfl
        |          from
        |            retail_stores_CN1
        |          group by
        |            dpqy
        |          order by
        |            dpqy
        |          limit
        |            2147483647
        |        )
        |        union
        |        (
        |          select
        |            spfl
        |          from
        |            retail_stores_by_transaction
        |          group by
        |            spfl
        |          order by
        |            spfl
        |          limit
        |            2147483647
        |        )
        |      )
        |    order by
        |      spfl
        |  ) AS qq
        |""".stripMargin

    compareResultsAgainstVanillaSpark(query, compareResult = true, _ => {})
  }
```
error like this:
![image](https://github.com/oap-project/gluten/assets/83992752/76b64ed0-cd4a-4c50-9e13-a78b40af1dfa)

### Spark version

3.2

PR #2659, opened 2023-08-08
[Gluten-2658][core]fix attribute bind error when "exchange rangepartitioning" get attributes from its child's output

lhuang09287750:

(Fixes: \#2658)

## How was this patch tested?

add a new test case


lhuang09287750:
already fix by \#2596 ,close this pr

Issue #2375, opened 2023-07-18
[CH] split function throw exception


shuai-xu:
**Describe the bug**
In clickhouse, split function require the first column to be constant or fixed string, but gluten will pass a StringColumn to it. And it will throw exception.
<img width="1439" alt="image" src="https://github.com/oap-project/gluten/assets/20960838/743f6d8e-9784-47ec-aa2b-74addf7942ce">

**To Reproduce**
select uid,split(op_id,concat('5_3_',cast(uid as string),'_'))[1] as order_id

PR #2660, opened 2023-08-08
[GLUTEN-2375][CH] Fix split function runtime exception in CH

taiyang-li:

(Fixes: \#2375) 

Current solution is falling back when some corner cases are not supported by CH. 

TODO: 
- align spark split and ch splitByRegexp when limit > 0
- align spark split and ch splitByRegexp when regexp is not literal


zzcclp (approved):
LGTM

PR #2665, opened 2023-08-09
[Core] Fix QAT build

marin-ma:


marin-ma:
cc: @jinchengchenghh

cpp/core/utils/qat/QatCodec.cc
```diff
@@ -21,6 +21,8 @@
 #include <qatseqprod.h>
 #include <qatzip.h>
 #include <zstd.h>
+#include <algorithm>
```
jinchengchenghh, line 24: Do you need to include this header?

marin-ma: Yes. `std::any_of` needs this header.

jinchengchenghh (approved):

marin-ma:
@jinchengchenghh I just found QAT build is broken on main branch. It's probably because my previous verification wasn't picking the latest commit. Also found it's possible to install QAT driver on a machine without QAT HW devices and I'm trying to enable QAT build in CI https://github.com/oap-project/gluten/pull/2651

PR #2666, opened 2023-08-09
[CORE] Minor code refactor for checking pushed down aggregation

PHILO-HE:
## How was this patch tested?
No regression issue on enabled UTs with pushed down aggregation.


zhouyuan (approved):
👍

PR #2674, opened 2023-08-09
[GLUTEN-2684][CORE][HOTFIX]: disable Spark-36726 parquet reading tests temporarily on Spark 3.3

marin-ma:
Temporarily disable test case "SPARK-36726: test incorrect Parquet row group file offset" in GlutenParquetIOSuite for Spark33


gaoyangxiaozhu:
curious what's the RCA why in gluten UT the spark read parquet always return `timestamp_ntz` column when reading test file ?
Is there anyway to fix it insteads of just disable it ?

marin-ma:
> curious what's the RCA in gluten UT the spark read parquet always return `timestamp_ntz` column when reading test file ? Is there anyway to fix it insteads of just disable it ?

I opened issue #2684 to track. Let's temporarily disable it until we fix it.

marin-ma:
cc: @zhouyuan

zhouyuan (approved):

zhouyuan:
@marin-ma thanks for the quick fix!

-yuan

PR #2683, opened 2023-08-09
[GLUTEN-2682][VL] support array extract

summaryzb:
Fixes: [ISSUE-2682](https://github.com/oap-project/gluten/issues/2682)

## How was this patch tested?
pass ut

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecHandler.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecHandler.scala
@@ -71,6 +71,14 @@ class SparkPlanExecHandler extends SparkPlanExecApi {
       leftNode: ExpressionNode,
       rightNode: ExpressionNode,
       original: GetArrayItem): ExpressionNode = {
+    if (original.dataType.isInstanceOf[DecimalType]) {
+      val decimalType = original.dataType.asInstanceOf[DecimalType]
+      val precision = decimalType.precision
+      if (precision > 18) {
+        throw new UnsupportedOperationException(
+          "GetArrayItem not support decimal precision more than 18")
+      }
+    }
     // ignore origin substraitExprName
     val functionName = ConverterUtils.makeFuncName(
       ExpressionMappings.expressionsMap.get(classOf[ElementAt]).get,
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecHandler.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecHandler.scala
@@ -21,7 +21,9 @@ import io.glutenproject.backendsapi.SparkPlanExecApi
 import io.glutenproject.columnarbatch.ColumnarBatches
 import io.glutenproject.execution._
 import io.glutenproject.expression._
+import io.glutenproject.expression.ConverterUtils.FunctionConfig
 import io.glutenproject.memory.alloc.NativeMemoryAllocators
+import io.glutenproject.substrait.expression.{ExpressionBuilder, ExpressionNode, IfThenNode}
 import io.glutenproject.vectorized.{ColumnarBatchSerializer, ColumnarBatchSerializerJniWrapper}
 
 import org.apache.spark.{ShuffleDependency, SparkException}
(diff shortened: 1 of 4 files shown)


gluten-core/src/main/scala/io/glutenproject/expression/ArrayExpressionTransformer.scala
```diff
@@ -93,10 +94,17 @@ class GetArrayItemTransformer(
...
 
-    val functionName = ConverterUtils.makeFuncName(
-      substraitExprName,
-      Seq(original.left.dataType, original.right.dataType),
-      FunctionConfig.OPT)
+    val functionName = if (BackendsApiManager.isVeloxBackend) {
+      ConverterUtils.makeFuncName(
+        ExpressionMappings.expressionsMap.get(classOf[ElementAt]).get,
+        Seq(original.dataType),
+        FunctionConfig.OPT)
```
jackylee-ch, line 101: Please add some uts for this implement.

summaryzb: ComplexTypeSuite does the ut

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecHandler.scala
```diff
@@ -57,6 +60,45 @@ import scala.collection.mutable.ArrayBuffer
...
+        FunctionConfig.OPT))
+    // right node already add 1
+    val literalNode = ExpressionBuilder.makeLiteral(1.toInt, IntegerType, false)
+    val lessThanFuncNode = ExpressionBuilder.makeScalarFunction(
+      lessThanFuncId,
+      Lists.newArrayList(rightNode, literalNode),
+      ConverterUtils.getTypeNode(BooleanType, true))
+    new IfThenNode(Lists.newArrayList(lessThanFuncNode), Lists.newArrayList(nullNode), resultNode)
+  }
+
```
yma11, line 101: Why not implement Spark `element_at` at velox side? Current solution seems like a workaround at transformer layer and it hides the difference between Spark `get_array_item` and Presto `element_at`. Furthermore, we may have to do such a workaround for Map `element_at` in future.

jackylee-ch: Actually, the `element_at` function in Spark is consistent with the one in Velox. It is the`get_array_item` function in Spark that has some semantic differences with the `element_at` function in Spark. 
Here’s an example to illustrate the difference between `get_array_item` and `element_at` in Spark.
GetArrayItem UT:
https://github.com/apache/spark/blob/master/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/ComplexTypeSuite.scala#L63
element_at UT:
https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/DataFrameFunctionsSuite.scala#L2130

yma11: So do we still need to do implementation of `get_array_item` at velox side? I see you don't change the function name in transformer.

jackylee-ch: AFAIK, it is not nessesary to support `get_array_item` in velox now. It seems that current implement has already solve the difference between `get_array_tiem` and `element_at`.

yma11: You resolve the difference here but still need to map `get_array_item` to `element_at` for offloading to velox, as changes in [2760](https://github.com/oap-project/gluten/pull/2760). The current status is fallback. You can confirm and I will reopen the PR if need.

summaryzb: After apply this pr, it'll not fallback

summaryzb: > So do we still need to do implementation of `get_array_item` at velox side? I see you don't change the function name in transformer.

function name changed in `genGetArrayItemExpressionNode`

jackylee-ch:
cc @zhouyuan @zhztheplayer

jackylee-ch (dismissed):

jackylee-ch (approved):
LGTM

ulysses-you (commented):
better to add some tests

jackylee-ch:
> better to add some tests

@ulysses-you There are already some uts in [GlutenComplexTypeSuite](https://github.com/oap-project/gluten/blob/main/gluten-ut/spark33/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenComplexTypeSuite.scala#L21).

PR #2700, opened 2023-08-10
[VL] Disable Velox spill file truncating by default

zhztheplayer:
This is a hot fix to disable spill truncating to avoid possible performance regression by too many small files.

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -955,7 +955,7 @@ object GlutenConfig {
       .internal()
       .doc("The maximum size of a single spill file created")
       .bytesConf(ByteUnit.BYTE)
-      .createWithDefaultString("0MB")
+      .createWithDefaultString("1GB")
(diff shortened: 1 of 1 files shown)


shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -955,7 +955,7 @@ object GlutenConfig {
       .internal()
       .doc("The maximum size of a single spill file created")
       .bytesConf(ByteUnit.BYTE)
-      .createWithDefaultString("20MB")
+      .createWithDefaultString("0MB")
```
jinchengchenghh, line 958: I would suggest to change to 1G, iceberg project compaction file default size is 1G

zhztheplayer: Yes sounds convicing. 1G is good for most cases. Let's use it.

Hopefully we can get some time to create some benchmarkings to decide on a optimal value of such configs.

jinchengchenghh (approved):

Issue #2699, opened 2023-08-10
[VL] org.apache.spark.sql.execution.ColumnarBuildSideRelation cannot be cast to org.apache.spark.sql.execution.joins.HashedRelation


kerwin-zk:
### Backend

VL (Velox)

### Bug description

In the scenario of tpcds 3T, test q10.sql in the case of scan fallback, and the following exception occurs.
```
Caused by: java.lang.ClassCastException: org.apache.spark.sql.execution.ColumnarBuildSideRelation cannot be cast to org.apache.spark.sql.execution.joins.HashedRelation
	at org.apache.spark.sql.execution.SubqueryBroadcastExec.$anonfun$relationFuture$2(SubqueryBroadcastExec.scala:81) ~[spark-sql_2.12-3.3.1.jar:3.3.1]
	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withExecutionId$1(SQLExecution.scala:145) ~[spark-sql_2.12-3.3.1.jar:3.3.1]
	at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:169) ~[spark-sql_2.12-3.3.1.jar:3.3.1]
	at org.apache.spark.sql.execution.SQLExecution$.withExecutionId(SQLExecution.scala:143) ~[spark-sql_2.12-3.3.1.jar:3.3.1]
	at org.apache.spark.sql.execution.SubqueryBroadcastExec.$anonfun$relationFuture$1(SubqueryBroadcastExec.scala:78) ~[spark-sql_2.12-3.3.1.jar:3.3.1]
	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659) ~[scala-library-2.12.15.jar:?]
	at scala.util.Success.$anonfun$map$1(Try.scala:255) ~[scala-library-2.12.15.jar:?]
	at scala.util.Success.map(Try.scala:213) ~[scala-library-2.12.15.jar:?]
	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292) ~[scala-library-2.12.15.jar:?]
	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33) ~[scala-library-2.12.15.jar:?]
	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33) ~[scala-library-2.12.15.jar:?]
[...]

PR #2701, opened 2023-08-10
[GLUTEN-2699][CORE] Fix a cast issue in DPP when gluten scan is disabled

kerwin-zk:

#2699

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -256,6 +256,7 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
       val newPartitionFilters =
         ExpressionConverter.transformDynamicPruningExpr(plan.partitionFilters, reuseSubquery)
       val newSource = plan.copy(partitionFilters = newPartitionFilters)
+      newSource.setLogicalLink(plan.logicalLink.get)
       TransformHints.tag(newSource, TransformHints.getHint(plan))
       newSource
     case plan: BatchScanExec =>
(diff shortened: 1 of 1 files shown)


gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -276,6 +276,13 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
             } else {
               return shj.withNewChildren(shj.children.map(replaceWithTransformerPlan))
             }
+          case scan: FileSourceScanExec =>
+            val newPartitionFilters = {
+              ExpressionConverter.transformDynamicPruningExpr(scan.partitionFilters, reuseSubquery)
+            }
+            val newSource = scan.copy(partitionFilters = newPartitionFilters)
+            TransformHints.tagNotTransformable(newSource, "fallback scan")
```
PHILO-HE, line 284: It looks `BatchScan/HiveTableScan` have the similar issue. It may be better to have a separate impl. like `applyScanTransformer` to handle these operators. The difference is scan will not be transformed. Thanks!

kerwin-zk: @PHILO-HE done.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -251,6 +251,34 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
     }
   }
 
+  def applyScanNotTransformable(plan: SparkPlan): SparkPlan = plan match {
+    case plan: FileSourceScanExec =>
+      val newPartitionFilters =
+        ExpressionConverter.transformDynamicPruningExpr(plan.partitionFilters, reuseSubquery)
+      val newSource = plan.copy(partitionFilters = newPartitionFilters)
+      TransformHints.tagNotTransformable(newSource, "fallback file source scan")
```
PHILO-HE, line 259: There are two minor tests failed related to unexpected fallback reason. Maybe, we need to re-use the tag of `plan` for `newSource` to keep the original fallback reason. e.g.,

`TransformHints.tag(newSource, TransformHints.getHint(plan))`

PHILO-HE (approved):
Looks great!

ulysses-you:
every time we fix a bug, better to add at least one test to ensure we will not make any regression in future

PHILO-HE:
> every time we fix a bug, better to add at least one test to ensure we will not make any regression in future

Yes, it is worthy to add some tests. I just did a trivial follow-up work at https://github.com/oap-project/gluten/pull/2788. Thanks!

PR #2713, opened 2023-08-10
[GLUTEN-2711][CH] Fix core dump when filter not has expression 

loneylee:
## How was this patch tested?

Test by ut


liuneng1994 (approved):
LGTM

PR #2714, opened 2023-08-10
[VL] Validate unsupported regex in re2

marin-ma:


cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -150,9 +151,18 @@ bool SubstraitToVeloxPlanValidator::validateScalarFunction(
   if (name == "extract") {
     return validateExtractExpr(params);
   }
-  if (name == "regexp_extract_all" && !scalarFunction.arguments()[1].value().has_literal()) {
-    logValidateMsg("native validation failed due to: pattern is not literal for regex_extract_all.");
-    return false;
+  if (name == "regexp_extract_all" || name == "regexp_extract" || name == "regexp_replace" || name == "rlike") {
+    auto patternArg = scalarFunction.arguments()[1].value();
+    if (!patternArg.has_literal()) {
+      logValidateMsg("native validation failed due to: pattern is not literal for regex_extract_all.");
```
PHILO-HE, line 157: Nit：
We should use `name` to replace `regex_extract_all`. Thanks!

PHILO-HE (commented):
Looks great!
Could you add some scala tests? e.g, with unsupported lookaround pattern used.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -150,9 +151,18 @@ bool SubstraitToVeloxPlanValidator::validateScalarFunction(
...
+      return false;
+    }
+    auto pattern = patternArg.literal().string();
+    auto re2 = RE2(pattern, RE2::Quiet);
+    if (!re2.ok()) {
+      logValidateMsg("native validation failed due to: pattern " + pattern + " compilation failed in RE2.");
```
PHILO-HE, line 163: Can we expose RE2's compilation error message to the validation message? It may be helpful to users to figure out the root reason. Thanks!

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -150,9 +151,18 @@ bool SubstraitToVeloxPlanValidator::validateScalarFunction(
...
-    logValidateMsg("native validation failed due to: pattern is not literal for regex_extract_all.");
-    return false;
+  if (name == "regexp_extract_all" || name == "regexp_extract" || name == "regexp_replace" || name == "rlike") {
+    auto patternArg = scalarFunction.arguments()[1].value();
+    if (!patternArg.has_literal()) {
+      logValidateMsg("native validation failed due to: pattern is not literal for regex_extract_all.");
+      return false;
+    }
+    auto pattern = patternArg.literal().string();
+    auto re2 = RE2(pattern, RE2::Quiet);
```
unigof, line 161: Could you extract **re2** related code to utils? Maybe it will be used frequently.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
marin-ma: Weird but the RE2 error message is always empty.

PHILO-HE: Maybe, `Quiet` option makes it silent. I'm not sure. You can try to change it. Thanks!

marin-ma: Tried default options with log print but not help. I'm not sure if it's an issue from the current RE2 version we use.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
marin-ma: Sorry for missing this. Will do.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
marin-ma: Just found it's due to my local env was broken. Already added.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -150,9 +152,18 @@ bool SubstraitToVeloxPlanValidator::validateScalarFunction(
   if (name == "extract") {
     return validateExtractExpr(params);
   }
-  if (name == "regexp_extract_all" && !scalarFunction.arguments()[1].value().has_literal()) {
-    logValidateMsg("native validation failed due to: pattern is not literal for regex_extract_all.");
-    return false;
+  if (name == "regexp_extract_all" || name == "regexp_extract" || name == "regexp_replace" || name == "rlike") {
```
jackylee-ch, line 155: nit: There are 10 spark build-in functions related to regex, it is better to use `list.contains` to define whether to check pattern.

marin-ma: Could you provide those function names or hint me where to find them? I only found these 4 functions.

jackylee-ch: We are making split to support regex, so split also needs to be added here later.

marin-ma: > Here is the list for functions that starts with regexp. https://spark.apache.org/docs/latest/api/sql/index.html#regexp And, `rlike` and `split` use regex too.

The url points to the latest spark function list. Many functions are not supported in spark 3.3.2 https://spark.apache.org/docs/3.3.2/api/sql/index.html#regexp_extract

Currently there should still be the 4 functions. `regex`, `regex_like` and `rlike` are the same and all aliased to `rlike.
`split` is in blacklist. I suppose there's no missing functions.

jackylee-ch: Okey, we can change it after `split` support regex with re2.

cpp/velox/utils/Common.cc
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "Common.h"
+
+namespace gluten {
+
+std::unique_ptr<re2::RE2> compilePattern(const std::string& pattern) {
+  return std::make_unique<re2::RE2>(re2::StringPiece(pattern), RE2::Quiet);
```
jackylee-ch, line 23: nit: add a check whether the pattern semantic diff between re2 and java.util.regex. Such as https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/RegexFunctions.cpp#L26

unigof: Hi, @jackylee-ch ,The check for regex semantic diff on [RegexFunctions.cpp#L26] maybe not enough, do you have any idea for diff check?
These are some other diff case：
'|' 
'.'
'.*'
'\\.*'
'\\?*'

cpp/velox/utils/Common.cc
```diff
@@ -0,0 +1,64 @@
...
+#include "Common.h"
+
+namespace gluten {
+
+// Note: This method is mostly copied from velox/functions/sparksql/RegexFunctions.cpp
+//
+// Blocks patterns that contain character class union, intersection, or
+// difference because these are not understood by RE2 and will be parsed as a
+// different pattern than in java.util.regex.
+bool ensureRegexIsCompatible(const std::string& pattern, std::string& error) {
```
marin-ma, line 27: @jackylee-ch Added here. @unigof As for your concern, could you submit another PR to enhance this function after this PR merge?

PHILO-HE (approved):
Looks great! This fix is much better!

PR #2718, opened 2023-08-11
[VL] Minor refactor c2r/r2c java code for return early

Yohahaha:

Let c2r/r2c could return early when input iterator is empty and initialize resources later.


ulysses-you (approved):

Yohahaha:
Thanks for merging!

PR #2719, opened 2023-08-11
[CH] Optimize shuffle writer

liuneng1994:

Support prefer cache columnar shuffle on ch backends
add more metrics on exchange node

## How was this patch tested?

UT


baibaichen (approved):
LGTM

Issue #2709, opened 2023-08-10
[CH] Can't repport error in case of reseting to unknow commit id.


baibaichen:
### Problem description

The current mechanism for cloning clickhouse source code is 
1. Clone latest code with `--depth 1`, which lead to only top commit id is
2. Reset to certain commit.

so, there is an error if the commit id isn't found, see following log. In this case, we always using latest codes  and didn't report an error.

Please note ,  there is only latest commit id could be found because of  `--depth 1`.  We also need fix this issue.

### System information

All

### CMake log

```bash
Please describe how you were tryiUpdating files: 100% (24797/24797), done.
[2023-08-10T06:14:42.126Z] fatal: ambiguous argument '9b1ae30a36e': unknown revision or path not in the working tree.
[2023-08-10T06:14:42.126Z] Use '--' to separate paths from revisions, like this:
[2023-08-10T06:14:42.126Z] 'git <command> [<revision>...] -- [<file>...]'ng to build velox and what issue occured
```
```

lwz9103:
### Fix Design

1. add `COMMAND_ERROR_IS_FATAL ANY` in each execute_process (there is no global config...) to fail fast
2. set depth to 3, so that custom CH branch can fetch latest 3 commits

lwz9103:
### Test Evidence
Set `COMMAND_ERROR_IS_FATAL ANY` and set CH_COMMIT to a not exist commit

![image](https://github.com/oap-project/gluten/assets/23639010/27dbb4c4-43b3-4699-a9e5-28aae28f355c)

PR #2722, opened 2023-08-11
[GLUTEN-2709][CH] enable cmake fail fast and set clone depth to 3

lwz9103:

(Fixes: \#2709)

## How was this patch tested?

manual tests


baibaichen (approved):
LGTM

PR #2726, opened 2023-08-11
[Gluten-CORE][MINNOR] Add the unsupported message for map functions

jackylee-ch:
This is a mirror pr. We found that the program does not print unsupported information correctly when using `GetMapValue`.

## How was this patch tested?
unit tests


ulysses-you (dismissed):

jackylee-ch:
retried UTs, now it can be merged @ulysses-you

ulysses-you (approved):

Issue #2740, opened 2023-08-14
The Gluten endpoint address should support IPV6


WangGuangxin:
### Backend

VL (Velox)

### Bug description

When we run on IPv6 env, the Gluten endpoint address is invalid. A spark exception is thrown with 
![image](https://github.com/oap-project/gluten/assets/1312321/10d9bbf2-bcfd-4917-bf7b-204fdb45971d)

The reason is that for IPv6 address, the endpoint address should be surrounded by square brackets like this

spark://GlutenDriverEndpoint@**[**IPV6 address**]**:33715

PR #2742, opened 2023-08-14
[GLUTEN-2740][CORE] Gluten endpoint should supports ipv6

WangGuangxin:

The Gluten endpoint address should be constructed using Spark buildin `RpcEndpointAddress` constructor, so that it can work on ipv6 env.

(Fixes: \#2740)


zhouyuan:
@WangGuangxin thanks for the quick fix!

zhouyuan (approved):
👍

PR #2747, opened 2023-08-14
Add spark round function

zhejiangxiaomai:
 velox round presto function can not handle round(0.575,2) = 0.57 case.  we need add spark round function to fix tpcds q78 resuslt not match issue.

## How was this patch tested?

### DIFF
--- a/cpp/velox/operators/functions/RegistrationAllFunctions.cc
+++ b/cpp/velox/operators/functions/RegistrationAllFunctions.cc
@@ -15,8 +15,10 @@
  * limitations under the License.
  */
 #include "RegistrationAllFunctions.h"
+#include "Arithmetic.h"
 #include "RowConstructor.h"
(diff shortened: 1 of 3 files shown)


cpp/velox/operators/functions/RegistrationAllFunctions.cc
```diff
@@ -37,6 +41,17 @@ void registerCustomFunctions() {
       std::make_unique<RowConstructor>());
 }
 
+// Register some functions that is not accepted by the community, but necessary for gluten.
+void registerFuntionsForGluten() {
```
rui-mo, line 45: Move to `registerFunctionOverwrite`? It is to override some Velox functions.

zhejiangxiaomai: Thanks. will fix it.

PR #2749, opened 2023-08-14
[GLUTEN-2031][VL] Support rand function

PHILO-HE:
This PR proposes the support for rand function, only considering the usage with no user-specified seed, i.e., `rand()`.

There will be a separate PR to support `rand(seed)`. It requires combing user-specified seed with partition index to serve as the final seed for random generator in native lib. Thus, we can achieve global cross-partitions random distribution. If only user-specified seed is used, same random sequence will be generated for each partition.
See [spark source code](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/randomExpressions.scala#L45).

## How was this patch tested?
Spark UTs.

### DIFF
--- a/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
+++ b/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
@@ -376,8 +376,8 @@ class GlutenClickHouseTPCHParquetSuite extends GlutenClickHouseTPCHAbstractSuite
   }
 
   test("test 'function rand'") {
-    // TODO: add rand(1), rand(null) test back.
-    runSql("select rand() from range(10)")(checkOperatorMatch[ProjectExecTransformer])
+    runSql("select rand(), rand(1), rand(null) from range(10)")(
+      checkOperatorMatch[ProjectExecTransformer])
   }
 
   test("test 'function date_add/date_sub/datediff'") {
--- a/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
+++ b/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
@@ -376,8 +376,8 @@ class GlutenClickHouseTPCHParquetSuite extends GlutenClickHouseTPCHAbstractSuite
   }
 
   test("test 'function rand'") {
-    runSql("select rand(), rand(1), rand(null) from range(10)")(
-      checkOperatorMatch[ProjectExecTransformer])
+    // TODO: add rand(1), rand(null) test back.
+    runSql("select rand() from range(10)")(checkOperatorMatch[ProjectExecTransformer])
   }
 
   test("test 'function date_add/date_sub/datediff'") {
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -22,10 +22,10 @@ import io.glutenproject.expression.WindowFunctionsBuilder
 import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat.{DwrfReadFormat, OrcReadFormat, ParquetReadFormat}
 
-import org.apache.spark.sql.catalyst.expressions.{Alias, CumeDist, DenseRank, Descending, Expression, Literal, NamedExpression, NthValue, PercentRank, RangeFrame, Rank, RowNumber, SortOrder, SpecialFrameBoundary, SpecifiedWindowFrame}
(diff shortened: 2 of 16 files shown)


backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -223,48 +223,32 @@ object BackendSettings extends BackendSettingsApi {
...
-   * Check whether plan is Count(1).
+   * Check whether a plan needs to be offloaded even though they have empty input schema, e.g,
+   * Sum(1), Count(1), rand(), etc.
    * @param plan:
    *   The Spark plan to check.
-   * @return
-   *   Whether plan is an Aggregation of Count(1).
    */
-  private def isCount1(plan: SparkPlan): Boolean = {
+  private def mayNeedOffload(plan: SparkPlan): Boolean = {
```
winningsix, line 231: Minor: ```isOffloadNeededWithoutSchema``` better?

winningsix (commented):
Do we need some tests for ```rand``` function?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -223,48 +223,32 @@ object BackendSettings extends BackendSettingsApi {
...
+            }
+          })
+      case p: ProjectExec if p.projectList.nonEmpty =>
+        p.projectList.forall(
+          expr =>
+            expr match {
+              // Block directly falling back the below functions by FallbackEmptySchemaRelation.
+              case _: Rand => true
+              case _ => false
+            })
```
zhouyuan, line 251: these changes seems not related with `rand()` ?

PHILO-HE: As `rand()` has no child output, we need this to avoid make it fall back by `fallbackOnEmptySchema` rule.

PHILO-HE:
Run Gluten Clickhouse CI

backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
```diff
@@ -376,8 +376,8 @@ class GlutenClickHouseTPCHParquetSuite extends GlutenClickHouseTPCHAbstractSuite
   }
 
   test("test 'function rand'") {
-    runSql("select rand(), rand(1), rand(null) from range(10)")(
-      checkOperatorMatch[ProjectExecTransformer])
+    // TODO: add rand(1), rand(null) test back.
+    runSql("select rand() from range(10)")(checkOperatorMatch[ProjectExecTransformer])
```
PHILO-HE, line 380: Hi @taiyang-li, this PR removes CH's testing for `rand(seed)`. This PR make such case fall back. See PR description. Please let me know if you have other ideas. Thanks!

PHILO-HE: cc @zzcclp

taiyang-li: @PHILO-HE Maybe we can add a if in RandTransformer for clickhouse backend to make sure CH do not fallback.

taiyang-li (changes requested):

backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
PHILO-HE: > @PHILO-HE Maybe we can add a if in RandTransformer for clickhouse backend to make sure CH do not fallback.

Hi @taiyang-li, does CH already handle the issue as mentioned in description? I'm assuming CH has the same issue which needs a separate fix, so I did this change.

taiyang-li: Thanks for reminding. Currently CH doesn't need the feature you mentioned in PR. Because CH rand function doesn't support user speicified seed.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -233,48 +233,35 @@ object BackendSettings extends BackendSettingsApi {
...
-              expression => {
-                expression.aggregateFunction match {
-                  case c: Count => c.children.size == 1 && c.children.head.equals(Literal(1))
-                  case _ => false
-                }
-              }) =>
-        true
-      case _ =>
-        false
+  private def mayNeedOffload(plan: SparkPlan): Boolean = {
```
rui-mo, line 241: About the naming, is there any case where this function returns true but cannot be offloaded? Why **may**NeedOffload?

rui-mo: OK, I see. Current name is good as whether to offload is not decided.

rui-mo (approved):

PR #2750, opened 2023-08-14
[GLUTEN-1434] [VL] Remove Arrow columnar to row

jinchengchenghh:


cpp/core/compute/Backend.h
```diff
@@ -74,17 +74,16 @@ class Backend : public std::enable_shared_from_this<Backend> {
...
+  virtual std::shared_ptr<ColumnarToRowConverter> getColumnar2RowConverter(MemoryAllocator* allocator) {
+    throw GlutenException("Not implement getColumnar2RowConverter");
+  };
 
   virtual std::shared_ptr<RowToColumnarConverter> getRowToColumnarConverter(
       MemoryAllocator* allocator,
       struct ArrowSchema* cSchema) {
-    return std::make_shared<gluten::RowToColumnarConverter>();
+    throw GlutenException("Not implement getRowToColumnarConverter");
+    ;
```
Yohahaha, line 86: nit: remove this

cpp/core/operators/c2r/ColumnarToRow.h
```diff
@@ -17,119 +17,36 @@
...
-#include <arrow/util/bit_util.h>
 #include "memory/ColumnarBatch.h"
 
-#include <boost/align.hpp>
-
 namespace gluten {
 
 class ColumnarToRowConverter {
  public:
-  explicit ColumnarToRowConverter() {}
```
Yohahaha, line 35: Why remove `explicit` keyword?

jinchengchenghh: It should be used for 1 argument constructor.
https://www.scaler.com/topics/cpp-explicit/
If we have a class that has a constructor with a single argument

zhztheplayer (approved):
Thanks! It would be good if we can rename and move scala code `VeloxColumnarToRowExec.scala` to module `gluten-data` as well in future.

PR #2754, opened 2023-08-14
[VL] Close output iterator early

Yohahaha:


Yohahaha:
Run Gluten Clickhouse CI

Yohahaha:
Hi @rui-mo @PHILO-HE @zhztheplayer could someone help take a look?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala
```diff
@@ -170,21 +170,12 @@ class IteratorHandler extends IteratorApi with Logging {
         res
       }
 
-      override def next(): Any = {
-        if (!hasNext) {
-          throw new java.util.NoSuchElementException("End of stream")
```
ulysses-you, line 175: why remove it ?

Yohahaha: I think Iterator ensure call order, and each `resIter.hasNext` operation will lookup native static map that not cheap.

Yohahaha:
Could someone help take a look? @ulysses-you @zhztheplayer @PHILO-HE

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala
ulysses-you: I agree it would save native calls, but better to let author @zhztheplayer take a look if it's safe enough

PHILO-HE (approved):
LGTM. Thanks!

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala
zhztheplayer: I'd suggest keeping the guard code like this unless we have special reason to remove it...

Since I don't make it sure that in gluten we always follow the legacy iterator convention, somehow a developer may write some code that calls next() without calling hasNext() at first.

For example you can look at the code in `ColumnarBatchInIterator.scala`, `next()` call will always return the same batch until `hasNext()` is called, that's definitely risky and not following the convention, but we don't have time to conduct a clean-up over the bad cases like that. So let's keep the guard code, temporarily?

Yohahaha: Make sense for me, thanks!

zhztheplayer (approved):

ulysses-you (approved):

PR #2763, opened 2023-08-15
[CORE] Refactor fallback policy

ulysses-you:

This pr refactor the fallback policy to resovle several issues:

1. A columnar BHJ fallback to row-based BHJ but it has a coumnar broadcast exchange 
2. A vanilla ColumnarToRow on the top of columnar shuffle exchange 
3. The fallback number does not count correctly
4. The fallback reason does not been catched

The main change is in `ColumnarOverrideRules`, it splits postColumnarTransitions into 3 parts:
1. fallbackPolicy
    Used to check if we should fallback according to the config `wholeStageFallbackThreshold` or `queryFallbackThreshold`. Make a new rule PlanFallbackPolicy to hold related code
2. postOverrides
    Same with previous postOverrides
3. finallyRules
    Used to report fallback reason and cleanup tags. These rules are moved from preOverrides.

This pr also removes the config `spark.gluten.sql.columnar.fallback.policy`, it is really confused to user. We can decide to use which config by whether we are inside AQE or not.

## How was this patch tested?

Add test

Before this pr, the bug can be reproduced by:

```shell
./bin/spark-shell --conf spark.gluten.sql.columnar.fallback.policy=stage --conf spark.gluten.sql.columnar.wholeStage.fallback.threshold=2

spark.sql("SELECT *, java_method('java.lang.Integer', 'sum', t1.c1, t2.c1) FROM t1 LEFT JOIN t2 on t1.c1 = t2.c1").show
```


Yohahaha:
> This pr also removes the config spark.gluten.sql.columnar.fallback.policy, it is really confused to user. 

I don think so, maybe fallback policy has some bugs, but it's better than threshold for me. 

We could keep two policy: stage, query, and remove operator policy, cause operator could bring lots of c2r/r2c and may break spark whole stage codegen.

ulysses-you:
@Yohahaha if you are looking at the usage of spark.gluten.sql.columnar.fallback.policy, you will find:
1. if the plan is AQEed and policy = query, then it will do nothing
2. if the plan is not AQEed and policy = stage, then it also do nothing

So, the behavior of policy config is according to the AQE. We just need apply stage threshold in AQE and apply query threshold in non-AQE. Is there any other cases can not been covered ?

Yohahaha:
> @Yohahaha if you are looking at the usage of spark.gluten.sql.columnar.fallback.policy, you will find:
> 
> 1. if the plan is AQEed and policy = query, then it will do nothing
> 2. if the plan is not AQEed and policy = stage, then it also do nothing
> 
> So, the behavior of policy config is according to the AQE. We just need apply stage threshold in AQE and apply query threshold in non-AQE. Is there any other cases can not been covered ?

Thanks for detailed explanation, I have not look related code deeply for now and sorry for can not provide more cases and feedback, my only opinion is keep fallback policy, it looks more comprehensible.

ulysses-you:
@Yohahaha the fallback policy is still existed, but just does not need to change it manually to free our time. The story is that, if we are in AQE, then we are using stage fallback policy. If we are not in AQE, then we are using query fallback policy.

ulysses-you:
also cc @zhouyuan @PHILO-HE thank you

gluten-core/src/main/scala/io/glutenproject/extension/PlanFallbackPolicy.scala
```diff
@@ -0,0 +1,164 @@
...
+        case ColumnarToRowExec(s: Exchange) if s.supportsColumnar && isAdaptiveContext =>
+          countFallback(s)
+        case ColumnarToRowExec(p: GlutenPlan) =>
+          logDebug(s"Find a columnar to row for gluten plan:\n$p")
+          fallbacks = fallbacks + 1
+          countFallback(p)
+        case leafPlan: LeafExecNode if !leafPlan.isInstanceOf[GlutenPlan] =>
+          // Possible fallback for leaf node.
+          fallbacks = fallbacks + 1
+        case p => p.children.foreach(countFallback)
```
zhztheplayer, line 86: Would you like to explain a bit more about this case? Thanks.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -798,99 +804,26 @@ case class ColumnarOverrideRules(session: SparkSession)
...
-        insertTransitions(originalPlan, outputsColumnar = false)
-      } else if (fallbackPolicy == "stage" && isAdaptiveContext && fallbackWholeStage(plan)) {
-        logWarning("Fall back the plan due to meeting the whole stage fallback threshold!")
-        insertTransitions(originalPlan, outputsColumnar = false)
-      } else {
-        transformPlan(postOverrides(), plan, "post")
+      val planWithFallbackPolicy = transformPlan(fallbackPolicy(), plan, "fallback")
+      val finalPlan = planWithFallbackPolicy match {
+        case FallbackNode(fallbackPlan) =>
+          fallbackPlan
```
zhztheplayer, line 819: Would you like to add one-line comment to describe why post overrides are skipped? Thanks.

zhouyuan:
@ulysses-you thanks for improving on this! will look today.

thanks, -yuan

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -798,99 +804,26 @@ case class ColumnarOverrideRules(session: SparkSession)
...
     PhysicalPlanSelector.maybe(session, plan) {
-      if (fallbackPolicy == "query" && !isAdaptiveContext && fallbackWholeQuery(plan)) {
-        logWarning("Fall back to run the query due to unsupported operator!")
-        insertTransitions(originalPlan, outputsColumnar = false)
-      } else if (fallbackPolicy == "stage" && isAdaptiveContext && fallbackWholeStage(plan)) {
-        logWarning("Fall back the plan due to meeting the whole stage fallback threshold!")
-        insertTransitions(originalPlan, outputsColumnar = false)
-      } else {
-        transformPlan(postOverrides(), plan, "post")
+      val planWithFallbackPolicy = transformPlan(fallbackPolicy(), plan, "fallback")
```
zhztheplayer, line 816: Just out of my curiosity, would it be possible to apply `fallbackPolicy()` in preColumnarTransitions?

gluten-core/src/main/scala/io/glutenproject/extension/PlanFallbackPolicy.scala
```diff
@@ -0,0 +1,164 @@
...
+ *              |
+ *       VeloxColumnarToRow
+ *              |
+ *           Project (unsupport columnar)
+ *
+ * @param isAdaptiveContext If is inside AQE
+ * @param originalPlan The vanilla SparkPlan without apply gluten transform rules
+ */
+// spotless:on
+case class PlanFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkPlan)
```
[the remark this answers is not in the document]
ulysses-you: how about `ExpandFallbackPolicy` ?

gluten-core/src/main/scala/io/glutenproject/extension/PlanFallbackPolicy.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  private def countFallbacks(plan: SparkPlan): Int = {
+    var fallbacks = 0
+    def countFallback(plan: SparkPlan): Unit = {
+      plan match {
+        case _: QueryStageExec => // Another stage.
+        case _: CommandResultExec | _: ExecutedCommandExec => // ignore
+        // we plan exchange to columnar exchange in columnar rules and the exchange does not
+        // support columnar, so the output columnar is always false in AQE postStageCreationRules
+        case ColumnarToRowExec(s: Exchange) if s.supportsColumnar && isAdaptiveContext =>
```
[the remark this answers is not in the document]
ulysses-you: There are not been counted.  Vanilla c2r is still existed even if we fallback, so we only care about gluten c2r.

gluten-core/src/main/scala/io/glutenproject/extension/PlanFallbackPolicy.scala
ulysses-you: sure, it is used to find the next cr2. Say a plan:
```
   Project
      |
RowToColumnar
      |
    Filter (unsupport columnar)
      |
ColumnarToRow
      |
     Scan

```

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
ulysses-you: The main issue is `postOverrides` will conver vanilla c2r to velox cr2. If we decide to fallback, we should use vanilla c2r

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
ulysses-you: There are two things need to do if we want to move `fallbackPolicy()` to preColumnarTransitions.
- We count the c2r inside `fallbackPolicy()`, and c2r only happens in postColumnarTransitions
- We want to skip `postOverrides` if we decide to fallback

So seems it's simpler to do it at postColumnarTransitions

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -0,0 +1,164 @@
...
+import org.apache.spark.sql.execution.exchange.Exchange
+
+// spotless:off
+/**
+ * Note, this rule should only fallback to row-based plan if there is no harm.
+ * The follow case should be handled carefully
+ *
+ * 1. A BHJ and the previous broadcast exchange is columnar
+ *    We should still make the BHJ columnar, otherwise it will fail if
+ *    the vanilla BHJ accept a columnar broadcast exchange, e.g.,
```
zhouyuan, line 39: maybe off topic: there's another case: columnar broadcast xchg + row broadcast join(due to unsupported). In this case, the possible fix is to add another adapter to convert the broadcast xchg back

docs/Configuration.md
```diff
@@ -40,7 +40,6 @@ You can add these configurations into spark-defaults.conf to enable or disable t
...
 | spark.gluten.sql.native.bloomFilter                     | Enable or Disable native runtime bloom filter.                                                                                                                                                                                                                                                                                                                                                                                                       | true                                                 |
-| spark.gluten.sql.columnar.fallback.policy               | setting different fallback policies. The options are: "operator" -> fallback on each operator if not supported. "stage" -> fallback on each stage if the number of unsupported operator/expressions in that stage >= `spark.gluten.sql.columnar.wholeStage.fallback.threshold` in AQE env . "query" -> fallback whole query if the number of unsupported operator/expressions >= `spark.gluten.sql.columnar.query.fallback.threshold` in non-AQE env | "operator"                                           |
```
zhouyuan, line 43: The initial idea is to define a basic cost model to calculate the cost on fallback, then apply the different policies at query/stage level. The current design is a sort of mixed, and it's confusing, I agree to remove this part and use the two ` threshold ` first.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -0,0 +1,164 @@
...
+    val planWithColumnarToRow = InsertTransitions.insertTransitions(originalPlan, false)
+    planWithColumnarToRow.transform {
+      case c2r @ ColumnarToRowExec(_: ShuffleQueryStageExec) =>
+        transformPostOverrides.transformColumnarToRowExec(c2r)
+    }
+  }
+
+  override def apply(plan: SparkPlan): SparkPlan = {
+    val reason = fallback(plan)
+    if (reason.isDefined) {
```
zhouyuan, line 150: may worth to add a warning log to indicate the stage/query is running fallback now

zhouyuan (commented):
the previous code on fallback policies is confusing indeed, I think this is a good change, but I have some doubts on AQE fallback code path. Will need to do some local tests to verify first and update back. 
Also ping @zzcclp

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: > We count the c2r inside fallbackPolicy(), and c2r only happens in postColumnarTransitions

Would it be feasible to count on the conjective non-transformable tags?

I think C2R is unlikely reliable to indicate fallback: it should be based on the assumption that Vanilla Spark is always row-based, while Gluten Spark is always not. However the assumption can be breaked since it's somewhat implicit. For example sometime Vanilla Spark has columanr processing either, some day Gluten may have the ability to fallback row processing to native, etc.

I saw you added "InsertTransitions" which does the similar job comparing to vanilla Spark's. Moving the logic to preColumnarTransitions can leave that job to vanilla Spark.

You can consider the comment non-blocking since it requires a lot of changes. Since the PR already makes things clearer than before.

gluten-core/src/main/scala/io/glutenproject/extension/PlanFallbackPolicy.scala
zhztheplayer: Please ignore, I was refering to the previous case branch, not this one. However I noticed that was for leaf nodes like scans. Looking good to me.

zhouyuan:
@ulysses-you 
update: 
I tried this on tpcds q28 which contains 5 BNLJ(not supported in Gluten) and it seems the last stage is not correctly fallback with wholestage.fallback.threashold = 4

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
ulysses-you: yeah, I like this idea. We can also change other rules. e.g., `FallbackBroadcastExchange` can be removed if we support transform columnar broadcast exchange to row-based or transform row-based broadcast exchange to columnar. I think we can do that in a new pr.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
ulysses-you: conjective non-transformable looks good to me, and we can even support partial fallback in one stage. We'd better to consider the cost of each operator. e.g., if one stage has a non-transformable project and several transformable joins, shall we fallack it.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
ulysses-you: yeah, it's wroth. Now, it follows the `GlutenFallbackReporter`. It would print log and post event to Spark UI.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
ulysses-you: addressed

backends-velox/src/test/scala/io/glutenproject/execution/FallbackSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+ */
+package io.glutenproject.execution
+
+import io.glutenproject.GlutenConfig
+
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.execution.{SparkPlan, VeloxColumnarToRowExec}
+import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
+
+class FallbackSuite extends WholeStageTransformerSuite with AdaptiveSparkPlanHelper {
```
zzcclp, line 25: why this suite only for  velox backend ?

ulysses-you: We have ut in `FallbackStrategiesSuite`. This test suite is due to I face a velox backend bug and I think I should leave some end-to-end tests here. I do not have any background/env about CH, so if you think CH backend should have these, then you can add some in followup.

zhouyuan (approved):
👍

backends-velox/src/test/scala/io/glutenproject/execution/FallbackSuite.scala
zhouyuan: https://github.com/oap-project/gluten/issues/2829

PHILO-HE (approved):
LGTM! Just some trivial comments, not blocking the merging.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  private def countFallbacks(plan: SparkPlan): Int = {
+    var fallbacks = 0
+    def countFallback(plan: SparkPlan): Unit = {
+      plan match {
+        case _: QueryStageExec => // Another stage.
+        case _: CommandResultExec | _: ExecutedCommandExec => // ignore
+        // we plan exchange to columnar exchange in columnar rules and the exchange does not
+        // support columnar, so the output columnar is always false in AQE postStageCreationRules
+        case ColumnarToRowExec(s: Exchange) if s.supportsColumnar && isAdaptiveContext =>
```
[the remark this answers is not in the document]
ulysses-you: I will do a followup for these.

PR #2768, opened 2023-08-15
[VL] Branch: upstream_velox: Adopt arbitrator factory API

zhztheplayer:


rui-mo:
@zhztheplayer Thanks for your fix!

Issue #2498, opened 2023-07-28
[CH] gluten submit additional jobs to get input paths


shuai-xu:
**Describe the bug**
Such as shown in bellow:
<img width="1439" alt="image" src="https://github.com/oap-project/gluten/assets/20960838/5b7fdb19-fd6f-4989-8e6f-6d1dabf23c6f">
<img width="1434" alt="image" src="https://github.com/oap-project/gluten/assets/20960838/a750fa7c-ae4f-46dd-8cc7-02f44b6f349f">

gluten submit some jobs to list directory and get the input files. But spark with same config didn't. This cause a preformance degradation。
The jobs are submitted by：
<img width="1189" alt="image" src="https://github.com/oap-project/gluten/assets/20960838/2fbaa264-96d7-4b53-9c61-8d64d36764d4">

shuai-xu:
These job are submitted by gluten When doing ColumnarOverride. 
<img width="916" alt="image" src="https://github.com/oap-project/gluten/assets/20960838/00f9d74d-a07c-48a2-83b9-08dc9e4cffea">

taiyang-li:
```
[...]

PR #2773, opened 2023-08-15
[GLUTEN-2498] Remove `FileScan` in `HiveTableScanExecTransformer`

baibaichen:

Currently, We  use `CatalogFileIndex` to list files, it eventually use `InMemoryFileIndex` which could create spark job. It is much slower than vanilla spark's implemation which uses `FileInputFormat.getSplits()`

This PR simulates  `FileInputFormat.getSplits()`,  but only use one thread. We can support mutiple threads latter.

(Fixes: \#2498)

## How was this patch tested?
UT


gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenTextBasedScanWrapper.scala
```diff
@@ -1,92 +0,0 @@
...
-
-  def withFilters(partitionFilters: Seq[Expression], dataFilters: Seq[Expression]): FileScan = {
-    scan match {
-      case t: TextScan => t.copy(partitionFilters = partitionFilters, dataFilters = dataFilters)
-      case j: JsonScan => j.copy(partitionFilters = partitionFilters, dataFilters = dataFilters)
-      case _ => scan
-    }
-  }
-
-  def compressionSplittable(path: Path): Boolean = {
```
KevinyhZou, line 70: Hello, Perhaps we should consider the compression text file not support split currently, like bzip2, lz4 which not support split by clickhouse. So it best to set unsplittable when the text file is compressed

baibaichen: Add test 
image

and using `BackendsApiManager` to decide whether split or not
image

KevinyhZou:
LGTM

zzcclp (approved):
LGTM

Issue #2762, opened 2023-08-15
TPCDS q4 need more memory and run slower after memory arbitrator 


Yohahaha:
### Backend

VL (Velox)

### Bug description

Before memory arbitrator enabled, we run sf1000 TPCDS with 4core 8g(off-heap) per executor successfully, and total time was `29.45`.

After that feature, we failed with same memory config, but success with 10g(off-heap), and total time was `37.6`.

Yohahaha:
CC @zhouyuan @zhztheplayer

Yohahaha:
Before: stage metrics shows peek memory was 56MB.
After: 2.4GiB.

Yohahaha:
@zhouyuan 
"before" means commits end up to 7.25, "after" means newly commits.

"before" not trigger spill, "after" trigger spill, but default spill strategy config does not passed to native correctly by default(will submit PR to fix), so spill failed.

Yohahaha:
https://github.com/oap-project/gluten/pull/2769

zhztheplayer:
> Before: stage metrics shows peek memory was 56MB. After: 2.4GiB.

This is something abnormal. Can you share more details about this case?

Yohahaha:
I probably know the root issue which cause task peek memory always increase, let me try to fix.

zhztheplayer:
> @zhouyuan "before" means commits end up to 7.25, "after" means newly commits.
> 
> "before" not trigger spill, "after" trigger spill, but default spill strategy config does not passed to native correctly by default(will submit PR to fix), so spill failed.

Sorry, after some investigation it's probably related to a recent Velox rebase action https://github.com/oap-project/gluten/pull/2370/files#diff-013353c1aa2ebe3f4c5c014b936dd52b2f91bef0a51421c1a901a57cac361f17R83 to change default strategy by mistake.

I've applied your PR https://github.com/oap-project/gluten/pull/2769, it should be back to normal.

Yohahaha:
I could share the memory config of TPCDS sf1000 in our benchmark, each executor has 4cores and 8G off-heap and 2G overhead and would not trigger spill.

As we talked in the meeting @weiting-chen  , your guys could find suitable memory config of your benchmark which has less memory but not trigger spill, thus we can guarantee follow changes would not cause peek memory increase.

PR #2786, opened 2023-08-17
[GLUTEN-2762] [VL] Release shuffle reader memory early to decrease peek memory usage

Yohahaha:

We found some query has high task peek memory usage after integrate memory arbitrator, e.g. before this patch, TPCDS sf1000 q4 use 2.4GiB peek memory in its biggest stage, after this patch, it use 384MiB instead.

And this patch only fix shuffle reader memory usage which is the most huge part, there still some memory consumers need release early to decrease peek usage, will fix these in followup patches.

(Fixes: #2762)

## How was this patch tested?

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleReader.cc
+++ b/cpp/velox/shuffle/VeloxShuffleReader.cc
@@ -321,6 +321,10 @@ VeloxShuffleReader::VeloxShuffleReader(
 arrow::Result<std::shared_ptr<ColumnarBatch>> VeloxShuffleReader::next() {
   ARROW_ASSIGN_OR_RAISE(auto batch, Reader::next());
   if (batch == nullptr) {
+    // Some empty batch has not been released here, so check is needed.
+    if (veloxPool_->currentBytes() == 0 || veloxPool_->reservedBytes() == 0) {
+      veloxPool_->shrinkManaged(veloxPool_.get(), veloxPool_->capacity());
+    }
     return nullptr;
   }
   auto rb = std::dynamic_pointer_cast<ArrowColumnarBatch>(batch)->getRecordBatch();
--- a/cpp/velox/shuffle/VeloxShuffleReader.cc
+++ b/cpp/velox/shuffle/VeloxShuffleReader.cc
@@ -322,7 +322,7 @@ arrow::Result<std::shared_ptr<ColumnarBatch>> VeloxShuffleReader::next() {
   ARROW_ASSIGN_OR_RAISE(auto batch, Reader::next());
   if (batch == nullptr) {
     // Some empty batch has not been released here, so check is needed.
-    if (veloxPool_->currentBytes() == 0 || veloxPool_->reservedBytes() == 0) {
+    if (veloxPool_->currentBytes() == 0 && veloxPool_->reservedBytes() == 0) {
       veloxPool_->shrinkManaged(veloxPool_.get(), veloxPool_->capacity());
     }
     return nullptr;
--- a/cpp/core/memory/ArrowMemoryPool.cc
+++ b/cpp/core/memory/ArrowMemoryPool.cc
@@ -61,7 +61,7 @@ int64_t ArrowMemoryPool::num_allocations() const {
 }
 
 std::string ArrowMemoryPool::backend_name() const {
-  return "gluten allocator";
+  return "gluten arrow allocator";
 }
 
 } // namespace gluten


Yohahaha:
CC @zhouyuan @zhztheplayer @kerwin-zk

cpp/velox/shuffle/VeloxShuffleReader.cc
```diff
@@ -321,6 +321,10 @@ VeloxShuffleReader::VeloxShuffleReader(
 arrow::Result<std::shared_ptr<ColumnarBatch>> VeloxShuffleReader::next() {
   ARROW_ASSIGN_OR_RAISE(auto batch, Reader::next());
   if (batch == nullptr) {
+    // Some empty batch has not been released here, so check is needed.
+    if (veloxPool_->currentBytes() == 0 || veloxPool_->reservedBytes() == 0) {
```
kerwin-zk, line 325: && is better.

zhouyuan:
@Yohahaha thanks for the patch, is there some performance difference with this fix?

cpp/velox/shuffle/VeloxShuffleReader.cc
```diff
@@ -321,6 +321,10 @@ VeloxShuffleReader::VeloxShuffleReader(
 arrow::Result<std::shared_ptr<ColumnarBatch>> VeloxShuffleReader::next() {
   ARROW_ASSIGN_OR_RAISE(auto batch, Reader::next());
   if (batch == nullptr) {
+    // Some empty batch has not been released here, so check is needed.
+    if (veloxPool_->currentBytes() == 0 && veloxPool_->reservedBytes() == 0) {
+      veloxPool_->shrinkManaged(veloxPool_.get(), veloxPool_->capacity());
+    }
```
zhztheplayer, line 327: veloxPool_ can't be reused by next shuffle reader, right? So the number of used memory keeps increasing.

We may avoid explicit calls to `shrinkManaged` since it's not a public/stable Velox API. Would you like to help find out if we can reuse the pool here, probably? Thanks!

Yohahaha: > veloxPool_ can't be reused by next shuffle reader, right?

Yes, each shuffle reader stream has its pool, and each `ShuffleReaderJniWrapper#make` call will create a new pool. 

> Would you like to help find out if we can reuse the pool here, probably?

That's a better way for me too, and it may bring lots of refactoring, current way is just a hot fix for now, if you accept this idea.

I actually want to a whole refactoring/optimization as I discussed with you, but I need learn both Gluten and Velox codes more deeply and will keep working on it.

Yohahaha: And I think reuse is not the key, each memory consumer should support spill is the key, and some spill operations just shrink itself.

Yohahaha:
> @Yohahaha thanks for the patch, is there some performance difference with this fix?

In TPCDS sf1000:
before this patch, q4,q23b,q24a/b will OOM, rest of successful query's total time is 509s.
after this patch, failed query will success, same query's total time is 498s.

cpp/velox/shuffle/VeloxShuffleReader.cc
zhztheplayer: > And I think reuse is not the key, each memory consumer should support spill is the key

The Velox pools should be all give their reservations back to Spark once destructed. Currently shuffle reader is a anti-case since all the pools are destructed at the point that task ends, that's too late. We actually have several solutions:

option 1: This patch
option 2: Reuse shuffle reader's memory pool
option 3: Close shuffle reader early, before task ends
option 4: Let memory pool reports it's used bytes to Spark, not capacity

I think we can first implement 1, maybe 2 and 3 too, to make this whole shuffle reader stuff more robust that it is.

Yohahaha: op3 is similar with op1, but check of currentBytes and TaskResources.addRecycle are still needed, cause I found some empty batch will exist when current shuffle read stream close, we need consider more careful way for batch later.

op4 seems same with origin design which make spark consumer bind with allocator, now spark consumer bind with arbitrator. I think bind with reservation/capacity changes is better.

zhztheplayer: Thanks for the explanation.

Let's apply this patch as a temporary solution, since the API `shrinkManaged` will not last for long time there for gluten.

Once op2 is implemented, we can drop this part of code. What do you think about it?

zhztheplayer (approved):
Thanks!

cpp/velox/shuffle/VeloxShuffleReader.cc
Yohahaha: > since the API shrinkManaged will not last for long time there for gluten.

Why and how do you shrink when nativeSpill be called if this API be removed?

> Once op2 is implemented, we can drop this part of code. What do you think about it?

LGTM.

zhouyuan:
I did some tests with 20 rounds of q67, the peak mem size of one big reduce stage decreased from 936MB  to 736MB indeed.
however overall the mem footprint(RSS) seems the same, I didn't dig deeper on this.
image

zhouyuan (approved):
👍  works for me

Yohahaha:
> however overall the mem footprint(RSS) seems the same.

what is mem footprint? could you share the way to measure it?

Yohahaha:
Thanks, I got your points. That's the reason for we can not simply close then erase shuffle reader handle, there still has lots of batches exists in somewhere. Since some helpful PR has merged, you can invoke TaskMemoryManager#showMemoryUsage in TaskResources.scala#L146 to find the footprint.

Issue #2785, opened 2023-08-17
[CH] Optimize the setting of lib path


exmy:
### Description

Currently, we configure the library(libch.so) path using `spark.gluten.sql.columnar.libpath`. 

When running in yarn mode, the library will be uploaded to the YARN container's working directory using --files, which leads to a situation where we must set the libpath as `./libch.so` in order to load the lib on the executor side. This also means that we need to set it in the same way on the driver side, which restricts the storage path of the library.

PR #2787, opened 2023-08-17
[GLUTEN-2785][CH] Optimize the setting of lib path

exmy:

Currently, we configure the library(libch.so) path using `spark.gluten.sql.columnar.libpath`.

When running in yarn mode, the library will be uploaded to the YARN container's working directory using `--files`, which leads to a situation where we must set the libpath as `./libch.so` in order to load the lib on the executor side. This also means that we need to set it in the same way on the driver side, which restricts the storage path of the library.

The patch fixes the loading path of the library on the executor side in yarn mode to be the filename of the library.

(Fixes: \#2785)

## How was this patch tested?

manual tests


exmy:
@zzcclp Could you help review this? thx

Issue #1925, opened 2023-06-13
[CH] allow all parquet write jobs (especially for hive ETL jobs) to use native parquet writer


binmahone:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

velox side has https://github.com/oap-project/gluten/pull/1663, but it's still not enough in a hive ETL scenario if a user wants to fully leverage the advantages of native parquet writer.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2799, opened 2023-08-17
[VL] Add RowToColumnarExec to FakeRowAdaptor when use velox native parquet write

WolverineJiang:

Some problems bring in by patch #1925 . This patch will resolve it.

## How was this patch tested?

Run UT by set spark.gluten.sql.native.writer.enabled=true

Run TPCDS by set spark.gluten.sql.native.parquet.writer.enabled=true && spark.sql.sources.useV1SourceList=parquet


gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -194,7 +194,15 @@ object GlutenWriterColumnarRules {
...
                     ))))
-            case other => rc.withNewChildren(Array(FakeRowAdaptor(other)))
+            case other =>
+              if (other.supportsColumnar) {
+                rc.withNewChildren(Array(FakeRowAdaptor(other)))
+              } else {
+                rc.withNewChildren(
+                  Array(
+                    FakeRowAdaptor(
+                      BackendsApiManager.getSparkPlanExecApiInstance.genRowToColumnarExec(other))))
```
PHILO-HE, line 204: Thanks for your PR!
I note `doExecuteColumnar` at `FakeRowAdaptor` has a similar handling for row-based child operator. Doesn't that piece of code work?
If there is any issue found in your test, could you please add some UTs? Thanks!

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -194,7 +194,15 @@ object GlutenWriterColumnarRules {
                       aqe.isSubquery,
                       supportsColumnar = true
                     ))))
-            case other => rc.withNewChildren(Array(FakeRowAdaptor(other)))
+            case other =>
+              if (other.supportsColumnar) {
+                rc.withNewChildren(Array(FakeRowAdaptor(other)))
```
jackylee-ch, line 199: Please add  some UTs to show why we should do this and confirm problem fixed.

jackylee-ch:
@PHILO-HE I just checked the problem @WolverineJiang  met. it has nothing to do with the current pr, he did not override the `ParquetFileFormat` through `extraClassPath`. Thus I will close this pr now.

PHILO-HE:
> @PHILO-HE I just checked the problem @WolverineJiang met. it has nothing to do with the current pr, he did not override the `ParquetFileFormat` through `extraClassPath`. Thus I will close this pr now.

No problem. Thanks for your clarification!

PR #2802, opened 2023-08-18
[GLUTEN-2803][VL] Use a single native shuffle reader instance per Spark reducer task

zhztheplayer:
A single shuffle reader instance in Gluten could hold one Velox memory pool that may reserve at least several MiBs data until Spark task ends. This could be a memory leak source in the case of large read amount.

In the patch we'll expand the shuffle reader scope to let one task create only a single shuffle reader instead of creating the reader per shuffle read fetch (which by default 48MiB).

Ideally this will decrease Gluten's overall Spark off-heap memory occupation.

A simple calculation:

Assuming shuffle read size is 1TiB, with 10 available Spark CPUs, read batch / buffer size is 4096 rows, 100 Byte per row:

Then the peak off-heap memory from shuffle reader (which could cause unnecessary spills even OOM when high):

Before:
1T / 48M * 4096 * 100B = 8.95 GiB

After:
10 * 4096 * 100B = 4 MiB

Note: the reserved Spark off-heap memory may already be released to OS, but not repaid to Spark. The main problem here we want to address is about the Spark off-heap counter. It's unrelated to high process RSS.

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -177,17 +177,14 @@ class JavaInputStreamAdaptor final : public arrow::io::InputStream {
 
 class JniColumnarBatchIterator : public ColumnarBatchIterator {
  public:
-  explicit JniColumnarBatchIterator(
-      JNIEnv* env,
-      jobject javaserializedColumnarBatchIterator,
-      std::shared_ptr<ArrowWriter> writer)
+  explicit JniColumnarBatchIterator(JNIEnv* env, jobject jColumnarBatchItr, std::shared_ptr<ArrowWriter> writer)
       : writer_(writer) {
     // IMPORTANT: DO NOT USE LOCAL REF IN DIFFERENT THREAD
     if (env->GetJavaVM(&vm_) != JNI_OK) {
       std::string errorMessage = "Unable to get JavaVM instance";
       throw gluten::GlutenException(errorMessage);
     }
-    javaserializedColumnarBatchIterator_ = env->NewGlobalRef(javaserializedColumnarBatchIterator);
+    jColumnarBatchItr_ = env->NewGlobalRef(jColumnarBatchItr);
   }
 
   // singleton, avoid stack instantiation
@@ -199,20 +196,20 @@ class JniColumnarBatchIterator : public ColumnarBatchIterator {
   virtual ~JniColumnarBatchIterator() {
     JNIEnv* env;
     attachCurrentThreadAsDaemonOrThrow(vm_, &env);
-    env->DeleteGlobalRef(javaserializedColumnarBatchIterator_);
+    env->DeleteGlobalRef(jColumnarBatchItr_);
     vm_->DetachCurrentThread();
   }
 
   std::shared_ptr<ColumnarBatch> next() override {
     JNIEnv* env;
     attachCurrentThreadAsDaemonOrThrow(vm_, &env);
-    if (!env->CallBooleanMethod(javaserializedColumnarBatchIterator_, serializedColumnarBatchIteratorHasNext)) {
+    if (!env->CallBooleanMethod(jColumnarBatchItr_, serializedColumnarBatchIteratorHasNext)) {
       checkException(env);
       return nullptr; // stream ended
     }
 
     checkException(env);
-    jlong handle = env->CallLongMethod(javaserializedColumnarBatchIterator_, serializedColumnarBatchIteratorNext);
+    jlong handle = env->CallLongMethod(jColumnarBatchItr_, serializedColumnarBatchIteratorNext);
     checkException(env);
     auto batch = columnarBatchHolder.lookup(handle);
     if (writer_ != nullptr) {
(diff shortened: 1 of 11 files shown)


zhouyuan:
@zhztheplayer this is a great improvement, can you please create one issue so we can track on it more easily?

-yuan

cpp/core/jni/JniWrapper.cc
```diff
@@ -959,19 +959,17 @@ JNIEXPORT void JNICALL Java_io_glutenproject_vectorized_OnHeapJniByteInputStream
 JNIEXPORT jlong JNICALL Java_io_glutenproject_vectorized_ShuffleReaderJniWrapper_make( // NOLINT
     JNIEnv* env,
     jobject,
-    jobject jniIn,
     jlong cSchema,
     jlong allocId,
     jstring compressionType,
     jstring compressionBackend) {
   JNI_METHOD_START
+  ReaderOptions options = ReaderOptions::defaults();
```
winningsix, line 967: hmm, why need to move from L974 to L967?

cpp/core/jni/JniWrapper.cc
```diff
@@ -179,15 +179,15 @@ class JniColumnarBatchIterator : public ColumnarBatchIterator {
  public:
   explicit JniColumnarBatchIterator(
       JNIEnv* env,
-      jobject javaserializedColumnarBatchIterator,
+      jobject javaSerializedColumnarBatchIterator,
```
winningsix, line 182: Minor: any way to simplify the name? e.g., ```jColumnarBatchItr``` better?

gluten-celeborn/src/main/scala/org/apache/spark/shuffle/CelebornColumnarBatchSerializer.scala
```diff
@@ -17,11 +17,10 @@
 package org.apache.spark.shuffle
 
 import io.glutenproject.GlutenConfig
-import io.glutenproject.columnarbatch.ColumnarBatches
 import io.glutenproject.memory.alloc.NativeMemoryAllocators
 import io.glutenproject.memory.arrowalloc.ArrowBufferAllocators
 import io.glutenproject.utils.ArrowAbiUtil
-import io.glutenproject.vectorized.{JniByteInputStreams, ShuffleReaderJniWrapper}
+import io.glutenproject.vectorized._
```
winningsix, line 23: Minor: only import necessary packages?

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
```diff
@@ -70,56 +69,66 @@ private class ColumnarBatchSerializerInstance(
   extends SerializerInstance
   with Logging {
 
+  private lazy val shuffleReaderHandle = {
```
winningsix, line 72: Not directly related to this PR. It seems [CelebornColumnarBatchSerializer.scala](https://github.com/oap-project/gluten/pull/2802/files#diff-89f2f350ebb871afee74677e5c52b05c85795d3f07080d0edd1895b2bf703f80) quite similar to this class. Can we reuse this?

winningsix (commented):
Just a few minor comments.

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
zhztheplayer: It makes a lot of sense to me.

The consolidation of the two serializers was in our original plan https://github.com/oap-project/gluten/issues/1205

> In Java, consolidate GlutenColumnarBatchSerializer and CelebornColumnarBatchSerializer

But somehow we didn't make the change yet. We should do it asap since the code brings extra maintenance burden to developers.

cpp/core/jni/JniWrapper.cc
zhztheplayer: fixed

cpp/core/jni/JniWrapper.cc
zhztheplayer: fixed

gluten-celeborn/src/main/scala/org/apache/spark/shuffle/CelebornColumnarBatchSerializer.scala
zhztheplayer: fixed

Yohahaha:
Do you tested the way that pass input stream as shuffle reader's function input args can ensure only initialize velox pool once?

I think we need modify createNativeMemoryAllocatorManager to createXXXPool instead, and new method return a created velox pool, so we can use same pool to each shuffle reader, cause each ShuffleReaderJni#make will create a new shuffle reader.

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
winningsix: #1205 was closed. How about creating another issues tracking this? Thoughts?

zhztheplayer:
> Do you tested the way that pass input stream as shuffle reader's function input args can ensure only initialize velox pool once?

I am still testing this but it should be a single pool per task. See below code:

https://github.com/oap-project/gluten/pull/2802/files#diff-c1147845fed4ad34faf580505712b04befdd3df0361297f93cb660069bedb7bdR360-R366

`veloxPool_` will be shared to the child streams (which are for different Spark read fetches).

> and new method return a created velox pool, so we can use same pool to each shuffle reader

That's promising. We need to have a design like that.

However that and this one could be two individual improvements to Gluten's shuffle.

Using a single C++ reader instance would provide a layer that manages all Spark read fetches so the layer could be used to cache any contexts that are shared among fetches, not only limited to Velox memory pool. E.g. Arrow pool, schema, codec, etc,

Yohahaha:
hmm... seems this PR is ready and passed CI, I just share the POC result of #2822 , biggest shuffle stage's peek memory decreased from 215MiB to 30MiB in my test case.

zhouyuan:
Benchmark Velox

zhouyuan (approved):
👍  works for me

Issue #2796, opened 2023-08-17
[VL] unix_timestamp("20230817", "yyyyMMdd") return wrong value


jackylee-ch:
### Backend

VL (Velox)

### Bug description

SQL
```
create table test(a string) using parquet;
insert into test values('20230817');
select unix_timestamp(a, 'yyyyMMdd') from test; 
```
Gluten return `1692230400` and Vanilla return `1692201600`.
```

### Spark version

3.3

### Spark configurations

_No response_

### System information

Latest commit: 688bc27f72e4d5ec03a85d3fafc74d80d7d6bab9
Linux: Centos7

### Relevant logs

_No response_

PR #2823, opened 2023-08-21
[GLUTEN-2796][VL] Set default session time zone if not explicitly set by user

PHILO-HE:

In this PR, we introduced a gluten config key to hold default session time zone, which will be used if spark config (spark.sql.session.timeZone) is not explicitly set by user (this handling is quite consistent with spark's).

(Fixes: \#2796)

## How was this patch tested?

Will refactor native config by introducing a dedicated class to hold properties. Will then add native tests for validation.

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -350,6 +350,11 @@ object GlutenConfig {
   val GLUTEN_EXTENDED_EXPRESSION_TRAN_CONF =
     "spark.gluten.sql.columnar.extended.expressions.transformer"
 
+  // A mirror property to Spark's spark.sql.session.timeZone. With this independent property, we
+  // can avoid some Spark UT failure where a specific session time zone is used by just setting
+  // the default time zone.
+  val GLUTEN_SESSION_LOCAL_TIMEZONE_KEY = "spark.gluten.sql.session.timeZone"
+
   // Principal of current user
   val GLUTEN_UGI_USERNAME = "spark.gluten.ugi.username"
   // Tokens of current user, split by `\0`
@@ -388,7 +393,7 @@ object GlutenConfig {
       GLUTEN_SAVE_DIR,
       GLUTEN_TASK_OFFHEAP_SIZE_IN_BYTES_KEY,
       GLUTEN_MAX_BATCH_SIZE_KEY,
-      SQLConf.SESSION_LOCAL_TIMEZONE.key
+      GLUTEN_SESSION_LOCAL_TIMEZONE_KEY
     )
     keys.forEach(
       k => {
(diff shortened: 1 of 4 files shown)


PHILO-HE:
@jackylee-ch, could you please take a review and verify the fix?

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -395,6 +395,12 @@ object GlutenConfig {
       k => {
         if (conf.contains(k)) {
           nativeConfMap.put(k, conf(k))
+        } else {
+          // Session's local time zone must be set. If not explicitly set by user,
+          // its default value is used, consistent with spark.
+          if (k.equals(SQLConf.SESSION_LOCAL_TIMEZONE.key)) {
+            nativeConfMap.put(k, SQLConf.SESSION_LOCAL_TIMEZONE.defaultValueString)
```
jackylee-ch, line 402: Maybe we need to verify if there are other cases with similar problems?

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -395,6 +395,12 @@ object GlutenConfig {
...
+          }
```
zhztheplayer, line 403: Probably can be placed to `GlutenDriverPlugin#setPredefinedConfigs`.

One of the reason is that the default time zone may vary from executor to driver, I assume the one from driver should be chosen? Here we use executor's time zone.

PHILO-HE: Hi @zhztheplayer, thanks for your comment!
Yes, it may be possible that driver/executors are in different default time zone due to the different time zone config for OS. And it is better to have a unique timezone setting from driver side. Just updated this part. Thanks!

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
PHILO-HE: After some checks, it looks there is no other similar issue. Only timezone config needs special handling as of now. Thanks!

PHILO-HE:
The directly setting for `spark.sql.session.timeZone` in gluten can cause some UT failure because some spark UTs do the tests by setting default timezone, instead of setting this spark config.
So in the latest code, we introduced a separate gluten config key to hold default session timezone, which will be used if that spark config is not set by user (this handling is quite consistent with spark's).

zhouyuan (approved):
👍

FelixYBW:
> The directly setting for `spark.sql.session.timeZone` in gluten can cause some UT failure because some spark UTs do the tests by setting default timezone, instead of setting this spark config. So in the latest code, we introduced a separate gluten config key to hold default session timezone, which will be used if that spark config is not set by user (this handling is quite consistent with spark's).

Do you mean spark also have a default timezone?

PHILO-HE:
> > The directly setting for `spark.sql.session.timeZone` in gluten can cause some UT failure because some spark UTs do the tests by setting default timezone, instead of setting this spark config. So in the latest code, we introduced a separate gluten config key to hold default session timezone, which will be used if that spark config is not set by user (this handling is quite consistent with spark's).
> 
> Do you mean spark also have a default timezone?

@FelixYBW, yes, Spark also has a default timezone.  And it's the OS timezone detected.

PR #2824, opened 2023-08-21
[GLUTEN-2836][VL] Fix: memory usage problems caused by complex type

guhaiyan0221:
Complex types of data will only be released when the shufflewriter is released, even if this part of the data has been stored in a RecordBatch This change allows each partition to have a StreamArena, so that after converting to RecordBatch, the memory in StreamArena can be freed immediately

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -964,8 +964,11 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
       auto partition = row2Partition_[row];
       if (complexTypeData_[partition] == nullptr) {
         // TODO: maybe memory issue, copy many times
+        if (arenas_[partition] == nullptr) {
+          arenas_[partition] = std::make_unique<facebook::velox::StreamArena>(veloxPool_.get());
+        }
         complexTypeData_[partition] = std::move(serde_->createSerializer(
-            complexWriteType_, partition2RowCount_[partition], arena_.get(), /* serdeOptions */ nullptr));
+            complexWriteType_, partition2RowCount_[partition], arenas_[partition].get(), /* serdeOptions */ nullptr));
       }
       rowIndexs[partition].emplace_back(IndexRange{row, 1});
     }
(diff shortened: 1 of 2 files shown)


zhztheplayer:
@jinchengchenghh Would you like to take a look?

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -964,8 +964,11 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
       auto partition = row2Partition_[row];
       if (complexTypeData_[partition] == nullptr) {
         // TODO: maybe memory issue, copy many times
+	if (arenas_[partition] == nullptr) {
+	  arenas_[partition] = std::make_unique<facebook::velox::StreamArena>(veloxPool_.get());
+	}
         complexTypeData_[partition] = std::move(serde_->createSerializer(
-            complexWriteType_, partition2RowCount_[partition], arena_.get(), /* serdeOptions */ nullptr));
+            complexWriteType_, partition2RowCount_[partition], arenas_[partition].get(), /* serdeOptions */ nullptr));
```
zhouyuan, line 971: so it will create many "arena_" here in each split(), is there any performance difference found here?

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1338,6 +1341,7 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
       complexTypeData_[partitionId]->flush(&out);
       allBuffers.emplace_back(valueBuffer);
       complexTypeData_[partitionId] = nullptr;
+      arenas_[partitionId] = nullptr;
```
jinchengchenghh, line 1344: Why we release complexTypeData_, but the arenas still occupy  memory?

guhaiyan0221: The variable complexTypeData_ is PrestoVectorSerializer and it doesn't own memory. All the memory used by PrestoVectorSerializer is managed by StreamArena.  StreamArena release its memory by deconstructing  memory::Allocation . Allocation (allocations_ and allocation_ of StreamArena) is deconstructed only when StreamArena is deconstructed.

jinchengchenghh (approved):

PR #2831, opened 2023-08-21
[VL] Remove shuffle write schema option

jinchengchenghh:
This option is default false, the true mode is only used in test code, and not suggest to change to the true mode.
If switch to true mode, shuffle will write serialized schema firstly, for some small batch, it is a high load. And we can get correct schema from spark plan when constructing ShuffleReader.

Refactor the test code to remove write_schema option.

### DIFF
--- a/cpp/core/shuffle/LocalPartitionWriter.cc
+++ b/cpp/core/shuffle/LocalPartitionWriter.cc
@@ -84,10 +84,6 @@ class PreferEvictPartitionWriter::LocalPartitionWriterInstance {
     const auto& dataFileOs = partitionWriter_->dataFileOs_;
     ARROW_ASSIGN_OR_RAISE(auto before_write, dataFileOs->Tell());
 
-    if (shuffleWriter_->options().write_schema) {
-      RETURN_NOT_OK(writeSchemaPayload(dataFileOs.get()));
-    }
-
     if (spilledFileOpened_) {
       RETURN_NOT_OK(spilledFileOs_->Close());
       RETURN_NOT_OK(mergeSpilled());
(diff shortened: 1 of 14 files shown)


marin-ma:
Thanks for the cleanup. Please rebase.

marin-ma:
Could you rebase again? There are several conflicts.

cpp/core/shuffle/LocalPartitionWriter.cc
```diff
@@ -335,10 +316,6 @@ arrow::Status PreferCachePartitionWriter::stop() {
       // 5. read if partition exists in the spilled file and write to the final file
       if (partitionSpillInfo.partitionId == pid) { // A hit
         if (firstWrite) {
```
winningsix, line 318: Do we still need to keep ```firstWrite```? It will do nothing after avoiding schema write.

cpp/core/shuffle/LocalPartitionWriter.cc
```diff
@@ -366,10 +339,6 @@ arrow::Status PreferCachePartitionWriter::stop() {
     ARROW_ASSIGN_OR_RAISE(auto rb, shuffleWriter_->createArrowRecordBatchFromBuffer(pid, true));
     if (rb) {
       if (firstWrite) {
```
winningsix, line 341: same as above.

cpp/core/shuffle/LocalPartitionWriter.cc
```diff
@@ -351,10 +328,6 @@ arrow::Status PreferCachePartitionWriter::stop() {
     auto cachedPayloadSize = shuffleWriter_->partitionCachedRecordbatchSize()[pid];
     if (cachedPayloadSize > 0) {
       if (firstWrite) {
```
winningsix, line 330: same as above.

jinchengchenghh: Need it, if not first write, will writeEOS.

winningsix: Oh, I was meaning this if-block seems always toggle ```firstWrite``` to ```false```. We can simplify it to ```firstWrite = false;```.

jinchengchenghh:
Can we merge it? Thanks! @marin-ma

marin-ma (approved):

Issue #2833, opened 2023-08-21
[CH] Avoid unnecessary copy of std::vector in c++ 


taiyang-li:
### Description

For example: 

![image](https://github.com/oap-project/gluten/assets/8181003/2e0ea446-7410-48c7-8af9-14cef9c9f743)

PR #2835, opened 2023-08-21
[GLUTEN-2833] [CH] Avoid unnecessary copy of temporary variables

baibaichen:

Avoid unnecessary copy of temporary variables

(Fixes: \#2833)

## How was this patch tested?

using existed sytle


zzcclp (approved):
LGTM

PR #2859, opened 2023-08-23
[VL] Add shuffle split buffer size metrics

marin-ma:


cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -134,9 +134,9 @@ int64_t getBatchNbytes(const arrow::RecordBatch& rb) {
 }
 
 std::shared_ptr<arrow::Array> makeNullBinaryArray(std::shared_ptr<arrow::DataType> type, ShuffleBufferPool* pool) {
-  std::shared_ptr<arrow::Buffer> offsetBuffer;
+  std::shared_ptr<arrow::ResizableBuffer> offsetBuffer;
   size_t sizeofBinaryOffset = sizeof(arrow::LargeStringType::offset_type);
-  GLUTEN_THROW_NOT_OK(pool->allocate(offsetBuffer, sizeofBinaryOffset * 2));
+  GLUTEN_THROW_NOT_OK(pool->allocateDirectly(offsetBuffer, sizeofBinaryOffset * 2));
```
marin-ma, line 139: @jinchengchenghh Change to `allocateDirectly` here to avoid allocating compression buffers from combined buffer.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -156,9 +156,9 @@ std::shared_ptr<arrow::Array> makeBinaryArray(
     return makeNullBinaryArray(type, pool);
   }
 
-  std::shared_ptr<arrow::Buffer> offsetBuffer;
+  std::shared_ptr<arrow::ResizableBuffer> offsetBuffer;
   size_t sizeofBinaryOffset = sizeof(arrow::LargeStringType::offset_type);
-  GLUTEN_THROW_NOT_OK(pool->allocate(offsetBuffer, sizeofBinaryOffset * 2));
+  GLUTEN_THROW_NOT_OK(pool->allocateDirectly(offsetBuffer, sizeofBinaryOffset * 2));
```
marin-ma, line 161: Same as above.

gluten-data/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -84,7 +84,7 @@ class ColumnarShuffleWriter[K, V](
 
   private var nativeShuffleWriter: Long = -1L
 
-  private var splitResult: SplitResult = _
+  private var splitResult: GlutenSplitResult = _
```
jinchengchenghh, line 87: Why rename `SplitResult` to `GlutenSplitResult`? @zhztheplayer has removed the prefix `Gluten` from other class

jinchengchenghh:
Others look good.

gluten-data/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
marin-ma: It's a new subclass of SplitResult. Because CH backend now has `CHSplitResult` that inherits `SplitResult`, and we both have different metrics in the subclasses.

jinchengchenghh (approved):

PR #2865, opened 2023-08-23
[VL] Fix duplicate metrics updating caused by R2C

rui-mo:

Metrics are updated when `hasNext` becomes false. In RowToVeloxColumnarExec, when child's `hasNext` becomes false, it is called for another time, which causes duplicate metrics updating.

## How was this patch tested?

Verified on Jenkins.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -170,9 +170,6 @@ class IteratorApiImpl extends IteratorApi with Logging {
       }
 
       override def next(): ColumnarBatch = {
-        if (!hasNext) {
-          throw new java.util.NoSuchElementException("End of stream")
-        }
         resIter.next()
       }
     }
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -170,6 +170,9 @@ class IteratorApiImpl extends IteratorApi with Logging {
       }
 
       override def next(): ColumnarBatch = {
+        if (!hasNext) {
+          throw new java.util.NoSuchElementException("End of stream")
+        }
         resIter.next()
       }
     }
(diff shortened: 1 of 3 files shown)


winningsix (commented):
Maybe off the topic. Do we still need [GeneralInIteratorV2](https://github.com/oap-project/gluten/blob/0edd7445781a144136a04502988bd0e1738cefa6/gluten-core/src/main/java/io/glutenproject/vectorized/GeneralInIteratorV2.java#L23)?

rui-mo:
> Maybe off the topic. Do we still need [GeneralInIteratorV2](https://github.com/oap-project/gluten/blob/0edd7445781a144136a04502988bd0e1738cefa6/gluten-core/src/main/java/io/glutenproject/vectorized/GeneralInIteratorV2.java#L23)?

@winningsix GeneralInIteratorV2 looks not in use and can be removed.

rui-mo:
This PR is under test from Meituan. Converting it to draft.

zhztheplayer (dismissed):
Thanks!

After the patch, we will guarantee each `hasNext` (from both Vanilla Spark's and our code) would be called only once for one single item. Is that correct?

rui-mo:
> After the patch, we will guarantee each `hasNext` (from both Vanilla Spark's and our code) would be called only once for one single item. Is that correct?

In R2C, `hasNext` is still called multiple times to append rows into a batch. But once `hasNext` becomes false, it should not be called anymore.

rui-mo:
@zzcclp @liuneng1994 It seems the change to simplify `GeneralInIterator` breaks CH backend. Do you have any clue?
https://github.com/oap-project/gluten/pull/2865/files#diff-fbc4c486f291dec71f3a8a7e0e1e501b9362ec99774d38342db02f2382c47a06L32-R38

zhztheplayer:
> @zzcclp @liuneng1994 It seems the change to simplify `GeneralInIterator` breaks CH backend. Do you have any clue? https://github.com/oap-project/gluten/pull/2865/files#diff-fbc4c486f291dec71f3a8a7e0e1e501b9362ec99774d38342db02f2382c47a06L32-R38

It's weird since the code is not used by CH backend...

zzcclp:
> @zzcclp @liuneng1994 It seems the change to simplify `GeneralInIterator` breaks CH backend. Do you have any clue? https://github.com/oap-project/gluten/pull/2865/files#diff-fbc4c486f291dec71f3a8a7e0e1e501b9362ec99774d38342db02f2382c47a06L32-R38

CH backend has fixed this issue in https://github.com/oap-project/gluten/blob/f1a0a3e6b489b7e3290e75fab5c427201c5198bb/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala#L135C17-L135C17

zhztheplayer (approved):
Thanks!

rui-mo:
A follow up PR for hasNext and next clean up: https://github.com/oap-project/gluten/pull/3490.

PR #2866, opened 2023-08-23
[VL] Enrich Spark task OOM error message

zhztheplayer:
The new OOM log message:

```
23/08/24 19:24:08 ERROR ManagedReservationListener: Error reserving memory from target
io.glutenproject.memory.memtarget.ThrowOnOomMemoryTarget$OutOfMemoryException: Not enough spark off-heap execution memory. Acquired: 8388608, granted: 0. Try tweaking config option spark.memory.offHeap.size to get larger space to run this application. 
Current config settings: 
spark.gluten.memory.offHeap.size.in.bytes=1024.0 MiB
spark.gluten.memory.task.offHeap.size.in.bytes=14.2 MiB
Memory consumer stats:
        from GlutenMemoryConsumer/BuildSideRelation#BatchSerializer@2d68d411: Current used bytes: 8.0 MiB, peak bytes: 8.0 MiB
        from                GlutenMemoryConsumer/WholeStageIterator@218fcda2: Current used bytes: 8.0 MiB, peak bytes: 8.0 MiB
        from              GlutenMemoryConsumer/ArrowContextInstance@32393da0: Current used bytes:   0.0 B, peak bytes: 8.0 MiB
        from                                OverAcquire.DummyTarget@5aa0300e: Current used bytes:   0.0 B, peak bytes: 2.4 MiB
	at io.glutenproject.memory.memtarget.ThrowOnOomMemoryTarget.borrow(ThrowOnOomMemoryTarget.java:77)
	at ...
```

### DIFF
--- a/gluten-core/src/main/java/io/glutenproject/memory/MemoryUsageStats.java
+++ b/gluten-core/src/main/java/io/glutenproject/memory/MemoryUsageStats.java
@@ -14,35 +14,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.glutenproject.memory.nmm;
+package io.glutenproject.memory;
 
-public interface ReservationListener {
-  ReservationListener NOOP =
-      new ReservationListener() {
-        @Override
-        public long reserve(long size) {
-          return 0L;
-        }
+public class MemoryUsageStats {
+  public final long peak;
+  public final long current;
 
-        @Override
-        public long unreserve(long size) {
-          return 0L;
-        }
-
-        @Override
-        public long getUsedBytes() {
-          return 0;
-        }
-
-        @Override
-        public void inactivate() {}
-      };
-
-  long reserve(long size);
-
-  long unreserve(long size);
-
-  void inactivate();
-
-  long getUsedBytes();
+  MemoryUsageStats(MemoryUsage usage) {
+    this.peak = usage.peak();
+    this.current = usage.current();
+  }
 }
--- a/gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
+++ b/gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.memory.memtarget;
+
+import io.glutenproject.GlutenConfig$;
+import io.glutenproject.memory.MemoryUsageStats;
+
+import org.apache.spark.memory.SparkMemoryUtil;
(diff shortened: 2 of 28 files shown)


gluten-core/src/main/java/io/glutenproject/memory/MemoryUsageStats.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.memory;
+
+public class MemoryUsageStats {
+  public final long peak;
+  public final long current;
+
+  public MemoryUsageStats(MemoryUsage usage) {
```
Yohahaha, line 23: Could we pass just long value?

gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public ThrowOnOomMemoryTarget(TaskManagedMemoryTarget target) {
+    this.target = target;
+  }
+
+  @Override
+  public long borrow(long size) {
+    long granted = target.borrow(size);
+    if (granted < size) {
+      // OOM
```
Yohahaha, line 34: Maybe add more description here, "Can not acquire enough memory after spill, throw OOM"

gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
```diff
@@ -0,0 +1,75 @@
...
+      if (granted != 0L) {
```
Yohahaha, line 35: this check can be removed, right?

gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
```diff
@@ -0,0 +1,75 @@
...
+                  + "Try tweaking config option spark.memory.offHeap.size to get larger space "
+                  + "to run this application. %n",
+              size, granted));
+      // Log memory usage
+      target.getTaskMemoryManager().showMemoryUsage();
+      // Dump usages to exception body
+      errorBuilder.append(SparkMemoryUtil.dumpMemoryConsumerStats(target.getTaskMemoryManager()));
+      throw new OutOfMemoryException(errorBuilder.toString());
+    }
+    return granted;
```
Yohahaha, line 52: nit: if (success acquired) { // logic } else { // OOM logic } would more readable.

gluten-core/src/main/java/io/glutenproject/memory/MemoryUsageStats.java
zhztheplayer: The class should be tightly coupled with MemoryUsage.java. We can treat it as a snapshot to MemoryUsage instances.

I'll enclose the ctor to package-private.

gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
zhztheplayer: `target.repay(0)` may lead to unexpected behavior if the underlying target wasn't coded with respect to this case.

Let's keep the check as some sort of defensive practice.

gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
zhztheplayer: Good suggestion however I'd omit the "else" block with fail-fast in if block. Let me make the change.

gluten-core/src/main/java/io/glutenproject/memory/memtarget/ThrowOnOomMemoryTarget.java
zhztheplayer: Sure, I'll add some comment here.

I'd use

```
// OOM happens.
// Note if the target type is a Spark memory consumer, spilling should already be requested but
// failed to reclaim enough memory. 
```

zhouyuan (approved):
👍

PR #2868, opened 2023-08-23
[VL] Move orc decimal filters to the remaining filter list in native scan

kerwin-zk:
At present, the orc filter of velox does not support decimal and timestamp. If the filter of decimal and timestamp is pushed down to scan, the task will directly fail. If scan fallback is set, serious performance regression will be caused. Therefore, it is currently necessary to push down only the types supported by the orc filter.

## How was this patch tested?
TPCDS  q33

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -829,7 +829,9 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
         remainingFunctions,
         singularOrLists,
         subfieldrOrLists,
-        remainingrOrLists);
+        remainingrOrLists,
+        veloxTypeList,
+        splitInfo->format);
 
     // Create subfield filters based on the constructed filter info map.
     auto subfieldFilters = createSubfieldFilters(colNameList, veloxTypeList, subfieldFunctions, subfieldrOrLists);
@@ -1321,7 +1323,9 @@ void SubstraitToVeloxPlanConverter::separateFilters(
     std::vector<::substrait::Expression_ScalarFunction>& remainingFunctions,
     const std::vector<::substrait::Expression_SingularOrList>& singularOrLists,
     std::vector<::substrait::Expression_SingularOrList>& subfieldOrLists,
-    std::vector<::substrait::Expression_SingularOrList>& remainingOrLists) {
+    std::vector<::substrait::Expression_SingularOrList>& remainingOrLists,
+    std::vector<TypePtr>& veloxTypeList,
+    dwio::common::FileFormat format) {
   for (const auto& singularOrList : singularOrLists) {
     if (!canPushdownSingularOrList(singularOrList)) {
       remainingOrLists.emplace_back(singularOrList);
@@ -1338,6 +1342,17 @@ void SubstraitToVeloxPlanConverter::separateFilters(
   for (const auto& scalarFunction : scalarFunctions) {
     auto filterNameSpec = SubstraitParser::findFunctionSpec(functionMap_, scalarFunction.function_reference());
     auto filterName = SubstraitParser::getSubFunctionName(filterNameSpec);
+    if (format == dwio::common::FileFormat::ORC && scalarFunction.arguments().size() > 0) {
+      auto value = scalarFunction.arguments().at(0).value();
+      if (value.has_selection()) {
+        uint32_t fieldIndex = SubstraitParser::parseReferenceSegment(value.selection().direct_reference());
+        auto type = value.selection().direct_reference().struct_field().field();
+        if (!veloxTypeList.empty() && veloxTypeList.at(fieldIndex)->isDecimal()) {
+          remainingFunctions.emplace_back(scalarFunction);
+          continue;
+        }
+      }
+    }
 
     // Check whether NOT and OR functions can be pushed down.
     // If yes, the scalar function will be added into the subfield functions.
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -1324,8 +1324,8 @@ void SubstraitToVeloxPlanConverter::separateFilters(
     const std::vector<::substrait::Expression_SingularOrList>& singularOrLists,
     std::vector<::substrait::Expression_SingularOrList>& subfieldOrLists,
     std::vector<::substrait::Expression_SingularOrList>& remainingOrLists,
-    std::vector<TypePtr>& veloxTypeList,
-    dwio::common::FileFormat format) {
+    const std::vector<TypePtr>& veloxTypeList,
+    const dwio::common::FileFormat& format) {
   for (const auto& singularOrList : singularOrLists) {
     if (!canPushdownSingularOrList(singularOrList)) {
       remainingOrLists.emplace_back(singularOrList);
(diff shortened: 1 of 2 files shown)


gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -540,4 +553,36 @@ object FilterHandler {
       case other =>
         throw new UnsupportedOperationException(s"${other.getClass.toString} is not supported.")
     }
+
+  def filterExpressions(filters: Seq[Expression]): Seq[Expression] = {
+    val resFilters = filters
+      .filter {
+        case Equality(ar: Attribute, l: Literal) => checkAttribute(ar)
```
PHILO-HE, line 560: Thanks for your fix! I think it may be better to do the check on native side. Right? cc @rui-mo.

rui-mo (commented):
Thanks, and I can understand the motivation of this PR. The goal is to offload scan to native if just some filters are supported, right?
But separating filters with some hard-coded rules looks fragile to me, and a better way could be to improve the fallback logic of scan. For example, if a pure scan can pass the validation, but a scan with filter pushdown can not, then a pure scan is offloaded to native, and its subsequent filter operator needs to be fallbacked. How do you think?

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
kerwin-zk: @PHILO-HE Could you please explain why check on native side is better? this part belongs to plan conversion, so theoretically it should be the job of java side.

kerwin-zk:
> Thanks, and I can understand the motivation of this PR. The goal is to offload scan to native if just some filters are supported, right? But separating filters with some hard-coded rules looks fragile to me, and a better way could be to improve the fallback logic of scan. For example, if a pure scan can pass the validation, but a scan with filter pushdown can not, then a pure scan is offloaded to native, and its subsequent filter operator needs to be fallbacked. How do you think?

@rui-mo  If the filter operator falls back, there will be additional c2r and r2c, which is time-consuming. At the same time, pure scan will lose its own available filters, resulting in reading more data. These two reasons will lead to performance degradation.

rui-mo:
> @rui-mo If the filter operator falls back, there will be additional c2r and r2c, which is time-consuming. At the same time, pure scan will lose its own available filters, resulting in reading more data. These two reasons will lead to performance degradation.

If a filter cannot be handled in Scan, then it cannot be handled in filter operator either. In Velox table scan, the remaing filter shares the same logic as filter node.

kerwin-zk:
> > @rui-mo If the filter operator falls back, there will be additional c2r and r2c, which is time-consuming. At the same time, pure scan will lose its own available filters, resulting in reading more data. These two reasons will lead to performance degradation.
> 
> If a filter cannot be handled in Scan, then it cannot be handled in filter operator either. In Velox table scan, the remaing filter shares the same logic as filter node. Please check [SubstraitToVeloxPlan.cc](https://github.com/oap-project/gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlan.cc#L848-L849).

@rui-mo The native filter I am talking about is divided into two parts, one part is the filter that supports push-down in ReadRel, and the other part does not support push-down. For example, the filters of decimal and timestamp are placed in FilterRel. The filter that does not support push down to scan will generate FilterRel, and decimal is handled as an int in velox, so the filter of decimal in FilterRel is supported, and the latest code of velox already supports the filter of timestamp in FilterRel.

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
PHILO-HE: We prefer the native validation for better code maintenance. Rui may have discussed with you. Thanks!

kerwin-zk:
@PHILO-HE @rui-mo cc

zhouyuan:
related: https://github.com/facebookincubator/velox/pull/6240

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -350,7 +350,9 @@ class SubstraitToVeloxPlanConverter {
       std::vector<::substrait::Expression_ScalarFunction>& remainingFunctions,
       const std::vector<::substrait::Expression_SingularOrList>& singularOrLists,
       std::vector<::substrait::Expression_SingularOrList>& subfieldrOrLists,
-      std::vector<::substrait::Expression_SingularOrList>& remainingrOrLists);
+      std::vector<::substrait::Expression_SingularOrList>& remainingrOrLists,
+      std::vector<TypePtr>& veloxTypeList,
```
rui-mo, line 354: const std::vector<TypePtr>&

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -350,7 +350,9 @@ class SubstraitToVeloxPlanConverter {
...
+      dwio::common::FileFormat format);
```
rui-mo, line 355: const dwio::common::FileFormat&

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1338,6 +1342,17 @@ void SubstraitToVeloxPlanConverter::separateFilters(
   for (const auto& scalarFunction : scalarFunctions) {
     auto filterNameSpec = SubstraitParser::findFunctionSpec(functionMap_, scalarFunction.function_reference());
     auto filterName = SubstraitParser::getSubFunctionName(filterNameSpec);
+    if (format == dwio::common::FileFormat::ORC && scalarFunction.arguments().size() > 0) {
```
rui-mo, line 1345: Please add some comment here, e.g.,
>// Add all decimal filters to remaining functions because their pushdown are not supported.

rui-mo (commented):
Thanks for the improvement!

cpp/velox/substrait/SubstraitToVeloxPlan.h
kerwin-zk: done.

cpp/velox/substrait/SubstraitToVeloxPlan.h
kerwin-zk: done.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
kerwin-zk: done.

kerwin-zk:
@rui-mo cc

rui-mo (approved):

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1338,6 +1342,18 @@ void SubstraitToVeloxPlanConverter::separateFilters(
   for (const auto& scalarFunction : scalarFunctions) {
     auto filterNameSpec = SubstraitParser::findFunctionSpec(functionMap_, scalarFunction.function_reference());
     auto filterName = SubstraitParser::getSubFunctionName(filterNameSpec);
+    // Add all decimal filters to remaining functions because their pushdown are not supported.
+    if (format == dwio::common::FileFormat::ORC && scalarFunction.arguments().size() > 0) {
+      auto value = scalarFunction.arguments().at(0).value();
+      if (value.has_selection()) {
+        uint32_t fieldIndex = SubstraitParser::parseReferenceSegment(value.selection().direct_reference());
+        auto type = value.selection().direct_reference().struct_field().field();
+        if (!veloxTypeList.empty() && veloxTypeList.at(fieldIndex)->isDecimal()) {
```
PHILO-HE, line 1351: Thanks so much for your fix!
According to the PR description, trying to push down filter in timestamp type can also fail. Shouldn't we also exclude it here?

kerwin-zk: @PHILO-HE  At present, when gluten encounters an orc read with timestamp type in the schema, it will fallback to java scan, so if there is timestamp in the filter, native scan will not be performed. ```Validation failed for class io.glutenproject.backendsapi.velox.BackendSettings$ due to: data type TimestampType. in file schema.```

PR #2878, opened 2023-08-24
[CORE] Change default value to 0.0D of spark.gluten.sql.columnar.backend.velox.overAcquiredMemoryRatio

Yohahaha:

This should be an advance config for user who knows what they want in memory management.

Let's change its default value to 0.0D as OverAcquire class described.


Yohahaha:
@zhztheplayer could you help take a look?

zhztheplayer:
The over-acquire feature used to workaround a Velox issue which has been fixed in https://github.com/facebookincubator/velox/pull/5780.

It might be safe to turn it off now, but more tests and benchmarks are needed.

Let's hold this off before we got some test results? Also did you verify the patch in your environment?

zhztheplayer (changes requested):
Let's hold this off.

PR #2880, opened 2023-08-24
[VL] Fix hash aggregate fail with mixed supported and unsupported aggregate functions

ulysses-you:

This pr fixes the bug that in native validation, when we check the aggregate signatures, we will return true once one of the aggregate functions matches. This pr changes to return ture only if all aggregate functions signatures match.

```sql
select c1, collect_set(c2), max(c2) from t group by c1;
```

```
23/08/24 13:27:45 ERROR TaskSetManager: Task 1 in stage 9.0 failed 1 times; aborting job
org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 9.0 failed 1 times, most recent failure: Lost task 1.0 in stage 9.0 (TID 23) (10.221.106.37 executor driver): java.lang.RuntimeException: Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Aggregate function 'array_distinct' not registered
Retriable: False
Function: intermediateType
File: /Users/cathy/Desktop/code/gluten/ep/build-velox/build/velox_ep/velox/exec/Aggregate.cpp
Line: 248
```

## How was this patch tested?

add test

### DIFF
--- a/gluten-ut/spark33/src/test/scala/org/apache/spark/sql/GlutenDataFrameAggregateSuite.scala
+++ b/gluten-ut/spark33/src/test/scala/org/apache/spark/sql/GlutenDataFrameAggregateSuite.scala
@@ -19,7 +19,10 @@ package org.apache.spark.sql
 import io.glutenproject.execution.HashAggregateExecBaseTransformer
 
 import org.apache.spark.sql.execution.aggregate.SortAggregateExec
+import org.apache.spark.sql.expressions.Aggregator
 import org.apache.spark.sql.functions._
+
+import java.lang.{Long => JLong}
 
 class GlutenDataFrameAggregateSuite extends DataFrameAggregateSuite with GlutenSQLTestsTrait {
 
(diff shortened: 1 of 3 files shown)


gluten-ut/spark33/src/test/scala/org/apache/spark/sql/GlutenDataFrameAggregateSuite.scala
```diff
@@ -200,4 +200,19 @@ class GlutenDataFrameAggregateSuite extends DataFrameAggregateSuite with GlutenS
           _.isInstanceOf[HashAggregateExecBaseTransformer]).isDefined)
     }
   }
+
+  test("mixed supported and unsupported aggregate functions") {
+    val df = spark.sql("SELECT a, collect_set(b), max(b) FROM testData2 group by a")
```
PHILO-HE, line 205: Thanks so much for your fix!

Currently, this test can well test the mixed supported/unsupported functions. But when collect_set is supported (actually WIP on my side), the test case will not meet our purpose. Right?
Maybe, we can add some c++ test by designing an agg function with input type that can never match with any velox function's signature. I can help do this.

PHILO-HE: Certainly, it is better to have an end-2-end test like you proposed. But I'm not sure whether we can design one spark test that always meet our test purpose.

ulysses-you: I agree, all built-in aggregte functions should be offload to native in future. I changed to use a UDAF to make sure it always running on jvm.

PHILO-HE (approved):
LGTM!

ulysses-you:
thank you @PHILO-HE for review , merging to main!

PR #2889, opened 2023-08-24
[VL][CELEBORN] Fallback to row-based celeborn shuffle writer

ulysses-you:

If the shuffle exchange is row-based and user enabled the celeborn, we should use vanilla celeborn shuffle writer rather than  fallback to spark shuffle writer. So we can make sure the dra can still work with celeborn.

## How was this patch tested?

manually test


ulysses-you:
cc @kerwin-zk thank you

gluten-celeborn/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
```diff
@@ -54,6 +54,7 @@ public class CelebornShuffleManager implements ShuffleManager {
   private LifecycleManager lifecycleManager;
   private ShuffleClient shuffleClient;
   private volatile ColumnarShuffleManager _columnarShuffleManager;
+  private volatile SparkShuffleManager _vanillaCelenornShuffleManager;
```
kerwin-zk, line 57: _vanillaCelenornShuffleManager -> _vanillaCelebornShuffleManager

gluten-celeborn/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
```diff
@@ -240,12 +260,17 @@ public <K, V> ShuffleWriter<K, V> getWriter(
...
-        if (ShuffleMode.HASH.equals(celebornConf.shuffleWriterMode())) {
+        if (!ShuffleMode.HASH.equals(celebornConf.shuffleWriterMode())) {
+          throw new UnsupportedOperationException(
+              "Unrecognized shuffle write mode!" + celebornConf.shuffleWriterMode());
+        }
+        if (h.dependency() instanceof ColumnarShuffleDependency) {
+          // columnar-based shuffle
           return new CelebornHashBasedColumnarShuffleWriter<>(
               h, context, celebornConf, client, metrics);
         } else {
```
kerwin-zk, line 271: ditto.

gluten-celeborn/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
```diff
@@ -160,33 +173,40 @@ private void initializeLifecycleManager() {
...
     if (dependency instanceof ColumnarShuffleDependency) {
-      appUniqueId = SparkUtils.appUniqueId(dependency.rdd().context());
-      initializeLifecycleManager();
-
       if (fallbackPolicyRunner.applyAllFallbackPolicy(
           lifecycleManager, dependency.partitioner().numPartitions())) {
         logger.warn("Fallback to ColumnarShuffleManager!");
         columnarShuffleIds.add(shuffleId);
         return columnarShuffleManager().registerShuffle(shuffleId, dependency);
       } else {
```
kerwin-zk, line 202: Adding concrete types here will make the logic clearer. ```else if (h.dependency() instanceof ShuffleDependency)```, and if none match, an exception is thrown in the else.

gluten-celeborn/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
ulysses-you: the dependency is awlays instanceof ShuffleDependency.. so it seems the clause is unnecessary

kerwin-zk (approved):

yaooqinn (approved):

Issue #2891, opened 2023-08-24
Velox doesn't work with Spark Delta Lake


felipepessoto:
### Backend

VL (Velox)

### Bug description

When I read a Delta Lake table it fallback:

GlutenFallbackReporter [Thread-32]: Validation failed for plan: Scan parquet , due to: does not support fileFormat: UnknownFormat.

Delta Lake tables are built on top of Parquet

### Spark version

Spark-3.3.x

PR #2892, opened 2023-08-24
[GLUTEN-2891][VL]fix: Add support to scan Delta Lake tables

felipepessoto:

Adds support to Delta Lake table

(Fixes: \#2891)

## How was this patch tested?

Manual test


Yohahaha:
Thanks for the fix, have you try this patch in your test? Velox does not support Delta format as I know, does Delta file can be read as a normal parquet file?

YannByron:
IMO, this is not enough. If enable column mapping, it will return NULL rows. And in Deletion Vector case, there will be some redundant rows.

felipepessoto:
@YannByron in that case would we need to re-implement Delta features in Velox?
I was expecting Velox to help with Native Scan and some other operators and that Delta would work on top of it applying filters for DV, or changing the column name to scan before sending to Velox. It seems it is not the case.

YannByron:
@felipepessoto 
Sorry for the late reply. I saw you have commented in https://github.com/oap-project/gluten/pull/2902. Let's discuss there.

Issue #2854, opened 2023-08-22
[VL] Shuffle writer should use tracked memory pool in native


Yohahaha:
### Description

Current use defaultLeafVeloxMemoryPool in VeloxShuffleWriter.cc#L570

PR #2898, opened 2023-08-25
[GLUTEN-2854] [VL] Refactor native defaultLeafVeloxMemoryPool usage for test and validation only

Yohahaha:

This patch limit defaultLeafPool method's usage, it only used in test and validation now.

After this patch, all memory pools used in real computation were tracked by Spark.

Fixes: #2854


Yohahaha:
@zhztheplayer could you help take a look? thanks!

cpp/velox/memory/VeloxMemoryManager.h
```diff
@@ -54,7 +54,16 @@ class VeloxMemoryManager final : public MemoryManager {
...
+  static auto memoryManager = std::make_shared<gluten::VeloxMemoryManager>(
+      "test",
+      gluten::defaultMemoryAllocator(),
+      std::shared_ptr<gluten::AllocationListener>(gluten::AllocationListener::noop()));
+  return memoryManager;
+}
+
+inline std::shared_ptr<facebook::velox::memory::MemoryPool> defaultLeafVeloxMemoryPool() {
+  return getDefaultMemoryManager()->getLeafMemoryPool();
+}
```
zhztheplayer, line 68: Why we inline the functions? They should not be called very frequently.

zhztheplayer (commented):
Only a minor comment. Thanks!

cpp/velox/memory/VeloxMemoryManager.h
Yohahaha: These codes were moved from BenchmarkUtils.h.

> Why we inline the functions?

Good question, and I can not give a reasonable answer, just follow other codes.

Yohahaha: Oh I see the origin codes was added by me... 

Inline or not are seems good to me.

zhztheplayer (approved):

PR #2907, opened 2023-08-25
[VL] Fix the incorrect result type when decimal arithmetic expressions are nested

marin-ma:
Fix the incorrect result type when decimal arithmetic expressions are nested. For example, in below case the input type got from divide is decimal(35, 20), which is not correct because cast was intentionally removed. Since the input types of divide are decimal(17, 2), the correct result type should be calculated from them.
> promote_precision(CheckOverflow((promote_precision(cast(promotions#850 as decimal(15,4))) / promote_precision(cast(total#851 as decimal(15,4)))), DecimalType(35,20))) * 100.00000000000000000000

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
@@ -414,15 +414,22 @@ object ExpressionConverter extends SQLConfHelper with Logging {
         val (left, right) = DecimalArithmeticUtil.rescaleCastForDecimal(
           DecimalArithmeticUtil.removeCastForDecimal(rescaleBinary.left),
           DecimalArithmeticUtil.removeCastForDecimal(rescaleBinary.right))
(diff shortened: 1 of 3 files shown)


rui-mo:
@marin-ma Could you rebase this PR? Thanks.

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
```diff
@@ -414,15 +414,22 @@ object ExpressionConverter extends SQLConfHelper with Logging {
         val (left, right) = DecimalArithmeticUtil.rescaleCastForDecimal(
           DecimalArithmeticUtil.removeCastForDecimal(rescaleBinary.left),
           DecimalArithmeticUtil.removeCastForDecimal(rescaleBinary.right))
+        val leftChild = replaceWithExpressionTransformer(left, attributeSeq)
```
winningsix, line 417: Nice catch. Do we need some UTs to cover this beyond TPC-H?

rui-mo (commented):
Verified locally. Thanks for the fix!

PR #2916, opened 2023-08-28
[GLUTEN-2974] [VL] Minor refactor for AllocationListener and VeloxMemoryMamanger to non copyable and non assignable

Yohahaha:
Fixes: #2974

### DIFF
--- a/cpp/core/jni/JniCommon.h
+++ b/cpp/core/jni/JniCommon.h
@@ -24,7 +24,7 @@
 
 #include "compute/ProtobufUtils.h"
 #include "config/GlutenConfig.h"
-#include "memory/ArrowMemoryPool.h"
+#include "memory/AllocationListener.h"
 #include "utils/compression.h"
 #include "utils/exception.h"
 
@@ -166,18 +166,20 @@ class SparkAllocationListener final : public gluten::AllocationListener {
  public:
   SparkAllocationListener(
       JavaVM* vm,
-      jobject javaListener,
-      jmethodID javaReserveMethod,
-      jmethodID javaUnreserveMethod,
+      jobject jListenerLocalRef,
+      jmethodID jReserveMethod,
+      jmethodID jUnreserveMethod,
       int64_t blockSize)
-      : vm_(vm),
-        javaReserveMethod_(javaReserveMethod),
-        javaUnreserveMethod_(javaUnreserveMethod),
-        blockSize_(blockSize) {
+      : vm_(vm), jReserveMethod_(jReserveMethod), jUnreserveMethod_(jUnreserveMethod), blockSize_(blockSize) {
     JNIEnv* env;
     attachCurrentThreadAsDaemonOrThrow(vm_, &env);
-    javaListener_ = env->NewGlobalRef(javaListener);
+    jListenerGlobalRef_ = env->NewGlobalRef(jListenerLocalRef);
   }
+
+  SparkAllocationListener(const SparkAllocationListener&) = delete;
+  SparkAllocationListener(SparkAllocationListener&&) = delete;
+  SparkAllocationListener& operator=(const SparkAllocationListener&) = delete;
+  SparkAllocationListener& operator=(SparkAllocationListener&&) = delete;
 
   ~SparkAllocationListener() override {
     JNIEnv* env;
@@ -186,7 +188,7 @@ class SparkAllocationListener final : public gluten::AllocationListener {
                 << "JNIEnv was not attached to current thread" << std::endl;
       return;
     }
-    env->DeleteGlobalRef(javaListener_);
+    env->DeleteGlobalRef(jListenerGlobalRef_);
   }
 
   void allocationChanged(int64_t size) override {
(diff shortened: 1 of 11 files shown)


cpp/core/memory/ReservationListener.h
```diff
@@ -0,0 +1,118 @@
...
+      JavaVM* vm,
+      jobject jListenerLocalRef,
+      jmethodID jReserveMethod,
+      jmethodID jUnreserveMethod,
+      int64_t blockSize)
+      : vm_(vm), jReserveMethod_(jReserveMethod), jUnreserveMethod_(jUnreserveMethod), blockSize_(blockSize) {
+    JNIEnv* env;
+    attachCurrentThreadAsDaemonOrThrow(vm_, &env);
+    jListenerGlobalRef_ = env->NewGlobalRef(jListenerLocalRef);
+  }
```
zhztheplayer, line 51: Why moving this out from `core/jni`?

`SparkReservationListener` relies on JNI it was in `core/jni`.

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -22,10 +22,10 @@
 namespace gluten {
 
 bool ListenableMemoryAllocator::allocate(int64_t size, void** out) {
-  listener_->allocationChanged(size);
+  listener_->reservationChanged(size);
   bool succeed = delegated_->allocate(size, out);
   if (!succeed) {
-    listener_->allocationChanged(-size);
+    listener_->reservationChanged(-size);
   }
```
zhztheplayer, line 29: Probably the terms should be clarified:

`allocation`: the actual memory used by the listener's caller
`reservation`: all reserved memory, including the over-borrowed memory from Spark's memory manager 

So here we'd better use `allocationChanged` rather than `reservationChanged`?

cpp/velox/memory/VeloxMemoryManager.h
```diff
@@ -29,7 +31,12 @@ class VeloxMemoryManager final : public MemoryManager {
   explicit VeloxMemoryManager(
       std::string name,
       std::shared_ptr<MemoryAllocator> allocator,
-      std::shared_ptr<AllocationListener> listener);
+      std::shared_ptr<ReservationListener> listener);
+
+  VeloxMemoryManager(const VeloxMemoryManager&) = delete;
+  VeloxMemoryManager(VeloxMemoryManager&&) = delete;
+  VeloxMemoryManager& operator=(const VeloxMemoryManager&) = delete;
+  VeloxMemoryManager& operator=(VeloxMemoryManager&&) = delete;
```
zhztheplayer, line 39: 👍

cpp/core/memory/ReservationListener.h
Yohahaha: ReservationListener concept was connected with memory management, so I think move to memory folder would make more sense, core/memory could depends on core/jni right? these are not independent.

cpp/core/memory/MemoryAllocator.cc
Yohahaha: In my view this API should align with its name, ReservationListener's caller use this API to reserve some virtual bytes,  if success, then caller can allocate real bytes.

Native memory manager report it's used bytes to Spark, which means reserve bytes from Spark, real allocate bytes should happens in memory manager internal.

cpp/core/memory/ReservationListener.h
zhztheplayer: > core/memory could depends on core/jni right?

Can you elaborate? I thought they were independent. Thanks.

Yohahaha: I got it, will fix later.

cpp/core/memory/MemoryAllocator.cc
zhztheplayer: > real allocate bytes should happens in memory manager internal.

That only applies for Velox backend. The base code is in `cpp/core`. Even for Velox backend it will probably not stand for long time since we should finally hook the allocations in Velox's pools rather than reservations.

Looking back to this code, the code here is for real "allocations", and the function has signature `::allocate(int64_t size, void** out)`.

Or if the codes are confusing, I'd suggest renaming `SparkReservationListener` to `SparkMemoryReporter` or something sounds more explanatory, it doesn't have to end with `Listener` based on my understanding.

cpp/core/memory/ReservationListener.h
Yohahaha: already revert.

cpp/core/memory/MemoryAllocator.cc
Yohahaha: already revert.

AllocationListener also make sense for me, I have no idea for more meaningful name. :)

zhztheplayer:
@Yohahaha Would you like to do rebase? Thanks!

cpp/core/jni/JniCommon.h
```diff
@@ -166,27 +166,29 @@ class SparkAllocationListener final : public gluten::AllocationListener {
  public:
   SparkAllocationListener(
       JavaVM* vm,
-      jobject javaListener,
-      jmethodID javaReserveMethod,
-      jmethodID javaUnreserveMethod,
+      jobject jListenerLocalRef,
```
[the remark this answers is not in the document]
Yohahaha: We should limit usage, create listener from java call native, and pass local ref in, only convert to global ref in listener internal, let listener manage this ref, otherwise, we can not manage this ref easily.

I'm not sure whether newGlobalRef is idempotent...

zhztheplayer:
Would you like to rebase ? Thanks! @Yohahaha

Yohahaha:
@zhztheplayer already rebased, but here has a network issue in CI, could you help re-run failed jobs?

zhztheplayer:
@rui-mo Would you like to take a look?

zhztheplayer (approved):

PR #2918, opened 2023-08-28
[VL] Add UT for nested decimal arithmetic

marin-ma:
A follow-up PR of #2907


marin-ma:
cc @winningsix @rui-mo

winningsix:
> cc @winningsix @rui-mo

Thanks @marin-ma Looks good to me.

marin-ma:
@rui-mo Just found the validation for "cast from date" is not aligned with Velox's judgement https://github.com/facebookincubator/velox/blob/main/velox/expression/CastExpr.cpp#L35-L92 
Velox only support cast from date to timestamp/varchar. I don't know the background of why cast from date to timestamp is disabled in Gluten's validation. Simply added fallback to all types other than varchar.

rui-mo:
> I don't know the background of why cast from date to timestamp is disabled in Gluten's validation.

Casts from/to timestamp were disabled due to failures and incorrect results. We need to go through each cast to dig out the reason, but before that, we just fallback them.

rui-mo (approved):

PR #2928, opened 2023-08-28
[VL] Fix UT: CheckOverflow

marin-ma:
On Gluten's main branch, `CheckOverflowExpression` is transformed into `CheckOverflowExpressionTransformer`, and is used to change the precision of a decimal value. But after rebase, the `CheckOverflowExpressionTransformer` is removed and the native computation will directly use the original scale/precision to avoid precision loss. Therefore after rebase, the UT failed because the `CheckOverflow` is nolonger used to change the precision of a decimal, and it should use `Cast` instead.

Reference:

Gluten main: https://github.com/oap-project/gluten/blob/0891942447560b26ad4374ad533b05b53bbb65c2/gluten-core/src/main/scala/io/glutenproject/expression/UnaryExpressionTransformer.scala#L159-L183

Gluten rebase: https://github.com/oap-project/gluten/blob/647bdb611eb1d81ac1269658d42f201911972723/gluten-core/src/main/scala/io/glutenproject/expression/UnaryExpressionTransformer.scala#L159-L168


marin-ma:
cc: @rui-mo @zhejiangxiaomai

rui-mo (commented):
Thanks for the fix. Why there is no this issue on the main branch?

zhejiangxiaomai:
> Thanks for the fix. Why there is no this issue on the main branch?

Maybe we can illustrate it on PR description.

rui-mo (approved):

Issue #2934, opened 2023-08-28
[CH] Scala.MatchError: knownfloatingpointnormalized


loneylee:
### Backend

CH (ClickHouse)

### Bug description

Run sql return error
```
select coalesce(t1.`i1`, 0) + coalesce(t2.`l1`, 0) `c1`,
       coalesce(t1.`d1`, t2.`d2`)                  sf
from (select double_field   d1,
             sum(int_field) i1
      from tt
      group by double_field) t1
         full join (select double_field    d2,
                           avg(long_field) l1
                    from tt
                    group by double_field) t2
                   on t1.d1 = t2.d2
```

error:
```text
scala.MatchError: knownfloatingpointnormalized(normalizenanandzero(d2#60)) (of class org.apache.spark.sql.catalyst.expressions.KnownFloatingPointNormalized)
	at org.apache.spark.sql.execution.utils.CHExecUtil$.$anonfun$buildHashPartitioning$1(CHExecUtil.scala:180)
	at org.apache.spark.sql.execution.utils.CHExecUtil$.$anonfun$buildHashPartitioning$1$adapted(CHExecUtil.scala:180)
```

### Spark version

None

PR #2935, opened 2023-08-28
[GLUTEN-2934][CH] Fix error knownfloatingpointnormalized

loneylee:

(Fixes: \#2934)

## How was this patch tested?

test by ut

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/utils/CHExecUtil.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/utils/CHExecUtil.scala
@@ -178,13 +178,12 @@ object CHExecUtil extends Logging {
       childOutput: Seq[Attribute],
       output: Seq[Attribute]): NativePartitioning = {
     val hashFields =
-      partitoining.expressions.map(ConverterUtils.getAttrFromExpr(_).toAttribute).map {
-        case a: Attribute =>
+      partitoining.expressions.map(
+        a =>
           BindReferences
-            .bindReference(a, childOutput)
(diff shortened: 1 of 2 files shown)


zzcclp:
@PHILO-HE please help to check whether this fix is suitable for Velox backend.

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/utils/CHExecUtil.scala
```diff
@@ -177,13 +177,14 @@ object CHExecUtil extends Logging {
...
-    val hashFields = partitoining.expressions.map {
-      case a: Attribute =>
-        BindReferences
-          .bindReference(a, childOutput)
-          .asInstanceOf[BoundReference]
-          .ordinal
-    }
+    val hashFields =
+      partitoining.expressions.map(ConverterUtils.getAttrFromExpr(_).toAttribute).map {
+        case a: Attribute =>
```
PHILO-HE, line 182: Nit:
It looks unnecessary to keep `case a: Attribute =>`, as the first map's output is always `Attribute` type.

PHILO-HE (approved):
Just tested velox backend in main branch. Looks no issue found at least for this test case.
@zzcclp, thanks for letting me know this.

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/utils/CHExecUtil.scala
loneylee: fixed

PHILO-HE (approved):
Thanks!

PR #2949, opened 2023-08-29
[VL] Avoid throwing exception from destructor

jackylee-ch:
`Terminated by ClosedByInterruptException` would be found while using speculation in Spark. The reason is that the `GlutenException` is not caught when the C++ object is destructed, which would cause the executor exit abnormally.

## How was this patch tested?
unit tests


jackylee-ch:
It seems we meet the same problem with https://github.com/oap-project/gluten/pull/1505 in new code. @zhztheplayer @zhouyuan

zhouyuan (approved):
👍

jackylee-ch:
Thanks for your review.

cpp/velox/jni/JniFileSystem.cc
```diff
@@ -75,13 +75,13 @@ class JniReadFile : public facebook::velox::ReadFile {
   ~JniReadFile() override {
     try {
       close0();
+      JNIEnv* env;
+      attachCurrentThreadAsDaemonOrThrow(vm, &env);
+      env->DeleteGlobalRef(obj_);
+      checkException(env);
     } catch (const std::exception& e) {
       LOG(WARNING) << "Error closing jni read file " << e.what();
     }
```
zhztheplayer, line 84: Thanks, and just out of my curiosity, do you actually use spill strategy `heap-over-local` ? @jackylee-ch

zhouyuan: it's possibly due to speculation is enabled in their test, thus executors will be killed "randomly" (?)

thanks, -yuan

jackylee-ch: > it's possibly due to speculation is enabled in their test, thus executors will be killed "randomly" (?)

@zhztheplayer @zhouyuan  The current problem is only related to speculation and I didn't use the spill strategy `heap-over-local`. 
As speculation is enabled, the task would be killed if another attempt success and would throw `ClosedByInterruptException` in JVM. The `checkException` would catch the exception and throw `GlutenException`, which was not catched by `try catch` and finally make executor failed.

zhztheplayer: OK but this file `JniFileSystem.cc` should only be used when `heap-over-local` is enabled. Am I missing something here?

jackylee-ch: https://github.com/oap-project/gluten/blob/main/cpp/velox/jni/VeloxJniWrapper.cc#L59C39-L59C39
Currently we would call `gluten::initVeloxJniFileSystem` in `JNI_OnLoad`.

jackylee-ch: > OK but this file `JniFileSystem.cc` should only be used when `heap-over-local` is enabled. Am I missing something here?

@zhztheplayer Maybe we should add a check for the `initVeloxJniFileSystem`

PR #2960, opened 2023-08-30
[GLUTEN-2959] Fix an issue in getting mapped substrait name for expression

PHILO-HE:


PHILO-HE:
@exmy, could you please take a review? I am assuming we still need to throw an exception to fall back if the mapped substrait name is not found. Right？

PHILO-HE:
> @exmy, could you please take a review? I am assuming we still need to throw an exception to fall back if the mapped substrait name is not found. Right？

Thought wrongly. Please ignore this PR. Thanks!

PR #2965, opened 2023-08-30
[VL] Avoid hold unmanaged velox memory pool for caller

Yohahaha:

Velox's addXXXPool only hold weak ptr in parent pool, returned shared ptr should be managed by caller.

Before this patch, WholeStageIter hold the shared pool and will release pool when we release iterator, but in some cases, there still some unused/unreleased batches before release iterator, so we may get bad memory usage error from Velox.

Since we implement MemoryManager to hold the shared ptr of memory pool, use getXXXPool of VeloxMemoryManager directly is more safe.

(Fixes: \#ISSUE-ID)


Yohahaha:
@zhztheplayer could you help take a look? although this patch can not fix speculation issue, but it still worth to fix.

zhztheplayer (approved):
Thanks!

Issue #2972, opened 2023-08-30
[CH] Read bool field from excel csv cause core dump


loneylee:
### Backend

CH (ClickHouse)

### Bug description

Read excel format csv
```text
true ,
```
cause coredump.

Expect return true.

### Spark version

None

PR #2973, opened 2023-08-30
[GLUTEN-2972][CH] Fix read field from excel format cause coredump

loneylee:
Fix read field from excel format cause coredump

(Fixes: \#2972)

## How was this patch tested?
Test by ut


baibaichen (approved):
LGTM

PR #2976, opened 2023-08-30
[VL] Remove combined buffer in shuffle split and use default allocator

marin-ma:


FelixYBW:
So the split buffer uses ShuffleMemoryPool, spill buffer uses arrow::default_pool, right?

marin-ma:
> So the split buffer uses ShuffleMemoryPool, spill buffer uses arrow::default_pool, right?

Yes. Only split buffer uses ShuffleMemoryPool.

zhztheplayer (approved):

zhztheplayer (approved):

zuochunwei:
Why we made this change?

FelixYBW:
This is part of the OOM solution. We need to shrink or reallocate split buffer on OOM. In this way large combine buffer can't be used any more.

PR #2982, opened 2023-08-31
[VL] Realloc shuffle split buffer if beyond avg threshold

marin-ma:
To boost performance, we calculate the average binary column's size per row and use the size + fixed column width to estimate the row count each reducer can allocate. Without the PR, we only calculate the binary column's size using the first batch. The problem is the binary size may varies a lot in batches, the first batch's size may be too small or too large. If it's too small, we need to reallocate memory during split which causes OOM.

The PR calculates the average batch size of each input record batch and compare with current allocated buffer size. If it's out of a threshold, we will firstly try to shrink or re-allocate the buffer based on new size, if the count is already larger than new size, then we spill the current record batch and reallocate a new buffer using the average batch size.


cpp/core/shuffle/ShuffleWriter.h
```diff
@@ -65,6 +67,60 @@ struct ShuffleWriterOptions {
   static ShuffleWriterOptions defaults();
 };
 
+class ShuffleMemoryPool : public arrow::MemoryPool {
+ public:
+  ShuffleMemoryPool(std::shared_ptr<arrow::MemoryPool> pool) : pool_(pool) {}
+
+  arrow::MemoryPool* delegated() {
```
jinchengchenghh, line 74: Why move these functions from ShuffleWriter.cpp to here?

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -651,21 +645,40 @@ arrow::Status VeloxShuffleWriter::updateInputHasNull(const velox::RowVector& rv)
...
         // allocate buffer if it's not yet allocated
-        auto newSize = std::max(calculatePartitionBufferSize(rv), partition2RowCount_[pid]);
         RETURN_NOT_OK(allocatePartitionBuffersWithRetry(pid, newSize));
+        // Update
+        partitionAvgBufferSize_[pid] = newSize;
+        partitionNumBuffers_[pid]++;
+      } else if (needRealloc(pid, newSize)) {
+        {
+          bool reuseBuffers = false;
+          ARROW_ASSIGN_OR_RAISE(auto rb, createArrowRecordBatchFromBuffer(pid, /*resetBuffers = */ !reuseBuffers));
```
jinchengchenghh, line 667: /*resetBuffers = */ -> /*reuseBuffers = */

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1134,24 +1152,29 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
...
+      auto stringViewColumn = column->asFlatVector<velox::StringView>();
+      assert(stringViewColumn);
+
+      // accumulate length
+      uint64_t length = stringViewColumn->values()->size();
+      for (auto& buffer : stringViewColumn->stringBuffers()) {
+        length += buffer->size();
       }
+
+      binaryArrayEmpiricalSize_[index] = (length + numRows - 1) / numRows;
```
jinchengchenghh, line 1177: Why is not `binaryArrayEmpiricalSize_[index] = length % numRows == 0 ? length / numRows : length / numRows + 1;`

docs/Configuration.md
```diff
@@ -9,44 +9,45 @@ nav_order: 3
...
+| spark.gluten.sql.columnar.shuffle.codecBackend          | Enable using hardware accelerators for shuffle de/compression. Valid options are QAT and IAA.                                                                                                                                                                                                                                                                     |                                                      |
+| spark.gluten.sql.columnar.shuffle.compressionMode       | Setting different compression mode in shuffle, Valid options are buffer and rowvector, buffer option compress each buffer of RowVector individually into one pre-allocated large buffer, rowvector option first copies each buffer of RowVector to a large buffer and then compress the entire buffer in one go.                                                  | buffer                                               |
```
jinchengchenghh, line 40: Looks like it introduce too many unrelated change, please only left yours.

cpp/velox/shuffle/VeloxShuffleWriter.cc
zuochunwei: it's just another implmentation

cpp/core/shuffle/ShuffleWriter.h
zuochunwei: these functions were removed from MemoryPoolWrapper, and the class `MemoryPoolWrapper` is unnecessary.

I agree this change because less is more.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -716,7 +733,7 @@ arrow::Status VeloxShuffleWriter::doSplit(const velox::RowVector& rv) {
   printPartitionBuffer();
 
   return arrow::Status::OK();
-}
+} // namespace gluten
```
jinchengchenghh, line 736: Unrelated change?

marin-ma: The namespace close comment is missing.

jinchengchenghh (dismissed):

cpp/velox/shuffle/VeloxShuffleWriter.cc
FelixYBW: we should count binary length using all history batches, instead of per rowvector. The value of each batch varies and still have the same issue

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -651,21 +645,40 @@ arrow::Status VeloxShuffleWriter::updateInputHasNull(const velox::RowVector& rv)
...
+          }
+        } // rb destructed
+        if (options_.prefer_evict) {
+          // if prefer_evict is set, evict current RowVector
+          RETURN_NOT_OK(evictPartition(pid));
+        }
+        RETURN_NOT_OK(allocatePartitionBuffersWithRetry(pid, newSize));
+        // Update
+        partitionAvgBufferSize_[pid] =
+            ((partitionAvgBufferSize_[pid] * partitionNumBuffers_[pid]) + newSize) / (partitionNumBuffers_[pid] + 1);
```
marin-ma, line 679: @FelixYBW Average buffer size is updated here.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1134,50 +1152,55 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
...
 
     for (size_t col = 0; col < simpleColumnIndices_.size(); ++col) {
       auto colIdx = simpleColumnIndices_[col];
       // `bool(1) >> 3` gets 0, so +7
-      sizePerRow += ((arrow::bit_width(arrowColumnTypes_[colIdx]->id()) + 7) >> 3);
+      bytesPerRow += ((arrow::bit_width(arrowColumnTypes_[colIdx]->id()) + 7) >> 3);
     }
 
-    VS_PRINTLF(sizePerRow);
+    VS_PRINTLF(bytesPerRow);
```
FelixYBW, line 1189: Let's use bytesPerRow = estimateFlatSize()/rv.size(); to calculate directly. It's a little over estimate the real size but it's fine. The function doesn't itereate the string_view array in binary column, instead it use buffer_array which several or all strings may share the same buffer.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1134,50 +1152,55 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
...
-        binaryArrayEmpiricalSize_[index] = length % numRows == 0 ? length / numRows : length / numRows + 1;
+    // Calculate average size bytes (bytes per row) for each binary array.
+    std::vector<uint64_t> binaryArrayAvgBytesPerRow(binaryColumnIndices_.size());
+    for (size_t i = 0; i < binaryColumnIndices_.size(); ++i) {
+      auto column = rv.childAt(binaryColumnIndices_[i]);
+      auto stringViewColumn = column->asFlatVector<velox::StringView>();
+      assert(stringViewColumn);
+
+      uint64_t binarySizeBytes = stringViewColumn->values()->size();
+      for (auto& buffer : stringViewColumn->stringBuffers()) {
```
marin-ma, line 1172: @FelixYBW We also iterates over stringBuffers, not each StringView.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -343,6 +354,13 @@ std::shared_ptr<arrow::RecordBatch> makeUncompressedRecordBatch(
   }
   return arrow::RecordBatch::Make(writeSchema, 1, {arrays});
 }
+
+inline bool shrinkBeforeSpill(SplitState state) {
+  return state == SplitState::kStop;
+}
```
zuochunwei, line 360: add a space line between two functions

cpp/velox/shuffle/VeloxShuffleWriter.cc
zuochunwei: It's each StringView here

FelixYBW:
@marin-ma Did you add the spill of split buffer in this PR?

marin-ma:
> @marin-ma Did you add the spill of split buffer in this PR?

No. Can we do it in next PR?

FelixYBW (approved):
Thank you for your work!

PR #2725, opened 2023-08-11
[GLUTEN-2723][CH] Improve Broadcast Join perfermance -- Part 1

baibaichen:
## What changes were proposed in this pull request?

Currently, when Broadcast size increased linearly, the  preformace degradates in a quadratic manner. 
image

One of reason is  we don't compress data when getting results from workers. This PR try **compression**.

(Fixes: \#2723)

## How was this patch tested?

Using existed UT

Issue #2723, opened 2023-08-11
Improve CH Backend Broadcast Join performance


baibaichen:
### Description

Per our test, when `broadcast size >= 60MB`,  the performance degradate.

PR #2986, opened 2023-08-31
[GLUTEN-2723][CH] Improve Broadcast Join performance -- Part 2

baibaichen:

This is follow up of  #2725.  What I  found is that calling java method in c++ thread is quite slow, Let's move such call in jni thread.

(Fixes: \#2723)

## How was this patch tested?
Using Existed UT


baibaichen:
While micro benchmark shows there is significant improvement, the end-to-end power test only reveals a slight improvement.

zzcclp (approved):

Issue #2987, opened 2023-08-31
[VL] Error: Native split: shuffle writer stop failed - Merging from spilled file NO.7 is not exhausted


zhztheplayer:
### Backend

VL (Velox)

### Bug description

reproduce using gluten-it

```
<gluten-it-bin> queries --local --preset=velox --benchmark-type=ds --error-on-memleak --disable-aqe --off-heap-size=5g -s=100.0 --shuffle-partitions=72 --threads=8 --iterations=1 --queries=q67 --enable-history --enable-ui
```

### Spark version

None

### Spark configurations

3.2.2

### Relevant logs

```bash
I0831 15:19:27.594770 497885 VeloxBackend.cc:43] session conf = {
 {spark.gluten.ugi.username = root}
 {spark.sql.caseSensitive = false}
 {spark.gluten.sql.session.timeZone.default = Asia/Shanghai}
 {spark.gluten.memory.task.offHeap.size.in.bytes = 671088640}
 {spark.gluten.ugi.tokens = }
}
23/08/31 15:19:27 ERROR TaskResources: Task 100 failed by error: 
java.lang.RuntimeException: Native split: shuffle writer stop failed - Merging from spilled file NO.7 is not exhausted.
	at io.glutenproject.vectorized.ShuffleWriterJniWrapper.stop(Native Method)
	at org.apache.spark.shuffle.ColumnarShuffleWriter.internalWrite(ColumnarShuffleWriter.scala:177)
	at org.apache.spark.shuffle.ColumnarShuffleWriter.write(ColumnarShuffleWriter.scala:218)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:52)
	at org.apache.spark.scheduler.Task.run(Task.scala:131)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:506)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1491)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:509)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #2990, opened 2023-08-31
[VL] Fix generated spilled file already created

marin-ma:
The current judgement of whether a new filename of local shuffle spill has already been created is wrong. Should use "O_EXCL" in the read flag to judge if the file has been created.

### DIFF
--- a/cpp/core/shuffle/utils.h
+++ b/cpp/core/shuffle/utils.h
@@ -25,8 +25,6 @@
 #include <arrow/type.h>
 #include <arrow/util/io_util.h>
 
-#include <bits/fcntl.h>
-#include <boost/thread.hpp>
 #include <boost/uuid/uuid_generators.hpp>
 #include <boost/uuid/uuid_io.hpp>
 #include <fcntl.h>
@@ -116,6 +114,8 @@ static inline arrow::Result<std::string> createTempShuffleFile(const std::string
         } else {
           return arrow::Status::IOError("Failed to open local file " + filePath + ", Reason: " + strerror(errno));
         }
+      } else {
+        close(fd);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


cpp/core/shuffle/utils.h
```diff
@@ -106,8 +107,16 @@ static inline arrow::Result<std::string> createTempShuffleFile(const std::string
     ARROW_ASSIGN_OR_RAISE(auto file_info, fs->GetFileInfo(filePath));
     if (file_info.type() == arrow::fs::FileType::NotFound) {
       exist = false;
-      ARROW_ASSIGN_OR_RAISE(auto s, fs->OpenOutputStream(filePath));
-      RETURN_NOT_OK(s->Close());
+      int fd = open(filePath.c_str(), O_CREAT | O_EXCL | O_RDWR, 0666);
+      if (fd < 0) {
+        if (errno == EEXIST) {
+          exist = true;
```
winningsix, line 113: Hmm, in which case it will return ```EEXIST``` given we have ```file_info.type() == arrow::fs::FileType::NotFound``` was true.

https://arrow.apache.org/docs/cpp/api/filesystem.html#_CPPv4N5arrow2fs8FileTypeE

marin-ma: `file_info.type() == arrow::fs::FileType::NotFound` only make sure the file doesn't exist, but it can't make sure the file is being created at the same time by other threads.

zhztheplayer:
Thanks Rong! Is this a fix for #2987 or not?

marin-ma:
> Thanks Rong! Is this a fix for #2987 or not?

Yes, I suspect this is the root cause. But I still can't reproduce using `sbin/gluten-it.sh queries --local --preset=velox --benchmark-type=ds --error-on-memleak --disable-aqe --off-heap-size=5g -s=100.0 --shuffle-partitions=72 --threads=8 --iterations=1 --queries=q67 --enable-history --enable-ui`

zhztheplayer (approved):
Thanks!

marin-ma:
@zhztheplayer This cannot fix #2987, but can be a potential risk.
I will merge this one first.

PR #2991, opened 2023-08-31
[VL] Fix decompress metrics

marin-ma:


zhztheplayer (approved):

zhztheplayer:
Thanks for catching!

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
```diff
@@ -177,9 +180,6 @@ private class ColumnarBatchSerializerInstance(
 
       override def close(): Unit = {
         if (!isClosed) {
-          // Collect Metrics
-          ShuffleReaderJniWrapper.INSTANCE.populateMetrics(shuffleReaderHandle, readerMetrics)
-          decompressTime += readerMetrics.getDecompressTime
```
ulysses-you, line 182: I reviewed the code but it seems we already have a hook for `this.close()`, so why we move these code to another hook ?  am I misunderstanding the fix ?

marin-ma: This one is called by InputStream.close. The other one is called at task end. This fix is due to 
native ShuffleReader is optimized from create per InputStream -> create per task. So, the metrics is also accumulated for all InputStream on native side. We need to count it once at task end instead of per InputStream.close(), otherwise the metrics would be wrong.

ulysses-you: got it, thank you for the deep explanation!

PR #2992, opened 2023-08-31
[CORE] Use PriorityQueue to hold task resource

ulysses-you:

This pr uses `PriorityQueue` to replace `java.util.HashMap[Long, java.util.List[TaskResource]]()` to simplify the code. And changes priority from long to int, as it seems the int type is enough.

## How was this patch tested?

PASS CI

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
@@ -25,10 +25,8 @@ import _root_.io.glutenproject.memory.SimpleMemoryUsageRecorder
 import _root_.io.glutenproject.utils.TaskListener
 
 import java.util
-import java.util.{Collections, UUID}
+import java.util.{Comparator, PriorityQueue, UUID}
 import java.util.concurrent.atomic.AtomicLong
-
-import scala.collection.JavaConverters._
 
 object TaskResources extends TaskListener with Logging {
   // And open java assert mode to get memory stack
@@ -69,11 +67,11 @@ object TaskResources extends TaskListener with Logging {
     }
   }
 
-  def addRecycler(name: String, prio: Long)(f: => Unit): Unit = {
+  def addRecycler(name: String, prio: Int)(f: => Unit): Unit = {
     addAnonymousResource(new TaskResource {
       override def release(): Unit = f
 
-      override def priority(): Long = prio
+      override def priority(): Int = prio
 
       override def resourceName(): String = name
     })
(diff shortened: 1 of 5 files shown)


gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
```diff
@@ -164,44 +162,39 @@ object TaskResources extends TaskListener with Logging {
 // thread safe
 class TaskResourceRegistry extends Logging {
   private val sharedUsage = new SimpleMemoryUsageRecorder()
-  private val resources = new java.util.LinkedHashMap[String, TaskResource]()
```
ulysses-you, line 167: no requirements to use LinkedHashMap, so changed to HashMap

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
```diff
@@ -164,44 +162,39 @@ object TaskResources extends TaskListener with Logging {
...
-          m =>
-            try { // LIFO
-              // logWarning(s"Prepare release ${m.resourceName()}")
-              m.release()
-            } catch {
-              case e: Throwable =>
-                logWarning("Failed to call release() on resource instance", e)
-            })
     }
+    resources.clear()
```
ulysses-you, line 197: before, we did not clear task resources after call release all, but seems not a big issue since we would remove the whole TaskResourceRegistry after task completion

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
zhztheplayer: 👍 

It was from a legacy code version using linkedhashmap to implement fifo, if I remember correctly.

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
```diff
@@ -164,44 +162,39 @@ object TaskResources extends TaskListener with Logging {
...
+    })
 
   private def addResource0(id: String, resource: TaskResource): Unit = synchronized {
     resources.put(id, resource)
-    if (!resourcesPriorityMapping.containsKey(resource.priority())) {
-      resourcesPriorityMapping.put(resource.priority(), new util.ArrayList[TaskResource]())
-    }
-    val list = resourcesPriorityMapping.get(resource.priority())
-    list.add(resource)
+    resourcesPriorityQueue.add((resource, resourcesPriorityQueue.size()))
```
zhztheplayer, line 183: This is buggy. The queue size would not be predictable for the same priority since it is the size for all priorities. If no other way you can find, probably dual nesting is not a bad choice.

zhztheplayer (changes requested):

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
ulysses-you: it always monotonically increasing as we never remove the element, so I think should be fine ?

zhztheplayer: I understand your point however the assumption relies on the fact that one should program carefully within this principle: never call `poll()` unless during task ends. 

It could be an option to have a accumulator to generated increasing integer IDs? That may align with your original purpose better I think.

ulysses-you: I'm fine to add one more accumulator, addressed

zhztheplayer (approved):
Thanks!

ulysses-you:
@zhztheplayer  thank you for review, merging to main branch

PR #3013, opened 2023-09-05
[CORE] Switch default Maven dependency scope of Spark to provided (by default) in gluten-it

winningsix:

FIXES https://github.com/oap-project/gluten/security/dependabot/37

## How was this patch tested?

Verify via local maven build.


winningsix:
cc @zhztheplayer

zhztheplayer (approved):
Thanks!

PR #3014, opened 2023-09-05
[GLUTEN-2808][VL] All managed Velox memory pools should be shrinkable

zhztheplayer:
So far we only adopt Velox reclaim/shrink APIs to plan execution pools.

We should extend the usage to all Velox pools using a common path.

https://github.com/oap-project/gluten/issues/2808

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -125,12 +125,12 @@ int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
   LOG(INFO) << logPrefix << "Pool has reserved " << pool_->currentBytes() << "/" << pool_->root()->reservedBytes()
             << "/" << pool_->root()->capacity() << "/" << pool_->root()->maxCapacity() << " bytes.";
   LOG(INFO) << logPrefix << "Shrinking...";
-  int64_t shrunk = pool_->shrinkManaged(pool_.get(), size);
-  LOG(INFO) << logPrefix << shrunk << " bytes released from shrinking.";
+  int64_t shrunken = pool_->shrinkManaged(pool_.get(), size);
+  LOG(INFO) << logPrefix << shrunken << " bytes released from shrinking.";
 
   // todo return the actual spilled size?
   if (spillStrategy_ == "auto") {
-    int64_t remaining = size - shrunk;
+    int64_t remaining = size - shrunken;
     LOG(INFO) << logPrefix << "Trying to request spilling for remaining " << remaining << " bytes...";
     // if we are on one of the driver of the spilled task, suspend it
     velox::exec::Driver* thisDriver = nullptr;
@@ -143,21 +143,21 @@ int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
       // not the driver, no need to suspend
       uint64_t spilledOut = pool_->reclaim(remaining);
       LOG(INFO) << logPrefix << "Successfully spilled out " << spilledOut << " bytes.";
-      uint64_t total = shrunk + spilledOut;
+      uint64_t total = shrunken + spilledOut;
       LOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
-      return shrunk + spilledOut;
+      return shrunken + spilledOut;
     }
     // suspend since we are on driver
     velox::exec::SuspendedSection noCancel(thisDriver);
     uint64_t spilledOut = pool_->reclaim(remaining);
     LOG(INFO) << logPrefix << "Successfully spilled out " << spilledOut << " bytes.";
-    uint64_t total = shrunk + spilledOut;
+    uint64_t total = shrunken + spilledOut;
     LOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
-    return shrunk + spilledOut;
+    return shrunken + spilledOut;
   }
 
-  LOG(INFO) << logPrefix << "Successfully reclaimed total " << shrunk << " bytes.";
-  return shrunk;
+  LOG(INFO) << logPrefix << "Successfully reclaimed total " << shrunken << " bytes.";
+  return shrunken;
 }
 
 void WholeStageResultIterator::getOrderedNodeIds(
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -118,20 +118,46 @@ std::shared_ptr<ColumnarBatch> WholeStageResultIterator::next() {
   return std::make_shared<VeloxColumnarBatch>(vector);
 }
 
+namespace {
+class ConditionalSuspendedSection {
+ public:
+  ConditionalSuspendedSection(velox::exec::Driver* driver, bool condition) {
+    if (condition) {
+      section_ = new velox::exec::SuspendedSection(driver);
+    }
+  }
+
+  virtual ~ConditionalSuspendedSection() {
+    if (section_) {
+      delete section_;
+    }
+  }
+
+  // singleton
+  ConditionalSuspendedSection(const ConditionalSuspendedSection&) = delete;
+  ConditionalSuspendedSection(ConditionalSuspendedSection&&) = delete;
+  ConditionalSuspendedSection& operator=(const ConditionalSuspendedSection&) = delete;
+  ConditionalSuspendedSection& operator=(ConditionalSuspendedSection&&) = delete;
+
+ private:
+  velox::exec::SuspendedSection* section_ = nullptr;
+};
+} // namespace
+
 int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
   std::string poolName{pool_->root()->name() + "/" + pool_->name()};
   std::string logPrefix{"Spill[" + poolName + "]: "};
-  LOG(INFO) << logPrefix << "Trying to reclaim " << size << " bytes of data...";
-  LOG(INFO) << logPrefix << "Pool has reserved " << pool_->currentBytes() << "/" << pool_->root()->reservedBytes()
-            << "/" << pool_->root()->capacity() << "/" << pool_->root()->maxCapacity() << " bytes.";
-  LOG(INFO) << logPrefix << "Shrinking...";
+  DLOG(INFO) << logPrefix << "Trying to reclaim " << size << " bytes of data...";
+  DLOG(INFO) << logPrefix << "Pool has reserved " << pool_->currentBytes() << "/" << pool_->root()->reservedBytes()
+             << "/" << pool_->root()->capacity() << "/" << pool_->root()->maxCapacity() << " bytes.";
+  DLOG(INFO) << logPrefix << "Shrinking...";
   int64_t shrunken = pool_->shrinkManaged(pool_.get(), size);
-  LOG(INFO) << logPrefix << shrunken << " bytes released from shrinking.";
+  DLOG(INFO) << logPrefix << shrunken << " bytes released from shrinking.";
 
   // todo return the actual spilled size?
   if (spillStrategy_ == "auto") {
     int64_t remaining = size - shrunken;
-    LOG(INFO) << logPrefix << "Trying to request spilling for remaining " << remaining << " bytes...";
+    LOG(INFO) << logPrefix << "Trying to request spilling for " << remaining << " bytes...";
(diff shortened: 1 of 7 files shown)


marin-ma:
LGTM. Could you also update the PR description to clarify the main change?

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -116,7 +116,7 @@ class ListenableArbitrator : public velox::memory::MemoryArbitrator {
       //
       // We are likely in destructor, do not throw. INFO log is fine since we have leak checks from Spark's memory
       //   manager
-      LOG(INFO) << "Memory pool " << pool->name() << " not completely shrunk when Memory::dropPool() is called";
+      LOG(INFO) << "Memory pool " << pool->name() << " not completely shrunken when Memory::dropPool() is called";
```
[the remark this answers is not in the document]
zhztheplayer: Yes. This is to limit the spill log outputs. After the patch shrink log will only be printed in debug build. While spill log remains in release build.

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -118,46 +118,64 @@ std::shared_ptr<ColumnarBatch> WholeStageResultIterator::next() {
...
+  }
+
+  // singleton
+  ConditionalSuspendedSection(const ConditionalSuspendedSection&) = delete;
+  ConditionalSuspendedSection(ConditionalSuspendedSection&&) = delete;
+  ConditionalSuspendedSection& operator=(const ConditionalSuspendedSection&) = delete;
+  ConditionalSuspendedSection& operator=(ConditionalSuspendedSection&&) = delete;
+
+ private:
+  velox::exec::SuspendedSection* section_ = nullptr;
```
Yohahaha, line 143: nit: why not use unique_ptr?

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -118,46 +118,64 @@ std::shared_ptr<ColumnarBatch> WholeStageResultIterator::next() {
...
-    LOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
-    return shrunk + spilledOut;
+    uint64_t total = shrunken + spilledOut;
+    DLOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
+    return shrunken + spilledOut;
   }
 
-  LOG(INFO) << logPrefix << "Successfully reclaimed total " << shrunk << " bytes.";
-  return shrunk;
+  DLOG(INFO) << logPrefix << "Successfully reclaimed total " << shrunken << " bytes.";
```
Yohahaha, line 177: nit: print a summary log may helpful, we are not always using DEBUG mode.

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -242,6 +242,24 @@ const MemoryUsageStats VeloxMemoryManager::collectMemoryUsageStats() const {
...
+  DLOG(INFO) << logPrefix << "Pool has reserved " << pool->currentBytes() << "/" << pool->root()->reservedBytes() << "/"
+             << pool->root()->capacity() << "/" << pool->root()->maxCapacity() << " bytes.";
+  DLOG(INFO) << logPrefix << "Shrinking...";
+  int64_t shrunken = pool->shrinkManaged(pool, size);
+  DLOG(INFO) << logPrefix << shrunken << " bytes released from shrinking.";
+  return shrunken;
+}
+} // namespace
+
+const int64_t VeloxMemoryManager::shrink(int64_t size) {
```
Yohahaha, line 259: nit: add override keyword.

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -242,6 +242,24 @@ const MemoryUsageStats VeloxMemoryManager::collectMemoryUsageStats() const {
...
 }
 
+namespace {
+int64_t shrinkVeloxMemoryPool(velox::memory::MemoryPool* pool, int64_t size) {
+  std::string poolName{pool->root()->name() + "/" + pool->name()};
+  std::string logPrefix{"Shrink[" + poolName + "]: "};
+  DLOG(INFO) << logPrefix << "Trying to shrink " << size << " bytes of data...";
+  DLOG(INFO) << logPrefix << "Pool has reserved " << pool->currentBytes() << "/" << pool->root()->reservedBytes() << "/"
+             << pool->root()->capacity() << "/" << pool->root()->maxCapacity() << " bytes.";
+  DLOG(INFO) << logPrefix << "Shrinking...";
```
Yohahaha, line 252: this log is confusing, could we remove it?

gluten-data/src/main/java/io/glutenproject/memory/nmm/NativeMemoryManagers.java
```diff
@@ -71,7 +72,21 @@ private static NativeMemoryManager createNativeMemoryManager(String name, Spille
...
+                                    () ->
+                                        new IllegalStateException(
+                                            ""
+                                                + "Shrink is requested before native "
+                                                + "memory manager is created. Try moving any "
+                                                + "actions about memory allocation out "
+                                                + "from the memory manager constructor.")),
+                        spiller // the input spiller, called after nmm.shrink was called
+                        ),
                     new GlutenMemoryConsumer.StatsBuilder() {
```
Yohahaha, line 90: Nice design! but we can decrease complexity by using spark's spill mechanism.
```
public long spill(long size, MemoryConsumer trigger) {
    if (trigger != this) {
        // shrink
    } else {
        // spill
    }
}
```

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: I'll open to both. Not a big deal. :)

cpp/velox/memory/VeloxMemoryManager.cc
zhztheplayer: > DLOG(INFO) << logPrefix << "Shrinking...";

Did you mean this line?

Yohahaha: yeah, this line has no meaning.

Yohahaha:
LGTM and some nits.

cpp/velox/memory/VeloxMemoryManager.cc
zhztheplayer: the keyword is only available inside class definition

Yohahaha: sry, please ignore.

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: I think a user should not be aware of the term "shrinking" which may be even confusing either. 

Anyway, is the log still be useful in your non-debug circumstance?

cpp/velox/compute/WholeStageResultIterator.cc
Yohahaha: unique_ptr is better to me, let's use modern cpp.

gluten-data/src/main/java/io/glutenproject/memory/nmm/NativeMemoryManagers.java
zhztheplayer: This would introduce bug. Since self-spilling always happens in the case of Velox's plan execution.

cpp/velox/compute/WholeStageResultIterator.cc
Yohahaha: I see there still has two LOG here, that's enough. 

BTW, print some warning logs when spillStrategy is `none` would helpful.

gluten-data/src/main/java/io/glutenproject/memory/nmm/NativeMemoryManagers.java
Yohahaha: Let's use c1 for consumer1 and c2 for consumer2.

c1 will trigger other to spill then spill itself, if c1 has unused reservations when trigger c2 to spill, c2 will shrink then spill if shrink is not enough right? in this case, actual spill happens when there still has enough memory.

we can talk this offline.

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: > unique_ptr is better to me, let's use modern cpp.

Can you elaborate? I am not sure modern cpp recommends using smart pointers in this case. `unique_ptr` should indeed be used when **ownership-transferring** is going to happen. However in this case we just define a owning pointer inside the class.

Just to make sure we don't create new rules casually to our CPP coding convention. The convention should be followed once built.

zhztheplayer: Please refer to https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: > BTW, print some warning logs when spillStrategy is none would helpful.

Yes this could be helpful. But let's use a new ticket for that and could you help open one? Thanks.

zhztheplayer:
Let's keep the discussion and merge this. We'd open new tickets once deciding adding new changes upon this PR.

The discussions will be helpful to further readers of the PR, feel free to continue. Thanks!

marin-ma (approved):

cpp/velox/compute/WholeStageResultIterator.cc
Yohahaha: https://www.oreilly.com/library/view/effective-modern-c/9781491908419/ch04.html

I'm not going to discuss which usage is the best and create more coding rules, current implementation is ok, but I prefer use unique_ptr to replace manually `delete`, let language provided ability help us.

You can keep this usage if you like, this not cause problem.

winningsix (commented):
Sorry to post review comments late.

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -118,46 +118,64 @@ std::shared_ptr<ColumnarBatch> WholeStageResultIterator::next() {
...
-      uint64_t spilledOut = pool_->reclaim(remaining);
-      LOG(INFO) << logPrefix << "Successfully spilled out " << spilledOut << " bytes.";
-      uint64_t total = shrunk + spilledOut;
-      LOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
-      return shrunk + spilledOut;
-    }
-    // suspend since we are on driver
-    velox::exec::SuspendedSection noCancel(thisDriver);
+    // suspend the driver when we are on it
+    ConditionalSuspendedSection noCancel(thisDriver, thisDriver != nullptr);
```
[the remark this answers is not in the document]
zhztheplayer: Sure. Let's do that in another patch.

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -118,46 +118,64 @@ std::shared_ptr<ColumnarBatch> WholeStageResultIterator::next() {
...
 int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
   std::string poolName{pool_->root()->name() + "/" + pool_->name()};
   std::string logPrefix{"Spill[" + poolName + "]: "};
-  LOG(INFO) << logPrefix << "Trying to reclaim " << size << " bytes of data...";
-  LOG(INFO) << logPrefix << "Pool has reserved " << pool_->currentBytes() << "/" << pool_->root()->reservedBytes()
-            << "/" << pool_->root()->capacity() << "/" << pool_->root()->maxCapacity() << " bytes.";
-  LOG(INFO) << logPrefix << "Shrinking...";
-  int64_t shrunk = pool_->shrinkManaged(pool_.get(), size);
-  LOG(INFO) << logPrefix << shrunk << " bytes released from shrinking.";
+  DLOG(INFO) << logPrefix << "Trying to reclaim " << size << " bytes of data...";
```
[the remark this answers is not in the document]
zhztheplayer: I was not ware of the term `acquireMemBySize`? But the target `size` is indeed used here and passed to Velox's reclaim and shrink API. Besides Velox doesn't necessarily reclaim or spill exactly the same size with the target size. It's just an expected number.

Issue #3017, opened 2023-09-05
[VL] Allow users to set spill partitions/levels


zhouyuan:
### Description

The spill partitions and spill levels are not exposed to user. The default settings may bring OOM under large table.

https://facebookincubator.github.io/velox/configs.html#spilling

<html>
<body>

max_spill_level | integer | 4 | The maximum allowed spilling level with zero being the initial spilling level. Applies to hash join build spilling which might use recursive spilling when the build table is very large. -1 means unlimited. In this case an extremely large query might run out of spilling partition bits. The max spill level can be used to prevent a query from using too much io and cpu resources.
-- | -- | -- | --

</body>
</html>

<html>
<body>

join_spiller_partition_bits | integer | 2 | The number of bits (N) used to calculate the spilling partition number for hash join: 2 ^ N. At the moment the maximum value is 3, meaning we only support up to 8-way spill partitioning.
-- | -- | -- | --

</body>
</html>

PR #3018, opened 2023-09-05
[GLUTEN-3017][VL] Allow users to set spill partitions/levels

zhouyuan:

This patch allows users to set spill partitions/levels, as on large table, the default values may bring OOM. 

(Fixes: #3017 )

## How was this patch tested?

manually tested

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -963,6 +963,21 @@ object GlutenConfig {
       .checkValues(Set("local", "heap-over-local"))
       .createWithDefaultString("local")
 
+  val COLUMNAR_VELOX_MAX_SPILL_LEVEL =
+    buildConf("spark.gluten.sql.columnar.backend.velox.maxSpillLevel")
+      .internal()
+      .doc("The maximum allowed spilling level with zero being the initial spilling level.")
+      .intConf
+      .createWithDefault(4)
+
+  val COLUMNAR_VELOX_SPILL_PARTITION_BITS =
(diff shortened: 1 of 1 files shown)


shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -963,6 +963,21 @@ object GlutenConfig {
       .checkValues(Set("local", "heap-over-local"))
       .createWithDefaultString("local")
 
+  val COLUMNAR_VELOX_SPILL_MAX_LEVEL =
+    buildConf("spark.gluten.sql.columnar.backend.velox.spillMaxLevel")
```
zhztheplayer, line 967: ~~spillMaxLevel~~ -> maxSpillLevel

Since in our cpp code

https://github.com/oap-project/gluten/blob/4bd1b04650d19844da3f56430918fa3c2c2a92b8/cpp/velox/compute/WholeStageResultIterator.cc#L51

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -963,6 +963,21 @@ object GlutenConfig {
...
+      .createWithDefault(4)
+
+  val COLUMNAR_VELOX_SPILL_PARTITION_BITS =
+    buildConf("spark.gluten.sql.columnar.backend.velox.spillPartitionBits")
+      .internal()
+      .doc("The number of bits used to calculate the spilling partition number." +
+        " At the moment the maximum value is 3.")
+      .intConf
+      .createWithDefault(2)
+
```
FelixYBW, line 980: Do the parameters pass to Velox?

PR #3029, opened 2023-09-05
[GLUTEN-2941] Enable spill support during shuffle splitter stop

marin-ma:
The stop function performs several tasks:
1. Opens the final data file.
2. Iterates over each partition ID (pid) to:
   a. Record the offset for each partition in the final file.
   b. Merge data from spilled files and write to the final file.
   c. Write cached payloads to the final file.
   d. Create the last payload from split buffer, and write to the final file.
   e. Optionally, write End of Stream (EOS) if any payload has been written.
3. Closes and deletes all the spilled files.
4. Records various metrics such as total write time, bytes evicted, and bytes written.
5. Clears any buffered resources and closes the final file.

Spill handling:
Spill is allowed during stop(). If OOM happens during stop(), it will first shrink the split buffers and then perform spill if necessary.
Among above steps, 1. and 2.d requires memory allocation and may trigger spill.
If spill is triggered by 1., cached payloads of all partitions will be spilled.
If spill is triggered by 2.d, cached payloads of the remaining unmerged partitions will be spilled.

### DIFF
--- a/cpp/core/shuffle/LocalPartitionWriter.cc
+++ b/cpp/core/shuffle/LocalPartitionWriter.cc
@@ -76,7 +76,7 @@ class PreferEvictPartitionWriter::LocalPartitionWriterInstance {
     RETURN_NOT_OK(ensureOpened());
 #endif
     RETURN_NOT_OK(writeRecordBatchPayload(spilledFileOs_.get()));
-    clearCache();
+    shuffleWriter_->clearCachedPayloads(partitionId_);
     return arrow::Status::OK();
   }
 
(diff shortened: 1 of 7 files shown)


marin-ma:
cc: @zhztheplayer

FelixYBW:
Can you add some info how the patch works? how we added spill during stop?

marin-ma:
> Can you add some info how the patch works? how we added spill during stop?

Thanks. Updated.

marin-ma:
@zhztheplayer Can we merge this one first? Then I can rebase #2982 and add the spill->shrink logic, or there will be lots of conflict.

zhztheplayer (approved):

Issue #3034, opened 2023-09-06
[VL] Fix allocating negative buffer in R2C


jackylee-ch:
### Backend

VL (Velox)

### Bug description

We are running gluten+velox in our environment with `bathSize=2048`, and then we found the Exception below. It seems that the `offset` becomes a negative value.
```
java.lang.IllegalArgumentException: the requested size must be non-negative
	at io.glutenproject.shaded.org.apache.arrow.util.Preconditions.checkArgument(Preconditions.java:136)
	at io.glutenproject.shaded.org.apache.arrow.memory.BaseAllocator.buffer(BaseAllocator.java:290)
	at io.glutenproject.shaded.org.apache.arrow.memory.RootAllocator.buffer(RootAllocator.java:29)
	at io.glutenproject.shaded.org.apache.arrow.memory.BaseAllocator.buffer(BaseAllocator.java:279)
	at io.glutenproject.shaded.org.apache.arrow.memory.RootAllocator.buffer(RootAllocator.java:29)
	at io.glutenproject.execution.RowToVeloxColumnarExec$$anon$1.nativeConvert(RowToVeloxColumnarExec.scala:139)
	at io.glutenproject.execution.RowToVeloxColumnarExec$$anon$1.next(RowToVeloxColumnarExec.scala:179)
	at io.glutenproject.execution.RowToVeloxColumnarExec$$anon$1.next(RowToVeloxColumnarExec.scala:93)
	at io.glutenproject.vectorized.CloseableColumnBatchIterator.next(CloseableColumnBatchIterator.scala:56)
```

After changing `offset` to Long, we found that the following errors. After tracing the problem, we found that the memory of `SpecificUnsafeProjection` also has an memory limit, which is slightly smaller than `Int.MaxValue`.
```
Reason: Operator::getOutput failed for [operator: ValueStream, plan node ID: 0]: Error during calling Java code from native code: java.lang.IllegalArgumentException: Cannot grow BufferHolder by size 1602585576 because the size after growing exceeds size limitation 2147483632
	at org.apache.spark.sql.catalyst.expressions.codegen.BufferHolder.grow(BufferHolder.java:71)
	at org.apache.spark.sql.catalyst.expressions.codegen.UnsafeWriter.grow(UnsafeWriter.java:63)
	at org.apache.spark.sql.catalyst.expressions.codegen.UnsafeWriter.writeUnalignedBytes(UnsafeWriter.java:127)
[...]

PR #3035, opened 2023-09-06
[GLUTEN-3034][CORE] Fix allocating negative buffer size

jackylee-ch:
Fixes: #3034 

## How was this patch tested?
unit tests


jackylee-ch:
This patch works in our environment, PTAL @zhztheplayer @jinchengchenghh

jinchengchenghh:
Can you add unit test to track this one?

jackylee-ch:
> Can you add unit test to track this one?

Sure, it's easy to add a unit test with large string.

backends-velox/src/main/scala/io/glutenproject/execution/RowToVeloxColumnarExec.scala
```diff
@@ -120,6 +120,9 @@ case class RowToVeloxColumnarExec(child: SparkPlan)
               val estimatedBufSize = Math.max(
                 Math.min(sizeInBytes.toDouble * numRows * 1.2, 31760L * numRows),
                 sizeInBytes.toDouble * 10)
+              // Here we limit the max buffer size to Int.MaxValue, which is the memory limit for
+              // UnsafeProject.
+              val estimatedNumRow = Math.min(Int.MaxValue / estimatedBufSize, numRows)
```
ulysses-you, line 125: Curious, after apply row to columnar, why do we still have `UnsafeProject`, should it be `ProjectExecTransformer` ?

jackylee-ch: Here we fix the memory used problem while converting `UnsafeProject` to `Columnar`.  If the row data is largeI, [iterator.next()](https://github.com/oap-project/gluten/blob/main/backends-velox/src/main/scala/io/glutenproject/execution/RowToVeloxColumnarExec.scala#L135) would failed with `Cannot grow BufferHolder by size 1602585576 because the size after growing exceeds size limitation 2147483632`.
More details can be found in #3034.

jackylee-ch: It's a good question, I would check the codegen to find out why `writer.reset()` is not called.

zhouyuan:
@jackylee-ch can you please do a rebase?

jackylee-ch:
Hi, @FelixYBW . Did you reproduce the same issue on your side?
I remember tuning the batch size could mitigate it to some extent; if this PR is still required for a stable fix, I’m happy to take another look.

abbas-gadhia:
I noticed that the title of the PR is about a negative buffer size but it seems that the fix is related to the "exceeds size limitation" error. 

Is this about the former error i.e negative size or size limitation?

We seem to be facing the negative size error with the following stacktrace
```
Reason: Operator::getOutput failed for [operator: ValueStream, plan node ID: 0]: Error during calling Java code from native code: java.lang.IllegalArgumentException: Cannot grow BufferHolder by size -588050728 because the size is negative
	at org.apache.spark.sql.catalyst.expressions.codegen.BufferHolder.grow(BufferHolder.java:67)
	at org.apache.spark.sql.catalyst.expressions.codegen.UnsafeWriter.grow(UnsafeWriter.java:63)
	at org.apache.spark.sql.catalyst.expressions.codegen.UnsafeWriter.writeUnalignedBytes(UnsafeWriter.java:127)
	at org.apache.spark.sql.catalyst.expressions.codegen.UnsafeWriter.write(UnsafeWriter.java:110)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.writeFields_0_3$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
	at org.apache.gluten.execution.RowToVeloxColumnarExec$$anon$1.next(RowToVeloxColumnarExec.scala:159)
	at org.apache.gluten.execution.RowToVeloxColumnarExec$$anon$1.next(RowToVeloxColumnarExec.scala:139)

```
I think the negative size error originates from velox i.e the `serializeVariableWidth` method, but i still need to confirm this.

jackylee-ch:
Hi, @abbas-gadhia. We previously encountered the issue with negative values. In my PR, I updated the type to Long, but realized it could then exceed the intended maximum. To address this, I’ve added a limit constraint to ensure the value stays within the required range. More detail can be found in #3034.

abbas-gadhia:
Hi @jackylee-ch
Correct me if i'm wrong, but it seems from the changes in the PR that it would reduce the number of rows to keep the buffer size in check. 

This differs from the negative size error i mentioned above, since it occurs when the BufferHolder tries to grow itself for a UTF8String and the numBytes of that string is negative.

jackylee-ch:
@abbas-gadhia Yes, you are right. For your problem, we should focus more on why the incoming `neededSize` became a negative value.

And the negative buffer size I met before has been fixed in https://github.com/apache/incubator-gluten/pull/5441

FelixYBW:
@jackylee-ch thank you for your info. I reopen it per @abbas-gadhia's request. Since it's different issue. Let's close it.

abbas-gadhia:
I'm trying to see if UnsafeRowFast.cpp#serializeRow can cause this overflow, specifically this code
`uint64_t sizeAndOffset = variableWidthOffset << 32 | size;`

It seems it 'might' overflow in the case of many varchar columns with large data > 125 columns. Still trying to confirm

abbas-gadhia:
FYI. This turned out to be a non-issue in OSS gluten. It was something related to a data type mismatch issue in my code that led to a timestamp_ntz data type being interpreted as a string type

PR #3036, opened 2023-09-06
[VL] Get avaliable off-heap memory for split buffer calculation everytime split() is called

zhztheplayer:
Rely on https://github.com/oap-project/gluten/pull/2982

Without the PR, we pass available off-heap memory size to shuffle writer when it's created. The problem is sometimes the shuffle writer is created before operators allocating large block memory like hash agg. Then the split buffer is preallocated according to whole off-heap memory instead of available memory when shuffle writer really works.

The PR pass the available memory to shuffle write with each batches, so we can recalculate the buffer size on each batch.

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -1275,6 +1275,8 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
     VS_PRINTLF(bytesPerRow);
 
     // remove the cache memory size since it can be spilled.
+    // The logic here is to keep the split buffer as large as possible, to get max batch size for reducer
+    // The risk is the cached buffer must be spilled once we evict a batch from split buffer
     // we can't use pool_->bytes_allocated() here because it's arrow::default_pool(). we should define a
     // pool for spill buffer, just like the pool for split buffer
     // memLimit+=pool_->bytes_allocated();
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -1218,17 +1218,17 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
 
     VS_PRINTLF(bytesPerRow);
 
-    //remove the cache memory size since it can be spilled.
-    //we can't use pool_->bytes_allocated() here because it's arrow::default_pool(). we should define a
+    // remove the cache memory size since it can be spilled.
+    // we can't use pool_->bytes_allocated() here because it's arrow::default_pool(). we should define a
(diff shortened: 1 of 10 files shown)


FelixYBW:
@zhztheplayer  pr2982 is merged. Can you rebase this one?

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1166,9 +1166,16 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
...
-    uint64_t preAllocRowCnt = options_.offheap_per_task > 0 && sizePerRow > 0
-        ? options_.offheap_per_task / sizePerRow / numPartitions_ >> 2
-        : options_.buffer_size;
+    //remove the cache memory size since it can be spilled.
+    //we can't use pool_->bytes_allocated() here because it's arrow::default_pool(). we should define a 
+    // pool for spill buffer, just like the pool for split buffer
+    // memLimit+=pool_->bytes_allocated();
+    memLimit += totalCachedPayloadSize();
+    //make sure split buffer uses 128M memory at least, let's hardcode it here for now
+    if (memLimit<128*1024*1024) memLimit=128*1024*1024;
```
FelixYBW, line 1175: Let's make 128M an configurable value

zhztheplayer: Is this required from this PR? Or I'd move to the topic assigning initial reservation to memory pools.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1275,9 +1274,17 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
...
     VS_PRINTLF(bytesPerRow);
 
-    uint64_t preAllocRowCnt = options_.offheap_per_task > 0 && bytesPerRow > 0
-        ? options_.offheap_per_task / bytesPerRow / numPartitions_ >> 2
-        : options_.buffer_size;
+    // remove the cache memory size since it can be spilled.
+    // we can't use pool_->bytes_allocated() here because it's arrow::default_pool(). we should define a
+    // pool for spill buffer, just like the pool for split buffer
+    // memLimit+=pool_->bytes_allocated();
+    memLimit += totalCachedPayloadSize();
```
FelixYBW, line 1283: @marin-ma can we use cached buffer memory pool directly?

FelixYBW (approved):

marin-ma (approved):

cpp/velox/shuffle/VeloxShuffleWriter.cc
marin-ma: Yes. Removed in #3159
https://github.com/oap-project/gluten/pull/3159/files#diff-55954c774f5d66d0d359c3c7baa29be65f75198b30fd27603b0456ccd11b6a79R1786

PR #3043, opened 2023-09-06
[GLUTEN-2877][VL]Feat: Support read iceberg cow table

liujiayi771:

Support read iceberg cow table in gluten. Resolve the first step in https://github.com/oap-project/gluten/issues/2877.

## How was this patch tested?
Run tpcds benchmark in 1T tpcds iceberg tables.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
@@ -25,11 +25,13 @@ import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.connector.read.{InputPartition, Scan}
-import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.{InSubqueryExec, ScalarSubquery, SparkPlan}
 import org.apache.spark.sql.execution.datasources.v2.{BatchScanExecShim, FileScan}
 import org.apache.spark.sql.execution.metric.SQLMetric
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.vectorized.ColumnarBatch
+
+import org.apache.iceberg.spark.source.GlutenSparkBatchQueryScan
 
 import java.util.Objects
 
@@ -48,6 +50,8 @@ class BatchScanExecTransformer(
   override def filterExprs(): Seq[Expression] = scan match {
     case fileScan: FileScan =>
       fileScan.dataFilters ++ pushdownFilters
+    case scan if scan.getClass.getSimpleName == "SparkBatchQueryScan" =>
+      Seq.empty
     case _ =>
       throw new UnsupportedOperationException(s"${scan.getClass.toString} is not supported")
   }
@@ -116,8 +120,19 @@ class BatchScanExecTransformer(
   override def metricsUpdater(): MetricsUpdater =
     BackendsApiManager.getMetricsApiInstance.genBatchScanTransformerMetricsUpdater(metrics)
 
-  @transient protected lazy val filteredFlattenPartitions: Seq[InputPartition] =
+  @transient protected lazy val filteredFlattenPartitions: Seq[InputPartition] = {
+    runtimeFilters.foreach {
+      case DynamicPruningExpression(inSubquery: InSubqueryExec) =>
+        executeInSubqueryForDynamicPruningExpression(inSubquery)
+      case e: Expression =>
+        e.foreach {
+          case s: ScalarSubquery => s.updateResult()
+          case _ =>
+        }
+      case _ =>
+    }
     filteredPartitions.flatten
+  }
 
   private def getPushdownFilters: Seq[Expression] = {
     pushdownFilters
@@ -128,6 +143,7 @@ class BatchScanExecTransformer(
     case "ParquetScan" => ReadFileFormat.ParquetReadFormat
     case "DwrfScan" => ReadFileFormat.DwrfReadFormat
     case "ClickHouseScan" => ReadFileFormat.MergeTreeReadFormat
+    case "SparkBatchQueryScan" => GlutenSparkBatchQueryScan.getFileFormat(scan)
     case _ => ReadFileFormat.UnknownFormat
   }
 }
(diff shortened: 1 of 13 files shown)


gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
```diff
@@ -48,6 +50,8 @@ class BatchScanExecTransformer(
   override def filterExprs(): Seq[Expression] = scan match {
     case fileScan: FileScan =>
       fileScan.dataFilters ++ pushdownFilters
+    case scan if scan.getClass.getSimpleName == "SparkBatchQueryScan" =>
```
jinchengchenghh, line 53: Why not use isInstanceOf ?

gluten-core/src/main/scala/org/apache/iceberg/spark/source/GlutenSparkBatchQueryScan.scala
```diff
@@ -0,0 +1,46 @@
...
+        tasks.map(_.asCombinedScanTask()).foreach {
+          task =>
+            val file = task.files().asScala.head.file()
+            file.format() match {
+              case FileFormat.PARQUET => return ReadFileFormat.ParquetReadFormat
+              case FileFormat.ORC => return ReadFileFormat.OrcReadFormat
+              case _ =>
+            }
+        }
+        throw new UnsupportedOperationException("iceberg Only support parquet and orc file format.")
```
jinchengchenghh, line 41: iceberg -> Iceberg

gluten-core/src/main/scala/org/apache/iceberg/spark/source/GlutenSparkBatchQueryScan.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object GlutenSparkBatchQueryScan {
+  def getFileFormat(scan: Scan): ReadFileFormat = {
+    scan match {
+      case scan: SparkBatchQueryScan =>
+        val tasks = scan.tasks().asScala
+        tasks.map(_.asCombinedScanTask()).foreach {
+          task =>
+            val file = task.files().asScala.head.file()
+            file.format() match {
```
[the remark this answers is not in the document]
jinchengchenghh: Maybe this is a constraint, we can document it. Check all the files is a too high expense and this situation is rarely.

jinchengchenghh:
Can you add the test about iceberg table?

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
liujiayi771: SparkBatchQueryScan is protected class in iceberg, we can not import this class here.

gluten-core/src/main/scala/org/apache/iceberg/spark/source/GlutenSparkBatchQueryScan.scala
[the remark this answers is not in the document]
liujiayi771: Yes, we should clarify the constraints.

gluten-core/src/test/scala/org/apache/spark/softaffinity/SoftAffinitySuite.scala
```diff
@@ -47,7 +47,7 @@ class SoftAffinitySuite extends QueryTest with SharedSparkSession with Predicate
       ).toArray
     )
 
-    val locations = SoftAffinityUtil.getFilePartitionLocations(partition)
+    val locations = SoftAffinityUtil.getFilePartitionLocations(partition, partition.files)
```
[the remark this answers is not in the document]
liujiayi771: The modification here is because the partition file of iceberg cannot be obtained directly through `icebergPartition.files`.

https://github.com/liujiayi771/gluten/blob/df0d79568da5561ba5b45a425903f21922558cc3/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala#L127

ulysses-you:
I'm afraid it would be hard to maintain if we mixed iceberg code into everywhere on gluten. What this pr does is something like:

```
Gluten    Iceberg
   \       /
     \   /
       |
   Spark SQL
```

I wonder if we can make it like:
```
    Iceberg
       |
     Gluten 
       |
   Spark SQL
```

The main difference is that, we can create new modules for iceberg or other Spark downstream projects. We can expose the power of Gluten as extension as much as possile, then these modules can be developed base on Gluten extension.

liujiayi771:
> Can you add the test about iceberg table?
Add a new test case VeloxTPCHIcebergSuite.

liujiayi771:
> Can you add the test about iceberg table?

Add a new test case VeloxTPCHIcebergSuite.

liujiayi771:
> I'm afraid it would be hard to maintain if we mixed iceberg code into everywhere on gluten. What this pr does is something like:
> 
> ```
> Gluten    Iceberg
>    \       /
>      \   /
>        |
>    Spark SQL
> ```
> 
> I wonder if we can make it like:
> 
> ```
>     Iceberg
>        |
>      Gluten 
>        |
>    Spark SQL
> ```
> 
> The main difference is that, we can create new modules for iceberg or other Spark downstream projects. We can expose the power of Gluten as extension as much as possile, then these modules can be developed base on Gluten extension.

I agree with you. We need to add the ability to inject new rules into the scan processing part in gluten-core. We will optimize this later. Our team are also supporting delta,  hudi and paimon. support for different lake formats requires a better architecture.

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
felipepessoto, line 53: Comparing simple name may be dangerous. What if Delta Lake (or other formats) has the same class name

liujiayi771: > Comparing simple name may be dangerous. What if Delta Lake (or other formats) has the same class name

https://github.com/oap-project/gluten/blob/ff79df0084aec978161e99359c3d4dd58376ec45/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala#L127

You are right, it will be more accurate to use a name containing package name, but basically the scan names of different formats are different. Gluten uses simple name when judging the scan type.

yma11:
@liujiayi771 Do you work on refactor this PR or plan to do so based on [discussion](https://github.com/oap-project/gluten/issues/3378)?

yma11:
> We will post the design in the next couple of days as it is almost finished. After discussing it, we will refactor this pull request.

I see. Look forward for your design. Thanks.

yma11:
Close this PR as related implementation all done in later PRs.

PR #3047, opened 2023-09-07
[VL] Add partial hash aggregation memory configurations to avoid OOM

kecookier:
HashAggregation will check these four config value to decide whether abandon partial aggregation or not. We may need fine-tune these config value to avoid OOM in Large-scale Spark job.

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -252,6 +252,14 @@ class GlutenConfig(conf: SQLConf) extends Logging {
   def enableParquetRowGroupMaxMinIndex: Boolean =
     conf.getConf(ENABLE_PARQUET_ROW_GROUP_MAX_MIN_INDEX)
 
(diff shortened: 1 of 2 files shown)


zuochunwei (approved):

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -1128,4 +1136,40 @@ object GlutenConfig {
...
+
+  val MAX_PARTIAL_AGGREGATION_MEMORY_RATIO =
+    buildConf("spark.gluten.sql.columnar.backend.velox.maxPartialAggregationMemoryRatio")
+      .internal()
+      .doc(
+        "Set the max memory of partial aggregation as "
+          + "maxPartialAggregationMemoryRatio of offheap size."
+      )
+      .doubleConf
+      .createOptional
```
zhztheplayer, line 1148: @kecookier Can we add default value here?

PR #3052, opened 2023-09-07
[GLUTEN-2808][VL][FOLLOWUP] All managed Velox memory pools should be shrinkable

Yohahaha:
Use builder pattern refactor NativeMemoryManager.

After this patch, caller can create a NativeMemoryManager with a optional spiller.


gluten-data/src/main/java/io/glutenproject/memory/nmm/NativeMemoryManager.java
```diff
@@ -87,4 +101,114 @@ public int priority() {
...
+                        }
+                      }),
+                  overAcquiredRatio));
+      // listener
+      ManagedReservationListener rl =
+          new ManagedReservationListener(target, TaskResources.getSharedUsage());
+      // native memory manager
+      out.set(
+          TaskResources.addResourceIfNotRegistered(
+              this.name, () -> NativeMemoryManager.create(this.name, rl)));
```
zhztheplayer, line 200: Any strong reason returning a cached instance from a builder?

Yohahaha: Just align with NativeMemoryManagers.contextInstance, and keep same usage in build now.
Do you think we need provide two API, one return cached instance and another return non cached instance?

zhztheplayer: > Do you think we need provide two API, one return cached instance and another return non cached instance?

Yes, but the factory class `NativeMemoryManagers` already contains the two types of API? Do we really need a builder pattern here as of now?

Yohahaha: > but the factory class NativeMemoryManagers already contains the two types of API?

`contextInstance` return cached instance, `create` return non-cached. Could you give more reason for the non-cached instance?

> Do we really need a builder pattern here as of now?

I think yes, use builder pattern can help us hide the optional spiller and reduce overload methods, 
 and `NativeMemoryManagers` factory was good but its naming was a little bit confused that not clearly.

Yohahaha: I found this origin codes.

```java
  public NativeMemoryAllocator contextInstance(String name) {
    if (!TaskResources.inSparkTask()) {
      return globalInstance();
    }
    final String id =
        NativeMemoryAllocatorManager.class + "-" + System.identityHashCode(global) + "-" + name;
    return TaskResources.addResourceIfNotRegistered(
            id,
            () ->
                createNativeMemoryAllocatorManager(
                    name,
                    TaskResources.getLocalTaskContext().taskMemoryManager(),
                    TaskResources.getSharedMetrics(),
                    0.0D,
                    Spiller.NO_OP,
                    global))
        .getManaged();
  }

  public NativeMemoryAllocator create(String name, double overAcquiredRatio, Spiller spiller) {
    if (!TaskResources.inSparkTask()) {
      throw new IllegalStateException("Spiller must be used in a Spark task");
    }

    final NativeMemoryAllocatorManager manager =
        createNativeMemoryAllocatorManager(
            name,
            TaskResources.getLocalTaskContext().taskMemoryManager(),
            TaskResources.getSharedMetrics(),
            overAcquiredRatio,
            spiller,
            global);
    return TaskResources.addAnonymousResource(manager).getManaged();
  }
```

zhztheplayer: > `NativeMemoryManagers` factory was good but its naming was a little bit confused that not clearly.

`<Class>s` is a very common naming pattern for factory / utility classes for class `<Class>`.

I think the code around `MemoryManager` indeed needs some reorganizations but the patch doesn't seem to help so far. Builder pattern may help in future, but not at this time, in the case that it returns a cached instance.

Yohahaha: > in the case that it returns a cached instance.

It's not with this patch, and I dont realize why you insist it. we already use memory manager's name as key to cache each instance, `contextInstnace` and `create` seems no difference for me, does there will have two memory manager instance created by `create` in one Spark task?

zhztheplayer: If we want to cache an object, please move the caching logic out from the builder. Or rename the relevant builder methods to let caller know it may return cached object. For example:

```java
A.newBuilder()
  .name(name)
  .setProperty(property)
  .findOrCreate()
```

The big deal here is that we didn't actually use the cached object for spillable instances, and the PR makes it cacheable. This needs some careful design to the APIs and please rethink about the current builder pattern and it's method naming.  Also the PR should be renamed either.

If you really want to alter this part of code, I think adding some UTs would be helpful to reviewing. 

For example:

1. Create for nameA twice without spiller;
2. Create for nameA twice, with the same spiller;
3. Create for nameA twice, with different spiller;

Beside, I still doubt that doing this helps on reducing memory manager code's complexity. I suggest we lower this kind of work's priority until it's really needed.

Yohahaha: Agree for this patch is not urgent and important, will close it. Not agree for this patch does not reduce code complexity.

Personally, I dont think community should always accept urgent and important patches.

zhztheplayer: > Personally, I dont think community should always accept urgent and important patches.

We definitely should not accept urgent patches only. But we should make sure we don't bring new complexities during resolve the old ones.

For this patch, it introduces a mm builder with cached output, and previously it was not cached. I understand that using consumer name as cache key is reliable in current usages, since we pass the same spiller ever time. But it's a new public API and open for all developers. One may pass another spiller for same name in future and we should make sure the API still returns meaningful instance or throw reasonably in that case. Let's continue this work once we have time to do a good design.

PR #3056, opened 2023-09-07
[VL] Gluten-it: Add option --random-kill-tasks to emulate the case that tasks are frequently killed and retried

zhztheplayer:
Umbrella: https://github.com/oap-project/gluten/issues/3048

To simulate speculative execution's task killing actions in local test, even using more aggressive kills.


zhouyuan (approved):
👍

Issue #3058, opened 2023-09-07
[VL] executor core dump on speculation on


zhouyuan:
### Backend

VL (Velox)

### Bug description

```
terminate called after throwing an instance of 'gluten::GlutenException'
  what():  Error during calling Java code from native code: io.glutenproject.exception.GlutenException: java.nio.channels.ClosedByInterruptException
  at io.glutenproject.vectorized.LowCopyFileSegmentJniByteInputStream.read(LowCopyFileSegmentJniByteInputStream.java:110)
  at io.glutenproject.vectorized.ColumnarBatchOutIterator.nativeNext(Native Method)
  at io.glutenproject.vectorized.ColumnarBatchOutIterator.nextInternal(ColumnarBatchOutIterator.java:57)
  at io.glutenproject.vectorized.GeneralOutIterator.next(GeneralOutIterator.java:36)
  at io.glutenproject.vectorized.ColumnarBatchSerializerInstance$$anon$1.liftedTree1$1(ColumnarBatchSerializer.scala:154)
  at io.glutenproject.vectorized.ColumnarBatchSerializerInstance$$anon$1.readValue(ColumnarBatchSerializer.scala:153)
  at org.apache.spark.serializer.DeserializationStream$$anon$2.getNext(Serializer.scala:188)
  at org.apache.spark.serializer.DeserializationStream$$anon$2.getNext(Serializer.scala:185)
```

Executor should be running fine, with some tasks failed

### Spark version

Spark-3.3.x

PR #3059, opened 2023-09-07
[GLUTEN-3058][VL] Fix shuffle reader on speculation run

zhouyuan:

this patch improved the exception handling on shuffle reading. If it's `ClosedByInterruptException` Gluten will print log only

https://github.com/apache/spark/pull/25674/files

(Fixes: #3058)

## How was this patch tested?

manually test, with speculation on


gluten-data/src/main/java/io/glutenproject/vectorized/LowCopyFileSegmentJniByteInputStream.java
```diff
@@ -106,6 +107,9 @@ public long read(long destAddress, long maxSize) {
       bytesRead += bytes;
       left -= bytes;
       return bytes;
+    } catch (ClosedByInterruptException ce) {
+      System.out.printf("Failed to create input stream from local block, " + ce.getMessage());
```
zhouyuan, line 111: CC: @jackylee-ch

PR #3060, opened 2023-09-07
[GLUTEN-3057][VL] Crash by dangling pointer to Arrow memory pool in shuffle writer, when task is killed

zhztheplayer:


cpp/core/shuffle/ShuffleWriter.h
```diff
@@ -234,15 +234,17 @@ class ShuffleWriter {
       : numPartitions_(numPartitions),
         partitionWriterCreator_(std::move(partitionWriterCreator)),
         options_(std::move(options)),
-        codec_(createArrowIpcCodec(options_.compression_type, options_.codec_backend)),
-        pool_(std::make_shared<ShuffleBufferPool>(options_.memory_pool)) {}
+        pool_(std::make_shared<ShuffleBufferPool>(options_.memory_pool)),
+        codec_(createArrowIpcCodec(options_.compression_type, options_.codec_backend)) {}
```
zhouyuan, line 238: so the fix is to move this line up?

zhztheplayer: Yes, the poll should be alive during the buffers are destructed,

zhouyuan:
@jackylee-ch

zhouyuan (approved):
👍

PR #3065, opened 2023-09-07
[GLUTEN-3062][VL] Check Java exception immediately when a call from C++ to Java ended

zhztheplayer:
May fix https://github.com/oap-project/gluten/issues/3062, still in testing

After the patch, we should abide by the rule that any C++ to Java invocations should be followed by a `checkException()` guard.

`checkException()` prevents its guarding code from running into undefined state and throw immediately if an Java exception was thrown.

For Java to C++, we have JNI_METHOD_START and JNI_METHOD_END to do the similar guarding job.

In future we may want to use macros to ease the manual `checkException` calls.


zhouyuan (approved):
👍

PR #3066, opened 2023-09-07
[WIP][CORE] Optimize shuffle partitioner algorithm to take advantage of  auto-vectorization in GCC

kecookier:
Why we make this change?

We made vs. benchmark in https://quick-bench.com/, And the result shows the speed of the improved Partitioner::computer is 2x faster than the old implementation.

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/core/shuffle/RoundRobinPartitioner.cc
+++ b/cpp/core/shuffle/RoundRobinPartitioner.cc
@@ -22,15 +22,27 @@ namespace gluten {
 arrow::Status gluten::RoundRobinPartitioner::compute(
     const int32_t* pidArr,
     const int64_t numRows,
-    std::vector<uint16_t>& partitionId,
-    std::vector<uint32_t>& partitionIdCnt) {
-  std::fill(std::begin(partitionIdCnt), std::end(partitionIdCnt), 0);
-  partitionId.resize(numRows);
-  for (auto& pid : partitionId) {
-    pid = pidSelection_;
-    partitionIdCnt[pidSelection_]++;
-    pidSelection_ = (pidSelection_ + 1) == numPartitions_ ? 0 : (pidSelection_ + 1);
+    std::vector<uint16_t>& row2Partition,
+    std::vector<uint32_t>& partition2RowCount) {
+  std::fill(std::begin(partition2RowCount), std::end(partition2RowCount), 0);
+  row2Partition.resize(numRows);
+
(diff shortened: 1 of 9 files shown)


cpp/core/shuffle/RoundRobinPartitioner.cc
```diff
@@ -22,15 +22,27 @@ namespace gluten {
...
-  for (auto& pid : partitionId) {
-    pid = pidSelection_;
-    partitionIdCnt[pidSelection_]++;
-    pidSelection_ = (pidSelection_ + 1) == numPartitions_ ? 0 : (pidSelection_ + 1);
+    std::vector<uint16_t>& row2Partition,
+    std::vector<uint32_t>& partition2RowCount) {
+  std::fill(std::begin(partition2RowCount), std::end(partition2RowCount), 0);
+  row2Partition.resize(numRows);
+
+  int32_t pidSelection = 0;
```
zuochunwei, line 30: We should use `pidSelection_` here.
You can use a local variable `int pidSelection` initialized from `pidSelection_` in the following loop, and store it into `pidSelection_` after the loop

cpp/core/shuffle/RoundRobinPartitioner.h
```diff
@@ -27,11 +27,8 @@ class RoundRobinPartitioner final : public ShuffleWriter::Partitioner {
   arrow::Status compute(
       const int32_t* pidArr,
       const int64_t numRows,
-      std::vector<uint16_t>& partitionId,
-      std::vector<uint32_t>& partitionIdCnt) override;
-
- private:
-  int32_t pidSelection_ = 0;
```
zuochunwei, line 34: don't delete it, it's used to save the status

zuochunwei:
supply some information about why we are making this change,  We made vs. benchmark in `https://quick-bench.com/`, and the result shows the speed of the improved `Partitioner::computer` is 2x faster than the old implement.

cpp/core/shuffle/RoundRobinPartitioner.cc
kecookier: ok

zuochunwei (approved):

PHILO-HE:
Just a small issue in merging code. Let's remove `[WIP]` in final commit message next time. Thanks!

PR #3074, opened 2023-09-08
[VL] Refine CompositeColumnarBatch

ulysses-you:

This pr refines the `CompositeColumnarBatch`, inlcuding:
- use  std::move to pass vector batch to avoid data copy
- use int32_t to replace long to avoid implicat conversion
- use underlying batch to get numBytes if exists

## How was this patch tested?

PASS CI

### DIFF
--- a/cpp/core/memory/ColumnarBatch.cc
+++ b/cpp/core/memory/ColumnarBatch.cc
@@ -118,7 +118,7 @@ std::shared_ptr<ColumnarBatch> CompositeColumnarBatch::create(std::vector<std::s
     }
     numColumns += batch->numColumns();
   }
-  return std::shared_ptr<ColumnarBatch>(new CompositeColumnarBatch(numColumns, numRows, batches));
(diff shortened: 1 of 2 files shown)


cpp/core/memory/ColumnarBatch.cc
```diff
@@ -118,19 +118,23 @@ std::shared_ptr<ColumnarBatch> CompositeColumnarBatch::create(std::vector<std::s
     }
     numColumns += batch->numColumns();
   }
-  return std::shared_ptr<ColumnarBatch>(new CompositeColumnarBatch(numColumns, numRows, batches));
+  return std::shared_ptr<ColumnarBatch>(new CompositeColumnarBatch(numColumns, numRows, std::move(batches)));
```
zhztheplayer, line 121: 👍 

Any performance improvement observed from your side?

ulysses-you: It seems Composite batch is only used by global sort. I did not see an obvious perfermance change.

zhztheplayer (approved):
Thanks!

Issue #1727, opened 2023-05-23
[CORE] Protocol message had too many levels of nesting


rui-mo:
**Describe the bug**
```
Caused by: com.google.protobuf.InvalidProtocolBufferException: Protocol message had too many levels of nesting.  May be malicious.  Use CodedInputStream.setRecursionLimit() to increase the depth limit.

        at com.google.protobuf.InvalidProtocolBufferException.recursionLimitExceeded(InvalidProtocolBufferException.java:148)

        at com.google.protobuf.CodedInputStream$ArrayDecoder.readMessage(CodedInputStream.java:869)

        at io.substrait.proto.Expression$FieldReference$Builder.mergeFrom(Expression.java:54368)

        at io.substrait.proto.Expression$FieldReference$Builder.mergeFrom(Expression.java:54151)

        at com.google.protobuf.CodedInputStream$ArrayDecoder.readMessage(CodedInputStream.java:873)

        at io.substrait.proto.Expression$Builder.mergeFrom(Expression.java:61881)

        at io.substrait.proto.Expression$Builder.mergeFrom(Expression.java:61567)

        at com.google.protobuf.CodedInputStream$ArrayDecoder.readMessage(CodedInputStream.java:873)

        at io.substrait.proto.FunctionArgument$Builder.mergeFrom(FunctionArgument.java:600)

        at io.substrait.proto.FunctionArgument$1.parsePartialFrom(FunctionArgument.java:1050)

        at io.substrait.proto.FunctionArgument$1.parsePartialFrom(FunctionArgument.java:1042)

        at com.google.protobuf.CodedInputStream$ArrayDecoder.readMessage(CodedInputStream.java:889)

        at io.substrait.proto.Expression$ScalarFunction$Builder.mergeFrom(Expression.java:19676)

        at io.substrait.proto.Expression$ScalarFunction$Builder.mergeFrom(Expression.java:19372)

```

**To Reproduce**

SELECT x2,x49,x18,x47,x29,x56,x60,x9,x31,x34,x53,x11

  FROM t11,t9,t47,t56,t53,t60,t34,t2,t29,t18,t31,t49

 WHERE a9=b11

   AND b53=a11

   AND a49=b9

   AND a31=3

   AND a60=b56

   AND a56=b34

   AND a29=b49

   AND a2=b18

   AND b2=a47

   AND a34=b47

   AND b29=a31

   AND b60=a53

zhouyuan:
[...]

PR #3079, opened 2023-09-08
[VL] Protobuf nesting limit error causes task failure when printing plan to stdout in debug build

zhztheplayer:
Note: This doesn't fix https://github.com/oap-project/gluten/issues/1727, just to avoid nesting limit error throwing during printing plans in debug build, to make sure the query would not fail by that.


zhztheplayer:
@rui-mo

rui-mo (approved):

loukey-lj:
#1727 Why close it if there is no fix

PR #3081, opened 2023-09-08
[GLUTEN-3077][VL] part-1: Refactor for rename native Backend to ExecutionCtx and VeloxInitializer to VeloxBackend

Yohahaha:

This is the first patch of issue 3077, only do some initial rename work to align with new concept, does not change any functionality.


Yohahaha:
> Besides, Velox has ExecCtx. We use longer name that it to distinguish with it

We have gluten namespace, it would be better to use gluten::ExecCtx for shorter naming, how do you think about it? @zhztheplayer 
btw, I use ExecutionCtx instead of ExecutionContext since ctx is a well know abbr of context.

ulysses-you:
If I understand it correctly, after this pr, the VeloxBackend will be per executor and the ExecutionCtx will be per task, right ?

Yohahaha:
> If I understand it correctly, after this pr, the VeloxBackend will be per executor and the ExecutionCtx will be per task, right ?

Correct.

ulysses-you:
thank you @Yohahaha for the work!

Yohahaha:
@zhztheplayer could you help re-run failed ci jobs? seems lose connection with celeborn server.

Yohahaha:
CI has passed, please help to review @zhztheplayer, cc @ulysses-you why may interested about it.

zhztheplayer:
> CI has passed, please help to review @zhztheplayer, cc @ulysses-you who may interested about it.

Thanks for your quick change! I'll take a look asap.

gluten-data/src/main/scala/io/glutenproject/exec/ExecutionCtx.scala
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package io.glutenproject.exec
+
+import io.glutenproject.init.BackendJniWrapper
+
+import org.apache.spark.util.TaskResource
+
+class ExecutionCtx extends TaskResource {
+  private val handle = BackendJniWrapper.createExecutionCtx()
```
ulysses-you, line 24: It seems this is a dead code, we did not have `BackendJniWrapper_createExecutionCtx` at jni side. Can you describe how/when to use this new class ? If it is just a part of future requirement, I prefer add this code when using it.

cpp/core/compute/ExecutionCtx.h
```diff
@@ -78,54 +78,40 @@ class Backend : public std::enable_shared_from_this<Backend> {
...
       const ShuffleWriterOptions& options,
-      MemoryManager* memoryManager) {
-    throw GlutenException("createShuffleWriter not implemented.");
-  }
+      MemoryManager* memoryManager) = 0;
 
-  virtual std::shared_ptr<Metrics> getMetrics(ColumnarBatchIterator* rawIter, int64_t exportNanos) {
-    return nullptr;
-  }
+  virtual std::shared_ptr<Metrics> getMetrics(ColumnarBatchIterator* rawIter, int64_t exportNanos) = 0;
```
ulysses-you, line 100: I'm fine with this change and as it already is. But I think in future, we should avoid touching unrelated code as far as possible to keep every pr minimum modification. Each pr should force on it's own topic to make easy to review, and history track.

gluten-data/src/main/scala/io/glutenproject/exec/ExecutionCtx.scala
Yohahaha: This patch does not change any functionality, and ExecutionCtx should be added in next patch as you said, thanks!

cpp/core/compute/ExecutionCtx.h
Yohahaha: +1 on your opinion, however I think this change is needed, ExecutionCtx.h should be a pure abstract class as much as possible, old codes may due to history reason that CH backend has shared these codes before.

gluten-data/src/main/scala/io/glutenproject/exec/ExecutionCtx.scala
Yohahaha: already revert.

Yohahaha:
Any more comments? this patch is as simple as it can be.

ulysses-you (approved):

ulysses-you:
@zhztheplayer I think any post review can be addressed in followups, so I'm going to merge to main

PR #3085, opened 2023-09-09
[CH] Fix memory leak

baibaichen:

We should call `ReleaseByteArrayElements`  before returning to JVM

(Fixes: \#ISSUE-ID)

## How was this patch tested?

Usigin Existed UT


loneylee:
LGTM

zzcclp (approved):
LGTM

PR #3088, opened 2023-09-11
[VL] Set a GHA workflow run to emulate random task kills/retries

zhztheplayer:
Using gluten-it `--random-kill-tasks` to emulate random task kills/retries on a complete TPC-H + TPC-DS run.

Within `--random-kill-tasks`, each task gets some chance to be killed after running for several seconds.

Temporarily, failures from this test will be ignored so doesn't block CI. We'll make it blocking after fixing all the relevant crash issues caused by task kill. See also https://github.com/oap-project/gluten/issues/3048.

### DIFF
--- a/.github/workflows/velox_be.yml
+++ b/.github/workflows/velox_be.yml
@@ -313,9 +313,9 @@ jobs:
         run: |
           docker exec centos8-test-$GITHUB_RUN_ID bash -c 'cd /opt/gluten/tools/gluten-it && \
           mvn clean install -Pspark-3.2 \
-          && GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
+          && GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries \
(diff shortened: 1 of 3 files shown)


.github/workflows/velox_be.yml
```diff
@@ -358,6 +358,13 @@ jobs:
             --local --preset=velox --benchmark-type=h --error-on-memleak --disable-aqe --off-heap-size=20g -s=1.0 --threads=16 --iterations=1 \
           && GLUTEN_IT_JVM_ARGS=-Xmx50G sbin/gluten-it.sh queries-compare \
             --local --preset=velox --benchmark-type=ds --error-on-memleak --off-heap-size=50g -s=30.0 --threads=32 --iterations=1'
+      - name: (To be fixed) TPC-H SF1.0 && TPC-DS SF30.0 Parquet local spark3.2 random kill tasks
```
winningsix, line 361: Just curious: why different SF for -H/-DS?

zhztheplayer: @rui-mo Was it because large SF for TPC-H doesn't make much difference?

rui-mo: We found different scales of TPC-DS could help uncover more possible issues, but for TPC-H, that could be not needed because it is simpler.

zhouyuan (approved):
👍 
The failures are ignored currently. it's better to put this in a nightly run.

PR #3089, opened 2023-09-11
[VL] Add config option spark.gluten.sql.columnar.backend.velox.enableSystemExceptionStacktrace

zhztheplayer:
Related to https://github.com/oap-project/gluten/issues/3048.

We observed some rare crashes in library `libunwind` when Velox is trying to handle exceptions.

After this patch, set both `spark.gluten.sql.columnar.backend.velox.enableSystemExceptionStacktrace=false` and `spark.gluten.sql.columnar.backend.velox.enableUserExceptionStacktrace=false` to workaround the crash issue to let the exceptions throw normally.

### DIFF
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -58,6 +58,10 @@ const std::string kEnableUserExceptionStacktrace =
     "spark.gluten.sql.columnar.backend.velox.enableUserExceptionStacktrace";
 const std::string kEnableUserExceptionStacktraceDefault = "true";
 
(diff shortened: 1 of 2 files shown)


zhouyuan (approved):
👍

FelixYBW:
Let's set them as false by default, since the infomation is used for developer only. From end customer usage it should be set to false to avoid the core dump issue.

zhztheplayer:
> Let's set them as false by default, since the infomation is used for developer only. From end customer usage it should be set to false to avoid the core dump issue.

I'll do that in another patch. Some similar issues needs to solve together.

zhouyuan (approved):

PR #3095, opened 2023-09-11
[GLUTEN-3077][VL] part-2: Use ExecutionCtx to manage resources' lifecycle

Yohahaha:
This is second patch of issue 3077, change list:
1. [cpp] refactor ExecutionCtx.h with three main APIs: `createXXX`, `getXXX`, `releaseXXX`, each resource's lifecycle will be managed by ExecutionCtx by these APIs.
2. [cpp] move old resources concurrent map into VeloxExecutionCtx.h, will try to replace with normal map in the following patch, and use ResourceHandle for readability.
3. [java] add ExecutionCtx class which implement TaskResource interface to align with native class, provide two methods to create it for different usage.
4. [java] bind all kinds of resources with ExecutionCtx.

note: MemoryManager has not been managed by ExecutionCtx for now.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
@@ -18,6 +18,7 @@ package org.apache.spark.sql.execution.datasources.velox
 
 import io.glutenproject.columnarbatch.ColumnarBatches
 import io.glutenproject.exception.GlutenException
+import io.glutenproject.exec.ExecutionCtxs
 import io.glutenproject.execution.datasource.GlutenRowSplitter
 import io.glutenproject.memory.arrowalloc.ArrowBufferAllocators
 import io.glutenproject.memory.nmm.NativeMemoryManagers
@@ -50,15 +51,17 @@ trait VeloxFormatWriterInjects extends GlutenFormatWriterInjectsBase {
     val arrowSchema =
       SparkArrowUtil.toArrowSchema(dataSchema, SQLConf.get.sessionLocalTimeZone)
     val cSchema = ArrowSchema.allocateNew(ArrowBufferAllocators.contextInstance())
-    var instanceId = -1L
+    var dsHandle = -1L
     val datasourceJniWrapper = new DatasourceJniWrapper()
     val allocator = ArrowBufferAllocators.contextInstance()
+    val executionCtxHandle = ExecutionCtxs.contextInstance().getHandle
     try {
       ArrowAbiUtil.exportSchema(allocator, arrowSchema, cSchema)
-      instanceId = datasourceJniWrapper.nativeInitDatasource(
+      dsHandle = datasourceJniWrapper.nativeInitDatasource(
         originPath,
         cSchema.memoryAddress(),
-        NativeMemoryManagers.contextInstance("VeloxWriter").getNativeInstanceId,
+        executionCtxHandle,
+        NativeMemoryManagers.contextInstance("VeloxWriter").getNativeInstanceHandle,
         nativeConf)
     } catch {
       case e: IOException =>
@@ -68,7 +71,13 @@ trait VeloxFormatWriterInjects extends GlutenFormatWriterInjectsBase {
     }
 
     val writeQueue =
-      new VeloxWriteQueue(instanceId, arrowSchema, allocator, datasourceJniWrapper, originPath)
+      new VeloxWriteQueue(
+        executionCtxHandle,
+        dsHandle,
+        arrowSchema,
+        allocator,
+        datasourceJniWrapper,
+        originPath)
 
     new OutputWriter {
       override def write(row: InternalRow): Unit = {
@@ -80,7 +89,7 @@ trait VeloxFormatWriterInjects extends GlutenFormatWriterInjectsBase {
 
       override def close(): Unit = {
         writeQueue.close()
-        datasourceJniWrapper.close(instanceId)
+        datasourceJniWrapper.close(executionCtxHandle, dsHandle)
       }
 
       // Do NOT add override keyword for compatibility on spark 3.1.
--- a/cpp/core/compute/ExecutionCtx.h
+++ b/cpp/core/compute/ExecutionCtx.h
@@ -29,6 +29,7 @@
 #include "shuffle/ShuffleReader.h"
 #include "shuffle/ShuffleWriter.h"
 #include "substrait/plan.pb.h"
+#include "utils/ConcurrentMap.h"
 
 namespace gluten {
 
@@ -47,11 +48,18 @@ class ExecutionCtx : public std::enable_shared_from_this<ExecutionCtx> {
   ExecutionCtx(const std::unordered_map<std::string, std::string>& confMap) : confMap_(confMap) {}
   virtual ~ExecutionCtx() = default;
 
-  virtual std::shared_ptr<ResultIterator> getResultIterator(
+  virtual ResourceHandle createResultIterator(
       MemoryManager* memoryManager,
       const std::string& spillDir,
       const std::vector<std::shared_ptr<ResultIterator>>& inputs,
       const std::unordered_map<std::string, std::string>& sessionConf) = 0;
+  virtual ResourceHandle addResultIterator(std::shared_ptr<ResultIterator>) = 0;
+  virtual std::shared_ptr<ResultIterator> getResultIterator(ResourceHandle) = 0;
+  virtual void releaseResultIterator(ResourceHandle) = 0;
+
+  virtual ResourceHandle addBatch(std::shared_ptr<ColumnarBatch>) = 0;
+  virtual std::shared_ptr<ColumnarBatch> getBatch(ResourceHandle) = 0;
+  virtual void releaseBatch(ResourceHandle) = 0;
 
   void parsePlan(const uint8_t* data, int32_t size) {
     parsePlan(data, size, {-1, -1, -1});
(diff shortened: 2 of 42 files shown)


Yohahaha:
This patch is ready for review, please help take a look, thanks! @zhztheplayer @ulysses-you

backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
```diff
@@ -80,7 +89,7 @@ trait VeloxFormatWriterInjects extends GlutenFormatWriterInjectsBase {
 
       override def close(): Unit = {
         writeQueue.close()
-        datasourceJniWrapper.close(instanceId)
+        datasourceJniWrapper.close(executionCtxHandle, dsHandle)
```
zhztheplayer, line 92: Such change makes code slightly more complicated than before (2 params vs 1 params). We'd find a better way later on.

cpp/core/compute/ExecutionCtx.h
```diff
@@ -47,11 +48,18 @@ class ExecutionCtx : public std::enable_shared_from_this<ExecutionCtx> {
...
   virtual ~ExecutionCtx() = default;
 
-  virtual std::shared_ptr<ResultIterator> getResultIterator(
+  virtual ResourceHandle createResultIterator(
       MemoryManager* memoryManager,
       const std::string& spillDir,
       const std::vector<std::shared_ptr<ResultIterator>>& inputs,
       const std::unordered_map<std::string, std::string>& sessionConf) = 0;
+  virtual ResourceHandle addResultIterator(std::shared_ptr<ResultIterator>) = 0;
+  virtual std::shared_ptr<ResultIterator> getResultIterator(ResourceHandle) = 0;
```
zhztheplayer, line 57: Should we make the `add...` member funcs private?

cpp/core/compute/ExecutionCtx.h
```diff
@@ -47,11 +48,18 @@ class ExecutionCtx : public std::enable_shared_from_this<ExecutionCtx> {
...
+  virtual void releaseResultIterator(ResourceHandle) = 0;
+
+  virtual ResourceHandle addBatch(std::shared_ptr<ColumnarBatch>) = 0;
+  virtual std::shared_ptr<ColumnarBatch> getBatch(ResourceHandle) = 0;
+  virtual void releaseBatch(ResourceHandle) = 0;
```
zhztheplayer, line 62: I think there was a way to use WeakReference / PhantomReference / `finalize` API / `Cleaner` API to implement Java-to-C++ object binding. We can see if the way can be leveraged by us in future.

cpp/core/compute/ExecutionCtx.h
```diff
@@ -178,7 +202,7 @@ struct ExecutionCtxFactoryContext {
     backendFactory = factory;
   }
 
-  std::shared_ptr<ExecutionCtx> create() {
+  ExecutionCtx* create() {
```
zhztheplayer, line 205: Would it be better to make `ExecutionCtxFactoryContext` a class as we have mutex to be hidden from callers?

cpp/core/tests/ExecutionCtxTest.cc
```diff
@@ -23,59 +23,104 @@ namespace gluten {
...
   }
   MemoryManager* createMemoryManager(
       const std::string& name,
       std::shared_ptr<MemoryAllocator> ptr,
       std::unique_ptr<AllocationListener> uniquePtr) override {
     return nullptr;
   }
-  std::shared_ptr<ColumnarToRowConverter> getColumnar2RowConverter(MemoryManager* memoryManager) override {
+  ResourceHandle addResultIterator(std::shared_ptr<ResultIterator> ptr) override {
+    return 0;
```
zhztheplayer, line 42: Does 0 have special meaning for a ResourceHandle ? If so we can create some const flags to avoid the magical uses.

cpp/velox/jni/VeloxJniWrapper.cc
```diff
@@ -84,6 +82,27 @@ JNIEXPORT void JNICALL Java_io_glutenproject_init_BackendJniWrapper_initializeBa
...
+JNIEXPORT void JNICALL Java_io_glutenproject_init_BackendJniWrapper_releaseExecutionCtx( // NOLINT
+    JNIEnv* env,
+    jclass,
+    jlong ctxHandle) {
+  JNI_METHOD_START
+  auto executionCtx = reinterpret_cast<gluten::ExecutionCtx*>(ctxHandle);
+  GLUTEN_CHECK(executionCtx != nullptr, "ExecutionCtx should not be null.");
+
+  gluten::releaseExecutionCtx(executionCtx);
+  JNI_METHOD_END()
```
zhztheplayer, line 103: Any reason putting the functions to `VeloxJniWrapper.cc` ? Can we use core module ?

gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
```diff
@@ -256,19 +262,27 @@ public static void close(long handle) {
...
   public static long compose(ColumnarBatch... batches) {
     long[] handles = Arrays.stream(batches).mapToLong(ColumnarBatches::getNativeHandle).toArray();
-    return ColumnarBatchJniWrapper.INSTANCE.compose(handles);
+    // we assume all input batches should be managed by same ExecutionCtx.
+    long executionCtxHandle =
+        Arrays.stream(batches)
+            .mapToLong(ColumnarBatches::getExecutionCtxHandle)
+            .findFirst()
+            .orElse(-1);
+    return ColumnarBatchJniWrapper.INSTANCE.compose(executionCtxHandle, handles);
```
zhztheplayer, line 271: Let's add a check for such assumptions.

zhztheplayer (commented):
Thanks!

But since so many new parameters added to functions for ExecutionCtx, a new way to bind C++ objects to Java becomes very necessary then. I think I can help to write some codes for that purpose too once I found some time.

backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
Yohahaha: Another solution is make each JniWrapper methods be member methods instead of static, so we can hide executionCtxHandle into instance, but it need lots changes for exist all JniWrapper.

cpp/core/compute/ExecutionCtx.h
Yohahaha: `addResultIterator` is needed for legacy codes, we may refactor later.

cpp/core/tests/ExecutionCtxTest.cc
Yohahaha: Just dummy implement for pass build, I'm ok for any changes here.

cpp/velox/jni/VeloxJniWrapper.cc
Yohahaha: Just follow `InitializeBackend`, I'm ok for put these functions to `core` or `velox` dir.

backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
zhztheplayer: Yes so let's keep the code as it is for now.

cpp/core/tests/ExecutionCtxTest.cc
zhztheplayer: Thanks! It was because I saw lots of `-1`s used in JniWrapper. Let's unify them all using a flag if possible.

backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
ulysses-you: can you use c++ `thread_local ` like what java did ? I think the `ExecutionCtx` should be got directly at c++ side without any handle.

Yohahaha: I dont really want to bind ExecutionCtx with thread, and we found PythonRunner will share same Spark TaskContext in different thread before.

cpp/core/compute/ExecutionCtx.h
Yohahaha: Could you explain more about that?

backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
ulysses-you: > we found PythonRunner will share same Spark TaskContext in different thread before.

It's not about this pr, just curious how can it happen.

Besides, I do not think they are the same thing. Unlike the python using a new process, the jni side is just a part of jvm evaluator, and they have to be in the same thread.

cpp/core/compute/ExecutionCtx.h
zhztheplayer: I meant to use `class` rather than `struct` on `ExecutionCtxFactoryContext`.

Yohahaha: Worth to try, let's try that in the following patch.

gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
Yohahaha: Done

cpp/core/tests/ExecutionCtxTest.cc
Yohahaha: Done

backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
Yohahaha: Indeed it happened, you can try debug with a python UDF.

Yohahaha:
Any more comments?

cpp/velox/jni/VeloxJniWrapper.cc
zhztheplayer: OK. Then let's open a new PR later on to do the clean-up.

Ideally `gluten-data` doesn't call `cpp/velox` directly. We should finally avoid that.

zhztheplayer (commented):
Thanks @Yohahaha ! Could you please address the current comments and rebase?

You can leave the functions in `VeloxJniWrappers.cc` as they are. We'd address in a later patch.

Yohahaha:
@zhztheplayer already rebased and passed CI, thanks!

zhztheplayer (approved):
Thanks @Yohahaha!

Issue #3099, opened 2023-09-11
[VL] building velox in ubuntu vm on Apple M2


XorSum:
### Problem description

When building velox in a ubuntu virtual machine that is running on Apple M2: 
```
$ uname -m
aarch64
$./build_velox.sh
......
+ '[' aarch64 == x86_64 ']'
+ '[' aarch64 == arm64 ']'
+ echo 'Unsupport arch: aarch64'
Unsupport arch: aarch64
+ exit 1
```

### System information

Velox System Info v0.0.2
Commit: da216c134470579f8e7019ab3f69802d94ea0370
CMake Version: 3.22.1
System: Linux-5.15.0-83-generic
Arch: aarch64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 11.4.0
C Compiler: /usr/bin/cc
C Compiler Version: 11.4.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

PR #3100, opened 2023-09-11
[GLUTEN-3099][VL] Support building velox on aarch64

XorSum:

Fixes: #3099 

## How was this patch tested?

manual tests

### DIFF
--- a/ep/build-velox/src/build_velox.sh
+++ b/ep/build-velox/src/build_velox.sh
@@ -122,6 +122,8 @@ function compile {
     make $COMPILE_TYPE EXTRA_CMAKE_FLAGS="${COMPILE_OPTION}"
   elif [ $ARCH == 'arm64' ]; then
     CPU_TARGET="arm64" make $COMPILE_TYPE EXTRA_CMAKE_FLAGS="${COMPILE_OPTION}"
+  elif [ $ARCH == 'aarch64' ]; then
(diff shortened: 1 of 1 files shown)


ep/build-velox/src/build_velox.sh
```diff
@@ -122,6 +122,8 @@ function compile {
     make $COMPILE_TYPE EXTRA_CMAKE_FLAGS="${COMPILE_OPTION}"
   elif [ $ARCH == 'arm64' ]; then
     CPU_TARGET="arm64" make $COMPILE_TYPE EXTRA_CMAKE_FLAGS="${COMPILE_OPTION}"
+  elif [ $ARCH == 'aarch64' ]; then
```
PHILO-HE, line 125: Nit:
It would be better to unify the handling for arm64/aarch64, i.e., use one `elif` block and let `CPU_TARGET = $ARCH`. Thanks!

PHILO-HE (dismissed):
Thanks for your PR!
Could you tell us your company name?

ep/build-velox/src/build_velox.sh
XorSum: updated

XorSum:
> Thanks for your PR! Could you tell us your company name?

I'm from [trip.com](https://www.trip.com/).

PHILO-HE (approved):
LGTM! Thanks!

PR #3101, opened 2023-09-11
[CORE][VL] Add option to limit the memory Gluten can use for each task to N = (memory / task slots)

zhztheplayer:
Add new option `spark.gluten.memory.isolation` (by default `false`):

Description of the option:

> Enable isolate memory mode. If true, Gluten controls the maximum off-heap memory can be used by each task to X, X = executor memory / max task slots. It's recommended to set true if Gluten serves concurrent queries within a single session, since not all memory Gluten allocated is guaranteed to be spillable. In the case, the feature should be enabled to avoid OOM.

The implementation inserts a complete memory management layer `TreeMemoryConsumer` between Spark's memory manager and Gluten. Once the task memory limit is hit, `TreeMemoryConsumer` will first try calling child spillers inside it's own scope without notifying Spark. After freeing some spaces, `TreeMemoryConsumer` continues to acquire memory from Spark.

User is supposed to use the feature to get rid of OOMs caused by pinned non-spillable Velox memory (https://github.com/oap-project/gluten/issues/3030) used by a historical task that is commanded to spill due to the coming of the other new tasks. This typically happens when the session is shared by a couple of concurrent queries.


gluten-core/src/main/java/io/glutenproject/memory/memtarget/IsolatedByTaskSlot.java
```diff
@@ -0,0 +1,66 @@
...
+package io.glutenproject.memory.memtarget;
+
+import io.glutenproject.proto.MemoryUsageStats;
+import org.apache.spark.memory.TaskMemoryManager;
+
+// A decorator to a task memory target, to restrict memory usage of the delegated
+//   memory target to X, X = free executor memory / task slots.
+// Using this to prevent OOMs if the delegated memory target could possibly
+//   hold large memory blocks that are not spillable.
+//   See https://github.com/oap-project/gluten/issues/3030
```
Yohahaha, line 27: How to restrict specific memory consumer's usage to X? return zero when acquireMemory or just OOM?

zhztheplayer: More designs needed here. But overall the consumer with the decorator should behave like a consumer registered to a task memory manager with a fixed limit (which is X).

Yohahaha: Great, hope following design could solve my question, when a consumer hit limit, what behavior is expected.

gluten-core/src/main/java/io/glutenproject/memory/memtarget/IsolatedByTaskSlot.java
```diff
@@ -0,0 +1,66 @@
...
+public class IsolatedByTaskSlot implements TaskMemoryTarget {
+
+  private final TaskMemoryTarget delegated;
+
+  public IsolatedByTaskSlot(TaskMemoryTarget delegated) {
```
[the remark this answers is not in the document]
zhztheplayer: The PR was not ready at that time, The class was just a placeholder for my initial thoughts of the design.

gluten-core/src/main/java/io/glutenproject/memory/memtarget/IsolatedByTaskSlot.java
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+package io.glutenproject.memory.memtarget;
+
+import io.glutenproject.proto.MemoryUsageStats;
+import org.apache.spark.memory.TaskMemoryManager;
+
+// A decorator to a task memory target, to restrict memory usage of the delegated
+//   memory target to X, X = free executor memory / task slots.
```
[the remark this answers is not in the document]
zhztheplayer: Yes. There is a calculation

```scala
def getTaskSlots(conf: SparkConf): Int = {
  val executorCores = SparkResourceUtil.getExecutorCores(conf)
  val taskCores = conf.getInt("spark.task.cpus", 1)
  executorCores / taskCores
}
```

gluten-core/src/main/java/io/glutenproject/memory/memtarget/spark/TreeMemoryConsumer.java
```diff
@@ -0,0 +1,307 @@
...
+              return -(diff > 0 ? 1 : diff < 0 ? -1 : 0); // descending
+            });
+    q.addAll(node.children().values());
+
+    long remainingBytes = bytes;
+    while (q.peek() != null && remainingBytes > 0) {
+      TreeMemoryConsumerNode head = q.remove();
+      long spilled = spillTree(head, remainingBytes);
+      remainingBytes -= spilled;
+    }
```
Yohahaha, line 132: This logic wants invoke spill from smallest consumer to largest consumer right?

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -186,10 +186,18 @@ class GlutenConfig(conf: SQLConf) extends Logging {
...
   def offHeapMemorySize: Long = conf.getConf(COLUMNAR_OFFHEAP_SIZE_IN_BYTES)
 
   def taskOffHeapMemorySize: Long = conf.getConf(COLUMNAR_TASK_OFFHEAP_SIZE_IN_BYTES)
 
+  def conservativeOffHeapMemorySize: Long =
+    conf.getConf(COLUMNAR_CONSERVATIVE_OFFHEAP_SIZE_IN_BYTES)
+
+  def conservativeTaskOffHeapMemorySize: Long =
+    conf.getConf(COLUMNAR_CONSERVATIVE_TASK_OFFHEAP_SIZE_IN_BYTES)
+
```
Yohahaha, line 200: Some questions here:
1. what's the difference for these two configs? 
2. what does `conservative` mean?
3. seems `conservativeOffHeapMemorySize` was not used?

gluten-core/src/main/java/io/glutenproject/memory/memtarget/spark/TreeMemoryConsumer.java
zhztheplayer: It's from largest to smallest.

In future we may want to follow vanilla Spark's rule, which uses the smallest one among those are larger than the target size.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
zhztheplayer: "Conservative" means the max size Gluten can consider is "safe" to use.  The two new options are set in `GlutenPlugin.scala` using the following code:

```java
// Pessimistic off-heap sizes, with the assumption that all non-borrowable storage memory
// determined by spark.memory.storageFraction was used.
val fraction = 1.0d - conf.getDouble("spark.memory.storageFraction", 0.5d)
val conservativeOffHeapSize = (offHeapSize
  * fraction).toLong
conf.set(
  GlutenConfig.GLUTEN_CONSERVATIVE_OFFHEAP_SIZE_IN_BYTES_KEY,
  conservativeOffHeapSize.toString)
val conservativeOffHeapPerTask = conservativeOffHeapSize / taskSlots
conf.set(
  GlutenConfig.GLUTEN_CONSERVATIVE_TASK_OFFHEAP_SIZE_IN_BYTES_KEY,
  conservativeOffHeapPerTask.toString)
```

The difference between the options and "non-conservative" options is that the "conservative" ones take storage memory into account. Assuming Spark had used 30% off-heap memory in storage memory pool, the memory would not be evict although a "borrow" is request from execution memory pool.

I think for stability, we may need to by default use "conservative" options since it's safter in most cases. I left the "non-conservative" options unchanged because of compatibility consideration.

zhztheplayer: Besides, I don't want users to set these "auto-generated" options. But we don't develop a general way to guard against that yet.

zhztheplayer: > seems conservativeOffHeapMemorySize was not used?

Yes. But it's worth to keep it for future use.

Yohahaha: Thanks for detailed explanation! I have realized the "conservative" meaning.
 
> we may need to by default use "conservative" options since it's safter in most cases.

Please keep a switch config for that, Spark already has `maybeGrowExecutionPool` to shrink storagePool and executionPool, I prefer let Spark control this logic. If use "conservative" option by default, we may not fully utilize the unused storage memory.

zhztheplayer: I meant to use "conservative" options for just Gluten's related codes that requires reading the off-heap size. For example shuffle writer and partial agg. It's not a goal to touch vanilla Spark's memory management.

zhztheplayer: Worth noting that Spark's "storage region size" determined by `spark.memory.storageFraction` is not evictable if in use. That's why we add the conservative options.

zhouyuan (approved):
👍 
Another big change on memory component!

gluten-core/src/main/java/io/glutenproject/memory/memtarget/spark/TreeMemoryConsumer.java
Yohahaha: Still not fully understanding...

We sort children in descending order and invoke spillTree on peek element's children recursively, which means we pick largest consumer and then pass smaller consumer into spillTree, when node has no children, we spill it and return.
I think spillTree loop from largest to smallest, but spill from smallest to largest, please correct me if I'm wrong, thanks!

Another misunderstanding logic, does the sort of root not sort its all children? seems  we sort node's children in every spillTree invoked.

zhztheplayer: Hi, thanks for helping checking this code. I haven't verified it carefully but let's take a simple example:
```
a 200 (self 80)
|-b 70
|-c 50 (self 10)
   |- d 30
   |- e 10
```

With the code we implemented post-order traversal on this tree (children-first, self-last), which means the visiting order is supposed to be

```
b (70) -> d (30) -> e (10) -> c (10) -> a (80)
```

Which seems to be aligned with my initial assumption: **largest to smallest, but self last**.
Is that the same with your thoughts?

> Another misunderstanding logic, does the sort of root not sort its all children? seems we sort node's children in every spillTree invoked.

Yes the data structure is not efficient since we only sorted the children of the current node (See code `q.addAll(node.children().values());`). So probably we can move to TreeMap/TreeSet later on.

Yohahaha: > Which seems to be aligned with my initial assumption: largest to smallest, but self last.
Is that the same with your thoughts?

Thanks, it's same.

PR #3117, opened 2023-09-12
[VL] Follow-up work for CI

PHILO-HE:


rui-mo (approved):

tools/gluten-te/ubuntu/dockerfile-buildenv
```diff
@@ -102,10 +102,14 @@ RUN set-login-env "LANG=en_US.UTF-8"
 RUN set-login-env "LANGUAGE=en_US:en"
 RUN set-login-env "LC_ALL=en_US.UTF-8"
 
-# Install Spark 3.2.2
+# Build & install Spark 3.2.2
```
winningsix, line 105: Do we want to pass-in Spark version as a docker build arg in  ```buildenv.sh```?

PHILO-HE: As spark 3.2, 3.3 CI job shared a same docker image, we may just keep it here. Thanks!

PR #3128, opened 2023-09-12
[CORE] Unify the logic of switching left/right keys for SHJ

waitinfuture:

This PR unifies the logic in `ShuffledHashJoinExecTransformer#exchangeTable` and
`ShuffledHashJoinExecTransformer#substraitJoinType`, because these two methods
follows the same switching table logic.Unifying these two methods makes the code less error prone.

Also renames `exchangeTable` to `needSwitchChildren`.

## How was this patch tested?
Passes GA.


backends-velox/src/main/scala/io/glutenproject/execution/ShuffledHashJoinExecTransformer.scala
```diff
@@ -87,7 +87,7 @@ case class ShuffledHashJoinExecTransformer(
    * Returns whether the build and stream table should be exchanged with consideration of build
    * type, planned build side and the preferred build side.
    */
-  override lazy val exchangeTable: Boolean = hashJoinType match {
+  override lazy val switchTable: Boolean = hashJoinType match {
```
Yohahaha, line 90: nit: `switchTable` -> `needSwitchTable`.

And switchTable confused me, maybe switchChild make more sense for me, join children does not always be table.

waitinfuture: Updated, thanks.

waitinfuture:
cc @rui-mo could you please take a look at this PR?

rui-mo (approved):

PHILO-HE (approved):
LGTM! Thanks!
cc @zzcclp  please note this change.

PR #3152, opened 2023-09-14
[VL] Support columnar table cache

ulysses-you:

This pr supports `dataframe.cache`.

Implement `ColumnarCachedBatchSerializer` as the Spark interface `CachedBatchSerializer`
 
 There is a issue that, if we allocate a small onheap memory, then it may be cause the
 cached block spill to disk frequently.
 
 Feature:
 1. This serializer supports column pruning without deserialize the `Array[Byte]`
 2. TODO: support push down filter
 3. Super TODO: cache offheap object directly if Spark framework support in the future
 
 The data transformation pipeline:
 
   - Serializer ColumnarBatch -> CachedColumnarBatch
     -> serialize to byte[]
 
   - Deserializer CachedColumnarBatch -> ColumnarBatch
     -> deserialize to byte[] to create Velox ColumnarBatch
 
   - Serializer InternalRow -> CachedColumnarBatch (support RowToColumnar)
     -> Convert InternalRow to ColumnarBatch
     -> Serializer ColumnarBatch -> CachedColumnarBatch
 
   - Serializer InternalRow -> DefaultCachedBatch (unsupport RowToColumnar)
     -> Convert InternalRow to DefaultCachedBatch using vanilla Spark serializer
 
   - Deserializer CachedColumnarBatch -> InternalRow (support ColumnarToRow)
     -> Deserializer CachedColumnarBatch -> ColumnarBatch
     -> Convert ColumnarBatch to InternalRow
 
   - Deserializer DefaultCachedBatch -> InternalRow (unsupport ColumnarToRow)
     -> Convert DefaultCachedBatch to InternalRow using vanilla Spark serializer

## How was this patch tested?

Add tests in `VeloxColumnarCacheSuite` and pass all tests in `GlutenCachedTableSuite`

For example:
`spark.table("t").cache.collect`

before
image

after
image


backends-velox/src/main/scala/io/glutenproject/execution/RowToVeloxColumnarExec.scala
```diff
@@ -58,132 +60,150 @@ case class RowToVeloxColumnarExec(child: SparkPlan) extends RowToColumnarExecBas
     val localSchema = schema
     child.execute().mapPartitions {
       rowIterator =>
-        if (rowIterator.isEmpty) {
```
ulysses-you, line 61: move the code to `RowToVeloxColumnarExec.toColumnarBatchIterator ` so that columnar serializer can reuse it

backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarToRowExec.scala
```diff
@@ -70,118 +69,106 @@ case class VeloxColumnarToRowExec(child: SparkPlan) extends ColumnarToRowExecBas
...
 
-class ColumnarToRowRDD(
-    @transient sc: SparkContext,
-    rdd: RDD[ColumnarBatch],
-    output: Seq[Attribute],
-    numOutputRows: SQLMetric,
-    numInputBatches: SQLMetric,
-    convertTime: SQLMetric)
-  extends RDD[InternalRow](sc, Seq(new OneToOneDependency(rdd))) {
+object VeloxColumnarToRowExec {
```
ulysses-you, line 88: move code to `VeloxColumnarToRowExec.toRowIterator` so that columnar serializer can reuse it

gluten-core/src/main/scala/io/glutenproject/execution/ColumnarInMemoryTableScanExec.scala
ulysses-you, line 1: We do not need to transform `InMemoryTableScanExec` as it already supports both row-based and columnar. We just need to implement the serializer interface.

gluten-core/src/main/scala/io/glutenproject/GlutenPlugin.scala
```diff
@@ -137,15 +137,26 @@ private[glutenproject] class GlutenDriverPlugin extends DriverPlugin with Loggin
...
       s"$GLUTEN_SESSION_EXTENSION_NAME"
     }
     conf.set(SPARK_SESSION_EXTS_KEY, String.format("%s", extensions))
 
+    // sql table cache serializer
+    if (conf.getBoolean(GlutenConfig.COLUMNAR_TABLE_CACHE_ENABLED.key, true)) {
+      if (BackendsApiManager.isVeloxBackend) {
+        conf.set(
+          StaticSQLConf.SPARK_CACHE_SERIALIZER.key,
+          "io.glutenproject.backendsapi.velox.ColumnarCachedBatchSerializer")
```
ulysses-you, line 154: it's the entrance

ulysses-you:
cc @zhztheplayer @rui-mo @PHILO-HE @Yohahaha @zhouyuan  It's ready to review, thank you

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
```diff
@@ -0,0 +1,303 @@
...
+
+import org.apache.arrow.vector.VectorSchemaRoot
+
+case class CachedColumnarBatch(
+    override val numRows: Int,
+    override val sizeInBytes: Long,
+    fieldBuffersArr: Array[FieldBuffers],
+    fieldNullCountsArr: Array[FiledNullCounts],
+    fieldNumRowsArr: Array[FieldNumRows])
+  extends CachedBatch {}
```
zhztheplayer, line 50: Thanks for working on this!

A question along with the PR description

> Spark does not support cache a offheap object, and it converts UnsafeRow to
Array[Byte] then cache it. So this serializer also use that way, by converting velox ColumnarBatch to
Array[Byte] then cache it.

Since Spark's developer API `CachedBatch` doesn't enforce a format for the underlying data, was there any special reason Array[Byte] is still used in the PR?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
```diff
@@ -0,0 +1,303 @@
...
+              CachedColumnarBatch(
+                numRows,
+                sizeInBytes,
+                fieldRecordBatch.fieldBuffersArr,
+                fieldRecordBatch.fieldNullCountsArr,
+                fieldRecordBatch.fieldNumRowsArr
+              )
+            } finally {
+              vsr.close()
+            }
```
zhztheplayer, line 232: When will the input batch get closed?

gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
```diff
@@ -130,6 +134,39 @@ public static ColumnarBatch ensureLoaded(BufferAllocator allocator, ColumnarBatc
...
+          handle, cSchema.memoryAddress(), cArray.memoryAddress());
+      VectorSchemaRoot vsr = VectorSchemaRoot.create(arrowSchema, allocator);
+      ArrowAbiUtil.importToVectorSchemaRoot(allocator, vsr, cArray);
+      for (long i = 0; i < iv.refCnt(); i++) {
+        input.close();
+      }
+      return vsr;
+    }
+  }
+
```
zhztheplayer, line 169: I assume the purpose of adding the new APIs is to implement a zero-copy way to convert Arrow data from/to Array[Byte], is that correct?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
ulysses-you: yeah, there is no constraint, but it seems Array[Byte] is a good option. Say, arrow based buf is offheap and we need a onheap buf.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
ulysses-you: inside `ColumnarBatches.loadArrowVectorSchemaRoot`, I have added comment for that method

gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
ulysses-you: yeah, I read the related code about Spark Arrow and Gluten. I think it's the shortest path to do the data transform...

gluten-data/src/main/java/io/glutenproject/vectorized/VeloxColumnarBatchBuilderJniWrapper.java
```diff
@@ -0,0 +1,26 @@
...
+package io.glutenproject.vectorized;
+
+import io.glutenproject.init.JniInitialized;
+
+public class VeloxColumnarBatchBuilderJniWrapper extends JniInitialized {
+  public static VeloxColumnarBatchBuilderJniWrapper INSTANCE =
+      new VeloxColumnarBatchBuilderJniWrapper();
+
+  public native long createVeloxColumnarBatch(long memoryManagerHandle, long cSchema, long cArray);
+}
```
zhztheplayer, line 26: We had an implicit convention to avoid Velox-related code from appearing in `gluten/data`

There was an issue related https://github.com/oap-project/gluten/issues/1539

Regarding this code, IIUC we do `byte-array -> arrow -> velox` conversion, is it possible to just pass a native Arrow batch (created using `ColumnarBatchJniWrapper.INSTANCE.createWithArrowArray`) to native, and let Velox backend do lazy loading (which means the loading procedure from native Arrow batch to native Velox batch)?

gluten-data/src/main/scala/io/glutenproject/utils/ArrowUtil.scala
```diff
@@ -86,4 +87,23 @@ object ArrowUtil extends Logging {
...
+  def toByteArray(buffer: ArrowBuf): Array[Byte] = {
+    val readableBytes = buffer.readableBytes()
+    if (readableBytes > Integer.MAX_VALUE) {
+      throw new GlutenException(
+        s"arrow buf readableBytes: $readableBytes is bigger than Integer.MAX_VALUE")
+    }
+    val buf = buffer.nioBuffer
+    val bytes: Array[Byte] = new Array[Byte](buf.remaining())
+    buf.get(bytes)
+    bytes
```
zhztheplayer, line 101: OK so it does conduct copy. This was intentional, right?

Copying also making sense to me as long as it's a intended design in the early version of the feature.

gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
zhztheplayer: OK but it's probably not possible to do this without copy. So copy is here and it's OK to me.

And I realized we once added `ColumnarBatchSerializerJniWrapper` as a Java-side serde for columnar batch. Would you think that will be helpful here, since we already can't avoid copies?

gluten-data/src/main/java/io/glutenproject/vectorized/VeloxColumnarBatchBuilderJniWrapper.java
ulysses-you: >  is it possible to just pass a native Arrow batch (created using ColumnarBatchJniWrapper.INSTANCE.createWithArrowArray) to native, and let Velox backend do lazy loading

The first version, that's how I did it. However, I can not find a good way to let downstream operator to do the lazy conversion(from arrow to velox ColumnarBatch). So, I create a new Jni method and convert it to velox ColumnarBatch directly.

Do you have any idea about it?

gluten-data/src/main/scala/io/glutenproject/utils/ArrowUtil.scala
ulysses-you: yes, I think it's necessary as the Arrow always use direct(offheap) buffer but we need a onheap buffer.

gluten-data/src/main/java/io/glutenproject/vectorized/VeloxColumnarBatchBuilderJniWrapper.java
zhztheplayer: Ideally it should be done implicitly in C++ Velox backend by using existing API `VeloxColumnarBatch::from`. If it's not working we may need to change the C++ backend codes to support that somehow, probably in another PR separately.

ulysses-you:
I hate rebase ..

ulysses-you:
@zhztheplayer I think I addressed the comments ? please point me if there are some remainings. thank you

ulysses-you:
Run Gluten Clickhouse CI

zhztheplayer:
> @zhztheplayer I think I addressed the comments ? please point me if there are some remainings. thank you

Thanks! Overall great besides I can help adding some minor changes on the PR, if it's OK to you?

ulysses-you:
thank you @zhztheplayer , that's ok,  I'm glad to see any imprvement

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
```diff
@@ -0,0 +1,311 @@
...
+      input: RDD[ColumnarBatch],
+      schema: Seq[Attribute],
+      storageLevel: StorageLevel,
+      conf: SQLConf): RDD[CachedBatch] = {
+    val timezoneId = SQLConf.get.sessionLocalTimeZone
+    val localSchema = toStructType(schema)
+    input.mapPartitions {
+      it =>
+        val executionCtxHandle = ExecutionCtxs.contextInstance().getHandle
+        val allocator = ArrowBufferAllocators.contextInstance()
```
liujiayi771, line 212: It will use ArrowBufferAllocator to allocate CachedBatch memory, so it will use offheap memory managed by gluten memory manager when doing df.cache. Is my understanding correct?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
```diff
@@ -0,0 +1,311 @@
...
+ * 3. Super TODO: cache offheap object directly if Spark framework support in the future
+ *
+ * The data transformation pipeline:
+ *
+ *   - Serializer ColumnarBatch -> CachedColumnarBatch
+ *     -> Export to Arrow VectorSchemaRoot
+ *     -> Unload to CachedColumnarBatch (onheap)
+ *
+ *   - Deserializer CachedColumnarBatch -> ColumnarBatch
+ *     -> Load to Arrow VectorSchemaRoot
```
JkSelf, line 75: Since we are planning to remove the Arrow dependency in Gluten, can we use Velox's [deserialization interface](https://github.com/oap-project/velox/blob/175870db403ba9159d588fc772a5c936c27d0910/velox/row/UnsafeRowDeserializers.h#L903) to construct a RowVector instead of relying on the Arrow VectorSchemaRoot to call Arrow API? @ulysses-you @zhztheplayer @zhouyuan

ulysses-you: I agree with you, but it's not the trivial things, this pr also uses the row to columnar which depends on arrow.. I think we can improve these plan step by step in future prs. The first thing is to make row to columnar decouple with arrow.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
ulysses-you: No, the arrow buf is an intermediate thing, we finally will store array[byte] as rdd block cache which is onheap.

zhztheplayer:
I added a commit @ulysses-you , but haven't have chance to test it yet. Would you like to continue the work following the direction?

Also please review @JkSelf

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
zhztheplayer: > I agree with you, but it's not the trivial things, this pr also uses the row to columnar which depends on arrow.. I think we can improve these plan step by step in future prs. The first thing is to make row to columnar decouple with arrow.

Let's try eliminating some of the Arrow usages in this patch. I just helped adding some code @ulysses-you

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
```diff
@@ -0,0 +1,303 @@
...
+              batch
+            )
+            val projected = new ColumnarBatch(requestedColumnIndices.map {
+              ordinal =>
+                heavy.column(ordinal)
+            }.toArray, heavy.numRows())
+            ColumnarBatches.retain(projected)
+            ColumnarBatches.release(heavy)
+
+            ColumnarBatches.ensureOffloaded(
```
ulysses-you, line 288: @zhztheplayer  The one more load and offload operator seems worse than before...  we should add a new jni method, something like 
```
deserialize(
    long execHandle,
    long deserializer,
    byte[] cachedBatch.encoded,
    long selectedCSchema
   )
```

it is used to deserialize the columnar batch and only return a offload columnar batch with the selected attribute.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
ulysses-you: After second thought, I think it's too hurried to decide to remove the Arrow dependency in Gluten. How to export/import schema ? How to manage the onheap memory ? If we can not clean up all the usage of Arrow in a short time, I think it's overkill to remove the usage of Arrow.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
ulysses-you: BTW, we should return a velox columnar batch but `ensureOffloaded` returns a arrow based.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
zhouyuan: agree, the key idea is to remove arrow from the I/O flow firstly, e.g., using velox native column vectors, thus the code logic will be simpler. the schema/metadata part indeed needs more refactor work - should be done in another epic 

thanks, -yuan

FelixYBW:
@zhouyuan 
In Gazelle, we cache the offheap object directly, No serialization or copy to heap. We can use the Velox spill to evict the data.

zhouyuan:
@FelixYBW correct, at Gazelle we introduced a class `ArrowWritableColumnVector` which allows to store vectors on off-heap, but again this will bring arrow depdencies on I/O code path. Hongze and Xiduo are discussing if we could setup another patch to simplify it with Spark's built-in api or sth similar
Here's the previous impl:
https://github.com/oap-project/gazelle_plugin/pull/246/files

thanks, -yuan

ulysses-you:
> In Gazelle, we cache the offheap object directly, No serialization or copy to heap. We can use the Velox spill to evict the data

IIUC, Spark block manager does not allow cache offheap object directly. There are some issues in Spark, e.g., How to estimate the offheap size ? Spark does not know the real offheap size usage in storage memory pool. How to spill offheap to disk when Spark want to shrink the storage memory pool ? Spark does not provide such interface. I'm not sure how Gazelle overcome this issue.

So the original idea of this pr is that, copy memory from offheap to onheap and then cache it. The way is similar with vanilla Spark unsaferow.

FelixYBW:
> @FelixYBW correct, at Gazelle we introduced a class `ArrowWritableColumnVector` which allows to store vectors on off-heap, but again this will bring arrow depdencies on I/O code path. Hongze and Xiduo are discussing if we could setup another patch to simplify it with Spark's built-in api or sth similar Here's the previous impl: https://github.com/oap-project/gazelle_plugin/pull/246/files
> 
> thanks, -yuan

Can we store the rowvector directly in offheap? spill it or call Velox's columnar2row to convert it if need?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
zhztheplayer: > @zhztheplayer The one more load and offload operator seems worse than before... we should add a new jni method, something like
> 
> ```
> deserialize(
>     long execHandle,
>     long deserializer,
>     byte[] cachedBatch.encoded,
>     long selectedCSchema
>    )
> ```
> 
> it is used to deserialize the columnar batch and only return a offload columnar batch with the selected attribute.

If we do this, we can implement filter-on-read to avoid unnecessary deserializations on the non-requested columns.

It it's tricky more or less, for the first version I recommend adding an abstract function to C++ `ColumnarBatch` API to do a simple ordinal-based projection which is very reasonable to any in memory columnar format. This way we can avoid the Arrow conversion either.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ColumnarCachedBatchSerializer.scala
```diff
@@ -0,0 +1,303 @@
...
+              Array(ColumnarBatches.getNativeHandle(batch)),
+              NativeMemoryManagers
+                .contextInstance("ColumnarCachedBatchSerializer")
+                .getNativeInstanceHandle
+            )
+            CachedColumnarBatch(
+              Math.toIntExact(serializeResult.getNumRows),
+              serializeResult.getSerialized.length,
+              serializeResult.getSerialized)
+          }
```
zhztheplayer, line 222:
```suggestion
CachedColumnarBatch(Math.toIntExact(serializeResult.getNumRows),
              serializeResult.getSerialized.length,
              Array(serialized col 0, serialized col 1, serialized col 2..))
```

Regarding serialized col x, it's serialized from a columnar batch that only includes a single column (with ordinal x) from the original columnar batch:
```
for (x <- (0 until batch.columns)) {
  batch for col x = ColumnarBatches.select(batch, Array(x)); // since only a "x" in input ordinal array, returned batch only contains one single column.
  serialized col x = ColumnarBatchSerializerJniWrapper.INSTANCE.serialize(ctx, batch for col x, mm);
}
```

zhztheplayer (commented):
@ulysses-you This way we can avoid deserializing the unwanted columns either.

zhztheplayer:
I went through Spark 3.2.2's code for the cached batch size part

1. 

> > @FelixYBW correct, at Gazelle we introduced a class `ArrowWritableColumnVector` which allows to store vectors on off-heap, but again this will bring arrow depdencies on I/O code path. Hongze and Xiduo are discussing if we could setup another patch to simplify it with Spark's built-in api or sth similar Here's the previous impl: https://github.com/oap-project/gazelle_plugin/pull/246/files
> > thanks, -yuan
> 
> Can we store the rowvector directly in offheap? spill it or call Velox's columnar2row to convert it if need?

It turns out to me that there were a couple of APIs which probably be useful to achieve the goal caching off-heap objects directly.

1. Spark's `KnownSizeEstimation`:
  To allows us to override the cached batch size with actual off-heap data size Spark uses to add to storage memory pool
2. Java's `Serializable`:
  To allow Spark's serializer serialize off-heap data correctly then evict to disk, in our customized way

@ulysses-you Do you think it's the right direction to improve this part of code in future?

ulysses-you:
thank you @zhztheplayer yes, we should make the cached batch implement `KnownSizeEstimation` and `Externalizable`. Besides, the default storage level always uses the onheap memory mode, we shuold replace it to offheap during plan transform. It really reminders me that we might be able to save the data to offheap directly, so I'm going to try it.

ulysses-you:
After some attempts, I find it's more complex than thought. The lifecycle of columnar batch will be changed if we cache it directly. Currently, the columnar batch will be removed after task completion. If we cache the columnar batch, then we can not remove it because it may be read by other tasks, however it will break Velox sandbox check that the memory is not fully released when closing memory pool.

So, for this pr, I plan to support cache columnar batch in a onheap way. We can do further optimization it in future prs.

zhztheplayer (approved):
Thanks @ulysses-you !

ulysses-you:
thank you all!

PR #3155, opened 2023-09-14
[GLUTEN-3154][CORE] Override doCanonicalize in scan operators

liujiayi771:
Fix https://github.com/oap-project/gluten/issues/3154

Spark will try to get reused queryStage based on plan.canonicalized.
https://github.com/apache/spark/blob/822f58f0d26b7d760469151a65eaf9ee863a07a1/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L523-L523

## How was this patch tested?

image
image

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
@@ -24,7 +24,6 @@ import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.expressions._
-import org.apache.spark.sql.catalyst.plans.QueryPlan
 import org.apache.spark.sql.connector.read.{InputPartition, Scan}
 import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.execution.datasources.v2.{BatchScanExecShim, FileScan}
@@ -37,8 +36,7 @@ import java.util.Objects
 class BatchScanExecTransformer(
     output: Seq[AttributeReference],
     @transient scan: Scan,
-    runtimeFilters: Seq[Expression],
-    pushdownFilters: Seq[Expression] = Seq())
+    runtimeFilters: Seq[Expression])
   extends BatchScanExecShim(output, scan, runtimeFilters)
   with BasicScanExecTransformer {
 
@@ -48,7 +46,7 @@ class BatchScanExecTransformer(
 
   override def filterExprs(): Seq[Expression] = scan match {
     case fileScan: FileScan =>
-      fileScan.dataFilters ++ pushdownFilters
+      fileScan.dataFilters
     case _ =>
       throw new UnsupportedOperationException(s"${scan.getClass.toString} is not supported")
   }
@@ -93,12 +91,11 @@ class BatchScanExecTransformer(
 
   override def equals(other: Any): Boolean = other match {
     case that: BatchScanExecTransformer =>
-      that.canEqual(this) && super.equals(that) &&
-      this.pushdownFilters == that.getPushdownFilters
+      that.canEqual(this) && super.equals(that)
     case _ => false
   }
 
-  override def hashCode(): Int = Objects.hash(batch, runtimeFilters, pushdownFilters)
+  override def hashCode(): Int = Objects.hash(batch, runtimeFilters)
 
   override def canEqual(other: Any): Boolean = other.isInstanceOf[BatchScanExecTransformer]
 
@@ -120,10 +117,6 @@ class BatchScanExecTransformer(
   @transient protected lazy val filteredFlattenPartitions: Seq[InputPartition] =
     filteredPartitions.flatten
 
-  private def getPushdownFilters: Seq[Expression] = {
-    pushdownFilters
-  }
-
   @transient override lazy val fileFormat: ReadFileFormat = scan.getClass.getSimpleName match {
     case "OrcScan" => ReadFileFormat.OrcReadFormat
     case "ParquetScan" => ReadFileFormat.ParquetReadFormat
@@ -137,10 +130,7 @@ class BatchScanExecTransformer(
     new BatchScanExecTransformer(
       canonicalized.output,
       canonicalized.scan,
-      canonicalized.runtimeFilters,
-      QueryPlan.normalizePredicates(
-        pushdownFilters.filterNot(_ == DynamicPruningExpression(Literal.TrueLiteral)),
-        output)
+      canonicalized.runtimeFilters
     )
   }
 }
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
@@ -24,6 +24,7 @@ import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.catalyst.plans.QueryPlan
 import org.apache.spark.sql.connector.read.{InputPartition, Scan}
 import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.execution.datasources.v2.{BatchScanExecShim, FileScan}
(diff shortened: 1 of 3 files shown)


PHILO-HE:
Thanks for your fix! Does `FileSourceScanExecTransformer` have the same issue?

kerwin-zk:
> Thanks for your fix! Does `FileSourceScanExecTransformer` have the same issue?

@PHILO-HE  It seems that both HiveTableScanExecTransformer and FileSourceScanExecTransformer may have this issue.

liujiayi771:
> > Thanks for your fix! Does `FileSourceScanExecTransformer` have the same issue?
> 
> @PHILO-HE It seems that both HiveTableScanExecTransformer and FileSourceScanExecTransformer may have this issue.

Yes, I have added doCanonicalize in FileSourceScanExecTransformer and HiveTableScanExecTransformer.

PHILO-HE:
@ulysses-you, @zzcclp, could you take a review?

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
```diff
@@ -130,4 +131,16 @@ class BatchScanExecTransformer(
...
   }
+
+  override def doCanonicalize(): BatchScanExecTransformer = {
+    val canonicalized = super.doCanonicalize()
+    new BatchScanExecTransformer(
+      canonicalized.output,
+      canonicalized.scan,
+      canonicalized.runtimeFilters,
+      QueryPlan.normalizePredicates(
+        pushdownFilters.filterNot(_ == DynamicPruningExpression(Literal.TrueLiteral)),
```
ulysses-you, line 142: not related to this pr, but it seems `pushdownFilters` is never been set and used... not sure the original idea, maybe we can just remove `pushdownFilters`

liujiayi771: Yes, I also noticed this when support iceberg read. @PHILO-HE @rui-mo  Can we remove it?

PHILO-HE: Yes, we can remove it in this PR. Thanks!

zzcclp (commented):
LGTM

ulysses-you (approved):

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
```diff
@@ -119,15 +117,20 @@ class BatchScanExecTransformer(
...
   @transient override lazy val fileFormat: ReadFileFormat = scan.getClass.getSimpleName match {
     case "OrcScan" => ReadFileFormat.OrcReadFormat
     case "ParquetScan" => ReadFileFormat.ParquetReadFormat
     case "DwrfScan" => ReadFileFormat.DwrfReadFormat
     case "ClickHouseScan" => ReadFileFormat.MergeTreeReadFormat
     case _ => ReadFileFormat.UnknownFormat
   }
+
+  override def doCanonicalize(): BatchScanExecTransformer = {
+    val canonicalized = super.doCanonicalize()
```
PHILO-HE, line 129: Just wondering whether we can directly use the members of `BatchScanExecTransformer` to doCanonicalize. There may have some differences. Thanks!

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
liujiayi771: @ulysses-you @PHILO-HE I encountered an error during the test. This parameter here cannot be deleted. Although it is not used, it is possible that the parameter was added unintentionally that avoid the error. 
```
Exception in thread "main" java.lang.IllegalStateException:
Failed to copy node.
Is otherCopyArgs specified correctly for BatchScanExecTransformer.
Exception message: wrong number of arguments
ctor: public io.glutenproject.execution.BatchScanExecTransformer(scala.collection.Seq,org.apache.spark.sql.connector.read.Scan,scala.collection.Seq)?
types: class scala.collection.immutable.$colon$colon, class org.apache.iceberg.spark.source.SparkBatchQueryScan, class scala.collection.immutable.$colon$colon, class scala.None$
args: List(application_id#10037, executor_id#10046, start_time#10047L, end_time#10048L, dt#10079, emr_cluster_id#10080, emr_user_id#10081), IcebergScan(table=default_iceberg.doctor_oss.spark_task_info, type=struct<1: application_id: optional string, 10: executor_id: optional string, 11: start_time: optional long, 12: end_time: optional long, 43: dt: optional string, 44: emr_cluster_id: optional string, 45: emr_user_id: optional string>, filters=[not_null(ref(name="dt")), not_null(ref(name="emr_user_id")), ref(name="dt") == "2023-08-01", not(ref(name="emr_user_id") == "1836697956323238"), not_null(ref(name="emr_cluster_id")), not_null(ref(name="application_id"))], runtimeFilters=[], caseSensitive=false), List(dynamicpruningexpression(true), dynamicpruningexpression(emr_cluster_id#10080 IN dynamicpruning#10622)), None

	at org.apache.spark.sql.catalyst.trees.TreeNode.makeCopy(TreeNode.scala:854)
	at org.apache.spark.sql.catalyst.trees.TreeNode.makeCopy(TreeNode.scala:797)
	at org.apache.spark.sql.execution.SparkPlan.super$makeCopy(SparkPlan.scala:100)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$makeCopy$1(SparkPlan.scala:100)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:779)
	at org.apache.spark.sql.execution.SparkPlan.makeCopy(SparkPlan.scala:100)
	at org.apache.spark.sql.execution.SparkPlan.makeCopy(SparkPlan.scala:60)
	at org.apache.spark.sql.catalyst.plans.QueryPlan.mapExpressions(QueryPlan.scala:223)
	at org.apache.spark.sql.catalyst.plans.QueryPlan.transformExpressionsUpWithPruning(QueryPlan.scala:188)
	at org.apache.spark.sql.execution.reuse.ReuseExchangeAndSubquery$$anonfun$org$apache$spark$sql$execution$reuse$ReuseExchangeAndSubquery$$reuse$1$1.applyOrElse(ReuseExchangeAndSubquery.scala:54)
	at org.apache.spark.sql.execution.reuse.ReuseExchangeAndSubquery$$anonfun$org$apache$spark$sql$execution$reuse$ReuseExchangeAndSubquery$$reuse$1$1.applyOrElse(ReuseExchangeAndSubquery.scala:44)
```

It may be a little difficult to explain here. I try to make it clear. When Spark is in TreeNode.makeCopy, it will call the productArity method of Product to obtain the number of parameters of the case class. The current number of parameters of BatchScanExecTransformer is 3, but productArity gets 4. After my test, productArity gets the number of parameters of the first specific case class, which is the number of parameters of BatchScanExec, which does not match the number of BatchScanExecTransformer constructors. 

I think the standard approach here is that if we need to inherit Spark's case class plan, we need to ensure that the number of parameters is consistent, otherwise it will cause problems when plan.transform calls makeCopy. Or implement the otherCopyArgs method to ensure that not transformed args can be copied.

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
liujiayi771: BatchScanExecTransformer inherits BatchScanExec, and their parameters should be exactly the same.

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
PHILO-HE: It's a subtle issue. Why `productArity` returns 4 for `BatchScanExec`? It looks `BatchScanExec` has only 3 constructor args.

liujiayi771: > It's a subtle issue. Why `productArity` returns 4 for `BatchScanExec`? It looks `BatchScanExec` has only 3 constructor args.

Spark 3.3 has 4 args.
https://github.com/apache/spark/blob/b170a670b68616af49b26cbcc84894c939713837/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/BatchScanExec.scala#L40

liujiayi771:
```scala
case class A(a1: Int, a2: Int) extends Product
class B(a1: Int, a2: Int, a3: Int) extends A(a1, a2)
val b = new B(1, 2, 3)
b.productArity
```
b.productArity will return 2.

PHILO-HE: I see. Please help fix it in a separate PR and leave some comments to explain in source code. Thanks!

PR #3159, opened 2023-09-15
[VL] Add memory pool to track shuffle cached memory usage

marin-ma:
After this patch, VeloxShuffleWriter has three memory pool.
- `options_.memory_pool`
- `partitionBufferPool_` :  Used to track memory usage of split buffers. The actual allocation is delegated to `options_.memory_pool`.
- `payloadPool_` : Used to track memory allocation of Arrow IPC payloads. The actual allocation is delegated to `options_.memory_pool`.

We used maintain `options_.ipc_memory_pool`, which can be assigned as `LargeMemoryPool` if needed. `LargeMemoryPool` was introduced to alleviate the occurrence of page fault, and it was only used for allocating Arrow IPC payload. But since the creating IPC payload part (compressing buffers, etc) has been refactored many times, we haven't done performance testing around the latest implementation, so the `LargeMemoryPool` will be removed as well.

depends on #2982


shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -774,6 +776,12 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val COLUMNAR_SHUFFLE_REALLOC_THRESHOLD =
+    buildConf("spark.gluten.sql.columnar.shuffle.realloc.threshold")
```
Yohahaha, line 780: nit: would you like copy docs from Configuration.md?

marin-ma: @Yohahaha This PR depends on #2982 Please check this commit only https://github.com/oap-project/gluten/pull/3159/commits/feb200942a9f505bbf7394a4a1117040fa3748f4

cpp/core/shuffle/ShuffleWriter.h
```diff
@@ -288,17 +257,17 @@ class ShuffleWriter {
       : numPartitions_(numPartitions),
         partitionWriterCreator_(std::move(partitionWriterCreator)),
         options_(std::move(options)),
-        pool_(std::make_shared<ShuffleMemoryPool>(options_.memory_pool)),
+        splitBufferPool_(std::make_shared<ShuffleMemoryPool>(options_.memory_pool)),
```
zuochunwei, line 260: changing the type of `splitBufferPool_` to `ShuffleMemoryPool` may be better.

the lifecycle of `splitBufferPool_` is binding with ShuffleWriter.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -343,6 +337,13 @@ std::shared_ptr<arrow::RecordBatch> makeUncompressedRecordBatch(
   }
   return arrow::RecordBatch::Make(writeSchema, 1, {arrays});
 }
+
+inline bool shrinkBeforeSpill(SplitState state) {
+  return state == SplitState::kStop;
+}
```
zuochunwei, line 343: add a space line

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -343,6 +337,13 @@ std::shared_ptr<arrow::RecordBatch> makeUncompressedRecordBatch(
   }
   return arrow::RecordBatch::Make(writeSchema, 1, {arrays});
 }
+
+inline bool shrinkBeforeSpill(SplitState state) {
```
zuochunwei, line 341: the `inline` is unnecessary, the compiler will make the decision when she sees the function defined in the inner of the class.

zuochunwei:
Could you provide some benchmark data for this PR?
I am curious about how much heap memory is saved after this change and the performance changes.

FelixYBW:
@zhztheplayer Let's add the buffers on OOM message like
shuffle write:
  spill buffer size:
  split buffer size:
  arrow::default memory pool size:

zhztheplayer:
> @zhztheplayer Let's add the buffers on OOM message once this PR is merged. something like shuffle write: spill buffer size: split buffer size: arrow::default memory pool size:

We can just use different Arrow pools for different purposes? Do you think it's technically possible? @marin-ma

FelixYBW (approved):

PR #3170, opened 2023-09-15
[GLUTEN-3077][VL] followup: add jniCastOrThrow method and codes clean up

Yohahaha:
Summary:
1. add `jniCastOrThrow` method to wrap reinterpret_cast and reduce duplicate codes.
2. move `create/releaseExecutionCtx` from VeloxJniWrapper.cc to JniWrapper.cc, `initializeBackend` will be refactored in spilt JniWrapper patch.

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -308,6 +308,26 @@ void JNI_OnUnload(JavaVM* vm, void* reserved) {
   env->DeleteGlobalRef(shuffleReaderMetricsClass);
 }
 
+JNIEXPORT jlong JNICALL Java_io_glutenproject_init_BackendJniWrapper_createExecutionCtx( // NOLINT
+    JNIEnv* env,
+    jclass) {
+  JNI_METHOD_START
+  auto executionCtx = gluten::createExecutionCtx();
+  return reinterpret_cast<jlong>(executionCtx);
+  JNI_METHOD_END(kInvalidResourceHandle)
+}
+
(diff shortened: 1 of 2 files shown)


Yohahaha:
@zhztheplayer would you help take a look?

zhztheplayer:
Thanks for the clean up! But do we need a macro for the purpose? Or a template function will be enough?

Yohahaha:
> But do we need a macro for the purpose? Or a template function will be enough?

I dont have strongly preference, and be fine if you insist on template function, but I'm not sure how to print template T to string as GLUTEN_CHECK's second arg.

```c++
template <typename T>
T* castBackResource(ResourceHandle handle) {
  auto resource = reinterpret_cast<T*>(handle);
  GLUTEN_CHECK(resource != nullptr, T "should not be null.");
  return resource;
}
```

zhztheplayer:
I think `GLUTEN_JNI_CAST(executionCtx, ctxHandle, ExecutionCtx)` is slightly less readable for developers.

Something like `executionCtx = jniCast<ExecutionCtx>(ctxHandle)` looks much better.

BTW the previous error message "... should not be null" is not accurate. It should be "FATAL:  ... is not a instance of ..." or something like that.

If it's hard to get the type name we can just make it `executionCtx = reintepretCastOrThrow<ExecutionCtx>(ctxHandle, "FATAL: ... is not a instance of ...")` which is too explanatory enough to me.

Yohahaha:
@zhztheplayer could you review again?

zhztheplayer (approved):

zhztheplayer:
Thanks!

PR #3177, opened 2023-09-18
[VL] Refactor split buffer allocation

marin-ma:
Extract the duplicated logic in both `resizePartitionBuffer` and `allocatePartitionBuffer`.
In prealloc partition buffer, use `resizePartitionBuffer` instead of `allocatePartitionBuffer` if need to preserve buffer data for the next split, and remove the resize branch in `allocate PartitionBuffer`.


marin-ma:
@FelixYBW Could you help to review? Thanks!

FelixYBW (approved):
Thank you for your work!

PR #3184, opened 2023-09-18
[GLUTEN-3154][FOLLOWUP] Add keyGroupedPartitioning in BatchScanExecTransformer for spark-3.3

liujiayi771:

Fix Spark-3.3 BatchScanExecTransformer makeCopy error after removing pushdownFilters parameter in BatchScanExecTransformer.
```
Exception in thread "main" java.lang.IllegalStateException:
Failed to copy node.
Is otherCopyArgs specified correctly for BatchScanExecTransformer.
Exception message: wrong number of arguments
ctor: public io.glutenproject.execution.BatchScanExecTransformer(scala.collection.Seq,org.apache.spark.sql.connector.read.Scan,scala.collection.Seq)?
types: class scala.collection.immutable.$colon$colon, class org.apache.iceberg.spark.source.SparkBatchQueryScan, class scala.collection.immutable.$colon$colon, class scala.None$
args: List(xxxxx#10037, xxxxx#10046, xxxxx#10047L, xxxxx#10048L, xxxxx#10079, xxxxx#10080, xxxxx#10081), IcebergScan(table=default_iceberg.xxxxx.xxxxx, type=struct<1: xxxxx: optional string, 10: xxxxx: optional string, 11: xxxxx: optional long, 12: xxxxx: optional long, 43: xxxxx: optional string, 44: xxxxx: optional string, 45: xxxxx: optional string>, filters=[not_null(ref(name="xxxxx")), not_null(ref(name="xxxxx")), ref(name="xxxxx") == "2023-08-01", not(ref(name="xxxxx") == "xxxxx"), not_null(ref(name="xxxxx")), not_null(ref(name="xxxxx"))], runtimeFilters=[], caseSensitive=false), List(dynamicpruningexpression(true), dynamicpruningexpression(xxxxx#10080 IN dynamicpruning#10622)), None

	at org.apache.spark.sql.catalyst.trees.TreeNode.makeCopy(TreeNode.scala:854)
	at org.apache.spark.sql.catalyst.trees.TreeNode.makeCopy(TreeNode.scala:797)
	at org.apache.spark.sql.execution.SparkPlan.super$makeCopy(SparkPlan.scala:100)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$makeCopy$1(SparkPlan.scala:100)
	at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:779)
	at org.apache.spark.sql.execution.SparkPlan.makeCopy(SparkPlan.scala:100)
	at org.apache.spark.sql.execution.SparkPlan.makeCopy(SparkPlan.scala:60)
	at org.apache.spark.sql.catalyst.plans.QueryPlan.mapExpressions(QueryPlan.scala:223)
	at org.apache.spark.sql.catalyst.plans.QueryPlan.transformExpressionsUpWithPruning(QueryPlan.scala:188)
	at org.apache.spark.sql.execution.reuse.ReuseExchangeAndSubquery$$anonfun$org$apache$spark$sql$execution$reuse$ReuseExchangeAndSubquery$$reuse$1$1.applyOrElse(ReuseExchangeAndSubquery.scala:54)
	at org.apache.spark.sql.execution.reuse.ReuseExchangeAndSubquery$$anonfun$org$apache$spark$sql$execution$reuse$ReuseExchangeAndSubquery$$reuse$1$1.applyOrElse(ReuseExchangeAndSubquery.scala:44)
```
In Spark-3.3, when Spark is in TreeNode.makeCopy, it will call the productArity method of Product to obtain the number of parameters of the case class. The current number of parameters of BatchScanExecTransformer is 3, but productArity gets 4. After my test, productArity gets the number of parameters of the first specific case class, which is the number of parameters of BatchScanExec, which does not match the number of BatchScanExecTransformer constructors.

I think the standard approach here is that if we need to inherit Spark's case class plan, we need to ensure that the number of parameters is consistent, otherwise it will cause problems when plan.transform calls makeCopy. Or implement the otherCopyArgs method to ensure that not transformed args can be copied.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
@@ -33,10 +33,17 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
 
 import java.util.Objects
 
+/**
+ * Columnar Based BatchScanExec. Although keyGroupedPartitioning is not used, it cannot be deleted,
+ * it can make BatchScanExecTransformer contain a constructor with the same parameters as
+ * Spark-3.3's BatchScanExec. Otherwise, the corresponding constructor will not be found when
+ * calling TreeNode.makeCopy and will fail to copy this node during transformation.
+ */
 class BatchScanExecTransformer(
     output: Seq[AttributeReference],
     @transient scan: Scan,
-    runtimeFilters: Seq[Expression])
+    runtimeFilters: Seq[Expression],
+    keyGroupedPartitioning: Option[Seq[Expression]] = None)
   extends BatchScanExecShim(output, scan, runtimeFilters)
   with BasicScanExecTransformer {
 


liujiayi771:
@PHILO-HE Please take a look.

PHILO-HE (approved):
Thanks for digging out the underneath reason!

gluten-core/src/main/scala/io/glutenproject/execution/BatchScanExecTransformer.scala
```diff
@@ -33,10 +33,17 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
...
+ * it can make BatchScanExecTransformer contain a constructor with the same parameters as
+ * Spark-3.3's BatchScanExec. Otherwise, the corresponding constructor will not be found when
+ * calling TreeNode.makeCopy and will fail to copy this node during transformation.
+ */
 class BatchScanExecTransformer(
     output: Seq[AttributeReference],
     @transient scan: Scan,
-    runtimeFilters: Seq[Expression])
+    runtimeFilters: Seq[Expression],
+    keyGroupedPartitioning: Option[Seq[Expression]] = None)
```
JkSelf, line 46: @liujiayi771 Does this change affect Spark3.2? Because spark 3.2 doesn't contain the `keyGroupedPartitioning `parameter.

liujiayi771: It will not affect spark3.2, it is an optional parameter. Did you encounter any problems?

JkSelf: @liujiayi771 We are currently working on upgrading to Spark 3.4 in Gluten. One of the changes in Spark 3.4 is the modification of BatchScanExec. It introduces a new table parameter. Currently, I have placed all the parameters in the constructor of BatchScanExecTransformer. However the table is not an optional parameter, I am unsure whether this change will affect Spark 3.2 and 3.3 and cause them to encounter this issue.  Do you have any suggestions regarding this? Thank you.

liujiayi771: @JkSelf According to the description of this PR, I think that such modification in 3.4 will cause the same problems that I encountered before. The reason is as explained in the description. Adding it directly to member variables will cause the lower version spark to not be able to find the corresponding constructors.

liujiayi771: @JkSelf It may be necessary to re-invent BatchScanTransformer so that it does not inherit Spark's BatchScanExec. Inheriting case class is not a good practice.

JkSelf: @liujiayi771 Yes, I will override the `otherCopyArgs ` method in the shim layer to bypass this issue when upgrading to Spark 3.4 [here](https://github.com/oap-project/gluten/pull/3360). We can further optimize the BatchScanExecTransformer in another pull request because it appears that the gluten override the case class for most XXTransformer.

PR #3185, opened 2023-09-18
[GLUTEN-3077][VL] Use ResourceMap to replace ConcurrentMap in VeloxExecutionCtx

Yohahaha:

Since ExecutionCtx does not expected to use in concurrent context, we can remove Concurrent which has mutex for each methods.

Added ResourceMap is backed by folly::F14FastMap.


Yohahaha:
@zhztheplayer could you help take a look? thanks!

zhztheplayer (approved):

PR #3191, opened 2023-09-18
[GLUTEN-3143][CH] Transform arrayJoin function to array join step to apply max_block_size

taiyang-li:

Notice that arrayJoin function and array join step are executing in different places. The former is in `ExpressionActions::execute`,  and latter is in `ArrayJoinAction::execute`. `max_block_size` is currently only valid in array join step, so we must transform arrayJoin function to array join step to apply max_block_size.  

Notice: This pr relies on https://github.com/ClickHouse/ClickHouse/pull/54664, but it can be merge firstly. 

(Fixes: \#3143)


taiyang-li:
@liuneng1994 @baibaichen There has been serveral uts about lateral view explode/posexplode, so this pr didn't add new uts.

liuneng1994 (approved):
LGTM

PR #3198, opened 2023-09-19
[GLUTEN-3048][VL] Disable Velox memory pool's own leak check since it may cause process crashes

zhztheplayer:
The leak check may throw exception from MemoryPool's destructor which could not be caught by user code.

The leak check should either be done by Spark's memory manager.

Enable the random kill tasks CI job to see if it could pass.


zhztheplayer:
See [CI log](https://github.com/oap-project/gluten/actions/runs/6230255116/job/16909991599?pr=3198), the PR solved a series of crashes but there still remains some.

marin-ma (approved):

PR #3199, opened 2023-09-19
[GLUTEN-3189] Fix shuffle spill failure and wrong result

marin-ma:
This patch fixes several cases:
1. If the buffer size after shrink is 0, do not shrink but reset the buffers.
2. If the buffer size before shrink is 0 or no space to shrink, skip shrink.
3. Fix split after spill wrong result.

Added several C++ UT to cover above cases.

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -1098,7 +1098,7 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
           binaryBuf.valueCapacity = capacity;
           dstValuePtr = binaryBuf.valuePtr + valueOffset - stringLen;
 
-          std::cout << "Split value buffer resized colIdx" << binaryIdx << std::endl;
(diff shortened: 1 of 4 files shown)


cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -1097,7 +1101,7 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const velox::RowVec
           binaryBuf.valueCapacity = capacity;
           dstValuePtr = binaryBuf.valuePtr + valueOffset - stringLen;
 
-          std::cout << "Split value buffer resized colIdx" << binaryIdx << std::endl;
+          DLOG(INFO) << "Split value buffer resized. Column index: " << binaryColumnIndices_[binaryIdx] << std::endl;
```
marin-ma, line 1104: @FelixYBW Our CI now prints too many logs from this line. Can we remove the log here?

cc: @Yohahaha https://github.com/oap-project/gluten/issues/3208#issuecomment-1726854275

marin-ma: Let's remove the log here. `VsPrintSplit` at below has given enough message.

FelixYBW (approved):

Issue #3206, opened 2023-09-20
[CH] SPARK-36926 is not supported by ch backend


lhuang09287750:
### Description

SPARK-36926 in GlutenDataFrameAggregateSuite cannot pass:
`  test("SPARK-36926") {
    val df = (1 to 10).map(_ => "9999999999.99").toDF("d")
    val res = df.select($"d".cast("decimal(12, 2)").as("d")).agg(avg($"d").cast("string"))
    checkAnswer(res, Row("9999999999.990000"))
  }`
gluten with chbackend get 9999999999.99,but spark get 9999999999.990000
That is because,during aggregation decimal(12, 2) will be upgraded to decimal(16, 6)
so spark will reserve six numbers for scale 6 after dot,while gluten ignore zero and only reserve two numbers

PR #3212, opened 2023-09-20
[GLUTEN-3206][CH] fix SPARK-36926 UT for chbackend

lhuang09287750:
Use ch func 'toDecimalString' instead of 'cast' expression when make cast from decimal to string

(Fixes: \#3206)

## How was this patch tested?

test with spark ut in GlutenDataFrameAggregateSuite


lhuang09287750:
Run Gluten Clickhouse CI

baibaichen (approved):

PR #3217, opened 2023-09-20
[VL] Refactor getMetrics to decrease metrics related redundant code

zuochunwei:

It seems to be a better way to use `enum`, it makes adding a new metric item easier and the code is also simpler.

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -431,95 +431,45 @@ JNIEXPORT jobject JNICALL Java_io_glutenproject_vectorized_ColumnarBatchOutItera
   auto executionCtx = jniCastOrThrow<ExecutionCtx>(ctxHandle);
 
   auto iter = executionCtx->getResultIterator(iterHandle);
-  std::shared_ptr<Metrics> metrics = iter->getMetrics();
+  auto metrics = iter->getMetrics();
+  unsigned int numMetrics = metrics->numMetrics;
 
-  int numMetrics = 0;
-  if (metrics) {
-    numMetrics = metrics->numMetrics;
-  }
-  auto inputRows = env->NewLongArray(numMetrics);
-  auto inputVectors = env->NewLongArray(numMetrics);
-  auto inputBytes = env->NewLongArray(numMetrics);
-  auto rawInputRows = env->NewLongArray(numMetrics);
-  auto rawInputBytes = env->NewLongArray(numMetrics);
-  auto outputRows = env->NewLongArray(numMetrics);
-  auto outputVectors = env->NewLongArray(numMetrics);
-  auto outputBytes = env->NewLongArray(numMetrics);
-  auto cpuCount = env->NewLongArray(numMetrics);
-  auto wallNanos = env->NewLongArray(numMetrics);
-  auto peakMemoryBytes = env->NewLongArray(numMetrics);
-  auto numMemoryAllocations = env->NewLongArray(numMetrics);
-  auto spilledBytes = env->NewLongArray(numMetrics);
-  auto spilledRows = env->NewLongArray(numMetrics);
-  auto spilledPartitions = env->NewLongArray(numMetrics);
-  auto spilledFiles = env->NewLongArray(numMetrics);
-  auto numDynamicFiltersProduced = env->NewLongArray(numMetrics);
-  auto numDynamicFiltersAccepted = env->NewLongArray(numMetrics);
-  auto numReplacedWithDynamicFilterRows = env->NewLongArray(numMetrics);
-  auto flushRowCount = env->NewLongArray(numMetrics);
-  auto scanTime = env->NewLongArray(numMetrics);
-  auto skippedSplits = env->NewLongArray(numMetrics);
-  auto processedSplits = env->NewLongArray(numMetrics);
-  auto skippedStrides = env->NewLongArray(numMetrics);
-  auto processedStrides = env->NewLongArray(numMetrics);
-
-  if (metrics) {
-    env->SetLongArrayRegion(inputRows, 0, numMetrics, metrics->inputRows);
-    env->SetLongArrayRegion(inputVectors, 0, numMetrics, metrics->inputVectors);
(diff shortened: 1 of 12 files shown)


cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -226,73 +222,76 @@ void WholeStageResultIterator::collectMetrics() {
...
+      statsNum += 1;
       continue;
     }
-    numOfStats += planStats.at(nodeId).operatorStats.size();
+    statsNum += planStats.at(nodeId).operatorStats.size();
   }
 
-  metrics_ = std::make_shared<Metrics>(numOfStats);
-  int metricsIdx = 0;
+  metrics_ = std::move(std::make_unique<Metrics>(statsNum));
```
Yohahaha, line 231: Could you explain the difference between `metrics_ = std::move(std::make_unique<Metrics>(statsNum));` and `metrics_ = std::make_unique<Metrics>(statsNum);`?

My IDE always complain about this usage.

zuochunwei: the `unique_ptr` does not support `unique_ptr& operator=(unique_ptr& rhs)` .
the compiler will report errors without `std::move`.

cpp/core/utils/metrics.h
```diff
@@ -17,104 +17,72 @@
...
+    kNumReplacedWithDynamicFilterRows,
+    kFlushRowCount,
+    kScanTime,
+    kSkippedSplits,
+    kProcessedSplits,
+    kSkippedStrides,
+    kProcessedStrides,
+
+    kEnd,
+    kNum = kEnd - kBegin
```
rui-mo, line 70: I believe `kNum` should be not an item of this enum because it is not one possible option. Maybe change it to a public variable of `Metrics`.

cpp/core/utils/metrics.h
```diff
@@ -17,104 +17,72 @@
...
-    flushRowCount = new long[numMetrics]();
-    scanTime = new long[numMetrics]();
-    skippedSplits = new long[numMetrics]();
-    processedSplits = new long[numMetrics]();
-    skippedStrides = new long[numMetrics]();
-    processedStrides = new long[numMetrics]();
+  unsigned int numMetrics = 0;
+  long veloxToArrow = 0;
+
+  std::unique_ptr<long[]> array; // underlying memory buffer
```
rui-mo, line 28: Please capitalize the first letter and add `.` at the end for all comments.

cpp/core/utils/metrics.h
```diff
@@ -17,104 +17,72 @@
...
+  long* arrayRawPtr = nullptr; // point to array.get() after array created
```
rui-mo, line 29: Comment style.

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -226,73 +222,76 @@ void WholeStageResultIterator::collectMetrics() {
...
     }
-    numOfStats += planStats.at(nodeId).operatorStats.size();
+    statsNum += planStats.at(nodeId).operatorStats.size();
   }
 
-  metrics_ = std::make_shared<Metrics>(numOfStats);
-  int metricsIdx = 0;
+  metrics_ = std::make_unique<Metrics>(statsNum);
+
+  int i = 0; // stands for Metric Index
```
rui-mo, line 233: Why not name it as `metricIndex ` and remove this comment?

rui-mo (commented):
Thanks for your refactor. Below are some comments.

cpp/core/utils/metrics.h
zuochunwei: kNum is used to stand for enum item count,  We can use it to determine the length of an array, it's a frequently used style, so let's keep this?

cpp/core/utils/metrics.h
zuochunwei: OK

cpp/core/utils/metrics.h
zuochunwei: OK

cpp/velox/compute/WholeStageResultIterator.cc
zuochunwei: OK

cpp/core/jni/JniWrapper.cc
```diff
@@ -431,95 +431,45 @@ JNIEXPORT jobject JNICALL Java_io_glutenproject_vectorized_ColumnarBatchOutItera
   auto executionCtx = jniCastOrThrow<ExecutionCtx>(ctxHandle);
 
   auto iter = executionCtx->getResultIterator(iterHandle);
-  std::shared_ptr<Metrics> metrics = iter->getMetrics();
+  auto metrics = iter->getMetrics();
+  unsigned int numMetrics = metrics->numMetrics;
```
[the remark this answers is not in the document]
zuochunwei: OK

cpp/core/jni/JniWrapper.cc
```diff
@@ -431,95 +431,50 @@ JNIEXPORT jobject JNICALL Java_io_glutenproject_vectorized_ColumnarBatchOutItera
   auto executionCtx = jniCastOrThrow<ExecutionCtx>(ctxHandle);
 
   auto iter = executionCtx->getResultIterator(iterHandle);
-  std::shared_ptr<Metrics> metrics = iter->getMetrics();
-
-  int numMetrics = 0;
+  auto metrics = iter->getMetrics();
```
rui-mo, line 434: Maybe we need to check and throw if `metrics` is nullptr, because it is still directly used as below.
> env->SetLongArrayRegion(longArray[i], 0, numMetrics, metrics->get((Metrics::TYPE)i));

cpp/core/utils/metrics.h
rui-mo: OK. Makes sense.

rui-mo (approved):

Issue #3213, opened 2023-09-20
[CH] null column determination bug when execute row to column transform


lhuang09287750:
### Backend

CH (ClickHouse)

### Bug description

when output has more than 31 columns and is followed by a row to column transform, the columns after the 31st column may be given wrong value,especially when those columns are null ,then they may be given their type's default value or a non-null value may be read as null
E.G, the 32nd column is string type and is non-null,but after row to column,it may be read as null

### Spark version

Spark-3.2.x

lhuang09287750:
nullbitmap read error when r2c

PR #3219, opened 2023-09-20
[GLUTEN-3213][CH] use wrong type(int32) to construct null bitmap mask when r2c

lhuang09287750:

change it to long(int64)

(Fixes: \#3213)

## How was this patch tested?

add a new testcase in GlutenDataFrameAggregateSuite


cpp-ch/local-engine/Parser/CHColumnToSparkRow.cpp
```diff
@@ -77,7 +77,7 @@ void bitSet(char * bitmap, size_t index)
 ALWAYS_INLINE bool isBitSet(const char * bitmap, size_t index)
 {
     assert(index >= 0);
-    int64_t mask = 1 << (index & 63);
+    int64_t mask = 1L << (index & 63);
```
zhanglistar, line 80: Use 1LL will be better. Since in x86_32, long is 32bit.

taiyang-li (approved):
LGTM.

cpp-ch/local-engine/Parser/CHColumnToSparkRow.cpp
lhuang09287750: err, I had no idea about the difference between L and LL. 
In the function bitSet(just before isBitSet) ,1L is used to make mask
And in the BItSetMethods.class which is in spark-unsafe jar ,1L is also used in isSet function,like this:
```
  public static boolean isSet(Object baseObject, long baseOffset, int index) {
    assert index >= 0 : "index (" + index + ") should >= 0";

    long mask = 1L << (index & 63);
    long wordOffset = baseOffset + (long)(index >> 6) * 8L;
    long word = Platform.getLong(baseObject, wordOffset);
    return (word & mask) != 0L;
  }
```
So I just use L for consistency. Seems LL is better

baibaichen (approved):
LGTM

PR #3220, opened 2023-09-20
[VL] Close pidBatch after composed in range partitioning

Yohahaha:

In range partitioning we create a batch and offloaded as pidBatch, but no place to close it.
Add batchHolder empty check in VeloxExecutionCtx dtor will find this issue.


Yohahaha:
@marin-ma could you help take a look?

marin-ma:
Thanks for the catch. Is this issue related memory leak? @zhztheplayer

Yohahaha:
Before ExecutionCtx lifecycle management it's an issue which may cause process memory increase slowly and released when process exit, after newly lifecycle management, it will be released normally.

marin-ma (approved):

PR #3223, opened 2023-09-20
[GLUTEN-2961][VL][CORE] Decimal sum avg use row_constructor

liujiayi771:

Decimal sum and avg have the meaning of overflow when the child field is null, and gluten's row_constructor_with_null cannot be used.

## How was this patch tested?
SPARK-28067: Aggregate sum should not return wrong results for decimal overflow
SPARK-35955: Aggregate avg should not return wrong results for decimal overflow


liujiayi771:
@JkSelf Please take a look.

JkSelf (approved):
LGTM

JkSelf:
@liujiayi771 Can you help to rebase? Thanks.

liujiayi771:
> @liujiayi771 Can you help to rebase? Thanks.

Done.

PR #3236, opened 2023-09-21
Fix a bug about replacing spark sort agg with gluten hash agg

weixiuli:

Fix a bug about replacing spark sort agg with gluten hash agg
- When the Final SortAggregate fallback, the Partial SortAggregate should also fallback to ensure data accuracy

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
@@ -141,6 +141,50 @@ class VeloxAggregateFunctionsSuite extends WholeStageTransformerSuite {
     }
   }
 
+  test(
+    "when the Final SortAggregate fallback, the Partial SortAggregate should also fallback " +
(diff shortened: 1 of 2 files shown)


Yohahaha:
Nice catch, and how about set `TRANSFORM_UNSUPPORTED` tag to logical agg node when we meet found a fallback node, then check that logical agg node before validate partial agg node?  I think this way is more make sense.

PHILO-HE:
Thanks for your fix! 
What if partial/final sort agg are separated by shuffle into different stages? It may need more considerations. Right?

ulysses-you:
thank you @weixiuli , do you mean, the velox does not support some aggregate functions with final mode but support partial mode ? Does ObjectHashAggregate has same issue ?

weixiuli:
Run Gluten Clickhouse CI

PHILO-HE:
Hi @weixiuli, as mentioned in https://github.com/oap-project/gluten/issues/3391, one proposed solution is, moving unsupported post project to an independent project operator. Thus, we can make final agg transformed and just make post project fall back.

PR #2605, opened 2023-08-04
[GLUTEN-2604][VL] Fix github dependabot alert for Spark version

xieqi:
## What changes were proposed in this pull request?

Fix dependabot alert

```
Apache Spark vulnerable to Improper Privilege Management

Upgrade org.apache.spark:spark-core_2.12 to fix [1 Dependabot alert](https://github.com/oap-project/gluten/security/dependabot?q=is%3Aopen+package%3Aorg.apache.spark%3Aspark-core_2.12+manifest%3Atools%2Fgluten-it%2Fpom.xml+has%3Apatch) in [tools/gluten-it/pom.xml](https://github.com/oap-project/gluten/blob/-/tools/gluten-it/pom.xml)
Upgrade org.apache.spark:spark-core_2.12 to version 3.4.0 or later.

In Apache Spark versions prior to 3.4.0, applications using spark-submit can specify a proxy-user to run as, limiting privileges. The application can execute code with the privileges of the submitting user, however, by providing malicious configuration-related classes on the classpath. This affects architectures relying on proxy-user, for example those using Apache Livy to manage submitted applications.

Update to Apache Spark 3.4.0 or later, and ensure that spark.submit.proxyUser.allowCustomClasspathInClusterMode is set to its
default of "false", and is not overridden by submitted applications.

```
## How was this patch tested?

Use exist UTs

PR #3238, opened 2023-09-21
[VL] Gluten-it: By default use -P spark-3.2

zhztheplayer:
In #2605 , we by default set -P spark-3.4 for gluten-it to fix the dependent bot alert. However it was more or less a deceit to go around dep bot's code scanning without actually fixing the issue.

Let's use spark-3.2 to avoid mismatching with gluten's main code until we came up with a better fix.


zhouyuan (approved):
👍

PR #3239, opened 2023-09-21
[VL] Simplify ExecutionCtx + Handle JNI code pattern

zhztheplayer:
A clean-up to JNI code by pulling up a common pattern `ExecutionCtx` + `Handle`.


zhztheplayer:
Closing, temporarily

zhztheplayer:
cc @Yohahaha

Yohahaha:
Thanks!

Overall goods, only one thoughts:
```java
  public static JniWrapper create(ExecutionCtx ctx) {
    return new JniWrapper(ExecutionCtxs.contextInstance);
  }

  public static JniWrapper create(ExecutionCtx ctx) {
    return new JniWrapper(ctx);
  }
```

First is good enough for user, but I prefer second one, it more like a develop API which bring more flexibility for developer who wants implement their own feature.

zhztheplayer:
> Thanks!
> 
> Overall looks goods, only one thoughts:
> 
> ```java
>   public static JniWrapper create(ExecutionCtx ctx) {
>     return new JniWrapper(ExecutionCtxs.contextInstance);
>   }
> 
>   public static JniWrapper create(ExecutionCtx ctx) {
>     return new JniWrapper(ctx);
>   }
> ```
> 
> First is good enough for user, but I prefer second one, it more like a develop API which bring more flexibility for developer who wants implement their own feature.

I added both `create()` and `forCtx(ctx)`. Was that same with what you suggested?

zhztheplayer:
Besides, developer can add `forCtx(ctx)` API to more JNI wrappers once needed.

rui-mo (approved):

INBreezefall:
> A clean-up to JNI code by pulling up a common pattern `ExecutionCtx` + `Handle`.
Hi, I am a Gluten newbie. I have a simple question, hope you can help me. 

I can see that the ExecutionCtxAware interface is implemented in many JNI classes, but its handle method is not called in the Gluten project. Can this interface be deleted from JNI, such as ColumnarBatchJniWrapper?
What is the purpose of this interface? Will the handle method be called?

PR #3255, opened 2023-09-22
[GLUTEN-3252][CH] fix race condition at the global variable of isAdaptiveContext in ColumnarOverrideRules

zheniantoushipashi:

[GLUTEN-3252][CH] fix race condition at the global variable of isAdaptiveContext in ColumnarOverrideRules

## How was this patch tested?

test  GlutenClickHouseTPCHParquetAQEConcurrentSuite  can  reproduce

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -731,8 +731,6 @@ case class ColumnarOverrideRules(session: SparkSession)
   private lazy val transformPlanLogLevel = GlutenConfig.getConf.transformPlanLogLevel
   @transient private lazy val planChangeLogger = new PlanChangeLogger[SparkPlan]()
 
-  // Tracks whether the columnar rule is called through AQE.
-  private var isAdaptiveContext: Boolean = false
   // This is an empirical value, may need to be changed for supporting other versions of spark.
   private val aqeStackTraceIndex = 13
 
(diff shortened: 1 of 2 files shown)


ulysses-you:
good catch!

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -731,8 +731,6 @@ case class ColumnarOverrideRules(session: SparkSession)
   private lazy val transformPlanLogLevel = GlutenConfig.getConf.transformPlanLogLevel
   @transient private lazy val planChangeLogger = new PlanChangeLogger[SparkPlan]()
 
-  // Tracks whether the columnar rule is called through AQE.
-  private var isAdaptiveContext: Boolean = false
```
PHILO-HE, line 735: Thanks so much for your fix!
Can we just use `ThreadLocal` type for `isAdaptiveContext` to fix this issue?

zzcclp: `localProperties` is also a `ThreadLocal` parameter, using `localProperties` is OK to me.

baibaichen: @PHILO-HE 
I think `LocalProperty `is more suitable here, and there will be some similar variables that need to be added in the future.

PHILO-HE (approved):

zzcclp (approved):
LGTM

PR #3265, opened 2023-09-24
[VL] Shrink min sized partition buffers and spill

marin-ma:
Currently, all partition buffers will be shrunk when spill is triggered. This patch refine the total reclaimable size with following modifications:
1. For `SplitState::kStop`, shrink happens before spill and the behavior is unchanged. Shrink all partition buffers to current size.
2. For `SplitState::kSplit` and `SplitState::kInit`, shrink happens after spill. Stop shrinking the buffers once the reclaimed size is not less than the requested size. For `SplitState::kInit`, the spill is not triggered during shuffle split and should be other operators. If shrinking partition buffers doesn't free enough memory, will spill partition buffers with minimum size.


cpp/core/shuffle/PartitionWriter.h
```diff
@@ -21,17 +21,36 @@
...
+class SpillHandle {
+ public:
+  virtual ~SpillHandle() = default;
+
+  virtual arrow::Status consume(uint32_t partitionId, std::unique_ptr<arrow::ipc::IpcPayload> payload) = 0;
+
+  virtual arrow::Status close() = 0;
+
+  virtual bool closed() = 0;
+};
```
zhztheplayer, line 33: `evict` could be preferred over `spill` regarding partition writer API?

cpp/core/shuffle/PartitionWriter.h
```diff
@@ -21,17 +21,36 @@
...
+  /// Request next spill.
+  /// \param flushOnSpill Whether to flush the spilled data immediately. If it's false,
+  /// the spilled data can be cached first.
+  virtual arrow::Status requestNextSpill(bool flushOnSpill) = 0;
+
+  /// Get the current managed SpillHandle. Returns nullptr if the current SpillHandle was closed,
+  /// or requestNextSpill has not been called.
+  /// \return
+  virtual std::shared_ptr<SpillHandle> getSpillHandle() = 0;
+
```
zhztheplayer, line 53: Was there any reason preventing using a single member function `virtual Result<std::shared_ptr<SpillHandle>> requestNextSpill`?

zhztheplayer:
@marin-ma FWIW the low mem GHA job downed at a suspicious place

https://github.com/oap-project/gluten/actions/runs/6293529693/job/17087545511?pr=3265

cpp/core/shuffle/PartitionWriter.h
marin-ma: The caller can use `requestNextSpill` to start a spill, and choose to call `getSpillHandle()->consume()` immediately, or to call it latter somewhere else. Only `requestNextSpill` cannot make the caller to start a new spill multiple times.

marin-ma:
> @marin-ma FWIW the low mem GHA job downed at a suspicious place
> 
> https://github.com/oap-project/gluten/actions/runs/6293529693/job/17087545511?pr=3265

Thanks. Let me try to reproduce locally.

cpp/core/shuffle/LocalPartitionWriter.cc
```diff
@@ -24,14 +24,155 @@
...
+      : LocalPartitionWriter::LocalEvictHandle(numPartitions, options, spillInfo) {}
+
+  arrow::Status evict(uint32_t partitionId, std::unique_ptr<arrow::ipc::IpcPayload> payload) override {
+    if (!os_) {
+      ARROW_ASSIGN_OR_RAISE(os_, arrow::io::FileOutputStream::Open(spillInfo_->spilledFile, true));
+    }
+    int32_t metadataLength = 0; // unused.
+
+    ARROW_ASSIGN_OR_RAISE(auto start, os_->Tell());
+#ifndef SKIPWRITE
```
FelixYBW, line 131: Let's remove the SKIPWRITE macro, it's used for debug purpose

FelixYBW (dismissed):
Thank you for your work! looks good to me

FelixYBW (approved):

PR #3268, opened 2023-09-25
[MINOR] Refine JoinUtils.createPreProjectionIfNeeded

waitinfuture:
Refine JoinUtils.createPreProjectionIfNeeded to be more readable.

## How was this patch tested?
Passes GA.


gluten-core/src/main/scala/io/glutenproject/execution/JoinUtils.scala
```diff
@@ -104,14 +104,14 @@ object JoinUtils {
       )
 
       // Compute index for join keys in join outputs.
-      val offset = joinOutput.size - inputNodeOutput.size + selectOrigins.size
```
waitinfuture, line 107: Since `inputNodeOutput` has the same size with `selectOrigins`, we can simplify here.

waitinfuture:
cc @JkSelf

JkSelf (approved):
Thanks for your good catch. LGTM.

PR #3269, opened 2023-09-25
[VL] Simplify the usage of debug print by providing a macro DEBUG_OUT

zuochunwei:

I designed a fake OStream class `FakeCout` to support the writing of `COUT << sth`, the empty implementation of `FakeCout::operator<<(T)` will be optimized out when the compiler's optimization switch is open. so don't worry about the unnecessary CPU time cost.

(Fixes: \#ISSUE-ID)

## How was this patch tested?


cpp/core/utils/Cout.h
```diff
@@ -0,0 +1,51 @@
...
+  }
+};
+
+extern FakeCout fakeCout;
+
+#define COUT fakeCout
+
+#endif
+
+} // namespace gluten
```
zhztheplayer, line 51: Please add a EoF line. Thanks!

cpp/core/utils/Cout.h
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+namespace gluten {
+
+#ifdef GLUTEN_PRINT_DEBUG
+
+#define COUT std::cout
```
zhztheplayer, line 24: `COUT` sounds like a regular print? Should it be named `DOUT` or something like that?

zhztheplayer (commented):
Thanks!

cpp/core/utils/Cout.h
zuochunwei: OK

cpp/core/utils/Cout.h
zuochunwei: how about DEBUG_OUT

zhztheplayer (approved):
Thanks!

PR #3290, opened 2023-09-26
[VL] Support columnar table cache with count(1)

ulysses-you:

Supports columnar `df.cache.count`. It is a special case that the selected schema is empty.

## How was this patch tested?

add test

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/VeloxColumnarCacheSuite.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/VeloxColumnarCacheSuite.scala
@@ -18,6 +18,8 @@ package io.glutenproject.execution
 
 import io.glutenproject.GlutenConfig
 
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.Row
 import org.apache.spark.sql.execution.SparkPlan
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.execution.columnar.InMemoryTableScanExec
@@ -31,6 +33,12 @@ class VeloxColumnarCacheSuite extends WholeStageTransformerSuite with AdaptiveSp
   override def beforeAll(): Unit = {
     super.beforeAll()
     createTPCHNotNullTables()
+  }
+
+  override protected def sparkConf: SparkConf = {
+    super.sparkConf
+      .set("spark.shuffle.manager", "org.apache.spark.shuffle.sort.ColumnarShuffleManager")
+      .set("spark.sql.shuffle.partitions", "3")
   }
 
   private def checkColumnarTableCache(plan: SparkPlan): Unit = {
@@ -97,6 +105,22 @@ class VeloxColumnarCacheSuite extends WholeStageTransformerSuite with AdaptiveSp
       checkAnswer(actual, expected)
     } finally {
       actual.unpersist()
+    }
+  }
+
+  test("Support transform count(1) with table cache") {
+    val cached = spark.table("lineitem").cache()
+    try {
+      val df = spark.sql("SELECT COUNT(*) FROM lineitem")
+      checkAnswer(df, Row(60175))
+      assert(
+        find(df.queryExecution.executedPlan) {
+          case _: RowToVeloxColumnarExec => true
+          case _ => false
+        }.isEmpty
+      )
+    } finally {
+      cached.unpersist()
     }
   }
 }
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -90,32 +90,7 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
    */
   private def genHashAggregateExec(plan: HashAggregateExec): SparkPlan = {
     val newChild = replaceWithTransformerPlan(plan.child)
-    // If child's output is empty, fallback or offload both the child and aggregation.
-    if (plan.child.output.isEmpty && BackendsApiManager.getSettings.fallbackAggregateWithChild()) {
-      newChild match {
-        case _: TransformSupport =>
-          // If the child is transformable, transform aggregation as well.
-          logDebug(s"Columnar Processing for ${plan.getClass} is currently supported.")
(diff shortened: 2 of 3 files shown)


gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
```diff
@@ -323,7 +323,7 @@ public static ColumnarBatch create(ExecutionCtx ctx, long nativeHandle) {
     int numColumns = Math.toIntExact(iv.getNumColumns());
     int numRows = Math.toIntExact(iv.getNumRows());
     if (numColumns == 0) {
-      return new ColumnarBatch(new ColumnVector[0], numRows);
+      return new ColumnarBatch(new ColumnVector[] {iv}, numRows);
```
ulysses-you, line 326: If the ColumnarBatch is light, then it must have one column. I think it won't break anything, but better to double check cc @zhztheplayer @PHILO-HE

backends-velox/src/test/scala/io/glutenproject/execution/VeloxColumnarCacheSuite.scala
```diff
@@ -99,4 +107,20 @@ class VeloxColumnarCacheSuite extends WholeStageTransformerSuite with AdaptiveSp
       actual.unpersist()
     }
   }
+
+  test("Support transform count(1) with table cache") {
+    val cached = spark.table("lineitem").cache()
+    try {
+      val df = spark.sql("SELECT COUNT(*) FROM lineitem")
```
PHILO-HE, line 114: Should use `COUNT(1)`?

ulysses-you: they are total same in Spark

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -90,42 +90,37 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
...
-              plan.requiredChildDistributionExpressions,
-              plan.groupingExpressions,
-              plan.aggregateExpressions,
-              plan.aggregateAttributes,
-              plan.initialInputBufferOffset,
-              plan.resultExpressions,
-              newChild
-            )
+          transformHashAggregate()
+        case i: InMemoryTableScanExec if i.supportsColumnar =>
```
PHILO-HE, line 113: Thanks so much for fix!
Shouldn't we add condition `BackendsApiManager.isVeloxBackend && GlutenConfig.getConf.columnarTableCacheEnabled` to make sure this is a scan for table cached by Gluten (i.e., child offloaded case)?

ulysses-you: yes, we have these checks, `InMemoryTableScanExec.supportsColumnar` will call `CachedBatchSerializer.supportsColumnarOutput`. So if supportsColumnar is true, that means we are in velox backend and enabled columnar table cache.

gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java
zhztheplayer: What happened without this change?

ulysses-you: `count(1)` requires child empty output, so the selected column for table cache is 0. Without this change, we can not make a valid light ColumnarBatch.

zhztheplayer: A batch with `new ColumnVector[] {iv}` creates inconsistency since it's both 1 column (by ColumnarBatch#numCols()) and 0 column (by iv.getNumColumns()) at the same time. So I am trying to understand what happened here. 

If it's about the call to `ColumnarBatches.create` in file `ColumnarCachedBatchSerializer.scala`, can we just improve `ColumnarBatches.select` to make the method support zero-column batch?

Also I think it's safe to consider a zero-column batch a heavy batch. Although our code didn't do that yet.

Besides FWIW the whole "light" "heavy" thing can be removed after stopping support Spark 3.2, there was a issue to track that https://github.com/oap-project/gluten/issues/540.

ulysses-you: >  If it's about the call to ColumnarBatches.create in file ColumnarCachedBatchSerializer.scala, can we just improve ColumnarBatches.select to make the method support zero-column batch?

It seems we can not.. Once we want to create a new light ColumnarBatch, we need to inject the `IndicatorVector`. That means we must have one column even if the acutally selected column size is 0.

> Also I think it's safe to consider a zero-column batch a heavy batch. Although our code didn't do that yet.

Yes, we can, and this method is only used to create light ColumnarBatch which should always have a `IndicatorVector`. So I think this change won't break this supposition.

> Besides FWIW the whole "light" "heavy" thing can be removed after stopping support Spark 3.2, there was a issue to track that https://github.com/oap-project/gluten/issues/540.

I see it, use a custom ColumnarBatch is better in future.

zhztheplayer: Thanks for the explanation. Let's use your code temporarily until finding a better way.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
PHILO-HE: Can it be vanilla reader for cached table when `supportsColumnar` is true?
I note this part of code makes agg plan & its child fall back or offload together. Is it necessary to add more check?
In addition, I note there is some check [here](
https://github.com/oap-project/gluten/blob/main/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala#L707) for `InMemoryTableScanExec`. Just wondering why they are handled differently. Thanks!

ulysses-you: ah, this is a good question. I changed the condition and unifed them. Hope it is clear now.

zhztheplayer (approved):
Thanks!

PR #3293, opened 2023-09-26
[VL] Update get_velox.sh for macos

zhangnew:

fix velox_ep/scripts/setup-macos.sh

folly needs to be installed after install_double_conversion

https://github.com/oap-project/velox/blob/1910a1722dcb5e7dde83e84defdc33f21327be84/scripts/setup-macos.sh#L116-L124

## How was this patch tested?

manual tests


ulysses-you (approved):
lgtm, if it works for your env

PR #3294, opened 2023-09-26
[CORE] Use thread local to hold original plan in ColumnarOverrideRules

ulysses-you:

`ColumnarOverrideRules` will be instanced per Spark session rather than per query, so better to use thread local to hold variable to make it thread safe. Including `oringalPlan` and `isAdaptiveContext`.

The local property of SparkContext is a little bit heavy that it would post it to executor side, so we should cleanup it before execution.

## How was this patch tested?

Pass CI

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -724,7 +724,7 @@ case class VanillaColumnarPlanOverrides(session: SparkSession) extends Rule[Spar
 }
 
 object ColumnarOverrideRules {
-  val GLUTEN_IS_ADAPTIVE_CONTEXT = "glutenIsAdaptiveContext"
+  val GLUTEN_IS_ADAPTIVE_CONTEXT = "gluten.isAdaptiveContext"
 }
 
 case class ColumnarOverrideRules(session: SparkSession)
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@baibaichen, @zzcclp

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -732,20 +732,55 @@ case class ColumnarOverrideRules(session: SparkSession)
   @transient private lazy val planChangeLogger = new PlanChangeLogger[SparkPlan]()
 
   // This is an empirical value, may need to be changed for supporting other versions of spark.
-  private val aqeStackTraceIndex = 13
+  private val aqeStackTraceIndex = 14
```
zzcclp, line 735: why change to 14 ?

zzcclp:
@zheniantoushipashi please take a review, thanks.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
ulysses-you: this pr adds one more function `setAdaptiveContext` and move `Thread.currentThread.getStackTrace` into it.

baibaichen:
related with https://github.com/oap-project/gluten/pull/3255

baibaichen:
> The local property of SparkContext is a little bit heavy that it would post it to executor side, so just use java thread local.

@ulysses-you 

We need disable or enable gluten per query.  If so, we should add similar mechanism at `ColumnarOverrideRules `, doesn't it messy？

ulysses-you:
@baibaichen @PHILO-HE I changed code based on the discussion

PHILO-HE (commented):
Just a trivial comment. Thanks!

PHILO-HE:
@zzcclp, @baibaichen, do you have any comment? Thanks!

baibaichen:
LGTM

zzcclp (approved):
LGTM

PR #3300, opened 2023-09-26
[VL]Avoid unnecessary filter binding for subfield

yma11:

Gluten tried to do a filter binding against subfield but actually not all subfields have corresponding filter. For example, `select * from table where a > 0` where `table` has cols `a` and `b`, so there is no need to do filter binding for `b` at all.

## How was this patch tested?

UT added

### DIFF
--- a/gluten-ut/spark32/src/test/scala/io/glutenproject/utils/velox/VeloxTestSettings.scala
+++ b/gluten-ut/spark32/src/test/scala/io/glutenproject/utils/velox/VeloxTestSettings.scala
@@ -806,6 +806,8 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude("Filter applied on merged Parquet schema with new column should work")
     .exclude("SPARK-23852: Broken Parquet push-down for partially-written stats")
     .exclude("SPARK-25207: exception when duplicate fields in case-insensitive mode")
+    // Rewrite for supported INT96 - timestamp
+    .exclude("filter pushdown - timestamp")
     .exclude("filter pushdown - date")
     // Ignore Spark's filter pushdown check.
     .exclude("Filters should be pushed down for vectorized Parquet reader at row group level")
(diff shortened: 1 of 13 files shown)


gluten-ut/spark32/src/test/scala/io/glutenproject/utils/velox/VeloxTestSettings.scala
```diff
@@ -768,6 +768,8 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude("Non-vectorized reader - without partition data column - SPARK-38977: schema pruning with correlated NOT IN subquery")
     .exclude("Non-vectorized reader - with partition data column - SPARK-38977: schema pruning with correlated NOT IN subquery")
   enableSuite[GlutenParquetColumnIndexSuite]
+    // Rewrite
```
rui-mo, line 771: Why this test is rewritten?

gluten-ut/spark33/src/test/scala/io/glutenproject/utils/velox/VeloxTestSettings.scala
```diff
@@ -616,6 +616,8 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude("Non-vectorized reader - without partition data column - SPARK-38977: schema pruning with correlated NOT IN subquery")
     .exclude("Non-vectorized reader - with partition data column - SPARK-38977: schema pruning with correlated NOT IN subquery")
   enableSuite[GlutenParquetColumnIndexSuite]
+    // Rewrite
```
rui-mo, line 619: Same.

yma11: This UT includes test against timestamp reading using Int128. Rewrite just remove this type.

rui-mo (approved):

rui-mo (approved):

Issue #3302, opened 2023-09-26
[VL] PR 3239 introduce a dead lock in our environment


Yohahaha:
### Backend

VL (Velox)

### Bug description

We found PR #3239 introduce a dead lock in our environment, but I'm not sure it can be reproduced in others env. I would try to investigate this issue.

### Spark version

None

### Relevant logs

```bash
Thread 827559: (state = BLOCKED)
 - org.apache.spark.shuffle.ColumnarShuffleWriter.<init>(org.apache.spark.shuffle.IndexShuffleBlockResolver, org.apache.spark.shuffle.BaseShuffleHandle, long, org.apache.spark.shuffle.ShuffleWriteMetricsReporter) @bci=286, line=83 (Interpreted frame)
 - org.apache.spark.shuffle.utils.ShuffleUtil$.genColumnarShuffleWriter(org.apache.spark.shuffle.GenShuffleWriterParameters) @bci=24, line=30 (Interpreted frame)
 - io.glutenproject.backendsapi.velox.SparkPlanExecHandler.genColumnarShuffleWriter(org.apache.spark.shuffle.GenShuffleWriterParameters) @bci=4, line=241 (Interpreted frame)
 - org.apache.spark.shuffle.GlutenShuffleWriterWrapper$.genColumnarShuffleWriter(org.apache.spark.shuffle.IndexShuffleBlockResolver, org.apache.spark.shuffle.sort.ColumnarShuffleHandle, long, org.apache.spark.shuffle.ShuffleWriteMetricsReporter) @bci=18, line=40 (Interpreted frame)
 - org.apache.spark.shuffle.sort.ColumnarShuffleManager.getWriter(org.apache.spark.shuffle.ShuffleHandle, long, org.apache.spark.TaskContext, org.apache.spark.shuffle.ShuffleWriteMetricsReporter) @bci=88, line=90 (Interpreted frame)
 - org.apache.spark.shuffle.ShuffleWriteProcessor.write(org.apache.spark.rdd.RDD, org.apache.spark.ShuffleDependency, long, org.apache.spark.TaskContext, org.apache.spark.Partition) @bci=29, line=57 (Interpreted frame)
 - org.apache.spark.scheduler.ShuffleMapTask.runTask(org.apache.spark.TaskContext) @bci=189, line=99 (Interpreted frame)
 - org.apache.spark.scheduler.ShuffleMapTask.runTask(org.apache.spark.TaskContext) @bci=2, line=52 (Interpreted frame)
[...]

PR #3303, opened 2023-09-26
[GLUTEN-3302][VL] Change all kinds of JniWrapper's create with required ExectionCtx to avoid race condition

Yohahaha:

Previous implementation need hold TaskResources' object lock in every JniWrapper#create method, but that method may be used in TaskResources internal, it cause race condition in some cases.

Fixes: \#3302

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala
@@ -18,6 +18,7 @@ package io.glutenproject.backendsapi.velox
 
 import io.glutenproject.GlutenNumaBindingInfo
 import io.glutenproject.backendsapi.IteratorApi
+import io.glutenproject.exec.ExecutionCtxs
 import io.glutenproject.execution._
 import io.glutenproject.metrics.IMetrics
 import io.glutenproject.substrait.plan.PlanNode
@@ -152,7 +153,7 @@ class IteratorHandler extends IteratorApi with Logging {
       new util.ArrayList[GeneralInIterator](inputIterators.map {
         iter => new ColumnarBatchInIterator(iter.asJava)
       }.asJava)
-    val transKernel = NativePlanEvaluator.create()
+    val transKernel = NativePlanEvaluator.create(ExecutionCtxs.contextInstance())
     val resIter: GeneralOutIterator =
       transKernel.createKernelWithBatchIterator(inputPartition.plan, columnarNativeIterators)
     pipelineTime += TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - beforeBuild)
@@ -201,7 +202,7 @@ class IteratorHandler extends IteratorApi with Logging {
 
     val beforeBuild = System.nanoTime()
 
-    val transKernel = NativePlanEvaluator.create()
+    val transKernel = NativePlanEvaluator.create(ExecutionCtxs.contextInstance())
     val columnarNativeIterator =
       new util.ArrayList[GeneralInIterator](inputIterators.map {
         iter => new ColumnarBatchInIterator(iter.asJava)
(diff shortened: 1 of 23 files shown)


backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala
```diff
@@ -201,7 +202,7 @@ class IteratorHandler extends IteratorApi with Logging {
 
     val beforeBuild = System.nanoTime()
 
-    val transKernel = NativePlanEvaluator.create()
+    val transKernel = NativePlanEvaluator.create(ExecutionCtxs.contextInstance())
```
zhztheplayer, line 205: The change doesn't seem to make real difference?

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -233,12 +233,13 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with SQLConfHe
...
         val nmm = NativeMemoryManagers
           .contextInstance("ColumnarCachedBatchSerializer read")
+        val execCtx = ExecutionCtxs.contextInstance()
         val schema = SparkArrowUtil.toArrowSchema(localSchema, timezoneId)
         val arrowAlloc = ArrowBufferAllocators.contextInstance()
         val cSchema = ArrowSchema.allocateNew(arrowAlloc)
         ArrowAbiUtil.exportSchema(arrowAlloc, schema, cSchema)
         val deserializerHandle = ColumnarBatchSerializerJniWrapper
-          .create()
+          .create(execCtx)
```
zhztheplayer, line 242: Is this change actually the one fixes the issue? Which stores a local variable of `ExecutionCtxs.contextInstance()`?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorHandler.scala
zhztheplayer: Since the current code `NativePlanEvaluator.create()` is just a shortcut to `NativePlanEvaluator.forCtx(ExecutionCtxs.contextInstance())`

Yohahaha:
I guess the root cause is related with broadcast relation, change all `create` methods just for quick fix to not block our internal release date.

If you dont accept this change and insist on provide `create` and `forCtx` methods, we could find the root cause too, but I still prefer provide only one static method for JniWrapper to keep readability, or refactor TaskResources to reduce lock scope as much as possible.

zhztheplayer:
>I still prefer provide only one static method for JniWrapper to keep readability

That's great to me, but the thing is issue #3302 is about deadlock, but I am feeling we don't make it clear which part of change fixes the deadlock? We must know exactly how the fix works before merging.

I won't block the code style changes as long as you think is reasonable. But let's find out how the patch fixes #3302 at first?

Yohahaha:
> > I still prefer provide only one static method for JniWrapper to keep readability
> 
> That's great to me, but the thing is issue #3302 is about deadlock, but I am feeling we don't make it clear which part of change fixes the deadlock? We must know exactly how the fix works before merging.
> 
> I won't block the code style changes as long as you think is reasonable. But let's find out how the patch fixes #3302 at first?

OK, I'm trying on that.

zhztheplayer:
> > > I still prefer provide only one static method for JniWrapper to keep readability
> > 
> > 
> > That's great to me, but the thing is issue #3302 is about deadlock, but I am feeling we don't make it clear which part of change fixes the deadlock? We must know exactly how the fix works before merging.
> > I won't block the code style changes as long as you think is reasonable. But let's find out how the patch fixes #3302 at first?
> 
> OK, I'm trying on that.

Thanks!

PR #3305, opened 2023-09-27
[GLUTEN-3302][VL] Fix dead lock of BackendJniWrapper static loading

Yohahaha:

BackendJniWrapper extend JniInitialized and JniInitializer invoke BackendJniWrapper#initializeBackend static method too, it will cause dead lock on static object lock.

PR #3929 change BackendJniWrapper#createExecutionCtx usage and let BackendJniWrapper init first in some case trigger this issue. Before 3929, JniInitializer's static lock was hold by other JniWrapper initialization which extend JniInitializer, it's safe to hold BackendJniWrapper static lock.

(Fixes: \#3302)

### DIFF
--- a/gluten-data/src/main/java/io/glutenproject/init/BackendJniWrapper.java
+++ b/gluten-data/src/main/java/io/glutenproject/init/BackendJniWrapper.java
@@ -16,13 +16,9 @@
  */
 package io.glutenproject.init;
 
-public class BackendJniWrapper extends JniInitialized {
+public class BackendJniWrapper {
 
   private BackendJniWrapper() {}
 
   public static native void initializeBackend(byte[] configPlan);
-
-  public static native long createExecutionCtx();
-
-  public static native void releaseExecutionCtx(long handle);
 }
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -315,7 +315,7 @@ void JNI_OnUnload(JavaVM* vm, void* reserved) {
   gluten::getJniCommonState()->close();
 }
 
-JNIEXPORT jlong JNICALL Java_io_glutenproject_init_ExecutionCtxJniWrapper_createExecutionCtx( // NOLINT
+JNIEXPORT jlong JNICALL Java_io_glutenproject_exec_ExecutionCtxJniWrapper_createExecutionCtx( // NOLINT
     JNIEnv* env,
     jclass) {
   JNI_METHOD_START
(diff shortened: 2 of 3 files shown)


Yohahaha:
@zhztheplayer

zhztheplayer:
> @zhztheplayer How about this?

Very convincing, thank you very much for catching.

gluten-data/src/main/java/io/glutenproject/init/BackendJniWrapper.java
```diff
@@ -16,13 +16,9 @@
  */
 package io.glutenproject.init;
 
-public class BackendJniWrapper extends JniInitialized {
+public class BackendJniWrapper {
```
zhztheplayer, line 19: I am feeling that we can remove this jni wrapper then move the method to class `JniInitialized`.

Then we can rename `JniInitialized` to `NativeBackendInitialized` or something to make things clearer.

How do you think about it?

Yohahaha: LGTM, let's do refactor in follow up patch.

I would like remove class hierarchy of JniInitialized too, expose a initBackend method and invoke this in ContextInitializer Object scala class, like below:
```scala
class ContextInitializer {
  override def initialize() = {
    init
    // ...
  }
}

object ContextInitializer {
  val init = JniInitialized.initBackend
}

```

zhztheplayer: > LGTM, let's do refactor in follow up patch.
> 
> I would like remove class hierarchy of JniInitialized too, expose a initBackend method and invoke this in ContextInitializer Object scala class, like below:
> 
> ```scala
> class ContextInitializer {
>   override def initialize() = {
>     init
>     // ...
>   }
> }
> 
> object ContextInitializer {
>   val init = JniInitialized.initBackend
> }
> ```
> 
> ContextInitializer's init variable will init in `<clinit>` as well as static initialization, thus only execute once.

Sounds good to me. Thanks.

zhztheplayer (approved):
Thanks!

zhouyuan:
@zhztheplayer @Yohahaha  it looks like the TPCDS test in github action is green when merging the previous patch, is it due to the scale factor too small? can we add a gate check on this? 

thanks, -yuan

Yohahaha:
> @zhztheplayer @Yohahaha it looks like the TPCDS test in github action is green when merging the previous patch, is it due to the scale factor too small? can we add a gate check on this?
> 
> thanks, -yuan

I found that issue in TPCDS sf1000 and can not repro every time, not sure how to add check.

zhztheplayer:
It's a random issue related to JVM's class loading order which is naturally randomized. Class loading happens only once typically in one JVM process that why it's more reproduceable in cluster. UT will hardly cover the case unless we enable forked processes to run UTs.

I think adding a GHA local-cluster job could help either and we can try adding one at a good time.

gluten-data/src/main/java/io/glutenproject/init/BackendJniWrapper.java
Yohahaha: #3307

PR #3307, opened 2023-09-27
[VL] Refine the way of initializing native backend

Yohahaha:

Changes:
1. move native backend initialization method from BackendJniWrapper to JniInitialized.
2. rename related file.
3. add ContextInitializer object class to guard native backend initialization to only execute once.
4. remove VeloxBackend#initUdf since these logic should under UdfJniWrapper


gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitialized.java
```diff
@@ -19,29 +19,29 @@
...
 
-  private static final Logger LOG = LoggerFactory.getLogger(JniInitialized.class);
+  private static final Logger LOG = LoggerFactory.getLogger(NativeBackendInitialized.class);
 
-  static {
+  public static void initializeBackend(scala.collection.Map<String, String> conf) {
     try {
       String prefix = BackendsApiManager.getSettings().getBackendConfigPrefix();
-      Map<String, String> nativeConfMap =
-          GlutenConfig.getNativeBackendConf(prefix, SQLConf.get().getAllConfs());
```
Yohahaha, line 37: I found `SQLConf.get().getAllConfig` return empty map in some test cases.

Yohahaha:
@zhztheplayer @marin-ma could you help take a look? thanks!

gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitialized.java
ulysses-you: it seems break the config propagation after moving the initialization to the beginning. For example, if people first call `spark.sql("set a=b")` then run query, then the native side can not aware of the config.

Yohahaha: The initialization here is for executor level and only pass need static config to VeloxBackend once, other dynamic config will be passed to native WholeStageIter in nativeCreateKernelWithIterator.

If `set a = b` is needed when initialize VeloxBackend, it should be added before intialize Spark session.

zhztheplayer: In addition, I suppose `set a = b` would cause an error when `a` is a "static conf key" in vanilla Spark. However in Gluten we didn't yet have a way to tell whether a Gluten config key is static or not. Probably we should somehow implement that.

gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitialized.java
```diff
@@ -19,29 +19,29 @@
...
-import org.apache.spark.sql.internal.SQLConf;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Map;
 
-// Initialize global contexts before calling any native methods from Java side.
-public abstract class JniInitialized {
+// Initialize native backend before calling any native methods from Java side.
+public class NativeBackendInitialized {
```
zhztheplayer, line 28: "NativeBackendInitialized" -> "NativeBackendInitializer" ?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ContextInitializer.scala
```diff
@@ -138,3 +142,14 @@ class ContextInitializer extends ContextApi {
...
+
+object ContextInitializer {
+  var init: Boolean = false
+  def initializeNative(conf: Map[String, String]): Unit = {
+    if (init) {
+      return
+    }
+    NativeBackendInitialized.initializeBackend(conf)
+    init = true
+  }
```
zhztheplayer, line 154: Looks like something needs to be guarded by a lock?

gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitialized.java
zhztheplayer: `public class` -> `public final class` ?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ContextInitializer.scala
zhztheplayer: BTW `AtomicBoolean` would be preferred if we want to avoid using non-final static variables.

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -251,8 +251,6 @@ void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf
     facebook::velox::serializer::presto::PrestoVectorSerde::registerVectorSerde();
   }
   velox::exec::Operator::registerOperator(std::make_unique<RowVectorStreamOperatorTranslator>());
-
-  initUdf(conf);
```
zhztheplayer, line 255: @marin-ma Would you please take a look at the change? Thanks!

gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitialized.java
Yohahaha: done.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ContextInitializer.scala
Yohahaha: It's safe here and impossible for multi thread invoke ContextInitializer#initializeNative.

gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitialized.java
ulysses-you: make sense to me, thank you

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ContextInitializer.scala
zhztheplayer: I think it's not enough to rely on outside code (e.g. from vanilla Spark) to guarantee the thread-safety here. Especially when driver and executor can be launched in a single JVM process in Spark's local mode. Would you please double-check?

Yohahaha: I modify these codes and verified when running UT, if run `NativeBackendInitialized.initializeBackend` twice in single JVM, it will core then exit. CI has prove its safety right?

zhztheplayer: > I modify these codes and verified when running UT, if run NativeBackendInitialized.initializeBackend twice in single JVM, it will core then exit. CI has prove its safety right?

Yes it probably can be considered safe in CI but I am not sure CI covers all cases. If it's hard to verify all the cases, I think it's OK to just add some guards to the code.

zhztheplayer: Or we can check vanilla Spark's code for its local mode. If the calls to the method strictly happens in the same thread, we can leave some comment here to explain why the code doesn't need to be guarded by lock.

Yohahaha: There are only two cases: local or cluster, let me double check the initialization of local mode.

Yohahaha: image
image

Verified and add comments.

zhztheplayer: OK. Thanks for making the change.

cpp/velox/compute/VeloxBackend.cc
Yohahaha: UDF related UT has passed, any comments?

Yohahaha: @marin-ma could you help confirm that?

marin-ma: @Yohahaha Could you follow the comment and run this suite? It failed in my local environment. https://github.com/oap-project/gluten/blob/00f83a40e753e3b33d8e9a38df40b48d22b8416e/backends-velox/src/test/scala/io/glutenproject/expression/VeloxUdfSuite.scala#L89-L100

Yohahaha: I revert changes of initUdf logic, it works.

The logic of load/register udf confused me lot...

Yohahaha:
@zhztheplayer @marin-ma please help review again, thanks!

marin-ma:
LGTM.

The timing of loading UDF libraries in Driver and Executors are different. Driver loads the libraries during applying SparkSessionExtension, but executors loads during backend initialization. The main difference is that driver need to register UDF signatures back to Spark (for SQL parsing) and to Velox (for validation) while executors only register the functions to Velox (for execution). Maybe next we can unify the timing of loading libraries to make the logic clearer.

zhztheplayer (approved):
Thanks!

PR #3308, opened 2023-09-27
[VL] Window transformer should require child ordering

ulysses-you:

OAP/Velox will skip sort for window, so we should always make sure the ordering requirements is satisfied to avoid potential issue.

## How was this patch tested?

PASS CI

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/WindowExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/WindowExecTransformer.scala
@@ -68,8 +68,14 @@ case class WindowExecTransformer(
     } else ClusteredDistribution(partitionSpec) :: Nil
   }
 
-  // We no longer require for sorted input for columnar window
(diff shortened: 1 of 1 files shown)


gluten-core/src/main/scala/io/glutenproject/execution/WindowExecTransformer.scala
```diff
@@ -68,8 +68,14 @@ case class WindowExecTransformer(
     } else ClusteredDistribution(partitionSpec) :: Nil
   }
 
-  // We no longer require for sorted input for columnar window
-  override def requiredChildOrdering: Seq[Seq[SortOrder]] = Seq.fill(children.size)(Nil)
+  override def requiredChildOrdering: Seq[Seq[SortOrder]] = {
```
zhouyuan, line 71: +1
Gluten replaced SMJ with SHJ, and replace SortAgg with HashAgg, will these bring some similar issue?

zhouyuan (approved):
👍

PR #3309, opened 2023-09-27
[VL] Wrong #numCols() result for offloaded columnar batch whose vector count is zero

zhztheplayer:
This will be a follow-up to https://github.com/oap-project/gluten/pull/3290, discussion https://github.com/oap-project/gluten/pull/3290#discussion_r1336963596.

Current method to offload zero-column batch doesn't result in real issues in tests, but it can still be risky since conflicted states about column count are created on the offloaded columnar batches.

The PR removes the risky code by adding a per-task cache for the zero-column batches to let the batch-offloading procedure pick instances directly from it. The cache will be released after task ends.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
@@ -261,7 +261,11 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with SQLConfHe
                 .deserialize(deserializerHandle, cachedBatch.bytes)
             val batch = ColumnarBatches.create(ExecutionCtxs.contextInstance(), batchHandle)
             if (shouldPruning) {
-              ColumnarBatches.select(nmm, batch, requestedColumnIndices.toArray)
+              try {
+                ColumnarBatches.select(nmm, batch, requestedColumnIndices.toArray)
+              } finally {
+                batch.close()
+              }
             } else {
               batch
             }
--- a/cpp/velox/compute/VeloxExecutionCtx.h
+++ b/cpp/velox/compute/VeloxExecutionCtx.h
@@ -73,6 +73,7 @@ class VeloxExecutionCtx final : public ExecutionCtx {
 
   ResourceHandle addBatch(std::shared_ptr<ColumnarBatch> ptr) override;
   std::shared_ptr<ColumnarBatch> getBatch(ResourceHandle handle) override;
+  ResourceHandle createOrGetEmptySchemaBatch(int32_t numRows) override;
   void releaseBatch(ResourceHandle handle) override;
   ResourceHandle select(MemoryManager* memoryManager, ResourceHandle batch, std::vector<int32_t> columnIndices)
       override;
@@ -142,6 +143,8 @@ class VeloxExecutionCtx final : public ExecutionCtx {
   ResourceMap<std::shared_ptr<ResultIterator>> resultIteratorHolder_;
 
   std::shared_ptr<const facebook::velox::core::PlanNode> veloxPlan_;
+
+  std::unordered_map<int32_t, ResourceHandle> emptySchemaBatchLoopUp_;
 };
 
 } // namespace gluten
--- a/gluten-data/src/main/java/io/glutenproject/vectorized/ColumnarBatchInIterator.java
+++ b/gluten-data/src/main/java/io/glutenproject/vectorized/ColumnarBatchInIterator.java
@@ -16,7 +16,9 @@
  */
 package io.glutenproject.vectorized;
 
+import io.glutenproject.columnarbatch.ColumnarBatchJniWrapper;
 import io.glutenproject.columnarbatch.ColumnarBatches;
+import io.glutenproject.memory.arrowalloc.ArrowBufferAllocators;
 
 import org.apache.spark.sql.vectorized.ColumnarBatch;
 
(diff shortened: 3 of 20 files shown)


zhztheplayer:
@ulysses-you

Would you like to take a look? Thanks.

cpp/velox/compute/VeloxExecutionCtx.h
```diff
@@ -142,6 +143,8 @@ class VeloxExecutionCtx final : public ExecutionCtx {
   ResourceMap<std::shared_ptr<ResultIterator>> resultIteratorHolder_;
 
   std::shared_ptr<const facebook::velox::core::PlanNode> veloxPlan_;
+
+  std::unordered_map<int32_t, ResourceHandle> emptySchemaBatchLoopUp_;
```
ulysses-you, line 147: can you point out to me, where do we release the resource ?

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -261,7 +261,11 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with SQLConfHe
                 .deserialize(deserializerHandle, cachedBatch.bytes)
             val batch = ColumnarBatches.create(ExecutionCtxs.contextInstance(), batchHandle)
             if (shouldPruning) {
-              ColumnarBatches.select(nmm, batch, requestedColumnIndices.toArray)
+              try {
+                ColumnarBatches.select(nmm, batch, requestedColumnIndices.toArray)
+              } finally {
+                batch.close()
```
ulysses-you, line 267: the batch is released at native side `select`, do we still need close it at jvm side ?

ulysses-you (commented):
thank you @zhztheplayer for the followup!

cpp/velox/compute/VeloxExecutionCtx.h
zhztheplayer: Batches should be released during destructing member `columnarBatchHolder_` from a JVM side call to release `ExecutionCtx` when task ends.

gluten-data/src/main/java/io/glutenproject/vectorized/ColumnarBatchInIterator.java
```diff
@@ -28,7 +30,13 @@ public ColumnarBatchInIterator(Iterator<ColumnarBatch> delegated) {
...
   public long next() {
-    final ColumnarBatch batch = nextColumnarBatch();
-    return ColumnarBatches.getNativeHandle(batch);
+    final ColumnarBatch next = nextColumnarBatch();
+    if (next.numCols() == 0) {
+      // the operation will find a zero column batch from a task-local pool
+      return ColumnarBatchJniWrapper.create().getForEmptySchema(next.numRows());
+    }
+    final ColumnarBatch offloaded =
+        ColumnarBatches.ensureOffloaded(ArrowBufferAllocators.contextInstance(), next);
```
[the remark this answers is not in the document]
zhztheplayer: I'll be OK not to call ensure but calling it looks like more reasonable since in-iterator is only used in native code.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
zhztheplayer: I see. I'd remove the release call from native code then let user to close the input batch manually. A batch-handle is usually mapped to a Java-side `IndicatorVector` which has a ref-cnt to manage the underlying buffers. Calling release from native code would break this kind of memory management.

gluten-data/src/main/java/io/glutenproject/vectorized/ColumnarBatchInIterator.java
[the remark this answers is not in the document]
zhztheplayer: I'd see benchmark report. I expect no visible overhead is brought since it's not a heavy operation if the batch was already offloaded.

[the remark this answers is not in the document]
ulysses-you: Thank you for the benchmark report. But I'm still worried for the sparse batch case, say, if each columnar batch only contain a few rows then the performance may be a little affected ?

cpp/velox/compute/VeloxExecutionCtx.h
ulysses-you: I did not see where release `emptySchemaBatchLoopUp_`.. do you mean it will be released when calling `delete executionCtx` ? it seems we did not override `~ExecutionCtx()`

zhztheplayer: > I did not see where release `emptySchemaBatchLoopUp_`.. do you mean it will be released when calling `delete executionCtx` ? it seems we did not override `~ExecutionCtx()`

You are right, we don't have to override dtor since the default dtor can handle it automatically.

gluten-data/src/main/java/io/glutenproject/vectorized/ColumnarBatchInIterator.java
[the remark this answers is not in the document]
zhztheplayer: Sounds reasonable to me. I am going to open a new issue for eliminating this overhead by adding some transaction-based operation for current utility methods in `ColumnarBatches`. Does that sound OK to you?

Besides, this sort of `ensureOffloaded` is required anyway although Velox backend doesn't need for it so far. Assuming we had another backend relying on `gluten-data` which could producing heavy batches for further native processing. The only thing is we should minimize the overhead for checking batch type.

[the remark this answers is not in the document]
zhztheplayer: https://github.com/oap-project/gluten/issues/3342

zhztheplayer:
@ulysses-you I'd use #3342 as follow-up.

zhouyuan (approved):
👍

ulysses-you (approved):
thank you @zhztheplayer !

PR #3311, opened 2023-09-27
[VL] Move session config to ExecutionCtx

Yohahaha:


Yohahaha:
@zhztheplayer please help take a look, thanks!

cpp/core/compute/ExecutionCtx.cc
```diff
@@ -20,28 +20,50 @@
...
+FactoryRegistry& executionCtxFactories() {
+  static FactoryRegistry registry;
+  return registry;
 }
+} // namespace
 
-void setExecutionCtxFactory(ExecutionCtxFactory factory) {
-  getExecutionCtxFactoryContext()->set(factory);
-  DEBUG_OUT << "Set execution context factory." << std::endl;
+void ExecutionCtx::registerFactory(const std::string& kind, ExecutionCtx::Factory factory) {
```
zhztheplayer, line 55: After the development iterations during the past few days I started to feel that `ExecutionCtx` is not a good name. Probably we can change to `Runtime` which has less conflictions with other parts of code in future.

cpp/core/jni/JniWrapper.cc
```diff
@@ -305,9 +305,11 @@ void JNI_OnUnload(JavaVM* vm, void* reserved) {
 
 JNIEXPORT jlong JNICALL Java_io_glutenproject_exec_ExecutionCtxJniWrapper_createExecutionCtx( // NOLINT
     JNIEnv* env,
-    jclass) {
+    jclass,
+    jbyteArray sessionConf) {
   JNI_METHOD_START
-  auto executionCtx = gluten::createExecutionCtx();
+  auto sparkConf = gluten::getConfMap(env, sessionConf);
+  auto executionCtx = gluten::ExecutionCtx::create(gluten::kVeloxExecutionCtxKind, sparkConf);
```
zhztheplayer, line 312: Velox-specific const `gluten::kVeloxExecutionCtxKind` appeared in common module?

gluten-data/src/main/java/io/glutenproject/exec/ExecutionCtxJniWrapper.java
```diff
@@ -20,7 +20,7 @@ public class ExecutionCtxJniWrapper {
 
   private ExecutionCtxJniWrapper() {}
 
-  public static native long createExecutionCtx();
+  public static native long createExecutionCtx(byte[] sessionConf);
```
zhztheplayer, line 23: Just as a TODO: We'd avoid relying on substrait's Protobuf definition for transmitting configs. We had added Gluten's own proto container `gluten/gluten-core/src/main/resources/io/glutenproject/proto`

zhztheplayer (changes requested):
Thanks! Just a few of comments.

zhztheplayer:
cc @PHILO-HE

cpp/core/compute/ExecutionCtx.cc
Yohahaha: both names are good for me as long as not change its responsibility.

Yohahaha:
could you help review again? @zhztheplayer @PHILO-HE

cpp/velox/compute/VeloxExecutionCtx.h
```diff
@@ -30,6 +30,8 @@
 
 namespace gluten {
 
+inline static const std::string kVeloxExecutionCtxKind{"velox"};
```
zhztheplayer, line 33: Please add a comment to note that the flag should be exactly the same to Java-side `VeloxBackend#getName()`

zhztheplayer (commented):
LGTM expect that a single comment is needed.

cpp/velox/compute/VeloxExecutionCtx.h
zhztheplayer: Or we can add a comment to the native registry API. Both is OK to me. This is to prevent the name mapping from being implicit to users.

Yohahaha: add a comment, thanks!

PHILO-HE (dismissed):
LGTM!

zhztheplayer (approved):
Thanks!

zhztheplayer:
@Yohahaha by the way, would you update the PR description? Thanks.

Yohahaha:
> @Yohahaha by the way, would you update the PR description? Thanks.

Done.

PR #3331, opened 2023-10-07
[VL] Add warn log when spill config not configured

Yohahaha:

Add warn log when spill-to-disk configuration is not configured.

## How was this patch tested?

No need

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -177,7 +177,9 @@ int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
     LOG(INFO) << logPrefix << "Successfully spilled out " << spilledOut << " bytes.";
     uint64_t total = shrunken + spilledOut;
     DLOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
-    return shrunken + spilledOut;
+    return total;
+  } else {
+    LOG(WARNING) << kSpillStrategy << " was not configured.";
   }
 
   DLOG(INFO) << logPrefix << "Successfully reclaimed total " << shrunken << " bytes.";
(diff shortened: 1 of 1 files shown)


cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -177,7 +177,9 @@ int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
     LOG(INFO) << logPrefix << "Successfully spilled out " << spilledOut << " bytes.";
     uint64_t total = shrunken + spilledOut;
     DLOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
-    return shrunken + spilledOut;
+    return total;
+  } else {
+    LOG(WARNING) << kSpillStrategy << " was not configured.";
```
zhztheplayer, line 182: "Skip spill-to-disk operation since .... was not configured"

Would this look better?

Yohahaha: ok.

zhztheplayer (approved):
Thanks!

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -177,7 +177,9 @@ int64_t WholeStageResultIterator::spillFixedSize(int64_t size) {
     LOG(INFO) << logPrefix << "Successfully spilled out " << spilledOut << " bytes.";
     uint64_t total = shrunken + spilledOut;
     DLOG(INFO) << logPrefix << "Successfully reclaimed total " << total << " bytes.";
-    return shrunken + spilledOut;
+    return total;
+  } else {
+    LOG(WARNING) << "Disable spill to disk since " << kSpillStrategy << " was not configured.";
```
zhztheplayer, line 182: > "Disable spill to disk since..." 

This still looks strange. Would you like to use "Spill-to-disk was disabled since..."?

zhztheplayer (changes requested):
Just another comment.

zhztheplayer (approved):
Thanks!

PR #3332, opened 2023-10-07
[GLUTEN-3328][VL]Add mutex for ResourceMap operators

JkSelf:

- Velox Parquet write operation now utilizes asynchronous writing for improved performance.  The batch will be placed into the VeloxWriteQueue in spark task. And a new thread will be spawned to retrieve the batch data, write it to the file, and close the batch after writing. Hence, we need to incorporate a locking mechanism in ResourceMap.

- Remove the twice close batch logic.

## How was this patch tested?

existing unit tests and jenkins tests.

### DIFF
--- a/cpp/velox/utils/ResourceMap.h
+++ b/cpp/velox/utils/ResourceMap.h
@@ -17,6 +17,7 @@
 
 #pragma once
 
+#include <mutex>
 #include "folly/container/F14Map.h"
 
 namespace gluten {
@@ -32,16 +33,19 @@ class ResourceMap {
   ResourceMap() : resourceId_(kInitResourceId) {}
 
   ResourceHandle insert(TResource holder) {
+    const std::lock_guard<std::mutex> lock(mtx_);
     ResourceHandle result = resourceId_++;
     map_.insert(std::pair<ResourceHandle, TResource>(result, holder));
     return result;
   }
 
   void erase(ResourceHandle moduleId) {
+    const std::lock_guard<std::mutex> lock(mtx_);
     map_.erase(moduleId);
   }
 
   TResource lookup(ResourceHandle moduleId) {
+    const std::lock_guard<std::mutex> lock(mtx_);
     auto it = map_.find(moduleId);
     if (it != map_.end()) {
       return it->second;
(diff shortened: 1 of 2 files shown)


JkSelf:
@FelixYBW @zhztheplayer @Yohahaha @zhouyuan  Please help to review. Thanks.

cpp/velox/utils/ResourceMap.h
```diff
@@ -33,15 +34,18 @@ class ResourceMap {
 
   ResourceHandle insert(TResource holder) {
     ResourceHandle result = resourceId_++;
+    const std::lock_guard<std::mutex> lock(mtx_);
```
Yohahaha, line 37: move lock at the beginning of method.

Yohahaha:
LGTM, thanks!

zhouyuan:
@JkSelf the previous patch does pass all GitHub action, is it convenient to add a check to ensure no regression in future? e.g. add one parquet write for velox backend

thanks, -yuan

zhztheplayer:
Slightly related to the topic, the member `veloxPlan_` probably breaks thread-safety of `ExecutionCtx` either. We might address that one in future. cc @Yohahaha

Yohahaha:
> Slightly related to the topic, the member `veloxPlan_` probably breaks thread-safety of `ExecutionCtx` either. We might address that one in future. cc @Yohahaha

ExecutionCtx were not expected to be thread-safe from my understanding, we could open a issue for that if needed.

zhztheplayer:
> > Slightly related to the topic, the member `veloxPlan_` probably breaks thread-safety of `ExecutionCtx` either. We might address that one in future. cc @Yohahaha
> 
> ExecutionCtx was not expected to be thread-safe from my understanding, we could open an discussion for that if needed.

Yes it was not designed to be thread-safe. But probably we need to change the design since issues like the one this PR addresses would emerge. I am not sure.

JkSelf:
> @JkSelf the previous patch does pass all GitHub action, is it convenient to add a check to ensure no regression in future? e.g. add one parquet write for velox backend
> 
> thanks, -yuan

@zhouyuan  Are you suggesting adding a unit test to replicate this issue?  It is difficult to reproduce the issue with small data sets; it only manifests when using 100GB of data. We can include performance testing for Parquet write in the Jenkins test before merging PR.

zhouyuan (approved):
:+1 works for me

zhouyuan:
@meharanjan318
Here's the docs on enabling parquet write:
https://github.com/oap-project/gluten/blob/main/docs/velox-backend-limitations.md#parquet-write

Also please check if this patch(#3332) is included, otherwise there will be a exception when writing. 

thanks, -yuan

PR #3336, opened 2023-10-07
[GLUTEN-3335] fix typo in ddlplan to dllplan

j7nhai:


zhztheplayer:
I assume "DDL" is the right spelling which stands for "Data Definition Language". @zzcclp may have comments here?

zzcclp:
> I assume "DDL" is the right spelling which stands for "Data Definition Language". @zzcclp may have comments here?

Yes, you are right, `DDL` is the right spelling. Or `DML` is OK.

BTW, the classes name in the package ` dllplan` are needed to modify ?

zhztheplayer:
So let's use `ddl` or `dml` uniformly? @zzcclp Which one would you prefer?

PR #3343, opened 2023-10-08
[VL] Pull data-source JNI wrapper up to common module

zhztheplayer:
Previously we mixed the use of Velox-related code in common JNI wrapper code. This is to clean up the code to make write API backend-independent.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/VeloxWriteQueue.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/VeloxWriteQueue.scala
@@ -16,7 +16,8 @@
  */
 package org.apache.spark.sql.execution.datasources
 
-import io.glutenproject.spark.sql.execution.datasources.velox.DatasourceJniWrapper
+import io.glutenproject.datasource.velox.DatasourceJniWrapper
+import io.glutenproject.vectorized.{CloseableColumnBatchIterator, ColumnarBatchInIterator}
 
 import org.apache.spark.sql.execution.datasources.VeloxWriteQueue.EOS_BATCH
 import org.apache.spark.sql.vectorized.ColumnarBatch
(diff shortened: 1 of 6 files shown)


backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/VeloxWriteQueue.scala
```diff
@@ -41,7 +45,9 @@ class VeloxWriteQueue(
   private val writeThread = new Thread(
     () => {
       try {
-        datasourceJniWrapper.write(dsHandle, scanner)
+        datasourceJniWrapper.write(
+          dsHandle,
+          new ColumnarBatchInIterator(new CloseableColumnBatchIterator(scanner).asJava))
       } catch {
         case e: Throwable =>
           writeException.set(e)
```
zhztheplayer, line 53: @JkSelf I think it's OK to remove all the write queue stuffs in future, since Velox's Parquet writer exposes a "write(rv)" API. Previously in Gazelle we had to pass a iterator to Arrow library for writing so that a queue was required.

zhztheplayer: Indeed, write performance might be impacted then. So some sort of PoCs will be needed.

JkSelf: @zhztheplayer If each batch calls a JNI function here, it may impact performance. However, I cannot recall the specific performance data at the moment. We can conduct a few POCs tests later to gather more information. Additionally, we are working on upgrading to Spark 3.4. With this update, we can directly incorporate a WriteTransformer to achieve Parquet write functionality, eliminating the need for these unnecessary lines of code.

zhztheplayer:
@JkSelf Please review, thanks.

JkSelf (approved):
LGTM. Thanks for your great work.

PR #3344, opened 2023-10-08
[VL] Improving VeloxShuffleWriter performance by just iterating valid partitions

zuochunwei:

The `numPartitions_` is set to `4096` for many jobs in our environment(`Meituan`), and the number of rows in a batch is also `4096`. 

The above configuration will result in some partitions not having rows, we need not iterate for all partitions(4096) but only the used partitions.

In order to avoid unnecessary iteration, I added a member data `partitionUsed` representing the partitions having values and then changed the loop based on all partitions to based on `partitionUsed`.

The performance will not be influenced if the `numPartitions_` configured a small value because the number of valid partitions will never be larger than the number of all partitions.

wait for a test under `meituan`'s environment, I will add the test result later.

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -362,7 +362,6 @@ arrow::Result<std::shared_ptr<VeloxShuffleWriter>> VeloxShuffleWriter::create(
   oss << " write_eos:" << options.write_eos;
   oss << " partition_writer_type:" << options.partition_writer_type;
   oss << " thread_id:" << options.thread_id;
-  oss << " offheap_per_task:" << options.offheap_per_task;
   LOG(INFO) << oss.str();
 #endif
   std::shared_ptr<VeloxShuffleWriter> res(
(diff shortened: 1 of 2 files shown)


cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -623,16 +622,22 @@ arrow::Status VeloxShuffleWriter::buildPartition2Row(uint32_t rowNum) {
...
+
+  // calc valid partition list
+  partitionUsed.clear();
+  for (auto pid = 0; pid != numPartitions_; ++pid) {
+    if (partition2RowCount_[pid] > 0) {
+      partitionUsed.push_back(pid);
+    }
+  }
+
+  std::cout << "[zcw] partitionUsed.size:" << partitionUsed.size() << " numPartitions_:" << numPartitions_ << std::endl;
```
marin-ma, line 640: Please remove this log, or change it to debug log if necessary.

zuochunwei: OK

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -42,7 +42,7 @@ using namespace facebook::velox;
 
 namespace gluten {
 
-#define VELOX_SHUFFLE_WRITER_LOG_FLAG 0
+#define VELOX_SHUFFLE_WRITER_LOG_FLAG 1
```
marin-ma, line 45: Please disable this flag.

zuochunwei: OK

zuochunwei:
@marin-ma ci passed

marin-ma (approved):

marin-ma:
Thanks! cc: @FelixYBW

zuochunwei:
We made a test for a big job, and the following is the result:

select the longest stage(21) of this job, the time cost(Median) is 7min vs 5.8min

test with this PR
image

test without this PR
image

marin-ma:
> We made a test for a big job, and the following is the result:
> 
> select the longest stage(21) of this job, the time cost(Median) is 7min vs 5.8min
> 
> test with this PR image
> 
> test without this PR image

@zuochunwei Is this the test result for the latest main branch, or an old version of main?

PR #3350, opened 2023-10-09
[VL] Run cpp test before maven test in CI

marin-ma:


zhztheplayer:
Sounds weird more or less to rely on Java code to generate data for C++ testing. May be we need to have a better design on this.

marin-ma:
> Sounds weird more or less to rely on Java code to generate data for C++ testing. Maybe we need to have a better design on that.

@zhztheplayer The substrait plan is generated by java code, and the generated json string may change. Do you have any suggestions?

zhztheplayer:
> > Sounds weird more or less to rely on Java code to generate data for C++ testing. Maybe we need to have a better design on that.
> 
> @zhztheplayer It is c++ micro benchmark that relies on Java test, not c++ UT. The substrait plan is generated by java code, and the generated json string may change. Do you have any suggestions?

No, so far so good. As long as the design doesn't block your work.

zhztheplayer (approved):

PR #3367, opened 2023-10-10
[VL] Improving performance by using simd gather instructions 

zuochunwei:

please don't review this PR now, wait for me to make a test.

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -83,6 +83,7 @@ add_compile_options(-Werror)
 add_compile_options(-Wno-error=parentheses)
 add_compile_options(-Wno-error=unused-function)
 add_compile_options(-Wno-error=unused-variable)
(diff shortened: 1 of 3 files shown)


zuochunwei:
The result does not conform to the expectation, and I will close later.

FelixYBW:
No benefit but no hurt either. We may keep it if you like

cpp/CMakeLists.txt
```diff
@@ -83,6 +83,7 @@ add_compile_options(-Werror)
 add_compile_options(-Wno-error=parentheses)
 add_compile_options(-Wno-error=unused-function)
 add_compile_options(-Wno-error=unused-variable)
+add_compile_options(-mavx2)
```
ulysses-you, line 86: macos arm64 does not support this option, can we only enable it for linux os ?

PR #3376, opened 2023-10-11
[Gluten-core][VL] Supports DeltaLake 2.2 Read

YannByron:

1. Supports Delta scan in Velox .
2. Delta 2.x supports Column Mapping, which is also supported in this PR.
3. Not support DeletionVector that is a new feature after Delta2.3

(Fixes: https://github.com/oap-project/gluten/issues/2891)

## How was this patch tested?

### DIFF
--- a/backends-velox/pom.xml
+++ b/backends-velox/pom.xml
@@ -185,6 +185,11 @@
       <version>1.3.0</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>io.delta</groupId>
+      <artifactId>delta-core_${scala.binary.version}</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/backends-velox/src/test/scala/io/glutenproject/execution/VeloxDeltaSuite.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/VeloxDeltaSuite.scala
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.execution
+
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.Row
+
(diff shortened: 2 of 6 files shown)


YannByron:
Follow https://github.com/oap-project/gluten/pull/2902

felipepessoto:
@YannByron do you have any suggestion how we can test how broken is Delta when running on Gluten/Velox?

I started an experiment, but then realized it doesn't follow the same pattern Gluten uses to test Spark. It seems we extends the existing classes and adds GlutenSQLTestsTrait, example: https://github.com/oap-project/gluten/pull/1381/files.
Should we do the same with Delta?


The experiment I started was to actually change Delta code to run the tests using Gluten:

1-) Cherry picked this to Delta 2.2. (this old workaround is not needed anymore and was messing the extension setting): https://github.com/delta-io/delta/commit/975daae690bce6295e5f8814f2515562553ed2ad
2-) Added gluten to the session, at DeltaSQLCommandTest.sparkConf:
```
  override protected def sparkConf: SparkConf = {
    super.sparkConf
      .set(StaticSQLConf.SPARK_SESSION_EXTENSIONS.key,
        classOf[DeltaSparkSessionExtension].getName)
      .set(SQLConf.V2_SESSION_CATALOG_IMPLEMENTATION.key,
        classOf[DeltaCatalog].getName)
      .set("spark.plugins", "io.glutenproject.GlutenPlugin")
      .set("spark.gluten.sql.columnar.backend.lib", "velox")
      .set("spark.memory.offHeap.enabled", "true")
      .set("spark.memory.offHeap.size", "10g")
      .set("spark.gluten.sql.columnar.forceshuffledhashjoin", "true")
      .set("spark.shuffle.manager", "org.apache.spark.shuffle.sort.ColumnarShuffleManager")
  }
```
3-) Added a reference to gluten jar in libraryDependencies for delta-core.

But it it not a complete solution, most of tests are failing or not running. And the ones which passes, I can't say if they are passing they Gluten falls back to non-native.

felipepessoto:
Found that _metadata is not working for Delta when column mapping is enabled.

_metadata has been recently changed to fall back: https://github.com/oap-project/gluten/issues/2618

yma11:
@YannByron will you work on update this PR based on [discussion](https://github.com/oap-project/gluten/issues/3378)?

YannByron:
Hi, @felipepessoto, sorry for the late reply.

I run TPCDS based on delta table with gluten/velox in EMR. I use Delta 2.2 only with some internal commits (there aren't related to gluten/velox), and add the delta jar in `$SPARK_HOME/jars`. And the configures are same with yours except that I don't use `spark.gluten.sql.columnar.backend.lib`. 
 
> Found that _metadata is not working for Delta when column mapping is enabled.

I'm not really aware of this _metadata. Where to use this during querying delta in gluten/velox env? If _metadata is used in parsing delta log, i think we can ignore for now because there are other factors (like UDFs) which make parsingdelt unsupported and the time cost for this phase was low.

YannByron:
@felipepessoto I test this case that query metadata on delta table that enable column mapping, and this can work correctly. Please check if there is this patch https://github.com/oap-project/gluten/pull/2563 in gluten you used. Cause gluten is not supported metaColumn, it will fallback the Scan to Spark.

YannByron:
And to support metadata column is tracked by https://github.com/oap-project/gluten/issues/2618.

PR #3384, opened 2023-10-11
[GLUTEN-3366][CH]Bug fix multi-thread unsafe problem for celeborn shuffle write

KevinyhZou:

(Fixes: \#3366)

## How was this patch tested?
manual tests

### DIFF
--- a/cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
+++ b/cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
@@ -313,7 +313,6 @@ size_t Partition::spill(DB::NativeWriter & writer)
         {
             raw_size += writer.write(block);
         }
-        spilled = true;
         return raw_size;
     }
     else
--- a/cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
+++ b/cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
@@ -230,29 +230,33 @@ void CelebornPartitionWriter::evictPartitions(bool for_memory_spill)
     auto spill_to_celeborn = [this]() -> void {
         Stopwatch serialization_time_watch;
         serialization_time_watch.start();
-        for (size_t partition_id = 0; partition_id < partition_buffer.size(); ++partition_id)
+        if (mtx.try_lock())
         {
-            const auto & partition = partition_buffer[partition_id];
-            size_t raw_size = 0;
-            if (partition.empty()) continue;
-            WriteBufferFromOwnString output;
-            auto codec = DB::CompressionCodecFactory::instance().get(boost::to_upper_copy(shuffle_writer->options.compress_method), {});
-            CompressedWriteBuffer compressed_output(output, codec, shuffle_writer->options.io_buffer_size);
-            NativeWriter writer(compressed_output, 0, shuffle_writer->output_header);
-            for (const auto & block : partition)
+            for (size_t partition_id = 0; partition_id < partition_buffer.size(); ++partition_id)
             {
-                raw_size += writer.write(block);
+                const auto & partition = partition_buffer[partition_id];
+                size_t raw_size = 0;
+                if (partition.empty()) continue;
+                WriteBufferFromOwnString output;
+                auto codec = DB::CompressionCodecFactory::instance().get(boost::to_upper_copy(shuffle_writer->options.compress_method), {});
+                CompressedWriteBuffer compressed_output(output, codec, shuffle_writer->options.io_buffer_size);
+                NativeWriter writer(compressed_output, 0, shuffle_writer->output_header);
+                for (const auto & block : partition)
+                {
+                    raw_size += writer.write(block);
+                }
+                compressed_output.sync();
+                Stopwatch push_time_watch;
+                push_time_watch.start();
+                celeborn_client->pushPartitionData(partition_id, output.str().data(), output.str().size());
+                shuffle_writer->split_result.partition_length[partition_id] += output.str().size();
+                shuffle_writer->split_result.raw_partition_length[partition_id] += raw_size;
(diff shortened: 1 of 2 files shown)


cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
```diff
@@ -230,29 +230,33 @@ void CelebornPartitionWriter::evictPartitions(bool for_memory_spill)
     auto spill_to_celeborn = [this]() -> void {
         Stopwatch serialization_time_watch;
         serialization_time_watch.start();
-        for (size_t partition_id = 0; partition_id < partition_buffer.size(); ++partition_id)
+        if (mtx.try_lock())
```
lgbo-ustc, line 233: use lock_guard

cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
```diff
@@ -271,10 +275,13 @@ void CelebornPartitionWriter::evictPartitions(bool for_memory_spill)
         spill_to_celeborn();
     }
     shuffle_writer->split_result.total_spill_time += spill_time_watch.elapsedNanoseconds();
-
-    for (auto & partition : partition_buffer)
+    if (mtx.try_lock())
```
taiyang-li, line 278: 建议使用std::lock_guard，参考https://www.cnblogs.com/joyer/p/15668695.html 
避免因提前退出时忘掉unlock导致死锁

taiyang-li (changes requested):

cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
KevinyhZou: done

cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
KevinyhZou: done

cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
```diff
@@ -228,6 +228,7 @@ CelebornPartitionWriter::CelebornPartitionWriter(CachedShuffleWriter * shuffleWr
 void CelebornPartitionWriter::evictPartitions(bool for_memory_spill)
 {
     auto spill_to_celeborn = [this]() -> void {
+        std::lock_guard<std::mutex> partition_spill_lock_guard(mtx);
```
zhanglistar, line 231: This lock scope is too large, need to just protect the race.

cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
```diff
@@ -271,10 +272,13 @@ void CelebornPartitionWriter::evictPartitions(bool for_memory_spill)
         spill_to_celeborn();
     }
     shuffle_writer->split_result.total_spill_time += spill_time_watch.elapsedNanoseconds();
-
-    for (auto & partition : partition_buffer)
+    if (!partition_buffer.empty())
```
taiyang-li, line 275: 这个判断也应该放在锁里。

taiyang-li (changes requested):

cpp-ch/local-engine/Shuffle/PartitionWriter.cpp
zhanglistar, line 275: Should, 1, lock 2,    ```
 for (auto & partition : partition_buffer)
    {
        partition.clear();
    }
```

Comment (KevinyhZou) [reply cpp-ch/local-engine/Shuffle/PartitionWriter.cpp]: done

**Review by liuneng1994 — APPROVED** [review]
LGTM

File: cpp-ch/local-engine/Shuffle/PartitionWriter.cpp [hunk 5]
```diff
@@ -300,5 +286,42 @@ void CelebornPartitionWriter::stop()
         shuffle_writer->split_result.total_bytes_written += item;
     }
 }
+
+void Partition::addBlock(DB::Block & block)
+{
+    std::unique_lock<std::mutex> lock(mtx, std::try_to_lock);
```
taiyang-li, line 289: 更常用的写法 
std::lock_guard<std::mutex> guard{mtx}; 
/// do what you want without owns_lock judgement.

KevinyhZou: if use std::lock_guard，it seems to lead to dead lock, as thread.join will suspend the main thread, if main thread hold the lock by lock_guard, but not finished, and the current thread try to get the lock which held by main thread, then the dead lock maybe appear. 

so i use try_to_lock, not to block if unable to get lock to avoid dead lock @taiyang-li

lgbo-ustc: This logic is not rgiht. If it doesn't get the lock, this block is not added in `blocks`

KevinyhZou: OK, i will look into this, but the test result is correct

lgbo-ustc:
The mutex has problems. At present, we notice that before the normal `evictPartitions` finish, the `stop` could be called by another thread.  In the case, this lock protection cannot avoid data inconsistent. explain as following
1. normal thread finisheds `evictPartitions`, but has not clear the `partition_buffer`. then the stop thread enter
2. stop thread add blocks and then call `evictPartitions`. Since the previous normal thread has not clear `partition_buffer`. This stop thread will flush duplicated  data into celeborn

lgbo-ustc:
Also I think we have chance to avoid lock here. For each thread, it has it own data to flush into celeborn, instead of use a shared `partition_buffer `

Issue #1283, opened 2023-04-04
[CORE] Throw error when a static config option gets changed


zhztheplayer:
Static config is not allowed to change in vanilla Spark's design.

However in Gluten the "backend conf" (distinguished with session conf) is designed in the similar concept with static config but is allowed to change without throwing error. We should make it act as Spark static conf either.

PR #3386, opened 2023-10-11
[GLUTEN-1283][VL] Distinguish static/dynamic configs introduced for backend

PHILO-HE:

Some configs should not allow users to change them during the session. We should make them static. And when getting session configs, static configs should be excluded.

(Fixes: \#1283)

## How was this patch tested?

Existing UTs.


zhztheplayer (approved):
Thanks!

PR #3394, opened 2023-10-12
[GLUTEN-3075][VL] Add getRowBytes(int32_t rowId) method in ColumnBatch

JkSelf:

Follow up https://github.com/oap-project/gluten/pull/3076.

Add getRowBytes(int32_t rowId) API in ColumnBatch.

## How was this patch tested?
Existing unit tests.


JkSelf:
@Yohahaha Can you help to review? Thanks.

Yohahaha:
How about move `getNonPartitionedColumnarBatch` to ColumnrBatch class too?

cpp/core/jni/JniWrapper.cc
```diff
@@ -1145,7 +1145,7 @@ Java_io_glutenproject_datasource_velox_DatasourceJniWrapper_splitBlockByPartitio
   }
   env->ReleaseIntArrayElements(partitionColIndice, pIndice, JNI_ABORT);
 
-  size = ctx->getColumnarBatchPerRowSize(batch);
+  size = batch->getColumnarBatchPerRowSize();
```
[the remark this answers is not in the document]
JkSelf: @zhztheplayer Updated. Thanks for your review.

JkSelf:
@Yohahaha @zhztheplayer Can you help to review again? Thanks.

Yohahaha:
LGTM, thanks!

cpp/core/memory/ColumnarBatch.h
```diff
@@ -47,7 +48,13 @@ class ColumnarBatch {
...
 
   virtual int64_t getExportNanos() const;
-  ;
+
+  virtual int32_t getRowSize() const;
+
+  virtual std::shared_ptr<ColumnarBatch> getNonPartitionedColumns(
+      std::vector<size_t> partitionColIndiceVec,
+      MemoryManager* memoryManager,
+      char*& rowBytes) const;
```
zhztheplayer, line 57: What's the purpose of this function?

zhztheplayer (changes requested):

cpp/core/memory/ColumnarBatch.h
JkSelf: @zhztheplayer Partition write only need to write the non partition columns into the file. This method is to retrieve the non partition columns  based on the partitionColIndiceVec.

zhztheplayer: We did add a `::select` API to ExecutionCtx. Is that one suitable for this work?

JkSelf: @zhztheplayer It appears that the ::select API can be useful for my purpose in ExecutionCtx. Spark needs to create the partition folder based on the first row in the column batch. Therefore, we only need to return the bytes of the first row to Spark, rather than the entire columnar batch. How about adding a getRowBytes method to ColumnarBatch specifically designed to retrieve the bytes of the first row?

zhztheplayer: If the output `getRowBytes` is compatible with Spark's UnsafeRow / InternalRow then it's OK to add it to `ColumnarBatch`.

The function should take row ordinal as its parameter, say `getRowBytes(int32_t rowId)` or something like that.

JkSelf: @zhztheplayer Yes. The getRowBytes method is compatible with Spark's UnsafeRow / InternalRow. I have made improvements by refining the method to getFirstRowBytes, which now only returns the bytes of the first row. Could you please review it again ? Thank you.

cpp/core/memory/ColumnarBatch.h
```diff
@@ -47,7 +48,8 @@ class ColumnarBatch {
   virtual std::shared_ptr<ArrowSchema> exportArrowSchema() = 0;
 
   virtual int64_t getExportNanos() const;
-  ;
+
+  virtual std::pair<char*, int> getFirstRowBytes() const;
```
zhztheplayer, line 52: Would it be better to use `getRowBytes(int32_t rowId)`?

Then we can change the calling code to `getRowBytes(0)`

JkSelf: @zhztheplayer Yes. Updated.

zhztheplayer: Thanks!

zhztheplayer (approved):
Thanks!

PR #3396, opened 2023-10-12
[GLUTEN-2845][VL] Separate jni method into different files

Yohahaha:
Changes:
1. Split methods of JniWrapper.cc into different `.cc` files, each new file's name is aligned with java file, only do movement without logic change. Keep `nativeValidateWithFailureReason` in VeloxJniWrapper, will refactor in follow up patch.
2. Move static jclass/jmethodId to JniCommonState.

After this patch, `core` dir structure will like below:
```
core
├── compute
│   ├── ExecutionCtxJniWrapper.cc
│   ├── PlanEvaluatorJniWrapper.cc
│   ├── batch
│   │   ├── ColumnarBatchJniWrapper.cc
│   │   └── ColumnarBatchOutIterJniWrapper.cc
│   ├── c2r
│   │   └── ColumnarToRowJniWrapper.cc
│   ├── r2c
│   │   └── RowToColumnarJniWrapper.cc
│   ├── serializer
│   │   └── ColumnarBatchSerializerJniWrapper.cc
│   ├── shuffle
│   └── rss
│   │   ├── ShuffleJniWrapper.cc
│   └── writer
│       └── DatasourceJniWrapper.cc
├── config
├── jni
├── memory
│   ├── MemoryJniWrapper.cc
```


Yohahaha:
This patch is ready. @zhztheplayer @ulysses-you @PHILO-HE

kecookier:
Why do this?

zuochunwei:
don't use a deep directory, we made this adjustment one year ago and added a rule in the cpp programming guide.
CPP is different from Java.

Yohahaha:
> don't use a deep directory, we made this adjustment one year ago and added a rule in the cpp programming guide. CPP is different from Java.

Main purpose is split JniWrapper, I prefer talk design in issue.

What the difference between `cpp/operators/c2r` and `cpp/compute/c2r` ? does second increase directory level?

zhztheplayer:
Probably the change is too big to match the volume of improvement it brings. Let's hold it off until the time is ripe? To avoid blocking other's pending work. @Yohahaha What do you think?

zhztheplayer:
> JniWrapper.cc already has 1372 lines

It is big but somewhat still organized. I think the complexity of maitaining it in future is basically predictable.

I agree that this sort of change improves the code but we should consider about the cost of applying the change. More concretely, a considerable amount of rebasing work will be needed in others' work if we merged the PR, also a lot of code lines' git author credit will be lost. So the cost of applying this change is high.

Yohahaha:
> > JniWrapper.cc already has 1372 lines
> 
> It is big but somewhat still organized. I think the complexity of maitaining it in future is basically predictable.
> 
> I agree that this sort of change improves the code but we should consider about the cost of applying the change. More concretely, a considerable amount of rebasing work will be needed in others' work if we merged the PR, also a lot of code lines' git author credit will be lost. So the cost of applying this change is high at the moment.

Make sense for me. Thanks!

zhztheplayer:
I am closing this. Feel free to reopen when time is ripe.

Thanks!

PR #3402, opened 2023-10-13
[VL] Branch upstream_velox: Untangle test case filter code

zhztheplayer:
@marin-ma


JkSelf:
@zhztheplayer Do we also need to update this 
`gluten-ut/spark33/src/test/scala/io/glutenproject/utils/velox/VeloxTestSettings.scala` file?

zhztheplayer:
cc @rui-mo

Issue #3267, opened 2023-09-25
[VL] Shuffle write size larger than vanilla spark


guhaiyan0221:
### Description

Conf:
spark.gluten.sql.columnar.shuffle.codec         zstd
spark.gluten.sql.columnar.shuffle.compressionMode    rowvector
spark.gluten.sql.columnar.shuffle.preferSpill     true

vanilla spark
![image](https://github.com/oap-project/gluten/assets/142767528/4b062131-c17f-4399-80ff-f8ff1929a599)

gluten
![image](https://github.com/oap-project/gluten/assets/142767528/48ab9b10-c3cd-4c93-90d9-59b02573a497)

Through more specific logging, we found that this is related to the size of the ;arge buffer eventually used for compression in rowvector mode.
{[471711,41789,11]}  means {[uncompressed buffer size, compressed buffer size, uncompressed/compressed]}
1. raw buffer size ~100k  
![image](https://github.com/oap-project/gluten/assets/142767528/dd460a15-118d-4258-b487-35a29617317c)

2. raw buffer size ~200k
![image](https://github.com/oap-project/gluten/assets/142767528/db434852-5330-451e-bd00-3e834d0388ff)

3. raw buffer size ~450k
![image](https://github.com/oap-project/gluten/assets/142767528/3478101d-29ea-4b1b-b0ac-5b3d3a84bacc)

marin-ma:
Thanks for the report. I have some questions first.
1. Is there any special reason to use `spark.gluten.sql.columnar.shuffle.preferSpill true` ? From the metrics, the shuffle bytes spilled is exactly equal to shuffle bytes written, which means there are lots of spills during shuffle write.
2. Have you tried `spark.gluten.sql.columnar.shuffle.compressionMode=buffer` and what's the compression ratio for that?
3. It would be nice if you can roughly describe the schema, like how many columns in total, how many columns of string?
4. From the metrics, the value of "shuffle bytes written total" doesn't match "local bytes read total + remote bytes read total". Do you know the reason?

cc: @FelixYBW

FelixYBW:
Can you print the raw data size of Vanilla Spark and Gluten? Let's check the raw data size

guhaiyan0221:
> Thanks for the report. I have some questions first.
> 
[...]

PR #3403, opened 2023-10-13
[VL] Shuffle splitter store string length instead of offset

marin-ma:
Issue: #3267 

If the data to compress has lots of duplicated values, both LZ4 and ZSTD perform high compression ratio( ~10 - 20). If no duplicated values, LZ4 can result in no compression and ZSTD compression ratio is ~2.

For fixed-length binary types, Spark UnsafeRow stores its value, offset in UnsafeRow and length. If the data schema consists of several columns of fixed-width data type and fixed-length string types, the offsets and lengths of a string field are identical for all rows. But currently, Gluten use arrow-like format for shuffle output, which stores value offsets in the string column and is considered unfriendly for data compression.

The metrics below compares the data size of splitting 2 columns of fixed-length binary type and 1 column of integer type. Before this PR, Gluten is 53.3 GB vs Vanilla Spark 47.9 GB. After this PR, Gluten is 46 GB.

Vanilla Spark shuffle bytes written = 47.9 GB:
image

Gluten shuffle bytes written Before = 53.3 GB:
image

Gluten shuffle bytes written After = 46 GB:
image

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleReader.cc
+++ b/cpp/velox/shuffle/VeloxShuffleReader.cc
@@ -118,20 +118,20 @@ VectorPtr readFlatVectorStringView(
     std::shared_ptr<const Type> type,
     memory::MemoryPool* pool) {
   auto nulls = buffers[bufferIdx++];
-  auto offsetBuffers = buffers[bufferIdx++];
-  auto valueBuffers = buffers[bufferIdx++];
-  const int32_t* rawLength = offsetBuffers->as<int32_t>();
+  auto offsetBuffer = buffers[bufferIdx++];
+  auto valueBuffer = buffers[bufferIdx++];
+  const int32_t* rawLength = offsetBuffer->as<int32_t>();
 
   std::vector<BufferPtr> stringBuffers;
   auto values = AlignedBuffer::allocate<char>(sizeof(StringView) * length, pool);
   auto rawValues = values->asMutable<StringView>();
-  auto rawChars = valueBuffers->as<char>();
+  auto rawChars = valueBuffer->as<char>();
   auto offset = 0;
   for (int32_t i = 0; i < length; ++i) {
     rawValues[i] = StringView(rawChars + offset, rawLength[i]);
     offset += rawLength[i];
   }
-  stringBuffers.emplace_back(valueBuffers);
+  stringBuffers.emplace_back(valueBuffer);
   if (nulls == nullptr || nulls->size() == 0) {
     return std::make_shared<FlatVector<StringView>>(
         pool, type, BufferPtr(nullptr), length, std::move(values), std::move(stringBuffers));
--- a/cpp/velox/shuffle/VeloxShuffleReader.cc
+++ b/cpp/velox/shuffle/VeloxShuffleReader.cc
@@ -118,9 +118,9 @@ VectorPtr readFlatVectorStringView(
     std::shared_ptr<const Type> type,
     memory::MemoryPool* pool) {
   auto nulls = buffers[bufferIdx++];
-  auto offsetBuffer = buffers[bufferIdx++];
+  auto lengthBuffer = buffers[bufferIdx++];
   auto valueBuffer = buffers[bufferIdx++];
-  const int32_t* rawLength = offsetBuffer->as<int32_t>();
+  const int32_t* rawLength = lengthBuffer->as<int32_t>();
 
   std::vector<BufferPtr> stringBuffers;
   auto values = AlignedBuffer::allocate<char>(sizeof(StringView) * length, pool);
--- a/cpp/velox/shuffle/VeloxShuffleReader.cc
+++ b/cpp/velox/shuffle/VeloxShuffleReader.cc
@@ -120,14 +120,16 @@ VectorPtr readFlatVectorStringView(
   auto nulls = buffers[bufferIdx++];
   auto offsetBuffers = buffers[bufferIdx++];
   auto valueBuffers = buffers[bufferIdx++];
-  const int32_t* rawOffset = offsetBuffers->as<int32_t>();
+  const int32_t* rawLength = offsetBuffers->as<int32_t>();
 
   std::vector<BufferPtr> stringBuffers;
   auto values = AlignedBuffer::allocate<char>(sizeof(StringView) * length, pool);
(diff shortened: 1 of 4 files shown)


zuochunwei:
what benefits will it bring? performance gains or other considerations?

marin-ma:
> what benefits will it bring? performance gains or other considerations?

Please check the updated PR description.

zuochunwei:
good job

marin-ma:
@guhaiyan0221 Can you try this patch to compare the "shuffle bytes written" metrics in your case?

zhouyuan:
Make sense to me. A fixed length string should be common, e.g., "20231010", in this case, the length is always 8, while the offsets would be 0, 7, 15, 23....
if the string is variable length, I assume this patch is not that beneficial?

thanks, -yuan

marin-ma:
> Make sense to me. A fixed length string should be common, e.g., "20231010", in this case, the length is always 8, while the offsets would be 0, 7, 15, 23.... if the string is variable length, I assume this patch is not that beneficial?
> 
> thanks, -yuan

Yes. Based on the observed data pattern and compression ratio, it's also impossible to get such high compression ratio in Vanilla Spark if the string is variable length.

cpp/velox/shuffle/VeloxShuffleReader.cc
```diff
@@ -117,18 +117,20 @@ VectorPtr readFlatVectorStringView(
     std::shared_ptr<const Type> type,
     memory::MemoryPool* pool) {
   auto nulls = buffers[bufferIdx++];
-  auto offsetBuffers = buffers[bufferIdx++];
-  auto valueBuffers = buffers[bufferIdx++];
-  const int32_t* rawOffset = offsetBuffers->as<int32_t>();
+  auto offsetBuffer = buffers[bufferIdx++];
```
zhztheplayer, line 120: offsetBuffer -> lengthBuffer ?

FelixYBW (approved):

FelixYBW:
Good catch!

Why the difference of gluten data size total vs. uncompressed data size total is so large? 185G vs. 112G.

Can you do an experiment to use uint6_t instead of uint32_t to save the offset buffer? uncompressed data size should be even less but let's see any difference on compressed data size

marin-ma:
@FelixYBW The "data size total" counts the input Velox RowVector size. The data size from table scan is 161GB, partial agg add an extra column (177G), and project add an extra column for partition id that makes the shuffle input data size 185G.

image

guhaiyan0221:
> @guhaiyan0221 Can you try this patch to compare the "shuffle bytes written" metrics in your case?

We have made similar change internally for https://github.com/oap-project/gluten/issues/2343. 
image

Issue #3267 was found after offset->length storing changed.

zhztheplayer (dismissed):

FelixYBW (approved):

FelixYBW:
> @FelixYBW The "data size total" counts the input Velox RowVector size. The data size from table scan is 161GB, partial agg add an extra column (177G), and project add an extra column for partition id that makes the shuffle input data size 185G.
> 
> image

Marked here: so the data set size before split is 177GB, which is Velox format and counted by buffer capacity instead of exact buffer size. 112G is the buffer size after split, which we actually use arrow's string buffer format, instead of Velox stringview. The difference comes from buffer capacity vs. buffer size and stringview vs. stringvalue buffer

marin-ma:
@FelixYBW I will open another PR for copy-compress small buffers.

zhztheplayer (approved):

cpp/velox/shuffle/VeloxShuffleReader.cc
```diff
@@ -118,18 +118,20 @@ VectorPtr readFlatVectorStringView(
...
+  auto lengthBuffer = buffers[bufferIdx++];
+  auto valueBuffer = buffers[bufferIdx++];
+  const int32_t* rawLength = lengthBuffer->as<int32_t>();
 
   std::vector<BufferPtr> stringBuffers;
   auto values = AlignedBuffer::allocate<char>(sizeof(StringView) * length, pool);
   auto rawValues = values->asMutable<StringView>();
-  auto rawChars = valueBuffers->as<char>();
+  auto rawChars = valueBuffer->as<char>();
+  auto offset = 0;
```
FelixYBW, line 129: Let's use uint64_t as offset

cpp/velox/shuffle/VeloxShuffleReader.cc
```diff
@@ -118,18 +118,20 @@ VectorPtr readFlatVectorStringView(
     std::shared_ptr<const Type> type,
     memory::MemoryPool* pool) {
   auto nulls = buffers[bufferIdx++];
-  auto offsetBuffers = buffers[bufferIdx++];
-  auto valueBuffers = buffers[bufferIdx++];
-  const int32_t* rawOffset = offsetBuffers->as<int32_t>();
+  auto lengthBuffer = buffers[bufferIdx++];
+  auto valueBuffer = buffers[bufferIdx++];
+  const int32_t* rawLength = lengthBuffer->as<int32_t>();
```
FelixYBW, line 123: change it to uint32_t

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -469,17 +469,16 @@ void collectFlatVectorBufferStringView(
...
   // last offset is the totalStringSize
-  auto offsetBufferSize = sizeof(int32_t) * (flatVector->size() + 1);
-  GLUTEN_ASSIGN_OR_THROW(auto offsetBuffer, arrow::AllocateResizableBuffer(offsetBufferSize, pool));
-  int32_t* rawOffset = reinterpret_cast<int32_t*>(offsetBuffer->mutable_data());
-  // first offset is 0
-  *rawOffset++ = 0;
+  auto lengthBufferSize = sizeof(int32_t) * flatVector->size();
+  GLUTEN_ASSIGN_OR_THROW(auto lengthBuffer, arrow::AllocateResizableBuffer(lengthBufferSize, pool));
+  int32_t* rawLength = reinterpret_cast<int32_t*>(lengthBuffer->mutable_data());
   int32_t offset = 0;
```
FelixYBW, line 475: let's change to uint64_t to support string buffer larger than 4G

PR #3414, opened 2023-10-16
[VL] Extract ORC reader compilation from GenericBenchmark

marin-ma:
Add `-DENABLE_ORC` cmake option. Disabled by default.


rui-mo:
Thanks!

PR #3420, opened 2023-10-17
[VL] Copy compress small partition buffer

marin-ma:
1. Currently, buffer size < 1024B (default threshold) will not be compressed. If partition buffers are small (e.g. < 100 rows), it's likely that the data is not compressed and result in the shuffle data size larger than Vanilla spark. This PR change the default threshold unit and make the default threshold value 100 rows. Partition buffers that fall below the threshold will be copied into one buffer and compress.
2. Change Columnar shuffle metrics "data size total" from collecting input Velox RowVector size to uncompressed size and remove the "uncompressed size" metric.

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -802,8 +802,10 @@ object GlutenConfig {
   val COLUMNAR_SHUFFLE_BUFFER_COMPRESS_THRESHOLD =
     buildConf("spark.gluten.sql.columnar.shuffle.bufferCompressThreshold")
       .internal()
+      .doc("If number of rows in a batch falls below this threshold," +
+        " will copy all buffers into one buffer to compress.")
       .intConf
-      .createWithDefault(1024)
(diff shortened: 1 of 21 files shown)


FelixYBW (approved):

marin-ma:
Test result for using int8_t/int16_t as type of binary type lengthBuffer:
Conclusion: Using int8_t/int16_t reduces uncompressed size a little, but has no effect on data size(compressed size).

int8_t
image

int16_t
image

int32_t
image

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -802,8 +802,10 @@ object GlutenConfig {
   val COLUMNAR_SHUFFLE_BUFFER_COMPRESS_THRESHOLD =
     buildConf("spark.gluten.sql.columnar.shuffle.bufferCompressThreshold")
       .internal()
+      .doc("If number of rows in a batch falls below this threshold," +
+        " will copy all buffers into one buffer to compress.")
```
FelixYBW, line 806: do we need to update document?

PR #3439, opened 2023-10-19
[VL] Increase kAbandonPartialAggregationMinRows

rui-mo:

To avoid partial aggreagation flushing on a low cardinality case.

## How was this patch tested?

Verified on TPC-DS.


zhouyuan (approved):

rui-mo:
@zhouyuan This change brings TPC-DS perf back nearly, but does not improve TPC-H.

zhouyuan:
> @zhouyuan This change brings TPC-DS perf back nearly, but does not improve TPC-H.

Thanks, yes, for H the issue is different, mostly on hash join I think.

PR #3442, opened 2023-10-19
[VL] Remove unnecessary registerConnectorFactory for HiveConnector

Yohahaha:
```
W1019 12:38:42.899380 996562 Connector.cpp:39] ConnectorFactory with name hive is already registered.
```

### DIFF
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -231,7 +231,6 @@ void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf
 #endif
 
   auto properties = std::make_shared<const velox::core::MemConfig>(configurationValues);
-  velox::connector::registerConnectorFactory(std::make_shared<velox::connector::hive::HiveConnectorFactory>());
   auto hiveConnector =
       velox::connector::getConnectorFactory(velox::connector::hive::HiveConnectorFactory::kHiveConnectorName)
           ->newConnector(kHiveConnectorId, properties, ioExecutor_.get());


Yohahaha:
@rui-mo could you help take a look? thanks!

rui-mo (commented):
Thanks for trying to fix this issue. We found on some environment, the factory will not be registered if without this line, but on the others, duplicate registration is noticed with it. This function should be called in [link](https://github.com/oap-project/velox/blob/3c3151235d91c57f0e0698c705a8d89cd487bd6a/velox/connectors/Connector.h#L452-L456). 
@zhztheplayer Is the factory still not registered without this line on your env?

zhztheplayer:
> @zhztheplayer Is the factory still not registered without this line on your env?

I used to experience a similar issue that the factory was not registered by Velox's static initializer. My intuition is the static initializers are not always reliable but indeed we can test against them. Let's do such optimization very carefully with plenty testing.

Yohahaha:
Any concern?

Yohahaha:
Previous static initializer issue has been refined in https://github.com/oap-project/gluten/pull/3307 https://github.com/oap-project/gluten/pull/3305. If remove such codes cause problem, I think it would be great that we found another implementation mistake, find issue then fix it, right?

rui-mo (approved):

rui-mo:
Velox change could be reverted with https://github.com/oap-project/velox/pull/422.

zhztheplayer:
> Previous static initializer issue has been refined in #3307 #3305. If remove such codes cause problem, I think it would be great that we found another implementation mistake, find issue then fix it, right? Expose problem always better than mystic logic for me.

Sounds convincing. And I couldn't remember how the previous issue was triggered so let's just apply and see what happens.

rz-vastdata:
Greetings,

We are seeing similar behaviour when adding a new connector to Velox - `VELOX_REGISTER_CONNECTOR_FACTORY` seems to not be included in the `libvelox.so` so `registerConnectorFactory()` is not called for our connector.

We are currently using Gluten 1.0 with https://github.com/oap-project/velox/tree/branch-1.0.

Is there a fix for this issue? We found https://github.com/oap-project/velox/commit/aadd693b0510ece0ff743bbd2e3208c2f6ebb264 but it seems to be reverted by https://github.com/oap-project/velox/commit/95deb2f027e7419b6411b5b48afa4c8358bb53bc.

All the best.

> > @zhztheplayer Is the factory still not registered without this line on your env?
> 
> I used to experience a similar issue that the factory was not registered by Velox's static initializer. My intuition is the static initializers are not always reliable but indeed we can test against them. Let's do such optimization very carefully with plenty testing.

Yohahaha:
Do you mean Gluten 1.0 has `connector not registered` bug? @rz-vastdata

rz-vastdata:
> Do you mean Gluten 1.0 has `connector not registered` bug?

Yes - when we add a new Velox connector, and register its factory with `VELOX_REGISTER_CONNECTOR_FACTORY` macro.

rz-vastdata:
Sounds good, thanks!
BTW, could you please elaborate why https://github.com/oap-project/gluten/commit/2085835a5b073b14ea7effa84e2e38b74206e243 resolves the issue? 
Is there a linker optimization that may cause this problem?

rui-mo:
@rz-vastdata Previously, we link the static library, in which the unused symbols will not be linked in. After this change, the unused symbol is linked in, and the registration can be ensured to be executed.

PR #3449, opened 2023-10-20
[VL] Add stream insertion operator for SparkTaskInfo on native side

Yohahaha:
Overload `<<` operator for SparkTaskInfo, after that, it can be used as `LOG(INFO) << taskInfo;`, hence, build mapping between java log and native log.
```
I1020 10:07:57.491497 2704501 VeloxExecutionCtx.cc:82] [Stage: 9 TID: 9]
23/10/20 10:07:57 WARN [Executor task launch worker for task 0.0 in stage 9.0 (TID 9)] xxx
```

Also pass it to WholeStageIterator as member variable.

### DIFF
--- a/cpp/core/compute/Runtime.h
+++ b/cpp/core/compute/Runtime.h
@@ -42,6 +42,10 @@ struct SparkTaskInfo {
   // Same as TID.
   int64_t taskId{0};
 
+  std::string toString() const {
+    return "[Stage: " + std::to_string(stageId) + " TID: " + std::to_string(taskId) + "]";
+  }
+
   friend std::ostream& operator<<(std::ostream& os, const SparkTaskInfo& taskInfo) {
     os << "[Stage: " << taskInfo.stageId << " TID: " << taskInfo.taskId << "]";
     return os;
--- a/cpp/core/compute/Runtime.h
+++ b/cpp/core/compute/Runtime.h
@@ -42,8 +42,9 @@ struct SparkTaskInfo {
   // Same as TID.
   int64_t taskId{0};
 
-  std::string toString() const {
-    return "[Stage: " + std::to_string(stageId) + " TID: " + std::to_string(taskId) + "]";
+  friend std::ostream& operator<<(std::ostream& os, const SparkTaskInfo& taskInfo) {
+    os << "[Stage: " << taskInfo.stageId << " TID: " << taskInfo.taskId << "]";
+    return os;
   }
 };
 
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -81,8 +81,9 @@ const std::string kHiveDefaultPartition = "__HIVE_DEFAULT_PARTITION__";
 WholeStageResultIterator::WholeStageResultIterator(
     std::shared_ptr<facebook::velox::memory::MemoryPool> pool,
     const std::shared_ptr<const facebook::velox::core::PlanNode>& planNode,
-    const std::unordered_map<std::string, std::string>& confMap)
-    : veloxPlan_(planNode), confMap_(confMap), pool_(pool) {
+    const std::unordered_map<std::string, std::string>& confMap,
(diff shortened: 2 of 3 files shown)


Yohahaha:
@rui-mo @zhztheplayer @PHILO-HE could someone help take a look? thanks!

cpp/velox/compute/WholeStageResultIterator.h
```diff
@@ -69,6 +70,8 @@ class WholeStageResultIterator : public ColumnarBatchIterator {
   /// A map of custom configs.
   std::unordered_map<std::string, std::string> confMap_;
 
+  const SparkTaskInfo taskInfo_;
```
rui-mo, line 73: What is main purpose to keep the task info? Seems previous temp variable is enough.

Yohahaha: It help use task info in other place not only in ctor.

cpp/core/compute/Runtime.h
```diff
@@ -42,8 +42,9 @@ struct SparkTaskInfo {
   // Same as TID.
   int64_t taskId{0};
 
-  std::string toString() const {
```
PHILO-HE, line 45: Thanks for your PR!
I think it may be better to keep `toString`. Then it can be called from other places, as well as from the implemented `<<` operator . Right?

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -500,14 +501,14 @@ WholeStageResultIteratorMiddleStage::WholeStageResultIteratorMiddleStage(
...
-    : WholeStageResultIterator(pool, planNode, confMap), streamIds_(streamIds) {
+    const SparkTaskInfo& taskInfo)
+    : WholeStageResultIterator(pool, planNode, confMap, taskInfo), streamIds_(streamIds) {
   std::unordered_set<velox::core::PlanNodeId> emptySet;
   velox::core::PlanFragment planFragment{planNode, velox::core::ExecutionStrategy::kUngrouped, 1, emptySet};
   std::shared_ptr<velox::core::QueryCtx> queryCtx = createNewVeloxQueryCtx();
 
   task_ = velox::exec::Task::create(
-      fmt::format("Gluten stage-{} task-{}", taskInfo.stageId, taskInfo.taskId),
+      fmt::format("Gluten stage-{} task-{}", taskInfo_.stageId, taskInfo_.taskId),
```
PHILO-HE, line 511: Nit: Directly use `toString()` if kept?

PHILO-HE (commented):
Just few trivial comment. Thanks!

cpp/core/compute/Runtime.h
Yohahaha: I keep toString and add new overload `<<` operator.

PHILO-HE (approved):
Looks good! Thanks!

PHILO-HE:
Just changed the title with "insertion operator". Suppose extraction operator means `>>`.

Yohahaha:
> Just changed the title with "insertion operator". Suppose extraction operator means `>>`.

hmm... I refer this doc https://en.cppreference.com/w/cpp/language/operators

```
The overloads of operator>> and operator<< that take a [std::istream](http://en.cppreference.com/w/cpp/io/basic_istream)& or [std::ostream](http://en.cppreference.com/w/cpp/io/basic_ostream)& as the left hand argument are known as insertion and extraction operators.
```

PHILO-HE:
> > Just changed the title with "insertion operator". Suppose extraction operator means `>>`.
> 
> hmm... I refer this doc https://en.cppreference.com/w/cpp/language/operators
> 
> ```
> The overloads of operator>> and operator<< that take a [std::istream](http://en.cppreference.com/w/cpp/io/basic_istream)& or [std::ostream](http://en.cppreference.com/w/cpp/io/basic_ostream)& as the left hand argument are known as insertion and extraction operators.
> ```

It may be wrong. Anyway, let's use insertion operator instead. Not a big issue. Thanks!

PR #3454, opened 2023-10-20
[VL] Add config for glog verbose level and severity level

Yohahaha:
Add new config `spark.gluten.sql.columnar.backend.velox.glogVerboseLevel` to control native glog verbose level and `spark.gluten.sql.columnar.backend.velox.glogSeverityLevel` to control native glog severity, default value is same as glog, [doc](https://github.com/google/glog).
By this config, user can easily control both Gluten and Velox log level at runtime.

Also change DLOG to VLOG to enable such logs without recompile, more bigger number in VLOG means more detailed message.

### DIFF
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -60,6 +60,9 @@ const std::string kEnableUserExceptionStacktrace =
     "spark.gluten.sql.columnar.backend.velox.enableUserExceptionStacktrace";
 const std::string kEnableUserExceptionStacktraceDefault = "true";
 
+const std::string kGlogVerboseLevel = "spark.gluten.sql.columnar.backend.velox.glogVerboseLevel";
+const std::string kGlogVerboseLevelDefault = "0";
+
 const std::string kEnableSystemExceptionStacktrace =
     "spark.gluten.sql.columnar.backend.velox.enableSystemExceptionStacktrace";
 const std::string kEnableSystemExceptionStacktraceDefault = "true";
@@ -116,6 +119,12 @@ void VeloxBackend::printConf(const std::unordered_map<std::string, std::string>&
 }
 
 void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf) {
+  // Init glog and verbose level.
+  auto vlogLevel = stoi(getConfigValue(conf, kGlogVerboseLevel, kGlogVerboseLevelDefault));
+  FLAGS_v = vlogLevel;
+  FLAGS_logtostderr = true;
+  google::InitGoogleLogging("gluten");
+
   // Avoid creating too many shared leaf pools.
   FLAGS_velox_memory_num_shared_leaf_pools = 0;
 
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -63,6 +63,9 @@ const std::string kEnableUserExceptionStacktraceDefault = "true";
 const std::string kGlogVerboseLevel = "spark.gluten.sql.columnar.backend.velox.glogVerboseLevel";
 const std::string kGlogVerboseLevelDefault = "0";
 
+const std::string kGlogSeverityLevel = "spark.gluten.sql.columnar.backend.velox.glogMinLogLevel";
+const std::string kGlogSeverityLevelDefault = "0";
+
 const std::string kEnableSystemExceptionStacktrace =
     "spark.gluten.sql.columnar.backend.velox.enableSystemExceptionStacktrace";
 const std::string kEnableSystemExceptionStacktraceDefault = "true";
@@ -119,11 +122,15 @@ void VeloxBackend::printConf(const std::unordered_map<std::string, std::string>&
 }
 
 void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf) {
-  // Init glog and verbose level.
-  auto vlogLevel = stoi(getConfigValue(conf, kGlogVerboseLevel, kGlogVerboseLevelDefault));
-  FLAGS_v = vlogLevel;
-  FLAGS_logtostderr = true;
-  google::InitGoogleLogging("gluten");
+  // Init glog and log level.
+  {
+    auto vlogLevel = stoi(getConfigValue(conf, kGlogVerboseLevel, kGlogVerboseLevelDefault));
+    auto severityLogLevel = stoi(getConfigValue(conf, kGlogSeverityLevel, kGlogSeverityLevelDefault));
+    FLAGS_v = vlogLevel;
+    FLAGS_minloglevel = severityLogLevel;
+    FLAGS_logtostderr = true;
+    google::InitGoogleLogging("gluten");
+  }
 
   // Avoid creating too many shared leaf pools.
   FLAGS_velox_memory_num_shared_leaf_pools = 0;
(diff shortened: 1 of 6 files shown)


ulysses-you:
+1, can we use the log level of log4j as the default glog level ?

Yohahaha:
> +1, can we use the log level of log4j as the default glog level ?

I have no idea about this cause log4j can set level for different module, could you describe expected usage?

ulysses-you:
I mean we should follow the default log level of log4j. If it is warn then the default level of glog should be warn. It is just a config mapping.

Yohahaha:
> I mean we should follow the default log level of log4j. If it is warn then the default level of glog should be warn. It is just a config mapping.

I have less knowledge about log4j, but use log4j's config to control native log is not make sense for me.

Yohahaha:
@rui-mo @zhouyuan could you help take a look? thanks!

zhouyuan (dismissed):
👍

ulysses-you:
> but use log4j's config to control native log is not make sense for me.

Why does not it make sense ? Say, if people only want to see the warn log, then he should change native config since he has changed the log4j config. If we can mapping the default log level of log4j to native side, then people do not need to change anything.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -1241,4 +1241,11 @@ object GlutenConfig {
         + "partial aggregation may be early abandoned.")
       .intConf
       .createOptional
+
+  val COLUMNAR_VELOX_GLOG_VERBOSE_LEVEL =
+    buildConf("spark.gluten.sql.columnar.backend.velox.glogVerboseLevel")
+      .internal()
+      .doc("Set glog verbose level in Velox backend.")
```
rui-mo, line 1248: Can we add some details for this config? Mainly about how should the user set this config if wanting to change the level to INFO/WARN/ERROR etc.

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -116,6 +119,12 @@ void VeloxBackend::printConf(const std::unordered_map<std::string, std::string>&
 }
 
 void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf) {
+  // Init glog and verbose level.
+  auto vlogLevel = stoi(getConfigValue(conf, kGlogVerboseLevel, kGlogVerboseLevelDefault));
```
rui-mo, line 123: We need to send this config from Scala side as [GlutenConfig.scala#L410-L458](https://github.com/oap-project/gluten/blob/main/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala#L410-L458), then it can extracted here, right?

Yohahaha:
```
log4j.rootCategory=INFO, console
log4j.logger.io.glutenproject=DEBUG
log4j.logger.io.glutenproject.vectorized.JniLibLoader=INFO
```
log4j can be set by module, I think there are no 'default' config.

if native use `native-log4j` it would be nice to use same config, give fine-grained config back to user is make sense for me.

> if people only want to see the warn log, then he should change native config since he has changed the log4j config. If we can mapping the default log level of log4j to native side, then people do not need to change anything.

I think user should set log4j for java and set provided config for native, mixup these not make sense for me.

cpp/velox/compute/VeloxBackend.cc
Yohahaha: I have test this patch in my local env, no problem.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
Yohahaha: hmm.. I dont really want to add how to use glog in GlutenConfig.scala, user should search it by own.

this config does not control LOG(INFO) or LOG(WARNING), it control [verbose log](https://github.com/google/glog#verbose-logging), I will add another config to control INFO/WARNING/ERROR.

Yohahaha: https://github.com/oap-project/gluten/pull/3466

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -1012,6 +1012,13 @@ object GlutenConfig {
       .intConf
       .createWithDefault(2)
 
+  val COLUMNAR_VELOX_GLOG_VERBOSE_LEVEL =
+    buildStaticConf("spark.gluten.sql.columnar.backend.velox.glogVerboseLevel")
+      .internal()
+      .doc("Set glog verbose level in Velox backend.")
+      .intConf
+      .createOptional
```
zhztheplayer, line 1020: We'd let user know about the default value of the option in the doc?

zhztheplayer: e.g.
```
.doc("Set glog verbose level in Velox backend. Default value: 0")
```

cpp/velox/compute/VeloxBackend.cc
rui-mo: Thanks for checking.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
rui-mo: I see. Maybe we can add sth. like below to let the user know how to adjust this value in Gluten.
`The lower the verbose level, the more likely messages are to be logged.`

Yohahaha:
already update PR description and config doc. @rui-mo @zhztheplayer

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -60,6 +60,12 @@ const std::string kEnableUserExceptionStacktrace =
     "spark.gluten.sql.columnar.backend.velox.enableUserExceptionStacktrace";
 const std::string kEnableUserExceptionStacktraceDefault = "true";
 
+const std::string kGlogVerboseLevel = "spark.gluten.sql.columnar.backend.velox.glogVerboseLevel";
+const std::string kGlogVerboseLevelDefault = "0";
+
+const std::string kGlogSeverityLevel = "spark.gluten.sql.columnar.backend.velox.glogMinLogLevel";
```
zhztheplayer, line 66: Should the key text be `spark.gluten.sql.columnar.backend.velox.glogSeverityLevel` ?

zhztheplayer (commented):
What's the relationship between `spark.gluten.sql.columnar.backend.velox.glogMinLogLevel` and `spark.gluten.sql.columnar.backend.velox.glogSeverityLevel` ? Or just a mistake?

Other parts LGTM

jackylee-ch:
> > but use log4j's config to control native log is not make sense for me.
> 
> Why does not it make sense ? Say, if people only want to see the warn log, then he should change native config since he has changed the log4j config. If we can mapping the default log level of log4j to native side, then people do not need to change anything.

+1 for using log4j config to control native log. It would make user easier to control all log printing, instead of learning glog and other log printing logic for log printing.

cpp/velox/compute/VeloxBackend.cc
Yohahaha: Yes, my mistake.

Yohahaha:
Could we merge?

zhztheplayer (approved):
Looks good to me but I agreed that we should simply the log level configuration within log4j's config considered either. But it's fine to me if we do that in another ticket or PR.

Yohahaha:
I will not do such patch that use log4j to config native log, if someone would, please leave a switch config for it. Two important gflags of glog has been introduced by this patch, just read glog README seems not a big issue.

zhouyuan (approved):
👍  works for me

zhztheplayer:
> I will not do such patch that use log4j to config native log, if someone would, please leave a switch config for it. Two important gflags of glog has been introduced by this patch, just read glog README seems not a big issue.

I didn't mean exactly that we should use log4j's level config for glog but it does seem to become an issue that we have been adding increasing number of Velox options to control logging. It may not be something very critical but doing some simplifications against the options will benefit on user experience.

Ideally in some cases a user may not be aware of glog while he/she just enables Gluten plugin and sees what happens.

PR #3457, opened 2023-10-20
[VL] Decouple JNI object storing code from ExecutionCtx

zhztheplayer:
In ExecutionCtx's code, the object storing procedure is pretty much specialized to support JNI calls. That says, only over JNI call, a C++ object is required to be stored temporarily because RAII is not applied to JNI coding model.

Since ExecutionCtx should be designed to be a layer to serve C++ calls only, in the patch we'll decouple the object storing code from it then move the logic to a newly added `ObjectStore` class.

This work can be a preparation to solving https://github.com/oap-project/gluten/issues/3142.

By the way, class `ExecutionCtx` will be renamed to `Runtime` in the patch based on previous discussion https://github.com/oap-project/gluten/pull/3311#discussion_r1351347883.


Yohahaha:
I prefer keep `ctx` suffix.

cpp/core/compute/Runtime.h
```diff
@@ -49,34 +47,16 @@ struct SparkTaskInfo {
   }
 };
 
-/// ExecutionCtx is stateful and manager all kinds of native resources' lifecycle during execute a computation fragment.
```
Yohahaha, line 52: why remove comments?

cpp/core/utils/ObjectStore.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace gluten {
+
+// A store for caching shared-ptrs and enlarging lifecycles of the ptrs to match lifecycle of the store itself by
+// default, and also serving release calls to release a ptr in advance. This is typically used in JNI scenario to bind
+// a shared-ptr's lifecycle to a Java-side object or some kind of resource manager.
+class ObjectStore {
+ public:
+  static std::unique_ptr<ObjectStore> create() {
+    return std::unique_ptr<ObjectStore>(new ObjectStore());
```
Yohahaha, line 32: why not use `std::make_unique` ?

cpp/core/utils/ObjectStore.h
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  virtual ~ObjectStore();
```
Yohahaha, line 35: if no sub-class derived from ObjectStore, I prefer remove this.

Yohahaha, line 35: if no sub-class derived from ObjectStore, I prefer remove this.

cpp/core/compute/Runtime.h
zhztheplayer: `Runtime` will no longer be about managing lifecycles of resources. It's more an C++ API for Spark-compatible query execution.

`objStore_` is still there to manage lifecycles but it will not be a long term solution. It might be removed once we developed a new way to ease Java-C++ object binding.

zhztheplayer:
> I prefer keep `ctx` suffix.

"Context" should be used when it's a container of contextual objects. Since lifecycle management will no longer be ExecutionCtx's own responsibility, we'd rethink if we still need "ctx" as the name suffix.

cpp/core/utils/ObjectStore.h
zhztheplayer: The dtor was overridden:
https://github.com/oap-project/gluten/pull/3457/files/bef40002c6ff0dfcbad539574cf3db6cca4c938d#diff-18f93239b8777fbe54615e69821a5f959b1a721f9be7279460e4881b077bb7fdR21-R31

cpp/core/utils/ObjectStore.h
zhztheplayer: The ctor has private access. In the case `std::make_unique` doesn't work.

cpp/core/utils/ObjectStore.h
zhztheplayer: > if no sub-class derived from ObjectStore, I prefer remove virtual.

`virtual` for dtor is usually required in C++ programming: 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c35-a-base-class-destructor-should-be-either-public-and-virtual-or-protected-and-non-virtual

cpp/core/compute/Runtime.h
```diff
@@ -164,7 +139,12 @@ class ExecutionCtx : public std::enable_shared_from_this<ExecutionCtx> {
     return taskInfo_;
   }
 
+  const std::unique_ptr<ObjectStore>& objectStore() {
```
Yohahaha, line 142: Runtime keep unique ref of ObjectStore, I think return raw pointer would be more make sense which mean caller should not hold it.

cpp/core/utils/ObjectStore.h
Yohahaha: I forget test class.. ignore this.

cpp/core/utils/ObjectStore.h
```diff
@@ -0,0 +1,56 @@
...
+
+  ResourceHandle save(std::shared_ptr<void> obj);
+
+  template <typename T>
+  std::shared_ptr<T> retrieve(ResourceHandle handle) {
+    const std::lock_guard<std::mutex> lock(mtx_);
+    std::shared_ptr<void> object = store_.lookup(handle);
+    // Programming carefully. This will lead to ub if wrong typename T as passed in.
+    auto casted = std::static_pointer_cast<T>(object);
+    return casted;
```
Yohahaha, line 45: Curious could we use such pattern to avoid wrong cast?
```c++
if (auto casted = dymanic_cast<T>(object)) {
  return casted;
}
GLUTEN_FAIL();
```

cpp/core/utils/ObjectStore.cc
```diff
@@ -0,0 +1,44 @@
...
+#include "ObjectStore.h"
+#include <iostream>
+
+gluten::ObjectStore::~ObjectStore() {
+  // destructing in reversed order (the last added object destructed first)
+  const std::lock_guard<std::mutex> lock(mtx_);
+  for (auto itr = aliveObjectHandles_.rbegin(); itr != aliveObjectHandles_.rend(); itr++) {
+    ResourceHandle handle = *itr;
+    if (store_.lookup(handle) == nullptr) {
+      std::cerr << "Fatal: resource handle " + std::to_string(handle) + " not found in store." << std::endl;
```
Yohahaha, line 27: I think this case it not acceptable and you use Fatal too, so use GLUTEN_FAIL or anyway to throw exception to kill task?

cpp/core/utils/ObjectStore.h
zhztheplayer: Not possible to use `dynamic_cast` over `void*` type. We may be able to have the relevant object classes inheriting one common parent then we can use `dynamic_cast`. We can try it in future as a follow-up to the PR.

cpp/core/utils/ObjectStore.cc
zhztheplayer: No, it's in destructor. We'd avoid throwing from destructor.

cpp/core/utils/ObjectStore.cc
```diff
@@ -0,0 +1,44 @@
...
+    }
+    store_.erase(handle);
+  }
+}
+
+gluten::ResourceHandle gluten::ObjectStore::save(std::shared_ptr<void> obj) {
+  const std::lock_guard<std::mutex> lock(mtx_);
+  ResourceHandle handle = store_.insert(obj);
```
Yohahaha, line 35: nit: store_.insert(std::move(obj));

ulysses-you:
TBH, I do not very care about the naming.. just hope do not change the code frequently.

> "Context" should be used when it's a container of contextual objects. Since lifecycle management will no longer be ExecutionCtx's own responsibility

Do you mean, now we do not hold `Iterator` and `ColumnarBatch` inside `ExecutionCtx` ?

ulysses-you:
then I'm fine if there is a plan to imporve this part further

rui-mo (approved):

zhztheplayer:
Since Jenkins is broken I'll keep tracking the result after merging.

Issue #3459, opened 2023-10-20
Skip unnecessary local sort for static partition write. From [SPARK-37194]


lwz9103:
### Description

This feature is supported in spark 3.4.  https://github.com/apache/spark/pull/37290
Please take this feature to gluten

PR #3460, opened 2023-10-20
[GLUTEN-3459] Skip unnecessary local sort for static partition write

lwz9103:

(Fixes: \#3459)

### DIFF
--- a/shims/spark33/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
+++ b/shims/spark33/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
@@ -100,6 +100,8 @@ object FileFormatWriter extends Logging {
    * @return
    *   The set of all partition paths that were updated during this write job.
    */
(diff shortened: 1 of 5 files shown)


shims/spark33/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -100,6 +100,32 @@ object FileFormatWriter extends Logging {
    * @return
    *   The set of all partition paths that were updated during this write job.
    */
+
+  // scalastyle:off argcount
+  def write(
```
JkSelf, line 105: @lwz9103 Can we also support in spark 32?

lwz9103: It's a good suggestion, and it has been added. Thanks.

JkSelf (approved):
LGTM.

PR #3461, opened 2023-10-20
[WIP][GLUTEN-3456][VL] Add config to disable columnar table cache if input is row-based

ulysses-you:

This config is a trade-off that to disable columnar table cache to prevent regression if the input is row-based. In general, if we use cache then there are multi-consumers to reference the cached plan, so it still has benefis for consumer side even if we add a row to columnar to do the cache.

Note that, if we disable columnar table cache then all the consumer plan will add row to columanr.

## How was this patch tested?

Pass CI


ulysses-you:
cc @gaoyangxiaozhu @zhouyuan @zhztheplayer

ulysses-you:
it seems hard to add this config. Say, we convert a row-based input using vanilla cache serializer, then at consumer side we did not know if the cached data is row-based or columanr..

PR #3466, opened 2023-10-20
[VL] Add config for glog severity log level in native side

Yohahaha:
Add `spark.gluten.sql.columnar.backend.velox.glogSeverityLevel` to control native glog severity log level, [doc](https://github.com/google/glog).
Default is 0 which means enable INFO level, set 1 means enable WARNING level.

User can set this config to suppress such Velox log
```
W1020 16:15:02.563387 216094 Connector.cpp:39] ConnectorFactory with name hive is already registered.
I1020 16:15:02.563445 216094 HiveConnector.cpp:69] Hive connector test-hive created with maximum of 20000 cached file handles.
```


Yohahaha:
depends #3454

PR #3471, opened 2023-10-20
[VL] Simplify the conversion from Substrait type to Velox type

rui-mo:

Previously, there are two kinds of conversion from Substrait type to Velox type.

1. From Substrait type to `SubstraitType`, which is composed of a type name and the nullability. The type name in `SubstraitType` is then used to convert to Velox Type.
2. From the short name of Substrait type to Velox type.

For `1`, string is used as the intermediate format to allow reusability of different backends. But as for now, this function is used only for Velox backend, and the intermediate string brings extra complexity for code maintainability.
For `2`, it firstly converts the short name to aligned name format with `SubstraitType`, and reuses the first conversion. The process of string-string conversion is unnecessary and increases the complexity.

With this PR, above two conversions will become more straight forward as below.
1. From Substrait type to Velox Type directly.
2. From the short name of Substrait type to Velox type directly.

## How was this patch tested?

under test


cpp/velox/substrait/VeloxSubstraitSignature.cc
```diff
@@ -60,6 +60,99 @@ std::string VeloxSubstraitSignature::toSubstraitSignature(const TypePtr& type) {
...
+      types.emplace_back(fromSubstraitSignature(typeStr));
+      names.emplace_back("");
+      childrenTypes.erase(0, pos + delimiter.length());
+    }
+    types.emplace_back(fromSubstraitSignature(childrenTypes));
+    names.emplace_back("");
+    return std::make_shared<RowType>(std::move(names), std::move(types));
+  }
+
+  VELOX_UNSUPPORTED("Substrait type signature conversion to Velox type not supported for {}.", signature);
```
zhztheplayer, line 153: Would the case of nested struct eventually fall here without crashes?

zhztheplayer (approved):
Thanks!

cpp/velox/substrait/VeloxSubstraitSignature.cc
rui-mo: Fixed the nested struct exception.

zhztheplayer (approved):

PR #3485, opened 2023-10-23
[VL] Refine cmake flags to decrease normal build time

Yohahaha:
Move duckdb and test_util targets under BUILD_TESTS flag to decrease build time, and modify related build scripts.
`normal` means without benchmark/test.


Yohahaha:
@rui-mo @PHILO-HE could you help review this patch? thanks!

dev/builddeps-veloxbe.sh
```diff
@@ -103,7 +103,7 @@ if [ "$SKIP_BUILD_EP" != "ON" ]; then
     cd $GLUTEN_DIR/ep/build-velox/src
     ./get_velox.sh --enable_hdfs=$ENABLE_HDFS --build_protobuf=$BUILD_PROTOBUF --enable_s3=$ENABLE_S3
     ./build_velox.sh --enable_s3=$ENABLE_S3 --build_type=$BUILD_TYPE --enable_hdfs=$ENABLE_HDFS \
-                   --enable_ep_cache=$ENABLE_EP_CACHE --build_benchmarks=$BUILD_BENCHMARKS
+                   --enable_ep_cache=$ENABLE_EP_CACHE --build_tests=$BUILD_TESTS
```
rui-mo, line 106: Why do we remove `build_benchmarks` option here?

ep/build-velox/src/build_velox.sh
```diff
@@ -50,8 +51,8 @@ for arg in "$@"; do
     ENABLE_EP_CACHE=("${arg#*=}")
     shift # Remove argument name from processing
     ;;
-  --build_benchmarks=*)
-    ENABLE_BENCHMARK=("${arg#*=}")
+  --build_tests=*)
+    ENABLE_TESTS=("${arg#*=}")
```
rui-mo, line 55: Same question.

ep/build-velox/src/build_velox.sh
```diff
@@ -90,9 +91,12 @@ function compile {
     exit 1
   fi
 
-  COMPILE_OPTION="-DVELOX_ENABLE_PARQUET=ON"
-  if [ $ENABLE_BENCHMARK == "OFF" ]; then
-    COMPILE_OPTION="$COMPILE_OPTION -DVELOX_BUILD_TESTING=OFF -DVELOX_BUILD_TEST_UTILS=ON"
+  COMPILE_OPTION="-DVELOX_ENABLE_PARQUET=ON -DVELOX_BUILD_TESTING=OFF -DVELOX_BUILD_TEST_UTILS=OFF -DVELOX_ENABLE_DUCKDB=OFF"
```
rui-mo, line 94: I suppose parquet compilation is not needed for ORC users. Better to keep it as optional.

dev/builddeps-veloxbe.sh
Yohahaha: I find we never use this option, and Gluten seems not need too, `BUILD_TESTS` is enough.

ep/build-velox/src/build_velox.sh
Yohahaha: previous parquet option is enabled by default, I keep it as original.

> I suppose parquet compilation is not needed for ORC users

We should not separate users with file format, more is better.

ep/build-velox/src/build_velox.sh
```diff
@@ -90,9 +91,12 @@ function compile {
...
+  if [ $ENABLE_BENCHMARK == "ON" ]; then
+    COMPILE_OPTION="$COMPILE_OPTION -DVELOX_BUILD_BENCHMARKS=ON"
```
PHILO-HE, line 96: Thanks for your fix!
Considering `--build_benchmarks` option will be removed, do we need to keep this part?

Yohahaha: Thanks for review, I keep both `build_benchmarks` and `build_tests` here.

PHILO-HE:
We just updated velox. Please do a code rebase. Thanks!

Yohahaha:
> We just updated velox. Please do a code rebase. Thanks!

Could you help review again? thanks!

PHILO-HE (approved):
Looks good! Thanks!

Yohahaha:
Thank you!

PR #3495, opened 2023-10-23
Fix a typo in error message of native backend initialization

j7nhai:
fix typo

### DIFF
--- a/gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitializer.java
+++ b/gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitializer.java
@@ -36,7 +36,7 @@ public final class NativeBackendInitializer {
       BackendsApiManager.getSettings().resolveNativeConf(nativeConfMap);
       initialize(JniUtils.toNativeConf(nativeConfMap));
     } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


gluten-data/src/main/java/io/glutenproject/init/NativeBackendInitializer.java
```diff
@@ -36,7 +36,7 @@ public static void initializeBackend(scala.collection.Map<String, String> conf)
       BackendsApiManager.getSettings().resolveNativeConf(nativeConfMap);
       initialize(JniUtils.toNativeConf(nativeConfMap));
     } catch (Exception e) {
-      LOG.error("Error calling InitializerJniWrapper.initialize(...)", e);
+      LOG.error("Error calling NativeBackendInitializer.initialize(...)", e);
```
PHILO-HE, line 39: Thanks for your fix!
Previously, we use `InitializerJniWrapper.initialize` to do the initialization. Now it becomes `initialize(...)`. So it may be better to change the error message to: "Failed to call native backend's initialize method" or something else.

j7nhai: updated

PHILO-HE (approved):
Looks good! Thanks!

PR #3496, opened 2023-10-23
[CORE] Do not fallback eagerly with ColumnarToRowExec

ulysses-you:

It has no meaning to add ColumnarToRow eagerly if we will add it finally. So we ignore the ColumnarToRow to avoid fallback eagerly. It is useful if people set fallback threshold to 1.
For example: 1 is better than 2

```
1. We first run native operator then add ColumnarToRow
      ColumnarExchange
            |
  HashAggregateTransformer
            |
      ColumnarToRow

2. We first add ColumnarToRow then run vanilla Spark operator
     ColumnarExchange
            |
      ColumnarToRow
            |
      HashAggregate
```

## How was this patch tested?

add test

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
@@ -18,14 +18,14 @@ package io.glutenproject.extension
 
 import io.glutenproject.GlutenConfig
 import io.glutenproject.backendsapi.BackendsApiManager
-import io.glutenproject.execution.BroadcastHashJoinExecTransformer
+import io.glutenproject.execution.{BasicScanExecTransformer, BroadcastHashJoinExecTransformer}
 import io.glutenproject.extension.columnar.TransformHints
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.Attribute
 import org.apache.spark.sql.catalyst.rules.Rule
-import org.apache.spark.sql.execution.{ColumnarBroadcastExchangeExec, ColumnarShuffleExchangeExec, ColumnarToRowExec, CommandResultExec, LeafExecNode, SparkPlan, UnaryExecNode}
+import org.apache.spark.sql.execution.{ColumnarBroadcastExchangeExec, ColumnarShuffleExchangeExec, ColumnarToRowExec, CommandResultExec, LeafExecNode, RowToColumnarExec, SparkPlan, UnaryExecNode}
 import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanExec, AQEShuffleReadExec, BroadcastQueryStageExec, ColumnarAQEShuffleReadExec, QueryStageExec, ShuffleQueryStageExec}
 import org.apache.spark.sql.execution.command.ExecutedCommandExec
 import org.apache.spark.sql.execution.exchange.{Exchange, ReusedExchangeExec}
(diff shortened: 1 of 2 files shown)


ulysses-you:
cc @zhouyuan @PHILO-HE thank you

zhouyuan:
@ulysses-you looks good to me, I'm curious if this is found on your product env? Could you please share the performance benefit on this change? 

thanks, -yuan

ulysses-you:
thank you @zhouyuan , yes, by default we set the fallabck threshold to 1. After this patch, the final stage will lazy fallback as the pr description said. So the benefits is obvious that we can apply more native operators before fallback.

PHILO-HE (commented):
Thanks for your proposal!

I am thinking whether we can broaden the optimizations, e.g. introduce a policy to only fall back partial plans to ensure consecutive cross-stage Gluten plans (see below graph) when use-specified fallback threshold is satisfied. Do you have any idea? Thanks!


```mermaid
  graph TD;
      ColumnarExchange-->GlutenPlan-A;
      GlutenPlan-A-->GlutenPlan-B;
      GlutenPlan-B-->ColumnarToRow;
      ColumnarToRow-->VanillaPlan-A;
      VanillaPlan-A-->VanillaPlan-B;
      VanillaPlan-B-->VanillaPlan-C;
```

ulysses-you:
@PHILO-HE If I understand it correct. You are talking about the the global optimum vs. partial optimum. Then we can do fallback with taking consecutive cross-stage into account. Currently, the policy of ExpandFallback is a kind of partial optimum which will fallback if match the threshold in one stage. If the next stage is offloaded, then it needs to add RowToColumnar by itself.

I think the main difference of global optimum vs. partial optimum is where do we add RowToColumnar. If we add it before shuffle exchange:
- a negative case is, after shuffle exchange it may add a ColumnarToRow if fallback
- a positive case is, after shuffle exchange all operators are offloaded so it improve the shuffle performance

It's more complex to reach global optimum, and the main issue are: 
- in columnar rules, we can only see the partial plan
- AQE may re-plan the query plan which may break the assumption of previous stage

To be clear, this pr itself is based on the current ExpandFallback and aims to lazy add ColumnarToRow as far as possible. It should solve the case you point out when threshold is 1.

PHILO-HE:
Hi @ulysses-you, thanks for your discussion!
Let me clarify a bit. what I meant is still to focus on the plan optimization for a single stage. For a given stage, there may have some transformable plans (GlutenPlan-A, GlutenPlan-B) at the beginning, then non-consecutive transformable plans (VanillaPlan-C, ...) (needs certain number of C2R/R2C, making fallback threshold satisfied). If last stage's output is columnar (gluten's columnar exchange is used), a new stage fallback policy will just keep GlutenPlan-A & GlutenPlan-B and then only fall back the following remaining plans for this stage. Thus, we can get consecutive cross-stage transformable plans (the exchange mentioned here is counted as last stage's plan, right?).
Just an immature idea for a new fallback policy, not completely matching the purpose of this PR.

> To be clear, this pr itself is based on the current ExpandFallback and aims to lazy add ColumnarToRow as far as possible. It should solve the case you point out when threshold is 1.

Yes, I agree.

ulysses-you:
@PHILO-HE got it, I think the idea makes sense. I think this pr is a improvement to reach the goal, is there any comments ?

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -67,6 +67,45 @@ import org.apache.spark.sql.execution.exchange.{Exchange, ReusedExchangeExec}
...
+    //
+    // spotless:on
+    var numColumnarToRow = 0
+    var numRowToColumnar = 0
+    plan.foreach {
+      case _: ColumnarToRowExec => numColumnarToRow += 1
+      case _: RowToColumnarExec => numRowToColumnar += 1
+      case _ =>
+    }
+    !hasScan && numColumnarToRow == 1 && numRowToColumnar == 0
```
PHILO-HE, line 106: I think it may be more straightforward to collect all leaf nodes that are query stage nodes (they represent last stages). And count all such leaf nodes who have columnar output to get `requiredColumnarToRowWhenFallbackWholeStage`. Then, we get

```
netFallbacks = fallbacks -  requiredColumnarToRowWhenFallbackWholeStage
```
We may just need to check whether `netFallbacks` meets user-specified `fallbackThreshold`.
For your case where `fallbackThreshold = 1`,  we can find `fallbacks = 1`, but `requiredColumnarToRowWhenFallbackWholeStage = 1`, so the fallback policy will not take effect for this case. 

Using `netFallbacks` looks reasonable also for other threshold settings, no matter whether this stage has scan node. Do you think so? Thanks!

ulysses-you: In one stage, there may be one more ColumnarToRow for one leaf node, for example:
```
  ColumnarExchange
         |
HashAggregateTransformer
         |
   ColumnarToRow
         |
      Project
         |
   ColumnarToRow
         |
 ProjectTransformer
         |
   ColumnarToRow
         | 
   HashAggregate
         |
        ... 
```
That's why I just improve the case that fallback threshold = 1. Things will be complex if the threshold is bigger than 1.

PHILO-HE: Let's merge this version firstly. I can try to do some optimization in a separate PR. Thanks!

PHILO-HE (approved):
Thanks for your continuous efforts!

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
ulysses-you: thank you @PHILO-HE !

PR #3497, opened 2023-10-23
[GLUTEN-3359] Add Spark3.4 unit test framework

JkSelf:

This PR consists of two commits. The first commit involves copying the spark 33 unit test into spark 34 within gluten-ut. The second commit addresses the compilation issues and disables the failed unit test. And the disabled unit test is recorded [here](https://github.com/oap-project/gluten/issues/3559).

## How was this patch tested?

Adding spark34 folder in gluten-ut module.


.github/workflows/velox_be.yml
```diff
@@ -193,6 +193,73 @@ jobs:
...
+          ./build_velox.sh --velox_home=/opt/velox --enable_ep_cache=ON'
+      - name: Build Gluten CPP library
+        run: |
+          docker exec ubuntu2004-test-spark34-slow-$GITHUB_RUN_ID bash -l -c '
+          cd /opt/gluten/cpp && \
+          ./compile.sh --build_velox_backend=ON --velox_home=/opt/velox --arrow_home=/opt/velox/_build/release/third_party/arrow_ep'
+      - name: Build and Run unit test for Spark 3.4.1(slow tests)
+        run: |
+          docker exec ubuntu2004-test-spark34-slow-$GITHUB_RUN_ID bash -l -c 'cd /opt/gluten && \
+          mvn clean install -Pspark-3.4 -Pbackends-velox -Prss -Pspark-ut -DargLine="-Dspark.test.home=/opt/spark331" -DtagsToInclude=org.apache.spark.tags.ExtendedSQLTest'
```
zhouyuan, line 220: looks like we should target for spark341 here?

JkSelf: @zhouyuan  Yes.  Only `GlutenParquetIOSuite#SPARK-40128 read DELTA_LENGTH_BYTE_ARRAY encoded strings `test throws a `Path does not exist` exception when reading the input parquet data. However, even when using Spark 3.4.1, it still fails with a velox exception. Therefore, can we use Spark 3.3.1 in this PR and use Spark 3.4.1 in the subsequent PR?

zhouyuan (approved):
👍

PR #3504, opened 2023-10-24
[VL] Fix cache output if selectedAttributes has wrong ordering with cacheAttributes

ulysses-you:

This pr fixes a bug when the consumer side selects the different ordering attributes with the original cached plan. We should reorganize the columnar batch.

## How was this patch tested?

add test


ulysses-you:
cc @zhztheplayer thank you

yaooqinn (approved):

PR #3515, opened 2023-10-24
[GLUTEN-3516] Fix centos8 build failure when hdfs enabled

j7nhai:


rui-mo (commented):
Thanks for the fix!
HDFS compilation can be enabled with `--enable_hdfs=ON` in get_velox, build_velox and compile  at [velox_be.yml#L290-L291](https://github.com/oap-project/gluten/blob/main/.github/workflows/velox_be.yml#L290-L291). Would you like to modify the command to get hdfs enabled?

j7nhai:
> Thanks for the fix! HDFS compilation can be enabled with `--enable_hdfs=ON` in get_velox, build_velox and compile at [velox_be.yml#L290-L291](https://github.com/oap-project/gluten/blob/main/.github/workflows/velox_be.yml#L290-L291). Would you like to modify the command to get hdfs enabled?

I have been used --enable_hdfs=ON. But As you can see from the origin file `get_velox.sh`:

```bash
sed -i '/^function install_protobuf.*/i function install_libhdfs3 {\n cd "\${DEPENDENCY_DIR}"\n github_checkout oap-project/libhdfs3 master \n cmake_install\n}\n' scripts/setup-centos8.sh
```

`function install_protobuf` may not exist in the file. So the function `install_libhdfs3` cannot be added to the file.

rui-mo:
@j7nhai This PR fixes above issue, right?
> Maybe build with hdfs enabled should be added to workflows?

As you mentioned in #3516,  I'm wondering if we can enable hdfs in the workflow. To achieve that, we need to modify the commands in [velox_be.yml#L290-L291](https://github.com/oap-project/gluten/blob/main/.github/workflows/velox_be.yml#L290-L291).

j7nhai:
> @j7nhai This PR fixes above issue, right?
> 
> > Maybe build with hdfs enabled should be added to workflows?
> 
> As you mentioned in #3516, I'm wondering if we can enable hdfs in the workflow. To achieve that, we need to modify the commands in [velox_be.yml#L290-L291](https://github.com/oap-project/gluten/blob/main/.github/workflows/velox_be.yml#L290-L291).

@rui-mo  Workflows updated. Now centos8 looks working well, but still some wrong with centos 7.

rui-mo:
@j7nhai Thank you. Could you please keep the modification for centos 8 only in this PR? We can get hdfs enabled in it firstly.

j7nhai:
> @j7nhai Thank you. Could you please keep the modification for centos 8 only in this PR? We can get hdfs enabled in it firstly.

OK @rui-mo , updated.

rui-mo (approved):

rui-mo:
Thanks!

rui-mo:
The remaining issue on centos7: https://github.com/oap-project/gluten/issues/3540.

PR #3520, opened 2023-10-25
[VL] Apply patch to Velox for compilation

rui-mo:

Removes build script changes to Velox by applying patch from Gluten to Velox.
https://github.com/oap-project/velox/pull/425

## How was this patch tested?

under test

### DIFF
--- a/ep/build-velox/src/build_velox.sh
+++ b/ep/build-velox/src/build_velox.sh
@@ -65,6 +65,19 @@ for arg in "$@"; do
   esac
 done
 
+function apply_compilation_fixes {
+  current_dir=$1
+  velox_home=$2
+  sudo cp ${current_dir}/modify_velox.patch ${velox_home}/
+  sudo cp ${current_dir}/modify_arrow.patch ${velox_home}/third_party/
+  cd ${velox_home}
+  git apply modify_velox.patch
+  if [ $? -ne 0 ]; then
+    echo "Failed to apply compilation fixes to Velox: $?."
+    exit 1
+  fi
+}
+
 function compile {
   TARGET_BUILD_COMMIT=$(git rev-parse --verify HEAD)
 
@@ -245,6 +258,7 @@ fi
 echo "Target Velox commit: $TARGET_BUILD_COMMIT"
 
 check_commit
+apply_compilation_fixes $CURRENT_DIR $VELOX_HOME
 compile
 
 echo "Successfully built Velox from Source."
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -16,8 +16,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=update
+VELOX_REPO=https://github.com/rui-mo/velox.git
+VELOX_BRANCH=revert
 VELOX_HOME=""
 
 #Set on run gluten on HDFS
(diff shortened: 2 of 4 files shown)


ep/build-velox/src/build_velox.sh
```diff
@@ -65,6 +65,16 @@ for arg in "$@"; do
...
 done
 
+function apply_compilation_fixes {
+  current_dir=$1
+  velox_home=$2
+  cp ${current_dir}/compilation.patch ${velox_home}/
+  sudo mkdir -p ${velox_home}/third_party/arrow_patches/
+  sudo cp ${current_dir}/helpers.patch ${velox_home}/third_party/arrow_patches/
+  cd ${velox_home}
+  git apply compilation.patch
```
PHILO-HE, line 75: If `git apply` fails, will the build script fail or continue the build even though the patch is not applied? Could you confirm it? Thanks!

rui-mo: If it fails, I suppose whether to continue or to stop, the build process will fail anyway. How do you think?

zhouyuan: not actually, as it may build the original code?

PHILO-HE: I think we can do a check after executing `git apply`. And leave some message to help developers easily figure out why build failed.

Reference: https://askubuntu.com/questions/29370/how-to-check-if-a-command-succeeded

rui-mo: Updated. Thanks.

ep/build-velox/src/build_velox.sh
```diff
@@ -65,6 +65,19 @@ for arg in "$@"; do
...
+function apply_compilation_fixes {
+  current_dir=$1
+  velox_home=$2
+  sudo cp ${current_dir}/compilation.patch ${velox_home}/
+  sudo mkdir -p ${velox_home}/third_party/arrow_patches/
+  sudo cp ${current_dir}/helpers.patch ${velox_home}/third_party/arrow_patches/
+  cd ${velox_home}
+  git apply compilation.patch
+  if [ $? -ne 0 ]; then
+    echo "Failed to apply compilation fixes to Velox: $?."
```
PHILO-HE, line 77: Please add `exit 1` to directly make the build exit. Suppose it is not acceptable to continue the build without the patch applied.

PHILO-HE (commented):
Nit:
Rename `compilation.patch` to some name like `modify_velox_build.patch`, which can be more read-friendly to developers.

ep/build-velox/src/build_velox.sh
```diff
@@ -65,6 +65,19 @@ for arg in "$@"; do
   esac
 done
 
+function apply_compilation_fixes {
+  current_dir=$1
+  velox_home=$2
+  sudo cp ${current_dir}/modify_velox.patch ${velox_home}/
+  sudo cp ${current_dir}/modify_arrow.patch ${velox_home}/third_party/
```
JkSelf, line 72: @rui-mo Do we need to git apply modify_arrow.patch like modify_velox.patch in  line 74?

ep/build-velox/src/build_velox.sh
rui-mo: Updated.

ep/build-velox/src/build_velox.sh
rui-mo: It is applied to Arrow during compilation: https://github.com/oap-project/gluten/pull/3520/files#diff-b41402fae332ead78c4646843c9bcf66a2813203cdc82521c9c9f130b48146dbR100.

PHILO-HE (commented):
LGTM!

PHILO-HE (approved):

PR #3522, opened 2023-10-25
[VL] Fix single partitioning evict error

marin-ma:
For single partitioning, the SplitState is always `kInit` during split. Therefore, the check of whether cached payloads are clear should not be applied to single partitioning.

This PR also adds `Evictable` interface and `SelfEvictedMemoryPool`, which can be used in unit test to mock the behavior of invoking the `evictFixedSize` API by Spark's memory management during memory allocation failure.
Below code demonstrates how to use in UT:
```
auto pool = defaultArrowPool().get();
auto mockingPool = SelfEvictedMemoryPool(pool);
auto evictable = Evictable(&mockingPool);
mockingPool.setEvictable(&evictable);

mockingPool.setCapcity(100); // When trying to allocate > 100 bytes, it will call evictable->evictFixedSize to trigger self-evict

```

### DIFF
--- a/cpp/core/memory/Evictable.h
+++ b/cpp/core/memory/Evictable.h
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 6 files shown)


JkSelf:
@marin-ma Can you help to add some unit tests to cover this change?

marin-ma:
@JkSelf Do you have further comments?

JkSelf (approved):
LGTM.

PR #3525, opened 2023-10-25
[Core] Shuffle the configured directories before use

marin-ma:
Make the configured directories in random order, to avoid each task creating local spilled files in the same directory.

### DIFF
--- a/cpp/core/shuffle/LocalPartitionWriter.cc
+++ b/cpp/core/shuffle/LocalPartitionWriter.cc
@@ -16,6 +16,7 @@
  */
 
 #include "shuffle/LocalPartitionWriter.h"
+#include <random>
 #include <thread>
 #include "shuffle/Utils.h"
 #include "utils/DebugOut.h"
@@ -32,6 +33,11 @@ std::string LocalPartitionWriterBase::nextSpilledFileDir() {
 
 arrow::Status LocalPartitionWriterBase::setLocalDirs() {
   ARROW_ASSIGN_OR_RAISE(configuredDirs_, getConfiguredLocalDirs());
+  // Shuffle the configured local directories. This prevents each task from using the same directory for spilled files.
+  std::random_device rd;
+  std::default_random_engine engine(rd());
+  std::shuffle(configuredDirs_.begin(), configuredDirs_.end(), engine);
+
   subDirSelection_.assign(configuredDirs_.size(), 0);
 
   // Both data_file and shuffle_index_file should be set through jni.


cpp/core/shuffle/LocalPartitionWriter.cc
```diff
@@ -32,6 +33,11 @@ std::string LocalPartitionWriterBase::nextSpilledFileDir() {
 
 arrow::Status LocalPartitionWriterBase::setLocalDirs() {
   ARROW_ASSIGN_OR_RAISE(configuredDirs_, getConfiguredLocalDirs());
+  // Shuffle the configured local directories. This prevents each task from using the same directory for spilled files.
+  std::random_device rd;
+  std::default_random_engine engine(rd());
+  std::shuffle(configuredDirs_.begin(), configuredDirs_.end(), engine);
+
```
jackylee-ch, line 40: nit: It would sound better to me if we generate a random ID for the configured directory.

marin-ma: Do you mean generating a random starting index and then cycling to get the next directory, or selecting a directory randomly each time?

jackylee-ch: The second one, selecting a directory randomly each time. Spark also use the same way to get the random directory.

marin-ma: Spark uses the mod of hash of the filename as the index. Here the spilled file is only used for native writer, so I think we don't need to completely follow spark's way as the spilling mechanism are also not the same. We only need to make sure the directory is randomly chosen.

jackylee-ch: The current idea seems fine to me. I'm just concerned about the shuffle cost for the configuredDirs, especially considering that we might have up to 10 directories in configuredDirs.

marin-ma: This is a one-time op at the beginning of each shuffle write task. After shuffling, it saves the result to `configuredDirs_`, and looping to get the next directory each time.

jackylee-ch: ok

JkSelf (approved):
LGTM.

zhztheplayer (approved):
Thanks. BTW I noticed that we rely on an env variable to set the dirs. Do you think the way needs to be optimized in future?

marin-ma:
> Thanks. BTW I noticed that we rely on an env variable to set the dirs. Do you think the way needs to be optimized in future?

Yes, we should avoid setting env var in the JNI codepath and need some cleanup. The env var should be only used for cpp unit tests and mirco benchmarks.

PR #3526, opened 2023-10-25
[VL] Iterator's and its payloads' lifecycle improvements

zhztheplayer:
One of the purpose of the patch is to close plan iterators (so Velox's task gets stopped) before stopping shuffle writer. 

Will do some cleanups against code.

Changes listed as following:

1. Add utility method `Iterators#wrap` for miscellaneous operations on iterator, e.g. handle iterator and payload's closing.
2. Remove CloseableColumnarBatchIterator, use utility method `#recyclePayload`.
3. Remove CloseablePairedColumnarBatchIterator, use utility method `#recyclePayload`.
4. Refactor code to use utility method `#recycleIterator`. The method would take early-close into account. Which means, once iterator's `hasNext` started to return `false`, iterator will be closed. Previously for Velox's out iterator we only close it when task ends.
5. Add API `NativeMemoryManager#hold()` to make memory manager hold references of all created memory pools to avoid them from becoming inaccessible after the task gets early-closed.

This will also fix some metrics issues when SQL `limit` clause is used since we rely on `#hasNext` result to collect iterator metrics.


gluten-core/src/main/scala/org/apache/spark/util/Iterators.scala
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  override def hasNext: Boolean = {
+    tryClose()
+    in.hasNext
+  }
+
+  override def next(): A = {
+    val a: A = in.next()
+    closer.synchronized {
```
marin-ma, line 38: I'm a bit confused why do we need `synchronized` here. Does it mean `hasNext` and `next` can be called at the same time?

zhztheplayer: > Does it mean hasNext and next can be called at the same time?

This case is unusual... However sometime another case might be possible that the iterator was read asynchronously by another thread (say, Velox sometime hand over it's sub-tasks to background thread), but the lambda passed into `TaskResources.addRecycler` will be called from the original Spark task thread.

BTW I didn't have all the cases tested but rather programed defensively... It will be hard to debug if we actually bumped into threading issues caused by lack of locking here.

zhztheplayer:
@zzcclp 

This impacts CH backend's execution either. Would you please help to review? Thanks.

gluten-core/src/main/scala/org/apache/spark/util/Iterators.scala
```diff
@@ -0,0 +1,154 @@
...
+    pipelineTime += accumulatedTime.getAndSet(
+      0L
+    ) // make sure the accumulated time is submitted once
+  }
+}
+
+class WrapperBuilder[A](in: Iterator[A]) { // FIXME how to make the ctor companion-private?
+  private var wrapped: Iterator[A] = in
+
+  def recyclePayload(closeCallback: (A) => Unit): WrapperBuilder[A] = {
```
ulysses-you, line 121: thank you @zhztheplayer for the refactor. Can you explain a bit more why naming a new word payload instead of ColumnarBatch in java ? Do you mean there is something else besides ColumnarBatch will call this method ? I'm wondering if we can make a `recycleColumnarBatch` method to make it clear.

zhztheplayer: In `Iterator[A]`, `A` should be type of "payloads" of the iterator. It's also possible that `A` is not `ColumnarBatch`, a typical case is C2R that creates a row iterator as output. 

Does `payload` sound very confusing here? `element` could be another choice but I felt it probably sounds too general for Gluten's use case.

backends-velox/src/main/scala/io/glutenproject/execution/VeloxColumnarToRowExec.scala
```diff
@@ -98,26 +98,13 @@ object VeloxColumnarToRowExec {
...
     val res: Iterator[Iterator[InternalRow]] = new Iterator[Iterator[InternalRow]] {
 
       override def hasNext: Boolean = {
-        val hasNext = batches.hasNext
-        if (!hasNext && !closed) {
-          jniWrapper.nativeClose(c2rId)
-          closed = true
-        }
-        hasNext
+        batches.hasNext
```
ulysses-you, line 107: It seems we didi not call `recyclePayload` for c2r after refactor since the iterator is row-based. Is there a leak ?

zhztheplayer: Didn't change any code about payload closing. The deleted lines were about iterator completion so moved to`recycleIterator`.

ulysses-you: I see it is closed by itself, then it seems c2r did not call `recyclePayload` ?

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
```diff
@@ -230,11 +230,12 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
   /**
    * Generate closeable ColumnBatch iterator.
    *
+   * FIXME: This no longer overrides parent's method
```
zzcclp, line 233: keep it in the `CHIteratorApi`

zhztheplayer: Any suggested change? I only removed `override` modifier. But not sure where to put this method yet.

zzcclp: > Any suggested change? I only removed `override` modifier. But not sure where to put this method yet.

This change is OK to me.

backends-velox/src/main/scala/io/glutenproject/execution/VeloxColumnarToRowExec.scala
zhztheplayer: > I see it is closed by itself, then it seems c2r did not call `recyclePayload` ?

Yes as you see the batches are closed manually within some `if-else` conditions. I don't how much effort needed to refactor the usages within `#recyclePayload()` so didn't change that part of code in this patch. Probably we can do that in a separate ticket, I am not sure.

zzcclp:
Run Gluten Clickhouse CI

marin-ma:
LGTM. Thanks!

marin-ma (approved):

PR #3529, opened 2023-10-25
[VL] Add EvictGuard to avoid spilling recursively in ShuffleWriter

marin-ma:
ShuffleWriter supports spilling initiated both internally and by other operators. However, it shouldn't self-trigger a spill during the spilling process. Because during a spill, it can invoke memory allocation APIs that triggers another spill, like when resizing the partition buffers."

This PR prevents the following call stack:

[Spill triggered by some op] -> ShuffleWriter.spill -> reserveMemory -> trigger spill another time -> ShuffleWriter.spill

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -319,6 +319,28 @@ std::shared_ptr<arrow::RecordBatch> makeUncompressedRecordBatch(
   }
   return arrow::RecordBatch::Make(writeSchema, 1, {arrays});
 }
+
+class EvictGuard {
+ public:
+  explicit EvictGuard(SplitState& splitState) : splitState_(splitState) {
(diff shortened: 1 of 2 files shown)


cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -319,6 +319,28 @@ std::shared_ptr<arrow::RecordBatch> makeUncompressedRecordBatch(
...
+class EvictGuard {
+ public:
+  explicit EvictGuard(SplitState& splitState) : splitState_(splitState) {
+    oldState_ = splitState;
+    splitState_ = SplitState::kUnevictable;
+  }
+
+  ~EvictGuard() {
+    splitState_ = oldState_;
+  }
```
zhztheplayer, line 332: Would you intend to save `splitState` to `splitState_` in ctor?

zhztheplayer (approved):

PR #3533, opened 2023-10-26
[CELEBORN] Optimize the performance of Celeborn's pushPartitionData

kerwin-zk:

Optimize the performance of Celeborn's pushPartitionData

## How was this patch tested?
Observing from the flame graph, it can be seen that the celebornConf.clientPushBufferMaxSize parameter in the pushPartitionData method is causing significant performance overhead.

image


ulysses-you (approved):
lgtm, thank you @kerwin-zk

ulysses-you (approved):

PR #3544, opened 2023-10-26
[VL] Refine shuffle writer return types and namespace

marin-ma:
Currently, there are inconsistencies in the shuffle writer code related to namespaces and return types.

1. **Namespace Usage:** In some instances, the prefix velox:: is used, such as `velox::RowVector`. In others, there is no prefixed namespace, simply using `RowVector`. It would be better to consistently use the full namespace: `facebook::velox::RowVector`.
2. **Return Types:** Some parts of the code return `arrow::Status` or `arrow::Result`, while others throw exceptions. This inconsistency complicates our C++ unit tests, making it unclear whether to check for exceptions or inspect the returned status code. A uniform approach should be adopted, returning `arrow::Status` and `arrow::Result` where applicable.

### DIFF
--- a/cpp/core/shuffle/ShuffleWriter.h
+++ b/cpp/core/shuffle/ShuffleWriter.h
@@ -97,10 +97,6 @@ class ShuffleWriter : public Evictable {
 
   virtual arrow::Status stop() = 0;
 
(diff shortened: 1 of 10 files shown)


zhztheplayer (approved):
Thanks!

PR #3564, opened 2023-10-30
[CORE] Optimize some methods in agg transformer

liujiayi771:

Implement methods needsPreProjection and needsPostProjection in HashAggregateExecBaseTransformer in a more elegant and concise manner.

## How was this patch tested?

CI

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
@@ -41,7 +41,6 @@ import com.google.protobuf.Any
 import java.util
 
 import scala.collection.mutable.ListBuffer
-import scala.util.control.Breaks.{break, breakable}
 
 /** Columnar Based HashAggregateExec. */
 abstract class HashAggregateExecBaseTransformer(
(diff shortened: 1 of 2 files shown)


gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -159,74 +158,48 @@ abstract class HashAggregateExecBaseTransformer(
...
-          break
+    groupingExpressions.exists {
+      case _: Attribute => false
+      case _ => true
+    } || aggregateExpressions.exists {
+      expr =>
+        expr.filter match {
+          case None =>
+          case Some(_: Attribute) =>
+          case Some(_: Literal) =>
```
rui-mo, line 171: Don't we need return value for these 3 branches?

rui-mo (commented):
Thanks！Just a question.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
liujiayi771: The original code, when encountering `expr.filter` defined and it is not `Attribute` and `Literal`, would set needsProjection to true and break to return directly. This corresponds to the modified code's `case _ => return true`. 
In other cases, it will not immediately break, but continue the next validation based on `expr.mode`. If these three cases explicitly return false, the subsequent validation will be skipped. If there is no return value, it will proceed to the validation based on `expr.mode` below.

rui-mo: Seems the CI fails. Could you take a look?

liujiayi771:
@rui-mo Can we merge this PR?

ulysses-you (approved):

rui-mo:
> @rui-mo Can we merge this PR?

Yes, sure. Already merged.

PR #3569, opened 2023-10-31
[CORE] Consider the cost when applying stage fallback policy

PHILO-HE:

When making a stage fallback, an extra ColumnarToRow may be required to adapt to the columnar output of last query stage(s). So this cost may need to be considered to get a net fallback number.

## How was this patch tested?

An existing UT covers it.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
@@ -67,47 +67,91 @@ import org.apache.spark.sql.execution.exchange.{Exchange, ReusedExchangeExec}
 case class ExpandFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkPlan)
   extends Rule[SparkPlan] {
 
-  private def countColumnarToRowWhenFallbackStage(plan: SparkPlan): Int = {
-    plan
-      .collectLeaves()
-      .filter(
-        p =>
-          p match {
-            case q: QueryStageExec if q.supportsColumnar => true
-            case _ => false
-          })
-      .size
-  }
-
-  private def countFallbacks(plan: SparkPlan): Int = {
+  private def countFallback(plan: SparkPlan): Int = {
     var fallbacks = 0
-    def countFallback(plan: SparkPlan): Unit = {
+    def countFallbackInternal(plan: SparkPlan): Unit = {
       plan match {
         case _: QueryStageExec => // Another stage.
         case _: CommandResultExec | _: ExecutedCommandExec => // ignore
         // we plan exchange to columnar exchange in columnar rules and the exchange does not
         // support columnar, so the output columnar is always false in AQE postStageCreationRules
         case ColumnarToRowExec(s: Exchange) if isAdaptiveContext =>
-          countFallback(s)
+          countFallbackInternal(s)
         case u: UnaryExecNode
             if !u.isInstanceOf[GlutenPlan] && InMemoryTableScanHelper.isGlutenTableCache(u.child) =>
           // Vanilla Spark plan will call `InMemoryTableScanExec.convertCachedBatchToInternalRow`
           // which is a kind of `ColumnarToRowExec`.
           fallbacks = fallbacks + 1
-          countFallback(u.child)
+          countFallbackInternal(u.child)
         case ColumnarToRowExec(p: GlutenPlan) =>
           logDebug(s"Find a columnar to row for gluten plan:\n$p")
           fallbacks = fallbacks + 1
-          countFallback(p)
+          countFallbackInternal(p)
         case leafPlan: LeafExecNode if InMemoryTableScanHelper.isGlutenTableCache(leafPlan) =>
         case leafPlan: LeafExecNode if !leafPlan.isInstanceOf[GlutenPlan] =>
           // Possible fallback for leaf node.
           fallbacks = fallbacks + 1
-        case p => p.children.foreach(countFallback)
+        case p => p.children.foreach(countFallbackInternal)
       }
     }
-    countFallback(plan)
+    countFallbackInternal(plan)
     fallbacks
+  }
+
+  /**
+   * When making a stage fall back, it's possible that we need a ColumnarToRow to adapt to last
+   * stage's columnar output. So we need to evaluate the cost, i.e., the number of required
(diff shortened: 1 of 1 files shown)


PHILO-HE:
Hi @ulysses-you, motivated by your previous work, this PR add some code to evaluate the cost when applying stage fallback policy. Could you please help review to see if it makes sense in practice? Thanks!

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -67,78 +67,93 @@ import org.apache.spark.sql.execution.exchange.{Exchange, ReusedExchangeExec}
...
+  private def countStageFallbackCost(plan: SparkPlan): Int = {
+    var stageFallbackCost = 0
+
+    /**
+     * Find a Gluten plan whose child is QueryStageExec. Then, increase stageFallbackCost if the
+     * last query stage's output is columnar.
+     */
+    def countStageFallbackCostInternal(plan: SparkPlan): Unit = {
+      plan match {
+        case p: GlutenPlan if p.children.find(_.isInstanceOf[QueryStageExec]).isDefined =>
```
ulysses-you, line 143: thank you @PHILO-HE. I think we should only use this cost if all leaf node are QueryStageExec with columnar output, and we should also consider table cache. For example, if the fallback threshold is 1

before this pr
```
ColumnarShuffle       
      |                 
  QueryStage       Shuffle
      |               |
 ColumnarToRow    QueryStage
      \             /
            Join
```

after this pr
```
                   Shuffle
                      | 
                  QueryStage  
                      |
ColumnarShuffle  RowToColumnar
   |                  |
 QueryStage       QueryStage
    \               /
     JoinTransformer
           |
      ColumnarToRow   
```

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -67,78 +67,93 @@ import org.apache.spark.sql.execution.exchange.{Exchange, ReusedExchangeExec}
...
+          p.children
+            .filter(_.isInstanceOf[QueryStageExec])
+            .foreach {
+              case stage: QueryStageExec if stage.supportsColumnar =>
+                stageFallbackCost = stageFallbackCost + 1
```
ulysses-you, line 148: it seems we should match default case here

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -175,11 +190,15 @@ case class ExpandFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkP
       return None
     }
 
-    val numFallback = countFallbacks(plan)
-    if (numFallback >= fallbackThreshold) {
+    val netFallbackNum = if (isAdaptiveContext) {
```
ulysses-you, line 193: the `net` do you mean final ? Does final sound more friendly ?

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
PHILO-HE: Thanks for your comment! Yes, it's possible that one last stage has row output. As RowToColumnar can be removed if current stage falls back, I think we can reduce the cost by 1 for such case to reflect the benefit of stage fallback. Just updated.

> we should also consider table cache.

Yes, just updated.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
PHILO-HE: Here, the net means "净值" to reflect stage fallback's benefit with cost deducted. I am afraid `final` may look like the fallback num at last after policy is applied. Do you think so?

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -67,78 +67,99 @@ import org.apache.spark.sql.execution.exchange.{Exchange, ReusedExchangeExec}
...
+     */
+    def countStageFallbackCostInternal(plan: SparkPlan): Unit = {
+      plan match {
+        case p: GlutenPlan if p.children.find(_.isInstanceOf[QueryStageExec]).isDefined =>
+          p.children
+            .filter(_.isInstanceOf[QueryStageExec])
+            .foreach {
+              case stage: QueryStageExec
+                  if stage.plan.isInstanceOf[GlutenPlan] ||
+                    InMemoryTableScanHelper.isGlutenTableCache(stage) =>
```
ulysses-you, line 149: we should also check it at line 143

ulysses-you (commented):
thank you @PHILO-HE , looks good to me except one comment

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
ulysses-you: thank you for theexplanation, I'm fine with net

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
PHILO-HE: Thanks for your comment! At line 143, we only find gluten plan whose one or more child is last stage. It seems this gluten plan cannot be gluten table cache since it's a leaf node without child. Right?

ulysses-you: I just consider such case if the leaf node is Gluten TableCache, for example:
```
    GlutenTableCache
           |
HashAggregateTrasnformer
           |
HashAggregateTrasnformer   
           |
     ColumnarToRow 
```

PHILO-HE: In this example, as `GlutenTableCache` cannot be connected to last stage, we don't need to consider two-stage's transition cost (ColumnarToRow) when falling back the current stage. Am I missing something?

ulysses-you (approved):

PR #3574, opened 2023-10-31
[CORE] Simplify WholeStageTransformer and BroadcastBuildSideRDD

ulysses-you:

In WholeStageTransformer, the columnarInputRDDs would contain BroadcastBuildSideRDD, but the dependency and partition of BroadcastBuildSideRDD is meaningless. BroadcastBuildSideRDD should only be used to hold the broadcast value and generate iterator for hash join.

## How was this patch tested?

PASS CI


ulysses-you:
cc @zhouyuan @PHILO-HE @marin-ma @zzcclp thank you

zzcclp (approved):
LGTM

PR #3584, opened 2023-11-01
[Core][VL] Add random parquet data generator and ShuffleWriterFuzzerTest

marin-ma:
Add `ShuffleWriterFuzzerTest`  which utilize `RandomParquetDataGenerator` to generate random schema, input batch size and data for shuffle. This test aims to evaluate the shuffle module's accuracy and spill behavior. Developers should first pass C++ unit tests and then manually execute this specific test after any modifications that could impact the shuffle module.

By default, each test will run with 10 iterations. Failed iterations will be printed as `Failed to run test 'testname' with seed: xxxxx, ...`. Developers can pick up the SQL query corresponding to the test name as well as the seeds from log and modify the `reproduce` test to reproduce. It's recommended to build cpp as Debug build type before running this test.
If any iterations are failed because of OOM, they will be printed as error log `Out of memory while running test 'testname' with seed: xxxxx, ...`. Iterations with OOM won't fail the test case.

`ShuffleWriterFuzzerTest` is tagged as `SkipTestTags` so it won't be run in CI.

### DIFF
--- a/backends-velox/pom.xml
+++ b/backends-velox/pom.xml
@@ -166,6 +166,13 @@
       <version>1.3.0</version>
       <scope>compile</scope>
     </dependency>
+    <!-- Java Faker for generating random data -->
+    <dependency>
+      <groupId>com.github.javafaker</groupId>
+      <artifactId>javafaker</artifactId>
+      <version>1.0.2</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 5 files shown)


marin-ma:
@zhouyuan Could you help to review? Thanks!

zhouyuan:
@marin-ma it seems there are several unit tests failed, I guess it's due to some gaps w/ main branch, can you please do a rebase to check again?

-yuan

marin-ma:
> @marin-ma it seems there are several unit tests failed, I guess it's due to some gaps w/ main branch, can you please do a rebase to check again?
> 
> -yuan

@zhouyuan The PR itself needs update. I will fix it ASAP. Thanks!

zhouyuan (approved):
👍 
this seems a standalone component, looks good to me

PR #3586, opened 2023-11-01
[VL] In OAP Velox fork, simplify shrinking/spilling code for Velox memory pool

zhztheplayer:
I've done some work to minimize the for-pick PR https://github.com/facebookincubator/velox/pull/5790, the change included in this patch would be needed for Gluten then.

### DIFF
--- a/cpp/velox/memory/VeloxMemoryManager.cc
+++ b/cpp/velox/memory/VeloxMemoryManager.cc
@@ -18,6 +18,7 @@
 #include "VeloxMemoryManager.h"
 #include "velox/common/memory/MallocAllocator.h"
 #include "velox/common/memory/MemoryPool.h"
+#include "velox/exec/MemoryReclaimer.h"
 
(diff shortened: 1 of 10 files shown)


cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -62,26 +63,28 @@ class ListenableArbitrator : public velox::memory::MemoryArbitrator {
   }
 
   void reserveMemory(velox::memory::MemoryPool* pool, uint64_t) override {
-    growPool(pool, memoryPoolInitCapacity_);
+    std::lock_guard<std::recursive_mutex> l(mutex_);
```
Yohahaha, line 66: 👍 
could we remove AllocationListener's mutex?

zhztheplayer: Let's keep that mutex for now since this one only works in Velox backend.

PHILO-HE (approved):

PR #3595, opened 2023-11-02
[VL] Fix a bug that may break bhj + exchange validation in a corner case

zhztheplayer:
Previous bhj + exchange joint validation may not work because of bugs. The patch is to fix the broken logic.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
@@ -313,8 +313,10 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
 
   /** Inserts a transformable tag on top of those that are not supported. */
   private def addTransformableTags(plan: SparkPlan): SparkPlan = {
-    addTransformableTag(plan)
-    plan.withNewChildren(plan.children.map(addTransformableTags))
+    // Walk the tree with post-order
+    val out = plan.withNewChildren(plan.children.map(addTransformableTags))
+    addTransformableTag(out)
+    out
   }
 
   private def addTransformableTag(plan: SparkPlan): Unit = {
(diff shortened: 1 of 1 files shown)


gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -551,7 +553,9 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
             maybeExchange match {
               case Some(exchange @ BroadcastExchangeExec(mode, child)) =>
                 TransformHints.tag(bhj, isBhjTransformable.toTransformHint)
-                TransformHints.tagNotTransformable(exchange, isBhjTransformable)
+                if (!isBhjTransformable.isValid) {
+                  TransformHints.tagNotTransformable(exchange, isBhjTransformable)
+                }
```
zhztheplayer, line 558: Once exchange was already tagged as **transformable**, This code may still re-tag it as **non-transformable** if bhj is considered **non-transformable** in validation at the time.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -313,8 +313,10 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
 
   /** Inserts a transformable tag on top of those that are not supported. */
   private def addTransformableTags(plan: SparkPlan): SparkPlan = {
-    addTransformableTag(plan)
-    plan.withNewChildren(plan.children.map(addTransformableTags))
+    // Walk the tree with post-order
+    val out = plan.withNewChildren(plan.children.map(addTransformableTags))
+    addTransformableTag(out)
+    out
```
zhztheplayer, line 319: This ensures bhj is validated after the corresponding exchange

PHILO-HE (approved):
Thanks!

PR #3599, opened 2023-11-02
[CORE] Add GlutenImplicits to get FallbackSummary easily

ulysses-you:

This pr adds a a helper class to get the Gluten fallback summary from a Spark Dataset. If AQE is enabled, but the query is not materialized, then this method will re-plan the query execution with disabled AQE. It is a workaround to get the final plan, and it may cause inconsistent results with a materialized query. However, we have no choice.

For example:
 
```scala
import org.apache.spark.sql.execution.GlutenImplicits._
val df = spark.sql("SELECT  FROM t")
df.fallbackSummary
```

## How was this patch tested?

add test

### DIFF
--- a/gluten-ut/spark33/src/test/scala/io/glutenproject/utils/clickhouse/ClickHouseTestSettings.scala
+++ b/gluten-ut/spark33/src/test/scala/io/glutenproject/utils/clickhouse/ClickHouseTestSettings.scala
@@ -2014,6 +2014,9 @@ class ClickHouseTestSettings extends BackendTestSettings {
       "SELECT structFieldSimple.key, arrayFieldSimple[1] FROM tableWithSchema a where int_Field=1")
     .exclude("SELECT structFieldComplex.Value.`value_(2)` FROM tableWithSchema")
   enableSuite[SparkFunctionStatistics]
-
+  enableSuite[GlutenImplicitsTest]
+    .exclude("fallbackSummary with shuffle")
+    .exclude("fallbackSummary with cache")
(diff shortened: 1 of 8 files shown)


ulysses-you:
cc @zhouyuan @PHILO-HE thank you

gluten-ut/spark33/src/test/scala/io/glutenproject/utils/clickhouse/ClickHouseTestSettings.scala
```diff
@@ -2014,6 +2014,9 @@ class ClickHouseTestSettings extends BackendTestSettings {
       "SELECT structFieldSimple.key, arrayFieldSimple[1] FROM tableWithSchema a where int_Field=1")
     .exclude("SELECT structFieldComplex.Value.`value_(2)` FROM tableWithSchema")
   enableSuite[SparkFunctionStatistics]
-
+  enableSuite[GlutenImplicitsTest]
+    .exclude("fallbackSummary with shuffle")
```
PHILO-HE, line 2018: This feature is not workable for CH backend? cc @zzcclp.

PHILO-HE (commented):
Thanks for your work!

PHILO-HE:
Hi @ulysses-you, could you also update the doc? This may be a good place: https://github.com/oap-project/gluten/blob/main/docs/get-started/Velox.md

ulysses-you:
@PHILO-HE yes, added the docs

PHILO-HE (approved):
Thanks!

PR #3601, opened 2023-11-02
[VL] Refine setting shuffle writer local directories

marin-ma:
Currently, shuffle writer local directories are setting through environment variable "GLUTEN_SPARK_LOCAL_DIRS", which is a historical issue and should be refined.

This PR removes setting/getting local directories through environment variable from JNI code path and keep the environment variable "GLUTEN_SPARK_LOCAL_DIRS" only for generic benchmark simulating spilling multiple disks.


zhztheplayer (approved):
Thanks!

PR #3603, opened 2023-11-02
[CORE] Use collection interface in method parameter and return type

liujiayi771:

It is recommended to use abstract interfaces for method parameters and return type instead of concrete implementations.

After modifying the method parameters and return type to abstract interface types, many codes can be rewritten in a more functional style. This PR also includes these modifications as well.

## How was this patch tested?

Existing test cases

### DIFF
--- a/backends-clickhouse/src/main/java/io/glutenproject/vectorized/StorageJoinBuilder.java
+++ b/backends-clickhouse/src/main/java/io/glutenproject/vectorized/StorageJoinBuilder.java
@@ -28,7 +28,6 @@ import org.apache.spark.sql.catalyst.expressions.Attribute;
 import org.apache.spark.sql.catalyst.expressions.Expression;
 import org.apache.spark.storage.CHShuffleReadStreamFactory;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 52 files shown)


gluten-core/src/main/scala/io/glutenproject/expression/WindowFunctionsBuilder.scala
```diff
@@ -21,11 +21,14 @@ import io.glutenproject.substrait.expression.ExpressionBuilder
 
 import org.apache.spark.sql.catalyst.expressions.{Expression, WindowExpression, WindowFunction}
 
+import java.lang.{Long => JLong}
+import java.util.{HashMap => JHashMap}
```
zhztheplayer, line 25: Can this import statement generated by some kind of tools?

liujiayi771: I made the modifications manually. However, upon carefully examining the code of Spark, I noticed that it actually has a lot of inconsistency in its usage. There are various conventions being used, with most of them using `JList`, but `java.util.List` and `util.List` are still present as well. I am now questioning whether it is necessary to make these modifications here, as it would be difficult to ensure that all future developers maintain a consistent style.

zhztheplayer: > as it would be difficult to ensure that all future developers maintain a consistent style.

Yes so it's not necessarily to become a coding style or restriction. Sounds fair to me.

liujiayi771:
@zhztheplayer @rui-mo It's ready for review. I have removed many unnecessary import modifications.

rui-mo (approved):

PR #3607, opened 2023-11-03
[CORE] Remove ColumnarAQEShuffleRead

ulysses-you:

AQEShuffleRead is compatible with columnar shuffle exchange, so it is unnecessary to add a custom ColumnarAQEShuffleRead. This pr removes all the usage of ColumnarAQEShuffleRead and make code simpler.

## How was this patch tested?

Pass CI


zhouyuan:
CC: @JkSelf

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
```diff
@@ -129,7 +129,7 @@ case class ColumnarShuffleExchangeExec(
   }
 
   override def getShuffleRDD(partitionSpecs: Array[ShufflePartitionSpec]): RDD[ColumnarBatch] = {
-    cachedShuffleRDD
+    new ShuffledColumnarBatchRDD(columnarShuffleDependency, readMetrics, partitionSpecs)
```
JkSelf, line 132: @ulysses-you Why need this change?

zhouyuan:
@JkSelf I remember we somehow backported some AQE patch to fix some bugs related with xchg reusing, is this valid?

-yuan

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
ulysses-you: It's the key change to use AQEShuffleRead directly. AQEShuffleRead will call getShuffleRDD to get the custom shuffle RDD. See:
https://github.com/apache/spark/blob/9cbc2d138a1de2d3ff399d224d817554ba0b9e18/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AQEShuffleReadExec.scala#L250C1-L265

JkSelf:
> @JkSelf I remember we somehow backported some AQE patch to fix some bugs related with xchg reusing, is this valid?
> 
> -yuan


@zhouyuan It seems to have no impact because the decision of whether to reuse the exchange in AQE depends on the `canonicalized `method of the plan. Whether it is the `ColumnarAQEShuffleReader `we added or the `AQEShuffleReader `of vanilla Spark, the `canonicalized ` method of `SparkPlan `is used. Therefore, there is no impact in this case.

JkSelf (approved):
LGTM.

PR #3611, opened 2023-11-03
[VL] Replace string comparisons with enum for shuffle partitioning

marin-ma:
This PR replaces string comparisons with enums for shuffle partitioning and includes some cleanup in the creation of partitioners. It does not introduce any changes to functionality.


JkSelf (approved):
LGTM.

PR #3629, opened 2023-11-06
[CORE] Remove unnecessary case match in getAttrForAggregateExpr

liujiayi771:
There are many processes in the `getAttrForAggregateExprs` method that construct aggregateAttr and index for specific aggFuncs. In fact, the patterns of these processes are consistent. They iterate over `inputAggBufferAttributes` of aggFunc, transform them, and insert them into `aggregateAttr`. Then, they increase the count in index for each insertion. The logic for validating the supported modes of different aggFuncs is the only specific part, which can be extracted separately.

## How was this patch tested?

Exists CI

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
@@ -34,7 +34,6 @@ import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.aggregate._
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.vectorized.ColumnarBatch
-import org.apache.spark.util.sketch.BloomFilter
 
 import com.google.protobuf.Any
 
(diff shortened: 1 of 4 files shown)


rui-mo (commented):
Thanks! Looks nice. Can be merged when our internal test passes.

rui-mo (approved):

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -417,195 +416,47 @@ abstract class HashAggregateExecBaseTransformer(
...
-            extendedAggFunc.getClass) =>
-        // get attributes from the extended aggregate functions
-        ExpressionMappings.expressionExtensionTransformer
-          .getAttrsForExtensionAggregateExpr(
-            aggregateFunc,
-            mode,
-            exp,
-            aggregateAttributeList,
-            aggregateAttr,
-            index)
```
zzcclp, line 432: Can not delete this part for the extension, when there are some custom agg functions which have the specified logic to implement.

zzcclp:
I will raise a pr to fix the extension problem.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
liujiayi771: I see that CustomSum should also follow this pattern.

zzcclp: CustomSum is only a simple custom ut case, in out internal product, we need to implement some custom logic.

liujiayi771: Thank you for the explanation. I see. Sorry for removing that portion of the logic.

PR #3638, opened 2023-11-07
[VL] Add find and build glog/gflags

marin-ma:
Currently, the CMake/Findglog.cmake module only finds libglog itself but doesn't resolve its dependency gflags. When libglog is found as static library, linking it to shared library such as `libvelox.so` will lead to undefined symbol.


marin-ma:
@Yohahaha Could you help to review? Thanks!

cpp/CMake/BuildGflags.cmake
```diff
@@ -0,0 +1,41 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+include_guard(GLOBAL)
+
+set(GLUTEN_GFLAGS_BUILD_SHA256_CHECKSUM
+  34af2f15cf7367513b352bdcd2493ab14ce43692d2dcd9dfc499492966c64dcf)
+string(CONCAT GLUTEN_GFLAGS_SOURCE_URL
+  "https://github.com/gflags/gflags/archive/refs/tags/"
+  "v${GLUTEN_GFLAGS_VERSION}.tar.gz")
```
Yohahaha, line 23: Could we use below pattern to allow user specify deps URL? same for glog.
```
if(DEFINED ENV{GLUTEN_GTEST_SOURCE_URL})
  set(GLUTEN_GTEST_SOURCE_URL "$ENV{GLUTEN_GTEST_SOURCE_URL}")
endif()
```

cpp/CMake/Findglog.cmake
```diff
@@ -1,37 +1,81 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set(GLUTEN_GLOG_MINIMUM_VERSION 0.4.0)
+set(GLUTEN_GLOG_VERSION 0.6.0)
+
+if(NOT BUILD_GLOG)
```
Yohahaha, line 21: Curious could we just find first, then check `glog_FOUND`? seems `BUILD_GLOG` can be removed.

Yohahaha:
Thank you! LGTM, only some comments.

cpp/CMake/Findglog.cmake
marin-ma: It's a bit tricky here. We cannot make sure user has both glog and gflags installed. If glog is found but gflags is missing, it fails. The option is for ignoring system glog then directly build glog and gflags from source. https://github.com/oap-project/gluten/pull/3638/files#diff-9df362ebf18ceb320fa6423ff9140fd2bd75906d57d92ca018febfc3a93864a9R66

Yohahaha: Thanks, I'm fine with it.

cpp/CMake/BuildGflags.cmake
Yohahaha: Let's merge this patch first, I will refine these deps download URL in follow up patch.

marin-ma:
@PHILO-HE @rui-mo Could you help to review? Thanks!

PHILO-HE (approved):
Thanks!

PR #3642, opened 2023-11-08
[VL] [Test] Test rebasing upstream velox 11_8

JkSelf:

In the upstream Velox, it is ensured that the sorting key is unique in TopNode [here](https://github.com/facebookincubator/velox/pull/7172). Therefore, we add a check for this in this PR.

## How was this patch tested?

Pass CI.

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.h
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.h
@@ -152,6 +152,12 @@ class SubstraitToVeloxPlanConverter {
   /// Get aggregation step from AggregateRel.
   core::AggregationNode::Step toAggregationStep(const ::substrait::AggregateRel& sAgg);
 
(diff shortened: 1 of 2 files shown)


rui-mo (commented):
How many tets are affected by Velox's change? If there are many of them, maybe we should consider removing duplicated keys in Gluten before sending to Velox.

JkSelf:
> How many tets are affected by Velox's change? If there are many of them, maybe we should consider removing duplicated keys in Gluten before sending to Velox.

@rui-mo Only Q72 in TPC-DS is affected.

rui-mo:
How does its fallback affect the performance of q72?

zhouyuan (approved):
👍

JkSelf:
> How does its fallback affect the performance of q72?

@rui-mo There is no performance impact.

PR #3646, opened 2023-11-08
[CORE][VL] Support RewriteTransformer Rules and DeltaLake Scan

YannByron:
1. support DeltaLake 2.2 scan
2. support the column mapping mode
3. provide `RewriteTransformerRules` to extend if needed.

(Fixes: #2891)

## How was this patch tested?


felipepessoto:
Moving my comment from old #3376:

I think _metadata is not heavily used in 2.2, or not at all, but may be needed to replace the input_file_name UDF. In recent versions like 2.4 it is used for deletion vectors, as it needs the _metadata_row_index.

I created this repro:

```
class TestGluten extends QueryTest
  with SharedSparkSession with DeltaSQLCommandTest {
  test("mytest") {
    withTempDir { inputDir =>
      val testPath = inputDir.getCanonicalPath
      spark.range(10)
        .write
        .format("delta")
        .save(testPath)

      spark.sql(s""" ALTER TABLE delta.`$testPath` SET TBLPROPERTIES (
    'delta.minReaderVersion' = '2',
    'delta.minWriterVersion' = '5',
    'delta.columnMapping.mode' = 'name'
  )""")

      spark.sql(s"""SELECT id, _metadata.file_name FROM delta.`$testPath`""").show(false)
    }
  }
}
```

It fails because we try to replace every column, and _metadata fields are not in the mapping:

> [info] - mytest *** FAILED *** (7 seconds, 852 milliseconds)
> [info]   java.util.NoSuchElementException: key not found: file_name
> [info]   at scala.collection.MapLike.default(MapLike.scala:236)
> [info]   at scala.collection.MapLike.default$(MapLike.scala:235)
> [info]   at scala.collection.AbstractMap.default(Map.scala:65)
> [info]   at scala.collection.mutable.HashMap.apply(HashMap.scala:69)
> [info]   at io.glutenproject.extension.RewritePlanIfNeeded.$anonfun$transformColumnMappingPlan$4(ColumnarOverrides.scala:109)
> [info]   at scala.collection.immutable.List.map(List.scala:297)
> [info]   at io.glutenproject.extension.RewritePlanIfNeeded.io$glutenproject$extension$RewritePlanIfNeeded$$transformColumnMappingPlan(ColumnarOverrides.scala:107)
> [info]   at io.glutenproject.extension.RewritePlanIfNeeded$$anonfun$apply$1.applyOrElse(ColumnarOverrides.scala:64)
> [info]   at io.glutenproject.extension.RewritePlanIfNeeded$$anonfun$apply$1.applyOrElse(ColumnarOverrides.scala:59)


`val newAttr = o.withName(columnNameMapping(o.name))`

gluten-delta/pom.xml
```diff
@@ -0,0 +1,155 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
felipepessoto, line 1: Do we need to make change to include this on gluten-<backend_type>-bundle-spark?

YannByron:
@felipepessoto  I have tested this case, and can work correctly. Please check if there is this patch https://github.com/oap-project/gluten/pull/2563 in gluten you used. And to support metadata column is tracked by https://github.com/oap-project/gluten/issues/2618. So let this pr focus on the common cases.

gluten-delta/src/main/scala/io/glutenproject/extension/DeltaRewriteTransformerRules.scala
```diff
@@ -0,0 +1,163 @@
...
+  override def rules: Seq[Rule[SparkPlan]] = columnMappingRule :: Nil
+}
+
+object DeltaRewriteTransformerRules {
+
+  private val COLUMN_MAPPING_RULE_TAG: TreeNodeTag[String] =
+    TreeNodeTag[String]("io.glutenproject.delta.column.mapping")
+
+  private def notAppliedColumnMappingRule(plan: SparkPlan): Boolean = {
+    plan.getTagValue(COLUMN_MAPPING_RULE_TAG).isEmpty
```
yma11, line 44: What's the logic here? Seems `COLUMN_MAPPING_RULE_TAG` is not empty at initialization?

yma11:
@YannByron LGTM except need to add documentation for this support, like additional configurations, etc.

gluten-delta/src/main/scala/io/glutenproject/extension/DeltaRewriteTransformerRules.scala
YannByron: Firstly, this `COLUMN_MAPPING_RULE_TAG` is used to avoid a delta scan applies this rule multiple times.
At initialization, the original transformer can't be tagged, so `COLUMN_MAPPING_RULE_TAG` is empty.

YannByron:
> @YannByron LGTM except need to add documentation for this support, like additional configurations, etc.

@yma11 thank you for your review. There is no configuration needed. Users just put the additional gluten-delta jar into the class path, then can query delta table in gluten/velox env.

yma11:
> @yma11 thank you for your review. There is no configuration needed. Users just put the additional gluten-delta jar into the class path, then can query delta table in gluten/velox env.

Yeah. Then just doc what you said in location like [here](https://github.com/oap-project/gluten/blob/main/docs/get-started/Velox.md#hdfs-support). You may can also make a short introduction about what cases supported.

gluten-delta/pom.xml
YannByron: No. I hope these gluten-lakeformat modules can be used in both backends.

YannByron:
@yma11 Doc is done. PTAL again.

yma11 (approved):

PR #3650, opened 2023-11-08
[GLUTEN-3378][CORE] Move getLocalFilesNode logic to scan transformer

liujiayi771:

The Lake format has its own custom partition type, with a processing logic that is different from the existing `FilePartition` and `GlutenMergeTreePartition`. In the unified design, the Lake format will have its own scan transformer. In order to ensure that the injection of the Lake format does not affect gluten-core, the process of generating `LocalFilesNode` from handling partitions needs to be moved to the scan transformer, allowing the Lake format's scan transformer to customize the processing logic of the partitions.

Introduce ReadSplit interface for ExtensionTableNode and LocalFileNode, so that we can get rid of using java.io.Serializable.

## How was this patch tested?

Exists CI.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -21,7 +21,7 @@ import io.glutenproject.backendsapi.IteratorApi
 import io.glutenproject.execution._
 import io.glutenproject.metrics.IMetrics
 import io.glutenproject.substrait.plan.PlanNode
-import io.glutenproject.substrait.rel.{LocalFilesBuilder, ReadSplit}
+import io.glutenproject.substrait.rel.{LocalFilesBuilder, SplitInfo}
 import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 import io.glutenproject.utils.Iterators
 import io.glutenproject.vectorized._
@@ -58,10 +58,10 @@ class IteratorApiImpl extends IteratorApi with Logging {
    *
    * @return
    */
-  override def genReadSplit(
+  override def genSplitInfo(
       partition: InputPartition,
       partitionSchemas: StructType,
-      fileFormat: ReadFileFormat): ReadSplit = {
+      fileFormat: ReadFileFormat): SplitInfo = {
     partition match {
       case f: FilePartition =>
         val (paths, starts, lengths, partitionColumns) =
@@ -200,7 +200,7 @@ class IteratorApiImpl extends IteratorApi with Logging {
   override def genNativeFileScanRDD(
       sparkContext: SparkContext,
       wsCxt: WholeStageTransformContext,
-      readSplits: Seq[ReadSplit],
+      splitInfos: Seq[SplitInfo],
       numOutputRows: SQLMetric,
       numOutputBatches: SQLMetric,
       scanTime: SQLMetric): RDD[ColumnarBatch] = {
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -76,6 +76,8 @@ class IteratorApiImpl extends IteratorApi with Logging {
           partitionColumns,
           fileFormat,
           preferredLocations.toList.asJava)
+      case _ =>
+        throw new UnsupportedOperationException(s"Unsupported input partition.")
     }
   }
 
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -21,7 +21,7 @@ import io.glutenproject.backendsapi.IteratorApi
 import io.glutenproject.execution._
 import io.glutenproject.metrics.IMetrics
 import io.glutenproject.substrait.plan.PlanNode
-import io.glutenproject.substrait.rel.LocalFilesBuilder
+import io.glutenproject.substrait.rel.{LocalFilesBuilder, ReadSplit}
 import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 import io.glutenproject.utils.Iterators
 import io.glutenproject.vectorized._
@@ -46,11 +46,10 @@ import java.lang.{Long => JLong}
 import java.net.URLDecoder
 import java.nio.charset.StandardCharsets
 import java.time.ZoneOffset
-import java.util.{ArrayList => JArrayList}
+import java.util.{ArrayList => JArrayList, HashMap => JHashMap, Map => JMap}
 import java.util.concurrent.TimeUnit
 
 import scala.collection.JavaConverters._
-import scala.collection.mutable
 
 class IteratorApiImpl extends IteratorApi with Logging {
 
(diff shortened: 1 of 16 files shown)


gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
```diff
@@ -244,26 +244,23 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
 
       // If these are two scan transformers, they must have same partitions,
       // otherwise, exchange will be inserted.
-      val allScanPartitions = basicScanExecTransformers.map(_.getPartitions)
-      val allScanPartitionSchemas = basicScanExecTransformers.map(_.getPartitionSchemas)
-      val partitionLength = allScanPartitions.head.size
-      if (allScanPartitions.exists(_.size != partitionLength)) {
+      val allScanLocalFilesNodes = basicScanExecTransformers.map(_.getLocalFilesNodes)
+      val partitionLength = allScanLocalFilesNodes.head.size
+      if (allScanLocalFilesNodes.exists(_.size != partitionLength)) {
```
YannByron, line 249: maybe extract a private method to encapsulate this logic that contains `getLocalFilesNodes`,  the assertion of whether the number of partitions is consistent, and the transposition. And doc for this method.

gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
```diff
@@ -54,6 +54,11 @@ trait BasicScanExecTransformer extends LeafTransformSupport with SupportFormat {
   // TODO: Remove this expensive call when CH support scan custom partition location.
   def getInputFilePaths: Seq[String]
 
+  def getLocalFilesNodes: Seq[(java.io.Serializable, Array[String])] =
```
YannByron, line 57: can we define a case class for `(java.io.Serializable, Array[String])`? To make `Serializable` as return type is so wired.

gluten-core/src/main/scala/org/apache/spark/softaffinity/SoftAffinityUtil.scala
```diff
@@ -23,42 +23,40 @@ import io.glutenproject.utils.LogLevelUtil
...
   private lazy val softAffinityLogLevel = GlutenConfig.getConf.softAffinityLogLevel
 
   /** Get the locations by SoftAffinityManager */
-  def getFilePartitionLocations(filePartition: FilePartition): Array[String] = {
-    // Get the original preferred locations
-    val expectedTargets = filePartition.preferredLocations()
-
+  def getFilePartitionLocations(
+      filePaths: Array[String],
+      preferredLocations: Array[String]): Array[String] = {
```
YannByron, line 34: can you explain this definition a little?

gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
liujiayi771: This is because CK is using `ExtensionTable` instead of `LocalFilesNode`, and their common superclass is `Serializable`. Now, a new interface called `ReadSplit` is introduced to represent the abstraction of both classes. Additionally, `preferredLocations` have been included in the class to avoid using Tuple.

gluten-core/src/main/scala/org/apache/spark/softaffinity/SoftAffinityUtil.scala
liujiayi771: Previously, the input of this method was `FilePartition`, but the `InputPartitions` for lake formats are custom-defined. This method primarily processes `FilePartition.files.filePath` and `preferredLocations`. Modifying the input type of this method would allow the lake formats to directly reuse this interface.

liujiayi771:
@yma11 @zhztheplayer Could you help review these changes?

liujiayi771:
@zzcclp Could you help review this PR?

gluten-core/src/main/java/io/glutenproject/substrait/rel/LocalFilesNode.java
```diff
@@ -30,12 +30,13 @@
 import java.util.List;
 import java.util.Map;
 
-public class LocalFilesNode implements Serializable {
+public class LocalFilesNode implements ReadSplit, Serializable {
```
yma11, line 33: Can we have a better name instead of `ReadSplit` which can show the inheritance relationship here? maybe `FilesNode` if there is kind of `IceBergFilesNode` / `DeltaFilesNode` exist?

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
```diff
@@ -336,6 +330,32 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
 
   override protected def withNewChildInternal(newChild: SparkPlan): WholeStageTransformer =
     copy(child = newChild, materializeInput = materializeInput)(transformStageId)
+
+  private def getReadSplitFromScanTransformer(
+      basicScanExecTransformers: Seq[BasicScanExecTransformer]): Seq[Seq[ReadSplit]] = {
+    // If these are two scan transformers, they must have same partitions,
```
yma11, line 336: When will there are multi scan transformers? Will they be against same table?

yma11:
@rui-mo can you also take a review on this PR as it has some substrait module related changes?

gluten-core/src/main/java/io/glutenproject/substrait/rel/ReadSplit.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+package io.glutenproject.substrait.rel;
+
+import com.google.protobuf.MessageOrBuilder;
+
+import java.util.List;
+
+public interface ReadSplit {
+  List<String> preferredLocations();
```
rui-mo, line 24: Maybe add some comments for this class and the member function `preferredLocations`.

gluten-core/src/main/java/io/glutenproject/substrait/rel/LocalFilesNode.java
yma11: I see. Then maybe `SplitInfo` is more suitable in this case and it has corresponding definition like "SplitInfo" or `HiveConnectorSplit` at velox side.

liujiayi771: @yma11 I think it makes sense. I will change it to `SplitInfo`.

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
ulysses-you: I do not thinkg bhj have two scan transformers in one stage, the build side should have a broadcast exchange. Bucket table join is a valid case which contain two scan transformers in one stage.

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
```diff
@@ -242,28 +243,21 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
...
-          })
+        val substraitPlanPartitions = allScanSplitInfos.zipWithIndex.map {
+          case (splitInfos, index) =>
+            wsCxt.substraitContext.initSplitInfosIndex(0)
+            wsCxt.substraitContext.setSplitInfos(splitInfos)
+            val substraitPlan = wsCxt.root.toProtobuf
+            GlutenPartition(
+              index,
+              substraitPlan.toByteArray,
+              splitInfos.head.preferredLocations().asScala.toArray)
```
ulysses-you, line 259: splitInfos(i).preferredLocations ?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
```diff
@@ -59,71 +58,61 @@ class IteratorApiImpl extends IteratorApi with Logging {
...
+          SoftAffinityUtil.getFilePartitionLocations(paths.asScala.toArray, f.preferredLocations())
+        LocalFilesBuilder.makeLocalFiles(
+          f.index,
+          paths,
+          starts,
+          lengths,
+          partitionColumns,
+          fileFormat,
+          preferredLocations.toList.asJava)
+    }
```
ulysses-you, line 79: nit: throw exception for default case

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
liujiayi771: Currently, the `GlutenPartition` only has one `preferredLocations`. However, within `GlutenPartition`, there are multiple `SplitInfo` instances, and each `SplitInfo` may have a different `preferredLocations`. However, the existing code only takes the first `preferredLocations`.

https://github.com/oap-project/gluten/blob/f21a96f727ab8368d1de0de25101b41c61d58038/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala#L126

ulysses-you: I see, it's for one partition but for several scans. I think the currently implementation loses some of scan preferredLocations. Shall we use `splitInfos.flatMap(_.preferredLocations).distinct` ?

liujiayi771: Yes. Merge all preferredLocations is better. I will fix.

liujiayi771: Additional note: distinct is not need as Spark will count the locations to determine which locations are preferred.

ulysses-you (approved):

PR #3652, opened 2023-11-09
[VL] Support custom cmake deps download url

Yohahaha:
Specify deps download url by `export GLUTEN_XXX_URL=path` if needed.

### DIFF
--- a/cpp/CMake/BuildGTest.cmake
+++ b/cpp/CMake/BuildGTest.cmake
@@ -8,9 +8,7 @@ set(GLUTEN_GTEST_SOURCE_URL
     "https://github.com/google/googletest/archive/refs/tags/v${GLUTEN_GTEST_VERSION}.tar.gz"
     )
 
-if(DEFINED ENV{GLUTEN_GTEST_SOURCE_URL})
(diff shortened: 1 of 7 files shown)


Yohahaha:
@PHILO-HE @marin-ma please help review this patch, thank you!

marin-ma:
If user set `GLUTEN_xxx_URL` environment variable, it's generally expected that they should also need to set `GLUTEN_xxx_SHA256`, or the checksum verification would fail. Is that correct?

Yohahaha:
> If user set `GLUTEN_xxx_URL` environment variable, it's generally expected that they should also need to set `GLUTEN_xxx_SHA256`, or the checksum verification would fail. Is that correct?

No, checksum is optional. checksum would be empty if user only set `GLUTEN_XXX_URL`.

marin-ma:
LGTM. Thanks!

marin-ma (approved):

PHILO-HE (approved):
Thanks!

Yohahaha:
could we meger?

Yohahaha:
thank you!

PR #3655, opened 2023-11-09
[GLUTEN-3654][VL]Fix the duplicated key exception in TopN

JkSelf:
The `TakeOrderedAndProjectExec `will be transformed into either a sort + limit or directly into a limit operation. If it is a sort + limit case, Velox will convert it into a `TopNNode`. Here, we need to do some validation on the `TopNNode`.

## How was this patch tested?

Pass Jenkins test

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -347,11 +347,23 @@ bool SubstraitToVeloxPlanValidator::validateExpression(
 }
 
 bool SubstraitToVeloxPlanValidator::validate(const ::substrait::FetchRel& fetchRel) {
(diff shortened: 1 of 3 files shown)


JkSelf:
@rui-mo Passed Jenkins TPC-DS test. http://sr168:8080/view/Gluten/job/Gluten_TPCDS_test/874/

zhouyuan:
@JkSelf it looks like the tpcds test in GHA is fine, is this due to the difference on scale factor?

-yuan

JkSelf:
> @JkSelf it looks like the tpcds test in GHA is fine, is this due to the difference on scale factor?
> 
> -yuan

@zhouyuan It seems that this issue is associated with the type of Spark application submission. I can reproduce  this problem using spark-shell with a data scale of 1GB.

rui-mo (approved):

PR #3659, opened 2023-11-09
[GLUTEN-3625][FOLLOWUP] Always use the hardcode module name

ulysses-you:

Before this pr the follow cmd can not change the project version for sub-module. It seems the maven plugin compatibility is not very good for using property value module name. So this pr reverts the change and go back to use the profile to specify the sub-module to avoid potential issue.

```
mvn versions:set -DgenerateBackupPoms=false \
  -DnewVersion="1.0.x" \
  -Pspark-3.2,spark-3.3,spark-3.4 -Pbackends-velox -Pspark-ut -Prss
```

## How was this patch tested?

manully test

### DIFF
--- a/shims/pom.xml
+++ b/shims/pom.xml
@@ -30,7 +30,6 @@
 
   <modules>
     <module>common</module>
-    <module>${spark.shim.module}</module>
   </modules>
 
   <dependencies>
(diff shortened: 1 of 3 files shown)


ulysses-you:
cc @PHILO-HE @zzcclp can you take a look again ? thank you

shims/pom.xml
```diff
@@ -30,7 +30,6 @@
 
   <modules>
     <module>common</module>
-    <module>${spark.shim.module}</module>
```
PHILO-HE, line 33: Thanks for your following up work!
Do we also need to remove the code where `spark.shim.module`/`spark.test.module` are set.
E.g.:
https://github.com/oap-project/gluten/blob/main/pom.xml#L48
https://github.com/oap-project/gluten/blob/main/pom.xml#L112

ulysses-you: yeah, removed

PHILO-HE (approved):
Thanks!

PR #3660, opened 2023-11-09
[VL] Flush intermediate results when partial aggregation runs out of memory

zhztheplayer:
Velox dependency https://github.com/zhztheplayer/velox/commits/wip-partial-flush


zhztheplayer:
Closing since we adopted partial spill https://github.com/oap-project/gluten/pull/3697

Issue #3637, opened 2023-11-07
WholeStage pipeline metric was broken


Yohahaha:
### Backend

VL (Velox)

### Bug description

![image](https://github.com/oap-project/gluten/assets/34979747/f56f3f65-29ca-44a2-ace9-e780ce9d29db)

### Spark version

None

PR #3661, opened 2023-11-09
[GLUTEN-3637][VL] Fix whole stage pipeline metric

Yohahaha:
(Fixes: \#3637)

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/utils/Iterators.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/utils/Iterators.scala
@@ -20,6 +20,7 @@ import org.apache.spark.{InterruptibleIterator, TaskContext}
 import org.apache.spark.sql.execution.metric.SQLMetric
 import org.apache.spark.util.TaskResources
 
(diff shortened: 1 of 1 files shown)


zhztheplayer (approved):
Thanks for catching!

zhouyuan (approved):
👍  works for me

gluten-core/src/main/scala/io/glutenproject/utils/Iterators.scala
```diff
@@ -110,8 +111,10 @@ private class PipelineTimeAccumulator[A](in: Iterator[A], pipelineTime: SQLMetri
   }
 
   private def tryFinish(): Unit = {
-    pipelineTime += accumulatedTime.getAndSet(
-      0L
+    pipelineTime += TimeUnit.NANOSECONDS.toMillis(
+      accumulatedTime.getAndSet(
```
zhouyuan, line 115: so this is due to `accumulatedTime` is changed during these days and it's nano-seconds(should use milliseconds)?

PR #3663, opened 2023-11-09
[VL] Avoid shrinking the binary buffer which triggers spill

marin-ma:
If spill is triggered by resizing binary partition buffer, should avoid shrinking itself during reclaiming memory from shrinking partition buffers. Because shrinking the buffer can possibly invalid the original underlying `ptr`.


marin-ma:
@jackylee-ch @zhztheplayer Could you help to review? Thanks!

zhouyuan (approved):
👍

PR #3665, opened 2023-11-10
[CORE] Fix ExpandExecTransformer miss input

ulysses-you:

Before this pr, the ExpandExecTransformer would miss input. It does not affect the result for now, but better to avoid potential issue in future.

image

## How was this patch tested?

Pass CI


JkSelf (approved):
Good catch!

PR #3673, opened 2023-11-10
[CORE] Deprecate RegularMemoryConsumer, use TreeMemoryConsumer for both isolated/shared memory mode

zhztheplayer:
Comparing to legacy `RegularMemoryConsumer` (which does the same with Spark's MemoryConsumer), Gluten's `TreeMemoryConsumer` is more customizable and already has following pros against the other:

(for developer)
1. Unlimited level of nesting
2. Able to set independent capacity for each node
3. (Planned) Able to control spiller's priority

The patch deprecates `RegularMemoryConsumer` and redirect dependent calls to `TreeMemoryConsumer`. Will remove `RegularMemoryConsumer` completely after some dev iterations.

The patch also divides Q67/Q95 low memory GHA job to two, one is expected to pass and another reports OOM and waits for fix.


gluten-core/src/main/scala/org/apache/spark/memory/SparkMemoryUtil.scala
```diff
@@ -58,6 +58,7 @@ object SparkMemoryUtil {
           overAcquire.getTarget.accept(this)
         }
 
+        @scala.annotation.nowarn
```
JkSelf, line 61: Is the "deprecated RegularMemoryConsumer" warning meant to be ignored here?

JkSelf (approved):
LGTM.

gluten-core/src/main/scala/org/apache/spark/memory/SparkMemoryUtil.scala
zhztheplayer: Yes. The annotation is required otherwise compilation will fail.

zhztheplayer:
@zzcclp I expect the patch won't impact CH backend. Please kindly report if any. Thanks.

PR #3679, opened 2023-11-10
[VL] Declare IntermediateTypes for specific agg function

liujiayi771:

We should not use intermediate types' size to distinguish between different agg functions, it is not rigorous. Different agg functions may have the same number of intermediate types, but these types may be completely different.

## How was this patch tested?

Exists CI.


liujiayi771:
@rui-mo Could you please take a look?

PHILO-HE (approved):
Looks good! Thanks!

PR #3682, opened 2023-11-11
[VL] Fix: add validate check for Generate

jackylee-ch:
1. Fix the validate check for Generate operator while parsing substrait to velox;
2. Fix the missing validate check for ifthen, such as `select explode(case a is null then array('test') else array(a) end) from test_string`;
3. Fix the validate failed while passing a `Alias` to Generate, such as `select all, a from (select split(a, '2') as al, a from test_string) lateral view explode(al) as all;`;
4. [mirror] add `from_json` to black_list since it is not support now;

## How was this patch tested?
unit and manul tests

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
@@ -18,7 +18,7 @@ package io.glutenproject.execution
 
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.Row
-import org.apache.spark.sql.execution.RDDScanExec
+import org.apache.spark.sql.execution.{GenerateExec, RDDScanExec}
 import org.apache.spark.sql.functions.{avg, col}
 import org.apache.spark.sql.types.{DecimalType, StringType, StructField, StructType}
 
@@ -563,9 +563,10 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
 
       // Testing unsupported case
       runQueryAndCompare("SELECT explode(from_json(cast(c1 as string),'ARRAY<STRING>')) from t;") {
-        df => {
-          getExecutedPlan(df).exists(plan => plan.exists(_.isInstanceOf[GenerateExec]))
-        }
+        df =>
+          {
+            getExecutedPlan(df).exists(plan => plan.exists(_.isInstanceOf[GenerateExec]))
+          }
       }
 
       // Testing unsupported case in case when
@@ -574,9 +575,10 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
           |SELECT explode(case when size(from_json(cast(c1 as string),'ARRAY<STRING>')) > 0
           |then array(c1) else array(c2) end) from t;
           |""".stripMargin) {
-        df => {
-          getExecutedPlan(df).exists(plan => plan.exists(_.isInstanceOf[GenerateExec]))
-        }
+        df =>
+          {
+            getExecutedPlan(df).exists(plan => plan.exists(_.isInstanceOf[GenerateExec]))
+          }
       }
     }
   }
--- a/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
@@ -545,6 +545,39 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
       runQueryAndCompare("SELECT c1, explode(array(c2)) FROM t") {
         checkOperatorMatch[GenerateExecTransformer]
       }
+
+      runQueryAndCompare("SELECT c1, explode(c3) FROM (SELECT c1, array(c2) as c3 FROM t)") {
+        checkOperatorMatch[GenerateExecTransformer]
+      }
+    }
+  }
+
+  test("Add the missing Generate validation check") {
+    withTable("t") {
+      spark
+        .range(10)
+        .selectExpr("id as c1", "id as c2")
+        .write
+        .format("parquet")
+        .saveAsTable("t")
+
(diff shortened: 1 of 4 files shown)


JkSelf:
@jackylee-ch Can we add some unit test to convert this change? Thanks.

jackylee-ch:
> @jackylee-ch Can we add some unit test to convert this change? Thanks.

Yeap, I have not finish it, will add it later.

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -545,6 +545,41 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
...
+    withTable("t") {
+      spark
+        .range(10)
+        .selectExpr("id as c1", "id as c2")
+        .write
+        .format("parquet")
+        .saveAsTable("t")
+
+      // Testing unsupported case
+      runQueryAndCompare("SELECT explode(from_json(cast(c1 as string),'ARRAY<STRING>')) from t;") {
```
PHILO-HE, line 565: This test will be broken after `from_json` is supported, right? Maybe, it is better to use a forever unsupported function for this test, e.g. a UDF.

Refer to Spark source [link](https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/SQLQuerySuite.scala#L3925-L3927).

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -545,6 +545,41 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
...
         checkOperatorMatch[GenerateExecTransformer]
       }
+
+      runQueryAndCompare("SELECT c1, explode(c3) FROM (SELECT c1, array(c2) as c3 FROM t)") {
+        checkOperatorMatch[GenerateExecTransformer]
+      }
+    }
+  }
+
+  test("Add the missing Generate validation check") {
```
PHILO-HE, line 555: Nit:
suggested test title:
validation should fail if unsupported expression is used for Generate.

PHILO-HE (commented):
Looks good! Just some trivial comments. Thanks!

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
jackylee-ch: Sure

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
jackylee-ch: Greate

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -545,6 +545,45 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
...
+
+      // Testing unsupported case
+      runQueryAndCompare("SELECT explode(intToArray(c1)) from t;") {
+        df =>
+          {
+            getExecutedPlan(df).exists(plan => plan.exists(_.isInstanceOf[GenerateExec]))
+          }
+      }
+
+      // Testing unsupported case in case when
```
JkSelf, line 576: This function tests the fallback behavior of the 'case when' statement. However, it appears that the fallback is due to the lack of support for 'intToArray', making it impossible to test the fallback of 'case when'.

JkSelf (commented):
LGTM. Do we need adding another test case that involves an `Alias`?

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
jackylee-ch: 'case when' would not cause fallback, this test is just used for testing whether we can fallback while there is a unsupported function in 'case when'.

jackylee-ch:
> LGTM. Do we need adding another test case that involves an `Alias`?

The test for alias is already added [here](https://github.com/oap-project/gluten/pull/3682/files#diff-d6f3816bd1517d3d8708ba5673298038b807726d79abf2b8f8f0584aac636752R549).

jackylee-ch:
@JkSelf @PHILO-HE any more suggestions?

JkSelf (approved):
LGTM. Thanks.

PHILO-HE (approved):
Looks good! Thanks for your fix!

jackylee-ch:
Thanks for your review. @JkSelf @PHILO-HE

supermem613:
@PHILO-HE, quick question: why was from_json marked as black-list with this PR? From a cursory glance it seems to run without issue. Are there any issues associated that made it be turned off?

PHILO-HE:
> @PHILO-HE, quick question: why was from_json marked as black-list with this PR? From a cursory glance it seems to run without issue. Are there any issues associated that made it be turned off?

@supermem613, it looks from_json is still not supported in velox. Did you confirm the operator is really offloaded to velox (e.g. by checking DAG in Spark UI)? Gluten has a fallback mechanism, for unsupported function, the relevant operator can fall back to vanilla Spark to still make the query run successfully.

supermem613:
> > @PHILO-HE, quick question: why was from_json marked as black-list with this PR? From a cursory glance it seems to run without issue. Are there any issues associated that made it be turned off?
> 
> @supermem613, it looks from_json is still not supported in velox. Did you confirm the operator is really offloaded to velox (e.g. by checking DAG in Spark UI)? Gluten has a fallback mechanism, for unsupported function, the relevant operator can fall back to vanilla Spark to still make the query run successfully.

Ah, you are right. I made the mistake of running it with a purely in-memory expression and it didn't fallback as it didn't really activate Gluten. Sorry.

PR #3687, opened 2023-11-13
[VL] Enable Spark functions for `translate`, `add_months`, `array_min`, `array_max`

PHILO-HE:
## How was this patch tested?

including translate, add_months, date_format, array_min, array_max.
These are already enabled in Velox, need to be removed from blacklist

Spark UTs covers the tests.


PHILO-HE:
Just aware of `date_format` depends on a velox patch. Moved out it.

rui-mo (approved):

PR #3694, opened 2023-11-13
[VL] use velox config lib to extract the configurations from spark

FelixYBW:
The PR is to avoid some other guy spending hours on stupid mistake by using 

`conf.set('spark.hadoop.fs.s3a.use.instance.credentials',True)`
instead of 
`conf.set('spark.hadoop.fs.s3a.use.instance.credentials','true')`

The root cause is that we use got=='true' to check if the flag is set.

The PR use velox's config lib to parse the configurations from Spark.

True can be set by 'true', 'True', 1, 'ON', 'Y', 'T', etc.

limitation: 'KB', 'MB', 'GB' is still not supported, only number is expected


cpp/velox/compute/VeloxBackend.cc
```diff
@@ -120,79 +120,49 @@ void VeloxBackend::printConf(const std::unordered_map<std::string, std::string>&
...
   // Init glog and log level.
-  {
-    auto vlogLevel = stoi(getConfigValue(conf, kGlogVerboseLevel, kGlogVerboseLevelDefault));
-    auto severityLogLevel = stoi(getConfigValue(conf, kGlogSeverityLevel, kGlogSeverityLevelDefault));
-    FLAGS_v = vlogLevel;
-    FLAGS_minloglevel = severityLogLevel;
-    FLAGS_logtostderr = true;
-    google::InitGoogleLogging("gluten");
-  }
+  auto veloxmmecfg = std::make_shared<facebook::velox::core::MemConfigMutable>(conf);
```
zhztheplayer, line 123: Looks like a typo? s/veloxmmecfg/veloxmemcfg/ ?

zhztheplayer (commented):
Thanks!

cpp/velox/compute/VeloxBackend.cc
FelixYBW: Thank you, good catch

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -208,50 +178,35 @@ void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf
   }
 
   std::unordered_map<std::string, std::string> s3Config({});
```
yma11, line 180: remove variable `s3Config` as it's not used anymore.

zhztheplayer (approved):

cpp/velox/compute/VeloxBackend.cc
FelixYBW: Oh, missed your comments. Let's remove it in next PR

Issue #3666, opened 2023-11-10
[VL] Integrate upstream DirectBufferedInput with executor


Yohahaha:
### Description

Upstream velox add DirectBufferedInput to replace BufferedInput, and use a optional executor do async load or prefetch, but its executor was passed from HiveConnector which means this executor is global instance, I'm not sure Gluten's thread mode or memory mode could fit it, will keep tracking.

Yohahaha:
DirectBufferedInput has less performance boost, my benchmark result in TPCH sf30 cloud storage is 160s(without executor) vs 152s(with executor).

FelixYBW:
@zhli1142015 Did you noted the perf regression?

zhli1142015:
yes, we noticed the regression also 10% totally for our TPCDS runs.

FelixYBW:
should we fallback to bufferInput?

directBufferedInput reads page by page, BufferInput reads row group by row group. 
rowgroup read ahead looks not work with direct buffer input.

zhli1142015:
I thought @Yohahaha opened an issue for this: https://github.com/facebookincubator/velox/issues/7535.

Yohahaha:
@FelixYBW have you try compare the perf between DBI and BI?

FelixYBW:
Yes, The issue I faced now is row group read ahead doesn't work with DBI. without the read ahead perf is worse

Yohahaha:
> Yes, The issue I faced now is row group read ahead doesn't work with DBI. without the read ahead perf is worse

thanks, I will reopen above issue to solve it.

we could use BI default in oap-velox for now. @zhouyuan

FelixYBW:
Just tested a IO bound query, DBI 4.6 min, BI 1.6 min

PR #3419, opened 2023-10-17
[GLUTEN-3364][VL] SplitPreloadPerDriver (SplitPreloadPerDriver = true, IOThreads > 0) causes unexpected crashes

zhztheplayer:
This fixes issue #3364.

~~After applying the patch, global memory pool will be used when IO threads > 0 for memory allocations in Hive connector. However side-effect that the allocations will be out-of-track by Spark memory manager will be brought.~~

Update: Switching to global memory pool doesn't help since the async executor holds task's memory pool anyway. Added a wait strategy (up to 1.5s) until async allocations are released. Some minor cleanups to connector initialization procedure.

PR #3696, opened 2023-11-13
[GLUTEN-3666][VL] Refine connector IO executor initialization

Yohahaha:
Change default value of spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver to 0, make startup message more clear.

(Fixes: \#3666)

### DIFF
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -89,7 +89,7 @@ const std::string kVeloxIOThreads = "spark.gluten.sql.columnar.backend.velox.IOT
 const std::string kVeloxIOThreadsDefault = "0";
 
 const std::string kVeloxSplitPreloadPerDriver = "spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver";
-const std::string kVeloxSplitPreloadPerDriverDefault = "2";
+const std::string kVeloxSplitPreloadPerDriverDefault = "0";
 
 // udf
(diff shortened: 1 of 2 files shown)


Yohahaha:
@zhli1142015 @zhouyuan please help review this, thank you!

zhztheplayer:
> Change default value of spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver to 0, make startup message more clear.

Would you like to share the reason of doing this? Thanks!

zhztheplayer (approved):

Yohahaha:
> > Change default value of spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver to 0
> 
> Would you like to share the reason of doing this? Thanks!

preload split does not fit Gluten's design perfectly, disable it by default avoid unexpected behavior.

zhztheplayer (changes requested):
> preload split does not fit Gluten's design perfectly, disable it by default avoid unexpected behavior.

Can you elaborate? Like it may lead memory issue or something?

Yohahaha:
> > preload split does not fit Gluten's design perfectly, disable it by default avoid unexpected behavior.
> 
> Can you elaborate? Like it may lead to memory issue or something?

Gluten use single thread execution mode and suppose all Velox object should be released when Spark task is finished, but the thread which preload split will run out live of Velox Task and it may access dangling pointer.
https://github.com/facebookincubator/velox/issues/6994
https://github.com/prestodb/presto/issues/20020

Yohahaha:
> Hi @Yohahaha yes, hongze is working on this, the split preload feature will be fixed by this: #3419

I think current patch does not change functionality.

PR #3634, opened 2023-11-06
[CI] Improve CI to auto close stale patch

zhouyuan:
## What changes were proposed in this pull request?

Auto close stale pull request after 45 days

## How was this patch tested?

not tests required

PR #3699, opened 2023-11-14
[CI] increase operations per run for stale action

zhouyuan:

follow up #3634 
Increase the operations per run to 300 

The default value is 30, which seems too small for gluten project

## How was this patch tested?

no tests required


PHILO-HE (approved):
LGTM!

PR #3721, opened 2023-11-15
[GLUTEN-3719][VL] Introduce VeloxIntermediateData to adjust velox agg func intermediate data

liujiayi771:

If there are inconsistencies in the order and type of intermediate data between Velox and Spark aggregation functions, special handling is required. We can introduce a `VeloxIntermediateData` object to handle these cases. For non-special cases, we can continue using the `aggBufferAttributes` of the aggregation function without any special matching. In future PRs, the remaining methods for handling such aggregation functions will be incorporated into this VeloxIntermediateData object.

In the `applyExtractStruct` function, a lot of code was written to match the intermediate data outputted by Velox with the column data in Spark's agg buffer. These code segments involved many index order adjustments, which made them difficult to read and understand why such ordering was necessary. For example (It is difficult to understand the significance of [1, 4, 5, 0, 2, 3]),
```scala
case _: Corr =>
  // Select count from Velox struct with count casted from LongType into DoubleType.
  expressionNodes.add(
    ExpressionBuilder
      .makeCast(
        ConverterUtils.getTypeNode(DoubleType, nullable = false),
        ExpressionBuilder.makeSelection(colIdx, 1),
        SQLConf.get.ansiEnabled))
  expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 4))
  expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 5))
  expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 0))
  expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 2))
  expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 3))
```
In this PR, all these code segments have been modified and improved.

## How was this patch tested?

Exists CI


liujiayi771:
@rui-mo @PHILO-HE Could you help review this PR?

backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
```diff
@@ -133,56 +123,29 @@ case class HashAggregateExecTransformer(
...
-          expressionNodes.add(
-            ExpressionBuilder
-              .makeCast(
-                ConverterUtils.getTypeNode(DoubleType, nullable = false),
-                ExpressionBuilder.makeSelection(colIdx, 1),
-                SQLConf.get.ansiEnabled))
-          expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 2))
-          expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 3))
-          expressionNodes.add(ExpressionBuilder.makeSelection(colIdx, 0))
+        case _ @VeloxIntermediateData.Type(veloxTypes: Seq[DataType]) =>
```
rui-mo, line 128: I'm confused here. Why can we match `aggregateFunction` against `VeloxIntermediateData.Type`?

backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
```diff
@@ -133,56 +123,29 @@ case class HashAggregateExecTransformer(
...
+          val (sparkOrders, sparkTypes) =
+            aggFunc.aggBufferAttributes.map(attr => (attr.name, attr.dataType)).unzip
+          val veloxOrders = VeloxIntermediateData.veloxIntermediateDataOrder(aggFunc)
+          val adjustedOrders = sparkOrders.map(veloxOrders.indexOf(_))
```
rui-mo, line 132: Is it enough to decide the order based on name equality? E.g., if attr.name contains suffix of exprId, would it fail to match with the string in veloxOrders?

backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
liujiayi771: It will use `VeloxIntermediateData.Type.unapply` method to extract `veloxTypes` from  aggFunc. This is equivalent to `val veloxTypes = VeloxIntermediateData.Type.unapply(aggFunc)`

backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
liujiayi771: I think the column names in `aggBufferAttributes` are fixed, it will not contains suffix of exprId. You can check the implementation of each agg. For example,
https://github.com/apache/spark/blob/22932385df3f5df216dcc510b4cdecad343f7642/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Average.scala#L82-L83

https://github.com/apache/spark/blob/22932385df3f5df216dcc510b4cdecad343f7642/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Sum.scala#L84-L86

backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
```diff
@@ -133,56 +123,29 @@ case class HashAggregateExecTransformer(
         case _ =>
           throw new UnsupportedOperationException(s"${expr.mode} not supported.")
       }
+      val aggFunc = expr.aggregateFunction
       expr.aggregateFunction match {
```
liujiayi771, line 127: Use aggFunc defined in the previous line.

rui-mo (commented):
Thanks for providing these details. Will merge after internal CI passes.

rui-mo (approved):

Issue #3705, opened 2023-11-14
[CORE] Support mapping one custom aggregate function to more than one backend functions


zzcclp:
### Description

Support mapping one custom aggregate function to more than one backend functions, like first/last function, they will be mapped to two backend function names according to the `ignoreNulls` parameter.

PR #3734, opened 2023-11-15
[GLUTEN-3705][FOLLOW][CH] Set the correct agg schema names after mapping one agg function to the other name

zzcclp:

With CH backend, in the final stage, the agg schema names must be the `agg_function#exprId#Partial#custom_sum`, after mapping the agg function to the other name, it does not modify according to the new one.

(Fixes: \#3705)

## How was this patch tested?


baibaichen (approved):
LGTM

Issue #3718, opened 2023-11-15
[VL] The latest code is very slow when reading HDFS data


liujiayi771:
### Backend

VL (Velox)

### Bug description

We test the latest code in TPCDS 1T in the same cluster and same configurations. Some queries like q74 q75 become very slow.
<img width="1122" alt="image" src="https://github.com/oap-project/gluten/assets/13622031/df0e9519-4780-4d5f-852f-6adf1c733e68">

### Spark version

Spark-3.3.x

rui-mo:
cc @zhouyuan

rui-mo:
The test is based on 5dabb5b66bd180cfc541c3ab652a863630481608.

zhouyuan:
Hi @liujiayi771 thanks for reporting, the notable changes on HDFS connector is https://github.com/facebookincubator/velox/pull/6140 which should improve the performance when reading.
Based on your description, it seems this only affects on some queries, so could this due to some special filter push down? 

I just checked TPCDS Q9 on our local cluster, it seems the performance on scan is not impacted a lot.  see below capture on new velox vs. old velox 
 
![Screenshot 2023-11-15 145418](https://github.com/oap-project/gluten/assets/43917/9b4000d5-2866-4d64-8d45-afbb73b3d981)

thanks, -yuan

zhouyuan:
@liujiayi771 btw, it seems the "total time scan total" on your new velox is "0ms", this seems not correct. 

-yuan

liujiayi771:
> Based on your description, it seems this only affects on some queries, so could this due to some special filter push down?
> 
> I just checked TPCDS Q9 on our local cluster, it seems the performance on scan is not impacted a lot. see below capture on new velox vs. old velox

@rui-mo checked the metrics code in gluten, it has some problems.

liujiayi771:
@zhouyuan Not only these queries, but many others have become slow as well. I have shared the test results with @rui-mo . Additionally, there is another scenario where if only queries 72-75 are executed, the execution time is normal. However, when running continuously from q1 to q75, it becomes very slow, indicating a possible memory leak issue.

zhouyuan:
[...]

PR #3740, opened 2023-11-16
[GLUTEN-3718][VL] Set connector file handle cache size to 20

Yohahaha:
20 is big enough for most case.

(Fixes: \#3718)


Yohahaha:
disable file handle is better.

PR #3758, opened 2023-11-17
[VL] Support s3 write

yma11:

Support s3 write for Velox backend

## How was this patch tested?

manually tested

### DIFF
--- a/cpp/velox/compute/VeloxRuntime.cc
+++ b/cpp/velox/compute/VeloxRuntime.cc
@@ -160,7 +160,11 @@ std::shared_ptr<Datasource> VeloxRuntime::createDatasource(
     MemoryManager* memoryManager,
     std::shared_ptr<arrow::Schema> schema) {
   auto veloxPool = getAggregateVeloxPool(memoryManager);
-  return std::make_shared<VeloxParquetDatasource>(filePath, veloxPool, schema);
+  // Pass a dedicate pool for S3 sink as can't share veloxPool
+  // with parquet writer.
+  auto s3SinkPool = getLeafVeloxPool(memoryManager);
+
+  return std::make_shared<VeloxParquetDatasource>(filePath, veloxPool, s3SinkPool, schema);
 }
 
 std::shared_ptr<ShuffleReader> VeloxRuntime::createShuffleReader(
--- a/cpp/velox/operators/writer/VeloxParquetDatasource.h
+++ b/cpp/velox/operators/writer/VeloxParquetDatasource.h
@@ -32,6 +32,10 @@
 #include "operators/writer/Datasource.h"
 
 #include "velox/common/file/FileSystems.h"
+#ifdef ENABLE_S3
+#include "velox/connectors/hive/storage_adapters/s3fs/S3FileSystem.h"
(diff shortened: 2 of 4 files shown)


cpp/velox/compute/VeloxRuntime.cc
```diff
@@ -152,7 +152,9 @@ std::shared_ptr<Datasource> VeloxRuntime::createDatasource(
     MemoryManager* memoryManager,
     std::shared_ptr<arrow::Schema> schema) {
   auto veloxPool = getAggregateVeloxPool(memoryManager);
-  return std::make_shared<VeloxParquetDatasource>(filePath, veloxPool, schema);
+  auto s3SinkPool = getLeafVeloxPool(memoryManager);
```
zhouyuan, line 155: nit: maybe it's worth to leave a note here to explain why this is necessary?

cpp/velox/operators/writer/VeloxParquetDatasource.h
```diff
@@ -49,8 +53,13 @@ class VeloxParquetDatasource final : public Datasource {
...
       const std::string& filePath,
       std::shared_ptr<facebook::velox::memory::MemoryPool> veloxPool,
+      std::shared_ptr<facebook::velox::memory::MemoryPool> s3SinkPool,
       std::shared_ptr<arrow::Schema> schema)
-      : Datasource(filePath, schema), filePath_(filePath), schema_(schema), pool_(std::move(veloxPool)) {}
+      : Datasource(filePath, schema),
+        filePath_(filePath),
+        schema_(schema),
+        pool_(std::move(veloxPool)),
+        s3SinkPool_(std::move(s3SinkPool)) {}
```
zhouyuan, line 62: This looks like adding the pool for all backends, including hdfs/local?

zhouyuan (commented):
Thank you for adding this feature!

cpp/velox/operators/writer/VeloxParquetDatasource.h
yma11: Yes. But this pool is actually not new created and it's only used in S3.

zhouyuan (approved):
👍

zhouyuan:
I'll follow up to add s3 write job in nightly run.

dcoliversun:
@yma11 @zhouyuan hi, I test s3 write with spark [InsertIntoHiveTableBenchmark](https://github.com/apache/spark/blob/master/sql/hive/src/test/scala/org/apache/spark/sql/execution/benchmark/InsertIntoHiveTableBenchmark.scala), and InsertIntoHadoopFsRelationCommand Node is fallback.
```plain
== Fallback Summary ==
(1) Range: Gluten does not touch it or does not support it
(6) Execute InsertIntoHadoopFsRelationCommand: Gluten does not touch it or does not support it

== Physical Plan ==
Execute InsertIntoHadoopFsRelationCommand (6)
+- VeloxColumnarToRowExec (5)
   +- ^ ProjectExecTransformer (3)
      +- RowToVeloxColumnar (2)
         +- Range (1)
```
Correct me if wrong. I think it still use s3 java sdk to write.  How should I write parquet into s3 with cpp sdk?

PR #3761, opened 2023-11-17
[VL] Activate random kill tasks GHA CI job

zhztheplayer:
Enable random-kill-task job to avoid stability regressions, e.g. core dump on task-killing.

The random-kill-task job: Use gluten-it's random task killing feature to emulate task kill requests randomly when a query is running. The query still has to succeed, without crashes or fatal errors since the killed tasks will be rescheduled by Spark's task scheduler.


PHILO-HE (approved):
Looks good! Thanks!

zhztheplayer:
Thanks for reviewing!

PR #3762, opened 2023-11-17
[VL] Activate CI test for distinct aggregation spill

zhztheplayer:
We had added a low memory CI job for TPC-DS q97 which was failing because at the time spilling support was not added for distinct aggregation. Enable it now to make sure the newly added spilling support could work.

Update: The job is still failing due to some other reasons although distinct spilling is functional. Adding a independent UT instead.


zhouyuan (approved):
👍

PR #3763, opened 2023-11-17
[GLUTEN-3760][VL]fix unsupported column name corner case

zhli1142015:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/FileSourceScanExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/FileSourceScanExecTransformer.scala
@@ -108,8 +108,9 @@ class FileSourceScanExecTransformer(
   override def hashCode(): Int = super.hashCode()
 
   override protected def doValidateInternal(): ValidationResult = {
-    if (hasUnsupportedColumns) {
-      return ValidationResult.notOk(s"Unsupported columns in native scan.")
+    if (!unsupportedColumns.isEmpty) {
+      return ValidationResult.notOk(
+        s"Unsupported columns in native scan: ${unsupportedColumns.mkString(",")}")
     }
 
     if (hasFieldIds) {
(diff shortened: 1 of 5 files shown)


zhli1142015:
@JkSelf  and @rui-mo, could you please help check this? Thanks.

gluten-core/src/main/scala/io/glutenproject/execution/FileSourceScanExecTransformer.scala
```diff
@@ -108,8 +108,9 @@ class FileSourceScanExecTransformer(
   override def hashCode(): Int = super.hashCode()
 
   override protected def doValidateInternal(): ValidationResult = {
-    if (hasMetadataColumns) {
-      return ValidationResult.notOk(s"Unsupported metadataColumns scan in native.")
+    if (!unsupportedColumns.isEmpty) {
+      return ValidationResult.notOk(
+        s"Unsupported columns in native scan: ${unsupportedColumns.mkString(",")}")
```
JkSelf, line 113: @zhli1142015 Can we modify the reason for unsupport to: 'Unsupported columns in native scan due to column name ambiguity in Velox. Please consider renaming the column.'? It is more readable.

JkSelf (commented):
LGTM. Just one minor comment.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/FileSourceScanExecShim.scala
```diff
@@ -53,6 +53,20 @@ class FileSourceScanExecShim(
   // Note: "metrics" is made transient to avoid sending driver-side metrics to tasks.
   @transient override lazy val metrics: Map[String, SQLMetric] = Map()
 
+  val unsupportedColumns: Seq[String] = {
```
rui-mo, line 56: FileSourceScan targets on datasource v1, do we have similar issue for datasource v2? If so, can this part of logic be put in `BasicScanExecTransformer`? Maybe by adding a function receives output as parameter.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/FileSourceScanExecShim.scala
```diff
@@ -53,6 +53,20 @@ class FileSourceScanExecShim(
...
+    // Below name has special meaning in Velox.
```
rui-mo, line 57: What is the special meaning, can we add more details?

zhli1142015:
Will submit another PR for this.

PR #3774, opened 2023-11-20
[VL] Remove corr in group-by.sql

marin-ma:
Velox `corr` has better computation logic but it fails Spark's precision check. Based on the discussion in https://github.com/facebookincubator/velox/pull/7204, we decide to use original Velox `corr` and mute the precision check failure in Gluten UT.

This PR contains following changes:
1. Moved supported SQL tests into Backends API `TestApi`.
2. Copied Spark's `inputs/group-by.sql`, `input/udf/udf-group-by.sql` and `results/group-by.sql.out`, `results/udf/udf-group-by.sql.out` and removed the query contains "corr".
3. Fixed SQL tests file name matching, using `equals` instead of `contains`.


marin-ma:
@PHILO-HE @rui-mo Could you help to review? Thanks!

gluten-core/src/main/scala/io/glutenproject/backendsapi/Backend.scala
```diff
@@ -38,4 +38,6 @@ trait Backend {
   def broadcastApi(): BroadcastApi
 
   def settings(): BackendSettingsApi
+
+  def testApi(): TestApi
```
PHILO-HE, line 42: Hi @marin-ma, as this interface is just for test use, can we move it to some place in test module? Maybe, into `BackendTestSettings`? Thanks!

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/GlutenSQLQueryTestSuite.scala
```diff
@@ -492,9 +307,9 @@ class GlutenSQLQueryTestSuite
     // If a test case is not in the test list, or it is in the ignore list, ignore this test case.
     if (
       !supportedList.exists(
-        t => testCase.name.toLowerCase(Locale.ROOT).contains(t.toLowerCase(Locale.ROOT))) ||
+        t => testCase.name.toLowerCase(Locale.ROOT).equals(t.toLowerCase(Locale.ROOT))) ||
       ignoreList.exists(
-        t => testCase.name.toLowerCase(Locale.ROOT).contains(t.toLowerCase(Locale.ROOT)))
+        t => testCase.name.toLowerCase(Locale.ROOT).equals(t.toLowerCase(Locale.ROOT)))
```
rui-mo, line 312: With this change, I suppose we wil be missing some sql files under subdirectory. Can we specially handle group-by.sql only?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/TestApiImpl.scala
```diff
@@ -0,0 +1,415 @@
...
+  override def getOverwriteSQLQueryTests(sparkVersion: SparkVersion): Set[String] = {
+    sparkVersion match {
+      case SparkVersion32() =>
+        OVERWRITE_SQL_QUERY_LIST_SPARK32
+      case SparkVersion33() =>
+        OVERWRITE_SQL_QUERY_LIST_SPARK33
+    }
+  }
+
+  private val SUPPORTED_SQL_QUERY_LIST_SPARK32: Set[String] =
```
marin-ma, line 46: @rui-mo Do you mean some of the filenames listed here might not specifying the subdirectory? If so, we should add the missing subdirectories.

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/GlutenSQLQueryTestSuite.scala
marin-ma: Using `contains` can lead to some unexpected files being added. For example, `group-by.sql` was in the supported list but `udf/udf-group-by.sql` wasn't. But the latter one was also tested. If we are sure that one file is supported, it should be explicitly added to the supported list.

rui-mo: Yes, that's just what I mean. With this change, we may be missing some tests for the files under subdirectory. There are some other sql files under subdirectory but share the same name with those in root directory.

rui-mo: Can we keep using `contains` and exclude `udf/udf-group-by.sql`? If there are more test cases in `udf/udf-group-by.sql`, maybe we can merge them into the `group-by.sql` added by ourselves.
This can make sure other tests are not missing.

marin-ma: Is it intended that we use `contains` to make `udf/udf-group-by.sql` a test? I think it would be better to explicitly add it to the supported list. Moreover, the current supported list inconsistently mixes some files specified with subdirectories and others with only the filename. It would be confusing for others who are trying to add new files to the list.

marin-ma:
@rui-mo I would suggest that I perform a final check to ensure that all files belonging to any subdirectories are specified correctly, and to make sure no files are missing. Also, we should use `equals` instead of `contains` to eliminate any uncertainty. What do you think?

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/GlutenSQLQueryTestSuite.scala
rui-mo: The way to use `contains` derives from Spark SQLQueryTestSuite. It makes sense to me if we need to change it in Gluten test.

rui-mo:
@marin-ma I think it makes sense as long as we don't miss any test.

marin-ma:
@rui-mo @PHILO-HE Could you help to review again? Thanks!

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/GlutenSQLQueryTestSuite.scala
```diff
@@ -768,35 +583,38 @@ class GlutenSQLQueryTestSuite
...
+      } else {
+        RegularTestCase(testCaseName, absPath, resultFile) :: Nil
+      }
     }
+    val overwriteTestCases = listFilesRecursively(new File(overwriteInputFilePath))
+      .flatMap(createTestCase(_, overwriteInputFilePath, overwriteGoldenFilePath))
+    val overwriteTestCaseNames = overwriteTestCases.map(_.name)
+    listFilesRecursively(new File(inputFilePath))
+      .flatMap(createTestCase(_, inputFilePath, goldenFilePath))
+      .filterNot(testCase => overwriteTestCaseNames.contains(testCase.name)) ++ overwriteTestCases
```
rui-mo, line 617: What are the thoughts here to use contains instead of equals?

marin-ma: overwriteTestCaseNames is `Seq[String]`, here it checks whether the `testCase.name`   is in the list.

rui-mo (approved):
Thanks. I believe commit https://github.com/oap-project/velox/commit/64b00b0d0e6125cd92b949a87f51e5e3319c751f could be removed after this PR. cc @JkSelf

PHILO-HE (approved):
Looks good!

marin-ma:
> Thanks. I believe commit [oap-project/velox@64b00b0](https://github.com/oap-project/velox/commit/64b00b0d0e6125cd92b949a87f51e5e3319c751f) could be removed after this PR. cc @JkSelf

Yes. The test PR to remove that commit also passed CI. https://github.com/oap-project/gluten/pull/3783

PR #3775, opened 2023-11-20
[VL] Align whole stage pipeline metric with Spark

Yohahaha:
Spark measure pipeline time since the iterator was created, but Gluten measure it in each call.

```
  /**
   * Returns the elapsed time since this object is created. This object represents a pipeline so
   * this is a measure of how long the pipeline has been running.
   */
  public long durationMs() {
    return (System.nanoTime() - startTimeNs) / (1000 * 1000);
  }
```

Following patch will add another metric to reflect native task execution time.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -47,7 +47,6 @@ import java.net.URLDecoder
 import java.nio.charset.StandardCharsets
 import java.time.ZoneOffset
 import java.util.{ArrayList => JArrayList, HashMap => JHashMap, Map => JMap}
-import java.util.concurrent.TimeUnit
 
 import scala.collection.JavaConverters._
 
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -129,7 +129,7 @@ class IteratorApiImpl extends IteratorApi with Logging {
       updateInputMetrics: (InputMetricsWrapper) => Unit,
       updateNativeMetrics: IMetrics => Unit,
       inputIterators: Seq[Iterator[ColumnarBatch]] = Seq()): Iterator[ColumnarBatch] = {
-    val beforeBuild = System.nanoTime()
+    val startTime = System.nanoTime()
     val columnarNativeIterators =
       new JArrayList[GeneralInIterator](inputIterators.map {
         iter => new ColumnarBatchInIterator(iter.asJava)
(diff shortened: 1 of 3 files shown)


zhztheplayer:
Any visible difference on the pipeline metrics after making this change?

@rui-mo

Yohahaha:
> Any visible difference on the pipeline metrics after making this change?
> 
> cc @rui-mo

before it record elapsed time of `hasNext` and `next`, after it record from iterator created to last `hasNext` align with Spark's behavior.

zhztheplayer:
> > Any visible difference on the pipeline metrics after making this change?
> > cc @rui-mo
> 
> before it record elapsed time of `hasNext` and `next`, after it record from iterator created to last `hasNext` align with Spark's behavior.

OK. So did you observe any difference on the metric numbers when running your queries from Spark GUI?

Yohahaha:
> > > Any visible difference on the pipeline metrics after making this change?
> > > cc @rui-mo
> > 
> > 
> > before it record elapsed time of `hasNext` and `next`, after it record from iterator created to last `hasNext` align with Spark's behavior.
> 
> OK. So did you observe any difference on the metric numbers when running your queries from Spark GUI?

Yes, sometimes stage UI show no difference between Gluten and Vanilla, but pipeline metric show Gluten has benefits. We want align metric between Gluten and Vanilla as much as possible, then find which part is slower.

zzcclp:
It seems that the changes do not impact the ch backend. `IteratorCompleter` is only for the velox backend.

rui-mo:
> before it record elapsed time of hasNext and next, after it record from iterator created to last hasNext align with Spark's behavior

@Yohahaha Compared with previous logic, looks more time is recorded in this metric, right?

Yohahaha:
> > before it record elapsed time of hasNext and next, after it record from iterator created to last hasNext align with Spark's behavior
> 
> @Yohahaha Compared with previous logic, looks more time is recorded in this metric, right?

Yes, previous logic more like record velox task execution time, but we may not ignore other parts' time.

rui-mo (approved):

Yohahaha:
could we merge?

Yohahaha:
@PHILO-HE @ulysses-you could you help merge? thanks!

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
```diff
@@ -184,8 +179,6 @@ class IteratorApiImpl extends IteratorApi with Logging {
         rootNode.toProtobuf.toByteArray,
         columnarNativeIterator)
 
-    pipelineTime += TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - beforeBuild)
```
ulysses-you, line 187: thanky you @Yohahaha . Shall we add one more metrics for the durationMs of preparing the native input iterator ? So the previous pipeline time can be broken into two parts, one is for preparing and the other is for executing.

Yohahaha: ok, I can open an issue to track it.

ulysses-you:
this pr does not affect CH backend, FWI @zzcclp

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
Yohahaha: https://github.com/oap-project/gluten/issues/3830

Issue #3653, opened 2023-11-09
SHJ metrics are updated wrongly when there is no shuffle on build side


rui-mo:
### Backend

VL (Velox)

### Bug description

https://github.com/oap-project/gluten/pull/3618#discussion_r1384692726
Could be reproduced with TPC-H q20 on Spark 3.3, with bucket scan, v2 datasource.

### Spark version

Spark-3.3.x

PR #3782, opened 2023-11-20
[GLUTEN-3653][VL] Fix SHJ metrics when there is no shuffle on build side

rui-mo:

If there is no shuffle on build side, we need to put the node id of stream input node together with join node id. Otherwise, the join metrics become non-adjacent, and cannot be updated in the correct order.

(Fixes: \#3653)

## How was this patch tested?

under test


rui-mo:
cc @zzcclp

zzcclp:
it's better use bucket tables case to add a ut for this :)

rui-mo:
InputIteratorTransformer was introduced, so current logic works for this case.

PR #3786, opened 2023-11-21
[VL] update pacakge.sh for spark34

zhouyuan:

update pacakge.sh to cover spark34

## How was this patch tested?

no tests required


JkSelf (approved):
Good catch. Thanks.

PR #3791, opened 2023-11-21
[VL] Support IsNull filter pushdown

rui-mo:

Previously, is_null filter was separated into remaining filter. With this PR, it can be pushed down into subfield filter.

Before:
> TableScan[table: hive_table, range filters: [(ss_item_sk, Filter(IsNotNull, deterministic, null not allowed))], remaining filter: (isnull("ss_addr_sk"))] -> n0_0:INTEGER, n0_1:INTEGER, n0_2:DECIMAL(7,2), n0_3:INTEGER

After:
> TableScan[table: hive_table, range filters: [(ss_item_sk, Filter(IsNotNull, deterministic, null not allowed)), (ss_store_sk, Filter(IsNull, deterministic, null allowed))]] -> n2_0:INTEGER, n2_1:INTEGER, n2_2:DECIMAL(7, 2), n2_3:INTEGER

## How was this patch tested?

Verified with TPC-DS q76.

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -94,6 +94,7 @@ std::string getMax<std::string>() {
 
 // Substrait function names.
 const std::string sIsNotNull = "is_not_null";
+const std::string sIsNull = "is_null";
 const std::string sGte = "gte";
 const std::string sGt = "gt";
 const std::string sLte = "lte";
@@ -1275,15 +1276,15 @@ bool SubstraitToVeloxPlanConverter::childrenFunctionsOnSameField(
   return false;
 }
 
-bool SubstraitToVeloxPlanConverter::canPushdownCommonFunction(
+bool SubstraitToVeloxPlanConverter::canPushdownFunction(
     const ::substrait::Expression_ScalarFunction& scalarFunction,
     const std::string& filterName,
     uint32_t& fieldIdx) {
   // Condtions can be pushed down.
-  static const std::unordered_set<std::string> supportedCommonFunctions = {sIsNotNull, sGte, sGt, sLte, sLt, sEqual};
+  static const std::unordered_set<std::string> supportedFunctions = {sIsNotNull, sIsNull, sGte, sGt, sLte, sLt, sEqual};
 
   bool canPushdown = false;
-  if (supportedCommonFunctions.find(filterName) != supportedCommonFunctions.end() &&
+  if (supportedFunctions.find(filterName) != supportedFunctions.end() &&
       fieldOrWithLiteral(scalarFunction.arguments(), fieldIdx)) {
     // The arg should be field or field with literal.
     canPushdown = true;
@@ -1417,7 +1418,7 @@ void SubstraitToVeloxPlanConverter::separateFilters(
     } else {
       // Check if the condition is supported to be pushed down.
       uint32_t fieldIdx;
-      if (canPushdownCommonFunction(scalarFunction, filterName, fieldIdx) &&
+      if (canPushdownFunction(scalarFunction, filterName, fieldIdx) &&
           rangeRecorders.at(fieldIdx).setCertainRangeForFunction(filterName)) {
         subfieldFunctions.emplace_back(scalarFunction);
       } else {
@@ -1465,6 +1466,12 @@ bool SubstraitToVeloxPlanConverter::RangeRecorder::setCertainRangeForFunction(
       // Is not null can always coexist with the other range.
       return true;
     }
+  } else if (functionName == sIsNull) {
+    if (reverse) {
+      return setCertainRangeForFunction(sIsNotNull, false, forOrRelation);
+    } else {
+      return setIsNull();
+    }
   } else {
     return false;
   }
@@ -1477,9 +1484,16 @@ void SubstraitToVeloxPlanConverter::setColumnFilterInfo(
     bool reverse) {
   if (filterName == sIsNotNull) {
     if (reverse) {
-      VELOX_NYI("Reverse not supported for filter name '{}'", filterName);
+      columnFilterInfo.setNull();
+    } else {
+      columnFilterInfo.forbidsNull();
     }
-    columnFilterInfo.forbidsNull();
+  } else if (filterName == sIsNull) {
+    if (reverse) {
+      columnFilterInfo.forbidsNull();
+    } else {
+      columnFilterInfo.setNull();
+    }
   } else if (filterName == sGte) {
     if (reverse) {
       columnFilterInfo.setUpper(literalVariant, true);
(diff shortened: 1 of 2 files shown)


zhli1142015:
Thanks, @rui-mo , with this fix, we should not need to change velox, is this right?

rui-mo:
@zhli1142015 That's true. But considering there's a similar function in Velox to extract conditions from remaining filter to subfield filter, maybe we can shift to that one when it covers enough cases in the long term. As for now, it looks all right to continue to improve Velox's implementation.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1472,9 +1479,14 @@ void SubstraitToVeloxPlanConverter::setColumnFilterInfo(
     bool reverse) {
   if (filterName == sIsNotNull) {
     if (reverse) {
```
marin-ma, line 1481: A bit confused here. Should it be:
```
    if (reverse) {
      columnFilterInfo.setNull();
    } else {
      columnFilterInfo.forbidsNull();
    }
```

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1472,9 +1479,14 @@ void SubstraitToVeloxPlanConverter::setColumnFilterInfo(
...
-      VELOX_NYI("Reverse not supported for filter name '{}'", filterName);
+      columnFilterInfo.setNull();
     }
     columnFilterInfo.forbidsNull();
+  } else if (filterName == sIsNull) {
```
marin-ma, line 1485: Same question as above.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
rui-mo: Thanks for catching this. Fixed.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1861,9 +1879,12 @@ void SubstraitToVeloxPlanConverter::constructSubfieldFilters(
...
     // Handle null filtering.
-    if (rangeSize == 0 && !nullAllowed) {
-      std::unique_ptr<common::IsNotNull> filter = std::make_unique<common::IsNotNull>();
-      filters[common::Subfield(inputName)] = std::move(filter);
+    if (rangeSize == 0) {
+      if (!nullAllowed) {
+        filters[common::Subfield(inputName)] = std::move(std::make_unique<common::IsNotNull>());
+      } else if (isNull) {
+        filters[common::Subfield(inputName)] = std::move(std::make_unique<common::IsNull>());
+      }
```
marin-ma, line 1887: Do we also need to add this check here? https://github.com/oap-project/gluten/pull/3791/files#diff-08ec1ecb10e48cd471188daf2152228da4b00a81c16acaf2e87b040fec16f8c5R1843

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -220,23 +220,33 @@ class SubstraitToVeloxPlanConverter {
...
         return false;
       }
       multiRange_ = true;
       return true;
     }
 
+    /// Set the existence of IsNull and returns whether it can coexist with
+    /// existing conditions for this field.
+    bool setIsNull() {
+      if (inRange_ || multiRange_ || leftBound_ || rightBound_) {
```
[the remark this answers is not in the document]
rui-mo: Yes, only one of them can be pushed down because IsNull cannot coexist with other range filter for a single field. But I doubt whether Spark generates such conditions. For query like `select l_returnflag from lineitem where l_returnflag > 1 and l_returnflag is null`, Spark only produces IsNotNull and IsNull conditions.
> (1) NativeScan parquet 
Output [1]: [l_returnflag#8]
Batched: true
Location: InMemoryFileIndex [hdfs://sr245:9001/tpch_date_10/lineitem]
PushedFilters: [IsNotNull(l_returnflag), IsNull(l_returnflag)]
ReadSchema: struct<l_returnflag:string>

cpp/velox/substrait/SubstraitToVeloxPlan.cc
rui-mo: Updated, thanks.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1812,17 +1824,21 @@ void SubstraitToVeloxPlanConverter::constructSubfieldFilters(
...
-      filters[common::Subfield(inputName)] = std::move(std::make_unique<common::IsNotNull>());
-    } else {
-      VELOX_NYI("constructSubfieldFilters only support IsNotNull for input type '{}'", inputType);
+    if (rangeSize == 0) {
+      if (!nullAllowed) {
+        filters[common::Subfield(inputName)] = std::move(std::make_unique<common::IsNotNull>());
+      } else if (isNull) {
+        filters[common::Subfield(inputName)] = std::move(std::make_unique<common::IsNull>());
+      }
+      VELOX_NYI("constructSubfieldFilters only support IsNotNull and IsNull for input type '{}'", inputType);
```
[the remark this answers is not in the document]
rui-mo: Thanks for catching this. Updated.

zhli1142015 (approved):
LGTM.

cpp/velox/substrait/SubstraitToVeloxPlan.h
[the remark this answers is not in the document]
marin-ma: @rui-mo The datatype for `l_returnflag` is string. If replacing `l_returnflag` with `l_orderkey` using Vanilla spark:
`spark.sql("select l_returnflag from lineitem where l_orderkey > 1 and l_orderkey is null")`

`df.explain()` prints
```
== Physical Plan ==
*(1) Project [l_returnflag#8]
+- *(1) Filter ((isnotnull(l_orderkey#0L) AND (l_orderkey#0L > 1)) AND isnull(l_orderkey#0L))
   +- *(1) ColumnarToRow
      +- FileScan parquet [l_orderkey#0L,l_returnflag#8] Batched: true, DataFilters: [isnotnull(l_orderkey#0L), (l_orderkey#0L > 1), isnull(l_orderkey#0L)], Format: Parquet, Location: InMemoryFileIndex(1 paths)[hdfs://user-S520-B3.sh.intel.com:8020/tpch_sf2000_parquet_snappy/linei..., PartitionFilters: [], PushedFilters: [IsNotNull(l_orderkey), GreaterThan(l_orderkey,1), IsNull(l_orderkey)], ReadSchema: struct<l_orderkey:bigint,l_returnflag:string>
```

And `df.collect()` prints
```
res11: Array[org.apache.spark.sql.Row] = Array()
```

It looks like both "IsNull" and "GreaterThan" are pushed down.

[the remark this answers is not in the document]
rui-mo: OK. Good to know!

Issue #3794, opened 2023-11-21
[VL] Remove GLUTEN_PRINT_DEBUG macro in native side


Yohahaha:
### Description

Replace GLUTEN_PRINT_DEBUG macro with Scala config, such as `spark.gluten.sql.debug`, let user easily print debug info without recompile.

PR #3803, opened 2023-11-22
[GLUTEN-3794][VL] Remove GLUTEN_PRINT_DEBUG usage

Yohahaha:
Remove GLUTEN_PRINT_DEBUG usage, avoid untested compile error.

(Fixes: \#3794)


cpp/core/utils/Print.h
```diff
@@ -1,298 +0,0 @@
...
-
-template <typename A, typename B>
-static inline void printEq(const A& a, const B& b) {
-  std::cout << a << " = " << b;
-}
-
-template <typename A, typename B>
-static inline void printEqlf(const A& a, const B& b) {
-  std::cout << a << " = " << b << std::endl;
-}
```
zhztheplayer, line 70: So there is no code relying on these APIs as of now?

cc @zuochunwei

Yohahaha: just wait CI.

PR #3807, opened 2023-11-22
[GLUTEN-3695][VL] Support building 3rd-party libs on Arm64

guyuqi:
In some scenarios, we need to build our own thirdparty-lib jar on **Arm64** when `spark.gluten.loadLibFromJar=true`

## How was this patch tested?

Build package:
`./dev/package.sh`


PHILO-HE (commented):
There is a CI failure. Please rebase the code and let's see the feedback again. Thanks!

PHILO-HE (approved):
Thanks for your fix!

PR #3809, opened 2023-11-22
[VL] Dump memory stats when memory leak is found during releasing native memory manager

zhztheplayer:


zhztheplayer:
The solution is invalid. Closing.

Issue #2414, opened 2023-07-20
Protocol message had too many levels of nesting


loukey-lj:
**Describe the bug**
org.apache.spark.SparkException: Job aborted due to stage failure: Task 5 in stage 2.0 failed 4 times, most recent failure: Lost task 5.3 in stage 2.0 (TID 17) (cnsz26plk8y0 executor 2): ja
va.lang.RuntimeException: Unable to understand proto buffer
      at io.glutenproject.shaded.com.google.protobuf.GeneratedMessageLite$SerializedForm.readResolve(GeneratedMessageLite.java:1463)
      at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
      at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
      at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
      at java.lang.reflect.Method.invoke(Method.java:498)
      at java.io.ObjectStreamClass.invokeReadResolve(ObjectStreamClass.java:1148)
      at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1817)
      at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
      at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2018)
      at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1942)
      at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1808)
      at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
      at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2018)
      at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1942)
      at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1808)
      at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
      at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2018)
      at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1942)
      at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1808)
      at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1353)
[...]

PR #3814, opened 2023-11-22
[CORE] Introduce custom protobuf dependency with recursion limit enlarged

PHILO-HE:

Protobuf has a hard-code limit (100) for recursion depth and runtime exception can occur when exceeding the limit. This PR proposes a custom protobuf dependency which has a much larger limit (100, 000, see source [here](https://github.com/oap-project/protobuf/tree/v3.23.4-0)). Thanks to @weiting-chen, the dependency jar has been published in maven repo (see [link](https://repo1.maven.org/maven2/io/glutenproject/protobuf-java/3.23.4-0/)).

Fixes: #1727, #2414. 

## How was this patch tested?

Existing tests.

### DIFF
--- a/gluten-core/pom.xml
+++ b/gluten-core/pom.xml
@@ -177,10 +177,9 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>com.google.protobuf</groupId>
+      <groupId>io.glutenproject</groupId>
       <artifactId>protobuf-java-util</artifactId>
-      <version>${protobuf.version}</version>
-      <scope>provided</scope>
+      <version>${custom.protobuf.version}</version>
     </dependency>
     <dependency>
       <groupId>commons-io</groupId>
--- a/gluten-core/pom.xml
+++ b/gluten-core/pom.xml
@@ -173,13 +173,13 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
-      <version>${forked.protobuf.version}</version>
+      <version>${protobuf.version}</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java-util</artifactId>
-      <version>${forked.protobuf.version}</version>
+      <version>${protobuf.version}</version>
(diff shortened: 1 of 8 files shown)


gluten-core/pom.xml
```diff
@@ -173,12 +173,14 @@
     <dependency>
       <groupId>com.google.protobuf</groupId>
       <artifactId>protobuf-java</artifactId>
-      <version>${protobuf.version}</version>
+      <version>${forked.protobuf.version}</version>
+      <scope>provided</scope>
```
PHILO-HE, line 177: We tried to replace this with custom protobuf, but got error (class symbol cannot be found) when building generated protobuf source. Maybe, there is some unknown requirement. So we just keep the original dependency in this module with "provided" scope used. The custom protobuf packed into gluten jar will be used at runtime.

zhztheplayer:
Thanks for the fix @PHILO-HE. But had we tried using Java's reflection utilities to rewrite the recursion limit forcibly at runtime? I went through [code](https://github.com/oap-project/protobuf/commit/1597b3bc7c3c398e8c67aeddad5b8f76efa2f7f2#diff-0aa3b0d34acf2404473476e234080a2293dcdd26c8865d5f33454c1d1a246b46) then realized it might be a typical case that reflection is enough to cover.

PHILO-HE:
> Thanks for the fix @PHILO-HE. But had we tried using Java's reflection utilities to rewrite the recursion limit forcibly at runtime? I went through [code](https://github.com/oap-project/protobuf/commit/1597b3bc7c3c398e8c67aeddad5b8f76efa2f7f2#diff-0aa3b0d34acf2404473476e234080a2293dcdd26c8865d5f33454c1d1a246b46) then realized it might be a typical case that reflection is enough to cover.

Hi @zhztheplayer, thanks for your suggestion! It should also work by using reflection to change the class static member in class's first loading. The proposed custom dependency is straightforward, but it requires some efforts to complete the maven publish process. Maybe, we can try to use reflection when protobuf is required to upgrade or needs some other similar changes in the future. Does it make sense to you?

zhztheplayer (approved):
Thanks for the explanation. I would be OK to both. Just thought that using some reflections could make things simpler.

PR #3828, opened 2023-11-23
[VL] Support StreamingAggregate if child output ordering is satisfied

ulysses-you:

If the child output ordering satisfies the grouping keys then the grouping keys are pregrouped, we can choose to use `StreamingAggregate`. Velox supports `StreamingAggregate` which requires the input is pregrouped, so `StreamingAggregate` could use less memory to do aggregate as it does not need to hold all groups in memory and it could avoid spill.

For example, the query can hit this optimization if the join is SMJ:
```
SELECT c1, count(*), sum(c2) FROM (
SELECT t1.c1, t2.c2 FROM t1 t1 JOIN t1 t2 ON t1.c1 = t2.c1
)
GROUP BY c1;
```

## How was this patch tested?

manually benchmark

```
OpenJDK 64-Bit Server VM 1.8.0_292-b10 on Linux 5.10.0-21-amd64
Intel(R) Core(TM) i5-10500 CPU @ 3.10GHz
streaming aggregate:                      Best Time(ms)   Avg Time(ms)   Stdev(ms)    Rate(M/s)   Per Row(ns)   Relative
------------------------------------------------------------------------------------------------------------------------
Enable streaming aggregate                         8074           8101          24          1.0        1009.2       1.0X
Disable streaming aggregate                       12338          12352          20          0.6        1542.2       0.7X

```

### DIFF
--- a/backends-velox/src/test/scala/org/apache/spark/sql/execution/benchmark/StreamingAggregateBenchmark.scala
+++ b/backends-velox/src/test/scala/org/apache/spark/sql/execution/benchmark/StreamingAggregateBenchmark.scala
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.execution.benchmark
+
+import io.glutenproject.GlutenConfig
+
+import org.apache.spark.benchmark.Benchmark
+import org.apache.spark.sql.internal.SQLConf
+
+/**
+ * Benchmark to measure performance for streaming aggregate. To run this benchmark:
+ * {{{
+ *    bin/spark-submit --class <this class> --jars <spark core test jar> <sql core test jar>
+ * }}}
+ */
+object StreamingAggregateBenchmark extends SqlBasedBenchmark {
+  private val numRows = {
+    spark.sparkContext.conf.getLong("spark.gluten.benchmark.rows", 8 * 1000 * 1000)
+  }
+
+  private val mode = {
+    spark.sparkContext.conf.getLong("spark.gluten.benchmark.remainder", 4 * 1000 * 1000)
+  }
+
+  private def doBenchmark(): Unit = {
+    val benchmark = new Benchmark("streaming aggregate", numRows, output = output)
+
+    val query =
+      """
+        |SELECT c1, count(*), sum(c2) FROM (
+        |SELECT t1.c1, t2.c2 FROM t t1 JOIN t t2 ON t1.c1 = t2.c1
+        |)
+        |GROUP BY c1
+        |""".stripMargin
+    benchmark.addCase(s"Enable streaming aggregate", 3) {
+      _ =>
+        withSQLConf(
+          GlutenConfig.COLUMNAR_PREFER_STREAMING_AGGREGATE.key -> "true",
+          SQLConf.AUTO_BROADCASTJOIN_THRESHOLD.key -> "-1",
+          GlutenConfig.COLUMNAR_FPRCE_SHUFFLED_HASH_JOIN_ENABLED.key -> "false"
+        ) {
+          spark.sql(query).noop()
+        }
+    }
+
+    benchmark.addCase(s"Disable streaming aggregate", 3) {
+      _ =>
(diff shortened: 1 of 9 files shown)


backends-velox/src/test/scala/org/apache/spark/sql/execution/benchmark/StreamingAggregateBenchmark.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  override def runBenchmarkSuite(mainArgs: Array[String]): Unit = {
+    spark
+      .range(numRows)
+      .selectExpr(s"id % $mode as c1", "id as c2")
+      .write
+      .saveAsTable("t")
+
+    try {
+      doBenchmark()
```
zhztheplayer, line 82: Seems to me that we don't check query result in the benchmark code. So should we add some simple UT code with result check?

ulysses-you: yeah, I will add a test later

ulysses-you:
cc @zhztheplayer @rui-mo @PHILO-HE it's ready to review, thank you

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -73,6 +74,24 @@ abstract class HashAggregateExecBaseTransformer(
...
+      return false
+    }
+
+    val childOrdering = child match {
+      case agg: HashAggregateExecBaseTransformer
+          if agg.groupingExpressions == this.groupingExpressions =>
+        agg.child.outputOrdering
+      case _ => child.outputOrdering
+    }
+    val requiredOrdering = groupingExpressions.map(expr => SortOrder.apply(expr, Ascending))
```
rui-mo, line 91: Why ascending by default?

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -976,7 +976,9 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::AggregateRel& ag
   if (aggRel.has_advanced_extension()) {
     std::vector<TypePtr> types;
     const auto& extension = aggRel.advanced_extension();
-    if (!validateInputTypes(extension, types)) {
+    // Aggregate always has advanced extension for steaming aggregate optimization,
```
rui-mo, line 979: Nit: steaming -> streaming

rui-mo (commented):
Thanks.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -73,6 +74,24 @@ abstract class HashAggregateExecBaseTransformer(
...
+      return false
+    }
+    if (groupingExpressions.isEmpty) {
+      return false
+    }
+
+    val childOrdering = child match {
+      case agg: HashAggregateExecBaseTransformer
+          if agg.groupingExpressions == this.groupingExpressions =>
+        agg.child.outputOrdering
```
zhztheplayer, line 88: Why propagating child's child's ordering here? It would be great if we could also leave some comments in code. Thanks.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
rui-mo: Does that mean it only fits with SMJ?

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -73,6 +74,24 @@ abstract class HashAggregateExecBaseTransformer(
       aggregateAttributes)
   }
 
+  protected def isGroupingKeysPreGrouped: Boolean = {
+    if (!conf.getConf(GlutenConfig.COLUMNAR_PREFER_STREAMING_AGGREGATE)) {
+      return false
+    }
```
zhztheplayer, line 80: The check doesn't do as the method name `isGroupingKeysPreGrouped` said?

I would suggest renaming `isGroupingKeysPreGrouped` to something like `isCapableForStreamingAggregation`, or just moving this config check out.

zhztheplayer (commented):
Thanks! It's already in good shape in my opinion.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
ulysses-you: added some comments, hope it is clear now

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
zhztheplayer: Is it by Spark's limitation that a plan node can't require "asc or desc" child ordering at the same time?

ulysses-you: yes, it's a kind of hard code, but so far, I did not see any requirements about changing it. Will Velox operators require or introduce Descending ordering ?

zhztheplayer (approved):

PR #3836, opened 2023-11-24
[VL] Link simdjson static lib built out by Velox

PHILO-HE:

Remove Simdjson.cmake and just re-use this static lib built out by Velox to avoid redundant build.


cpp/velox/CMakeLists.txt
```diff
@@ -200,6 +200,7 @@ macro(ADD_VELOX_DEPENDENCIES)
   endif()
   add_velox_dependency(common::compression "${VELOX_COMPONENTS_PATH}/common/compression/libvelox_common_compression.a")
   add_velox_dependency(common::io "${VELOX_COMPONENTS_PATH}/common/io/libvelox_common_io.a")
+  add_velox_dependency(external::simdjson "${VELOX_BUILD_PATH}/_deps/simdjson-build/libsimdjson.a")
```
Yohahaha, line 203: what happens when velox use system simdjson? maybe we should use cmake way to link simdjson dep.

PHILO-HE: > what happens when velox use system simdjson? maybe we should use cmake way to link simdjson dep.

Hi @Yohahaha, thanks for your review! As we let velox build from simdjson source (because `simdjson_SOURCE=BUNDLED` is used, see [link](https://github.com/oap-project/gluten/blob/main/ep/build-velox/src/build_velox.sh#L127)), this static lib should always be found here. Right?

Yohahaha:
```
find_package(simdjson)
if (NOT simdjson_FOUND)
  include(Simdjson)
endif ()
```

this works in my env.

Yohahaha: > > what happens when velox use system simdjson? maybe we should use cmake way to link simdjson dep.
> 
> Hi @Yohahaha, thanks for your review! As we let velox always build from simdjson source (because `simdjson_SOURCE=BUNDLED` is used, see [link](https://github.com/oap-project/gluten/blob/main/ep/build-velox/src/build_velox.sh#L127)), this static lib should always be found here. Right?

yeah, I got it. we always build velox by myself, and not sure if others will install simdjson in their env.

PHILO-HE: Yes, no matter whether simdjson is installed in user env., we always link the one built out by velox.

Yohahaha: If user have installed simdjson and recompile velox, `_deps/simdjson-build` may not exists.

PHILO-HE: As simdjson_SOURCE=BUNDLED is set, Velox should always build simdjson from source instead of finding package in system. So I think `_deps/simdjson-build` should always exist. See velox doc [here]( https://github.com/facebookincubator/velox/blob/2efd4e5ac7483230a39ae91392b4896c4d71a033/CMake/ResolveDependency.cmake#L61).

Yohahaha: > As simdjson_SOURCE=BUNDLED is set, Velox should always build simdjson from source instead of finding package in system

just want clarify, it means user must use gluten provided build script or set simdjson_SOURCE=BUNDLED in their build process, right? 

and curious why gluten need to force use BUNDLED simdjson?

PHILO-HE: Yes, we expect user always uses the script provided by gluten to build velox. For example, you may note the provided script will apply a patch named `modify_velox.patch` before building velox, which is required by Gluten.

I'm not sure why BUNDED is set by someone. Maybe, it is for fixing some unknown issue.

Yohahaha: Hi @rui-mo could you give more information about why force use BUNDLED simdjson?

rui-mo: I remember it was to solve Gluten compilation issues caused by different library versions.

rui-mo (approved):

cpp/velox/CMakeLists.txt
PHILO-HE: @Yohahaha, I will create another PR to try to remove the force use of BUNDLED and let gluten find package in system path or _deps/ path. Thanks!

PR #3840, opened 2023-11-24
[VL] Fix QAT codec not closed

marin-ma:
Some of the inner structures of `qzSession_` are not released without calling `qzTeardownSession`.


zhouyuan (approved):

zhouyuan:
it seems there is another flaky unit tests, the failure is not related

PR #3842, opened 2023-11-24
[WIP][VL] fix RoundRobinPartitioner by setting seed pre partition

zhli1142015:

`spark.range(100).withColumn("pCol", 'id % 2).repartition(10).write.format("parquet").mode("overwrite").save("/tmp/veloxtest7")`

This query should generate 10 files under directory /tmp/veloxtest7, but in my local shell, only 6 files are genereated (16 cores, so 16 partition by default, and average 6 rows each partition).

RoundRobinPartitioner always start with 0 as selected partition id for all partitions this is not expected. Make this change to align with spark behavior: to set seed as `xshift(partitionId)`.

## How was this patch tested?

TBD.

### DIFF
--- a/cpp/core/shuffle/Partitioner.cc
+++ b/cpp/core/shuffle/Partitioner.cc
@@ -23,12 +23,15 @@
 
 namespace gluten {
 
-arrow::Result<std::shared_ptr<Partitioner>> Partitioner::make(Partitioning partitioning, int32_t numPartitions) {
+arrow::Result<std::shared_ptr<Partitioner>> Partitioner::make(
+    Partitioning partitioning,
+    int32_t numPartitions,
+    int32_t partitionKeySeed) {
   switch (partitioning) {
     case Partitioning::kHash:
       return std::make_shared<HashPartitioner>(numPartitions);
     case Partitioning::kRoundRobin:
-      return std::make_shared<RoundRobinPartitioner>(numPartitions);
+      return std::make_shared<RoundRobinPartitioner>(numPartitions, partitionKeySeed);
     case Partitioning::kSingle:
       return std::make_shared<SinglePartitioner>();
     case Partitioning::kRange:
--- a/gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornHashBasedColumnarShuffleWriter.scala
+++ b/gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornHashBasedColumnarShuffleWriter.scala
@@ -29,6 +29,7 @@ import org.apache.spark.scheduler.MapStatus
 import org.apache.spark.shuffle.celeborn.CelebornShuffleHandle
 import org.apache.spark.sql.vectorized.ColumnarBatch
 import org.apache.spark.util.SparkResourceUtil
+import org.apache.spark.util.random.XORShiftRandom
 
 import org.apache.celeborn.client.ShuffleClient
 import org.apache.celeborn.common.CelebornConf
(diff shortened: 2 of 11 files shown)


zhli1142015:
@PHILO-HE and @JkSelf , could you please review this PR?

JkSelf (commented):
LGTM.

JkSelf:
cc @marin-ma

gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornHashBasedColumnarShuffleWriter.scala
```diff
@@ -72,6 +73,12 @@ class VeloxCelebornHashBasedColumnarShuffleWriter[K, V](
       } else {
         val handle = ColumnarBatches.getNativeHandle(cb)
         if (nativeShuffleWriter == -1L) {
+          val startPartitionId = dep.nativePartitioning.getShortName match {
+            case "rr" =>
+              new XORShiftRandom(context.partitionId())
+                .nextInt(dep.partitioner.numPartitions)
+            case _ => 0
+          }
```
ulysses-you, line 81: @zhli1142015  thank you for the fix. Can we move this code block into NativePartitioning ? It should also be reused for other shuffle service.

cpp/core/shuffle/Partitioner.cc
```diff
@@ -23,12 +23,13 @@
...
 
-arrow::Result<std::shared_ptr<Partitioner>> Partitioner::make(Partitioning partitioning, int32_t numPartitions) {
+arrow::Result<std::shared_ptr<Partitioner>>
+Partitioner::make(Partitioning partitioning, int32_t numPartitions, int32_t startPartitionId) {
   switch (partitioning) {
     case Partitioning::kHash:
       return std::make_shared<HashPartitioner>(numPartitions);
     case Partitioning::kRoundRobin:
-      return std::make_shared<RoundRobinPartitioner>(numPartitions);
+      return std::make_shared<RoundRobinPartitioner>(numPartitions, startPartitionId);
```
PHILO-HE, line 32: Thanks so much for your fix!
As just required by `RoundRobinPartitioner`, can we move the code of generating `startPartitionId` from scala side to native side, i.e., into `RoundRobinPartitioner`? It may be a cleaner way if feasible. Thanks!

zhli1142015: Thanks for your comments. Yes, I'm firstly also looking to implement this in `RoundRobinPartitioner`, but looks there is no native implement of `XORShiftRandom`. Please point me if there is.
Besides if we move this logic to native side we still need to add a parameter to pass partition id from JVM to native.
Now the native side implment is more like the `HashPartitioner`, and `getPartitionKeyExtractor` are executed in JVM.

marin-ma: Yes. There is no native `XORShiftRandom` in Gluten cpp. Since it's just a simple computation of a starting value, we don't need to bother maintaining the same logic on native side. However, the current argument list does look cumbersome. To make it cleaner, we should find a better way to separate parameters for the native `ShuffleWriter` and `Partitioner`, perhaps in a subsequent patch.

PHILO-HE: Thanks for the clarifications. The current change is OK to me.
BTW, I am wondering whether `XORShiftRandom` must be used or can use a replacement (I'm not suggesting to change it).

zhli1142015: This impacts the partition that rows are placed, I think it's better to use the same algorithm with spark. And it may be required if "SORT_BEFORE_REPARTITION" is true. But there looks another issue, as now we just ignore it, we may add local sort before round robin shuffle in future if `SORT_BEFORE_REPARTITION == true`.
https://github.com/apache/spark/blob/e1a2255f99be88e776295f30f995b339c3e4b5af/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L3367

zhouyuan (approved):
👍

gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornHashBasedColumnarShuffleWriter.scala
zhli1142015: Moved this to shuffleUtils, thanks.

ulysses-you (approved):

Issue #3378, opened 2023-10-11
Unified design for data lake read support in Gluten + Velox


yma11:
### Description

Currently there are 2 PRs opened in Gluten to support [Iceberg COW table read](https://github.com/oap-project/gluten/pull/3043) and [Delta Lake read](https://github.com/oap-project/gluten/pull/2902). There is also one hot discussion in Velox about [Iceberg read](https://github.com/facebookincubator/velox/issues/5977) support. By consolidating the ideas and based on Gluten's position, we would like to share a draft unified design for data lake read support in Gluten.

As addressed in this project's home page, one of Gluten key function is to transform Spark’s whole stage physical plan to Substrait plan and send to native. It applies to data lake read support, thus:

1. We'd better avoid hacking of original Spark physical plan node. Gluten core has plan transformer to generate correct plan info into substrait format and then pass it to Velox for read and computation. So no matter what kind of the hack is, it should can be done in the transformer layer, such as column mapping. IMO, we should try best pass original info in spark plan to Velox as a bridge and do correct consumption at Velox side, unless it's not doable or velox can't support. By the way, one issue for feature like column mapping is, it's a common feature for kinds of file format reading, velox can handle this at its datasource level and the community has plan to do so. 

3. Clear transformer hierarchy is need for different data lake backends. In the [Iceberg COW table read](https://github.com/oap-project/gluten/pull/3043) PR, a new branch is added to do specific process for Iceberg and leverage an utility class put in a dedicated folder, and in future, I believe more branches will be needed to support other cases, like MOR. So introducing a new transformer inherited from `BatchScanExecTransformer` would be a better way. The possible hierarchy should be like following:

```
                                IcebergDataSoure?                                     DeltaLakeDatasoure?                        ?
                                                    \                                     |                                 / 
                                                            \                             |                           /
                                                                   \                      |                      /
                                                                                       velox
                                                                                          |
                                                                                     substrait
                                                               /                          |                    \
                                                  /                                       |                             \
        SparkBatchQueryScanExecTransformer                           DeltaLakeScanExecTransformer                                 \
                                  |                                                       |
                  BatchScanExecTransformer                              FileSourceScanExecTransformer         HiveTableScanExecTransformer
                                    \                                                     |                                    /
                                                                         BasicScanExecTransformer
```
@YannByron @felipepessoto @liujiayi771 @ulysses-you, please give comments on above suggestions. Thanks.

liujiayi771:
[...]

PR #3843, opened 2023-11-24
[GLUTEN-3378][CORE] Datasource V2 data lake read support

liujiayi771:

* Implement datasource v2 data lake read based on #3378 .
* Introduce gluten-iceberg module, `IcebergScanTransformer` is defined that extends `BatchScanExecTransformer`. Although the `IcebergLocalFilesNode` is currently the same as the `LocalFilesNode`, it is being prepared for future implementation of the "delete file" functionality.
* The `ScanTransformerFactory` is used to construct various types of `ScanTransformer`. For Iceberg, the `IcebergScanTransformer` will be constructed using service loader in order to avoid the dependency of `gluten-core` on `gluten-iceberg`.
* The logic that pushes down the Filter conditions to the BatchScan runtimeFilters has been removed from the original code. This logic was unnecessary because `BatchScan` only supports DPP's runtime filter for partition filtering.
* The logic for executing subqueries has been placed in the `transformDynamicPruningExpr` method. This allows both `BatchScan` and `FileSourceScan` to share this logic, as it is also required by `BatchScan`.

## How was this patch tested?

* Add `VeloxTPCHIcebergSuite` and `VeloxIcebergSuite`. 
* The modifications to the interfaces can be tested using the existing CI.


liujiayi771:
@YannByron @yma11 @rui-mo Could you help review?

docs/get-started/Velox.md
```diff
@@ -260,6 +260,23 @@ After the two steps, you can query delta table by gluten/velox without scan's fa
 Gluten with velox backends also support the column mapping of delta tables.
 About column mapping, see more [here](https://docs.delta.io/latest/delta-column-mapping.html).
 
+## Iceberg Support
+
+Gluten with velox backend supports [Iceberg](https://iceberg.apache.org/) table. Currently, only reading COW (Copy-On-Write) tables is supported.
+
+### How to use
+
+First of all, compile gluten-delta module by a `delta` profile, as follows:
```
rz-vastdata, line 269:
```suggestion
First of all, compile gluten-delta module by a `iceberg` profile, as follows:
```

gluten-core/src/main/scala/io/glutenproject/execution/ScanTransformerFactory.scala
```diff
@@ -0,0 +1,115 @@
...
+      reuseSubquery: Boolean,
+      validation: Boolean = false): BatchScanTransformer = {
+    val newPartitionFilters = if (validation) {
+      batchScanExec.runtimeFilters
+    } else {
+      ExpressionConverter.transformDynamicPruningExpr(batchScanExec.runtimeFilters, reuseSubquery)
+    }
+    val scan = batchScanExec.scan
+    scan match {
+      case _ if scan.getClass.getName == IcebergScanClassName =>
```
rz-vastdata, line 90: IIUC, supporting a new `Scan` type (e.g. adding a non-Iceberg data source) will require modifying this `match` code, as well as the `supportedBatchScan()` method below.
Would it please be possible to allow other plugins to register themselves, so that adding a new format will not require changing `ScanTransformerFactory` code?

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+package io.glutenproject.substrait.rel;
+
+import java.util.List;
+import java.util.Map;
+
+public class IcebergLocalFilesNode extends LocalFilesNode {
+
+  class DeleteFile {
```
rz-vastdata, line 24: Not sure if it's used by this PR...

gluten-core/src/main/scala/io/glutenproject/execution/ScanTransformerFactory.scala
liujiayi771: @rz-vastdata We can achieve this using the Service Loader, similar to `DataSourceRegister` in Spark. However, the current `ScanTransformer` in Gluten extends Spark's `ScanExec` and requires a constructor with parameters. This makes it a bit difficult to handle with Service Loader. I will think about whether there is another way to use the Service Loader.

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
liujiayi771: I didn't use it. This is a TODO.

YannByron:
My fault in design. I noticed that there are many modification that rename `XXXExecTransformer` to `XXXTransformer`. Maybe we can use `XXXExecTransformer` as these class names directly to avoid it.

yma11:
> My fault in design. I noticed that there are many modification that rename `XXXExecTransformer` to `XXXTransformer`. Maybe we can use `XXXExecTransformer` as these class names directly to avoid it.

+1

gluten-core/src/main/scala/io/glutenproject/execution/DatasourceScanTransformer.scala
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.execution
+
+import org.apache.spark.sql.catalyst.TableIdentifier
+import org.apache.spark.sql.sources.BaseRelation
+
+trait DatasourceScanTransformer extends BaseScanTransformer {
```
yma11, line 22: Is this added to just keep consistent with vanilla Spark? If so, I think we can deduct this hierarchy to make the inheritance not so complicated.

gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
```diff
@@ -52,7 +52,7 @@ class HiveTableScanExecTransformer(
     relation: HiveTableRelation,
     partitionPruningPred: Seq[Expression])(session: SparkSession)
   extends HiveTableScanExec(requestedAttributes, relation, partitionPruningPred)(session)
-  with BasicScanExecTransformer {
+  with BaseScanTransformer {
```
yma11, line 55: Better to still use `BasicScanExecTransformer` as this change is not necessary.

gluten-core/src/main/scala/io/glutenproject/execution/BaseDataSource.scala
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+package io.glutenproject.execution
+
+import io.glutenproject.substrait.SupportFormat
+
+import org.apache.spark.sql.connector.read.InputPartition
+import org.apache.spark.sql.types.StructType
+
+trait BaseDataSource extends SupportFormat {
```
yma11, line 24: Seems `BasicScanTransformer` is the only implementation of `SupportFormat`, maybe you can add the fields directly?

gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
liujiayi771: OK. I will revert these changes first.

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
```diff
@@ -0,0 +1,63 @@
...
+      this.fileSize = fileSize;
+      this.recordCount = recordCount;
+      this.lowerBounds = lowerBounds;
+      this.upperBounds = upperBounds;
+    }
+  }
+
+  // TODO: Add delete file support for MOR iceberg table
+
+  IcebergLocalFilesNode(
```
yma11, line 53: Will we also need to use service loader for this LocalFilesNode serialization?

gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
```diff
@@ -70,11 +70,11 @@ class HiveTableScanExecTransformer(
 
   override def getPartitions: Seq[InputPartition] = partitions
 
-  override def getPartitionSchemas: StructType = relation.tableMeta.partitionSchema
```
liujiayi771, line 73: @yma11 I think the names of these interfaces need to be changed. Here, you can refer to the corresponding interfaces in Spark, and we should not use plurals.

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
liujiayi771: No. It is not used by `gluten-core`. The incremental part of the delta and iceberg definitions of the MOR table is expected to have significant differences.

gluten-core/src/main/scala/io/glutenproject/execution/BaseDataSource.scala
liujiayi771: Add this interface to `BaseDataSource`?

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
yma11: There should be a `toProtobuf()` method for `IcebergLocalFilesNode` and it will be called in gluten-core/substrait/../PlanNode for serialization. Or how do you plan to pass these new added fields?

gluten-core/src/main/scala/io/glutenproject/execution/BaseDataSource.scala
yma11: I mean `SupportFormat`.

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
liujiayi771: The current PR does not involve deleting files. To support deleting files in the future, we will need to implement a specific `toProtobuf` method and modify the proto file.

liujiayi771:
@yma11 @YannByron @rz-vastdata I have modified the reflection to use the service loader. Can you please help review?

gluten-core/src/main/scala/io/glutenproject/execution/DataSourceV2TransformerRegister.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.execution.datasources.v2.BatchScanExec
+
+/**
+ * Data sources v2 transformer should implement this trait so that they can register an alias to
+ * their data source v2 transformer. This allows users to give the data source v2 transformer alias
+ * as the format type over the fully qualified class name.
+ */
+trait DataSourceV2TransformerRegister {
```
YannByron, line 27: `DataSourceTransformerRegister` ?

gluten-core/src/main/scala/io/glutenproject/execution/ScanTransformerFactory.scala
```diff
@@ -0,0 +1,135 @@
...
+import java.util.ServiceLoader
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.collection.JavaConverters._
+
+object ScanTransformerFactory {
+
+  private val dataSourceV2TransformerMap = new ConcurrentHashMap[String, Class[_]]()
+
+  def createFileSourceScanTransformer(
```
YannByron, line 36: can we combine the two `createFileSourceScanTransformer` methods ?

liujiayi771: Of course, initially they were combined together, but we need to add some optional "Option" parameters.

gluten-core/src/main/scala/io/glutenproject/execution/DataSourceV2TransformerRegister.scala
liujiayi771: Yes. DataSource V1 has differrent interface parameters.

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -469,52 +468,31 @@ object FilterHandler {
...
-          fileSourceScan.dataFilters ++ leftFilters,
-          fileSourceScan.tableIdentifier,
-          fileSourceScan.disableBucketedScan
-        )
+          getLeftFilters(fileSourceScan.dataFilters, flattenCondition(filter.condition))
+        ScanTransformerFactory.createFileSourceScanTransformer(
+          fileSourceScan,
+          reuseSubquery,
+          extraFilters = leftFilters)
       case batchScan: BatchScanExec =>
```
rui-mo, line 480: Thanks for your fix. Can be avoid the extra filter pushdown for BatchScan by not calling `applyFilterPushdownToScan` for it?

https://github.com/oap-project/gluten/blob/c531abd94045db71a8f8ef692e5c5a80cbcd118f/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala#L140-L145

liujiayi771: Fix this in https://github.com/oap-project/gluten/commit/d879fd097dada9aba9c21046b01b1104f916188a.

yma11 (approved):
Hi @liujiayi771, thanks for your update. Can you do a rebase for a round CI? I will merge it once it's done.

PR #3854, opened 2023-11-27
[CORE] Add InputIteratorTransformer to decouple ReadRel and iterator index

ulysses-you:

This pr adds a new transformer `InputIteratorTransformer` to connect SparkPlan and TransformSupport. It replace the original `ColumnarInputAdapter` to provide the substrait plan ReadRel for the child columnar iterator, so that the TransformSupport always has input. It would be transformed to ValueStreamNode at native Velox side.

The reason is that:
1. decopule with scan transofrmer, the scan transofrmer does not need iterator index
2. make the whole stage transform framework more clear, all the TransformSupport does not need to check if the child is a TransformSupport and build a ReadRel by theirself
3. make the metrics framework more clear, the metrics should belong to the input iterator rather than the operators
4. it aligns with native side, e.g., velox backend will transform it to ValueStreamNode

## How was this patch tested?

Pass CI and manually test

image

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/MetricsApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/MetricsApiImpl.scala
@@ -41,7 +41,7 @@ class MetricsApiImpl extends MetricsApi with Logging {
       sparkContext: SparkContext): Map[String, SQLMetric] = {
     Map(
       "cpuCount" -> SQLMetrics.createMetric(sparkContext, "cpu wall time count"),
-      "wallNanos" -> SQLMetrics.createNanoTimingMetric(sparkContext, "totaltime of batch scan"),
+      "wallNanos" -> SQLMetrics.createNanoTimingMetric(sparkContext, "totaltime of input iterator"),
       "outputVectors" -> SQLMetrics.createMetric(sparkContext, "number of output vectors")
     )
   }
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/MetricsApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/MetricsApiImpl.scala
@@ -37,6 +37,20 @@ class MetricsApiImpl extends MetricsApi with Logging {
     MetricsUtil.updateNativeMetrics(child, relMap, joinParamsMap, aggParamsMap)
   }
 
+  override def genInputIteratorTransformerMetrics(
+      sparkContext: SparkContext): Map[String, SQLMetric] = {
+    Map(
+      "cpuCount" -> SQLMetrics.createMetric(sparkContext, "cpu wall time count"),
+      "wallNanos" -> SQLMetrics.createNanoTimingMetric(sparkContext, "totaltime of batch scan"),
+      "outputVectors" -> SQLMetrics.createMetric(sparkContext, "number of output vectors")
+    )
+  }
+
+  override def genInputIteratorTransformerMetricsUpdater(
+      metrics: Map[String, SQLMetric]): MetricsUpdater = {
+    InputIteratorMetricsUpdater(metrics)
+  }
+
   override def genBatchScanTransformerMetrics(sparkContext: SparkContext): Map[String, SQLMetric] =
     Map(
       "inputRows" -> SQLMetrics.createMetric(sparkContext, "number of input rows"),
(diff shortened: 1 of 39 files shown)


backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/MetricsApiImpl.scala
```diff
@@ -37,6 +37,20 @@ class MetricsApiImpl extends MetricsApi with Logging {
...
   }
 
+  override def genInputIteratorTransformerMetrics(
+      sparkContext: SparkContext): Map[String, SQLMetric] = {
+    Map(
+      "cpuCount" -> SQLMetrics.createMetric(sparkContext, "cpu wall time count"),
+      "wallNanos" -> SQLMetrics.createNanoTimingMetric(sparkContext, "totaltime of batch scan"),
+      "outputVectors" -> SQLMetrics.createMetric(sparkContext, "number of output vectors")
+    )
+  }
```
Yohahaha, line 47: could you also add `inputVectors` ? seems shuffle read does not show that.

ulysses-you: Does not ValueStream inputVectors always same with outputVectors ?

Yohahaha: you are right.

ulysses-you:
Run Gluten Clickhouse CI

ulysses-you:
cc @Yohahaha @zhztheplayer @zhouyuan @zzcclp thank you

zhztheplayer:
cc @rui-mo

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/MetricsApiImpl.scala
```diff
@@ -37,6 +37,20 @@ class MetricsApiImpl extends MetricsApi with Logging {
     MetricsUtil.updateNativeMetrics(child, relMap, joinParamsMap, aggParamsMap)
   }
 
+  override def genInputIteratorTransformerMetrics(
+      sparkContext: SparkContext): Map[String, SQLMetric] = {
+    Map(
+      "cpuCount" -> SQLMetrics.createMetric(sparkContext, "cpu wall time count"),
+      "wallNanos" -> SQLMetrics.createNanoTimingMetric(sparkContext, "totaltime of batch scan"),
```
zhztheplayer, line 44: Is it accurate to use "totaltime of batch scan"? I thought the transformer is common for both scan and other inputs.

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
```diff
@@ -19,13 +19,58 @@ package org.apache.spark.sql.execution
...
 import java.util.concurrent.atomic.AtomicInteger
 
+import scala.collection.JavaConverters._
+
+/**
+ * A bridge to connect [[SparkPlan]] and [[TransformSupport]] and provide the substrait plan
+ * `ReadRel` for the child columnar iterator, so that the [[TransformSupport]] always has input. It
+ * would be transformed to `ValueStreamNode` at native side.
+ */
+case class InputIteratorTransformer(child: SparkPlan) extends UnaryTransformSupport {
```
zhztheplayer, line 43: Should we combine `InputIteratorTransformer` and `ColumnarInputAdapter` later on?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/MetricsApiImpl.scala
ulysses-you: good catch, I think it is copied from a wrong code path..

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
ulysses-you: It seems we do not need `ColumnarInputAdapter` any more

Yohahaha:
LGTM! thank you for this great patch.

zzcclp:
Good patch. Sorry for the late review, if there is any issue for ch backend, I will fix later.

PR #3872, opened 2023-11-29
[VL] Fix assertion failures when mixing use of partial aggregation spilling and flushing 

zhztheplayer:
Depends on https://github.com/oap-project/velox/pull/453


JkSelf (approved):
LGTM. Thanks.

PR #3873, opened 2023-11-29
[VL] Fix incorrect url path caused by decoding on non-original encoded path in PartitionedFile

yma11:

When constructing Spark `PartitionedFile`, the path used is the not the original encoded path so the decoded path in `IteratorApiImpl` is not correct in some cases. Current we observed the case for '+'.  It's encoded to `%2B` in `URLEncoder.encode` but it becomes `+` when passed into `PartitionedFile` and finally becomes ` ` after decode. This PR tries to fix this issue with UT added.

## How was this patch tested?

Test in customer S3 environment.

### DIFF
--- a/backends-velox/src/main/java/io/glutenproject/utils/GlutenURLDecoder.java
+++ b/backends-velox/src/main/java/io/glutenproject/utils/GlutenURLDecoder.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.utils;
+
+import java.io.*;
+import java.net.URLEncoder;
+
+public class GlutenURLDecoder {
+  // This Util is copied from
+  // https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/net/URLDecoder.java
+  // Just remove the special handle for '+'
+  /**
+   * Decodes a <code>application/x-www-form-urlencoded</code> string using a specific encoding
+   * scheme. The supplied encoding is used to determine what characters are represented by any
+   * consecutive sequences of the form "<code>%<i>xy</i></code>".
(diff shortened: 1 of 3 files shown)


cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -452,8 +452,13 @@ WholeStageResultIteratorFirstStage::WholeStageResultIteratorFirstStage(
       auto partitionColumn = partitionColumns[idx];
       std::unordered_map<std::string, std::optional<std::string>> partitionKeys;
       constructPartitionColumns(partitionKeys, partitionColumn);
+#ifdef ENABLE_S3
+      auto path = UrlDecode(paths[idx]);
```
PHILO-HE, line 456: Maybe, we can call a folly function to do the decoding.
See links:
https://github.com/facebook/folly/blob/775967aafc9c68af3d8f38619eac1cede43f0be7/folly/test/StringTest.cpp#L226C68-L226C68
https://github.com/facebook/folly/blob/775967aafc9c68af3d8f38619eac1cede43f0be7/folly/String-inl.h#L247

zhouyuan: @yma11 please check if this is necessary

zhouyuan: @yma11 could this patch also apply to gcs/abfs case?

FelixYBW (approved):

cpp/velox/compute/WholeStageResultIterator.cc
zhli1142015: I think we also need this, we found this case caused failure for both localfs and abfs, when partition key contains '+'.

zhli1142015:
This is our repro test case:
```

val c = "?.+<_>|/"
val path = "XXXXX"

val key1 = s"${c}key1${c}${c}"
val key2 = s"${c}key2${c}${c}"

val valueA = s"${c}some${c}${c}value${c}A"
val valueB = s"${c}some${c}${c}value${c}B"
val valueC = s"${c}some${c}${c}value${c}C"
val valueD = s"${c}some${c}${c}value${c}D"

val df1 = spark.range(3).withColumn(key1, lit(valueA)).withColumn(key2, lit(valueB))
val df2 = spark.range(4, 7).withColumn(key1, lit(valueC)).withColumn(key2, lit(valueD))
val df = df1.union(df2)
df.write.partitionBy(key1, key2).format("parquet").mode("overwrite").save(path)

spark.read.format("parquet").load(path).where(s"`$key1` = '${c}some${c}value${c}A'").select("id").collect

```

cpp/velox/compute/WholeStageResultIterator.cc
zhli1142015: We already do url decode in https://github.com/oap-project/gluten/blob/6d7d01fe0a433b82d3f216c992d244eb4686949a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala#L90
looks we will do it again, can we merge it to one place?

yma11:
Recover from deletion by mistake.

zhli1142015 (approved):
LGTM, thanks.

zhli1142015 (approved):

PR #3895, opened 2023-11-30
[VL] Break up spilling operation to two phases: shrink phase and spill phase

zhztheplayer:
This will make sure the memory manager shrinks all unused reservations before calling a spiller's `spill()` method. `shrink()` must be called before `spill()` since the latter costs computation and IO resources much higher than the former.


zhztheplayer:
@JkSelf @zhouyuan

JkSelf (approved):
LGTM. Thanks.

PR #3897, opened 2023-11-30
[GLUTEN-3580][VL] support read data from abfs with account key

gaoyangxiaozhu:

Fixes: https://github.com/oap-project/gluten/issues/3580

## How was this patch tested?

test maunally end to end


zhouyuan:
@gaoyangxiaozhu thank you for adding ABFS support! Can you please help to fix the dependency installation also? something like gcs I think
https://github.com/oap-project/gluten/blob/main/ep/build-velox/src/get_velox.sh#L99

thanks, -yuan

zhouyuan (approved):
👍

zhouyuan:
Note: need follow up on
- adding missing docs for abfs w/ gluten
- enable the abfs build by default in GHA

Issue #3906, opened 2023-12-01
[VL] dev/package.sh build failed for x86 cause by CPU_ARCH env not correct


lzjqsdd:
### Problem description

package.sh export CPU_ARCH env from 'uname -m' ,  this will get `x86_64` for intel platform。 velox need CPU_ARCH for aarch64/aarch64/avx/sse,  this will cause velox build fmt dependency failed 。 Velox will automatically check CPU_ARCH environment if not specified

### System information

Velox System Info v0.0.2
Commit: d2980b73afbf28800727172200e36a84c4c83536
CMake Version: 3.16.3
System: Linux-5.15.0-84-generic
Arch: x86_64
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 9.4.0
C Compiler: /usr/bin/cc
C Compiler Version: 9.4.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

PR #3907, opened 2023-12-01
[GLUTEN-3906] [VL] fix: fix package.sh failed for x86

lzjqsdd:
package.sh export CPU_ARCH env from 'uname -m' ,  this will get `x86_64` for intel platform。 velox need CPU_ARCH for aarch64/aarch64/avx/sse,  this will cause velox build fmt dependency failed 。 Velox will automatically check CPU_ARCH environment if not specified

Fix: #3906


zhouyuan (approved):
👍  thank you

PR #3916, opened 2023-12-04
[VL] Avoid reapplying code patches for external projects when ENABLE_EP_CACHE=ON

zhztheplayer:
This is to fix error "Failed to apply compilation fixes to Velox: ..." when running build scripts with `ENABLE_EP_CACHE=ON`.

Also, remove option SKIP_BUILD_EP since its functionality overlaps ENABLE_EP_CACHE. One can use `ENABLE_EP_CACHE=ON` to skip building external projects which were already built earlier.

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -21,7 +21,6 @@ ENABLE_GCS=OFF
 ENABLE_S3=OFF
 ENABLE_HDFS=OFF
 ENABLE_EP_CACHE=OFF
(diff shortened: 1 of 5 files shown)


zhouyuan:
will this also check the compile flags?
e.g. if velox commit does not change, but the compile flags are modified, `--enable_hdfs=ON/OFF`

zhztheplayer:
@zhouyuan @PHILO-HE @rui-mo

PHILO-HE (approved):
Looks good! Thanks!

PR #3926, opened 2023-12-05
[VL] ENABLE_EP_CACHE=ON still uses cached Velox build although the build arguments were changed

zhztheplayer:
Prior to this fix, ENABLE_EP_CACHE=ON only uses the target commit hash as cache key, which means for example, when we change BUILD_TYPE from `Debug` to `Release`, the cached debug build is still used as cached build incorrectly.

This patch is a fix for the bug.

### DIFF
--- a/ep/build-velox/src/build_velox.sh
+++ b/ep/build-velox/src/build_velox.sh
@@ -143,7 +143,7 @@ function compile {
   fi
 }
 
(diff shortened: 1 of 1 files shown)


zhztheplayer:
@zhouyuan

zhouyuan (approved):
👍

PR #3932, opened 2023-12-05
[VL] Decouple ShuffleWriter and PartitionWriter

marin-ma:
In the previous implementation, The life cycle of `PartitionWriter` is managed by `ShuffleWriter`. However, `PartitionWriter` also holds the pointer to `ShuffleWriter` and manipulates its internal state, which is considered as bad design. This PR removes the raw pointer of `ShuffleWriter` in `PartitionWrtier`.

Detailed changes:
1. In the previous implementation, during `ShuffleWriter::Stop`, it will call `PartitionWriter::Stop` to merge spilled files and flush cached payloads, and calls back `ShuffleWriter` to write in-memory split buffers. After this change, `ShuffleWrtier` will transfer the ownership of split buffers to `PartitionWriter` at the beginning of `ShuffleWriter::Stop`, so the callback can be removed.
2. `PartitionWriter` will update `ShuffleWriter` metrics. This PR adds `ShuffleWriterMetrics` and `PartitionWriter::populateMetrics` to update metrics data only with `ShuffleWriterMetrics` instance.

With above changes, the raw pointer of `ShuffleWriter` can be removed in `PartitionWriter`.


FelixYBW:
ready to review?

marin-ma:
Resolved by #4099

PR #3933, opened 2023-12-05
[GLUTEN-3854][CORE][FOLLOWUP] Add ColumnarInputAdapter back to recover UI graph

ulysses-you:

This pr is followup fix of https://github.com/oap-project/gluten/pull/3854. We can not remove `ColumnarInputAdapter`, otherwise the UI graph would be broken.

## How was this patch tested?

before:
image

after:
image

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
@@ -24,10 +24,12 @@ import io.glutenproject.substrait.SubstraitContext
 import io.glutenproject.substrait.rel.RelBuilder
 
 import org.apache.spark.broadcast.Broadcast
+import org.apache.spark.rdd.RDD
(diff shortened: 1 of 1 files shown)


ulysses-you:
cc @zhztheplayer thank you

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
```diff
@@ -66,6 +71,48 @@ case class InputIteratorTransformer(child: SparkPlan) extends UnaryTransformSupp
...
+      depth,
+      lastChildren,
+      append,
+      verbose,
+      prefix = "",
+      addSuffix = false,
+      maxFields,
+      printNodeId,
+      indent)
+  }
```
zhztheplayer, line 113: It seems that we only need to override `#supportCodegen` ? Can we remove others to simply code? Thanks!

ulysses-you: sure, addressed

zhztheplayer (approved):
Thanks for the fix!

PR #3937, opened 2023-12-05
[CORE][GLUTEN-3936] Support collapse project transformer

liujiayi771:

Fix https://github.com/oap-project/gluten/issues/3936

## How was this patch tested?

Add CollapseProjectExecTransformerSuite


liujiayi771:
@ulysses-you Could you help review?

gluten-core/src/main/scala/io/glutenproject/extension/CollapseProjectTransformer.scala
```diff
@@ -0,0 +1,42 @@
...
+      return plan
+    }
+    plan.transformUp {
+      case p1 @ ProjectExecTransformer(_, p2: ProjectExecTransformer)
+          if CollapseProjectShim.canCollapseExpressions(
+            p1.projectList,
+            p2.projectList,
+            alwaysInline = false) =>
+        p2.copy(projectList =
+          CollapseProjectShim.buildCleanedProjectList(p1.projectList, p2.projectList))
```
ulysses-you, line 39: To be conservative, better to check if the collapsed project can be transformed, and fallback to original plan if fail to validate. Also, do not forget to copy tags.

backends-velox/src/test/scala/io/glutenproject/extension/CollapseProjectExecTransformerSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+        |    WHERE
+        |      l_orderkey = o_orderkey
+        |      AND l_commitdate < l_receiptdate
+        |  )
+        |ORDER BY
+        | o_orderpriority
+        |LIMIT
+        | 100;
+        |""".stripMargin
+    withSQLConf(GlutenConfig.ENABLE_COLUMNAR_PROJECT_COLLAPSE.key -> "false") {
```
ulysses-you, line 72: how about use a loop to combine two test cases `Seq(true, false).foreach { collapsed => }`

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -1324,6 +1326,13 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val ENABLE_COLUMNAR_PROJECT_COLLAPSE =
+    buildConf("spark.gluten.sql.columnar.project.collapse")
+      .internal()
+      .doc("Combines two columnar project operators into one and perform alias substitution")
+      .booleanConf
+      .createWithDefault(false)
```
ulysses-you, line 1334: We should enable it by default and pass CI to see if there are some bad case.

backends-velox/src/test/scala/io/glutenproject/extension/CollapseProjectExecTransformerSuite.scala
ulysses-you, line 1: better to put it in spark-ut module as it affects both velox and ch backend.

ulysses-you:
thank you @liujiayi771 , the idea looks good to me

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
liujiayi771: Originally it was enabled by default, but the unit test for ck failed because some tests in ck directly used the index to retrieve the plan. After my optimization, the number of plans decreased, causing an IndexOutOfBoundsException in the unit test for ck. If it's just for validation, it can be enabled by default and then disabled later.
```
00:39:58  16:39:58.131 WARN io.glutenproject.metrics.MetricsUtil: Updating native metrics failed due to the wrong size of metrics data: 7
00:39:58  - Check TPCH Q2 metrics updater *** FAILED ***
00:39:58    java.lang.IndexOutOfBoundsException: 10
00:39:58    at scala.collection.mutable.ResizableArray.apply(ResizableArray.scala:46)
00:39:58    at scala.collection.mutable.ResizableArray.apply$(ResizableArray.scala:45)
00:39:58    at scala.collection.mutable.ArrayBuffer.apply(ArrayBuffer.scala:49)
00:39:58    at io.glutenproject.execution.metrics.GlutenClickHouseTPCHMetricsSuite.$anonfun$new$9(GlutenClickHouseTPCHMetricsSuite.scala:210)
00:39:58    at io.glutenproject.execution.metrics.GlutenClickHouseMetricsUTUtils$.executeMetricsUpdater(GlutenClickHouseMetricsUTUtils.scala:117)
00:39:58    at io.glutenproject.execution.metrics.GlutenClickHouseTPCHMetricsSuite.$anonfun$new$7(GlutenClickHouseTPCHMetricsSuite.scala:205)
00:39:58    at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
00:39:58    at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
00:39:58    at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
00:39:58    at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
```

ulysses-you: We can just fix it, e.g., change 10 to 9 if we have collapsed two projects into one. I think the code change should be test only.

gluten-core/src/main/scala/io/glutenproject/extension/CollapseProjectTransformer.scala
liujiayi771: > do not forget to copy tags.

Do you mean that after executing `val newProject = p2.copy`, I still need to call `newProject.copyTagsFrom(p1)`?

ulysses-you: it seems there is no way to show the fallback reason for a transform node.. so it's ok to skip copying tags.

liujiayi771:
@ulysses-you Ready for review. CK's UT needs to be adapted when the optimization is enabled by default, but this optimization will not be enabled by default later on. Unlike Velox, CK does not insert a project to calculate hash value before shuffle, so there are very few cases that meet the criteria for project collapsing. Currently, all UTs can pass. After completing the review, I will turn off the parameter by default and roll back the modifications made to CK's UT.

ulysses-you:
I think enabled it by default is fine even if the clickhouse backend has less case can be optimized.

ulysses-you (approved):
lgtm, cc @zzcclp if you have other comments

PR #3945, opened 2023-12-06
[VL] Fix syntax error in stale.yml

zhztheplayer:
This is a follow-up to https://github.com/oap-project/gluten/pull/3888


zhztheplayer:
@zhouyuan

Issue #3944, opened 2023-12-06
[CH] gluten.jar contains delta-20 adapter classes, it failed when use spark 33 to create mergetree table


lwz9103:
### Problem description

RT

### System information

RT

PR #3947, opened 2023-12-06
[GLUTEN-3944][CH]Fix gluten.jar with delta20 when use spark 3.3

lwz9103:

(Fixes: \#3944)

## How was this patch tested?

manual tests


baibaichen (approved):
LGTM

PR #3964, opened 2023-12-07
[VL] gluten-te: Enable BuildKit to support --cache-from

zhztheplayer:
The patch is a preparation of further CI docker script enhancements and should not impact current build procedure.

Ref of Docker external cache:

https://docs.docker.com/engine/reference/commandline/build/#cache-from

With the patch, we can perform the following in dedicated image-building job:

```
DOCKER_PUSH_REGISTRY=<registry> build.sh
```

Then we can use the cache to speed up build from another machine that have access to the registry:

```
DOCKER_CACHE_REGISTRY=<registry> build.sh
```


zhouyuan (approved):
👍

PR #3967, opened 2023-12-07
[CORE] Fix wrong fallback cost

ulysses-you:

Since we have removed AQE columnar shuffle read, so it's possible the gluten plan and query stage are not adjacent. This pr changes to collect leaves for gluten plan to make code robust.

## How was this patch tested?

add test


ulysses-you:
cc @PHILO-HE thank you

PHILO-HE (approved):
Looks good! Thanks!
We are fixing CI issue. Stay tuned.

PR #3982, opened 2023-12-08
[GLUTEN-3378][CORE] DeltaScanTransformer to support delta table

YannByron:

Based on the new framework defined in https://github.com/oap-project/gluten/pull/3843, modify the implementation of delta scan. 

## How was this patch tested?


gluten-core/src/main/scala/io/glutenproject/execution/DataSourceScanTransformerRegister.scala
```diff
@@ -0,0 +1,59 @@
...
+   * For example:
+   * {{{
+   *   override val scanClassName: String = "org.apache.iceberg.spark.source.SparkBatchQueryScan"
+   * }}}
+   */
+  val scanClassName: String
+
+  def createDataSourceTransformer(
+      batchScan: FileSourceScanExec,
+      partitionFilters: Seq[Expression]): FileSourceScanExecTransformer = {
```
yma11, line 48: Rename `partitionFilters` to `newPartitionFilters` so that avoid confusion as `FileSourceScanExec` also includes this kind of member?

yma11:
By the way, is it reasonable to move previous column mapping overrides into `DeltaScanTransformer`?

YannByron:
> By the way, is it reasonable to move previous column mapping overrides into `DeltaScanTransformer`?

They are two things. `DeltaScanTransformer` is used to replace this line: 
```
case "DeltaParquetFileFormat" => ReadFileFormat.ParquetReadFormat
```

to remove all the codes related to specified lake format in gluten-core.

leesf (approved):
+1

yma11 (approved):

sezruby:
@YannByron  are you planning to add protocol reader version check? 
It won't support a scan with deletion vector.

PR #3987, opened 2023-12-09
[GLUTEN-3378][VL] Iceberg support set partition columns in split info

liujiayi771:

Velox support specifying `partitionColumns` in the split info, which improves efficiency when reading partition columns as there's no need to read from the parquet file.

### DIFF
--- a/gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
+++ b/gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
@@ -140,8 +140,11 @@ object GlutenIcebergSourceUtil {
       partitionFields.zipWithIndex.foreach {
         case (field, index) =>
           val partitionValue = partition.get(index, field.`type`().typeId().javaClass())
+          val partitionType = field.`type`()
           if (partitionValue != null) {
-            partitionColumns.put(field.name(), partitionValue.toString)
+            partitionColumns.put(
+              field.name(),
+              TypeUtil.getPartitionValueString(partitionType, partitionValue))
           } else {
             partitionColumns.put(field.name(), ExternalCatalogUtils.DEFAULT_PARTITION_NAME)
           }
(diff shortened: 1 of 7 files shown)


liujiayi771:
@YannByron @yma11 Could you help review?

gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
```diff
@@ -45,7 +47,7 @@ object GlutenIcebergSourceUtil {
           paths.add(task.file().path().toString)
           starts.add(task.start())
           lengths.add(task.length())
-          partitionColumns.add(new JHashMap[String, String]())
+          partitionColumns.add(getPartitionColumns(task))
```
yma11, line 50: Can we assume the tasks share same `PartitionColumns` so that we don't need to get it for each task?

liujiayi771: The partition column is generally the same, but the partition values may vary.
image

gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxTPCHIcebergSuite.scala
```diff
@@ -84,3 +93,21 @@ class VeloxTPCHIcebergSuite extends VeloxTPCHSuite {
     }
   }
 }
+
+class VeloxPartitionedTableTPCHIcebergSuite extends VeloxTPCHIcebergSuite {
```
liujiayi771, line 97: @yma11 Added a test case that will execute all TPCH tests under a partitioned iceberg table.

gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
yma11: Get it. Thanks for show case.

liujiayi771:
@zzcclp ClickHouse CI failed. My changes should not have caused this failure. I see that #4037 also has this failure. Could you help fix it?

yma11:
@liujiayi771 please rebase this PR to resolve conflicts.

liujiayi771:
@yma11 Rebased and CI passed.

yma11 (approved):

PR #3990, opened 2023-12-10
[VL] Add config to velox's file read

FelixYBW:
Velox PR7217(https://github.com/facebookincubator/velox/pull/7217) added directbufferinput, which leads to performance regression seriously. The root cause is that the default config in the PR is not optimal for remote storage. You may find more talk here: https://github.com/facebookincubator/velox/pull/7873

The PR added 3 config:
loadQuantum: 256M (make sure it's larger than row group size, parquet default is 128M)
maxCoalesceDistance: 1M ( in case the columns are not load contieneously, like select a, c from table_with_column_a_b_c. If b is mall column than 1M, then we can load it to make a large block
CoalesceBytes: 64M, break the row group fetches into small chunks

With these configuration, here is the final traceview: You can see the S3 read is totally in parallel with data processing.

image


FelixYBW:
Velox PR created. https://github.com/facebookincubator/velox/pull/7978

FelixYBW:
With directinputbuffer, velox fetches data in iothreads (number is configed by  spark.gluten.sql.columnar.backend.velox.IOThreads), the task's memory pool is passed from query context, which is used to allocate memory for fetched data. There is a risk of memory leak check may fails.

The new PR https://github.com/oap-project/gluten/pull/4005 disabled the executor crash on memory leak, it put a warning instead

FelixYBW:
details are shown here: https://github.com/facebookincubator/velox/issues/8041

FelixYBW:
Velox https://github.com/facebookincubator/velox/pull/7978 merged, waiting for rebase

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -272,6 +284,30 @@ void VeloxBackend::initConnector(const std::shared_ptr<const facebook::velox::Co
...
+      velox::connector::hive::HiveConfig::kLoadQuantum, conf->get<std::string>(kLoadQuantum, "268435456")); // 256M
+  mutableConf->setValue(
+      velox::connector::hive::HiveConfig::kFooterEstimatedSize,
+      conf->get<std::string>(kDirectorySizeGuess, "32768")); // 32K
+  mutableConf->setValue(
+      velox::connector::hive::HiveConfig::kFilePreloadThreshold,
+      conf->get<std::string>(kFilePreloadThreshold, "1048576")); // 1M
+
+  // set cache_prefetch_min_pct = 0 to force all loads are prefetched in DirectBufferInput.
+  FLAGS_cache_prefetch_min_pct = 0;
```
zhli1142015, line 309: Could you help add gluten conf for this property also instead of hard code 0 here?

FelixYBW: Added, thank you

zhli1142015 (approved):
LGTM, thanks

fzhedu:
@FelixYBW how can we generate the traceview figure?

FelixYBW:
> @FelixYBW how can we generate the traceview figure?

Do you mean the chart in https://github.com/apache/incubator-gluten/pull/3990#issue-2034241461?

it needs to print log in velox code, then use a script to parse the log and generate traceview's json format.  you may find the json format spec here: 

https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview?tab=t.0#heading=h.yr4qxyxotyw

Here is the tool we used to generate eventlog's traceview:
https://github.com/apache/incubator-gluten/blob/main/tools/workload/benchmark_velox/analysis/sparklog.ipynb

PR #3994, opened 2023-12-11
[VL] Make bloom_filter_agg fall back when might_contain is not transformable

zhli1142015:
1. Reapply the change "[VL] Make bloom_filter_agg fall back when might_contain is not transformable".
2. Fix the issue only might_contain fallback.
3. Add flag to disable this in CK.
4. Add more UTs.

## How was this patch tested?

UT.


zhli1142015:
cc @zhouyuan , @baibaichen , thanks.

zhouyuan:
CC: @loneylee

zhli1142015:
@zhouyuan , @baibaichen , are you ok to reapply this change?
@baibaichen and @loneylee , bloom_filter_agg fallback rule is disabled for CK so there is no impact to you. You can remove the flag `enableBloomFilterAggFallbackRule` after you confirm this change works ok for CK.
Thanks.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -269,6 +270,52 @@ case class FallbackEmptySchemaRelation() extends Rule[SparkPlan] {
...
+          TransformHints.tagNotTransformable(agg, "related BloomFilterMightContain falls back")
+          tagNotTransformableRecursive(agg.child)
+        case a: org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec =>
+          tagNotTransformableRecursive(a.executedPlan)
+        case _ =>
+          p.children.map(tagNotTransformableRecursive)
+      }
+    }
+
+    plan.transformExpressions {
```
zhouyuan, line 311: @zhli1142015 is this the key change? 

`Fix the issue only might_contain fallback.`

zhouyuan (approved):
+1

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
zhli1142015: yes.

PR #4005, opened 2023-12-12
[VL] Disable checkUsageLeak in Velox memory manager

zhztheplayer:
`checkUsageLeak = true` causes crash since exception would be thrown from destructor once OOM.

We'd rely on Spark's own memory leak detection. (`spark.unsafe.exceptionOnMemoryLeak`)

In Velox, it might be practical to have an option to turn leak exceptions to warning messages.


FelixYBW:
we need to set spark.unsafe.exceptionOnMemoryLeak=false to disable the check from spark side as well

PR #4009, opened 2023-12-12
[VL] Open all glog when debug mode enabled

Yohahaha:


Yohahaha:
@zhztheplayer @marin-ma @PHILO-HE  @ulysses-you please help review this patch, thank you!

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -138,10 +138,14 @@ void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf
...
-  uint32_t severityLogLevel = veloxcfg->get<uint32_t>(kGlogSeverityLevel, kGlogSeverityLevelDefault);
-  FLAGS_v = vlogLevel;
-  FLAGS_minloglevel = severityLogLevel;
+  if (!veloxcfg->get<bool>(kDebugModeEnabled, false)) {
+    uint32_t vlogLevel = veloxcfg->get<uint32_t>(kGlogVerboseLevel, kGlogVerboseLevelDefault);
+    FLAGS_v = vlogLevel;
+    uint32_t severityLogLevel = veloxcfg->get<uint32_t>(kGlogSeverityLevel, kGlogSeverityLevelDefault);
+    FLAGS_minloglevel = severityLogLevel;
+  } else {
+    FLAGS_v = 99;
```
zhztheplayer, line 147: Is there one case that user set `kDebugModeEnabled` to `true` then tended to set glog level to lower value? With the change user would not be able to change glog level in debug mode.

Yohahaha: Do you mean `FLAGS_v=99` is strange? if yes, I can revert, verbose log is different with severity log, I just open them all for convenience.

Most debug info be printed by LOG(INFO), I can't image someone enable debug mode but disable log print.

zhztheplayer: > Do you mean FLAGS_v=99 is strange? 

This is OK to me. I am feeling that we have a couple of different options to control logging. Probably it's necessary to do some sort of simplifications.

zhztheplayer (approved):
@Yohahaha Please fill in the PR description. Thanks!

Yohahaha:
> @Yohahaha Please fill in the PR description. Thanks!

already updated, thank you!

Issue #3743, opened 2023-11-16
[VL] Support AttributeReference filter


Yohahaha:
### Description

select id from tbl where a = true; // col a is bool type

Function:flattenConditions, Expression: GetFlatConditions not supported for type '2', Source: RUNTIME, ErrorCode: NOT_IMPLEMENTED

PR #4017, opened 2023-12-12
[GLUTEN-3743][VL] Support pushed bool filter in scan

Yohahaha:

(Fixes: \#3743)

## How was this patch tested?

Add a test.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
@@ -26,7 +26,8 @@ import io.glutenproject.substrait.rel.{ReadRelNode, RelBuilder, SplitInfo}
 import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.catalyst.expressions.{And, Attribute, Expression}
+import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.types.BooleanType
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
 import com.google.common.collect.Lists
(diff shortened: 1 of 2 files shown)


Yohahaha:
Run Gluten Clickhouse CI

gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
```diff
@@ -113,6 +114,11 @@ trait BasicScanExecTransformer extends LeafTransformSupport with BaseDataSource
     }.asJava
     // Will put all filter expressions into an AND expression
     val transformer = filterExprs()
+      .map {
+        case ar: AttributeReference if ar.dataType == BooleanType =>
+          EqualNullSafe(ar, Literal.TrueLiteral)
```
rui-mo, line 119: Could you help double-check if null or false is returned for null input?

rui-mo (commented):
Thanks. Just one question, and please resolve the conflicts.

gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
Yohahaha: add ut to cover this case.

rui-mo (approved):

gluten-core/src/main/scala/io/glutenproject/execution/BasicScanExecTransformer.scala
```diff
@@ -113,6 +114,11 @@ trait BasicScanExecTransformer extends LeafTransformSupport with BaseDataSource
     }.asJava
     // Will put all filter expressions into an AND expression
     val transformer = filterExprs()
+      .map {
+        case ar: AttributeReference if ar.dataType == BooleanType =>
```
ulysses-you, line 118: does it work with not ? e.g., `WHERE !b`

Yohahaha: yes, `where !b` equal to `where b=null`

PR #4025, opened 2023-12-13
[VL] Increase async wait time to 30s when stopping tasks

zhztheplayer:
This will fix some core dump issues in the case of turning on split preloading, and to query for remote data source like S3 or remote HDFS cluster.

### DIFF
--- a/cpp/velox/memory/VeloxMemoryManager.cc
+++ b/cpp/velox/memory/VeloxMemoryManager.cc
@@ -286,7 +286,6 @@ bool VeloxMemoryManager::tryDestructSafe() {
 
 VeloxMemoryManager::~VeloxMemoryManager() {
   static const uint32_t kWaitTimeoutMs = FLAGS_gluten_velox_aysnc_timeout_on_task_stopping; // 30s by default
-  GLUTEN_CHECK(kWaitTimeoutMs > 0, "Aysnc timeout must be greater than 0");
   uint32_t accumulatedWaitMs = 0UL;
   for (int32_t tryCount = 0; accumulatedWaitMs < kWaitTimeoutMs; tryCount++) {
     if (tryDestructSafe()) {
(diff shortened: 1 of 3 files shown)


Yohahaha:
I remember you proposed use static memory pool for connector before, curious why it isn't being used now?

zhztheplayer:
> I remember you proposed use static memory pool for connector before, curious why it isn't being used now?

The PR was not merged. Since async allocations could happen on any memory pools including the task memory pool which cannot be directly replaced. This waiting solution is the one that was observed to be practical so far.

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -283,18 +283,20 @@ bool VeloxMemoryManager::tryDestructSafe() {
 }
 
 VeloxMemoryManager::~VeloxMemoryManager() {
-  // Wait (50 + 100 + 200 + 400 + 800)ms = 1550ms to let possible async tasks (e.g. preload split) complete.
-  for (int32_t tryCount = 0; tryCount < 5; tryCount++) {
+  static const uint32_t kWaitTimeoutMs = 30000UL; // 30s
```
zhouyuan, line 286: it's better to make this configurable

zhli1142015:
Is there any side effect, if we increase this value to 30s?
I think here we assume the threads would be released from IO threadpool in 30s, as task object is hold by io threads. If wrong, please correct me.
Thanks.

zhztheplayer:
Hi @zhli1142015

Some remote data sources could be slow in responding if network is bad. So I think it's reasonable to enlarge the value to a number that is close to network timeout which could usually be up to tens of seconds.

I'd make it configurable then user can keep it larger than the data source's connection timeout - which I think is practical to deal with previous core dump issues.

FelixYBW:
The fundamental solution is to make sure io threads release all memory allocated by memory pool before executor close. In theory IO thread should only allocate large memory block using memory pool and tracked by Spark. For tiny allocation just use std::malloc and counted them in the executor.overhead memory. All large block of memory allocation should be released when task ends, instead of executor quit.

FelixYBW:
what I observed is at the end only several Bytes are not released.

PR #4027, opened 2023-12-13
[VL][Delta]add more UT for delta lake and enable in CI

yma11:

Add time travel and partition filter UT for delta and enable in CI

## How was this patch tested?

UT

### DIFF
--- a/.github/workflows/velox_be.yml
+++ b/.github/workflows/velox_be.yml
@@ -24,6 +24,7 @@ on:
       - 'gluten-celeborn/**'
       - 'gluten-core/**'
       - 'gluten-data/**'
+      - 'gluten-delta/**'
       - 'gluten-ut/**'
       - 'shims/**'
       - 'tools/gluten-it/**'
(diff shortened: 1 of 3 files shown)


gluten-delta/src/main/scala/io/glutenproject/execution/DeltaScanTransformer.scala
```diff
@@ -14,9 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.glutenproject.extension
+package io.glutenproject.execution
```
zhztheplayer, line 17: So `extension` was by mistake?

zhztheplayer:
Thanks @yma11. And by the way, we may need to add the newly added module folders to [here](https://github.com/oap-project/gluten/blob/0f5d9e18615fc2978bfa286f5dada8e54e0f1acc/.github/workflows/velox_be.yml#L21-L36) and [here](https://github.com/oap-project/gluten/blob/0f5d9e18615fc2978bfa286f5dada8e54e0f1acc/.github/workflows/clickhouse_be_trigger.yml#L22-L31). So that a PR only changes the new modules would trigger CI.

gluten-delta/src/main/scala/io/glutenproject/execution/DeltaScanTransformer.scala
yma11: I think so. I synced with @YannByron and he will take a further look as there may still be some other issues related with column mapping.

yma11:
@zhztheplayer I updated this PR. please help review it again.

zhztheplayer (approved):
Thanks!

.github/workflows/velox_be.yml
```diff
@@ -24,6 +24,7 @@ on:
       - 'gluten-celeborn/**'
       - 'gluten-core/**'
       - 'gluten-data/**'
+      - 'gluten-delta/**'
```
zhztheplayer, line 27: But should we add `gluten-iceberg/**` too?

zhztheplayer (approved):

yma11:
Run Gluten Clickhouse CI

PR #4031, opened 2023-12-13
[VL] Add config to switch on/off spilling for partial aggregation

zhztheplayer:
After the patch, one can set `spark.gluten.sql.columnar.backend.velox.partialAggregationSpillEnabled=false` to disable spilling support for partial aggregation.


JkSelf (approved):
LGTM. Thanks.

Issue #4041, opened 2023-12-14
[CH] Use if instead of multiIf function if condtion size is 1


zhanglistar:
Since if function is faster than multiIf function, we should use if function if there is only one condition.

PR #4042, opened 2023-12-14
[CH] Use if function if condition size is 1

zhanglistar:

(Fixes: \#4041)
If performance is better than mulitiIf, so we use if function if we have only on condition.
## How was this patch tested?

CI
Function `if` has better performance than `mulitiIf` when there is only one condition, especially after PR:
https://github.com/ClickHouse/ClickHouse/pull/57885

On SQL 7954_0
Before:
image

After PR https://github.com/ClickHouse/ClickHouse/pull/57745
image

Still if is faster than multiIf:
image

### DIFF
--- a/cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
+++ b/cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
@@ -1993,10 +1993,10 @@ ASTPtr ASTParser::parseArgumentToAST(const Names & names, const substrait::Expre
             {
                 const auto & ifs = if_then.ifs(i);
                 auto if_node = parseArgumentToAST(names, ifs.if_());
-                args.emplace_back(std::move(if_node));
+                args.emplace_back(if_node);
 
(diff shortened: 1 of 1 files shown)


cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
```diff
@@ -1976,11 +1984,15 @@ ASTPtr ASTParser::parseArgumentToAST(const Names & names, const substrait::Expre
         }
         case substrait::Expression::RexTypeCase::kIfThen: {
             const auto & if_then = rel.if_then();
-            const auto * ch_function_name = "multiIf";
+            auto condition_nums = if_then.ifs_size();
+            std::string ch_function_name;
+            if (condition_nums == 1)
+                ch_function_name = "if";
+            else
+                ch_function_name = "multiIf";
```
taiyang-li, line 1992: shorter codes: std::string ch_function_name = condition_nums <=1 ?  "if" : "multiIf";

liuneng1994 (approved):
LGTM

PR #4047, opened 2023-12-14
[VL]Use velox Config to extract QueryContextConf

yma11:

Use velox Config to extract values for `QueryContextConf`

## How was this patch tested?

CI/CD

### DIFF
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -131,8 +131,8 @@ void VeloxBackend::init(const std::unordered_map<std::string, std::string>& conf
   gluten::Runtime::registerFactory(gluten::kVeloxRuntimeKind, veloxRuntimeFactory);
 
   // Init glog and log level.
-  auto veloxmemcfg = std::make_shared<facebook::velox::core::MemConfigMutable>(conf);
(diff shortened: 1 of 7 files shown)


cpp/velox/compute/VeloxBackend.cc
```diff
@@ -192,7 +192,7 @@ facebook::velox::cache::AsyncDataCache* VeloxBackend::getAsyncDataCache() const
 }
 
 // JNI-or-local filesystem, for spilling-to-heap if we have extra JVM heap spaces
-void VeloxBackend::initJolFilesystem(const facebook::velox::Config* conf) {
+void VeloxBackend::initJolFilesystem(const std::shared_ptr<const facebook::velox::Config> conf) {
```
FelixYBW, line 195: Let's use const std::shared_ptr<const facebook::velox::Config>& for this func and all below ones

FelixYBW:
Thank you for the fix!

cpp/velox/compute/VeloxBackend.cc
yma11: Updated.

FelixYBW (approved):

FelixYBW (approved):

PR #4056, opened 2023-12-15
[GLUTEN-4055][CORE] Wrapper reason with Option instead of Some to avoid reason is null

### LINKED ISSUE #4055: [CORE] Wrapper reason with `Option` instead of `Some` to avoid reason is null
### Description

Wrapper reason with `Option` instead of `Some` to avoid reason is null.

zwangsheng:

Wrapper reason with Option instead of Some to avoid reason is null

Close #4055 

## How was this patch tested?

CI


ulysses-you (approved):
lgtm

Issue #4058, opened 2023-12-15
Support JDK 11 and 17 for gluten


surnaik:
### Description

Lot of Spark Runtimes are now using JDK 11 and above instead of 8. The only issue currently is the removal of finalize method in JDK 11.

PR #2011, opened 2023-06-19
[VL][DNM] test for JDK 17

zhouyuan:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#ISSUE-ID)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #4059, opened 2023-12-15
[GLUTEN-4058][CORE] Support JDK 11 and 17 for Gluten

surnaik:

Support for JDK 11 and 17

Fixes: #4058 

## How was this patch tested?

Existing UTs

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
@@ -18,12 +18,18 @@ package org.apache.spark.util
 
 object GlutenShutdownManager {
 
+  private def isJava8: Boolean = Runtime.version().version().get(0) == 8
+
   def addHook(hook: () => Unit): AnyRef = {
     ShutdownHookManager.addShutdownHook(ShutdownHookManager.DEFAULT_SHUTDOWN_PRIORITY)(hook)
   }
 
+  // Java 11+ registers the Native Libraries for
+  // cleaning while loading in java/lang/ClassLoader.java#L2438
   def addHookForLibUnloading(hook: () => Unit): AnyRef = {
-    ShutdownHookManager.addShutdownHook(ShutdownHookManager.SPARK_CONTEXT_SHUTDOWN_PRIORITY)(hook)
+    ShutdownHookManager.addShutdownHook(ShutdownHookManager.SPARK_CONTEXT_SHUTDOWN_PRIORITY) {
+      if (isJava8) hook else () => Unit
+    }
   }
 
   def addHookForTempDirRemoval(hook: () => Unit): AnyRef = {
--- a/gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
@@ -18,7 +18,7 @@ package org.apache.spark.util
 
 object GlutenShutdownManager {
 
-  private def isJava8: Boolean = Runtime.version().version().get(0) == 8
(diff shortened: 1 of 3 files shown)


surnaik:
@zhouyuan Please review. Thanks

FelixYBW:
Thank you for your PR. Did you test with spark3.2, 3.3 and 3.4?

surnaik:
> Thank you for your PR. Did you test with spark3.2, 3.3 and 3.4?

Thanks @FelixYBW for taking a look, yes, I have tested with 3.2 and 3.3. I haven't tested 3.4

FelixYBW:
Is there performance difference?

surnaik:
> Is there performance difference?

No, there isn't any significant difference

FelixYBW:
Thank you for sharing. @zhouyuan @weiting-chen Can you have a review what should we do next?

surnaik:
Test failure seems unrelated

weiting-chen:
> Thank you for sharing. @zhouyuan @weiting-chen Can you have a review what should we do next?

Currently our CI system only run test with JDK8 and most of Gluten users are still using JDK8 as well.
We may consider to support JDK11 in next release.
If we decide moving to JDK11 and JDK17,we should also add CI system testing for JDK11 and 17 as well.
For JDK11, we can pick up this PR to run more testing and check if there is any issues for JDK11.
For JDK17, there was an unmerged PR(#2011 ) may also need to be added with some parameters, we should create a profile in pom file to manage which version to be used in Gluten.

surnaik:
I see, thanks @weiting-chen. Sure, we can take this up later. Yes, parameter needs to be added, I thought of adding it directly to the CI run using -DargLine when the CI supports JDK 11/17

zhouyuan:
@zzcclp @lgbo-ustc This patch tries to add JDK11 support, could you please kindly help to check the issue on CK backend?  There's a minor change on the shutdown hook
The log from CK backend showed a core dump on exiting:
```
13:16:23  # A fatal error has been detected by the Java Runtime Environment:
13:16:23  #
13:16:23  #  SIGSEGV (0xb) at pc=0x00007f302735b95e, pid=1020, tid=0x00007f2ff177f700
13:16:23  #
13:16:23  # JRE version: OpenJDK Runtime Environment (8.0_312-b07) (build 1.8.0_312-b07)
13:16:23  # Java VM: OpenJDK 64-Bit Server VM (25.312-b07 mixed mode linux-amd64 compressed oops)
13:16:23  # Problematic frame:
13:16:23  # C  [libch.so+0xf6b595e]  DB::CHJIT::deleteCompiledModule(DB::CHJIT::CompiledModule const&)+0xde
13:16:23  #
13:16:23  # Core dump written. Default location: /home/jenkins/agent/workspace/gluten/gluten-ci/gluten2/backends-clickhouse/core or core.1020
```

gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
```diff
@@ -18,12 +18,18 @@ package org.apache.spark.util
 
 object GlutenShutdownManager {
 
+  private def isJava8: Boolean = System.getProperty("java.version").equals("8")
```
baibaichen, line 21: @lwz9103 colud you please check whether we set this environment variable our CI environment?

surnaik: This is a property exposed by the JVM, it should be present in all the JVM

surnaik: But I think before Java 9, the java.version is set to 1.major.minor_subminor. I'll change this to parse the major version.

PR #4061, opened 2023-12-15
[GLUTEN-4060][MINOR] Remove the override method that is consistent with the parent

### LINKED ISSUE #4060: [MINOR] Remove the override method that is consistent with the parent
### Description

As title, to keep code clean

zwangsheng:

Remove the override method that is consistent with the parent to keep code clean

1. As unSupport `doExecuteColumnar` methods, let SparkPlan throw exception to keep output same

Close #4060

## How was this patch tested?

CI


zhouyuan (approved):
👍

PR #4064, opened 2023-12-15
[UI] Collapse the collapse table arrow of gluten build information

wForget:

`Gluten Build Information` is not very important, we can collapse it by default.

## How was this patch tested?

before:

image

after:

image


wForget:
@ulysses-you  could you please take a look?

FelixYBW (approved):

Issue #4065, opened 2023-12-15
Refactor soft affinity


baibaichen:
### Description

Currently `GlutenMergeTreePartition` is defined in core, let's move it to backend-clickhouse

PR #4066, opened 2023-12-15
[GLUTEN-4065][CH] Support forcedHostLocationTag

baibaichen:

In our internal Spark version, we have introduced support for a new `TaskLocality` called `FORCED_NODE_LOCAL`. We have also refactored `SoftAffinityUtil` to utilize node locality computing. Additionally, the definition of GlutenMergeTreePartition currently resides in the core; therefore, we should relocate it to a backend-clickhouse.

(Fixes: \#4065)

## How was this patch tested?

Using existed UT.


zzcclp (approved):
LGTM

PR #4067, opened 2023-12-15
[VL] Avoid early return when spark.gluten.sql.columnar.libpath is not blank

wForget:

`ListenerApiImpl#initialize` should not return early when libPath is empty.

## How was this patch tested?

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ListenerApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ListenerApiImpl.scala
@@ -149,11 +149,11 @@ class ListenerApiImpl extends ListenerApi {
     val libPath = conf.get(GlutenConfig.GLUTEN_LIB_PATH, StringUtils.EMPTY)
     if (StringUtils.isNotBlank(libPath)) { // Path based load. Ignore all other loadees.
       JniLibLoader.loadFromPath(libPath, false)
-      return
+    } else {
+      val baseLibName = conf.get(GlutenConfig.GLUTEN_LIB_NAME, "gluten")
+      loader.mapAndLoad(baseLibName, false)
+      loader.mapAndLoad(VeloxBackend.BACKEND_NAME, false)
     }
-    val baseLibName = conf.get(GlutenConfig.GLUTEN_LIB_NAME, "gluten")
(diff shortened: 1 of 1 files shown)


wForget:
@zhztheplayer could you please take a look?

wForget:
> What's the issue the PR is going to solve? Does it cause executor failure?

This is just an improvement that I noticed while reading the code. If I understand correctly, the subsequent initialization codes should be executed even if libPath is not blank.

https://github.com/oap-project/gluten/blob/b0085045d84de6d39969aa4e488626e23198a9be/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ListenerApiImpl.scala#L158-L164

So I think we should not return immediately when libPath is not blank

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ListenerApiImpl.scala
```diff
@@ -149,11 +149,11 @@ class ListenerApiImpl extends ListenerApi {
     val libPath = conf.get(GlutenConfig.GLUTEN_LIB_PATH, StringUtils.EMPTY)
     if (StringUtils.isNotBlank(libPath)) { // Path based load. Ignore all other loadees.
       JniLibLoader.loadFromPath(libPath, false)
-      return
+    } else {
```
PHILO-HE, line 152: It looks a bug. We usually don't set GLUTEN_LIB_PATH in config, so the issue is not exposed.
Maybe, this config is useless now for velox backend, but may keep it here for future use possibly. cc @zhztheplayer

zhztheplayer: @wForget So are you actually using `GlutenConfig.GLUTEN_LIB_PATH`?

Anyway the change still looks reasonable even if no one is using the option.

zhztheplayer (approved):
Thanks!

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/ListenerApiImpl.scala
wForget: > @wForget So are you actually using `GlutenConfig.GLUTEN_LIB_PATH`?

no, just noticed while reading the code

PR #4077, opened 2023-12-18
[CORE] Prefer to use Gluten plan in fallback policy

ulysses-you:

If the vanilla Spark plan also contains ColumnarToRow and the number of ColumnarToRow is not smaller than Gluten ColumnarToRow, we should better to prefer to use Gluten plan. For example,
```
Scan Parquet
     |
ColumnarToRow
```

This pr adds a new config `spark.gluten.sql.columnar.fallback.preferColumnar` and enables it by default.

To make it clear with `RowToColumnar`, this pr adds a new config `spark.gluten.sql.columnar.fallback.ignoreRowToColumnar` to control if we consider `RowToColumnar` as the fallback operator.

## How was this patch tested?

Add tests


ulysses-you:
cc @PHILO-HE @zhouyuan thank you

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -151,10 +157,7 @@ case class ExpandFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkP
               // we make the stage fall back.
               case tableCache if InMemoryTableScanHelper.isGlutenTableCache(tableCache) =>
                 stageFallbackCost = stageFallbackCost + 1
-              // For other case, table cache will save internal RowToColumnar if we make
-              // the stage fall back.
               case _ =>
-                stageFallbackCost = stageFallbackCost - 1
```
ulysses-you, line 157: @PHILO-HE I move it to the `countFallback` and add a config to control if we consider `RowToColumnar` as the fallback operator. Before, it's possbile that the `fallbackCost` is a negative number which looks a bit confused. After, we just count it to the fallback number if ignoreRowToColumnar is false.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -243,18 +246,44 @@ case class ExpandFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkP
     }
   }
 
+  private def countFallbackForVanillaSparkPlan(plan: SparkPlan): Int = {
```
PHILO-HE, line 249: Using word "Fallback" for vanilla spark may be misleading, as in Gluten it means going back to vanilla spark plan instead of using gluten plan.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
PHILO-HE: I think a negative fallback cost may make sense as it votes for falling back the stage, although the name is confusing.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
PHILO-HE: Can we put it into the calculation of fallback cost?

ulysses-you: There is a conflict, we do not know the fallback vanilla plan when we are at fallback cost. We should first do fallback then we can compare the fallback vanilla plan vs. gluten plan.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
ulysses-you: The point is that it's possible the gluten plan contains other RowToColumnar operators and we should handle it like fallback cost, so I move it to countFallback. If we want to keep a negative fallback cost here, we should change it to count the RowToColumnar operators instead of the row-based exchange. Does it sound good ?

PHILO-HE: In the initial design, we just considered fallback times which is measured by counting `ColumanrToRow` as you know. In Gluten's current code, as `Row2Columnar` is considered, negative fallback cost looks confusing indeed. But it also looks a bit confusing if we count it as fallback. Maybe, we can use some different concepts to make the code clear, e.g., transition cost.

For a given stage transformed by Gluten, I think we can calculate `transitionCost` by counting all `ColumnarToRow` & `RowToColumnar`. Then,  we can calculate `StageFallbackTransitionCost` by counting all `ColumnarToRow` for the fallback stage. Here, the usage of `ColumnarToRow` comes from: 1) last stage's columnar output; 2) vanilla spark's vectorized reader (use a config to control whether to count this?).

By comparing `transitionCost` - `StageFallbackTransitionCost` with user-specified threshold, we can let gluten fall back the stage or not.

It is basically equivalent to Gluten's current approach of evaluating fallback count & fallback cost. But I think it is more clear logically. Does it make sense?

ulysses-you: thank you @PHILO-HE  agree, I will refine the related code and naming.

ulysses-you:
cc @PHILO-HE any more comments ? thank you

yaooqinn (approved):

PR #4083, opened 2023-12-18
[DNM]remove vectorgraphics2d from xchart

weiting-chen:

Remove vectorgraphcis2d dependency from xchart since GPL license issue

## How was this patch tested?

Tested in my local server, but need to test pass in Github Action


weiting-chen:
This library "vectorgraphics2d" has used GPL license only and no other option can be used, so I remove it to avoid any concern from Apache's license check.

zhouyuan (approved):

Issue #4084, opened 2023-12-18
[VL] Verify supported file system


wForget:
### Description

The velox backend currently only supports `local/hdfs/s3a` file systems. We can check whether the schema of the scan file path is supported to avoid failure to read unsupportted files.

We have a custom file system internally that causes scanning files to fail:

```
23/12/18 14:10:24 ERROR TaskResources: Task 2 failed by error: 
io.glutenproject.exception.GlutenException: java.lang.RuntimeException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: No registered file system matched with file path 'qbfs://test01/XXX/part-00000-d67646ca-9136-4c51-b67f-7d087b020d19-c000.snappy.parquet'
Retriable: False
Context: Split [Hive: qbfs://test01/XXX/part-00000-d67646ca-9136-4c51-b67f-7d087b020d19-c000.snappy.parquet 0 - 651] Task Gluten_Stage_2_TID_2
Top-Level Context: Same as context.
Function: getFileSystem
../../velox/common/file/FileSystems.cpp
Line: 61
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKSsEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox11filesystems13getFileSystemESt17basic_string_viewIcSt11char_traitsIcEESt10shared_ptrIKNS0_6ConfigEE
# 4  _ZN8facebook5velox19FileHandleGeneratorclERKSs
# 5  _ZN8facebook5velox13CachedFactoryISsSt10shared_ptrINS0_10FileHandleEENS0_19FileHandleGeneratorEE8generateERKSs
# 6  _ZN8facebook5velox9connector4hive14HiveDataSource8addSplitESt10shared_ptrINS1_14ConnectorSplitEE
# 7  _ZN8facebook5velox4exec9TableScan9getOutputEv
# 8  _ZN8facebook5velox4exec6Driver11runInternalERSt10shared_ptrIS2_ERS3_INS1_13BlockingStateEERS3_INS0_9RowVectorEE
[...]

PR #4086, opened 2023-12-18
[GLUTEN-4084][CORE][VL] Verify supported file system

wForget:

Verify supported file system

Fixes: #4084

## How was this patch tested?

added test.

After this pr:

unsupported file system:

image
image
image

supported file system:

image


gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
```diff
@@ -79,6 +80,10 @@ class HiveTableScanExecTransformer(
     Seq.empty
   }
 
+  override def getRootPathsInternal: Seq[Path] = {
+    relation.tableMeta.storage.locationUri.map(new Path(_)).toSeq
```
PHILO-HE, line 84: Thanks for your work!
Could you clarify the overhead required for validating root paths? Does each file have a root path?

wForget: I think the overhead of rootPath is smaller, at least it doesn't need to be listed. For nodes without rootPath we will pass conversion.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -53,9 +56,10 @@ object VeloxBackend {
   val BACKEND_NAME = "velox"
 }
 
-object BackendSettings extends BackendSettingsApi {
+object BackendSettings extends BackendSettingsApi with Logging {
 
   val SHUFFLE_SUPPORTED_CODEC = Set("lz4", "zstd")
+  val SUPPORTED_FILE_SYSTEM = Set("file", "hdfs", "s3a")
```
leesf, line 62: oss can also be supported via s3a scheme.

PHILO-HE: Maybe, we should include all schemes in the below link and name this set `SUPPORTED_SCHEMES`. @leesf, does it make sense?
https://github.com/facebookincubator/velox/blob/aaaa0791008267d8be0a5c141db6da261aba8892/velox/connectors/hive/storage_adapters/s3fs/S3Util.h#L34-L45

PHILO-HE: If it is named `SUPPORTED_SCHEMES`, maybe each scheme name should be xxx://.

wForget: how about we provide a blacklist configuration to exclude?

wForget: all schemes:

```
Set(
 "file", // local
 "hdfs", // hdfs
 "s3", "s3a", "s3n", "oss", "oss", "cosn",  // s3
 "gs", // gcs
 "abfs", "abfss"  // abfs
) 
```

wForget: > If it is named `SUPPORTED_SCHEMES`, maybe each scheme name should be xxx://.

Do we need that? The scheme obtained by `java.net.URI#getScheme` does not have the '://' suffix.

PHILO-HE: Maybe, using a configurable blacklist is not necessary.

BTW, I note velox registers S3/HDFS/GCS/ABFS according to the corresponding build option. If the validation passes on Gluten side, but velox doesn't actually register the file system. Can the expected fallback still happen?
https://github.com/facebookincubator/velox/blob/aaaa0791008267d8be0a5c141db6da261aba8892/velox/connectors/hive/HiveConnector.cpp#L151-L159

PHILO-HE: > > If it is named `SUPPORTED_SCHEMES`, maybe each scheme name should be xxx://.
> 
> Do we need that? The scheme obtained by `java.net.URI#getScheme` does not have the '://' suffix.

OK to me if "://" is lacking.

wForget: > BTW, I note velox registers S3/HDFS/GCS/ABFS according to the corresponding build option. If the validation passes on Gluten side, but velox doesn't actually register the file system. Can the expected fallback still happen?

Good point. How about we add relevant comments in the code? It prompts to recompile velox for support.

PHILO-HE: OK. Please add some comments to clarify. Thanks!

wForget: > OK. Please add some comments to clarify. Thanks!

Added, please help to continue the review, thanks.

PHILO-HE:

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
leesf: @PHILO-HE agree that we should not break the capabilities velox already have. In addition, if users add new schemes to velox(which is compatible with s3a), we must also fix it in gluten, which would make it a little complex. But I am open with PR.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -53,18 +58,46 @@ object VeloxBackend {
...
+    // local
+    "file",
+    // hdfs
+    "hdfs",
+    // s3
+    "s3",
+    "s3a",
+    "s3n",
+    "oss",
+    "oss",
```
leesf, line 83: duplicate items.

wForget: Fixed, thanks.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
PHILO-HE: >  if users add new schemes to velox(which is compatible with s3a), we must also fix it in gluten

For making file system support state always consistent with velox, one way is to do the validation on native side, i.e., calling velox [getFileSystem](https://github.com/facebookincubator/velox/blob/5144f141cfe2cc832d324085ed570457c22a9a77/velox/common/file/FileSystems.cpp#L52) to validate. @wForget, would you like to improve this PR?

wForget: Sure, I will try it.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -83,33 +116,43 @@ object BackendSettings extends BackendSettingsApi {
...
+        logWarning(
           s"Validation failed for ${this.getClass.toString}" +
             s" due to: data type $unsupportedDataType. in file schema. ")
-        // scalastyle:on println
       }
       unsupportedDataTypes.isEmpty
     }
 
+    // check if supported file system
+    val unsupportedPath = rootPaths.find(!isSupportedFileSystem(_))
```
rui-mo, line 127: Do we need to verify all the root paths? Is it a usual case for files belonging to different systems?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -307,4 +350,6 @@ object BackendSettings extends BackendSettingsApi {
   override def staticPartitionWriteOnly(): Boolean = true
 
   override def allowDecimalArithmetic: Boolean = SQLConf.get.decimalOperationsAllowPrecisionLoss
+
+  override def requiredRootPaths(): Boolean = true
```
rui-mo, line 354: Do we need this variable, and in what case the root paths are not required?

rui-mo (commented):
Thanks for working on this. Just two quesitons.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
wForget: > Do we need to verify all the root paths? Is it a usual case for files belonging to different systems?

Maybe first is enough. I think the rootPaths of one datasource are mostly the same filesystem.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
wForget: > Do we need this variable, and in what case the root paths are not required?

I'm not familiar with clickhouse backend so I kept this and enabled it only for velox backend.

rui-mo: @zzcclp Could you help check if it's needed for CH backend? Thanks.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
rui-mo: @PHILO-HE What do you think?

PHILO-HE: Maybe, multiple kinds of file system are not commonly used in scan. We may just validate the first one. If doing so, let's leave some comment to clarify.

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHBackend.scala
```diff
@@ -129,7 +131,8 @@ object CHBackendSettings extends BackendSettingsApi with Logging {
       format: ReadFileFormat,
       fields: Array[StructField],
       partTable: Boolean,
-      paths: Seq[String]): Boolean = {
+      paths: Seq[String],
+      rootPaths: Seq[Path]): Boolean = {
```
zhztheplayer, line 135: Minor: Do we have to have `paths` and  `rootPaths` together in the parameter list? Would it be reasonable to use the latter to extract the former?

wForget: `paths` has a large initialization cost and is empty by default in velox backend.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -53,18 +58,46 @@ object VeloxBackend {
...
 
+  // scalastyle:off line.size.limit
+  /**
+   * Supported file schemes for Velox backend.
+   *
+   * Note: Need to turn on the required file systems when compiling velox to obtain support. Refer:
+   * https://github.com/facebookincubator/velox/blob/fa6de06c826084ce5d938c71b7450c59204c864d/CMakeLists.txt#L93-L96
+   */
+  // scalastyle:on line.size.limit
+  val SUPPORTED_FILE_SCHEMES = Set(
```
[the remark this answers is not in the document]
wForget: Maybe you missed the discussion in https://github.com/oap-project/gluten/pull/4086#discussion_r1430817063. We will check if filesystem is supported via native method. What do you think?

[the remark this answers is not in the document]
zhli1142015: I see.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
wForget: Fixed, thanks

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
wForget: > > if users add new schemes to velox(which is compatible with s3a), we must also fix it in gluten
> 
> For making file system support state always consistent with velox, one way is to do the validation on native side, i.e., calling velox [getFileSystem](https://github.com/facebookincubator/velox/blob/5144f141cfe2cc832d324085ed570457c22a9a77/velox/common/file/FileSystems.cpp#L52) to validate. @wForget, would you like to improve this PR?

@PHILO-HE  I added a native method to check paths, could you please help me take a look?

wForget:
@PHILO-HE @rui-mo could you please take a look again if you have time?

PHILO-HE:
Thanks so much for the iterations! Can we do the validation in `SubstraitToVeloxPlanValidator`? I think it would be better.

wForget:
> Thanks so much for the iterations! Can we do the validation in `SubstraitToVeloxPlanValidator`? I think it would be better.

Thanks for your guidance I will try that.

zhli1142015:
Hello @wForget,
I saw this PR is closed, will you continue this PR?
Thanks.

wForget:
> Hello @wForget, I saw this PR is closed, will you continue this work? Thanks.

I disabled columnar filescan by adding `spark.gluten.sql.columnar.filescan=false;` config, so this PR is unnecessary for me, if this is useful I can continue to work on it. cc @PHILO-HE WDYT?

zhli1142015:
We need this PR, when reading from for not supported file systems, falling back is always better than failing.

PHILO-HE:
> > Hello @wForget, I saw this PR is closed, will you continue this work? Thanks.
> 
> I disabled columnar filescan by adding `spark.gluten.sql.columnar.filescan=false;` config, so this PR is unnecessary for me, if this is useful I can continue to work on it. cc @PHILO-HE WDYT?

@wForget, do you remember the landing gap of this pr? Thanks!

wForget:
> @wForget, do you remember the landing gap of this pr? Thanks!

I remember the current status should be waiting for review, but I didn't move forward with it because I found `spark.gluten.sql.columnar.filescan=false` configuration. Do we need to reopen this PR?

PHILO-HE:
> > @wForget, do you remember the landing gap of this pr? Thanks!
> 
> I remember the current status should be waiting for review, but I didn't move forward with it because I found `spark.gluten.sql.columnar.filescan=false` configuration. Do we need to reopen this PR?

@wForget, I see. Thanks for your work!
@zhli1142015, could you take over this work if it is really necessary to you? You can create a new pr based on this one, with @wForget's work highlighted there.

zhli1142015:
> > > @wForget, do you remember the landing gap of this pr? Thanks!
> > 
> > 
> > I remember the current status should be waiting for review, but I didn't move forward with it because I found `spark.gluten.sql.columnar.filescan=false` configuration. Do we need to reopen this PR?
> 
> @wForget, I see. Thanks for your work! @zhli1142015, could you take over this work if it is really necessary to you? You can create a new pr based on this one, with @wForget's work highlighted there.

Sure, thanks @wForget 's work.

Issue #4100, opened 2023-12-19
[VL] Should not apply `collect_set` to `array_distinct`


zwangsheng:
### Description

`collect_set` is a agg func in Spark, `array_distinct` is a array func not agg func.

So we can't simply use `array_distinct` instead of `collect_set`.

PR #4101, opened 2023-12-19
[GLUTEN-4100][VL] Should not simply apply `collect_set` to `array_distinct`

zwangsheng:

`collect_set` is a agg func in Spark, `array_distinct` is a array func not agg func.

So we can't simply use array_distinct instead of collect_set.

Close: #4100 

## How was this patch tested?

CI


zhouyuan:
CC: @PHILO-HE

PHILO-HE:
This will be fixed by anther drafted PR. However, we can merge this PR to avoid misunderstanding.

PHILO-HE (approved):

PR #4102, opened 2023-12-19
[VL] Mark SplitPreloadPerDriver conf as dynamic config

zhli1142015:

Mark SplitPreloadPerDriver conf as dynamic config, then it can be passed down to native in `getNativeSessionConf`. Now this config is consumed in [getQueryContextConf](https://github.com/oap-project/gluten/blob/main/cpp/velox/compute/WholeStageResultIterator.cc#L404).

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -515,9 +515,6 @@ object GlutenConfig {
       (
         COLUMNAR_VELOX_CONNECTOR_IO_THREADS.key,
         COLUMNAR_VELOX_CONNECTOR_IO_THREADS.defaultValueString),
-      (
-        COLUMNAR_VELOX_SPLIT_PRELOAD_PER_DRIVER.key,
-        COLUMNAR_VELOX_SPLIT_PRELOAD_PER_DRIVER.defaultValueString),
       (COLUMNAR_SHUFFLE_CODEC.key, ""),
       (COLUMNAR_SHUFFLE_CODEC_BACKEND.key, ""),
       ("spark.hadoop.input.connect.timeout", "180000"),
@@ -1095,7 +1092,7 @@ object GlutenConfig {
       .createWithDefault(30000)
 
   val COLUMNAR_VELOX_SPLIT_PRELOAD_PER_DRIVER =
-    buildStaticConf("spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver")
+    buildConf("spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver")
       .internal()
       .doc("The split preload per task")
       .intConf


zhli1142015:
cc @zhouyuan @FelixYBW

zhli1142015:
> cc @zhouyuan @FelixYBW

And looks split reploading is not enabled by default, should we change default value of `spark.gluten.sql.columnar.backend.velox.IOThreads` to some number > 0 (for example: 8)

zhli1142015:
> will the split prefetch only fetch the first rowgroup of a split or fetch all rowgroups? if only fetch the first rowgroup, we may enable it by default, otherwise it depends on customer data. If a split has multiple rowgroups, we should only prefetch the first one, then set rowgroup prefetch


I think only first row group.

Yohahaha:
SplitPreload will prefetch row groups according to config, not only first row group.

zhli1142015:
I see.

FelixYBW:
> SplitPreload will prefetch row groups according to config, not only first row group.

does it use the same directinputbuffer to fetch? if so it should have the same behavior as first split. It may prefetch prefetch_rowgroup# of rowgroups.

zhouyuan:
> > cc @zhouyuan @FelixYBW
> 
> And looks split reploading is not enabled by default, should we change default value of `spark.gluten.sql.columnar.backend.velox.IOThreads` to some number > 0 (for example: 8)

@zhztheplayer  helped to turn on this by default but failed some tests, will take a look on this
https://github.com/oap-project/gluten/pull/3876

zhli1142015:
Thanks for the discussion, can this PR be merged? @zhouyuan , @FelixYBW 
We set `spark.gluten.sql.columnar.backend.velox.SplitPreloadPerDriver` as 4, but found default value 2 is used in velox. This pr fixes this issue.

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -515,9 +515,6 @@ object GlutenConfig {
       (
         COLUMNAR_VELOX_CONNECTOR_IO_THREADS.key,
         COLUMNAR_VELOX_CONNECTOR_IO_THREADS.defaultValueString),
-      (
-        COLUMNAR_VELOX_SPLIT_PRELOAD_PER_DRIVER.key,
-        COLUMNAR_VELOX_SPLIT_PRELOAD_PER_DRIVER.defaultValueString),
```
zhztheplayer, line 520: Can you explain this change? Thanks.

zhztheplayer (approved):

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
zhli1142015: Now we need to pass this config to native by `getNativeSessionConf`, so these two lines should be removed.
https://github.com/oap-project/gluten/blob/main/cpp/velox/compute/WholeStageResultIterator.cc#L404

FelixYBW:
record here, split preload will use one of the IO thread to start the preload work, then the other IO threads are used to fetch data. Which leads more complex memory allocation actions in preload thread, and risk to get memory pool issue when task thread ends.

PR #4104, opened 2023-12-19
[CORE] Fix miss RowToColumnar with columnar table cache in AQE

ulysses-you:

Since https://github.com/apache/spark/pull/43484 Spark supports cache columnar batch, then outputsColumnar of AQE can be true. It breaks the assumption that the output is always row-based. This pr makes fallback policy compaible with this behavior.

## How was this patch tested?

add test

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
@@ -247,7 +247,7 @@ case class ExpandFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkP
   }
 
   override def apply(plan: SparkPlan): SparkPlan = {
-    // By default, the outputsColumnar if always false.
+    // By default, the outputsColumnar is always false.
     // The outputsColumnar will be true if it is a cached plan and we are going to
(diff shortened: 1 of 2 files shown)


yaooqinn (dismissed):

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -240,13 +241,20 @@ case class ExpandFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkP
       case ColumnarToRowExec(child: SparkPlan)
           if InMemoryTableScanHelper.isGlutenTableCache(child) =>
         child
+      case plan: RowToColumnarExec =>
+        BackendsApiManager.getSparkPlanExecApiInstance.genRowToColumnarExec(plan.child)
     }
   }
 
   override def apply(plan: SparkPlan): SparkPlan = {
+    // By default, the outputsColumnar if always false.
```
PHILO-HE, line 250: typo: is always false.

PHILO-HE (commented):
Looks good! Just a comment for typo. Thanks!

PHILO-HE (approved):
Thanks！

PR #4112, opened 2023-12-19
[VL] Use partial companion functions for distinct aggregation

rui-mo:

Flush of intermediate aggregation needs to be disabled if it is used for distinct computing. But disabling flushing is tricky because doing so will require enabling spilling. In this PR, we use partial companion functions instead to make sure that final aggregation returns "intermediate" results.

https://github.com/oap-project/velox/pull/465

## How was this patch tested?

Verfied on Jenkins.


backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
```diff
@@ -174,7 +174,14 @@ case class HashAggregateExecTransformer(
...
   override protected def modeToKeyWord(aggregateMode: AggregateMode): String = {
-    super.modeToKeyWord(if (mixedPartialAndMerge) Partial else aggregateMode)
+    super.modeToKeyWord(if (mixedPartialAndMerge) {
+      Partial
+    } else {
+      aggregateMode match {
+        case PartialMerge => Final
+        case _ => aggregateMode
+      }
+    })
```
zhztheplayer, line 184: Can we use `Final` for all aggregate modes?

zhztheplayer (approved):
Thanks!

Issue #4115, opened 2023-12-20
[CH] Grouping without aggregate functions may have wrong result


lgbo-ustc:
### Backend

CH (ClickHouse)

### Bug description

[Expected behavior] and [actual behavior].

Consider following sql without aggregate functions
```sql
select n_regionkey, n_nationkey from nation group by n_regionkey, n_nationkey
```
At present, the 1st and 2nd aggregate stages are both executed by `StreamingAggregatingStep` or `AggregatingStep` . If the grouping keys set is very large and cause spill in 2nd aggregate stage, the result will be wrong.

BTW, There is no information to indicate in which stage it is.

### Spark version

None

PR #4124, opened 2023-12-20
[GLUTEN-4115][CH] Identify the final aggregate stage of aggregating without aggregate functions

lgbo-ustc:

Fixes: #4115

For aggregate without any aggregate functions, CH needs to know in which aggregate stage it is.

## How was this patch tested?

unit tests

### DIFF
--- a/backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashAggregateExecTransformer.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashAggregateExecTransformer.scala
@@ -16,11 +16,13 @@
  */
 package io.glutenproject.execution
 
+import io.glutenproject.backendsapi.BackendsApiManager
 import io.glutenproject.execution.CHHashAggregateExecTransformer.getAggregateResultAttributes
 import io.glutenproject.expression._
-import io.glutenproject.substrait.`type`.TypeNode
+import io.glutenproject.substrait.`type`.{TypeBuilder, TypeNode}
 import io.glutenproject.substrait.{AggregationParams, SubstraitContext}
 import io.glutenproject.substrait.expression.{AggregateFunctionNode, ExpressionBuilder, ExpressionNode}
+import io.glutenproject.substrait.extensions.{AdvancedExtensionNode, ExtensionBuilder}
 import io.glutenproject.substrait.rel.{RelBuilder, RelNode}
 
 import org.apache.spark.sql.catalyst.expressions._
@@ -29,7 +31,11 @@ import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.aggregate.BaseAggregateExec
 import org.apache.spark.sql.types._
 
+import com.google.protobuf.{Any, StringValue}
+
 import java.util
+
+import scala.collection.JavaConverters._
 
 object CHHashAggregateExecTransformer {
   def getAggregateResultAttributes(
@@ -388,5 +394,26 @@ case class CHHashAggregateExecTransformer(
   override protected def withNewChildInternal(
       newChild: SparkPlan): CHHashAggregateExecTransformer = {
     copy(child = newChild)
+  }
+
+  override protected def getAdvancedExtension(
+      validation: Boolean = false,
+      originalInputAttributes: Seq[Attribute] = Seq.empty): AdvancedExtensionNode = {
+    val enhancement = if (validation) {
+      // Use a extension node to send the input types through Substrait plan for validation.
+      val inputTypeNodeList = originalInputAttributes
+        .map(attr => ConverterUtils.getTypeNode(attr.dataType, attr.nullable))
+        .asJava
+      Any.pack(TypeBuilder.makeStruct(false, inputTypeNodeList).toProtobuf)
+    } else {
+      null
+    }
+    val optimizationContent = s"has_required_child_distribution_expressions=" +
+      s"${requiredChildDistributionExpressions.isDefined}\n"
+    val optimization =
+      BackendsApiManager.getTransformerApiInstance.packPBMessage(
+        StringValue.newBuilder.setValue(optimizationContent).build)
+    ExtensionBuilder.makeAdvancedExtension(optimization, enhancement)
+
   }
 }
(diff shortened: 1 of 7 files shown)


backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
```diff
@@ -2253,5 +2253,13 @@ class GlutenClickHouseTPCHParquetSuite extends GlutenClickHouseTPCHAbstractSuite
     }
   }
 
+  test("GLUTEN-4115 aggregate without any function") {
+    val sql =
+      """
+        | select n_regionkey, n_nationkey from nation group by n_regionkey, n_nationkey
```
rui-mo, line 2259: For case like this, can we just treat it as complete stage? If so, distinct outputs can be got from partial stage.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -59,6 +59,12 @@ abstract class HashAggregateExecBaseTransformer(
     child.output ++ aggregateBufferAttributes ++ aggregateAttributes ++
       aggregateExpressions.flatMap(_.aggregateFunction.inputAggBufferAttributes)
 
+  // According to planAggregateWithoutDistinct in AggUtils.scala, an aggregate without aggregate
+  // functions will only be organized in two stages, partial and final. In partial stage
+  // requiredChildDistributionExpressions is None, but in final stage it is Some(Seq[Expression]).
+  protected lazy val isFinalStage: Boolean = aggregateExpressions.exists(_.mode == Final) ||
+    (aggregateExpressions.isEmpty && requiredChildDistributionExpressions.isDefined)
```
zhztheplayer, line 66: I am feeling that Velox doesn't require for this information as of now. Would it be possible to move it to advanced extension? Or you think it could be valuable as a common feature?

zhztheplayer:
> For aggregate without any aggregate functions, CH needs to know in which aggregate stage it is.

Would you like to share a little bit more information about this in CH? How does CH handle differently between non-final stage and final stage of a distinct?

backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseTPCHParquetSuite.scala
lgbo-ustc: It will cause performance issue

lgbo-ustc:
> > For aggregate without any aggregate functions, CH needs to know in which aggregate stage it is.
> 
> Would you like to share a little bit more information about this in CH? How does CH handle differently between non-final stage and final stage of a distinct?

In partial aggregating stage, CH doesn't wait all input data to complete. Once the memory usage has reache limit, it will spill partial Intermediate result to final stage. Only the final stage could guarantee the grouping keys are unique in the result。

For example，in partial stage, we run `select a from t group by a`. (1), (1), (2) come in, and trigger spill, it generates (1), (2) to downstream. And then comes (1), (3), it generate (1), (3) to downstream. The result have duplicate keys.

At present, CH use partial aggregating operator to handle all the stages that aggregate without aggregate functions. That means if spill happen in the final stage, the aggregate result will be wrong.

If we use final aggregating operator  to handle all the stages, there will be performance problems when the grouping keys are high cardinality.

There is no information is current subtraint protocol to tell us, in which the stage it is

zhztheplayer:
@lgbo-ustc 

Thank you for sharing the context!

> How does velox handle this case?

Currently in Velox backend, we are leveraging Velox's `PartialAggreation` that works pretty much CH's partial step which could emit (or "spill" as you mentioned) intermediate data. So we too observed a couple of wrong result bugs like your case in the past few days.

> Treat them all as a final stage ?

We happen to be working on treating them all as final stage, please refer to our pending PRs https://github.com/oap-project/gluten/pull/4112 https://github.com/oap-project/gluten/pull/4130. So after the work get merged the relevant incorrect result issues will be resolved.

Also we will add `allowIntermediateOutput`  (in #4130) flag only for Velox backend through Substrait advanced extension. The flag looks like similar to `is_final_stage` you are about to add within this patch but they don't come from the same calculation. So I am inclined we can both use Substrait advanced extension without changing common module code, in the first version? We can still try extracting common part in future.

lgbo-ustc:
@zhztheplayer I think I should wait #4130 finished to minimize code changes

zhztheplayer:
> @zhztheplayer I think I should wait #4130 finished to minimize code changes

Aha, #4130 basically doesn't change common module's code. I'll be good if you'd like to extract some part out from `backend-velox` after it gets merged.

backends-clickhouse/src/main/scala/io/glutenproject/execution/CHHashAggregateExecTransformer.scala
```diff
@@ -389,4 +395,26 @@ case class CHHashAggregateExecTransformer(
...
+    val enhancement = if (validation) {
+      // Use a extension node to send the input types through Substrait plan for validation.
+      val inputTypeNodeList = originalInputAttributes
+        .map(attr => ConverterUtils.getTypeNode(attr.dataType, attr.nullable))
+        .asJava
+      Any.pack(TypeBuilder.makeStruct(false, inputTypeNodeList).toProtobuf)
+    } else {
+      null
+    }
+    val optimizationContent = s"isStreaming=$isStreaming\n" +
```
zzcclp, line 411: why needs this `isStreaming` parameter

lgbo-ustc: keep it from hash base transform

zzcclp: but this `isStreaming` is different from the one in the `HashAggregateExecBaseTransformer`

zzcclp (approved):
LGTM

PR #4139, opened 2023-12-21
[VL] add data types test for ORC file format

yma11:

add data types test for ORC file format, note that timestamp is not supported in ORC.

## How was this patch tested?

UT


JkSelf (approved):
LGTM. Thanks.

PR #4145, opened 2023-12-21
[GLUTEN-4144][CORE]Use Weak to Weak ReferenceMap instead of WeakHashMap

srinivasst:
tmm: TaskMemoryManager will never be garbage collected as WeakHashMap holds a hard reference over TreeMemoryTarget which holds a reference of TaskMemoryManager. The map entries are hence never removed.

This causes an on-heap memory leak.
Before Fix:
image
```
java.lang.OutOfMemoryError: Java heap space
	at java.util.Arrays.copyOf(Arrays.java:3745) ~[?:?]
	at java.io.ByteArrayOutputStream.grow(ByteArrayOutputStream.java:120) ~[?:?]
	at java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:95) ~[?:?]
	at java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:156) ~[?:?]
	at org.apache.spark.util.ByteBufferOutputStream.write(ByteBufferOutputStream.scala:41) ~[spark-core_2.12-3.3.2.jar:3.3.2]
	at java.io.ObjectOutputStream$BlockDataOutputStream.drain(ObjectOutputStream.java:1873) ~[?:?]
	at java.io.ObjectOutputStream$BlockDataOutputStream.setBlockDataMode(ObjectOutputStream.java:1782) ~[?:?]
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1180) ~[?:?]
	at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:345) ~[?:?]
	at org.apache.spark.serializer.JavaSerializationStream.writeObject(JavaSerializer.scala:46) ~[spark-core_2.12-3.3.2.jar:3.3.2]
	at org.apache.spark.serializer.JavaSerializerInstance.serialize(JavaSerializer.scala:115) ~[spark-core_2.12-3.3.2.jar:3.3.2]
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:593) ~[spark-core_2.12-3.3.2.jar:3.3.2]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[?:?]
	at java.lang.Thread.run(Thread.java:829) ~[?:?]
```
After Fix
image


srinivasst:
@zhztheplayer please take a look at this PR

ulysses-you:
thank you @srinivasst , please reformat the code style using
 `mvn spotless:apply -Pspark-3.2,spark-3.3,spark-3.4 -Pbackends-velox -Pbackends-clickhouse -Prss -Pspark-ut`

Yohahaha:
thank you for the fix and curious how do you measure off-heap usage? could you share more about this?

surnaik:
@srinivasst Could you run `mvn spotless:apply`

gluten-core/src/main/java/io/glutenproject/memory/memtarget/spark/TreeMemoryConsumers.java
```diff
@@ -62,7 +64,8 @@ public static Factory shared() {
 
   public static class Factory {
 
-    private static final WeakHashMap<TaskMemoryManager, TreeMemoryTarget> MAP = new WeakHashMap<>();
+    private static final ReferenceMap<TaskMemoryManager, TreeMemoryTarget> MAP =
+            new ReferenceMap<>(ReferenceStrength.WEAK, ReferenceStrength.WEAK);
```
zhztheplayer, line 68: This looks very suitable for our case except I have one single question: Since the map key is `tmm` and map value is usually `TreeMemoryConsumer(tmm)`, would it mean the map value could ideally get garbage-collected before the corresponding key? When this happens, is it possible to have `null` returned from `#getSharedAccount` which lead to NPE?

zhztheplayer (commented):
Thank you very much for catching this @srinivasst!

srinivasst:
@Yohahaha I used the following script to measure memory.
```shell
#!/bin/bash

# Usage: script.sh <PID> <Output File> <Sleep Duration in Seconds> <Number of Iterations>

PID=$1
OUTPUT_FILE=$2
SLEEP_DURATION=$3
ITERATIONS=$4

rm $OUTPUT_FILE
for i in $(seq 1 $ITERATIONS); do
#while true; do
    current_time=$(date +%s)
    echo $PID | xargs -I {} sh -c 'echo $(date +%s | awk "{print \$1 \",\"}")$(sudo jstat -gc {} | awk "NR==1 {for (i=1; i<=NF; i++) idx[\$i] = i} NR==2 {printf \"%s,%s,%s,%s,%s,\", \$idx[\"S0U\"], \$idx[\"S1U\"], \$idx[\"EU\"], \$idx[\"OU\"], \$idx[\"MU\"]}")$(sudo top -p {} -b -n 1 | grep {} | awk "{print \$6}")' >> $OUTPUT_FILE
    sleep $SLEEP_DURATION
done
```
Output is a csv file with follwing columns: Epoch, S0U, S1U, EU, OU, MU, RES. We can derive onHeap by S0U + S1U + EU + OU + MU and offHeap by RES - onHeap

gluten-core/src/main/java/io/glutenproject/memory/memtarget/spark/TreeMemoryConsumers.java
srinivasst: As per the documentation of org.apache.commons.collections4.map.ReferenceMap: [If non-hard references are used, then the garbage collector can remove mappings if a key or value becomes unreachable](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/ReferenceMap.html#:~:text=If%20non%2Dhard%20references%20are%20used%2C%20then%20the%20garbage%20collector%20can%20remove%20mappings%20if%20a%20key%20or%20value%20becomes%20unreachable)

If value is garbage collected, both key value pair will be removed.

zhztheplayer (approved):
Thanks!

zhztheplayer (changes requested):
Would you correct the code style? I see @ulysses-you has given instructions. Thanks.

srinivasst:
Spark 3.2 uses an older version of the ReferenceMap. Will create a shim to address this.

srinivasst:
@ulysses-you @zhztheplayer 
The code has been reformatted. 
I'm using an older version of ReferenceMap to allow compatibility with Spark 3.2.

zhztheplayer (approved):
Thanks!

ulysses-you (approved):

Issue #4155, opened 2023-12-22
[VL] Limit the number of compile jobs in `builddeps-veloxbe.sh`


dcoliversun:
### Description

It is better to limit the number of compile jobs in `builddeps-veloxbe.sh`
https://github.com/oap-project/gluten/blob/ed843672bc1223aa2dbe44281d51db5906740882/dev/builddeps-veloxbe.sh#L159

Just like `compile.sh`  
https://github.com/oap-project/gluten/blob/ed843672bc1223aa2dbe44281d51db5906740882/cpp/compile.sh#L146

Because compile broke caused by many jobs when exec `./dev/builddeps-veloxbe.sh --build_tests=ON --build_benchmarks=ON --enable_s3=ON --enable_hdfs=ON`.

dcoliversun:
similar with #4123 so close this

PR #4123, opened 2023-12-20
[GLUTEN-4122][VL] Add num_threads option to limit build threads for velox

wForget:
## What changes were proposed in this pull request?

Add NUM_THREADS argument to limit compilation parallelism

Fixes: #4122

PR #4157, opened 2023-12-22
[GLUTEN-4155][VL] Limit the number of compile jobs in builddeps-veloxbe.sh

dcoliversun:

This PR aims to limit the number of compile jobs in builddeps-veloxbe.sh to reduce compile error caused by resources.

(Fixes: \#4155)

## How was this patch tested?

Pass build test in github action


wForget:
There is a similar PR #4123. If possible, can we also consider parallelism in compiling velox?

dcoliversun:
@wForget Sure, your pr can solve this issure, I will cose this pr

PR #4165, opened 2023-12-22
[CORE] fix UnsupportedOperationException caused by gluten C2R

zhli1142015:

We encountered UnsupportedOperationException with internal custom plan, the custom plan node can only handle Vanilla spark's columnar data. Now gluten extensions recognize columnar operators (except reader) as gluten plan, this caused wrong VeloxColumnarToRowExec and RowToVeloxColumnar are added around the custom plan node.
I wrote a UT to simulate our query (`DummyFilterColumnar(supportsColumnar = true)`). The issue case we encountered is both parent (Agg) and child (scan) are fallback because other reasons, then VeloxColumnarToRowExec trigered the exceptions. Please see below:

Error:
```
java.lang.UnsupportedOperationException
        at io.glutenproject.columnarbatch.IndicatorVector.hasNull(IndicatorVector.java:82)
        at org.apache.spark.sql.DummyFilterColumnarExec.$anonfun$doExecuteColumnar$1(GlutenSparkSessionExtensionSuite.scala:54)
        at io.glutenproject.execution.VeloxColumnarToRowExec$$anon$1.next(VeloxColumnarToRowExec.scala:111)
        at io.glutenproject.execution.VeloxColumnarToRowExec$$anon$1.next(VeloxColumnarToRowExec.scala:104)
        at scala.collection.TraversableOnce$FlattenOps$$anon$2.hasNext(TraversableOnce.scala:521)
        at io.glutenproject.utils.InvocationFlowProtection.hasNext(Iterators.scala:135)
        at io.glutenproject.utils.IteratorCompleter.hasNext(Iterators.scala:69)
        at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage2.hashAgg_doAggregateWithoutKey_0$(Unknown Source)
        at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage2.processNext(Unknown Source)
        at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
        at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext(WholeStageCodegenExec.scala:760)
        at scala.collection.Iterator.isEmpty(Iterator.scala:387)
        at scala.collection.Iterator.isEmpty$(Iterator.scala:387)
        at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.isEmpty(WholeStageCodegenExec.scala:758)
        at io.glutenproject.execution.RowToVeloxColumnarExec$.toColumnarBatchIterator(RowToVeloxColumnarExec.scala:87)
        at io.glutenproject.execution.RowToVeloxColumnarExec.$anonfun$doExecuteColumnarInternal$1(RowToVeloxColumnarExec.scala:70)
        at org.apache.spark.rdd.RDD.$anonfun$mapPartitions$2(RDD.scala:853)
        at org.apache.spark.rdd.RDD.$anonfun$mapPartitions$2$adapted(RDD.scala:853)
        at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
        at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:364)
        at org.apache.spark.rdd.RDD.iterator(RDD.scala:328)
        at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
        at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:364)
        at org.apache.spark.rdd.RDD.iterator(RDD.scala:328)
        at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
        at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:101)
        at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:53)
        at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:161)
        at org.apache.spark.scheduler.Task.run(Task.scala:139)
        at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:554)
        at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1529)
        at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:557)

```
Issue plan:
 ```
  *(3) HashAggregate(keys=[], functions=[max(id#192L)], output=[max(id)#194L])
   +- VeloxColumnarToRowExec
      +- ShuffleQueryStage 0
         +- ColumnarExchange SinglePartition, ENSURE_REQUIREMENTS, [plan_id=707], [id=#707], [OUTPUT] List(max:LongType), [OUTPUT] List(max:LongType)
            +- RowToVeloxColumnar
               +- *(2) HashAggregate(keys=[], functions=[partial_max(id#192L)], output=[max#200L])
                  +- VeloxColumnarToRowExec
                     +- DummyFilterColumnar
                        +- RowToVeloxColumnar
                           +- *(1) ColumnarToRow
                              +- FileScan parquet spark_catalog.default.a[id#192L] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/var/git/gluten/gluten-ut/spark34/spark-warehouse/a], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
```

Expected plan:
```
   *(2) HashAggregate(keys=[], functions=[max(id#192L)], output=[max(id)#194L])
   +- VeloxColumnarToRowExec
      +- ShuffleQueryStage 0
         +- ColumnarExchange SinglePartition, ENSURE_REQUIREMENTS, [plan_id=693], [id=#693], [OUTPUT] List(max:LongType), [OUTPUT] List(max:LongType)
            +- RowToVeloxColumnar
               +- *(1) HashAggregate(keys=[], functions=[partial_max(id#192L)], output=[max#200L])
                  +- *(1) ColumnarToRow
                     +- DummyFilterColumnar
                        +- FileScan parquet spark_catalog.default.a[id#192L] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/var/git/gluten/gluten-ut/spark34/spark-warehouse/a], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
```
Before gluten only handles correctly for Vanilla vectorized readers, this pr to handle other Vanilla vectorized operators (maybe added by customers). Other fallback cases (fallback avg / scan or not) are covered by new UT also. The plan looks not optimal but it would not fail by now.

```
               +- ^(5) HashAggregateTransformer(keys=[], functions=[partial_max(id#192L)], output=[max#222L])
                  +- ^(5) InputIteratorTransformer
                     +- RowToVeloxColumnar
                        +- *(1) ColumnarToRow
                           +- DummyFilterColumnar
                              +- FileScan parquet spark_catalog.default.a[id#192L] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/var/git/gluten/gluten-ut/spark34/spark-warehouse/a], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
```

```
               +- *(1) HashAggregate(keys=[], functions=[partial_max(id#192L)], output=[max#211L])
                  +- *(1) ColumnarToRow
                     +- DummyFilterColumnar
                        +- RowToColumnar
                           +- VeloxColumnarToRowExec
                              +- ^(4) NativeFileScan parquet spark_catalog.default.a[id#192L] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/var/git/gluten/gluten-ut/spark34/spark-warehouse/a], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
```

```
               +- ^(2) HashAggregateTransformer(keys=[], functions=[partial_max(id#192L)], output=[max#200L])
                  +- ^(2) InputIteratorTransformer
                     +- RowToVeloxColumnar
                        +- *(1) ColumnarToRow
                           +- DummyFilterColumnar
                              +- RowToColumnar
                                 +- VeloxColumnarToRowExec
                                    +- ^(1) NativeFileScan parquet spark_catalog.default.a[id#192L] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/var/git/gluten/gluten-ut/spark34/spark-warehouse/a], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<id:bigint>
```

## How was this patch tested?

New UT.

### CODE REVIEW

**Comment by FelixYBW** [comment]
is the DummyFilterColumnar a specific operator which can take both Vanilla spark's columnar data and Gluten's columnar data?

**Comment by zhli1142015** [comment]
> is the DummyFilterColumnar a specific operator which can take both Vanilla spark's columnar data and Gluten's columnar data?

It can only handle Vanilla spark's columnar data.
The issue here is Gluten extenstion recognize all columnar operator except readers as gluten plan and wrapper it with `VeloxColumnarToRowExec`  and `RowToVeloxColumnar` and also extra `ColumnarToRow`.
image

**Comment by FelixYBW** [comment]
> luten extenstion recognize all columnar operator except readers as gluten plan and wrapper it

got it. thank you

**Comment by zhli1142015** [comment]
@PHILO-HE , @rui-mo , @baibaichen , @zzcclp , @ulysses-you  could you please help to take a look?

File: gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala [hunk 1]
```diff
@@ -615,15 +615,33 @@ case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[Spark
...
       val child = replaceWithTransformerPlan(plan.child)
-      logDebug(s"ColumnarPostOverrides ColumnarToRowExecBase(${child.nodeName})")
-      val nativeConversion =
-        BackendsApiManager.getSparkPlanExecApiInstance.genColumnarToRowExec(child)
-      val validationResult = nativeConversion.doValidate()
-      if (validationResult.isValid) {
-        nativeConversion
-      } else {
-        TransformHints.tagNotTransformable(plan, validationResult)
+      def isChildGlutenPlan(p: SparkPlan): Boolean = {
```
ulysses-you, line 618: Can we pull out `isGlutenPlan` and `isVanillaColumnarPlan` ?

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -615,15 +615,33 @@ case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[Spark
...
+        p match {
+          case a: AQEShuffleReadExec => isChildGlutenPlan(a.child)
+          case s: QueryStageExec => isChildGlutenPlan(s.plan)
+          case s: ReusedExchangeExec => isChildGlutenPlan(s.child)
+          case s: InputAdapter => isChildGlutenPlan(s.child)
+          case s: WholeStageCodegenExec => isChildGlutenPlan(s.child)
+          case s: AdaptiveSparkPlanExec => isChildGlutenPlan(s.executedPlan)
+          case _: GlutenPlan => true
```
ulysses-you, line 626: we should also handle columnar table cache

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -694,42 +712,59 @@ object InMemoryTableScanHelper {
 }
 
 // This rule will try to add RowToColumnarExecBase and ColumnarToRowExec
-// to support vanilla columnar scan.
+// to support vanilla columnar operators.
```
ulysses-you, line 715: I'm afraid the scope is bigger than Gluten: we are trying to be compatible with vanilla columnar operators which is not Spark built-in operators.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhli1142015: Sure.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhli1142015: Thanks for reviewing, can you point me the plan class name?

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhli1142015: This change is for supporting "vanilla columnar operators which is not Spark built-in operators.". I think Gluten should support this or declare this as a limitation. Please let me know if any concern for supporting it.
Now gluten logic wrpper all columnar ops (except readers) with `VeloxColumnarToRowExec` and `VeloxRowToColumnarExec` and caused the exceptions. This PR just fix the behavior to make sure, only gluten plan is wrappered with velox version C2R and R2C. other columnar plans are still wrppered with vanilla version C2R and R2C.

PHILO-HE (commented):
I'm not sure how common such custom operator is used. Could you clarify? cc @zzcclp @zhouyuan

zhli1142015:
> I'm not sure how common such custom operator is used. Could you clarify? cc @zzcclp @zhouyuan

Thanks for reviewing. Here is the background: our partner team added such custom operators for optimization, as I known there is only one custom op. And it failed when running with gluten. We expect there is no perf improvement when integrating with gluten, as gluten doesn't cover it. But it should not cause exceptions to existing logic / mechanism which vanilla spark supports. This is the reason we post this PR.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhli1142015: Updated, pull out plan related util methods to a separate file.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -694,42 +712,59 @@ object InMemoryTableScanHelper {
...
+              BackendsApiManager.getSparkPlanExecApiInstance.genColumnarToRowExec(child))
+          } else {
+            child
+          }
+      })
+    case _ =>
+      plan.withNewChildren(plan.children.map(replaceWithVanillaRowToColumnar))
+  }
+
+  private def isVanillaColumnarOp(plan: SparkPlan): Boolean = plan match {
```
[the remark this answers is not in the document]
zhli1142015: These two should be a little different, renamed 'isChildGlutenPlan' to another name to avoid confusing.

PHILO-HE:
Thanks for your contribution! Maybe, it needs some more discussion after the holiday.

zhztheplayer:
> It can only handle Vanilla spark's columnar data.
The issue here is Gluten extenstion recognize all columnar operator except readers as gluten plan and wrapper it with VeloxColumnarToRowExec and RowToVeloxColumnar and also extra ColumnarToRow.

Gluten provides utility method `ColumnarBatches.ensureLoaded` for converting native columnar batch into Arrow columnar batch which is compatible with Vanilla Spark. Would that help on solving your issue?

https://github.com/oap-project/gluten/blob/fcb31fc6b09ece3be76131caaba81681a3b7b53c/gluten-data/src/main/java/io/glutenproject/columnarbatch/ColumnarBatches.java#L160-L165

If we wanted to completely solve this kind of issue, I am inclined to develop a good mechanism to mark a plan with its underlying type, probably using Spark's tag? Then we can add corresponding transitions including c2r, r2c, or c2c afterwards. However I am not sure whether Spark's columnar API is powerful enough for us to implement that.

zhli1142015:
> +- *(2) HashAggregate(keys=[], functions=[partial_max(id#192L)], output=[max#200L])
>                  +- VeloxColumnarToRowExec
>                     +- DummyFilterColumnar
>                        +- RowToVeloxColumnar
>                           +- *(1) ColumnarToRow
>                              +- FileScan parquet

Thanks fo your comments, may be we can split this to two issues.
the first one, we should wrapper correct C2R & R2c to gluten plan and non-gluten plan.
the second one, use VanillaC2VeloxC to replace such conversion [`BackendsApiManager.getSparkPlanExecApiInstance.genRowToColumnarExec(ColumnarToRowExec(plan))`](https://github.com/oap-project/gluten/blob/main/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala#L707), and maybe VeloxC2VanillaC.

How about we solve the first issue in this PR, then track second one by another issue?

zhli1142015:
> > +- *(2) HashAggregate(keys=[], functions=[partial_max(id#192L)], output=[max#200L])
> > +- VeloxColumnarToRowExec
> > +- DummyFilterColumnar
> > +- RowToVeloxColumnar
> > +- *(1) ColumnarToRow
> > +- FileScan parquet
> 
> Thanks fo your comments, may be we can split this to two issues. the first one, we should wrapper correct C2R & R2c to gluten plan and non-gluten plan. the second one, use VanillaC2VeloxC to replace such conversion [`BackendsApiManager.getSparkPlanExecApiInstance.genRowToColumnarExec(ColumnarToRowExec(plan))`](https://github.com/oap-project/gluten/blob/main/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala#L707), and maybe VeloxC2VanillaC.
> 
> How about we solve the first issue in this PR, then track second one by another issue?

https://github.com/oap-project/gluten/issues/4241

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -674,62 +679,43 @@ case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[Spark
...
   private def replaceWithVanillaColumnarToRow(plan: SparkPlan): SparkPlan = plan match {
-    case c2r: ColumnarToRowExecBase if isVanillaColumnarReader(c2r.child) =>
+    case c2r: ColumnarToRowExecBase if PlanUtil.isVanillaColumnarOp(c2r.child) =>
       ColumnarToRowExec(c2r.child)
-    case c2r: ColumnarToRowExec if isVanillaColumnarReader(c2r.child) =>
+    case c2r: ColumnarToRowExec if PlanUtil.isVanillaColumnarOp(c2r.child) =>
       c2r
-    case _ if isVanillaColumnarReader(plan) =>
+    case _ if PlanUtil.isVanillaColumnarOp(plan) =>
       BackendsApiManager.getSparkPlanExecApiInstance.genRowToColumnarExec(ColumnarToRowExec(plan))
```
zhztheplayer, line 693: How does the code handle consecutive vanilla columnar plan nodes?

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -674,62 +679,43 @@ case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[Spark
...
-      } else {
-        !plan.isInstanceOf[GlutenPlan] && plan.supportsColumnar
-      }
-    case _ => false
+  private def replaceWithVanillaRowToColumnar(plan: SparkPlan): SparkPlan = plan match {
+    case _ if PlanUtil.isVanillaColumnarOp(plan) =>
+      plan.withNewChildren(plan.children.map {
+        c =>
+          val child = replaceWithVanillaRowToColumnar(c)
+          if (child.isInstanceOf[GlutenPlan]) {
```
zhztheplayer, line 703: If there is `PlanUtil.isVanillaColumnarOp`, should we add `PlanUtil.isGlutenColumnarOp`?

PHILO-HE:
Hi @zhli1142015, can `FilterColumnar` be replaced by Gluten columnar filter? In your PR description, you mentioned both parent (Agg) and child (scan) fall back in your case for some reasons. I'm assuming, with Gluten's development, such fallback will be removed at some time point. So always keeping vanilla op `FilterColumnar` used may cause some perf. regression in the future due to the data transition cost.

zhli1142015:
> Hi @zhli1142015, can `FilterColumnar` be replaced by Gluten columnar filter?

I think this can not be replaced by now velox filter. cc @felipepessoto to clarify more.

> In your PR description, you mentioned both parent (Agg) and child (scan) fall back in your case for some reasons. I'm assuming, with Gluten's development, such fallback will be removed at some time point. So always keeping vanilla op `FilterColumnar` used may cause some perf. regression in the future due to the data transition cost.

Yes, agree on this. I think we can either support C2C as @zhztheplayer suggested or let gluten support to register custom velox operator, then we can implement it by velox data format. I think velox has such [mechanism](https://github.com/facebookincubator/velox/blob/main/velox/exec/Operator.h#L258C4-L259C15). Maybe another follow up~

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -674,62 +679,48 @@ case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[Spark
...
-  def apply(plan: SparkPlan): SparkPlan =
-    if (GlutenConfig.getConf.enableVanillaVectorizedReaders) {
-      val newPlan = replaceWithVanillaColumnarToRow(plan)
-      planChangeLogger.logRule(ruleName, plan, newPlan)
-      newPlan
-    } else {
-      plan
-    }
+  def apply(plan: SparkPlan): SparkPlan = {
+    val newPlan = replaceWithVanillaRowToColumnar(replaceWithVanillaColumnarToRow(plan))
```
ulysses-you, line 720: Can we add config `enableVanillaVectorizedReaders` back ?

gluten-core/src/main/scala/io/glutenproject/utils/PlanUtil.scala
```diff
@@ -0,0 +1,82 @@
...
+    plan match {
+      case i: InMemoryTableScanExec =>
+        if (PlanUtil.isGlutenTableCache(i)) {
+          // `InMemoryTableScanExec` do not need extra RowToColumnar or ColumnarToRow
+          false
+        } else {
+          !plan.isInstanceOf[GlutenPlan] && plan.supportsColumnar
+        }
+      case _: AQEShuffleReadExec => false
+      case _: QueryStageExec => false
```
ulysses-you, line 68: What happens if there is a columnar shuffle exchange based on the vanilla Spark columnar framework ? Does your case only happen on the first stage ?

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhli1142015: I think here this config check is not needed. If this conf is false, then scan ops' `supportsColumnar` should be false, then the logic here would not change the plan. Please let me know if anything I missed.

gluten-core/src/main/scala/io/glutenproject/utils/PlanUtil.scala
zhli1142015: yes, my case only happens in first stage, the op is more like a filter.
This function is only used to handle below cases:
gluten op --> vanilla columnar op  => gluten op --> R2VeloxC --> C2R --> vanilla columnar op
or 
vanilla columnar op --> gluten op => vanilla columnar op --> R2C --> VeloxC2R --> gluten op

So there is columnar shuffle exchange based on the vanilla Spark columnar framework, I think it would add  C2R and R2C.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
ulysses-you: make sense

ulysses-you:
@zhli1142015 it seems the failed tests are related, can you check it ?

felipepessoto:
> > Hi @zhli1142015, can `FilterColumnar` be replaced by Gluten columnar filter?
> 
> I think this can not be replaced by now velox filter. cc @felipepessoto to clarify more.
> 
> > In your PR description, you mentioned both parent (Agg) and child (scan) fall back in your case for some reasons. I'm assuming, with Gluten's development, such fallback will be removed at some time point. So always keeping vanilla op `FilterColumnar` used may cause some perf. regression in the future due to the data transition cost.
> 
> Yes, agree on this. I think we can either support C2C as @zhztheplayer suggested or let gluten support to register custom velox operator, then we can implement it by velox data format. I think velox has such [mechanism](https://github.com/facebookincubator/velox/blob/main/velox/exec/Operator.h#L258C4-L259C15). Maybe another follow up~

I'm not familiar with Gluten columnar filters. Do we have any doc or example?

If it isn't implemented in Scala, I'd say no. If it is Scala and straightforward to port, I think we would need to consider pros and cons for each case. But it is not ideal as we already have two implementations for our internal feature, row-based, columnar, and adding another one is not desirable.

Support to Spark Columnar filter probably would help with Delta Deletion Vector support in the future as well.

`+` @sezruby

gluten-core/src/main/scala/io/glutenproject/utils/PlanUtil.scala
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+          !plan.isInstanceOf[GlutenPlan] && plan.supportsColumnar
+        }
+      case a: AQEShuffleReadExec => isVanillaColumnarOp(a.child)
+      case s: QueryStageExec => isVanillaColumnarOp(s.plan)
+      case _: RowToColumnarExec => false
+      case _: InputAdapter => false
+      case _: WholeStageCodegenExec => false
+      case r: ReusedExchangeExec => isVanillaColumnarOp(r.child)
+      case p if p.getClass.getSimpleName == "VeloxColumnarWriteFilesExec" => false
```
ulysses-you, line 73: it is a known issue and has been fixed, can you rebase and remove this pattern ? thank you

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -674,62 +679,52 @@ case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[Spark
...
 case class VanillaColumnarPlanOverrides(session: SparkSession) extends Rule[SparkPlan] {
   @transient private val planChangeLogger = new PlanChangeLogger[SparkPlan]()
 
   private def replaceWithVanillaColumnarToRow(plan: SparkPlan): SparkPlan = plan match {
-    case c2r: ColumnarToRowExecBase if isVanillaColumnarReader(c2r.child) =>
+    case c2r: ColumnarToRowExecBase if PlanUtil.isVanillaColumnarOp(c2r.child) =>
       ColumnarToRowExec(c2r.child)
-    case c2r: ColumnarToRowExec if isVanillaColumnarReader(c2r.child) =>
+    case c2r: ColumnarToRowExec if PlanUtil.isVanillaColumnarOp(c2r.child) =>
       c2r
```
zhztheplayer, line 691: Do we need to handle c2r's children recursively? Probably we can use `SparkPlan`'s `transformUp` API to avoid handwritten recursions.

PHILO-HE:
> I'm not familiar with Gluten columnar filters. Do we have any doc or example?
> 
> If it isn't implemented in Scala, I'd say no. If it is Scala and straightforward to port, I think we would need to consider pros and cons for each case. But it is not ideal as we already have two implementations for our internal feature, row-based, columnar, and adding another one is not desirable.

Hi @felipepessoto, we have a scala class (see [here](https://github.com/oap-project/gluten/blob/8797350233304cfaef570a7fef83e6b1965ce609/backends-velox/src/main/scala/io/glutenproject/execution/FilterExecTransformer.scala)) to represent gluten columnar filter and it can replace row based filter plan of vanilla spark currently. The underneath execution is on native side and it only supports gluten columnar data (i.e., velox data format for velox backend). If your `FilterColumnar` can be replaced by gluten columnar filter, we may instantiate a gluten filter on gluten side and then do the replacement. I don't know the details of `FilterColumnar`, so not sure whether my thought is feasible actually.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhli1142015: c2r cases can be removed, as they are used to prevent c2r is converted in original below case match. Now we don't need them.
Here we need to look two level plans: parent is gluten op and child is vanilla columnar op. I think transformUp may not be suitable.

zhztheplayer: Sounds good to me. Thanks for the explanation.

zhztheplayer (approved):
Thanks!

ulysses-you (approved):

Issue #3942, opened 2023-12-06
RecordBatch field 0 should be integer when using HashPartitioning


felipepessoto:
### Backend

VL (Velox)

### Bug description

When I use ColumnarShuffleExchangeExec and ShuffleExchangeExec with HashPartitioning I receive the error below. I tried to isolate the issue and created this repro code. It only fails when partitioning is HashPartitioning. Single and RoundRobin works fines.

I created it based on examples I found in the Gluten source, like https://github.com/oap-project/gluten/blob/81bb6c9b0652ec4df39e6f50c0405756b59d5a3d/gluten-core/src/main/scala/io/glutenproject/execution/TakeOrderedAndProjectExecTransformer.scala#L103

Repro code:

```
val sourcePath = "/tmp/test/source"
val outputPath = "/tmp/test/target"
val pColumn = "colC"
val isTablePartitioned = true
val targetFileSize = 134217728
val numShufflePartitions = spark.sessionState.conf.numShufflePartitions

val dfSource = spark
.range(5000)
.map { _ =>
  (10L,
  11,
  scala.util.Random.nextInt(2))
}
.repartition(100)
.toDF("colA", "colB", "colC")

dfSource.write.partitionBy(pColumn).format("parquet").mode("overwrite").save(sourcePath)

val df = spark.read.format("parquet").load(sourcePath)

val physicalPlan = df.queryExecution.executedPlan

println("\n\n physicalPlan: " + physicalPlan)

val originalPlan = physicalPlan

//Remove top C2R
val planWithoutC2R = physicalPlan.children(0)

val partitionColumnsExpr = Array(planWithoutC2R.output.find(c => c.name.equals(pColumn)).get)
val partitioning = HashPartitioning(partitionColumnsExpr, numShufflePartitions)
// These two works
// SinglePartition
// RoundRobinPartitioning(numShufflePartitions)

val shuffleExec = ShuffleExchangeExec(partitioning, planWithoutC2R)
val transformedShuffleExec =
      ColumnarShuffleExchangeExec(shuffleExec, planWithoutC2R, shuffleExec.child.output)

// TransformHints.tag(shuffleExec, transformedShuffleExec.doValidate().toTransformHint)

// Add C2R back
val planWithC2R = VeloxColumnarToRowExec(transformedShuffleExec)

val planToWrite = planWithC2R

println("\n\n planToWrite: " + planToWrite)
println("X: " + planToWrite.execute().first().getLong(0))
```

### Spark version

[...]

PR #4167, opened 2023-12-22
[GLUTEN-3942][CORE]fix: Change columnar overrides to accept ShuffleExchangeLike

felipepessoto:

TransformPreOverrides.replaceWithTransformerPlan only tranform ShuffleExchangeExec, which is a case class, meaning I can't extend it.

This PR changes it to support ShuffleExchangeLike.

Partially fix the #3942

## How was this patch tested?

Existing tests


marin-ma (approved):

marin-ma:
Thanks for this fix!

PR #4175, opened 2023-12-25
[VL] Daily Update Velox Version (2023-12-25)

zhztheplayer:
```
c5bcc33e9 (upstream/main) Improve the capacity management in memory arbitrator (#8110)
a1e5c346d Add support for multiple S3FileSystems (#7388)
50187434e Add write file create config to hive config (#8133)
9d7d4e8a5 Allow marking an argument as constant in the simple function API (#7901)
8948b9e94 Add TypeSignature Parser using Flex and Bison (#7590)
b93dd31c9 Add Status class (#8084)
e87462633 Register abfs filesystem during Hive Connector registration (#8144)
32e21c9f8 Fix reading MAP_KEY_VALUE Parquet SchemaElement (#7966)
```

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -17,7 +17,7 @@
 set -exu
 
 VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=update
+VELOX_BRANCH=2023_12_25
 VELOX_HOME=""
 
 #Set on run gluten on HDFS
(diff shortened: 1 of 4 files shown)


JkSelf (approved):
LGTM. Thanks.

ep/build-velox/src/get_velox.sh
```diff
@@ -17,7 +17,7 @@
 set -exu
 
 VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=update
+VELOX_BRANCH=2023_12_25
```
zhztheplayer, line 20: I'll evaluate if using a rolling-updated branch name here could be better than the previous way in the next few days.

PR #4177, opened 2023-12-25
[GLUTEN-4170][VL] Decouple partitions from plan to avoid driver stalled

Yohahaha:

There are two parts will lead driver stalled when scan contains lots of partitions,
1. plan serialization happens in every GlutenPartition construction.
2. GlutenWholeStageColumnarRDD#getPartitions

| |  | 22374 partitions | 44611 partitions |
|--------|--------|--------|--------|
| before | #1 | 880ms | 1352ms |
| | #2 | 3662ms | 17186ms |
| after | #1 | 21ms | 106ms |
|  | #2 | 6ms | 25ms | 

This patch decouple scan splitInfo(LocalFileNodes) from ReadRel to avoid serialize substrait plan for each partition in Driver, when the plan is complex or the number of partitions is particularly large, the cost of this serialization cannot be ignored.

Stream splitInfo(inputIterator) still kept in ReadRel for now.

(Fixes: \#4170)

### DIFF
--- a/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
@@ -47,11 +47,6 @@ import scala.collection.JavaConverters._
 
 class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
 
-  /**
-   * Generate native row partition.
-   *
-   * @return
-   */
   override def genSplitInfo(
       partition: InputPartition,
       partitionSchema: StructType,
@@ -92,6 +87,24 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
           preferredLocations.toList.asJava)
       case _ =>
         throw new UnsupportedOperationException(s"Unsupported input partition.")
+    }
+  }
+
+  /**
+   * Generate native row partition.
+   *
+   * @return
+   */
+  override def genPartitions(wsCtx: WholeStageTransformContext, splitInfos: Seq[Seq[SplitInfo]]): Seq[BaseGlutenPartition] = {
+    splitInfos.zipWithIndex.map {
+      case (splitInfos, index) =>
+        wsCtx.substraitContext.initSplitInfosIndex(0)
+        wsCtx.substraitContext.setSplitInfos(splitInfos)
+        val substraitPlan = wsCtx.root.toProtobuf
+        GlutenPartition(
+          index,
+          substraitPlan.toByteArray,
+          locations = splitInfos.flatMap(_.preferredLocations().asScala).toArray)
     }
   }
 
--- a/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
@@ -95,7 +95,9 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
    *
    * @return
    */
-  override def genPartitions(wsCtx: WholeStageTransformContext, splitInfos: Seq[Seq[SplitInfo]]): Seq[BaseGlutenPartition] = {
+  override def genPartitions(
+      wsCtx: WholeStageTransformContext,
+      splitInfos: Seq[Seq[SplitInfo]]): Seq[BaseGlutenPartition] = {
     splitInfos.zipWithIndex.map {
       case (splitInfos, index) =>
         wsCtx.substraitContext.initSplitInfosIndex(0)
@@ -265,7 +267,7 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
           GlutenPartition(
             index,
             substraitPlan.toByteArray,
-            splitInfo.preferredLocations().asScala.toArray)
+            locations = splitInfo.preferredLocations().asScala.toArray)
       }
     }(t => logInfo(s"Generating the Substrait plan took: $t ms."))
 
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -202,8 +202,10 @@ class IteratorApiImpl extends IteratorApi with Logging {
     val nativeResultIterator =
       transKernel.createKernelWithBatchIterator(
         rootNode.toProtobuf.toByteArray,
+        // Final iterator does not contain scan split, so pass empty split info to native here.
         new Array[Array[Byte]](0),
-        columnarNativeIterator)
+        columnarNativeIterator
+      )
 
     Iterators
       .wrap(nativeResultIterator.asScala)
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
@@ -21,7 +21,7 @@ import io.glutenproject.backendsapi.IteratorApi
 import io.glutenproject.execution._
 import io.glutenproject.metrics.IMetrics
 import io.glutenproject.substrait.plan.PlanNode
-import io.glutenproject.substrait.rel.{LocalFilesBuilder, SplitInfo}
+import io.glutenproject.substrait.rel.{LocalFilesBuilder, LocalFilesNode, SplitInfo}
 import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 import io.glutenproject.utils.Iterators
 import io.glutenproject.vectorized._
@@ -47,16 +47,12 @@ import java.net.URLDecoder
 import java.nio.charset.StandardCharsets
 import java.time.ZoneOffset
 import java.util.{ArrayList => JArrayList, HashMap => JHashMap, Map => JMap}
+import java.util.concurrent.TimeUnit
 
 import scala.collection.JavaConverters._
 
 class IteratorApiImpl extends IteratorApi with Logging {
 
-  /**
-   * Generate native row partition.
-   *
-   * @return
-   */
   override def genSplitInfo(
       partition: InputPartition,
       partitionSchema: StructType,
@@ -77,6 +73,24 @@ class IteratorApiImpl extends IteratorApi with Logging {
           preferredLocations.toList.asJava)
       case _ =>
         throw new UnsupportedOperationException(s"Unsupported input partition.")
+    }
+  }
+
+  /** Generate native row partition. */
+  override def genPartitions(
+      wsCtx: WholeStageTransformContext,
+      splitInfos: Seq[Seq[SplitInfo]]): Seq[BaseGlutenPartition] = {
+    // Only serialize plan once, save lots time when plan is complex.
+    val planByteArray = wsCtx.root.toProtobuf.toByteArray
+
+    splitInfos.zipWithIndex.map {
+      case (splitInfos, index) =>
+        GlutenPartition(
+          index,
+          planByteArray,
+          splitInfos.map(_.asInstanceOf[LocalFilesNode].toProtobuf.toByteArray).toArray,
+          splitInfos.flatMap(_.preferredLocations().asScala).toArray
+        )
     }
   }
 
@@ -121,11 +135,7 @@ class IteratorApiImpl extends IteratorApi with Logging {
     transKernel.injectWriteFilesTempPath(path)
   }
 
-  /**
-   * Generate Iterator[ColumnarBatch] for first stage.
-   *
-   * @return
-   */
+  /** Generate Iterator[ColumnarBatch] for first stage. */
   override def genFirstStageIterator(
       inputPartition: BaseGlutenPartition,
       context: TaskContext,
@@ -133,13 +143,26 @@ class IteratorApiImpl extends IteratorApi with Logging {
       updateInputMetrics: (InputMetricsWrapper) => Unit,
       updateNativeMetrics: IMetrics => Unit,
       inputIterators: Seq[Iterator[ColumnarBatch]] = Seq()): Iterator[ColumnarBatch] = {
+    assert(
+      inputPartition.isInstanceOf[GlutenPartition],
+      "Velox backend only accept GlutenPartition.")
+
+    val beforeBuild = System.nanoTime()
     val columnarNativeIterators =
       new JArrayList[GeneralInIterator](inputIterators.map {
         iter => new ColumnarBatchInIterator(iter.asJava)
       }.asJava)
     val transKernel = NativePlanEvaluator.create()
+
+    val splitInfoByteArray = inputPartition
+      .asInstanceOf[GlutenPartition]
+      .splitInfosByteArray
     val resIter: GeneralOutIterator =
-      transKernel.createKernelWithBatchIterator(inputPartition.plan, columnarNativeIterators)
+      transKernel.createKernelWithBatchIterator(
+        inputPartition.plan,
+        splitInfoByteArray,
+        columnarNativeIterators)
+    pipelineTime += TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - beforeBuild)
 
     Iterators
       .wrap(resIter.asScala)
@@ -157,11 +180,7 @@ class IteratorApiImpl extends IteratorApi with Logging {
 
   // scalastyle:off argcount
 
-  /**
-   * Generate Iterator[ColumnarBatch] for final stage.
-   *
-   * @return
-   */
+  /** Generate Iterator[ColumnarBatch] for final stage. */
   override def genFinalStageIterator(
       context: TaskContext,
       inputIterators: Seq[Iterator[ColumnarBatch]],
@@ -183,6 +202,7 @@ class IteratorApiImpl extends IteratorApi with Logging {
     val nativeResultIterator =
       transKernel.createKernelWithBatchIterator(
         rootNode.toProtobuf.toByteArray,
+        new Array[Array[Byte]](0),
         columnarNativeIterator)
 
     Iterators
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -382,9 +382,6 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithI
 
   auto spillDirStr = jStringToCString(env, spillDir);
 
-  auto planData = reinterpret_cast<const uint8_t*>(env->GetByteArrayElements(planArr, nullptr));
-  auto planSize = env->GetArrayLength(planArr);
-
   if (jsize splitInfoArraySize = env->GetArrayLength(splitInfosArr); splitInfoArraySize != 0) {
     for (auto i = 0; i < splitInfoArraySize; i++) {
       jbyteArray splitInfoArray = reinterpret_cast<jbyteArray>(env->GetObjectArrayElement(splitInfosArr, i));
@@ -393,6 +390,9 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithI
       ctx->parseSplitInfo(splitInfoData, splitInfoSize);
     }
   }
+
+  auto planData = reinterpret_cast<const uint8_t*>(env->GetByteArrayElements(planArr, nullptr));
+  auto planSize = env->GetArrayLength(planArr);
   ctx->parsePlan(planData, planSize, {stageId, partitionId, taskId});
 
   auto& conf = ctx->getConfMap();
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -368,6 +368,7 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithI
     jobject wrapper,
     jlong memoryManagerHandle,
     jbyteArray planArr,
+    jobjectArray splitInfosArr,
     jobjectArray iterArr,
     jint stageId,
     jint partitionId,
@@ -384,7 +385,16 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithI
   auto planData = reinterpret_cast<const uint8_t*>(env->GetByteArrayElements(planArr, nullptr));
   auto planSize = env->GetArrayLength(planArr);
 
+  if (jsize splitInfoArraySize = env->GetArrayLength(splitInfosArr); splitInfoArraySize != 0) {
+    for (auto i = 0; i < splitInfoArraySize; i++) {
+      jbyteArray splitInfoArray = reinterpret_cast<jbyteArray>(env->GetObjectArrayElement(splitInfosArr, i));
+      jsize splitInfoSize = env->GetArrayLength(splitInfoArray);
+      auto splitInfoData = reinterpret_cast<const uint8_t*>(env->GetByteArrayElements(splitInfoArray, nullptr));
+      ctx->parseSplitInfo(splitInfoData, splitInfoSize);
+    }
+  }
   ctx->parsePlan(planData, planSize, {stageId, partitionId, taskId});
+
   auto& conf = ctx->getConfMap();
 
   // Handle the Java iters
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -382,13 +382,11 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithI
 
   auto spillDirStr = jStringToCString(env, spillDir);
 
-  if (jsize splitInfoArraySize = env->GetArrayLength(splitInfosArr); splitInfoArraySize != 0) {
-    for (auto i = 0; i < splitInfoArraySize; i++) {
-      jbyteArray splitInfoArray = reinterpret_cast<jbyteArray>(env->GetObjectArrayElement(splitInfosArr, i));
-      jsize splitInfoSize = env->GetArrayLength(splitInfoArray);
-      auto splitInfoData = reinterpret_cast<const uint8_t*>(env->GetByteArrayElements(splitInfoArray, nullptr));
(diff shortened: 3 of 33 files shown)


Yohahaha:
@FelixYBW @rui-mo @PHILO-HE @ulysses-you 
let's discuss the solution based on current patch, the key is decouple scan partition from plan when serialization. I know this patch may seems tricky, and open for more better design.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -159,16 +159,18 @@ message ReadRel {
...
-     repeated partitionColumn partition_columns = 16;
+      }
+      repeated partitionColumn partition_columns = 16;
+
+      /// File schema
+      NamedStruct schema = 17;
 
-     /// File schema
-     NamedStruct schema = 17;
+      bool stream_input = 18;
```
ulysses-you, line 173: is it necessary ? we check if it is a iterator input by comparing the prefix string is equal to iterator

gluten-core/src/main/java/io/glutenproject/substrait/rel/LocalFilesNode.java
```diff
@@ -50,7 +50,7 @@ public enum ReadFileFormat {
   }
 
   private ReadFileFormat fileFormat = ReadFileFormat.UnknownFormat;
-  private Boolean iterAsInput = false;
+  private Boolean streamInput = false;
```
ulysses-you, line 53: TBH, the original variable name seems easy to follow. streamInput can be used when Gluten support Flink.

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -33,6 +34,8 @@ struct SplitInfo {
   /// Whether the split comes from arrow array stream node.
   bool isStream = false;
 
+  int32_t idx{-1};
```
ulysses-you, line 37: `iteratorIndex`

ulysses-you (commented):
thank you @Yohahaha for the idea, looks reasonable to me. One concern is about the name, can we avoid using `streamXxx` ? I think `iteratorXxx` is easy to follow. We are using iterator in both java and native side.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
Yohahaha: it's not necessary, I add it before for debugging purposes, already removed.

gluten-core/src/main/java/io/glutenproject/substrait/rel/LocalFilesNode.java
Yohahaha: fixed.

cpp/velox/substrait/SubstraitToVeloxPlan.h
Yohahaha: removed.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
```diff
@@ -178,6 +210,7 @@ class IteratorApiImpl extends IteratorApi with Logging {
     val nativeResultIterator =
       transKernel.createKernelWithBatchIterator(
         rootNode.toProtobuf.toByteArray,
+        new Array[Array[Byte]](0),
```
rui-mo, line 213: Could you explain a bit what happens here, and is split info needed for final iterator?

gluten-data/src/main/java/io/glutenproject/vectorized/NativePlanEvaluator.java
```diff
@@ -61,7 +61,8 @@ public NativePlanValidationInfo doNativeValidateWithFailureReason(byte[] subPlan
   // Used by WholeStageTransform to create the native computing pipeline and
   // return a columnar result iterator.
   public GeneralOutIterator createKernelWithBatchIterator(
-      byte[] wsPlan, List<GeneralInIterator> iterList) throws RuntimeException, IOException {
+      byte[] wsPlan, byte[][] splitInfo, List<GeneralInIterator> iterList)
+      throws RuntimeException, IOException {
```
rui-mo, line 65: Could we ensure `splitInfo` is not empty?

rui-mo (commented):
Thanks for the improvement.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
Yohahaha: SplitInfo only exists when stage contain scan, so the final iterator does not need it I think.

`new Array[Array[Byte]](0)` is an empty byte array which represents empty split infos, I know it's looks wired, maybe use a new proto class for array of split infos would be better, but needs change lots codes of substrait conversion. How do you think about that?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
```diff
@@ -128,13 +145,28 @@ class IteratorApiImpl extends IteratorApi with Logging {
...
     val columnarNativeIterators =
       new JArrayList[GeneralInIterator](inputIterators.map {
         iter => new ColumnarBatchInIterator(iter.asJava)
       }.asJava)
     val transKernel = NativePlanEvaluator.create()
+
+    val splitInfoByteArray = inputPartition
+      .asInstanceOf[GlutenPartition]
+      .splitInfos
+      .map(_.asInstanceOf[LocalFilesNode].toProtobuf.toByteArray)
```
ulysses-you, line 162: shall we call LocalFilesNode to byte array at driver side ?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -809,21 +844,20 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
...
+    return constructValueStreamNode(readRel, streamIdx);
   }
 
   // Otherwise, will create TableScan node for ReadRel.
+  auto splitInfo = std::make_shared<SplitInfo>();
+  if (!validationMode_) {
+    VELOX_CHECK_LT(splitInfoIdx_, splitInfos_.size(), "Plan must have readRel and related split info.");
+    splitInfo = splitInfos_[splitInfoIdx_++];
+  }
+  fileFormat_ = splitInfo->format;
```
ulysses-you, line 859: it seems `fileFormat_` is unused

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -809,21 +844,20 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
         !readRel.common().has_emit(), "Emit not supported for ValuesNode and TableScanNode related Substrait plans.");
   }
 
-  // Check if the ReadRel specifies an input of stream. If yes, the pre-built
-  // input node will be used as the data source.
-  auto splitInfo = std::make_shared<SplitInfo>();
+  // Check if the ReadRel specifies an input of stream. If yes, build ValueStreamNode as the data source.
   auto streamIdx = getStreamIndex(readRel);
```
rui-mo, line 848: `getStreamIndex` relies on local files to extract the iterator index. Could you help confirm?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
rui-mo: I'm wondering if split info is also needed for final iterator if ReadRel does not contain LocalFiles. See https://github.com/oap-project/gluten/pull/4177#discussion_r1439213063.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
Yohahaha: yes.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
Yohahaha: scan split info is not needed for final iterator, but stream split info is needed.

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
```diff
@@ -95,6 +90,24 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
...
+   * Generate native row partition.
+   *
+   * @return
+   */
+  override def genPartitions(wsCtx: WholeStageTransformContext, splitInfos: Seq[Seq[SplitInfo]]): Seq[BaseGlutenPartition] = {
+    splitInfos.zipWithIndex.map {
+      case (splitInfos, index) =>
+        wsCtx.substraitContext.initSplitInfosIndex(0)
+        wsCtx.substraitContext.setSplitInfos(splitInfos)
+        val substraitPlan = wsCtx.root.toProtobuf
```
PHILO-HE, line 103: I'm assuming the proposed optimization can also be applied for CH backend. If so, it will need some follow-up work from CH engineer. @baibaichen

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
Yohahaha: good point, I will check later.

Yohahaha:
I assume this solution looks well for all your guys.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
rui-mo: Does this PR cover this change or some extra change is needed?

Yohahaha: > Does this PR cover this change or some extra change is needed?

this patch does change stream split info's behavior, it will be extracted from ReadRel as before.

zhouyuan:
@lgbo-ustc hi, this patch tries to refactor on gen file partitions, please take a look if it will impact CK backend

thanks, -yuan

cpp/velox/substrait/SubstraitToVeloxPlan.cc
Yohahaha: yes.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
```diff
@@ -80,6 +76,27 @@ class IteratorApiImpl extends IteratorApi with Logging {
     }
   }
 
+  /**
+   * Generate native row partition.
+   *
+   * @return
```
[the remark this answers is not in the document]
Yohahaha: removed.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -159,16 +159,16 @@ message ReadRel {
...
         string value = 2;
-     }
-     repeated partitionColumn partition_columns = 16;
+      }
+      repeated partitionColumn partition_columns = 16;
 
-     /// File schema
-     NamedStruct schema = 17;
+      /// File schema
+      NamedStruct schema = 17;
```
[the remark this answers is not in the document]
Yohahaha: it's needed, there should be 6 leading space.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1157,10 +1149,7 @@ int32_t SubstraitToVeloxPlanConverter::getStreamIndex(const ::substrait::ReadRel
       VELOX_FAIL(err.what());
     }
   }
-  if (validationMode_) {
-    return -1;
-  }
-  VELOX_FAIL("Local file is expected.");
```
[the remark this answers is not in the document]
Yohahaha: ReadRel may not have local file node when it's a real scan node.

Yohahaha:
I see QueryBenchmark is similar with GenericBenchmark and QueryBenchmark does not covered by CI, could we remove it? @marin-ma @jinchengchenghh

marin-ma:
> I see QueryBenchmark is similar with GenericBenchmark and QueryBenchmark does not covered by CI, could we remove it? @marin-ma @jinchengchenghh

cc: @rui-mo

rui-mo:
> I see QueryBenchmark is similar with GenericBenchmark and QueryBenchmark does not covered by CI, could we remove it?

@Yohahaha GenericBenchmark uses arrow to read files, while QueryBenchmark uses Velox. So QueryBenchmark is useful when we want to test Velox TableScan. I think the better option is to enable QueryBenchmark on CI. @marin-ma Please help to confirm, thanks.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
rui-mo: > Final iterator does not need split infos, and an empty byte array is used to represent empty split infos.

Maybe better to add some comments as above, thanks.

cpp/velox/benchmarks/GenericBenchmark.cc
```diff
@@ -126,7 +127,9 @@ auto BM_Generic = [](::benchmark::State& state,
   auto memoryManager = getDefaultMemoryManager();
   auto runtime = Runtime::create(kVeloxRuntimeKind, conf);
   const auto& filePath = substraitJsonFile;
-  auto plan = getPlanFromFile(filePath);
+  auto plan = getPlanFromFile("Plan", filePath);
+  const auto& splitPath = splitFile;
```
rui-mo, line 131: Why do we need this variable, can we use splitFile instead?

cpp/velox/compute/VeloxPlanConverter.h
```diff
@@ -35,45 +35,19 @@ class VeloxPlanConverter {
       const std::unordered_map<std::string, std::string>& confMap,
       bool validationMode = false);
 
-  std::shared_ptr<const facebook::velox::core::PlanNode> toVeloxPlan(::substrait::Plan& substraitPlan);
+  std::shared_ptr<const facebook::velox::core::PlanNode> toVeloxPlan(
+      ::substrait::Plan& substraitPlan,
```
rui-mo, line 39: Not your change. Maybe better to add const here.

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -137,6 +140,15 @@ class SubstraitToVeloxPlanConverter {
     planNodeId_ = planNodeId;
   }
 
+  void setSplitInfos(std::vector<std::shared_ptr<SplitInfo>> splitInfos) {
+    splitInfoIdx_ = 0;
```
rui-mo, line 144: > int32_t splitInfoIdx_{0};

Why do we need to set split info index as 0? Its initial value should already be 0.

marin-ma:
> > I see QueryBenchmark is similar with GenericBenchmark and QueryBenchmark does not covered by CI, could we remove it?
> 
> @Yohahaha GenericBenchmark uses arrow to read files, while QueryBenchmark uses Velox. So QueryBenchmark is useful when we want to test Velox TableScan. I think the better option is to enable QueryBenchmark on CI. @marin-ma Please help to confirm, thanks.

@rui-mo If input is from middle stage, GenericBenchmark will use arrow reader to load the input iterator. If input is from first stage, the whole pipeline is offloaded including table scan. Here's the doc https://github.com/oap-project/gluten/blob/main/docs/developers/MicroBenchmarks.md#generate-substrait-plan-and-input-for-any-query

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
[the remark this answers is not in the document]
rui-mo: Is it about the format? How do you meet these errors?

rui-mo:
@marin-ma Thanks for confirming. @Yohahaha We can remove QueryBenchmark because its functionality is covered by `GenericBenchmark`.

FelixYBW:
> @marin-ma Thanks for confirming. @Yohahaha We can remove QueryBenchmark because its functionality is covered by `GenericBenchmark`.

Agree

cpp/velox/benchmarks/GenericBenchmark.cc
Yohahaha: just follow `filePath`, will remove.

cpp/velox/compute/VeloxPlanConverter.h
Yohahaha: ok.

cpp/velox/substrait/SubstraitToVeloxPlan.h
Yohahaha: make sure each call to `setSplitInfos` will reset `splitInfoIdx_` correctly.

cpp/velox/compute/VeloxPlanConverter.cc
```diff
@@ -35,183 +34,87 @@ VeloxPlanConverter::VeloxPlanConverter(
...
   }
-}
 
-void VeloxPlanConverter::setInputPlanNode(const ::substrait::FilterRel& sfilter) {
-  if (sfilter.has_input()) {
-    setInputPlanNode(sfilter.input());
-  } else {
-    throw std::runtime_error("Child expected");
-  }
+  planConverter->setSplitInfos(std::move(splitInfos));
```
rui-mo, line 107: Will we call `setSplitInfos` multiple times for an instance of planConverter?
https://github.com/oap-project/gluten/pull/4177#discussion_r1441506167

yma11:
@Yohahaha Can you rebase this PR? then I think the CI/CD should can pass.

Yohahaha:
> @Yohahaha Can you rebase this PR? then I think the CI/CD should can pass.

I will request review&&merge when I fix all ci issue, thanks.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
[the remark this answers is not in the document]
Yohahaha: Just found this format issue and fix it.

cpp/velox/compute/VeloxPlanConverter.cc
Yohahaha: should not, but not sure.

Yohahaha:
I believe CI could pass, any more comments? cc who still watching this patch.

FelixYBW:
looks the micro benchmark failed, @marin-ma any insight?

Yohahaha:
> looks the micro benchmark failed, @marin-ma any insight?

I will fix it later.

Yohahaha:
please help review again, CI has passed. @FelixYBW @rui-mo @PHILO-HE @ulysses-you

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
```diff
@@ -429,6 +419,9 @@ class ColumnarInputRDDsWrapper(columnarInputRDDs: Seq[RDD[ColumnarBatch]]) exten
   }
 
   def getPartitions(index: Int): Seq[Partition] = {
+    if (columnarInputRDDs.isEmpty) {
+      return Seq.empty
+    }
```
ulysses-you, line 424: is this change necessary ?

cpp/core/jni/JniWrapper.cc
```diff
@@ -384,7 +385,16 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithI
   auto planData = reinterpret_cast<const uint8_t*>(env->GetByteArrayElements(planArr, nullptr));
   auto planSize = env->GetArrayLength(planArr);
 
+  if (jsize splitInfoArraySize = env->GetArrayLength(splitInfosArr); splitInfoArraySize != 0) {
```
ulysses-you, line 388: move these code to line 384

cpp/velox/compute/VeloxPlanConverter.cc
```diff
@@ -34,183 +33,87 @@ VeloxPlanConverter::VeloxPlanConverter(
...
-  if (sagg.has_input()) {
-    setInputPlanNode(sagg.input());
-  } else {
-    throw std::runtime_error("Child expected");
-  }
-}
+void parseLocalFileNodes(
+    SubstraitToVeloxPlanConverter* planConverter,
+    std::vector<::substrait::ReadRel_LocalFiles>& localFiles) {
+  std::vector<std::shared_ptr<SplitInfo>> splitInfos;
```
ulysses-you, line 98: `splitInfos.reserve(localFiles.size())`

cpp/velox/compute/VeloxPlanConverter.cc
```diff
@@ -34,183 +33,87 @@ VeloxPlanConverter::VeloxPlanConverter(
...
+  // avoid include RowVectorStream.h in SubstraitToVeloxPlan.cpp, it may cause redefintion of array abi.h.
+  auto factory = [](std::string nodeId,
+                    memory::MemoryPool* pool,
+                    std::shared_ptr<ResultIterator> inputIter,
+                    RowTypePtr outputType) {
+    auto vectorStream = std::make_shared<RowVectorStream>(pool, std::move(inputIter), outputType);
+    return std::make_shared<ValueStreamNode>(nodeId, outputType, std::move(vectorStream));
+  };
+  substraitVeloxPlanConverter_.setValueStreamNodeFactory(std::move(factory));
+  substraitVeloxPlanConverter_.setInputIterators(inputIters);
```
zhztheplayer, line 48: If using this way, should we remove `setInputIterators` then capture `inputIters` in factory body?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/IteratorApiImpl.scala
```diff
@@ -80,6 +76,24 @@ class IteratorApiImpl extends IteratorApi with Logging {
...
+      splitInfos: Seq[Seq[SplitInfo]]): Seq[BaseGlutenPartition] = {
+    // Only serialize plan once, save lots time when plan is complex.
+    val planByteArray = wsCtx.root.toProtobuf.toByteArray
+
+    splitInfos.zipWithIndex.map {
+      case (splitInfos, index) =>
+        GlutenPartition(
+          index,
+          planByteArray,
+          splitInfos.map(_.asInstanceOf[LocalFilesNode].toProtobuf.toByteArray).toArray,
```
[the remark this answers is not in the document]
Yohahaha: yes.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1297,7 +1286,7 @@ int32_t SubstraitToVeloxPlanConverter::getStreamIndex(const ::substrait::ReadRel
     std::string filePath = fileList[0].uri_file();
     std::string prefix = "iterator:";
     std::size_t pos = filePath.find(prefix);
-    if (pos == std::string::npos) {
+    if (pos == std::string::npos || validationMode_) {
```
zhztheplayer, line 1289: Can we let the function return `-1` fast in `validationMode_` ?

E.g., 
```cpp
if (validationMode_) {
  return -1;
}
```

At the entrance of the function.

cc @rui-mo

zhztheplayer (commented):
Thanks for working on this!

cpp/velox/compute/VeloxPlanConverter.cc
zhztheplayer: Sorry, I still don't quite understand the point here. Capturing `inputIters` won't make the factory stateful since once it is created, nobody could change its internal creation logic of the value stream node. And e could even eliminate stateful `setInputIterators` by the change as well. Any special reason you wanted to keep it?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
zhztheplayer: It should effectively be the same with the patch's code
```cpp
if (pos == std::string::npos || validationMode_) {
  return -1;
}
```

The code looks to return -1 when `validationMode_ == true` anyway?

Yohahaha: you are right, thanks!

cpp/velox/compute/VeloxPlanConverter.cc
Yohahaha: got it, thanks!

ulysses-you (dismissed):
lgtm, thank you @Yohahaha

marin-ma (changes requested):
Thank you for this work! Could you please also update the micro benchmark documentation? Noticed that we need to specify split files for first stages.

marin-ma (approved):

marin-ma:
LGTM. Thanks!

ulysses-you (approved):

ulysses-you:
We can create followups if there are some new finding issues

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -960,6 +960,36 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
   }
 }
 
+core::PlanNodePtr SubstraitToVeloxPlanConverter::constructValueStreamNode(
+    const ::substrait::ReadRel& readRel,
+    int32_t streamIdx) {
+  // Get the input schema of this iterator.
+  uint64_t colNum = 0;
+  std::vector<TypePtr> veloxTypeList;
+  if (readRel.has_base_schema()) {
```
jinchengchenghh, line 969: I find here you change the input iterator names, and comment ValueStreamNode in Velox does not support name change, but here updates the output type based on output type, do you meet any problem? @Yohahaha 
https://github.com/apache/incubator-gluten/blob/c5b7e59335201f960cda49dff7edc315b36ed05e/cpp/velox/operators/plannodes/RowVectorStream.h#L100

jinchengchenghh: And this update only take effect on the top level columns, what if there is nested column?

jinchengchenghh: I know this is used for constructing ValueStreamNode outputType, may also need to update nested column

Issue #4182, opened 2023-12-26
[VL] Slow CI


zhztheplayer:
### Bug description

Observed slowed down CI due to bad performance of some gluten-it TPC jobs. Will propose a fix soon.

zhztheplayer:
The issue is caused by build option `ENABLE_HDFS=ON`:

*  with --enable_hdfs=ON

```
dev/builddeps-veloxbe.sh --build_type=Release --enable_hdfs=ON --enable_ep_cache=OFF && mvn -Pbackends-velox -Pspark-3.2 -DskipTests clean install
```

```
RAM statistics: JVM Heap size: 283860 KiB (total 10315776 KiB), Process RSS: 14406900 KiB

Test report: 

Summary: 5 out of 5 queries passed. 

|       Query ID|     Was Passed|            Expected Row Count|              Actual Row Count|  Baseline Query Time (Millis)|           Query Time (Millis)|          Query Time Variation|
|             q1|           true|                           100|                           100|                          4428|                          6567|                       -32.57%|
|             q2|           true|                          2513|                          2513|                          2370|                          8580|                       -72.38%|
|             q3|           true|                           100|                           100|                          1884|                          2559|                       -26.38%|
|             q4|           true|                           100|                           100|                         18137|                         23398|                       -22.48%|
|             q5|           true|                           100|                           100|                          3480|                         31549|                       -88.97%|

No failed queries. 

Overall: 

|       Query ID|     Was Passed|            Expected Row Count|              Actual Row Count|  Baseline Query Time (Millis)|           Query Time (Millis)|          Query Time Variation|
|            all|           true|                          2913|                          2913|                         30299|                         72653|                       -58.30%|
```

* without  --enable_hdfs=ON

```
dev/builddeps-veloxbe.sh --build_type=Release --enable_ep_cache=ON && mvn -Pbackends-velox -Pspark-3.2 -DskipTests clean install
```

```
RAM statistics: JVM Heap size: 300330 KiB (total 10484224 KiB), Process RSS: 13775500 KiB

Test report: 

Summary: 5 out of 5 queries passed. 

|       Query ID|     Was Passed|            Expected Row Count|              Actual Row Count|  Baseline Query Time (Millis)|           Query Time (Millis)|          Query Time Variation|
|             q1|           true|                           100|                           100|                          4595|                          2611|                        75.99%|
|             q2|           true|                          2513|                          2513|                          2057|                          1557|                        32.11%|
[...]

PR #4184, opened 2023-12-26
[GLUTEN-4182][VL] Disallow resetting HDFS client credential to fix performance issue / to avoid race condition

zhztheplayer:
`updateHdfsTokens` may lead to race condition since `spark.gluten.ugi.tokens` / `spark.gluten.ugi.username` are both resettable in Spark session while Velox `hdfsSetDefautUserName` and `hdfsSetTokenForDefaultUser` are global static API.

The patch adds a check to prevent user from resetting the two options in the same session.

Also, we will call Velox's relevant APIs mentioned above only once in the same session after the patch. This will fix the severe performance issue #4182.

Related to https://github.com/oap-project/gluten/pull/1706

### DIFF
--- a/cpp/velox/utils/HdfsUtils.cc
+++ b/cpp/velox/utils/HdfsUtils.cc
@@ -16,6 +16,7 @@
  */
 
 #include "HdfsUtils.h"
+#include <hdfs/hdfs.h>
 #include "config/GlutenConfig.h"
 #include "utils/exception.h"
 
@@ -39,7 +40,7 @@ void updateHdfsTokens(const facebook::velox::Config* veloxCfg) {
   static std::mutex mtx;
   std::lock_guard lock{mtx};
 
-  static std::optional<Credential> credential{std::nullopt};
+  static std::optional<Credential> activeCredential{std::nullopt};
 
   const auto& newUserName = veloxCfg->get<std::string>(gluten::kUGIUserName);
   const auto& newAllTokens = veloxCfg->get<std::string>(gluten::kUGITokens);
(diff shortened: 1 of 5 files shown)


zhztheplayer:
cc @ccat3z

cpp/velox/utils/HdfsUtils.h
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <velox/core/Config.h>
+#include <memory>
```
JkSelf, line 19: @zhztheplayer Please insert empty line between line 19 and line 20.

JkSelf (approved):
LGTM. Just one small comment. Also can fix in following PR. Thanks.

cpp/velox/utils/HdfsUtils.h
zhztheplayer: Will address probably in another patch. Thanks!

cpp/velox/utils/HdfsUtils.cc
```diff
@@ -0,0 +1,69 @@
...
+    for (auto& token : tokens)
+      hdfsSetTokenForDefaultUser(token.data());
+    activeCredential.emplace(newCredential);
+    return;
+  }
+
+  // active credential already set
+  GLUTEN_CHECK(
+      activeCredential.value() == newCredential,
+      "Gluten currently doesn't allow resetting HDFS activeCredential in session");
```
ulysses-you, line 67: @zhztheplayer  We encounter this check after rebase.. Can we reset token when `activeCredential.value() != newCredential` and remove this check ? I think it can also fix the original issue.

Issue #4151, opened 2023-12-22
[VL] Use `dwio::common::FileSink::create` for `LocalFileSink` and `HDFSFileSink`


dcoliversun:
### Description

As discussion as https://github.com/oap-project/gluten/pull/4129#discussion_r1433784703,  `LocalFileSink`/`HDFSFileSink` should use `dwio::common::FileSink::create`

dcoliversun:
cc @yma11

PR #4191, opened 2023-12-26
[GLUTEN-4151][VL] initialize local&hdfs file write sink with FileSink::create function

dcoliversun:

The WriteFileSink is not the API Velox supposed to expose, we need to initialize local&hdfs file write sink with FileSink::create function.

(Fixes: \#4151)

## How was this patch tested?

Original unit test in GA.

### DIFF
--- a/cpp/velox/operators/writer/VeloxParquetDatasource.cc
+++ b/cpp/velox/operators/writer/VeloxParquetDatasource.cc
@@ -45,9 +45,7 @@ const int32_t kGzipWindowBits4k = 12;
 
 void VeloxParquetDatasource::init(const std::unordered_map<std::string, std::string>& sparkConfs) {
   if (strncmp(filePath_.c_str(), "file:", 5) == 0) {
-    auto path = filePath_.substr(5);
-    auto localWriteFile = std::make_unique<LocalWriteFile>(path, true, false);
-    sink_ = std::make_unique<WriteFileSink>(std::move(localWriteFile), path);
+    sink_ = dwio::common::FileSink::create(filePath_, {.pool = pool_.get()});
   } else if (isSupportedS3SdkPath(filePath_)) {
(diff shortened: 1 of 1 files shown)


dcoliversun:
cc @yma11 It would be nice if you have time to review this

cpp/velox/operators/writer/VeloxParquetDatasource.cc
```diff
@@ -45,9 +45,7 @@ const int32_t kGzipWindowBits4k = 12;
 
 void VeloxParquetDatasource::init(const std::unordered_map<std::string, std::string>& sparkConfs) {
   if (strncmp(filePath_.c_str(), "file:", 5) == 0) {
-    auto path = filePath_.substr(5);
-    auto localWriteFile = std::make_unique<LocalWriteFile>(path, true, false);
-    sink_ = std::make_unique<WriteFileSink>(std::move(localWriteFile), path);
+    sink_ = dwio::common::FileSink::create(filePath_, {.pool = pool_.get()});
```
yma11, line 48: `pool` should not be needed for `localFileSink` and `hdfsFileSink`.  You can take a check at velox layer.

dcoliversun: Correct me if I'm wrong. I refer [velox code](https://github.com/facebookincubator/velox/blob/c5bcc33e93678e5bf0a7479ba16212035008a302/velox/dwio/common/tests/LocalFileSinkTest.cpp#L41-L50) to initialize the sink, could you please share velox layer?

yma11: Yes. Seems `pool` is always provided in velox's usage even it's not actually used.

yma11 (approved):

PR #4199, opened 2023-12-27
[VL] Daily Update Velox Version (2023-12-27) 

zhztheplayer:
```
79c9754bc (upstream/main) Add to log reference count of active root pools (#8180)
601b98812 Add max_spill_bytes config (#8171)
75617ae1a Fix memory reservation in hash join spill (#8174)
```

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -17,7 +17,7 @@
 set -exu
 
 VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=231226
+VELOX_BRANCH=2023_12_27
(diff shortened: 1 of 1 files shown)


ep/build-velox/src/get_velox.sh
```diff
@@ -17,7 +17,7 @@
 set -exu
 
 VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=231226
+VELOX_BRANCH=2023_12_27
```
zhztheplayer, line 20: Changed back to branch since tag is not that convenient for merging PRs.

JkSelf (approved):
LGTM. Thanks.

Issue #3943, opened 2023-12-06
[CH] Swith default shuffle hash algorithm from cityHash64 to sparkMurmurHash3_32 because the former impl in CH is not consistent with vanilla spark 


taiyang-li:
### Description

可能会有部分ut失败，需要修复。

PR #4205, opened 2023-12-27
[GLUTEN-3943][CH] Support complex input arguments for sparkMurmurHash3_32, and switch default hash algorithm to sparkMurmurHash3_32 

taiyang-li:

Fixes: \#3943, also close https://github.com/oap-project/gluten/issues/2341

### DIFF
--- a/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHBackend.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHBackend.scala
@@ -99,7 +99,7 @@ object CHBackendSettings extends BackendSettingsApi with Logging {
     GlutenConfig.GLUTEN_CONFIG_PREFIX + CHBackend.BACKEND_NAME +
       ".shuffle.hash.algorithm"
   // valid values are: cityHash64 or sparkMurmurHash3_32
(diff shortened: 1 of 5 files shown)


taiyang-li:
```
- gluten Exchange reuse *** FAILED ***
  2 did not equal 1 (GlutenAdaptiveQueryExecSuite.scala:450)
```

taiyang-li:
> ```
> - gluten Exchange reuse *** FAILED ***
>   2 did not equal 1 (GlutenAdaptiveQueryExecSuite.scala:450)
> ```

@zzcclp I can't reproduce this ut with sqls below, can you help solve it, thanks
``` sql 
set spark.sql.adaptive.enabled = true;
set spark.sql.adaptive.autoBroadcastJoinThreshold = 90;
set spark.sql.shuffle.partitions = 5;

CREATE TEMPORARY VIEW testData AS
select cast(id+1 as int) as key,  cast(id as string) as value from range(100);

CREATE TEMPORARY VIEW testData2 AS
SELECT *
FROM VALUES
  (1, 1),
  (1, 2),
  (2, 1),
  (2, 2),
  (3, 1),
  (3, 2)
AS testData2(a, b);



CREATE TEMPORARY VIEW testData3 AS
SELECT *
FROM VALUES
  (1, null),
  (2, 2)
as testData3(a, b);


explain EXTENDED SELECT value FROM testData join testData2 ON key = a join (SELECT value v from testData join testData3 ON key = a) on value = v

```

zzcclp (approved):
LGTM

PR #4207, opened 2023-12-27
[CORE] Add preProjection and postProjection in WindowExecTransformer

liujiayi771:

Window and agg are similar, and in Velox, they also do not support input that contains expressions. The input must be an attribute. Therefore, if there are non-attribute columns in the input for window, we need to add a preProjection to execute them. And similar to the approach in `SortExecTransformer`, the output of the window, which is `child.output ++ windowExpr`, requires a postProjection to remove any extra columns from child.output in the final output (where child refers to the added preProjection).

```sql
SELECT
  c_custkey,
  row_number() OVER (
    PARTITION BY c_nationkey,
    "a"
    ORDER BY
      c_custkey,
      "a"
  ) AS row_num
FROM
   customer
ORDER BY 1, 2;
```
Before this PR, the window in above query will fallback due to the reason "native validation failed due to: Only field is supported in window functions".

When converting window to a substrait plan, there are three parts that need to be handled: partitionExprs, sortExprs, and windowExprs. Expressions may exist in any of these three parts, and they all need to be verified. However, in the current PR, only partitionExprs and sortExprs are considered. I plan to refactor the existing logic of preProjection and postProjection in Gluten to be corrected based on Rules in the future. Therefore, in the future, support for windowExprs will be added.

## How was this patch tested?

Add a new test case GlutenSQLWindowFunctionSuite."Literal in window partition by and sort" for Velox and ClickHouse.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/WindowExecTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/WindowExecTransformer.scala
@@ -101,7 +101,7 @@ case class WindowExecTransformer(
     BackendsApiManager.getTransformerApiInstance.packPBMessage(message)
   }
 
-  private def needsPreProject: Boolean = (partitionSpec ++ orderSpec.map(_.child)).exists {
+  private def needsProject: Boolean = (partitionSpec ++ orderSpec.map(_.child)).exists {
     case _: Attribute => false
     case _ => true
   }
@@ -172,7 +172,7 @@ case class WindowExecTransformer(
     }
   }
 
-  private def getWindowRelWithPreProjection(
+  private def getWindowRelWithProjection(
       context: SubstraitContext,
       originalInputAttributes: Seq[Attribute],
       operatorId: Long,
@@ -230,7 +230,7 @@ case class WindowExecTransformer(
         operatorId,
         emitStartIndex)
     }
-    getWindowRelAfterProject(
+    createWindowRelAfterProjection(
       context,
       originalInputAttributes,
       preExprNodes,
@@ -240,7 +240,7 @@ case class WindowExecTransformer(
       validation)
   }
 
-  private def getWindowRelAfterProject(
+  private def createWindowRelAfterProjection(
       context: SubstraitContext,
       originalInputAttributes: Seq[Attribute],
       preExprNodes: JList[ExpressionNode],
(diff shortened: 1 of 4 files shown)


liujiayi771:
Run Gluten Clickhouse CI

rui-mo:
@PHILO-HE Is this PR relevant to your change on window literal support?

PHILO-HE:
> @PHILO-HE Is this PR relevant to your change on window literal support?

No, it's not relevant to my work.

liujiayi771:
Run Gluten Clickhouse CI

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/execution/GlutenSQLWindowFunctionSuite.scala
```diff
@@ -16,6 +16,71 @@
...
+    (60865, 5),
+    (61065, 13),
+    (127412, 13),
+    (148303, 10),
+    (9954, 5),
+    (95337, 12)
+  )
 
-class GlutenSQLWindowFunctionSuite extends SQLWindowFunctionSuite with GlutenSQLTestsTrait {}
+  test("Literal in window partition by and sort") {
```
rui-mo, line 41: > in the current PR, only partitionExprs and sortExprs are considered

Is it possible to also add test for window expressions? At least to make sure it is runnable with fallback.

gluten-core/src/main/scala/io/glutenproject/execution/WindowExecTransformer.scala
```diff
@@ -182,6 +172,206 @@ case class WindowExecTransformer(
...
+      context,
+      originalInputAttributes,
+      preExprNodes,
+      selections,
+      inputRel,
+      operatorId,
+      validation)
+  }
+
+  private def getWindowRelAfterProject(
```
rui-mo, line 243: The name `getWindowRelAfterProject` seems to be easily mixed up with `getWindowRelWithPreProjection`. Could we use a more precise name?

liujiayi771: Similar to agg, naming functions is hard, just simply modified to 'create'. This part will be removed after inserting pre/post-project through Rule.

rui-mo (approved):

Issue #4147, opened 2023-12-22
[CH] support function getarraystructfield


taiyang-li:
### Description

support function getarraystructfield

taiyang-li:
Let's do it!

PR #4210, opened 2023-12-27
[GLUTNE-4147][CH] Support GetArrayStructFields function for CH

taiyang-li:

(Fixes: \#4147)
 

Notice: current pr must be tested and merged after https://github.com/Kyligence/ClickHouse/pull/477 is merged


taiyang-li:
@zzcclp can you help review this change, thanks!

liuneng1994 (commented):
LGTM

gluten-core/src/main/scala/io/glutenproject/expression/UnaryExpressionTransformer.scala
```diff
@@ -215,3 +215,27 @@ case class RandTransformer(
     ExpressionBuilder.makeScalarFunction(functionId, inputNodes, typeNode)
   }
 }
+
+case class GetArrayStructFieldsTransformer(
```
rui-mo, line 219: Have you met any issue to use the default transformation? If not, maybe we don't need to add a specific transformer for this function.

gluten-core/src/main/scala/io/glutenproject/expression/UnaryExpressionTransformer.scala
```diff
@@ -215,3 +215,27 @@ case class RandTransformer(
...
+    substraitExprName: String,
+    child: ExpressionTransformer,
+    ordinal: Int,
+    numFields: Int,
+    containsNull: Boolean,
```
rui-mo, line 224: It seems numFields and containsNull are not added to the function arguments. Are they needed?

gluten-core/src/main/scala/io/glutenproject/expression/UnaryExpressionTransformer.scala
taiyang-li: Default transformation is not acceptable for current expr. Because ordinal should be passed as an extra argument in substrait scalar function, at least for CH backend(I'm not sure if it is needed by Velox backend)

gluten-core/src/main/scala/io/glutenproject/expression/UnaryExpressionTransformer.scala
taiyang-li: numFields and containsNull  are not needed in CH backend.

rui-mo (approved):

Issue #4213, opened 2023-12-28
Refactoring the pre/post projection logic based on Spark Rule


liujiayi771:
### Description

Due to the limitation that only the project operator in the native engine supports evaluating expressions, there is currently a lot of code logic in the transformer to handle inputs with expressions. To address this, we can refactor the code by unifying the handling logic using Spark Rule. We can extract the pre/post projection logic from transformers like agg, sort, etc., and create a separate project transformer. This approach can also be combined with the project collapse optimization #3937 I previously implemented, which allows for merging redundant project operations to improve performance.

liujiayi771:
cc @rui-mo

zhztheplayer:
Thank you @liujiayi771 . 

First of all, just a naive question: It sounds like "logical optimization" in Spark to merge filters / projects or to pull up / down pre/post projects from other plan nodes. Why do we have to inject "physical optimizations" for doing so? I believe vanilla Spark already had a logical rule to merge consecutive projects.

liujiayi771:
@zhztheplayer The main purpose of extracting pre/post project is to optimize the current code logic. Currently, the inserted pre/post code in the transformer is quite messy, and each operator has its own separate implementation. The ability to merge is just an additional benefit.

Spark does have an optimizer for merging project/filter operations. However, Gluten has made modifications to the physical plan, which includes unmerged `projectExecTransformers`. For example, before #3937, redundant `projectExecTransformers` can be seen before shuffling. As for pre/post project operations, the current approach is to directly insert a project RelNode into the substrait plan. This means that there is a project operation at the execution layer, but we cannot perceive it in the plan UI and cannot collapse the project operation.

![image](https://github.com/oap-project/gluten/assets/13622031/ffd68f12-4670-40fc-9acf-65c6a2433ca3)

zhztheplayer:
Thanks for the explanation. Then I understand why you wrote a physical rule for project collapse.

So regarding the feature you proposed in the issue (unified way to pull pre/post projects), would you think we can implement within logical rules? In the way we may be able to leverage other logical optimizations from vanilla Spark's query planner.

Or if we had no choice other than using physical rules, I would be OK to that way either.

liujiayi771:
[...]

PR #3649, opened 2023-11-08
[WIP][CORE] Add PullOutPreProject rule to decouple substrait pre-project

ulysses-you:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#ISSUE-ID)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #4245, opened 2024-01-02
[GLUTEN-4213][CORE] Refactoring insertion process of pre/post projection

liujiayi771:

Implement [#4213](https://github.com/oap-project/gluten/issues/4213). 
Introduced three kinds of Rules.
* `PullOutProject`. Pulling out pre-project at the `LogicalPlan` level. Currently, it only supports the Velox backend and can reduce the number of pre-projects when agg includes distinct.
* `ColumnarPullOutProject` (`ColumnarPullOutPostProject` + `ColumnarPullOutPreProject`). Pulling out pre/post-project at the `SparkPlan` level. `PullOutProject` cannot handle all scenarios (e.g., `Aggregate` introduced by `InjectRuntimeFilter` will be executed before `PullOutProject`, and some `Expression`s will be generated in `Strategy`). The missing parts will be handled completely by `ColumnarPullOutProject`. Some information required for post-project is more easily obtained at the physical plan level, hence it is handled there.
* `GlutenPlanPullOutProject`. Handling the case of constructing a Gluten transformer directly in `TakeOrderedAndProjectExecTransformer`.

Currently, only agg and sort have been incorporated into this framework. In the future, support for operators such as join and window that require pre/post projection will be added.

Next steps:
1. Support join/window in this framework
2. Add a rule for row_constructor required by velox backend.

## How was this patch tested?

Exists CI.


liujiayi771:
@zhztheplayer @rui-mo Could you help review? I have already validated this modification on TPCDS. Using this framework to insert pre/post projection can eliminate a significant amount of redundant code in the transformer. The previous approach required many if-else branches based on whether to insert projection and whether it was for validation. It also eliminated the need to construct projection based on an index.

liujiayi771:
@waitinfuture Could you help review?

liujiayi771:
Design doc https://github.com/oap-project/gluten/issues/4213#issuecomment-1873669924

rui-mo (commented):
Thanks for the nice refactor. In the meantime, could you also check if metrics work well? Some relevant code to handle the metrics of pre/post projection could be removed.

rui-mo:
Seems there is a PR https://github.com/oap-project/gluten/pull/3649 proposing the similar refactor. @ulysses-you @liujiayi771 Could you help check on that? Thanks.

liujiayi771:
The work done by these two PRs is essentially the same, with the difference being that #3649 modifies the logical plan, while my PR modifies the physical plan. For the sort operator, both pre and post projection can be modified in the logical plan. For agg, the pre projection can be modified in the logical plan, but the post projection can only be modified in the physical plan if the native output doesn't match with the resultExpressions in Spark's output. 

Initially, I also considered doing it in the logical plan to avoid impacting validation and AQE. I think we can combine both approaches, doing the parts that can be done in the logical plan in the logical rule, but for the sort in TakeOrderedAndProjectExecTransformer, it should be done only in the physical plan. @ulysses-you I didn't notice your PR before. I searched for issues related to project but did not check the pull requests. I would like to hear your opinion again, as our approaches are fairly similar.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -157,14 +158,12 @@ abstract class HashAggregateExecBaseTransformer(
...
-      case _: Attribute => false
-      case _ => true
-    } || aggregateExpressions.exists {
+  override def needsPreProjection: Boolean = {
+    needsPreProjection(groupingExpressions) ||
+    aggregateExpressions.exists {
       expr =>
         expr.filter match {
-          case None | Some(_: Attribute) | Some(_: Literal) =>
+          case None | Some(_: Attribute) =>
```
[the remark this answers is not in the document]
liujiayi771: There are two reasons for this:
- If the filter condition is a Literal, it can only be of boolean type. Such filters are typically removed during the optimization process in Spark. You can refer to the "EliminateAggregateFilter" function in Spark for more information.
- filter in velox only support `FieldAccessTypedExpr`, but it is possible that CK supports Literal filter.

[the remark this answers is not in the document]
rui-mo: Thanks. Makes sense.

ulysses-you:
I think the main goal to pull out pre/post project is:
1. make the transformer plan tree align with native plan tree; e.g., if we have a native project then we must have a project transformer
2. decouple pre/post project fallback with the original operator; e.g., if we support transform aggregate but does not support post project, then we should only fallback post project
3. avoid expression multi-evaluation; e.g., `t1 join t2 on c1 + 1 = c2`, say it's a shuffled hash join then we will evaluate `c1 + 1` multi-times, one for shuffle, one for pre-project

One option is that, we can do pull out pre-project at logical side and do pull out post-project at columnar side.

liujiayi771:
@ulysses-you Aggree with you. I can continue to modify the pre-projection part into logical rule if you'd like, or would you prefer to continue working on #3649?

ulysses-you:
@liujiayi771  it's fine to go ahead in this pr, thank you

liujiayi771:
@ulysses-you I have identified an issue where, if we modify the logical plan, the extendedOperatorOptimizationRules we insert is placed before DecimalAggregates. DecimalAggregates converts sum/avg(decimal attr) into sum/decimal(unscaledValue(decimal attr)), but the unscaledValue cannot be seen in our rule. This results in the required pre-project not being added.

Maybe we should use ExperimentalMethods.extraOptimizations or postHocOptimizationBatches? I currently do not know how to use postHocOptimizationBatches.

ulysses-you:
`sparkSession.experimental.extraOptimizations = sparkSession.experimental.extraOptimizations ++ Seq(yourRule)`
Does it work ?

liujiayi771:
> sparkSession.experimental.extraOptimizations ++

Yes, this approach will work, but do you think it is reasonable to use `ExperimentalMethods`? Will Spark remove this class in the future? However, there is no other way to add rule at the end. Currently, it seems that there is no mechanism in Gluten that allows modifying the extraOptimizations in the spark session right after it is launched.

liujiayi771:
@ulysses-you One method I can think of is to add a rule through injectResolutionRule, before the optimization step. This rule would only perform the modifications on the sparkSession. However, this approach might be considered as a hack.

liujiayi771:
Hi @waitinfuture, @rui-mo @ulysses-you, it's ready for review. Originally, I planned to submit the PR after completing the Join. However, testing revealed that there are multiple changes involved, resulting in a large PR. Therefore, I would prefer to complete the Aggregate and Sort parts first before proceeding with the Join and Window.

The current code has undergone some changes compared to the previous design, let me explain.

There are a total of 2 rules: `PullOutProject` (`LogicalPlan`` Rule) and `ColumnarPullOutProject` (`GlutenPlan`` Rule). We have also discussed whether to insert pre/post-project in the logical plan or the physical plan. The previous conclusion was that it is better to do it in the logical plan because it can reduce the number of pre-projects in scenarios like distinct aggregations, improving performance, and it also allows for fallback targeted only at the project during validation. However, during actual development, I found that some plans cannot be processed in and out at the logical plan level.

* To maintain the correctness of the `outputOrdering` for operators downstream of sort, it is necessary to override the `outputOrdering`` of sort to handle cases where its child is a pre-project.
* When an Aggregate is included in DPP, it will be executed at an early stage and cannot be processed by PullOutProject. It can only be seen in ColumnarPullOutProject.
* Regarding the Join support that will be provided in the future, I have also conducted some tests. When the condition of the Join contains `EqualNullSafe`, `Coalesce` and `IsNull` will be wrapped around it during the conversion from logical plan to physical plan, which needs to be handled through pre-project. These cannot be seen in `PullOutProject` either. Furthermore, the `BuildSide` information of Join cannot be obtained in the logical plan, and post-project, which requires the exchange of output order, cannot be handled either. Therefore, when Join is supported later, it will also need to be addressed in `ColumnarPullOutProject`.

Taking into account the above reasons, the current approach is to handle what can be processed in the logical plan within the `LogicalPlan` Rule. The handling of the physical plan is considered as a fallback solution to address any missed cases (currently only applicable to the Aggregate scenario). The parts that can only be handled in the physical plan will be addressed within the `GlutenPlan` Rule.

liujiayi771:
Benchmark Velox

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarPullOutProject.scala
```diff
@@ -0,0 +1,183 @@
...
+  private def tag(plan: SparkPlan, hint: ProjectProcessedHint): Unit = {
+    plan.setTagValue(TAG, hint)
+  }
+}
+
+object ColumnarPullOutPostProject extends Rule[SparkPlan] with PullOutProjectHelper {
+
+  override def apply(plan: SparkPlan): SparkPlan = plan.transform {
+    case agg: HashAggregateExecBaseTransformer
+        if !ProjectProcessedHint.isPostProjectProcessed(agg) &&
```
ulysses-you, line 61: why do we need this hint ? after we pull out post project, it should nothing happen when we go in this rule next time

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -791,8 +791,7 @@ case class ColumnarOverrideRules(session: SparkSession)
         (_: SparkSession) => TransformPreOverrides(isAdaptiveContext),
         (_: SparkSession) => RemoveNativeWriteFilesSortAndProject(),
         (spark: SparkSession) => RewriteTransformer(spark),
-        (_: SparkSession) => EnsureLocalSortRequirements,
-        (_: SparkSession) => CollapseProjectExecTransformer
+        (_: SparkSession) => EnsureLocalSortRequirements
       ) :::
       BackendsApiManager.getSparkPlanExecApiInstance.genExtendedColumnarPreRules() :::
```
ulysses-you, line 796: We should apply pull out rules between `AddTransformHintRule` and `TransformPreOverrides`.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -462,7 +462,11 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
                 plan.resultExpressions,
                 plan.child
               )
-            TransformHints.tag(plan, transformer.doValidate().toTransformHint)
+            val allTransformable = ColumnarPullOutProject
+              .getTransformedPlan(transformer)
+              .map(_.asInstanceOf[GlutenPlan].doValidate())
+              .reduce(ValidationResult.merge)
```
ulysses-you, line 468: I think we should only validate the aggregate, so we can decouple the fallback for project and aggregate.

gluten-core/src/main/scala/io/glutenproject/execution/SortExecTransformer.scala
```diff
@@ -52,124 +50,26 @@ case class SortExecTransformer(
 
   override def output: Seq[Attribute] = child.output
 
-  override def outputOrdering: Seq[SortOrder] = sortOrder
+  override def outputOrdering: Seq[SortOrder] = child match {
+    case project: ProjectExecTransformer if project.isPreProject =>
```
ulysses-you, line 54: do we really need `isPreProject` flag ? it seems we can do it for all projects

gluten-core/src/main/scala/io/glutenproject/extension/AddExtraOptimizations.scala
```diff
@@ -0,0 +1,30 @@
...
+package io.glutenproject.extension
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+
+case class AddExtraOptimizations(sparkSession: SparkSession) extends (LogicalPlan => Unit) {
+
+  override def apply(plan: LogicalPlan): Unit = {
+    if (!sparkSession.experimental.extraOptimizations.exists(_.isInstanceOf[PullOutProject])) {
+      sparkSession.experimental.extraOptimizations = sparkSession.experimental.extraOptimizations ++
```
ulysses-you, line 26: If I get it corrent, the reason why we add the pull out project rule to `extraOptimizations` is `DecimalAggregates` ? Is there other rules affect ?

Can we inject the pull out project rule to `preCBORules` and apply `DecimalAggregates` first manually ?

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarPullOutProject.scala
liujiayi771: It can be removed here. In a previous version of implementation, I did not rewrite the `resultExpressions` of agg, which caused it to always match here. I will remove this check in agg.

However, the hint might still be needed for implementing join because join requires a post-project when the order of left and right need to be changed. This condition will always be satisfied for join.

gluten-core/src/main/scala/io/glutenproject/extension/AddExtraOptimizations.scala
liujiayi771: Yes, currently we have only found that `DecimalAggregates` can cause an impact. I haven't figured out why Spark decides to place `DecimalAggregates` in this position, but it could be because adding `UnscaledValue` might affect the matching of certain optimizers. As a cautious approach, can we handle the expressions added by `DecimalAggregates` in `ColumnarPullOutPreProject`?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
liujiayi771: However, currently we can only obtain the Pre`ProjectExecTransformer` and Post`ProjectExecTransformer` generated by the Columnar Rule at this point. We cannot handle them separately like we do for verifying regular `ProjectExecTransformer`. I think we need to add a hint tag in the agg to record its pre/post-project fallback requirements, and then directly generate a fallback plan when generating pre/post-project.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarPullOutProject.scala
ulysses-you: I'm not sure it is required for join, at least can add it when needed

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
ulysses-you: It should be easy to get the aggregate after applying pull out project rules. We pass an aggregate in and it should only output two options, 1) aggregate, 2) project-aggregate. So we can strip the top project if exists.

gluten-core/src/main/scala/io/glutenproject/extension/AddExtraOptimizations.scala
ulysses-you: yes, I think it's better to apply `DecimalAggregates` in `PullOutPreProject` manually at logical side

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
liujiayi771: Currently, the "pull out" is applied to the columnar plan and needs to be performed after `TransformPreOverrides`.

gluten-core/src/main/scala/io/glutenproject/execution/SortExecTransformer.scala
liujiayi771: For a normal `ProjectTransformer`, executing this process poses no problems, but the resulting `sortOrder` remains the same as the original.

gluten-core/src/main/scala/io/glutenproject/extension/AddExtraOptimizations.scala
liujiayi771: This will cause some issues.
```
Max iterations (100) reached for batch Operator Optimization before Inferring Filters.
```

ulysses-you: @liujiayi771 are you injecting this rule using `preCBORules` ? it should not affect the Inferring Filters.

gluten-core/src/main/scala/io/glutenproject/execution/SortExecTransformer.scala
liujiayi771: I find a case that before the sort there is a project to rename the sort key, we should not replace alias in this case. So we still need pre-project hint for sort.

gluten-core/src/main/scala/io/glutenproject/extension/AddExtraOptimizations.scala
liujiayi771: @ulysses-you Sorry, I used the `OptimizerRules`. Let me try using the `preCBORules` instead.

zhztheplayer (commented):
Thanks!

And would you like to update the PR description to add some words to summarize the newly added rules? This would help others understand the changes.

(probably including)
1. ColumnarPullOutProject (ColumnarPullOutPostProject + ColumnarPullOutPreProject)
2. GlutenPlanPullOutProject
3. PullOutProject

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -755,12 +768,17 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
               val limitPlan = LimitTransformer(plan.child, 0, plan.limit)
               tagged = limitPlan.doValidate()
             } else {
+              val sortExec = SortExec(plan.sortOrder, false, plan.child)
+              val pulledOutSortExec =
+                ColumnarPullOutProject.getPulledOutPlanLocally[SortExec](sortExec)
```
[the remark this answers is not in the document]
liujiayi771: `ColumnarPullOutProject` will pull out `ProjectExec` and need to verify if `ProjectExec` can be converted to native plan. If we want to decouple `ColumnarPullOutProject` from `TransformHintRule`, we can place it before `TransformHintRule`, which is also feasible. Initially, I implemented it this way, but I encountered an issue where ClickHouse's [custom agg](https://github.com/oap-project/gluten/blob/25f0d69f1205ac3a64d6c2d8bf36b5b4ae6263aa/gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala#L423) https://github.com/oap-project/gluten/pull/3629#discussion_r1384889177 throws an exception when determining if post-project is needed. It may require ClickHouse's assistance to redesign the API for custom agg and not rely on throwing exceptions in `getAttrsIndexForExtensionAggregateExpr` for fallback, but instead trigger the fallback logic in `doValidationInternal`. 

We can proceed with the modifications step by step. For now, let's place the validation of `ProjectExec` within the rule itself. `ColumnarPullOutProject` will only validate `ProjectExec`. I understand that there are similar codes in [other places](https://github.com/oap-project/gluten/blob/25f0d69f1205ac3a64d6c2d8bf36b5b4ae6263aa/gluten-core/src/main/scala/io/glutenproject/execution/ScanTransformerFactory.scala#L119) in Gluten that tag hints, and we can handle them together later.

gluten-core/src/main/scala/io/glutenproject/extension/GlutenPlan.scala
```diff
@@ -46,6 +46,22 @@ object ValidationResult {
...
+   *   - invalid merge invalid = invalid
+   */
+  def merge(first: ValidationResult, second: ValidationResult): ValidationResult = {
+    if (first.isValid && second.isValid) {
+      ok
+    } else {
+      val reasonStr = first.reason.getOrElse("") + second.reason.getOrElse("")
+      notOk(reasonStr)
+    }
+  }
```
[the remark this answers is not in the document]
liujiayi771: Yes, we can remove it now.

liujiayi771:
@JkSelf Could you also help to take a look?

gluten-core/src/main/scala/io/glutenproject/execution/SortExecTransformer.scala
```diff
@@ -52,123 +51,36 @@ case class SortExecTransformer(
 
   override def output: Seq[Attribute] = child.output
 
-  override def outputOrdering: Seq[SortOrder] = sortOrder
+  override def outputOrdering: Seq[SortOrder] = child match {
+    case project: ProjectExecTransformer if ProjectTypeHint.isPreProject(project) =>
```
liujiayi771, line 55: @JkSelf The issue with `outputOrdering` encountered earlier is currently resolved as follows.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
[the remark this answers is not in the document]
zhztheplayer: > I understand that there are similar codes in [other places](https://github.com/oap-project/gluten/blob/25f0d69f1205ac3a64d6c2d8bf36b5b4ae6263aa/gluten-core/src/main/scala/io/glutenproject/execution/ScanTransformerFactory.scala#L119) in Gluten that tag hints, and we can handle them together later.

Actually I think we may allow adding validation tags to the original physical plan in rules other than `AddTransformHintRule` as of now, just as if multiple rules are doing validation together. However I am not sure whether the code you mentioned above can be a good case so it might still be needed to optimize, I'll take a look as well.

> I encountered an issue where ClickHouse's [custom agg](https://github.com/oap-project/gluten/blob/25f0d69f1205ac3a64d6c2d8bf36b5b4ae6263aa/gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala#L423) https://github.com/oap-project/gluten/pull/3629#discussion_r1384889177 throws an exception when determining if post-project is needed.

This is interesting... So my feeling is we may have to re-think how should we handle backend-specific pre/post project creation code when doing the refactor. Say, if backend A has some specialized conditions to decide whether a project should be pulled out from a plan node, the we'd provide extensibility to have it customized?

Also in the patch, code of the new feature is currently located into several places, including logical optimization, validation (transform hint), and physical optimization (the actual pulling logic). So I slightly feel that the complexity added to Gluten is a little bit higher than what we expected? Do we have chance to reduce? At the same time am just more worried about coupling of the rules in this PR. Do you think we can add some new methods to backend API to deal with the CH Agg issue you mentioned?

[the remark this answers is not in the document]
liujiayi771: The part about "custom agg" may not have been expressed clearly. Currently, both CH and Velox require a "post project" process. The logic is the same, which is to convert the output on the native side to a consistent output for Spark. What I mean is that CH throws an exception when retrieving the native output, which requires validation before pulling it out. It doesn't mean that CH "post project" process is different from Velox's and has custom requirements. In fact, it is about the `getAttrForAggregateExprs` method that retrieves the actual output of the aggregation. Based on this output, a "post-project" is constructed. Velox also has validation logic that throws exceptions. I also hope to include this logic in the `doValidationInternal` method like [this](https://github.com/oap-project/gluten/pull/4459). CH has custom aggregation requirements, and validation is also performed when retrieving the output for custom agg. For example, the `CustomSum` only supports Final [code](https://github.com/oap-project/gluten/blob/19a863c0d3781034b4def57be899e5bd4a8825b6/backends-clickhouse/src/test/scala/io/glutenproject/execution/extension/CustomAggExpressionTransformer.scala#L49). Since I am not familiar with the specific logic of other custom agg in CH, I cannot move the validation logic into `doValidationInternal` like in Velox. It may require CH's developers to redesign this part. However, this part is not essential and can be improved in future development.

gluten-core/src/main/scala/io/glutenproject/extension/PullOutProject.scala
```diff
@@ -0,0 +1,124 @@
...
+          // We cannot pull out the children of TypedAggregateExpression to pre-project,
+          // and Gluten cannot support TypedAggregateExpression.
+          false
+        case ae: AggregateExpression
+            if ae.filter.exists(isNotAttribute) || ae.aggregateFunction.children.exists(
+              isNotAttributeAndLiteral) =>
+          true
+        case _ => false
+      }.isDefined)
+    case Sort(order, _, _) =>
```
JkSelf, line 58: @liujiayi771 We have added a sort check in the `needPreProject` method. However, it appears that the logic for handling sort operators is not being added in this context [here](https://github.com/oap-project/gluten/pull/4245/files#diff-fa6b47ebbfa2f2ef8742dc00ce6ca270e050eb6e9d731704b0f0c5d89db37b9dR114. ).

gluten-core/src/main/scala/io/glutenproject/extension/columnar/ColumnarPullOutProject.scala
```diff
@@ -0,0 +1,279 @@
...
+          case (exprAttr: Attribute, resAttr) =>
+            // If the result attribute and result expression has different name or type,
+            // post-projection is needed.
+            exprAttr.name != resAttr.name || exprAttr.dataType != resAttr.dataType
+          case _ =>
+            // If result expression is not instance of Attribute,
+            // post-projection is needed.
+            true
+        }
+      case _ => false
```
JkSelf, line 143: Do we need the sort check here?

liujiayi771: Sort is different from Agg, if it has pre-project, it always needs a post-project, so I pullout pre and post project together in `ColumnarPullOutPreProject` for `SortExec`.

gluten-core/src/main/scala/io/glutenproject/extension/PullOutProject.scala
JkSelf: Could you provide more information about the `outputOrdering` issue you mentioned? Maybe i missed some context. Thanks.

liujiayi771: @JkSelf You can check this [discussion](https://github.com/oap-project/gluten/discussions/4308#discussioncomment-8036736).

rui-mo (commented):
Thanks for your work. Added several comments.

PR #4247, opened 2024-01-02
[VL] Follow-up: decouple velox test build from gluten cpp test build and use VELOX_BUILD_TEST_UTILS instead

PHILO-HE:

This is a follow-up work for https://github.com/oap-project/gluten/pull/4153. There are some dependencies lacked for gluten cpp test build after that PR. Here, we turn on VELOX_BUILD_TEST_UTILS when gluten cpp build test is on. Maybe, we can deprecate `build_velox_test`, `build_velox_benchmark` in the future.

## How was this patch tested?

Pass CI build and local build.


dcoliversun (approved):
LGTM

rui-mo (approved):

PR #4248, opened 2024-01-02
Upgrade project version from 1.1.0-SNAPSHOT to 1.2.0-SNAPSHOT

yaooqinn:

Upgrade project version from 1.1.0-SNAPSHOT to 1.2.0-SNAPSHOT to distinguish itself from branch-1.1

## How was this patch tested?

Pass existing CI workflows

### DIFF
--- a/backends-clickhouse/pom.xml
+++ b/backends-clickhouse/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <artifactId>gluten-parent</artifactId>
     <groupId>io.glutenproject</groupId>
-    <version>1.1.0-SNAPSHOT</version>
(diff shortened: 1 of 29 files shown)


PHILO-HE:
Looks duplicated to https://github.com/oap-project/gluten/pull/3950. We may need some time to do some alignment work in internal test environment. cc @zhouyuan

zhouyuan:
@yaooqinn many thanks for this patch! it's missed in the past two weeks indeed - just got LGTM from @baibaichen should be handled now. 

thanks, -yuan

yaooqinn:
NVM, I'll close this as a duplication.

PR #4253, opened 2024-01-02
[VL] Remove file format check for pushed down filters

PHILO-HE:

The check logic seems doing nothing actually.

## How was this patch tested?

Existing tests.


cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -894,26 +894,6 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
   }
 }
 
-bool isPushDownSupportedByFormat(
```
PHILO-HE, line 897: Hi @zhztheplayer, this check seems not taking effect. Could you help confirm? Thanks!

zhztheplayer (approved):

PR #4259, opened 2024-01-03
[VL] Add orc char type validation

kerwin-zk:

Currently velox does not support the char type of orc.
```
Error Source: RUNTIME
Error Code: UNKNOWN
Reason: {} not supported yet.CHAR
Retriable: False
Context: Split [Hive: oss:///user/hive/warehouse/tpcds_1000.db/customer/000003_0 64101152 - 7199294] Task Gluten stage-3 task-2073
Top-Level Context: Same as context.
Function: kind
File: ../../velox/dwio/dwrf/common/FileMetadata.cpp
Line: 100
```


kerwin-zk:
@rui-mo  cc

kerwin-zk:
> Thanks for your fix. Could you share some background? Is the Parquet generated with certain option?

@rui-mo 
When Spark reads the orc table created by Hive, if the table contains the char type, this exception will occur when Velox  reads the orc file, because velox does not support the char type of orc. Velox parquet supports char type, so there is no such problem.

rui-mo (approved):

PR #4262, opened 2024-01-03
[VL] Disable system curl to build cpr

boneanxs:

Velox disables system curl to build cpr to avoid any conflicts from pr: https://github.com/facebookincubator/velox/pull/7718
But it introduce other tests issues like https://github.com/facebookincubator/velox/pull/7718#issuecomment-1850056124
Which is being resolved in https://github.com/facebookincubator/velox/pull/7965

So it would be nice we can disable system curl also to avoid conflicts.

## How was this patch tested?

Existing tests

### DIFF
--- a/ep/build-velox/src/modify_velox.patch
+++ b/ep/build-velox/src/modify_velox.patch
@@ -35,17 +35,6 @@ index d49115f12..1aaa8e532 100644
 +          IMPORTED_LOCATION_DEBUG "${LZ4_LIBRARY_DEBUG}")
 +  endif()
  endif()
-diff --git a/CMake/resolve_dependency_modules/cpr.cmake b/CMake/resolve_dependency_modules/cpr.cmake
-index 4e5bf23ed..1f8deb0d2 100644
---- a/CMake/resolve_dependency_modules/cpr.cmake
-+++ b/CMake/resolve_dependency_modules/cpr.cmake
-@@ -30,5 +30,5 @@ FetchContent_Declare(
-   PATCH_COMMAND git apply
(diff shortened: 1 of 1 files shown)


zhouyuan:
@boneanxs thanks for the fix - it's missed indeed. 

-yuan

Yohahaha:
curious have you encounter curl issue like you mentioned in PR description? 

I think keep this patch which set use_system_curl=on is needed for Gluten developer, cpr will never be used in Gluten I believe, but cpr introduce lots extra deps when compile it, and these extra deps could not be downloaded easily for us due to network issue.

We must consider the developers who work with limited network environments on their build machines, I do not want to spend a lot of time dealing with network issues.

cc @zhouyuan

zhouyuan:
@Yohahaha actually on a clean centos 7 docker image gluten build will fail w/o this patch, but your inputs makes sense to me - developer efficiency is very important for community growing. Let me do more clean up and see what is the best configuration for this, our goal is to minimize the private patch for meta/velox as I know in some custom env they also have a modified velox. 
I saw there's a patch from jiayi to enable offline build for curl/cpr, would this work for a temporary fix? 
https://github.com/facebookincubator/velox/pull/7853

thanks, -yuan

Yohahaha:
> I saw there's a patch from jiayi to enable offline build for curl/cpr, would this work for a temporary fix?

thank you for reply!

I have discussed with jiayi offline, cpr will build zlib by default, set(CURL_ZLIB OFF) could disable build zib. I will try submit to meta/velox.

PR #4268, opened 2024-01-03
[VL]Update individual parameters for Spark and Velox

weiting-chen:

Update README only

## How was this patch tested?

Change README only


PHILO-HE (approved):
Thanks!

PR #4276, opened 2024-01-04
[VL] Remove DeserializationStream from TaskResource when close() is called

marin-ma:
Below code can cause heavy GC, especially when AQE is on with large number of coalesced partitions.
Assuming 8000 shuffle mapper tasks, 8000 shuffle partitions and 200 reducer tasks (mapper output coalesced, each reducer reads 8000/200 = 40 partitions from each mapper), 8 tasks per executor.
There will be ~ 40 partitions * 8000 mappers * 8 tasks = 2560000 objects being added to `TaskResource` per executor JVM.
```
      TaskResources.addRecycler(s"ShuffleReaderDeserializationStream_${wrappedOut.getId}", 50) {
        this.close()
      }
```

image

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
@@ -94,8 +94,24 @@ object TaskResources extends TaskListener with Logging {
     })
   }
 
+  def addRecycler(id: String, name: String, prio: Int)(f: => Unit): Unit = {
+    addResource(
+      id,
+      new TaskResource {
+        override def release(): Unit = f
+
+        override def priority(): Int = prio
+
+        override def resourceName(): String = name
+      })
+  }
+
   def addResource[T <: TaskResource](id: String, resource: T): T = {
     getTaskResourceRegistry().addResource(id, resource)
+  }
+
+  def removeResource(id: String): Unit = {
+    getTaskResourceRegistry().removeResource(id)
   }
 
   def addResourceIfNotRegistered[T <: TaskResource](id: String, factory: () => T): T = {
@@ -177,8 +193,8 @@ object TaskResources extends TaskListener with Logging {
 // thread safe
 class TaskResourceRegistry extends Logging {
   private val sharedUsage = new SimpleMemoryUsageRecorder()
-  private val resources = new util.HashMap[String, TaskResource]()
   type TaskResourceWithOrdering = (TaskResource, Int)
+  private val resources = new util.HashMap[String, TaskResourceWithOrdering]()
   // A monotonically increasing accumulator to specify the task resource ordering
   // when inserting into queue
   private var resourceOrdering = 0
@@ -197,8 +213,9 @@ class TaskResourceRegistry extends Logging {
     })
 
   private def addResource0(id: String, resource: TaskResource): Unit = synchronized {
-    resources.put(id, resource)
-    resourcesPriorityQueue.add((resource, resourceOrdering))
+    val resourceWithOrdering = (resource, resourceOrdering)
+    resources.put(id, resourceWithOrdering)
+    resourcesPriorityQueue.add(resourceWithOrdering)
     resourceOrdering += 1
   }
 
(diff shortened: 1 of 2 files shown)


Yohahaha:
we may remove all TaskResource when close it from TaskResources.

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
```diff
@@ -245,10 +262,19 @@ class TaskResourceRegistry extends Logging {
...
   private[util] def getSharedUsage(): SimpleMemoryUsageRecorder = synchronized {
     sharedUsage
   }
+
+  private[util] def removeResource(id: String): Unit = synchronized {
+    if (!resources.containsKey(id)) {
+      throw new IllegalArgumentException(
+        String.format("TaskResource with ID %s is not registered", id))
+    }
+    resourcesPriorityQueue.remove(resources.get(id))
```
zhztheplayer, line 277: Looks to be that this is a O(n) call for `PriorityQueue#remove(...)`, I am not sure. Could you confirm? If it's uneasy to optimize, probably we can use a `TreeSet` instead.

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
```diff
@@ -198,18 +206,23 @@ private class ColumnarBatchSerializerInstance(
...
-          if (numBatchesTotal > 0) {
-            readBatchNumRows.set(numRowsTotal.toDouble / numBatchesTotal)
-          }
-          numOutputRows += numRowsTotal
-          wrappedOut.close()
-          byteIn.close()
-          if (cb != null) {
-            cb.close()
-          }
+          TaskResources.removeResource(resourceId)
```
zhztheplayer, line 209: So the call to `#removeResource` may happen inside a call to `TaskResourceRegistry#releaseAll`. Should we double-check whether the code brings inconsistency?

To simplify the procedure, I'd suggest returning a `TaskResourceHandle[T]` (or something like that) from TaskResource's `addResource...` `addRecycler...` APIs, then caller could keep the handle and manually release the resource and remove it from registry to make it GC-able by calling `TaskResourceHandle#release` API when needed. Then we don't need to add a new property `ID` to task resource definition.

zhztheplayer (commented):
Thanks in advance!

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
marin-ma: > So the call to `#removeResource` may happen inside a call to `TaskResourceRegistry#releaseAll`. Should we double-check whether the code brings inconsistency?

As `TaskResourceRegistry#releaseAll` is called at the end of taks, should calling 'close` explicitly always be earlier than the former one? Based on my understanding, calling `close` after `TaskResourceRegistry#releaseAll` should be considered as invalid code path.

marin-ma: I see. It's possible that releasing other `TaskResource` from `TaskResourceRegistry#releaseAll` calls this `close` function.

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
marin-ma: Thanks for advice. I agree that we use `TreeSet` here. However, an O(n) complexity shouldn't become a hotspot, as the value of n won't be large if we prevent caching lots of objects. If n is large, then the GC can be a more serious problem.

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
zhztheplayer: Sorry I got it wrong. You are right that it's not possible to be inside another operation on `TaskResourceRegistry` since you are calling `close0` rather than `close` in the recycler.

gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
zhztheplayer: Thank you for your explanation. It makes sense to me except that I think a structure like `PriorityQueue` should be used for specific reasons. If we'll introduce inefficient operations then a user may question us why we are using it. Would you agree on that?

zhztheplayer (approved):
I think current comments are all minor. I got something wrong in the last review. LGTM.

PR #4280, opened 2024-01-04
[VL] Enable native write files by default for Spark3.4

ulysses-you:

This pr enables native write files by default for Spark3.4 with Velox backend.

## How was this patch tested?

Pass CI

enable tests:

```
GlutenStatisticsCollectionSuite
 column stats collection for null columns
 store and retrieve column stats in different time zones
 "SPARK-38140: describe column stats (min, max) for timestamp column: desc results should be consistent with the written value if writing and desc happen in the same time zone")
 "SPARK-38140: describe column stats (min, max) for timestamp column: desc should show different results if writing in UTC and desc in other time zones
 Gluten - store and retrieve column stats in different time zones

GlutenSQLQuerySuite
 SPARK-26709: OptimizeMetadataOnlyQuery does not handle empty records correctly

GlutenFileSourceSQLInsertTestSuite
 SPARK-30844: static partition should also follow StoreAssignmentPolicy

GlutenParquetV2QuerySuite
 SPARK-11997 parquet with null partition values

GlutenParquetV1QuerySuite
 SPARK-11997 parquet with null partition values

GlutenParquetIOSuite
 SPARK-7837 Do not close output writer twice when commitTask() fails

GlutenInsertSuite
 SPARK-35106: Throw exception when rename custom partition paths returns false
```

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/VeloxTPCHSuite.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/VeloxTPCHSuite.scala
@@ -243,6 +243,11 @@ class VeloxTPCHV2BhjSuite extends VeloxTPCHSuite {
 }
 
 class VeloxPartitionedTableTPCHSuite extends VeloxTPCHSuite {
+
+  override protected def sparkConf: SparkConf = {
+    super.sparkConf.set("spark.memory.offHeap.size", "4g")
+  }
+
   override protected def createTPCHNotNullTables(): Unit = {
     TPCHTableDataFrames = TPCHTables.map {
       table =>
--- a/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteSuite.scala
@@ -21,7 +21,6 @@ import io.glutenproject.sql.shims.SparkShimLoader
 import io.glutenproject.utils.FallbackUtil
 
 import org.apache.spark.SparkConf
-import org.apache.spark.sql.functions.lit
 
 import org.junit.Assert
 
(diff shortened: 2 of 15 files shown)


backends-velox/src/test/scala/io/glutenproject/execution/VeloxTPCHSuite.scala
```diff
@@ -243,6 +243,11 @@ class VeloxTPCHV2BhjSuite extends VeloxTPCHSuite {
 }
 
 class VeloxPartitionedTableTPCHSuite extends VeloxTPCHSuite {
+
+  override protected def sparkConf: SparkConf = {
+    super.sparkConf.set("spark.memory.offHeap.size", "4g")
```
ulysses-you, line 248: this change is due to, velox dynamiac partition write requires more memory as it would create writer instance for each partition value.

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/GlutenSQLQueryTestSuite.scala
```diff
@@ -254,7 +254,7 @@ class GlutenSQLQueryTestSuite
     "datetime-parsing.sql",
     "datetime-special.sql",
 //    "decimalArithmeticOperations.sql",
-    "describe-part-after-analyze.sql",
+//    "describe-part-after-analyze.sql",
```
ulysses-you, line 257: this test failed due to 
```
Expected "...son] bytes, 3 rows  []	                   ...", but got "...son] bytes, 3 rows  [ ]	                   ..." Result did not match for query #6
```

it seems the number of spaces changed

JkSelf:
Do we need to enable the failed unit tests firstly and then enable native write by default for spark34? https://github.com/oap-project/gluten/issues/4110

ulysses-you:
Wait for velox rebase to enable write null partition values tests

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteSuite.scala
```diff
@@ -114,15 +113,4 @@ class VeloxParquetWriteSuite extends VeloxWholeStageTransformerSuite {
       Assert.assertTrue(FallbackUtil.hasFallback(df.queryExecution.executedPlan))
     }
   }
-
-  ignore("parquet write with empty dataframe") {
```
ulysses-you, line 118: this case is already covered by Spark test

ulysses-you:
cc @JkSelf the main tests in https://github.com/oap-project/gluten/issues/4110 are passed

JkSelf (approved):
LGTM. Thanks.

Issue #4266, opened 2024-01-03
[VL] Support cross join type with Merge join and Hash join


Surbhi-Vijay:
### Description

A join gets converted to hash join or merge join only when there are equi join conditions.

Consider query `select * from Tbl1 cross join Tbl2 on Tbl1.c1 = Tbl2.c1` This join will be converted to hash join or merge join depending on data type and size. This query is equivalent to `select * from Tbl1 inner join Tbl2 on Tbl1.c1 = Tbl2.c1`.

Cross join with non equi join conditions will never get converted to hash or merge join. Considering these facts cross join with equi join conditions can safely be added to hash/merge joins. 

There will only be need to add it in supported join type and it can be mapped to `JoinRel.JoinType.JOIN_TYPE_INNER` in substrait.

Surbhi-Vijay:
There is `TODO` in code to support it with `crossRel` but this will not be required.
```
object SubstraitUtil {
  def toSubstrait(sparkJoin: JoinType): JoinRel.JoinType = sparkJoin match {
    case Inner =>
      JoinRel.JoinType.JOIN_TYPE_INNER
    case FullOuter =>
      JoinRel.JoinType.JOIN_TYPE_OUTER
    case LeftOuter | RightOuter =>
      // The right side is required to be used for building hash table in Substrait plan.
      // Therefore, for RightOuter Join, the left and right relations are exchanged and the
      // join type is reverted.
      JoinRel.JoinType.JOIN_TYPE_LEFT
    case LeftSemi =>
      JoinRel.JoinType.JOIN_TYPE_LEFT_SEMI
    case LeftAnti =>
      JoinRel.JoinType.JOIN_TYPE_ANTI
    case _ =>
      // TODO: Support cross join with Cross Rel
      // TODO: Support existence join
      JoinRel.JoinType.UNRECOGNIZED
  }
}

```

Cross join with non equi join conditions will get converted to either `BroadcastNestedLoopJoinExec` or `CartesianProductExec`. Both of these will be supported with `crossRel`.

Surbhi-Vijay:
@zhouyuan @rui-mo Can you please check this issue and let me know if you see any issue with it?

zhouyuan:
@Surbhi-Vijay I think you are correct, the code is introduced by CK backend - maybe something different in CK impl

PR #4281, opened 2024-01-04
[GLUTEN-4266][VL] Support cross join type with Merge join and Hash join

Surbhi-Vijay:

Added cross join type in hash and merge join:

Cross join type will only get converted to hash or merge join when there are equi join conditions. Cross join with equi join conditions is actually inner join only. Ideally, users should not even write such queries.

Currently, queries like `select * from TBL1 cross join TBL2 on TBL1.c1 == TBL2.c1` falls back to Spark. which can easily be supported by Gluten by converting cross join substrait inner join type.

(Fixes: \#4266)

## How was this patch tested?
Tested using dummy unit tests

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/execution/ShuffledHashJoinExecTransformer.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/execution/ShuffledHashJoinExecTransformer.scala
@@ -118,7 +118,7 @@ case class ShuffledHashJoinExecTransformer(
   }
 
   override protected lazy val substraitJoinType: JoinRel.JoinType = joinType match {
-    case Inner =>
+    case _: InnerLike =>
       JoinRel.JoinType.JOIN_TYPE_INNER
     case FullOuter =>
       JoinRel.JoinType.JOIN_TYPE_OUTER
@@ -143,7 +143,6 @@ case class ShuffledHashJoinExecTransformer(
     case LeftAnti =>
       JoinRel.JoinType.JOIN_TYPE_ANTI
     case _ =>
-      // TODO: Support cross join with Cross Rel
       JoinRel.JoinType.UNRECOGNIZED
   }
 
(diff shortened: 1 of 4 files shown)


backends-velox/src/main/scala/io/glutenproject/execution/ShuffledHashJoinExecTransformer.scala
```diff
@@ -172,24 +173,8 @@ case class GlutenBroadcastHashJoinExecTransformer(
...
-    case LeftSemi | ExistenceJoin(_) =>
-      JoinRel.JoinType.JOIN_TYPE_LEFT_SEMI
-    case LeftAnti =>
-      JoinRel.JoinType.JOIN_TYPE_ANTI
-    case _ =>
-      // TODO: Support cross join with Cross Rel
-      JoinRel.JoinType.UNRECOGNIZED
-  }
+  override protected lazy val substraitJoinType: JoinRel.JoinType =
+    SubstraitUtil.toSubstrait(joinType)
```
rui-mo, line 177: We exchange tables sometimes to adjust the build side of suffled hash join for performance optimization. For BHJ and SMJ, we may need to keep the original left and right tables.

rui-mo (commented):
> Currently, queries like select * from TBL1 cross join TBL2 on TBL1.c1 == TBL2.c1 falls back to Spark.

Thanks for supporting this case. Shall we add a unit test maybe in [TestOperator](https://github.com/oap-project/gluten/blob/main/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala) and check the query plan to make sure it won't fallback?

backends-velox/src/main/scala/io/glutenproject/execution/ShuffledHashJoinExecTransformer.scala
Surbhi-Vijay: Reverted this change. Instead changed `Inner` to `InnerLike` which will work for both `Inner` & `Cross` joins.

rui-mo (approved):
Thanks.

PR #4285, opened 2024-01-05
[VL] Build & install spark-3.4.2 in docker and correct spark version in pom.xml

PHILO-HE:

We have made spark-3.4.2 the official support version at https://github.com/oap-project/gluten/commit/8d46bec8105ac02bbcdb05ff21e1c57c3fab1365. Therefore, we need to build & install spark-3.4.2 inside docker for test to use.

## How was this patch tested?

Exiting tests.


PHILO-HE:
@JkSelf, could you take a review? This is just a small follow-up work for upgrading to spark-3.4.2.

JkSelf (approved):
Thanks for your great work. LGTM. Thanks.

PR #4290, opened 2024-01-05
[VL] Remove native write files sort and project

ulysses-you:

This pr adds a new rule to remove the V1Writes added sort and Empty2Null in project. The reason is that:

- Velox table write does not require the data is ordered by partition columns
- Velox table write treats empty string as the `__HIVE_DEFAULT_PARTITION__`, so we do not need `Empty2Null`

After this pr, we can transform the whole write files for dynamic partition write.

## How was this patch tested?

add tests

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/RemoveNativeWriteFilesSortAndProject.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/RemoveNativeWriteFilesSortAndProject.scala
@@ -1,0 +1,121 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.extension
+
+import io.glutenproject.GlutenConfig
+import io.glutenproject.execution.{ProjectExecTransformer, SortExecTransformer, WriteFilesExecTransformer}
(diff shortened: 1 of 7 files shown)


ulysses-you:
cc @JkSelf @zhouyuan thank you

gluten-core/src/main/scala/io/glutenproject/extension/RemoveNativeWriteFilesSortAndProject.scala
```diff
@@ -0,0 +1,100 @@
...
+import io.glutenproject.GlutenConfig
+import io.glutenproject.execution.{SortExecTransformer, WriteFilesExecTransformer}
+
+import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, AttributeMap, AttributeSet, SortOrder}
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.{ProjectExec, SortExec, SparkPlan}
+
+/**
+ * This rules remove the `V1Writes` added sort and project because:
+ *   - Velox table write does not require the data is ordered by partition columns
```
JkSelf, line 28: @ulysses-you It appears that Velox only performs sorting by the partition columns for bucket writes, but not for partition writes [here](https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/HiveDataSink.cpp#L334-L350). Therefore, if we remove the sort operator in Spark, how can we ensure that the data is ordered by the partition columns in the same way as vanilla Spark?"

ulysses-you: Velox will create new writer instance per partition which is similar with the feature of Spark `spark.sql.maxConcurrentOutputFileWriters`, so it does not require the partition values are ordered.

One difference things with Spark is that, Velox does not support fallback to sorting mode if there are so many partitions in a task which may cause OOM.

JkSelf: @ulysses-you If the `spark.sql.maxConcurrentOutputFileWriters` parameter is set to 0, which means concurrent writer is disabled, the data written by Velox will be unordered, while the data written by Spark will be ordered. Do we need to consider not removing the Sort operator in this case?

ulysses-you: The data is also unordered in each partition with vanilla Spark. Vanilla Spark only sorts the data by partition column, so the actually data column is unordered. The ordering of partition column is useless if we create new writer instance for each partition value.

JkSelf (approved):
Thanks for your great work. LGTM. Thanks.

PR #4291, opened 2024-01-05
[VL] remove validate for timestamp in sortRel

zhli1142015:

This is added before because timestamp is not fully supported, now looks we can remove this check.

## How was this patch tested?

Existing UT.


rui-mo (approved):
Thanks.

PR #4292, opened 2024-01-05
[VL] A couple of code optimizations to TaskResources

zhztheplayer:
Following #4276:

1. In resources' priority table, replace `PriorityQueue` with `LinkedHashSet` for efficient removing operation; (`PriorityQueue` was good until we added API to remove a single instance in #4276)
2. Make `DeserializationStream` a resource type to replace advanced use case of `addRecycler`
3. Mirror relevant changes to `VeloxCelebornColumnarBatchSerializer`

Update:

The patch will also add restriction to user defined `release()` method to disallow it from accessing task registry recursively, which may cause internal inconsistency leading to invalid state or memory leak.

### DIFF
--- a/gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarBatchSerializer.scala
+++ b/gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarBatchSerializer.scala
@@ -83,9 +83,9 @@ private class CelebornColumnarBatchSerializerInstance(
       }
     val compressionCodecBackend =
       GlutenConfig.getConf.columnarShuffleCodecBackend.orNull
+    val jniWrapper = ShuffleReaderJniWrapper.create()
     val batchSize = GlutenConfig.getConf.maxBatchSize
-    val handle = ShuffleReaderJniWrapper
-      .create()
+    val handle = jniWrapper
       .make(
         cSchema.memoryAddress(),
         nmm.getNativeInstanceHandle,
@@ -98,7 +98,7 @@ private class CelebornColumnarBatchSerializerInstance(
     // was used to create all buffers read from shuffle reader. The pool
     // should keep alive before all buffers to finish consuming.
     TaskResources.addRecycler(s"CelebornShuffleReaderHandle_$handle", 50) {
-      ShuffleReaderJniWrapper.create().close(handle)
+      jniWrapper.close(handle)
       cSchema.release()
       cSchema.close()
       allocator.close()
--- a/gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarBatchSerializer.scala
+++ b/gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarBatchSerializer.scala
@@ -21,7 +21,7 @@ import io.glutenproject.exec.Runtimes
 import io.glutenproject.memory.arrowalloc.ArrowBufferAllocators
 import io.glutenproject.memory.nmm.NativeMemoryManagers
 import io.glutenproject.utils.ArrowAbiUtil
-import io.glutenproject.vectorized.{ColumnarBatchOutIterator, GeneralOutIterator, JniByteInputStream, JniByteInputStreams, ShuffleReaderJniWrapper}
+import io.glutenproject.vectorized._
 
 import org.apache.spark.SparkEnv
 import org.apache.spark.internal.Logging
@@ -31,7 +31,7 @@ import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.utils.SparkSchemaUtil
 import org.apache.spark.sql.vectorized.ColumnarBatch
-import org.apache.spark.util.TaskResources
+import org.apache.spark.util.{TaskResource, TaskResources}
 
 import org.apache.arrow.c.ArrowSchema
 import org.apache.arrow.memory.BufferAllocator
@@ -39,6 +39,8 @@ import org.apache.celeborn.client.read.CelebornInputStream
 
 import java.io._
 import java.nio.ByteBuffer
+import java.util.UUID
+import java.util.concurrent.atomic.AtomicBoolean
 
 import scala.reflect.ClassTag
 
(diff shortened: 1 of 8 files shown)


zhztheplayer:
@marin-ma

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
```diff
@@ -124,106 +123,102 @@ private class ColumnarBatchSerializerInstance(
...
-        throw new UnsupportedOperationException
-      }
+    override def readObject[T: ClassTag](): T = {
+      // This method is never called by shuffle code.
+      throw new UnsupportedOperationException
+    }
 
-      override def close(): Unit = {
-        if (!isClosed) {
-          TaskResources.removeResource(resourceId)
```
marin-ma, line 209: If remove this call, when will `releaseResource` be called?

marin-ma:
As `Treeset` is not the replacement data structure, please also update the PR description. Thanks!

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
marin-ma: I'm still confused. Shouldn't somewhere explicitly call `releaseResource`? If nowhere calls `releaseResource`, when will the `TaskResource` be removed from the registry?

zhztheplayer: @marin-ma We release all corresponding resources when the Spark task is getting stopped. https://github.com/oap-project/gluten/blob/3fe4c959e07739ac142433485691f92a8a75bfd1/gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala#L171

By adding `releaseResource(id: String)` we can releasing a single registered resource by ID.

So after then `releaseAll` could be considered batch version of `releaseResource`.

marin-ma: The idea of https://github.com/oap-project/gluten/pull/4276 is to remove the single resource immediately when `close` is called, to avoid caching lots of object references. We shouldn't rely on `releaseAll` to release the object at task end, because with lots of objects being cached, GC will take a very long time.

zhztheplayer: Thanks for the explanation. And IIUC This PR doesn't alter the core logic you mentioned above. 

See `remove` will be finally called through `releaseResource`:

https://github.com/oap-project/gluten/pull/4292/files#diff-8f21fe60b00b9ff709fad96d28ddb7a06c7f5e20ae7929f9216844de663e636aR103-R105

https://github.com/oap-project/gluten/pull/4292/files#diff-8f21fe60b00b9ff709fad96d28ddb7a06c7f5e20ae7929f9216844de663e636aR232-R233

zhztheplayer: @marin-ma , I double-check the PR and noticed it was my mistake not adding `releaseResource` into deserializer code. It should be fixed in latest commit.

Thanks for catching!

gluten-data/src/main/scala/io/glutenproject/vectorized/ColumnarBatchSerializer.scala
```diff
@@ -124,106 +124,110 @@ private class ColumnarBatchSerializerInstance(
...
-          close0()
-          isClosed = true
-        }
+    override def close(): Unit = {
```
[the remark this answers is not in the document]
zhztheplayer: The call flow is as follow:

`close` calls `releaseResource` which calls `release` which calls `tryClose` which checks `isClosed` state then calls `close0` if needed.

So I assume the current code is fine. I feel that we may be able to add some kind of testing here to guard against the GC issue, or to make futher effort to simplify code. It's a bit complicated as of now.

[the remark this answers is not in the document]
marin-ma: The `close` can be called from other code paths more than once, not this code path, which leads to the CI failure. I met the same CI failure before :(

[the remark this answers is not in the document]
zhztheplayer: Then we may require for another guard to avoid calling `releaseResource` twice. I'll see we can develop a optimal way.

zhztheplayer:
@marin-ma

gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarBatchSerializer.scala
```diff
@@ -96,128 +98,155 @@ private class CelebornColumnarBatchSerializerInstance(
...
+        return
       }
+      // Would remove the resource object from registry to lower GC pressure.
+      TaskResources.releaseResource(resourceId)
+    }
 
-      override def readObject[T: ClassTag](): T = {
-        // This method is never called by shuffle code.
-        throw new UnsupportedOperationException
+    override def release(): Unit = {
```
marin-ma, line 230: What's the benefit of implementing `TaskResource` interface? The code looks a bit confusing, that `close()` calls `releaseResource()` while `release()` calls `close0()`.

zhztheplayer: > The code looks a bit confusing, that close() calls releaseResource() while release() calls close0().

Yes this is a little bit implicit to me and be worth to be discussed.

Let's summarize the core change first. Previous code:

``` scala
  if (!isClosed) {
    TaskResources.removeResource(resourceId)
    close0()
    isClosed = true
  }
}
```
and the current code:
```scala
  if (!closeCalled.compareAndSet(false, true)) {
    return
  }
  // Would remove the resource object from registry to lower GC pressure.
  TaskResources.releaseResource(resourceId)
}
```

So basically in the patch I consolidated the two operations `removeResource` and `close0`, into a single API `releaseResource`. You can consider `removeResource` + `close0` a common pattern since `removeResource` omits running `release()` method of the registered TaskResource. So after calling `removeResource` you should call `close0` which does exactly the same with resource's `release()` method. Since the code here can be a common pattern, I added `releaseResource` for better encapsulation. Also, exposing `removeResouce` API may lead to unexpected usages, e.g. removing a resource and go around the resource clean-up procedure, we should only allow this kind of usage for certain strong reasons.

marin-ma (approved):

PR #4312, opened 2024-01-08
[VL] Enable FlushableHashAggregateRule

zhztheplayer:
A follow-up to https://github.com/oap-project/gluten/pull/4130

Enable FlushableHashAggregateRule to convert eligible hash agg to flushable agg.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/catalyst/FlushableHashAggregateRule.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/catalyst/FlushableHashAggregateRule.scala
@@ -16,7 +16,7 @@
  */
 package org.apache.spark.sql.catalyst
 
-import io.glutenproject.execution.{FlushableHashAggregateExecTransformer, RegularHashAggregateExecTransformer}
+import io.glutenproject.execution.{FlushableHashAggregateExecTransformer, ProjectExecTransformer, RegularHashAggregateExecTransformer}
 
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.expressions.aggregate.{Partial, PartialMerge}
@@ -27,8 +27,6 @@ import org.apache.spark.sql.execution.exchange.ShuffleExchangeLike
 /**
  * To transform regular aggregation to intermediate aggregation that internally enables
  * optimizations such as flushing and abandoning.
- *
- * Currently not in use. Will be enabled via a configuration after necessary verification is done.
  */
 case class FlushableHashAggregateRule(session: SparkSession) extends Rule[SparkPlan] {
   override def apply(plan: SparkPlan): SparkPlan = plan.transformUp {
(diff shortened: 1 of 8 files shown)


backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
```diff
@@ -700,7 +701,26 @@ abstract class VeloxAggregateFunctionsSuite extends VeloxWholeStageTransformerSu
   }
 }
 
-class VeloxAggregateFunctionsDefaultSuite extends VeloxAggregateFunctionsSuite {}
+class VeloxAggregateFunctionsDefaultSuite extends VeloxAggregateFunctionsSuite {
+  override protected def sparkConf: SparkConf = {
+    super.sparkConf
+      // To test flush behaviors, set low flush threshold to ensure flush happens.
+      .set(GlutenConfig.VELOX_FLUSHABLE_PARTIAL_AGGREGATION_ENABLED.key, "false")
```
rui-mo, line 708: Unmatched comment and code.

backends-velox/src/main/scala/org/apache/spark/sql/catalyst/FlushableHashAggregateRule.scala
```diff
@@ -27,30 +27,69 @@ import org.apache.spark.sql.execution.exchange.ShuffleExchangeLike
...
  * To transform regular aggregation to intermediate aggregation that internally enables
  * optimizations such as flushing and abandoning.
- *
- * Currently not in use. Will be enabled via a configuration after necessary verification is done.
  */
 case class FlushableHashAggregateRule(session: SparkSession) extends Rule[SparkPlan] {
   override def apply(plan: SparkPlan): SparkPlan = plan.transformUp {
     case shuffle: ShuffleExchangeLike =>
       // If an exchange follows a hash aggregate in which all functions are in partial mode,
       // then it's safe to convert the hash aggregate to intermediate hash aggregate.
```
rui-mo, line 35: intermediate ->flushable?

backends-velox/src/main/scala/org/apache/spark/sql/catalyst/FlushableHashAggregateRule.scala
```diff
@@ -27,30 +27,69 @@ import org.apache.spark.sql.execution.exchange.ShuffleExchangeLike
...
+    if (!plan.isInstanceOf[ProjectExecTransformer]) {
+      return None
+    }
+    val proj = plan.asInstanceOf[ProjectExecTransformer]
+    val child = proj.child
+    if (!child.isInstanceOf[RegularHashAggregateExecTransformer]) {
+      return None
+    }
+    val agg = child.asInstanceOf[RegularHashAggregateExecTransformer]
+    if (!agg.aggregateExpressions.forall(p => p.mode == Partial || p.mode == PartialMerge)) {
```
rui-mo, line 77: Can the correctness of result be ensured with this rule? E.g. if it flushes, the result of distinct count could be wrong?

zhztheplayer: So far the rule would work correctly. The principle is that if a partial aggregation is directly sending data to exchange from mapper side, then a single reducer could always receive partial-aggregated data from several mappers. So enabling flushing on mapper side's agg would act as if adding some mapper instances. 

However there might be some corner cases we should consider, e.g. if partial aggregation is already handing data that is partitioned by the exchange keys, and we may enhance the rule later on for the corner cases. But so far the rule works well on existing tests.

rui-mo (approved):

PR #4318, opened 2024-01-08
[VL] Support write empty rdd and empty iterator

ulysses-you:

This pr makes native write support write empty rdd and empty iterator. Vanilla Spark will generate an empty file for metadata only if write empty. Gluten should respect the behavior.

## How was this patch tested?

Enable related spark tests


ulysses-you:
cc @JkSelf this pr is to solve the write empty data issue

JkSelf (approved):
Thanks for your fix. LGTM except two small comments.

JkSelf:
@ulysses-you So we don't need to upstream this PR https://github.com/oap-project/velox/pull/370 to velox community. Right?

ulysses-you:
@JkSelf yes, we do not need the Velox code change any more.

JkSelf (approved):
LGTM. Thanks.

PR #4322, opened 2024-01-08
[CORE] add missing boost-locale dependency in VCPKG. Otherwise fail to build in ubuntu22.04/Docker.

HannanKan:
bugfix: missing boost-locale dependency in vcpkg which lead to compiling failure in ubuntu22.04/Docker

image

## How was this patch tested?
manual test: with this pr, compile gluten with `./dev/buildbundle-veloxbe.sh --enable_vcpkg=ON` command will succeed. Otherwise, no more error will be thrown.


zhouyuan (approved):
👍 
indeed this is missed

zhouyuan:
@HannanKan no worries, the failure is not related

thanks, -yuan

Issue #4327, opened 2024-01-09
[CH] A lot of small blocks are generated by `StreamingAggregatingTransform`


lgbo-ustc:
### Backend

CH (ClickHouse)

### Bug description

[Expected behavior] and [actual behavior].

Enable streaming aggregating, a lot of small blocks are generated by `StreamingAggregatingTransform`
![image](https://github.com/oap-project/gluten/assets/2599447/e6a75b33-5966-46ae-a142-580c9a99d96d)

### Spark version

None

PR #4328, opened 2024-01-09
[GLUTEN-4327][CH] Avoid generating a lot of small blocks in `StreamingAggregatingTransform`

lgbo-ustc:

Fixes: #4327 

## How was this patch tested?

unit tests

### DIFF
--- a/cpp-ch/local-engine/Parser/AggregateRelParser.cpp
+++ b/cpp-ch/local-engine/Parser/AggregateRelParser.cpp
@@ -256,6 +256,7 @@ void AggregateRelParser::addMergingAggregatedStep()
     bool enable_streaming_aggregating = getContext()->getConfigRef().getBool("enable_streaming_aggregating", true);
     if (enable_streaming_aggregating)
     {
+        params.group_by_two_level_threshold = settings.group_by_two_level_threshold;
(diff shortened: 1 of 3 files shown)


cpp-ch/local-engine/Parser/AggregateRelParser.cpp
```diff
@@ -289,14 +289,16 @@ void AggregateRelParser::addAggregatingStep()
...
         // Disable spilling to disk.
+        // We let group_by_two_level_threshold_bytes = 0, avoid converting the hash table into two level
+        // in Aggregator when the hash table is not large enough.
         Aggregator::Params params(
             grouping_keys,
             aggregate_descriptions,
             false,
             settings.max_rows_to_group_by,
             settings.group_by_overflow_mode,
             settings.group_by_two_level_threshold,
```
zhanglistar, line 300: what about hash table size?

lgbo-ustc: The size limit of hash table does not cause this problem. `Aggregator` will convert a hash table into a two level one when the memory usage the hash table  reaches the limit. But the memory usage statistics is not right

liuneng1994 (approved):
LGTM

PR #4329, opened 2024-01-09
[VL] Improve native write files fallback

ulysses-you:

If people set maxRecordsPerFile, then we should fallback to vanilla Spark write files. This pr also moves the write files transformer validation to velox backend.

## How was this patch tested?

enable test

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -111,7 +111,7 @@ object BackendSettings extends BackendSettingsApi {
             "StructType as element in ArrayType"
           case StructField(_, arrayType: ArrayType, _, _)
               if arrayType.elementType.isInstanceOf[ArrayType] =>
-            "A rrayType as element in ArrayType"
+            "ArrayType as element in ArrayType"
           case StructField(_, mapType: MapType, _, _) if mapType.keyType.isInstanceOf[StructType] =>
             "StructType as Key in MapType"
           case StructField(_, mapType: MapType, _, _)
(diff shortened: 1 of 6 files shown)


ulysses-you:
cc @JkSelf thank you

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -109,7 +111,7 @@ object BackendSettings extends BackendSettingsApi {
             "StructType as element in ArrayType"
           case StructField(_, arrayType: ArrayType, _, _)
               if arrayType.elementType.isInstanceOf[ArrayType] =>
-            "A rrayType as element in ArrayType"
+            "ArrayType as element in ArrayType"
```
JkSelf, line 114: @ulysses-you Thanks for your fix.

JkSelf (commented):
Thanks for your improvement. LGTM except one small comments.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -157,24 +163,45 @@ object BackendSettings extends BackendSettingsApi {
...
+        Some("Unsupported native write: bucket write is not supported.")
+      } else {
+        None
+      }
+    }
+
+    validateCompressionCodec()
+      .orElse(validateFileFormat())
+      .orElse(validateFieldMetadata())
+      .orElse(validateDateTypes(fields))
```
JkSelf, line 202: Can we also remove the fields parameters in validateDateTypes method?

ulysses-you: sure, removed

JkSelf (approved):
LGTM. Thanks.

Issue #4326, opened 2024-01-09
[CH] java.lang.UnsupportedOperationException: Not supported operator ShuffleQueryStage for BroadcastRelation


zhanglistar:
SQL: d_669_0
Driver log:
```
24/01/09 09:13:41,887 INFO [dag-scheduler-event-loop] YarnScheduler: Killing all running tasks in stage 40: Stage cancelled
24/01/09 09:13:41,887 ERROR [main] FileFormatWriter: Aborting job ef27cbe5-a6d3-425a-9799-2149acef9d23.
java.lang.UnsupportedOperationException: Not supported operator ShuffleQueryStage for BroadcastRelation
        at io.glutenproject.backendsapi.clickhouse.CHSparkPlanExecApi.createBroadcastRelation(CHSparkPlanExecApi.scala:314) ~[gluten-1.2.0-SNAPSHOT-spark-3.3-jar-with-dependencies.jar:?]
        at org.apache.spark.sql.execution.ColumnarBroadcastExchangeExec.$anonfun$relationFuture$2(ColumnarBroadcastExchangeExec.scala:79) ~[gluten-1.2.0-SNAPSHOT-spark-3.3-jar-with-dependencies.jar:3.3.2]
        at io.glutenproject.utils.Arm$.withResource(Arm.scala:25) ~[gluten-1.2.0-SNAPSHOT-spark-3.3-jar-with-dependencies.jar:?]
        at io.glutenproject.metrics.GlutenTimeMetric$.millis(GlutenTimeMetric.scala:37) ~[gluten-1.2.0-SNAPSHOT-spark-3.3-jar-with-dependencies.jar:?]
        at org.apache.spark.sql.execution.ColumnarBroadcastExchangeExec.$anonfun$relationFuture$1(ColumnarBroadcastExchangeExec.scala:69) ~[gluten-1.2.0-SNAPSHOT-spark-3.3-jar-with-dependencies.jar:3.3.2]
        at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withThreadLocalCaptured$1(SQLExecution.scala:191) ~[spark-sql_2.12-3.3.2.jar:3.3.2]
        at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[?:1.8.0_292]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) ~[?:1.8.0_292]
[...]

PR #4334, opened 2024-01-09
[GLUTEN-4326][CH] Fix unsupported operator ShuffleQueryStage for BroadcastRelation

exmy:

When AQE is enabled but `spark.sql.adaptive.coalescePartitions.enabled` is false, AQE converts shj to bhj in runtime but the `ShuffleQueryStageExec` of shj has been generated, which causes the exception `java.lang.UnsupportedOperationException: Not supported operator ShuffleQueryStage for BroadcastRelation` when `createBroadcastRelation` for it.

(Fixes: \#4326)

## How was this patch tested?

manual tests, ut seems not easy to construct.


zzcclp:
please add a ut

exmy:
> please add a ut

I have attempted to add it, but it's challenging to construct suitable examples.

PR #4335, opened 2024-01-09
[VL]Fallback complex type in native write

JkSelf:

Velox doesn't support nested type encoding when converting the rowVector to arrow data https://github.com/facebookincubator/velox/blob/main/velox/vector/arrow/Bridge.cpp#L795. So we fallback nested type in native write.

## How was this patch tested?

existing tests.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -109,7 +109,7 @@ object BackendSettings extends BackendSettingsApi {
             "StructType as element in ArrayType"
           case StructField(_, arrayType: ArrayType, _, _)
               if arrayType.elementType.isInstanceOf[ArrayType] =>
-            "ArrayType as element in ArrayType"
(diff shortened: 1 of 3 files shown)


JkSelf:
@zhli1142015 @ulysses-you Can you help to review? Thanks.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
```diff
@@ -109,7 +109,7 @@ object BackendSettings extends BackendSettingsApi {
             "StructType as element in ArrayType"
           case StructField(_, arrayType: ArrayType, _, _)
               if arrayType.elementType.isInstanceOf[ArrayType] =>
-            "ArrayType as element in ArrayType"
+            "A rrayType as element in ArrayType"
```
ulysses-you, line 112: unnecessary change

JkSelf: Good catch. I will fix in following PRs. Thanks.

PR #4344, opened 2024-01-10
[VL] Pass partition id to velox functions

zhli1142015:

Pass partition id to velox functions.

## How was this patch tested?

UT.

### DIFF
--- a/cpp/velox/jni/VeloxJniWrapper.cc
+++ b/cpp/velox/jni/VeloxJniWrapper.cc
@@ -110,9 +110,8 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeValidateWithFailu
   gluten::parseProtobuf(planData, planSize, &subPlan);
 
   // A query context used for function validation.
-  velox::core::QueryCtx queryCtx;
   std::unordered_map<std::string, std::string> configs{{velox::core::QueryConfig::kSparkPartitionId, "0"}};
-  queryCtx.testingOverrideConfigUnsafe(std::move(configs));
+  velox::core::QueryCtx queryCtx(nullptr, core::QueryConfig(configs));
   auto pool = gluten::defaultLeafVeloxMemoryPool().get();
   // An execution context used for function validation.
   velox::core::ExecCtx execCtx(pool, &queryCtx);
--- a/cpp/velox/jni/VeloxJniWrapper.cc
+++ b/cpp/velox/jni/VeloxJniWrapper.cc
@@ -111,6 +111,8 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeValidateWithFailu
 
   // A query context used for function validation.
   velox::core::QueryCtx queryCtx;
+  std::unordered_map<std::string, std::string> configs{{velox::core::QueryConfig::kSparkPartitionId, "0"}};
+  queryCtx.testingOverrideConfigUnsafe(std::move(configs));
   auto pool = gluten::defaultLeafVeloxMemoryPool().get();
   // An execution context used for function validation.
   velox::core::ExecCtx execCtx(pool, &queryCtx);
--- a/cpp/velox/jni/VeloxJniWrapper.cc
+++ b/cpp/velox/jni/VeloxJniWrapper.cc
@@ -111,7 +111,7 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeValidateWithFailu
 
   // A query context used for function validation.
   std::unordered_map<std::string, std::string> configs{{velox::core::QueryConfig::kSparkPartitionId, "0"}};
-  velox::core::QueryCtx queryCtx(nullptr, core::QueryConfig(configs));
+  velox::core::QueryCtx queryCtx(nullptr, velox::core::QueryConfig(configs));
   auto pool = gluten::defaultLeafVeloxMemoryPool().get();
   // An execution context used for function validation.
   velox::core::ExecCtx execCtx(pool, &queryCtx);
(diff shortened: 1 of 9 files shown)


zhli1142015:
we should use rdd partition id instead of task partition id for this function.
```
- SPARK-14393: values generated by non-deterministic functions shouldn't change after coalesce or union *** FAILED ***
  Array([0,0], [1,1], [2,2], [3,3]) did not equal Array([0,0], [1,0], [2,1], [3,1]) Values changed after coalesce when codegenFallback=false and wholeStage=false. (DataFrameFunctionsSuite.scala:3591)
```

zhli1142015:
@PHILO-HE @rui-mo could you help review this PR?

cpp/velox/operators/functions/SparkPartitionId.cc
```diff
@@ -0,0 +1,55 @@
...
+
+#include "SparkPartitionId.h"
+#include "config/GlutenConfig.h"
+
+#include "velox/expression/VectorFunction.h"
+#include "velox/type/Type.h"
+
+namespace gluten {
+
+class SparkPartitionId final : public facebook::velox::exec::VectorFunction {
```
rui-mo, line 26: Just wondering why we prefer to implement this function in Gluten. Is there any chance to upstream it in Velox?

rui-mo (commented):
Thanks.

cpp/velox/operators/functions/SparkPartitionId.cc
zhli1142015: I thought this function is tightly bound to gluten. And you see the logic of this function is very simple. The only meaning to support this is to avoid fallback because of this function.

zhli1142015: I can split this to two changes and add it in upstream velox, if you prefer this way, thanks.

PHILO-HE: It seems a bit strange to me. The function just returns the value stored in config.  Do we really need to implement a native version of spark_partition_id?

PHILO-HE:
> we should use rdd partition id instead of task partition id for this function.
> 
> ```
> - SPARK-14393: values generated by non-deterministic functions shouldn't change after coalesce or union *** FAILED ***
>   Array([0,0], [1,1], [2,2], [3,3]) did not equal Array([0,0], [1,0], [2,1], [3,1]) Values changed after coalesce when codegenFallback=false and wholeStage=false. (DataFrameFunctionsSuite.scala:3591)
> ```

How was this issue produced? This spark_partition_id() function should fall back in this test. Could you clarify a bit? Thanks!

PHILO-HE (commented):
Thanks!

cpp/velox/operators/functions/SparkPartitionId.cc
zhli1142015: We observed the fallback with this function, as the implement is very simple, so I think we should avoid this fallback.
Please let me know if you have other suggestion, like we should handle this in gluten side, etc.

zhli1142015:
> > we should use rdd partition id instead of task partition id for this function.
> > ```
> > - SPARK-14393: values generated by non-deterministic functions shouldn't change after coalesce or union *** FAILED ***
> >   Array([0,0], [1,1], [2,2], [3,3]) did not equal Array([0,0], [1,0], [2,1], [3,1]) Values changed after coalesce when codegenFallback=false and wholeStage=false. (DataFrameFunctionsSuite.scala:3591)
> > ```
> 
> How was this issue produced? This spark_partition_id() function should fall back in this test. Could you clarify a bit? Thanks!

This issue is found after i offload spark_partition_id to velox. native spark_partition_id is simple which returns a config as you mentioned. Before the partition id we passed to native is task partition id, not rdd partition id. They are different when there are coalesce  or union operatiors.

cpp/velox/operators/functions/SparkPartitionId.cc
FelixYBW: Yes, let's try to upstream to Velox firstly.

zhli1142015: Sure, get it.

cpp/velox/jni/VeloxJniWrapper.cc
```diff
@@ -111,6 +111,8 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeValidateWithFailu
 
   // A query context used for function validation.
   velox::core::QueryCtx queryCtx;
+  std::unordered_map<std::string, std::string> configs{{velox::core::QueryConfig::kSparkPartitionId, "0"}};
+  queryCtx.testingOverrideConfigUnsafe(std::move(configs));
```
PHILO-HE, line 115: Seems `testingOverrideConfigUnsafe` should be used in test code only.

cpp/velox/jni/VeloxJniWrapper.cc
```diff
@@ -111,6 +111,8 @@ Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_nativeValidateWithFailu
 
   // A query context used for function validation.
   velox::core::QueryCtx queryCtx;
+  std::unordered_map<std::string, std::string> configs{{velox::core::QueryConfig::kSparkPartitionId, "0"}};
```
PHILO-HE, line 114: Do we need to pass this config for validator? Seems if lacked, it doesn't have any impact.

PHILO-HE (commented):
Looks good! Just two comments, could you help check and clarify? BTW, could you let us know how `get_partition_id()` function is used on your side? Directly used in sql? Thanks! cc @zhouyuan

cpp/velox/jni/VeloxJniWrapper.cc
zhli1142015: Updated, thanks.

cpp/velox/jni/VeloxJniWrapper.cc
zhli1142015: Added, thanks

zhli1142015:
> Looks good! Just two comments, could you help check and clarify? BTW, could you let us know how `get_partition_id()` function is used on your side? Directly used in sql? Thanks! cc @zhouyuan

This function is used for data profiling, thanks.

PHILO-HE (approved):
Thanks for your work! cc @rui-mo

PR #4346, opened 2024-01-10
Revert "[GLUTEN-3942][CORE]fix: Change columnar overrides to accept ShuffleExchangeLike"

felipepessoto:
Reverts oap-project/gluten#4167

My initial understanding was that ColumnarShuffleExchangeExec would preserve the behavior of ShuffleExchangeExec/ShuffleExchangeLike, but it actually replaces it.

In this case I think my change is incorrect, and the `case plan: ShuffleExchangeLike` in `replaceWithTransformerPlan` should be `case plan: ShuffleExchangeExec` as before. Because ColumnarShuffleExchangeExec is made specifically to replace ShuffleExchangeExec and not any ShuffleExchangeLike.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -163,7 +163,7 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
    * result columns from the projection.
    */
   private def addProjectionForShuffleExchange(
-      plan: ShuffleExchangeLike): (Int, Partitioning, SparkPlan) = {
+      plan: ShuffleExchangeExec): (Int, Partitioning, SparkPlan) = {
     def selectExpressions(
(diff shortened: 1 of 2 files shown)


marin-ma:
`ColumnarShuffleExchangeExec` is also an implementation of `ShuffleExchangeLike`. In which case would `ColumnarShuffleExchangeExec` fail to replace `ShuffleExchangeLike`?

felipepessoto:
Currently it doesn't make difference because ShuffleExchangeExec is the only ShuffleExchangeLike, but if you create your own class extending ShuffleExchangeLike, or future Spark version has different implementations, ColumnarShuffleExchangeExec won't honor the custom logic.

PHILO-HE:
This revert pr looks good to me.
@marin-ma, do you have any other comment? Thanks!

PHILO-HE (approved):
Thanks for your fix!

marin-ma:
LGTM. Thanks.

PR #4355, opened 2024-01-10
[CORE] Add RemoveKnownFloatingPointNormalized Rule

liujiayi771:

We plan to pull out the pre-project of agg in the logical plan. However, due to some historical reasons, Spark's agg `groupingExpressions` will add `KnownFloatingPointNormalized` to certain specific data types when converting to a physical plan. These expressions need to be pull out into the pre-project during native execution, but we have not pulled out the pre-project of agg in the physical plan. `KnownFloatingPointNormalized` does not actually serve any purpose in Gluten, and during plan conversion, we directly processes its child, so we can remove it.

## How was this patch tested?

ClickHouse backend already have a "knownfloatingpointnormalized" test case in `GlutenClickHouseFileFormatSuite`. Add a "knownfloatingpointnormalized" test case in Velox backend `TestOperator`.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -787,6 +787,7 @@ case class ColumnarOverrideRules(session: SparkSession)
         (spark: SparkSession) => PlanOneRowRelation(spark),
         (_: SparkSession) => FallbackEmptySchemaRelation(),
         (_: SparkSession) => AddTransformHintRule(),
+        (_: SparkSession) => RemoveKnownFloatingPointNormalized,
         (_: SparkSession) => FallbackBloomFilterAggIfNeeded(),
         (_: SparkSession) => TransformPreOverrides(isAdaptiveContext),
         (_: SparkSession) => RemoveNativeWriteFilesSortAndProject(),
--- a/gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
@@ -57,20 +57,19 @@ object RemoveKnownFloatingPointNormalized extends Rule[SparkPlan] {
       // Only remove KnownFloatingPointNormalized when agg is transformable.
       case agg: BaseAggregateExec
           if normalizeExpressionExists(agg.groupingExpressions) &&
-            (validation || supportTransform(plan)) =>
+            (validation || supportTransform(agg)) =>
         val newGroupingExprs = agg.groupingExpressions.toIndexedSeq.map {
           expr =>
             val transformedExpr = expr.transform {
-              case _ @KnownFloatingPointNormalized(_ @NormalizeNaNAndZero(child)) => child
-              case _ @KnownFloatingPointNormalized(_ @If(_ @IsNull(expr), _, _)) => expr
-              case _ @KnownFloatingPointNormalized(_ @ArrayTransform(expr, _)) => expr
-              case other => other
+              case KnownFloatingPointNormalized(NormalizeNaNAndZero(child)) => child
+              case KnownFloatingPointNormalized(If(IsNull(expr), _, _)) => expr
+              case KnownFloatingPointNormalized(ArrayTransform(expr, _)) => expr
             }
-            removeRedundantAlias(transformedExpr)
+            removeRedundantAlias(transformedExpr).asInstanceOf[NamedExpression]
         }
         agg match {
           case hash: HashAggregateExec =>
-            hash.copy(groupingExpressions = newGroupingExprs.asInstanceOf[Seq[NamedExpression]])
+            hash.copy(groupingExpressions = newGroupingExprs)
           case objectHash: ObjectHashAggregateExec =>
             objectHash.copy(groupingExpressions =
               newGroupingExprs.asInstanceOf[Seq[NamedExpression]])
--- a/gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.extension
+
+import io.glutenproject.extension.columnar.TransformHints
+
+import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.catalyst.optimizer.NormalizeNaNAndZero
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.aggregate.{BaseAggregateExec, HashAggregateExec, ObjectHashAggregateExec, SortAggregateExec}
+
+/**
+ * [[KnownFloatingPointNormalized]] in Gluten is useless. When converting plans, it directly takes
+ * its child. We can reverse the KnownFloatingPointNormalized added by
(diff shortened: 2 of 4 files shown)


gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -786,6 +786,7 @@ case class ColumnarOverrideRules(session: SparkSession)
       List(
         (spark: SparkSession) => PlanOneRowRelation(spark),
         (_: SparkSession) => FallbackEmptySchemaRelation(),
+        (spark: SparkSession) => RemoveKnownFloatingPointNormalized(spark),
```
ulysses-you, line 789: We should apply this rule after `AddTransformHintRule` and check that we apply this rule only if we support transform aggregate

gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  override def apply(plan: SparkPlan): SparkPlan = PhysicalPlanSelector.maybe(session, plan) {
+    plan.transform {
+      case agg: BaseAggregateExec
+          if normalizeExpressionExists(agg.groupingExpressions) &&
+            !TransformHints.isAlreadyTagged(plan) =>
+        val newGroupingExprs = agg.groupingExpressions.toIndexedSeq.map {
+          expr =>
+            val transformedExpr = expr.transform {
+              case _ @KnownFloatingPointNormalized(_ @NormalizeNaNAndZero(child)) => child
```
ulysses-you, line 60: Can we remove `KnownFloatingPointNormalized` and `NormalizeNaNAndZero` in `ExpressionConverter` ?

liujiayi771: Yes. Removed.

liujiayi771: @ulysses-you We cannot remove, both window and join still use them. They are handled in the logical plan rule `NormalizeFloatingNumbers`, so we don't need to eliminate it. We only need to remove the `NormalizeFloatingNumbers` in agg. Although we could also include join and window in `RemoveKnownFloatingPointNormalized`, I think it's unnecessary. In the future, when Spark handles agg within the logical plan as well, we can remove `RemoveKnownFloatingPointNormalized` in Gluten.

ulysses-you: I see, after we pull out join/window, the NormalizeNaNAndZero will be moved to project and we skip tranform it at the end

gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
```diff
@@ -0,0 +1,83 @@
...
+      case agg: BaseAggregateExec
+          if normalizeExpressionExists(agg.groupingExpressions) &&
+            (validation || supportTransform(plan)) =>
+        val newGroupingExprs = agg.groupingExpressions.toIndexedSeq.map {
+          expr =>
+            val transformedExpr = expr.transform {
+              case _ @KnownFloatingPointNormalized(_ @NormalizeNaNAndZero(child)) => child
+              case _ @KnownFloatingPointNormalized(_ @If(_ @IsNull(expr), _, _)) => expr
+              case _ @KnownFloatingPointNormalized(_ @ArrayTransform(expr, _)) => expr
+              case other => other
```
ulysses-you, line 67: transform do not need other pattern

gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
```diff
@@ -0,0 +1,83 @@
...
+    expressions.exists(_.find(_.isInstanceOf[KnownFloatingPointNormalized]).isDefined)
+  }
+
+  /**
+   * If the groupingExpressions of the aggregate are expressions, an Alias will be added on the
+   * outside. After removing the KnownFloatingPointNormalized, it is necessary to remove this
+   * redundant Alias.
+   */
+  private def removeRedundantAlias(expression: Expression): Expression = expression match {
+    case _ @Alias(child: Attribute, _) => child
```
ulysses-you, line 46: `_ @` is unused

gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
```diff
@@ -0,0 +1,83 @@
...
+      // Only remove KnownFloatingPointNormalized when agg is transformable.
+      case agg: BaseAggregateExec
+          if normalizeExpressionExists(agg.groupingExpressions) &&
+            (validation || supportTransform(plan)) =>
+        val newGroupingExprs = agg.groupingExpressions.toIndexedSeq.map {
+          expr =>
+            val transformedExpr = expr.transform {
+              case _ @KnownFloatingPointNormalized(_ @NormalizeNaNAndZero(child)) => child
+              case _ @KnownFloatingPointNormalized(_ @If(_ @IsNull(expr), _, _)) => expr
+              case _ @KnownFloatingPointNormalized(_ @ArrayTransform(expr, _)) => expr
```
ulysses-you, line 66: `_ @` is unused

gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
```diff
@@ -0,0 +1,83 @@
...
+    TransformHints.isAlreadyTagged(plan) && TransformHints.isTransformable(plan)
+
+  override def apply(plan: SparkPlan): SparkPlan = applyWithValidation(plan)
+
+  def applyWithValidation(plan: SparkPlan, validation: Boolean = false): SparkPlan =
+    plan.transform {
+      // Only remove KnownFloatingPointNormalized when agg is transformable.
+      case agg: BaseAggregateExec
+          if normalizeExpressionExists(agg.groupingExpressions) &&
+            (validation || supportTransform(plan)) =>
```
ulysses-you, line 60: `supportTransform(agg)`

gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
```diff
@@ -0,0 +1,83 @@
...
+        val newGroupingExprs = agg.groupingExpressions.toIndexedSeq.map {
+          expr =>
+            val transformedExpr = expr.transform {
+              case _ @KnownFloatingPointNormalized(_ @NormalizeNaNAndZero(child)) => child
+              case _ @KnownFloatingPointNormalized(_ @If(_ @IsNull(expr), _, _)) => expr
+              case _ @KnownFloatingPointNormalized(_ @ArrayTransform(expr, _)) => expr
+              case other => other
+            }
+            removeRedundantAlias(transformedExpr)
```
ulysses-you, line 69: we can do `. asInstanceOf[NamedExpression]` here

liujiayi771:
@ulysses-you Aside from the situation in this PR, it is also possible to insert agg into the physical plan within the DPP. Therefore, I am planning to perform an additional pre-project check in the physical plan. In most cases, pre-project can be inserted in the logical plan, and any omissions can be compensated for in the physical plan. This approach can also preserve the performance advantages of processing in the logical plan. In that case, this PR would no longer be necessary.

gluten-core/src/main/scala/io/glutenproject/extension/RemoveKnownFloatingPointNormalized.scala
```diff
@@ -0,0 +1,82 @@
...
+            val transformedExpr = expr.transform {
+              case KnownFloatingPointNormalized(NormalizeNaNAndZero(child)) => child
+              case KnownFloatingPointNormalized(If(IsNull(expr), _, _)) => expr
+              case KnownFloatingPointNormalized(ArrayTransform(expr, _)) => expr
+            }
+            removeRedundantAlias(transformedExpr).asInstanceOf[NamedExpression]
+        }
+        agg match {
+          case hash: HashAggregateExec =>
+            hash.copy(groupingExpressions = newGroupingExprs)
```
ulysses-you, line 72: It seems we should fail in `TransformPreOverrides` as  we apply `RemoveKnownFloatingPointNormalized` first and do not copy tags, but the tests passed.. can you check if this rule is actually applied ?
https://github.com/oap-project/gluten/blob/451b1cee89b9425443924ae69f9dd04a186c85a4/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala#L282C1-L283

ulysses-you:
@liujiayi771 dpp should not be affected as it is a subquery. This new rule should be applied in it's own AQE.

liujiayi771:
> @liujiayi771 dpp should not be affected as it is a subquery. This new rule should be applied in it's own AQE.

I find a unit test error with ClickHouse. Upon debugging, I noticed that agg is inserted at a physical rule. So, the logical plan rule wouldn't be able to see this agg, right?

https://github.com/apache/spark/blob/86bac8911a31a98eb6ad1f5365554f4f095c0376/sql/core/src/main/scala/org/apache/spark/sql/execution/dynamicpruning/PlanDynamicPruningFilters.scala#L84C19-L84C19

Issue #4336, opened 2024-01-09
[GLUTEN-4336][VL]Questions about Parquet compression format


rhh777:
### Backend

VL (Velox)

### Bug description

As described in velox-backend-limitations.md, parquetscan does not support lzo, lz4, and zlib compression formats. During the actual test, there was no problem in reading.

https://github.com/oap-project/gluten/blob/main/docs/velox-backend-limitations.md#exception-caused-by-file-compression-codec

create table sql:
```
create table parquet_compress_lzo (col1 string, col2 int) stored as parquet tblproperties ("parquet.compression"="lzo");

create table parquet_compress_lz4 (col1 string, col2 int) stored as parquet tblproperties ("parquet.compression"="lz4");
```

table  data file
```
(venvs) [hdfs@offline1 ~]$ hadoop fs -ls /warehouse/tablespace/managed/hive/parquet_compress_lzo
-rw-rw----+  2 hdfs hadoop          0 2024-01-09 16:58 /warehouse/tablespace/managed/hive/parquet_compress_lzo/_SUCCESS
-rw-rw----+  2 hdfs hadoop        757 2024-01-09 16:58 /warehouse/tablespace/managed/hive/parquet_compress_lzo/part-00000-ad96f08c-9d21-4764-a659-65bb9ed59de6-c000.lzo.parquet
(venvs) [hdfs@offline1 ~]$ hadoop fs -ls /warehouse/tablespace/managed/hive/parquet_compress_lz4
-rw-rw----+  2 hdfs hadoop          0 2024-01-09 17:01 /warehouse/tablespace/managed/hive/parquet_compress_lz4/_SUCCESS
-rw-rw----+  2 hdfs hadoop        752 2024-01-09 17:01 /warehouse/tablespace/managed/hive/parquet_compress_lz4/part-00000-5f5b03ef-b2f3-422d-8066-f7252abba841-c000.lz4.parquet
```

query sql:
```
spark-sql> select * from parquet_compress_lzo;
compress_lzo    5
Time taken: 0.187 seconds, Fetched 1 row(s)
spark-sql> select * from parquet_compress_lz4;
compress_lz4    6
Time taken: 0.216 seconds, Fetched 1 row(s)
```

sparkui 
![image](https://github.com/oap-project/gluten/assets/41908930/3901a045-dce6-4126-83da-f4c0f13692e0)

### Spark version

Spark-3.3.x

### Spark configurations

spark-sql 
[...]

PR #4358, opened 2024-01-10
[GLUTEN-4336][VL][DOC] Remove unsupported codecs of reader

marin-ma:
parquet and dwrf now both use https://github.com/oap-project/velox/blob/2024_01_09/velox/dwio/common/compression/Compression.cpp#L644-L727 to decompress data

fixes: #4336


zhouyuan (approved):
👍

PR #4377, opened 2024-01-11
[CORE] Support native write with noop

jackylee-ch:
With this pull request, we can disregard the impact of the last ColumnarToRow when we only want to evaluate the performance of operators such as Scan, Project functions, or aggregate functions.

## How was this patch tested?
unit tests.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
@@ -32,6 +32,7 @@ import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec
 import org.apache.spark.sql.execution.command.{CreateDataSourceTableAsSelectCommand, DataWritingCommand, DataWritingCommandExec}
 import org.apache.spark.sql.execution.datasources.orc.OrcFileFormat
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
+import org.apache.spark.sql.execution.datasources.v2.{AppendDataExec, OverwriteByExpressionExec}
 import org.apache.spark.sql.hive.execution.{CreateHiveTableAsSelectCommand, InsertIntoHiveDirCommand, InsertIntoHiveTable}
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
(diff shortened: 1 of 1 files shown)


zhouyuan:
@jackylee-ch thanks for adding this feature, would it be convenient for you to add some tests? 
also copy @JkSelf 

thanks, -yuan

gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -144,7 +145,16 @@ object GlutenWriterColumnarRules {
   }
 
   case class NativeWritePostRule(session: SparkSession) extends Rule[SparkPlan] {
+
+    private val NOOP_WRITE = "org.apache.spark.sql.execution.datasources.noop.NoopWrite$"
```
zhouyuan, line 149: is this class name valid for spark 32/33/34 ?

JkSelf: @zhouyuan The class name has been verified to exist in Spark [3.2](https://github.com/apache/spark/blob/v3.2.2/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/noop/NoopDataSource.scala#L61) and [3.3](https://github.com/apache/spark/blob/v3.3.2/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/noop/NoopDataSource.scala#L59). However, starting from Spark 3.4, this code will no longer be executed.

JkSelf:
@jackylee-ch Thanks for your improvement. Can you add some unit test for spark 32 and spark33? Thanks.

zhouyuan (approved):
👍 
note, spark34 is not using the code path, so it should be treated differently

JkSelf (approved):

jackylee-ch:
@zhouyuan @JkSelf Thanks for your review.

liujiayi771:
@zhouyuan Can we add this to `OffloadSingleNode` for Spark 3.4/3.5?

jackylee-ch:
> @zhouyuan Can we add this to `OffloadSingleNode` for Spark 3.4/3.5?

Sure, I can try this on Spark 3.4/3.5

PR #4400, opened 2024-01-15
[VL] minor fix for velox cache

zhli1142015:
We observed performance gain for velox cache (`CacheBufferInput`  class) with config values specified in https://github.com/oap-project/gluten/commit/b57f56030eef2496e8d871c8f0e766b96661da25.
Here are some test results for 1TB TPCDS (3 runs eache query get fareast one as result), 8 executors(8 cores 56Gb memory each):
Velox cache enabled (mem only 8GB): 4.9%.
Velox cache enabled (mem only 8GB) + soft affinity: 9.7%.
```
"spark.gluten.sql.columnar.backend.velox.cacheEnabled": "true",
"spark.gluten.sql.columnar.backend.velox.memCacheSize": "8589934592",
"spark.gluten.sql.columnar.backend.velox.ssdCacheSize": "0",
"spark.gluten.sql.columnar.backend.velox.fileHandleCacheEnabled": "true",
"spark.gluten.soft-affinity.enabled": "true",
```

The limit of this approach is now the memory used by velox cache is not managed by spark.


zhli1142015:
cc @zhouyuan

gluten-core/src/main/scala/io/glutenproject/GlutenPlugin.scala
```diff
@@ -180,6 +180,15 @@ private[glutenproject] class GlutenDriverPlugin extends DriverPlugin with Loggin
       conf.set("spark.sql.orc.enableVectorizedReader", "false")
       conf.set("spark.sql.inMemoryColumnarStorage.enableVectorizedReader", "false")
     }
+
+    if (
```
zhouyuan, line 184: would be better if you could add a note here

zhouyuan (approved):
@zhli1142015 
thanks for the fix! The local cache has been broken for a while as i know.

gluten-core/src/main/scala/io/glutenproject/GlutenPlugin.scala
zhli1142015: Added, thanks

zhouyuan (approved):
👍 
works for me

PR #4401, opened 2024-01-15
[VL] Change row2Partition to uint32_t to deal with num partitions ...

marin-ma:
...larger than 64k and partition buffer size type to uint16_t


FelixYBW (approved):
Thank you for your work!

FelixYBW:
The PR does two things: 

- Remove the limitation of 64K partitions in max. Partition number may be more than 64K if cluster and dataset is too huge
- Add limitation of row number per batch to 32K in max. The bigger batch size, the less switch during batches which may benefit performance, but CPU cache miss will increase. In our test 32K is already too large. If we set it to less than 32K, there are several struct we can use uint16_t instead of uint32_t which can decrease the data foot print and these data struct is hot and assumed to be cached in CPU L2 cache always.

Issue #4405, opened 2024-01-15
[CORE] InExpressionTransformer should let element in order then pass to backend to keep unify


zwangsheng:
### Description

As title

PHILO-HE:
Closing as fixed. Thanks!

PR #4406, opened 2024-01-15
[GLUTEN-4405][CORE] Sort values of In/InSet transformer for deterministic behavior

zwangsheng:

#4405

Spark has a declaration logic for In Set functions that keeps the output in a consistent order, so let's make input array sorted when we call transformer.

https://github.com/apache/spark/blob/1bdb81b2b5473793697083fdd9ec1a07ae605528/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala#L605-L612

## How was this patch tested?

None

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/expression/PredicateExpressionTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/expression/PredicateExpressionTransformer.scala
@@ -60,6 +60,9 @@ object InExpressionTransformer {
     val expressionNodes = new java.util.ArrayList[ExpressionNode](
       values
         .map(value => ExpressionBuilder.makeLiteral(value, valueType, value == null))
+        .toSeq
+        // Sort elements for deterministic behaviours
+        .sortBy(_.toString)
         .asJava)
 
     ExpressionBuilder.makeSingularOrListNode(leftNode, expressionNodes)
--- a/gluten-core/src/main/scala/io/glutenproject/expression/PredicateExpressionTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/expression/PredicateExpressionTransformer.scala
@@ -60,7 +60,7 @@ object InExpressionTransformer {
     val expressionNodes = new java.util.ArrayList[ExpressionNode](
       values.toSeq
         // Sort elements for deterministic behaviours
-        .sortBy(value => if (value == null) "NULL"  else value.toString)
+        .sortBy(Literal(_, valueType).toString())
         .map(value => ExpressionBuilder.makeLiteral(value, valueType, value == null))
         .asJava)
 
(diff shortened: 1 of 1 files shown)


zwangsheng:
PTAL @PHILO-HE @ulysses-you

gluten-core/src/main/scala/io/glutenproject/expression/PredicateExpressionTransformer.scala
```diff
@@ -58,7 +58,9 @@ object InExpressionTransformer {
       values: Set[Any],
       valueType: DataType): ExpressionNode = {
     val expressionNodes = new java.util.ArrayList[ExpressionNode](
-      values
+      values.toSeq
+        // Sort elements for deterministic behaviours
+        .sortBy(value => if (value == null) "NULL"  else value.toString)
```
PHILO-HE, line 63: It looks elements are sorted based the strings converted for values. Can we just do the sorting based on values? I'm assuming their type is comparable.

PHILO-HE (commented):
Thanks for your fix!
I note CI reported code style issue. You can follow the below guide to generate a fix:
https://github.com/oap-project/gluten/blob/main/docs/developers/NewToGluten.md#javascala-code-style

gluten-core/src/main/scala/io/glutenproject/expression/PredicateExpressionTransformer.scala
zwangsheng: Using `values.toSeq.sorted` directly seems to have compilation problems.

How about we follow the Spark logic in the following way:

```scala
      values.toSeq
        // Sort elements for deterministic behaviours
        .sortBy(Literal(_, valueType).toString())
```

PHILO-HE: I think it may be ok. Thanks!

gluten-core/src/main/scala/io/glutenproject/expression/PredicateExpressionTransformer.scala
```diff
@@ -58,7 +58,9 @@ object InExpressionTransformer {
       values: Set[Any],
       valueType: DataType): ExpressionNode = {
     val expressionNodes = new java.util.ArrayList[ExpressionNode](
-      values
+      values.toSeq
```
PHILO-HE, line 61: I note this method is shared by `In/InSet`. But the purpose is to only fix `InSet`. Should we keep no change for `In`?

zwangsheng: Well, for now, our goal is to maintain consistent behavior, and the Gluten In Function is also converted by Set, so it is acceptable for changes to involve In Function.

PHILO-HE:
cc @exmy
If not sorted, the result can be inconsistent for each run? Could you please clarify a bit? Thanks!

zwangsheng:
> If not sorted, the result can be inconsistent for each run? Could you please clarify a bit? Thanks!

It won't cause different result or performance loss, we just need this PR to make sure that the final toString output stays the same

PHILO-HE (approved):
Looks good! Thanks!

exmy:
> cc @exmy If not sorted, the result can be inconsistent for each run? Could you please clarify a bit? Thanks!

I think it's ok.

PR #4407, opened 2024-01-15
[Velox] soft affinity support placing duplicate reading to same executors

zhli1142015:

We add two APIs in Soft Affinity to implement the optimization of handling duplicate data reads by preferentially scheduling them on the same executor. This differs from the hash-based method in that it only sets the prefer location during repeated reads; the initial read is not affected by this. This approach reduces the observed uneven distribution of tasks caused by hash conflicts.
With the new APIs and spark events consumption, we obtain information about the relationships between partitions, RDDs, tasks, and executors. Eventually, we simplify this information into a mapping of partitions to executors. This allows us to determine which partition has been read on which executor. When duplicate partitions arrive, we can then allocate the corresponding tasks to the executor where the partition has been read as much as possible.

Its limitation is in assuming that the cache size will always be sufficient, and all reads will be cached. However, this assumption may not always hold true in practice.

## How was this patch tested?
UT.

Velox cache enabled (mem only 8GB) + soft affinity: 9.7%.
Velox cache enabled (mem only 8GB) + soft affinity (duplicate reading detecting): 14.3%.
For details please see, https://github.com/oap-project/gluten/pull/4400#issue-2081288878

Fixes: #4762

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/softaffinity/SoftAffinityManager.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/softaffinity/SoftAffinityManager.scala
@@ -25,7 +25,8 @@ import org.apache.spark.SparkEnv
 import org.apache.spark.internal.Logging
 import org.apache.spark.scheduler.{SparkListenerStageCompleted, SparkListenerStageSubmitted, SparkListenerTaskEnd}
 import org.apache.spark.sql.execution.datasources.FilePartition
-import org.apache.spark.sql.execution.ui._
+
+import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
 
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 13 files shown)


zhli1142015:
@baibaichen and @zhouyuan , could you help to review?
Thanks.

zhli1142015:
General ping, @zhouyuan and @baibaichen , thanks.

gluten-core/src/main/scala/io/glutenproject/softaffinity/SoftAffinityManager.scala
```diff
@@ -47,6 +53,30 @@ abstract class AffinityManager extends LogLevelUtil with Logging {
...
+  val totalSoftAffinitySetRecords = new AtomicInteger(0)
+  val totalSoftAffinityMissRecords = new AtomicInteger(0)
+  val totalSoftAffinityEffiencyRecords = new AtomicInteger(0)
+
+  val querySoftAffinityDetectInfo = new ConcurrentHashMap[Long, Int]()
+  val querySoftAffinityDetectMissInfo = new ConcurrentHashMap[Long, Int]()
+  val querySoftAffinityDetectResultInfo = new ConcurrentHashMap[Long, Int]()
+
+  // rdd id -> patition id, file path, start, length
+  val rddPartitionInfoMap = new ConcurrentHashMap[Int, Array[(Int, String, Long, Long)]]()
```
zhouyuan, line 71: With large spark job this could be very big? especially the file path is also incuded

zhouyuan:
CC: @zzcclp

gluten-core/src/main/scala/io/glutenproject/softaffinity/SoftAffinityManager.scala
zhli1142015: We add config `GLUTEN_SOFT_AFFINITY_MAX_DUPLICATE_READING_RECORDS` to control the map size. If the records in map > this threshold, then recording logic would be skipped.

gluten-core/src/main/scala/org/apache/spark/softaffinity/SoftAffinity.scala
```diff
@@ -44,6 +45,25 @@ abstract class Affinity(val manager: AffinityManager) extends LogLevelUtil with
     }
   }
 
+  def getFilePartitionLocations(filePartition: FilePartition): Array[String] = {
+    val filePaths = filePartition.files.map(_.filePath.toString)
+    val preferredLocations = filePartition.preferredLocations()
+    if (shouldUseSoftAffinity(filePaths, preferredLocations)) {
+      if (manager.detectDuplicateReading) {
```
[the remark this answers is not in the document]
zhli1142015: yes, soft affiity is enabled by a separate config different one from velox local cache configs. We add a new API for velox, so it should not impact CH cases. In this new API, if `detectDuplicateReading` is true, new logic would be called, otherwise the original logic is executed.

zzcclp:
> CC: @zzcclp

I will review later

gluten-core/src/main/scala/io/glutenproject/softaffinity/SoftAffinityManager.scala
zzcclp: can we consider to use `LoadingCache` to save these records, and then set the `maximumSize` of the `LoadingCache` to control the max records count, while it will use the  LRU policy to remove unused key.

zhli1142015: Thanks, I updated `duplicateReadingInfos` with `LoadingCache`. For others maps they are used to store temp informations and would be clear after stages ending, so I keep them as before.

zzcclp:
LGTM

zhouyuan (approved):
👍

zhouyuan:
@zhli1142015 could you please also help to create one issue to track this patch?

PR #4414, opened 2024-01-16
[VL] add missed validation for in expr and pushdown filter

zhli1142015:

```
(0 to 3).toDF("x").write.parquet("time_x")
spark.read.parquet("time_x").createOrReplaceTempView("timestamps")
spark.sql("SELECT x FROM timestamps WHERE cast(x as timestamp) IN ('1970-01-01 08:00:01','1970-01-01 08:00:02')").collect
```
Because timestamp cast is not fallback, wrong result is generated.
```
(1) Scan parquet 
Output [1]: [x#0]
Batched: true
Location: InMemoryFileIndex [file:/var/git/time_x]
ReadSchema: struct<x:int>

(2) FilterExecTransformer
Input [1]: [x#0]
Arguments: cast(x#0 as timestamp) IN (1970-01-01 08:00:01,1970-01-01 08:00:02)

```

## How was this patch tested?

UT.

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
@@ -795,4 +795,15 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
       checkOperatorMatch[BatchScanExecTransformer]
     }
   }
+
+  test("timestamp cast fallback") {
+    withTempPath {
+      path =>
+        (0 to 3).toDF("x").write.parquet(path.getCanonicalPath)
+        spark.read.parquet(path.getCanonicalPath).createOrReplaceTempView("view")
+        runQueryAndCompare(
+          "SELECT x FROM view WHERE cast(x as timestamp) " +
+            "IN ('1970-01-01 08:00:00.001','1970-01-01 08:00:00.2')")(_)
+    }
+  }
 }
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -315,7 +315,7 @@ bool SubstraitToVeloxPlanValidator::validateSingularOrList(
     const ::substrait::Expression::SingularOrList& singularOrList,
     const RowTypePtr& inputType) {
   for (const auto& option : singularOrList.options()) {
-    if(!option.has_literal()) {
+    if (!option.has_literal()) {
       logValidateMsg("native validation failed due to: Option is expected as Literal.");
       return false;
     }
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -271,7 +271,6 @@ bool SubstraitToVeloxPlanValidator::validateCast(
   }
 
   core::TypedExprPtr input = exprConverter_->toVeloxExpr(castExpr.input(), inputType);
-
   // Casting from some types is not supported. See CastExpr::applyCast.
   if (input->type()->isDate()) {
     if (toType->kind() == TypeKind::TIMESTAMP) {
@@ -311,6 +310,22 @@ bool SubstraitToVeloxPlanValidator::validateIfThen(
   return true;
 }
 
+bool SubstraitToVeloxPlanValidator::validateSingularOrList(
+    const ::substrait::Expression::SingularOrList& singularOrList,
+    const RowTypePtr& inputType) {
+  for (const auto& option : singularOrList.options()) {
+    if(!option.has_literal()) {
+      logValidateMsg("native validation failed due to: Option is expected as Literal.");
+      return false;
+    }
+    if (!validateLiteral(option.literal(), inputType)) {
+      return false;
+    }
+  }
+
+  return validateExpression(singularOrList.value(), inputType);
+}
+
 bool SubstraitToVeloxPlanValidator::validateExpression(
     const ::substrait::Expression& expression,
     const RowTypePtr& inputType) {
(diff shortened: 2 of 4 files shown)


jinchengchenghh (approved):

rui-mo (commented):
> Because timestamp cast is not fallback, wrong result is generated.

@zhli1142015 Thanks for your work. Does above mean casting from int to timestamp gives wrong output? If so, why do we fix it on IN expr?

> cast(cast(x#7 as timestamp) as string) IN (1970-01-01 08:00:00.001,1970-01-01 08:00:00.2)

I also notice the related IN expr is on string type. Are you noticing wrong result for that?

zhli1142015:
> > Because timestamp cast is not fallback, wrong result is generated.
> 
> @zhli1142015 Thanks for your work. Does above mean casting from int to timestamp gives wrong output? If so, why do we fix it on IN expr?
> 
> > cast(cast(x#7 as timestamp) as string) IN (1970-01-01 08:00:00.001,1970-01-01 08:00:00.2)
> 
> I also notice the related IN expr is on string type. Are you noticing wrong result for that?

Look both cast(timestamp as string) and cast(int as timestamp) are related to the wrong result. I saw we have logic to fallback both. but as in is missed in the validation logic, so they are not executed. does these two cast should work as expected now?

Wrong result --> before this change, this query returns empty.
```
scala> spark.sql("SELECT x FROM timestamps WHERE cast(x as timestamp) IN ('1970-01-01 08:00:01','1970-01-01 08:00:02')").collect
res1: Array[org.apache.spark.sql.Row] = Array()
```
after the fallback it returns 
```
scala> spark.sql("SELECT x FROM timestamps WHERE cast(x as timestamp) IN ('1970-01-01 08:00:01','1970-01-01 08:00:02')").collect
24/01/17 10:45:43 WARN GlutenFallbackReporter: Validation failed for plan: Scan parquet [QueryId=7], due to: native check failure:native validation failed due to: Casting to TIMESTAMP is not supported..
24/01/17 10:45:43 WARN GlutenFallbackReporter: Validation failed for plan: ColumnarToRow[QueryId=7], due to: child is not gluten plan.
24/01/17 10:45:43 WARN GlutenFallbackReporter: Validation failed for plan: Filter[QueryId=7], due to: native check failure:native validation failed due to: Casting to TIMESTAMP is not supported..
res7: Array[org.apache.spark.sql.Row] = Array([2], [1])
```

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -774,4 +775,27 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
...
+        (0 to 3).toDF("x").write.parquet(path.getCanonicalPath)
+        spark.read.parquet(path.getCanonicalPath).createOrReplaceTempView("view")
+        runQueryAndCompare(
+          "SELECT x FROM view WHERE cast(x as timestamp) " +
+            "IN ('1970-01-01 08:00:00.001','1970-01-01 08:00:00.2')") {
+          df =>
+            {
+              val plans = getExecutedPlan(df)
+              val result =
+                plans.exists(plan => plan.find(_.isInstanceOf[BatchScanExec]).isDefined) &&
```
mskapilks, line 791: `BatchScanExec` Does this mean it is falling back to Spark?

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -311,6 +311,23 @@ bool SubstraitToVeloxPlanValidator::validateIfThen(
   return true;
 }
 
+bool SubstraitToVeloxPlanValidator::validateSingularOrList(
+    const ::substrait::Expression::SingularOrList& singularOrList,
+    const RowTypePtr& inputType) {
+  auto options = singularOrList.options();
+  for (const auto& option : options) {
```
rui-mo, line 318: Maybe replace `options` with `singularOrList.options()` as it is used only once.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.h
```diff
@@ -127,6 +127,10 @@ class SubstraitToVeloxPlanValidator {
   /// Validate Substrait if-then expression.
   bool validateIfThen(const ::substrait::Expression_IfThen& ifThen, const RowTypePtr& inputType);
 
+  bool validateSingularOrList(
```
rui-mo, line 130: Maybe add a comment for this method.
> /// Validate Substrait IN expression.

rui-mo (commented):
Thanks.

rui-mo:
> Look both cast(timestamp as string) and cast(int as timestamp) are related to the wrong result.

I see. Opened https://github.com/oap-project/gluten/issues/4433 to track this bug.

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
rui-mo: Maybe we don't need to verify a fallback query plan as offloaded execution should be the preferred.

zhli1142015: Ok, let me remove it.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
zhli1142015: Updated, thanks

cpp/velox/substrait/SubstraitToVeloxPlanValidator.h
zhli1142015: Updated, thanks

rui-mo (approved):

rui-mo:
@zhli1142015 Just noticed the unit test was removed. Is there any reason for that?

zhli1142015:
> @zhli1142015 Just noticed the unit test was removed. Is there any reason for that? `runQueryAndCompare` could check the query result without plan verification.

Add i back, thanks.

rui-mo (approved):
Thanks!

PR #4415, opened 2024-01-16
[VL] Optimize ensureFlattened

marin-ma:
Use upstream `facebook::velox::BaseVector::flattenVector` to avoid unnecessary copy.

### DIFF
--- a/cpp/velox/memory/VeloxColumnarBatch.cc
+++ b/cpp/velox/memory/VeloxColumnarBatch.cc
@@ -16,6 +16,7 @@
  */
 #include "VeloxColumnarBatch.h"
 #include "compute/VeloxRuntime.h"
+#include "utils/Timer.h"
 #include "utils/VeloxArrowUtils.h"
 #include "velox/row/UnsafeRowFast.h"
 #include "velox/type/Type.h"
@@ -47,20 +48,17 @@ void VeloxColumnarBatch::ensureFlattened() {
   if (flattened_ != nullptr) {
     return;
   }
-  auto startTime = std::chrono::steady_clock::now();
-  // Make sure to load lazy vector if not loaded already.
-  for (auto& child : rowVector_->children()) {
-    child->loadedVector();
-  }
 
-  // Perform copy to flatten dictionary vectors.
-  velox::RowVectorPtr copy = std::dynamic_pointer_cast<velox::RowVector>(
-      velox::BaseVector::create(rowVector_->type(), rowVector_->size(), rowVector_->pool()));
-  copy->copy(rowVector_.get(), 0, 0, rowVector_->size());
-  flattened_ = copy;
-  auto endTime = std::chrono::steady_clock::now();
-  auto duration = std::chrono::duration_cast<std::chrono::nanoseconds>(endTime - startTime).count();
-  exportNanos_ += duration;
+  ScopedTimer timer(&exportNanos_);
+  for (auto& child : rowVector_->children()) {
+    facebook::velox::BaseVector::flattenVector(child);
+    if (child->isLazy()) {
+      child = child->as<facebook::velox::LazyVector>()->loadedVectorShared();
+      VELOX_DCHECK_NOT_NULL(child);
+    }
+  }
+  // Invalid the original rowVector_ after being flattened.
+  flattened_ = std::move(rowVector_);
 }
 
 std::shared_ptr<ArrowSchema> VeloxColumnarBatch::exportArrowSchema() {
@@ -83,6 +81,7 @@ int64_t VeloxColumnarBatch::numBytes() {
 }
 
 velox::RowVectorPtr VeloxColumnarBatch::getRowVector() const {
+  VELOX_CHECK_NOT_NULL(rowVector_);
   return rowVector_;
 }
 
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -91,7 +91,11 @@ const int32_t* getFirstColumn(const facebook::velox::RowVector& rv) {
   VELOX_CHECK(rv.childrenSize() > 0, "RowVector missing partition id column.");
 
   auto& firstChild = rv.childAt(0);
-  VELOX_CHECK(firstChild->type()->isInteger(), "RecordBatch field 0 should be integer");
+  VELOX_CHECK(firstChild->isFlatEncoding(), "Partition id (field 0) is not flat encoding.");
+  VELOX_CHECK(
+      firstChild->type()->isInteger(),
+      "Partition id (field 0) should be integer, but got {}",
+      firstChild->type()->toString());
 
   // first column is partition key hash value or pid
   return firstChild->asFlatVector<int32_t>()->rawValues();
@@ -472,9 +476,7 @@ arrow::Status VeloxShuffleWriter::splitRowVector(const facebook::velox::RowVecto
 arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const facebook::velox::RowVector& rv) {
   for (auto col = 0; col < fixedWidthColumnCount_; ++col) {
     auto colIdx = simpleColumnIndices_[col];
-    auto column = rv.childAt(colIdx);
-    assert(column->isFlatEncoding());
-
+    auto& column = rv.childAt(colIdx);
     const uint8_t* srcAddr = (const uint8_t*)column->valuesAsVoid();
     const auto& dstAddrs = partitionFixedWidthValueAddrs_[col];
 
@@ -679,7 +681,7 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const facebook::vel
   arrow::Status VeloxShuffleWriter::splitValidityBuffer(const facebook::velox::RowVector& rv) {
     for (size_t col = 0; col < simpleColumnIndices_.size(); ++col) {
       auto colIdx = simpleColumnIndices_[col];
-      auto column = rv.childAt(colIdx);
+      auto& column = rv.childAt(colIdx);
       if (vectorHasNull(column)) {
         auto& dstAddrs = partitionValidityAddrs_[col];
         for (auto& pid : partitionUsed_) {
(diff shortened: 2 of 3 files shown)


cpp/velox/memory/VeloxColumnarBatch.cc
```diff
@@ -47,20 +48,13 @@ void VeloxColumnarBatch::ensureFlattened() {
   if (flattened_ != nullptr) {
     return;
   }
-  auto startTime = std::chrono::steady_clock::now();
-  // Make sure to load lazy vector if not loaded already.
+
+  ScopedTimer timer(&exportNanos_);
```
Yohahaha, line 52: curious could we use velox's `MicrosecondTimer` to avoid repeated implementation of same functionality.

cpp/velox/memory/VeloxColumnarBatch.cc
```diff
@@ -83,6 +77,7 @@ int64_t VeloxColumnarBatch::numBytes() {
 }
 
 velox::RowVectorPtr VeloxColumnarBatch::getRowVector() const {
+  VELOX_CHECK_NOT_NULL(rowVector_);
```
jinchengchenghh, line 80: VELOX_DCHECK_NOT_NULL

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -91,12 +91,24 @@ const int32_t* getFirstColumn(const facebook::velox::RowVector& rv) {
   VELOX_CHECK(rv.childrenSize() > 0, "RowVector missing partition id column.");
 
   auto& firstChild = rv.childAt(0);
-  VELOX_CHECK(firstChild->type()->isInteger(), "RecordBatch field 0 should be integer");
+  VELOX_CHECK(firstChild->isFlatEncoding(), "Partition id (field 0) is not flat encoding.");
```
jinchengchenghh, line 94: VELOX_DCHECK, this logic is guaranteed by code, so the DCHECK is enough

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -91,12 +91,24 @@ const int32_t* getFirstColumn(const facebook::velox::RowVector& rv) {
...
+      "Partition id (field 0) should be integer, but got {}",
+      firstChild->type()->toString());
 
   // first column is partition key hash value or pid
   return firstChild->asFlatVector<int32_t>()->rawValues();
 }
 
+facebook::velox::VectorPtr flatChildAt(const facebook::velox::RowVector& rv, facebook::velox::column_index_t idx) {
+  auto column = rv.childAt(idx);
+  if (column->isLazy()) {
```
jinchengchenghh, line 106: What if column is complex datatype? Could this situation happens? The `struct` column is not lazy and its child column is lazy.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -764,10 +774,9 @@ arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const facebook::vel
       auto binaryIdx = col - fixedWidthColumnCount_;
       auto& dstAddrs = partitionBinaryAddrs_[binaryIdx];
       auto colIdx = simpleColumnIndices_[col];
-      auto column = rv.childAt(colIdx);
-      auto stringColumn = column->asFlatVector<facebook::velox::StringView>();
-      assert(stringColumn);
-      RETURN_NOT_OK(splitBinaryType(binaryIdx, *stringColumn, dstAddrs));
+      auto column = flatChildAt(rv, colIdx)->asFlatVector<facebook::velox::StringView>();
+      VELOX_CHECK_NOT_NULL(column);
```
jinchengchenghh, line 778: DCHECK

cpp/velox/shuffle/VeloxShuffleWriter.cc
marin-ma: The complex types will be handled by `PrestoSerializer` properly.

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -466,9 +478,7 @@ arrow::Status VeloxShuffleWriter::splitRowVector(const facebook::velox::RowVecto
 arrow::Status VeloxShuffleWriter::splitFixedWidthValueBuffer(const facebook::velox::RowVector& rv) {
   for (auto col = 0; col < fixedWidthColumnCount_; ++col) {
     auto colIdx = simpleColumnIndices_[col];
-    auto column = rv.childAt(colIdx);
-    assert(column->isFlatEncoding());
-
+    auto column = flatChildAt(rv, colIdx);
```
zhztheplayer, line 481: Why `flatChildAt` becomes required in the patch?

marin-ma: Because in `facebook::velox::BaseVector::flattenVector`, it loads lazy vectors but preserves lazy encoding.

zhztheplayer: I see. Then is it possible to handle the case inside `VeloxColumnarBatch::ensureFlattened` ?

marin-ma:
One Spark AQE UT failed, because w/o copy, `facebook::velox::RowVector::estimateFlatSize` can produce larger data size than the actual uncompressed data size as it calculated the acutal underly buffer size. In the failed test case, the number of rows of RowVector is 6, but the underlying buffer lengths are 4096. https://github.com/oap-project/gluten/pull/4428 will address this by using "uncompressed size" as the "dataSize" metrics.

Yohahaha:
> One Spark AQE UT failed, because w/o copy, `facebook::velox::RowVector::estimateFlatSize` can produce larger data size than the actual uncompressed data size as it calculated the acutal underly buffer size. In the failed test case, the number of rows of RowVector is 6, but the underlying buffer lengths are 4096. #4428 will address this by using "uncompressed size" as the "dataSize" metrics.

One question here: origin `dataSize` is split call returned value, which is estimateFlatSize. `splitResult.getRawPartitionLengths` is arrow buffer size, correct me if wrong.
So, estimateFlatSize is not accurate, arrow buffer size is accurate?

We have observed some query shows the uncompressed size is much smaller than data size, such as 58G(uncompressed size) vs 3.1T(data size).

marin-ma:
> One question here: origin `dataSize` is split call returned value, which is estimateFlatSize. `splitResult.getRawPartitionLengths` is arrow buffer size, correct me if wrong. So, estimateFlatSize is not accurate, arrow buffer size is accurate?

@Yohahaha The "dataSize" metric in Vanilla Spark refers to the uncompressed size of the InternalRow. As it's used in the AQE "LocalShuffleReader" optimization, my understanding is that they consider this value as the raw uncompressed **output** data size. On the other hand, the size of the RowVector can be considered as the **input** size for the shuffle. The uncompressed Arrow buffers align more closely with the concept of the raw uncompressed size in this context.


> We have observed some query shows the uncompressed size is much smaller than data size, such as 58G(uncompressed size) vs 3.1T(data size).

This seems to exceed normal expectations. Could you provide some more detailed workload data? For example, data size, data schema, shuffle partitions, etc.

zhztheplayer (approved):

Yohahaha:
Vanilla Spark update `dataSize` by UnsafeRow::getSizeInBytes https://github.com/apache/spark/blob/38fc127a232ed4005e6436cdef4c5a1f1bba95f7/sql/core/src/main/scala/org/apache/spark/sql/execution/UnsafeRowSerializer.scala#L63-L71 and use `shuffle bytes written` as compressed size.

> On the other hand, the size of the RowVector can be considered as the input size for the shuffle.

So you think estimateFlatSize is not the real uncompressed size? From my understanding, shuffle input size is same as uncompressed size, Vanilla Spark use `dataSize` as QueryStageExec's runtimeStatistics instead of `shuffle bytes written`.

What I want confirmed is whether arrow buffers really reflect real uncompressed size? is any reuse logic for arrow buffers?

Yohahaha:
> >We have observed some query shows the uncompressed size is much smaller than data size, such as 58G(uncompressed size) vs 3.1T(data size).

> This seems to exceed normal expectations. Could you provide some more detailed workload data? For example, data size, data schema, shuffle partitions, etc.

any query before #4428 can reproduce, above result is from TPCDS 10T with 8000 partitions.
image

marin-ma:
> So you think estimateFlatSize is not the real uncompressed size?

Yes. And it's true. This PR has one failed UT w/o #4428, that the RowVector length is 6, but the underlying buffer lengths are 4096. `esitmateFlatSize` counts all buffer size. Before this patch, the RowVector is flattened by copying, and the copied buffers are of length 6. After this patch, the copy is removed so the `esitmateFlatSize` is much larger than before. It's more likely that this value represents the memory usage, but it cannot represent the "uncompressed size".

> What I want confirmed is whether arrow buffers really reflect real uncompressed size? is any reuse logic for arrow buffers?

The reuse logic has nothing to do with the buffer size computing. We accumulate the total raw arrow buffer size each time before the buffers is destroyed or reused.


I would think the case you provided is reasonable, because Velox buffer size can be larger than arrow buffers. One known case is that in Velox, any string <= 12 characters will use 12 bytes, while arrow stores real length. But 58G(uncompressed size) vs 3.1T(data size) is very abnormal and needs further investigation. Could you reproduce this case and compare the data size with Vanilla spark? What's the compressed data size (shuffle bytes written)?

Yohahaha:
I rerun a new case of TPCDS sf100 q99.

| Vanilla | Gluten |
|--------|--------|
| image | image |

you can see Gluten's `dataSize` is smaller than `shuffle bytes written`, is this expected?

marin-ma:
> I rerun a new case of TPCDS sf100 q99.
> 
> Vanilla	Gluten
> image	image
> you can see Gluten's `dataSize` is smaller than `shuffle bytes written`, is this expected?

@Yohahaha Yes. The "dataSize" is the original "uncompressed size". If the data after compression is larger than it's uncompressed size, then we directly write the original uncompressed data, but with a few metadata added. Therefore the "shuffle bytes written" can be greater than the "dataSize" in some cases.

Yohahaha:
> > you can see Gluten's `dataSize` is smaller than `shuffle bytes written`, is this expected?
> 
> @Yohahaha Yes. The "dataSize" is the original "uncompressed size". If the data after compression is larger than it's uncompressed size, then we directly write the original uncompressed data, but with a few metadata added. Therefore the "shuffle bytes written" can be greater than the "dataSize" in some cases.

got it, thank you! but still make no sense for me, will keep tracking this issue.

marin-ma:
@Yohahaha Thanks. Let's discuss in https://github.com/oap-project/gluten/issues/4576 if you have more inputs.

PR #4428, opened 2024-01-17
[VL] Use uncompressed size for "dataSize" metrics in shuffle

marin-ma:
Some AQE test failed due to this change:
`gluten Exchange reuse`
`gluten Avoid changing merge join to broadcast join if too many empty partitions on build plan`

Separate `GlutenAdaptiveQueryExecSuite` for different backends and modified above test cases.


marin-ma:
@zzcclp We discussed offline that we agree on separating AQE tests for backends because they produce different shuffle statistics. Please kindly review. Thanks!

PHILO-HE (approved):
Looks good! Thanks!

gluten-data/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -195,6 +194,8 @@ class ColumnarShuffleWriter[K, V](
     dep.metrics("bytesSpilled").add(splitResult.getTotalBytesSpilled)
     dep.metrics("splitBufferSize").add(splitResult.getSplitBufferSize)
     dep.metrics("uncompressedDataSize").add(splitResult.getRawPartitionLengths.sum)
+    dep.metrics("dataSize").add(splitResult.getRawPartitionLengths.sum)
+    logWarning(s"Shuffle ${dep.shuffleId} dataSize: ${splitResult.getRawPartitionLengths.sum}")
```
Yohahaha, line 198: does this needed? or forgot to remove?

marin-ma: Ooops sorry. Let me quick fix.

gluten-data/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -195,6 +194,8 @@ class ColumnarShuffleWriter[K, V](
     dep.metrics("bytesSpilled").add(splitResult.getTotalBytesSpilled)
     dep.metrics("splitBufferSize").add(splitResult.getSplitBufferSize)
     dep.metrics("uncompressedDataSize").add(splitResult.getRawPartitionLengths.sum)
+    dep.metrics("dataSize").add(splitResult.getRawPartitionLengths.sum)
```
ulysses-you, line 197: Why not use columnar batch size ? IIUC, the getRawPartitionLengths is from arrow payload, but what we would hold in memory is columnar batch, e.g., we use velox columnar batch serializer in broadcast join. The dataSize affects AQE tune shuffled join to broadcast join.

Yohahaha: Some discussion in #4415, I'm confused too.

marin-ma: @ulysses-you @Yohahaha  Some background:
The purpose of another PR https://github.com/oap-project/gluten/pull/4415 was to eliminate the "copy" operation in the native 'ensureFlattened' function. However, this removal led to inaccuracies in the data size obtained from the Velox API, which became significantly larger than the expected size based on the splitted data and is also inaccurate comparing with the output size of Velox columnar batch serializer. Therefore we consider the original "uncompressed size", which is the arrow buffer size, should be more accurate.

But recently we found this removal can trigger UT failure in some corner cases, therefore the changes in https://github.com/oap-project/gluten/pull/4415 has been partially removed, and the "copy" code path is still being used on main branch.

zhli1142015: To me, the datasize calculation now and also the datasize calculation before this change both are not accurate. We observed same joins is converted to broadcast join in Vanilla spark but not with gluten.

Yohahaha: > therefore the changes in https://github.com/oap-project/gluten/pull/4415 has been partially removed, and the "copy" code path is still being used on main branch.

may I ask which patch bring back "copy" path codes? current main branch still use `flattenVector`

https://github.com/oap-project/gluten/blob/fb10c4c7a32293a4ead7f790f19f3b0f088a35f0/cpp/velox/memory/VeloxColumnarBatch.cc#L46-L64

ulysses-you: I see, thank you @marin-ma.

PR #4429, opened 2024-01-17
[VL][CI] Ensure latest Gluten is cloned & compiled in docker build

PZD-CHINA:
Use ARG CACHEBUST=$(date +%s) in Docker builds to break caching, ensuring latest Gluten are fetched with each build.


PHILO-HE (approved):
Thanks for your fix!

PR #4441, opened 2024-01-18
[VL] Fix Filter node falling back to Spark whenever there is Timestamp column in input (child node)

mskapilks:
Currently if the input to filter node contains any timestamp column, it fallbacks to Spark even though filter predicate may not be on timestamp column. Current check is very strict.

(Fixes: \#ISSUE-ID)

## How was this patch tested?
Unit tests in PR

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
@@ -21,7 +21,7 @@ import io.glutenproject.sql.shims.SparkShimLoader
 
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{AnalysisException, DataFrame, Row}
-import org.apache.spark.sql.execution.{FileSourceScanExec, GenerateExec, RDDScanExec}
+import org.apache.spark.sql.execution.{GenerateExec, RDDScanExec}
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.functions.{avg, col, lit, udf}
 import org.apache.spark.sql.internal.SQLConf
--- a/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
@@ -21,7 +21,7 @@ import io.glutenproject.sql.shims.SparkShimLoader
 
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{AnalysisException, DataFrame, Row}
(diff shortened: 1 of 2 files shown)


cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -829,12 +830,6 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::FilterRel& filte
     logValidateMsg("native validation failed due to: Validation failed for input types in FilterRel.");
     return false;
   }
-  for (const auto& type : types) {
```
mskapilks, line 832: Removed this completely to check if all UTs passes, as there are specific validation for timestamp cases. If some case are missing, I can add better validation check which can validate if predicate is on timestamp column.

mskapilks:
The one failure is not related to my change.
```
error: building krb5:x64-linux-avx failed with: BUILD_FAILED
Elapsed time to handle krb5:x64-linux-avx: 1.5 s
Please ensure you're using the latest port files with `git pull` and `vcpkg update`.
Then check for known issues at:
```

rui-mo:
@mskapilks Just restart this test.

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -734,6 +734,34 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
...
+      runQueryAndCompare("select c1, c2 from ts where c1 = 1") {
+        checkOperatorMatch[FileSourceScanExecTransformer]
+      }
+
+      // Fallback should only happen when there is a filter on timestamp column
+      runQueryAndCompare(
+        "select c1, c2 from ts where" +
+          " c2 = timestamp'1965-01-01 10:11:12.123456'") {
+        df: DataFrame =>
+          assert(df.queryExecution.executedPlan.find(_.isInstanceOf[FileSourceScanExec]).isDefined)
```
rui-mo, line 753: I feel we don't need to check a fallback plan as native execution is preferred. To make sure the query is runnable with `runQueryAndCompare` is enough. How do you think?

rui-mo (commented):
Thanks. It makes sense to apply more specific fallback logic on timestamp.

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
mskapilks: Make sense

mskapilks:
The current failure seems due to some package download failure

zhli1142015:
Retriggered the tests.

rui-mo (approved):

zhli1142015 (approved):

PR #4444, opened 2024-01-18
[VL] Gluten-te: Invalidate docker local cache only when branch code is updated

zhztheplayer:
…s updated

This is an optimization following https://github.com/oap-project/gluten/pull/4429. We can still use cached build when code is not updated.


zhztheplayer:
@PZD-CHINA @PHILO-HE

PHILO-HE (approved):
This looks better! Thanks! cc @PZD-CHINA

PR #4447, opened 2024-01-18
[VL] Make velox orc scan configurable and keep it on by default.

SinghAsDev:

Velox ORC scan is leading to data correctness check failures. While ORC scan is fixed, this PR makes enabling of velox ORC scans configurable and turn it off by default.

## How was this patch tested?

Local build.

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -94,6 +94,9 @@ class GlutenConfig(conf: SQLConf) extends Logging {
   def enableCommonSubexpressionEliminate: Boolean =
     conf.getConf(ENABLE_COMMON_SUBEXPRESSION_ELIMINATE)
 
+  def veloxOrcScanEnabled: Boolean =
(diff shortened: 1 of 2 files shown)


shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -1594,4 +1597,11 @@ object GlutenConfig {
       .doc("Log granularity of AWS C++ SDK in velox.")
       .stringConf
       .createWithDefault("FATAL")
+
+  val VELOX_ORC_SCAN_ENABLED =
+    buildStaticConf("spark.gluten.sql.columnar.backend.velox.orc.scan.enabled")
+      .internal()
+      .doc(" Enable velox orc scan. If disabled, vanilla spark orc scan will be used.")
+      .booleanConf
+      .createWithDefault(false)
```
FelixYBW, line 1606: can we set default to true? ORC format is already used by some companies.

SinghAsDev: Sure, updated. I thought we would want to keep it off until the data correctness is resolved.

rui-mo (approved):

Issue #4452, opened 2024-01-19
[CH] get wrong hash table from cache


shuai-xu:
### Backend

CH (ClickHouse)

### Bug description

When run sql 12745, it failed with 
<img width="296" alt="image" src="https://github.com/oap-project/gluten/assets/20960838/f8806a2e-f61a-47cc-a26e-6d2479c20da1">
We found the root cause is that if a task has two joins, it may get a wrong hash table from cache for the second join.

### Spark version

Spark-3.3.x

PR #4453, opened 2024-01-19
[GLUTEN-4452] [CH] fix may get wrong hash table when multi joins in a task

shuai-xu:

This pr fix the bug that it may get wrong hash table when there are multi joins in a task

(Fixes: \#4452)

## How was this patch tested?

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/joins/ClickHouseBuildSideRelation.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/joins/ClickHouseBuildSideRelation.scala
@@ -45,14 +45,17 @@ case class ClickHouseBuildSideRelation(
       broadCastContext: BroadCastHashJoinContext): ClickHouseBuildSideRelation = this
 
   private var hashTableData: Long = 0L
+  private var hashTableName: String = ""
+
   def buildHashTable(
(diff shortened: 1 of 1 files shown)


backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/joins/ClickHouseBuildSideRelation.scala
```diff
@@ -45,14 +45,17 @@ case class ClickHouseBuildSideRelation(
       broadCastContext: BroadCastHashJoinContext): ClickHouseBuildSideRelation = this
 
   private var hashTableData: Long = 0L
+  private var hashTableName: String = ""
+
   def buildHashTable(
       broadCastContext: BroadCastHashJoinContext): (Long, ClickHouseBuildSideRelation) =
     synchronized {
-      if (hashTableData == 0) {
+      if (hashTableData == 0 || hashTableName != broadCastContext.buildHashTableId) {
```
zzcclp, line 53: use equals, not `!=`

shuai-xu: In scala , it seems that '!=' is better than equals.

PR #4455, opened 2024-01-19
[VL][Minor] Fix a format issue of unclosed braces

marin-ma:
`VeloxShuffleWriter.cc` has an unclosed braces in "PROCESSAVX".


marin-ma:
@FelixYBW Could you help to review? I'm not sure if we should keep the code in `PROCESSAVX`, as the it's out of maintenance for a long time.

FelixYBW:
> @FelixYBW Could you help to review? I'm not sure if we should keep the code in `PROCESSAVX`, as the it's out of maintenance for a long time.

Let's remove it since it's not used

jinchengchenghh (approved):

FelixYBW (approved):
thank you for clean up

PR #4459, opened 2024-01-19
[CORE] Optimize some code in HashAggregateExecBaseTransformer

liujiayi771:

1. We should move `aggregateFunction` validation to `doValidateInternal`.
2. For `getAttrForAggregateExprs`, we can use simpler implementation to get the correct attributes from `aggregateExpressions`.
3. For `ConverterUtils.getAttrFromExpr`, we can use the `Expression.references` method to obtain the desired Attribute. In the original implementation, when iterating over `BinaryLike` types, the left child is used for iteration. We can use `reference.head` to get the same Attribute.

Due to the inclusion of custom agg in ClickHouse, the original logic has been retained, and the `getAttrForAggregateExprs` has been abstracted as an interface to accommodate different logics implemented by Velox and ClickHouse.

## How was this patch tested?

Exists CI.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
@@ -23,8 +23,6 @@ import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.expressions._
-import org.apache.spark.sql.catalyst.expressions.aggregate._
-import org.apache.spark.sql.catalyst.optimizer._
 import org.apache.spark.sql.catalyst.plans._
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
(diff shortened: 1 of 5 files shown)


liujiayi771:
@rui-mo @zzcclp @PHILO-HE @ulysses-you Could you help review?

gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
```diff
@@ -23,61 +23,24 @@ import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
...
-      case u: Upper =>
-        getAttrFromExpr(u.child)
-      case ss: Substring =>
-        getAttrFromExpr(ss.children.head)
-      case other =>
-        throw new UnsupportedOperationException(
-          s"makeStructField is unable to parse from $other (${other.getClass}).")
+  def getAttrFromExpr(expr: Expression, skipAlias: Boolean = false): Attribute = {
+    if (skipAlias) {
+      expr.references.head
```
ulysses-you, line 41: what happens if the child if alias is not attribute ? it seems the behavior is different with previous

gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
```diff
@@ -23,61 +23,24 @@ import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
...
+    } else {
+      expr.transformDown { case alias: Alias => alias.toAttribute }.references.head
```
ulysses-you, line 43: do we support all the expressions ? previous we only support some

liujiayi771: The main purpose of this function is to retrieve the `Attribute` output by partial agg `aggregateExpressions`. The existing code only lists some of the current cases encountered. However, from the perspective of the function itself, it can adapt to all situations through reference.

gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
liujiayi771: f we don't skip the Alias and continue to reference it, the reference will also traverse its child.

gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
liujiayi771: @rui-mo Maybe you are more familiar with the logic here. Do we need to match supported `Expression`s here?

gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
ulysses-you: before, if skipAlias: alias(add(c1, c2)) => alias.toAttribute
this pr, if skipAlias: alias(add(c1, c2)) => c1

liujiayi771: @ulysses-you You are right, strictly speaking, the semantics are indeed inconsistent. I thought that "notSkipAlias" meant to traverse Alias, but I didn't notice the restriction that the child must be `AttributeReference`. However, I noticed that there is no call with "skipAlias" set to true in Gluten. So, we can directly remove this parameter.

gluten-core/src/main/scala/io/glutenproject/expression/ConverterUtils.scala
rui-mo: It would be better to support more expressions here. Previously we only listed the expressions encountered.

liujiayi771: @ulysses-you The original strategy was to add support for any encountered `Expression`s here, so there may be some cases that were overlooked. Now, it is able to cover all possible expressions.

ulysses-you (approved):

ulysses-you:
cc @rui-mo if you have other comments

PR #4460, opened 2024-01-19
[DNM debugging] [ICEBERG] Use reflection to access Iceberg local classes to avoid IlegalAccessError

SinghAsDev:

Use reflection to access Iceberg local classes to avoid IlegalAccessError

## How was this patch tested?

Local build and explain on a iceberg table scan.


FelixYBW (approved):
make sense to me

FelixYBW:
The issue is caused by class loader. We need to copy the gluten-iceberg jar to spark/jars

FelixYBW:
gluten-iceberg jar is build into a single gluten jar now. We need to copy the gluten jar to `spark/jars`, remove it from commandline `--jars`

PR #4471, opened 2024-01-22
[CORE] Add rule to rewrite multi-children count

ulysses-you:

Velox does not support multi-children count for now, see [CountAggregate](https://github.com/facebookincubator/velox/blob/b9af6b4c85a1c809c393bef1cbb02d2d8af86ed5/velox/functions/prestosql/aggregates/CountAggregate.cpp#L177-L178) . So this pr adds a new rule to rewrite the multi-children count to one child. The behavior follows the Spark, see [Count](https://github.com/apache/spark/blob/1f68dc73f69fa577619d9fdf99482c108e69dbe1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Count.scala#L90-L98).

For example
```
SELECT c1, count(distinct c2,c3), count(c3,c4) FROM t GROUP BY c1; 
```

## How was this patch tested?

Add tests


ulysses-you:
cc @PHILO-HE @rui-mo @zhztheplayer thank you

gluten-core/src/main/scala/io/glutenproject/extension/RewriteCount.scala
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * This Rule is used to rewrite the count for aggregates if:
+ *   - the count is partial
+ *   - the count has more than one child
+ *   - the aggregate expressions do not need row construct
+ *
+ * TODO: Remove this rule when Velox support multi-children Count
+ */
+object RewriteCount extends Rule[SparkPlan] {
```
rui-mo, line 38: Thanks! Generally LGTM. Can we add some notes here about how we rewrite the plan? From my reading, we will use If and isNull to replace the multiple children.

rui-mo, line 38: I'm having a question about the design. Why rule is preferred? It seems wrapping the multiple children of Count into a If clause during plan transformation should also work.

ulysses-you: sure, added the notes

ulysses-you: Rewrite count using rule is helpful to pull out pre-project. If we do it during transform then at the native side we would have one more project.

rui-mo: Makes sense. Thanks.

yaooqinn (approved):

PR #4473, opened 2024-01-22
[VL] Enable partition path lower unit test

JkSelf:

Enable the following two suite:

```
enableSuite[GlutenFileIndexSuite]
    .include("SPARK-20367 - properly unescape column names in inferPartitioning")
enableSuite[GlutenFileSourceSQLInsertTestSuite]
    .include(
      "SPARK-34556: checking duplicate static partition columns should respect case sensitive conf")
```

## How was this patch tested?

unit test


rui-mo (approved):
Thanks!

Issue #4480, opened 2024-01-22
[CH] Decouple LocalFiles from plan to improve driver generating substrait plan


exmy:
### Description

as velox does https://github.com/oap-project/gluten/pull/4177

PR #4457, opened 2024-01-19
[GLUTEN-4454][CH] Support write mergetree data for CH backend

zzcclp:
## What changes were proposed in this pull request?

Support write mergetree data for CH backend:

1. create table for the mergetree data, and also support to config the orderByKey, primaryKey for CH;
2. now support to write to partition table and bucket table;
3. insert overwrite doesn't support now

Example:

```
CREATE TABLE IF NOT EXISTS lineitem_mergetree_partition
(
 l_orderkey      bigint,
 l_partkey       bigint,
 l_suppkey       bigint,
 l_linenumber    bigint,
 l_quantity      double,
 l_extendedprice double,
 l_discount      double,
 l_tax           double,
 l_returnflag    string,
 l_linestatus    string,
 l_commitdate    date,
 l_receiptdate   date,
 l_shipinstruct  string,
 l_shipmode      string,
 l_comment       string,
 l_shipdate      date
)
USING clickhouse
PARTITIONED BY (l_shipdate, l_returnflag)
TBLPROPERTIES (orderByKey='l_orderkey',
               primaryKey='l_orderkey')
LOCATION '$basePath/lineitem_mergetree_partition'
```

```
insert into table lineitem_mergetree_partition
   select
   l_orderkey      ,
 l_partkey       ,
 l_suppkey       ,
 l_linenumber    ,
 l_quantity      ,
 l_extendedprice ,
 l_discount      ,
 l_tax           ,
 l_returnflag    ,
 l_linestatus    ,
 l_commitdate    ,
 l_receiptdate   ,
 l_shipinstruct  ,
 l_shipmode      ,
 l_comment,
 l_shipdate
    from lineitem
   where l_shipdate BETWEEN date'1992-01-01' AND date'1992-06-10'
```

(Fixes: #4454)

Close #4454 

Co-authored-by: liuneng <1398775315@qq.com>
Co-authored-by: Shuai li <loneylee@live.cn>

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #4481, opened 2024-01-22
[GLUTEN-4480][CH] Decouple LocalFiles from plan to improve driver generating substrait plan

exmy:

Applied for CH backend following velox does  https://github.com/oap-project/gluten/pull/4177

The time taken to generate the substrait plan has decreased from `249938ms` to `4486ms` for `163180` partitions after this pach.

(Fixes: \#4480)

## How was this patch tested?

Pass CI

### DIFF
--- a/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
@@ -19,11 +19,12 @@ package io.glutenproject.backendsapi.clickhouse
 import io.glutenproject.{GlutenConfig, GlutenNumaBindingInfo}
 import io.glutenproject.backendsapi.IteratorApi
 import io.glutenproject.execution._
+import io.glutenproject.expression.ConverterUtils
 import io.glutenproject.metrics.{GlutenTimeMetric, IMetrics, NativeMetrics}
 import io.glutenproject.substrait.plan.PlanNode
-import io.glutenproject.substrait.rel.{ExtensionTableBuilder, LocalFilesBuilder, SplitInfo}
+import io.glutenproject.substrait.rel.{ExtensionTableBuilder, ExtensionTableNode, LocalFilesBuilder, LocalFilesNode, SplitInfo}
 import io.glutenproject.substrait.rel.LocalFilesNode.ReadFileFormat
-import io.glutenproject.utils.{LogLevelUtil, SubstraitPlanPrinterUtil}
+import io.glutenproject.utils.LogLevelUtil
 import io.glutenproject.vectorized.{CHNativeExpressionEvaluator, CloseableCHColumnBatchIterator, GeneralInIterator, GeneralOutIterator}
 
 import org.apache.spark.{InterruptibleIterator, SparkConf, SparkContext, TaskContext}
@@ -35,7 +36,7 @@ import org.apache.spark.sql.connector.read.InputPartition
 import org.apache.spark.sql.execution.datasources.FilePartition
 import org.apache.spark.sql.execution.joins.BuildSideRelation
 import org.apache.spark.sql.execution.metric.SQLMetric
-import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.types.{StructField, StructType}
 import org.apache.spark.sql.utils.OASPackageBridge.InputMetricsWrapper
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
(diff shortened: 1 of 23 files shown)


cpp-ch/local-engine/Storages/SubstraitSource/SubstraitFileSource.cpp
```diff
@@ -65,15 +65,15 @@ static DB::Block getRealHeader(const DB::Block & header)
 }
 
 SubstraitFileSource::SubstraitFileSource(
-    DB::ContextPtr context_, const DB::Block & header_, const substrait::ReadRel::LocalFiles & file_infos)
+    DB::ContextPtr context_, const DB::Block & header_, const std::shared_ptr<substrait::ReadRel::LocalFiles> & file_infos)
```
lgbo-ustc, line 68: It seems no special reason to change file_info into a shared ptr if there is no need to hold it in `SubstraitFileSource`

zhztheplayer:
Thank you for working on this @exmy.

I noticed the PR doesn't make changes to the following two parts of code yet:

1. https://github.com/oap-project/gluten/blob/19a863c0d3781034b4def57be899e5bd4a8825b6/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala#L247-L280

2. https://github.com/oap-project/gluten/blob/19a863c0d3781034b4def57be899e5bd4a8825b6/backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/v2/ClickHouseAppendDataExec.scala#L234-L252

Do you have a plan to make cleanup to the code as well? Especially 2 since some of the code seems to be obsolete so can be removed probably. I am no sure so would you like to help doing a check? Thanks.

After that, we may be able to remove a series of APIs on `splitInfos` from `SubstraitContext` to get a cleaner design.

exmy:
> Do you have a plan to make cleanup to the code as well? Especially 2 since some of the code seems to be obsolete so can be removed probably. I am no sure so would you like to help doing a check? Thanks.
> 
> After that, we may be able to remove a series of APIs on `splitInfos` from `SubstraitContext` to get a cleaner design.

@zzcclp Could you help check if we need cleanup the two parts of code?

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
```diff
@@ -97,16 +97,26 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
...
     splitInfos.zipWithIndex.map {
-      case (splitInfos, index) =>
-        wsCtx.substraitContext.initSplitInfosIndex(0)
-        wsCtx.substraitContext.setSplitInfos(splitInfos)
-        val substraitPlan = wsCtx.root.toProtobuf
+      case (splits, index) =>
+        val splitInfosByteArray = splits.zipWithIndex.map {
+          case (split, i) =>
+            val filesNode = split.asInstanceOf[LocalFilesNode]
+            filesNode.setFileSchema(scans(i).getDataSchema)
```
exmy, line 109: @Yohahaha It seems LocalFilesNode doesn't set fileSchema and fileReadProperties in velox after decouping LocalFiles from plan, could you please check and verify whether velox needs it?

zzcclp:
> 1. genNativeFileScanRDD

change the code in the part of the `genNativeFileScanRDD` , and the `ClickHouseAppendDataExec.scala` is already removed in other pr #4457

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHIteratorApi.scala
Yohahaha: Velox use `base_schema` defined in ReadRel, instead of `schema` defined in LocalFiles.FileOrFiles, one read rel should only has one schema I think.

`fileReadProperties` is not needed in Velox for now.

exmy:
@zhztheplayer @zzcclp Have cleaned up the code in the part of the `genNativeFileScanRDD` except `ClickHouseAppendDataExec`. I'd like to remove the related APIs on `splitInfos` from `SubstraitContext` in the following pr.

zzcclp:
@exmy PR #4457 has been merged, please rebase.
BTW, there are some changes for LocalFileNode when query data from mergetree data

zhztheplayer:
LGTM for changes to common module 

@zzcclp If you want to review CH changes, thanks.

cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
```diff
@@ -458,9 +475,12 @@ QueryPlanPtr SerializedPlanParser::parseOp(const substrait::Rel & rel, std::list
         }
         case substrait::Rel::RelTypeCase::kRead: {
             const auto & read = rel.read();
-            assert(read.has_local_files() || read.has_extension_table() && "Only support local parquet files or merge tree read rel");
-            if (read.has_local_files())
+            // TODO: We still maintain the old logic of parsing LocalFiles or ExtensionTable in RealRel
+            // to be compatiable with some suites about metrics.
+            // Remove this compatiability in later and then only java iter has local files in ReadRel.
+            if (read.has_local_files() || (!read.has_extension_table() && !isReadFromMergeTree(read)))
```
zzcclp, line 481: don't add this parameter `isMergeTree`, can ignore the suites about metrics first and raist a issue to fix.

zzcclp: the paramter `isMergeTree` is confused.

liuneng1994 (approved):
LGTM

cpp-ch/local-engine/Parser/SerializedPlanParser.cpp
exmy: > don't add this parameter `isMergeTree`, can ignore the suites about metrics first and raist a issue to fix.

We need a way to identify whether to parse according to `LocalFiles` or `ExtensionTable` here. If we don't have `isMergeTree` parameter, are there alternative ways to achieve the same thing?

zzcclp: `read.has_extension_table()` is for mergetree, and `read.has_local_files()` is for other file format, parquet or orc don't need the extension table part.

exmy: `ReadRel` no longer includes `LocalFiles` or `ExtensionTable` since they have been decoupled from `RealRel` which is exactly the goal this pr want to achieve, right?

zzcclp: Got, can we judge the `split_info` is `localfiles` or `extendsiontable` according to the `split_info string`?

exmy: I tried before, but unfortunately, there wasn't a suitable way to judge it solely based on `split_info` string information.

baibaichen (approved):

Issue #4479, opened 2024-01-22
[VL] Map unbase64 to a Velox function name


fyp711:
### Description

The`from_base64` function provided by perstosql, it's the same functionality as `unbase64`, and I think we can map the sparksql function `unbase64`   to `from_base64`.

The Spark functions doc link is : https://spark.apache.org/docs/latest/api/sql/index.html#unbase64 
The Presto functions doc link is : https://prestodb.github.io/docs/current/functions/binary.html

PR #4482, opened 2024-01-22
[GLUTEN-4479][VL] Map unbase64 to a presto function

fyp711:

Thefrom_base64 function provided by perstosql, it's the same functionality as unbase64, and I think we can map the sparksql function unbase64 to from_base64.

The Spark functions doc link is : https://spark.apache.org/docs/latest/api/sql/index.html#unbase64
The Presto functions doc link is : https://prestodb.github.io/docs/current/functions/binary.html

(Issue-id: \#4479)

## How was this patch tested?

select unbase64(xxx) from tbName;


Yohahaha:
please also add a test in backend-velox module, like https://github.com/oap-project/gluten/blob/19a863c0d3781034b4def57be899e5bd4a8825b6/backends-velox/src/test/scala/io/glutenproject/execution/VeloxFunctionsValidateSuite.scala#L368-L386

fyp711:
> please also add a test in backend-velox module, like
> 
> https://github.com/oap-project/gluten/blob/19a863c0d3781034b4def57be899e5bd4a8825b6/backends-velox/src/test/scala/io/glutenproject/execution/VeloxFunctionsValidateSuite.scala#L368-L386

@Yohahaha I'm going to perfect it now.Thanks!

backends-velox/src/test/scala/io/glutenproject/execution/VeloxFunctionsValidateSuite.scala
```diff
@@ -383,4 +383,26 @@ class VeloxFunctionsValidateSuite extends VeloxWholeStageTransformerSuite {
...
+          .toDF("e")
+          .write
+          .parquet(path.getCanonicalPath)
+
+        spark.read.parquet(path.getCanonicalPath).createOrReplaceTempView("base64_tbl")
+
+        runQueryAndCompare("select unbase64(e) from base64_tbl") {
+          checkOperatorMatch[ProjectExecTransformer]
+        }
+    }
```
zhztheplayer, line 404: Why we need to write/read file for testing this function?

fyp711: Looks like we don't neet to write and read file test , just write a few fixed-value tests? Thanks!

fyp711:
Hi everyone, since the Base64-related functions provided by Velox do the some validations, I think we need to read and write the file to test, generating some base64-encoded strings. Thanks.

see : https://github.com/facebookincubator/velox/blob/1deb0368b222e1074d23ed3204762ad1940afecc/velox/common/encode/Base64.cpp#L342

backends-velox/src/test/scala/io/glutenproject/execution/VeloxFunctionsValidateSuite.scala
```diff
@@ -389,4 +389,22 @@ class VeloxFunctionsValidateSuite extends VeloxWholeStageTransformerSuite {
...
+          .write
+          .parquet(path.getCanonicalPath)
+
+        spark.read.parquet(path.getCanonicalPath).createOrReplaceTempView("unbase64_tbl")
+
+        runQueryAndCompare("SELECT unbase64(e) from unbase64_tbl") {
+          checkOperatorMatch[ProjectExecTransformer]
+        }
+    }
+  }
```
Yohahaha, line 409: maybe use such way
```java
withTable("t") {
      withSQLConf("gluten disable") {
        sql("create table && write base64 string from normal string")
      }
      withSQLConf("gluten enable") {
        runQueryAndCompare("select unbase64") {
          checkOperatorMatch[FileSourceScanExecTransformer]
        }
      }
    }
```

fyp711: It looks great, but I have a question. Why should we consider gluten disabled? 
Maybe we can directly like this? Thanks for your guidance!
> withTable("t") {
        sql("create table && write base64 string from normal string")
        runQueryAndCompare("select unbase64(xx) from t") {
          checkOperatorMatch[ProjectExecTransformer]
        }
    }

Yohahaha: you are right.

PR #4485, opened 2024-01-22
[VL] Update generic benchmark usage and doc

marin-ma:
Follow-up of #4177 

Main changes:
- Removed the `--skip-input` flag for the first stage, and added `--plan`, `--split` and `--data` command line flags. `--plan [path]` to specify the plan file of Substrait JSON string. To simulate a first stage, similarly to how users obtain the Substrait JSON string, they need to extract the split JSON string from the executor log, save it to a separate file, and use `--split [path]` to specify the file. To simulate a middle stage, user need to specify the parquet file of inputs with `--data [input_file_1] [input_file_2] ...`
- Refined the error log and documentation.


FelixYBW:
what's the main change?

marin-ma:
> what's the main change?

@FelixYBW Updated the PR description. Could you help to review? Thanks!

PR #4489, opened 2024-01-23
[CORE] Propagate fallback reason to union and shuffle exchange

ulysses-you:

This pr does two things

- propagate fallback reason to union and shuffle exchange during validation
- rename UnionExecTransformer to ColumnarUnionExec since we do not transform union in fact 

## How was this patch tested?

Pass CI


ulysses-you:
cc @PHILO-HE @zhouyuan thank you

PHILO-HE (approved):
Looks good. cc @rui-mo

PR #4507, opened 2024-01-24
[VL] Correct unit displayed on SQL UI for spill-related metrics

zhztheplayer:
Some "spilled bytes" metrics are using wrong type so displayed with wrong unit on SQL UI. This patch fixes the issue.


zhztheplayer:
@JkSelf @FelixYBW

FelixYBW (approved):
Thank you for fix

Issue #4502, opened 2024-01-24
[VL] Native validation failed due to: Invalid input type in casting: ARRAY/MAP/ROW/VARBINARY


PHILO-HE:
### Description

Will fix it.

Yohahaha:
we encounter same issue which is cast map to string.

PR #4509, opened 2024-01-24
[GLUTEN-4502][VL] Allow TIMESTAMP & complex types in cast expression

PHILO-HE:

This is just an attempt to offload cast with such types involved to velox. Maybe, at least, we still need to block nested complex types. 

(Fixes: #4502)

## How was this patch tested?

Existing tests & new test.


zhouyuan:
CC @wangyum

zhouyuan:
note: most of the failures seems due to window related unit tests

PHILO-HE:
> Casting TIMESTAMP is taken care of by #8743.

@JunhyungSong, actually we have merged several patches to support casting between timestamp and other types. And there are also some relevant patches merged into Velox. If you are using Gluten 1.2, it may be not easy to pick and apply all these patches.

Issue #4458, opened 2024-01-19
[CH] not support join on complex expression


shuai-xu:
### Backend

CH (ClickHouse)

### Bug description

When run sql 12701 which contains t1 join t2 on t1.uid=t2.uid and datediff(t2.day, t1.day) in (1,3,7), it throws exception:
<img width="365" alt="image" src="https://github.com/oap-project/gluten/assets/20960838/714c6383-af2b-4927-8606-4c7b34f72a98">
We should make it fallback to vanilia spark if clickhouse can not support it.

### Spark version

Spark-3.3.x

PR #4523, opened 2024-01-25
[GLUTEN-4458] [CH] fallback when has IN in join

shuai-xu:

Clickhouse doesn't support join like `select * from a join b on a.id=b.id and datediff(a.day,b.day) in (1,3); select * from a join b on a.value-b.value=1`, rollback it first.

(Fixes: \#4458)

## How was this patch tested?


baibaichen:
如果时 inner join 可以把不支持的部分搞成 post condition 去过滤

zhanglistar:
在这个PR里支持了issue里提到的join表达式，https://github.com/ClickHouse/ClickHouse/pull/46479  等CH社区进。

zzcclp:
LGTM if ci pass

zzcclp (approved):
LGTM

PR #4524, opened 2024-01-25
[GLUTEN-4751][VL] Enable storage partitioned join UTs with correct settings

yma11:

In storage partitioned join cases, SMJ shouldn't be replaced by `ShuffledHashJoin` thus we will set `spark.gluten.sql.columnar.forceShuffledHashJoin` to `false`. And in current test path for `GlutenKeyGroupedPartitioningSuite.scala`, AQE is enabled so actually the ops are all fallbacked. In this PR, we enable storage partitioned join UTs with correct settings and also introduce `collectColumnarShuffleExchangeExec` to check shuffle counts in Gluten plan.

## How was this patch tested?

UT


gluten-ut/spark33/src/test/scala/org/apache/spark/sql/connector/GlutenKeyGroupedPartitioningSuite.scala
```diff
@@ -15,16 +15,275 @@
  * limitations under the License.
  */
 package org.apache.spark.sql.connector
+import io.glutenproject.execution.SortMergeJoinExecTransformer
```
zhouyuan, line 18: should enable this in Spark34 also?

yma11:
@zhouyuan all updated. Please take a review again. Thanks.

zhouyuan (approved):
👍

PR #4531, opened 2024-01-26
[VL] Remove fmt lib from vcpkg.json

PHILO-HE:

I note fmt is already statically linked to velox libs. And gluten cpp doesn't depend on it.

## How was this patch tested?

Static build test in CI.


rui-mo (commented):
Thanks!

rui-mo (approved):

zhouyuan:
Note: we should try to link with the libs provided in vcpkg if possible, this will ensure the functionality in static build

PR #4532, opened 2024-01-26
[VL] Fix wrong native plan string

ulysses-you:

In validationMode, we should also respect the iterator index, otherwise the native plan string will always be `TableScan`.

## How was this patch tested?

add test


ulysses-you:
cc @Yohahaha @PHILO-HE  thank you

Yohahaha:
LGTM!

PHILO-HE (approved):
Thanks!

PR #4533, opened 2024-01-26
[VL] Optimize resource management for driver/test code 

zhztheplayer:
1. Remove unsafe APIs `tmpInstance` / `globalInstance` from resource classes;
2. Add a new unified API `TaskResource#runUnsafe` for all unsafe operations against resources, used in driver or test code. The new API would protect the unsafe code with a code block then the code should conform to a specific exit point for releasing all the resource registered in the code block. This way should be safer than the previous `tmpInstance` / `globalInstance` APIs.


gluten-core/src/main/scala/org/apache/spark/util/TaskResources.scala
```diff
@@ -39,18 +40,61 @@ object TaskResources extends TaskListener with Logging {
...
+  // be created and used. Since fallback registry is not managed by Spark's task memory manager,
+  // Spark may not be aware of the allocations happened inside the user code.
+  //
+  // The API should only be used in the following cases:
+  //
+  // 1. Run code on driver
+  // 2. Run test code
+  def runUnsafe[T](body: => T): T = {
+    TaskResources.setUnsafeTaskContext()
+    onTaskStart()
```
zhztheplayer, line 76: Should call `onTaskSucceeded0` /  `onTaskFailed`  although they are no-ops

shims/spark34/src/main/scala/org/apache/spark/TaskContextUtils.scala
```diff
@@ -0,0 +1,48 @@
...
+package org.apache.spark
+
+import org.apache.spark.executor.TaskMetrics
+import org.apache.spark.memory.{TaskMemoryManager, UnifiedMemoryManager}
+import org.apache.spark.metrics.MetricsSystem
+import org.apache.spark.network.util.ByteUnit
+
+import java.util.Properties
+
+object TaskContextUtils {
```
marin-ma, line 26: Why is it necessary to place this class in the shim layer?

marin-ma (dismissed):
Thanks!

shims/spark34/src/main/scala/org/apache/spark/TaskContextUtils.scala
zhztheplayer: It's needed mainly for 2 reasons:

1. Spark requires `SparkEnv` to be set to instantiate `TaskContextImpl` in its later version. Since we don't plan to set `SparkEnv` which is not thread-local in `runUnsafe` routine, we should manually pass `cpus` to skip `SparkEnv.get`
https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/TaskContextImpl.scala#L58

2. Spark 3.4 added `numPartitions` to the arg list
https://github.com/apache/spark/commit/f48602912b4d967a59085226345e056bc39d865e

marin-ma (approved):

PR #4538, opened 2024-01-26
[VL] Offload byte type scan

rui-mo:
Offload byte data type in parquet scan. It's already supported in velox parquet scan as tinyint.


FelixYBW (approved):
test passed

PR #4541, opened 2024-01-26
[GLUTEN-3559][VL] Fix "cannot divide by 0" exception for remainder function

ulysses-you:

Velox has a dedicated remainder function for spark sql, so we should better to use it instead of presto function.

## How was this patch tested?

enable related test

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
@@ -929,4 +929,16 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
         |""".stripMargin
     )(df => checkFallbackOperators(df, 1))
   }
+
+  test("Remainder with non-foldable right side") {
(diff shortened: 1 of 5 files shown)


ulysses-you:
cc @PHILO-HE @rui-mo thank you

PHILO-HE:
Can we fix it in Velox? With this fix, if the right input is not foldable, exception is still thrown. Right?

PHILO-HE:
I note velox has a function called [RemainderFunction](https://github.com/facebookincubator/velox/blob/main/velox/functions/sparksql/Arithmetic.h#L31), which produces null if right input is 0. It's registered name is "remainder". Maybe, we can just change the below code to map to this function.
https://github.com/oap-project/gluten/blob/main/cpp/velox/substrait/SubstraitParser.cc#L377

ulysses-you:
@PHILO-HE Velox won't throw expcetion if the right child is not foldable. I think it's better to use the existed velox spark function. Thank you for the reminder.

PHILO-HE (approved):
Thanks for your fix!

PR #4544, opened 2024-01-26
[VL] Support C2R and R2C between broadcast relations

zhztheplayer:
The patch removes the restriction that broadcast hash join and broadcast exchange should be enabled and validated at the same time. `spark.gluten.sql.columnar.broadcastExchange` `spark.gluten.sql.columnar.broadcastJoin` can be then turned on/off individually.

C2Rs / R2Cs will work as expected to convert between vanilla Spark broadcast relation and Gluten (Velox, as of now)'s broadcast relation.

By doing this, the related broadcast exchange+join coherent validation rules will be removed from core module. They will continue existing in `backends-clickhouse` module until we implement this feature for CH backend.

~~Require for https://github.com/oap-project/gluten/pull/4533~~


gluten-ut/spark33/src/test/scala/io/glutenproject/utils/velox/VeloxTestSettings.scala
```diff
@@ -1005,6 +1005,7 @@ class VeloxTestSettings extends BackendTestSettings {
       "SPARK-9083: sort with non-deterministic expressions"
     )
   enableSuite[GlutenDataFrameTimeWindowingSuite]
+    .exclude("time window joins") // FIXME hongze
```
zhztheplayer, line 1008: The test fails by
```
Input schema contains unsupported type when convert row to columnar for StructType(StructField(window,StructType(StructField(start,TimestampNTZType,true),StructField(end,TimestampNTZType,true)),false),StructField(othervalue,IntegerType,false)) due to do not support data type: TimestampNTZType
java.lang.UnsupportedOperationException: Input schema contains unsupported type when convert row to columnar for StructType(StructField(window,StructType(StructField(start,TimestampNTZType,true),StructField(end,TimestampNTZType,true)),false),StructField(othervalue,IntegerType,false)) due to do not support data type: TimestampNTZType
	at io.glutenproject.execution.RowToVeloxColumnarExec.$anonfun$doExecuteColumnarInternal$1(RowToVeloxColumnarExec.scala:52)
	at scala.Option.foreach(Option.scala:407)
	at io.glutenproject.execution.RowToVeloxColumnarExec.doExecuteColumnarInternal(RowToVeloxColumnarExec.scala:49)
	at io.glutenproject.execution.RowToColumnarExecBase.doExecuteColumnar(RowToColumnarExecBase.scala:62)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeColumnar$1(SparkPlan.scala:221)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:232)
	at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:229)
	at org.apache.spark.sql.execution.SparkPlan.executeColumnar(SparkPlan.scala:217)
	at io.glutenproject.backendsapi.velox.SparkPlanExecApiImpl.createBroadcastRelation(SparkPlanExecApiImpl.scala:332)
	at org.apache.spark.sql.execution.ColumnarBroadcastExchangeExec.$anonfun$relationFuture$2(ColumnarBroadcastExchangeExec.scala:79)
	at io.glutenproject.utils.Arm$.withResource(Arm.scala:25)
	at io.glutenproject.metrics.GlutenTimeMetric$.millis(GlutenTimeMetric.scala:37)
	at org.apache.spark.sql.execution.ColumnarBroadcastExchangeExec.$anonfun$relationFuture$1(ColumnarBroadcastExchangeExec.scala:69)
	at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withThreadLocalCaptured$1(SQLExecution.scala:191)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
```

Will fix in a later patch.

zhztheplayer:
@marin-ma @zhouyuan

zhztheplayer:
cc @Surbhi-Vijay

zhztheplayer:
@zzcclp Please review CH's changes, thanks.

backends-velox/src/main/scala/io/glutenproject/execution/ShuffledHashJoinExecTransformer.scala
```diff
@@ -152,6 +157,20 @@ case class ShuffledHashJoinExecTransformer(
     copy(left = newLeft, right = newRight)
 }
 
+case class VeloxBroadcastBuildSideRDD(
```
Surbhi-Vijay, line 160: This case class is not specific to hash join and can be moved to generic place or to a new file.

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  def sparkToVeloxUnsafe[F, T](
+      context: SparkContext,
+      mode: BroadcastMode,
+      schema: StructType,
+      from: Broadcast[F],
+      fn: Iterator[InternalRow] => Iterator[ColumnarBatch]): Broadcast[T] = {
+    // HashedRelation to ColumnarBuildSideRelation.
+    val fromBroadcast = from.asInstanceOf[Broadcast[HashedRelation]]
```
Surbhi-Vijay, line 66: Can you please add a check here for BroadcastMode? HashedRelation will only be present in case of `HashedRelationBroadcastMode` . It will be `Array[InternalRow]` in case of `IdentityBroadcastmode`

Surbhi-Vijay: just checked your comment  `FIXME: Add checking for broadcast mode.`

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
```diff
@@ -327,48 +325,24 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
...
-    val batches = countsAndBytes.filter(_._1 != 0).map(_._2)
-    val rawSize = batches.map(_.length).sum
-    if (rawSize >= BroadcastExchangeExec.MAX_BROADCAST_TABLE_BYTES) {
-      throw new SparkException(
-        s"Cannot broadcast the table that is larger than 8GB: ${rawSize >> 30} GB")
+    mode match {
+      case IdentityBroadcastMode =>
+        throw new IllegalStateException("Unreachable code")
+      case HashedRelationBroadcastMode(_, _) =>
+        val serialized: Array[ColumnarBatchSerializeResult] = child
```
Surbhi-Vijay, line 332: There is no hash table created in Gluten for the broadcast relation. `IdentityBroadcastMode` and `hashedRelationBroadcastMode` are actually same in case of Gluten.

Please check below comment  in `ColumnarBroadcastExchangeExec`
```
 //this created relation ignore HashedRelationBroadcastMode isNullAware, because we
            // cannot get child output rows, then compare the hash key is null, if not null,
            // compare the isNullAware, so gluten will not generate HashedRelationWithAllNullKeys
            // or EmptyHashedRelation, this difference will cause performance regression in some
            // cases.
```

This match block for mode can be removed safely when BNLJ is supported. Please correct me if my understanding is wrong.

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
```diff
@@ -0,0 +1,120 @@
...
+        throw new IllegalArgumentException("Unexpected partitioning: " + partitioning.toString)
+    }
+  }
+
+  def serializeStream(batches: Iterator[ColumnarBatch]): ColumnarBatchSerializeResult = {
+    val filtered = batches
+      .filter(_.numRows() != 0)
+      .map(
+        b => {
+          ColumnarBatches.retain(b)
```
Surbhi-Vijay, line 97: Can there be out of memory scenarios? since the original broadcast will be  there in heap memory and then again memory will be allotted until the serialization is finished.

backends-velox/src/test/scala/io/glutenproject/execution/VeloxTPCHSuite.scala
```diff
@@ -271,3 +273,23 @@ class VeloxPartitionedTableTPCHSuite extends VeloxTPCHSuite {
...
+      .set("spark.gluten.sql.columnar.broadcastExchange", "false")
+  }
+}
+
+class VeloxTPCHV1VanillaBhjGlutenBeSuite extends VeloxTPCHSuite {
+  override protected def sparkConf: SparkConf = {
+    super.sparkConf
+      .set("spark.sql.sources.useV1SourceList", "parquet")
+      .set("spark.sql.autoBroadcastJoinThreshold", "30M")
+      .set("spark.gluten.sql.columnar.broadcastJoin", "false")
```
Surbhi-Vijay, line 292: Should there be one more test case where both the configs `broadcastJoin` and `broadcastExchange` are enabled ?

Surbhi-Vijay (commented):
@zhztheplayer the changes looks good to me. Added some queries and minor suggestions.
Thanks.
- Surbhi

zzcclp:
LGTM, we will support this feature for CH backend later

backends-velox/src/main/scala/io/glutenproject/execution/ShuffledHashJoinExecTransformer.scala
zhztheplayer: I'm moving it out from this file. If you think it can be reused for BNLJ.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
zhztheplayer: The assumption is correct to me. Currently Gluten doesn't handle anything about `IdentityBroadcastMode` so it's reasonable to throw when it sees one. If BNLJ requires the same code with BHJ in this code block, then the one for `IdentityBroadcastMode` can be removed.

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
zhztheplayer: Sure. Will add some checks for mode.

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
zhztheplayer: If I understand correctly, you are considering the amplified memory consumption by the original hash relation + converted columnar batch + serialized byte arrays?

If so, it should become a problem when the broadcast threshold is set to a large number. We can optimize it in further development iterations with some benchmarks.

backends-velox/src/test/scala/io/glutenproject/execution/VeloxTPCHSuite.scala
zhztheplayer: The other existing suites should already cover the case you mentioned. I'll add some explicit config settings to those suits to make it clear.

PHILO-HE (approved):
Thanks!

PR #4550, opened 2024-01-27
[Core][Minor] Do not transform SortAggExec if columnar hashagg is disabled

surnaik:

When the plan contains SortAggregateExec, Gluten will tag the plan as not transformable if replaceSortAggWithHashAgg is disabled or columnarHashAgg is disabled. Even when the plan is marked as not transformable, Gluten will try to generateHashAggregateExecTransformer which is redundant in this case since plan is already marked as not transformable.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
@@ -471,24 +471,24 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
         case plan: SortAggregateExec =>
           if (!BackendsApiManager.getSettings.replaceSortAggWithHashAgg) {
             TransformHints.tagNotTransformable(plan, "replaceSortAggWithHashAgg is not enabled")
-          }
-          if (!enableColumnarHashAgg) {
+          } else if (!enableColumnarHashAgg) {
             TransformHints.tagNotTransformable(
               plan,
               "columnar HashAgg is not enabled in SortAggregateExec")
+          } else {
+            val rewrittenAgg = RewriteMultiChildrenCount.applyForValidation(plan)
+            val transformer = BackendsApiManager.getSparkPlanExecApiInstance
+              .genHashAggregateExecTransformer(
+                rewrittenAgg.requiredChildDistributionExpressions,
+                rewrittenAgg.groupingExpressions,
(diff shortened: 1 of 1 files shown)


surnaik:
@PHILO-HE @zhouyuan - Please review.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -471,24 +471,24 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
...
+              .genHashAggregateExecTransformer(
+                rewrittenAgg.requiredChildDistributionExpressions,
+                rewrittenAgg.groupingExpressions,
+                rewrittenAgg.aggregateExpressions,
+                rewrittenAgg.aggregateAttributes,
+                rewrittenAgg.initialInputBufferOffset,
+                rewrittenAgg.resultExpressions,
+                rewrittenAgg.child
+              )
+            TransformHints.tag(plan, transformer.doValidate().toTransformHint)
```
zhztheplayer, line 490: Does this mean this code always tags `TRANSFORM_UNSUPPORTED` currently, when the two if-conditions get hit? Since we have a check to avoid re-tagging `TRANSFORM_UNSUPPORTED` as `TRANSFORM_SUPPORTED`:

https://github.com/oap-project/gluten/blob/487d07951e5e11040e5450a7ca4b0128b7aea259/gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala#L97-L99

surnaik: Ya, that's right, if conditions pass, it'll always mark as unsupported. Isn't that the right behaviour? Since we're converting SortAggExec to Hashaggexe transformer, we shouldn't do it if the two conditions are disabled. The check to prevent TRANSFORM_UNSUPPORTED plan to get tagged again is currently the reason this works as expected, if not it would have tagged it as supported again. This PR adds an explicit check.

PHILO-HE (approved):
Yes, it's unnecessary to try to transform the plan if any of the two conditions is satisfied. Even though the re-tagging failure can also make it fallback, it is better to handle it explicitly as this pr proposed.

PHILO-HE:
Hi @zhztheplayer, do you have any other comment?

zhztheplayer (approved):

PR #4557, opened 2024-01-29
[VL] Put fmt lib back to vcpkg.json and bump the version

PHILO-HE:

Add fmt lib back to vcpkg.json to let vcpkg build the static lib. In velox, fmt_SOURCE=AUTO, so the lib will be firstly searched from system. If not found, velox will build from SOURCE. With fmt managed by vcpkg, velox will directly use the pre-built  lib found from vcpkg installation path. This pr also bumped the version to align with velox.

## How was this patch tested?

Compile pass.

### DIFF
--- a/dev/vcpkg/vcpkg.json
+++ b/dev/vcpkg/vcpkg.json
@@ -109,6 +109,6 @@
     }
   },
   "overrides": [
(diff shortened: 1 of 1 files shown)


zhouyuan (approved):
👍

PR #4564, opened 2024-01-29
[VL][CORE] Refactor and fix the columnar cartesian product

ulysses-you:

This pr refactors the code of `CartesianProductExecTransformer` to fix some issues, e.g., before the `CartesianProductExecTransformer` work with other join together would fail.

This pr adds a bridge `ColumnarCartesianProductBridge` connect left/right children and `CartesianProductExecTransformer` to avoid wrapping  them into one `WholeStageTransformer`. The reason is that, `WholeStageTransformer` always assume the input partitions are same, but it is not true for `CartesianProductExec`. The dag after this pr would like.

image

## How was this patch tested?

add tests

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
@@ -150,12 +150,6 @@ case class ColumnarCollapseTransformStages(
   /** Inserts an InputIteratorTransformer on top of those that do not support transform. */
   private def insertInputIteratorTransformer(plan: SparkPlan): SparkPlan = {
     plan match {
-      case p: CartesianProductExecTransformer =>
-        p.withNewChildren(
-          p.children.map(
-            child =>
-              ColumnarCollapseTransformStages.wrapInputIteratorTransformer(
-                insertWholeStageTransformer(child))))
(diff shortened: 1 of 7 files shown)


gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
```diff
@@ -150,12 +150,6 @@ case class ColumnarCollapseTransformStages(
   /** Inserts an InputIteratorTransformer on top of those that do not support transform. */
   private def insertInputIteratorTransformer(plan: SparkPlan): SparkPlan = {
     plan match {
-      case p: CartesianProductExecTransformer =>
-        p.withNewChildren(
-          p.children.map(
-            child =>
-              ColumnarCollapseTransformStages.wrapInputIteratorTransformer(
-                insertWholeStageTransformer(child))))
```
ulysses-you, line 158: We should avoid changing the core code for dedicated operator which makes the code hard to read and maintain.

ulysses-you:
cc @Surbhi-Vijay @zhouyuan thank you

zhouyuan (commented):
@ulysses-you thanks for the help! works for me. @Surbhi-Vijay Please also take a look. 

thanks, -yuan

Surbhi-Vijay (approved):

yaooqinn (approved):

PR #4573, opened 2024-01-30
[CORE] Add junitxml path of scalatest-maven-plugin for backends-velox, spark33 and spark34

Yohahaha:

See https://www.scalatest.org/user_guide/using_the_scalatest_maven_plugin, it helps CI tools collect UT results.

### DIFF
--- a/backends-velox/pom.xml
+++ b/backends-velox/pom.xml
@@ -215,6 +215,7 @@
         <groupId>org.scalatest</groupId>
         <artifactId>scalatest-maven-plugin</artifactId>
         <configuration>
+          <junitxml>.</junitxml>
           <tagsToExclude>${tagsToExclude}</tagsToExclude>
           <systemProperties>
             <velox.udf.lib.path>${cpp.build.dir}/velox/udf/examples/libmyudf.so</velox.udf.lib.path>
--- a/gluten-ut/spark33/pom.xml
+++ b/gluten-ut/spark33/pom.xml
@@ -156,6 +156,9 @@
       <plugin>
         <groupId>org.scalatest</groupId>
         <artifactId>scalatest-maven-plugin</artifactId>
+        <configuration>
+          <junitxml>.</junitxml>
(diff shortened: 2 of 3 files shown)


Yohahaha:
@PHILO-HE please help review this minor patch, thank you!

PHILO-HE:
Thanks for your work!
With this patch, some xml files will be generated under the configured path on CI server, right? Can we cover Spark32 tests?

Yohahaha:
> Thanks for your work! With this patch, some xml files will be generated under the configured path on CI server, right? Can we cover Spark32 tests?

Spark32 already has such config, https://github.com/oap-project/gluten/blob/bc4c1ac753c4ca1822f6fcb23eeadee448912ac9/gluten-ut/spark32/pom.xml#L155

zhztheplayer (commented):
Thanks. We already set the option in root pom:

https://github.com/oap-project/gluten/blob/bc4c1ac753c4ca1822f6fcb23eeadee448912ac9/pom.xml#L567

Would you figure out whether it's working or not?

Yohahaha:
> Thanks. We already set the option in root pom:
> 
> https://github.com/oap-project/gluten/blob/bc4c1ac753c4ca1822f6fcb23eeadee448912ac9/pom.xml#L567
> 
> Would you figure out whether it's working or not?

I think not, you can check your CI machine's `backends-velox/target/scalatest-reports` before this patch and this patch.

zhztheplayer:
Did you find out why it's not working? May it relate to relative/abstract path or something?

Yohahaha:
> Did you find out why it's not working? May it relate to relative/abstract path or something?

nope, `-DtestFailureIgnore=true` also not work.

Yohahaha:
can we merge?

PHILO-HE (approved):
Looks good! Thanks!
If there is any other comment, let's fix it in follow-up PR.

PR #4582, opened 2024-01-30
[GLUTEN-3582][CORE][VL][CH] Refactor filter pushdown logic

baibaichen:

Vanilla spark just push down part of filter condition into scan, however gluten can push down all filters in some cases, it was done in #4132. But, Clickhouse backend can not do this in parquet file format. we can add this later

This PR add `postProcessPushDownFilter` in `SparkPlanExecApi`,  `CHSparkPlanExecApi` overwrite this funtion to add its own logic.

### Note for CH
Consider TPCH 22 `c_acctbal > (select avg(c_acctbal) from customer where ...)`.   it can be push down before this PR.  I disable it for ch backend, since I want to make push down functionality same as vanilla spark first.

## Update for `ScanTransformerFactory`
Before this PR, `createxxxTransformer` assume pass the extra ilters, after this PR, we pass all push down filters, this is because we may remove unsupported filters  

```scala 
def createFileSourceScanTransformer(
      scanExec: FileSourceScanExec,
      reuseSubquery: Boolean,
      allPushDownFilters: Option[Seq[Expression]] = None,  // extraFilters: Seq[Expression] = Seq.empty,
      validation: Boolean = false): FileSourceScanExecTransformer = {}

  def createBatchScanTransformer(
      batchScan: BatchScanExec,
      reuseSubquery: Boolean,
      allPushDownFilters: Option[Seq[Expression]] = None,  //pushdownFilters: Seq[Expression] = Seq.empty,
      validation: Boolean = false): SparkPlan = {}
```

## How was this patch tested?
Using existed UT

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
@@ -384,30 +384,27 @@ object FilterHandler extends PredicateHelper {
     (ExpressionSet(filters) -- ExpressionSet(scanFilters)).toSeq
 
   // Separate and compare the filter conditions in Scan and Filter.
-  // Push down the remaining conditions in Filter into Scan.
+  // Try push down the remaining conditions in Filter into Scan.
   def applyFilterPushdownToScan(filter: FilterExec, reuseSubquery: Boolean): SparkPlan =
     filter.child match {
       case fileSourceScan: FileSourceScanExec =>
-        val remainingFilters =
-          getRemainingFilters(
-            fileSourceScan.dataFilters,
-            splitConjunctivePredicates(filter.condition))
+        val pushDownFilters =
+          BackendsApiManager.getSparkPlanExecApiInstance.postProcessPushDownFilter(
+            splitConjunctivePredicates(filter.condition),
+            fileSourceScan)
         ScanTransformerFactory.createFileSourceScanTransformer(
           fileSourceScan,
           reuseSubquery,
-          extraFilters = remainingFilters)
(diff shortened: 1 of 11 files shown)


baibaichen:
@liujiayi771 please review it.

zhztheplayer:
Changes to common module looks good to me. Thanks.

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -384,30 +384,27 @@ object FilterHandler extends PredicateHelper {
     (ExpressionSet(filters) -- ExpressionSet(scanFilters)).toSeq
 
   // Separate and compare the filter conditions in Scan and Filter.
-  // Push down the remaining conditions in Filter into Scan.
+  // Try push down the remaining conditions in Filter into Scan.
```
liujiayi771, line 387: nit: Try to push down.

liujiayi771:
LGTM. Thanks.

gluten-core/src/main/scala/io/glutenproject/execution/BasicPhysicalOperatorTransformer.scala
baibaichen: Let's fix in the follow up PR

zhouyuan (approved):
👍

abbas-gadhia:
Hello @baibaichen 
you mentioned above that "however gluten can push down all filters in some cases". How is this possible when the underlying parquest scan does not support filters such as `concat(name, surname) = "XY"` or `abs(profit) = 20`

PR #4607, opened 2024-02-01
[CORE] Fix limit with offset for Spark3.4

ulysses-you:

We should pass the specified offset in global limit to the native operator,  since Spark3.4 supports offset feature. Native topK does not offset, so this pr add fallback if topK offset is not 0.

## How was this patch tested?

Enable some tests

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
@@ -27,7 +27,7 @@ import io.glutenproject.utils.PhysicalPlanSelector
 import org.apache.spark.api.python.EvalPythonExecTransformer
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression, SortOrder}
+import org.apache.spark.sql.catalyst.expressions.{AttributeReference, Expression}
 import org.apache.spark.sql.catalyst.plans.FullOuter
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.catalyst.trees.TreeNodeTag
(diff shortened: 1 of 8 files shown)


gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -694,28 +696,14 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
...
-              val sortPlan = SortExecTransformer(plan.sortOrder, false, inputTransformer)
-              val limitPlan = LimitTransformer(sortPlan, 0, plan.limit)
-              tagged = limitPlan.doValidate()
-            }
-
-            if (tagged.isValid) {
-              val projectPlan = ProjectExecTransformer(plan.projectList, plan.child)
-              tagged = projectPlan.doValidate()
-            }
-            TransformHints.tag(plan, tagged.toTransformHint)
```
ulysses-you, line 718: move to `TakeOrderedAndProjectExecTransformer`

ulysses-you (commented):
cc @JkSelf @PHILO-HE thank you

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -432,6 +432,10 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::FetchRel& fetchR
   if (fetchRel.has_input()) {
     topNFlag = fetchRel.input().has_sort();
     if (topNFlag) {
+      if (fetchRel.offset() != 0) {
```
JkSelf, line 435: @ulysses-you Do we need this check in native side? It seems we have validate it in scala side [here](https://github.com/oap-project/gluten/pull/4607/files#diff-5024a9ba02c16355af248920e6bd13de93a1149d1e0510e10bc6f7d31de9045fR50).

ulysses-you: I think we can remove it

PHILO-HE (approved):
Looks good! Thanks!

PR #4613, opened 2024-02-02
[VL] Move RewriteMultiChildrenCount Rule before AddTransformHintRule

liujiayi771:

Because `RewriteMultiChildrenCount` Rule will create new `Expression` in aggregation that needs to be pull out, we should move this rule before `AddTransformHintRule`, so that we could pull out pre/post-project before `AddTransformHintRule`.

## How was this patch tested?

Exists CI in `TestOperator`.


liujiayi771:
@ulysses-you Could you help review?

PR #4619, opened 2024-02-02
[CORE] Move PullOutPreProject after AddTransformHintRule

liujiayi771:
It is a better choice to modify the `SparkPlan` only when `SparkPlan` itself supports execution by Gluten, so that the fallback `SparkPlan` remains unchanged.

For agg, we found that there are some corner cases that cannot be executed correctly when codegen is enabled after pulling out the project, therefore, we will put the pulling out process after `AddTransformHintRule`, perform the validation process first, and then pull out the `SparkPlan` that has passed validation.

## How was this patch tested?

Exists CI.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
@@ -16,9 +16,8 @@
  */
 package io.glutenproject.extension.columnar
 
-import io.glutenproject.utils.{PhysicalPlanSelector, PullOutProjectHelper}
+import io.glutenproject.utils.PullOutProjectHelper
 
-import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.expressions.{Expression, NamedExpression, SortOrder}
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.{ProjectExec, SortExec, SparkPlan, TakeOrderedAndProjectExec}
@@ -32,15 +31,12 @@ import scala.collection.mutable
  * to transform the SparkPlan at the physical plan level, constructing a SparkPlan that supports
  * execution by the native engine.
  */
-case class PullOutPreProject(session: SparkSession)
-  extends Rule[SparkPlan]
-  with PullOutProjectHelper {
+object PullOutPreProject extends Rule[SparkPlan] with PullOutProjectHelper {
 
   private def needsPreProject(plan: SparkPlan): Boolean = {
     if (notSupportTransform(plan)) {
-      // TagBeforeTransformHits may tag the plan as not transformable for some reason,
-      // for example, when ansi mode is enabled. In such cases, there is no need to
-      // pull out the project.
+      // We should not pull out pre-project for SparkPlan that already been tagged as
+      // not transformable.
       return false
     }
     plan match {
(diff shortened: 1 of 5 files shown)


liujiayi771:
@zhztheplayer Could you help review?

liujiayi771:
@ulysses-you Could you help to review?

ulysses-you:
> For agg, we found that there are some corner cases that cannot be executed correctly when codegen is enabled after pulling out the project

What's wrong with vanilla Spark after pulling out project? How does it work after we changing the applying order?

liujiayi771:
@ulysses-you As with offline communication, it was found that some UTs were stuck in codegen. Disabling codegen allows for normal execution. The current process is to first validate the modified sort and agg, and if transform is not supported, the project will not be pulled out. If transform is supported, it will be pulled out later, and the `AddTransformHintRule` will be executed again to validate the project. 

I think that we should try to avoid modifying the Spark plan as much as possible, especially if the plan is ultimately executed by Spark itself.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
```diff
@@ -83,38 +79,53 @@ case class PullOutPreProject(session: SparkSession)
...
   }
 
-  override def apply(plan: SparkPlan): SparkPlan = PhysicalPlanSelector.maybe(session, plan) {
-    plan.transform {
-      case sort: SortExec if needsPreProject(sort) =>
-        val expressionMap = new mutable.HashMap[Expression, NamedExpression]()
-        val newSortOrder = getNewSortOrder(sort.sortOrder, expressionMap)
+  override def apply(plan: SparkPlan): SparkPlan = {
+    val transformedPlan = plan.transform(applyLocally)
+    AddTransformHintRule().apply(transformedPlan)
```
ulysses-you, line 84: Can we move `AddTransformHintRule` out ? it seems we just apply `AddTransformHintRule` one more time after pull out pre project.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
```diff
@@ -83,38 +79,53 @@ case class PullOutPreProject(session: SparkSession)
...
-        newTopK
-    }
+      val newSort = sort.copy(sortOrder = newSortOrder, child = preProject)
+      newSort.copyTagsFrom(sort)
+
+      // The pre-project and post-project of SortExec always appear together, so it's more
+      // convenient to handle them together. Therefore, SortExec's post-project will no longer
+      // be pulled out separately.
+      val postProject = ProjectExec(sort.child.output, newSort)
+      postProject.setTagValue(PULLOUT_PROJECT_TAG, true)
```
ulysses-you, line 118: Can we do `unsetTagValue` original plan manutally after pulling out project ? Better to avoid introducing new tag.

liujiayi771: Do you mean inserting `AddTransformHintRules` again in `preOverrides` after `PullOutPreProject`?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
liujiayi771: Do you mean inserting `AddTransformHintRules` again in `preOverrides` after `PullOutPreProject`?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
liujiayi771: After the plan transformation is completed, it's difficult to identify which projects are pulled out projects.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
ulysses-you: yeah, just pull it out to make `preOverrides` clear

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
ulysses-you: why do we need to know whether a project is a pull out project ?

liujiayi771: @ulysses-you When pulling out a project, because there is no tag, Spark will copy the tag of the "sort" to the pulled-out project. We need to find this project in order to unset the tag values in it.

We can also unset the `TRANSFORM_SUPPORTED` tag for sort, but this will require an additional validation for sort. Is that what you mean?

ulysses-you: > Can we do unsetTagValue original plan manutally after pulling out project ? Better to avoid introducing new tag.

I mean we can unsetTagValue for sort after we pulling out project, so there is no tag Spark would copy.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
liujiayi771: OK. It is better.

ulysses-you (approved):
lgtm, thank you

PR #4620, opened 2024-02-02
[CORE] Untangle C2R / R2C transforming code

zhztheplayer:
This patch reworks `TransformPostOverrides` which converts C2R / R2C to native version to simplify code. Will pull out 2 rules from `TransformPostOverrides`:

 * RemoveTopmostColumnarToRow
 * RemoveGlutenTableCacheColumnarToRow

Current UTs should be able to cover the change:
1. GlutenFallbackSuite
2. VeloxAdaptiveQueryExecSuite
3. VeloxColumnarCacheSuite

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
@@ -593,68 +593,22 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
 
 // This rule will try to convert the row-to-columnar and columnar-to-row
 // into native implementations.
-case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[SparkPlan] {
+case class TransformPostOverrides() extends Rule[SparkPlan] {
   val columnarConf: GlutenConfig = GlutenConfig.getConf
   @transient private val planChangeLogger = new PlanChangeLogger[SparkPlan]()
 
-  def transformColumnarToRowExec(plan: ColumnarToRowExec): SparkPlan = {
-    if (columnarConf.enableNativeColumnarToRow) {
-      val child = replaceWithTransformerPlan(plan.child)
-
-      if (!PlanUtil.outputNativeColumnarData(child)) {
-        TransformHints.tagNotTransformable(plan, "child is not gluten plan")
-        plan.withNewChildren(plan.children.map(replaceWithTransformerPlan))
-      } else {
-        logDebug(s"ColumnarPostOverrides ColumnarToRowExecBase(${child.nodeName})")
-        val nativeConversion =
-          BackendsApiManager.getSparkPlanExecApiInstance.genColumnarToRowExec(child)
-        val validationResult = nativeConversion.doValidate()
-        if (validationResult.isValid) {
-          nativeConversion
-        } else {
-          TransformHints.tagNotTransformable(plan, validationResult)
-          plan.withNewChildren(plan.children.map(replaceWithTransformerPlan))
-        }
-      }
-    } else {
-      plan.withNewChildren(plan.children.map(replaceWithTransformerPlan))
-    }
-  }
-
-  def replaceWithTransformerPlan(plan: SparkPlan): SparkPlan = plan match {
-    case plan: RowToColumnarExec =>
-      val child = replaceWithTransformerPlan(plan.child)
+  def replaceWithTransformerPlan(plan: SparkPlan): SparkPlan = plan.transformDown {
+    case RowToColumnarExec(child) =>
       logDebug(s"ColumnarPostOverrides RowToColumnarExec(${child.getClass})")
       BackendsApiManager.getSparkPlanExecApiInstance.genRowToColumnarExec(child)
-    // The ColumnarShuffleExchangeExec node may be the top node, so we cannot remove it.
-    // e.g. select /* REPARTITION */ from testData, and the AQE create shuffle stage will check
-    // if the transformed is instance of ShuffleExchangeLike, so we need to remove it in AQE
-    // mode have tested gluten-it TPCH when AQE OFF
-    case ColumnarToRowExec(child: ColumnarShuffleExchangeExec) if isAdaptiveContext =>
-      replaceWithTransformerPlan(child)
-    case ColumnarToRowExec(child: ColumnarBroadcastExchangeExec) =>
-      replaceWithTransformerPlan(child)
-    case ColumnarToRowExec(child: BroadcastQueryStageExec) =>
-      replaceWithTransformerPlan(child)
-    // `InMemoryTableScanExec` internally supports ColumnarToRow
-    case ColumnarToRowExec(child: SparkPlan) if PlanUtil.isGlutenTableCache(child) =>
-      child
-    case plan: ColumnarToRowExec =>
-      transformColumnarToRowExec(plan)
-    case r: SparkPlan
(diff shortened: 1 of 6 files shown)


gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -594,68 +594,22 @@ case class TransformPreOverrides(isAdaptiveContext: Boolean)
...
-    case ColumnarToRowExec(child: BroadcastQueryStageExec) =>
-      replaceWithTransformerPlan(child)
-    // `InMemoryTableScanExec` internally supports ColumnarToRow
-    case ColumnarToRowExec(child: SparkPlan) if PlanUtil.isGlutenTableCache(child) =>
-      child
-    case plan: ColumnarToRowExec =>
-      transformColumnarToRowExec(plan)
-    case r: SparkPlan
-        if !r.isInstanceOf[QueryStageExec] && !r.supportsColumnar &&
-          r.children.exists(_.isInstanceOf[ColumnarToRowExec]) =>
```
zhztheplayer, line 647: Should double-check if the check `!r.isInstanceOf[QueryStageExec]` is still needed.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -666,6 +620,42 @@ case class TransformPostOverrides(isAdaptiveContext: Boolean) extends Rule[Spark
...
+  override def apply(plan: SparkPlan): SparkPlan = {
+    if (!isAdaptiveContext) {
+      // The rule only applies in AQE. If AQE is off the topmost C2R will be strictly required
+      // by Spark.
+      return plan
+    }
+    plan match {
+      case ColumnarToRowLike(child: ColumnarShuffleExchangeExec) => child
+      case ColumnarToRowLike(child: ColumnarBroadcastExchangeExec) => child
+      case ColumnarToRowLike(child: BroadcastQueryStageExec) => child
```
zhztheplayer, line 636: Ditto. Should check the code source of this line.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: Should be safe to remove since it's impossible to have QueryStageExec's child set as C2R because we always remove the C2R in rule `RemoveTopmostColumnarToRow`. Otherwise Spark throws error from `AdaptiveSparkPlanExec#newQueryStage`.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: The statement seems to be redundant after going through the code, removing

rui-mo (approved):

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
```diff
@@ -236,19 +235,9 @@ case class ExpandFallbackPolicy(isAdaptiveContext: Boolean, originalPlan: SparkP
...
+    val transformPostOverrides = TransformPostOverrides()
     val planWithTransitions = InsertTransitions.insertTransitions(originalPlan, outputsColumnar)
-    planWithTransitions.transform {
-      case c2r @ ColumnarToRowExec(_: ShuffleQueryStageExec) =>
-        transformPostOverrides.transformColumnarToRowExec(c2r)
-      case c2r @ ColumnarToRowExec(_: AQEShuffleReadExec) =>
-        transformPostOverrides.transformColumnarToRowExec(c2r)
-      // `InMemoryTableScanExec` itself supports columnar to row
-      case ColumnarToRowExec(child: SparkPlan) if PlanUtil.isGlutenTableCache(child) =>
-        child
```
ulysses-you, line 248: We should apply `RemoveGlutenTableCacheColumnarToRow` here manually since we sperate `TransformPostOverrides` into 3 rules.
BTW, some tests which I added in `VeloxColumnarCacheSuite` are used for Spark3.5 so may not work for now...

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -808,8 +797,10 @@ case class ColumnarOverrideRules(session: SparkSession)
 
   private def postOverrides(): List[SparkSession => Rule[SparkPlan]] =
     List(
-      (_: SparkSession) => TransformPostOverrides(isAdaptiveContext),
-      (s: SparkSession) => VanillaColumnarPlanOverrides(s)
+      (_: SparkSession) => TransformPostOverrides(),
+      (s: SparkSession) => VanillaColumnarPlanOverrides(s),
+      (s: SparkSession) => RemoveGlutenTableCacheColumnarToRow(s),
+      (s: SparkSession) => RemoveTopmostColumnarToRow(s, isAdaptiveContext)
```
ulysses-you, line 803: Shall we apply `RemoveGlutenTableCacheColumnarToRow` and `RemoveTopmostColumnarToRow` first ? To reduce the tree depth before going to transform.

gluten-core/src/main/scala/io/glutenproject/extension/ExpandFallbackPolicy.scala
zhztheplayer: I assume `RemoveGlutenTableCacheColumnarToRow` would eventually be applied? By looking into this code:

https://github.com/oap-project/gluten/blob/67978a666835dd83989cecd4168cf67276b723f1/gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala#L832-L847

`fallbackPolicy()` would apply before `postOverrides()` in which we populate ``RemoveGlutenTableCacheColumnarToRow` rule is applied.

zhztheplayer: Aha I see. The fallback plan is not doing `postOverrides()`.  It was my mistake.

ulysses-you: It won't be applied if the plan is fallback.. we should put it into `finallyRules`  if we want to apply one rule finally.

zhztheplayer: > It won't be applied if the plan is fallback.. we should put it into `finallyRules` if we want to apply one rule finally.

Yes that's just what I am trying to do. PR updated, please take another look. Thanks.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: It's feasible but I actually be inclined to put it a little bit back since there might be some rules adding C2R / R2C by themselves after `TransformPostOverrides`. E.g., `VanillaColumnarPlanOverrides` is adding C2Rs and R2Cs (of course, it may not add topmost C2R). So `RemoveTopmostColumnarToRow` can guard against all bad cases to prevent the AQE error.

rui-mo (approved):

ulysses-you (dismissed):

rui-mo (dismissed):

ulysses-you (dismissed):

rui-mo (approved):

PR #4621, opened 2024-02-02
[VL] Use the same result type for decimal round

rui-mo:

Spark 3.2 provides different result type with Spark 3.3 and 3.4 for decimal round. To make them work, a result type check in Velox was disabled. This PR changes to use the same result type for decimal round.

## How was this patch tested?

Verified unit tests with https://github.com/oap-project/velox/pull/476.


JkSelf:
@rui-mo Do we need to enable related uts?

rui-mo:
> @rui-mo Do we need to enable related uts in gluten?

Relating tests for decimal round are enabled in Gluten.

JkSelf (approved):
LGTM. Thanks.

PR #4629, opened 2024-02-03
[GLUTEN-4213][UT] Overwrite udf/udf-group-by.sql in spark34

liujiayi771:

https://github.com/oap-project/gluten/pull/4628 will pull out unsupported expressions to project, and agg will no longer fallback.
So the answer of udf(corr) is not completely identical to Spark. We need to overwrite `udf/udf-group-by.sql` in spark34, the same as what we did in spark 33 and spark 32.

* Remove corr case in `udf/udf-group-by.sql`
* Remove corr result in `udf-group-by.sql.out`

## How was this patch tested?

Exists CI.


liujiayi771:
@rui-mo Can you take a look?

rui-mo (commented):
> So the answer of udf(corr) is not completely identical to Spark. 

Could you share more details about the result difference? Thanks.

liujiayi771:
@rui-mo Velox answer is 0.99999999... while Spark answer is 1.0.

liujiayi771:
@rui-mo https://github.com/oap-project/gluten/pull/3774

liujiayi771:
```
- udf/udf-group-by.sql - Scala UDF *** FAILED ***
  udf/udf-group-by.sql - Scala UDF
  Expected "[1.0	1.0]	3", but got "[0.9999999999999999	0.9999999999999999]	3" Result did not match for query #26
  SELECT corr(DISTINCT x, y), udf(corr(DISTINCT y, x)), count(*)
    FROM (VALUES (1, 1), (2, 2), (2, 2)) t(x, y) (GlutenSQLQueryTestSuite.scala:803)
```

marin-ma (approved):
Thanks!

PR #4641, opened 2024-02-04
[VL] Use unscaled value as partition value for decimal type

rui-mo:

When parsing partition values, Velox regards decimal string is formatted with unscaled value ([link](https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/SplitReader.cpp#L280)). This PR aligns the string format with that assumption.

## How was this patch tested?

under test


PHILO-HE (approved):

rui-mo:
@PHILO-HE Thanks!

PR #4663, opened 2024-02-06
[CORE] Support pull out project for WindowExec

liujiayi771:

* Pulling out pre/post-project for `WindowExec`.
* `Expression` in `WindowExpression` will no longer cause a fallback.

## How was this patch tested?

Enable `Expression in WindowExpression that will fallback` test case in `GlutenSQLWindowFunctionSuite` as not fallback case.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
@@ -26,7 +26,6 @@ import org.apache.spark.sql.execution.aggregate.{BaseAggregateExec, TypedAggrega
 import org.apache.spark.sql.execution.window.WindowExec
 
 import scala.collection.mutable
-import scala.collection.mutable.ArrayBuffer
 
 /**
  * The native engine only supports executing Expressions within the project operator. When there are
@@ -165,27 +164,9 @@ object PullOutPreProject extends Rule[SparkPlan] with PullOutProjectHelper {
         window.partitionSpec.map(replaceExpressionWithAttribute(_, expressionMap))
 
       // Handle windowExpressions.
-      val postWindowExpressions = new ArrayBuffer[NamedExpression]()
-      val newWindowExpressions = window.windowExpression.toIndexedSeq
-        .map {
-          _.transform { case we: WindowExpression => rewriteWindowExpression(we, expressionMap) }
-        }
-        .map {
-          case alias @ Alias(_: WindowExpression, _) =>
-            postWindowExpressions += alias.toAttribute
-            alias
-          case other =>
-            // Directly use the output of WindowExpression, and move expression evaluation to
-            // post-project for computation.
-            assert(hasWindowExpression(other))
-            val we = other.collectFirst { case w: WindowExpression => w }.get
-            val rewriteWindowExpr = rewriteWindowExpression(we, expressionMap)
-            val alias = Alias(rewriteWindowExpr, generatePostAliasName)()
-            postWindowExpressions += other
-              .transform { case _: WindowExpression => alias.toAttribute }
-              .asInstanceOf[NamedExpression]
-            alias
-        }
+      val newWindowExpressions = window.windowExpression.toIndexedSeq.map {
+        _.transform { case we: WindowExpression => rewriteWindowExpression(we, expressionMap) }
+      }
 
       val newWindow = window.copy(
         orderSpec = newOrderSpec,
@@ -196,7 +177,7 @@ object PullOutPreProject extends Rule[SparkPlan] with PullOutProjectHelper {
           window.child)
       )
 
-      ProjectExec(window.child.output ++ postWindowExpressions, newWindow)
+      ProjectExec(window.output, newWindow)
 
     case _ => plan
   }
--- a/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
@@ -18,13 +18,15 @@ package io.glutenproject.extension.columnar
 
 import io.glutenproject.utils.PullOutProjectHelper
 
-import org.apache.spark.sql.catalyst.expressions.{Expression, Literal, NamedExpression, SortOrder}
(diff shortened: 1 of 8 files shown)


liujiayi771:
cc @ulysses-you, could you help review?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
```diff
@@ -149,6 +155,49 @@ object PullOutPreProject extends Rule[SparkPlan] with PullOutProjectHelper {
...
+      val postWindowExpressions = new ArrayBuffer[NamedExpression]()
+      val newWindowExpressions = window.windowExpression.toIndexedSeq
+        .map {
+          _.transform { case we: WindowExpression => rewriteWindowExpression(we, expressionMap) }
+        }
+        .map {
+          case alias @ Alias(_: WindowExpression, _) =>
+            postWindowExpressions += alias.toAttribute
+            alias
+          case other =>
```
ulysses-you, line 177: It should not happen. The window expression must be `Alias(_: WindowExpression, _)`

liujiayi771: I seem to have encountered a unit test that does not match `Alias(_: WindowExpression, _)`. I will post the corresponding plan later.

liujiayi771: @ulysses-you `GlutenSQLWindowFunctionSuite.Gluten - Expression in WindowExpression` has a `Alias(Cast(Divide(WindowExpression, Literal)))` case.
```
Window [cast((avg(UnscaledValue(c_acctbal#5)) windowspecdefinition(c_nationkey#4, c_custkey#3 ASC NULLS FIRST, specifiedwindowframe(RangeFrame, unboundedpreceding$(), currentrow$())) / 100.0) as decimal(11,6)) AS avg(c_acctbal) OVER (PARTITION BY c_nationkey ORDER BY c_custkey ASC NULLS FIRST RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)#10], [c_nationkey#4], [c_custkey#3 ASC NULLS FIRST]
```
image

This situation will occur when the input is decimal.

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/execution/GlutenSQLWindowFunctionSuite.scala
```diff
@@ -40,7 +39,7 @@ class GlutenSQLWindowFunctionSuite extends SQLWindowFunctionSuite with GlutenSQL
     Row(4953, 10, decimal(6037.28)),
     Row(35403, 5, decimal(6034.70)),
     Row(35803, 12, decimal(5284.87)),
-    Row(60865, 5, decimal(-227.83)),
+    Row(60865, 5, decimal(-227.82)),
```
liujiayi771, line 42: The modification here is because the original input resulted in an output of Velox being 4464.706667, while ClickHouse was 4464.706666, which should be due to the difference in backend calculation precision and is unrelated to the functionality being tested here. After adjusting the input, the results from Velox and ClickHouse can be made consistent.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
ulysses-you: ah I see

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
```diff
@@ -149,6 +155,49 @@ object PullOutPreProject extends Rule[SparkPlan] with PullOutProjectHelper {
...
+            // Directly use the output of WindowExpression, and move expression evaluation to
+            // post-project for computation.
+            assert(hasWindowExpression(other))
+            val we = other.collectFirst { case w: WindowExpression => w }.get
+            val rewriteWindowExpr = rewriteWindowExpression(we, expressionMap)
```
ulysses-you, line 182: do we need to rewrite it again? it seems we have rewritten all window expressions in previous transform line171

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
ulysses-you: is it possible that we only need post project ? it seems we only go into the pattern if need pre project.

liujiayi771: @ulysses-you Window and sort are similar, if the `orderSpec` contains an `Expression`, then both pre-project and post-project are definitely required. For `partitionSpec`, only pre-project is needed; for `windowExpression`, only post-project is required, but currently, they are handled together. If we were to extract the `windowExpression` part and handle it separately in the `PullOutPostProject` rule, it could also work. When both `orderSpec` and `windowExpression` contain Expressions, it might produce two post-projects, but the `CollapseProjectTransformer` rule would merge them together. Do you think the part that handles `windowExpression` should be extracted to the `PullOutPostProject` rule?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
liujiayi771: Yes, no need to rewrite it again, we can just use `we`.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
ulysses-you: I'm fine to handle post project at here together. My concern is that, the `needsPreProject` method does not cover the case if the window does not need pre project but need post project. We should add post project if the windowExpression child is not the `Alias(_: WindowExpression, _)`.

liujiayi771: Ok, I know. I think it is better to move this part to `PullOutPostProject` and add this check in `needsPostProject`, I will fix this later.

ulysses-you (approved):

Issue #4670, opened 2024-02-07
[VL] Fix split time metric for celeborn 


kerwin-zk:
### Backend

VL (Velox)

### Bug description

The split time of Celeborn should be similar to that of Columnar Shuffle, but the gap is currently quite large.
Celeborn：
![image](https://github.com/oap-project/gluten/assets/107825064/dcc95f2a-a6fc-4202-9e60-471d382f44e6)

Columnar Shuffle：
![image](https://github.com/oap-project/gluten/assets/107825064/da9387c7-3ddb-401a-bd61-9dc664da604e)

### Spark version

Spark-3.3.x

PR #4671, opened 2024-02-07
[GLUTEN-4670][VL] Fix split time metric for celeborn

kerwin-zk:
Fixes: #4670

## How was this patch tested?
CI

### DIFF
--- a/gluten-core/src/main/java/io/glutenproject/vectorized/SplitResult.java
+++ b/gluten-core/src/main/java/io/glutenproject/vectorized/SplitResult.java
@@ -71,7 +71,7 @@ public class SplitResult {
   }
 
   public long getTotalPushTime() {
-    return totalBytesEvicted;
+    return totalEvictTime;
   }
 
   public long[] getPartitionLengths() {


kerwin-zk:
@ulysses-you cc

cpp/core/shuffle/rss/CelebornPartitionWriter.cc
```diff
@@ -55,24 +55,28 @@ arrow::Status CelebornPartitionWriter::evict(
     Evict::type evictType,
     bool reuseBuffers,
     bool hasComplexType) {
-  rawPartitionLengths_[partitionId] += inMemoryPayload->getBufferSize();
+  int64_t spillTime{0};
```
ulysses-you, line 58: I do not get this issue, can you point out why this local variable can fix it ?

cpp/core/shuffle/rss/CelebornPartitionWriter.cc
```diff
@@ -55,24 +55,28 @@ arrow::Status CelebornPartitionWriter::evict(
...
+  {
+    ScopedTimer timer(&spillTime);
```
kerwin-zk, line 60: @ulysses-you  This should be ```&spillTime```, which has been modified. ```spillTime``` will record the time of each ```evict```, and then accumulate and record the total time through ```spillTime_```.

ulysses-you: I did not find the difference.. Each evict would call this method, and the previous spillTime_ would record the method elapsed time.

kerwin-zk: The difference is that before, ```spillTime_``` only recorded the time of one evict after the shuffle write was completed, but after the modification, it recorded the total time of evict.

ulysses-you (approved):

PR #4705, opened 2024-02-18
[VL] Adapting the bind reference of agg that contains subquery in agg expressions

liujiayi771:

When `aggregateExpressions` includes a subquery, Spark's `PlanAdaptiveSubqueries` Rule will transform the subquery within the final aggregation. The `aggregateFunction` in the `aggregateExpressions` of the final aggregation will be cloned, resulting in creating new `aggregateFunction` objects. The `inputAggBufferAttributes` will also generate new `AttributeReference` instances with larger `exprId`, which leads to a failure in binding with the output of the partial aggregation. We need to adapt to this situation; when encountering a failure to bind, it is necessary to allow the binding of `inputAggBufferAttribute` with the same name but different `exprId`.

## How was this patch tested?

Add a test case in `VeloxAggregateFunctionsSuite`.

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
@@ -706,6 +706,12 @@ abstract class VeloxAggregateFunctionsSuite extends VeloxWholeStageTransformerSu
                          |from values (5), (-10), (15) AS tab(c);
                          |""".stripMargin)(
       df => assert(getExecutedPlan(df).count(_.isInstanceOf[HashAggregateExecTransformer]) == 2))
+
+    runQueryAndCompare("""
+                         |select sum(if(c > (select sum(a) from values (1), (-1) AS tab(a)), 1, -1))
+                         |from values (1L, 5), (1L, -10), (2L, 15) AS tab(sum, c) group by sum;
+                         |""".stripMargin)(
+      df => assert(getExecutedPlan(df).count(_.isInstanceOf[HashAggregateExecTransformer]) == 2))
   }
 }
 
--- a/backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
@@ -699,6 +699,14 @@ abstract class VeloxAggregateFunctionsSuite extends VeloxWholeStageTransformerSu
           getExecutedPlan(df).count(plan => plan.isInstanceOf[HashAggregateExecTransformer]) >= 2)
     }
   }
+
+  test("bind reference failed when subquery in agg expressions") {
+    runQueryAndCompare("""
+                         |select sum(if(c > (select sum(a) from values (1), (-1) AS tab(a)), 1, -1))
+                         |from values (5), (-10), (15) AS tab(c);
+                         |""".stripMargin)(
+      df => assert(getExecutedPlan(df).count(_.isInstanceOf[HashAggregateExecTransformer]) == 2))
+  }
 }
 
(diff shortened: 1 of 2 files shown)


liujiayi771:
@PHILO-HE Could you help to review? This bug causes some agg fallback.

liujiayi771:
Before this PR:
image
After this PR:
image

PHILO-HE:
Seems there are some relevant test failure. Please fix it. Thanks!

ulysses-you (commented):
It seems aggregate buffer attribute name is not unique, e.g., two sum aggregate functions in one aggregate `SELECT sum(select c1 from t), sum(select c2 from t) FROM t`. How about use aggregate buffer attribute index ?

liujiayi771:
> It seems aggregate buffer attribute name is not unique, e.g., two sum aggregate functions in one aggregate `SELECT sum(select c1 from t), sum(select c2 from t) FROM t`. How about use aggregate buffer attribute index ?

I just switched to using index, can you help review the latest code?

liujiayi771:
> Seems there are some relevant test failure. Please fix it. Thanks!

These failed UTs include some other scenarios that need to be addressed. The current code has been modified to be more comprehensive, and these failed UTs used to be fallback.

gluten-core/src/main/scala/io/glutenproject/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -244,12 +244,36 @@ abstract class HashAggregateExecBaseTransformer(
...
+                  // in the aggregateExpressions of the final aggregation will be cloned, resulting
+                  // in creating new aggregateFunction object. The inputAggBufferAttributes will
+                  // also generate new AttributeReference instances with larger exprId, which leads
+                  // to a failure in binding with the output of the partial agg. We need to adapt
+                  // to this situation; when encountering a failure to bind, it is necessary to
+                  // allow the binding of inputAggBufferAttribute with the same name but different
+                  // exprId.
+                  val attrsWithSameName = originalInputAttributes.collect {
+                    case a if a.name == attr.name => a
+                  }
```
ulysses-you, line 262: can we pull out this code to avoid doing it for every inputAggBufferAttribute ?

ulysses-you, line 262: What happens if originalInputAttributes (grouping keys) has an attribute naing `sum`,`count`, etc ? Can we add a test case for that case ?

liujiayi771: This is a bad case; perhaps we should remove the group key in front of `originalInputAttributes`, because the output of the final agg's child is always the group key ++ `inputAggBufferAttributes`.

backends-velox/src/test/scala/io/glutenproject/execution/VeloxAggregateFunctionsSuite.scala
```diff
@@ -699,6 +699,20 @@ abstract class VeloxAggregateFunctionsSuite extends VeloxWholeStageTransformerSu
...
   }
+
+  test("bind reference failed when subquery in agg expressions") {
+    runQueryAndCompare("""
+                         |select sum(if(c > (select sum(a) from values (1), (-1) AS tab(a)), 1, -1))
+                         |from values (5), (-10), (15) AS tab(c);
+                         |""".stripMargin)(
+      df => assert(getExecutedPlan(df).count(_.isInstanceOf[HashAggregateExecTransformer]) == 2))
+
+    runQueryAndCompare("""
```
liujiayi771, line 710: @ulysses-you Add a new test case.

ulysses-you:
lgtm if test pass

ulysses-you (approved):

PHILO-HE:
Thanks for your great work! I just created an issue for possible reference in the future.

PR #4719, opened 2024-02-20
[CORE] Code refactoring for HashAggregateExecBaseTransformer

liujiayi771:

As the HashAggregate transformer code has been gradually modified, many of the originally defined interfaces have diversified into implementations across different backends, and many that are only used in subclasses no longer need to be abstracted as interfaces. It is necessary to clean them up.

## How was this patch tested?

No logic changing, only moving codes.
Exists CI.


liujiayi771:
cc @PHILO-HE @ulysses-you, thanks.

PHILO-HE (approved):
Looks good! Thanks!

PR #4731, opened 2024-02-21
[CORE] Remove tag TRANSFORM_SUPPORTED and rename transform-hint to fallback-hint: Phase 1

zhztheplayer:
Since we now introduced a set of validation rules and backend-specific validation rules that all tend to add transform-hint tags to plan nodes, it more or less becomes confusing to tag a plan as `TRANSFORM_SUPPORTED`. Before the plan is finally implemented within Gluten transformers, a plan node can be turned from `TRANSFORM_SUPPORTED` to `TRANSFORM_UNSUPPORTED` in any particular validation rule. So if a plan is tagged as `TRANSFORM_SUPPORTED`, it's never guaranteed to be executed natively by Gluten.

To eliminate this confusing state and ease understanding, the PR (and next PR as phase 2) targets to remove tag `TRANSFORM_SUPPORTED` and rename the term `transform-hint` to `fallback-hint`. When a plan is tagged with the hint, then it's considered a plan to be fallback and being run in Vanilla Spark.

Phase 1 (this PR):
Remove tag TRANSFORM_SUPPORTED.

Phase 2 (in next PR):
Rename transform-hint to fallback-hint.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/RewriteSparkPlanRulesManager.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/RewriteSparkPlanRulesManager.scala
@@ -16,7 +16,7 @@
  */
 package io.glutenproject.extension
 
-import io.glutenproject.extension.columnar.{AddTransformHintRule, TRANSFORM_SUPPORTED, TransformHint, TransformHints}
+import io.glutenproject.extension.columnar.{AddTransformHintRule, TRANSFORM_UNSUPPORTED, TransformHint, TransformHints}
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
@@ -46,7 +46,7 @@ case class RewrittenNodeWall(originalChild: SparkPlan) extends LeafExecNode {
 class RewriteSparkPlanRulesManager(rewriteRules: Seq[Rule[SparkPlan]]) extends Rule[SparkPlan] {
 
   private def mayNeedRewrite(plan: SparkPlan): Boolean = {
-    TransformHints.getHintOption(plan).isEmpty && {
+    TransformHints.isTransformable(plan) && {
       plan match {
         case _: SortExec => true
         case _: TakeOrderedAndProjectExec => true
@@ -58,13 +58,19 @@ class RewriteSparkPlanRulesManager(rewriteRules: Seq[Rule[SparkPlan]]) extends R
     }
   }
 
-  private def getTransformHintBack(origin: SparkPlan, rewrittenPlan: SparkPlan): TransformHint = {
+  private def getTransformHintBack(
+      origin: SparkPlan,
+      rewrittenPlan: SparkPlan): Option[TransformHint] = {
     // The rewritten plan may contain more nodes than origin, here use the node name to get it back
     val target = rewrittenPlan.collect {
       case p if p.nodeName == origin.nodeName => p
     }
     assert(target.size == 1)
-    TransformHints.getHint(target.head)
+    if (TransformHints.isTransformable(target.head)) {
+      None
+    } else {
+      Some(TransformHints.getHint(target.head))
+    }
   }
 
   private def applyRewriteRules(origin: SparkPlan): (SparkPlan, Option[String]) = {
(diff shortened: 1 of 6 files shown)


zhztheplayer:
@rui-mo @ulysses-you

gluten-core/src/main/scala/org/apache/spark/sql/execution/GlutenFallbackReporter.scala
```diff
@@ -58,38 +58,40 @@ case class GlutenFallbackReporter(glutenConfig: GlutenConfig, spark: SparkSessio
...
-                        // add the new reason
-                        lastReason + "; " + reason
-                      }
-                  }
-                  .getOrElse(reason)
-                logicalPlan.setTagValue(FALLBACK_REASON_TAG, newReason)
-            }
-
-          case _ =>
+        if (TransformHints.isNotTransformable(plan)) {
```
ulysses-you, line 61: We can merge this if to previous line to reduce the code diff,
`case plan: SparkPlan if TransformHints.isNotTransformable(plan) =>`

gluten-core/src/main/scala/io/glutenproject/extension/RewriteSparkPlanRulesManager.scala
```diff
@@ -58,13 +58,19 @@ class RewriteSparkPlanRulesManager(rewriteRules: Seq[Rule[SparkPlan]]) extends R
...
     val target = rewrittenPlan.collect {
       case p if p.nodeName == origin.nodeName => p
     }
     assert(target.size == 1)
-    TransformHints.getHint(target.head)
+    if (TransformHints.isTransformable(target.head)) {
+      None
+    } else {
+      Some(TransformHints.getHint(target.head))
+    }
```
ulysses-you, line 73: do we need this change ? it seems verbose than before

zhztheplayer: The code looks simpler before, but implicitly hid the fact that the rule's logic depends on `TRANSFORM_SUPPORTED` which is considered unreliable after this patch.

The if/else just pulled the unreliable logic closer to API calls, say, to call `isTransformable` which should be finally deprecated and removed. So it becomes easier for further code cleanups because of this change.

zhztheplayer: Can refer to the code comment https://github.com/oap-project/gluten/pull/4731/files/3f03b483e23d8b2f64a924ab473b1630b95d4379#diff-af786f23393cb8ac3fb3018ea6062722ce6d7388b1a69d3cd512ac7f1ab69435R71-R77

ulysses-you: thank you for the explanation

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -62,16 +61,25 @@ object TransformHints {
...
   def isTransformable(plan: SparkPlan): Boolean = {
-    plan.getTagValue(TAG).get.isInstanceOf[TRANSFORM_SUPPORTED]
-  }
-
-  def isNotTransformable(plan: SparkPlan): Boolean = {
-    plan.getTagValue(TAG).get.isInstanceOf[TRANSFORM_UNSUPPORTED]
+    getHintOption(plan) match {
+      case None => true
+      case _ => false
+    }
```
rui-mo, line 82: If `isNotTransformable(plan)` returns true, it means plan cannot be offloaded for sure. If `isNotTransformable(plan)` returns false, it means the plan could or could not be offloaded. Is this correct?

zhztheplayer: > If `isNotTransformable(plan)` returns true, it means plan cannot be offloaded for sure. If `isNotTransformable(plan)` returns false, it means the plan could or could not be offloaded. Is this correct?

That's correct.

rui-mo: Could we add some comments for this function to clarify? Thanks.

zhztheplayer: Sure, will add some in next commit. Thanks.

rui-mo (dismissed):

rui-mo (approved):

gluten-core/src/main/scala/org/apache/spark/sql/execution/GlutenFallbackReporter.scala
```diff
@@ -88,9 +88,11 @@ case class GlutenFallbackReporter(glutenConfig: GlutenConfig, spark: SparkSessio
                   .getOrElse(reason)
                 logicalPlan.setTagValue(FALLBACK_REASON_TAG, newReason)
             }
-
+          case TRANSFORM_UNSUPPORTED(_, _) =>
```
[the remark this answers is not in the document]
zhztheplayer: That's a good idea. Let's do that in another patch? Since the change here is just a code cleanup which doesn't change the behavior.

PR #4735, opened 2024-02-21
[CORE] Add metadata exclusion list for native write

ulysses-you:

This pr adds a new config `spark.gluten.sql.native.writeFilesColumnMetadataWhiteList` to allow native write files even if there are some metadata. By default, set with `comment` which is quite common if the source table column has comment and the write files column reference it directly. 

## How was this patch tested?

add test

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -1326,6 +1326,14 @@ object GlutenConfig {
       .booleanConf
       .createOptional
 
+  val NATIVE_WRITE_FILES_COLUMN_METADATA_WHITE_LIST =
+    buildConf("spark.gluten.sql.native.writeFilesColumnMetadataWhiteList")
+      .doc(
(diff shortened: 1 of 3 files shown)


ulysses-you:
cc @PHILO-HE thank you

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -1326,6 +1326,14 @@ object GlutenConfig {
       .booleanConf
       .createOptional
 
+  val NATIVE_WRITE_FILES_COLUMN_METADATA_WHITE_LIST =
+    buildConf("spark.gluten.sql.native.writeFilesColumnMetadataWhiteList")
```
PHILO-HE, line 1330: "whiteList" looks misleading, as the listed items are actually removed from output. Maybe, `xxx.WriteColumnMetadataExclusionList`?

ulysses-you: I'm fine with this new name, changed it

PHILO-HE (approved):
Looks good! Thanks!

PR #4736, opened 2024-02-21
[VL] Allow updating native hdfs credential

ulysses-you:

Allow update native hdfs credential if the new credential is different with before.

Related to https://github.com/oap-project/gluten/pull/4184

## How was this patch tested?

Pass CI


zhztheplayer (approved):
Thanks!

Issue #3559, opened 2023-10-30
[VL] Track all the failed unit test in Spark 3.4.


JkSelf:
### Description

Our plan is to ensure that Spark 3.4 passes all unit tests. This issue will track all the failed unit tests in Spark 3.4, and we will work on fixing them one by one.

**GlutenSQLQueryTestSuite**

- [ ] - describe-query.sql *** FAILED ***
- [ ] - sql-compatibility-functions.sql *** FAILED ***
- [x] - cast.sql *** FAILED ***
- [x] - show-tables.sql *** FAILED ***
- [x] - intersect-all.sql *** FAILED ***
- [x] - cte-nested.sql *** FAILED ***
- [x] - random.sql *** FAILED ***
- [ ] - datetime-formatting-invalid.sql *** FAILED ***
- [x] - group-by.sql *** FAILED ***
- [x] - show-create-table.sql *** FAILED ***
- [x] - cte-nonlegacy.sql *** FAILED ***
- [x] - inline-table.sql *** FAILED ***
- [x] - describe-table-column.sql *** FAILED ***
- [x] - map.sql *** FAILED ***
- [x] - describe.sql *** FAILED ***
- [ ] - datetime-parsing-invalid.sql *** FAILED ***
- [x] - show_columns.sql *** FAILED ***
- [x] - comments.sql *** FAILED ***
- [x] - udaf.sql *** FAILED ***
- [x] - json-functions.sql *** FAILED ***
- [ ] - describe-table-after-alter-table.sql *** FAILED ***
- [x] - show-tblproperties.sql *** FAILED ***
- [x] - grouping_set.sql *** FAILED ***
- [x] - charvarchar.sql *** FAILED ***
- [x] - table-valued-functions.sql *** FAILED ***
- [x] - regexp-functions.sql *** FAILED ***
- [x] - ilike-any.sql *** FAILED ***
- [x] - change-column.sql *** FAILED ***
- [x] - string-functions.sql *** FAILED ***
- [x] - columnresolution-negative.sql *** FAILED ***
- [x] - group-by-ordinal.sql *** FAILED ***
- [x] - timestamp-ntz.sql *** FAILED ***
- [x] - interval.sql *** FAILED ***
- [x] - date.sql *** FAILED ***
- [x] - bitwise.sql *** FAILED ***
- [x] - natural-join.sql *** FAILED ***
- [x] - timezone.sql *** FAILED ***
- [x] - ilike-all.sql *** FAILED ***
[...]

PR #4737, opened 2024-02-21
[GLUTEN-3559][VL] Rewrite GlutenInsertSuite test cases with default values

Surbhi-Vijay:
Additional support was added in Spark-3.4 for default values in parquet file scans.  
https://issues.apache.org/jira/browse/SPARK-39265

While scanning the files if the column with default value does not have any value then reader appends the default value to it. So, even if the column with default value was added later, file scan still provides values for all records (existing as well as new ones).

Velox does not support back filling the existing records while scan. So, if the column with default value was added later then it will provide null as column value for existing records.  
This is a behavior difference and not an inconsistent behavior. Users can update the existing data by running DML commands.

This PR, rewrites those testcases with default value in Gluten.

(Fixes:  #3559)

## How was this patch tested?
Unit tests are passing


Surbhi-Vijay:
@JkSelf @PHILO-HE Please review!

ayushi-agarwal (approved):

ayushi-agarwal:
@Surbhi-Vijay An issue for this can be opened: Velox does not support back filling the existing records while scan

PHILO-HE:
@Surbhi-Vijay, could you please rebase the code and resolve the conflicts? Thanks!

PHILO-HE (approved):
Thanks for your efforts!

PR #4743, opened 2024-02-22
[VL] Preserve JNI libraries after driver/executor process exits in debug mode

zhztheplayer:
Some debugging utilities (e.g., jemalloc memory profiler) may require for the libraries to obtain debug symbols. Currently the libraries are deleted after process exits, the patch helps Gluten preserve them in debug mode (spark.gluten.sql.debug=true, spark.gluten.sql.debug.keepJniWorkspace=true).


ep/build-velox/src/modify_velox.patch
```diff
@@ -69,20 +69,21 @@ index 82fed089f..91e4cc0df 100644
...
 +++ b/third_party/CMakeLists.txt
-@@ -26,7 +26,8 @@ if(VELOX_ENABLE_ARROW)
+@@ -26,7 +26,9 @@ if(VELOX_ENABLE_ARROW)
    endif()
    set(ARROW_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/arrow_ep")
    set(ARROW_CMAKE_ARGS
 -      -DARROW_PARQUET=OFF
 +      -DARROW_PARQUET=ON
 +      -DARROW_FILESYSTEM=ON
++      -DARROW_PROTOBUF_USE_SHARED=OFF
```
zhztheplayer, line 82: This fixes a regression once fixed by https://github.com/oap-project/gluten/pull/174

cpp/core/CMakeLists.txt
```diff
@@ -328,3 +328,8 @@ install(TARGETS gluten
         DESTINATION ${CMAKE_INSTALL_LIBDIR})
 install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/resources/libhdfs.so
         DESTINATION ${LIBHDFS3_DESTINATION})
+
+add_custom_command(TARGET gluten POST_BUILD COMMAND ld $<TARGET_FILE:gluten> || true
+        COMMENT "Checking ld result of libgluten.so")
+add_custom_command(TARGET gluten POST_BUILD COMMAND ldd $<TARGET_FILE:gluten> || true
+        COMMENT "Checking ldd result of libgluten.so")
```
zhztheplayer, line 335: Adding short CMake messages to check ld / ldd result of the built library. Could be useful to help troubleshooting symbol-related issues in CI.

zhztheplayer:
@PHILO-HE @zhouyuan

cpp/core/CMakeLists.txt
Yohahaha: any particular reason to keep these commands in git repo? build/link issue should be fixed before changes merged, keep above commands in git bring confused/noisy information when first build and every after.

if it's really useful, maybe we could move it to developer docs.

PR #4747, opened 2024-02-22
[VL] Link lib jemalloc produced by custom building

PHILO-HE:

Fix undefined symbol issue related to jemalloc. Currently, we need to add `--build_jemalloc=ON` when building gluten if some jemalloc functions are expected to be explicitly called. 

## How was this patch tested?

Local test.


cpp/CMake/Buildjemalloc_pic.cmake
```diff
@@ -0,0 +1,77 @@
...
+
+  file(MAKE_DIRECTORY "${JEMALLOC_INCLUDE_DIR}")
+  add_library(jemalloc::libjemalloc STATIC IMPORTED)
+  set_target_properties(
+    jemalloc::libjemalloc
+    PROPERTIES INTERFACE_LINK_LIBRARIES Threads::Threads
+               IMPORTED_LOCATION "${JEMALLOC_STATIC_LIB}"
+               INTERFACE_INCLUDE_DIRECTORIES
+               "${JEMALLOC_INCLUDE_DIR}")
+  add_dependencies(jemalloc::libjemalloc protobuf_ep)
```
marin-ma, line 76: Looks like a bug. Should be `add_dependencies(jemalloc::libjemalloc jemalloc_ep)`

FelixYBW:
error reported when build with vcpkg=on on ubuntu20.04:
```
[100%] Linking CXX shared library ../releases/libvelox.so
/usr/bin/ld: CMakeFiles/velox.dir/udf/UdfLoader.cc.o: in function `gluten::UdfLoader::loadUdfLibraries0(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)':
/root/gluten/cpp/velox/udf/UdfLoader.cc:53: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /usr/lib/x86_64-linux-gnu/libresolv.a(res_send.o): relocation R_X86_64_PC32 against symbol `in6addr_any@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status

```

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -280,6 +284,9 @@ VeloxMemoryManager::~VeloxMemoryManager() {
     usleep(waitMs * 1000);
     accumulatedWaitMs += waitMs;
   }
+#ifdef ENABLE_JEMALLOC
+  je_gluten_malloc_stats_print(NULL, NULL, NULL);
```
FelixYBW, line 288: It printed too many logs if we put here. Every task will print a log.

@PHILO-HE  @zhztheplayer Does Gluten has callback of each query? If so let's put the dump to the callback. if not see if we can add one.

PHILO-HE: @zhztheplayer, as mentioned in the meeting, could you provide some details about how CH backend does similarly when query execution is exiting?

zhztheplayer: Hi, please take a loot at this PR  https://github.com/oap-project/gluten/pull/1604 to see if the facility it added could help.

FelixYBW:
@PHILO-HE can the PR still necessary? I remember the issue is fixed in my env but I'm not sure if I ported the PR or not.

PHILO-HE:
> @PHILO-HE can the PR still necessary? I remember the issue is fixed in my env but I'm not sure if I ported the PR or not.

@FelixYBW, yes, it can be used for debugging. I will do a rebase and re-verification before merging it.

PHILO-HE:
Let's merge this version. It's just for debugging purpose.

FelixYBW (approved):

FelixYBW:
For record, to print state we need to use `je_gluten_malloc_stats_print(NULL, NULL, NULL);`

PR #4726, opened 2024-02-20
[GLUTEN-4713][CORE] Not trim the Alias from resultExpressions when determining whether a post-project is needed

liujiayi771:
## What changes were proposed in this pull request?

Fix #4713.

## How was this patch tested?

Exists CI. Test 1T TPCDS q23b.

Issue #4713, opened 2024-02-19
[VL] TPCDS-Q23b failed on large scale


zhouyuan:
### Backend

VL (Velox)

### Bug description

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007feb2f043bcb, pid=1573553, tid=0x00007feb1e9ff700
#
# JRE version: Java(TM) SE Runtime Environment (8.0_161-b12) (build 1.8.0_161-b12)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.161-b12 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libvelox.so+0x4b84bcb]  facebook::velox::exec::GroupingSet::extractGroups(folly::Range<char**>, std::shared_ptr<facebook::velox::RowVector> const&)+0xdb
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /mnt/DP_disk5/tpcds/yarn/usercache/sparkuser/appcache/application_1702277591581_0578/container_1702277591581_0578_01_000016/hs_err_pid1573553.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

### Spark version

None

zhouyuan:
may related with #4628

PR #4753, opened 2024-02-23
[GLUTEN-4713][CORE] Fix invalid children caused by std::move in RowVectorStream

liujiayi771:

In #4713, we discovered that executing q23b on large scale TPCDS dataset would result in a core dump. After investigation, we concluded that an optimization introduced in #4628, which removed some unnecessary post-projects, led to the occurrence of the core dump. However, we believe that this is only a symptom and not the root cause of the problem. In #4726, we restored the post-project and continued to investigate the reason behind the core dump that was triggered by the absence of the post-project.

We ultimately identified the root cause of the issue to be that the `next` method of `RowVectorStream` moves away the children of `RowVector`, leaving the `RowVector`'s children in an indeterminate state. Since the aggregation operation reuses the output `RowVector`, accessing a child that has been moved away during reuse can lead to a core dump.

1. Why doesn't a core dump occur when there is a post-project?
`FilterProject` operator in Velox does not reuse the output `RowVector`, it always creates a new `RowVector` to return.

2. Why doesn't the bug get triggered on small datasets?
The bug is only triggered when there is a reuse scenario occurring with multiple batches being outputted by the aggregation. With small data, the aggregation would only output a single batch, hence no issue arises. However, if the batch size is reduced, this bug can also be triggered.

3. Does the bug occur when the aggregation outputs multiple batches?
The bug is triggered when the downstream of the aggregation is a `RowVectorStream` operator.

After the fix, we can revert the optimization of removing unnecessary post-projects introduced in #4628.

## How was this patch tested?

Testing 1T TPCDS q23b.

### DIFF
--- a/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseHiveTableSuite.scala
+++ b/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickHouseHiveTableSuite.scala
@@ -1146,7 +1146,7 @@ class GlutenClickHouseHiveTableSuite()
         |if(xxx1 in (39,40),6,
         |if(xxx1 in (38,47),xxx1,-1))))))) as string)
         """.stripMargin
-    runQueryAndCompare(querySql)(df => checkOperatorCount[ProjectExecTransformer](3)(df))
+    runQueryAndCompare(querySql)(df => checkOperatorCount[ProjectExecTransformer](2)(df))
   }
 
   test(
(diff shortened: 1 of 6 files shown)


liujiayi771:
cc @zhouyuan, @rui-mo, @ulysses-you, thanks.

rui-mo (commented):
Thanks! Verified on internal TPC-DS test.

ulysses-you (approved):
lgtm, thank you

PR #4757, opened 2024-02-23
[GLUTEN-4748][VL] Fix JNI array elements memory leak

### LINKED ISSUE #4748: [VL] JNI code memory leak when getting array elements
Some JNI code calls `Get<PrimitiveType>ArrayElements` but not calls `Release<PrimitiveType>ArrayElements` after that. This leaks memory while a session is long-running and serving a lot of queries.

Probably related to https://github.com/oap-project/gluten/issues/4144 (requires confirmation)

Ref: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#Get_PrimitiveType_ArrayElements_routines

zhztheplayer:
This fixes #4748

Introduce a set of `gluten::get<primitive-type>ArrayElementsSafe` functions that would release the gotten array elements while the returned object is destructed. So user would not need to manually release the array elements which can be easily forgotten during programming.

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -336,7 +336,8 @@ JNIEXPORT jstring JNICALL Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapp
     jboolean details) {
   JNI_METHOD_START
 
-  auto planData = reinterpret_cast<const uint8_t*>(env->GetByteArrayElements(planArray, 0));
+  auto safeArray = gluten::getByteArrayElementsSafe(env, planArray);
+  auto planData = reinterpret_cast<const uint8_t*>(safeArray.elems());
   auto planSize = env->GetArrayLength(planArray);
   auto ctx = gluten::getRuntime(env, wrapper);
   ctx->parsePlan(planData, planSize, {}, std::nullopt);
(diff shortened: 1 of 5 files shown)


zhztheplayer:
@FelixYBW @zhouyuan

cpp/core/jni/JniWrapper.cc
```diff
@@ -354,11 +356,10 @@ JNIEXPORT void JNICALL Java_io_glutenproject_vectorized_PlanEvaluatorJniWrapper_
   JNI_METHOD_START
 
   auto len = env->GetArrayLength(path);
-  jbyte* bytes = env->GetByteArrayElements(path, 0);
-  std::string pathStr(reinterpret_cast<char*>(bytes), len);
+  auto safeArray = gluten::getByteArrayElementsSafe(env, path);
+  std::string pathStr(reinterpret_cast<char*>(safeArray.elems()), len);
   auto ctx = gluten::getRuntime(env, wrapper);
   ctx->injectWriteFilesTempPath(pathStr);
-  env->ReleaseByteArrayElements(path, bytes, JNI_ABORT);
```
zhouyuan, line 361: does this mean the `env->ReleaseByteArrayElements` not work as expected?

zhouyuan (approved):
👍

cpp/core/jni/JniWrapper.cc
zhztheplayer: It should work but some of our calls to `Get` doesn't have a corresponding call to `Release`. So this code was a good case not leading to memory leak.

zhztheplayer: As a comparison the code in `nativeCreateKernelWithIterator` may leak considerable amount of memory since it's called frequently and missed several `Release` calls.

PR #4764, opened 2024-02-23
[GLUTEN-4763][VL] Add RewriteTypedImperativeAggregate rule for collect_list

liujiayi771:

For some typed imperative aggregate functions, like `collect_list`/`collect_set`, Spark will make agg buffer stored as `BinaryType` in partial/ partial merge phase, even though these two functions' output is` ArrayType`. Gluten has no such special handling. For Velox backend, agg function's raw data type is used for agg buffer in partial/partial merge. Thus, there will be a type mismatch issue in the latter shuffle operator, as Gluten plan inherits attributes from its corresponding Spark plan. For `collect_list`/`collect_set`, shuffle expects `BinaryType`, but gets `ArrayType`.

So, we need a rule `RewriteTypedImperativeAggregate` to rewrite `collect_list` and `collect_set` for Velox backend. Referenced the implementation in https://github.com/oap-project/gluten/pull/2669.

Velox does not register the companion functions for `collect_set`, to enable `collect_set`, modifications need to be made in Velox.

Currently, there are still some issues, such as the semantics of null, where Spark and Velox have many differences. At present, `collect_list` supports ignoring null values, but when all inputs are null, Spark returns an empty array, while Velox returns null. `collect_set` does not yet support ignoring null values, and there is the same problem when all inputs are null. Adaptations on the Velox side are still needed.

## How was this patch tested?

* Exists CI.
* Add `collect_list` in `VeloxAggregateFunctionsSuite.distinct functions`
* Add a new test case `VeloxAggregateFunctionsSuite.collect_list null inputs`

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/VeloxBackend.scala
@@ -443,4 +443,10 @@ object BackendSettings extends BackendSettingsApi {
   override def supportCartesianProductExec(): Boolean = true
 
   override def supportBroadcastNestedLoopJoinExec(): Boolean = true
+
+  override def shouldRewriteTypedImperativeAggregate(): Boolean = {
+    // The intermediate type of collect_list, collect_set in Velox backend is not consistent with
+    // vanilla Spark, we need to rewrite the aggregate to get the correct data type.
+    true
+  }
 }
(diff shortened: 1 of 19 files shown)


liujiayi771:
cc @PHILO-HE, thanks.

PHILO-HE (approved):
Looks good to me. Thanks so much for your work!
I note collect_list/set were excluded for window operator. Maybe, they can be removed now. It can be done in a separate pr. Thanks!
https://github.com/oap-project/gluten/blob/bda60be4c44621c3722ecfbe55e98fd22f06b04a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc#L645

PR #4777, opened 2024-02-26
[VL] Fix Iceberg getPartitionValueString

liujiayi771:
The original code had a typo when retrieving the partition value of the iceberg table. The reason why the tpch partitioned table tests did not catch this is mainly because the tpch table's partition column type is primarily date, which did not trigger the `case _` code.

## How was this patch tested?

Add new test case for partition key type of string and integer.

### DIFF
--- a/.github/workflows/velox_be.yml
+++ b/.github/workflows/velox_be.yml
@@ -25,6 +25,7 @@ on:
       - 'gluten-core/**'
       - 'gluten-data/**'
       - 'gluten-delta/**'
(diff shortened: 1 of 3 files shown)


liujiayi771:
cc @yma11 @rui-mo , thanks.

SinghAsDev:
Thanks for the change @liujiayi771 , I verified that this fixes partitioned iceberg table scans.

SinghAsDev (approved):

yma11 (approved):

PR #4779, opened 2024-02-26
[GLUTEN-3378][VL] Feat: Support read iceberg mor table for Velox backend

liujiayi771:

Velox add iceberg mor table read support in https://github.com/facebookincubator/velox/pull/7847. This PR supports read iceberg mor table for Velox backend.

## How was this patch tested?

Add mor table read test case "iceberg read mor table".

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -158,9 +158,7 @@ WholeStageResultIterator::WholeStageResultIterator(
       if (auto icebergSplitInfo = std::dynamic_pointer_cast<IcebergSplitInfo>(scanInfo)) {
         // Set Iceberg split.
         std::unordered_map<std::string, std::string> customSplitInfo{{"table_format", "hive-iceberg"}};
-        auto deleteFilesFind = icebergSplitInfo->deleteFilesMap.find(paths[idx]);
-        auto deleteFiles = deleteFilesFind != icebergSplitInfo->deleteFilesMap.end() ? deleteFilesFind->second
-                                                                                     : std::vector<IcebergDeleteFile>{};
+        auto deleteFiles = icebergSplitInfo->deleteFilesVec[idx];
         split = std::make_shared<velox::connector::hive::iceberg::HiveIcebergSplit>(
             kHiveConnectorId,
             paths[idx],
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -154,19 +154,39 @@ WholeStageResultIterator::WholeStageResultIterator(
       auto metadataColumn = metadataColumns[idx];
       std::unordered_map<std::string, std::optional<std::string>> partitionKeys;
       constructPartitionColumns(partitionKeys, partitionColumn);
-      auto split = std::make_shared<velox::connector::hive::HiveConnectorSplit>(
-          kHiveConnectorId,
-          paths[idx],
-          format,
-          starts[idx],
-          lengths[idx],
-          partitionKeys,
-          std::nullopt,
-          std::unordered_map<std::string, std::string>(),
-          nullptr,
-          std::unordered_map<std::string, std::string>(),
-          0,
-          metadataColumn);
+      std::shared_ptr<velox::connector::ConnectorSplit> split;
+      if (auto icebergSplitInfo = std::dynamic_pointer_cast<IcebergSplitInfo>(scanInfo)) {
+        // Set Iceberg split.
+        std::unordered_map<std::string, std::string> customSplitInfo{{"table_format", "hive-iceberg"}};
+        auto deleteFilesFind = icebergSplitInfo->deleteFilesMap.find(paths[idx]);
+        auto deleteFiles = deleteFilesFind != icebergSplitInfo->deleteFilesMap.end() ? deleteFilesFind->second
+                                                                                     : std::vector<IcebergDeleteFile>{};
+        split = std::make_shared<velox::connector::hive::iceberg::HiveIcebergSplit>(
+            kHiveConnectorId,
+            paths[idx],
+            format,
+            starts[idx],
+            lengths[idx],
+            partitionKeys,
+            std::nullopt,
+            customSplitInfo,
+            nullptr,
+            deleteFiles);
+      } else {
+        split = std::make_shared<velox::connector::hive::HiveConnectorSplit>(
+            kHiveConnectorId,
+            paths[idx],
+            format,
+            starts[idx],
+            lengths[idx],
+            partitionKeys,
+            std::nullopt,
+            std::unordered_map<std::string, std::string>(),
+            nullptr,
+            std::unordered_map<std::string, std::string>(),
+            0,
+            metadataColumn);
+      }
       connectorSplits.emplace_back(split);
     }
 
--- a/cpp/velox/compute/iceberg/IcebergPlanConverter.cc
+++ b/cpp/velox/compute/iceberg/IcebergPlanConverter.cc
@@ -72,7 +72,10 @@ std::shared_ptr<IcebergSplitInfo> IcebergPlanConverter::parseIcebergSplitInfo(
       deletes.emplace_back(IcebergDeleteFile(
           fileContent, deleteFile.filepath(), format, deleteFile.recordcount(), deleteFile.filesize()));
     }
-    icebergSplitInfo->deleteFilesMap[file.uri_file()] = std::move(deletes);
+    icebergSplitInfo->deleteFilesVec.emplace_back(deletes);
+  } else {
+    // Add an empty delete files vector to indicate that this data file has no delete file.
+    icebergSplitInfo->deleteFilesVec.emplace_back(std::vector<IcebergDeleteFile>{});
   }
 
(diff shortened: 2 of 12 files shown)


liujiayi771:
cc @yma11 @YannByron, thanks.

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
```diff
@@ -57,7 +35,79 @@ class DeleteFile {
...
+      deleteFileBuilder.setFileSize(delete.fileSizeInBytes());
+      deleteFileBuilder.setRecordCount(delete.recordCount());
+      switch (delete.format()) {
+        case PARQUET:
+          ReadRel.LocalFiles.FileOrFiles.ParquetReadOptions parquetReadOptions =
+              ReadRel.LocalFiles.FileOrFiles.ParquetReadOptions.newBuilder()
+                  .setEnableRowGroupMaxminIndex(
+                      GlutenConfig.getConf().enableParquetRowGroupMaxMinIndex())
+                  .build();
+          deleteFileBuilder.setParquet(parquetReadOptions);
```
yma11, line 98: Should the deletion files share same read options as the data file?

gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxIcebergSuite.scala
```diff
@@ -42,15 +43,54 @@ class VeloxIcebergSuite extends WholeStageTransformerSuite {
...
+                    |  'write.merge.mode' = 'merge-on-read'
+                    |)
+                    |partitioned by (p);
+                    |""".stripMargin)
+        // Insert some test rows.
+        spark.sql("""
+                    |insert into table iceberg_mor_tb
+                    |values (1, 'a1', 'p1'), (2, 'a2', 'p1'), (3, 'a3', 'p2');
+                    |""".stripMargin)
+        // Delete row.
```
yma11, line 82: add cases for multi deletions?

gluten-iceberg/src/main/java/io/glutenproject/substrait/rel/IcebergLocalFilesNode.java
liujiayi771: Iceberg allows the format of delete file and data file to be different, but in most cases they are consistent.

cpp/velox/compute/iceberg/IcebergPlanConverter.cc
```diff
@@ -0,0 +1,81 @@
...
+
+std::shared_ptr<IcebergSplitInfo> IcebergPlanConverter::parseIcebergSplitInfo(
+    substrait::ReadRel_LocalFiles_FileOrFiles file,
+    std::shared_ptr<SplitInfo> splitInfo) {
+  using SubstraitFileFormatCase = ::substrait::ReadRel_LocalFiles_FileOrFiles::IcebergReadOptions::FileFormatCase;
+  using SubstraitDeleteFileFormatCase =
+      ::substrait::ReadRel_LocalFiles_FileOrFiles::IcebergReadOptions::DeleteFile::FileFormatCase;
+  auto icebergSplitInfo = std::dynamic_pointer_cast<IcebergSplitInfo>(splitInfo)
+      ? std::dynamic_pointer_cast<IcebergSplitInfo>(splitInfo)
+      : std::make_shared<IcebergSplitInfo>(*splitInfo);
```
yma11, line 30: When will it happen that it's not a `IcebergSplitInfo`?

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -152,8 +152,28 @@ WholeStageResultIterator::WholeStageResultIterator(
...
       std::unordered_map<std::string, std::optional<std::string>> partitionKeys;
       constructPartitionColumns(partitionKeys, partitionColumn);
-      auto split = std::make_shared<velox::connector::hive::HiveConnectorSplit>(
-          kHiveConnectorId, paths[idx], format, starts[idx], lengths[idx], partitionKeys);
+      std::shared_ptr<velox::connector::ConnectorSplit> split;
+      if (auto icebergSplitInfo = std::dynamic_pointer_cast<IcebergSplitInfo>(scanInfo)) {
+        // Set Iceberg split.
+        std::unordered_map<std::string, std::string> customSplitInfo{{"table_format", "hive-iceberg"}};
+        auto deleteFilesFind = icebergSplitInfo->deleteFilesMap.find(paths[idx]);
+        auto deleteFiles = deleteFilesFind != icebergSplitInfo->deleteFilesMap.end() ? deleteFilesFind->second
```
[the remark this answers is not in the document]
liujiayi771: For normal `SplitInfo`, file information is stored in lists, it is unable to obtain the mapping relationship between the data file and the delete file. I have actually considered placing the delete file map under protobuf's `LocalFiles`. The current processing logic is to extract the map information into the `IcebergReadOptions` of each `FileOrFiles` in Java, and then combine them back into a map in C++. It is actually somewhat redundant, but this approach has the least impact on the current protobuf changes. 
We can also add a new definition for `table_format` in `LocalFiles`, with the default being `hive`. It can also be `hive-iceberg`. When the format is `hive-iceberg`, the delete file map contained within will be read.

cpp/velox/compute/iceberg/IcebergPlanConverter.cc
liujiayi771: Since `substrait::ReadRel_LocalFiles_FileOrFiles` contains multiple files, each file will enter this function during iteration. The first time it enters as `SplitInfo`, and subsequently, it is replaced by `IcebergSplitInfo`.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -149,6 +149,28 @@ message ReadRel {
...
+      message IcebergReadOptions {
+        enum FileContent {
+          DATA = 0;
+          POSITION_DELETES = 1;
+          EQUALITY_DELETES = 2;
+        }
+        message DeleteFile {
+          FileContent fileContent = 1;
+          string filePath = 2;
+          uint64 fileSize = 5;
```
YannByron, line 161: maybe a stupid question: why it skips 3 and 4?

gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxIcebergSuite.scala
YannByron: +1. At least, should cover `UPDATE` operation. `MergeInto` is nice to have.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
liujiayi771: I removed some redundant fields and forgot to update the sequence numbers; it shouldn't skip 3 and 4.

gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxIcebergSuite.scala
liujiayi771: @yma11 @YannByron While adding test cases, I discovered a bug in the Velox code. I will fix this bug first before updating the current PR. Later on, I will add test cases for MERGE INTO and UPDATE, as well as multiple DELETE operations.

cpp/velox/compute/WholeStageResultIterator.cc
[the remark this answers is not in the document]
yma11: In current code, we put all the fields like `paths`, `starts`, `lengths`, etc of all files for each task together, with each as a list. I think we can refactor it by using a single list of `file` or `split` which contains `path`, `start`, `length`,  a list of `deleteFiles` so that we won't need this map. But how will it affect the protobuf part? any idea?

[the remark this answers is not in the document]
liujiayi771: @yma11 This is also a feasible solution, but not all data files have a delete file, and some data files may have multiple delete files. We would need a two-dimensional vector to maintain this relationship, and the two-dimensional vector would need to have empty vector for data files that do not have a corresponding delete file. Do you think this is a better solution? I'm OK to change to using this approach.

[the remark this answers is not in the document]
yma11: Yeah. `scanInfo` will have a list of `SplitInfo` and each `SplitInfo` contains its own `path`, `start`, `length` as well as a list of `deleteFiles` which may be empty. I think it will be clear in conception and without using `idx` anywhere. `Spark-delta` has a similar structure called `AddFile` which organizes like this. It's okay to merge this PR first and we do it as a follow up.

yma11 (approved):

yma11:
@liujiayi771 Seems code has scala style violations. Please update.

yma11 (approved):

zhouyuan:
@liujiayi771 There's a small conflict, could you please help to do a rebase? 

thanks,
-yuan

liujiayi771:
> @liujiayi771 There's a small conflict, could you please help to do a rebase?
> 
> thanks, -yuan

Done.

zhouyuan (approved):
👍

PR #4783, opened 2024-02-26
[VL] Fix and use flattenVector

marin-ma:
This patch re-enables the flattern vector optimizations.

The flattenVector optimization firstly landed in https://github.com/oap-project/gluten/pull/4415
but partially reverted in https://github.com/oap-project/gluten/pull/4474 due to some bugs on Celeborn code path.

Celeborn integration tests should check the code.

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -222,6 +222,7 @@ arrow::Status VeloxShuffleWriter::init() {
 
   ARROW_ASSIGN_OR_RAISE(
       partitioner_, Partitioner::make(options_.partitioning, numPartitions_, options_.startPartitionId));
+  DLOG(INFO) << "Create partitioning type: " << std::to_string(options_.partitioning);
 
   // pre-allocated buffer size for each partition, unit is row count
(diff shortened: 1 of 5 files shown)


cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -222,6 +222,7 @@ arrow::Status VeloxShuffleWriter::init() {
 
   ARROW_ASSIGN_OR_RAISE(
       partitioner_, Partitioner::make(options_.partitioning, numPartitions_, options_.startPartitionId));
+  DLOG(INFO) << "Create partitioning type: " << std::to_string(options_.partitioning);
```
zhouyuan, line 225: this looks like a debug log?

zhouyuan (approved):
👍

cpp/velox/shuffle/VeloxShuffleWriter.cc
marin-ma: Yes. The `DLOG` ensures that it only get printed with debug mode. We probably need this log for debugging, because sometimes the shuffle operator can get omitted on Spark UI, such as a single partitioning after limit operator.

PR #4784, opened 2024-02-26
[VL] Add write path option for generic benchmark

Yohahaha:


marin-ma:
Thanks! Could you also update the document?

Yohahaha:
> Thanks! Could you also update the document? It looks like user must specify this option if simulating the write stage.

updated, default write path is `/tmp`.

marin-ma (approved):

PR #4788, opened 2024-02-27
[GLUTEN-4789] [VL] Wrong preceding/following convert in window function node

WangGuangxin:

The convert logic in `WindowFuncitonNode.java`  make the assumption that the lower bound of a row/range frame are always `preceding offset`, and the upper bound are always `following offset`,  which is not the case.

https://github.com/oap-project/gluten/blob/6acd1b367a669bcca8af3708c00328cc3e707d2a/gluten-core/src/main/java/io/glutenproject/substrait/expression/WindowFunctionNode.java#L80

For example, following case is also a valid row frame.
```
select max(l_suppkey) over  (partition by l_suppkey order by l_orderkey rows between 2 preceding and 1 preceding) from lineitem
```

In fact, the spark row frame will convert preceding offset A to -A, so that the lower and upper bound params in `SpecifiedWindowFrame` are always **following offset**. 

Since Velox requires the input offset must be positive, so we just need to convert negative following offset to preceding offset, and vice versa.

## How was this patch tested?
UT

### DIFF
--- a/backends-velox/src/test/scala/io/glutenproject/execution/VeloxWindowExpressionSuite.scala
+++ b/backends-velox/src/test/scala/io/glutenproject/execution/VeloxWindowExpressionSuite.scala
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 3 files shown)


WangGuangxin:
cc @lgbo-ustc @PHILO-HE @rui-mo @ulysses-you

ulysses-you:
Run Gluten Clickhouse CI

ulysses-you:
lgtm if ch backend tests pass

ulysses-you (approved):

PHILO-HE (approved):
Thanks!

PR #4790, opened 2024-02-27
[CORE] Move all columnar rules to post-columnar transitions

zhztheplayer:
This patch moves all columnar rules to post-columnar transition API. This works like we'll inject Gluten's custom rules to Spark's preparation rule list rather than in to columnar rules.

By doing this, we can treat all of Gluten's columnar rules as an individual Spark preparation rule that works on a complete Spark plan with C2Rs and R2Cs added . So we get full control of plan conversion rather than being clamped by Spark's rule `ApplyColumnarRulesAndInsertTransitions`. This will help on continuously improving Gluten's plan optimization capability and to make it not only limited to doing columnar conversions.


gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -520,39 +524,27 @@ private[extension] object ColumnarToRowLike {
...
-            child
-          }
+        case child if PlanUtil.isGlutenColumnarOp(child) =>
+          RowToColumnarExec(
+            BackendsApiManager.getSparkPlanExecApiInstance.genColumnarToRowExec(child))
+        case other => other
       })
-    case _ =>
-      plan.withNewChildren(plan.children.map(replaceWithVanillaRowToColumnar))
   }
```
zhztheplayer, line 548: cc @ulysses-you

zhztheplayer:
@rui-mo @PHILO-HE

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -520,39 +524,27 @@ private[extension] object ColumnarToRowLike {
 }
 // This rule will try to add RowToColumnarExecBase and ColumnarToRowExec
 // to support vanilla columnar operators.
-case class VanillaColumnarPlanOverrides(session: SparkSession) extends Rule[SparkPlan] {
+case class InsertColumnarToColumnarTransitions(session: SparkSession) extends Rule[SparkPlan] {
```
ulysses-you, line 527: It seems the previous rule name is more readable. It is used to be compatible with vanilla columnar related things.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -684,29 +754,54 @@ case class ColumnarOverrideRules(session: SparkSession)
...
+   *
+   * See: https://github.com/oap-project/gluten/pull/4790
+   */
+  final override def preColumnarTransitions: Rule[SparkPlan] = plan => {
+    // To infer caller's property: ApplyColumnarRulesAndInsertTransitions#outputsColumnar.
+    OutputsColumnarTester.wrap(plan)
+  }
+
+  override def postColumnarTransitions: Rule[SparkPlan] = plan => {
+    val outputsColumnar = OutputsColumnarTester.inferOutputsColumnar(plan)
```
ulysses-you, line 780: why not use `plan.supportsColumnar` ?

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -684,29 +754,54 @@ case class ColumnarOverrideRules(session: SparkSession)
...
+    val unwrapped = OutputsColumnarTester.unwrap(plan)
+    val vanillaPlan = ColumnarTransitions.insertTransitions(unwrapped, outputsColumnar)
+    withSuggestRules(suggestRules(outputsColumnar)).apply(vanillaPlan)
+  }
+
+  // Visible for testing.
+  def withSuggestRules(suggestRules: List[SparkSession => Rule[SparkPlan]]): Rule[SparkPlan] =
```
ulysses-you, line 787: How about naming `transformRules` ? The `suggest rules` is a bit hard to follow..

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: Thanks for the explanation. 

Can we temporarily make the name describe what it actually does? If it's now converts from columnar to columnar, I will be inclined to use this new name.

I understand it could do more than columnar transitions but we can discuss on whether to add new logics to this rule or into other independent rules at that time.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: Spark has `supportsRowBased` since 3.3 . A plan node can be both `supportsColumnar == true` and `supportsRowBased == true`.  We should know about the caller's intention having which property requested.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: I'll be good to both. I'd change to `withTransformRules` if it's confusing.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
ulysses-you: If the plan support both columnar and row, then Spark should do thing. Does `plan.supportsColumnar` return different value with `OutputsColumnarTester.inferOutputsColumnar` ? It seems we never touch supportsRowBased.

zhztheplayer: It's a rare case but it does covered by existing UTs...

Consider an `UnionExec` which supports both columnar and row-based execution, then even caller requested `outputsColumnar=false` through `ApplyColumnarRulesAndInsertTransitions#outputsColumnar`, we don't have a way to be aware of that since the `UnionExec` returns `true` for its `supportsColumnar` method.

So we should infer `outputsColumnar`'s exact value here rather than just calling `supportsColumnar`.

ulysses-you: Can we move OutputsColumnarTester related code to a new file with the comment ? The ColumnarOverrides file is too big.

JkSelf (approved):
Great work. LGTM. Thanks.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: I was planning on splitting the file too. Let's do that in another patch. Thanks for the suggestion.

Issue #4808, opened 2024-02-29
[CH] Support sort merge join


zhanglistar:
As title.

loudongfeng:
I'm working on #4745 . And is almost ready to merge. I see your version is still in draft state.Would you mind review #4745 instead?

zhanglistar:
@loudongfeng oK.

PR #4809, opened 2024-02-29
[CH] add more join type support

zhanglistar:

In d_2046_0, partial merge join is better than hash join, so we need support more join implementation.

(Fixes: \#4808)

## How was this patch tested?

CI


zhanglistar:
On SQL d_2046_0, partial_merge is better than hash join. So we need support more join implementation.

Issue #4241, opened 2024-01-02
[VL] Add plan node to convert Vanilla spark columnar format data to Velox columnar format data


zhli1142015:
### Description

Add plan node to convert Vanilla spark columnar format data to Velox columnar format data
C2C to replace `RowToVeloxColumnarExec(ColumnarToRowExec())` maybe also `RowToColumnarExec(VeloxColumnarToRowExec())`

zhztheplayer:
Closed in https://github.com/oap-project/gluten/commit/dd8e872973eb11ccb1ddb73a373d6bd4508f25ae

boneanxs:
Hey @zhli1142015 , are you working on this? If not, I'm willing to driver this.

zhli1142015:
I'm not working on it, please go ahead. Thanks.

boneanxs:
Hey @acvictor, sorry to update here, I'm already work on this(doing some tests now), and the pr will be ready soon(should be finished next week)

PR #4818, opened 2024-02-29
[GLUTEN-4241][VL] Add plan node to convert Vanilla spark columnar format data to Velox columnar format data

boneanxs:

Add plan node to convert Vanilla spark columnar format data to Velox columnar format data

This pr tries to convert from Spark columnar batch -> arrow Array -> velox columnar batch, all primitive types(decimal types requires: https://github.com/facebookincubator/velox/pull/8957) and Map/Array types are supported.

Users can enable it by setting `spark.gluten.sql.columnar.vanillaColumnarToNativeColumnar`, which is by default false.

(Fixes: \#4241)

## How was this patch tested?

Added tests


cpp/velox/tests/VeloxColumnarToColumnarTest.cc
```diff
@@ -0,0 +1,100 @@
...
+}
+
+TEST_F(VeloxColumnarToColumnarTest, decimal) {
+  auto vector = makeRowVector({
+      makeNullableFlatVector<int64_t>({1.00, 2.00, 3.00}, DECIMAL(5, 2))
+  });
+
+  ArrowSchema cSchema;
+  ArrowArray arrowArray;
+  exportToArrow(vector, cSchema, ArrowUtils::getBridgeOptions());
```
boneanxs, line 96: If I'm understanding well, there might be a bug that velox doesn't handle short decimal well when converting `arrowArray`-> `rowVector`

In `exportToArrow`, we can see it needs to convert to `int128_t` since it's arrow short decimal.

```c++
rows.apply([&](vector_size_t i) {
      int128_t value = buf.as<int64_t>()[i];
      memcpy(dst + (j++) * sizeof(int128_t), &value, sizeof(int128_t));
    });
```

But in `importFromArrow`, velox directly handle it as `int64_t`, whereas this test won't pass.

```c++
values = wrapInBufferView(
        arrowArray.buffers[1], arrowArray.length * type->cppSizeInBytes());
```

cc @PHILO-HE @zhouyuan pls correct me if I'm wrong

zhouyuan: +1 looks a bug

zhztheplayer (commented):
Thanks! I didn't take look in the PR in detail but overall the direction looks promising.

Also, would you like to add a little bit more explanation in PR description ? Like it supports all datatypes or just some of them, it will be turned on or off by default, what's the config key to turn it on/off, etc. That will help further users getting started with the feature quickly.

backends-velox/src/main/scala/io/glutenproject/execution/ColumnarToVeloxColumnarExec.scala
```diff
@@ -0,0 +1,158 @@
...
+
+case class ColumnarToVeloxColumnarExec(child: SparkPlan) extends ColumnarToColumnarExecBase(child) {
+
+  override def doExecuteColumnarInternal(): RDD[ColumnarBatch] = {
+    new ValidatorApiImpl().doSchemaValidate(schema).foreach {
+      reason =>
+        throw new UnsupportedOperationException(
+          s"Input schema contains unsupported type when performing columnar" +
+            s" to columnar for $schema " + s"due to $reason")
+    }
```
zhztheplayer, line 47: It looks a little bit weird to do validation during execution. Would you like to share the reason doing this?

cpp/velox/tests/VeloxColumnarToColumnarTest.cc
boneanxs: I see, let me try to fix velox firstly

backends-velox/src/main/scala/io/glutenproject/execution/ColumnarToVeloxColumnarExec.scala
boneanxs: Yea, I follow `RowToVeloxColumnarExec` to do the validation here, you're right, it looks weird to do so during execution, I also have the same doubt when seeing `RowToVeloxColumnarExec`...

Let me fix it by overriding `doValidateInternal` instead.

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -547,8 +547,28 @@ case class InsertColumnarToColumnarTransitions(session: SparkSession) extends Ru
...
+    val newPlan =
+      if (
+        GlutenConfig.getConf.enableNativeColumnarToColumnar && BackendsApiManager.getSettings
+          .supportColumnarToColumnarExec()
+      ) {
+        replaceWithColumnarToColumnar(
+          replaceWithVanillaRowToColumnar(replaceWithVanillaColumnarToRow(plan)))
+      } else {
+        replaceWithVanillaRowToColumnar(replaceWithVanillaColumnarToRow(plan))
+      }
```
zhztheplayer, line 571: Could we simplify the code to something like
```scala
val newPlan = replaceWithVanillaRowToColumnar(replaceWithVanillaColumnarToRow(plan))
if (...) {
newPlan = replaceWithColumnarToColumnar(newPlan)
}
```
?

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
```diff
@@ -547,8 +547,28 @@ case class InsertColumnarToColumnarTransitions(session: SparkSession) extends Ru
       })
   }
 
+  private def replaceWithColumnarToColumnar(plan: SparkPlan): SparkPlan = {
+    plan match {
+      case p: RowToColumnarExecBase if p.child.isInstanceOf[ColumnarToRowExec] =>
+        val replacedChild = replaceWithColumnarToColumnar(
+          p.child.asInstanceOf[ColumnarToRowExec].child)
+        BackendsApiManager.getSparkPlanExecApiInstance.genColumnarToColumnarExec(replacedChild)
```
zhztheplayer, line 555: The case condition says it matches `vanilla-columnar->row->gluten-columnar` but it calls a general interface to do `columnar->columnar`. Do we need to specialize the interface name to make sure we exactly request for `vanilla-columnar->gluten-columnar` transition here?

FelixYBW:
Do you convert to Velox format directly? or convert to Arrow then to Velox?

boneanxs:
> Do you convert to Velox format directly? or convert to Arrow then to Velox?

Convert to Arrow firstly, then to velox

FelixYBW:
> > Do you convert to Velox format directly? or convert to Arrow then to Velox?
> 
> Convert to Arrow firstly, then to velox

make sense. We may upstream the parquet columnar format => arrow format to Spark

cpp/velox/tests/VeloxColumnarToColumnarTest.cc
boneanxs: fix it here: https://github.com/facebookincubator/velox/pull/8957

gluten-core/src/main/scala/io/glutenproject/extension/ColumnarOverrides.scala
zhztheplayer: Thanks for the explanation.

If it's for ColumnarToRow + RowToVeloxColumnar,

Would it be better to rename the method `genColumnarToColumnarExec` with `genVanillaColumnarToGlutenColumnarExec` or something?

zhztheplayer: So it's clearer that we are not doing `Gluten` -> `Vanilla`  c2c transition here.

backends-velox/src/main/scala/io/glutenproject/execution/VanillaColumnarToVeloxColumnarExec.scala
```diff
@@ -0,0 +1,162 @@
...
+        }
+      }
+
+      override def hasNext: Boolean = {
+        if (arrowArray != null) {
+          arrowArray.release()
+          arrowArray.close()
+          arrowConverter.reset()
+          arrowArray = null
+        }
```
zhztheplayer, line 128: Seems to be that the file includes a bunch of manual life cycle management calls like `_.close` `_.release` `_.reset`, etc. Would you like to review them to see if we can eliminate some? I recall that `recycleIterator` / `recyclePayloads` can do some of the clean up jobs automatically.

It's a non-blocking comment so feel free to do that in another PR. But anyway we'd try best to make code here easier to maintain from further developers.

zhztheplayer (approved):
Overall it's good to me. Just a non-blocking comment. Thanks!

backends-velox/src/main/scala/io/glutenproject/execution/VanillaColumnarToVeloxColumnarExec.scala
boneanxs: We need to reset arrowArray and arrowConverter when iterate each item, so `recyclePayload` is reasonable, but it only passes the result of the `next()`, which is `ColumnarBatch`, we cannot pass arrowArray to close them.

So here instead I add an extra method `releaseArrowBuf` to release them to reduce duplicates.

FelixYBW:
@zhztheplayer can you check how the memory is allocated during the conversion? Where the arrow memory is allocated? how many memcpy during the conversion? Is there onheap=>offheap copy?

zhztheplayer:
> @zhztheplayer can you check how the memory is allocated during the conversion? Where the arrow memory is allocated? how many memcpy during the conversion? Is there onheap=>offheap copy?

@boneanxs If you'd like to address the questions also, thanks.

I believe the patch reused our old `ArrowWritableColumnarVector` code to write Spark columnar data to native so there should be a bunch of "onheap => offheap" copies. And we should count on how much of copies the implementation exactly does ideally. @boneanxs You can also check on this part.

What I was worried about is `ArrowWritableColumnarVector` have not actually been under active maintenance for a period of time so we should have more tests here especially for complex data types.

Also would be great if you could share thoughts about the risk of memory leaks this approach may bring. Overall the PR's writing looks find to me and we had removed most of the unsafe APIs but still there might be some. Let's check this part carefully too.

FelixYBW:
Let's document the conversion clearly here. I have a impression that parquet-mr can take use of offheap memory for columnar data. If so the best case is that we can avoid any memcpy during the conversion (not considered about arrow=>velox conversion). But we will needs some work to avoid it.

If parquet scan is onheap, we can't avoid the onheap->offheap copy, but then we should reuse the offheap buffer. 

If we can't make sure how the conversion happens, let's use memcpy as long we need, make sure it's 100% no memleak for now.

boneanxs:
> @zhztheplayer can you check how the memory is allocated during the conversion? Where the arrow memory is allocated? how many memcpy during the conversion? Is there onheap=>offheap copy?

Hey, @FelixYBW for each columnar batch, before the conversion, this pr tries to allocate offheap memory to perform `spark columnar batch` -> `arrow array`(here this pr doesn't treat onheap/offheap spark columnar batch, it uses `ArrowFieldWriter`(which is implemented by spark) to do this transformation, can see `ArrowColumnarBatchConverter#write`, which is this pr newly added).

So yes, there will be one memcpy from `spark columnar batch` -> `arrow array`, no matter `spark columnar batch` is on heap or off heap.

In native side, I simply uses `importFromArrow` to convert `arrow array` to `velox columnar batch`, there will be some memcpy either for String, timestamp, shortDecimal, etc(I'm not fully checked).

> What I was worried about is ArrowWritableColumnarVector have not actually been under active maintenance for a period of time so we should have more tests here especially for complex data types.

Sry @zhztheplayer, I might miss something, are you saying `ArrowFieldWriter`? This pr uses `ArrowFieldWriter` to do the conversion, it's wildly used by pyspark, so reusing it here should be safe.

> Also would be great if you could share thoughts about the risk of memory leaks this approach may bring @boneanxs .
The extra memory here acquired are `arrowArray`, `cSchema`, `velox columnar batch`, and they're all well handled by `TaskResources` and `recyclePayload`, `arrowArray` and `velox columnar batch` will be released during each iterator, and `cSchema` will be released until the iterator ends, `TaskResources` is the extra ensureance that all allocated memory will be released if the iterator interrupted abnormally.

> I have a impression that parquet-mr can take use of offheap memory for columnar data. If so the best case is that we can avoid any memcpy during the conversion
I'm not sure whether parquet-mr can directly use offheap memory, but spark can export parquet data as offheap columnar batch by enabling `spark.sql.columnVector.offheap.enabled`, but directly reuse offheap columnar batch to arrow array is not supported in this pr(by the way, can we do so? Not sure abt this)

I can also try to do benchmarks comparing with VanillarColumnarBatchToRow->RowToVeloxColumnar if necessary, I think at least for memcpy, `RowToVeloxColumnar` also does an extra copy even if the row might from an offheap ColumnarBatch

zhztheplayer:
> This pr uses ArrowFieldWriter to do the conversion, it's wildly used by pyspark, so reusing it here should be safe.

Ahh then it's fine enough. Some of the code is so similar to Gluten's `ArrowWritableColumnarVector` which had pasted some code from vanilla Spark so I was led by wrong intuition. Sorry for the mistake.

FelixYBW:
> does an extra copy even if the row might from an off

Thank you for explanation. You may try to enable spark.sql.columnVector.offheap.enabled.

onheap to offheap memcpy is more expensive than offheap to offheap. Do you know how ArrowFieldWriter allocate memory? is it from unsafe API or direct memory? 

@zhztheplayer Do you know?

boneanxs:
Hey @FelixYBW `ArrowFieldWriter` calls arrow `ValueVector`, which is internally uses `ArrowBuf` to store values, so it should be offheap memory.
https://github.com/apache/spark/blob/1f58f4c68e8de03a8b4c314488dd4f342beb8de2/sql/catalyst/src/main/scala/org/apache/spark/sql/execution/arrow/ArrowWriter.scala#L53

for example, IntVector
https://github.com/apache/arrow/blob/b202ede131e3c54628616330162f7854ba0c0d70/java/vector/src/main/java/org/apache/arrow/vector/IntVector.java#L151

ValueBuffer is allocated from arrow `ArrowBuf`
https://github.com/apache/arrow/blob/b202ede131e3c54628616330162f7854ba0c0d70/java/vector/src/main/java/org/apache/arrow/vector/BaseFixedWidthVector.java#L73

zhztheplayer:
I think it's OK to have it disabled by default.

@boneanxs Can you add a CI case for the feature to run TPC-H / TPC-DS tests?

Example:

https://github.com/apache/incubator-gluten/blob/8ab9b012db7ebbd7110ba1288b5c4e8a702ccc09/.github/workflows/velox_docker.yml#L197C3-L197C26


You can use gluten-it arg `--extra-conf=...` to enable C2C during testing.

And sorry for late response. Let's have this merged asap.

FelixYBW:
Oh, just noted the PR is still open and have many conflict. @boneanxs would you like to continue?

boneanxs:
Hey @FelixYBW @zhztheplayer, yea, I'm willing to continue this pr. After last comment, I actually have run some benchmarks in my local environment, and found there's no obviously improvement comparing VanillaColumnar->row->VeloxColumnar(Also I notice there's a compatible issue since this pr relies on spark arrow version, which is conflict with gluten). I'm still checking why this(performance not improved) could happen(but didn't catch time recently), will update here once I'm done other things.

zhouyuan:
@boneanxs hi, could you please help to do rebase? There's a big commit(renaming io.glutneproject to org.apache.gluten) when we migrate to apache repo.

thanks
-yuan

zhztheplayer:
comment to keep the PR open as it could be a valuable topic

PR #4822, opened 2024-03-01
[GLUTEN-4424][CORE] Upgrade spark version to 3.5.1 in Gluten

JkSelf:

This PR is a follow-up to [PR#4425](https://github.com/oap-project/gluten/pull/4425). Many thanks to @holdenk for her previous work

The mainly changes as below:

1. `WhostageTransfomer`: Spark 3.5 has changed the parameter type of the `generateTreeString `API in TreeNode, which prevents `WhostageTransfomer ` from directly overriding the `generateTreeString `method. A `GenerateTreeStringShim `trait is defined in the shim to override `generateTreeString`.
2. Spark 3.5 has changed the `PartitionDirectory `API and `PartitionedFileUtil.splitFiles` API. To be compatible with 3.5, we have made these changes in the shim layer.
3. `ColumnarShuffleExchangeExec`: Spark 3.5 has added a new interface, `advisoryPartitionSize`, in `ShuffleExchangeLike`.
4. `GlutenQueryTest`: Spark 3.5 has introduced ExtendedAnalysisException to replace the previous AnalysisException. The solution is to introduce ExtendedAnalysisException in versions prior to 3.5.
5. `ColumnarBuildSideRelation` and `BroadcastUtils`: Spark 3.5 has modified the `fromAttributes `and `toAttributes `interfaces in `StructType`.

## How was this patch tested?

Add tpcds/h test

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
@@ -100,8 +100,6 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
   with UnaryTransformSupport {
   assert(child.isInstanceOf[TransformSupport])
 
-  def childPlan: SparkPlan = child
-
   def stageId: Int = transformStageId
 
   def substraitPlanJsonValue: String = substraitPlanJson
--- a/gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
@@ -96,8 +96,17 @@ trait UnaryTransformSupport extends TransformSupport with UnaryExecNode {
 
 case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = false)(
     val transformStageId: Int
-) extends UnaryTransformSupport {
+) extends GenerateTreeStringShim
+  with UnaryTransformSupport {
   assert(child.isInstanceOf[TransformSupport])
+
+  def childPlan: SparkPlan = child
+
+  def stageId: Int = transformStageId
+
+  def substraitPlanJsonValue: String = substraitPlanJson
+
+  def wholeStageTransformerContextDefined: Boolean = wholeStageTransformerContext.isDefined
 
   // For WholeStageCodegen-like operator, only pipeline time will be handled in graph plotting.
   // See SparkPlanGraph.scala:205 for reference.
@@ -157,34 +166,6 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
   override def outputOrdering: Seq[SortOrder] = child.outputOrdering
 
   override def otherCopyArgs: Seq[AnyRef] = Seq(transformStageId.asInstanceOf[Integer])
-
-  override def generateTreeString(
-      depth: Int,
-      lastChildren: Seq[Boolean],
-      append: String => Unit,
-      verbose: Boolean,
-      prefix: String = "",
-      addSuffix: Boolean = false,
-      maxFields: Int,
-      printNodeId: Boolean,
-      indent: Int = 0): Unit = {
-    val prefix = if (printNodeId) "^ " else s"^($transformStageId) "
-    child.generateTreeString(
-      depth,
-      lastChildren,
-      append,
-      verbose,
-      prefix,
-      addSuffix = false,
-      maxFields,
-      printNodeId = printNodeId,
-      indent)
-    if (verbose && wholeStageTransformerContext.isDefined) {
-      append(prefix + "Substrait plan:\n")
-      append(substraitPlanJson)
-      append("\n")
-    }
-  }
 
   // It's misleading with "Codegen" used. But we have to keep "WholeStageCodegen" prefixed to
   // make whole stage transformer clearly plotted in UI, like spark's whole stage codegen.
--- a/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
@@ -41,7 +41,8 @@ case class ColumnarShuffleExchangeExec(
     override val outputPartitioning: Partitioning,
     child: SparkPlan,
     shuffleOrigin: ShuffleOrigin = ENSURE_REQUIREMENTS,
-    projectOutputAttributes: Seq[Attribute])
+    projectOutputAttributes: Seq[Attribute],
+    advisoryPartitionSize: Option[Long] = None)
   extends ShuffleExchangeLike
   with GlutenPlan {
   private[sql] lazy val writeMetrics =
--- a/pom.xml
+++ b/pom.xml
@@ -139,6 +139,16 @@
       </properties>
     </profile>
     <profile>
+      <id>spark-3.5</id>
+      <properties>
+        <sparkbundle.version>3.5</sparkbundle.version>
+        <sparkshim.artifactId>spark-sql-columnar-shims-spark35</sparkshim.artifactId>
+        <spark.version>3.5.1</spark.version>
+        <delta.version>2.4.0</delta.version>
+        <delta.binary.version>24</delta.binary.version>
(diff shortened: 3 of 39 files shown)


ulysses-you:
the failed tests in `ubuntu2004-test-spark32` are due to https://github.com/oap-project/gluten/pull/4815, please do a rebase

JkSelf:
@holdenk @ulysses-you @PHILO-HE @zhouyuan Can you help to review? Thanks.

shims/spark32/src/main/scala/io/glutenproject/execution/GenerateTreeStringShim.scala
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.glutenproject.execution
+
+import org.apache.spark.sql.execution.SparkPlan
+
+trait GenerateTreeStringShim extends SparkPlan {
```
ulysses-you, line 21: can we extends `UnaryExecNode` ?

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
```diff
@@ -163,6 +163,8 @@ case class ColumnarShuffleExchangeExec(
 
   protected def withNewChildInternal(newChild: SparkPlan): ColumnarShuffleExchangeExec =
     copy(child = newChild)
+
+  def advisoryPartitionSize: Option[Long] = None
```
ulysses-you, line 167: shall we add it to constructor to respect `advisoryPartitionSize` ?

pom.xml
```diff
@@ -45,7 +45,7 @@
     <scala.version>2.12.15</scala.version>
     <spark.major.version>3</spark.major.version>
     <sparkbundle.version>3.2</sparkbundle.version>
-    <spark.version>3.4.2</spark.version>
+    <spark.version>3.5.1</spark.version>
```
ulysses-you, line 48: it's a bit hurry to change the default Spark version, I think leave it to 3.4.2 is better for now

ulysses-you (commented):
thank you @JkSelf for the work!

pom.xml
PHILO-HE: @ulysses-you, do we need this line? I note each profile already specifies its spark version, including the default activated profile, spark-3.2.

JkSelf: It seems we can delete this line.

ulysses-you: it can make idea happy without profile..

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
JkSelf: Sure. Updated.

shims/spark32/src/main/scala/io/glutenproject/execution/GenerateTreeStringShim.scala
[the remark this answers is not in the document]
JkSelf: @PHILO-HE @ulysses-you 

If we extend `TreeNode`, we would also need to specify [BaseType](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala#L69). However, these operations have already been specified when `SparkPlan `extends `TreeNode`, so there is no need to do it again here. I am more inclined to extend `UnaryExecNode`, because `WhostageTransformer ` also extends the `UnaryExecNode ` abstract class from Spark.

pom.xml
JkSelf: Revert to 3.4.2.

holdenk:
Awesome, thanks for taking on the upgrade :)
If it's still open on Thursday I'll take a look (otherwise I'm going to be riding motorcycles or on the beach hiding from computers)

P.S. my pronouns are she/her :)

zhouyuan:
There's a new operator for partial window need to be supported, fired one issue to track this https://github.com/oap-project/gluten/issues/4836

zhouyuan:
CC: @zzcclp

shims/spark34/src/main/scala/io/glutenproject/execution/GenerateTreeStringShim.scala
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Spark 3.5 has changed the parameter type of the generateTreeString API in TreeNode. In order to
+ * support multiple versions of Spark, we cannot directly override the generateTreeString method in
+ * WhostageTransformer. Therefore, we have defined the GenerateTreeStringShim trait in the shim to
+ * allow different Spark versions to override their own generateTreeString.
+ */
+
+trait GenerateTreeStringShim extends UnaryExecNode {
+  def childPlan: SparkPlan
```
ulysses-you, line 29: do we need `childPlan` ? can not just use child ?

ulysses-you (commented):
lgtm except one comment

shims/spark34/src/main/scala/io/glutenproject/execution/GenerateTreeStringShim.scala
JkSelf: Good catch. Removed.

ulysses-you (approved):

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
```diff
@@ -96,9 +96,16 @@ trait UnaryTransformSupport extends TransformSupport with UnaryExecNode {
...
 case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = false)(
     val transformStageId: Int
-) extends UnaryTransformSupport {
+) extends GenerateTreeStringShim
+  with UnaryTransformSupport {
   assert(child.isInstanceOf[TransformSupport])
 
+  def stageId: Int = transformStageId
+
+  def substraitPlanJsonValue: String = substraitPlanJson
```
PHILO-HE, line 105: Can we remove the new method `substraitPlanJsonValue` and just use the existing `substraitPlanJson`?  We may just need to declare `substraitPlanJson` in the extending trait.

shims/spark32/src/main/scala/org/apache/spark/sql/catalyst/ExtendedAnalysisException.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.trees.Origin
+
+/**
+ * Thrown when a query fails to analyze, usually because the query itself is invalid.
+ *
+ * @since 1.3.0
+ */
+@Stable
+class ExtendedAnalysisException protected[sql] (
```
PHILO-HE, line 30: Can we just keep one for this new exception class? Assume they have no difference. Maybe, we can move it into a common place.

PHILO-HE (commented):
Some trivial comments. Could you check and see if it makes sense? Thanks!

gluten-core/src/main/scala/io/glutenproject/execution/WholeStageTransformer.scala
JkSelf: Sure. I will updated in the following PRs. Thanks @PHILO-HE

shims/spark32/src/main/scala/org/apache/spark/sql/catalyst/ExtendedAnalysisException.scala
JkSelf: Do you mean to move it in shim/common? This class only need in 32/33/34. And 35 is not needed. It maybe not make sense to move to shim/common.

PHILO-HE: Not a strong feeling on this. Keeping the current change is OK to me. BTW, it would be better to also add some comment for these classes to let people know why they are introduced. Thanks!

PR #4843, opened 2024-03-04
[CORE] Support In list option contains non-foldable expression

ulysses-you:

This pr adds a rule to rewrite `In` if the list option contain non-foldable value. A rewrite example:

```
SELECT * FROM t WHERE c in (1, 2, c2)
=>
SELECT * FROM t WHERE c in (1, 2) or c = c2
```

## How was this patch tested?

add test

### DIFF
--- a/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickhouseFunctionSuite.scala
+++ b/backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickhouseFunctionSuite.scala
@@ -20,14 +20,17 @@ import io.glutenproject.GlutenConfig
 import io.glutenproject.utils.UTSystemParameters
 
 import org.apache.spark.SparkConf
-import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.execution.datasources.v2.clickhouse.ClickHouseLog
 
 import org.apache.commons.io.FileUtils
 
 import java.io.File
 
-class GlutenClickhouseFunctionSuite extends GlutenClickHouseTPCHAbstractSuite {
+class GlutenClickhouseFunctionSuite
+  extends GlutenClickHouseTPCHAbstractSuite
+  with AdaptiveSparkPlanHelper {
 
   override protected val resourcePath: String =
     "../../../../gluten-core/src/test/resources/tpch-data"
@@ -37,9 +40,9 @@ class GlutenClickhouseFunctionSuite extends GlutenClickHouseTPCHAbstractSuite {
     rootPath + "../../../../gluten-core/src/test/resources/tpch-queries"
   override protected val queriesResults: String = rootPath + "queries-output"
 
-  override protected def createTPCHNullableTables(): Unit = {}
-
-  override protected def createTPCHNotNullTables(): Unit = {}
+  override protected def createTPCHNotNullTables(): Unit = {
+    createTPCHParquetTables(tablesPath)
+  }
 
   private var _hiveSpark: SparkSession = _
   override protected def spark: SparkSession = _hiveSpark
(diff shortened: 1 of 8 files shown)


ulysses-you:
cc @rui-mo @PHILO-HE thank you

ulysses-you:
@rui-mo it would fallback for both scan and filter

rui-mo (approved):

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
```diff
@@ -275,6 +275,10 @@ object ExpressionConverter extends SQLConfHelper with Logging {
           cw
         )
       case i: In =>
+        if (i.list.exists(!_.foldable)) {
+          throw new UnsupportedOperationException(
+            s"In list option does not support non-foldable expression, ${i.list.map(_.sql)}")
+        }
```
rui-mo, line 281: @zzcclp Is non-foldable expression not supported by CH backend either?

gluten-core/src/main/scala/io/glutenproject/extension/RewriteIn.scala
```diff
@@ -0,0 +1,73 @@
...
+          Some(in.copy(list = foldableExprs))
+        }
+        (newIn ++ Seq(rewrittenExpr)).reduce(Or)
+    }
+  }
+
+  override def apply(plan: SparkPlan): SparkPlan = {
+    plan match {
+      // TODO: Support datasource v2
+      case scan: FileSourceScanExec if scan.dataFilters.exists(_.find(shouldRewrite).isDefined) =>
```
PHILO-HE, line 64: Maybe, we can make a rule to only focus on handling spark expression instead of Spark plan if the rewriting should always take effect regardless of what Spark plan (scan, filter, etc.) it comes from.
If feasible, we can do it in the future.

PHILO-HE (approved):
Thanks for your good work!
Let's wait for CH developers to verify. cc @baibaichen

gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
zzcclp: yes, not supported.

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -1086,4 +1086,47 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
     }
   }
 
+  test("Support In list option contains non-foldable expression") {
```
zzcclp, line 1089: can you add this similar ut for the ch backend ?

gluten-core/src/main/scala/io/glutenproject/extension/RewriteIn.scala
ulysses-you: good point, if we are sure it's safe, we can call transformExpressions instead

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
ulysses-you: I try to add test for ch backend, let's see if the tests can pass..

backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickhouseFunctionSuite.scala
```diff
@@ -137,4 +140,59 @@ class GlutenClickhouseFunctionSuite extends GlutenClickHouseTPCHAbstractSuite {
       assert(diffCount == 0)
     }
   }
+
+  private def checkFallbackOperators(df: DataFrame, num: Int): Unit = {
```
zzcclp, line 144: it's better to put this tool function into WholeStageTransformerSuite ?

zzcclp:
LGTM

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -51,7 +50,7 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
       .set("spark.memory.offHeap.size", "2g")
       .set("spark.unsafe.exceptionOnMemoryLeak", "true")
       .set("spark.sql.autoBroadcastJoinThreshold", "-1")
-      .set("spark.sql.sources.useV1SourceList", "avro")
+      .set("spark.sql.sources.useV1SourceList", "avro,parquet")
```
ulysses-you, line 53: cc @PHILO-HE  I'm not sure why are we using v2 to read parquet. I changed to use v1 if it does not break your original idea.

backends-clickhouse/src/test/scala/io/glutenproject/execution/GlutenClickhouseFunctionSuite.scala
ulysses-you: moved

ulysses-you:
Run Gluten Clickhouse CI

yaooqinn (approved):

PR #4847, opened 2024-03-04
[VL](WIP) Support inline function

marin-ma:


marin-ma:
@zhouyuan Please kindly review. Thanks!

gluten-core/src/main/scala/io/glutenproject/execution/GenerateExecTransformer.scala
```diff
@@ -152,19 +154,65 @@ case class GenerateExecTransformer(
...
+        generatorNode,
+        childOutput,
+        extensionNode,
+        context,
+        operatorId)
+    }
+    applyPostProjectOnGenerator(generateRel, context, operatorId, childOutput, validation)
+  }
+
+  private def applyPostProjectOnGenerator(
```
zhouyuan, line 180: Could you please add one comment on when this should be called, any other cases than `inline`

zhouyuan (approved):
quite neat impl!

gluten-core/src/main/scala/io/glutenproject/execution/GenerateExecTransformer.scala
marin-ma: There are 3 types of CollectionGenerator in spark: `Explode`, `PosExplode` and `Inline`. Looks like only `inline` needs the post projection according Velox `UnnestNode` implementation. Besides, it would be better to implement this function in backends API. I will try `PosExplode` next and refine this part of code in another PR.

zhouyuan (approved):
👍 
The failure is not related

PR #4857, opened 2024-03-05
[GLUTEN-4853][CORE] Only trim Alias when its child is semantically equal to resAttr

liujiayi771:

Fix https://github.com/oap-project/gluten/issues/4853.
When an aggregate operation computes both `agg_func(expression)` and `agg_func(distinct expression)` on the same expression, the final determination of whether a post-project is needed can be incorrect. This is because the corresponding outputs in `resultExpressions` have `Alias` wrappers with `AttributeReference` that share the same name and type, leading to the conclusion that a post-project is not necessary. If we want to ensure that `Alias` do not generate additional post-projects, we should set a stricter verification. It is required that the child of the `Alias` is semantically equal to the native result attribute, and we cannot simply trim the `Alias`.

## How was this patch tested?
Add a new test case in `GlutenSQLAggregateFunctionSuite`.


liujiayi771:
cc @KevinyhZou, @ulysses-you, thanks.

gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPostProject.scala
```diff
@@ -49,7 +49,11 @@ object PullOutPostProject extends Rule[SparkPlan] with PullOutProjectHelper with
         agg.resultExpressions.size != allAggregateResultAttributes.size ||
         // Compare each item in result expressions and output attributes. Attribute in Alias
         // should be trimmed before checking.
-        agg.resultExpressions.map(trimAliases).zip(allAggregateResultAttributes).exists {
```
ulysses-you, line 52: shall we remove `AliasHelper` ?

liujiayi771: Removed.

ulysses-you (approved):

PR #4859, opened 2024-03-05
[VL]Bucket join support for Iceberg tables

SinghAsDev:

Bucket join support for Iceberg tables for Spark 3.4+

(Fixes: \#ISSUE-ID)

## How was this patch tested?

Added unit tests. Also, had run some tests on private cluster with a custom build and ensured shuffles are avoided where applicable and results are valid.

### DIFF
--- a/gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxIcebergSuite.scala
+++ b/gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxIcebergSuite.scala
@@ -43,6 +43,21 @@ class VeloxIcebergSuite extends WholeStageTransformerSuite {
       .set("spark.sql.catalog.spark_catalog.warehouse", s"file://$rootPath/tpch-data-iceberg-velox")
   }
 
+  private def isSparkVersionAtleast(version: String): Boolean = {
+    val currentVersion = spark.version
+    val currentVersionSplit = currentVersion.split("\\.")
+    val versionSplit = version.split("\\.")
+    currentVersionSplit.zip(versionSplit).foreach {
+      case (current, required) =>
+        if (current.toInt > required.toInt) {
+          return true
+        } else if (current.toInt < required.toInt) {
+          return false
+        }
+    }
+    true
+  }
+
   test("iceberg transformer exists") {
     spark.sql("""
                 |create table iceberg_tb using iceberg as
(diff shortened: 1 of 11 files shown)


shims/spark34/src/main/scala/io/glutenproject/sql/shims/spark34/Spark34Shims.scala
```diff
@@ -252,4 +254,45 @@ class Spark34Shims extends SparkShims {
...
+  }
+
+  override def orderPartitions(
+      scan: Scan,
+      keyGroupedPartitioning: Option[Seq[Expression]],
+      filteredPartitions: Seq[Seq[InputPartition]],
+      outputPartitioning: Partitioning): Seq[InputPartition] = {
+    scan match {
+      case _ if keyGroupedPartitioning.isDefined =>
+        var finalPartitions = filteredPartitions
```
yma11, line 274: Nit: may name to `newPartitions` to comply with vanilla spark?

shims/spark34/src/main/scala/io/glutenproject/sql/shims/spark34/Spark34Shims.scala
```diff
@@ -252,4 +254,45 @@ class Spark34Shims extends SparkShims {
...
+  override def getKeyGroupedPartitioning(batchScan: BatchScanExec): Option[Seq[Expression]] = {
+    batchScan.keyGroupedPartitioning
+  }
+
+  override def getCommonPartitionValues(
+      batchScan: BatchScanExec): Option[Seq[(InternalRow, Int)]] = {
+    batchScan.commonPartitionValues
+  }
+
+  override def orderPartitions(
```
yma11, line 267: Shouldn't  have the partitions distribution been done at plan analysis phase? Why do we need to add this step at transformer layer?

yma11:
Actually, we have made support for Spark3.5 even w/o UTs added. Can you also help apply the shim layer change against this version?

shims/spark34/src/main/scala/io/glutenproject/sql/shims/spark34/Spark34Shims.scala
SinghAsDev: The partitions are ordered when partitions are accessed in wholestagecodegen, since we override wholestagecodegen we need to handle in transformer layer.

SinghAsDev:
> Actually, we have made support for Spark3.5 even w/o UTs added. Can you also help apply the shim layer change against this version?

Hi @yma11 is it ok if I handle spark3.5 in a separate diff. It will require a bit more change as spark's `BatchScanExec` api changed in spark 3.5 and removed some of the params like keyGroupPartitioning, etc.

gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxIcebergSuite.scala
```diff
@@ -56,44 +71,246 @@ class VeloxIcebergSuite extends WholeStageTransformerSuite {
     }
   }
 
-  test("iceberg partitioned table") {
-    withTable("p_str_tb", "p_int_tb") {
+  test("iceberg bucketed join") {
+    assume(isSparkVersionAtleast("3.4"))
```
yma11, line 75: So these tests doesn't work for Spark3.3? should we fix them, maybe in late PR?

yma11 (approved):

gluten-iceberg/src/test/scala/io/glutenproject/execution/VeloxIcebergSuite.scala
SinghAsDev: Yea, in my view the storage partitioned join was usable mostly from Spark 3.4. So, I didn't bother making it work for 3.3. Let me know if you think some users are using spj on 3.3 and I can look more. Maybe in 3.5 follow up I will add some documentation on README about SPJ support.

FelixYBW: Let's add the limitation to Readme that it only work on Spark3.2 now. The other gluten user may add support to 3.3 or later if they need

FelixYBW (approved):

PR #4871, opened 2024-03-06
[VL] Add flushable decimal sum test case

liujiayi771:

Add flushable decimal sum test case.

## How was this patch tested?

Add new test case.


liujiayi771:
cc @rui-mo, thanks.

rui-mo (approved):

PR #4872, opened 2024-03-06
[GLUTEN-4875][VL]Support spark sql conf sortBeforeRepartition to avoid stage partial retry casuing result mismatch

zjuwangg:

Spark introduced `spark.sql.execution.sortBeforeRepartition` config in https://issues.apache.org/jira/browse/SPARK-23207 to keep the result correct, and we should do the same thing in gluten plan to achieve the same affact.

(Fixes: https://github.com/oap-project/gluten/issues/4875)

## How was this patch tested?

* Added ut to verify the executed plan is as expected.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
@@ -239,15 +239,14 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
         val sortOrder = SortOrder(projectTransformer.output.head, Ascending)
         val sortByHashCode = SortExecTransformer(Seq(sortOrder), global = false, projectTransformer)
         val dropSortColumnTransformer = ProjectExecTransformer(projectList.drop(1), sortByHashCode)
-        if (dropSortColumnTransformer.doValidate().isValid) {
+        val validationResult = dropSortColumnTransformer.doValidate()
+        if (validationResult.isValid) {
           ColumnarShuffleExchangeExec(
             shuffle,
             dropSortColumnTransformer,
             dropSortColumnTransformer.output)
         } else {
-          TransformHints.tagNotTransformable(
-            shuffle,
-            dropSortColumnTransformer.doValidate().reason.get)
+          TransformHints.tagNotTransformable(shuffle, validationResult)
           shuffle.withNewChildren(newChild :: Nil)
         }
       case _ =>
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
@@ -36,12 +36,12 @@ import org.apache.spark.sql.catalyst.{AggregateFunctionRewriteRule, FlushableHas
 import org.apache.spark.sql.catalyst.analysis.FunctionRegistry.FunctionBuilder
 import org.apache.spark.sql.catalyst.catalog.BucketSpec
 import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
-import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, Cast, CreateNamedStruct, ElementAt, Expression, ExpressionInfo, Generator, GetArrayItem, GetMapValue, GetStructField, If, IsNaN, Literal, Murmur3Hash, NamedExpression, NaNvl, PosExplode, Round, StringSplit, StringTrim}
+import org.apache.spark.sql.catalyst.expressions.{Alias, Ascending, Attribute, Cast, CreateNamedStruct, ElementAt, Expression, ExpressionInfo, Generator, GetArrayItem, GetMapValue, GetStructField, If, IsNaN, Literal, Murmur3Hash, NamedExpression, NaNvl, PosExplode, Round, SortOrder, StringSplit, StringTrim}
 import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, HLLAdapter}
 import org.apache.spark.sql.catalyst.optimizer.BuildSide
 import org.apache.spark.sql.catalyst.plans.JoinType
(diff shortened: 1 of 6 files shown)


cpp/core/jni/JniWrapper.cc
```diff
@@ -830,7 +830,8 @@ JNIEXPORT jlong JNICALL Java_io_glutenproject_vectorized_ShuffleWriterJniWrapper
     jint startPartitionId,
     jint pushBufferMaxSize,
     jobject partitionPusher,
-    jstring partitionWriterTypeJstr) {
+    jstring partitionWriterTypeJstr,
+    jboolean sortBeforeRR) {
```
marin-ma, line 834: Would the argument `sortBeforeRR` be used on the native side?

cpp/core/shuffle/RoundRobinPartitioner.h
```diff
@@ -23,8 +23,8 @@ namespace gluten {
 
 class RoundRobinPartitioner final : public Partitioner {
  public:
-  RoundRobinPartitioner(int32_t numPartitions, int32_t startPartitionId)
-      : Partitioner(numPartitions, false), pidSelection_(startPartitionId % numPartitions) {}
+  RoundRobinPartitioner(int32_t numPartitions, int32_t startPartitionId, bool hasPid)
```
marin-ma, line 26: `hasPid` should be `sortBeforeRR`, if we need this parameter.

cpp/core/jni/JniWrapper.cc
zjuwangg: Yes. RoundRobinPartitioner implement should know whether drop the first sorted column.

cpp/core/shuffle/RoundRobinPartitioner.h
zjuwangg: addressed.

cpp/core/jni/JniWrapper.cc
marin-ma: Would it be possible to add another projection before to drop the extra column?

zjuwangg: It's possible, but we can leverage current design to eliminate  the first column and avoid the performance loss by providing extra project.

marin-ma: I doubt that the impact of performance loss by adding one project which is only used to drop one column. The most part of extra overhead should come from the sort node. By adding this projection, we can also avoid adding argument `sortBeforeRR` to the several layers in the call stack.

zjuwangg: Sounds great. I will address this comment later.

zjuwangg:
@marin-ma plz help review again, there is a ci pre-check failure due to network problem.

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/execution/GlutenReplaceHashWithSortAggSuite.scala
```diff
@@ -72,6 +72,7 @@ class GlutenReplaceHashWithSortAggSuite
                |   SORT BY key
                |)
                |GROUP BY key
+               |ORDER BY key
```
marin-ma, line 75: Why is this change needed? Could you add some comments?

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/GlutenReplaceHashWithSortAggSuite.scala
```diff
@@ -71,6 +71,7 @@ class GlutenReplaceHashWithSortAggSuite
                |   SORT BY key
                |)
                |GROUP BY key
+               |ORDER BY key
```
marin-ma, line 74: Same as above.

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/execution/GlutenReplaceHashWithSortAggSuite.scala
zjuwangg: Because repartition modification causing the result sort order not same and casuing the result not same.

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/GlutenReplaceHashWithSortAggSuite.scala
zjuwangg: ditto

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/execution/GlutenReplaceHashWithSortAggSuite.scala
marin-ma: Thanks. Please document in the code as this modification makes our test sql different from the original one https://github.com/apache/spark/blob/922f5f686dc72433a9028bbe471a35a5b84f2855/sql/core/src/test/scala/org/apache/spark/sql/execution/ReplaceHashWithSortAggSuite.scala#L61-L72

marin-ma:
LGTM. Please document the sql query modification in the code and I will proceed with merging. Thanks!

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
```diff
@@ -232,7 +232,24 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
...
+        val dropSortColumnTransformer = ProjectExecTransformer(projectList.drop(1), sortByHashCode)
+        if (dropSortColumnTransformer.doValidate().isValid) {
+          ColumnarShuffleExchangeExec(
+            shuffle,
+            dropSortColumnTransformer,
+            dropSortColumnTransformer.output)
+        } else {
+          TransformHints.tagNotTransformable(
+            shuffle,
+            dropSortColumnTransformer.doValidate().reason.get)
```
ulysses-you, line 250: Can we reuse the previous validation result ?

zjuwangg: addressed

ulysses-you (approved):

PR #4879, opened 2024-03-07
[VL] Remove fallback for rand function with user-specified seed

PHILO-HE:

Allow offloading rand(seed) to native.

## How was this patch tested?

Spark UTs & one added test.


PHILO-HE:
Pending on till partition id is passed to velox config.

Issue #4884, opened 2024-03-07
[VL] Call getPartitions once in WholeStageTransformer


acvictor:
### Description

In `WholeStageTransformer` `getPartitions` is called twice - once in `getSplitInfos` and another in `val allScanPartitions = basicScanExecTransformers.map(_.getPartitions)`. This results in a code loop that is executed for twice for the same sequence of partitions including https://github.com/oap-project/gluten/blob/f775d42b5c96709f7beacbd521baebf33337c0d6/gluten-core/src/main/scala/io/glutenproject/utils/InputPartitionsUtil.scala#L37. We should call `getPartitions` once and reuse it.

PR #4885, opened 2024-03-07
Call getPartitions once in WholeStageTransformer

acvictor:

In `WholeStageTransformer` `getPartitions` is called twice - once in `getSplitInfos` and another in `val allScanPartitions = basicScanExecTransformers.map(_.getPartitions)`. This results in a code loop that is executed for twice for the same sequence of partitions including https://github.com/oap-project/gluten/blob/f775d42b5c96709f7beacbd521baebf33337c0d6/gluten-core/src/main/scala/io/glutenproject/utils/InputPartitionsUtil.scala#L37. This PR makes a fix to call `getPartitions` only once and reuse it.

(Fixes: \#4884)


acvictor:
CC @zhli1142015

zhli1142015 (approved):
LGTM, thanks

PR #4893, opened 2024-03-08
[VL] Fix wrong plan equality due to case class inheritance

zhztheplayer:
We had some case classes inheriting Spark's case class `BatchScanExec` or `FileSourceScanExec` or `HiveTableScanExec`. The case class inheritance is usually considered a bad practice since it breaks case class's equality convention.

The patch will fix the issue by putting Vanilla spark's code as abstract classes into shim layers.

Another pending [PR](https://github.com/apache/incubator-gluten/pull/4888) would require for this change so as the test of this patch, along with all the existing UTs.

The following UTs will be disabled for this path since they are based on vanilla Spark's non-overridable code that checks against  exact scan class types:

```
File source v2: support passing data filters to FileScan without partitionFilters
File source v2: support partition pruning
disable bucketing when the output doesn't contain all bucketing columns
Fallback Parquet V2 to V1
Aggregates with no groupby over tables having 1 BUCKET, return multiple rows
SPARK-32859: disable unnecessary bucketed table scan - other operators test
SPARK-32859: disable unnecessary bucketed table scan - multiple bucketed columns test
SPARK-32859: disable unnecessary bucketed table scan - multiple joins test
SPARK-32859: disable unnecessary bucketed table scan - basic test
```

### DIFF
--- a/gluten-ut/spark33/src/test/scala/io/glutenproject/utils/clickhouse/ClickHouseTestSettings.scala
+++ b/gluten-ut/spark33/src/test/scala/io/glutenproject/utils/clickhouse/ClickHouseTestSettings.scala
@@ -339,6 +339,10 @@ class ClickHouseTestSettings extends BackendTestSettings {
     .exclude("Option recursiveFileLookup: disable partition inferring")
     .exclude("SPARK-31116: Select nested schema with case insensitive mode")
     .exclude("SPARK-35669: special char in CSV header with filter pushdown")
+    // DISABLED: GLUTEN-4893 Vanilla UT checks scan operator by exactly matching the class type
+    .exclude("File source v2: support passing data filters to FileScan without partitionFilters")
+    // DISABLED: GLUTEN-4893 Vanilla UT checks scan operator by exactly matching the class type
+    .exclude("File source v2: support partition pruning")
     .excludeGlutenTest("Spark native readers should respect spark.sql.caseSensitive - parquet")
     .excludeGlutenTest("SPARK-25237 compute correct input metrics in FileScanRDD")
     .excludeGlutenTest("Option recursiveFileLookup: disable partition inferring")
@@ -994,6 +998,8 @@ class ClickHouseTestSettings extends BackendTestSettings {
     .exclude("SPARK-33267: push down with condition 'in (..., null)' should not throw NPE")
   enableSuite[GlutenDeleteFromTableSuite]
   enableSuite[GlutenFileDataSourceV2FallBackSuite]
+    // DISABLED: GLUTEN-4893 Vanilla UT checks scan operator by exactly matching the class type
+    .exclude("Fallback Parquet V2 to V1")
   enableSuite[GlutenKeyGroupedPartitioningSuite]
     .exclude("partitioned join: number of buckets mismatch should trigger shuffle")
     .exclude("partitioned join: only one side reports partitioning")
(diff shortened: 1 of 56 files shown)


ulysses-you:
Thank you @zhztheplayer for the improvement. I'm wondering why we need to make our transformer plan inherit spark operator. Is there any history reason ?

Is it possible that just pass spark operator to transformer plan as a parameter ? For example, like the comet: [CometScanExec](https://github.com/apache/arrow-datafusion-comet/blob/main/spark/src/main/scala/org/apache/spark/sql/comet/CometScanExec.scala)

zhztheplayer:
> Thank you @zhztheplayer for the improvement. I'm wondering why we need to make our transformer plan inherit spark operator. Is there any history reason ?

Yes it was mainly because of historical reasons of this project. We used to have different backgrounds of developers with various knowledge on Scala at the very first launching phase of this project.

> Is it possible that just pass spark operator to transformer plan as a parameter ? For example, like the comet: [CometScanExec](https://github.com/apache/arrow-datafusion-comet/blob/main/spark/src/main/scala/org/apache/spark/sql/comet/CometScanExec.scala)

That looks like a good idea and we may need to check if it's feasible to use that approach for Gluten too. Probably it can be done in shim layer without altering too much core module code. Although it may require for considerable refactor work. Once doing that, we can just remove the pasted files from this PR.

gluten-ut/spark33/src/test/scala/io/glutenproject/utils/clickhouse/ClickHouseTestSettings.scala
```diff
@@ -339,6 +339,10 @@ class ClickHouseTestSettings extends BackendTestSettings {
     .exclude("Option recursiveFileLookup: disable partition inferring")
     .exclude("SPARK-31116: Select nested schema with case insensitive mode")
     .exclude("SPARK-35669: special char in CSV header with filter pushdown")
+    // DISABLED: GLUTEN-4893 Vanilla UT checks scan operator by exactly matching the class type
+    .exclude("File source v2: support passing data filters to FileScan without partitionFilters")
+    // DISABLED: GLUTEN-4893 Vanilla UT checks scan operator by exactly matching the class type
+    .exclude("File source v2: support partition pruning")
```
zhztheplayer, line 345: @zzcclp 

I am disabling some UTs (including CH ones) that are based on strict type checking of the scan plan. We'd fix them case by case in later PRs but probably by pasting Spark's code to our UT folder. Or there might be better solutions, we can keep thinking of it. At this time it's more important to have the case class inheritance issue corrected, I don't have enough resource to do everything in one patch. So would that work for you?

zzcclp: It's ok to me.

Yohahaha:
seems copy Apache Spark source file into Gluten brings a limitation that require user must use same commit id or tag of these copied source file, if user has modified these source file in their own Spark, they may need apply changes in Gluten copied source file too, correct me if I was wrong.

I believe the way below is better than copy more and more Spark source files into Gluten.

> Is it possible that just pass spark operator to transformer plan as a parameter ? For example, like the comet: [CometScanExec](https://github.com/apache/arrow-datafusion-comet/blob/main/spark/src/main/scala/org/apache/spark/sql/comet/CometScanExec.scala)

PR #4920, opened 2024-03-11
[VL] Fix protobuf configure arguments in get_velox.sh

liujiayi771:

Velox already has install_protobuf in [setup-centos8.sh](https://github.com/facebookincubator/velox/blob/02ca9b0b4f554868b533d2f6526a480ea1e7d035/scripts/setup-centos8.sh#L107-L116), we should only use sed to modify the configure arguments.

## How was this patch tested?

N/A


liujiayi771:
cc @PHILO-HE, thanks.

PHILO-HE (approved):
Thanks!

PR #4931, opened 2024-03-12
[CORE] Prior to #4893, add vanilla Spark's original scan source code to keep git history

zhztheplayer:
This should be merged as an individual commit before merging https://github.com/apache/incubator-gluten/pull/4893.

To keep the git history then further developers would know how much of changes we applied to the source files.

The relevant Spark tags to refer to:

https://github.com/apache/spark/tree/v3.2.2
https://github.com/apache/spark/tree/v3.3.1
https://github.com/apache/spark/tree/v3.4.2
https://github.com/apache/spark/tree/v3.5.1

### DIFF
--- a/shims/spark32/src/main/scala/org/apache/spark/sql/execution/AbstractFileSourceScanExec.scala
+++ b/shims/spark32/src/main/scala/org/apache/spark/sql/execution/AbstractFileSourceScanExec.scala
@@ -1,0 +1,524 @@
+package org.apache.spark.sql.execution
+
+import org.apache.hadoop.fs.Path
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.catalog.BucketSpec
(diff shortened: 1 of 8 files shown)


Yohahaha:
could you add Spark commit id as well?

zhztheplayer:
> could you add Spark commit id as well?

It should be from Spark GitHub repo's version tags, but I am not 100% sure. Some of them were fetched from their released source jar, I didn't have a check about commit ID of them. 

I'll update the PR description to refer to the relevant Spark tags.

Yohahaha:
> > could you add Spark commit id as well?
> 
> It should be from Spark GitHub repo's version tags, but I am not 100% sure. Some of them were fetched from their released source jar, I didn't have a check about commit ID of them.
> 
> I'll update the PR description to refer to the relevant Spark tags, in case we bumped the versions after some time.

tag id is ok for me, thanks!

PR #4901, opened 2024-03-08
[VL] Support posexplode function and code refactoring on GenerateExecTransformer

marin-ma:
This patch contains 2 major changes:

- Support `posexplode` for Velox backend.
- Separate `GenerateExecTransformer` for Velox/CH backends. For Velox backend, the RelNode creation requires pre/projection for different generators. For CH backend, the RelNode creation is kept clean.

PR #574, opened 2022-11-17
[CH-191] Support generate exec

taiyang-li:
## What changes were proposed in this pull request?

Support generate exec. close https://github.com/Kyligence/ClickHouse/issues/191 

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

test by CH[[194]]

PR #4952, opened 2024-03-14
[CORE] Pullout pre/post project for generate

marin-ma:
A follow-up of https://github.com/apache/incubator-gluten/pull/4901

Pullout pre/post project for `GenerateExec`. The added projections are moved into Backends API. With this approach, `Literal` and `CreateArray` for `Inline` are also supported as they can be directly constructed by the expression transformer in the project.

Other fixes: 1. Fix coredump on array of null Struct Literal; fallback this case. 2. Add field name for constructing struct literal in `SubstratiToVeloxExpr` to avoid the downstream Velox operator from referencing unexpected struct fields by name.

### DIFF
--- a/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
+++ b/backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
@@ -764,4 +764,8 @@ class CHSparkPlanExecApi extends SparkPlanExecApi {
   ): GenerateExecTransformerBase = {
     CHGenerateExecTransformer(generator, requiredChildOutput, outer, generatorOutput, child)
   }
+
+  override def genPreProjectForGenerate(generate: GenerateExec): SparkPlan = generate
+
+  override def genPostProjectForGenerate(generate: GenerateExec): SparkPlan = generate
 }
--- a/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
@@ -44,7 +44,7 @@ import org.apache.spark.sql.catalyst.plans.JoinType
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.plans.physical.{BroadcastMode, HashPartitioning, Partitioning, RoundRobinPartitioning}
 import org.apache.spark.sql.catalyst.rules.Rule
-import org.apache.spark.sql.execution.{BroadcastUtils, ColumnarBuildSideRelation, ColumnarShuffleExchangeExec, SparkPlan, VeloxColumnarWriteFilesExec}
+import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.datasources.{FileFormat, WriteFilesExec}
 import org.apache.spark.sql.execution.exchange.{BroadcastExchangeExec, ShuffleExchangeExec}
 import org.apache.spark.sql.execution.joins.BuildSideRelation
@@ -676,5 +676,13 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
       child: SparkPlan
   ): GenerateExecTransformerBase = {
     GenerateExecTransformer(generator, requiredChildOutput, outer, generatorOutput, child)
+  }
+
+  override def genPreProjectForGenerate(generate: GenerateExec): SparkPlan = {
+    PullOutGenerateProjectHelper.pullOutPreProject(generate)
+  }
+
+  override def genPostProjectForGenerate(generate: GenerateExec): SparkPlan = {
+    PullOutGenerateProjectHelper.pullOutPostProject(generate)
   }
 }
--- a/backends-velox/src/main/scala/io/glutenproject/execution/GenerateExecTransformer.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/execution/GenerateExecTransformer.scala
@@ -17,25 +17,24 @@
 package io.glutenproject.execution
 
 import io.glutenproject.backendsapi.BackendsApiManager
-import io.glutenproject.expression.{ConverterUtils, ExpressionConverter, ExpressionNames}
-import io.glutenproject.expression.ConverterUtils.FunctionConfig
+import io.glutenproject.execution.GenerateExecTransformer.supportsGenerate
 import io.glutenproject.extension.ValidationResult
 import io.glutenproject.metrics.{GenerateMetricsUpdater, MetricsUpdater}
-import io.glutenproject.substrait.`type`.TypeBuilder
 import io.glutenproject.substrait.SubstraitContext
-import io.glutenproject.substrait.expression.{ExpressionBuilder, ExpressionNode}
+import io.glutenproject.substrait.expression.ExpressionNode
 import io.glutenproject.substrait.extensions.{AdvancedExtensionNode, ExtensionBuilder}
 import io.glutenproject.substrait.rel.{RelBuilder, RelNode}
+import io.glutenproject.utils.PullOutProjectHelper
 
 import org.apache.spark.sql.catalyst.expressions._
-import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.{GenerateExec, ProjectExec, SparkPlan}
 import org.apache.spark.sql.execution.metric.SQLMetrics
-import org.apache.spark.sql.types.{ArrayType, LongType, MapType, StructType}
+import org.apache.spark.sql.types.IntegerType
 
-import com.google.common.collect.Lists
 import com.google.protobuf.StringValue
 
 import scala.collection.JavaConverters._
+import scala.collection.mutable
 
 case class GenerateExecTransformer(
     generator: Generator,
@@ -62,26 +61,12 @@ case class GenerateExecTransformer(
   override protected def doGeneratorValidate(
       generator: Generator,
       outer: Boolean): ValidationResult = {
-    if (outer) {
-      return ValidationResult.notOk(s"Velox backend does not support outer")
-    }
-    generator match {
-      case _: JsonTuple =>
-        ValidationResult.notOk(s"Velox backend does not support this json_tuple")
-      case _: ExplodeBase =>
-        ValidationResult.ok
-      case Inline(child) =>
-        child match {
-          case AttributeReference(_, ArrayType(_: StructType, _), _, _) =>
-            ValidationResult.ok
-          case _ =>
-            // TODO: Support Literal/CreateArray.
-            ValidationResult.notOk(
-              s"Velox backend does not support inline with expression " +
-                s"${child.getClass.getSimpleName}.")
-        }
-      case _ =>
-        ValidationResult.ok
+    if (!supportsGenerate(generator, outer)) {
+      ValidationResult.notOk(
+        s"Velox backend does not support this generator: ${generator.getClass.getSimpleName}" +
+          s", outer: $outer")
+    } else {
+      ValidationResult.ok
     }
   }
 
@@ -91,30 +76,13 @@ case class GenerateExecTransformer(
       generatorNode: ExpressionNode,
       validation: Boolean): RelNode = {
     val operatorId = context.nextOperatorId(this.nodeName)
-
-    val newInput = if (!validation) {
-      applyPreProject(inputRel, context, operatorId)
-    } else {
-      // No need to validate the pre-projection. The generator output has been validated in
-      // doGeneratorValidate.
-      inputRel
-    }
-
-    val generateRel = RelBuilder.makeGenerateRel(
-      newInput,
+    RelBuilder.makeGenerateRel(
+      inputRel,
       generatorNode,
       requiredChildOutputNodes.asJava,
       getExtensionNode(validation),
       context,
       operatorId)
-
-    if (!validation) {
-      applyPostProject(generateRel, context, operatorId)
-    } else {
-      // No need to validate the post-projection on the native side as
-      // it only flattens the generator's output.
-      generateRel
-    }
   }
 
   private def getExtensionNode(validation: Boolean): AdvancedExtensionNode = {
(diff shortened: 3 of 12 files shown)


marin-ma:
@liujiayi771 Could you help to review?

I also have concerns about the fallback mechanism after pulling out the pre/post project. As the rules are applied before `TransformHintRule`, do we have a framework or new rule to remove the pre/post project if this operator fails its validation?

marin-ma:
cc: @zhouyuan

liujiayi771:
@marin-ma `RewriteSparkPlanRulesManager` will fallback these pre/post project.

liujiayi771:
cc @ulysses-you

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
```diff
@@ -750,4 +750,8 @@ class CHSparkPlanExecApi extends SparkPlanExecApi {
   ): GenerateExecTransformerBase = {
     CHGenerateExecTransformer(generator, requiredChildOutput, outer, generatorOutput, child)
   }
+
+  override def genPreProjectForGenerate(generate: GenerateExec): SparkPlan = generate
```
liujiayi771, line 754: The logic of pull out should also apply to CH? It's just that CH hasn't encountered a scenario that requires pull out yet?

liujiayi771, line 754: The logic of pull out should also apply to CH? It's just that CH hasn't encountered a scenario that requires pull out yet? If that is the case, then we can centralize these pieces of code within the pull out rule.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
```diff
@@ -659,4 +662,81 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
   ): GenerateExecTransformerBase = {
     GenerateExecTransformer(generator, requiredChildOutput, outer, generatorOutput, child)
   }
+
+  override def genPreProjectForGenerate(generate: GenerateExec): SparkPlan = {
+    if (supportsGenerate(generate)) {
+      val newGeneratorChild = generate.generator.asInstanceOf[UnaryExpression].child match {
```
liujiayi771, line 668: Some `Generator`s are not `UnaryExpression`, use match case, and throw exception if it is not `UnaryExpression`.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
```diff
@@ -659,4 +662,81 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
...
+        case attr: Attribute => attr
+        case expr @ (Literal(_, _) | CreateMap(_, _) | CreateArray(_, _)) =>
+          Alias(expr, generatePreAliasName)()
```
liujiayi771, line 671: It would be better to maintain consistency with the code patterns of other 'pull out' operations, using methods such as `isNotAttribute` and `replaceExpressionWithAttribute` from `PullOutProjectHelper`.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
```diff
@@ -659,4 +662,81 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
...
+        case other =>
+          throw new UnsupportedOperationException(
```
liujiayi771, line 673: Are these the only three scenarios that require a 'pull out'? Is it the case that in other situations, as long as it's not an Attribute, a 'pull out' can be performed without necessarily throwing an exception?

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
```diff
@@ -659,4 +662,81 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
...
+            s"Fail to execute ${generate.generator.getClass.getSimpleName} " +
+              s"with child type ${other.getClass.getSimpleName}")
+      }
+      generate.copy(
+        generator =
+          generate.generator.withNewChildren(Seq(newGeneratorChild)).asInstanceOf[Generator],
+        child = ProjectExec(generate.child.output :+ newGeneratorChild, generate.child)
```
liujiayi771, line 680: Use `eliminateProjectList` to get the `projectList` of pre project.

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
marin-ma: So far, CH doesn't need pre/post project for their already supported generator type. But not sure if it would be needed in the future with more `Generator` supported.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
marin-ma: These are all kind of child expression I've found from the existing UTs.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
marin-ma: Do you mean by using `elimainateProjectList(generate.child.output, newGeneratorChild)`? `newGeneratorChild` can be a duplicated Attribute in `generate.child.output`. It shouldn't be eliminated because the native backend identifies the last field as generator's input.

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
marin-ma: `Explode` and `PosExplode` are already supported in CH, and pre/post projection are not needed. Looks like we cannot unify the implementation.

gluten-core/src/main/scala/io/glutenproject/execution/GenerateExecTransformerBase.scala
```diff
@@ -26,11 +26,28 @@ import io.glutenproject.substrait.expression.{ExpressionBuilder, ExpressionNode}
...
 
+import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, ExprCode}
 import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.types.StructType
 
 import scala.collection.JavaConverters._
 
+case class GlutenGeneratorWrapper(unwrapped: Generator) extends Generator {
```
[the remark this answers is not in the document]
marin-ma: This wrapper is used to stop applying the `PullOutPostProject` on the same `GenerateExec` once it has been applied.

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
liujiayi771: At least for pre-project, CH and Velox are consistent; they can only execute expressions within the project operator. I think that the logic of pre-project can also be applied to CH at present, it's just that CH has not yet started to adapt `GeneratorExec` that include expressions. I think you can put `GeneratorExec` into `needsPreProject` for validation, and only apply the pull-out logic when the child is not an `Attribute`. This should not affect the current CH, and later, if CH wants to adapt GeneratorExec` with expressions, it can be directly reused.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
liujiayi771: I think for non-`Attribute` cases, you can simply hand them over to the `replaceExpressionWithAttribute` method to handle, without the need to match the current cases only encountered in UTs. If other cases arise in the future, they can also be supported. If the new case is not yet supported by Gluten, it will only cause pre-project to fallback.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
liujiayi771: Could you add some comments for this explanation?

gluten-core/src/main/scala/io/glutenproject/execution/GenerateExecTransformerBase.scala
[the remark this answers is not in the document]
ulysses-you: We do not need to worry about this. `PullOutPostProject` would not apply same operator twice. The `RewriteSparkPlanRulesManager` framework makes sure only transforming once for each operator.

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
marin-ma: Here's the implementation in the initial commit. Could you help to check if that matches your expectation?
https://github.com/apache/incubator-gluten/pull/4952/commits/c3529b8a050fd40db18624c09f8c20e514934e03#diff-c49dd56d8ac0d29278ad2e6a33cc0cb97d86a4c74d3e5c011bad4733ba1203feR186-R195

The reason of giving up using `replaceExpressionWithAttribute` is that this function also matches `BouldReference`, which is not the case used by `Generator.child`, and the output cannot be directly used as the input to pre-project. The new child should be either the original `Attribute`, or an `Alias` to other expressions.

I would think removing the type list `@ (Literal(_, _) | CreateMap(_, _) | CreateArray(_, _))` here is a better way.

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
marin-ma: `GenerateExec` was initially implemented for CH, and they should have the native implementation for the generator expressions. Please check https://github.com/apache/incubator-gluten/pull/574

Velox backend does not have the native function implementation to the generator expressions. But it has an `UnnestNode` operator that can help to generate the expected output. That's why a pre/post projection is added to adopt the input/output for the `UnnestNode`.

I believe CH doesn't require the pre/post project logic for the supported generator. cc @taiyang-li

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
liujiayi771: @marin-ma Do you mean to say that if `Generator.child` is a `BoundReference`, it also needs to be wrapped with an Alias? 
The `BoundReference` within `replaceExpressionWithAttribute` is a special optimization case that was initially placed within the `case other`. I think if that's the only reason, we could add a parameter to `replaceExpressionWithAttribute` named `replaceBoundReference: Boolean`, with the default value of false. The `Generator` could set this to true. Then it could be changed to case `e: BoundReference if !replaceBoundReference => e`.

liujiayi771: I find that in your initial commit, the main logic codes are put in pull-out rule. Why do you move it to `SparkPlanExecApi` now? I think it’s better for `SparkPlanExecApi` to only control whether it's necessary to pull out.

marin-ma: @liujiayi771 This pull-out logic fully depends on the native implementation of Velox library. It's not a general case.

backends-clickhouse/src/main/scala/io/glutenproject/backendsapi/clickhouse/CHSparkPlanExecApi.scala
taiyang-li: ch doesn't need it for now. thanks for reminding me

gluten-core/src/main/scala/io/glutenproject/execution/GenerateExecTransformerBase.scala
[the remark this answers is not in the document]
marin-ma: After transforming, the `GenerateExec` becomes `ProjectExec(newChild: GenerateExec)`, and the rule continues to apply on the `newChild: GenerateExec`, which results in infinite recursion. By adding this wrapper, the next time the `supportsGenerate` will return false, preventing it from infinitely applying on the `GenerateExec` https://github.com/apache/incubator-gluten/pull/4952/files#diff-4ebd3cbecb20dc83596c28d20936d40f48ed1b41f334525fd5b5bdca3ac3bf3fR730

backends-velox/src/main/scala/io/glutenproject/execution/GenerateExecTransformer.scala
```diff
@@ -141,92 +108,100 @@ case class GenerateExecTransformer(
...
+  }
+}
 
-    RelBuilder.makeProjectRel(
-      inputRel,
-      projectExpressions.asJava,
-      context,
-      operatorId,
-      child.output.size)
+object PullOutGenerateProjectHelper extends PullOutProjectHelper {
```
marin-ma, line 129: @liujiayi771 Could you help to review again? I've made some code structure changes since your last review. This class is created for velox backend to generate the pre/post projection for GenerateExec. By extending `PullOutProjectHelper` here, the protected scope of the helper functions can be preserveed.

gluten-core/src/main/scala/io/glutenproject/execution/GenerateExecTransformerBase.scala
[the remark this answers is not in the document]
ulysses-you: I see, shall we change `transform` to `transformUp` ?

https://github.com/apache/incubator-gluten/blob/bb0464797a2fb129b345cc5e0b815ddf53da9b2f/gluten-core/src/main/scala/io/glutenproject/extension/columnar/RewriteSparkPlanRulesManager.scala#L75-L79

gluten-core/src/main/scala/io/glutenproject/utils/PullOutProjectHelper.scala
```diff
@@ -57,12 +57,13 @@ trait PullOutProjectHelper {
 
   protected def replaceExpressionWithAttribute(
       expr: Expression,
-      projectExprsMap: mutable.HashMap[Expression, NamedExpression]): Expression =
+      projectExprsMap: mutable.HashMap[Expression, NamedExpression],
+      replaceBoundReference: Boolean = true): Expression =
```
liujiayi771, line 61: The default value should be false, and `GenerateExec` set it to true. "replace" means use an `Alias` wrapper.

marin-ma:
@liujiayi771 @ulysses-you Could you help to review again? Thanks!

ulysses-you:
It seems the failed tests are irrelevant, I tried to re-run the failed tests.

liujiayi771:
@marin-ma LGTM. Thanks.

liujiayi771 (approved):

backends-velox/src/main/scala/io/glutenproject/backendsapi/velox/SparkPlanExecApiImpl.scala
```diff
@@ -659,4 +659,14 @@ class SparkPlanExecApiImpl extends SparkPlanExecApi {
...
+  override def genPreProjectForGenerate(generate: GenerateExec): SparkPlan = {
+    PullOutGenerateProjectHelper.pullOutPreProject(generate)
+  }
+
+  override def genPostProjectForGenerate(generate: GenerateExec): SparkPlan = {
+    PullOutGenerateProjectHelper.pullOutPostProject(generate)
+  }
 }
+
+object SparkPlanExecApiImpl {}
```
ulysses-you, line 672: nit: unnecessary change

ulysses-you (approved):

PR #4958, opened 2024-03-14
[VL] Remove glog level config in unit test

liujiayi771:

Currently, in the Velox backend, the configuration of glog is reused at the process level. If the glog level is set in a test case, it will cause all subsequent test cases to run with glog=0, generating a large amount of redundant log information. The glog level should not be set in test cases unless necessary.

## How was this patch tested?

N/A


liujiayi771:
After `NativeBenchmarkPlanGenerator`, all test case will use `glog=0`.
iShot_2024-03-14_17 51 23

liujiayi771:
Hi @marin-ma, Could you help check if we can remove the settings for glog and debug mode?

liujiayi771:
cc @PHILO-HE, @rui-mo, thanks.

PHILO-HE (approved):
Looks good! Thanks!

PR #4961, opened 2024-03-14
[VL] Add velox decimal avg sum large precision test case

liujiayi771:

When the input decimal precision is very large, the precision of sum and avg may exceed 38, and we should not infer the intermediate type based on the result type in decimal sum/avg functions, we should always infer the intermediate type and result type based on the raw input type.

For example, in average aggregate function:
input = decimal(p, s)
intermediate = row(decimal(min(38, p + 10), s), bigint)
result = decimal(min(38, p + 4), min(38, s + 4))

We should not infer intermediate type by using result precision + 6, it is not correct when p > 34.

Add some test cases with large precision to prevent future modifications to decimal avg sum from introducing errors.

## How was this patch tested?

Add large precision test cases in avg and sum `VeloxAggregateFunctionsSuite`.


liujiayi771:
cc @rui-mo, thanks.

rui-mo (approved):
Thanks!

Issue #4933, opened 2024-03-12
Bump Iceberg version


SteNicholas:
### Description

Iceberg has already released [1.4.0](https://github.com/apache/iceberg/releases/tag/apache-iceberg-1.4.0), [1.4.1](https://github.com/apache/iceberg/releases/tag/apache-iceberg-1.4.1), [1.4.2](https://github.com/apache/iceberg/releases/tag/apache-iceberg-1.4.2), [1.4.3](https://github.com/apache/iceberg/releases/tag/apache-iceberg-1.4.3) and [1.5.0](https://github.com/apache/iceberg/releases/tag/apache-iceberg-1.5.0). Meanwhile, Gluten only supports the Iceberg 1.3.1 version, which could support bumping the released version.

SteNicholas:
cc @liujiayi771.

j7nhai:
> The current implementation has compatibility issues with the latest version of Iceberg that need to be resolved.

Hi, @liujiayi771 Can you explain some detail of compatibility issues? We are currently using both iceberg 1.3 & 1.4, and we look forward to resolve these issues.

liujiayi771:
@j7nhai We have already upgraded the Iceberg version of Spark 3.4 to 1.4.3. The compatibility issues mentioned above might have been partially resolved. I don't remember the specific details, but you can try upgrading the version. If the end-to-end tests and unit tests can pass, it should be fine.

PR #4967, opened 2024-03-15
[GLUTEN-4933][VL] Update iceberg version to 1.4.3 for Spark 3.4 and above

yma11:

Update iceberg version to 1.4.3 for Spark 3.3 and above, as v1.3.1 is not compatible with Spark 3.5.
Note that iceberg's latest version is 1.5.0 but as it's not available in mirror repo, we use 1.4.3 instead. So the version mapping is:

```
Spark 3.2 : v1.3.1
Spark 3.3 : v1.3.1
Spark 3.4 : v1.4.3
Spark 3.5 : v1.4.3  
```

## How was this patch tested?

Existing UTs

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,6 @@
   <properties>
     <caffeine.version.java8>2.9.3</caffeine.version.java8>
     <delta.version>2.0.1</delta.version>
-    <iceberg.version>1.3.1</iceberg.version>
     <delta.binary.version>20</delta.binary.version>
     <scala.binary.version>2.12</scala.binary.version>
     <scala.version>2.12.15</scala.version>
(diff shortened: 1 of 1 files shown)


zhouyuan:
related: https://github.com/apache/incubator-gluten/issues/4933

yma11:
@liujiayi771 UT shows incompatible `ParquetWriter` API in Spark 3.3 + iceberg 1.4.3. I found that iceberg 1.4.3 is using parquet `1.13.1` while Spark 3.3 is using parquet `1.12.2`, so we should keep using iceberg 1.3.1 for Spark 3.3, any comments?

zhouyuan:
@yma11 can we use different iceberg for different spark? sounds this would fix the issue

liujiayi771:
@yma11 I think we should keep iceberg 1.3.1 for Spark 3.3. This compatibility issue was discovered before, which is why the Iceberg version 1.3.1 was chosen at that time.

zhouyuan (approved):
👍

PR #4968, opened 2024-03-15
[CORE] Add Complete case match in PullOutPreProject

liujiayi771:

Since ClickHouse has an optimizer rule `MergeTwoPhasesHashBaseAggregate` which will merge two aggregates and change the aggregate mode to `Complete`. `PullOutPreProject` rule should also match `Complete` mode.

## How was this patch tested?

Exists CI.


liujiayi771:
cc @zzcclp @ulysses-you, thanks.

ulysses-you (approved):

PR #4972, opened 2024-03-15
[VL] Fix generic benchmark usage when both split and data exists

Yohahaha:
This patch fixes the generic benchmark usage when simulate stage which contain scan task and shuffle read task at same time, e.g. BHJ.


marin-ma:
Thanks. I'm a bit confused about which kind of stage would get such plan that contains both scan and shuffle read. Is it BHJ stage? Could you update in the PR description?

Yohahaha:
> Thanks. I'm a bit confused about which kind of stage would get such plan that contains both scan and shuffle read. Is it BHJ stage? Could you update in the PR description?

yes, BHJ stage contains scan and shuffle read at same time.

marin-ma (approved):

PR #4978, opened 2024-03-15
[VL] Copy the libsodium.so with version number in dynamic packaging

liujiayi771:
#4870 Add libsodium.so in `SharedLibraryLoaderCentos8`. However, it should not link `libsodium.so` to `libsodium.so`, as this would cause the shared object to fail to load at runtime. I discovered the following warning while using GDB to debug:
```
warning: Could not load shared library symbols for /tmp/gluten-81e940a2-74ff-4c8d-8098-112e930af733/jni/5d511fe8-1801-4338-86dc-8b7e31613bac/gluten-9058260266840333844/libsodium.so.
```

We should load and create link for `libsodium.so.23` to `libsodium.so` or `libsodium.so.23.3.0` to `libsodium.so`.

## How was this patch tested?

N/A


liujiayi771:
cc @zhouyuan, @PHILO-HE, thanks.

PHILO-HE (commented):
Hi @liujiayi771, do we really need to pack and load lib sodium?  Looks it is not handled for other OS.

PHILO-HE (approved):
Just note https://github.com/apache/incubator-gluten/pull/4870 which introducing this lib.
The PR looks good. Thanks!

liujiayi771:
@PHILO-HE If we not pack and load libsodium, and not use yum to install sodium-devel, it will failed to load libvelox.so just like the description in #4870.

liujiayi771:
image

Issue #4990, opened 2024-03-18
[CH] data loss when write dynamic partitions


shuai-xu:
### Backend

CH (ClickHouse)

### Bug description

When run sql d_1039_0, which writer dynamic partitions, part of the data is lost. This is because the output file will be reopen each time when write for dynamic partition, and the open mode is overwrite.

### Spark version

Spark-3.3.x

PR #4991, opened 2024-03-18
[GLUTEN-4990] [CH] fix data loss for dynamic patition

shuai-xu:

(Fixes: \#4990)

## How was this patch tested?

this patch was tested by manual tests.


binmahone (changes requested):
add a test case?

shuai-xu:
> add a test case?

done

PR #4996, opened 2024-03-18
[CORE] Introduce GlutenNotSupportException for expected fallback behavior

PHILO-HE:

Currently, `java.lang.UnsupportedOperationException` will be thrown in Gluten when an operation is not supported. And Gluten will then collectively handle such exception and make fallback behavior happen. As a JDK class commonly used, such exception can be thrown by any code (including third-party code) in Gluten's validation/transform path, then will cause a fallback at last. But for some situation, we need developers to dig the reason and fix the exception. With a dedicated Gluten exception introduced, it may be helpful to debug by distinguishing the situations where developer expects fallback happens and other situations where developer doesn't expect.

## How was this patch tested?

Existing tests.

### DIFF
--- a/backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
+++ b/backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
@@ -17,6 +17,7 @@
 package io.glutenproject.execution
 
 import io.glutenproject.backendsapi.BackendsApiManager
+import io.glutenproject.exception.{GlutenException, GlutenNotSupportException}
 import io.glutenproject.expression._
 import io.glutenproject.expression.ConverterUtils.FunctionConfig
 import io.glutenproject.extension.columnar.RewriteTypedImperativeAggregate
@@ -26,16 +27,18 @@ import io.glutenproject.substrait.expression.{AggregateFunctionNode, ExpressionB
 import io.glutenproject.substrait.extensions.{AdvancedExtensionNode, ExtensionBuilder}
 import io.glutenproject.substrait.rel.{RelBuilder, RelNode}
 import io.glutenproject.utils.VeloxIntermediateData
+
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
+
 import com.google.protobuf.StringValue
-import io.glutenproject.exception.{GlutenException, GlutenNotSupportException}
 
 import java.lang.{Long => JLong}
 import java.util.{ArrayList => JArrayList, HashMap => JHashMap, List => JList}
+
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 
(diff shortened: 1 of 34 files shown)


PHILO-HE:
@ulysses-you, @zzcclp, could you review this pr? Thanks!

gluten-core/src/main/scala/io/glutenproject/GlutenPlugin.scala
```diff
@@ -144,7 +145,7 @@ private[glutenproject] class GlutenDriverPlugin extends DriverPlugin with Loggin
 
     // off-heap bytes
     if (!conf.contains(GlutenConfig.GLUTEN_OFFHEAP_SIZE_KEY)) {
-      throw new UnsupportedOperationException(s"${GlutenConfig.GLUTEN_OFFHEAP_SIZE_KEY} is not set")
+      throw new GlutenException(s"${GlutenConfig.GLUTEN_OFFHEAP_SIZE_KEY} is not set")
```
ulysses-you, line 148: `GlutenNotSupportException` ?

backends-velox/src/main/scala/io/glutenproject/execution/HashAggregateExecTransformer.scala
```diff
@@ -369,7 +370,7 @@ abstract class HashAggregateExecTransformer(
           exprNodes.addAll(childNodes)
 
         case _: HyperLogLogPlusPlus if aggFunc.aggBufferAttributes.size != 1 =>
-          throw new UnsupportedOperationException("Only one input attribute is expected.")
+          throw new GlutenException("Only one input attribute is expected.")
```
ulysses-you, line 373: `GlutenNotSupportException` ?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
```diff
@@ -737,11 +738,14 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
         // Currently we assume a plan to be transformable by default.
       }
     } catch {
-      case e: UnsupportedOperationException =>
+      case e @ (_: GlutenNotSupportException | _: UnsupportedOperationException) =>
```
ulysses-you, line 741: shall we remove `UnsupportedOperationException` ?

ulysses-you (commented):
thank you @PHILO-HE , lgtm except some comments

gluten-core/src/main/scala/io/glutenproject/GlutenPlugin.scala
PHILO-HE: Looks the exception here also cannot make fall back happen, but just let the execution fail. So I think `GlutenException` may be good. Thanks!

gluten-core/src/main/scala/io/glutenproject/extension/columnar/TransformHintRule.scala
PHILO-HE: I guess `UnsupportedOperationException` can still be thrown unexpectedly for some reasons in user's environment. And by still catching it here, the execution can possibly pass with the help of Gluten's fallback mechanism. So I feel it may be better to temporarily keep it. Thanks!

PHILO-HE (commented):
@ulysses-you, thanks so much for your review! Just updated and posted some thoughts.

ulysses-you (approved):

zzcclp (approved):
LGTM

Issue #4999, opened 2024-03-18
In ColumnarUnionExec, it would be better to be transformed to PartitionerAwareUnionRDD than UnionRDD when they has same partitioner.


guixiaowen:
### Description

In ColumnarUnionExec, it would be better to be transformed to PartitionerAwareUnionRDD than UnionRDD when they has same partitioner.

In ColumnarUnionExec, it will be transformed to UnionRDD.

For example:

select * from test a 
union all
select * from test b 
union all
select * from test c

They have the same partitioner.

In ColumnarUnionExec,  they will be transformed to rdd1.union(rdd2).union(rdd3).

After this pr,

In ColumnarUnionExec, they will be transformed to PartitionerAwareUnionRDD(sc, Seq(rdd1, rdd2, rdd3)) if they have the same partitioner.

PR #5000, opened 2024-03-18
In ColumnarUnionExec, it would be better to be transformed to Partiti…

guixiaowen:
…onerAwareUnionRDD than UnionRDD when they has same partitioner. #4999

For example:

select * from test a
union all
select * from test b
union all
select * from test c

They have the same partitioner.

In ColumnarUnionExec, they will be transformed to rdd1.union(rdd2).union(rdd3).

After this pr,

In ColumnarUnionExec, they will be transformed to PartitionerAwareUnionRDD(sc, Seq(rdd1, rdd2, rdd3)) if they have the same partitioner.

(Fixes: \#4999)

## How was this patch tested?


guixiaowen:
Update commit message info.

Issue #5003, opened 2024-03-18
Null literal Fallback


WangGuangxin:
### Backend

VL (Velox)

### Bug description

fix null literal fallback

### Spark version

None

PR #5004, opened 2024-03-18
[GLUTEN-5003][VL] Fix Null literal fallback

WangGuangxin:

Fix NULL literal fallback unexpectly

(Fixes: \#5003)

## How was this patch tested?

existing UT


marin-ma (approved):
Thanks.

marin-ma:
Note that I found `validateOffloadResult("SELECT array(struct(1, '2'), null)")` still generate coredump with this patch. I will try to fix that in another PR.

cc: @PHILO-HE

Issue #4889, opened 2024-03-08
[VL] Add support of percentile_approx / approx_percentile


zhztheplayer:
### Description

Part of https://github.com/apache/incubator-gluten/issues/4039

Velox(presto)'s `approx_percentile` has different intermediate types with Spark's function with same name.

Velox's signature code of `approx_percentile` :

```cpp
void addSignatures(
    const std::string& inputType,
    const std::string& percentileType,
    const std::string& returnType,
    std::vector<std::shared_ptr<exec::AggregateFunctionSignature>>&
        signatures) {
  auto intermediateType = fmt::format(
      "row(array(double), boolean, double, integer, bigint, {0}, {0}, array({0}), array(integer))",
      inputType);
  signatures.push_back(exec::AggregateFunctionSignatureBuilder()
                           .returnType(returnType)
                           .intermediateType(intermediateType)
                           .argumentType(inputType)
                           .argumentType(percentileType)
                           .build());
  signatures.push_back(exec::AggregateFunctionSignatureBuilder()
                           .returnType(returnType)
                           .intermediateType(intermediateType)
                           .argumentType(inputType)
                           .argumentType("bigint")
                           .argumentType(percentileType)
                           .build());
  signatures.push_back(exec::AggregateFunctionSignatureBuilder()
                           .returnType(returnType)
                           .intermediateType(intermediateType)
                           .argumentType(inputType)
                           .argumentType(percentileType)
                           .argumentType("double")
                           .build());
  signatures.push_back(exec::AggregateFunctionSignatureBuilder()
                           .returnType(returnType)
                           .intermediateType(intermediateType)
                           .argumentType(inputType)
                           .argumentType("bigint")
                           .argumentType(percentileType)
                           .argumentType("double")
                           .build());
}
```

Link https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/ApproxPercentileAggregate.cpp#L790C1-L828C1

Spark's code of `approx_percentile` / `percentile_approx`:
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/ApproximatePercentile.scala

ulysses-you:
Thank you @zhztheplayer for driving this, I think @liujiayi771 is the better one to take it :)

liujiayi771:
> Thank you @zhztheplayer for driving this, I think @liujiayi771 is the better one to take it :)

[...]

PR #5007, opened 2024-03-18
[GLUTEN-4889][VL] Support approx_percentile

WangGuangxin:

Support approx_percentile for VL backend

(Fixes: \#4889)

## How was this patch tested?

UT

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
@@ -39,7 +39,7 @@ import org.apache.spark.shuffle.utils.CHShuffleUtil
 import org.apache.spark.sql.{SparkSession, Strategy}
 import org.apache.spark.sql.catalyst.{CHAggregateFunctionRewriteRule, EqualToRewrite}
 import org.apache.spark.sql.catalyst.expressions._
-import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, CollectList, CollectSet}
+import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, ApproximatePercentile, CollectList, CollectSet}
 import org.apache.spark.sql.catalyst.optimizer.BuildSide
 import org.apache.spark.sql.catalyst.parser.ParserInterface
 import org.apache.spark.sql.catalyst.plans.JoinType
@@ -623,7 +623,8 @@ class CHSparkPlanExecApi extends SparkPlanExecApi {
   override def extraExpressionMappings: Seq[Sig] = {
     List(
       Sig[CollectList](ExpressionNames.COLLECT_LIST),
-      Sig[CollectSet](ExpressionNames.COLLECT_SET)
+      Sig[CollectSet](ExpressionNames.COLLECT_SET),
+      Sig[ApproximatePercentile](ExpressionNames.APPROX_PERCENTILE)
     ) ++
       SparkShimLoader.getSparkShims.bloomFilterExpressionMappings()
   }
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -22,7 +22,7 @@ import org.apache.gluten.datasource.ArrowConvertorRule
 import org.apache.gluten.exception.GlutenNotSupportException
 import org.apache.gluten.execution._
 import org.apache.gluten.expression._
-import org.apache.gluten.expression.aggregate.{HLLAdapter, VeloxBloomFilterAggregate, VeloxCollectList, VeloxCollectSet}
+import org.apache.gluten.expression.aggregate.{HLLAdapter, VeloxApproximatePercentile, VeloxBloomFilterAggregate, VeloxCollectList, VeloxCollectSet}
 import org.apache.gluten.extension._
 import org.apache.gluten.extension.columnar.FallbackTags
 import org.apache.gluten.extension.columnar.transition.Convention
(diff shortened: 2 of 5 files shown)


zhouyuan:
CC: @zhztheplayer

WangGuangxin:
I believe the change on Gluten side is ready now. But the `ApproxPercentileAggregate` implements in velox has to make some minor modification, becasue when add intermediate data, it checks the row vector encoding in `addIntermediateImpl`
https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/ApproxPercentileAggregate.cpp#L652 ,
I believe the encoding is guaranteed by `extractAccumulators` https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/aggregates/ApproxPercentileAggregate.cpp#L262.

But when it comes in Gluten, the output of PartialAgg are processed by Shuffle, which will flatten all vectors, so that the encoding is not guaranteed when added to FinalAgg. 
https://github.com/apache/incubator-gluten/blob/main/cpp/velox/shuffle/VeloxShuffleWriter.cc#L343

cc @zhztheplayer @liujiayi771 @ulysses-you

zhztheplayer:
Thank you in advance!

> I believe the change on Gluten side is ready now.

Did that mean this PR can be merged prior to merging Velox changes?

backends-velox/src/test/scala/io/glutenproject/execution/TestOperator.scala
```diff
@@ -1233,4 +1233,12 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
       checkOperatorMatch[HashAggregateExecTransformer]
     }
   }
+
+  test("Support ApproximatePercentile") {
+    runQueryAndCompare("""
+                         |SELECT approx_percentile(col, array(0.5, 0.4, 0.1), 100)
```
zhztheplayer, line 1239: I remember Spark has  `percentile_approx`. Is that one supported by this patch either?

zhztheplayer (dismissed):
Looking good if test passes. Thanks a lot for helping on this.

@PHILO-HE If you have further comments

liujiayi771:
@WangGuangxin Perhaps we could create an issue in the upstream Velox community to remove this verification.

zhztheplayer:
Hi @WangGuangxin so the current plan is to merge https://github.com/WangGuangxin/gluten/commit/97db869a56e8d42c0782e4e8218e1937e40682ce to upstream Velox before this ? Am I understanding correctly?

WangGuangxin:
> this

yes, we should merge to upsteam first before merge this PR

gluten-core/src/main/scala/io/glutenproject/extension/columnar/RewriteTypedImperativeAggregate.scala
```diff
@@ -40,6 +41,38 @@ object RewriteTypedImperativeAggregate extends Rule[SparkPlan] with PullOutProje
     }
   }
 
+  def shouldRewriteForPercentileLikeExpr(ae: AggregateExpression): Boolean = {
+    ae.aggregateFunction match {
+      case _: ApproximatePercentile =>
+        ae.mode match {
+          case Partial | PartialMerge => true
```
liujiayi771, line 48: It only support `Partial` and `Final` in `HashAggregateExecTransformer.scala`. Shall we add `PartialMerge` here?

WangGuangxin: yeah, it should be removed

liujiayi771:
@WangGuangxin Any progress of this PR?

gluten-core/src/main/scala/io/glutenproject/extension/columnar/RewriteTypedImperativeAggregate.scala
```diff
@@ -67,6 +100,28 @@ object RewriteTypedImperativeAggregate extends Rule[SparkPlan] with PullOutProje
...
+                ae =>
+                  attr.copy(dataType = getPercentileLikeInterminateDataType(ae.aggregateFunction))(
+                    exprId = attr.exprId,
+                    qualifier = attr.qualifier
+                  )
+              }
+              .getOrElse(attr)
+          case other => other
+        }
+        copyBaseAggregateExec(agg)(newResultExpressions = newResultExpressions)
```
zhztheplayer, line 123: IIUC this could make the offloaded partial aggregate operator incompatible with vanilla Spark's final aggregate operator? Say if final aggregation is fallen back then UB will be led.

I am considering whether we could using the approach like we had done for `approx_count_distinct`, to replace `approx_percentile` with something like `velox_approx_percentile` that has different intermediate type definition at the very early planning phase? Then we can distinguish between vanilla's and Velox's implementations for this function.

WangGuangxin: ok, I'll rebase and refact it

boneanxs:
@WangGuangxin Thanks for supporting this! Any new process abt this pr?

WangGuangxin:
> @WangGuangxin Thanks for supporting this! Any new process abt this pr?

@boneanxs Hi, I'll do rebase today since main branch changed a lot

PHILO-HE:
@WangGuangxin, could you let us know this work's state and the blocking issues? If you have no bandwidth, I can help cover it. Thanks!

WangGuangxin:
> @WangGuangxin, could you let us know this work's state and the blocking issues? If you have no bandwidth, I can help cover it. Thanks!

@PHILO-HE Sorry for this. Before this I have a PR to velox and I'll rebase it and send out today.

Yizhou-Yang:
@WangGuangxin 
hi, just to double-check why this pr is not merged into main branch, since I would like to support approx_percentile... 

Based on the results of my debug:
1)  The signature is different, which is why you need this rewrite rule to fit into gluten
Clipboard_Screenshot_1769764660
Clipboard_Screenshot_1769764751

2) Velox uses KLL algorithm, which is more space-efficient but not deterministic, but Spark uses GK algorithm, which is slower but deterministic. The results might differ if you choose to use the current velox implementation.

Is there any other differences which I have not yet noticed?

If my observation is correct, can I take over your existing work, make my changes to it, and implement a velox native SparkApproximatePercentileAggregate function in the velox project (based on spark's GK algorithm, and I will do the work to translate it into native), then adapt it to Gluten? I will add a gluten config to make sure that the user always have to option to choose between GK, VLL and fallback.

WangGuangxin:
> 1. rewrite

@Yizhou-Yang Feel free to take over this work. When I was working on this, besides changes from gluten side, we also need to make change to velox, similiar to this https://github.com/facebookincubator/velox/pull/9208/changes

Issue #5020, opened 2024-03-19
[VL] INSTALL xsimd: Permission   denied


xumingming:
### Problem description

Build command:

```
./dev/buildbundle-veloxbe.sh --enable_ep_cache=ON --build_type=RelWithDebInfo
```

Error message:

```
+ echo 'INSTALL xsimd.'
INSTALL xsimd.
+ '[' Darwin == Linux ']'
+ '[' Darwin == Darwin ']'
+ cmake --install xsimd-build/
-- Install configuration: "RelWithDebInfo"
-- Installing: /usr/local/include/xsimd
CMake Error at xsimd-build/cmake_install.cmake:41 (file):
  file INSTALL cannot make directory "/usr/local/include/xsimd": Permission
  denied.
```

### System information

OS: macOS

PR #5031, opened 2024-03-19
[GLUTEN-5020][VL] Add sudo for macOS related commands

xumingming:

Add sudo for macOS related commands, without it installing xsimd and gtest will fail.

Fixes: #5020

## How was this patch tested?

Manual tested on macOS.


PHILO-HE (approved):
Thanks for your fix!

Issue #5024, opened 2024-03-19
[VL] Enhance buildbundle-veloxbe.sh to have the ability to run only one specific step


xumingming:
### Problem description

buildbundle-veloxbe.sh is composed of at least two steps:

- Compile Velox
- Compile Gluten CPP code

Currently it is not smooth to run on some platforms such as MacOS, we might need to run this script multiple times to fix a build issue, e.g. `Compile Velox` succeed, but `Compile Gluten CPP code` failed, we'd better have the ability to only run `Compile Gluten CPP code` rather than run the whole script every time.

### System information

MacOS

PR #5032, opened 2024-03-19
[GLUTEN-5024][VL] Enhance buildbundle-veloxbe.sh to run single step

xumingming:
buildbundle-veloxbe.sh is composed of at least two steps:

- Get and build Velox
- Build Gluten CPP code

Currently it is not smooth to run on some platforms such as MacOS, we might need to run this script multiple times to fix a build issue, e.g. `Get and build Velox` succeed, but `Build Gluten CPP code` failed, in this PR we introduced the ability to specify the commands to run rather than run the whole script every time. By default we still run all the steps. Usage samples are the following:

```bash
-- Run all the steps
./dev/buildbundle-veloxbe.sh --build_type=RelWithDebInfo
-- Only run build_gluten_cpp
./dev/buildbundle-veloxbe.sh --build_type=RelWithDebInfo build_gluten_cpp
```

Fixes: #5024

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -173,21 +173,44 @@ if [ "$ENABLE_VCPKG" = "ON" ]; then
     eval "$envs"
 fi
 
-## build velox
 concat_velox_param
-cd $GLUTEN_DIR/ep/build-velox/src
-./get_velox.sh --enable_hdfs=$ENABLE_HDFS --build_protobuf=$BUILD_PROTOBUF --enable_s3=$ENABLE_S3 --enable_gcs=$ENABLE_GCS --enable_abfs=$ENABLE_ABFS $VELOX_PARAMETER
-# When BUILD_TESTS is on for gluten cpp, we need turn on VELOX_BUILD_TEST_UTILS via build_test_utils.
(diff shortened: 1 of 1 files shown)


FelixYBW:
Thank you. It's something we just planed. Since build_velox is always needed, can you add one flag like -Dfetch_velox=ON/OFF which can fetch velox or not before build? by default let's set it ON. We may also check if velox is fetched if not it always fetch.

Another issue is that currently script skip the velox build even if we modified some file in velox because it check the commit id to continue. So we need a git commit --amend in velox dir to force to build. Is there any better way to build velox automatically?

FelixYBW:
@zhztheplayer

zhztheplayer:
Thanks @xumingming for the enhancement.

> can you add one flag like -Dfetch_velox=ON/OFF which can fetch velox or not before build? by default let's set it ON.

Yes that would be one of the commen case. I assume there could probably be a `fetch_velox` before

>   build_velox
  build_gluten_cpp 

Though I am not sure if it's trivial to add. If this tricky we can apply this PR and do the further step by step.

FelixYBW:
>Another issue is that currently script skip the velox build even if we modified some file in velox because it check the commit id to continue. So we need a git commit --amend in velox dir to force to build. Is there any better way to build velox automatically?

Let's fix this issue at least. It's really not good experience to run git commit --amend each time

FelixYBW (approved):

PR #5033, opened 2024-03-19
[GLUTEN-4917][VL] static packaging refine

zhouyuan:

refine the static package in vcpkg
- bump boost to 1.84
- adding duckdb
- fix xsimd/simdjson version 
 
## How was this patch tested?

pass GHA static build test


ulysses-you:
@zhouyuan Does this pr ready ?

zhouyuan:
> @zhouyuan Does this pr ready ?

Hi @ulysses-you 

This patch adds the arrow build in vcpkg. it's not full ready for your case yet. The gap is:
- currently Velox will force build libarrow/libparquet, need to add one feature to allow velox to reuse the pre-built arrow
- I also find the compile flags for arrow are different in velox vs. vcpkg, may need some adjustments

the first patch may need some discussion with velox community. I'll try to do some tests in my own branch first.

Thanks,
-yuan

ulysses-you:
thank you @zhouyuan !

zhouyuan:
Velox is able to find the pre-built arrow with https://github.com/apache/incubator-gluten/pull/5033/commits/6b0fb02332cf3a882dd13a1f8c4a3cb6b351fb1c
gluten-cpp and the library loader needs to be modified for next

ulysses-you:
hi @zhouyuan  if there any blocker on this ?

zhouyuan:
> hi @zhouyuan if there any blocker on this ?

@ulysses-you 

hi Xiduo, we are actually breaking up this patch into smaller ones to get merged. still following on this.

The first half is merged already however we do find some unit tests failed due to the new compile method. `(free(): invalid pointer:)`
https://github.com/apache/incubator-gluten/pull/5515
We should fix the unit tests firstly, @JkSelf is working on this.

For static building with arrow and parquet(the second half of this patch) it's working only local test. However it requires to modify the library loading code path - which is actually conflicting with the original dynamic packaging method(as we still need to load libarrow/libparquet there). We may need to keep the both code path for a while as static packaging is not verified in production env. 

thanks
-yuan

PR #5036, opened 2024-03-20
[CORE] Catch more specific exceptions in Gluten plan validation

PHILO-HE:

This is a follow-up to use the newly introduced GlutenNotSupportException in plan validation. The `Exception` caught in GlutenPlan validation is too general, which makes some exceptions that require us to handle hidden in fall back behavior.

## How was this patch tested?

Existing tests.

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
@@ -83,6 +83,9 @@ object ExpressionConverter extends SQLConfHelper with Logging {
       udf: ScalaUDF,
       attributeSeq: Seq[Attribute],
       expressionsMap: Map[Class[_], String]): ExpressionTransformer = {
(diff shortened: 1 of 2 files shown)


gluten-core/src/main/scala/io/glutenproject/expression/ExpressionConverter.scala
```diff
@@ -83,6 +83,9 @@ object ExpressionConverter extends SQLConfHelper with Logging {
       udf: ScalaUDF,
       attributeSeq: Seq[Attribute],
       expressionsMap: Map[Class[_], String]): ExpressionTransformer = {
+    if (!udf.udfName.isDefined) {
+      throw new GlutenNotSupportException("UDF name is not found!")
```
PHILO-HE, line 87: This is to fix NoSuchElementException that can be thrown in some Spark UTs, where scala UDF is used with no name provided.

PHILO-HE:

rui-mo (approved):

Issue #4946, opened 2024-03-13
[CH] avg(bigint) return wrong result when overflow


loudongfeng:
### Backend

CH (ClickHouse)

### Bug description

```sql
create table big_int(id bigint) using parquet;
insert into big_int 
values 
(9223372036854775807),
(9223372036854775807),
(9223372036854775807),
(9223372036854775807);
select avg(id) from big_int
```
Spark result: 9.223372036854776E18
Gluten + CH result: -1

### Spark version

None

loudongfeng:
Average is calculated by sum/count.
Internally, spark use double for sum DataType ; while ClickHouse use bigint, which is just the input type.

PR #5048, opened 2024-03-20
[GLUTEN-4946][CH] Fix avg(bigint) overflow

loudongfeng:

cast avg input type to double when type is integer types: bigint/int/short...
Fixes: \#4946

## How was this patch tested?
 unit tests

### DIFF
--- a/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
+++ b/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
@@ -353,6 +353,8 @@ class GlutenConfig(conf: SQLConf) extends Logging {
   def enableColumnarProjectCollapse: Boolean = conf.getConf(ENABLE_COLUMNAR_PROJECT_COLLAPSE)
 
   def awsSdkLogLevel: String = conf.getConf(AWS_SDK_LOG_LEVEL)
+
+  def enableCastAvgAggregateFunction: Boolean = conf.getConf(COLUMNAR_NATIVE_CAST_AGGREGATE_ENABLED)
 }
 
 object GlutenConfig {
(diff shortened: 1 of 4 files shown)


loudongfeng:
CI failures unrelated. Ready for review. @zzcclp thanks

zzcclp (approved):
LGTM

binmahone (approved):

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
```diff
@@ -1691,4 +1693,10 @@ object GlutenConfig {
       .doc("Force fallback for orc char type scan.")
       .booleanConf
       .createWithDefault(true)
+
+  val COLUMNAR_NATIVE_CAST_AGGREGATE_ENABLED =
+    buildConf("spark.gluten.sql.columnar.cast.avg")
+      .internal()
+      .booleanConf
+      .createWithDefault(true)
```
rui-mo, line 1701: nit: do we need this config if the cast behavior is default on CH backend?

rui-mo (approved):

shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala
loudongfeng: We can disable this type cast when we are sure no overflow will happen, thereby improving the performance of some  cases.

zzcclp: @rui-mo agree with this? if yes, merge this pr ?

rui-mo: Yes, I understand. Thanks.

PR #5054, opened 2024-03-20
[VL] Fix bug where session config is lost when benchmark is enabled

FelixYBW:
It's a quick bug fix.

unordered_map::merge removes the duplicated items from source.

https://en.cppreference.com/w/cpp/container/unordered_map/merge


marin-ma (approved):
Thanks!

PR #5065, opened 2024-03-21
[DOC] Remove arrow version setting with 11.0.0-gluten

PHILO-HE:


PHILO-HE:
Assuming Gluten's dependency on 11.0.0-gluten was depreciated long time ago, this pr is simply removing it. @zhztheplayer, could you help review? Thanks!

Issue #5074, opened 2024-03-21
[VL] UDF load failed in yarn-cluster mode


kecookier:
### Backend

VL (Velox)

### Bug description

Based on my tests, some bugs have caused this issue:
1. Currently, `VeloxBackend::initUdf()` is called before `UDFResolver.loadAndGetFunctionDescriptions()`, so the `udflibPath` cannot be resolved on the driver.
2. When using `SparkFiles.get(f)` to retrieve a local file, it does not work on the driver in `yarn-client` or `yarn-cluster` mode. 
This is because the `--files/--archives` options set different Spark configuration keys, which are determined by the type of cluster manager and the deploy mode. If we use `--master=yarn`, then the Spark configuration keys `spark.yarn.dist.files`/`spark.yarn.dist.archives` will be set, and the file will be distributed to the working directory on all nodes(both driver and executors). Otherwise, `spark.files`/`spark.archives` will be set, and the files are added by `SparkContext.addFile`, which can be accessed by `SparkFiles.get()`.

The Spark code that processes the --files and --archives arguments can be found at: https://github.com/apache/spark/blob/25ecde948bebf01d2cb1e160516238e1d949ffdb/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L653

### Spark version

Spark-3.2.x

PR #5075, opened 2024-03-21
[GLUTEN-5074][VL] fix: UDF load error in yarn-cluster mode

kecookier:
1.  Remove `driverUdfLibPath` and retain only `udfLibPath`, as there is no need to differentiate between the driver and executor. The method used to access the file is determined by whether --master=yarn is specified.
> For more details on the reason, refer to Issue link #5074 

3. The UDF is loaded by `VeloxBackend::initUdf` in both the driver and the executor. `UdfResolver` does not load the UDF repeatedly; it only retrieves the function signatures.

(Fixes: \#5074)

## How was this patch tested?
I tested with and without the --files/--archives arguments in local, yarn-client, and yarn-cluster modes.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
@@ -132,8 +132,9 @@ object UDFResolver extends Logging {
       case None =>
     }
 
-    logInfo(s"after resolve path " +
-      s"is:${sparkConf.getOption(BackendSettings.GLUTEN_VELOX_UDF_LIB_PATHS)}")
+    logInfo(
+      s"after resolve path " +
+        s"is:${sparkConf.getOption(BackendSettings.GLUTEN_VELOX_UDF_LIB_PATHS)}")
   }
 
   // Try to unpack archive. Throws exception if failed.
--- a/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
@@ -37,7 +37,8 @@ import org.apache.spark.util.Utils
 import com.google.common.collect.Lists
 
 import java.io.File
-import java.nio.file.{Files, Paths}
+import java.net.URI
+import java.nio.file.{Files, FileVisitOption, Paths}
 
 import scala.collection.JavaConverters.asScalaIteratorConverter
 import scala.collection.mutable
@@ -107,7 +108,7 @@ object UDFResolver extends Logging {
 
   def getFilesWithExtension(directory: java.nio.file.Path, extension: String): Seq[String] = {
     Files
-      .walk(directory)
+      .walk(directory, FileVisitOption.FOLLOW_LINKS)
       .iterator()
       .asScala
       .filter(p => Files.isRegularFile(p) && p.toString.endsWith(extension))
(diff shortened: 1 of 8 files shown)


backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
```diff
@@ -152,34 +143,39 @@ object UDFResolver extends Logging {
...
       .map {
         f =>
-          val file = new File(f)
-          // Relative paths should be uploaded via --files or --archives
-          // Use SparkFiles.get to download and unpack
-          if (!file.isAbsolute) {
-            if (!canAccessSparkFiles) {
-              throw new IllegalArgumentException(
-                "On yarn-client mode, driver only accepts absolute paths, but got " + f)
+          val uri = Utils.resolveURI(f)
```
marin-ma, line 152: Looks like `f` can also be a relative filename or file tag here without a scheme. Would this logic be able to handle such case?

e.g.
```
```

backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
```diff
@@ -152,34 +143,39 @@ object UDFResolver extends Logging {
...
+          var file: File = null
```
marin-ma, line 153: Let's avoid using `var` here.

backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
kecookier: yes，if f is `libmyudf.so`, Utils.resolveURI() will return `file://${PWD}/libmyudf.so`

backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
kecookier: ok, I'll fix it.

backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
marin-ma: But `file://${PWD}/libmyudf.so` is not the expected path, right? It should be `/path/to/gluten/cpp/build/velox/udf/examples/libmyudf.so`

kecookier: The call to Utils.resolveURI() is used to determine whether the file is local or remote.

In your example, libmyudf.so is a local file, and we will not use uri.path directly.

The --files argument copies this file to a different destination directory. When --master=yarn is specified, the file is copied to the working directory on all nodes (both the driver and executors). In local mode, the files are added using SparkContext.addFile, and they can then be accessed using SparkFiles.get(f).

marin-ma: > In your example, libmyudf.so is a local file

That doesn't make sense to me. `libmyudf.so` here should refer to the file uploaded by `--files`. But here it's resolved as a relative path to the runtime directory.

> The call to Utils.resolveURI() is used to determine whether the file is local or remote.

It would be better to check whether it's a relative path first. Although it's resolved as a local file and pass the `if` condition, the path doesn't even exist.

> When --master=yarn is specified, the file is copied to the working directory on all nodes (both the driver and executors).

Based on my previous experience, for yarn client mode, the files will be copied to all executor container + AM container, so they won't be copied to the driver node. In this case, if we only use the two configurations below
```
```
the driver will fail to get the actual path for `libmyudf.so`, and that's the reason for adding  `spark.gluten.sql.columnar.backend.velox.driver.udfLibraryPaths`

kecookier: > It would be better to check whether it's a relative path first. Although it's resolved as a local file and pass the if condition, the path doesn't even exist.

kecookier: Thank you for your patient explanation, Rong! 

I previously thought that under yarn-client mode, `spark.yarn.dist.files` would also copy the files to the driver. I got the AM and driver mixed up.  

Let's fix this issue.

marin-ma: I would suggest preserving 2 kinds of configuration for `spark.gluten.sql.columnar.backend.velox.udfLibraryPaths`

- relative path
- URI

And make `spark.gluten.sql.columnar.backend.velox.udfLibraryPaths=/path/to/..." invalid.

kecookier: @marin-ma  I have already fix the code according to the previous discussion in the comments, could you help review it again?

backends-velox/src/test/scala/io/glutenproject/expression/VeloxUdfSuite.scala
```diff
@@ -82,7 +82,8 @@ class VeloxUdfSuiteLocal extends VeloxUdfSuite {
   override val master: String = "local[2]"
   override protected def sparkConf: SparkConf = {
     super.sparkConf
-      .set("spark.gluten.sql.columnar.backend.velox.udfLibraryPaths", udfLibPath)
+      .set("spark.files", udfLibPath)
+      .set("spark.gluten.sql.columnar.backend.velox.udfLibraryPaths", "libmyudf.so")
```
marin-ma, line 86: Please do not hard code "libmyudf.so" here. Use a helper function to extract the filename from `udfLibPath`. Note the `udfLibPath` can also be a comma-separated string with multiple paths.

marin-ma:
Could you also update the examples in the document with either relative or URI path?

marin-ma (approved):
LGTM. Thanks!

Issue #5085, opened 2024-03-22
[VL] buildbundle-veloxbe.sh fails on macOS


xumingming:
### Problem description

Build command:

```bash
./dev/buildbundle-veloxbe.sh --build_type=RelWithDebInfo
```

Error message:

```bash
+ sed -i '/facebook\/fbthrift/{n;s/cmake_install -DBUILD_TESTS=OFF/cmake_install -DBUILD_TESTS=OFF -DBUILD_SHARED_LIBS=OFF/;}' scripts/setup-macos.sh
sed: 1: "scripts/setup-macos.sh": unterminated substitute in regular expression
```

### System information

N/A

PR #5086, opened 2024-03-22
[GLUTEN-5085] [VL] Fix get_velox.sh on macOS

xumingming:

sed's `-i` param is different on macOS than on Linux, we need to use `-i ''` rather than `-i` to make it work.

Fixes: #5085

## How was this patch tested?

Manual tested on macOS


PHILO-HE (approved):
Thanks!

PR #5087, opened 2024-03-22
[VL] Remove unused openssl setup script change for centos8

yma11:

function `install_openssl` is not actually used as `openssl-devel` is already installed in previous step in the script.

## How was this patch tested?

CI


PHILO-HE (approved):
Looks good! Thanks!

Issue #4039, opened 2023-12-14
[VL] Spark function support list [please leave comment/mark if you plan to implement]


PHILO-HE:
### Description

Here listed some functions we need to support in Gluten and Velox, will keep updated in the future. You may reply in the thread if you'd like to work on anyone.

- [x] size https://github.com/facebookincubator/velox/pull/5461 fix return type
- [ ] percentile_approx
- [ ] explode
- [ ] concat_ws
- [x] collect_list (velox supported, needs Gluten to enable array for project plan node)
- [ ] date_sub (one velox upstream PR near to merge)
- [ ] date_format: can map to format_datatime https://github.com/facebookincubator/velox/issues/5734
- [x] regexp_replace (supported, only few incompatible issue from RE2 used by velox)
- [x] collect_set
- [x] current_timestamp (foldable expression, optimized by spark, no need to offload)
- [ ] unix_timestamp: "Only supports string type, with session timezone considered, todo: support date type"
- [ ] to_json
- [x] str_to_map https://github.com/facebookincubator/velox/pull/5842
- [ ] json_tuple
- [ ] to_date
- [x] rank
- [ ] hour
- [ ] locate
- [ ] mod
- [ ] conv: "PR ready“
- [ ] parse_url
- [ ] pow
- [ ] ifnull
- [ ] translate (upstream velox PR merged, Gluten PR pending for future velox rebase)
- [ ] urldecoder: "UDF, supported by spark as a built-in funciton since 3.4.0."
- [ ] distinct
- [ ] normalizenanandzero
- [ ] isarrayintersects
- [ ] default.json_split (udf, no need to impl.): "external UDF"
- [ ] parsejsonarray: "external UDF"
- [x] <=> (EqualNullSafe) https://github.com/oap-project/velox/pull/357 https://github.com/oap-project/gluten/pull/2259
- [ ] struct
- [ ] rand (supported except seed is provided)
- [x] now (foldable expression, optimized by spark, no need to offload)
- [ ] lead: "PR ready"
- [ ] add_months
- [ ] percentile
- [ ] last_day
- [ ] lag: "PR ready"
- [ ] minute
- [ ] year
- [ ] month
- [ ] weekofyear
- [ ] next_day
- [ ] posexplode
- [ ] trunc
- [ ] months_between
- [ ] array_min
- [ ] dense_rank
- [x] current_date (foldable expression, optimized by spark, no need to offload)
- [ ] day
[...]

PR #5093, opened 2024-03-24
[VL] Support map_concat spark function

Surbhi-Vijay:
Adds support for `map_concat` spark function as part of #4039 

Added supporting test case and updated documentation


rui-mo (dismissed):
Thanks. Could be merged after CI passes.

Surbhi-Vijay:
`df.collect()` works as expected for `map_concat` function. (converts to velox plan)

`df.show()` is still executing in Spark due to the fallback. The reason of fallback is, Spark adds limit of 20 for `df.show` and tries to cast all the columns to string data type. 
(Code snippet from Dataset.scala)
```
private[sql] def getRows(
      numRows: Int,
      truncate: Int): Seq[Seq[String]] = {
    val newDf = toDF()
    val castCols = newDf.logicalPlan.output.map { col =>
      // Since binary types in top-level schema fields have a specific format to print,
      // so we do not cast them to strings here.
      if (col.dataType == BinaryType) {
        Column(col)
      } else {
        Column(col).cast(StringType)
      }
    }
```

The CAST function is then transformed into Gluten to CASTTransformer but fails in validation since the cast is not supported for map type in Gluten yet.
(Code snippet from SubstraitToVeloxPlanValidator.cc)
```
switch (input->type()->kind()) {
    case TypeKind::ARRAY:
    case TypeKind::MAP:
    case TypeKind::ROW:
    case TypeKind::VARBINARY:
      LOG_VALIDATION_MSG("Invalid input type in casting: ARRAY/MAP/ROW/VARBINARY.");
      return false;
    case TypeKind::TIMESTAMP: {
      LOG_VALIDATION_MSG("Casting from TIMESTAMP is not supported or has incorrect result.");
      return false;
    }
    default: {
    }
  }
```

Surbhi-Vijay:
@rui-mo can you please check this patch again. I have added few comments and rewritten the spark test case since there is no exception thrown for duplicate values. Velox already handles it by keeping last value.

rui-mo:
Hi @Surbhi-Vijay, I notice `MapKeyDedupPolicy.EXCEPTION` seems to be the default behavior of Spark, is that right? In your opinion, do we need to customize the map_concact function in Velox to allow configurable behavior?

Surbhi-Vijay:
> Hi @Surbhi-Vijay, I notice `MapKeyDedupPolicy.EXCEPTION` seems to be the default behavior of Spark, is that right? In your opinion, do we need to customize the map_concact function in Velox to allow configurable behavior?

Yes, `MapKeyDedupPolicy.EXCEPTION` is the default behavior in spark. Exception will have to thrown from Velox if the behavior should be same as Spark. The velox implementation is based on Presto where LAST_WIN is the default behavior.

Maybe we can introduce change in Velox to throw exception based on a config which will only be used in Gluten.

Please let me know your thoughts.

Surbhi-Vijay:
> > Hi @Surbhi-Vijay, I notice `MapKeyDedupPolicy.EXCEPTION` seems to be the default behavior of Spark, is that right? In your opinion, do we need to customize the map_concact function in Velox to allow configurable behavior?
> 
> Yes, `MapKeyDedupPolicy.EXCEPTION` is the default behavior in spark. Exception will have to thrown from Velox if the behavior should be same as Spark. The velox implementation is based on Presto where LAST_WIN is the default behavior.
> 
> Maybe we can introduce change in Velox to throw exception based on a config which will only be used in Gluten.
> 
> Please let me know your thoughts.

@rui-mo @PHILO-HE can you please check this comment? and we can finalize the next set of action for this patch.

PHILO-HE:
> Maybe we can introduce change in Velox to throw exception based on a config which will only be used in Gluten.
Please let me know your thoughts.

@Surbhi-Vijay, yes, better to introduce a config in Velox to support those two behaviors. I guess Spark's default EXCEPTION behavior is commonly used, so please firstly fix it in Velox. cc @rui-mo

Surbhi-Vijay:
I will be updating these PRs in few days.

Issue #4988, opened 2024-03-18
UDF with one more return type can't be registered to gluten


liujp:
### Backend

VL (Velox)

### Bug description

In UDF register to gluten, using unordered_map{udf_name, udf_return_type} to register the udf to gluten, as follows:
<img width="937" alt="image" src="https://github.com/apache/incubator-gluten/assets/11934146/3852f557-d0b5-4ed0-9c64-bf1c39e1086c">
so, in this method, udf registered to gluten can only using one return type. but if I have a UDF that int myudf(int, int) and string myudf(string, int).if so, I can't register one udf with one more return type

zhouyuan:
CC @marin-ma

liujp:
done

liujp:
done

PR #5104, opened 2024-03-25
[VL] Supports register udf with different signatures

marin-ma:
issue #4988

Native side (UDF defination and registration) 

**Before**: Only support specifying function name and return type in `UdfEntry`. **After**: Add 2 fields `numArgs` and `argTypes` to `UdfEntry`. Serialize both return type and arg types during registration from native to JVM.

Java side (Resolving UDF expression) 

**Before**: Create the `UDFExpression` based on the function name with the only supported return type. **After**: Get the registered return type based on function name and argument types. Create the `UDFExpression` based on the function name and registered return type.


liujp:
so, can this use the variable-length arguments in UDF？

marin-ma:
> so, can this use the variable-length arguments in UDF？

@liujp I'm afraid not. Will try to resolve it in another patch.

marin-ma:
@kecookier @zhouyuan Could you please help to review? Thanks!

marin-ma:
@PHILO-HE Could you help to review? Thanks!

PHILO-HE (approved):
Thanks for your great work!

Issue #4263, opened 2024-01-03
[VL] Compression type 2 not supported


rhh777:
### Backend

VL (Velox)

### Bug description

When running TPC DS query67, a large number of tasks failed. The reasons for failure are Compression type 2 not supported and OutOfMemoryException.

What compression type is type 2? 

I tried two ways to run successfully。

1. Increase offHeap memory(15g->20g) and the task can run successfully. 

2. Modify spark.io.compression.codec from the default lz4 to zstd.  Then modify spark.gluten.sql.columnar.backend.velox.orderBySpillMemoryThreshold from the default value 0 to 256MB. After modifying these two configurations, the task can run successfully and the speed is improved.

### Spark version

Spark-3.3.x

### Spark configurations

spark-sql \
--master yarn \
--deploy-mode client \
--driver-cores 5 \
--driver-memory 5G \
--executor-cores 5 \
--executor-memory 4G \
--num-executors 10 \
--conf spark.executor.memoryOverhead=1g \
--conf spark.plugins=io.glutenproject.GlutenPlugin \
--conf spark.shuffle.manager=org.apache.spark.shuffle.sort.ColumnarShuffleManager \
--conf spark.memory.offHeap.enabled=true \
--conf spark.memory.offHeap.size=15G \
--class org.apache.kyuubi.tpcds.benchmark.RunBenchmark \
--conf spark.executorEnv.JAVA_HOME=/usr/local/jdk1.8.0_382 \
--conf spark.yarn.appMasterEnv.JAVA_HOME=/usr/local/jdk1.8.0_382 \
--files /data/hrh/tpcds/hdfs-client.xml \
--conf spark.sql.autoBroadcastJoinThreshold=10MB

### System information

```
Caused by: java.lang.RuntimeException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Operator::addInput failed for [operator: Window, plan node ID: 2]: Error during calling Java code from native code: io.glutenproject.memory.memtarget.ThrowOnOomMemoryTarget$OutOfMemoryException: Not enough spark off-heap execution memory. Acquired: 8388608, granted: 0. Try tweaking config option spark.memory.offHeap.size to get larger space to run this application. 
Current config settings: 
	spark.gluten.memory.offHeap.size.in.bytes=15.0 GiB
	spark.gluten.memory.task.offHeap.size.in.bytes=3.0 GiB
[...]

PR #5121, opened 2024-03-26
[GLUTEN-4263][VL] Fix compression type 2 not supported in static build

zhztheplayer:
This fixes error "Compression type 2 not supported" when spilling kicks in, in static build mode.

https://github.com/apache/incubator-gluten/issues/4263


FelixYBW:
Thank you for the fix

FelixYBW (approved):

Issue #5123, opened 2024-03-26
Build bundle package in Github actions failed with centos:8


dcoliversun:
### Problem description

https://github.com/dcoliversun/gluten/actions/runs/8431129692/job/23089193957

With centos:8, `Build bundle package` fails with error  
```plain
apt-get update && \
  apt-get install -y openjdk-8-jdk maven && \
  apt remove openjdk-11* -y
  shell: sh -e {0}
/__w/_temp/c5ea88ee-41ec-4[3](https://github.com/dcoliversun/gluten/actions/runs/8431129692/job/23089193957#step:5:3)35-b73[4](https://github.com/dcoliversun/gluten/actions/runs/8431129692/job/23089193957#step:5:4)-c[7](https://github.com/dcoliversun/gluten/actions/runs/8431129692/job/23089193957#step:5:8)0b17e7d818.sh: line 1: apt-get: command not found
Error: Process completed with exit code 127.
```

### System information

no need

### CMake log

```bash
no need
```

dcoliversun:
@wangyum could you please have a look :)

Issue #4942, opened 2024-03-13
[VL] remove dynamic packaging and use static packaing


zhouyuan:
### Description

The dynamic packaging will make a jar with all necessary dependencies for shared libs. On gluten start it will try to extract the libs and then load into JVM.
This method is introduced in early stage of Gluten. The production env is quite different: Cent7, Cent8, Ubuntu 20, Debian, ArchLinux and some customized Linux releases. For different OS the dependencies are different, especially on the library path and the library versions. The code is quite difficult to maintain as Velox is bumping the dependencies.

For better packaging releasing/maintaining, gluten may should pick the static packaging(vcpkg based) so all dependencies are all statically linked and can work on all OS.
  
CC: @weiting-chen 

Thanks,
-yuan

FelixYBW:
We may mention this in dev maillist.

static build makes the deployment much easier. The only issue may be the jar size but now even with debug info, it's just ~88MB. 

For developer with current build script, all static library will only be built once. Or build again after vcpkg script changed. velox will be built when commit id changed. So not an issue to development either. It may increase the link time somehow.

See if anyone have concerns about this.

boneanxs:
> We may mention this in dev maillist.

Hey @zhouyuan @FelixYBW Since we're in Apache incubator community now, if we decide to use dev maillist, better update https://github.com/apache/incubator-gluten/blob/62746c40e6e725f1c20ad06251ef976350d2bfc3/README.md?plain=1#L113 to reflect email address whereas others can join it.

zhouyuan:
> > We may mention this in dev maillist.
> 
> Hey @zhouyuan @FelixYBW Since we're in Apache incubator community now, if we decide to use dev maillist, better update
> 
> https://github.com/apache/incubator-gluten/blob/62746c40e6e725f1c20ad06251ef976350d2bfc3/README.md?plain=1#L113
> 
> to reflect email address whereas others can join it.

Hi Rex, yes, will add the maillist to the doc, thanks!

Thanks,
-yuan

PHILO-HE:
FYI.
Just tried using static link on centos-8, then running test on centos-7. An issue was reported at runtime:
```
java.lang.UnsatisfiedLinkError: 
[...]

PR #5124, opened 2024-03-26
[GLUTEN-5123][INFRA]set up java and maven according to os in build_bundle_package.yml

dcoliversun:

Set up java and maven according to os in build_bundle_package.yml

(Fixes: \#5123)

## How was this patch tested?

Pass the build bundle package in github action.

Manual test:  
* centos7: https://github.com/dcoliversun/gluten/actions/runs/8436722430
* ubuntu: https://github.com/dcoliversun/gluten/actions/runs/8438294441
* centos8: https://github.com/dcoliversun/gluten/actions/runs/8439025928

### DIFF
--- a/.github/workflows/build_bundle_package.yml
+++ b/.github/workflows/build_bundle_package.yml
@@ -70,11 +70,18 @@ jobs:
         with:
           name: velox-native-lib-${{github.sha}}
           path: ./cpp/build/releases
-      - name: Setup java and maven
+      - name: Setup java and maven for Ubuntu
+        if: startsWith(github.event.inputs.os, 'ubuntu')
         run: |
           apt-get update && \
           apt-get install -y openjdk-8-jdk maven && \
           apt remove openjdk-11* -y
+      - name: Setup java and maven for CentOS
+        if: startsWith(github.event.inputs.os, 'centos')
+        run: |
+          yum update -y && \
+          yum install -y java-1.8.0-openjdk-devel maven && \
+          yum remove java-11* -y
       - name: Build for Spark ${{ github.event.inputs.spark }}
         run: |
           cd $GITHUB_WORKSPACE/ && \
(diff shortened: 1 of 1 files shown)


dcoliversun:
@zhouyuan @wangyum please review this PR if have time

wangyum (approved):
LGTM

zhouyuan:
Hi @dcoliversun 
This patch seems trying to generate package for each OS, the package built from centos7 should be to work on other platforms as it's using static packaging via vcpkg. 
Can you please to check the centos7 package can work on your case?

thanks,
-yuan

zhouyuan:
The feature itself looks good to me
CC @PHILO-HE

dcoliversun:
@zhouyuan I'm ok to check the centos7 package. I want to make sure I need to test with bundle jar in https://github.com/dcoliversun/gluten/actions/runs/8436722430

zhouyuan (approved):
👍

zhouyuan:
> @zhouyuan I'm ok to check the centos7 package. I want to make sure I need to test with bundle jar in https://github.com/dcoliversun/gluten/actions/runs/8436722430

Hi @dcoliversun 
Yes, the static package on centos7 should be work on other platforms. There maybe some requirements for glibc versions, but it should be OK for most of the recent releases.


Initially gluten tried to do a package for each platform, with all the dependencies packed into a third-party jar. This is not required in static packaging. You may also noticed in the recent releases there's only one jar file for specific Spark version(which can work on all platformts). In fact we are in the progress to migrate to static packaging. Here's the detail issue 
https://github.com/apache/incubator-gluten/issues/4942

Thanks,
-yuan

dcoliversun:
@zhouyuan Thanks and I think it's very useful for us. In our production environment, we used the `bundle` and `third-lib` jar. I will follow up this feature. If I need to do anything, I will also do it

PR #5128, opened 2024-03-26
[CORE] Move memory off-heap conf checks to driver plugin

wForget:

Move memory off-heap conf checks to driver plugin to fail quickly when misconfigured


zhztheplayer (approved):
Thanks!

PR #5129, opened 2024-03-26
[CORE] Move BackendBuildInfo case class from GlutenPlugin to Backend class file

wForget:

The `BackendBuildInfo` case class seemed more appropriate in `Backend` class file, so I moved it.


zhztheplayer (approved):
Thanks!

Issue #5110, opened 2024-03-25
Register  a aggregate UDF but an error occurs


liujp:
### Backend

VL (Velox)

### Bug description

[Expected behavior] and [actual behavior].

### Spark version

None

liujp:
@marin-ma

marin-ma:
@liujp UDAF is not supported yet.

liujp:
> @liujp UDAF is not supported yet.
👌,

liujp:
@marin-ma thx~

PR #5130, opened 2024-03-26
[VL] Support native UDAF

marin-ma:
Add `UserDefinedAggregateFunction` as wrapper for all native udaf.

User should define and register UDAFs in the same way as UDFs. Gluten framework will handle the differences.

issue: #5110

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
@@ -27,10 +27,13 @@ import org.apache.gluten.vectorized.JniWorkspace
 import org.apache.spark.{SparkConf, SparkContext, SparkEnv, SparkFiles}
 import org.apache.spark.deploy.SparkHadoopUtil
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.catalyst.FunctionIdentifier
+import org.apache.spark.sql.catalyst.{FunctionIdentifier, InternalRow}
 import org.apache.spark.sql.catalyst.analysis.FunctionRegistry.FunctionBuilder
-import org.apache.spark.sql.catalyst.expressions.{Expression, ExpressionInfo}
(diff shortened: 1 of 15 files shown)


backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
```diff
@@ -78,9 +106,12 @@ case class UDFExpression(
 
 object UDFResolver extends Logging {
   private val UDFNames = mutable.HashSet[String]()
-
   private val UDFMap = mutable.HashMap[(String, Seq[DataType]), ExpressionType]()
 
+  private val UDAFNames = mutable.HashSet[String]()
+  private val UDAFMap =
+    mutable.HashMap[(String, Seq[DataType]), (ExpressionType, Seq[AttributeReference])]()
```
PHILO-HE, line 113: Better to add comments for this map & the above UDFMap.

cpp/velox/udf/UdfLoader.cc
```diff
@@ -77,11 +78,34 @@ std::unordered_set<std::shared_ptr<UdfLoader::UdfSignature>> UdfLoader::getRegis
...
+    return {};
+  }
+  if (!names_.empty()) {
+    return names_;
+  }
+  if (signatures_.empty()) {
+    getRegisteredUdfSignatures();
+  }
+  for (const auto& sig : signatures_) {
+    if (!sig->intermediateType.empty()) {
```
PHILO-HE, line 104: Maybe, better to not mix UDF/UDAF signatures together?

PHILO-HE (approved):
Looks good! Just some suggestions. Please check whether it makes sense (I'm OK to merge this first code version). Thanks!

cpp/velox/udf/UdfLoader.cc
marin-ma: Thanks. I've also considered this design. Will address that.

Issue #5136, opened 2024-03-27
[VL] Vanilla Spark broadcast exchange + R2C is slow sometimes


zhztheplayer:
### Backend

VL (Velox)

### Bug description

This is because the code to convert vanilla Spark's hashed relation to Gluten's sometimes produced duplicated rows.

The fix will be incorporated in https://github.com/apache/incubator-gluten/pull/5058 (at commit https://github.com/apache/incubator-gluten/pull/5058/commits/96c3fc7b282370e49012bcfff17ab243f2bdc33f) since it can be tested by the ACBO changes.

ulysses-you:
Thank you @zhztheplayer It's a good point, columnar broadcast would broadcast the origin binary data but vanilla Spark would broadcast hash relation. So I think this issue is a common case even if there is no r2c.

Is it possbile to create a new pr for this issue ?

zhztheplayer:
I don't have dedicated UTs for it so it was incorporated into the other PR.

Still I can open one for it if you think it's needed: https://github.com/apache/incubator-gluten/pull/5141.

The change was already tested so I will proceed to merge after code style check is passed if it's OK to you.

zhztheplayer:
The major issue I have found is that the `flatMap` approach would cause `UnsafeHashedRelation` to produce duplicated rows in my case (TPCDS q14a with current version of ACBO)
While the `map` approach would cause `LongHashedRelation` to loss rows (TPCDS q2).

The following fix (the same with #5141) can work but I didn't dive into it deeply to find the root reason of the inconsistency (maybe related to `keyIsUnique`? I am not sure).

```scala
  private def reconstructRows(relation: HashedRelation): Iterator[InternalRow] = {
    // It seems that LongHashedRelation and UnsafeHashedRelation don't follow the same
    //  criteria while getting values from them.
    // Should review the internals of this part of code.
    relation match {
      case relation: LongHashedRelation if relation.keyIsUnique =>
        relation.keys().map(k => relation.getValue(k))
      case relation: LongHashedRelation if !relation.keyIsUnique =>
        relation.keys().flatMap(k => relation.get(k))
      case other => other.valuesWithKeyIndex().map(_.getValue)
    }
  }
```

zhztheplayer:
Fixed in https://github.com/apache/incubator-gluten/pull/5141. I assume we can close this now.

PR #5058, opened 2024-03-21
[CORE] Basic runnable version of ACBO (Advanced CBO)

zhztheplayer:
See proposal https://github.com/apache/incubator-gluten/issues/5057

This is the first runnable version of ACBO with TPCH SF 1.0 and TPCDS 10.0 passed.

After this patch, one could set `spark.gluten.sql.advanced.cbo.enabled=true` to enable ACBO. It's by default disabled.

Issues: 

1. It's now only replacing `TransformPreOverrides()` with a rough cost model to do fallback;
2. It's not tested with CH backend yet. @baibaichen @zzcclp If you would like to evaluate it for CH; Otherwise I could disable it for CH in next patch;
3. It may generates slow plan since some operators like aggregation are not considered in ACBO. The side-effect of this will be amplified in performance test;
4. The first version would only be **runnable**. Further integration works are required to make it ready for production.

The following improvements are on the way:

1. Enable group mutation (for ops like AQEShuffleReadExec which could propagate children's convention to parent);
2. Enable pattern based rule-matching;

The required facilities of the above were already added but not enabled yet. Will enable and test them in PRs respectively.

PR #5141, opened 2024-03-27
[GLUTEN-5136][VL] Duplicated output from Spark-to-Velox broadcast relation conversion

zhztheplayer:
https://github.com/apache/incubator-gluten/issues/5136

Will be tested by https://github.com/apache/incubator-gluten/pull/5058. It's probably not a trivial code path and may delay adding dedicated UTs.


zhztheplayer:
cc @ulysses-you

ulysses-you:
Thank you @zhztheplayer for the quick fix. After this pr if there is no c2r, the duplicate keys issue is still existed right ?

zhztheplayer:
> Thank you @zhztheplayer for the quick fix. After this pr if there is no c2r, the duplicate keys issue is still existed right ?

After the fix is applied we should no longer have any relevant issues on BHJ unless unknown. 

The issue this PR tried to fix only happened when broadcast exchange is fallen back but bhj is not. Which is a corner case for current Gluten, usually they are both fallen back or both not. Thus ideally we shouldn't have this issue in usual bhj processing.

ulysses-you:
I see, thank you for the explaination!

PR #5144, opened 2024-03-27
[Core] minor change for eliminateProjectList

zhli1142015:

`eliminateProjectList` removed the project function, like below:

`HashAggregate(keys=[knownfloatingpointnormalized(normalizenanandzero(f#38)) AS f#38], functions=[partial_count(1)], output=[f#38, count#46L])`
-->
```
+- ^(5) FlushableHashAggregateTransformer(keys=[f#38], functions=[partial_count(1)], output=[f#38, count#46L])
                                  +- ^(5) ProjectExecTransformer [f#38]
```
`knownfloatingpointnormalized(normalizenanandzero` are removed after the transfer. This is found when doing the change https://github.com/facebookincubator/velox/pull/9272, it can be reproduced by the query:
`Seq(0.0f, -0.0f, 0.0f/0.0f, Float.NaN).toDF("f").groupBy("f").count().collect`

### DIFF
--- a/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
+++ b/gluten-core/src/main/scala/io/glutenproject/extension/columnar/PullOutPreProject.scala
@@ -121,7 +121,7 @@ object PullOutPreProject extends Rule[SparkPlan] with PullOutProjectHelper {
       // post-projection, the additional columns need to be removed, leaving only the original
       // output of the child.
       val preProject = ProjectExec(
-        eliminateProjectList(sort.child.outputSet, expressionMap.values.toSeq),
+        eliminateProjectList(expressionMap.values.toSeq, sort.child.outputSet),
         sort.child)
       val newSort = sort.copy(sortOrder = newSortOrder, child = preProject)
       // The pre-project and post-project of SortExec always appear together, so it's more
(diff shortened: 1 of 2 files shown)


zhli1142015:
cc @liujiayi771 , thanks.

liujiayi771:
@zhli1142015 Thanks for your fix. We should always keep the contents of `expressionMap.values` and append the `child.outputSet`. But the word 'append' denotes additional content, so it seems a bit odd to consider it as content to be retained in your code. And I think the 'append' should be the last parameter of the method.

Perhaps change the function to the below code, using the set difference method of the `AttributeSet` seems to be more concise, and swap the order of all the arguments that call this function.
```scala
protected def eliminateProjectList(
    projectAttributes: Seq[NamedExpression],
    childOutput: AttributeSet): Seq[NamedExpression] = {
  projectAttributes ++ (childOutput -- projectAttributes).toIndexedSeq
    .sortWith(_.exprId.id < _.exprId.id)
}
```

zhli1142015:
> ```scala
> sortWith
> ```

Thanks, I'm ok with the signature change, but should the logic like below?
```
(projectAttributes ++ (childOutput -- projectAttributes).toIndexedSeq)
    .sortWith(_.exprId.id < _.exprId.id)
```

liujiayi771:
@zhli1142015 Yes, I missed a pair of parentheses.

liujiayi771 (approved):

zhli1142015:
Run Gluten Clickhouse CI

PR #5145, opened 2024-03-27
[VL] CI: Enable GHA dependency cache on static Velox build

zhztheplayer:
To speed up CI static build if velox and vcpcg's code not gets changed.

Ref: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

Dynamic build is not impacted by this patch so the overall Velox CI duration would not change.

Some links to inspect the cache:

https://api.github.com/repos/apache/incubator-gluten/actions/caches
https://api.github.com/repos/apache/incubator-gluten/actions/cache/usage

### DIFF
--- a/.github/workflows/velox_docker.yml
+++ b/.github/workflows/velox_docker.yml
@@ -52,7 +52,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Generate cache key
         run: |
-          echo ${{ hashFiles('./ep/build-velox/src/**', './dev/vcpkg/**') }} > cache-key
+          echo ${{ hashFiles('./ep/build-velox/src/**', './dev/**', './cpp/*') }} > cache-key
       - name: Cache
         id: cache
         uses: actions/cache@v3
--- a/.github/workflows/velox_docker.yml
+++ b/.github/workflows/velox_docker.yml
@@ -60,7 +60,7 @@ jobs:
           path: ./cpp/build/releases/
           key: cache-velox-build-${{ hashFiles('./cache-key') }}
       - name: Build Gluten velox third party
-        if: ${{ steps.cache.cache-hit != 'true' }}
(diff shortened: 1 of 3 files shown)


zhztheplayer:
@zhouyuan

zhztheplayer:
I can think of a problem that when we update the Velox branch for some reason without changing Gluten's code, the cache should be manually invalidated otherwise will still be restored.

Delete cache:
https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-github-actions-caches-for-a-repository-using-a-cache-key

But the way doesn't seem to be so friendly to developer thus we may better to create a new Velox branch if appending changes.

zhouyuan:
> I can think of a problem that when we update the Velox branch for some reason without changing Gluten's code, the cache should be manually invalidated otherwise will still be restored.
> 
> Delete cache: https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-github-actions-caches-for-a-repository-using-a-cache-key
> 
> But the way doesn't seem to be so friendly to developer thus we may better to create a new Velox branch if appending changes.

Yes, this will happen when we do a rebase in velox, then find some Spark UT failed in gluten, then did some fixes.
can we filter on the pull request title, like if there's a key word(like forcebuildvelox)?

zhztheplayer:
@zhouyuan A way to trigger rebuilding is needed anyway. I will raise a new PR for that. Thanks for the suggestion.

zhouyuan (approved):

zhztheplayer:
Update: This cache doesn't seem to be able to share among PRs. https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache.

It's OK for now and may need further improvements though.

PR #5147, opened 2024-03-27
[VL][MINOR] Don't link libjemalloc_extension.a of DuckDB if BUILD_JEMALLOC_EXTENSION is OFF

wankunde:

Minor change

## How was this patch tested?

Fix compile issue


PHILO-HE:
Maybe, we can directly remove linking this lib.

PR #5150, opened 2024-03-27
[VL] Enable bitwise_and, bitwise_not, bitwise_or, bitwise_xor for tinyint & int

zhli1142015:

Enable bitwise_and, bitwise_not, bitwise_or, bitwise_xor for tinyint and int.

## How was this patch tested?

UT


zhli1142015:
Hello @FelixYBW , @PHILO-HE , @rui-mo ,
This pr failed with error `Validation failed due to exception caught at file:SubstraitToVeloxPlanValidator.cc line:804 function:validate, thrown from file:ExprCompiler.cpp line:426 function:compileRewrittenExpression, reason:Found incompatible return types for 'bitwise_not' (BIGINT vs. TINYINT) for input types (TINYINT)..` this is very wierd as the function with correct signatures are already added in Velox.

In Gluten we reigister presto fucntion first then spark functions. Is this because of the overwrite behavior from `registerStatefulVectorFunction`? But for simple functions they use vectors to store function signatures `registerFunctionInternal`.  So for vector functions the last signaturewins, but for simple functions, the first wins. 

This is too confusing, should we make them same behaviors, maybe set `overwrite` as false for all spark vector functions, and register spark fucntions first in gluten?
I saw there are several discussions and fixes for such issue, please suggest any guidence to handle such case.
Thanks.

PHILO-HE:
@zhli1142015, 
So Velox is somehow trying to use prestosql's `bitwise_not`, not the one you added after the below resolving.
https://github.com/facebookincubator/velox/blob/main/velox/expression/SimpleFunctionRegistry.cpp#L143

We always let presto function registered firstly, thus spark function registration can overwrite the previous one if they have same signature (including return type, see [link](https://github.com/facebookincubator/velox/blob/main/velox/expression/SimpleFunctionRegistry.cpp#L48)). So in your case, the overwriting does not happen.

We have a discussion for some other conflict cases at https://github.com/apache/incubator-gluten/discussions/4759 where two fix approaches are suggested.

zhli1142015:
> @zhli1142015, So Velox is somehow trying to use prestosql's `bitwise_not`, not the one you added after the below resolving. https://github.com/facebookincubator/velox/blob/main/velox/expression/SimpleFunctionRegistry.cpp#L143
> 
> We always let presto function registered firstly, thus spark function registration can overwrite the previous one if they have same signature (including return type, see [link](https://github.com/facebookincubator/velox/blob/main/velox/expression/SimpleFunctionRegistry.cpp#L48)). So in your case, the overwriting does not happen.
> 
> We have a discussion for some other conflict cases at #4759 where two fix approaches are suggested.

Looks the overwrite happens only if the function signature exactly matches. 
https://github.com/facebookincubator/velox/blame/0c86a0aedb90bfa966087def7551a40fc3571a01/velox/core/SimpleFunctionMetadata.h#L498-L515C4
Is this some regression?  I test bitwise_and, looks it also failed with same error.
 `Validation failed due to exception caught at file:SubstraitToVeloxPlanValidator.cc line:804 function:validate, thrown from file:ExprCompiler.cpp line:426 function:compileRewrittenExpression, reason:Found incompatible return types for 'bitwise_and' (BIGINT vs. TINYINT) for input types (TINYINT, TINYINT)..`

PHILO-HE:
> Is this some regression?

@zhli1142015, seems before this [commit](https://github.com/facebookincubator/velox/commit/9c7eb6c1732de4240b90ad0bea03ff221eb8e39a), the overwriting still requires signatures are logically equivalent. We are considering to separate the sparksql/prestosql functions (see mentioned discussion link), which can also fix other discovered conflict cases.

zhli1142015:
> > Is this some regression?
> 
> @zhli1142015, seems before this [commit](https://github.com/facebookincubator/velox/commit/9c7eb6c1732de4240b90ad0bea03ff221eb8e39a), the overwriting still requires signatures are logically equivalent. We are considering to separate the sparksql/prestosql functions (see mentioned discussion link), which can also fix other discovered conflict cases.

For this change, I think I need to register bitwise_* functions with spark prefix in gluten separately to work around , is this ok to you or I should wait for the long term solution?
Thanks.

PHILO-HE:
> > > Is this some regression?
> > 
> > 
> > @zhli1142015, seems before this [commit](https://github.com/facebookincubator/velox/commit/9c7eb6c1732de4240b90ad0bea03ff221eb8e39a), the overwriting still requires signatures are logically equivalent. We are considering to separate the sparksql/prestosql functions (see mentioned discussion link), which can also fix other discovered conflict cases.
> 
> For this change, I think I need to register bitwise_* functions with spark prefix in gluten separately to work around , is this ok to you or I should wait for the long term solution? Thanks.

@zhli1142015, it's ok to firstly use this workaround. Thanks!

PHILO-HE (approved):
Thanks for your work!

PR #5159, opened 2024-03-28
[CORE][VL] ACBO: Add GlutenMetadataModel, move Gluten schema def from property model to metadata model

zhztheplayer:
Metadata model would be used to retrieve nodes' logical properties, specifically the collected statistics from Spark in future. With this it would become possible to write heuristics in rule for more directed cost-based optimization.


JkSelf (approved):
LGTM. Thanks.

PR #5167, opened 2024-03-28
[VL] Fix wrong task info when log split info

Yohahaha:

Extract TaskInfo param from `parsePlan` to a new method `setSparkTaskInfo`.


Yohahaha:
@marin-ma please help take a look, thank you!

marin-ma:
Thanks! This dependency seems implicit to developers. Can we add a new method `setTaskInfo()` and remove the `taskInfo` argument from `parsePlan`?

marin-ma (approved):

PR #5171, opened 2024-03-28
[CH] Move project transformer rewriting code to CH backend

zhztheplayer:
Since the rewrite logic is currently specialized to CH backend. ACBO would require for this fix to make sure a node's metadata doesn't get changed after applying rule (although it's not a MUST for ACBO, it's better to keep metadata not changed after applying rules).


zhztheplayer:
@baibaichen @zzcclp Thanks

zzcclp (approved):
LGTM

PR #5175, opened 2024-03-28
[VL] Restore the test cases for corr in group-by.sql and udf-group-by.sql

liujiayi771:

The calculation formula of `corr` in Velox is slightly different from that in Spark, which causes some differences in the accuracy of the final calculation results, but these differences are within the allowed precision range. Previously, the inconsistent test cases were deleted, but now it has been found that other statistics-related agg functions such as `skewness`, `stddev`, and `variance` also have differences in calculation accuracy. We cannot delete all of these test cases that have differences. A better approach is to modify the content in the result file.

In this PR, we restore the `corr` test cases in group-by.sql and udf-group-by.sql, and modify the correct results in .out file.

## How was this patch tested?

Exists CI.

### DIFF
--- a/gluten-ut/spark32/src/test/resources/sql-tests/inputs/group-by.sql
+++ b/gluten-ut/spark32/src/test/resources/sql-tests/inputs/group-by.sql
@@ -75,6 +75,10 @@ SELECT 1 from (
 ) b
 where b.z != b.z;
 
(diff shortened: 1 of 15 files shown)


liujiayi771:
cc @rui-rui, thanks.

PHILO-HE (approved):
Thanks! cc @rui-mo

rui-mo (approved):
Thanks!

Issue #5182, opened 2024-03-28
[CH] Fail to parse post join filter


shuai-xu:
### Backend

CH (ClickHouse)

### Bug description

When run sql d_1340_1, it throw an exception: 
<img width="844" alt="image" src="https://github.com/apache/incubator-gluten/assets/20960838/bcc27473-53e7-47d6-8d8b-969b6a2310c8">
This is because the post join filter may be complex expressions, not only scalar function.

### Spark version

Spark-3.3.x

PR #5183, opened 2024-03-28
[GLUTEN-5182] [CH] fix fail to parse post join filter

shuai-xu:

This pr fix the bug that may fail to parse post join filter if contains complex expressions.

(Fixes: \#5182)

## How was this patch tested?

This patch was tested by ut.


liuneng1994 (approved):
LGTM

PR #5191, opened 2024-03-29
[CORE] Restore the function signature for eliminateProjectList

liujiayi771:

Because we still need to append the pulled-out `NamedExpression`s after the child output, it would be more appropriate to revert back to the original function signature.

## How was this patch tested?

Exists CI tests.


liujiayi771:
Hi @zhli1142015. If we still need to append the pulled-out `NamedExpression`s after the child output, and we want to maintain the order of the project list in the existing golden file tests without modification, I think it would be better to revert back to the original function signature to clearly express that the second parameter "appends after" the first parameter.

zhli1142015 (approved):

PR #5193, opened 2024-03-29
[GLUTEN-5192][VL] Copy bundle jars to dist dir

### LINKED ISSUE #5192: Copy bundle jar to dist dir
### Description

Although multiple spark versions are specified for compilation in `package.sh`, only the last bundle jar is retained because `mvn clean` cleans the previous build.
So I want to copy the bundle jar generated by each build into `dist` directory.

wForget:

closes #5192

## How was this patch tested?

run `./dev/package.sh` after this change: 

image

### DIFF
--- a/dev/buildbundle-veloxbe.sh
+++ b/dev/buildbundle-veloxbe.sh
@@ -4,7 +4,19 @@ BASEDIR=$(dirname $0)
 source "$BASEDIR/builddeps-veloxbe.sh"
 
 cd $GLUTEN_DIR
+
+DISTDIR="$GLUTEN_DIR/dist"
+rm -rf "$DISTDIR"
+mkdir -p "$DISTDIR"
+
 mvn clean package -Pbackends-velox -Prss -Pspark-3.2 -DskipTests
+cp package/target/gluten-*-bundle-spark*.jar "$DISTDIR"
+
 mvn clean package -Pbackends-velox -Prss -Pspark-3.3 -DskipTests
+cp package/target/gluten-*-bundle-spark*.jar "$DISTDIR"
+
 mvn clean package -Pbackends-velox -Prss -Pspark-3.4 -DskipTests
+cp package/target/gluten-*-bundle-spark*.jar "$DISTDIR"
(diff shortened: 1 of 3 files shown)


PHILO-HE (approved):
Looks good! Thanks! cc @ulysses-you

dev/package.sh
```diff
@@ -4,17 +4,28 @@ set -eux
 
 CURRENT_DIR=$(cd "$(dirname "$BASH_SOURCE")"; pwd)
 GLUTEN_DIR="$CURRENT_DIR/.."
-THIRDPARTY_LIB="$GLUTEN_DIR/package/target/thirdparty-lib"
+DISTDIR="$GLUTEN_DIR/dist"
+THIRDPARTY_LIB="$DISTDIR/thirdparty-lib"
```
ulysses-you, line 8: I guess there are some docs should be updated

dev/buildbundle-veloxbe.sh
```diff
@@ -4,7 +4,19 @@ BASEDIR=$(dirname $0)
 source "$BASEDIR/builddeps-veloxbe.sh"
 
 cd $GLUTEN_DIR
+
+DISTDIR="$GLUTEN_DIR/dist"
+rm -rf "$DISTDIR"
+mkdir -p "$DISTDIR"
+
 mvn clean package -Pbackends-velox -Prss -Pspark-3.2 -DskipTests
+cp package/target/gluten-*-bundle-spark*.jar "$DISTDIR"
```
ulysses-you, line 13: should we use `mv` ? It will make binary size bigger

wForget: > should we use `mv` ? It will make binary size bigger

The `package/target` directory is just the mvn build output directory of `package` module, I think we don't need to care about its size, it will be cleaned after `mvn clean`.

dev/package.sh
wForget: Thanks, there seems to be no more documentation, just a commented script..

image

PR #5221, opened 2024-03-31
[GLUTEN-5219][CH]Fix the table metadata sync issue for the CH backend

### LINKED ISSUE #5219: [CH]Fix the table metadata sync issue for the CH backend
### Backend

CH (ClickHouse)

### Bug description

[CH]Fix the table metadata sync issue for the CH backend.
The CH backend receives the table metadata by the substrait plan when executing create table and query, but now there is no  version info to represent the table metadata, when it meets the below case, it will execute failed:
```
1. create a table_1 with the path1, insert data
2. query from the table_1
3. drop the table_1 and re-create a table named table_1, but the table path is path2;
4. insert data, will be found that the inserted data will be located in the path1 but not in the path2;
```
So add a timestamp field to represent the version of the table metadata, this timestamp will be get from the Delta snapshot. The CH backend will update the table metadata according to this timestamp.

### Spark version

None

### Spark configurations

_No response_

### System information

_No response_

### Relevant logs

_No response_

zzcclp:
The CH backend receives the table metadata by the substrait plan when executing create table and query, but now there is no version info to represent the table metadata, when it meets the below case, it will execute failed:
```
1. create a table_1 with the path1, insert data
2. query from the table_1
3. drop the table_1 and re-create a table named table_1, but the table path is path2;
4. insert data, will be found that the inserted data will be located in the path1 but not in the path2;
```
So add a timestamp field to represent the version of the table metadata, this timestamp will be get from the Delta snapshot. The CH backend will update the table metadata according to this timestamp.

Close #5219.

(Fixes: #5219)

## How was this patch tested?


backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/utils/MergeTreePartsPartitionsUtil.scala
```diff
@@ -55,6 +55,9 @@ object MergeTreePartsPartitionsUtil extends Logging {
     }
     val fileIndex = relation.location.asInstanceOf[TahoeFileIndex]
 
+    // when querying, use deltaLog.update(true) to get the staleness acceptable snapshot
+    val latestCommitTimestamp = table.deltaLog.update(true).timestamp
```
binmahone, line 59: will this impact perf benchmark ? need perf result

zzcclp: set `spark.databricks.delta.stalenessLimit` to non-zero, it will launch a thread to update the snapshot:
image

liuneng1994 (commented):
LGTM

binmahone (commented):
LGTM , better to pass gluten-perf

binmahone (approved):

liuneng1994 (approved):
LGTM

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/utils/MergeTreePartsPartitionsUtil.scala
zzcclp: Merge this pr first, will test perf later

PR #5226, opened 2024-04-01
[VL] gluten-it: Remove specific result matching code for Q65

zhztheplayer:
Try directly remove the hacky code to see if jobs can pass.


zhztheplayer:
@rui-mo Please review, thanks!

rui-mo (approved):

PR #5234, opened 2024-04-01
[VL] CI: Update dependency cache only when main branch is updated

zhztheplayer:
After this patch, only main branch is allowed to generate cache then can be used in PR build.

This can solve some inconsistency issue when PR's Velox branch gets changed.


zhztheplayer:
@zhouyuan

PR #5235, opened 2024-04-01
[Gluten-5152][CH] fix core dump issues when running in parallel

binmahone:

When testing querying and optimizing with TPCH 100 data with 30 core,
we may encounter core dump issues, this PR fixes it.

Also fix some potential memory leak.

## How was this patch tested?


zzcclp (approved):
LGTM

PR #5239, opened 2024-04-01
[CH] Support csc/sec/cot function

exmy:

Support csc/sec/cot function

## How was this patch tested?

PASS CI


liuneng1994 (approved):
LGTM

PR #5246, opened 2024-04-02
[GLUTEN-5252] Fix condition eliminate for Iceberg/delta scan with filter rel

yma11:

There is an optimization that eliminate common expression between scan and filter, but it doesn't apply on `IcebergScanTransformer`/`DeltaScanTransformer ` as they are not inherited from `BasicScanExecTransformer`. This PR will address this issue.

## How was this patch tested?

CI

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala
@@ -22,13 +22,22 @@ import org.apache.spark.sql.execution.SparkPlan
 case class FilterExecTransformer(condition: Expression, child: SparkPlan)
   extends FilterExecTransformerBase(condition, child) {
   override protected def getRemainingCondition: Expression = {
-    val scanFilters = child match {
-      // Get the filters including the manually pushed down ones.
-      case basicScanExecTransformer: BasicScanExecTransformer =>
-        basicScanExecTransformer.filterExprs()
-      // For fallback scan, we need to keep original filter.
-      case _ =>
-        Seq.empty[Expression]
+    var scanFilters: Seq[Expression] = Seq()
+    if (child.isInstanceOf[BatchScanExecTransformerBase]) {
+      scanFilters = child.asInstanceOf[BatchScanExecTransformerBase].filterExprs();
+    } else if (child.isInstanceOf[FileSourceScanExecTransformerBase]) {
+      scanFilters = child.asInstanceOf[FileSourceScanExecTransformerBase].filterExprs();
+    } else {
+      scanFilters = child match {
+        // Get the filters including the manually pushed down ones.
+        case basicScanExecTransformer: BasicScanExecTransformer =>
+          basicScanExecTransformer.filterExprs()
+        // For fallback scan, we need to keep original filter.
+        //      case batchScanExecTransformer: BatchScanExecTransformerBase =>
+        //        batchScanExecTransformer.filterExprs()
+        case _ =>
+          Seq.empty[Expression]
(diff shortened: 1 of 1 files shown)


yma11:
@liujiayi771 @YannByron Can you help review on this change? Thanks.

backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala
```diff
@@ -22,13 +22,22 @@ import org.apache.spark.sql.execution.SparkPlan
...
   extends FilterExecTransformerBase(condition, child) {
   override protected def getRemainingCondition: Expression = {
-    val scanFilters = child match {
-      // Get the filters including the manually pushed down ones.
-      case basicScanExecTransformer: BasicScanExecTransformer =>
-        basicScanExecTransformer.filterExprs()
-      // For fallback scan, we need to keep original filter.
-      case _ =>
-        Seq.empty[Expression]
+    var scanFilters: Seq[Expression] = Seq()
```
liujiayi771, line 25: Would you use match case instead of using `isInstanceOf`?

yma11: `IcebergScanTransformer`/`DeltaScanTransformer` is inherited from abstract class and can't be matched using case.

YannByron: An abstract class can be used in `match-case`.

liujiayi771: Can this not work?
```scala
val scanFilters = child match {
      case base: BatchScanExecTransformerBase =>
        base.filterExprs();
      case base: FileSourceScanExecTransformerBase =>
        base.filterExprs();
      case _ =>
        child match {
          // Get the filters including the manually pushed down ones.
          case basicScanExecTransformer: BasicScanExecTransformer =>
            basicScanExecTransformer.filterExprs()
          // For fallback scan, we need to keep original filter.
          //      case batchScanExecTransformer: BatchScanExecTransformerBase =>
          //        batchScanExecTransformer.filterExprs()
          case _ =>
            Seq.empty[Expression]
        }
    }
```

backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala
```diff
@@ -22,13 +22,22 @@ import org.apache.spark.sql.execution.SparkPlan
...
+    } else if (child.isInstanceOf[FileSourceScanExecTransformerBase]) {
+      scanFilters = child.asInstanceOf[FileSourceScanExecTransformerBase].filterExprs();
+    } else {
+      scanFilters = child match {
+        // Get the filters including the manually pushed down ones.
+        case basicScanExecTransformer: BasicScanExecTransformer =>
+          basicScanExecTransformer.filterExprs()
+        // For fallback scan, we need to keep original filter.
+        //      case batchScanExecTransformer: BatchScanExecTransformerBase =>
+        //        batchScanExecTransformer.filterExprs()
```
liujiayi771, line 37: Is this intentionally commented out here?

yma11: Yes.

yma11:
@liujiayi771 @YannByron I found the existing case match can work for iceberg/delta because of recent [PR](https://github.com/apache/incubator-gluten/pull/4893) so current patch not needed any more. Sorry for your time.

PR #5276, opened 2024-04-03
[VL] RAS: Add EnumeratedApplier to manage columnar rule applications when ras is enabled

zhztheplayer:
1. Add API `ColumnarRuleApplier`
2. Move heuristic rule application code into implementation `HeuristicApplier`
3. Move enumerated rule application code into implementation `EnumeratedApplier`
4. Move adaptive context code to independent class `AdaptiveContext` (original PR: https://github.com/apache/incubator-gluten/pull/3294)

After this, most of the further development work on Gluten ras integration will take place inside a single source folder `gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic`.


gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/HeuristicApplier.scala
```diff
@@ -0,0 +1,173 @@
...
+      adaptiveContext.resetOriginalPlan()
+      adaptiveContext.resetAdaptiveContext()
+    }
+  }
+
+  /**
+   * Rules to let planner create a suggested Gluten plan being sent to `fallbackPolicies` in which
+   * the plan will be breakdown and decided to be fallen back or not.
+   */
+  private def transformRules(outputsColumnar: Boolean): List[SparkSession => Rule[SparkPlan]] = {
```
PHILO-HE, line 101: Looks there are few difference between the rule sets (`transformRules`, `fallbackPolicies`, `postRules`, `finalRules`) of `HeuristicApplier` & `EnumeratedApplier`. Will some more differences be introduced in the future development? If not, may be better to put common rules in some common place or super class?

PHILO-HE (commented):
Looks good! Just some small suggestions. Please check whether they make sense. Thanks!

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/HeuristicApplier.scala
zhztheplayer: Actually it's intentional to have some duplicated code in `HeuristicApplier` and `EnumeratedApplier` for this PR. In my plan `EnumeratedApplier`'s code will be frequently updated in the future PRs of RAS, and end up with very different design with `HeuristicApplier` so the two will end up sharing less (or event no) code. 

So the duplicated code is for better isolation of the two implementations to ease future development.

PHILO-HE: @zhztheplayer, I see. Thanks for the clarification!

PHILO-HE (approved):
Thanks!

zhztheplayer:
Thanks for reviewing @PHILO-HE . I'll add some comments to newly added components in further PR.

PR #5280, opened 2024-04-03
[VL][CI] Add a nightly build job

PHILO-HE:

This PR proposes a nightly job to verify gluten build script from a clean environment.

1. Gluten's existing CI jobs are using docker images with pre-installed dependencies to accelerate validation process, but it is possible the code doesn't work in a clean environment.
E.g., Gluten changes velox's setup script by using "sed" command to match some code, then modifying it. The expected change sometimes doesn't happen due to upstream's code changes.
2. dev/package.sh is still necessary to some users, even though static build is recommended. It lacks CI verification. 

To avoid slowing down PR verification process, this job will be triggered only if the added yml file is changed or scheduled nightly test time comes.

## How was this patch tested?

N/A


PHILO-HE:

rui-mo (approved):

PR #5283, opened 2024-04-03
[CORE] [UT] Ensure GlutenBroadcastJoinSuite run with Gluten enabled

Yohahaha:


Yohahaha:
@PHILO-HE please help take a look, thank you!

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/joins/GlutenBroadcastJoinSuite.scala
```diff
@@ -74,5 +72,6 @@ class GlutenBroadcastJoinSuite extends BroadcastJoinSuite with GlutenTestsCommon
         .config("spark.unsafe.exceptionOnMemoryLeak", "true")
         .getOrCreate()
     }
+    super.beforeAll()
```
zhztheplayer, line 75: Thanks although I am trying to get the rationale of the fix. Would you like to explain a bit? Given that `super.beforeAll` also sets the test spark session:

https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/execution/joins/BroadcastJoinSuite.scala#L58-L65

Yohahaha: SparkContext only initializes once in JVM, before this patch, this suites only run with vanilla Spark, because `super.beforeAll` generate a SparkContext instance without Gluten plugins. After this patch, we can inject Gluten's config before SparkContext initialization.

zhztheplayer: Thanks for the explanation. Though it's so implicit to me if the call order of `super.beforeAll()` matters. Should we add some comments in code to explain that?

And maybe we should also check other test files to see if similar issues exist.

zhztheplayer (approved):
Thanks for catching!

PR #5291, opened 2024-04-04
[CH][Minor] Fix build due to Clickhouse Refactor

baibaichen:

Fix build due to https://github.com/ClickHouse/ClickHouse/pull/62170, which avoids public dependency on `CurrentThread.h`  and hence cause our build failed.

We can avoid https://github.com/apache/incubator-gluten/pull/5289 upgrading clickhouse backend by this PR.

## How was this patch tested?
Using existed UT


baibaichen:
lgtm

PR #5305, opened 2024-04-07
[CORE][VL] RAS: Refactor memo cache to look up on cluster-canonical node rather than on group-canonical node

zhztheplayer:
Explanation on terms:

1. cluster-canonical node: 
A node with all children replaced by cluster key.
2. group-canonical node:
A node with all children replaced by group id.

After the patch, memo will use cluster-canonical node as the cache key to look up for resident cluster of new node.

This will make the memo size and memo cache smaller during optimization since we always create new cluster if cache is not hit when adding node to memo.

Consider two plans are inserted to memo respectively:

```
Project(col1, col2)
  \-  Group(cluster0, ordering by col1)
```

```
Project(col1, col2)
  \- Group(cluster0, ordering by col2)
```

Previously the second insertion would create a new cluster since `ordering by col2` is considered in cache key. After the patch the second insertion will no longer create new cluster.


gluten-ras/common/src/test/scala/org/apache/gluten/ras/PropertySuite.scala
```diff
@@ -37,6 +38,30 @@ abstract class PropertySuite extends AnyFunSuite {
...
+    memo.memorize(ras, PassNodeType(1, PassNodeType(1, PassNodeType(1, TypedLeaf(TypeA, 1)))))
+    val leafGroup = memo.memorize(ras, TypedLeaf(TypeA, 1))
+    memo
+      .openFor(leafGroup.clusterKey())
+      .memorize(ras, TypedLeaf(TypeB, 1))
+    memo.memorize(ras, PassNodeType(1, PassNodeType(1, PassNodeType(1, TypedLeaf(TypeB, 1)))))
+    val state = memo.newState()
+    assert(state.allClusters().size == 4)
+    assert(state.getGroupCount() == 11)
+  }
```
zhztheplayer, line 63: Memo dump of this case:

Before:
graphviz

After:
graphviz (1)

Issue #5215, opened 2024-03-30
[CH]fix precision issue in statistical aggregation functions


loudongfeng:
### Backend

CH (ClickHouse)

### Bug description

The precision is different between Spark and Gluten+CH back end for the following statistical aggregation functions:
var_samp
var_pop
covar_samp
covar_pop
stddev_samp
stddev_pop
corr
variance

### Spark version

Spark-3.3.x

PR #5311, opened 2024-04-07
[GLUTEN-5215][CH]stable precision in statistical aggregation functions

loudongfeng:
1. use varSampStable,varPopStable ... to replace varSamp, varPop ...
2. use fixed nullable property for corr like aggregation's input
3. convert Infinite to Null to keep compatiable with Spark

Fixes: \#5215

## How was this patch tested?

tested by UT


backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHHashAggregateExecTransformer.scala
```diff
@@ -349,13 +349,18 @@ case class CHHashAggregateExecTransformer(
...
               var fields = Seq[(DataType, Boolean)]()
-              fields = fields :+ (covar.left.dataType, covar.left.nullable)
-              fields = fields :+ (covar.right.dataType, covar.right.nullable)
+              // when COVAR_SAMP's input is non-nullable , due to its output maybe null in spark,
+              // backend parser will cast input type to nullable for intermediate agg,
+              // While final stage still use input type no changed.
+              // It causes type miss match between partial and final AggregationFunction
+              // so directly assign attr.nullable here, which should always be nullable
+              fields = fields :+ (covar.left.dataType, attr.nullable)
+              fields = fields :+ (covar.right.dataType, attr.nullable)
```
lgbo-ustc, line 358: is there any case?

lgbo-ustc:
Is there any chance to modify the implementation in `CH` ? It will bring additional costs to add a`ifInfinite` at the end.
Or could you give some performance test result to know how much is the cost

backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHHashAggregateExecTransformer.scala
lgbo-ustc: I am concerned about this change. If the nullability of input is different from result, it could cause an exception when CH parses the intemediate aggregate result in the final stage. How could this  avoid ?

cpp-ch/local-engine/Parser/aggregate_function_parser/CommonAggregateFunctionParser.cpp
```diff
@@ -25,17 +25,17 @@ REGISTER_COMMON_AGGREGATE_FUNCTION_PARSER(Sum, sum, sum)
 REGISTER_COMMON_AGGREGATE_FUNCTION_PARSER(Avg, avg, avg)
 REGISTER_COMMON_AGGREGATE_FUNCTION_PARSER(Min, min, min)
 REGISTER_COMMON_AGGREGATE_FUNCTION_PARSER(Max, max, max)
-REGISTER_COMMON_AGGREGATE_FUNCTION_PARSER(StdDev, stddev, stddev_samp)
```
taiyang-li, line 28: How much performance degradation if replace them all to stable implementations?  

According to https://clickhouse.com/docs/en/sql-reference/aggregate-functions/reference/stddevsamp: use the stddevSampStable function. It works slower but provides a lower computational error.

cpp-ch/local-engine/Functions/FunctionIfInfinite.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <Core/ColumnNumbers.h>
+
+
+namespace local_engine
+{
+using namespace DB;
+namespace
+{
+
+    /// IfInfinite(x, y) is equivalent to isInfinite(x) ? y : x.
```
taiyang-li, line 31: No need to implement a new function. It is too heavy. All we need is just update ActionsDAGs in  convertInfiniteIfNeeded. Pls refer to cpp-ch/local-engine/Parser/scalar_function_parser/*.cpp

cpp-ch/local-engine/Parser/AggregateFunctionParser.cpp
```diff
@@ -174,6 +174,28 @@ const DB::ActionsDAG::Node * AggregateFunctionParser::convertNodeTypeIfNeeded(
     return func_node;
 }
 
+const DB::ActionsDAG::Node * AggregateFunctionParser::convertInfiniteIfNeeded(
+    const CommonFunctionInfo & func_info,
+    const DB::ActionsDAG::Node * func_node,
+    DB::ActionsDAGPtr & actions_dag) const
+{
+    static std::unordered_set<String> infinite_aggs{"varSampStable", "stddevSampStable", "covarSampStable"};
```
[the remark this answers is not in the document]
taiyang-li: Maybe reusing `convertNodeTypeIfNeeded` interface is better choice than adding another not so common interface `convertInfiniteIfNeeded`

backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHHashAggregateExecTransformer.scala
loudongfeng: Please refer to the newly added test case "covar_samp non-nullable input", which use a temp inline table  and all columns are non-nullable.Before this fix, it will fail.

lgbo-ustc: Is it that the intermediate result will be nullable any way in `CH` ?

loudongfeng:
Find that spark also return Infinity in some cases, so convert Infinity to null does not work.
spark-sql --master local[4] --conf  spark.sql.shuffle.partitions=3
```sql
drop table if exists example_table1;
CREATE TABLE example_table1(id int, value double) USING parquet;
INSERT INTO example_table1 VALUES
(1, 4.9E-322),
(2, 1.7976931348623157E308),
(3, 1.7976931348623157E308);
SELECT var_samp(value) FROM example_table1; 
```
the result will be Infinity., while return NaN for the following:
spark-sql --master local[2]

loudongfeng:
Nullable property issue only affect stable statistical aggregation functions, so no need to change for now.

PR #5314, opened 2024-04-08
[VL] Upgrade folly to v2024.04.01.00

PHILO-HE:

To align with Velox's change: https://github.com/facebookincubator/velox/commit/5fed974784a2883dbe61535aa1111957dd0effbb.

If the patches under ports/ cannot be applied, no error is reported. So we need to manually apply the patches (uses the nearest version in vcpkg repo) to see the validity.

To fix the below build error caused by exception tracer module, we need to set `-DFOLLY_NO_EXCEPTION_TRACER=ON`.

```
(.text.__cxa_begin_catch+0x0): multiple definition of `__cxa_begin_catch'
/usr/local/lib/libfolly.a(ExceptionTracerLib.cpp.o):/tmp/build/folly/folly/experimental/exception_tracer/ExceptionTracerLib.cpp:123: first defined here
```

## How was this patch tested?

Existing tests.


zhouyuan:
@PHILO-HE  It looks like this patch is missed

rui-mo (approved):

Issue #5307, opened 2024-04-07
[VL] Potential Overflow Issue in `VeloxShuffleWriter` Due to Mismatched Data Types of RowNumber


yangzhg:
### Backend

VL (Velox)

### Bug description

In the `VeloxShuffleWriter` component of the project, several variables are declared with the `uint16_t` data type, which includes `partition2RowCount_`, `partition2RowOffsetBase_`, `rowOffset2RowId_`, `partitionBufferSize_`, and `partitionBufferBase_`. These variables serve various purposes, such as tracking the row count for each partition, managing row offsets, and holding partition buffer sizes and write positions.

However, there's a potential overflow risk associated with these `uint16_t` variables when interfacing with Velox's RowVector data structure. The size attribute of a RowVector in Velox is of type `vector_size_t` (`int32_t`), which allows for a significantly larger range of values compared to `uint16_t`. This discrepancy in data type sizes could lead to overflow issues when `RowVector` sizes exceed the maximum value that a `uint16_t` variable can represent (65,535).

Moreover, an additional issue has been identified with Velox's `HashAggregation::getOutput` method, which does not strictly control the number of rows outputted. This lack of stringent control further exacerbates the potential for overflow within `VeloxShuffleWriter`, considering the existing data type mismatches. Especially in ` HashAggregation::getDistinctOutput` when `newDistincts_` is `true`, there is no limit.
 https://github.com/facebookincubator/velox/blob/84ae6bfd6f61775b9e20742cdc5bc73ecd0829c3/velox/exec/HashAggregation.cpp#L266

The query may be like `select  c1, c2_t from t lateral view explode(split(c2, ',')) tmp as c2_t group by c1, c2_t`. And the table `t` only have two columns of string type `c1` ,`c2`  while `c2` is a `,` splited string

### Spark version

None

FelixYBW:
So the root cause is that Velox may output rowvector with unlimited batch size.

If so we need raise the issue to Velox.

PR #5326, opened 2024-04-08
[GLUTEN-5307][VL] Fix Potential Overflow Issue in VeloxShuffleWriter Due to Mismatched Data Types of RowNumber

yangzhg:

Fixes: #5307

### DIFF
--- a/cpp/core/shuffle/FallbackRangePartitioner.cc
+++ b/cpp/core/shuffle/FallbackRangePartitioner.cc
@@ -24,7 +24,7 @@ arrow::Status gluten::FallbackRangePartitioner::compute(
     const int32_t* pidArr,
     const int64_t numRows,
     std::vector<uint32_t>& row2Partition,
-    std::vector<uint16_t>& partition2RowCount) {
+    std::vector<uint32_t>& partition2RowCount) {
   row2Partition.resize(numRows);
   std::fill(std::begin(partition2RowCount), std::end(partition2RowCount), 0);
   for (auto i = 0; i < numRows; ++i) {
--- a/cpp/core/shuffle/FallbackRangePartitioner.h
+++ b/cpp/core/shuffle/FallbackRangePartitioner.h
@@ -29,7 +29,7 @@ class FallbackRangePartitioner final : public Partitioner {
       const int32_t* pidArr,
       const int64_t numRows,
       std::vector<uint32_t>& row2partition,
-      std::vector<uint16_t>& partition2RowCount) override;
+      std::vector<uint32_t>& partition2RowCount) override;
 };
 
 } // namespace gluten
--- a/cpp/core/shuffle/HashPartitioner.cc
+++ b/cpp/core/shuffle/HashPartitioner.cc
@@ -23,7 +23,7 @@ arrow::Status gluten::HashPartitioner::compute(
     const int32_t* pidArr,
     const int64_t numRows,
     std::vector<uint32_t>& row2partition,
-    std::vector<uint16_t>& partition2RowCount) {
+    std::vector<uint32_t>& partition2RowCount) {
   row2partition.resize(numRows);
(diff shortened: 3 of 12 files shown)


marin-ma:
Thanks. Does this cause overflow in real cases? The uint16_t is used on purpose to benefit L2 cache, and is not supposed to cause the overflow with the 32k limitation of the batch size. Please check https://github.com/apache/incubator-gluten/pull/4401#issuecomment-1895360554

yangzhg:
> Thanks. Does this cause overflow in real cases? The uint16_t is used on purpose to benefit L2 cache, and is not supposed to cause the overflow with the 32k limitation of the batch size. Please check [#4401 (comment)](https://github.com/apache/incubator-gluten/pull/4401#issuecomment-1895360554)
> 
> cc: @FelixYBW

Indeed, using uint16_t in this scenario can lead to overflow.   The specific statements and conditions for reproducing this issue have already been outlined in issue #5307, and I will be providing a dataset shortly to further illustrate this point.

marin-ma:
@yangzhg Thanks for your detailed explanation. There are configurations in Gluten to explicitly limit the batch size within 32k https://github.com/marin-ma/gluten/blob/cf391fae10da22d73c451aced5cd05e474e3bccd/shims/common/src/main/scala/io/glutenproject/GlutenConfig.scala#L929-L946, (the configurations are passed to velox QueryContext here https://github.com/apache/incubator-gluten/blob/c60b5d90e0be23430672d732f9d574267b67c06e/cpp/velox/compute/WholeStageResultIterator.cc#L477-L480), and the prerequisites of doing that is we expect the batch size produced by Velox pipeline always <= 32k. If the input batch size doesn't exceed this limitation, there shouldn't be a problem to use uint16_t. 

However, it seems like there are cases that the output from Velox pipeline can exceed the configured batch size. We are not sure if it's an expected case in Velox. Could you help to raise an issue in Velox to illustrate the case you've, that the Hashagg output can exceed the configured batch size? Thanks!

yangzhg:
> @yangzhg Thanks for your detailed explanation. There are configurations in Gluten to explicitly limit the batch size within 32k
> 
> https://github.com/apache/incubator-gluten/blob/c60b5d90e0be23430672d732f9d574267b67c06e/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala#L478-L487
> 
> , the configurations are passed to velox QueryContext here
> https://github.com/apache/incubator-gluten/blob/c60b5d90e0be23430672d732f9d574267b67c06e/cpp/velox/compute/WholeStageResultIterator.cc#L477-L480
> 
> , and the prerequisites of doing that is we expect the batch size produced by Velox pipeline always <= 32k. If the input batch size doesn't exceed this limitation, there shouldn't be a problem to use uint16_t.
> However, it seems like there are cases that the output from Velox pipeline can exceed the configured batch size. We are not sure if it's an expected case in Velox. Could you help to raise an issue in Velox to illustrate the case you've met, that the Hashagg output can exceed the configured batch size? Thanks!

OK! Just out of curiosity, how much performance improvement can using uint16_t bring compared to uint32_t?

FelixYBW:
> OK! Just out of curiosity, how much performance improvement can using uint16_t bring compared to uint32_t?

It's up to queries and CPU. You may test in your case. Memory latency can be >50x of L2 cache latency. Split operation is top 5 hotspot in our TPCH test.

If batch size is unlimited, it may more hurt Velox pipeline's performance. Velox is vectorized engine, basic policy is that one operator's output is used by following operator's input. So if batch size is too large it will leads to big cache miss and hurt performance. It's why we use 4K rows as default now.

marin-ma:
@yangzhg Sorry that I didn't notice this PR wasn't got merged. There's no performance impact. Thanks for your fix!

FelixYBW:
@yangzhg Did you create an issue in Velox community?

PR #5330, opened 2024-04-09
[CH] Fix diff of factorial function

exmy:

* Fix diff of factorial function, `factorial(x)` should return null when x > 20 or x < 0

* enable spark ut about `ceil` and `floor`

## How was this patch tested?

PASS CI


liuneng1994 (approved):
LGTM

PR #5345, opened 2024-04-09
[GLUTEN-5344][VL] Add some parquet example files from parquet-mr for native read test

yma11:

Add some parquet example files from parquet-mr for native read test. 

## How was this patch tested?

UT


yma11:
@FelixYBW I firstly port the example data files from parquet-mr in velox backend test, please take a review.

FelixYBW:
Can they pass on current Gluten?

yma11:
> Can they pass on current Gluten?
The `test-file-with-no-column-indexes-1.parquet` can't be read by Velox so I exclude it in test for the time being. There should be a bug. Interesting thing is that some of the example files can't be read by Spark32/33/34 but can be read by Velox successfully.

FelixYBW:
> > Can they pass on current Gluten?
> > The `test-file-with-no-column-indexes-1.parquet` can't be read by Velox so I exclude it in test for the time being. There should be a bug. Interesting thing is that some of the example files can't be read by Spark32/33/34 but can be read by Velox successfully.

Interesting! Spark should use upstream parquet-mr but needs to specify some parameters. May because of these parameters. We may remove them since not compatible with Spark.

yma11:
> > > Can they pass on current Gluten?
> > > The `test-file-with-no-column-indexes-1.parquet` can't be read by Velox so I exclude it in test for the time being. There should be a bug. Interesting thing is that some of the example files can't be read by Spark32/33/34 but can be read by Velox successfully.
> 
> Interesting! Spark should use upstream parquet-mr but needs to specify some parameters. May because of these parameters. We may remove them since not compatible with Spark.

That file has an empty row group which isn't allowed by Spark3.2/3.3/3.4. But it can be read by Spark 3.5. So I keep it and limit the test on Spark 3.5 version for a broader coverage.

yma11:
@FelixYBW okay to merge this PR?

FelixYBW (approved):

FelixYBW:
> @FelixYBW okay to merge this PR?

Yes, let's merge it

PR #5356, opened 2024-04-10
[VL] fix wrong result for try_add

zhli1142015:

After spark 3.4, try_add would not be converted to try_eval(add), so it isn't fallbacked and output wrong results.
Fix this by adding extra check, and try_add for IntegralTypes can be offload to velox.
TryMultiply, TrySubtract, TryDivide should have silimar issues, will fix them in separate PRs.

## How was this patch tested?
UT.


shims/common/src/main/scala/org/apache/gluten/expression/ExpressionNames.scala
```diff
@@ -81,6 +81,8 @@ object ExpressionNames {
   final val NOT = "not"
   final val IS_NAN = "isnan"
   final val NANVL = "nanvl"
+  final val TRY_EVAL = "try"
+  final val CHECK_ADD = "plus"
```
PHILO-HE, line 85: I note this is for Spark's `TryAdd`. And considering "plus" is velox-specific, it may be better to change it to:

`final val TRY_ADD = "try_add"`

Then, mapping it to velox function name in `SubstraitParser.cc`

PHILO-HE (commented):
@zhli1142015, looks good! Just one small suggestion. Thanks!

PHILO-HE (approved):
Looks good! cc @rui-mo

PR #5358, opened 2024-04-10
[VL] Update to_utc_timestamp and from_utc_timestamp tests

acvictor:

The earlier implementation of to/from_utc_timestamp excluded timezones specified as an offset from UTC. This is fixed here https://github.com/facebookincubator/velox/pull/9403 so removing the test overrides which passed timezone as full string only.


acvictor:
@PHILO-HE can you please review?

PHILO-HE (approved):
@acvictor, thanks for your follow-up!

PR #5367, opened 2024-04-11
[VL] RAS: Integrate filter rules into enumerated transform

zhztheplayer:
Note: **It's a small PR**. Most of the changed lines are from golden files.

Using following 3 RAS rules to do filter transform / pushdown:

1. ImplementFilter
This rule transforms a single vanilla Spark filter to Gluten filter
2. PushFilterToScan
This rule matches on Filter+Scan pattern and do filter push-down
3. FilterRemoveRule
This rule matches on Filter+Scan pattern and removes the top filter when the filter conditions were already pushed to scan

As a result, heuristic rule of filter transformation is removed from `EnumerableApplier`.

Review with golden files excluded:
https://github.com/apache/incubator-gluten/pull/5367/files?file-filters%5B%5D=.scala&show-viewed-files=true

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+backends-velox/src/test/resources/tpch-approved-plan/** linguist-generated
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-backends-velox/src/test/resources/tpch-approved-plan/* linguist-generated
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+backends-velox/src/test/resources/tpch-approved-plan/* linguist-generated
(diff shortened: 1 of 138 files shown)


zwangsheng:
> @zwangsheng It feels messy that golden files are being considered into PR/contribution line count. Do you think we can binarize the golden files?
> 

Sounds good to me, we don't care about the specific storage format, as long as it's good enough for valid comparisons during unit testing.

.gitattributes
```diff
@@ -0,0 +1 @@
+backends-velox/src/test/resources/tpch-approved-plan/** linguist-generated
```
zwangsheng, line 1: Nice job!

zhztheplayer: But sadly by `linguist-generated` we can only fold the files in PR review view. It still doesn't reduce the changed line count. So this is still needed to be enhanced.

PHILO-HE (approved):
Looks good! Thanks!

PR #5371, opened 2024-04-11
[VL] Enable weekofyear function

Yohahaha:


backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
```diff
@@ -747,4 +747,14 @@ class ScalarFunctionsValidateSuite extends FunctionsValidateTest {
     }
   }
 
+  test("weekofyear") {
+    withTable("t") {
+      sql("create table t (dt date) using parquet")
+      sql("insert into t values(date '2008-02-20')")
```
yma11, line 753: can include values like `null`?

Yohahaha: DateExpressionSuite already has null tests, this test only verify function be offloaded.

PHILO-HE (approved):
Looks good! It's strange `YEAR_OF_WEEK` is mapped to "week_of_year" also. I will confirm.

PR #5376, opened 2024-04-12
[VL] Fix timestamp precision loss in serializer

zhli1142015:

By default `PrestoVectorSerde` use millisecond precision for timestamp data. We observed timestamp prescision loseing when doing broadcast serialization. We need to fix this by setting `useLosslessTimestamp` for `PrestoVectorSerde`.

```
import org.apache.spark.sql.functions.expr
import org.apache.spark.sql.catalyst.expressions._
spark.range(0, 5).toDF("id").createOrReplaceTempView("target")
spark.range(0, 10)
        .toDF("id")
        .withColumn("timestampOne", expr("timestamp_micros(id)"))
        .createOrReplaceTempView("source")

spark.sql("SELECT unix_micros(source.timestampOne) FROM source RIGHT OUTER JOIN target ON target.id = source.id").show(false)
```
Wrong result:
```
+-------------------------+
|unix_micros(timestampOne)|
+-------------------------+
|0                        |
|0                        |
|0                        |
|0                        |
|0                        |
+-------------------------+
```
Expected result:
```
+-------------------------+
|unix_micros(timestampOne)|
+-------------------------+
|0                        |
|1                        |
|2                        |
|3                        |
|4                        |
+-------------------------+
```

## How was this patch tested?
UT

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
@@ -1307,4 +1307,15 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
     assert(plan1.find(_.isInstanceOf[ProjectExecTransformer]).isDefined)
     assert(plan2.find(_.isInstanceOf[ProjectExecTransformer]).isDefined)
   }
+
+  test("timestamp broadcast join") {
+    spark.range(0, 5).createOrReplaceTempView("right")
+    spark.sql("SELECT id, timestamp_micros(id) as ts from right").createOrReplaceTempView("left")
+    val expected = spark.sql("SELECT unix_micros(ts) from left")
+    val df = spark.sql(
+      "SELECT unix_micros(ts)" +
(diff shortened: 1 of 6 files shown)


acvictor (approved):

zhli1142015:
cc @PHILO-HE and @marin-ma , thanks.

zhli1142015:
BTW, from this setting's comments, it's used for spilling also, not sure if this should be set to true for spilling also?
https://github.com/facebookincubator/velox/blob/main/velox/serializers/PrestoSerializer.h#L62

zhli1142015:
> Thanks for this fix. Looks like it's already set to true in Velox Spill https://github.com/facebookincubator/velox/blob/b1252f26e112e9632e5a3e11df536387282c0259/velox/exec/SpillFile.cpp#L197-L198
> 
> https://github.com/facebookincubator/velox/blob/b1252f26e112e9632e5a3e11df536387282c0259/velox/exec/SpillFile.cpp#L23-L28
> 
> cc: @zhztheplayer

Get it. Thanks for your info.

PHILO-HE (approved):
Looks good! cc @zhztheplayer

PR #5382, opened 2024-04-12
[GLUTEN-5381] Refine testWithSpecifiedSparkVersion to compare major version

yma11:

Currently in the test method testWithSpecifiedSparkVersion, it uses SPARK_VERSION_SHORT for comparison directly which with value like 3.3.1, thus in test case like testWithSpecifiedSparkVersion(XXX, Some("3.3"), Some("3.3") will be ignored but it's expected to run. We'd better use major version 3.3 for comparison.

## How was this patch tested?

CI

### DIFF
--- a/gluten-core/src/test/scala/org/apache/spark/sql/GlutenQueryTest.scala
+++ b/gluten-core/src/test/scala/org/apache/spark/sql/GlutenQueryTest.scala
@@ -44,14 +44,35 @@ abstract class GlutenQueryTest extends PlanTest {
 
   protected def spark: SparkSession
 
+  def shouldRun(
+      minSparkVersion: Option[String] = None,
+      maxSparkVersion: Option[String] = None): Boolean = {
+    val version = SPARK_VERSION_SHORT.split("\\.")
+    var shouldRun = true
+    if (!minSparkVersion.isEmpty) {
+      val minVersion = minSparkVersion.get.split("\\.");
+      shouldRun =
+        minVersion(0) < version(0) || (minVersion(0) == version(0) && minVersion(1) <= version(1))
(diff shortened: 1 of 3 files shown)


gluten-core/src/test/scala/org/apache/spark/sql/GlutenQueryTest.scala
```diff
@@ -44,13 +44,18 @@ abstract class GlutenQueryTest extends PlanTest {
...
+
   def ignore(
       testName: String,
       minSparkVersion: Option[String] = None,
       maxSparkVersion: Option[String] = None)(testFun: => Any): Unit = {
     if (
-      minSparkVersion.forall(_ <= SPARK_VERSION_SHORT)
-      && maxSparkVersion.forall(_ >= SPARK_VERSION_SHORT)
+      minSparkVersion.forall(_ <= sparkVersion)
+      && maxSparkVersion.forall(_ >= sparkVersion)
```
zhztheplayer, line 58: It's like comparing by plain string was wrong...

I ran

```scala
println("3.3" > "3.2")
println("3.1" > "3.2")
println("3.12" > "3.2")
```

Output:
```
true
false
false
```

Apparently `println("3.12" > "3.2")` should output `true` rather than `false`.

yma11: Seems this is the case, let me update the comparison method.

yma11: Updated. Check again?

zhztheplayer (approved):
Thanks!

PR #5388, opened 2024-04-12
[CH] [Minor] fix ut due to https://github.com/ClickHouse/ClickHouse/pull/61216

baibaichen:

After https://github.com/ClickHouse/ClickHouse/pull/61216, We will failed at [`tryPushDownFilter()` in filterPushDown.cpp](https://github.com/ClickHouse/ClickHouse/blob/db46866db749018466da2de114c4a6e21b853900/src/Processors/QueryPlan/Optimizations/filterPushDown.cpp#L272)

Here is a workaround, refer to `chooseJoinAlgorithm()` in PlannerJoins.cpp, it always call `TableJoin::setRename` to create aliases for columns in the right table. By using right table header name sets, so `TableJoin::deduplicateAndQualifyColumnNames` can do same thing as `chooseJoinAlgorithm()`

ffected UT fixed bh this workaround:
 - GlutenClickHouseTPCHParquetRFSuite:TPCH Q17, Q19, Q20, Q21

## How was this patch tested?

Using existed UT

### DIFF
--- a/cpp-ch/local-engine/Parser/JoinRelParser.cpp
+++ b/cpp-ch/local-engine/Parser/JoinRelParser.cpp
@@ -286,8 +286,16 @@ DB::QueryPlanPtr JoinRelParser::parseJoin(const substrait::JoinRel & join, DB::Q
 
 void JoinRelParser::addConvertStep(TableJoin & table_join, DB::QueryPlan & left, DB::QueryPlan & right)
 {
+
+    /// After https://github.com/ClickHouse/ClickHouse/pull/61216, We will failed at tryPushDownFilter() in filterPushDown.cpp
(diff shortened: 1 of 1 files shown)


liuneng1994 (approved):
LGTM

ulysses-you:
cc @zhztheplayer @PHILO-HE @zhouyuan it seems the prs in 4.12 have regression, I'm not sure which commit cause it since we are doing benchmark each day rather than each commit. Do you have time to take a look ?

zhli1142015:
I think this is related to this change: https://github.com/facebookincubator/velox/pull/9079 , for some cases the improvement in probe is less than the overhead introduced in build side. I'm still looking this, sorry for inconvenience.

ulysses-you:
thank you @zhli1142015 , looking forward to the improvement

PR #5395, opened 2024-04-13
[VL] fix bug of string buffer size calculation in shuffle

FelixYBW:
Flatten buffer doesn't update vector<stringbuffers>, so we can't use the vector to calculate the string buffer size. Instead to get the strlength from stringview and nullbit.


FelixYBW:
It's a full stringview scan for each binary column. But it's sequential access. Need to check the penalty in future.

marin-ma (approved):
Thanks!

Issue #4836, opened 2024-03-02
[VL] support partial window in Spark 3.5


zhouyuan:
### Description

It's very similar with Window
https://issues.apache.org/jira/browse/SPARK-37099

ayushi-agarwal:
@zhouyuan Is someone working on this, if not I can work on this.

zhouyuan:
I know @Jkself may have some plans. @JKself please suggest

Thanks -yuan

________________________________

JkSelf:
@ayushi-agarwal I have done some research before, and can map the `RankLimit `operator in Spark to the `TopNRowNumber `operator in Velox. However, I am currently busy with other things and have not started the relevant implementation. If you have time, you are welcome to help support this feature.

ayushi-agarwal:
@JkSelf There are three functions, row_number, rank and dense_rank which are supported by WindowGroupLimit operator in spark. We can map row_number to TopNRowNumber but to support other two it would need more changes.
There can be two approaches:
1. To convert this WindowGroupLimit to Window plus Filter node. This won't require any change in velox.
2. Modify TopNRowNumber or create two more operators to support rank and dense_rank. This would require change in velox.
Want to know your input on this that which approach would be better?

JkSelf:
@ayushi-agarwal Apologies for the delayed response. If method 1 is feasible, I would prefer method 1. Regarding method 1, I have a question I would like to ask: It seems that the Window operator in Velox does not have an option for partial mode. If that is the case, are we unable to map to Window + Filter without modifying the code? Correct me if wrong understand. Thanks.

ayushi-agarwal:
@JkSelf The plan after optimization in spark looks like Plan -> Sort -> WindowGroupLimit partial -> Exchange -> Sort -> WindowGroupLimit final -> WindowExec -> Filter.
In Gluten this can be converted to Plan -> Sort -> Window -> Filter -> Exchange -> Sort -> Window -> Filter. Mode will not be required in this case. Please let me know if my understanding is not correct or if there is any issue with this approach.

JkSelf:
@ayushi-agarwal For the Window + Filter in the Gluten plan, can it be understood that after the Window operation is completed, the Filter selects the top n records from the limit? If that's the case, do we still need to calculate all the results within the Window and then filter out the final results in the Filter? If so, the outcome would be correct, but it likely wouldn't result in a performance improvement.

ayushi-agarwal:
[...]

PR #5398, opened 2024-04-14
[GLUTEN-4836][VL]Add support for WindowGroupLimitExec in gluten

ayushi-agarwal:

There is a new operator named WindowGroupLimitExec introduced in spark 3.5 for optimizing window with rank like function with filter on it. https://github.com/apache/spark/pull/38799. This PR maps this operator with row number function to TopNRowNumberNode in velox. Currently velox doesn't support TopN for rank and dense_rank. There is an issue opened in velox for it https://github.com/facebookincubator/velox/discussions/9404. Once that support gets added in velox this new operator in Gluten can be extended to support those.

(Partially Fixes: \#4836)

## How was this patch tested?

Added new UT

### DIFF
--- a/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
+++ b/gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
@@ -328,6 +328,15 @@ message WindowRel {
   }
 }
 
+message WindowGroupLimitRel {
+  RelCommon common = 1;
+  Rel input = 2;
+  repeated Expression partition_expressions = 3;
+  repeated SortField sorts = 4;
+  int32 limit = 5;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
+
 // The relational operator capturing simple FILTERs (as in the WHERE clause of SQL)
 message FilterRel {
   RelCommon common = 1;
@@ -495,6 +504,7 @@ message Rel {
     GenerateRel generate = 17;
     WriteRel write = 18;
     TopNRel top_n = 19;
+    WindowGroupLimitRel windowGroupLimit = 20;
   }
 }
 
--- a/gluten-core/src/main/scala/org/apache/gluten/execution/WindowGroupLimitExecTransformer.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/execution/WindowGroupLimitExecTransformer.scala
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gluten.execution
+
+import org.apache.gluten.backendsapi.BackendsApiManager
+import org.apache.gluten.expression.{ConverterUtils, ExpressionConverter}
+import org.apache.gluten.extension.ValidationResult
+import org.apache.gluten.metrics.MetricsUpdater
+import org.apache.gluten.substrait.`type`.TypeBuilder
+import org.apache.gluten.substrait.SubstraitContext
+import org.apache.gluten.substrait.extensions.ExtensionBuilder
+import org.apache.gluten.substrait.rel.{RelBuilder, RelNode}
+
+import org.apache.spark.sql.catalyst.expressions.{Ascending, Attribute, Expression, SortOrder}
+import org.apache.spark.sql.catalyst.plans.physical.{AllTuples, ClusteredDistribution, Distribution, Partitioning}
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.window.{Final, Partial, WindowGroupLimitMode}
+
+import io.substrait.proto.SortField
+
+import scala.collection.JavaConverters._
+
+case class WindowGroupLimitExecTransformer(
+    partitionSpec: Seq[Expression],
+    orderSpec: Seq[SortOrder],
+    rankLikeFunction: Expression,
+    limit: Int,
+    mode: WindowGroupLimitMode,
+    child: SparkPlan)
+  extends UnaryTransformSupport {
+
+  @transient override lazy val metrics =
+    BackendsApiManager.getMetricsApiInstance.genWindowTransformerMetrics(sparkContext)
+
+  override protected def withNewChildInternal(newChild: SparkPlan): SparkPlan =
+    copy(child = newChild)
+
+  override def metricsUpdater(): MetricsUpdater =
+    BackendsApiManager.getMetricsApiInstance.genWindowTransformerMetricsUpdater(metrics)
+
+  override def output: Seq[Attribute] = child.output
+
+  override def requiredChildDistribution: Seq[Distribution] = mode match {
+    case Partial => super.requiredChildDistribution
+    case Final =>
+      if (partitionSpec.isEmpty) {
+        AllTuples :: Nil
+      } else {
+        ClusteredDistribution(partitionSpec) :: Nil
+      }
+  }
+
+  override def requiredChildOrdering: Seq[Seq[SortOrder]] = {
+    if (BackendsApiManager.getSettings.requiredChildOrderingForWindowGroupLimit()) {
+      // Velox StreamingTopNRowNumber need to require child order.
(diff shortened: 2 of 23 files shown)


ayushi-agarwal:
@JkSelf Could you please help with review?

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -319,6 +319,15 @@ message WindowRel {
   }
 }
 
+message WindowGroupLimitRel {
```
JkSelf, line 322: @ayushi-agarwal It seems that the only difference between WindowGroupLimitRel and WindowRel is the "limit" parameter? If that's the case, can we simply use WindowRel and pass the "limit" parameter through advanced_extension? This would prevent the introduction of a new operator in the Substrait code.

gluten-core/src/main/scala/org/apache/gluten/execution/WindowGroupLimitExecTransformer.scala
```diff
@@ -0,0 +1,158 @@
...
+      if (partitionSpec.isEmpty) {
+        AllTuples :: Nil
+      } else {
+        ClusteredDistribution(partitionSpec) :: Nil
+      }
+  }
+
+  override def requiredChildOrdering: Seq[Seq[SortOrder]] = {
+    if (BackendsApiManager.getSettings.requiredChildOrderingForWindowGroupLimit()) {
+      // Velox StreamingTopNRowNumber need to require child order.
```
JkSelf, line 69: @ayushi-agarwal In Velox, the TopNRowNumber here will reorder the input channels based on the passed partition key and order by key [here](https://github.com/facebookincubator/velox/blob/main/velox/exec/TopNRowNumber.cpp#L23). Do we still need to set the sorting order in scala side? Moreover, it seems that this method is not used anywhere; can we remove it? If it is indeed necessary, can we construct it on the Velox native side based on the partition key and order by key?

JkSelf (commented):
@ayushi-agarwal Thanks for your great work.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
ayushi-agarwal: @JkSelf For now we are only supporting row number but later when we support rank and dense rank, then I was thinking that we would need an extra parameter for passing the function name, so I created the new operator. Shall that info also be passed in advanced_extension?

gluten-core/src/main/scala/org/apache/gluten/execution/WindowGroupLimitExecTransformer.scala
ayushi-agarwal: In this case it will go in else part as TopNRowNumber does the sorting based on partition and order keys passed. But as spark already adds a sort operator, we don't need sorting to be done by TopNRowNumber operator as you mentioned it in point 3 here https://github.com/apache/incubator-gluten/issues/4836#issuecomment-2041285118. In that case we will need the else part. Please let me know if I understood it incorrectly.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
JkSelf: @ayushi-agarwal Yes, it can be included in the advanced_extension. However, it appears that the WindowGroupLimitRel is intended to map to the TopNRowNumber operator, not the Window operator. Therefore, it make sense to introduce a new operator in Substrait for this purpose. @zhztheplayer Do you have any input?

zhztheplayer: > Therefore, it make sense to introduce a new operator in Substrait for this purpose.

I am inclined to this approach, like adding a `TopNRowNumberRel` or something in proto. 1:1 mapping from Velox's operator to Substrait / Gluten operator is definitely encouraged, like discussed in https://github.com/apache/incubator-gluten/pull/5409#discussion_r1566884037.

zhztheplayer: To make things fancier, we can even add a `TopNRowNumberExec` in `backends-velox` module then use a rule to convert Spark's `WindowGroupLimitExec` conditionally to `TopNRowNumberExec`. That would be the best method to solve this kind of operator-mapping problems.

However if you think `WindowGroupLimitExec` and `TopNRowNumberRel` doesn't have much semantic difference apart of the window function types they support, then we can do it without rules and it's also fine.

JkSelf:
@ayushi-agarwal Can you help to resolve the conflicts? Thanks.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
ayushi-agarwal: I kept it as WindowGroupLimitRel and not TopNRowNumberRel so that later we can use the same for rank and denseRank.

ayushi-agarwal:
> @ayushi-agarwal Can you help to resolve the conflicts? Thanks.

@JkSelf I have updated it.

JkSelf:
@ayushi-agarwal Can you help to fix the following compile error in clickhouse backend?

```
17:47:31  [INFO] --- scala-maven-plugin:4.8.0:compile (scala-compile-first) @ backends-clickhouse ---
17:47:31  [WARNING] joda-time:joda-time/maven-metadata.xmlfailed to transfer from http://0.0.0.0/ during a previous attempt. This failure was cached in the local repository and resolution will not be reattempted until the update interval of maven-default-http-blocker has elapsed or updates are forced. Original error: Could not transfer metadata joda-time:joda-time/maven-metadata.xml from/to maven-default-http-blocker (http://0.0.0.0/): transfer failed for http://0.0.0.0/joda-time/joda-time/maven-metadata.xml
17:47:31  [INFO] /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/backends-clickhouse/src/main/java:-1: info: compiling
17:47:31  [INFO] /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/backends-clickhouse/src/main/delta-20:-1: info: compiling
17:47:31  [INFO] /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/backends-clickhouse/src/main/scala:-1: info: compiling
17:47:31  [INFO] Compiling 122 source files to /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/backends-clickhouse/target/scala-2.12/classes at 1713779251191
17:47:41  [ERROR] /home/jenkins/agent/workspace/gluten/gluten-ci/gluten2/backends-clickhouse/src/main/scala/org/apache/spark/sql/delta/catalog/ClickHouseTableV2.scala:312: error: not found: type Expression
17:47:41  [ERROR]       filterExprs: Seq[Expression]): Seq[InputPartition] = {
17:47:41  [ERROR]                        ^
17:47:41  [ERROR] /home/jenkins/agent/workspace/gluten/gluten-ci/gluten/backends-clickhouse/src/main/scala/org/apache/spark/sql/delta/catalog/ClickHouseTableV2.scala:312: error: not found: type Expression
17:47:41  [ERROR]       filterExprs: Seq[Expression]): Seq[InputPartition] = {
17:47:41  [ERROR]                        ^
17:47:41  [ERROR] one error found
17:47:41  [ERROR] one error found
17:47:41  [ERROR] exception compilation error occurred!!!
```

JkSelf (approved):
LGTM. Thanks for your great work.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -495,6 +504,7 @@ message Rel {
     GenerateRel generate = 17;
     WriteRel write = 18;
     TopNRel top_n = 19;
+    WindowGroupLimitRel windowGroupLimit = 20;
```
EpsilonPrime, line 507: Substrait already has a field #20.  This means that your Substrait is not going to be compatible with any other Substrait consumers or tools.  The Substrait project is open to adding new relations -- all it takes is a PR, issue, or email to start the discussion.

acvictor: @EpsilonPrime can you explain what you mean and if it's still an issue? Does any change to algebra.proto need to be updated in the Substrait project too?

EpsilonPrime: Right now the this copy differs in about a half dozen ways from the original project.  A protobuf saved using this version will be loaded incorrectly since DdlRel has field number 20.  What should happen is that the differences introduced here get applied back to the main project.  I've started the process of the CSV text format there.  The Substrait project introduced ConsistentPartitionWindowRel a while back (as field number 17) which may actually do what you want here.

As long as you only talk to other consumers using this version of Substrait your code will work.  But you're missing out on the other tools.  For instance, the Substrait Validator is great for checking that you've constructed a conforming plan.  I run all of the plans generated by my end to end tests through it and catches issues all the time.

acvictor: Got it, I'm working on a PR that also touches algebra.proto - https://github.com/apache/incubator-gluten/pull/5632/files#diff-632cd96e81af6938b453cb8ec69d66123284f17c6e47f3f16b8f08e8046afc39. Can the reconciliation happen in one shot? I can try to port these changes over with your guidance!

Issue #5405, opened 2024-04-15
[CH] Optimze todate function from bi query


loneylee:
### Description

If users query data through BI tools.
 The BI tools may generate SQL similar to
 ```
to_date(
   from_unixtime(
     unix_timestamp(stringType, 'yyyyMMdd')
   )
 )
```
 to convert string strings to dates.
 Under ch backend, the StringType can be directly converted into DateType,
     and the functions `from_unixtime` and `unix_timestamp` can be optimized here.
Optimized result is 
```
to_date(stringType)
```

PR #5406, opened 2024-04-15
[GLUTEN-5405][CH] Add rewrite todate function

loneylee:

If users query data through BI tools.
The BI tools may generate SQL similar to
```
to_date(
  from_unixtime(
    unix_timestamp(stringType, 'yyyyMMdd')
  )
 )
```
to convert string strings to dates.
Under ch backend, the StringType can be directly converted into DateType,
    and the functions `from_unixtime` and `unix_timestamp` can be optimized here.
Optimized result is `to_date(stringType)` by use
```
spark.gluten.sql.rewrite.dateConversion=true
```

Notice:
When this parameter is turned on, there may be inconsistent results, because ch is compatible with a lot of date parsing, and when format is yyyy-MM-dd , spark returns null, but ch parses normally.

(Fixes: \#5405)

## How was this patch tested?

Test by ut

| OptimizeRuleBenchmark    | Best Time(ms) | Avg Time(ms) | Stdev(ms) | Rate(M/s) | Per Row(ns) | Relative |
| -------- | ------- | ------- | ------- | ------- | ------- | ------- |
| ClickHouse rewrite dateConversion: false  | 6597 | 7576 | 1278 | 0.0 | 659716914.8 | 1.0X| 
| ClickHouse rewrite dateConversion: true | 3780 | 3865 | 98 | 0.0 | 378002746.3 | 1.7X |

### DIFF
--- a/backends-clickhouse/src/test/scala/org/apache/spark/sql/execution/benchmarks/CHOptimizeRuleBenchmark.scala
+++ b/backends-clickhouse/src/test/scala/org/apache/spark/sql/execution/benchmarks/CHOptimizeRuleBenchmark.scala
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.execution.benchmarks
+
+import org.apache.spark.benchmark.Benchmark
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.execution.benchmark.SqlBasedBenchmark
+
+object CHOptimizeRuleBenchmark extends SqlBasedBenchmark with CHSqlBasedBenchmark {
+
+  protected lazy val appName = "CHOptimizeRuleBenchmark"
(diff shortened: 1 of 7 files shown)


lgbo-ustc:
Is it similar to https://github.com/apache/incubator-gluten/pull/4617?

loneylee:
> similar

Not the same. This pr is to cancel two conversions of timestamp.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1588,6 +1591,16 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val ENABLE_REWRITE_DATE_CONVERSION =
+    buildConf("spark.gluten.sql.rewrite.dateConversion")
+      .internal()
+      .doc(
+        "Rewrite the conversion between date and string."
```
taiyang-li, line 1598: It is useful for velox backend?

backends-clickhouse/src/test/scala/org/apache/spark/sql/execution/benchmarks/CHOptimizeRuleBenchmark.scala
```diff
@@ -0,0 +1,96 @@
...
+
+    parquetReadBenchmark.addCase(s"ClickHouse rewrite dateConversion: true", executedCnt) {
+      _ => testToDateOptimize(parquetDir)
+    }
+
+    parquetReadBenchmark.run()
+  }
+
+  def testToDateNotOptimize(parquetDir: String): Unit = {
+    withSQLConf(("spark.gluten.sql.rewrite.dateConversion", "false")) {
```
taiyang-li, line 64: duplicated codes in testToDateNotOptimize and testToDateOptimize, better to reuse.

loneylee: fix it

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
loneylee: Not yet, beacuse this behavior may cause an inconsistent results. I have already change the config name to backend.ch.xx

taiyang-li (approved):
LGTM

PR #5408, opened 2024-04-15
[VL] Support collect_list in window

liujiayi771:

Support collect_list in window.

## How was this patch tested?

Add new test case.


liujiayi771:
cc @PHILO-HE, thanks.

The `collect_set` function still has an issue with null values, and because `windowExec` does not have `resultExpressions`, it cannot handle the conversion of null to an empty list through rewrite `resultExpressions`. This will be supported in the future when implementing the new `collect_set` aggregate function in Velox.

PHILO-HE (approved):
Thanks!

PR #5409, opened 2024-04-15
[VL] Add independent operator for top-n processing in TakeOrderedAndProjectExecTransformer

zhztheplayer:
This PR adds a new physical operator TopNTransformer in Velox backend mapping to Velox's `core::TopNNode`.

By doing this, we can remove the hacky limit + sort coherent validation code, and the proxied executed plan used by `TakeOrderedAndProjectExecTransformer` will correctly map to the actual physical plan in Velox task.

This PR is a phase 1 because the actual Velox plan is still hidden by `TakeOrderedAndProjectExecTransformer` although limit+sort was replaced by top-n. We may eventually deprecate `TakeOrderedAndProjectExecTransformer` and expose more execution details to query planner.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/TopNTransformer.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/TopNTransformer.scala
@@ -26,7 +26,7 @@ import org.apache.gluten.substrait.extensions.ExtensionBuilder
 import org.apache.gluten.substrait.rel.{RelBuilder, RelNode}
 
 import org.apache.spark.sql.catalyst.expressions.{Attribute, SortOrder}
-import org.apache.spark.sql.catalyst.plans.physical.{Partitioning, SinglePartition}
+import org.apache.spark.sql.catalyst.plans.physical.{AllTuples, Distribution, Partitioning, SinglePartition, UnspecifiedDistribution}
 import org.apache.spark.sql.catalyst.util.truncatedString
 import org.apache.spark.sql.execution.SparkPlan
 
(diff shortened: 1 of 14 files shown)


zhztheplayer:
@rui-mo @ulysses-you thanks

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -268,6 +268,15 @@ message FetchRel {
...
 }
 
+// The relational operator representing TOP N calculation
+message TopNRel {
+  RelCommon common = 1;
+  Rel input = 2;
+  int64 n = 3;
+  repeated SortField sorts = 4;
+  substrait.extensions.AdvancedExtension advanced_extension = 10;
+}
```
rui-mo, line 278: Could you document this change?
https://github.com/apache/incubator-gluten/blob/main/docs/developers/SubstraitModifications.md

backends-velox/src/main/scala/org/apache/gluten/execution/TopNTransformer.scala
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.spark.sql.catalyst.expressions.{Attribute, SortOrder}
+import org.apache.spark.sql.catalyst.plans.physical.{AllTuples, Distribution, Partitioning, UnspecifiedDistribution}
+import org.apache.spark.sql.catalyst.util.truncatedString
+import org.apache.spark.sql.execution.SparkPlan
+
+import io.substrait.proto.SortField
+
+import scala.collection.JavaConverters._
+
+case class TopNTransformer(
```
rui-mo, line 37: I remember a transformer aims to convert Spark plan as Substrait plan. As Spark does not keep a TopN operator, I'm wondering why do we need a TopNTransformer. With this change, it seems we are corresponding with Velox operator instead of Spark. How do you think?

rui-mo (commented):
Thanks!

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
JkSelf: Could we utilize the SortRel operator to implement TopNRel, incorporating the 'n' parameter within the advanced_extension?

backends-velox/src/main/scala/org/apache/gluten/execution/TopNTransformer.scala
zhztheplayer: > I remember a transformer aims to convert Spark plan as Substrait plan. As Spark does not keep a TopN operator, I'm wondering why we need a TopNTransformer. With this change, it seems we are corresponding with Velox operator instead of Spark. How do you think?

This is what I was thinking about either. My opinion is that we have some debatable design from the very beginning.

Physical plan is basically an abstraction for the actual query execution procedure. The planner / optimizer tend to obtain more information form physical plan for better optimization. A simplest example, optimizer may place exchanges to alter data distribution in the middle of two physical operators on-demand. Top-N can also be an example, if optimizer can distinguish between OrderBy+Limit and TopN, it could optimize the former to the latter when it's needed. Say if N in TopN is abnormally large and TopN algorithm may not help much in that case, optimizer can decide not converting OrderBy+Limit into it.

We have limited capability to extend our physical plan if we don't extend Substrait or using upstream Substrait. I am inclined that in long term we should not follow our previous criteria and allow developers to extend Gluten's Substrait proto whenever it's needed - then the physical plan can carry more execution detail and create more potential to optimize.

In the case of we want to keep compatibility with Substrait ecosystem, we'd have a new criteria that when we extend Gluten's Substrait proto file, it's recommended to add new operators instead of modifying the old one. By doing this if we want to support a new backend that compatible to upstream Substrait, we still can use the same proto definition for it. Indeed, better solutions may include one Substrait proto (or whatever Spark physical plan serde you like) per backend, but that may require for more refactor work.

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
zhztheplayer: Refer to this comment https://github.com/apache/incubator-gluten/pull/5409#discussion_r1566884037, let's make a new operator for better compatibility with other potential Substrait implementers.

rui-mo (approved):

Issue #5341, opened 2024-04-09
UT failures for Spark3.5


yma11:
### Backend

VL (Velox)

### Bug description
Note that following list doesn't include the possible failures for the UTs newly added Spark3.5. Will update the list once them collected.
- [x] generate hash join plan - v2
- [ ] test write parquet with compression codec
- [ ] regr_r2
- [ ] regr_slope
- [ ] regr_intercept
- [ ] regr_sxy
- [x] input row
- [x] test hive static partition write table
- [x] test hive write table
- [x] test hive write dir

- [x] column mapping mode = id
- [x] column mapping mode = name
- [x] delta: time travel
- [x] delta: partition filters
- [x] basic test with stats.skipping disabled
- [x] column mapping with complex type

- [x] iceberg bucketed join
- [x] iceberg bucketed join with partition
- [x] iceberg bucketed join with partition filter

- [x] GlutenExpressionMappingSuite
- [x] SPARK-26708 Cache data and cached plan should stay consistent
- [x] SPARK-46590 adaptive query execution works correctly with broadcast join and union
- [x] SPARK-46590 adaptive query execution works correctly with cartesian join and union
- [x] SPARK-40618: Regression test for merging subquery bug with nested subqueries
- [x] SPARK-40615: Check unsupported data type when decorrelating subqueries
- [x] SPARK-40618: Regression test for merging subquery bug with nested subqueries

- [x] Gluten - test for FAILFAST parsing mode
- [x] SPARK-36612: Support left outer join build left or right outer join build right in shuffled hash join
- [x] A cached table preserves the partitioning and ordering of its cached SparkPlan
- [ ] SPARK-42782: Hive compatibility check for get_json_object
- [x] SPARK-43876: Enable fast hashmap for distinct queries
- [x] SPARK-41896: Filter on row_index and a stored column at the same time
- [x] SPARK-43450: Filter on full _metadata column struct
- [x] SPARK-43450: Filter on aliased _metadata.row_index

- [x] SPARK-45171: Handle evaluated nondeterministic expression (VeloxRuntimeError: the key in unnest Operator only support field)

### Spark version

None

FelixYBW:
does it cover all the UT failures of 3.5? If so let's pin to top

yma11:
> does it cover all the UT failures of 3.5? If so let's pin to top

Not quite, there should some Spark3.5 specific UTs(newly added in Spark3.5) we haven't ported yet.

yma11:
[...]

PR #5425, opened 2024-04-16
[VL][UT] Fix scalar-subquery-select.sql in spark35

liujiayi771:

Currently in the gluten-ut of spark35, `scalar-subquery-select.sql` and `scalar-subquery-select.sql.out` were directly copied from spark34. However, these two files have changed in spark 3.5 version and cannot be used as-is from spark34. They need to be modified to match the spark 3.5 version.

## How was this patch tested?

N/A


liujiayi771:
cc @yma11, thanks.

yma11 (approved):

zhouyuan:
related: https://github.com/apache/incubator-gluten/issues/5341

zhouyuan:
@liujiayi771 thanks for taking care of this, did you check other files? Is this the only change in SQLQuerySuite?

thanks,
-yuan

liujiayi771:
@zhouyuan For the overwrite sql, this file is the only change.

PR #5433, opened 2024-04-17
[VL] Add a bad test case when bloom_filter_agg is fallen back while might_contain is not

zhztheplayer:
So far the case fails with error

```
org.apache.gluten.exception.GlutenException: java.lang.RuntimeException: Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: (1 vs. 0)
Retriable: False
Expression: kBloomFilterV1 == version
Function: merge
../.././velox/common/base/BloomFilter.h
Line: 67
Stack trace:
# 0  std::shared_ptr<facebook::velox::VeloxException::State const> facebook::velox::VeloxException::State::make<facebook::velox::VeloxException::make(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, facebook::velox::VeloxException::Type, std::basic_string_view<char, std::char_traits<char> >)::{lambda(auto:1&)#1}>(facebook::velox::VeloxException::Type, facebook::velox::VeloxException::make(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, facebook::velox::VeloxException::Type, std::basic_string_view<char, std::char_traits<char> >)::{lambda(auto:1&)#1})
# 1  facebook::velox::VeloxException::VeloxException(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, facebook::velox::VeloxException::Type, std::basic_string_view<char, std::char_traits<char> >)
# 2  facebook::velox::VeloxUserError::VeloxUserError(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, std::basic_string_view<char, std::char_traits<char> >)
[...]

PR #5435, opened 2024-04-17
[VL] Rework co-fallback mechanism of bloom-filter might_contain/agg

zhztheplayer:
Velox's bloom-filter agg/filter functions are logically different with Spark's version. This makes their resident Gluten/Spark Filter/Aggregate operators logically different with Spark's version. For such logical differences, we should use different functions to distinguish between implementations rather than reusing Spark's function type in Velox backend.

Patch incorporates:

1. Fix https://github.com/apache/incubator-gluten/pull/5433.
2. Add `VeloxBloomFilterMightContain` / `VeloxBloomFilterAggregate`.
3. Add `VeloxBloomFilter` inheriting Spark's `BloomFilter` to make sure fallen-back`velox_might_contain` expression work correctly. Should support fallen-back `velox_bloom_filter_agg` in future.
4. Remove rule `FallbackBloomFilterAggIfNeeded` (https://github.com/apache/incubator-gluten/pull/3994).
5. Add logical rule `BloomFilterMightContainJointRewriteRule` in Velox backend module only, to convert  `BloomFilterMightContain` / `BloomFilterAggregate` to `VeloxBloomFilterMightContain` / `VeloxBloomFilterAggregate` when Gluten is enabled.

The patch makes the relevant code safer than before, since we'll have explicit function pair of  `BloomFilterMightContain` / `BloomFilterAggregate` and `VeloxBloomFilterMightContain` / `VeloxBloomFilterAggregate`. Thus we can easierly detect a mismatch between agg/filter function by checking their function types, rather than failing the query execution at runtime.


zhztheplayer:
Example Velox bf with vanilla code-gen

```java
public Object generate(Object[] references) {
return new GeneratedIteratorForCodegenStage1(references);
}

/*wsc_codegenStageId*/
final class GeneratedIteratorForCodegenStage1 extends org.apache.spark.sql.execution.BufferedRowIterator {
private Object[] references;
private scala.collection.Iterator[] inputs;
private scala.collection.Iterator localtablescan_input_0;
private org.apache.spark.sql.catalyst.expressions.codegen.UnsafeRowWriter[] filter_mutableStateArray_1 = new org.apache.spark.sql.catalyst.expressions.codegen.UnsafeRowWriter[1];
private org.apache.spark.util.sketch.VeloxBloomFilter[] filter_mutableStateArray_0 = new org.apache.spark.util.sketch.VeloxBloomFilter[1];

public GeneratedIteratorForCodegenStage1(Object[] references) {
this.references = references;
}

public void init(int index, scala.collection.Iterator[] inputs) {
partitionIndex = index;
this.inputs = inputs;
localtablescan_input_0 = inputs[0];

filter_mutableStateArray_1[0] = new org.apache.spark.sql.catalyst.expressions.codegen.UnsafeRowWriter(1, 0);

filter_mutableStateArray_0[0] = org.apache.spark.util.sketch.VeloxBloomFilter.readFrom(((byte[]) references[2] /* bloomFilterData */));
}

protected void processNext() throws java.io.IOException {
while ( localtablescan_input_0.hasNext()) {
InternalRow localtablescan_row_0 = (InternalRow) localtablescan_input_0.next();
((org.apache.spark.sql.execution.metric.SQLMetric) references[0] /* numOutputRows */).add(1);
do {
long localtablescan_value_0 = localtablescan_row_0.getLong(0);

boolean filter_isNull_0 = false;
boolean filter_value_0 = false;
if (!filter_isNull_0) {
filter_value_0 = filter_mutableStateArray_0[0].mightContainLong((Long)localtablescan_value_0);
}
if (filter_isNull_0 || !filter_value_0) continue;

((org.apache.spark.sql.execution.metric.SQLMetric) references[1] /* numOutputRows */).add(1);

filter_mutableStateArray_1[0].reset();

filter_mutableStateArray_1[0].write(0, localtablescan_value_0);
append((filter_mutableStateArray_1[0].getRow()));

} while(false);
if (shouldStop()) return;
}
}

}
```

backends-velox/src/main/scala/org/apache/spark/sql/catalyst/BloomFilterMightContainJointRewriteRule.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.expressions.VeloxBloomFilterMightContain
+import org.apache.spark.sql.catalyst.expressions.aggregate.VeloxBloomFilterAggregate
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+
+case class BloomFilterMightContainJointRewriteRule(spark: SparkSession) extends Rule[LogicalPlan] {
+  override def apply(plan: LogicalPlan): LogicalPlan = {
+    if (
+      !(GlutenConfig.getConf.enableNativeBloomFilter &&
```
zhli1142015, line 31: do we fix the failed test by this check?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/ListenerApiImpl.scala
```diff
@@ -45,12 +46,16 @@ class ListenerApiImpl extends ListenerApi {
...
     initialize(conf)
   }
 
   override def onDriverShutdown(): Unit = shutdown()
 
-  override def onExecutorStart(conf: SparkConf): Unit = initialize(conf)
+  override def onExecutorStart(conf: SparkConf): Unit = {
+    UDFResolver.resolveUdfConf(conf, isDriver = false)
+    initialize(conf)
+  }
```
zhztheplayer, line 58: Moved UDF resolution to this file

@marin-ma

backends-velox/src/main/scala/org/apache/spark/sql/catalyst/BloomFilterMightContainJointRewriteRule.scala
zhztheplayer: So far yes. 

I'll raise another PR at a good time to support fallen-back `velox_bloom_filter_agg` then we can remove the `enableColumnarHashAgg` check safely.

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/GlutenBloomFilterAggregateQuerySuite.scala
```diff
@@ -106,11 +107,52 @@ class GlutenBloomFilterAggregateQuerySuite
           assert(
             collectWithSubqueries(df.queryExecution.executedPlan) {
               case h if h.isInstanceOf[HashAggregateExecBaseTransformer] => h
-            }.size == 0,
+            }.size == 2,
```
marin-ma, line 110: Could you document this change?

zhztheplayer: No need to add documents. Will change the test name to `Test bloom_filter_agg offloaded with filter fallen back` to clarify in next patch

PR #5437, opened 2024-04-17
[GLUTEN-5434] Add script for parquet read fuzzer test

yma11:

Leverage different version of parquet-mr to generate parquet files with random data and then verify the data using Gluten UT

## How was this patch tested?

N/A

### DIFF
--- a/tools/parquet-read/data-generate-v1.11.sh
+++ b/tools/parquet-read/data-generate-v1.11.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# install dependencies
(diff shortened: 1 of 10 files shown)


zhouyuan:
CC @marin-ma

marin-ma:
Thanks! Not sure if this fuzzer test can cover many cases. Looks like the data generator only generates some simple data types https://github.com/apache/parquet-mr/blob/master/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/DataGenerator.java#L156-L165
But issues like #9242 can be related to complex data types.

yma11:
> Thanks! Not sure if this fuzzer test can cover many cases. Looks like the data generator only generates some simple data types https://github.com/apache/parquet-mr/blob/master/parquet-benchmarks/src/main/java/org/apache/parquet/benchmarks/DataGenerator.java#L156-L165 But issues like [facebookincubator/velox#9242](https://github.com/facebookincubator/velox/issues/9242) can be related to complex data types.

Yes. The parquet-mr DataGenerator only cover simple data types as it's used for benchmark test. Complex type support is a gap. Parquet read related issues are mainly about 1) legacy format/schema that velox haven't fully covered  and 2) complex type reading if the schema can be recognized correctly, especially null/empty values. For 1), we can use script in this PR as it will generate files with different schemas as different parquet version used. We may can add more options to control generating more kinds of schemas. For 2), we can add complex type generation in it and try upstream or we can leverage the data generation in Gluten to cover null/empty values. Both are fine. What do you think?

marin-ma:
> For 1), we can use script in this PR as it will generate files with different schemas by different parquet version used. 

Thanks for your explanation! Just one more question. Looks like the DataGenerator always generates fixed schema with random data. After adding complex types in the `DataGenerator`, will it still generate a fixed schema or we plan to support generating random schemas?

yma11:
@zhouyuan @FelixYBW @marin-ma Scripts for building parquet read fuzzer job are all here. Please help take a review. I will need monitor the job for some days and collect initial issues. We will refine the data generator to increase more coverage. Currently it by default use parquet v2 protocol so the main problem is an issue about new v2 encoding support.

FelixYBW:
Let's run this on our internal server only

Issue #5438, opened 2024-04-17
Dynamically sizing off-heap memory


supermem613:
### Description

When using Gluten with Velox and Spark, today we specify the off-heap memory size and accordingly adjust the on-heap memory as well. In practice, this means that the amount of memory we set aside for on-heap cannot be used for off-heap and vice-versa, which can lead to situations where we are not optimally using the machine's memory since we may be doing processing mostly using on-heap or off-heap memory, but rarely both at the same time in great quantities.

This is particularly painful, for example, when we fall back execution to "vanilla" Spark.

For example, for a 64GB machine where we want to use 56GB of memory for Spark, we would set on-heap memory (via the spark.executor.memory setting) to, say, 14GB and set the off-heap (via the spark.memory.offHeap.size) to 42GB. In this case, if we fallback execution to Spark, we will be constrained by the 14GB of on-heap memory. If we don't fall back, we are using up to 42GB, leaving a fair number of unused GBs of memory that could be used.

We propose to leverage the existing off-heap allocation tracking in Gluten, paired with JDK APIs (Runtime.getRuntime().totalMemory() and freeMemory() APIs) that show on-heap utilization to provide unified memory managed utilization control. However, it is important to notice that this approach does not actively control Java allocations, so it can in practice allow some over subscription of memory to happen until a native allocation comes along and is failed accordingly.

From a configuration perspective, there will be a new gluten Boolean configuration to turn on this new feature, which in turn obviates any off-heap configuration. This means that the setting for off-heap enabling and sizing will no longer be used. Instead, we will continue to configure the executor memory – the on-heap sizing – to use as much memory as possible, as is done today with "vanilla" Spark.

zhli1142015:
cc @FelixYBW , thanks.

Issue #3055, opened 2023-09-07
[VL] Use mixed memory (off-heap and on-heap) for native


Yohahaha:
### Description

Gluten only use off-heap memory for now, and I see some PR want extend spill with spill to heap. However, we could use mixed memory which means use off-heap first then use on-heap when off-heap will OOM to utilize memory as much as possible.

zhztheplayer:
This is a big topic but indeed very helpful to the project. So it would be good if you can elaborate on the design if you have some, and update the PR description to make it more concrete.

What I could imagine is that we can hack into Spark's memory manager, then dynamically borrow some heap caps when JVM heap is shrinking to a number below `-Xmx`, when G1 GC or Parallel GC are used. CMS GC doesn't support heap shrinking AFAIK.

Hacking to deeper level of JVM is also a potential solution, but that may require some techniques like `heap-pining` of Hotspot JVM which seems to be a very old feature in its earlier versions. I am not sure whether the feature works well in JDK 8+ but probably we can do some path-findings around it anyway.

Yohahaha:
I'm not ready to submit a PR, just use issue to record the idea.

Yohahaha:
> What I could imagine is that we can hack into Spark's memory manager, then dynamically borrow some heap caps when JVM heap is shrinking to a number below -Xmx, when G1 GC or Parallel GC are used. CMS GC doesn't support heap shrinking AFAIK.

What if we declare a MemoryConsumer with ON_HEAP mode? just report to Spark how much memory we need, but not do real allocation in JVM.

zhztheplayer:
> > What I could imagine is that we can hack into Spark's memory manager, then dynamically borrow some heap caps when JVM heap is shrinking to a number below -Xmx, when G1 GC or Parallel GC are used. CMS GC doesn't support heap shrinking AFAIK.
> 
> What if we declare a MemoryConsumer with ON_HEAP mode? just report to Spark how much memory we need, but not do real allocation in JVM.

To me that would be a little bit dangerous: If you don't actually allocate from JVM, it will finally full then trigger GC. During which the process RSS will go beyond the limit.

But if we can somehow keep the GCs inside young generation, the approach could be comparatively safter however still be risky if we want to control the memory limit very accurately.

Yohahaha:
Reservation and allocation are totally different things, reservation not trigger GC but allocation.
[...]

PR #5439, opened 2024-04-17
[GLUTEN-5438] feat: Dynamically sizing off-heap memory

supermem613:

Today, in Spark we specify the on-heap and off-heap memory sizes as a configuration value read at the beginning of executing a job.

With this change, we are exposing a new feature that is enabled with a new spark.gluten.memory.dynamic.offHeap.sizing.enabled setting. When this setting is configured to true, the offheap setting will be ignored in Gluten and we will size the offheap as the same size as the spark.executor.memory setting.

We will then proceed to enforcing a total memory quota, calculated by the sum of what memory is committed and in use in the Java heap (calculated with Runtime.getRuntime().totalMemory() - Runtime.GetRuntime().freeMemory()) plus the tracked off-heap memory in TreeMemoryConsumer. When there is an allocation that would tide us over this total amount of committed memory, we will fail the allocation and trigger an OOM.

Note that with this change, we perform the "quota check" when an allocation in the native engine is informed to Gluten. In practice, this means that it is possible that the Java codebase can oversubscribe memory as it allocates, which is under the on-heap quota, although there is enough off-heap usage where we should fail the allocation.

A test exercising this setting is part of this change.

Fixes: #5438 

## How was this patch tested?

Manual testing with Spark and included test


FelixYBW:
Thank you for your PR! @zhanglistar may do a test quickly

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
```diff
@@ -0,0 +1,71 @@
...
+  DynamicOffHeapSizingPolicyChecker() {}
+
+  public boolean canBorrow(long size) {
+    if (size == 0) {
+      return true;
+    }
+
+    long totalMemory = Runtime.getRuntime().totalMemory();
+    long freeMemory = Runtime.getRuntime().freeMemory();
+    long usedOnHeapBytes = (totalMemory - freeMemory);
```
zhztheplayer, line 41: Is this calculation safe enough? Say if total memory is 10G and free memory is 5G, then the JVM could still occupy 10G RAM from host machine, isn't it?

zhztheplayer (commented):
Thank you for raising up this.

Though I could think about some limitations within the implementation:

1. When heap is increasing by real JVM heap allocation, JVM will not respect Gluten's fake heap allocation then the overall RSS of the process may go beyond limit unexpectedly;
2. Spill is still only be triggered by **maximum** off-heap limit only;

I only read the code so may miss something and please add if you know some. 

And I know it's challenging to have a perfect solution for such topics so if we decide to provide this feature, I'd suggest we document the known limitations in code or doc files (can also add some experimental marks).

Also you can refer to some previous discussions if you missed them https://github.com/apache/incubator-gluten/issues/3055#issuecomment-1709439477.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
supermem613: As per https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Runtime.html, totalMemory is the "the total amount of memory currently available for current and future objects, measured in bytes", i.e., the on-heap size set for Spark via spark.executor.memory in our case. freeMemory is the "an approximation to the total amount of memory currently available for future allocated objects", i.e., memory NOT currently in use by Java objects. 

In this calculation if the on-heap was set to 10GB and 5GB is unused by any Java objects, usedOnHeapBytes would give us 5GB. The JVM memory footprint in this case is 10GB. With the calculations above, we would allow up to 5GB of memory to be used for off-heap.

Does that help?

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
```diff
@@ -0,0 +1,71 @@
...
+package org.apache.gluten.memory.memtarget.spark;
+
+import org.apache.gluten.GlutenConfig;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public final class DynamicOffHeapSizingPolicyChecker {
```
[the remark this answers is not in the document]
supermem613: Will investigate.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
supermem613: To be transparent, however, you are 100% correct that this is not a strict / perfect quota. For example, if allocations happen on Java code, the logic above doesn't kick in, so we may oversubscribe memory usage.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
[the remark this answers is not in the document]
supermem613: Can you please elaborate your suggestion a bit? The approach here leans on the policy checker being static, accounting for used off-heap bytes across all memory targets (say, if memory isolation is on). Are you suggesting a new TreeMemoryTarget or something that is an alternative to it, which presumably would have similar complexity since we have to support, say, memory isolation?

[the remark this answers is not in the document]
zhztheplayer: > Are you suggesting a new TreeMemoryTarget or something that is an alternative to it, which presumably would have similar complexity since we have to support, say, memory isolation?

We don't have to add something as heavy as TreeMemoryTarget. If I understand correctly, `DynamicOffHeapSizingPolicyChecker` was designed to be a global singleton, right? If that's true, we can make the policy checker extend API https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTarget.java#L23 and having the allocations accounted into a global static counter in policy checker.

Just like the logging memory target https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/LoggingMemoryTarget.java, it logs allocation through the static `LOGGER` object, and delegates all the allocate to the subsequent target.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
zhztheplayer: Thank for the explanation.

> In this calculation if the on-heap was set to 10GB and 5GB is unused by any Java objects, usedOnHeapBytes would give us 5GB. The JVM memory footprint in this case is 10GB. With the calculations above, we would allow up to 5GB of memory to be used for off-heap.

So in this case, the important one is that whether the total RSS of the JVM is calculated by which one of the following:

1. 10G total-heap + 5G off-heap = 15G
2. 5G used-heap + 5G off-heap = 10G

The code seems to map to the second calculation. But I am concerned that probably the first one is the real situation. If user configured Gluten/Spark with a 10G memory limit but the actual RSS goes 15G, that is unexpected.

zhztheplayer: > To be transparent, however, you are 100% correct that this is not a strict / perfect quota. For example, if allocations happen on Java code, the logic above doesn't kick in, so we may oversubscribe memory usage.

Yes, but this one is another issue that doesn't seem to be easily solved in terms of the PR's implementation. Maybe we can temporarily just document this to let user know the risk.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
```diff
@@ -0,0 +1,71 @@
...
+  private final AtomicLong usedOffHeapBytes = new AtomicLong();
+
+  DynamicOffHeapSizingPolicyChecker() {}
+
+  public boolean canBorrow(long size) {
+    if (size == 0) {
+      return true;
+    }
+
+    long totalMemory = Runtime.getRuntime().totalMemory();
```
srinivasst, line 39: Wouldn't Runtime.getRuntime().maxMemory() - Runtime.getRuntime().totalMemory() be a better estimation for bytes used for on heap? 

Although (totalMemory - freeMemory) is the bytes used by _live_ objects, totalMemory is more representative of RSS memory. 

Any allocations by native would add on top of totalMemory.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
```diff
@@ -0,0 +1,71 @@
...
+
+public final class DynamicOffHeapSizingPolicyChecker {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(DynamicOffHeapSizingPolicyChecker.class);
+  private final long maxOnHeapMemoryInBytes = GlutenConfig.getConf().onHeapMemorySize();
+  private final AtomicLong usedOffHeapBytes = new AtomicLong();
+
+  DynamicOffHeapSizingPolicyChecker() {}
+
+  public boolean canBorrow(long size) {
```
ulysses-you, line 34: This code says when native requires memory, we borrow onheap memory to it.  But then later onheap requires memory, how can onheap get the actually memory before native repay? I think It would be killed by yarn or k8s due to it uses double memory at this time. And It's a very general case, e.g., there are stages running in parallel or there are many queries running in parallel.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
supermem613: I see. Yes, the amount of memory that we are calculating as "committed but unused by the JVM", with the freeMemory API call, likely would be counted by the OSS as part of the RSS since the JVM likely touches those pages as they are added to the JVM heap free list. 

However, the counterpoint is that these pages will not see further memory access unless they are actually put to use, so if we consider them against the overall "used" quota, we are essentially leaving unused memory capacity at the table.

Perhaps a good way to approach this is to add a specific config with the proposed change that optionally allows for this more aggressive use of memory. If the config is off, say, the default, we will consider totalMemory as the total used memory, not discounting the freeMemory API value. If the setting is off, we will discount the freeMemory API value as the proposed changes stand now.

Thoughts?

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
supermem613: I believe this is the same point that zhztheplayer made above. Please see the reply to that thread.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/spark/DynamicOffHeapSizingPolicyChecker.java
supermem613: I believe you are referring to the caveat raised above that "it is possible that the Java codebase can oversubscribe memory as it allocates, which is under the on-heap quota, although there is enough off-heap usage where we should fail the allocation". Feedback was also given above that this should be documented in the code as well to be very explicit about the risk. I will do that.

ulysses-you: Oversubscribe memory is quite risk. I think we should at least respect the Spark idea. The max execution memory in Spark is:
```
(spark.executor.memory - 300MB) * spark.memory.fraction
```

`totalMemory() - freeMemory()` breaks this idea. Even if the execution offload to native, there are still many things depend on on-heap memory, e.g., the metadata(plan, split, map status, event...), broadcast block, rdd cache. One thought is that, we'd better to follow the Spark, and change it to:
```
(totalMemory() - freeMemory() - 300MB) * spark.memory.fraction
```

supermem613: Agreed. With the proposed change, we are seeing the "max on heap" as being the spark.executor.memory setting, but that does go against Spark's setting aside memory accordingly. I'll change the proposed logic to follow what Spark has determined as the max on-heap.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargets.java
```diff
@@ -54,6 +54,13 @@ public static MemoryTarget newConsumer(
...
     }
-    return factory.newConsumer(tmm, name, spillers, virtualChildren);
+
+    final MemoryTarget memoryTarget = factory.newConsumer(tmm, name, spillers, virtualChildren);
+
+    if (GlutenConfig.getConf().dynamicOffHeapSizingEnabled()) {
+      return new DynamicOffHeapSizingMemoryTarget(memoryTarget);
+    }
+
+    return memoryTarget;
```
zhztheplayer, line 64: Can we add another method to do this wrapping? E.g.,

```java
public static MemoryTarget dynamicOffHeapSizing(MemoryTarget target) {
  ...
}
```

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1821,4 +1832,42 @@ object GlutenConfig {
...
+      .doc(
+        "Experimental: Determines the memory fraction used to determine the total " +
+          "memory available for offheap and onheap allocations when the dynamic offheap " +
+          "sizing feature is enabled. The default is set to match spark.executor.memoryFraction.")
+      .doubleConf
+      .checkValue(v => v >= 0 && v <= 1, "offheap sizing memory fraction must between [0, 1]")
+      .createWithDefault(0.6)
+
+  val COLUMNAR_ONHEAP_SIZE_IN_BYTES =
+    buildConf(GlutenConfig.GLUTEN_ONHEAP_SIZE_IN_BYTES_KEY)
```
ulysses-you, line 1865: do we still need this config since we have `DYNAMIC_OFFHEAP_SIZING_MEMORY_FRACTION` ?

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
```diff
@@ -0,0 +1,95 @@
...
+    if (size == 0) {
+      return 0;
+    }
+
+    long totalMemory = Runtime.getRuntime().totalMemory();
+    long freeMemory = Runtime.getRuntime().freeMemory();
+    long usedOnHeapBytes = (totalMemory - freeMemory);
+    long usedOffHeapBytesNow = USED_OFFHEAP_BYTES.get();
+
+    if (size + usedOffHeapBytesNow + usedOnHeapBytes > MAX_MEMORY_IN_BYTES) {
```
ulysses-you, line 49: Shall we return `MAX_MEMORY_IN_BYTES - (size + usedOffHeapBytesNow + usedOnHeapBytes)` ?

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
```diff
@@ -0,0 +1,95 @@
...
+    long reserved = delegated.borrow(size);
+
+    USED_OFFHEAP_BYTES.addAndGet(reserved);
+
+    return reserved;
+  }
+
+  @Override
+  public long repay(long size) {
+    long unreserved = delegated.repay(size);
```
ulysses-you, line 75: IIUC, after calling `repay`, it is possible that the unreserved memory is from vanilla Spark rather than native. Is it expected for `USED_OFFHEAP_BYTES` ?

ulysses-you: @zhztheplayer , am my understanding right ?

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
supermem613: I don't think so - we were asked to allocate a specific amount of memory. The check here is for what's overall allowed and returning it would deviate from current behavior.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
supermem613: I don't believe this is possible. All repay/borrow flow here is not coming from Spark Vanilla.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
supermem613: You're correct - we no longer need this. Removing it.

zhztheplayer (commented):
LGTM for now % the unresolved comments. cc @ulysses-you If you have more.

I actually still be a little bit concerned about the calculation `totalMemory - freeMemory`. But you can proceed on more integration tests to see if it leads to problems. 

So overall I think it's expected to see more test results for this feature as the next step. Did you have some plan on that @supermem613? And anyway we can give the feature some time for more development iterations.

ulysses-you:
I'm fine with this pr, looking forward some best practices, thank you @supermem613 @zhztheplayer

zhztheplayer (approved):
Thanks!

ulysses-you (approved):

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -164,20 +165,71 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
...
-    // determined by spark.memory.storageFraction was used.
-    val fraction = 1.0d - conf.getDouble("spark.memory.storageFraction", 0.5d)
-    val conservativeOffHeapPerTask = (offHeapSize * fraction).toLong / taskSlots
-    conf.set(
-      GlutenConfig.GLUTEN_CONSERVATIVE_TASK_OFFHEAP_SIZE_IN_BYTES_KEY,
-      conservativeOffHeapPerTask.toString)
+    // If we are using dynamic off-heap sizing, we should also enable off-heap memory
+    // officially.
+    if (conf.getBoolean(GlutenConfig.GLUTEN_DYNAMIC_OFFHEAP_SIZING_ENABLED, false)) {
+      conf.set(GlutenConfig.GLUTEN_OFFHEAP_ENABLED, "true")
```
Yohahaha, line 213: I'm a bit confused here, even user set `spark.memory.offHeap.enabled=false`, it will still be enabled when dynamic off-heap sizing enabled?

I'm not sure whether Runtime#freeMemory could account the memory allocated from Unsafe class, which means if we re-enable spark off-heap memory, this portion of memory will not be accounted for within JVM. In another word, Runtime#freeMemory does not account Spark off-heap memory usage.

I try ask GPT and the answer is 
```
The memory allocated through Unsafe#allocateMemory is not accounted for by the Runtime#freeMemory method.
```

ulysses-you: I try to do an explanation if nobody ack here. Author can correct me if anything worng.

In short, the `spark.memory.offHeap.size` **is not used** if `spark.gluten.memory.dynamic.offHeap.sizing.enabled` is enabled.

The magic of this pr did is that, it said we can leverage the power of JVM to control both onheap and offheap memory. We only need to configure `spark.executor.memory` which is the unified memory for onheap and offheap. I think the core idea is that, if there is enough free onhep memory, then we can borrow(logical borrow) it to offheap side and prevent the onheap memory to be allocated.

Yohahaha: @ulysses-you thank you for the explanation.

> if there is enough free onhep memory, then we can borrow(logical borrow) it to offheap side and prevent the onheap memory to be allocated.

I got the core idea, but I dont understand why it can prevent the onheap memory allocation totally, would you give more explanation? thank you!

ulysses-you: As far as I can see, it can not. Actually, I have asked the question before. For now, we add a fraction to preserve some memory for onheap, [references](https://github.com/apache/incubator-gluten/pull/5439#discussion_r1590476656).

Heatao:
I'm also a bit confused here, if it cannot prevent onheap memory allocation, will this prevent GC from being triggered?
JVM thought it could allocate memory, so it did not trigger GC, resulting in being killed by k8s or yarn

zhztheplayer:
> I'm also a bit confused here, if it cannot prevent onheap memory allocation, will this prevent GC from being triggered? JVM thought it could allocate memory, so it did not trigger GC, resulting in being killed by k8s or yarn

That's the case that this approach doesn't handle yet. The faked on-heap allocations from this feature don't count into JVM's heap size counter so never trigger GC. I assume this feature isn't production-ready before being improved.

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
yikf, line 213: If `GLUTEN_OFFHEAP_ENABLED` is set, Spark will request this portion of memory, so the total memory requested is approximately `spark.executor.memory + spark.memory.offHeap.size`. However, in this feature, `spark.executor.memory` is managed uniformly for onheap and offheap memory. So the final effect is that Spark requests `spark.executor.memory + spark.memory.offHeap.size` memory from the resource manager, but this feature only uses a portion of that memory. In other words, it seems to be limiting itself? @supermem613

PR #5442, opened 2024-04-17
[VL] Make the maxBatchSize check consistent

WangGuangxin:
A minor change to make the `maxBatchSize` consistent with the check in `VeloxShuffleWriter` 

https://github.com/apache/incubator-gluten/blob/main/cpp/velox/shuffle/VeloxShuffleWriter.cc#L222

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -481,10 +481,10 @@ object BackendSettings extends BackendSettingsApi {
   private def checkMaxBatchSize(nativeConf: java.util.Map[String, String]): Unit = {
     if (nativeConf.containsKey(GlutenConfig.GLUTEN_MAX_BATCH_SIZE_KEY)) {
       val maxBatchSize = nativeConf.get(GlutenConfig.GLUTEN_MAX_BATCH_SIZE_KEY).toInt
-      if (maxBatchSize > MAXIMUM_BATCH_SIZE) {
+      if (maxBatchSize >= MAXIMUM_BATCH_SIZE) {
         throw new IllegalArgumentException(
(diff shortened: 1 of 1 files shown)


WangGuangxin:
cc @marin-ma

marin-ma:
@FelixYBW Should we completely remove the batch size limitation on both java and native code as https://github.com/apache/incubator-gluten/pull/5326 has been merged?

FelixYBW:
> @FelixYBW Should we completely remove the batch size limitation on both java and native code as #5326 has been merged?

Yes, let's comment it and marked as issue to be solved in Velox.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -481,10 +481,10 @@ object BackendSettings extends BackendSettingsApi {
   private def checkMaxBatchSize(nativeConf: java.util.Map[String, String]): Unit = {
     if (nativeConf.containsKey(GlutenConfig.GLUTEN_MAX_BATCH_SIZE_KEY)) {
       val maxBatchSize = nativeConf.get(GlutenConfig.GLUTEN_MAX_BATCH_SIZE_KEY).toInt
-      if (maxBatchSize > MAXIMUM_BATCH_SIZE) {
+      if (maxBatchSize >= MAXIMUM_BATCH_SIZE) {
```
marin-ma, line 484: The check in native shuffle writer requires the split buffer (aka. output batch size from shuffle writer) <= 32k, which means the maximum batch size in the pipeline can equal to 32k. But this change makes the maximum batch size < 32k.

WangGuangxin: Got it

FelixYBW: @marin-ma we should remove the check here, right?

PR #5446, opened 2024-04-18
[VL] Remove batch size limit

marin-ma:
Remove the batch size limitation as Velox pipeline can produce large outputs against the configured batch size.
See issue https://github.com/apache/incubator-gluten/issues/5307 and PR #5326 

Velox issue: https://github.com/facebookincubator/velox/issues/9540


FelixYBW (approved):
Let's put the Velox issue id here in case we need to add the check in future

PR #5447, opened 2024-04-18
[GLUTEN-5414][VL] FEAT: Support read CSV

jinchengchenghh:
The PR use Arrow's CSV reader to parse the CSV file then feed the arrow format data into Velox pipeline.
Quey plan is
```
*(1) ColumnarToRow
+- ArrowFileScan arrowcsv [Name#17,Language#18] Batched: true, DataFilters: [], Format: org.apache.gluten.datasource.ArrowCSVFileFormat@7772ec28, Location: InMemoryFileIndex(1 paths)[file:/mnt/DP_disk1/code/incubator-gluten/backends-velox/target/scala-2..., PartitionFilters: [], PushedFilters: [], ReadSchema: struct<Name:string,Language:string>


VeloxColumnarToRowExec
+- ^(2) FilterExecTransformer (isnotnull(Name#17) AND (Name#17 = Peter))
   +- ^(2) InputIteratorTransformer[Name#17, Language#18]
      +- ^(2) InputAdapter
         +- ^(2) ArrowFileScan arrowcsv [Name#17,Language#18] Batched: true, DataFilters: [isnotnull(Name#17), (Name#17 = Pet
```
`ArrowFileScan arrowcsv` can indicate the file format has already been changed to arrow

If the specified schema is different with file schema, will fallback to vanilla Spark to generate UnsafeRow and then convert to velox ColumnarBatch then to ArrowRecordBatch, because now we don't have row -> arrow converter, and `supportsBatch` true means we should output a Arrow columnarbatch

This PR introduces protobuf to compile jni code, https://github.com/apache/arrow/pull/36929/files, but the higher protobuf version will cause `UnsatisfiedLinkError: /tmp/jnilib-4372912739792055919.tmp: /tmp/jnilib-4372912739792055919.tmp: undefined symbol: _ZTIN6google8protobuf7MessageE`, so we need to compile Arrow java dataset module


gluten-core/src/main/scala/org/apache/spark/memory/SparkMemoryUtil.scala
```diff
@@ -209,9 +209,50 @@ object SparkMemoryUtil {
...
+      val hasNext = delegate.hasNext
+      hasNext
+    }
+
+    override def next(): T = {
+      val b = delegate.next()
+      holder.retain(b)
+      b
+    }
+  }
```
zhztheplayer, line 257: Had some new APIs to replace this sort of wrappers

https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/utils/Iterators.scala

gluten-data/src/main/scala/org/apache/gluten/datasource/ArrowFileFormat.scala
```diff
@@ -0,0 +1,163 @@
...
+  extends FileFormat
+  with DataSourceRegister
+  with Logging
+  with Serializable {
+
+  override def isSplitable(
+      sparkSession: SparkSession,
+      options: Map[String, String],
+      path: Path): Boolean = {
+    true
```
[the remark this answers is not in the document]
jinchengchenghh: super is false, but here is true. when codec is not empty, it cannot split, I will refactor here

backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
```diff
@@ -458,6 +458,17 @@ class TestOperator extends VeloxWholeStageTransformerSuite {
     }
   }
 
+  test("csv scan") {
```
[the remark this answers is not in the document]
jinchengchenghh: yes, this PR supports compression codec

gluten-data/src/main/scala/org/apache/gluten/utils/ArrowUtil.scala
```diff
@@ -86,4 +101,223 @@ object ArrowUtil extends Logging {
...
+    }
+    val rewritten = new URI(sch, ssp, uri.getFragment)
+    rewritten.toString
+  }
+
+  def makeArrowDiscovery(encodedUri: String, format: FileFormat): FileSystemDatasetFactory = {
+    val allocator = ArrowBufferAllocators.contextInstance()
+    val factory = new FileSystemDatasetFactory(
+      allocator,
+      NativeMemoryPool.getDefault, // TODO: wait to change
```
jinchengchenghh, line 149: This memory pool usage is not tracked. Looks like we does not have a wrapper in Gluten. @zhztheplayer

zhztheplayer: We can try `NativeMemoryPool#createListenable`. Though In Gluten we already had a listener named `ReservationListener` so we should do some tricks on naming.

gluten-data/src/main/scala/org/apache/gluten/utils/ArrowUtil.scala
```diff
@@ -86,4 +101,223 @@ object ArrowUtil extends Logging {
...
+    vectors.foreach {
+      vector =>
+        vector.putNulls(0, rowCount)
+        vector.setValueCount(rowCount)
+    }
+
+    SparkMemoryUtil.addLeakSafeTaskCompletionListener(
+      (_: TaskContext) => {
+        vectors.foreach(_.close())
+      })
```
zhztheplayer, line 243: Is there a way to avoid closing these vectors through task completion listener?

gluten-data/src/main/scala/org/apache/gluten/utils/ArrowUtil.scala
```diff
@@ -86,4 +101,223 @@ object ArrowUtil extends Logging {
...
+      i => {
+        ArrowColumnVectorUtils.populate(partitionColumns(i), partitionValues, i)
+        partitionColumns(i).setValueCount(rowCount)
+        partitionColumns(i).setIsConstant()
+      })
+
+    SparkMemoryUtil.addLeakSafeTaskCompletionListener(
+      (_: TaskContext) => {
+        partitionColumns.foreach(_.close())
+      })
```
zhztheplayer, line 263: ditto

gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala
```diff
@@ -70,6 +70,9 @@ object PlanUtil {
       case _: InputAdapter => false
       case _: WholeStageCodegenExec => false
       case r: ReusedExchangeExec => isVanillaColumnarOp(r.child)
+      case f: FileSourceScanExec =>
+        !f.relation.fileFormat.getClass.getSimpleName.equals("ArrowCSVFileFormat") &&
+        f.supportsColumnar
```
zhztheplayer, line 75: Would you like to explain this change? Thanks.

zhztheplayer (commented):
Most of the code is in good shape to me. Thanks for working on this!

gluten-data/src/main/scala/org/apache/gluten/utils/ArrowUtil.scala
zhztheplayer: > We can try `NativeMemoryPool#createListenable`. Though in Gluten we already had a listener named `ReservationListener` so we should do some tricks on naming.

This can be done in separate PR if we can leave some TODO notes in code.

gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala
jinchengchenghh: This will cause add extra c2r and r2c, the function is invoked in fallback to add c2r operator

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1460,6 +1462,13 @@ object GlutenConfig {
       .booleanConf
       .createOptional
 
+  val NATIVE_ARROW_READER_ENABLED =
+    buildConf("spark.gluten.sql.native.arrow.reader.enabled")
+      .internal()
+      .doc("This is config to specify whether to enable the native columnar csv reader")
+      .booleanConf
+      .createWithDefault(true)
```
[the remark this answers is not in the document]
jinchengchenghh: ok , default true is used to pass CI, I will only set it to true in `GlutenCSVSuite` and other relevant tests

gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala
zhztheplayer: So what if we add a rule to convert `FileSourceScanExec` to `GlutenFileSourceScanExec` (or something like that) that outputs Arrow columnar data by default? Which is reasonable since Gluten is guaranteed compatible with Arrow at arch level. The code can be placed in `gluten-data` and `cpp/core`.

zhztheplayer: If we don't want to copy&paste code from  `FileSourceScanExec`, we can adopt decorator design pattern to wrap a  `FileSourceScanExec` instance inside `ArrowFileSourceScanExec`.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -748,6 +749,14 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
     SparkShimLoader.getSparkShims.getExtendedColumnarPostRules() ::: List()
   }
 
+  override def genInjectPostHocResolutionRules(): List[SparkSession => Rule[LogicalPlan]] = {
+    val buf: ListBuffer[SparkSession => Rule[LogicalPlan]] = ListBuffer()
+    if (BackendsApiManager.getSettings.enableNativeArrowReadFiles()) {
+      buf += ArrowConvertorRule
```
zhztheplayer, line 755: Would you like to explain a bit about the reason to inject the rule as PostHocResolutionRule? Thanks.

gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala
jinchengchenghh: Then we should change the inject rule logic, replace FileSourceScanExec to other class. Like https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala,  https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scalathis implement looks complex, do we have easier way to inject the rule?

zhztheplayer: We can simply add a validation line to https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala

And add some transformation code to https://github.com/apache/incubator-gluten/blob/c4b4c72732f2d662c6c7b0b41c46d47f77b281d0/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala#L289

Could just follow the normal way we transform file scan to file scan transformer. 

It's awkward that we still need to take time to find a way to correctly inject some rules after such long time of development of Gluten. We need some complete refactors against the base arch eventually...

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
jinchengchenghh: Because it replace the LogicalRelation( HadoopFsRelation(_, CSVFileFormat))

zhouyuan (approved):
👍

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
```diff
@@ -303,7 +303,15 @@ object TransformOthers {
...
+      val plan = p match {
+        case plan: FileSourceScanExec
+            if plan.relation.fileFormat.getClass.getSimpleName == "ArrowCSVFileFormat" =>
+          val arrowScan = ArrowFileSourceScanExec(plan)
+          TransformHints.tagNotTransformable(arrowScan, "Arrow scan cannot transform")
+          return arrowScan
+        case p => p
+      }
+
       if (TransformHints.isNotTransformable(plan)) {
```
ulysses-you, line 315: shall we handle `ArrowFileSourceScanExec` inside this branch ?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
```diff
@@ -322,6 +330,8 @@ object TransformOthers {
           applyScanTransformer(plan)
         case plan: FileSourceScanExec =>
           applyScanTransformer(plan)
+        case plan: ArrowFileSourceScanExec =>
+          applyScanTransformer(plan)
```
ulysses-you, line 334: it seems we can never reach here ?

ulysses-you (commented):
thank you @jinchengchenghh , I have a few comments

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
jinchengchenghh: No, we the partitionFilters is null, it is taged not transformable, https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala#L343, otherwise, the plan has no tag

ulysses-you: The `ArrowFileSourceScanExec` is always tagged as fallback. So if we do not handle `ArrowFileSourceScanExec` here,  what happens if `ArrowFileSourceScanExec` has a dynamic partition filter ?

jinchengchenghh: The csv arrow reader does not support filter pushdown now, so all the filter will left to FilterExec

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
```diff
@@ -303,7 +303,15 @@ object TransformOthers {
...
     def doReplace(p: SparkPlan): SparkPlan = {
-      val plan = p
+      val plan = p match {
+        case plan: FileSourceScanExec
+            if plan.relation.fileFormat.getClass.getSimpleName == "ArrowCSVFileFormat" =>
+          val arrowScan = ArrowFileSourceScanExec(plan)
+          TransformHints.tagNotTransformable(arrowScan, "Arrow scan cannot transform")
+          return arrowScan
+        case p => p
+      }
```
zhztheplayer, line 313: The code is placed in `gluten-core` but relies on code in `backend-velox`. Would you like to check whether we can add a rule via backend API?

Probably through `BackendsApiManager.getSparkPlanExecApiInstance.genExtendedColumnarValidationRules` ?

backends-velox/src/main/scala/org/apache/gluten/datasource/ArrowConvertorRule.scala
```diff
@@ -0,0 +1,85 @@
...
+        ) {
+          l.copy(relation = r.copy(fileFormat = new ArrowCSVFileFormat())(session))
+        } else l
+      case r =>
+        r
+    }
+  }
+
+  private def checkCsvOptions(csvOptions: CSVOptions, timeZone: String): Boolean = {
+    csvOptions.headerFlag && !csvOptions.multiLine && csvOptions.delimiter == "," &&
```
liujiayi771, line 64: Does Arrow only support "," delimiter?

liujiayi771:
> the higher protobuf version will cause UnsatisfiedLinkError: /tmp/jnilib-4372912739792055919.tmp: /tmp/jnilib-4372912739792055919.tmp: undefined symbol: _ZTIN6google8protobuf7MessageE

Do we have a workaround for this issue? I also encountered it during testing.

backends-velox/src/main/scala/org/apache/gluten/datasource/ArrowConvertorRule.scala
jinchengchenghh: Yes, this is the default delimiter

jinchengchenghh:
> > the higher protobuf version will cause UnsatisfiedLinkError: /tmp/jnilib-4372912739792055919.tmp: /tmp/jnilib-4372912739792055919.tmp: undefined symbol: _ZTIN6google8protobuf7MessageE
> 
> Do we have a workaround for this issue? I also encountered it during testing.

Yes, I have fixed it by compiling Arrow https://github.com/apache/incubator-gluten/pull/5447/files#diff-e2b6a0ac0f3384d96902fbc545a697d0aa305c7ed711493df5e1bddf6c6d401bR295, and if your platform is centos, you may get the openssl link zlib issue, you could compile static openssl and set `export PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig` to find correct `libcrypto.pc` `https://github.com/apache/incubator-gluten/pull/5447/files#diff-38eda6989d285e6a3d79a3a66cc656d0b6d483b3d5c77bd0969d878642fd0f0eR824

backends-velox/src/main/scala/org/apache/gluten/datasource/ArrowConvertorRule.scala
jinchengchenghh: And I will discuss with Arrow community to support more options, https://github.com/apache/arrow/issues/28866#issuecomment-2095354665

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
jinchengchenghh: Can I change it in a following PR? This PR is too big

liujiayi771:
We need to add shaded exclude in `package/pom.xml`.
```xml
<exclude>org.apache.arrow.dataset.*</exclude>
```
Otherwise, JNI cannot find the correct dataset java method.

liujiayi771:
We also need to add HDFS support for arrow dataset.
```
caused by: java.lang.RuntimeException: Got HDFS URI but Arrow compiled without HDFS support                                                                                                                                  
        at org.apache.arrow.dataset.file.JniWrapper.makeFileSystemDatasetFactory(Native Method)                                                                                                                              
        at org.apache.arrow.dataset.file.FileSystemDatasetFactory.createNative(FileSystemDatasetFactory.java:40)                                                                                                             
        at org.apache.arrow.dataset.file.FileSystemDatasetFactory.<init>(FileSystemDatasetFactory.java:31)                                                                                                                   
        at org.apache.gluten.utils.ArrowUtil$.makeArrowDiscovery(ArrowUtil.scala:149)                                                                                                                                        
        at org.apache.gluten.datasource.ArrowCSVFileFormat.$anonfun$buildReaderWithPartitionValues$3(ArrowCSVFileFormat.scala:281)                                                                                           
        at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.org$apache$spark$sql$execution$datasources$FileScanRDD$$anon$$readCurrentFile(FileScanRDD.scala:231)                                               
        at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.nextIterator(FileScanRDD.scala:293)                                                                                                                
        at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:125)
```

jinchengchenghh:
> We also need to add HDFS support for arrow dataset.
> 
> ```
> caused by: java.lang.RuntimeException: Got HDFS URI but Arrow compiled without HDFS support                                                                                                                                  
>         at org.apache.arrow.dataset.file.JniWrapper.makeFileSystemDatasetFactory(Native Method)                                                                                                                              
>         at org.apache.arrow.dataset.file.FileSystemDatasetFactory.createNative(FileSystemDatasetFactory.java:40)                                                                                                             
>         at org.apache.arrow.dataset.file.FileSystemDatasetFactory.<init>(FileSystemDatasetFactory.java:31)                                                                                                                   
>         at org.apache.gluten.utils.ArrowUtil$.makeArrowDiscovery(ArrowUtil.scala:149)                                                                                                                                        
>         at org.apache.gluten.datasource.ArrowCSVFileFormat.$anonfun$buildReaderWithPartitionValues$3(ArrowCSVFileFormat.scala:281)                                                                                           
>         at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.org$apache$spark$sql$execution$datasources$FileScanRDD$$anon$$readCurrentFile(FileScanRDD.scala:231)                                               
>         at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.nextIterator(FileScanRDD.scala:293)                                                                                                                
>         at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:125)
> ```
> 
> But even after enabling ARROW_HDFS, there are still issues when accessing csv in HDFS.
> 
> ```
> [libprotobuf ERROR google/protobuf/descriptor_database.cc:642] File already exists in database: Security.proto                                                                                                               
> [libprotobuf FATAL google/protobuf/descriptor.cc:1986] CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):                                                                                                
> terminate called after throwing an instance of 'google::protobuf::FatalException'                                                                                                                                            
>   what():  CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):   
> ```

This may be in case of protobuf version, I will try to test CSV in HDFS

jinchengchenghh:
After I enable HDFS in ARROW, I can successfully read csv file in HDFS.
```
scala> val filePath = "/input/student.csv"
filePath: String = /input/student.csv

scala> val df = spark.read.format("csv").option("header", "true").load(filePath)
E0508 10:25:42.841267 3005137 Exceptions.h:69] Line: /mnt/DP_disk1/code/incubator-gluten/ep/build-velox/build/velox_ep/velox/exec/Task.cpp:1850, Function:terminate, Expression:  Cancelled, Source: RUNTIME, ErrorCode: INVALID_STATE
df: org.apache.spark.sql.DataFrame = [Name: string, Language: string]

scala>
     | df.show()
+-----+--------+
| Name|Language|
+-----+--------+
| Juno|    Java|
|Peter|  Python|
|Celin|     C++|
+-----+--------+


scala> print(df.queryExecution.executedPlan)
*(1) ColumnarToRow
+- ArrowFileScan arrowcsv [Name#17,Language#18] Batched: true, DataFilters: [], Format: org.apache.gluten.datasource.ArrowCSVFileFormat@485f3327, Location: InMemoryFileIndex(1 paths)[hdfs://0.0.0.0:9000/input/student.csv], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<Name:string,Language:string>


```
My local protobuf version, I'm not sure if it is related to protobuf version @liujiayi771 
```
root@sr249:/mnt/DP_disk2/tpcds/scripts# protoc --version
libprotoc 3.21.4
```

liujiayi771:
@jinchengchenghh 
I have successfully read CSV files in HDFS. Thanks.

jinchengchenghh:
Can you help merge it? Thanks! @zhztheplayer

zhouyuan (approved):
👍

FelixYBW:
@jinchengchenghh can you do a test of performance improvement? csv parser only and csv + Gluten. You may create a large table table and run TPCH Q6.

jinchengchenghh:
TPCH SF2000 Q6 performance, query:
`select        sum(l_extendedprice * l_discount) as revenue from       lineitem where  l_shipdate >= '1994-01-01'      and l_shipdate < '1995-01-01'   and l_discount between .06 - 0.01 and .06 + 0.01        and l_quantity < 24`

lineitem data: 622G
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/cjin/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/cjin/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


csv gluten without native reader | csv gluten native csv reader
-- | --
8333.039907 | 2456



</body>

</html>

Test script:
```
val schema = new StructType().add("l_orderkey", LongType).add("l_partkey", LongType).add("l_suppkey", LongType).add("l_linenumber", LongType).add("l_quantity", DoubleType).add("l_extendedprice", DoubleType).add("l_discount", DoubleType).add("l_tax", DoubleType).add("l_returnflag", StringType).add("l_linestatus", StringType).add("l_shipdate", DateType).add("l_commitdate", DateType).add("l_receiptdate", DateType).add("l_shipinstruct", StringType).add("l_shipmode", StringType).add("l_comment", StringType)

val lineitem = spark.read.format("csv").option("header","true").schema(schema).load("file:///mnt/DP_disk2/tpch/csvdata/")
spark.sql(q6)
```
Note: because the file schema should match Arrow schema, so we should specify the schema by `.schema(arrow_matched_schema)`

FelixYBW:
Th

> TPCH SF2000 Q6 performance, query: `select sum(l_extendedprice * l_discount) as revenue from lineitem where l_shipdate >= '1994-01-01' and l_shipdate < '1995-01-01' and l_discount between .06 - 0.01 and .06 + 0.01 and l_quantity < 24`
> 
> lineitem data: 622G
> 
> csv gluten without native reader	csv gluten native csv reader
> 8333.039907	2456
> Test script:
> 
> ```
> val schema = new StructType().add("l_orderkey", LongType).add("l_partkey", LongType).add("l_suppkey", LongType).add("l_linenumber", LongType).add("l_quantity", DoubleType).add("l_extendedprice", DoubleType).add("l_discount", DoubleType).add("l_tax", DoubleType).add("l_returnflag", StringType).add("l_linestatus", StringType).add("l_shipdate", DateType).add("l_commitdate", DateType).add("l_receiptdate", DateType).add("l_shipinstruct", StringType).add("l_shipmode", StringType).add("l_comment", StringType)
> 
> val lineitem = spark.read.format("csv").option("header","true").schema(schema).load("file:///mnt/DP_disk2/tpch/csvdata/")
> spark.sql(q6)
> ```
> 
> Note: because the file schema should match Arrow schema, so we should specify the schema by `.schema(arrow_matched_schema)`

Thank you, Chengcheng. What's the vanilla spark performance in this case? And how many task threads did you use?

PR #5449, opened 2024-04-18
[GLUTEN-5448][VL] Fix the issue of cleaning when left with previous build in velox-backends part

Donvi:

Just add sudo command to the clean step to make sure all of those files removed

## How was this patch tested?

mannual test in local build and find it can work well

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -189,7 +189,7 @@ function build_velox {
 function build_gluten_cpp {
   echo "Start to Gluten CPP"
   cd $GLUTEN_DIR/cpp
-  rm -rf build
+  sudo rm -rf build
   mkdir build
   cd build
   cmake -DBUILD_VELOX_BACKEND=ON -DCMAKE_BUILD_TYPE=$BUILD_TYPE \
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -189,7 +189,7 @@ function build_velox {
 function build_gluten_cpp {
   echo "Start to Gluten CPP"
   cd $GLUTEN_DIR/cpp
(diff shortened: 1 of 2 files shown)


dev/builddeps-veloxbe.sh
```diff
@@ -189,7 +189,7 @@ function build_velox {
 function build_gluten_cpp {
   echo "Start to Gluten CPP"
   cd $GLUTEN_DIR/cpp
-  rm -rf build
+  sudo rm -rf build
   mkdir build
   cd build
```
Yohahaha, line 194: do we really need this change? if you need sudo rm, you will need sudo mkdir too, could you double check it?

ep/build-velox/src/build_velox.sh
```diff
@@ -209,7 +209,7 @@ function check_commit {
     fi
   else
     # Branch-new build requires all untracked files to be deleted. We only need the source code.
-    git clean -dffx :/
+    sudo git clean -dffx :/
```
Yohahaha, line 212: great catch!

dev/builddeps-veloxbe.sh
```diff
@@ -213,4 +213,4 @@ commands_to_run=${OTHER_ARGUMENTS:-}
        "${cmd}"
     done
   fi
-)
\ No newline at end of file
+)
```
Yohahaha, line 216: just avoid introducing unnecessary changes.

Yohahaha (commented):
thank you! some minor comments

dev/builddeps-veloxbe.sh
Donvi: Thanks for reminding, fixed.  But this is only for removing, so this should not influence the build result.

Yohahaha (approved):
LGTM!

PR #5456, opened 2024-04-19
[VL] Enable split preloading by default

zhli1142015:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
@@ -1232,7 +1232,7 @@ object GlutenConfig {
       .internal()
       .doc("The IO threads for connector split preloading")
       .intConf
-      .createWithDefault(0)
+      .createWithDefaultFunction(() => SQLConf.get.getConfString("spark.executor.cores", "1").toInt)
 
   val COLUMNAR_VELOX_ASYNC_TIMEOUT =
     buildStaticConf("spark.gluten.sql.columnar.backend.velox.asyncTimeoutOnTaskStopping")
--- a/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
@@ -1230,7 +1230,8 @@ object GlutenConfig {
   val COLUMNAR_VELOX_CONNECTOR_IO_THREADS =
     buildStaticConf("spark.gluten.sql.columnar.backend.velox.IOThreads")
       .internal()
-      .doc("The IO threads for connector split preloading")
+      .doc("The Size of the IO thread pool in the Connector. This thread pool is used for split:" +
+        " preloading and DirectBufferedInput.")
       .intConf
       .createWithDefaultFunction(() => SQLConf.get.getConfString("spark.executor.cores", "1").toInt)
 
--- a/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
@@ -1233,7 +1233,11 @@ object GlutenConfig {
       .doc("The Size of the IO thread pool in the Connector. This thread pool is used for split" +
         " preloading and DirectBufferedInput.")
       .intConf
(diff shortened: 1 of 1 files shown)


shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1211,7 +1211,7 @@ object GlutenConfig {
       .internal()
       .doc("The IO threads for connector split preloading")
       .intConf
-      .createWithDefault(0)
+      .createWithDefaultFunction(() => SQLConf.get.getConfString("spark.executor.cores", "1").toInt)
```
Yohahaha, line 1214: may I ask why use spark.executor.cores as default values? why not set 2 to align with Velox?

Yohahaha, line 1214: this thread executor will be used in DirectBufferedInput, not just for split preloading.

Yohahaha:
I remember split preloading has memory issue you mentioned before, does that fixed?

Yohahaha:
> I remember split preloading has memory issue you mentioned before, does that fixed?

I see the PR has merged in upstream Velox, thank you!

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
zhli1142015: Updated the comments, thanks.

zhli1142015:
cc @FelixYBW  and @zhouyuan , thanks.

zhli1142015:
@zhouyuan  and @zhztheplayer , could you help check the memory leak failure? From the log I added and my understaning, this is related to the change https://github.com/apache/incubator-gluten/pull/5558/files#diff-1f31118af1f2bba25c508f4499fbe0304a65f1a681669ab555ce5ba03ed004c4.  The reservation is added twice (`pool->grow(0, bytes)` and `pool->grow(bytes, bytes)`) and only one is counted by child pool and can be released. 
BTW,  the memory leask looks not triggered by spilt preloading, but the `facebook::velox::cache::CoalescedLoad::loadOrFuture`.

image

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1230,9 +1230,10 @@ object GlutenConfig {
   val COLUMNAR_VELOX_CONNECTOR_IO_THREADS =
     buildStaticConf("spark.gluten.sql.columnar.backend.velox.IOThreads")
       .internal()
-      .doc("The IO threads for connector split preloading")
+      .doc("The Size of the IO thread pool in the Connector. This thread pool is used for split" +
+        " preloading and DirectBufferedInput.")
       .intConf
-      .createWithDefault(0)
+      .createWithDefaultFunction(() => SQLConf.get.getConfString("spark.executor.cores", "1").toInt)
```
ulysses-you, line 1236: Do you mean if the executor cores is 4, then there are at most 4 * 4 IO Threads at native side in one executor?

ulysses-you: Or the IOThread is shared by all tasks in one executor ?

zhli1142015: It's shared by all tasks in one exectuor.

zhouyuan:
related:
https://github.com/apache/incubator-gluten/pull/3876

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
jackylee-ch: Does the IO Threads is related to the task num in one executor? The task num in one executor is determined by `spark.executor.cores`/`spark.task.cpus`, so we may get 8 IO Threads while there are only 4 tasks running in one executor.

Yohahaha (approved):
thank you! 

users need to be aware that this config can have a huge impact on performance.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
zhli1142015: Make sense, updated.

PR #5463, opened 2024-04-19
[GLUTEN-5341] [Part 2] fix fail 3.5 ut of  VeloxParquetWriteSuite

gaoyangxiaozhu:

fix test write parquet with compression codec ut of VeloxParquetWriteSuite

(Fixes: \#5341)

## How was this patch tested?

manually run


backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteSuite.scala
```diff
@@ -77,7 +72,7 @@ class VeloxParquetWriteSuite extends VeloxWholeStageTransformerSuite {
                   val files = f.list()
                   assert(files.nonEmpty, extension)
 
-                  if (!SparkShimLoader.getSparkVersion.startsWith("3.4")) {
+                  if (!isSparkVersionGE("3.4")) {
```
yma11, line 75: I noticed this too, seems snappy doesn't exist, do you know the reason? here you seems you skip the check for Spark3.5?

gaoyangxiaozhu: hey @yma11 for spark 3.4+ the generated parquet file name changed

before for example in 3.3 it is 

```
part-00000-1e54f8f2-0858-489c-b43f-32e6f48b3b49-c000.snappy.parquet
```

but now it is 
```
Gluten_Stage_8_TID_8_0_1_e34cee0f-625d-484e-887f-c42d0f57a8d6.parquet
```
the file name not contains the codec type anymore, that's why i skip the check for this part

gaoyangxiaozhu:
@yma11 / @zhouyuan for help sign

zhouyuan:
CC @JkSelf

JkSelf (approved):
LGTM. Thanks for the fix.

PR #5464, opened 2024-04-19
Rename velox backendsapi to make consistent with CH

yma11:
Rename velox backendsapi to make consistent with CH and make it easier to distinguish it's Velox's impl

## How was this patch tested?

N/A

### DIFF
--- a/backends-velox/src/main/resources/META-INF/services/org.apache.gluten.backendsapi.Backend
+++ b/backends-velox/src/main/resources/META-INF/services/org.apache.gluten.backendsapi.Backend
@@ -1,1 +1,1 @@
-org.apache.gluten.backendsapi.VeloxBackend
+org.apache.gluten.backendsapi.velox.VeloxBackend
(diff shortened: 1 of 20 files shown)


yma11:
@zhztheplayer @zhouyuan can you help review this PR? Thanks.

backends-velox/src/main/resources/META-INF/services/org.apache.gluten.backendsapi.Backend
```diff
@@ -1 +1 @@
-org.apache.gluten.backendsapi.velox.VeloxBackend
+org.apache.gluten.backendsapi.VeloxBackend
```
zhztheplayer, line 1: I see CH backend's impl is located at package `org.apache.gluten.backendsapi.clickhouse`. Should we do such movement only for Velox backend?

https://github.com/apache/incubator-gluten/blob/972004395369224e572690b0fe53d90c19cd3a94/backends-clickhouse/src/main/resources/META-INF/services/org.apache.gluten.backendsapi.Backend#L1C1-L1C51

zhztheplayer: Also, what's the point of doing the movement?

zhztheplayer (commented):
Other changes LGTM. Thanks!

backends-velox/src/main/resources/META-INF/services/org.apache.gluten.backendsapi.Backend
yma11: I only want to change the API names but not include this folder change. Let me recover it back.

zhztheplayer (approved):
Thanks!

PR #5477, opened 2024-04-22
[VL] Support fallback processing of velox_bloom_filter_agg

zhztheplayer:
Support fallback row-based processing of agg function `velox_bloom_filter_agg`. Which means we will always use Velox's bloom-filter implementation not matter the resident operators of bloom filter functions are fallen back / disabled or not, so long as Gluten plugin + Velox backend is enabled. So we will get rid of all possible bloom-filter compatibility issues in Velox backend once and for all.


marin-ma (approved):
LGTM. Thanks!

PR #5480, opened 2024-04-22
[VL] Fix case-class inheritance for VeloxColumnarWriteFilesExec 

zhztheplayer:
Currently in code, class `VeloxColumnarWriteFilesExec` extends case class `WriteFilesExec` which is bad practice speaking of Scala.

The patch fixes the issue.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarWriteFilesExec.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarWriteFilesExec.scala
@@ -255,6 +255,10 @@ class VeloxColumnarWriteFilesRDD(
   }
 }
 
+// The class inherits from "BinaryExecNode" instead of "UnaryExecNode" because
+// we need to expose a dummy child (as right child) with type "WriteFilesExec" to let Spark
+// choose the new write code path (version >= 3.4). The actual plan to write is the left child
+// of this operator.
 case class VeloxColumnarWriteFilesExec(
     child: SparkPlan,
     fileFormat: FileFormat,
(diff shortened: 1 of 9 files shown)


zhztheplayer:
@JkSelf

backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarWriteFilesExec.scala
```diff
@@ -254,20 +255,28 @@ class VeloxColumnarWriteFilesRDD(
   }
 }
 
-class VeloxColumnarWriteFilesExec(
+case class VeloxColumnarWriteFilesExec(
```
JkSelf, line 258: @zhztheplayer Could you also add a comment at the class level explaining why it is necessary to extend BinaryExecNode?

JkSelf (approved):
LGTM. Thanks.

backends-velox/src/main/scala/org/apache/spark/sql/execution/VeloxColumnarWriteFilesExec.scala
```diff
@@ -308,12 +321,39 @@ class VeloxColumnarWriteFilesExec(
...
+  // a plan that has at least one node exactly of type `WriteFilesExec` that is a Scala case-class,
+  // to decide to choose new `#executeWrite` code path over the legacy `#execute` for write
+  // operation.
+  //
+  // So we add a no-op `WriteFilesExec` child to let Spark pick the new code path.
+  //
+  // See: FileFormatWriter#write
+  // See: V1Writes#getWriteFilesOpt
+  override val right: SparkPlan =
+    WriteFilesExec(NoopLeaf(), fileFormat, partitionColumns, bucketSpec, options, staticPartitions)
```
ulysses-you, line 336: @zhztheplayer sorry for the post review. It seems quick hack to me and really hard to read. I prefer to inherit `case class WriteFileExec`. BTW, This workaround has two issues:

1. What does UI show the new SQL graph ?
2. It would cause extra local sort if the ordering matches, see the [code](https://github.com/apache/spark/blob/79357c8ccd22729a074c42f700544e7e3f023a8d/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala#L157-L160), Spark would get the ordering of WriteFileExec's child, but this workaround miss the information.

PR #5481, opened 2024-04-22
[VL] Fix negative function mapping

zhli1142015:

Fix negative function mapping.

## How was this patch tested?

UT.


zhli1142015:
cc @PHILO-HE and @rui-mo  , thanks

rui-mo (approved):
Thanks.

PR #5493, opened 2024-04-23
[CH] Fix CH CI

zhztheplayer:
There is a CI error related to https://github.com/apache/incubator-gluten/pull/5482 but was not detected since CH CI was broken at that time. The patch fixes the error..


liuneng1994:
Does the singleton implementation of SparkDirectoryUtil lack the necessary thread synchronization?

liuneng1994 (approved):
LGTM

PR #5496, opened 2024-04-23
[VL] Improve checkNativeWrite in VeloxParquetWriteForHiveSuite

Zouxxyy:

Currently, there is no verification on whether used native write in VeloxParquetWriteForHiveSuite on spark3.4+, this PR improves it.

## How was this patch tested?

VeloxParquetWriteForHiveSuite case

### DIFF
--- a/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
@@ -16,17 +16,17 @@
  */
 package org.apache.spark.sql.execution
 
-import org.apache.gluten.sql.shims.SparkShimLoader
-
 import org.apache.spark.SparkConf
 import org.apache.spark.internal.config
 import org.apache.spark.internal.config.UI.UI_ENABLED
 import org.apache.spark.sql.{GlutenQueryTest, Row, SparkSession}
 import org.apache.spark.sql.catalyst.expressions.CodegenObjectFactoryMode
 import org.apache.spark.sql.catalyst.optimizer.ConvertToLocalRelation
+import org.apache.spark.sql.execution.datasources.FakeRowAdaptor
 import org.apache.spark.sql.hive.HiveUtils
 import org.apache.spark.sql.internal.{SQLConf, StaticSQLConf}
 import org.apache.spark.sql.test.SQLTestUtils
+import org.apache.spark.sql.util.QueryExecutionListener
 
 class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
   private var _spark: SparkSession = null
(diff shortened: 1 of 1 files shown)


Zouxxyy:
CI passed, can someone help with review? Thanks

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
```diff
@@ -130,60 +122,30 @@ class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
...
           " STORED AS PARQUET partitioned by (c, d)")
-      withSQLConf("spark.sql.hive.convertMetastoreParquet" -> "true") {
-        checkNativeStaticPartitionWrite(
-          "INSERT OVERWRITE TABLE t partition(c=1, d)" +
-            " SELECT 3 as e, 2 as e",
-          native = false)
-      }
+      spark.sql(
+        "INSERT OVERWRITE TABLE t partition(c=1, d)" +
+          " SELECT 3 as e, 2 as e")
```
Yohahaha, line 127: why not use checkNativeWriter?

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
```diff
@@ -130,60 +122,30 @@ class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
...
-          SparkShimLoader.getSparkVersion.startsWith("3.5")
-        ) {
-          checkNativeWrite("INSERT OVERWRITE TABLE t SELECT 1 as c", native = false)
-        } else {
-          checkNativeWrite("INSERT OVERWRITE TABLE t SELECT 1 as c", native = true)
-        }
-
-      }
+      checkNativeWrite("INSERT OVERWRITE TABLE t SELECT 1 as c")
       checkAnswer(spark.table("t"), Row(1))
```
Yohahaha, line 136: seems we do not check the case of using native writer for spark.sql.hive.convertMetastoreParquet=false with spark33/32 right? why?

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
```diff
@@ -130,60 +122,30 @@ class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
...
     }
   }
 
   test("test hive write dir") {
     withTempPath {
       f =>
-        // compatible with Spark3.3 and later
-        withSQLConf("spark.sql.hive.convertMetastoreInsertDir" -> "false") {
```
Yohahaha, line 165: same question.

Yohahaha (commented):
some comments, thank you!

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
```diff
@@ -77,50 +77,42 @@ class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
       .set("spark.memory.offHeap.enabled", "true")
       .set("spark.memory.offHeap.size", "1024MB")
       .set("spark.gluten.sql.native.writer.enabled", "true")
+      .set("spark.sql.hive.convertMetastoreParquet", "true")
+      .set("spark.sql.hive.convertMetastoreInsertDir", "true")
```
Yohahaha, line 81: these configs are true by default, seems we can remove it.

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
Zouxxyy: I modified the PR and now no configuration logic has been changed

liujiayi771: Perhaps gluten not support write parquet using Hive serde.

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
```diff
@@ -79,24 +79,30 @@ class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
...
+            qe.executedPlan.find(_.isInstanceOf[VeloxColumnarWriteFilesExec]).isDefined
+          } else {
+            qe.executedPlan.find(_.isInstanceOf[FakeRowAdaptor]).isDefined
+          }
+        }
+      }
     }
-    assert(
-      testAppender.loggingEvents.exists(
-        _.getMessage.toString.contains("Use Gluten parquet write for hive")) == native)
```
liujiayi771, line 89: It is great to remove this hardcoding.

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
Zouxxyy: > Perhaps gluten does not support write parquet using Hive serde.

I have tested it, 3.2/3.3 is supported, 3.4+ is not supported yet

ulysses-you: The reason why Gluten with Spark 3.4 and later does not support write with hive format is that, Spark will convert hive table to datasource table by default, so if people does not change config e.g., `spark.sql.hive.convertMetastoreParquet`, the hive format table will never happen. On the otherhand, if there is a hive format table, that means people does not want to use Spark datasource table, so we should respect it and not use velox to write. In short, we only transform datasource table to native.

PR #5498, opened 2024-04-23
[CH] Speed up mergetree metadata reloading

liuneng1994:

Speed up mergetree metadata reloading. The principle is to increase the writing speed of metadata through the thread pool.
The default thread pool size is minimum 10, which can be specified through `spark.gluten.sql.columnar.backend.ch.runtimesettings.max_threads`

```
2000+parts
before:
runTPCHQueryBySQL cost time: 136886 ms
after:
runTPCHQueryBySQL cost time: 12017 ms
```

## How was this patch tested?

manual tests


loneylee:
LGTM

PR #5500, opened 2024-04-23
[VL] Enable window.sql in GlutenSQLQueryTestSuite

JkSelf:

Enable window.sql test in GlutenSQLQueryTestSuite.

## How was this patch tested?

new unit tests.

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -16,8 +16,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=2024_04_23
+VELOX_REPO=https://github.com/JkSelf/velox.git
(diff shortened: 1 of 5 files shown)


FelixYBW:
Do we have you list of window functions/features which still fallback?

JkSelf:
> Do we have you list of window functions/features which still fallback?

@FelixYBW 
- Window Function fallback: Support for all window functions is available. However, additional testing may be required for specific data types and core case scenarios.
- Feature fallback: RangeFrame Fallback Considerations: **1)** Literal Fallback: For instance, 'RANGE BETWEEN 5 PRECEDING AND 5 FOLLOWING'; **2)** Unsupported DESC in ORDER BY. **3)** ORDER BY Data Type Limitation: Only integral data types are supported for ordering (byte, short, integer, long). **4)** Negative upper bounds and positive lower bounds are not supported.

PR #5514, opened 2024-04-24
[CORE] Untangle AddTransformHintRule to extract pre-validation code out

zhztheplayer:
Definitions (they are just for explaining what happens in this patch):

- pre-validation: Decide whether a Spark plan node can be transformed in terms of the vanilla Spark plan node it self.
- post-validation: Decide whether a Spark plan node can be transformed by calling `#doValidate` API on Gluten plan node after a tentative transform.

The PR extracts all pre-validation code out from class `AddTransformHintRule` for future use by RAS. RAS needs this part only and does post-validation together with RAS implementation rules.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
@@ -1,0 +1,212 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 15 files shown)


gluten-core/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  private class FallbackByBackendSettings(settings: BackendSettingsApi) extends Validator {
+    override def validate(plan: SparkPlan): Validator.OutCome = plan match {
+      case p: ShuffleExchangeExec if !settings.supportColumnarShuffleExec() => fail(p)
+      case p: SortMergeJoinExec if !settings.supportSortMergeJoinExec() => fail(p)
+      case p: WriteFilesExec
+          if !(settings.enableNativeWriteFiles() && settings.supportTransformWriteFiles) =>
+        fail(p)
+      case p: SortAggregateExec if !settings.replaceSortAggWithHashAgg =>
```
PHILO-HE, line 142: Looks no need to keep `replaceSortAggWithHashAgg` in backend api now. Maybe, move the check (depends on config) for `SortAggregateExec` to `FallbackByUserOptions`?

If it makes sense, it's ok to do the refactor in follow-up pr.

PHILO-HE (approved):
Looks great! Just one comment. Please confirm it. Thanks!

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
zhztheplayer: I got your point except that it seems to be an individual patch indeed. This PR just moves the options/settings checks around without optimizing them. Let's do that in another patch.

PR #5279, opened 2024-04-03
[CH]feat: Support external sort shuffle, reduce shuffle memory usage when the number of partitions is high

liuneng1994:
## What changes were proposed in this pull request?

add external sort shuffle writer

## How was this patch tested?

unit tests

Three new configurations are introduced
```
spark.gluten.sql.columnar.backend.ch.maxSortBufferSize
```
The maximum size of sort shuffle buffer in CH backend.
```
spark.gluten.sql.columnar.backend.ch.spillFirstlyBeforeStop
```
Whether to spill the sort buffers before stopping the shuffle writer.
```
spark.gluten.sql.columnar.backend.ch.forceSortShuffle
```
Whether to force to use sort shuffle in CH backend. Sort shuffle disabled by default

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #5516, opened 2024-04-24
[CH] Support celeborn on external sort shuffle

liuneng1994:

now we can use external sort shuffle with celeborn, relate to #5279 

new configuration
`spark.gluten.sql.columnar.backend.ch.runtime_config.spill_memory_overhead`, type int, default value 50MB
The new configuration item is used to control the amount of additional memory overhead when spilling. There are some stages where it is not allowed to enter the spill again. Need to prevent this from happening by adding a certain amount of spill_memory_overhead.

## How was this patch tested?

unit tests


loneylee:
LGTM

zzcclp (approved):
LGTM

PR #5517, opened 2024-04-24
[VL] CI: Minor optimizations for cache build settings

zhztheplayer:
1. Do not cancel cache job when new job are submitted. Which may cause a job that is about to finish to be aborted and renewed so waste time.
2. Check existing cache before building cache. Build only when it's not built yet.


zhztheplayer:
@zhouyuan @PHILO-HE

zhztheplayer:
Have tested this change in my own repo.

PR #5519, opened 2024-04-24
[GLUTEN-5525][BUILD] Run `mvn clean` fails if a spark profile is not specified

zhouyifan279:

Fix issue #5525

## How was this patch tested?
Run following command locally:
- `mvn clean`
- `mvn clean install -DskipTests`
- `mvn clean -Pbackends-velox -DskipTests`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,9 +47,14 @@
     <scala.binary.version>2.12</scala.binary.version>
     <scala.version>2.12.15</scala.version>
     <spark.major.version>3</spark.major.version>
-    <sparkbundle.version>3.2</sparkbundle.version>
+    <sparkbundle.version>3.4</sparkbundle.version>
     <spark.version>3.4.2</spark.version>
-    <sparkshim.artifactId>spark-sql-columnar-shims-spark32</sparkshim.artifactId>
+    <sparkshim.module.name>spark34</sparkshim.module.name>
+    <sparkshim.artifactId>spark-sql-columnar-shims-spark34</sparkshim.artifactId>
+    <iceberg.version>1.5.0</iceberg.version>
+    <delta.package.name>delta-core</delta.package.name>
+    <delta.version>2.4.0</delta.version>
+    <delta.binary.version>24</delta.binary.version>
     <celeborn.version>0.3.2-incubating</celeborn.version>
     <arrow.version>14.0.1</arrow.version>
     <uniffle.version>0.8.0</uniffle.version>
@@ -129,11 +134,9 @@
     </profile>
     <profile>
       <id>spark-3.2</id>
-      <activation>
-        <activeByDefault>true</activeByDefault>
-      </activation>
       <properties>
         <sparkbundle.version>3.2</sparkbundle.version>
+        <sparkshim.module.name>spark32</sparkshim.module.name>
         <sparkshim.artifactId>spark-sql-columnar-shims-spark32</sparkshim.artifactId>
         <spark.version>3.2.2</spark.version>
 	<iceberg.version>1.3.1</iceberg.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -129,11 +134,9 @@
     </profile>
     <profile>
       <id>spark-3.2</id>
-      <activation>
```
zhouyifan279, line 132: When profile java-8 is activated, all profiles that are active by default are automatically deactivated.
As a result, `delta.package.name` is undefined and we get a invalid artifactId in delta dependency.

Ref: https://maven.apache.org/guides/introduction/introduction-to-profiles.html#details-on-profile-activation

shims/pom.xml
```diff
@@ -68,31 +69,4 @@
       </plugins>
     </pluginManagement>
   </build>
-
-  <profiles>
```
zhouyifan279, line 72: Move into parent profiles

pom.xml
```diff
@@ -47,9 +47,14 @@
...
     <scala.version>2.12.15</scala.version>
     <spark.major.version>3</spark.major.version>
-    <sparkbundle.version>3.2</sparkbundle.version>
+    <sparkbundle.version>3.4</sparkbundle.version>
     <spark.version>3.4.2</spark.version>
-    <sparkshim.artifactId>spark-sql-columnar-shims-spark32</sparkshim.artifactId>
+    <sparkshim.module.name>spark34</sparkshim.module.name>
+    <sparkshim.artifactId>spark-sql-columnar-shims-spark34</sparkshim.artifactId>
+    <iceberg.version>1.5.0</iceberg.version>
+    <delta.package.name>delta-core</delta.package.name>
```
zhouyifan279, line 55: If run `mvn clean` without this line, `delta.package.name` is undefined and we get a invalid artifactId in delta dependency..
Because profile java-8 is activated and all profiles that are active by default are automatically deactivated.

Ref: https://maven.apache.org/guides/introduction/introduction-to-profiles.html#details-on-profile-activation

pom.xml
```diff
@@ -47,9 +47,14 @@
     <scala.binary.version>2.12</scala.binary.version>
     <scala.version>2.12.15</scala.version>
     <spark.major.version>3</spark.major.version>
-    <sparkbundle.version>3.2</sparkbundle.version>
+    <sparkbundle.version>3.4</sparkbundle.version>
```
zhztheplayer, line 50: Does it mean when no `-Pspark-<version>` specified, Spark 3.4 will be used by default? So the change includes a default version bump (from 3.2 to 3.4) for Spark?

zhouyifan279: @zhztheplayer Thanks for your review.  Yes, default Spark version will be changed.
PR #3727 already changed `spark.version` to 3.4.1 because reported security issue by local code scan. 
So I think it's better to make `sparkbundle.version` consistent with `spark.version`.

zhztheplayer (approved):
Thanks!

acvictor:
I think this PR breaks running `mvn versions:set -DnewVersion=<>`. It seems the command is not able to resolve the placeholder while updating the pom, or is there some specific way to do it? `mvn versions:set -DnewVersion=<> -Dsparkshim.module.name=spark34` does not update shims/spark34/pom.xml.

boneanxs:
> I think this PR breaks running `mvn versions:set -DnewVersion=<>`. It seems the command is not able to resolve the placeholder while updating the pom, or is there some specific way to do it? `mvn versions:set -DnewVersion=<> -Dsparkshim.module.name=spark34` does not update shims/spark34/pom.xml.

I also met the same issue, have you got any way to fix/bypass it? @acvictor

acvictor:
> > I think this PR breaks running `mvn versions:set -DnewVersion=<>`. It seems the command is not able to resolve the placeholder while updating the pom, or is there some specific way to do it? `mvn versions:set -DnewVersion=<> -Dsparkshim.module.name=spark34` does not update shims/spark34/pom.xml.
> 
> I also met the same issue, have you got any way to fix/bypass it? @acvictor

We just reverted this commit internally.

PR #5520, opened 2024-04-24
[VL] No need to increment the fixedWidthIdx variable

XinShuoWang:
If increase the fixedWidthIdx variable here, it will cause a coredump in the subsequent partitionBuffers_[fixedWidthIdx][partitionId].

(Fixes: \#ISSUE-ID)

## How was this patch tested?


marin-ma:
cc: @PHILO-HE

PHILO-HE:
> cc: @PHILO-HE Is it possible to add unittest for NullType?

Yes, @XinShuoWang, do you have a reproducible case that can be provided to us?

XinShuoWang:
> > cc: @PHILO-HE Is it possible to add unittest for NullType?
> 
> Yes, @XinShuoWang, do you have a reproducible case that can be provided to us?

I think the reproducible case should be very easy to find. It can be reproduced by shuffle null columns and fixed-length columns at the same time.

PR #5523, opened 2024-04-25
[VL] Use slice instead of resize in ensureFlattened

marin-ma:
The children of input RowVector, can reference same BaseVector before and after flattening. e.g. `select explode(a), a from ...`, which a is of map type  will output 3 columns: col_0: a.mapKeys col_1: a.mapValues and col_2: a. In this case, col_0/co_1 is referencing the same BaseVector as the mapKeys/mapValues in col_2. Resizing col_0/col_1 will cause expected failure in accessing col_2.


jinchengchenghh:
Can you add a unit test?

FelixYBW (approved):

PR #5526, opened 2024-04-25
[VL] CI: Split SF30 job to 4 jobs to speed up execution

zhztheplayer:
Added a new `gluten-it` option `--shard` to achieve this.


zhztheplayer:
@zzcclp @liuneng1994 CH CI is failed but the error doesn't seem to be related.

zhztheplayer:
cc @zhouyuan

Issue #5534, opened 2024-04-25
[VL] shuffle writer OOM if input RowVector is extremely large


guhaiyan0221:
### Backend

VL (Velox)

### Bug description

Shuffle will  OOM if the flattened rowvector's size is larger than memlimit
The case is 500 partitions, row length 315k, memlimit 2.3G, and the calculated row count for every partition is 3,
![image](https://github.com/apache/incubator-gluten/assets/142767528/8f9bd7b0-13b4-49f2-8a49-76f9f81253fa)
while the input batch has 11840 rows which can not fit into memory.

### Spark version

None

PR #5536, opened 2024-04-25
fix shuffle OOM if input batch is extremely large

guhaiyan0221:

Fix https://github.com/apache/incubator-gluten/issues/5534
Calculating the maximum number of rows (maxBatchSize) in a batch based on the memory limit of the previous batch and the number of pre-allocated rows. If row count of current batch larger than maxBatchSize, slice it into small ones

## How was this patch tested?

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxShuffleWriter.cc
@@ -340,21 +340,38 @@ arrow::Status VeloxShuffleWriter::split(std::shared_ptr<ColumnarBatch> cb, int64
     START_TIMING(cpuWallTimingList_[CpuWallTimingFlattenRV]);
     rv = veloxColumnBatch->getFlattenedRowVector();
     END_TIMING();
-    if (partitioner_->hasPid()) {
-      auto pidArr = getFirstColumn(*rv);
-      START_TIMING(cpuWallTimingList_[CpuWallTimingCompute]);
-      RETURN_NOT_OK(partitioner_->compute(pidArr, rv->size(), row2Partition_, partition2RowCount_));
-      END_TIMING();
-      auto strippedRv = getStrippedRowVector(*rv);
-      RETURN_NOT_OK(initFromRowVector(*strippedRv));
-      RETURN_NOT_OK(doSplit(*strippedRv, memLimit));
+    if (isExtremelyLargeBatch(rv)) {
+      auto numRows = rv->size();
+      int32_t offset = 0;
+      do {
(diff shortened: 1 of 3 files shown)


cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -892,6 +909,8 @@ uint32_t VeloxShuffleWriter::calculatePartitionBufferSize(const facebook::velox:
 
   totalInputNumRows_ += numRows;
 
+  maxBatchSize_ = preAllocRowCnt == 0 ? numPartitions_ : preAllocRowCnt * numPartitions_;
```
marin-ma, line 912: Can be simplified to `maxBatchSize_ = preAllocRowCnt * numPartitions_` ?

marin-ma: Why adding 'numPartitions_` here? Could you add some comments here for this calculation?

cpp/velox/shuffle/VeloxShuffleWriter.cc
```diff
@@ -340,25 +340,42 @@ arrow::Status VeloxShuffleWriter::split(std::shared_ptr<ColumnarBatch> cb, int64
...
-      START_TIMING(cpuWallTimingList_[CpuWallTimingCompute]);
-      RETURN_NOT_OK(partitioner_->compute(pidArr, rv->size(), row2Partition_, partition2RowCount_));
-      END_TIMING();
-      auto strippedRv = getStrippedRowVector(*rv);
-      RETURN_NOT_OK(initFromRowVector(*strippedRv));
-      RETURN_NOT_OK(doSplit(*strippedRv, memLimit));
+    if (isExtremelyLargeBatch(rv)) {
+      auto numRows = rv->size();
+      int32_t offset = 0;
+      do {
```
marin-ma, line 346: Is this logic covered by existing unit test? If not, could you add a new one?

cpp/velox/shuffle/VeloxShuffleWriter.cc
guhaiyan0221: if 0 < preAllocRowCnt < 1, currently use 1 for simplicity. This calculation is not accurate.

cpp/velox/shuffle/VeloxShuffleWriter.cc
guhaiyan0221: ut added

marin-ma (approved):
Thanks!

PR #5540, opened 2024-04-25
Extend adapter setup to support Debian distributions

Preetesh2110:
Extend adapter setup to support Debian distributions

This change allows the setup script to work on both Ubuntu and Debian,
increasing its portability across common Linux environments.

## How was this patch tested?
Manual build was successful.

### DIFF
--- a/ep/build-velox/src/modify_velox.patch
+++ b/ep/build-velox/src/modify_velox.patch
@@ -152,3 +152,17 @@ index 2cabfc29a..54329ce23 100644
  
  add_library(
    velox_dwio_arrow_parquet_writer_test_lib
+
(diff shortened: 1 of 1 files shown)


Preetesh2110:
@acvictor could you please take a look at it. Thanks

acvictor:
> @acvictor could you please take a look at it. Thanks

@Preetesh2110 let's update this in Velox directly? It will become painful to maintain every time the upstream file changes. I can raise a PR.

surnaik:
Hi @acvictor, nice to hear from you. Sure, please raise it in upstream velox. That seems like the right place. Thanks @Preetesh2110 for this change.

PR #5544, opened 2024-04-26
[GLUTEN-4652][VL] Fix min_by/max_by result mismatch

yma11:

Fix `min_by`/`max_by` result mismatch. Take `max_by` for example, we need to keep intermediate result row like `<null, 11>` which will be compared with another result like `<5, 8>` and assure final result is `<null, 11>`. 

## How was this patch tested?

New UT added


yma11:
@rui-mo Can you help review this PR? Thanks.

backends-velox/src/main/scala/org/apache/gluten/utils/VeloxIntermediateData.scala
```diff
@@ -159,7 +159,10 @@ object VeloxIntermediateData {
    * row_constructor_with_null.
    */
   def getRowConstructFuncName(aggFunc: AggregateFunction): String = aggFunc match {
-    case _: Average | _: Sum if aggFunc.dataType.isInstanceOf[DecimalType] => "row_constructor"
+    case _: Average | _: Sum if aggFunc.dataType.isInstanceOf[DecimalType] =>
+      "row_constructor"
+    case _: MaxMinBy =>
+      "row_constructor_with_all_null"
```
rui-mo, line 165: @yma11 Is this change due to the semantics difference between Velox and Spark? I wonder if it is possible to adjust the semantics of the implementation in Velox and avoid introducing more hack in Gluten.

yma11: I think by default, the upstream `row_constructor` doesn't set null value of the struct, it just take the input children and wrap them as a row. These issues are mainly caused by the additional projects we added, so it's reasonable to handle it at our side.

rui-mo: I see your point. Could we add a comment to explain why `row_constructor_with_all_null` is needed by MaxMinBy?

cpp/velox/operators/functions/RowConstructorWithAllNull.h
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/expression/FunctionCallToSpecialForm.h"
+#include "velox/expression/SpecialForm.h"
+
+namespace gluten {
+class RowConstructorWithAllNullCallToSpecialForm : public facebook::velox::exec::FunctionCallToSpecialForm {
```
rui-mo, line 24: Is it possible to extend `RowConstructorWithNullCallToSpecialForm` and reuse most of the code?

cpp/velox/operators/functions/RowFunctionWithNull.h
```diff
@@ -42,15 +44,26 @@ class RowFunctionWithNull final : public facebook::velox::exec::VectorFunction {
...
         for (size_t c = 0; c < argsCopy.size(); c++) {
           auto arg = argsCopy[c].get();
           if (arg->mayHaveNulls() && arg->isNullAt(i)) {
-            // If any argument of the struct is null, set the struct as null.
-            facebook::velox::bits::setNull(nullsPtr, i, true);
-            cntNull++;
-            break;
+            // For row_constructor_with_null, if any argument of the struct is null,
+            // set the struct as null.
+            if (!allNull) {
```
rui-mo, line 53: constexpr

cpp/velox/operators/functions/RowFunctionWithNull.h
```diff
@@ -42,15 +44,26 @@ class RowFunctionWithNull final : public facebook::velox::exec::VectorFunction {
...
+              facebook::velox::bits::setNull(nullsPtr, i, true);
+              cntNull++;
+              break;
+            } else {
+              argsNullCnt++;
+            }
           }
         }
+        // For row_constructor_with_all_null, set the struct to be null when all arguments are all
+        if (allNull && argsNullCnt == argsCopy.size()) {
```
rui-mo, line 63: nit: constexpr

rui-mo (commented):
Thanks!

cpp/velox/operators/functions/RowConstructorWithAllNull.h
yma11: updated.

yma11:
@rui-mo your comments are addressed. Please help review again. Thanks.

rui-mo (approved):
Thanks.

zhouyifan279:
I found the result still not right after applying this patch.
Run SQL:
```
set spark.sql.leafNodeDefaultParallelism=2;
select min_by(a, b), max_by(a, b) from values (5, 6), (null, 11), (null, 5) test(a, b);
```
Expected result is
```
NULL    NULL
```
Actual result is 
```
5      5
```

zhouyifan279:
@yma11 @rui-mo I created a follow up PR #5711. Can you help to review it?

zhouyifan279:
Also cc @ulysses-you

PR #5545, opened 2024-04-26
[VL] [BUG fix] Make the hasNext method can be called multi times

JkSelf:
Currently, the iterator is designed to close once the hasNext() method returns false. Therefore, if `hasNext()` is called a second time, it will throw exception [here](https://github.com/apache/incubator-gluten/blob/8ade7a9cb50910021a2c0cb6977aaaeb9a17b5ce/cpp/core/jni/JniWrapper.cc#L452). This PR adds a `protectInvocationFlow()` method to ensure that the iterator's `hasNext()` can be safely called multiple times without causing an exception.

## How was this patch tested?
locally test.


JkSelf:
@FelixYBW @zhztheplayer Can you help to review? Thanks.

zhztheplayer (approved):
Thanks!

zhztheplayer:
Please don't forget adding a prefix to PR title and commit message.

FelixYBW (approved):

PR #5552, opened 2024-04-26
[VL] RAS: Remove AddTransformHintRule route from EnumeratedApplier

zhztheplayer:
RAS planner doesn't need tag to mark whether a vanilla Spark plan is offload-able . It just does every essential checks whenever it tries to generate offloaded plan.

The patch does some optimizations to code to remove AddTransformHintRule from EnumeratedApplier and leave the pre-validation work to RAS planner.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
@@ -155,7 +155,7 @@ case class TransformJoin() extends TransformSingleNode with LogLevelUtil {
       plan match {
         case shj: ShuffledHashJoinExec =>
           if (BackendsApiManager.getSettings.recreateJoinExecOnFallback()) {
-            // Since https://github.com/apache/incubator-gluten/pull/5042
+            // Since https://github.com/apache/incubator-gluten/pull/408
             // Because we manually removed the build side limitation for LeftOuter, LeftSemi and
             // RightOuter, need to change the build side back if this join fallback into vanilla
             // Spark for execution.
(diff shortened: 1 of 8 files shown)


gluten-core/src/main/scala/org/apache/gluten/planner/cost/GlutenCostModel.scala
```diff
@@ -52,22 +55,26 @@ object GlutenCostModel {
...
-      case ColumnarTransitions.ColumnarToRowLike(child) => 3L
-      case ColumnarTransitions.RowToColumnarLike(child) => 3L
-      case p if PlanUtil.isGlutenColumnarOp(p) => 2L
-      case p if PlanUtil.isVanillaColumnarOp(p) => 3L
-      // Other row ops. Usually a vanilla row op.
-      case _ => 5L
+    private def selfLongCostOf(node: SparkPlan): Long = {
+      node match {
+        case p: ShuffledHashJoinExec if !TransformJoin.isLegal(p) =>
+          infLongCost
```
zhztheplayer, line 61: doc needed here

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformSingleNode.scala
```diff
@@ -237,6 +223,20 @@ case class TransformJoin() extends TransformSingleNode with LogLevelUtil {
 
 }
 
+object TransformJoin {
+  private def getSparkSupportedBuildSide(plan: ShuffledHashJoinExec): BuildSide = {
```
zhztheplayer, line 227: doc needed here

PR #5564, opened 2024-04-28
[CORE] Fix delta.package.name error

ulysses-you:

The `delta.package.name` of Spark3.5 is `delta-spark`, we should use property to specify the delta artifactId.

## How was this patch tested?

Pass `mvn spotless:apply -Pspark-3.2,spark-3.3,spark-3.4,spark-3.5 -Pbackends-velox -Pbackends-clickhouse -Prss -Pspark-ut`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -139,8 +139,8 @@
         <sparkshim.module.name>spark32</sparkshim.module.name>
         <sparkshim.artifactId>spark-sql-columnar-shims-spark32</sparkshim.artifactId>
         <spark.version>3.2.2</spark.version>
-	<iceberg.version>1.3.1</iceberg.version>
-	<delta.package.name>delta-core</delta.package.name>
+        <iceberg.version>1.3.1</iceberg.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -139,8 +139,8 @@
         <sparkshim.module.name>spark32</sparkshim.module.name>
         <sparkshim.artifactId>spark-sql-columnar-shims-spark32</sparkshim.artifactId>
         <spark.version>3.2.2</spark.version>
-	<iceberg.version>1.3.1</iceberg.version>
-	<delta.package.name>delta-core</delta.package.name>
+        <iceberg.version>1.3.1</iceberg.version>
```
yma11, line 142: Seems the change should only be needed for CH backend.  Why diff exists here?

ulysses-you: this is a small style fix, before it looks like:
image

yma11: Oh, didn't notice this. Thanks for change.

yma11 (approved):

PR #5575, opened 2024-04-29
[VL] RAS: Include rewrite rules used by RewriteSparkPlanRulesManager in EnumeratedTransform

zhztheplayer:
The patch improves EnumerabledApplier to move RewriteSparkPlanRulesManager from heuristic rule list to RAS rule list.

1. Rename API `TransformSingleNode` to `OffloadSingleNode`.
2. Minor refactor on `RewriteSparkPlanRulesManager` by adding API `RewriteSingleNode`.
3. In EnumerableApplier, remove `RewriteSparkPlanRulesManager`, and integrate relevant rewrite rules to RAS offload rule list in `EnumeratedTransform`.

After the patch, in RAS planner, when about to offload a Spark plan node, it will firstly be rewritten by the rewrite rules, then do pre-validation, offload, post-validation respectively.

For example, an HashAggregate will be handled as follow in rule `RasOfflodAggregate`:

1. Before:

```
HashAggregate
\- Leaf
```

2. After (suppose aggregate can be offloaded):

```
Project (post)
\- HashAggregateTransformer
   \- Project (pre)
      \- Leaf
```

3. After (suppose aggregate cannot be offloaded):

```
Project (post)
\- HashAggregate
   \- Project (pre)
      \- Leaf
```

When aggregate **cannot be** offloaded, then planner will consider plan 1 over plan 3 since 1 has lower cost than 3. If the aggregate **can be** offloaded, planner will usually prefer plan 2 over 1 if the pre-project and post-project can be offloaded (by rule `new AsRasOffload(OffloadOthers())`) as well, since it will contain less transitions and more Gluten plan nodes which makes it has lower cost than 1.


gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteSingleNode.scala
```diff
@@ -0,0 +1,48 @@
...
+ * the output plan.
+ *
+ * Note: Only the current plan node is supposed to be open to modification. Do not access or modify
+ * the children node. Tree-walking is done by caller of this trait.
+ *
+ * TODO: Ideally for such API we'd better to allow multiple alternative outputs.
+ */
+trait RewriteSingleNode {
+  def rewrite(plan: SparkPlan): SparkPlan
+}
```
zhztheplayer, line 36: cc @ulysses-you

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPostProject.scala
```diff
@@ -33,17 +32,17 @@ import scala.collection.mutable.ArrayBuffer
  * the output of Spark, ensuring that the output data of the native plan can match the Spark plan
  * when a fallback occurs.
  */
-object PullOutPostProject extends Rule[SparkPlan] with PullOutProjectHelper {
+object PullOutPostProject extends RewriteSingleNode with PullOutProjectHelper {
 
-  private def needsPostProjection(plan: SparkPlan): Boolean = {
+  def needsPostProjection(plan: SparkPlan): Boolean = {
     plan match {
```
zhztheplayer, line 38: cc @liujiayi771

zhztheplayer:
cc @marin-ma @rui-mo

rui-mo (commented):
Thanks.

gluten-core/src/main/scala/org/apache/gluten/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -172,12 +172,32 @@ abstract class HashAggregateExecBaseTransformer(
...
+        agg.resultExpressions,
+        childConverter(agg.child)
+      )
+  }
+}
+
+trait HashAggregateExecPullOutBaseHelper {
   // The direct outputs of Aggregation.
-  lazy val allAggregateResultAttributes: List[Attribute] =
+  def allAggregateResultAttributes(groupingExpressions: Seq[NamedExpression]): List[Attribute] =
```
ulysses-you, line 200: how about define `agg: BaseAggregateExec` in this trait ?

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
```diff
@@ -60,6 +60,12 @@ abstract class HashAggregateExecTransformer(
     resultExpressions,
     child) {
 
+  override def output: Seq[Attribute] = {
+    // TODO: We should have a check to make sure the returned schema actually matches the output
+    //  data. Since "resultExpressions" is not actually in used by Velox.
+    super.output
```
ulysses-you, line 66: Why do we need to highlight this, did not we add post project with `resultExpressions` ?

liujiayi771: The `projectList` of agg post project is the `resultExpressions`.

zhztheplayer: > The `projectList` of agg post-project is the `resultExpressions`. Is the intended expression here something like, when post-project hasn't been inserted, the output of `HashAggregateExecTransformer` isn't actually equal to `resultExpressions`?

That's correct. We don't have strict verification between `groupingExpressions+aggregateExpressions` and `resultExpressions` to ensure them matching together when instantiating `HashAggregateExecTransformer`, so there will be risk that unexpected errors happens at runtime when a `HashAggregateExecTransformer` is created without pre/post project. 

I think it will be fine if we can have a check for `needsPreProjection == false & needsPostProjection == false` at planning time in `HeuristicApplier` as well to fix this (I used very similar code to pick executable plan in RAS in this PR) in future.

Anyway it's just a soft `TODO` since current code doesn't actually impact anything.

gluten-core/src/main/scala/org/apache/gluten/execution/HashAggregateExecBaseTransformer.scala
zhztheplayer: I'll open to both because actually the change is not needed by this PR anymore... But once I needed `getAttrForAggregateExprs` to be exposed for RAS's agg validation but after sometime I reverted that approach since it messed up code. 

Back to this trait, I am feeling it's not that bad to only include required arguments to get more flexibility. However again I no longer need this change so feel free to change to another approach in future development if needed.

PR #5587, opened 2024-04-30
[GLUTEN-5586] Fix multiple generate functions failure

marin-ma:
After adding the pull-out pre project for `GenerateExecTransformer`, the child expression name can be changed. Therefore should use exprId while matching the output from child.


ulysses-you (approved):
the change lgtm, the description should be `after adding the pull-out pre project` ?

Issue #5613, opened 2024-05-06
[CH] CH function SparkCheckoverflow return type not equals with spark


loneylee:
### Backend

CH (ClickHouse)

### Bug description

Same behavior with spark

### Spark version

None

PR #5614, opened 2024-05-06
[GLUTEN-5613][CH] Fix CH function SparkCheckoverflow return type not equals with spark

loneylee:

(Fixes: \#5613)

## How was this patch tested?

Test by history ut


baibaichen (approved):
LGTM

PR #5615, opened 2024-05-06
[VL] CI: Reformat gluten-it code with Spark331's scalafmt configuration

zhztheplayer:
As title. The patch reformats gluten-it's code with Spark331's scalafmt configuration.


zhztheplayer:
Will merge this directly if no comments since it's just a code style optimization.

PR #5617, opened 2024-05-06
[VL] Add a config to ignore fallback cost for scan and improve fallback cost evaluation

PHILO-HE:
* It's common that the file format isn't supported by Gluten like Hive Row format which won't be supported in near future, but the rest of the operators are already supported by Gluten. It's high possibility we can still get perf gain from the rest operators - the R2C cost. So let's add a specific config for the scan only.
* Remove the fallback check for leaf node, as the fallback cost has been measured with transition cost (C2R/R2C).


shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -208,7 +208,7 @@ class GlutenConfig(conf: SQLConf) extends Logging {
 
   def queryFallbackThreshold: Int = conf.getConf(COLUMNAR_QUERY_FALLBACK_THRESHOLD)
 
-  def fallbackIgnoreRowToColumnar: Boolean = conf.getConf(COLUMNAR_FALLBACK_IGNORE_ROW_TO_COLUMNAR)
+  def ignoreScanFallbackCost: Boolean = conf.getConf(COLUMNAR_IGNORE_SCAN_FALLBACK_COST)
```
FelixYBW, line 211: Is the Row_to_column still used? I think it's only used by table scan

PHILO-HE: The config for ignoring RowToColumnar looks useless. So I'm removing it. It needs some discussion with our customer.

ulysses-you: I think it's still useful, fallbackIgnoreRowToColumnar allows convert shuffle to columnar shuffle

PHILO-HE: @ulysses-you, I feel it is more clear to keep considering Row2Columnar, because we are calculating all transition cost (C2R/R2C) in the current evaluation logic. The Row2Columnar doesn't only exist prior to columnar shuffle. So it may be not that good to keep `fallbackIgnoreRowToColumnar` just for the purpose of using columnar shuffle. If we prefer using columnar shuffle, maybe we can add some dedicate code to realize the purpose.

ulysses-you: @PHILO-HE the reason is that, we can some cases depend on this config. When we set transitionCost=1 and enable `fallbackIgnoreRowToColumnar`,  the following query can be transformed since we ignore row to columnar.
```
ColumnarShuffle    Shuffle
      |               |
       \        RowToColumnar
        \           /
       JoinTransformer
```

If we drop `fallbackIgnoreRowToColumnar`, when transitionCost=1 is a way to allow transforming this plan ?

PHILO-HE: @ulysses-you, looks in this example, RowToCoumnar is not counted in the Join stage.  
For some other cases, with `RowToColumnar` considered, it is easy to understand the current stage plan's transition cost. User can increase the configured threshold if transformed plan is preferred.

ulysses-you: > looks in this example, RowToCoumnar is not counted in the Join stage.

This is because we have `fallbackIgnoreRowToColumnar`..  if we drop `fallbackIgnoreRowToColumnar` then this plan can not be transformed.

>  User can increase the configured threshold if transformed plan is preferred.

It's hard to let user tune config one by one. In general, we have a default config that set transition cost to 1 and set fallbackIgnoreRowToColumnar to true, so that all application prefer to use columnar and avoid performance regression as far as possible.

ulysses-you: I think `fallbackIgnoreRowToColumnar` does not block what this pr want to do. `fallbackIgnoreRowToColumnar` is a kind of generic case for `ignore fallback cost`. It's not reasonable to allow a dedicated case but disallow a generic case.. .

PR #5624, opened 2024-05-06
[CORE] Only return columns of partitions that require read for iceberg

Zouxxyy:

Only return columns of partitions that require read for iceberg
Actually, with this PR, we won't need this: https://github.com/facebookincubator/velox/pull/9450

## How was this patch tested?

### DIFF
--- a/gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
+++ b/gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
@@ -34,7 +34,10 @@ import scala.collection.JavaConverters._
 
 object GlutenIcebergSourceUtil {
 
-  def genSplitInfo(inputPartition: InputPartition, index: Int): SplitInfo = inputPartition match {
+  def genSplitInfo(
+      inputPartition: InputPartition,
+      index: Int,
+      readPartitionSchema: StructType): SplitInfo = inputPartition match {
     case partition: SparkInputPartition =>
       val paths = new JArrayList[String]()
       val starts = new JArrayList[JLong]()
@@ -50,8 +53,8 @@ object GlutenIcebergSourceUtil {
           paths.add(filePath)
           starts.add(task.start())
           lengths.add(task.length())
-          partitionColumns.add(getPartitionColumns(task))
-          deleteFilesList.add(task.deletes());
+          partitionColumns.add(getPartitionColumns(task, readPartitionSchema))
+          deleteFilesList.add(task.deletes())
           val currentFileFormat = convertFileFormat(task.file().format())
           if (fileFormat == ReadFileFormat.UnknownFormat) {
             fileFormat = currentFileFormat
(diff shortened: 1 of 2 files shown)


gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
```diff
@@ -94,15 +97,24 @@ object GlutenIcebergSourceUtil {
       throw new GlutenNotSupportException("Only support iceberg SparkBatchQueryScan.")
   }
 
-  def getPartitionSchema(sparkScan: Scan): StructType = sparkScan match {
+  def getReadPartitionSchema(sparkScan: Scan): StructType = sparkScan match {
```
yma11, line 100: Maybe rename to `getReferredPartitionSchema` to indicates that we only retrieve the partition schemas referred  by the query?

yma11:
@Zouxxyy Thanks for fix. We may can leave alone the velox side change as it's not quite necessary by itself.

gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
Zouxxyy: It just imitates the style of spark, see `Scan.readSchema`

gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
```diff
@@ -94,15 +97,24 @@ object GlutenIcebergSourceUtil {
...
+            val readFields = scan.readSchema().fields.map(_.name).toSet
+            // Iceberg will generate some non-table fields as partition fields, such as x_bucket,
+            // which will not appear in readFields, they also cannot be filtered.
+            val tableFields = spec.schema().columns().asScala.map(_.name()).toSet
+            val partitionFields =
+              spec
+                .partitionType()
+                .fields()
+                .asScala
+                .filter(f => !tableFields.contains(f.name) || readFields.contains(f.name()))
```
yma11, line 117: Can you explain the filter here?  should it be `!tableFields.contains(f.name) && readFields.contains(f.name())`?

Zouxxyy: For iceberg's bucket join, 
spec.partitionType() will return id_bucket, p
tableFields are id, name, p
readFields are id ,name, p
will need both keep id_bucket and p

yma11: see. Thanks!

yma11 (approved):

gluten-iceberg/src/main/scala/org/apache/iceberg/spark/source/GlutenIcebergSourceUtil.scala
```diff
@@ -94,15 +97,24 @@ object GlutenIcebergSourceUtil {
...
+  def getReadPartitionSchema(sparkScan: Scan): StructType = sparkScan match {
     case scan: SparkBatchQueryScan =>
       val tasks = scan.tasks().asScala
       asFileScanTask(tasks.toList).foreach {
         task =>
           val spec = task.spec()
           if (spec.isPartitioned) {
             var partitionSchema = new StructType()
-            val partitionFields = spec.partitionType().fields().asScala
+            val readFields = scan.readSchema().fields.map(_.name).toSet
```
liujiayi771, line 108: Can we use `AtrributeSet.exists` directly?

Issue #5625, opened 2024-05-06
[VL] Support window range frame


WangGuangxin:
### Description

Support window range frame

PR #5626, opened 2024-05-06
[GLUTEN-5625][VL] Support window range frame

WangGuangxin:

Support window range frame by pre-project frame offset for non-SpecialFrameBoundary, based on the discussion https://github.com/facebookincubator/velox/issues/7557

(Fixes: \#5625)

## How was this patch tested?

UT

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -840,6 +840,14 @@ const core::WindowNode::Frame SubstraitToVeloxPlanConverter::createWindowFrame(
       VELOX_FAIL("the window type only support ROWS and RANGE, and the input type is ", std::to_string(type));
   }
 
+  auto specifiedBound = [&](bool hasOffset, int64_t offset, const ::substrait::Expression& columnRef) -> core::TypedExprPtr {
+    if (hasOffset) {
+      return std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(offset));
+    } else {
+      return exprConverter_->toVeloxExpr(columnRef, inputType);
+    }
+  };
+
   auto boundTypeConversion = [&](::substrait::Expression_WindowFunction_Bound boundType)
       -> std::tuple<core::WindowNode::BoundType, core::TypedExprPtr> {
     if (boundType.has_current_row()) {
@@ -849,25 +857,15 @@ const core::WindowNode::Frame SubstraitToVeloxPlanConverter::createWindowFrame(
     } else if (boundType.has_unbounded_preceding()) {
       return std::make_tuple(core::WindowNode::BoundType::kUnboundedPreceding, nullptr);
     } else if (boundType.has_following()) {
-      if (boundType.following().has_offset()) {
-        return std::make_tuple(
+      auto following = boundType.following();
+      return std::make_tuple(
             core::WindowNode::BoundType::kFollowing,
-            std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.following().offset())));
-      } else {
-        return std::make_tuple(
-            core::WindowNode::BoundType::kFollowing,
-            exprConverter_->toVeloxExpr(boundType.following().ref(), inputType));
-      }
+            specifiedBound(following.has_offset(), following.offset(), following.ref()));
     } else if (boundType.has_preceding()) {
-      if (boundType.preceding().has_offset()) {
-        return std::make_tuple(
+      auto preceding = boundType.preceding();
+      return std::make_tuple(
             core::WindowNode::BoundType::kPreceding,
-            std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.preceding().offset())));
-      } else {
-        return std::make_tuple(
-            core::WindowNode::BoundType::kPreceding,
-            exprConverter_->toVeloxExpr(boundType.preceding().ref(), inputType));
-      }
+            specifiedBound(preceding.has_offset(), preceding.offset(), preceding.ref()));
     } else {
       VELOX_FAIL("The BoundType is not supported.");
     }
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -823,10 +823,11 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
       nextPlanNodeId(), replicated, unnest, std::move(unnestNames), ordinalityName, childNode);
 }
 
-const core::WindowNode::Frame createWindowFrame(
+const core::WindowNode::Frame SubstraitToVeloxPlanConverter::createWindowFrame(
     const ::substrait::Expression_WindowFunction_Bound& lower_bound,
     const ::substrait::Expression_WindowFunction_Bound& upper_bound,
-    const ::substrait::WindowType& type) {
+    const ::substrait::WindowType& type,
+    const RowTypePtr& inputType) {
   core::WindowNode::Frame frame;
   switch (type) {
     case ::substrait::WindowType::ROWS:
@@ -839,9 +840,8 @@ const core::WindowNode::Frame createWindowFrame(
       VELOX_FAIL("the window type only support ROWS and RANGE, and the input type is ", std::to_string(type));
   }
 
-  auto boundTypeConversion = [](::substrait::Expression_WindowFunction_Bound boundType)
+  auto boundTypeConversion = [&](::substrait::Expression_WindowFunction_Bound boundType)
       -> std::tuple<core::WindowNode::BoundType, core::TypedExprPtr> {
-    // TODO: support non-literal expression.
     if (boundType.has_current_row()) {
       return std::make_tuple(core::WindowNode::BoundType::kCurrentRow, nullptr);
     } else if (boundType.has_unbounded_following()) {
@@ -849,13 +849,25 @@ const core::WindowNode::Frame createWindowFrame(
     } else if (boundType.has_unbounded_preceding()) {
       return std::make_tuple(core::WindowNode::BoundType::kUnboundedPreceding, nullptr);
     } else if (boundType.has_following()) {
-      return std::make_tuple(
-          core::WindowNode::BoundType::kFollowing,
-          std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.following().offset())));
+      if (boundType.following().has_offset()) {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kFollowing,
+            std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.following().offset())));
+      } else {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kFollowing,
+            exprConverter_->toVeloxExpr(boundType.following().ref(), inputType));
+      }
     } else if (boundType.has_preceding()) {
-      return std::make_tuple(
-          core::WindowNode::BoundType::kPreceding,
-          std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.preceding().offset())));
+      if (boundType.preceding().has_offset()) {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kPreceding,
+            std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.preceding().offset())));
+      } else {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kPreceding,
+            exprConverter_->toVeloxExpr(boundType.preceding().ref(), inputType));
+      }
     } else {
       VELOX_FAIL("The BoundType is not supported.");
     }
@@ -906,7 +918,7 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
     windowColumnNames.push_back(windowFunction.column_name());
 
     windowNodeFunctions.push_back(
-        {std::move(windowCall), std::move(createWindowFrame(lowerBound, upperBound, type)), ignoreNulls});
+        {std::move(windowCall), std::move(createWindowFrame(lowerBound, upperBound, type, inputType)), ignoreNulls});
   }
 
   // Construct partitionKeys
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -843,8 +843,13 @@ const core::WindowNode::Frame SubstraitToVeloxPlanConverter::createWindowFrame(
   auto specifiedBound =
       [&](bool hasOffset, int64_t offset, const ::substrait::Expression& columnRef) -> core::TypedExprPtr {
     if (hasOffset) {
+      VELOX_CHECK(
+          frame.type != ::substrait::WindowType::RANGE,
+          "for RANGE frame offset, we should pre-calculate the range frame boundary and pass the column reference, but got a constant offset.")
       return std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(offset));
     } else {
+      VELOX_CHECK(
+          frame.type != ::substrait::WindowType::ROWS, "for ROW frame offset, we should pass a constant offset.")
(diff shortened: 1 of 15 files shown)


WangGuangxin:
cc @PHILO-HE @rui-mo

FelixYBW:
@PHILO-HE is the PR offload SQL like below?
```
  sum(x) OVER (
    PARTITION BY id,sur
    ORDER BY date RANGE BETWEEN 6 preceding AND CURRENT ROW
  )
```

PHILO-HE:
> @PHILO-HE is the PR offload SQL like below?
> 
> ```
>   sum(x) OVER (
>     PARTITION BY id,sur
>     ORDER BY date RANGE BETWEEN 6 preceding AND CURRENT ROW
>   )
> ```

@FelixYBW, yes, this pr can remove the fallback for such case.

FelixYBW:
Let me do a test once merged. Thank you!

FelixYBW:
@WangGuangxin can you help to resolve conflicts?

WangGuangxin:
> @WangGuangxin can you help to resolve conflicts?

@FelixYBW done

PHILO-HE (commented):
Will spare some time to review in next week. Sorry for late response.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -793,23 +794,34 @@ const core::WindowNode::Frame createWindowFrame(
...
-          std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.following().offset())));
+      if (boundType.following().has_offset()) {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kFollowing,
+            std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.following().offset())));
+      } else {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kFollowing,
+            exprConverter_->toVeloxExpr(boundType.following().ref(), inputType));
+      }
```
rui-mo, line 814: It seems the only difference of the two branches is the expression. Can we differentiate the expressions only and reuse the rest of code?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -793,23 +794,34 @@ const core::WindowNode::Frame createWindowFrame(
...
-          std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.preceding().offset())));
+      if (boundType.preceding().has_offset()) {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kPreceding,
+            std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.preceding().offset())));
+      } else {
+        return std::make_tuple(
+            core::WindowNode::BoundType::kPreceding,
+            exprConverter_->toVeloxExpr(boundType.preceding().ref(), inputType));
+      }
```
rui-mo, line 824: ditto

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
```diff
@@ -16,41 +16,50 @@
...
+
   WindowFunctionNode(
       Integer functionId,
       List<ExpressionNode> expressionNodes,
       String columnName,
       TypeNode outputTypeNode,
-      String upperBound,
-      String lowerBound,
+      org.apache.spark.sql.catalyst.expressions.Expression upperBound,
+      org.apache.spark.sql.catalyst.expressions.Expression lowerBound,
```
rui-mo, line 59: Can we import `org.apache.spark.sql.catalyst.expressions.Expression` and use it here?

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
```diff
@@ -80,20 +91,50 @@ private Expression.WindowFunction.Bound.Builder setBound(
...
+            throw new UnsupportedOperationException(
+                "Unsupported Window Function Frame Type:" + boundType);
+          }
+        } else if (boundType.foldable()) {
+          try {
+            Long offset = Long.valueOf(boundType.eval(null).toString());
+            if (offset < 0) {
+              Expression.WindowFunction.Bound.Preceding.Builder offsetPrecedingBuilder =
+                  Expression.WindowFunction.Bound.Preceding.newBuilder();
+              offsetPrecedingBuilder.setOffset(0 - offset);
```
rui-mo, line 125: Do we need -offset?

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -989,18 +989,28 @@ message Expression {
...
-        int64 offset = 1;
+        oneof kind {
+          // A strictly positive integer specifying the number of records that
+          // the window extends back from the current record. Use
+          // CurrentRow for offset zero and Following for negative offsets.
+          int64 offset = 1;
+
+          // the reference to pre-project range frame boundary.
+          Expression ref = 2;
+        }
```
rui-mo, line 1000: Would you document this change to Substrait in https://github.com/apache/incubator-gluten/blob/main/docs/developers/SubstraitModifications.md?

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
WangGuangxin: No need. In fact, this branch is the existing logic

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
WangGuangxin: The `Expression` is conflict with `io.substrait.proto.Expression` which alreay imported in this class, and Java doen't have the alias grammer like scala

cpp/velox/substrait/SubstraitToVeloxPlan.cc
WangGuangxin: done. 
I tried to reuse the logic here, but since the type of `boundType.preceding()` and `boundType.following()` are different, so I have to pass the `offset` and `ref` instead of the expression itself

gluten-core/src/main/resources/substrait/proto/substrait/algebra.proto
WangGuangxin: added

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
```diff
@@ -80,20 +91,50 @@ private Expression.WindowFunction.Bound.Builder setBound(
...
+              builder.setPreceding(refPrecedingBuilder.build());
+            } else {
+              Expression.WindowFunction.Bound.Following.Builder refFollowingBuilder =
+                  Expression.WindowFunction.Bound.Following.newBuilder();
+              refFollowingBuilder.setRef(refNode.toProtobuf());
+              builder.setFollowing(refFollowingBuilder.build());
+            }
+          } catch (NumberFormatException e) {
+            throw new UnsupportedOperationException(
+                "Unsupported Window Function Frame Type:" + boundType);
```
PHILO-HE, line 117: Do we need to add statements to catch this exception? I think Spark parser has already checked the validity of foldable bound (e.g., 100a is not valid). The catching looks unnecessary.

gluten-core/src/main/scala/org/apache/gluten/utils/PullOutProjectHelper.scala
```diff
@@ -151,6 +183,22 @@ trait PullOutProjectHelper {
       case ae: AggregateExpression => rewriteAggregateExpression(ae, expressionMap)
       case other => other
     }
-    we.copy(windowFunction = newWindowFunc)
+
+    val newWindowSpec = we.windowSpec.frameSpecification match {
+      case swf: SpecifiedWindowFrame if needPreComputeRangeFrame(swf) =>
+        // This is guaranteed by Spark, but we still check it here
+        if (orderSpecs.size != 1) {
+          throw new GlutenNotSupportException(
```
PHILO-HE, line 191: Let's use `GlutenException`. Generally `GlutenNotSupportException` is caught in fallback handling for some limitation cases. But for this case, I think we can directly let a runtime exception happen.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -839,23 +840,33 @@ const core::WindowNode::Frame createWindowFrame(
...
     } else if (boundType.has_unbounded_following()) {
       return std::make_tuple(core::WindowNode::BoundType::kUnboundedFollowing, nullptr);
     } else if (boundType.has_unbounded_preceding()) {
       return std::make_tuple(core::WindowNode::BoundType::kUnboundedPreceding, nullptr);
     } else if (boundType.has_following()) {
+      auto following = boundType.following();
       return std::make_tuple(
           core::WindowNode::BoundType::kFollowing,
-          std::make_shared<core::ConstantTypedExpr>(BIGINT(), variant(boundType.following().offset())));
+          specifiedBound(following.has_offset(), following.offset(), following.ref()));
```
PHILO-HE, line 864: Can `has_offset` be true for Velox backend after the java side handling?

PHILO-HE (commented):
Thanks for your great work! Just posted some comments. Please check whether it makes sense.

gluten-core/src/main/scala/org/apache/gluten/utils/PullOutProjectHelper.scala
WangGuangxin: done

cpp/velox/substrait/SubstraitToVeloxPlan.cc
WangGuangxin: It's reused both for ROW and RANGE frame, so offset is used for ROW frame here.

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
```diff
@@ -80,20 +91,50 @@ private Expression.WindowFunction.Bound.Builder setBound(
...
+          }
+        } else if (boundType.foldable()) {
```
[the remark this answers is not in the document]
WangGuangxin: For clickhouse backend and velox backend with ROW frame. Comments added

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
```diff
@@ -80,20 +92,53 @@ private Expression.WindowFunction.Bound.Builder setBound(
...
+                  Expression.WindowFunction.Bound.Preceding.newBuilder();
+              offsetPrecedingBuilder.setOffset(0 - offset);
+              builder.setPreceding(offsetPrecedingBuilder.build());
+            } else {
+              Expression.WindowFunction.Bound.Following.Builder offsetFollowingBuilder =
+                  Expression.WindowFunction.Bound.Following.newBuilder();
+              offsetFollowingBuilder.setOffset(offset);
+              builder.setFollowing(offsetFollowingBuilder.build());
+            }
+          } catch (NumberFormatException e) {
```
PHILO-HE, line 137: Looks we can also remove this try/catch.

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
```diff
@@ -80,20 +92,53 @@ private Expression.WindowFunction.Bound.Builder setBound(
...
+            throw new UnsupportedOperationException(
+                "Unsupported Window Function Frame Type:" + boundType);
           }
-        } catch (NumberFormatException e) {
+        } else {
           throw new UnsupportedOperationException(
               "Unsupported Window Function Frame Type:" + boundType);
```
PHILO-HE, line 143: Not frame type, but bound type, please help fix it in your pr.
```
"Unsupported Window Function Frame Bound Type: " + boundType
```

PHILO-HE (commented):
Two minor comments. We can merge it after CI is fixed. Thanks!

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
WangGuangxin: done

gluten-core/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
WangGuangxin: done

WangGuangxin:
@PHILO-HE @rui-mo Can you please review it again?

PHILO-HE:
> @PHILO-HE @rui-mo Can you please review it again?

@WangGuangxin, the pr looks good to me. Could you fix the CI failure? Looks relevant to this pr.

PHILO-HE:
@WangGuangxin, please also rebase the code. Thanks!

gluten-core/src/main/scala/org/apache/gluten/utils/PullOutProjectHelper.scala
```diff
@@ -143,8 +144,39 @@ trait PullOutProjectHelper {
...
+      orderSpec: SortOrder,
+      expressionMap: mutable.HashMap[Expression, NamedExpression]): Expression = {
+    bound match {
+      case _: PreComputeRangeFrameBound => bound
+      case _ if !bound.foldable => bound
+      case _ if bound.foldable =>
+        val a = expressionMap
+          .getOrElseUpdate(
+            bound.canonicalized,
+            Alias(Add(orderSpec.child, bound), generatePreAliasName)())
```
ulysses-you, line 166: Do we really support reuse the reference for the same bound literal ? What happens if the two window expressions have the same frame but requires different ordering column.
```
  sum(x) OVER (
    PARTITION BY c1
    ORDER BY c2 RANGE BETWEEN 1 preceding AND CURRENT ROW
  ), 
  sum(x) OVER (
    PARTITION BY c1
    ORDER BY c3 RANGE BETWEEN 1 preceding AND CURRENT ROW
  )
```

WangGuangxin: > Do we really support reuse the reference for the same bound literal ? What happens if the two window expressions have the same frame but requires different ordering column.
> 
> ```
>   sum(x) OVER (
>     PARTITION BY c1
>     ORDER BY c2 RANGE BETWEEN 1 preceding AND CURRENT ROW
>   ), 
>   sum(x) OVER (
>     PARTITION BY c1
>     ORDER BY c3 RANGE BETWEEN 1 preceding AND CURRENT ROW
>   )
> ```

let me check

WangGuangxin: The frame bouding is a reference, which will be pre-project before window operator. And the sort/partition are calculated in window operator. So it's ok to reuse it.
I add a UT for this.

WangGuangxin: The new added UT succeed in spark3.2, but failed in spark3.4/3.5. I'll fix it

WangGuangxin: Sorry for late response, all UT passed now, the failed one seems unrelated. Please help review again @PHILO-HE @ulysses-you 

The reason why it was failed is because  sort column `l_discount` is a decimal, and when we do `Add(sort_col, offset)`, the precision of decimal type will change, so that the data type check failed in Spark (https://github.com/apache/spark/blob/82a84ede6a47232fe3af86672ceea97f703b3e8a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L83)

In fact, we only support `ByteType | ShortType | IntegerType | LongType | DateType` by design, which is checked in https://github.com/apache/incubator-gluten/pull/5626/files#diff-68925b6a5ee67a942a976971ae6727a45a0b166d200d8d983454169e7c3bd1f6R316, but even when WindowExec fallback by this check, the `PullOutPreProjection` rule will still be executed.

ulysses-you (approved):
thank you @WangGuangxin

PR #5632, opened 2024-05-07
DNM

acvictor:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


FelixYBW:
Can you comment a bit how these info are used by Velox? Is it aligned with Spark's behavior?

gaoyangxiaozhu:
can't we leverage `metadatacolumns` for this since it already contains `fileSize` and `modifyTime` ?

acvictor:
> can't we leverage `metadatacolumns` for this since it already contains `fileSize` and `modifyTime` ?

Does `metadatacolumns` always contain file size and modification time? I understood it as containing these only if they are being queried. In this case I want it to always be passed in the split.

acvictor:
> Can you comment a bit how these info are used by Velox? Is it aligned with Spark's behavior?

In Spark, properties are present in the split and obtained during listing. Earlier Velox would make one additional call to remote storage to fetch file length per openFileForRead call on a path even though this information is already present upstream wihle constructing the split. Velox now allows these values to be passed from the caller. By making this change we can eliminate one additional call to remote storage per path (RTT is in the order of 10s of milli seconds).

PR #5633, opened 2024-05-07
[VL] Add test for getbit Spark function

acvictor:

Add a test for `getbit` which an alias for `bit_get` Spark function.

## How was this patch tested?

Added a UT.


backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
```diff
@@ -41,6 +41,12 @@ class ScalarFunctionsValidateSuite extends FunctionsValidateTest {
     }
   }
 
+  test("Test getbit function") {
+    runQueryAndCompare("SELECT getbit(l_partkey, 0) from lineitem limit 1") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
```
PHILO-HE, line 46: I think it's better to move this into the above test. You can rename that test name to cover `getbit`.

acvictor: Done!

PHILO-HE (approved):
Thanks!

PR #5634, opened 2024-05-07
[VL] Spark width_bucket function support

gaoyangxiaozhu:
velox already support  `width_bucket`  function

#4039

## How was this patch tested?


gaoyangxiaozhu:
fail due to Velox `width_bucket` treat `bucket_number <=0` as a user error  (invalid parameter) https://github.com/facebookincubator/velox/blob/668d578420f20568e2f6c152c8eed77d3ad64c59/velox/functions/prestosql/Arithmetic.h#L384C1-L388C75 however spark would return null for this case https://github.com/apache/spark/blob/08c6bb9bf32f31b5b9870d56cc4c16ab97616da6/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/mathExpressions.scala#L1749

@zhouyuan / @zhli1142015 , may basic c++ question, do you how to modify velox widthBatch function to let it support return null value for those specifal cases as result type defination is int64_t.

FelixYBW:
> hose specifal cases as the result type already be set to
Velox uses Presto implementation. You may need to create a new function for Velox/Spark, and pass the reference of null set to the function.

gaoyangxiaozhu:
@rui-mo help review , thanks!

rui-mo (approved):
Thanks.

Issue #5309, opened 2024-04-07
[VL] Spark 3.5 Unit Tests track


yma11:
### Description

UTs for Spark3.5 are not ported yet. This ticket will track:
1) add them under `gluten-ut/spark3.5`.
2) fix failed UTs.
3) enable them in CI/CD.

**Update on 5/8:**
Currently all gaps for Spark3.5 UTs are filled. The failed UTs include: 

- [ ] reading _tmp_metadata_row_index - not present in a table
- [ ] reading _tmp_metadata_row_index - present in a table
- [ ] GlutenParquetRowIndexSuite

FelixYBW:
Not fully solved, rely on PR
https://github.com/facebookincubator/velox/pull/9174

yma11:
Currently all gaps for Spark3.5 UTs are filled. The failed UTs include: `reading _tmp_metadata_row_index - not present in a table`, `reading _tmp_metadata_row_index - present in a table` and `GlutenParquetRowIndexSuite`.

yma11:
Hi @gaoyangxiaozhu, the left failed UTs for Spark3.5 are all about row index metadata and I see your support [PR](https://github.com/facebookincubator/velox/commit/49c3ebb2028c8db380a5a414538baba38f14ef7b) is already merged. Will there be anything need to be done at Gluten side and can you help on these UTs? Thanks.

gaoyangxiaozhu:
> Hi @gaoyangxiaozhu, the left failed UTs for Spark3.5 are all about row index metadata and I see your support [PR](https://github.com/facebookincubator/velox/commit/49c3ebb2028c8db380a5a414538baba38f14ef7b) is already merged. Will there be anything need to be done at Gluten side and can you help on these UTs? Thanks.

the failure of `GlutenParquetRowIndexSuite` is not related my velox change, but the issue of parquet native write. Let me send another PR to enable `GlutenParquetRowIndexSuite` firstly.

gaoyangxiaozhu:
The PR for enable `GlutenParquetRowIndexSuite` https://github.com/apache/incubator-gluten/pull/5740

PR #5637, opened 2024-05-07
[GLUTEN-5603] Add new added Spark3.4 UTs in Gluten for Spark3.5

yma11:

Add the new added UTs in Spark3.4 for Gluten Spark35 profile

## How was this patch tested?

CI


FelixYBW:
If there are failed UTs, let's track here:
https://github.com/apache/incubator-gluten/issues/5309

PR #5641, opened 2024-05-07
[VL][DOC] Suggest enable vcpkg for docker build in centos

zml1206:
As title.

## How was this patch tested?
Local build test.

### DIFF
--- a/docs/developers/docker_centos7.md
+++ b/docs/developers/docker_centos7.md
@@ -26,6 +26,7 @@ yum -y install \
     ninja-build \
     wget \
     ca-certificates \
+    patch \
     sudo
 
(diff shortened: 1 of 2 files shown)


FelixYBW:
Actually we are going to enable vcpkg=on as default. The dynamic build script is still there but CI won't track it.
@weiting-chen

ulysses-you:
@FelixYBW @weiting-chen @zhouyuan  I believe there is an issue https://github.com/apache/incubator-gluten/pull/5033 before going to enbale vcpkg by default ?

zml1206:
Enable vcpkg=on as default is going on, so close it.

PR #5648, opened 2024-05-08
[GLUTEN-2620][VL] Enable compile_arrow_java by default to avoid invalid pointer error

PHILO-HE:

Though the reported issue is on centos-7 by now, it's still possible that user encounters such issue on other OS. So let's directly enable this option by default. 

## How was this patch tested?

Verified by customer.


PHILO-HE:
@zhouyuan, @weiting-chen, please take a review. Thanks!

Yohahaha (approved):

zhouyuan:
Note: the new CSV reader and the comming json/text support should also requires arrow java setup, these functions are in different scripts now, we may need to have a new script for this purpose - can be done as the next commit
https://github.com/apache/incubator-gluten/blob/main/ep/build-velox/src/build_arrow_deps_centos8.sh

PR #5650, opened 2024-05-08
[VL] Fix clang-format version

PHILO-HE:
~clang-format-12 is recommended to use according to Gluten doc, but the version used by CI is 15.
See:
https://github.com/apache/incubator-gluten/blob/main/dev/formatcppcode.sh~

Fix inconsistency in script and doc.

### DIFF
--- a/.github/workflows/code_style.yml
+++ b/.github/workflows/code_style.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Run clang-format style check for C/C++ programs.
         uses: jidicula/clang-format-action@v4.11.0
         with:
-          clang-format-version: '15'
+          clang-format-version: '12'
           check-path: ${{ matrix.path['check'] }}
           fallback-style: 'Google' # optional
 
--- a/.github/workflows/code_style.yml
+++ b/.github/workflows/code_style.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Run clang-format style check for C/C++ programs.
         uses: jidicula/clang-format-action@v4.11.0
         with:
-          clang-format-version: '12'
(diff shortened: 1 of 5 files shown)


PHILO-HE:
@yma11, could you take a look? Thanks!

yma11:
What's the difference between `12` and `15`? `15` is used as consistent with CH backend.

PHILO-HE:
> What's the difference between `12` and `15`? `15` is used as consistent with CH backend.

Certainly, there are some differences between 12 and 15. It's better to use one uniform version. I think either 12 or 15 is acceptable. If 15 is used, format script and doc needs to be fixed.

@liuneng1994, is there any reason for using 15 for CH cpp code?

baibaichen:
> clang-format-12 is recommended to use, but the version used by CI is 15.

why use `12`?

PHILO-HE:
> > clang-format-12 is recommended to use, but the version used by CI is 15.
> 
> why use `12`?

No strong reason. Just note a gluten velox doc recommends to use 12, but CI uses 15. Maybe, we can uniformly use 15. I will fix the inconsistency.

liuneng1994:
> > What's the difference between `12` and `15`? `15` is used as consistent with CH backend.
> 
> Certainly, there are some differences between 12 and 15. It's better to use one uniform version. I think either 12 or 15 is acceptable. If 15 is used, format script and doc needs to be fixed.
> 
> @liuneng1994, is there any reason for using 15 for CH cpp code?

There is no special reason. CH initially used clang-format-15.

yma11 (approved):
nit: update the PR description for new change.

PR #5649, opened 2024-05-08
[VL] Add a bad test case that final aggregate of collect_list is fallen back while partial aggregate is not

zhztheplayer:
With whole stage fallback enabled, we can manage to create a plan (AE=on) like:

```
Exchange SinglePartition, ENSURE_REQUIREMENTS, [plan_id=393]
+- HashAggregate(keys=[], functions=[partial_sum(ele#42L)], output=[sum#48L])
   +- ObjectHashAggregate(keys=[c1#14], functions=[collect_list(c2#15L, 0, 0)], output=[ele#42L])
      +- VeloxColumnarToRowExec
         +- AQEShuffleRead coalesced
            +- ShuffleQueryStage 0
               +- ColumnarExchange hashpartitioning(c1#14, 5), ENSURE_REQUIREMENTS, [c1#14, buf#50], [plan_id=315], [id=#315], [OUTPUT] List(c1:IntegerType, buf:ArrayType(LongType,false)), [OUTPUT] List(c1:IntegerType, buf:ArrayType(LongType,false))
                  +- ^(3) ProjectExecTransformer [hash(c1#14, 42) AS hash_partition_key#56, c1#14, buf#50]
                     +- ^(3) FlushableHashAggregateTransformer(keys=[c1#14], functions=[partial_collect_list(c2#15L, 0, 0)], output=[c1#14, buf#50])
                        +- ^(3) NativeFileScan parquet spark_catalog.default.tmp1[c1#14,c2#15L] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/opt/gluten/spark-warehouse/org.apache.gluten.execution.FallbackS..., PartitionFilters: [], PushedFilters: [], ReadSchema: struct<c1:int,c2:bigint>
```

In the plan, only partial aggregate phase of aggregate function `collect_list` is offloaded to native. 

The we got error:

```
java.lang.NullPointerException
	at org.apache.spark.sql.catalyst.expressions.aggregate.Collect.deserialize(collect.scala:81)
	at org.apache.spark.sql.catalyst.expressions.aggregate.Collect.deserialize(collect.scala:39)
	at org.apache.spark.sql.catalyst.expressions.aggregate.TypedImperativeAggregate.merge(interfaces.scala:592)
	at org.apache.spark.sql.execution.aggregate.AggregationIterator$$anonfun$1.$anonfun$applyOrElse$3(AggregationIterator.scala:200)
[...]

PR #5655, opened 2024-05-08
[VL] Fix NullPointerException when collect_list / collect_set are partially fallen back

zhztheplayer:
Update: Fixes https://github.com/apache/incubator-gluten/pull/5649. Added vanilla implementation of velox_collect_list and velox_collect_set.

Velox backend's collect_list / collect_set implementations require for ARRAY intermediate data however Spark uses BINARY. To address this we did some tricks to forcibly modify the physical plan to change the output schema of partial aggregate operator to align with Velox, but that way the actual information for the two functions in Velox backend is still hidden from query plan so advanced optimizations or compatibility checks are made difficult during planning phase.

This patch adds new functions velox_collect_list / velox_collect_set to correctly map to Velox backend's implementation for the two functions and does essential code cleanup and refactors.

1. Add functions velox_collect_list / velox_collect_set.
2. Remove physical rule `RewriteCollect` / `RewriteTypedImperativeAggregate`, add logical rule `CollectRewriteRule` to incorporate functionalities of the formers.
3. `CollectRewriteRule` will replace collect_list / collect_set with velox_collect_list / velox_collect_set.
4. Since velox_collect_list / velox_collect_set becomes `DeclarativeAggregate`, some UTs should be disabled as they do some plan checks for existence of `ObjectHashAggregateExec`.
5. Some optimizations for UT facility code.


zhouyuan:
CC @zhli1142015

zhztheplayer:
@liujiayi771

zhztheplayer:
@PHILO-HE

liujiayi771:
@zhztheplayer  Do you think this approach is better, or is it better for us to convert it to binary in the project during row extraction?

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/GlutenDataFrameAggregateSuite.scala
```diff
@@ -279,4 +283,55 @@ class GlutenDataFrameAggregateSuite extends DataFrameAggregateSuite with GlutenS
       randn(Random.nextLong())
     ).foreach(assertNoExceptions)
   }
+
+  Seq(true, false).foreach {
+    value =>
+      testGluten(s"SPARK-31620: agg with subquery (whole-stage-codegen = $value)") {
```
PHILO-HE, line 289: Do we need to also add this rewritten test for other spark versions? Generally, we do that even though duplicate code is introduced.

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
```diff
@@ -106,15 +112,38 @@ class FallbackSuite extends VeloxWholeStageTransformerSuite with AdaptiveSparkPl
     }
   }
 
-  // java.lang.NullPointerException
-  ignore("fallback final aggregate of collect_list") {
+  test("fallback collect_list") {
+    withSQLConf(
+      GlutenConfig.EXPRESSION_BLACK_LIST.key -> "collect_list"
```
PHILO-HE, line 117: Is it possible to only make partial fallback happen in the test? Maybe, needs a config to control.

PHILO-HE (commented):
Thanks!

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
```diff
@@ -0,0 +1,65 @@
...
+  protected lazy val buffer: AttributeReference = AttributeReference("buffer", dataType)()
+
+  override def dataType: DataType = ArrayType(child.dataType, false)
+
+  override def aggBufferAttributes: Seq[AttributeReference] = List(buffer)
+
+  override lazy val initialValues: Seq[Expression] = List(Literal.create(Seq.empty, dataType))
+
+  override lazy val updateExpressions: Seq[Expression] = List(
+    ArrayUnion(buffer, CreateArray(List(child), useStringTypeWhenEmpty = false))
```
ulysses-you, line 34: For each row, we wrap a array. Will it cause performance regression with vanilla Spark ?

ulysses-you: Does the semaintics of `ArrayUnion` equal to `Collect` ?
- `ArrayUnion` always returns distinct value in an array
- `ArrayUnion` allows null but CollectList/CollectSet do not allow

ulysses-you (commented):
thank you @zhztheplayer , can you also provide the npe case to help reproduce ? It seems this pr did not contain that case ?

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/GlutenDataFrameAggregateSuite.scala
zhztheplayer: > Do we need to also add this rewritten test for other spark versions? Generally, we do that even though duplicate code is introduced.

For this case I felt it's OK to just test with one version (as long as it's newer version, e.g., 3.4, 3.5). `collect_list` / `collect_set` in Spark shouldn't have actual differences among different Spark versions. Having it tested with all Spark version definitely provides better coverage but we can also make trade-off taking maintenance effort into account.

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
zhztheplayer: One of the following tests is for partial fallback:

https://github.com/apache/incubator-gluten/pull/5655/files/c40a5ad359fe14b030f85627e0d983a2848c8bd6#diff-493b5fa71a5befc5adcd67b51e47e802b41870287619a7b752bbf85f53b1880cR139

Would you mean that?

zhztheplayer:
> thank you @zhztheplayer , can you also provide the npe case to help reproduce ? It seems this pr did not contain that case ?

NPE can be repeated from current main code test by enabling some ignored tests.

See https://github.com/apache/incubator-gluten/pull/5649.

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
zhztheplayer: > For each row, we wrap a array. Will it cause performance regression with vanilla Spark ?

Yes that's possible. But I assume it should be a comparatively rare case when user enables Gluten however these two functions are fallen back. So I think it's fine until getting real complaints from users.

zhztheplayer: > ArrayUnion always returns distinct value in an array

Oops, if it does de-dup then `velox_collect_list` may return wrong result. I'll take a look on this.

zhztheplayer: So [concat](https://spark.apache.org/docs/latest/api/sql/index.html#concat) could be the right choice here?

ulysses-you: concat allows null but collect_list does not allow:
- the result is null
- the null element value in array

I think we should at least pass test `collect_set/collect_list with null` in `VeloxAggregateFunctionsSuite` with partially fallback.

ulysses-you: Instead of adding more cutsom Gluten expression, cann't we just add project to adapt data type (convert array to binary, etc) in columnar rule ?

zhztheplayer: Sure, let me add more test against this case.

zhztheplayer:
> @zhztheplayer Do you think this approach is better, or is it better for us to convert it to binary in the project during row extraction after partial agg?

I am not 100% sure but seems to be that the PR's approach looks more portable? Row extraction would work but if it's done at operator level, the function still outputs ARRAY intermediate data which doesn't match Spark's CollectList / CollectSet definition.

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
zhztheplayer: @ulysses-you Please take a look again. 

Change:
https://github.com/apache/incubator-gluten/pull/5655/files#diff-1ebbabef839649a226bf6bed298759976ab8adafe31d3dbfe07c559a1a95842dR33-R38

Test case:
https://github.com/apache/incubator-gluten/pull/5655/files#diff-742ce9027f0d774d324125a130d9993bad09d4ea5eb6d35d791b39bb57138c77R983-R1057

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
zhztheplayer: Added a set of cases here https://github.com/apache/incubator-gluten/pull/5655/files#diff-742ce9027f0d774d324125a130d9993bad09d4ea5eb6d35d791b39bb57138c77R983-R1057 for better coverage.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
```diff
@@ -0,0 +1,123 @@
...
+  override def apply(plan: LogicalPlan): LogicalPlan = LogicalPlanSelector.maybe(spark, plan) {
+    val out = plan.transformUp {
+      case node =>
+        val out = replaceCollectSet(replaceCollectList(node))
+        out
+    }
+    if (out.fastEquals(plan)) {
+      return plan
+    }
+    spark.sessionState.analyzer.checkAnalysis(out)
```
ulysses-you, line 48: what does this check for ? I did not see Spark call it at optimizer phase...

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
```diff
@@ -0,0 +1,68 @@
...
+  )
+
+  override lazy val mergeExpressions: Seq[Expression] = List(
+    Concat(List(buffer.left, buffer.right))
+  )
+
+  override def defaultResult: Option[Literal] = Option(Literal.create(Array(), dataType))
+}
+
+case class VeloxCollectSet(override val child: Expression) extends VeloxCollect {
```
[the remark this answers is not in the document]
zhztheplayer: It internally uses Array. A single de-dup operation should cost ~O(n). I would do a distinct at end of aggregation.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
```diff
@@ -0,0 +1,123 @@
...
+      case agg: Aggregate =>
+        agg.transformExpressions {
+          case ToVeloxCollectSet(newAggFunc) =>
+            val out = ensureNonNull(newAggFunc)
+            out
+        }
+      case w: Window =>
+        w.transformExpressions {
+          case func @ WindowExpression(ToVeloxCollectSet(newAggFunc), _) =>
+            val out = ensureNonNull(func.copy(newAggFunc))
```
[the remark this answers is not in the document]
zhztheplayer: That doesn't work. Spark has some checks in `checkAnalysis` to enforce "WindowExpression(WindowFunction, ...)" pattern.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
```diff
@@ -0,0 +1,123 @@
...
+    //    vanilla Spark.
+    // 2. Filter out null inputs from VeloxCollectSet to align with vanilla Spark.
+    //
+    // Since https://github.com/apache/incubator-gluten/pull/4805
+    node match {
+      case agg: Aggregate =>
+        agg.transformExpressions {
+          case ToVeloxCollectSet(newAggFunc) =>
+            val out = ensureNonNull(newAggFunc)
+            out
```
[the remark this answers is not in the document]
zhztheplayer: I didn't find a proper way for placing breakpoint for debugging code without doing this. Do you know some?

If there is not a good way, Personally I would keep something like this and it doesn't increase code complexity.

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
```diff
@@ -106,15 +112,44 @@ class FallbackSuite extends VeloxWholeStageTransformerSuite with AdaptiveSparkPl
...
+        assert(collect(plan) { case v: HashAggregateExecBaseTransformer => v }.isEmpty)
+      }
+    }
+  }
+
+  test("fallback collect_set") {
+    withSQLConf(
+      GlutenConfig.EXPRESSION_BLACK_LIST.key -> "collect_set"
+    ) {
+      CollectRewriteRule.forceEnableAndRun {
```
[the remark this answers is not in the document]
zhztheplayer: I'll remove this API and the two tests. They are not useful for now.

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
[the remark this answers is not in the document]
ulysses-you: It's mainly for reduce shuffle data if the partial mode aggreagte fallback.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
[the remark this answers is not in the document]
ulysses-you: I usually dig the internal method returned value.

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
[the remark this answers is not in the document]
zhztheplayer: Yes, It costs more space. May be we can use Spark Map type for collect_set?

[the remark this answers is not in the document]
ulysses-you: Is it possible to fully copy `Collect` code and change `serialize`, `deserialize` to adapt datatype ? It seems we can re-point binary to a UnsafeArrayData and then getArray.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
zhztheplayer: I'll try removing this. It should not be required with current approach either.

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
```diff
@@ -0,0 +1,68 @@
...
+  override def prettyName: String = "velox_collect_set"
+
+  override def nullable: Boolean = true
```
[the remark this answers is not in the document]
zhztheplayer: Added some comment to VeloxCollectSet.

For now we don't need the `withNullability` part of the comment.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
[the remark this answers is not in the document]
zhztheplayer: It's evaluated with the project created in `WindoeExecBase#createResultProjection`. Velox's window implementation doesn't support `collect_set` yet. I'd add a case for vanilla Spark + velox_collect_set if you are concerned about this part of code.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
```diff
@@ -0,0 +1,106 @@
...
+      case aggFunc @ AggregateExpression(s: CollectSet, _, _, filter, _) if has[VeloxCollectSet] =>
+        val newFilter = (filter ++ Some(IsNotNull(s.child))).reduceOption(And)
+        val newAggFunc =
+          aggFunc.copy(aggregateFunction = VeloxCollectSet(s.child), filter = newFilter)
+        Some(newAggFunc)
+      case _ => None
+    }
+  }
+
+  private def has[T <: Expression: ClassTag]: Boolean = {
```
ulysses-you, line 102: After droping test, do we still need this check ? It seems this rule is only added by velox backend and the same is VeloxCollectSet. So when we going to this rule, there must have VeloxCollectSet ?

zhztheplayer: I agree with removing the check. Although we can do that later in some version... 

For example `VeloxCollectSet` is still expensive than vanilla Spark so removing this check may cause a user who disabled `collect_set` experience performance regression.

BTW I'll raise some optimizations on `VeloxCollectSet` after this patch is merged. When I am confident with vanilla Spark + velox_collect_list / velox_collect_set being used in production enough, I'll remove this check.

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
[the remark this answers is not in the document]
zhztheplayer: Isn't binary data only able to obtain when it's a `ImperativeAggregate`? Would you elaborate on the suggestion?

JkSelf (approved):
LGTM. Thanks.

ulysses-you (approved):

zhztheplayer:
Thank you all for reviewing!

Issue #5656, opened 2024-05-08
[CORE][VL] Subqueries are executed during explain phase for complex result type


zhztheplayer:
We usually do a [tentative offload](https://github.com/apache/incubator-gluten/blob/c9018cdd884c5e93a171ba1561888e719026898c/gluten-core/src/main/scala/org/apache/gluten/execution/BasicPhysicalOperatorTransformer.scala#L121-L127) while validating query plans.

Given that in Gluten, offloading a subquery expression may cause `collect` to be called immediately for complex types (MAP, ARRAY, STRUCT) ([example](https://github.com/apache/incubator-gluten/blob/c9018cdd884c5e93a171ba1561888e719026898c/gluten-core/src/main/scala/org/apache/gluten/expression/ScalarSubqueryTransformer.scala#L37)), thus in the case, plan validation always executes subqueries, so explaining always executes subqueries.

This is not a big issue as of now but can be considered something counter-intuitive. We may need to rework the validation procedure a little bit to fix this.

Related to : https://github.com/apache/incubator-gluten/pull/1202

PR #5658, opened 2024-05-08
[GLUTEN-5656][CORE] Avoid executing subqueries with complex data type during validation

zhztheplayer:
Fixes https://github.com/apache/incubator-gluten/issues/5656

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/expression/ScalarSubqueryTransformer.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/expression/ScalarSubqueryTransformer.scala
@@ -20,17 +20,13 @@ import org.apache.gluten.substrait.expression.{ExpressionBuilder, ExpressionNode
 
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.execution.{BaseSubqueryExec, ScalarSubquery}
-import org.apache.spark.sql.types.{ArrayType, DataType, MapType, StructType}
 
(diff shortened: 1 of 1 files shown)


zhztheplayer:
There is an error may not relate to this change 

```
GlutenOrcV2SchemaPruningSuite:
- Spark vectorized reader - without partition data column - select only top-level fields !!! IGNORED !!!
- Spark vectorized reader - with partition data column - select only top-level fields !!! IGNORED !!!
- Non-vectorized reader - without partition data column - select only top-level fields
- Non-vectorized reader - with partition data column - select only top-level fields
- Spark vectorized reader - without partition data column - select a single complex field with disabled nested schema pruning !!! IGNORED !!!
- Spark vectorized reader - with partition data column - select a single complex field with disabled nested schema pruning !!! IGNORED !!!
- Non-vectorized reader - without partition data column - select a single complex field with disabled nested schema pruning !!! IGNORED !!!
- Non-vectorized reader - with partition data column - select a single complex field with disabled nested schema pruning !!! IGNORED !!!
- Spark vectorized reader - without partition data column - select only input_file_name()
- Spark vectorized reader - with partition data column - select only input_file_name()
- Non-vectorized reader - without partition data column - select only input_file_name()
- Non-vectorized reader - with partition data column - select only input_file_name()
pure virtual method called
terminate called without an active exception
```

https://github.com/apache/incubator-gluten/actions/runs/8998521030/job/24718799749?pr=5658

Yohahaha (approved):
great fix, thank you!

PR #5661, opened 2024-05-08
[VL] Generate hdfs-client.xml for libhdfs

ulysses-you:

When using Gluten + HDFS, we have to make a hdfs-client.xml manually and specify it by --files. This pr makes hdfs-client.xml when initializing Gluten driver plugin so that people do not need do anything.  It supports yarn/k8s master (does not support local).

## How was this patch tested?

manually test


ulysses-you:
cc  @zhztheplayer  @PHILO-HE @Yohahaha thank you

yaooqinn (approved):

PHILO-HE (commented):
After the generation, do we still depend on the setting for LIBHDFS3_CONF to let libhdfs3 obtain the config?

Also posted a stale similar pr here.
https://github.com/apache/incubator-gluten/pull/2130

ulysses-you:
@PHILO-HE , no,  we do not need it. libhdfs would read hdfs-client.xml by default in working directory and that's where this pr generates hdfs-client.xml.

PR #5667, opened 2024-05-09
[GLUTEN-5414] [VL] Move ArrowFileScanExec class to module backends-velox

jinchengchenghh:
Arrow related class should not be in module gluten-core


zhztheplayer:
@jinchengchenghh Thanks for keeping improving the code.

My suggestion: Let's see if it's feasible to directly move ArrowFileSourceScanExec into module `backends-velox`.

Then move [this part of code](https://github.com/apache/incubator-gluten/blob/69fa591d48f6e085e2b4059c1588b845d34c5d1f/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala#L300-L307) into a individual physical rule plugged into Gluten through [this API](https://github.com/apache/incubator-gluten/blob/69fa591d48f6e085e2b4059c1588b845d34c5d1f/gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala#L412).

jinchengchenghh:
https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala#L53

If I move `ArrowFileSourceScanExec` to backends-velox, here I won't access it

jinchengchenghh:
And the genExtendedColumnarValidationRules may take effect after `add c2r r2c`, so the test failed

zhztheplayer:
> https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala#L53
> 
> If I move `ArrowFileSourceScanExec` to backends-velox, here I won't access it

Am I missing something? I assume `ArrowFileSourceScanExec` should return `true` for `#outputNativeColumnarData`?

jinchengchenghh:
Can you help review this one again? Thanks! @zhztheplayer

zhztheplayer (approved):

PR #5670, opened 2024-05-09
[GLUTEN-5682][VL] Fix incorrect result when isNull & isNotNull coexist in filter

zjuwangg:
Fix the bug founded in issue (Fixes: \#5682)

## How was this patch tested?

Add ut to cover the change!

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.h
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.h
@@ -316,6 +316,10 @@ class SubstraitToVeloxPlanConverter {
       if (!initialized_) {
         initialized_ = true;
       }
+      forbidsNullSet_ = true;
+      if (isNullSet_) {
+        existIsNullAndIsNotNull_ = true;
+      }
     }
 
     // Only null is allowed.
@@ -324,6 +328,10 @@ class SubstraitToVeloxPlanConverter {
       nullAllowed_ = true;
       if (!initialized_) {
         initialized_ = true;
+      }
+      isNullSet_ = true;
+      if (forbidsNullSet_) {
+        existIsNullAndIsNotNull_ = true;
       }
     }
 
@@ -375,6 +383,9 @@ class SubstraitToVeloxPlanConverter {
 
     bool nullAllowed_ = false;
     bool isNull_ = false;
+    bool forbidsNullSet_ = false;
+    bool isNullSet_ = false;
+    bool existIsNullAndIsNotNull_ = false;
 
     // If true, left bound will be exclusive.
     std::vector<bool> lowerExclusives_;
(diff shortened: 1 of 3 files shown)


FelixYBW:
Is it only a Spark UT issue? or observed in some SQL?

zjuwangg:
> Is it only a Spark UT issue? or observed in some SQL?

It's not just a Spark UT issue. We encounter this bug causing result not mismatch.
We can reproduce this bug by executing following sql in VL moudule.
`select l_orderkey from lineitem where l_comment is null and l_comment is not null`

image

zjuwangg:
velox related change has been merged via https://github.com/facebookincubator/velox/pull/9718

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -375,6 +383,9 @@ class SubstraitToVeloxPlanConverter {
 
     bool nullAllowed_ = false;
     bool isNull_ = false;
+    bool forbidsNullSet_ = false;
+    bool isNullSet_ = false;
+    bool existIsNullAndIsNotNull_ = false;
```
PHILO-HE, line 388: Can we just depend on the check for `forbidsNullSet_ && isNullSet_` instead of introducing a new flag?

PHILO-HE (commented):
Thanks for your fix! Just one comment. Thanks!

cpp/velox/substrait/SubstraitToVeloxPlan.h
zjuwangg: I'll address this later soon.

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -375,6 +377,8 @@ class SubstraitToVeloxPlanConverter {
 
     bool nullAllowed_ = false;
     bool isNull_ = false;
+    bool forbidsNullSet_ = false;
+    bool isNullSet_ = false;
```
PHILO-HE, line 381: Is `forbidsNullSet_` logically equivalent to `!nullAllowed`? And `isNullSet_` equivalent to `isNull_`? If true, we can just use the existing flags. Thanks!

zjuwangg: Actually not. initialized_ can be set other place not just here two null related setter method.

zjuwangg: > Actually not. initialized_ can be set other place not just here two null related setter method.

So we can not according to `initialized_ ` variable to ensure the null behavior

PHILO-HE (approved):
Looks good! Thanks!

Issue #5673, opened 2024-05-09
[VL] Unexpected memory issue found in CI


Yohahaha:
### Backend

VL (Velox)

### Bug description

```
W20240509 07:32:15.819299 59531 MemoryAllocator.cpp:186] [MEM] Exceeded memory reservation limit when reserve 1 new pages when allocate 1 pages
E20240509 07:32:15.819386 59531 Executor.cpp:31] ThreadPoolExecutor: func threw unhandled gluten::GlutenException: Unexpected: Failed to reserve 1048576 bytes although there is enough space, free bytes: 1048576
```

https://github.com/apache/incubator-gluten/actions/runs/9013102975/job/24763373229?pr=5672

### Spark version

None

Yohahaha:
We can give `memoryPoolInitCapacity` a proper default value(32MB) to decrease memory request times, and may relieve concurrent memory request from DBI.

And could combine `memoryPoolTransferCapacity_` with `spark.gluten.memory.reservationBlockSize` then remove AllocationListener's reservation compute logic.

FelixYBW:
@zhztheplayer

PR #5674, opened 2024-05-09
[GLUTEN-5673][VL] Fix arbitrator grow logic when exist concurrent memory request

Yohahaha:

Connector IO thread executor is enabled by default and be used in DirectBufferedInput, which bring concurrent memory request issue when do arbitration.

One possible scenario: 
T1: thread 1 enter arbitration, get freeBytes and find it's ok to grow without capacity increment.
T2: thread 2 enter maybeIncrementReservation and increase reservationBytes with a small bytes requests successfully.
T3: thread 1 invoke mempool#grow() but current freeBytes is insufficient.

Solution:
try grow without capacity increment first, if failed, then shrink pool to prevent following memory request and do real grow. 

(Fixes: \#5673)

### DIFF
--- a/cpp/velox/memory/VeloxMemoryManager.cc
+++ b/cpp/velox/memory/VeloxMemoryManager.cc
@@ -95,16 +95,22 @@ class ListenableArbitrator : public velox::memory::MemoryArbitrator {
     // Since
     // https://github.com/facebookincubator/velox/pull/9557/files#diff-436e44b7374032f8f5d7eb45869602add6f955162daa2798d01cc82f8725724dL812-L820,
     // We should pass bytes as parameter "reservationBytes" when calling ::grow.
-    auto reclaimedFreeBytes = pool->shrink(0);
-    if (reclaimedFreeBytes >= bytes) {
-      bool ret = pool->grow(reclaimedFreeBytes, bytes);
-      VELOX_CHECK(ret, "{} failed to grow {} bytes, current state {}", pool->name(), velox::succinctBytes(bytes), pool->toString())
-      return ret;
+    auto freeByes = pool->freeBytes();
+    if (freeByes > bytes) {
+      if (pool->grow(0, bytes)) {
+        return 1;
+      }
     }
+    auto reclaimedFreeBytes = pool->shrink(0);
     auto neededBytes = bytes - reclaimedFreeBytes;
     listener_->allocationChanged(neededBytes);
     auto ret = pool->grow(bytes, bytes);
-    VELOX_CHECK(ret, "{} failed to grow {} bytes, current state {}", pool->name(), velox::succinctBytes(bytes), pool->toString())
+    VELOX_CHECK(
+        ret,
+        "{} failed to grow {} bytes, current state {}",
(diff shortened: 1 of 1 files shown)


zhztheplayer:
Thanks for helping investigate this @Yohahaha !

What I am curious about is that, since we are on single-threaded execution, each `ListenableArbitrator` should be only used by a single thread. By the PR description did you mean there will be multiple threads calling the arbitrator?

Yohahaha:
> Thanks for helping investigate this @Yohahaha !
> 
> What I am curious about is that, since we are on single-threaded execution, each `ListenableArbitrator` should be only used by a single thread. By the PR description did you mean there will be multiple threads calling the arbitrator?

DirectBufferedInput will allocate memory in each thread of connector IO executor, thus concurrent memory request may occur. 

Velox invoke maybeIncrementReservation when growCapacityCb success before, it's ok for ListenableArbitrator since it does not care pool's reservation incrementation, but after https://github.com/facebookincubator/velox/pull/9557, Arbitrator need commit reservation incrementation, so the issue happens.

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -94,20 +95,22 @@ class ListenableArbitrator : public velox::memory::MemoryArbitrator {
...
-      bool reserved = pool->grow(0, bytes);
-      GLUTEN_CHECK(
-          reserved,
-          "Unexpected: Failed to reserve " + std::to_string(bytes) +
-              " bytes although there is enough space, free bytes: " + std::to_string(freeBytes));
-      return 0;
+    auto freeByes = pool->freeBytes();
+    if (freeByes > bytes) {
+      if (pool->grow(0, bytes)) {
+        return 1;
```
zhztheplayer, line 101: What's the meaning of `return 1`? If the return value is unused, how about making the function return `void`?

zhztheplayer:
> > Thanks for helping investigate this @Yohahaha !
> > What I am curious about is that, since we are on single-threaded execution, each `ListenableArbitrator` should be only used by a single thread. By the PR description did you mean there will be multiple threads calling the arbitrator?
> 
> DirectBufferedInput will allocate memory in each thread of connector IO executor, thus concurrent memory request may occur.
> 
> Velox invoke maybeIncrementReservation when growCapacityCb success before, it's ok for ListenableArbitrator since it does not care pool's reservation incrementation, but after [facebookincubator/velox#9557](https://github.com/facebookincubator/velox/pull/9557), Arbitrator need commit reservation incrementation, so the issue happens.

Thank you for the explanation. I think we probably should align up with Velox community for arbitrator API. I feel the API is not constantly stable for `ListenableArbitrator` as we should frequently change our code to match their changes. :(

cpp/velox/memory/VeloxMemoryManager.cc
Yohahaha: ok, growCapacity must success in Gluten or throw OOM, we could omit the return value.

zhztheplayer (approved):

PR #5690, opened 2024-05-10
[VL] Refine memory changes logic when growCapacity

Yohahaha:

Changes:
1. add new config `spark.gluten.sql.columnar.backend.velox.memInitCapacity` to reduce memory arbitration times, default value is 8M.
2. set `memoryPoolTransferCapacity_` with `spark.gluten.memory.reservationBlockSize`, move reservation computation logic from SparkAllocationListener to Arbitrator.
3. refactor backend-wise config usage.


cpp/core/jni/JniCommon.h
```diff
@@ -348,47 +343,30 @@ class SparkAllocationListener final : public gluten::AllocationListener {
...
-      newBlockCount = 0;
-    } else {
-      // ceil to get the required block number
-      newBlockCount = (bytesReserved_ - 1) / blockSize_ + 1;
-    }
-    int64_t bytesGranted = (newBlockCount - blocksReserved_) * blockSize_;
-    blocksReserved_ = newBlockCount;
-    maxBytesReserved_ = std::max(maxBytesReserved_, bytesReserved_);
-    return bytesGranted;
-  }
```
zhztheplayer, line 365: Does `VeloxMemoryManager::arrowPool_` still use this code? Can we move it into another independent `gluten::AllocationListener` implementation which can wrap another listener? Like the way `BacktraceAllocationListener` adopted.

Yohahaha: I move such logic to ListenableMemoryAllocator.

PR #5711, opened 2024-05-12
[GLUTEN-4652][VL] Fix min_by/max_by result mismatch when RDD partition num > 1

zhouyifan279:

A follow-up of PR #5544 to fix min_by/max_by result mismatch when RDD partition num > 1.

Run SQL:
```
set spark.sql.leafNodeDefaultParallelism=2;
select min_by(a, b), max_by(a, b) from values (5, 6), (null, 11), (null, 5) test(a, b);
```
Current result is 
```
5      5
```
Expected result is
```
NULL    NULL
```

## How was this patch tested?

Modified test case `VeloxAggregateFunctionsSuite - min_by/max_by`


yma11:
I think I used to test the `"spark.default.parallelism", "2"` in previous fix and it worked. So what's the root cause of the failure as you tested? which part is the corresponding fix?

zhouyifan279:
@yma11 Did you reproduced the problem I mentioned [here](https://github.com/apache/incubator-gluten/pull/5544#issuecomment-2106285804) based on the master gluten branch?
That is what I'm trying to fix.

zhouyifan279:
> I think I used to test the `"spark.default.parallelism", "2"` in previous fix and it worked. So what's the root cause of the failure as you tested? which part is the corresponding fix?

I guess you didn't reproduce the problem because there was only 1 file under test table dir.
In that case, only 1 task existed in the Spark aggregation stage and the result was correct.

yma11:
> @yma11 Did you reproduced the problem I mentioned above based on the master gluten branch? That is what I'm trying to fix.

I mean which part of your code changes in this PR is the fix? Seems only some code refactor?

zhouyifan279:
> > @yma11 Did you reproduced the problem I mentioned above based on the master gluten branch? That is what I'm trying to fix.
> 
> I mean which part of your code changes in this PR is the fix? Seems only some code refactor?

Before this PR, `RowConstructorWithAllNullCallToSpecialForm.constructSpecialForm` was not called when function name was `row_constructor_with_all_null`.
I think it was related to C++ overload/override mechanism.

In this PR, I take function name as a member of `RowConstructorWithNullCallToSpecialForm` to avoid the overload/override issues and fix the result mismatch.

yma11:
> > > @yma11 Did you reproduced the problem I mentioned above based on the master gluten branch? That is what I'm trying to fix.
> > 
> > 
> > I mean which part of your code changes in this PR is the fix? Seems only some code refactor?
> 
> Before this PR, `RowConstructorWithAllNullCallToSpecialForm.constructSpecialForm`(the one defined in `RowConstructorWithAllNull.h`) was not called when function name was `row_constructor_with_all_null`. I think it was related to C++ overload/override mechanism.
> 
> In this PR, I take function name as a member of `RowConstructorWithNullCallToSpecialForm` to avoid the overload/override issues and fix the result mismatch.

That's interesting. Actually, I verified the previous code using real workload but with single parquet file so may not trigger your scenario. Thanks for your fix!

yma11 (approved):

zhouyifan279:
@yma11 CI `run-spark-test-spark34`'s failure seems irrelevant with this PR. Can you rerun it?

PR #5722, opened 2024-05-13
[VL] Add test for shuffle function

zhli1142015:
Add test for shuffle function.

## How was this patch tested?

UT.


zhli1142015:
cc @PHILO-HE and @rui-mo , thanks

rui-mo (approved):
Thanks.

gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -211,6 +211,13 @@ trait SparkPlanExecApi {
     GenericExpressionTransformer(substraitExprName, Seq(), original)
   }
 
+  def genShuffleTransformer(
+      substraitExprName: String,
+      child: ExpressionTransformer,
+      original: Shuffle): ExpressionTransformer = {
+    GenericExpressionTransformer(substraitExprName, Seq(child), original)
```
PHILO-HE, line 218: Why not pass seed? I guess CH backend also needs the seed. Then, we can remove genShuffleTransformer in API class.

cc @exmy, I note `shuffle` is mapped to `arrayshuffle` in CH. The doc says it supports seed. See https://clickhouse.com/docs/en/sql-reference/functions/array-functions#arrayshuffle

exmy: Look it's ok to pass seed for CH backend, I will check it later. @PHILO-HE

PR #5723, opened 2024-05-13
[CORE] Add a compilation-time check to forbid case-class inheritance 

zhztheplayer:
This uses Scala linter `wartremover` and its maven plugin. And I wrote a custom wart for this.

The jar including the custom wart:
https://central.sonatype.com/artifact/io.github.zhztheplayer.scalawarts/scalawarts

Source code:
https://github.com/zhztheplayer/scalawarts

Wartremover:
https://github.com/wartremover/wartremover

After this patch, once a class tends to inherit a case class, compiler will raise an error:

```
error: [wartremover:InheritFromCaseClass] case class is not inheritable: MyClass is trying to inherit List(class CaseClazz)
```

To suppress that error, add an annotation before the class definition:

```
@SuppressWarnings(Array("io.github.zhztheplayer.scalawarts.InheritFromCaseClass"))
class MyClass extends CaseClazz{...}
```

In the other hand, I'll try upstreaming the code of `InheritFromCaseClass` to wartremover main stream. After that we'll remove the dependency to `io.github.zhztheplayer.scalawarts`.

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/spark/shuffle/HashPartitioningWrapper.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/spark/shuffle/HashPartitioningWrapper.scala
@@ -21,6 +21,7 @@ import org.apache.spark.sql.catalyst.plans.physical.HashPartitioning
 
 // A wrapper for HashPartitioning to remain original hash expressions.
 // Only used by CH backend when shuffle hash expressions contains non-field expression.
+@SuppressWarnings(Array("io.github.zhztheplayer.scalawarts.InheritFromCaseClass"))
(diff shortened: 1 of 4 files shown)


zhztheplayer:
cc @PHILO-HE @JkSelf @zzcclp

PHILO-HE (approved):
Looks good!

Do we need to tell developer how to suppress the error in the error message? I'm afraid some developer really needs such inheritance for some class, but don't know how to suppress the error.

zhztheplayer:
> InheritFromCaseClass

I'll add some documents addressing this, In the next PR. Thank you for the suggestion.

PR #5729, opened 2024-05-13
[CORE] Add a compilation-time check to forbid case-class inheritance

zhztheplayer:
This reopens https://github.com/apache/incubator-gluten/pull/5723 with a fix for building with Intellij Idea.

Ref: https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -633,7 +633,7 @@
             <args>
               <arg>-Wconf:msg=While parsing annotations in:silent</arg>
               <arg>-Ywarn-unused:imports</arg>
-              <arg>-Xfatal-warnings</arg>
+              <arg>-Wconf:any:e</arg>
               <arg>-deprecation</arg>
               <arg>-feature</arg>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -615,13 +615,28 @@
...
+                <artifactId>scalawarts</artifactId>
+                <version>0.1.0</version>
+              </dependency>
+            </dependencies>
             <recompileMode>${scala.recompile.mode}</recompileMode>
             <args>
               <arg>-Wconf:msg=While parsing annotations in:silent</arg>
               <arg>-Ywarn-unused:imports</arg>
-              <arg>-Xfatal-warnings</arg>
+              <arg>-Wconf:any:e</arg>
```
PHILO-HE, line 636: Will this override the arg at line 634?

zhztheplayer: Oops, I missed :634. Let me do a fix.

PR #5737, opened 2024-05-14
[VL] Drop the test table after all tests in FallbackSuite

ivoson:

Follow up of https://github.com/apache/incubator-gluten/pull/5655 to drop new added test table after all.

## How was this patch tested?

N/A


ivoson:
cc @zhztheplayer can you please take a look? Thanks

zhztheplayer (approved):
Thanks for catching!

PR #5740, opened 2024-05-14
[VL] Enable GlutenParqutRowIndexSuite for Spark 3.4/3.5

gaoyangxiaozhu:

The PR use to enable GlutenParquetRowIndexSuite. 
The suite do below steps to run tests:
1. Run parquet write api `df.write` to generate specific parquet test data file with specific option as block size, page size, etc.
2. Check if the generated parquet data file meets the requirements as the block size is expected or not.
3. If the generated parquet files are ok for use, then run all `row index generation xx` and `invalid row index type xxx` uts.

However, gluten parquet native write currenrly only does not support customize the `block_size`, `page_size`, that means 
the generated parquet file would always has hard-code row group number / page size per file which causes the tests run fail in parquet file generation stage. 

Checking here for the default block_size, page_size use in velox parquet write.

https://github.com/facebookincubator/velox/blob/b29d93374d8e75a595c85550428eb008a371d456/velox/dwio/parquet/writer/Writer.h#L90C3-L102C74

https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/writer/Writer.cpp#L420C3-L420C52

The PR re-write the suite to disable the native write to always generate the test parquet file content with vinaila spark and some addition check to fix the uts.

(Fixes: \#5309)

## How was this patch tested?

manually test / local run


gaoyangxiaozhu:
@yma11 / @FelixYBW / @zhouyuan

zhouyuan:
CC @JkSelf

gaoyangxiaozhu:
add @zhli1142015

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetRowIndexSuite.scala
```diff
@@ -16,6 +16,337 @@
...
+        rcs.reverse.tail.foreach {
+          rc =>
+            assert(
+              rc > DEFAULT_MINIMUM_RECORD_COUNT_FOR_CHECK,
+              "expected row groups larger than minimal row count")
+        }
+    }
+  }
+
+  case class GlutenRowIndexTestConf(
```
JkSelf, line 104: @gaoyangxiaozhu Can we leave a TODO to remove this change after https://github.com/facebookincubator/velox/pull/8864 merged?

JkSelf (approved):
Thanks for the fix. LGTM.

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetRowIndexSuite.scala
zhli1142015: @gaoyangxiaozhu  please open an issue for above. Thanks.

gaoyangxiaozhu: @JkSelf even 8864 merged, for 3.4/3.5 we still can't remove this chagne right ? Since I only see it been set here https://github.com/apache/incubator-gluten/blob/37be4aa026c0aa9265a89fb7ae6b4210a1eaafed/cpp/velox/operators/writer/VeloxParquetDatasource.cc#L101

but looks `veloxParquetDatasource.cc` is not used for 3.4/3.5 parquet write ?

gaoyangxiaozhu: @zhli1142015 and @JkSelf creating a issue to track https://github.com/apache/incubator-gluten/issues/5751

gaoyangxiaozhu:
@zhli1142015 could you help merge ?

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetRowIndexSuite.scala
JkSelf: @gaoyangxiaozhu Yes, 3.4 will not reach the code of V[VeloxParquetDatasource.cc](https://github.com/apache/incubator-gluten/blob/37be4aa026c0aa9265a89fb7ae6b4210a1eaafed/cpp/velox/operators/writer/VeloxParquetDatasource.cc#L101). However, with the https://github.com/facebookincubator/velox/pull/8864, we can set the flush policy where the TableWriteNode is created https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlan.cc#L583. Currently, this does not support pageSize, but we can add it in Velox in a similar manner."

JkSelf: @gaoyangxiaozhu It appears that [VeloxParquetDatasource.cc](https://github.com/apache/incubator-gluten/blob/37be4aa026c0aa9265a89fb7ae6b4210a1eaafed/cpp/velox/operators/writer/VeloxParquetDatasource.cc#L101) does not set the pageSize either. How is it that this unit test can pass in both Spark 3.2 and Spark 3.3? Perhaps I miss something. Thanks.

gaoyangxiaozhu:
@JkSelf / @zhli1142015 could you help check if can merge this change ?

PR #5743, opened 2024-05-14
[VL] Move velox related configs to VeloxConfig.h

Yohahaha:
Move velox related configs to VeloxConfig.h, thus these configs can be used by include VeloxConfig.h

### DIFF
--- a/cpp/core/config/GlutenConfig.h
+++ b/cpp/core/config/GlutenConfig.h
@@ -61,13 +61,6 @@ const std::string kShuffleCompressionCodecBackend = "spark.gluten.sql.columnar.s
 const std::string kQatBackendName = "qat";
 const std::string kIaaBackendName = "iaa";
 
-// Velox conf
-const std::string kGlogVerboseLevel = "spark.gluten.sql.columnar.backend.velox.glogVerboseLevel";
-const uint32_t kGlogVerboseLevelDefault = 0;
-const uint32_t kGlogVerboseLevelMaximum = 99;
-const std::string kGlogSeverityLevel = "spark.gluten.sql.columnar.backend.velox.glogSeverityLevel";
-const uint32_t kGlogSeverityLevelDefault = 1;
-
 std::unordered_map<std::string, std::string>
 parseConfMap(JNIEnv* env, const uint8_t* planData, const int32_t planDataLength);
 
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -24,8 +24,6 @@
 #include "operators/plannodes/RowVectorStream.h"
 #include "utils/ConfigExtractor.h"
 
-#include "shuffle/VeloxShuffleReader.h"
-
 #ifdef GLUTEN_ENABLE_QAT
 #include "utils/qat/QatCodec.h"
 #endif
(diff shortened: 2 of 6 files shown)


FelixYBW:
Not sure if we can have a better way to use a config file then automatically generate the GlutenConfig.h and GlutenConfig.scala.

The initial design is to add all configs to .scala, put native config into .h/.cc

zhztheplayer:
I think there was some initial consideration about storing only general configurations in `cpp/core` as well as `GlutenConfig.h`.

I do believe we don't yet have a very clear design for relationship between these Scala/CPP modules. But until we start a complete refactor, I think we can just follow the rules we already set so far.

Regarding the change, I am inclined to not moving all confs to `GlutenConfig.h` (unless we have strong reason), but just moving [these Velox confs](https://github.com/apache/incubator-gluten/blob/37be4aa026c0aa9265a89fb7ae6b4210a1eaafed/cpp/core/config/GlutenConfig.h#L64-L69) from `cpp/core` to `cpp/velox`.

Yohahaha:
> I think there was some initial consideration about storing only general configurations in `cpp/core` as well as `GlutenConfig.h`.
> 
> I do believe we don't yet have a very clear design for relationship between these Scala/CPP modules. But until we start a complete refactor, I think we can just follow the rules we already set so far.
> 
> Regarding the change, I am inclined to not moving all confs to `GlutenConfig.h` (unless we have strong reason), but just moving [these Velox confs](https://github.com/apache/incubator-gluten/blob/37be4aa026c0aa9265a89fb7ae6b4210a1eaafed/cpp/core/config/GlutenConfig.h#L64-L69) from `cpp/core` to `cpp/velox`.

All configs are defined in GlutenConfig.scala, include each banckend, but I still need search much place to find related native configs. If I want use/test some behavior with these hidden config,  it's inconvenience to move it to open place and change in each PR. This is the reason for me.

And there is no reason to separate native configs into different place, since all configs has already been unified in java side. If some design is not solid now, why it can not be changed?

I don't treat this PR as a huge refactor, it's the first PR to add tests for memory manager in native side.

Yohahaha:
another way is put common configs into GlutenConfig.h and velox configs into velox/VeloxConfig.h, but I dont find the real difference between these.

zhztheplayer:
> And there is no reason to separate native configs into different place, since all configs has already been unified in java side. If some design is not solid now, why it can not be changed?

This is why I mentioned that we haven't yet do well on module separation. Before we have a clearer design or do a refactor against the base arch of Gluten, I'd suggest keeping Velox backend's code inside its own module, including those in `GlutenConfig.scala`. 

I had a rough thinking on an ideal shape of code architecture of Gluten, e.g.,


```
gluten
|- core
|  |- java
|  |  |- main
|  |  \- shim
|  |- scala
|  |  |- main
|  |  \- shim
|  \- cpp
|- velox-backend
|  |- java
|  |  |- main
|  |  \- shim
|  |- scala
|  |  |- main
|  |  \- shim
|  \- cpp
|
...
``` 

Which intentionally requires for code separation between kernel and backends. Whether or not we go like this, it's very possible that Gluten keeps attempting to be extensible for different backends. So I see keeping the code separated in their own modules better than union them into core module anyway.

> another way is put common configs into GlutenConfig.h and velox configs into velox/VeloxConfig.h, but I dont find the real difference between these.

I would second this. Would you like to use that approach, temporarily?

Yohahaha:
> I would second this. Would you like to use that approach, temporarily?

ok, I will do this.

FelixYBW:
> I do believe we don't yet have a very clear design for relationship between these Scala/CPP modules. But until we start a complete refactor, I think we can just follow the rules we already set so far.

Let's make sure Gluten_config.scala has all the configurations and short description.

Yohahaha:
@FelixYBW @zhztheplayer please help review, thank you!

zhztheplayer (approved):
Thanks!

cc @PHILO-HE

PHILO-HE (approved):
Looks good!

PR #674, opened 2022-12-06
Remove/Rename the overridden classes against vanilla Spark's code

zhztheplayer:
Temporarily remove some classes that override vanilla Spark's code since the class loading order is not always predictable in Java 8 & HotSpot JVM. We'll find another way to achieve the same if possible.

PR #408, opened 2022-09-28
[OPPRO-266] Use the smaller table to build hashmap in shuffled hash join

rui-mo:
## What changes were proposed in this pull request?

Removed Spark's limitation for the build side of left and right outer join in shuffled hash join, so we could use the smaller table to build the hashmap. That means, LeftOuter and LeftSemi join can build left, and RightOuter join can build right when the left or right table is smaller.

## How was this patch tested?

Verified on Jenkins.

PR #5750, opened 2024-05-15
[CORE] Use the smaller table to build hashmap in shuffled hash join

zml1206:
Background history: [SPARK-36612](https://issues.apache.org/jira/browse/SPARK-36612) support left outer join build left or right outer join build right in shuffled hash join, it first took effect in spark 3.5.
  #408 support use the smaller table to build hashmap in shuffled hash join in velox backend, and it overwrite `HashJoin`, so when fallback to vanilla spark all versions support, not just spark 3.5.  But since the class loading order is not always predictable in Java 8 & HotSpot JVM, Support has been temporarily removed in #674.
This PR finally select build side when `genShuffledHashJoinExecTransformer` to support use the smaller table to build hashmap in shuffled hash join in velox backend. 
`JoinSelectionOverrides` will no longer generate illegal build side, causing `EnsureRequirements` verification to fail.


zml1206:
Test failure is unrelated. cc @rui-mo @zhztheplayer

zhztheplayer:
> Background history: [SPARK-36612](https://issues.apache.org/jira/browse/SPARK-36612) support left outer join build left or right outer join build right in shuffled hash join, it first took effect in spark 3.5.

@zml1206 

Then how does this PR work with lower Spark version?

zml1206:
> > Background history: [SPARK-36612](https://issues.apache.org/jira/browse/SPARK-36612) support left outer join build left or right outer join build right in shuffled hash join, it first took effect in spark 3.5.
> 
> @zml1206
> 
> Then how does this PR work with lower Spark version?


If there is no fallback, all versions of velox bakcend support using small tables to build hashmaps. If there is a fallback, it will remain consistent with vanilla spark. It is supported after 3.5 and not supported before 3.5. @zhztheplayer

zhztheplayer:
> > > Background history: [SPARK-36612](https://issues.apache.org/jira/browse/SPARK-36612) support left outer join build left or right outer join build right in shuffled hash join, it first took effect in spark 3.5.
> > 
> > 
> > @zml1206
> > Then how does this PR work with lower Spark version?
> 
> If there is no fallback, all versions of velox bakcend support using small table to build hashmap. If there is a fallback, it will remain consistent with vanilla spark, it is supported after 3.5 and not supported before 3.5. @zhztheplayer

Does that mean we don't actually need strategy `JoinSelectionOverrides` to alter join build side on vanilla query plan ? Can we completely remove that strategy?

zhztheplayer (approved):
Thanks!

WangGuangxin:
Do we still need `preferredBuildSide` in `ShuffledHashJoinExecTransformer`?  https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/execution/ShuffledHashJoinExecTransformer.scala#L78

In fact, I'm quite confused this logic here. In fact, if we can choose the right build side, why prefer new build side based on pattern there?

rui-mo:
> if we can choose the right build side, why prefer new build side based on pattern there?

@WangGuangxin The reasoning behind this is that, in cases where the child is an aggregation or filter, we discovered that Spark's smaller table selection was inconsistent with reality. If we follow Spark's recommendation in this case, we will actually be using the larger table as the build table, which will result in performance regression.

zml1206:
> > > > Background history: [SPARK-36612](https://issues.apache.org/jira/browse/SPARK-36612) support left outer join build left or right outer join build right in shuffled hash join, it first took effect in spark 3.5.
> > > 
> > > 
> > > @zml1206
> > > Then how does this PR work with lower Spark version?
> > 
> > 
> > If there is no fallback, all versions of velox bakcend support using small table to build hashmap. If there is a fallback, it will remain consistent with vanilla spark, it is supported after 3.5 and not supported before 3.5. @zhztheplayer
> 
> Does that mean we don't actually need strategy `JoinSelectionOverrides` to alter join build side on vanilla query plan ? Can we completely remove that strategy?

First of all, the judgment of JoinSelectionOverrides and Spark Join Selection are not the same. I am not sure why JoinSelectionOverrides does not have the judgment of size. Secondly, clickhouse backend has special processing.

zhztheplayer:
>  I am not sure why JoinSelectionOverrides does not have the judgment of size.

@rui-mo Do you have some contexts on this difference? If Velox backend doesn't require for the strategy then we may be able to move it to CH module.

WangGuangxin:
> > I am not sure why JoinSelectionOverrides does not have the judgment of size.
> 
> @rui-mo Do you have some contexts on this difference? If Velox backend doesn't require for the strategy then we may be able to move it to CH module.

IIUC, `JoinSelectionOverride` does two things:
1. force shuffled hash join
2. fallback if there are continue joins ( TPC-DS badcase?)

Both of them can be ( and should be ?) moved to  `ColumnOverrides`.

In fact, if we reuse the `JoinSelection` rule in Spark instead of override it, it can naturally support choose build side by size, right? cc @zml1206 @rui-mo

rui-mo:
> it can naturally support choose build side by size, right?

@WangGuangxin A deeper issue might be the size estimation could be inaccurate if there is an aggregate or filter, e.g. in some case, the aggregated data has much less rows than its input and becomes the smaller table, but Spark still treats it as the larger table.

WangGuangxin:
> > it can naturally support choose build side by size, right?
> 
> @WangGuangxin A deeper issue might be the size estimation could be inaccurate if there is an aggregate or filter, e.g. in some case, the aggregated data has much less rows than its input and becomes the smaller table, but Spark still treats it as the larger table.

@rui-mo AQE will also go through `JoinSelection` to reoptimize the new submit stage. At this time, the `stats` is not estimated, but `ShuffleQueryStage`'s real datasize.
image

If there is no `Exchange` before `Join`, yes the stats is estimated and we cann't have an accurate stats.

zml1206:
I also think that the ShuffledHashJoinExecTransformer.`preferredBuildSide` is unnecessary when AQE is enabled.

jackylee-ch:
> Then how does this PR work with lower Spark version?

@zhztheplayer We have already use this pr in our environment last year with Spark 3.3, and they work well now.

> I also think that the ShuffledHashJoinExecTransformer.preferredBuildSide is unnecessary when AQE is enabled.

Agree, we have some bad case for this, such as we have a large table and all data are not same, then we still have a large data after agg.

zml1206:
I still have a doubt, why is the design forced to use ShuffledHashJoin instead of judging based on the threshold? If the left and right tables are both very large, will it not cause velox memory problems?

zhztheplayer:
> Both of them can be ( and should be ?) moved to ColumnOverrides.

If the custom strategy can be removed by moving the code to ColumnarOverrides (without more workarounds), I will be inclined to do that since it:

1. Simplifies code
2. Creates "more vanilla" plan when the join operators are falling back

> why is the design forced to use ShuffledHashJoin instead of judging based on the threshold?

It was mainly because hash join is faster / more well-maintained than sort join in Velox, if I remember it correctly.

> If the left and right tables are both very large, will it not cause velox memory problems?

This is very possible. Velox's  grace hash join has maximum spill level (be default 4) so OOM will happen for huge input. We may need to finally provide production-ready SMJ support in Gluten.

zml1206:
> If the custom strategy can be removed by moving the code to ColumnarOverrides (without more workarounds), Personally I will be inclined to do that since it:
> 
> 1. Simplifies code
> 2. Creates "more vanilla" plan when the join operators are falling back

Do it in this PR or next? @zhztheplayer

zhztheplayer:
> > If the custom strategy can be removed by moving the code to ColumnarOverrides (without more workarounds), Personally I will be inclined to do that since it:
> > 
> > 1. Simplifies code
> > 2. Creates "more vanilla" plan when the join operators are falling back
> 
> Do it in this PR or next? @zhztheplayer

Thank you for willing to take this. Of course it can be done in a separate PR.

And did you do benchmark on this change? I think the benchmark (called by `/Benchmark Velox`) is broken now.

zml1206:
> And did you do benchmark on this change? I think the benchmark tool (called by `/Benchmark Velox`) is broken now.

Sorry, I didn't do benchmark. There are some problems with the local environment of my mac m1.

zhztheplayer:
> > And did you do benchmark on this change? I think the benchmark tool (called by `/Benchmark Velox`) is broken now.
> 
> Sorry, I didn't do benchmark. There are some problems with the local environment of my mac m1.

I see. Then maybe we should wait a little bit before merging this to see if the tool could recover.

rui-mo:
Just triggered benchmark tests manually. Let us wait for the performance result.

zhztheplayer:
@zml1206 Would you like to rebase the code? I believe benchmark tool is online now. Thanks!

zml1206:
> @zml1206 Would you like to rebase the code? I believe benchmark tool is online now. Thanks!

Of course, done.

zhztheplayer (approved):

rui-mo:
We notice large regression for q14a/q14b on TPC-DS benchmark. We may need to double-check. Thanks.
```
query	log/native_5750_time.csv	log/native_master_05_26_2024_time.csv	difference
q1	8.54	9.57	1.033
q2	8.81	8.39	-0.424
q3	2.76	2.91	0.149
q4	47.48	47.51	0.037
q5	5.23	6.27	1.046
q6	1.51	2.55	1.035
q7	4.40	4.19	-0.215
q8	2.37	2.59	0.220
q9	15.51	15.36	-0.150
q10	6.73	7.07	0.345
q11	24.12	23.95	-0.169
q12	1.86	0.98	-0.878
q13	4.04	4.01	-0.028
q14a	385.94	33.16	-352.776
q14b	375.88	30.84	-345.040
```

rui-mo (changes requested):
We found performance regression for q14a/q14b. Would you take a look?

zml1206:
> We found performance regression for q14a/q14b. Would you take a look?

I will take a look later.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -376,13 +376,8 @@ object VeloxBackendSettings extends BackendSettingsApi {
...
           // OPPRO-266: For Velox backend, build right and left are both supported for
-          // LeftOuter and LeftSemi.
-          // FIXME Hongze 22/12/06
-          //  HashJoin.scala in shim was not always loaded by class loader.
-          //  The file should be removed and we temporarily disable the improvement
-          //  introduced by OPPRO-266 by commenting out the following prerequisite
-          //  condition.
-//          case LeftOuter | LeftSemi => true
+          // LeftOuter.
+          case LeftOuter => true
```
rui-mo, line 380: Build left is supported by LeftSemi on Velox backend. Shall we add it?

zml1206: The previous regression of b14a/b14b was due to left semi join build left side. Although build left is supported by LeftSemi on Velox backend, but the performance is not as good as build right, vanilla spark also choose build right.

rui-mo: This function returns if building left is supported, which I assume indicates the backend's capability and shouldn't be mixed up the build side selection.

> vanilla spark also choose build right.

It is due to some special logic, e.g. preferredBuildSide, Gluten chooses to build left? If so, I think we need to fix there.

zml1206: I was planning to remove `preferredBuildSide` in the next PR because it is unnecessary and not as accurate as AQE. Like vanilla spark, its performance improvement is compared to sort merge join, for left outer and join outer, if both are shuffle hash joins, choosing a small table will not improve performance.
I don’t know the logic of velox’s support for small table build.  Shuffle hash join is forced to be used here, 
what I'm worried about is that for shuffle hash join of LeftSemi, build left is much slower than build right.

zml1206: https://github.com/apache/spark/blob/branch-3.2/sql/core/src/test/resources/tpcds-plan-stability/approved-plans-v1_4/q14a/explain.txt
Both left and right contain filter, so `isPreferred(left)` and `isPreferred(right)` both are true, so I think it is not caused `preferredBuildSide`.

rui-mo: Not exactly. The two sides of the time-consuming join have a large difference on data size, so we need to make sure the smaller table is used as build side. I'm still confirming that.

image

rui-mo: Could you add a TODO here to describe why we don't add LeftSemi? Maybe also include the issue link.
https://github.com/facebookincubator/velox/issues/9980

zml1206: Added.

rui-mo (approved):
Thanks.

zml1206:
Thanks all for review.

zml1206:
> > > If the custom strategy can be removed by moving the code to ColumnarOverrides (without more workarounds), Personally I will be inclined to do that since it:
> > > 
> > > 1. Simplifies code
> > > 2. Creates "more vanilla" plan when the join operators are falling back
> > 
> > 
> > Do it in this PR or next? @zhztheplayer
> 
> Thank you for willing to take this. Of course it can be done in a separate PR.

I looked it up and it seems like I can't move custom strategy to ColumnarOverrides. Custom strategy is used to transform a LogicalPlan into a SparkPlan. Columnar rule implements Spark plan into Gluten plan. Custom strategy runs before Columnar rule. @zhztheplayer

zhztheplayer:
> > > > If the custom strategy can be removed by moving the code to ColumnarOverrides (without more workarounds), Personally I will be inclined to do that since it:
> > > > 
> > > > 1. Simplifies code
> > > > 2. Creates "more vanilla" plan when the join operators are falling back
> > > 
> > > 
> > > Do it in this PR or next? @zhztheplayer
> > 
> > 
> > Thank you for willing to take this. Of course it can be done in a separate PR.
> 
> I looked it up and it seems like can't move custom strategy to ColumnarOverrides. Custom strategy is used to transform a LogicalPlan into a SparkPlan. Columnar rule implements Spark plan into Gluten plan. Custom strategy runs before Columnar rule. @zhztheplayer

I think SHJ and SMJ in Spark share the same `requiredChildDistribution` code. Which means we can just convert one to another in columnar rule then adjust its children to add/remove sorts on top of them. (Ideally, you could double check the relevant code)

In the simplest case the topic could be very similar to the one about hash/sort agg, see [this code](https://github.com/apache/incubator-gluten/blob/00dda1b67ec81dc841643e1c02323f32df5525d9/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala#L289-L297)

zml1206:
> > > > > If the custom strategy can be removed by moving the code to ColumnarOverrides (without more workarounds), Personally I will be inclined to do that since it:
> > > > > 
> > > > > 1. Simplifies code
> > > > > 2. Creates "more vanilla" plan when the join operators are falling back
> > > > 
> > > > 
> > > > Do it in this PR or next? @zhztheplayer
> > > 
> > > 
> > > Thank you for willing to take this. Of course it can be done in a separate PR.
> > 
> > 
> > I looked it up and it seems like can't move custom strategy to ColumnarOverrides. Custom strategy is used to transform a LogicalPlan into a SparkPlan. Columnar rule implements Spark plan into Gluten plan. Custom strategy runs before Columnar rule. @zhztheplayer
> 
> I think SHJ and SMJ in Spark share the same `requiredChildDistribution` code. Which means we can just convert one to another in columnar rule then adjust its children to add/remove sorts on top of them. (Ideally, you could double check the relevant code)
> 
> In the simplest case the topic could be very similar to the one about hash/sort agg, see [this code](https://github.com/apache/incubator-gluten/blob/00dda1b67ec81dc841643e1c02323f32df5525d9/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala#L289-L297)

This is a good way for me, thank you, I'll do it later.

PR #5752, opened 2024-05-15
[CORE] Add decimal precision tests

ulysses-you:

- Port `DecimalPrecisionSuite` into gluten.
- Add dataType in `ExpressionTransformer` to help check data type

## How was this patch tested?

Pass CI


gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -450,14 +450,6 @@ trait SparkPlanExecApi {
     GenericExpressionTransformer(substraitExprName, children, original)
   }
 
-  def genEqualNullSafeTransformer(
-      substraitExprName: String,
-      left: ExpressionTransformer,
-      right: ExpressionTransformer,
-      original: EqualNullSafe): ExpressionTransformer = {
-    GenericExpressionTransformer(substraitExprName, Seq(left, right), original)
```
ulysses-you, line 458: no backends override this, remove it and use GenericExpressionTransformer

ulysses-you:
cc @jinchengchenghh @baibaichen @PHILO-HE  thank you

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala
```diff
@@ -18,6 +18,15 @@ package org.apache.gluten.expression
 
 import org.apache.gluten.substrait.expression.ExpressionNode
 
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.types.DataType
+
 trait ExpressionTransformer {
```
jinchengchenghh, line 24: Does this one still should be used? Can we use `ExpressionTransformerWithOrigin` to replace `ExpressionTransformer` and rename `ExpressionTransformerWithOrigin` to `ExpressionTransformer`?

ulysses-you: Some expression transformer did not catch orginal expression. e.g., `BoundReferenceTransformer`, `AttributeReferenceTransformer`, etc. So I leave it in this pr.

jinchengchenghh: We can change it to catch original, do you have other example?
```
val b = bindReference.asInstanceOf[BoundReference]
          AttributeReferenceTransformer(
            a.name,
            b.ordinal,
            a.dataType,
            b.nullable,
            a.exprId,
            a.qualifier,
            a.metadata)

      case b: BoundReference =>
        BoundReferenceTransformer(b.ordinal, b.dataType, b.nullable)
```

jinchengchenghh: For the code clean, I prefer to hold origin for all the `ExpressionTransformer`, what do you think? @zhztheplayer

yaooqinn (approved):

PR #5761, opened 2024-05-16
[VL] Enable length function for binary type

zhli1142015:

Enable length function for binary type.

## How was this patch tested?
UT.


FelixYBW (approved):

rui-mo (approved):
Thanks.

PR #5763, opened 2024-05-16
[CORE] Avoid copy in ByteLiteralNode

jinchengchenghh:
The serialized bloom filter is binary node, it may be very large, then affect the performance.

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/substrait/expression/BinaryLiteralNode.java
+++ b/gluten-core/src/main/java/org/apache/gluten/substrait/expression/BinaryLiteralNode.java
@@ -22,6 +22,9 @@ import org.apache.gluten.substrait.type.TypeNode;
 import com.google.protobuf.ByteString;
 import io.substrait.proto.Expression.Literal.Builder;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
 public class BinaryLiteralNode extends LiteralNodeWithValue<byte[]> {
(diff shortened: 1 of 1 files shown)


jinchengchenghh:
Can not use ByteBuffer to serialize because of exception
```
java.io.NotSerializableException: java.nio.DirectByteBuffer
```

gluten-core/src/main/java/org/apache/gluten/substrait/expression/BinaryLiteralNode.java
```diff
@@ -33,6 +36,14 @@ public BinaryLiteralNode(byte[] value, TypeNode typeNode) {
...
-    literalBuilder.setBinary(ByteString.copyFrom(value));
+    ByteString byteValue;
+    try {
+      Method m = ByteString.class.getDeclaredMethod("wrap", byte[].class);
+      m.setAccessible(true);
+      byteValue = (ByteString) m.invoke(null, value);
+    } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException e) {
+      throw new RuntimeException(e);
+    }
+    literalBuilder.setBinary(byteValue);
```
zhouyuan, line 47: shall we fallback to use the "copy" code path here?

jinchengchenghh: These exceptions only occur when the code is wrong, so we should not hide the exception

zhouyuan (approved):

Issue #5764, opened 2024-05-16
[CORE] Rework planner C2R / R2C code with new transition facilities


zhztheplayer:
`Transition` will stand for interchanges between different in-memory data formats.

The enhancement includes:

1. Union current C2R / R2C / C2R2C / (future) C2C code together with centralized management for `transition`s.
   - Remove old rules for inserting C2R / R2C / C2R2C transitions.
   - Remove C2R / R2C creation APIs from SparkPlanExecApi.
2. Extensible `convention` API:
   - `Convention` will be bound to each individual plan operator to describe its supported output in-memory format.
   - `Convention` will consist of `RowType` and `BatchType`
   - `BatchType` is extensible with predefined API
   - When a new `BatchType` is defined, the relevant `to-row` / `from-row` / `to-other-batch` / `from-other-batch` transitions can be defined at the same time.
3. Insert transitions:
   - Remove old rules for inserting C2R / R2C / C2R2C transitions.
   - Remove code that will remain unused from `PlanUtil`.
   - Add a new rule to add transitions using the new facility. In the new rule, Gluten will find the cheapest defined transition for two adjacent plan operators.

zhztheplayer:
Fixed in https://github.com/apache/incubator-gluten/pull/5767

Issue #1440, opened 2023-04-21
Error while reading iceberg:  `batch is not Arrow columnar batch`


jinhaichen:
the error info is `batch is not Arrow columnar batch`.

```
23/04/20 13:32:53 INFO BlockManagerMaster: BlockManagerMaster stopped
23/04/20 13:32:53 INFO OutputCommitCoordinator$OutputCommitCoordinatorEndpoint: OutputCommitCoordinator stopped!
23/04/20 13:32:53 INFO SparkContext: Successfully stopped SparkContext
Exception in thread "main" org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 0.0 failed 4 times, most recent failure: Lost task 0.3 in stage 0.0 (TID 3) (1.2.200.100 executor 3): java.lang.RuntimeException: Error during calling Java code from native code: java.lang.IllegalArgumentException: batch is not Arrow columnar batch
	at io.glutenproject.columnarbatch.ArrowColumnarBatches.offload(ArrowColumnarBatches.java:121)
	at io.glutenproject.columnarbatch.ArrowColumnarBatches.ensureOffloaded(ArrowColumnarBatches.java:172)
	at io.glutenproject.vectorized.ArrowInIterator.next(ArrowInIterator.java:36)
	at io.glutenproject.vectorized.ArrowOutIterator.nativeHasNext(Native Method)
	at io.glutenproject.vectorized.ArrowOutIterator.hasNextInternal(ArrowOutIterator.java:48)
	at io.glutenproject.vectorized.GeneralOutIterator.hasNext(GeneralOutIterator.java:37)
	at io.glutenproject.backendsapi.glutendata.GlutenIteratorApi$$anon$3.hasNext(GlutenIteratorApi.scala:300)
	at io.glutenproject.vectorized.CloseableColumnBatchIterator.hasNext(CloseableColumnBatchIterator.scala:41)
```

Please have a check.

zhouyuan:
hi @jinhaichen 

I assume the test is using Velox backend(?). iceberg/hudi/deltalake is not supported yet. Velox backend do have some rough plans to support those lakehouse based solutions in 2H23. The main gap is Velox does not have "connectors" for those lakehouse. 

Thanks, -yuan

jinhaichen:
> hi @jinhaichen
> 
[...]

PR #5767, opened 2024-05-16
[CORE] Rework C2R / R2C code with new transition facilities

zhztheplayer:
This is the 1st part fix for enhancement https://github.com/apache/incubator-gluten/issues/5764.

New trait `Convention` uses same name with the one used in RAS. They share similar functionalities. Will raise another PR to cleanup RAS code to rename and redirect to the new API.


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -73,6 +75,22 @@ import scala.collection.mutable.ListBuffer
...
+
+  /**
+   * Overrides [[org.apache.gluten.extension.columnar.transition.ConventionFunc]] Gluten is using to
+   * determine the convention (its row-based processing / columnar-batch processing support) of a
+   * plan with a user-defined function that accepts a plan then returns batch type it outputs.
+   */
+  override def batchTypeFunc(): BatchOverride = {
+    case i: InMemoryTableScanExec
+        if i.relation.cacheBuilder.serializer.isInstanceOf[ColumnarCachedBatchSerializer] =>
+      VeloxBatch
```
zhztheplayer, line 91: cc @ulysses-you

backends-velox/src/main/scala/org/apache/gluten/execution/datasource/v2/ArrowBatchScanExec.scala
```diff
@@ -31,6 +33,10 @@ case class ArrowBatchScanExec(original: BatchScanExec)
 
   @transient lazy val batch: Batch = original.batch
 
+  override protected def batchType0(): Convention.BatchType = {
+    ArrowBatch
```
zhztheplayer, line 37: cc @jinchengchenghh

jinchengchenghh:
Can we remove https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala#L59?

zhztheplayer:
> Can we remove https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/utils/PlanUtil.scala#L59?

It was already removed in this PR

backends-velox/src/main/scala/org/apache/gluten/execution/datasource/v2/ArrowBatchScanExec.scala
```diff
@@ -31,6 +33,10 @@ case class ArrowBatchScanExec(original: BatchScanExec)
 
   @transient lazy val batch: Batch = original.batch
 
+  override protected def batchType0(): Convention.BatchType = {
```
rui-mo, line 36: nit: the name batchType0 is a bit confusing, and can we use a more informative name?

rui-mo (approved):

Yohahaha:
> With out 3, once batch-to-batch transition is needed, batch-to-row-to-batch transition will be promoted instead.

why batch-to-batch be replaced with batch-to-row-to-batch?

Yohahaha:
and I want to say that it's to hard to read these codes, so many new concept, so many new abstraction, so many "design".

I'm not sure who besides Hongze can maintain and fully understand this part of the code. cc @FelixYBW

zhztheplayer:
> and I want to say that it's to hard to read these codes

Please start from [this test](https://github.com/zhztheplayer/gluten/blob/0cb05533f1d32abe289b301093ad8f9f82b6b5c8/gluten-core/src/test/scala/org/apache/gluten/extension/columnar/transition/TransitionSuite.scala).

> so many new concept, so many new abstraction, so many "design".

Query planner / optimizer is always hard. Spark has design issue when dealing with different batch types and Gluten's transition system perfectly solves the problem. That's why we no longer see issues like https://github.com/apache/incubator-gluten/issues/1440 that could easily crash our applications.

> so many "design"

You can call for documentation about the feature and I am glad to do, but I won't suggest using this kind of double-quoting for technical discussion in community.

FelixYBW:
@Yohahaha Did you see any issue with this PR? We can setup a meeting to go through the details if needed.

@zhztheplayer was working on couple of PRs to refactor the JNI call. The end goal is to to have a set of APIs to easily add 3rd libraries and accelerators as native engine.   
  
We do need to add more document and blogs in future.

PR #5768, opened 2024-05-16
[VL][CI] Cache native libraries to re-use them in Spark test jobs

### LINKED ISSUE #5721: [VL] Reuse native build artifact for different spark version to save resources and time
### Description

native build artifact (velox and gluten cpp) is common for different spark version, we can reuse it and trigger java module when success.

cc @zhouyuan @PHILO-HE

PHILO-HE:
Fix #5721.


.github/workflows/velox_docker.yml
```diff
@@ -494,13 +494,74 @@ jobs:
           GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
             --local --preset=velox-with-celeborn --benchmark-type=ds --error-on-memleak --off-heap-size=10g -s=1.0 --threads=8 --iterations=1
 
+  build-native-lib-centos-8:
```
zhztheplayer, line 497: Why centos-8 is needed?

PHILO-HE: @zhztheplayer, it's just for distinguishing it with the other static build job for centos-7. We originally planned to just use one cache, produced by static build, but there still exist some issues. So we can firstly keep the two build jobs to produce libraries cache for tpc test jobs (on centos-7) & spark test jobs (on centos-8) respectively.

PR #5770, opened 2024-05-16
[VL] Move memory reservation block computation logic into ListenableAllocator

Yohahaha:

Changes:
1. set memoryPoolTransferCapacity_ with spark.gluten.memory.reservationBlockSize which used when grow capacity.
2. add BlockAllocationListener for block reservation logic, it's safe to keep AllocationListener be non-thread safe since it will be invoked by whether ListenableAllocator or ListenableArbitrator at same time.

## How was this patch tested?

Add MemoryManagerTest.cc

### DIFF
--- a/cpp/core/config/GlutenConfig.h
+++ b/cpp/core/config/GlutenConfig.h
@@ -42,6 +42,9 @@ const std::string kSparkOffHeapMemory = "spark.gluten.memory.offHeap.size.in.byt
 
 const std::string kSparkTaskOffHeapMemory = "spark.gluten.memory.task.offHeap.size.in.bytes";
 
+const std::string kMemoryReservationBlockSize = "spark.gluten.memory.reservationBlockSize";
+const uint64_t kMemoryReservationBlockSizeDefault = 8 << 20;
+
 const std::string kSparkBatchSize = "spark.gluten.sql.columnar.maxBatchSize";
 
 const std::string kParquetBlockSize = "parquet.block.size";
--- a/cpp/core/jni/JniCommon.h
+++ b/cpp/core/jni/JniCommon.h
@@ -322,13 +322,8 @@ static inline gluten::CompressionMode getCompressionMode(JNIEnv* env, jstring co
 
 class SparkAllocationListener final : public gluten::AllocationListener {
  public:
-  SparkAllocationListener(
-      JavaVM* vm,
-      jobject jListenerLocalRef,
-      jmethodID jReserveMethod,
-      jmethodID jUnreserveMethod,
-      int64_t blockSize)
-      : vm_(vm), jReserveMethod_(jReserveMethod), jUnreserveMethod_(jUnreserveMethod), blockSize_(blockSize) {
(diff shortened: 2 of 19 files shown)


zhztheplayer:
@Yohahaha Seems to be that the block reservation code can be very suitable to become a individual `AllocationListener`. Would you like to check on that?

Yohahaha:
> @Yohahaha Seems to be that the block reservation code can be very suitable to become a individual `AllocationListener`. Would you like to check on that?

I see it, listener can return the real changed bytes after move block reservation logic into it.

zhztheplayer:
> > @Yohahaha Seems to be that the block reservation code can be very suitable to become a individual `AllocationListener`. Would you like to check on that?
> 
> I see it, listener can return the real changed bytes after move block reservation logic into it.

I think we can wrap the underlying listener with a block-enabled listener only for creating ListenableMemoryAllocator. And use the underlying listener directly for ListenableArbitrator.

The block-reservation code at least includes some kind of algorithms so let's have it isolated from other code.

Yohahaha:
After some tries I would like to say ReservationListener is a hook to another memory management system and should be as simple as possible, MemoryAllocator is a common class at top level which may be implemented with different allocate/reserved details, each supported native backend has its own memory management way and can compose both MemoryAllocator and ReservationListener.

For this PR:
1. SparkAllocationListener aim to hook in Spark memory management system and do memory report work.
2. ListenableMemoryAllocator provide a specific allocator implementation at cpp module level which its memory changes can be tracked by listener.
3. VeloxMemoryManager track arrow memory usage by ListenableMemoryAllocator and velox part's memory usage by velox::memory::MemoryManager, memoryPoolTransferCapacity is velox's details and could be used in ListenableArbitrator.

> Seems to be that the block reservation code can be very suitable to become a individual AllocationListener. Would you like to check on that?

So I think the responsibility and boundaries of above classes are clear.

Yohahaha:
> ListenableMemoryAllocator

make sense for me.

Yohahaha:
> > > @Yohahaha Seems to be that the block reservation code can be very suitable to become a individual `AllocationListener`. Would you like to check on that?
> > 
> > 
> > I see it, listener can return the real changed bytes after move block reservation logic into it.
> 
> I think we can wrap the underlying listener with a block-enabled listener only for creating ListenableMemoryAllocator. And use the underlying listener directly for ListenableArbitrator.
> 
> The block-reservation code at least includes some kind of algorithms so let's have it isolated from other code.

make sense for me.

zhztheplayer (approved):
Great. Thank you for the update.

And let's wait for the benchmark result as well.

zhztheplayer:
Hi @Yohahaha the benchmark is currently unstable because of network issues. Let's wait until we get a valid result after the issues get fixed.

Yohahaha:
> Hi @Yohahaha the benchmark is currently unstable because of network issues. Let's wait until we get a valid result after the issues get fixed.

ok

zhztheplayer (approved):

Issue #5771, opened 2024-05-16
Add metric for ColumnarArrowEvalPythonExec


yma11:
### Description

We will add following metrics for ColumnarArrowEvalPythonExec:
`numInputRows`, `numOutputRows`, `numOutputBatches`, `processTime`

Issue #5461, opened 2024-04-19
ColumnarArrowPythonEvalExec support for Velox backend


yma11:
### Description

Currently for `PythonEvalExec`, Gluten will try to offload if backend has same implementation as udf in its built-in functions and fallback if native validation failed. But for Velox, it will mostly fallback since there are no such functions implemented. Besides for Velox backend, the data between each op is `VeloxColumnarBatch` and can export to `ArrowRecordBatch` through velox's bridge, ideally w/o memory copy. So we can take benefit of this and adopt a `ColumnarArrowPythonEvalExec` which can consume the `ArrowRecordBatch` directly, to avoid the overhead of `C2R` and `R2C`.
<img width="485" alt="image" src="https://github.com/apache/incubator-gluten/assets/11849056/3c2afcbd-a4bc-4e32-99fd-a2822bc5feb0">

As above diagram shows, we mainly have 2 things to do:

1. Additional projections: Pre-projections happen in this node when it need do some calculation against the input cols before python UDF. In this case, vanilla spark has a dedicated `Projector` to do this work. In current implementation, we add a pre project node to handle this. The output of the project node will keep the original cols as well as projected cols. In `ColumnarArrowPythonEvalExec`, we put the original cols in cache and then join them with the outputs of `ColumnarArrowPythonRunner` as the final output.
2. Transition between `VeloxColumnarBatch` and `ArrowRecordBatch`: In this node, we leveraging existing API `ColumnarBatches.ensureLoaded` to transfer velox batch into arrow format and then send it to `ColumnarArrowPythonRunner` as the python worker's input. The python worker will evaluate the input based on `PythonUDF` and return a `ArrowRecordBatch`. Then we use `ColumnarBatches.ensureOffloaded` to convert this batch to Velox batch format for next node.

### Performance

We tested performance of arrow udf and collected some performance:
```
from pyspark.sql.functions import pandas_udf, PandasUDFType
import pyspark.sql.functions as F
import os
@pandas_udf('long')
def pandas_plus_one(v):
    return (v + 1)

@pandas_udf('string')
def pd_get_first(v):
    return v.str.split(':').str[1]
# test int
df = spark.read.orc("file:///xx/yy").select("user_id").withColumn("processed_user_id", pandas_plus_one("user_id")).select("processed_user_id")
# test string
df = spark.read.orc("file:///xx/yy").select("url").withColumn("processed_url", pd_get_first("url")).select("processed_url")
```

[...]

PR #5772, opened 2024-05-16
[GLUTEN-5771][VL] Add metrics for ColumnarArrowEvalPythonExec

yma11:

Add metric for ColumnarArrowEvalPythonExec

(Fixes: \#5771)

## Spark UI

image

## How was this patch tested?

We tested performance of arrow udf and collected some performance:
```
from pyspark.sql.functions import pandas_udf, PandasUDFType
import pyspark.sql.functions as F
import os
@pandas_udf('long')
def pandas_plus_one(v):
    return (v + 1)

@pandas_udf('string')
def pd_get_first(v):
    return v.str.split(':').str[1]
# test int
df = spark.read.orc("file:///xx/yy").select("user_id").withColumn("processed_user_id", pandas_plus_one("user_id")).select("processed_user_id")
# test string
df = spark.read.orc("file:///xx/yy").select("url").withColumn("processed_url", pd_get_first("url")).select("processed_url")
```

The perf shows ~20% perf gain compared with vanilla spark.

image


FelixYBW:
@yma11 can you add a UI chart for the pyarrow UDF? Also add some implementation details?

In theory we can convert Velox to Arrow in Velox pipeline, then pass the arrow pointer to Spark where it's send to python process. There is no C2R and R2C in the whole process and no memcpy between Velox and Spark. Can we achieve this?

yma11:
> @yma11 can you add a UI chart for the pyarrow UDF? Also add some implementation details?
> 
> In theory we can convert Velox to Arrow in Velox pipeline, then pass the arrow pointer to Spark where it's send to python process. There is no C2R and R2C in the whole process and no memcpy between Velox and Spark. Can we achieve this?

Yes. There is no C2R and R2C in current implementation. There is a VeloxColumnar to Arrow only. But for memcpy, it depends on the arrow bridge. I found there are still some memory allocation at velox for data types like string. Let me add the implementation under the feature track.

yma11:
> > @yma11 can you add a UI chart for the pyarrow UDF? Also add some implementation details?
> > In theory we can convert Velox to Arrow in Velox pipeline, then pass the arrow pointer to Spark where it's send to python process. There is no C2R and R2C in the whole process and no memcpy between Velox and Spark. Can we achieve this?
> 
> Yes. There is no C2R and R2C in current implementation. There is a VeloxColumnar to Arrow only. But for memcpy, it depends on the arrow bridge. I found there are still some memory allocation at velox for data types like string. Let me add the implementation under the feature track.

@FelixYBW The implementation details are now added in [5461](https://github.com/apache/incubator-gluten/issues/5461). Perf data is also wrapped there. FYI.

zhztheplayer (commented):
I just noticed that this file's package is `package org.apache.spark.api.python` which is wrong. Would you like to fix it? @yma11

backends-velox/src/main/scala/org/apache/gluten/execution/python/ColumnarArrowEvalPythonExec.scala
```diff
@@ -341,11 +355,20 @@ case class ColumnarArrowEvalPythonExec(
               val joinedVectors = (0 until inputCb.numCols).toArray.map(
                 i => inputCb.column(i)) ++ (0 until outputCb.numCols).toArray.map(
                 i => outputCb.column(i))
+              // Columns in outputCb has random 0 or 1 refCnt and will fail checks in ensureOffload,
+              // so we do a hard reset here.
```
zhztheplayer, line 359: Did you figure out why "random ref count" happens?

gluten-data/src/main/java/org/apache/gluten/vectorized/ArrowWritableColumnVector.java
```diff
@@ -345,6 +345,10 @@ public void retain() {
     refCnt.getAndIncrement();
   }
 
+  public void resetRefCnt(long value) {
+    refCnt.set(value);
+  }
```
zhztheplayer, line 350: There was already a `#forceClose` API in `ColumnarBatches`, I assume that one could cover your use case.

Although let's figure out why we need to forcibly close the vectors first.

backends-velox/src/main/scala/org/apache/gluten/execution/python/ColumnarArrowEvalPythonExec.scala
yma11: Not quite. The outputCb is directly retrieved from python worker so can't debug why it happens.

yma11:
> I just noticed that this file (ColumnarArrowEvalPythonExec.scala)'s package is `package org.apache.spark.api.python` which is wrong. Would you like to fix it? @yma11

Fixed.

docs/developers/VeloxUDF.md
```diff
@@ -179,3 +179,48 @@ The output from spark-shell will be like
 +------------------+----------------+
 ```
 
+# Pandas UDFs (a.k.a. Vectorized UDFs)
```
yma11, line 182: @zhouyuan Can you take a look at this contents for pandas udf?

yma11:
@zhztheplayer Please help take a look again. Thanks.

gluten-data/src/main/java/org/apache/gluten/vectorized/ArrowWritableColumnVector.java
```diff
@@ -345,6 +345,10 @@ public void retain() {
     refCnt.getAndIncrement();
   }
 
+  public AtomicLong getRefCntObj() {
+    return refCnt;
+  }
```
zhztheplayer, line 350: It still looks risky... Let me add a fix on your PR

zhztheplayer (approved):

Issue #5773, opened 2024-05-16
[VL] update aws-sdk-cpp to version 1.11.285


yma11:
### Description

We need to update aws-sdk-cpp version for S3 RetryStrategy support.

PR #5774, opened 2024-05-16
[GLUTEN-5773] Update aws-sdk-cpp version to 1.11.285 (from 1.11.169)

yma11:

update aws-sdk-cpp to version 1.11.285

(Fixes: \#5773)

## How was this patch tested?
N/A


yma11:
@PHILO-HE  please help take a review. Thanks.

FelixYBW:
So we only updated the the aws-sdk-cpp in vcpkg.

@PHILO-HE Is there any risk if we only update individual components in vcpkg but not vcpkg itself in long term?

PHILO-HE:
> So we only updated the the aws-sdk-cpp in vcpkg.
> 
> @PHILO-HE Is there any risk if we only update individual components in vcpkg but not vcpkg itself in long term?

@FelixYBW, I think there is no risk. The only inconvenience is, we have to copy the ports files for some lib from vcpkg and maintain them in Gluten. I can take a try to upgrade vcpkg.

PHILO-HE:
@yma11, I note this lib is upgraded to 1.11.321 in upstream velox. Do we need to align with it in Gluten?
https://github.com/facebookincubator/velox/pull/9756

PHILO-HE (dismissed):
Looks good!

yma11:
> Looks good!

Thanks but wait for a while. May need some extra change. Thanks.

zhztheplayer (approved):

dev/vcpkg/ports/aws-sdk-cpp/vcpkg.json
```diff
@@ -1,8 +1,8 @@
 {
   "$note": "Automatically generated by generateFeatures.ps1",
   "name": "aws-sdk-cpp",
-  "version": "1.11.169",
-  "port-version": 2,
+  "version": "1.11.285",
```
marin-ma, line 4: Do you know the reason why the version was upgraded to 1.11.285 here but in Velox it was 1.11.321? https://github.com/facebookincubator/velox/pull/9756

marin-ma: Seems like they don't have 1.11.321 in ports. https://vcpkg.link/ports/aws-sdk-cpp/versions The nearest version is 1.11.321

PHILO-HE: @marin-ma, yes, vcpkg skips the support for 1.11.321. If there is no reason that Velox must use 1.11.321, maybe, we can upgrade it to a newer version, e.g., 1.11.474, for both Velox and Gluten vcpkg.

PHILO-HE: @marin-ma, it was just upgraded to the latest release (it's 1.11.321 at that time point) in Velox. But I guess we were not aware that vcpkg doesn't support this version, so use a lower version (1.11.285) in Gluten instead. So it's ok to use 1.11.285 or some other higher version at both sides. Maybe, it's better to change Gluten & velox to use a same higher version.

zhouyuan: https://github.com/apache/incubator-gluten/pull/9169

Issue #5757, opened 2024-05-16
Unnecessary ProjectExec generated for Generate function


xumingming:
### Description

For the following test case:

```scala
  test("test explode1") {
    withTempView("t1") {
      sql(
        """select * from values (array(1)), (array(2)), (array(3)), (array(4))
          |as tbl(a)
     """.stripMargin).createOrReplaceTempView("t1")
      runQueryAndCompare(
        s"""
           |SELECT explode(a) from t1;
           |""".stripMargin) {
        checkGlutenOperatorMatch[GenerateExecTransformer]
      }
    }
  }
```

The input of `explode` is already an Attribute, while current implementation would still plug in a new ProjectExec, which seems not necessary?

```
Generate explode(a#122), false, [col#128]
+- Project [a#122, a#122]
   +- RewrittenNodeWall Filter (size(a#122, true) > 0)
```

PR #5782, opened 2024-05-17
[GLUTEN-5757][CORE] Remove unnecessary ProjectExecTransformer for Generate

xumingming:

If generator function's input is already Attribute reference, we omit the
introduction of the ProjectExec.

Previous implementation always assume there is Project under Generate. In
the new implementation we added a metadata(injectedProject) in Substrait
plan to tell us whether there is a dedicated Project under Generate.

(Fixes: \#5757)

### DIFF
--- a/gluten-core/src/test/scala/org/apache/gluten/execution/WholeStageTransformerSuite.scala
+++ b/gluten-core/src/test/scala/org/apache/gluten/execution/WholeStageTransformerSuite.scala
@@ -23,7 +23,7 @@ import org.apache.gluten.utils.Arm
 import org.apache.spark.SparkConf
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.{DataFrame, GlutenQueryTest, Row}
-import org.apache.spark.sql.execution.{CommandResultExec, SparkPlan}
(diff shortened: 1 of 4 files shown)


gluten-core/src/test/scala/org/apache/gluten/execution/WholeStageTransformerSuite.scala
```diff
@@ -252,6 +252,34 @@ abstract class WholeStageTransformerSuite
...
+   * @param tag
+   *   : class of the expected plan.
+   * @param childTag
+   *   : class of the expected plan's child.
+   * @tparam T
+   *   : type of the expected plan.
+   * @tparam PT
+   *   : type of the expected plan's child.
+   */
+  def checkSparkOperatorChainMatch[T <: SparkPlan, PT <: SparkPlan](
```
zhztheplayer, line 269: Should we use `T <: UnaryExecNode` ?

zhztheplayer:
> spark-test-spark32-sl

I think it's in suite `GlutenSQLQueryTestSuite` and you could try modifying `supportedList` to run a single test.

zhztheplayer:
cc @marin-ma

marin-ma (approved):
LGTM. Thanks!

PR #5675, opened 2024-05-09
[VL] Support celeborn sort based shuffle

kerwin-zk:
## What changes were proposed in this pull request?

#4048 

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #5786, opened 2024-05-17
[VL] Refine evict logic in sort shuffle writer

marin-ma:
A follow-up PR of https://github.com/apache/incubator-gluten/pull/5675

Remove the mapping from input vector to row index ranges. The process can be done in one loop.
Also addresses some code clean up.

### DIFF
--- a/cpp/velox/shuffle/VeloxSortBasedShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxSortBasedShuffleWriter.cc
@@ -15,46 +15,19 @@
  * limitations under the License.
  */
 
-#include "VeloxSortBasedShuffleWriter.h"
-#include "memory/ArrowMemory.h"
+#include "shuffle/VeloxSortBasedShuffleWriter.h"
 #include "memory/VeloxColumnarBatch.h"
 #include "memory/VeloxMemoryManager.h"
 #include "shuffle/ShuffleSchema.h"
 #include "utils/Common.h"
 #include "utils/VeloxArrowUtils.h"
 #include "utils/macros.h"
+
 #include "velox/common/base/Nulls.h"
 #include "velox/type/Type.h"
 #include "velox/vector/ComplexVector.h"
 
-#if defined(__x86_64__)
-#include <immintrin.h>
-#include <x86intrin.h>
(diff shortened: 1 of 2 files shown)


marin-ma:
Could you help to review commit https://github.com/apache/incubator-gluten/pull/5786/commits/8c195f2b45e7e302014b25da630040e870d5bac0? Thanks!

marin-ma:
@kerwin-zk Could you help to review commit https://github.com/apache/incubator-gluten/pull/5786/commits/8c195f2b45e7e302014b25da630040e870d5bac0? Thanks!

marin-ma:
@kerwin-zk Just rebased. Could you help to review again? Thanks!

cpp/velox/shuffle/VeloxSortBasedShuffleWriter.cc
```diff
@@ -154,83 +127,71 @@ arrow::Status VeloxSortBasedShuffleWriter::write(std::shared_ptr<ColumnarBatch>
...
-        if (tempVectorIndex == -1) {
-          tempVectorIndex = vectorIndex;
-          baseRowIndex = rowIndex;
-          tempRowIndex = rowIndex;
-        } else {
-          if (vectorIndex == tempVectorIndex && rowIndex == tempRowIndex + 1) {
-            size += 1;
-            tempRowIndex = rowIndex;
+        int32_t numRowsInRange = 1;
+        std::vector<facebook::velox::IndexRange> groupedIndices;
```
kerwin-zk, line 158: Is there a need to reserve here?

marin-ma: Post reply here:

> @kerwin-zk I don't think so. Sort shuffle applies when the shuffle partition is large, so there won't be may rows to the partition in a input RowVector, therefore the actual size of groupedIndices should be small. Reserving maxRowsPerBatch can be a waste of memory.

What do you think?

kerwin-zk: @marin-ma  I think it's okay.

kerwin-zk (approved):
LGTM. Thanks!

Issue #5744, opened 2024-05-14
[CH] Crash on exit with Poco exception


zhanglistar:
### Backend

CH (ClickHouse)

### Bug description

Log is:
```
2024/05/14 16:27:30,975 INFO [dispatcher-Executor] JniLibLoader: Library /data9/hadoop/yarn/local/usercache/xumingyong/appcache/application_1710462985812_739851/container_e57_1710462985812_739851_01_000015/./libch.so has been loaded using path-loading method
2024-05-14 16:27:31.236 <Information> ClickHouseBackend: Init environment variables.
2024-05-14 16:27:31.238 <Debug> CHUtil: Set settings from config key:spark.gluten.sql.columnar.backend.ch.runtime_config.local_engine.settings.log_processors_profiles value:true
2024-05-14 16:27:31.238 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.input_format_orc_row_batch_size value:10000
2024-05-14 16:27:31.238 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.join_algorithm value:grace_hash
2024-05-14 16:27:31.239 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.max_block_size value:8192
2024-05-14 16:27:31.239 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.max_bytes_before_external_group_by value:1610612736
2024-05-14 16:27:31.239 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.max_bytes_before_external_sort value:1073741824
2024-05-14 16:27:31.239 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.max_bytes_in_join value:1073741824
2024-05-14 16:27:31.239 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.max_memory_usage value:3221225472
2024-05-14 16:27:31.239 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.max_threads value:1
2024-05-14 16:27:31.239 <Debug> CHUtil: Set settings key:spark.gluten.sql.columnar.backend.ch.runtime_settings.output_format_orc_compression_method value:snappy
[...]

Issue #5741, opened 2024-05-14
[CH] Crash on gluten exit


zhanglistar:
### Backend

CH (ClickHouse)

### Bug description

<img width="1231" alt="image" src="https://github.com/apache/incubator-gluten/assets/985418/3a5353ce-b82a-44f5-b26b-178fde3943a5">

But, when I gdb the core, get, 
```
BFD: Warning: /data/corefiles/core.VM Thread.78310.1715039755 is truncated: expected core file size >= 1950076928, found: 1391566848.
```

The core is not complete to trace it, need to dig why.

We first found the core on occur 2024.2.12.

### Spark version

Spark-3.3.x

zhanglistar:
ulimit -a
```
yarn@hello-dn119:/data/corefiles$ ulimit -a
core file size          (blocks, -c) unlimited
data seg size           (kbytes, -d) unlimited
scheduling priority             (-e) 0
file size               (blocks, -f) unlimited
pending signals                 (-i) 1030220
max locked memory       (kbytes, -l) 64
max memory size         (kbytes, -m) unlimited
open files                      (-n) 1000000
pipe size            (512 bytes, -p) 8
POSIX message queues     (bytes, -q) 819200
real-time priority              (-r) 0
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 65536
virtual memory          (kbytes, -v) unlimited
file locks                      (-x) unlimited
```
Not the limit problem.

zhanglistar:
This `yarn.nodemanager.sleep-delay-before-sigkill.ms` parameter of yarn cause this.

zhanglistar:
(gdb) bt
#0  0x00007fda35788428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007fda3578a02a in __GI_abort () at abort.c:89
#2  0x00007fda33e8fc15 in __gnu_cxx::__verbose_terminate_handler() ()
   from /data7/hadoop/yarn/local/filecache/187072/jdk-8u391-linux-x64.tar.gz/jdk1.8.0_391/jre/lib/amd64/server/libjvm.so
#3  0x00007fda33e8de36 in __cxxabiv1::__terminate(void (*)()) ()
[...]

PR #5787, opened 2024-05-17
[GLUTEN-5741][CH] Fix core dump when executor exits

exmy:

(Fixes: \#5741 #5744)

https://github.com/apache/incubator-gluten/blob/10182a52d409cb659169cbfe1d7f1869d9205dce/gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java#L131

When running on yarn cluster, `libPath` is a symbolic link existed on the working directory of yarn container, like `/data5/hadoop/yarn/local/usercache/xumingyong/appcache/application_1710462985812_777680/container_e57_1710462985812_777680_01_000321/./libch.so`. It's an `APPLICATION` level resource and will be removed when container exits. So it may have been removed here when reading real path, which causes `Files.isSymbolicLink(Paths.get(libPath))` return false in that the library (e.g. libch.so) cannot be unload and the method `JNI_OnUnload` cannot be invoked (which casuses core dump). As a workaround, we can read real path when loading the library.

## How was this patch tested?

 manual tests

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
+++ b/gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
@@ -82,7 +82,21 @@ public class JniLibLoader {
     loaded.forEach(JniLibLoader::unloadFromPath);
   }
 
+  private static String toRealPath(String libPath) {
+    String realPath = libPath;
+    try {
+      while (Files.isSymbolicLink(Paths.get(realPath))) {
+        realPath = Files.readSymbolicLink(Paths.get(realPath)).toString();
+      }
+      LOG.info("Read real path {} for libPath {}", realPath, libPath);
+      return realPath;
+    } catch (Throwable th) {
+      throw new GlutenException("Error to read real path for libPath: " + libPath, th);
+    }
+  }
+
   private static synchronized void loadFromPath0(String libPath, boolean requireUnload) {
+    libPath = toRealPath(libPath);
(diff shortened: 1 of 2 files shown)


gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
```diff
@@ -125,13 +139,10 @@ public static synchronized void unloadFromPath(String libPath) {
       return;
     }
 
+    LOG.info("Starting unload library path: {} ", libPath);
     REQUIRE_UNLOAD_LIBRARY_PATHS.remove(libPath);
 
     try {
-      while (Files.isSymbolicLink(Paths.get(libPath))) {
```
exmy, line 131: When running on yarn cluster, `libPath` is a symbolic link existed on the working directory of yarn container, like `/data5/hadoop/yarn/local/usercache/xumingyong/appcache/application_1710462985812_777680/container_e57_1710462985812_777680_01_000321/./libch.so`. It's an `APPLICATION` level resource and will be removed when container exists. So it may have been removed here when reading real path, which causes `Files.isSymbolicLink(Paths.get(libPath))` return false in that the library cannot be unload. As a workaround, we can read real path when loading the library.

exmy:
@baibaichen @zhztheplayer Could you help review this change? Thanks!

baibaichen (approved):
LGTM.

baibaichen:
@exmy please update comment and merge it.

exmy:
> @exmy please update comment and merge it.

done

PR #4562, opened 2024-01-29
[GLUTEN-1439][CH] Refactor: move some backend-specific expression transformer to cpp-ch/local-engine/Parser to prevent duplicated expression evaluation

taiyang-li:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#1439)  

Refactor: move some backend-specific expression transformer to cpp-ch/Parser to prevent duplicated expression evaluation

ClickHouse backend will benefit from performance improvements brought by these changes. Below is a typical case: 

``` sql 
CREATE TEMPORARY VIEW lineitem
USING org.apache.spark.sql.parquet
OPTIONS (
  path "/data1/liyang/cppproject/gluten/gluten-core/src/test/resources/tpch-data/lineitem"
) ;

select count(1) from (select y,z from lineitem lateral view posexplode(split(l_comment, ' ')) x as y,z); 
```

Time cost:
```
New: 1 row selected (0.443 seconds) 
Old: 1 row selected (0.613 seconds) 
``` 

ClickHouse plan: 

New: 
```
      Expression (Pre-projection In Generate)
      Header: mapFromArrays(range(length(assumeNotNull(ifNull(splitByRegexp( _0,l_comment,-1_1),[]_2)))),assumeNotNull(ifNull(splitByRegexp( _0,l_comment,-1_1),[]_2))) Map(UInt64, String)
      Actions: INPUT : 0 -> l_comment Nullable(String) : 0
               COLUMN Const(String) ->  _0 String : 1
               COLUMN Const(Int32) -> -1_1 Int32 : 2
               COLUMN Const(Array(String)) -> []_2 Array(String) : 3
               FUNCTION splitByRegexp( _0 :: 1, l_comment :: 0, -1_1 :: 2) -> splitByRegexp( _0,l_comment,-1_1) Nullable(Array(String)) : 4
               FUNCTION ifNull(splitByRegexp( _0,l_comment,-1_1) :: 4, []_2 :: 3) -> ifNull(splitByRegexp( _0,l_comment,-1_1),[]_2) Array(String) : 2
               FUNCTION assumeNotNull(ifNull(splitByRegexp( _0,l_comment,-1_1),[]_2) :: 2) -> assumeNotNull(ifNull(splitByRegexp( _0,l_comment,-1_1),[]_2)) Array(String) : 3
[...]

PR #5796, opened 2024-05-17
[CORE] Refactor ExpressionTransformer

ulysses-you:

This pr enriches the trait `ExpressionTransformer` so that we can reuse much codebase when transforming expression.

Hightlight the trait change:
```scala
trait ExpressionTransformer {
  def substraitExprName: String
  def children: Seq[ExpressionTransformer]
  def original: Expression
  def dataType: DataType = original.dataType
  def nullable: Boolean = original.nullable

  def doTransform(args: java.lang.Object): ExpressionNode = {
    val functionMap = args.asInstanceOf[java.util.HashMap[String, java.lang.Long]]
    val funcName: String =
      ConverterUtils.makeFuncName(substraitExprName, original.children.map(_.dataType))
    val functionId = ExpressionBuilder.newScalarFunction(functionMap, funcName)
    val childNodes = children.map(_.doTransform(args)).asJava
    val typeNode = ConverterUtils.getTypeNode(dataType, nullable)
    ExpressionBuilder.makeScalarFunction(functionId, childNodes, typeNode)
  }
}

trait LeafExpressionTransformer extends ExpressionTransformer {
  final override def children: Seq[ExpressionTransformer] = Nil
}

trait UnaryExpressionTransformer extends ExpressionTransformer {
  def child: ExpressionTransformer
  final override def children: Seq[ExpressionTransformer] = child :: Nil
}

trait BinaryExpressionTransformer extends ExpressionTransformer {
  def left: ExpressionTransformer
  def right: ExpressionTransformer
  final override def children: Seq[ExpressionTransformer] = left :: right :: Nil
}
```

## How was this patch tested?

Pass CI


gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala
```diff
@@ -16,17 +16,74 @@
...
+
+trait BinaryExpressionTransformer extends ExpressionTransformer {
+  def left: ExpressionTransformer
+  def right: ExpressionTransformer
+  final override def children: Seq[ExpressionTransformer] = left :: right :: Nil
+}
+
+// ==== Expression transformer basic interface end ====
+
+case class GenericExpressionTransformer(
```
ulysses-you, line 64: Most expression can leverage this transformer rather than make a custom expression. Take an example `MakeDecimal`, it is a unary expression in vanilla Spark but native requires one more arg `nullOnOverflow`, then we can pass one more child `LiteralTransformer` to 
`GenericExpressionTransformer`:
```scala
GenericExpressionTransformer(
  substraitExprName,
  Seq(
    replaceWithExpressionTransformerInternal(m.child, attributeSeq, expressionsMap),
    LiteralTransformer(m.nullOnOverflow)),
  m
)
```

gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -439,13 +437,6 @@ trait SparkPlanExecApi {
     GenericExpressionTransformer(substraitExprName, children, original)
   }
 
-  def genMd5Transformer(
```
ulysses-you, line 442: no backend override these since https://github.com/apache/incubator-gluten/pull/4562

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala
```diff
@@ -16,17 +16,74 @@
...
+  def children: Seq[ExpressionTransformer]
   def original: Expression
   def dataType: DataType = original.dataType
+  def nullable: Boolean = original.nullable
+
+  def doTransform(args: java.lang.Object): ExpressionNode = {
+    val functionMap = args.asInstanceOf[java.util.HashMap[String, java.lang.Long]]
+    // TODO: the funcName seems can be simplified to `substraitExprName`
+    val funcName: String =
+      ConverterUtils.makeFuncName(substraitExprName, original.children.map(_.dataType))
```
ulysses-you, line 39: Just highlight the implicit change, the function config of function name is never been used. The default function config of makeFuncName is `NON`.

gluten-core/src/main/scala/org/apache/gluten/expression/UnaryExpressionTransformer.scala
```diff
@@ -203,43 +102,32 @@ case class RandTransformer(
     substraitExprName: String,
     explicitSeed: ExpressionTransformer,
     original: Rand)
-  extends ExpressionTransformerWithOrigin {
+  extends LeafExpressionTransformer {
```
ulysses-you, line 105: This is an another example to show how to work with `ExpressionTransformer`, rand is a unary expression in vanilla Spark but native does not accpet arg, then we mark it as leaf expression transformer.

ulysses-you:
cc @jinchengchenghh @zhztheplayer @rui-mo  @zzcclp  thank you

zhztheplayer (approved):
Thanks!

jinchengchenghh (approved):

zzcclp (approved):
LGTM, please rebase.

ulysses-you:
thank you all!

PR #5799, opened 2024-05-17
[VL] Use conf to control C2R occupied memory

XinShuoWang:

In the current design, the Column2Row operation is completed in one go, which consumes a lot of memory and causes the program OOM. In this commit, I modified the C2R operation into multiple operations, which will greatly reduce the peak memory of the C2R operation. In addition, there should be some performance advantages from memory reuse.

## How was this patch tested?


FelixYBW:
@jinchengchenghh

XinShuoWang:
@jinchengchenghh @zhztheplayer Can you review this pr?

cpp/core/jni/JniWrapper.cc
```diff
@@ -580,17 +583,28 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
...
+    jlong rowId) {
   JNI_METHOD_START
   auto ctx = gluten::getRuntime(env, wrapper);
+  auto& conf = ctx->getConfMap();
+
   auto columnarToRowConverter = ctx->objectStore()->retrieve<ColumnarToRowConverter>(c2rHandle);
   auto cb = ctx->objectStore()->retrieve<ColumnarBatch>(batchHandle);
-  columnarToRowConverter->convert(cb);
+
+  int64_t column2RowMemThreshold = 200 * 1024 * 1024;
```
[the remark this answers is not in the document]
XinShuoWang: done

cpp/core/jni/JniWrapper.cc
```diff
@@ -580,17 +583,28 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
...
+    jlong rowId) {
   JNI_METHOD_START
   auto ctx = gluten::getRuntime(env, wrapper);
+  auto& conf = ctx->getConfMap();
+
   auto columnarToRowConverter = ctx->objectStore()->retrieve<ColumnarToRowConverter>(c2rHandle);
   auto cb = ctx->objectStore()->retrieve<ColumnarBatch>(batchHandle);
-  columnarToRowConverter->convert(cb);
+
+  int64_t column2RowMemThreshold = 256 * 1024 * 1024;
```
jinchengchenghh, line 595: Can you place this config computation to GlutenConfig.cc

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -16,46 +16,61 @@
...
   numCols_ = rowVector->childrenSize();
 
   fast_ = std::make_unique<velox::row::UnsafeRowFast>(rowVector);
 
   size_t totalMemorySize = 0;
   if (auto fixedRowSize = velox::row::UnsafeRowFast::fixedRowSize(velox::asRowType(rowVector->type()))) {
-    totalMemorySize += fixedRowSize.value() * numRows_;
+    GLUTEN_CHECK(
+        memoryThreshold > fixedRowSize.value(),
+        "Only 1 row exceed Column2RowMemoryThreshold, which is " + velox::succinctBytes(memoryThreshold));
```
jinchengchenghh, line 44: Drop `Only `

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -16,46 +16,61 @@
...
+    auto rowSize = fixedRowSize.value();
+    numRows_ = std::min<int64_t>(memoryThreshold / rowSize, vectorLength - rowId);
```
jinchengchenghh, line 46: Why is it vectorLength - rowId

FelixYBW:
I met similar issue recently, reports as ArrowContext OOM. Is the root cause that the RowVector too larger? or because the data in each row is too large? By default GLuten configure Velox as 4K batch, but looks many Velox operators may exceed 4K limit.

Does the PR hold on the batch and output part of the batch each time?

FelixYBW:
We may need couple of UTs to cover this

XinShuoWang:
> I met similar issue recently, reports as ArrowContext OOM. Is the root cause that the RowVector too larger? or because the data in each row is too large? By default GLuten configure Velox as 4K batch, but looks many Velox operators may exceed 4K limit.
> 
> Does the PR hold on the batch and output part of the batch each time?

- In my test case, each row is only 20KB in size, but the size of RowVector is very large, which eventually leads to OOM. 
- In my test case the number of output lines exceeds the 4K limit is because of the Generate Operator.  
- This PR is aimed to hold on the batch and output part of the batch each time

cpp/core/jni/JniWrapper.cc
```diff
@@ -580,17 +583,28 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
     JNIEnv* env,
     jobject wrapper,
     jlong batchHandle,
-    jlong c2rHandle) {
+    jlong c2rHandle,
+    jlong rowId) {
```
ulysses-you, line 587: Can we pass a range [startRowId, endRowId) to naitve ? I think it should be more clear to use the config at java side.

FelixYBW:
> > I met similar issue recently, reports as ArrowContext OOM. Is the root cause that the RowVector too larger? or because the data in each row is too large? By default GLuten configure Velox as 4K batch, but looks many Velox operators may exceed 4K limit.
> > Does the PR hold on the batch and output part of the batch each time?
> 
> * In my test case, each row is only 20KB in size, but the size of RowVector is very large, which eventually leads to OOM.
> * In my test case the number of output lines exceeds the 4K limit is because of the Generate Operator.
> * This PR is aimed to hold on the batch and output part of the batch each time

Can we use the batch size (4K row) by default as the config? We can still add the threshold but if customer doesn't config threshold, let's use 4K row each time.

FelixYBW:
why the PR is closed?

XinShuoWang:
> why the PR is closed?

Sorry, I accidentally forced push the main branch, which caused the PR to be closed. The new PR is here: https://github.com/apache/incubator-gluten/pull/5952.

PR #5808, opened 2024-05-20
[CORE]Add branch protection rule

weiting-chen:

Add branch protection rule with at least 1 review pass.

## How was this patch tested?
N/A

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,7 +31,14 @@ github:
     squash: true
     merge: false
     rebase: false
-  protected_branches: ~
+  protected_branches:
+    main:
+      required_pull_request_reviews:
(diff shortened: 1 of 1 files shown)


zhouyuan (approved):
👍

.asf.yaml
```diff
@@ -31,7 +31,14 @@ github:
     squash: true
     merge: false
     rebase: false
-  protected_branches: ~
+  protected_branches:
+    main:
+      required_pull_request_reviews:
+        dismiss_stale_reviews: true
+        required_approving_review_count: 1
+      required_signatures: true
```
Yohahaha, line 39: may I ask why signatures is needed? seems developer need re-import ssh key with signed key in Github

PR #5809, opened 2024-05-20
[CORE] Unify the aggregate function name mapping

ulysses-you:

Unify the special aggregate function name mapping to reduce the duplicated code, e.g., fisrt/last with ignoreNulls.

## How was this patch tested?

Pass CI


ulysses-you:
cc @PHILO-HE @marin-ma thank you

PR #5810, opened 2024-05-20
[CORE] ASF repo config: Set required_signatures to false

zhztheplayer:
`required_signatures:true` blocks a couple of PRs from merging. Let's disable this rule at the moment, until we come up with a BKM about signing PR commits.


weiting-chen (approved):
Looks good to me.

PR #5811, opened 2024-05-20
[VL] Enable local sort-based shuffle

marin-ma:
This is an experimental feature. In TPCH benchmark, we observed significant performance drop and higher memory pressure 
 using default shuffle partitions (2x~4x vcore#) comparing with hash-based shuffle. The sort-based shuffle will be enabled if shuffle partitions is greater than `spark.gluten.sql.columnar.shuffle.sort.threshold` (default value 100'000). It's recommended not to enable local sort-base shuffle until we fix the performance.


marin-ma:
The UT failure of commit https://github.com/apache/incubator-gluten/pull/5811/commits/56580aa690fd5099c55fcf01e213b9cad1267fad is due to metrics changed which causes AQE suites failure https://github.com/apache/incubator-gluten/actions/runs/9185004959

kerwin-zk (approved):
LGTM. Thanks!

PR #5815, opened 2024-05-20
[VL] Add config for memory pool init capacity to reduce arbitration times

Yohahaha:

Add new config `spark.gluten.sql.columnar.backend.velox.memInitCapacity` to reduce memory arbitration times, default value is 8M.

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
@@ -1207,6 +1207,13 @@ object GlutenConfig {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefaultString("1GB")
 
+  val COLUMNAR_VELOX_MEM_INIT_CAPACITY =
+    buildConf("spark.gluten.sql.columnar.backend.velox.memInitCapacity")
+      .internal()
+      .doc("The initial memory capacity.")
+      .bytesConf(ByteUnit.BYTE)
+      .createWithDefaultString("8MB")
+
   val COLUMNAR_VELOX_SSD_CACHE_PATH =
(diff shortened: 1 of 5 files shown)


Yohahaha:
> We previously hardcoded it as 8M, this PR make it configurable, right?

preciously is 0.

zhztheplayer:
> Add new config spark.gluten.sql.columnar.backend.velox.memInitCapacity to reduce memory arbitration times

Is there any difference on performance in your test observed within the change? @Yohahaha

Yohahaha:
> > Add new config spark.gluten.sql.columnar.backend.velox.memInitCapacity to reduce memory arbitration times
> 
> Is there any difference on performance in your test observed within the change? @Yohahaha

currently, no, but it can reduce the frequency of enter arbitration in theory, and may have benefits when set this config larger, eg 128M.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1207,6 +1207,13 @@ object GlutenConfig {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefaultString("1GB")
 
+  val COLUMNAR_VELOX_MEM_INIT_CAPACITY =
+    buildConf("spark.gluten.sql.columnar.backend.velox.memInitCapacity")
+      .internal()
+      .doc("The initial memory capacity.")
```
zhztheplayer, line 1213: Could rephase to “The initial memory capacity to reserve for a newly created Velox query memory pool."

Reference https://github.com/facebookincubator/velox/blob/9c9712eac7e335b3636f5193b0a061c6015f9d10/velox/common/memory/Memory.h#L162

zhztheplayer (dismissed):

zhztheplayer (approved):

PR #5824, opened 2024-05-21
[VL] RAS: Reuse same code path with heuristic planner for convention enforcement

zhztheplayer:
1. In RAS, reuse transition facilities added in https://github.com/apache/incubator-gluten/pull/5767 to do convention enforcement.
2. Rename RAS's `Convention` and relevant classes to new names starting with `Conv`.
3. New trait `KnownChildrenConventions` to allow user-defined children batch-type requirements for a specific `SparkPlan` node. To better understand the need of this, one can think about `convention` as something similar to an existing physical property vanilla Spark already implements, say `ordering`, `SparkPlan` has both method `outputOrdering` and `requiredChildOrdering` to define the ordering and allowed children's orderings for a single plan node.
4. Other essential code refactors.


gluten-core/src/main/scala/org/apache/gluten/planner/property/Conv.scala
```diff
@@ -0,0 +1,106 @@
...
+ */
+package org.apache.gluten.planner.property
+
+import org.apache.gluten.extension.columnar.transition.{Convention, ConventionReq, Transition}
+import org.apache.gluten.ras.{Property, PropertyDef}
+import org.apache.gluten.ras.rule.{RasRule, Shape, Shapes}
+
+import org.apache.spark.sql.execution._
+
+sealed trait Conv extends Property[SparkPlan] {
```
rui-mo, line 25: It would be nice to add some description for the new API and its members.

rui-mo (approved):

gluten-core/src/main/scala/org/apache/gluten/planner/property/Conv.scala
zhztheplayer: Thanks. I will add some then.

BTW it's a RAS component as of now (and all source files in package `org.apache.gluten.extension.columnar.enumerated` and `org.apache.gluten.planner`) that isn't used by regular code path.

PR #5825, opened 2024-05-21
[GLUTEN-5720][VL] Enable left and right semi join type in smj

JkSelf:

Support left  and right semi join in SMJ.

## How was this patch tested?

Adding new unit test.


rui-mo (approved):
Thanks!

Issue #5823, opened 2024-05-21
[CH] Make shuffle writer exit gracefully when tasks in executors are killed 


taiyang-li:
### Backend

CH (ClickHouse)

### Bug description

When spark speculation execution is enabled, some tasks maybe killed by driver because another attempt had finished successfully.
Those tasks failed because they are not exiting gracefully.

![image](https://github.com/apache/incubator-gluten/assets/8181003/4cfb597d-135b-438e-9c05-785ffbced130)
 

### Spark version

Spark-3.3.x

### Spark configurations

``` 
--conf spark.speculation=true
    --conf spark.speculation.multiplier=1.1
    --conf spark.speculation.quantile=0.5
```

### Relevant logs

```
2024-05-22 14:33:45.853 <Information> LocalExecutor: build pipeline 0.314 ms; create executor 0.026 ms;
2024-05-22 14:33:47.230 <Debug> jni: [{"id":1,"name":"kFilter","time":2068,"input_wait_time":2748222,"output_wait_time":1767,"steps":[{"name":"Filter","description":"WHERE","processors":[{"name":"FilterTransform","time":1762,"output_rows":605146,"output_bytes":10892628,"input_rows":605146,"input_bytes":10892628}]},{"name":"Expression","description":"Remove nullable properties","processors":[{"name":"ExpressionTransform","time":306,"output_rows":605146,"output_bytes":10287482,"input_rows":605146,"input_bytes":10892628}]}]},{"id":0,"name":"kRead","time":1332017,"input_wait_time":770,"output_wait_time":3352,"steps":[{"name":"SubstraitFileSourceStep","description":"read local files","processors":[{"name":"SubstraitFileSource","time":1332017,"output_rows":605146,"output_bytes":10892628,"input_rows":0,"input_bytes":0}]}]}]
2024/05/22 14:33:47.244 INFO [dispatcher-Executor] spark.executor.Executor: Executor is trying to kill task 91.0 in stage 6.0 (TID 2398), reason: another attempt succeeded
2024/05/22 14:33:47.245 ERROR [Executor task launch worker for task 91.0 in stage 6.0 (TID 2398)] celeborn.client.ShuffleClientImpl: Exception raised while registering shuffle 6 with 136 mapper and 200 partitions.
java.lang.InterruptedException
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1039)
[...]

PR #5839, opened 2024-05-22
[GLUTEN-5787][CH]Make pipeline and shuffle exit gracefully when tasks in executors are killed or interrupted

taiyang-li:

Changes:
- Clean code:  remove useless JNIs and classes under cpp-ch
- Support cancel for all gluten processors. It was triggered when task is killed or shut down.
- Make sure offheap memory free does not throw exception. Ref: https://zhuanlan.zhihu.com/p/65454580 

(Fixes: \#5787 \#5823) 

## How was this patch tested?

Manual


liuneng1994 (dismissed):
LGTM

backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
```diff
@@ -305,13 +311,24 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
       }
     }
     var closed = false
+    val cancelled = false
```
zhanglistar, line 314: whta's usage of this var?

taiyang-li: Useless, just to keep the same style with previous codes...

liuneng1994 (dismissed):
LGTM

gluten-core/src/main/java/org/apache/gluten/vectorized/GeneralOutIterator.java
```diff
@@ -60,10 +61,18 @@ public final void close() {
     }
   }
 
+  public final void cancel() {
```
taiyang-li, line 64: @rui-mo I added a new interface for GeneralOutIterator. For velox it is by default empty. It is needed by clickhouse backend to cancel the pipeline before task is killed, to ensure some resources are released gracefully. Could you help review the change? Thanks!

zhztheplayer: Is it possible to place the API directly to [CH's out iterator impl](https://github.com/apache/incubator-gluten/blob/e1469f0c3859c9dd5f0edb9d2f5890794a540cee/backends-clickhouse/src/main/java/org/apache/gluten/vectorized/BatchIterator.java#L27)?

We can pull up it later when Velox backend gets to have similar demand.

taiyang-li: It seems  hard. Currently `GeneralOutIterator` is used in spark stages to get output both for Velox and CH.  Adding `cancel` only for CH `BatchIterator` will introduce a series of `if (iter.isInstanceOf[BatchIterator])`. And it it ugly.

rui-mo: Sorry, I missed this message. It is possible to create resIter as BatchIterator instead of GeneralOutIterator in CH backend? In this way, we could use its custom member function directly.

> val resIter: GeneralOutIterator =
      transKernel.createKernelWithBatchIterator(
        inputPartition.plan,
        splitInfoByteArray,
        inBatchIters,
        false)

zhztheplayer: > Sorry, I missed this message. It is possible to create resIter as BatchIterator instead of GeneralOutIterator in CH backend? In this way, we could use its custom member function directly.

+1

It's safe and ideal now to make `org.apache.gluten.vectorized.CHNativeExpressionEvaluator#createKernelWithBatchIterator` directly return a `BatchIterator` than `GeneralOutIterator`. We share `GeneralOutIterator`'s code only for legacy reason. It just includes some JNI method definition and doesn't have to be reused.

taiyang-li: Thanks for your advice. I'll try.

liuneng1994 (approved):

PR #5849, opened 2024-05-23
[CORE] Remove static modifier on TreeMemoryConsumers.Factory.map

zhztheplayer:
The factory objects are already globally static. Its inner map doesn't need to be static. Make it non-static to take better encapsulation.

Also add some basic tests for `TreeMemoryConsumer`.


zhztheplayer:
cc @marin-ma @JkSelf

marin-ma (approved):
LGTM. Thanks!

Issue #5852, opened 2024-05-23
[CH] throw GlutenException: Missmatch result columns size


shuai-xu:
### Backend

CH (ClickHouse)

### Bug description

When run sql d_917_1, it throws GlutenException: Missmatch result columns size.
<img width="1388" alt="image" src="https://github.com/apache/incubator-gluten/assets/20960838/a99f46ea-a0b2-452b-aede-b926ca1caa3b">

### Spark version

Spark-3.3.x

PR #5619, opened 2024-05-06
[GLUTEN-5618][CH] Fix 'Position x is out of bound in Block' error when executing count distinct

zzcclp:
## What changes were proposed in this pull request?

When excuting count distinct, the group by keys are also in the count distinct expression, it will throw 'Position x is out of bound in Block' error or core dump.

RC:
CH backend will remove the duplicated column when executing pipeline.

Close #5618.

(Fixes: #5618)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #5853, opened 2024-05-23
[GLUTEN-5852] [CH] fix mismatch result columns size exception

shuai-xu:

For sql whose agg columns contains two same const value, they will be transfromed to same name, In #5619 , it called distinct on agg's keys and outputs, so the const columns may be removed. 

(Fixes: \#5852)

## How was this patch tested?

This patch was tested by unit tests.


zzcclp:
@ulysses-you @liujiayi771 please help to review, thanks.

liujiayi771:
@shuai-xu I currently don't have the env for CH, so I used Velox to test the case you provided. The result is consistent with vanilla Spark. Does this issue only occur in CH? Could you tell me what CH returns before the fix?

In addition, the original logic here is to reuse some duplicated pulled-out `Attribute`s. If the issue you encountered is only related to `Literal`, could you only add a special handling for `Literal` here?

shuai-xu:
> @shuai-xu I currently don't have the env for CH, so I used Velox to test the case you provided. The result is consistent with vanilla Spark. Does this issue only occur in CH? Could you tell me what CH returns before the fix? I also tried debugging and did not enter into the code you modified.
> 
> In addition, the original logic here is to reuse some duplicated pulled-out `Attribute`s. If the issue you encountered is only related to `Literal`, could you only add a special handling for `Literal` here?

This case only happens in CH. It will throw an exception as shown in the issue before fix. The problem here is that the resue logic consider two different attibutes with same name as same one, so later in CH, it can't know whether they are really same.

ulysses-you:
The literal should not happen in group keys.  Can we add `RemoveLiteralFromGroupExpressions` into extended optimzier rule ? It seems Spark only optimize the group by case but miss to optimize the distinct case.

ulysses-you (approved):

zzcclp (approved):
LGTM

liujiayi771 (approved):

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
```diff
@@ -158,9 +158,11 @@ object PullOutPreProject extends RewriteSingleNode with PullOutProjectHelper {
       val newAgg = copyBaseAggregateExec(agg)(
         newGroupingExpressions = newGroupingExpressions,
         newAggregateExpressions = newAggregateExpressions)
-      val preProject = ProjectExec(
-        eliminateProjectList(agg.child.outputSet, expressionMap.values.toSeq),
-        agg.child)
+      // ISSUE-5852: literals with same names are lost in expressionMap, need addback to Project
```
liujiayi771, line 161: Was the modification of the logic in this code unnecessary in your earliest version?

gluten-core/src/main/scala/org/apache/gluten/utils/PullOutProjectHelper.scala
```diff
@@ -70,6 +76,19 @@ trait PullOutProjectHelper {
           .toAttribute
     }
 
+  protected def getMissingLiterals(exprs: Seq[NamedExpression]): Seq[NamedExpression] = {
+    val literalMap = new mutable.HashMap[Expression, NamedExpression]()
+    exprs.filter {
+      case alias: Alias =>
+        alias.child match {
+          case _: Literal =>
+            literalMap.put(alias.child.canonicalized, alias).nonEmpty
```
liujiayi771, line 85: Can you add a comment here explaining that the purpose of the HashMap is to obtain duplicate literals?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
shuai-xu: Yes, the literals is not in the project outputs, it will case fallback.

liujiayi771 (approved):

liujiayi771:
LGTM. cc @ulysses-you Do you have any other suggestions?

gluten-core/src/main/scala/org/apache/gluten/utils/PullOutProjectHelper.scala
```diff
@@ -61,7 +61,13 @@ trait PullOutProjectHelper {
       replaceBoundReference: Boolean = false): Expression =
     expr match {
       case alias: Alias =>
-        projectExprsMap.getOrElseUpdate(alias.child.canonicalized, alias).toAttribute
+        alias.child match {
+          case _: Literal =>
+            projectExprsMap.getOrElseUpdate(alias.child.canonicalized, alias)
```
ulysses-you, line 66: Does `getMissingLiterals` method do the same things with this code ? This code and `getMissingLiterals` say if there is an `Alias(Literal)` then we add one more `Alias(Literal)` to project. What happens if there are three `Alias(Literal)` ?

Do you mean we should always include `Alias(Literal)` ? if so we can change `projectExprsMap.getOrElseUpdate(alias.child.canonicalized, alias)` to `projectExprsMap.getOrElseUpdate(alias, alias)` and remove `getMissingLiterals`.

shuai-xu: Yes, all literal alias should be added.

ulysses-you (approved):

PR #5854, opened 2024-05-23
[VL] Enable NaN tests for array functions

rui-mo:
A bunch of commits are added in Velox upstream to fix the semantics of NaN in array functions, where NaNs are changed to be treated as equal, which becomes aligned with Spark semantics. Tests on corresponding array functions could be enabled with these modifications.


JkSelf (approved):
LGTM. Thanks for your fix.

Issue #5315, opened 2024-04-08
[VL] Add SampleExec operator support


gaoyangxiaozhu:
### Description

Using this issue to track support for sampleExec operation. https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/basicPhysicalOperators.scala#L313

gaoyangxiaozhu:
Hey @zhouyuan / @FelixYBW / @rui-mo ,

I'm starting to work on this feature, and here's a quick draft to share my ideas and one basic but crucial question need you help confirm before i start the detailed design doc and code implement.

In Velox, there's no such sampleNode available, but it does support random sampling push down with jimmy's this PR [Table sampling push down](https://github.com/facebookincubator/velox/commit/3d9cf528b065052e3d7ff6d0603035f5b56ebfc3#diff-58b64d1e01b72d7e092a092708f13e1a2785f0151709bce18b1c736a0c8d28ee) , it support accelerate random sampling based on Bernoulli trials by push the random sampling operations to table scan.

So, my current idea is to leverage the existing sample filter pushdown logic. This involves transforming the vanilla Spark sampleExec node into a filter node, with the sample operation transformed into a sample filter expression., ultimately pushing the random sampling filter  down to the scan filter.

The issue here is that Spark uses  [XORShiftRandom](https://github.com/apache/spark/blob/207d675110e6fa699a434e81296f6f050eb0304b/core/src/main/scala/org/apache/spark/util/random/RandomSampler.scala#L57C1-L58C49) pseudorandom number generator and [Bernoulli trials based sampler](https://github.com/apache/spark/blob/207d675110e6fa699a434e81296f6f050eb0304b/core/src/main/scala/org/apache/spark/util/random/RandomSampler.scala#L116C1-L123C6) for sampling, while Velox employs__gnu_cxx::sfmt19937 random number generator with geometric distribution for sampling, checking here [RandomUtil](https://github.com/facebookincubator/velox/blob/main/velox/common/base/RandomUtil.h#L93C1-L95C33). Thus, even with the same `fraction` and `seed` (which user can specify), the sampled results will differ from the vinalla Spark results vs velox.

So, my basic but crucial question here is whether it's acceptable for such correctness issues to exist in the sample scenario when offload to velox. 

If it is acceptable, I'll then start by drafting a design document in the gluten channel and then proceed with the code implementation.

gaoyangxiaozhu:
[...]

PR #5856, opened 2024-05-23
[VL][Core] SampleExec Operator Native Support 

gaoyangxiaozhu:

The PR introduce the `SampleExec` native support (`SampleExecTransformer`) for velox backend, basically it transfer the `SampleExec` spark physical operator to velox `ProjectFilter` operator with `rand(seed) < fraction` expression.

The feature is under feature flag `spark.gluten.sql.columnarSampleEnabled` with default is false now.

SampleExec support two sample way
 1. Sampling with replacement (withReplacement is set to true)
 2. sampling without replacement  (default)

Right now, we only support for samling without replacement.

(Fixes: \#5315)

## How was this patch tested?
locally test with turning on feature flag 

A case use `df.sample(0.5, 10)`

image


gaoyangxiaozhu:
@rui-mo / @zhouyuan / @zhli1142015  help review

gluten-core/src/main/scala/org/apache/gluten/execution/SampleExecTransformer.scala
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.gluten.substrait.rel.{RelBuilder, RelNode}
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression, LessThan, Literal, Rand}
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.types.DoubleType
+
+import scala.collection.JavaConverters._
+
+case class SampleExecTransformer(
```
rui-mo, line 35: Would you add some description for how sample operator is translated into Substrait?

gaoyangxiaozhu:
> > Right now, we only support for samling without replaceme
> 
> Will Sampling with replacement (withReplacement is set to true) fallback? how complex if we add support to Velox? how useful it is?

sample with relpacement allow pick up same row multiple times, that means it actually not simple filter but re-generating the data set,  that is not worked by simple convert to a filter rand expression. I don't yet figur out a good way to supprot that part unless to make a new `sampleNode` in Velox side.

gluten-core/src/main/scala/org/apache/gluten/execution/SampleExecTransformer.scala
gaoyangxiaozhu: sure, will do

gaoyangxiaozhu: done

rui-mo (approved):
Thanks.

gaoyangxiaozhu:
@zhli1142015 help merge

Issue #4422, opened 2024-01-16
io.glutenproject.vectorized.ColumnarBatchOutIterator.nativeSpill Coredump


yixi-gu:
### Backend

VL (Velox)

### Bug description

run tpcds benchmark 30 hours, it occured the coredump problem

### Spark version

Spark-3.3.x

### Spark configurations

spark.master yarn
spark.deploy-mode client
spark.eventLog.enabled true
spark.eventLog.dir hdfs://master:9000/sparklogs
spark.driver.cores 8
spark.driver.memory 9g
spark.driver.maxResultSize 27g
spark.executor.instances 24
spark.executor.cores 4
spark.executor.memory 6g
spark.executor.memoryOverhead 1g
spark.memory.offHeap.enabled true
spark.gluten.enabled true
spark.plugins io.glutenproject.GlutenPlugin
spark.gluten.sql.columnar.backend.lib velox
spark.shuffle.manager org.apache.spark.shuffle.sort.ColumnarShuffleManager
spark.executorEnv.VELOX_HDFS hdfs://master:9000
spark.gluten.loadLibFromJar true
spark.io.compression.codec lz4
spark.memory.offHeap.size 10g
spark.task.maxFailures 1
spark.default.parallelism 192
spark.kryoserializer.buffer 288k
spark.memory.fraction 0.55
spark.shuffle.file.buffer 128k
spark.sql.adaptive.maxShuffledHashJoinLocalMapThreshold 8m
spark.sql.autoBroadcastJoinThreshold 1m
spark.sql.broadcastTimeout 6900
spark.sql.files.maxPartitionBytes 384m
spark.sql.files.openCostInBytes 3m
spark.sql.shuffle.partitions 192

### System information

![Uploading image.png…]()

### Relevant logs

```bash
Stack: [0x0000fffeeadfe000,0x0000fffeeaffe000],  sp=0x0000fffeeaff7de0,  free space=2023k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libgluten.so+0x4da324]  Java_io_glutenproject_vectorized_ColumnarBatchOutIterator_nativeSpill+0xdc
J 23934  io.glutenproject.vectorized.ColumnarBatchOutIterator.nativeSpill(JJ)J (0 bytes) @ 0x0000ffffaae144f8 [0x0000ffffaae14480+0x78]
C  0x0000000000000000

[...]

PR #5874, opened 2024-05-27
[GLUTEN-4422][CORE] Fix core dump caused by spill on closed iterator

WangGuangxin:

There is a chance that `nativeSpill` is called after iterator is closed.  We can ignore this.

(Fixes: \#4422)

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -544,7 +544,12 @@ JNIEXPORT jlong JNICALL Java_org_apache_gluten_vectorized_ColumnarBatchOutIterat
   auto ctx = gluten::getRuntime(env, wrapper);
 
   auto it = ctx->objectStore()->retrieve<ResultIterator>(iterHandle);
-  return it->spillFixedSize(size);
+  if (it == nullptr) {
+    LOG(WARNING) << "Try to spill on a closed iterator!";
+    return 0L;
+  } else {
+    return it->spillFixedSize(size);
+  }
   JNI_METHOD_END(kInvalidResourceHandle)
 }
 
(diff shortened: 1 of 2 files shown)


cpp/core/jni/JniWrapper.cc
```diff
@@ -544,7 +544,12 @@ JNIEXPORT jlong JNICALL Java_org_apache_gluten_vectorized_ColumnarBatchOutIterat
   auto ctx = gluten::getRuntime(env, wrapper);
 
   auto it = ctx->objectStore()->retrieve<ResultIterator>(iterHandle);
-  return it->spillFixedSize(size);
+  if (it == nullptr) {
```
zhztheplayer, line 547: `it == nullptr` is not that clear to indicate `it` is closed. Could we leverage some kind of closing flags? Probably from Java / Scala side?

WangGuangxin: ok

zhztheplayer (approved):

zhztheplayer:
Thank you for the fix!

gluten-data/src/main/java/org/apache/gluten/vectorized/ColumnarBatchOutIterator.java
```diff
@@ -80,7 +80,11 @@ public IMetrics getMetricsInternal() throws IOException, ClassNotFoundException
   }
 
   public long spill(long size) {
-    return nativeSpill(iterHandle, size);
+    if (!closed.get()) {
```
boneanxs, line 83: Recently I also meet this issue(we haven't include this pr yet). But I'm curious in this way, IIUC, we can't spill memory that this `NativeMemoryManager` holding if the iterator is closed? Since we still hold the memory after the iterator is closed.

If it's an issue, can we directly get `NativeMemoryManager` to spill itself? By calling `shrinkCapacity({pool}, remaining)` in the native.

@zhztheplayer @WangGuangxin

PR #5888, opened 2024-05-28
[GLUTEN-5890][Core] Enhance jni signature with a more readable way

Donvi:

This is to optimize current const text in code like "([B)V", which means the jbooleanArray with void return.
So we currently have this way for optimize:
We use typeid to get the hash of the type and  build a map from type_info.hash_code to the signature.

## How was this patch tested?

This is test in ut, for all the value equals to previous values.
And check for all single value.

### DIFF
--- a/cpp/core/jni/JniCommon.h
+++ b/cpp/core/jni/JniCommon.h
@@ -31,6 +31,7 @@
 #include "utils/exception.h"
 
 static jint jniVersion = JNI_VERSION_1_8;
+static map<int, string> type2sig = buildTypeMapping();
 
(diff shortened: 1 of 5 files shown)


cpp/core/jni/JniCommon.h
```diff
@@ -31,6 +31,11 @@
 #include "utils/exception.h"
 
 static jint jniVersion = JNI_VERSION_1_8;
+static map<int, string> type2sig = buildTypeMapping();
```
Yohahaha, line 34: maybe used `constexpr`?

PR #5914, opened 2024-05-29
[CORE] Move driver/executor endpoint to CH backend

ulysses-you:

The driver/executor endpoint only used for ch backend but live in gluten-core, this pr move these to ch backend. After this change, we can also remove the `BroadcastApi`,  `BroadcastApi` is used to handle some special code for ch backend broadcast join with driver endpoint and we can put these code to ch broadcast join.

## How was this patch tested?

Pass CI


gluten-core/pom.xml
```diff
@@ -24,10 +24,6 @@
       <artifactId>gluten-ui</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <dependency>
-      <groupId>com.github.ben-manes.caffeine</groupId>
-      <artifactId>caffeine</artifactId>
-    </dependency>
```
ulysses-you, line 30: Required for CH backend, so Velox backend can get rid of it.

ulysses-you:
cc @zhztheplayer @PHILO-HE @binmahone @zzcclp thank you

zhztheplayer:
So this lifecycle feature is not really that common for all backends? (cc @baibaichen)

baibaichen:
> So this lifecycle feature is not really that common for all backends? (cc @baibaichen)

@ulysses-you @zhztheplayer 

We use such mechanism to free memory allocated at Broadcast case. Don't Velox need it?

ulysses-you:
@zhztheplayer @baibaichen IIUC, Velox backend follows the vanilla Spark broadcast exchange way. I'm not sure why do we need additional lifecycle.

zhztheplayer:
> @zhztheplayer @baibaichen IIUC, Velox backend follows the vanilla Spark broadcast exchange way. I'm not sure why do we need additional lifecycle.

I assume CH backend uses the feature for storing broadcast data at process wise. If so it's OK to me to move the code if we don't tend do the same in Velox backend.

zhztheplayer (dismissed):
+0.9 since we can still pull up some code once Velox backend requires that feature

gluten-core/src/main/scala/org/apache/gluten/backendsapi/ListenerApi.scala
```diff
@@ -16,11 +16,11 @@
...
-import org.apache.spark.SparkConf
+import org.apache.spark.{SparkConf, SparkContext}
 
 trait ListenerApi {
-  def onDriverStart(conf: SparkConf): Unit = {}
+  def onDriverStart(sc: SparkContext, conf: SparkConf): Unit = {}
   def onDriverShutdown(): Unit = {}
-  def onExecutorStart(conf: SparkConf): Unit = {}
+  def onExecutorStart(executorID: String, conf: SparkConf): Unit = {}
   def onExecutorShutdown(): Unit = {}
```
zhztheplayer, line 25: nit

```suggestion
  def onDriverStart(sc: SparkContext, pc: PluginContext): Unit = {}
  def onDriverShutdown(): Unit = {}
  def onExecutorStart(pc: PluginContext): Unit = {}
  def onExecutorShutdown(): Unit = {}
```

to keep up with Spark's plugin APIs

gluten-core/src/main/scala/org/apache/gluten/backendsapi/TransformerApi.scala
```diff
@@ -70,4 +70,6 @@ trait TransformerApi {
   def getNativePlanString(substraitPlan: Array[Byte], details: Boolean): String
 
   def packPBMessage(message: Message): Any
+
+  def invalidateSQLExecutionResource(executionId: String): Unit = {}
```
zhztheplayer, line 74: nit

Probably add a comment saying that the method is only for testing so far?

ulysses-you: addressed

zhztheplayer (approved):
Thanks!

gluten-core/src/main/scala/org/apache/gluten/backendsapi/ListenerApi.scala
ulysses-you: addressed

zhztheplayer (approved):

PR #5889, opened 2024-05-28
[VL] Gluten-it: Improve test report table rendering

zhztheplayer:
This just optimizes the representation of test report table of subcommand `queries-compare`.

Before:
```
...
| Query ID| Was Passed| Expected Row Count| Actual Row Count| Baseline Planning Time (Millis)| Planning Time (Millis)| Baseline Query Time (Millis)| Query Time (Millis)| Query Time Variation|
|       q1|       true|                100|              100|                             967|                    184|                         3530|                2372|               48.82%|
|       q2|       true|               2513|             2513|                             275|                    182|                         1916|                1516|               26.39%|
|       q3|       true|                100|              100|                              38|                     54|                          984|                 726|               35.54%|
|       q4|       true|                100|              100|                             577|                    496|                        11244|                4606|              144.12%|
|       q5|       true|                100|              100|                             340|                    159|                         2996|                1970|               52.08%|
...
```

After: 

```
...
| Query ID | Passed |    Row Count     | Planning Time (Millis)  |  Query Time (Millis)  | Speedup |
|          |        | Vanilla | Gluten |  Vanilla   |   Gluten   |  Vanilla  |  Gluten   |         |
|----------|--------|---------|--------|------------|------------|-----------|-----------|---------|
|        q1|    true|      100|     100|         682|         157|       3899|       2366|   64.79%|
|        q2|    true|     2513|    2513|         250|         155|       2195|       1575|   39.37%|
|        q3|    true|      100|     100|          44|          38|        871|        502|   73.51%|
|        q4|    true|      100|     100|         567|         395|      10284|       3978|  158.52%|
|        q5|    true|      100|     100|         262|         144|       2810|       1458|   92.73%|
...
```

PR #5917, opened 2024-05-30
[VL] Following #5889, correct / simplify the table indenting algorithm

zhztheplayer:
The patch just corrects the identing algorithm of #5889.


JkSelf (approved):
LGTM. Thanks.

PR #5923, opened 2024-05-30
[GLUTEN-5656][CORE][FOLLOWUP] Support GetStructField with NullLiteralNode as subqueries not executing during validation

jackylee-ch:
After #5658 , we met validation failed with q24a ad q24b while using bloomFilter.

## How was this patch tested?
manul test


zhztheplayer (approved):
Thank you for catching.

PR #5925, opened 2024-05-30
[VL] Gluten-it: Optimize Maven dependency list

zhztheplayer:
As title. Optimize Maven dependency list, clarify dependency scopes and visibilities.


zhouyuan (approved):
👍

PR #5926, opened 2024-05-30
[GLUTEN-4942][VL] Adding arrow port in VCPKG

zhouyuan:

this patch adds arrow port in vcpkg list
also restore the change in package-vcpkg.sh 

## How was this patch tested?

local verified

### DIFF
--- a/dev/package-vcpkg-gha.sh
+++ b/dev/package-vcpkg-gha.sh
@@ -6,7 +6,8 @@ CURRENT_DIR=$(cd "$(dirname "$BASH_SOURCE")"; pwd)
 GLUTEN_DIR="$CURRENT_DIR/.."
 
 cd "$GLUTEN_DIR"
-source ./dev/vcpkg/env.sh
+source /opt/rh/devtoolset-9/enable
+source /opt/gluten/dev/vcpkg/env.sh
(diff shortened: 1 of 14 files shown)


FelixYBW:
@zhouyuan how far we can move to static build as default?

zhouyuan:
@FelixYBW 
Spark UT is working on static building now. https://github.com/apache/incubator-gluten/pull/5906
We may need extra 2-3 patches for 1) allow Velox to use external arrow build 2) improve the library loading code path to work on both static and dynamic build, as in static build no libarrow.so/libparquet.so is required 3) improve GHA to cover both packaging method

CC @ulysses-you @Yohahaha

ulysses-you (approved):
thank you @zhouyuan

Yohahaha:
thank you!

> how far we can move to static build as default?

static build by default is ok, but let's stay aligned on this matter: Gluten needs to support both dynamic and static, and maybe keep one build checks for dynamic link in CI is better. @zhouyuan

zhouyuan:
> thank you!
> 
> > how far we can move to static build as default?
> 
> static build by default is ok, but let's stay aligned on this matter: Gluten needs to support both dynamic and static, and maybe keep one build checks for dynamic link in CI is better. @zhouyuan

@Yohahaha yes, I hear the inputs, will keep the dynamic packaging 

thanks
-yuan

PR #5928, opened 2024-05-30
[VL] Fix shuffle with round robin partitioning fail

ulysses-you:

We should validate the project before sort rather than using ProjectExecTransformer directly. The hash expression may not support offload to native, e.g., the child output contains null type or something else.

The following query would fail:

```sql
spark.sql("select 1 as c, null as x").repartition().collect
```

## How was this patch tested?

add test


ulysses-you:
cc @zhztheplayer @marin-ma thank you

marin-ma:
Thanks. Seems like this failure is because we creates a hash computation for all input columns. If there are any NullType in the input, those column types will be converted to UNKNOWN Type in Velox, but Velox doesn't support hash computing on UNKNOWN types. I would suggest we keep this check, meanwhile drop the NullType input columns in the hash computation as the null values doesn't affect the hash computation (`hash(x)` and `hash(x, null)` produce the same result). Then we can avoid fallback for NullTypes.

ulysses-you:
@marin-ma  Is null type the only unsupported data type for hash expression ? If so, I can drop null column before going to hash expression.

ulysses-you:
@marin-ma never mind.. I will drop null type column in this pr

marin-ma (approved):
LGTM. Thanks for the fix!

PR #5935, opened 2024-05-31
[VL] Remove reselect build side in ShuffledHashJoinExecTransformer

zml1206:

After #5750 , velox backend supported use the smaller table to build hashmap in shuffled hash join ,`preferredBuildSide` in `ShuffledHashJoinExecTransformer` is unnecessary and not as accurate as AQE, so remove it.

## How was this patch tested?

TPC-DS


jackylee-ch (approved):

zhztheplayer (approved):
Thanks!

PR #5942, opened 2024-05-31
[VL] Enable sort-based shuffle in micro benchmark

marin-ma:
Removed `ShuffleSplitBenchmark.cc` as it's no longer maintained. We can use the GenericBenchmark.cc to test the shuffle write performance. Steps are added to the doc.


Yohahaha (approved):
👍

PR #5943, opened 2024-05-31
[CH] Fix left and substring with length -1

liuneng1994:

Fix left and substring with length -1

expect empty string

## How was this patch tested?

unit tests


zzcclp (approved):
LGTM

PR #5952, opened 2024-06-02
[VL] Use conf to control C2R occupied memory

XinShuoWang:

In the current design, the Column2Row operation is completed in one go, which consumes a lot of memory and causes the program OOM. In this commit, I modified the C2R operation into multiple operations, which will greatly reduce the peak memory of the C2R operation. In addition, there should be some performance advantages from memory reuse.

## How was this patch tested?


cpp/core/jni/JniWrapper.cc
```diff
@@ -580,17 +583,28 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
...
+    jlong rowId) {
   JNI_METHOD_START
   auto ctx = gluten::getRuntime(env, wrapper);
+  auto& conf = ctx->getConfMap();
+
   auto columnarToRowConverter = ctx->objectStore()->retrieve<ColumnarToRowConverter>(c2rHandle);
   auto cb = ctx->objectStore()->retrieve<ColumnarBatch>(batchHandle);
-  columnarToRowConverter->convert(cb);
+
+  int64_t column2RowMemThreshold = 256 * 1024 * 1024;
```
Yohahaha, line 595: move default value to GlutenConfig.h

cpp/core/operators/c2r/ColumnarToRow.h
```diff
@@ -27,7 +28,15 @@ class ColumnarToRowConverter {
 
   virtual ~ColumnarToRowConverter() = default;
 
-  virtual void convert(std::shared_ptr<ColumnarBatch> cb = nullptr) = 0;
+  // We will start conversion from the 'rowId' row of 'cb'. The maximum memory consumption during the grabbing and
+  // swapping process is 'memoryThreshold' bytes. The number of rows successfully converted is stored in the 'numRows_'
+  // variable.
+  virtual void
+  convert(std::shared_ptr<ColumnarBatch> cb = nullptr, int64_t rowId = 0, int64_t memoryThreshold = INT64_MAX) = 0;
```
Yohahaha, line 35: could we configure `memoryThreshold` for ColumnarToRowConverter when initializing it?

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1031,6 +1038,12 @@ object GlutenConfig {
       .checkValue(_ > 0, s"$GLUTEN_MAX_BATCH_SIZE_KEY must be positive.")
       .createWithDefault(4096)
 
+  val GLUTEN_COLUMNAR_TO_ROW_MEM_THRESHOLD =
+    buildConf(GLUTEN_COLUMNAR_TO_ROW_MEM_THRESHOLD_KEY)
+      .internal()
+      .longConf
+      .createWithDefault(256 * 1024 * 1024)
+
```
Yohahaha, line 1046: we have support the usage of `set spark.gluten.sql.columnarToRowMemoryThreshold=256m`

Yohahaha (commented):
> In the current design, the Column2Row operation is completed in one go, which consumes a lot of memory and causes the program OOM

do you mean current implementation will convert all columnar batches into unsafe rows in one run? it's seems unbelievable, could you double check it and give a case which c2r conversion's peak memory is extremely high?

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -16,46 +16,64 @@
...
+    totalMemorySize = rowSize * numRows_;
   } else {
-    for (auto i = 0; i < numRows_; ++i) {
-      totalMemorySize += fast_->rowSize(i);
+    int64_t i = rowId;
+    for (; i < vectorLength; ++i) {
+      auto rowSize = fast_->rowSize(i);
+      if (UNLIKELY(totalMemorySize + rowSize > memoryThreshold)) {
+        GLUTEN_CHECK(
+            i >= 1,
```
ulysses-you, line 55: i > rowId + 1

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -16,46 +16,64 @@
...
   }
 
   if (veloxBuffers_ == nullptr) {
-    // First allocate memory
-    veloxBuffers_ = velox::AlignedBuffer::allocate<uint8_t>(totalMemorySize, veloxPool_.get());
-  }
-
-  if (veloxBuffers_->capacity() < totalMemorySize) {
-    velox::AlignedBuffer::reallocate<uint8_t>(&veloxBuffers_, totalMemorySize);
+    veloxBuffers_ = velox::AlignedBuffer::allocate<uint8_t>(memoryThreshold, veloxPool_.get());
```
ulysses-you, line 67: The first parameter of `velox::AlignedBuffer::allocate` is numElements rather than size, it will allocate the sizeof(uint8_t) * memoryThreshold.

gluten-data/src/main/java/org/apache/gluten/vectorized/NativeColumnarToRowJniWrapper.java
```diff
@@ -38,8 +38,8 @@ public long handle() {
 
   public native long nativeColumnarToRowInit(long memoryManagerHandle) throws RuntimeException;
 
-  public native NativeColumnarToRowInfo nativeColumnarToRowConvert(long batchHandle, long c2rHandle)
-      throws RuntimeException;
+  public native NativeColumnarToRowInfo nativeColumnarToRowConvert(
+      long batchHandle, long c2rHandle, long rowId) throws RuntimeException;
```
ulysses-you, line 42: Can we pass memoryThreshold to jni ? It should be easy to track the config at java side.

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
ulysses-you, line 67: It seems the 256MB memory is a bit large ? In general, one columnar batch is usually less than 64MB.(shuffle advisory size is 64MB)

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -16,46 +16,64 @@
...
 
   fast_ = std::make_unique<velox::row::UnsafeRowFast>(rowVector);
 
   size_t totalMemorySize = 0;
   if (auto fixedRowSize = velox::row::UnsafeRowFast::fixedRowSize(velox::asRowType(rowVector->type()))) {
-    totalMemorySize += fixedRowSize.value() * numRows_;
+    GLUTEN_CHECK(
+        memoryThreshold > fixedRowSize.value(),
+        "spark.gluten.sql.columnarToRowMemoryThreshold(" + velox::succinctBytes(memoryThreshold) +
+            ") is too small, it can't hold even one row(" + velox::succinctBytes(fixedRowSize.value()) + ")");
```
FelixYBW, line 45: Let's not report exception here, just try to allocate the new size for one row. After all, user needs to increases the number.

FelixYBW:
@XinShuoWang Can you update?

FelixYBW:
can you fix the UT?

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -16,46 +16,66 @@
...
   numCols_ = rowVector->childrenSize();
 
   fast_ = std::make_unique<velox::row::UnsafeRowFast>(rowVector);
 
   size_t totalMemorySize = 0;
   if (auto fixedRowSize = velox::row::UnsafeRowFast::fixedRowSize(velox::asRowType(rowVector->type()))) {
-    totalMemorySize += fixedRowSize.value() * numRows_;
+    if (memoryThreshold < fixedRowSize.value()) {
+      memoryThreshold = fixedRowSize.value();
+      LOG(ERROR) << "spark.gluten.sql.columnarToRowMemoryThreshold(" + velox::succinctBytes(memoryThreshold) +
```
FelixYBW, line 44: Let's change it to LOG(INFO) as a reminder only.

FelixYBW:
@XinShuoWang can you fix the UT?

cpp/core/jni/JniWrapper.cc
```diff
@@ -494,17 +497,33 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
...
+  int64_t column2RowMemThreshold;
+  auto it = conf.find(kColumnToRowMemoryThreshold);
+  bool confIsLeagal =
+      ((it == conf.end()) ? false : std::all_of(it->second.begin(), it->second.end(), [](unsigned char c) {
+        return std::isdigit(c);
+      }));
+  if (confIsLeagal) {
+    column2RowMemThreshold = std::stoll(it->second);
+  } else {
+    column2RowMemThreshold = std::stoll(kColumnToRowMemoryDefaultThreshold);
```
yma11, line 518: Add a warning here?

yma11:
@XinShuoWang The UT fails as the default buffer size is 64MB and it's too large when task memory is small. You may can  change the threshold value for these UTs.

FelixYBW:
@XinShuoWang can you fix the UT? the bug triggered many queries failed in our test.

FelixYBW:
@XinShuoWang Would you continue to fix the issue? If not we can pickup and fix it. The bug blocked our adoption

cpp/core/jni/JniWrapper.cc
```diff
@@ -530,16 +533,36 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
     JNIEnv* env,
     jobject wrapper,
     jlong c2rHandle,
-    jlong batchHandle) {
+    jlong batchHandle,
+    jlong rowId) {
   JNI_METHOD_START
   auto columnarToRowConverter = ObjectStore::retrieve<ColumnarToRowConverter>(c2rHandle);
   auto cb = ObjectStore::retrieve<ColumnarBatch>(batchHandle);
   columnarToRowConverter->convert(cb);
```
zhztheplayer, line 541: look like a mistake not removing this line

cpp/core/jni/JniWrapper.cc
```diff
@@ -530,16 +533,36 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
...
+  if (confIsLeagal) {
+    column2RowMemThreshold = std::stoll(it->second);
+  } else {
+    LOG(INFO)
+        << "Because the spark.gluten.sql.columnToRowMemoryThreshold configuration item is invalid, the kColumnToRowMemoryDefaultThreshold default value is used, which is "
+        << kColumnToRowMemoryDefaultThreshold << " byte";
+    column2RowMemThreshold = std::stoll(kColumnToRowMemoryDefaultThreshold);
+  }
+
+  columnarToRowConverter->convert(cb, rowId, column2RowMemThreshold);
```
zhztheplayer, line 560: The operations could become expensive given that the JNI function is called for every single row?

cpp/core/operators/c2r/ColumnarToRow.h
FelixYBW: @zhztheplayer can you fix it?

cpp/core/jni/JniWrapper.cc
```diff
@@ -530,16 +533,35 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
     JNIEnv* env,
     jobject wrapper,
     jlong c2rHandle,
-    jlong batchHandle) {
+    jlong batchHandle,
+    jlong rowId) {
```
FelixYBW, line 537: Let's rename rowId as startRow

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -16,46 +16,66 @@
...
+        }
+        break;
+      } else {
+        totalMemorySize += rowSize;
+      }
     }
+    numRows_ = i - rowId;
   }
 
   if (veloxBuffers_ == nullptr) {
```
FelixYBW, line 68: @zhztheplayer we need to reallocate memory if memoThreshold is changed

cpp/core/jni/JniWrapper.cc
```diff
@@ -530,16 +533,35 @@ Java_org_apache_gluten_vectorized_NativeColumnarToRowJniWrapper_nativeColumnarTo
...
-  columnarToRowConverter->convert(cb);
+  auto ctx = gluten::getRuntime(env, wrapper);
+
+  auto& conf = ctx->getConfMap();
+  int64_t column2RowMemThreshold;
+  auto it = conf.find(kColumnToRowMemoryThreshold);
+  bool confIsLegal =
+      ((it == conf.end()) ? false : std::all_of(it->second.begin(), it->second.end(), [](unsigned char c) {
+        return std::isdigit(c);
+      }));
```
FelixYBW, line 549: The check is called every batch, it's unnecessary. Let's save the threshold

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
zhztheplayer: Passing a memory threshold for a iterator's `next` operation is not an optimal way if the iterator holds some memory across invocations. I am afraid we should redesign the code more or less.

FelixYBW:
Looks like the generator operator already limit the output rowvector size to the one we set.

The OOM isn't caused by C2R, but the R2C after parquet scan, because the struct column is too large. The way to solve OOM is to decrease the batch. 

Another operator which may output large batch size is hashjoin. So maybe the PR is still useful.

image

zhztheplayer:
> So maybe the PR is still useful.

I am thinking about using the similar mechanism with the one added in https://github.com/apache/incubator-gluten/pull/6009 as a general solution. So we don't have to add such kind of pre-processors for operators one by one @FelixYBW

zhztheplayer:
> > So maybe the PR is still useful.
> 
> I am thinking about using the similar mechanism with the one added in #6009 as a general solution. So we don't have to add such kind of pre-processors for operators one by one @FelixYBW

And the operator may need to be added in Velox to fit into Velox's task processing. We don't want to break whole-stage transformer by the operator in query plan.

FelixYBW:
> #6009

It should be more efficient to break or combine batches in the next operator, instead of implement a dedicated one. Dedicated operator must have a memcpy.

FelixYBW:
> ay need to be added in Velox to fit into Velox's task processing. We don't want to break whole-stage transformer by the operator in query plan.

Velox should hornor the max batch size always, or limit the batch by memory size.

FelixYBW:
> ow operation is completed in one go, which consumes a lot of memory and causes the program OOM. In this commit, I modified the C2

@XinShuoWang what's your case to have large batch size? My issue can be fixed by reduce spark.gluten.sql.columnar.maxBatchSize and spark.sql.parquet.columnarReaderBatchSize

FelixYBW:
It's caused by a bug when calculating row size.

FelixYBW:
OK, it's still the generator operator's issue, I set the batch size to 16, but actually the output row number may be as high as 500.

image

Another issue is that to some columns, the internalrow format may have >11 memory size than a columnar format. Still investigating

image

FelixYBW:
> Another issue is that to some columns, the internalrow format may have >11x memory size than a columnar format. Still investigating
> 

The output should be correct. Looks it's a big advantage of columnar format.

zhztheplayer:
Error still occurs at case `SPARK-16995: flat mapping on Dataset containing a column created with lit/expr`, should investigate.


```
Stack: [0x00007f9cf43da000,0x00007f9cf44db000],  sp=0x00007f9cf44d89e0,  free space=1018k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0xb04f5a]  Unsafe_GetLong+0x7a
J 7081  sun.misc.Unsafe.getLong(Ljava/lang/Object;J)J (0 bytes) @ 0x00007f9f6d53728e [0x00007f9f6d5371c0+0xce]
j  org.apache.spark.unsafe.Platform.getLong(Ljava/lang/Object;J)J+5
j  org.apache.spark.unsafe.bitset.BitSetMethods.isSet(Ljava/lang/Object;JI)Z+66
j  org.apache.spark.sql.catalyst.expressions.UnsafeRow.isNullAt(I)Z+14
j  org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificOrdering.compare(Lorg/apache/spark/sql/catalyst/InternalRow;Lorg/apache/spark/sql/catalyst/InternalRow;)I+27
j  org.apache.spark.sql.execution.GroupedIterator.fetchNextGroupIterator()Z+127
j  org.apache.spark.sql.execution.GroupedIterator.hasNext()Z+8
J 8209 C2 scala.collection.Iterator$$anon$11.hasNext()Z (35 bytes) @ 0x00007f9f6e6bf538 [0x00007f9f6e6bf340+0x1f8]
j  org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext()V+42
j  org.apache.spark.sql.execution.BufferedRowIterator.hasNext()Z+11
j  org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext()Z+4
j  org.apache.spark.sql.execution.SparkPlan.$anonfun$getByteArrayRdd$1(ZILscala/collection/Iterator;)Lscala/collection/Iterator;+207
j  org.apache.spark.sql.execution.SparkPlan$$Lambda$3173.apply(Ljava/lang/Object;)Ljava/lang/Object;+12
j  org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2(Lscala/Function1;Lorg/apache/spark/TaskContext;ILscala/collection/Iterator;)Lscala/collection/Iterator;+2
j  org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2$adapted(Lscala/Function1;Lorg/apache/spark/TaskContext;Ljava/lang/Object;Lscala/collection/Iterator;)Lscala/collection/Iterator;+7
j  org.apache.spark.rdd.RDD$$Lambda$2870.apply(Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;+13
j  org.apache.spark.rdd.MapPartitionsRDD.compute(Lorg/apache/spark/Partition;Lorg/apache/spark/TaskContext;)Lscala/collection/Iterator;+27
j  org.apache.spark.rdd.RDD.computeOrReadCheckpoint(Lorg/apache/spark/Partition;Lorg/apache/spark/TaskContext;)Lscala/collection/Iterator;+24
j  org.apache.spark.rdd.RDD.iterator(Lorg/apache/spark/Partition;Lorg/apache/spark/TaskContext;)Lscala/collection/Iterator;+40
j  org.apache.spark.scheduler.ResultTask.runTask(Lorg/apache/spark/TaskContext;)Ljava/lang/Object;+201
j  org.apache.spark.TaskContext.runTaskWithListeners(Lorg/apache/spark/scheduler/Task;)Ljava/lang/Object;+2
j  org.apache.spark.scheduler.Task.run(JILorg/apache/spark/metrics/MetricsSystem;ILscala/collection/immutable/Map;Lscala/Option;)Ljava/lang/Object;+254
j  org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Lorg/apache/spark/executor/Executor$TaskRunner;Lscala/runtime/BooleanRef;)Ljava/lang/Object;+43
j  org.apache.spark.executor.Executor$TaskRunner$$Lambda$2820.apply()Ljava/lang/Object;+8
j  org.apache.spark.util.Utils$.tryWithSafeFinally(Lscala/Function0;Lscala/Function0;)Ljava/lang/Object;+4
j  org.apache.spark.executor.Executor$TaskRunner.run()V+457
j  java.util.concurrent.ThreadPoolExecutor.runWorker(Ljava/util/concurrent/ThreadPoolExecutor$Worker;)V+95
j  java.util.concurrent.ThreadPoolExecutor$Worker.run()V+5
j  java.lang.Thread.run()V+11
v  ~StubRoutines::call_stub
V  [libjvm.so+0x69fb72]  JavaCalls::call_helper(JavaValue*, methodHandle*, JavaCallArguments*, Thread*)+0xe32
V  [libjvm.so+0x69d183]  JavaCalls::call_virtual(JavaValue*, KlassHandle, Symbol*, Symbol*, JavaCallArguments*, Thread*)+0x263
V  [libjvm.so+0x69d787]  JavaCalls::call_virtual(JavaValue*, Handle, KlassHandle, Symbol*, Symbol*, Thread*)+0x57
V  [libjvm.so+0x73de8c]  thread_entry(JavaThread*, Thread*)+0x6c
V  [libjvm.so+0xad7b0b]  JavaThread::thread_main_inner()+0xdb
V  [libjvm.so+0xad7dd9]  JavaThread::run()+0x299
V  [libjvm.so+0x9651f2]  java_start(Thread*)+0xe2
C  [libpthread.so.0+0x8609]  start_thread+0xd9
```

FelixYBW:
I fixed a bit. Now if the batch is small, we needn't alloate 64M memory, but the accurate memory

FelixYBW:
> Error still occurs at case `SPARK-16995: flat mapping on Dataset containing a column created with lit/expr`, should investigate.

Is it related? I can't see

zhztheplayer (approved):

Issue #5953, opened 2024-06-02
[VL] Prevent pushdown filters with unsupported data types to scan node 


WangGuangxin:
### Description

Prevent pushdown filters with unsupported data types to scan node, in case scan fallback

PR #5954, opened 2024-06-02
[GLUTEN-5953][VL] Prevent pushdown filters with unsupported data types to scan node

WangGuangxin:

Prevent pushdown filters with unsupported data types to scan node in case the scan fallback to vanilla scan operator

(Fixes: \#5953)

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
@@ -18,11 +18,13 @@ package org.apache.gluten.execution
 
 import org.apache.gluten.GlutenConfig
 import org.apache.gluten.backendsapi.velox.VeloxBackendSettings
+import org.apache.gluten.benchmarks.RandomParquetDataGenerator
 import org.apache.gluten.utils.VeloxFileSystemValidationJniWrapper
 
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.catalyst.expressions.GreaterThan
 import org.apache.spark.sql.execution.ScalarSubquery
+import org.apache.spark.sql.types._
 
 class VeloxScanSuite extends VeloxWholeStageTransformerSuite {
   protected val rootPath: String = getClass.getResource("/").getPath
@@ -117,5 +119,35 @@ class VeloxScanSuite extends VeloxWholeStageTransformerSuite {
     assert(
       !VeloxFileSystemValidationJniWrapper.allSupportedByRegisteredFileSystems(
         Array("file:/test_path/", "unsupported://test_path")))
+  }
+
+  test("unsupported data type scan filter pushdown") {
+    withTempView("t") {
+      withTempDir {
+        dir =>
+          val path = dir.getAbsolutePath
+          val schema = StructType(
+            Array(
+              StructField("short_decimal_field", DecimalType(5, 2), true),
+              StructField("long_decimal_field", DecimalType(32, 8), true),
+              StructField("binary_field", BinaryType, true),
+              StructField("timestamp_field", TimestampType, true)
+            ))
+          RandomParquetDataGenerator(0).generateRandomData(spark, schema, 10, Some(path))
+          spark.catalog.createTable("t", path, "parquet")
+          runQueryAndCompare(
+            """select * from t where long_decimal_field = 3.14)""".stripMargin
+          )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
+          runQueryAndCompare(
+            """select * from t where short_decimal_field = 3.14""".stripMargin
+          )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
+          runQueryAndCompare(
+            """select * from t where binary_field = '3.14'""".stripMargin
+          )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
+          runQueryAndCompare(
+            """select * from t where timestamp_field = current_timestamp()""".stripMargin
+          )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
+      }
+    }
   }
 }
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
@@ -136,7 +136,7 @@ class VeloxScanSuite extends VeloxWholeStageTransformerSuite {
           RandomParquetDataGenerator(0).generateRandomData(spark, schema, 10, Some(path))
           spark.catalog.createTable("t", path, "parquet")
           runQueryAndCompare(
-            """select * from t where long_decimal_field = 3.14)""".stripMargin
+            """select * from t where long_decimal_field = 3.14""".stripMargin
           )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
           runQueryAndCompare(
             """select * from t where short_decimal_field = 3.14""".stripMargin
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -1709,8 +1709,8 @@ bool SubstraitToVeloxPlanConverter::isPushdownSupported(TypePtr inputType) {
 }
 
 bool SubstraitToVeloxPlanConverter::canPushdownScalarFunction(
-  const ::substrait::Expression_ScalarFunction& function,
-  const std::vector<TypePtr>& veloxTypeList) {
+    const ::substrait::Expression_ScalarFunction& function,
+    const std::vector<TypePtr>& veloxTypeList) {
   for (const auto& arg : function.arguments()) {
     if (arg.value().has_scalar_function()) {
       const auto& scalarFunction = arg.value().scalar_function();
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -1705,7 +1705,7 @@ bool SubstraitToVeloxPlanConverter::isPushdownSupported(TypePtr inputType) {
     case TypeKind::HUGEINT:
       return false;
     default:
-      return false;
+      return true;
   }
 }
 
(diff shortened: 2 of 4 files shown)


FelixYBW:
Do you mean these fields are not supported in filter pushdown?
```
              StructField("short_decimal_field", DecimalType(5, 2), true),
              StructField("long_decimal_field", DecimalType(32, 8), true),
              StructField("binary_field", BinaryType, true),
              StructField("timestamp_field", TimestampType, true)
```

WangGuangxin:
> Do you mean these fields are not supported in filter pushdown?
> 
> ```
>               StructField("short_decimal_field", DecimalType(5, 2), true),
>               StructField("long_decimal_field", DecimalType(32, 8), true),
>               StructField("binary_field", BinaryType, true),
>               StructField("timestamp_field", TimestampType, true)
> ```

@FelixYBW Before this PR, Gluten will try to pushdown them, but throw `Subfield filters creation not supported for input type **`, which make FileSourceScan fallback to Vanilla Spark.
After this PR, Gluten will not try to pushdown them.

BTW, `short_decimal_field` is supported to pushdown, the other three fields don't.

WangGuangxin:
cc @PHILO-HE @rui-mo can you help review this?

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
```diff
@@ -74,4 +77,34 @@ class VeloxScanSuite extends VeloxWholeStageTransformerSuite {
...
+            """select * from t where long_decimal_field = 3.14""".stripMargin
+          )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
+          runQueryAndCompare(
+            """select * from t where short_decimal_field = 3.14""".stripMargin
+          )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
+          runQueryAndCompare(
+            """select * from t where binary_field = '3.14'""".stripMargin
+          )(checkGlutenOperatorMatch[FileSourceScanExecTransformer])
+          runQueryAndCompare(
+            """select * from t where timestamp_field = current_timestamp()""".stripMargin
```
rui-mo, line 105: We have made some updates on the timestamp support. Could you help check if the pushdown of timestamp is supported?

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -468,6 +468,13 @@ class SubstraitToVeloxPlanConverter {
   /// 'or' expression are effective on the same column.
   static bool childrenFunctionsOnSameField(const ::substrait::Expression_ScalarFunction& function);
 
+  /// Check whether the data type is supported to pushdown.
+  static bool isPushdownSupported(TypePtr inputType);
+  /// Check whether the scalar function contains data type that doesn't to pushdown.
```
rui-mo, line 473: nit: add one empty line above for readability.

PHILO-HE:
@kecookier, @rui-mo

jiangjiangtian:
I think we should also prevent pushdown filters from singularOrLists.
https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlan.cc#L1694

WangGuangxin:
> I think we should also prevent pushdown filters from singularOrLists. https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlan.cc#L1694

@jiangjiangtian maybe yes, but since there is already type check, I'm not quite sure. https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlan.cc#L2469 .  I think it can be in another PR if needed?

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
WangGuangxin: Yes, I think it's already supported in velox side. But as for now, Gluten can still fallback since `mapToFilters` doesn't support yet.  I think we can support it in next PR.

rui-mo: Got it. Would you like to open an issue to track its support? Thanks.

rui-mo: Thanks.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1681,6 +1681,48 @@ bool SubstraitToVeloxPlanConverter::canPushdownOr(
...
+    case TypeKind::TINYINT:
+    case TypeKind::SMALLINT:
+    case TypeKind::INTEGER:
+    case TypeKind::BIGINT:
+    case TypeKind::REAL:
+    case TypeKind::DOUBLE:
+    case TypeKind::BOOLEAN:
+    case TypeKind::VARCHAR:
+    case TypeKind::ARRAY:
+    case TypeKind::MAP:
```
rui-mo, line 1699: It looks like most type are supported. Would you list the unsupported types instead?

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -473,6 +473,14 @@ class SubstraitToVeloxPlanConverter {
   /// 'or' expression are effective on the same column.
   static bool childrenFunctionsOnSameField(const ::substrait::Expression_ScalarFunction& function);
 
+  /// Check whether the data type is supported to pushdown.
+  static bool isPushdownSupported(TypePtr inputType);
+
+  /// Check whether the scalar function contains data type that doesn't to pushdown.
+  static bool canPushdownScalarFunction(
```
rui-mo, line 480: Could we merge it with `canPushdownFunction` which checks if a scalar function can be pushed down.

jiangjiangtian:
> > I think we should also prevent pushdown filters from singularOrLists. https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlan.cc#L1694
> 
> @jiangjiangtian maybe yes, but since there is already type check, I'm not quite sure. https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlan.cc#L2469 . I think it can be in another PR if needed?

Yes, I think you are right. There is already type check.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1712,19 +1733,6 @@ void SubstraitToVeloxPlanConverter::separateFilters(
   for (const auto& scalarFunction : scalarFunctions) {
     auto filterNameSpec = SubstraitParser::findFunctionSpec(functionMap_, scalarFunction.function_reference());
     auto filterName = SubstraitParser::getNameBeforeDelimiter(filterNameSpec);
-    // Add all decimal filters to remaining functions because their pushdown are not supported.
-    if (format == dwio::common::FileFormat::ORC && scalarFunction.arguments().size() > 0) {
```
rui-mo, line 1716: Looks like previous logic only excludes decimal type for ORC format, but now we are excluding HUGEINT for all file formats. Could you confirm if it is expected?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1600,17 +1600,26 @@ bool SubstraitToVeloxPlanConverter::childrenFunctionsOnSameField(
...
+    return false;
+  }
+
+  // The arg should be field or field with literal.
+  if (!fieldOrWithLiteral(scalarFunction.arguments(), fieldIdx)) {
+    return false;
   }
-  return canPushdown;
+
+  // check whether data type is supported or not
```
rui-mo, line 1617: nit: capitalize the first letter and add `.` at the end for comments.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1686,6 +1695,18 @@ bool SubstraitToVeloxPlanConverter::canPushdownOr(
   return true;
 }
 
+bool SubstraitToVeloxPlanConverter::isPushdownSupported(TypePtr inputType) {
+  // keep the same with mapToFilters
```
rui-mo, line 1699: Ditto.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
kecookier: The ORC format already supports short-decimal predicate pushdown. @jiangjiangtian Can you help test this?

WangGuangxin: The main idea of these three types (`Timestamp/VarBinary/HugeInt`) in this PR is that if we don't put them in `remainingFilter`, then when we call `mapToFilters`, a exception `Subfield filters creation not supported for input type '{}' in mapToFilters`  will throw, which will cause scan fallback. 

The reason why `mapToFilters` don't support these three types is because there is no `MultiRangeType` or related type traits defined, which has nothing to do with the fileformat. cc @rui-mo @jiangjiangtian @kecookier

rui-mo: That means actually we don't support the pushdown of long decimal in Parquet and ORC. Is that correct?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1600,17 +1600,26 @@ bool SubstraitToVeloxPlanConverter::childrenFunctionsOnSameField(
...
+  }
+
+  // The arg should be field or field with literal.
+  if (!fieldOrWithLiteral(scalarFunction.arguments(), fieldIdx)) {
+    return false;
   }
-  return canPushdown;
+
+  // Check whether data type is supported or not
+  if (!veloxTypeList.empty() && fieldIdx < veloxTypeList.size() && !isPushdownSupported(veloxTypeList.at(fieldIdx))) {
```
rui-mo, line 1618: Is it a bug if 'fieldIdx >= veloxTypeList.size()'? Shall we just add a check to ensure it does not happen?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
WangGuangxin: > That means actually we don't support the pushdown of long decimal in Parquet and ORC. Is that correct?

@rui-mo Currently yes. But to support a `HugeintMultiRange` in velox side is not a big issue, I can work on it later.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
WangGuangxin: ok

PR #5969, opened 2024-06-04
[VL] Fix build package script on GHA

zhouyuan:

This patch fixes the build package script by 1) adding the arrow jar 2) use the latest docker image 

## How was this patch tested?

local verified

### DIFF
--- a/.github/workflows/build_bundle_package.yml
+++ b/.github/workflows/build_bundle_package.yml
@@ -38,7 +38,7 @@ on:
 jobs:
   build-native-lib:
     runs-on: ubuntu-20.04
-    container: apache/gluten:gluten-vcpkg-builder_2024_03_17
+    container: apache/gluten:gluten-vcpkg-builder_2024_05_29
     steps:
       - uses: actions/checkout@v2
       - name: Build Gluten velox third party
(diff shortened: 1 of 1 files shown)


zhztheplayer (dismissed):

.github/workflows/build_bundle_package.yml
```diff
@@ -53,11 +53,17 @@ jobs:
...
+      - name: Upload native libs
+        uses: actions/upload-artifact@v2
         with:
           path: ./cpp/build/releases/
           name: velox-native-lib-${{github.sha}}
           retention-days: 1
+      - name: Upload Artifact Arrow Jar
+        uses: actions/upload-artifact@v2
+        with:
+          path: /root/.m2/repository/org/apache/arrow/
```
zhztheplayer, line 65: @jinchengchenghh Would you please confirm whether we could replace this line with 2 explicit gluten jars? E.g.,

```
path: |
    /root/.m2/repository/org/apache/arrow/.../arrow-dataset-<version>.jar
    /root/.m2/repository/org/apache/arrow/.../arrow-cdata-<version>.jar
```

jinchengchenghh: Maybe not, I tried before, but failed by cannot find jar. I don't remember the detailed error message.

.github/workflows/build_bundle_package.yml
```diff
@@ -38,7 +38,7 @@ on:
 jobs:
   build-native-lib:
     runs-on: ubuntu-20.04
-    container: apache/gluten:gluten-vcpkg-builder_2024_03_17
+    container: gluten:gluten-vcpkg-builder_2024_05_29
```
wangyum, line 41: It seems the container should be `apache/gluten:gluten-vcpkg-builder_2024_05_29`?
```
 /usr/bin/docker pull gluten:gluten-vcpkg-builder_2024_05_29
  Error response from daemon: pull access denied for gluten, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

wangyum (approved):

PHILO-HE (approved):

PR #5971, opened 2024-06-04
[CORE] ExpandFallbackPolicy should propagate fallback reason to vanilla SparkPlan

ulysses-you:

This pr makes ExpandFallbackPolicy propagate fallback reason to vanilla SparkPlan. The real fallback reason should not be overwritten by ExpandFallbackPolicy's fallback reason.

## How was this patch tested?

add test


ulysses-you:
cc @zhztheplayer @PHILO-HE thank you

zhztheplayer (approved):
Thanks!

PR #5976, opened 2024-06-04
[VL] Remove unnecessary arrow compile option

Yohahaha:
`-am` maven options is helpful, just compile needed module.


FelixYBW (dismissed):

Yohahaha:
convert status to draft, will update when I back from OOO.

Issue #5982, opened 2024-06-04
Issue while building gluten with velox and S3 enabled 


deepashreeraghu:
### Problem description

While I was trying to build gluten with velox as backend and enabling S3 flag, it fails as I already have minio package installed on the system. 
Since the package is already present the `rpm -i` results into non-zero exit code and build exits. 

### System information

```
Velox System Info v0.0.2
Commit: 7999b61a40dd57163f8fe5ade30087fcacc22bec
CMake Version: 3.20.2
System: Linux-4.18.0-477.27.1.el8_8.x86_64
Arch: x86_64
C++ Compiler: /opt/rh/gcc-toolset-9/root/usr/bin/c++
C++ Compiler Version: 9.2.1
C Compiler: /usr/lib64/ccache/cc
C Compiler Version: 9.2.1
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt
```

PR #5983, opened 2024-06-04
[GLUTEN-5982][VL] Allow replacing installed minio package in 1.1.1 branch

deepashreeraghu:

Allowing replace of minio package for branch 1.1.1

Fixes: https://github.com/apache/incubator-gluten/issues/5982

## How was this patch tested?

I ran a build with this fix and now it proceeds further and does not fail at rpm -i minio step.

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -146,6 +146,8 @@ function process_setup_centos8 {
   sed -i "s/yum -y install/sudo yum -y install/" ${VELOX_HOME}/scripts/setup-adapters.sh
   if [ $ENABLE_S3 == "ON" ]; then
     sed -i '/^  run_and_time install_fbthrift/a \ \ '${VELOX_HOME}/scripts'/setup-adapters.sh aws' scripts/setup-centos8.sh
+    # Replace minio if already installed.
+    sed -i 's/rpm -i minio-20220526054841.0.0.x86_64.rpm/rpm -i --replacepkgs minio-20220526054841.0.0.x86_64.rpm/g' scripts/setup-adapters.sh
   fi
   if [ $ENABLE_GCS == "ON" ]; then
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@deepashreeraghu, could you change the target to main branch? There will be no more release in 1.1 branch.

FelixYBW:
It's an environment issue. It's actually not only minio, but all the dependency libraries which causes version conflict.

So we shouldn't put it in get_velox.sh. Current recommendation is to build Gluten on a clean OS or docker. Otherwise there will be plenty of issues. So let's put the solution in issue.

@PHILO-HE

PHILO-HE:
> It's an environment issue. It's actually not only minio, but all the dependency libraries which causes version conflict.
> 
> So we shouldn't put it in get_velox.sh. Current recommendation is to build Gluten on a clean OS or docker. Otherwise there will be plenty of issues. So let's put the solution in issue.
> 
> @PHILO-HE

@FelixYBW, yes, all other libs installed similarly should also have this potential issue.

FelixYBW:
> 
> @FelixYBW, yes, all other libs installed similarly should also have this potential issue.

@deepashreeraghu can you put the error message and the solution in the issue? Looks it's not a good idea to solve the conflict in get_velox.sh. We should keep the script as simple as possible because we need to maintain it during daily velox rebase.

deepashreeraghu:
Thanks for your response. 
I did start with a clean image to build. But while building I hit into some failures and then I retried the build locally ( to find what was the build issue actually) That retry fails even before it reaches the actual failure because it has minio on it already. 
I did not face the same trouble with other packages. Hence requested this Pull request so that builds can go through fine when being retried.

deepashreeraghu:
Also, I am building on 1.1.1 because I wanted S3 support which is not available in the released jar. 
But, this branch does have a lot of other goodness in it and hence I opted for this branch and trying a dev build.

deepashreeraghu:
In the main branch the same fix is already present.

deepashreeraghu:
Since the fix will get into Velox as part of the PR - https://github.com/facebookincubator/velox/pull/10044. 
Closing this PR.

PR #5985, opened 2024-06-04
[VL] Handle try_subtract, try_multiply, try_divide

zhli1142015:

Follow up https://github.com/apache/incubator-gluten/pull/5356

## How was this patch tested?

UT.


zhli1142015:
cc @rui-mo and @PHILO-HE , thanks.

cpp/velox/operators/functions/RegistrationAllFunctions.cc
```diff
@@ -67,6 +68,10 @@ void registerFunctionOverwrite() {
       kRowConstructorWithAllNull,
       std::make_unique<RowConstructorWithNullCallToSpecialForm>(kRowConstructorWithAllNull));
   velox::functions::sparksql::registerBitwiseFunctions("spark_");
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedPlusFunction>({"check_plus"});
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedMinusFunction>({"check_minus"});
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedMultiplyFunction>({"check_multiply"});
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedDivideFunction>({"check_divide"});
```
PHILO-HE, line 74: Can we just register these functions with spark-aligned names: try_add, try_subtract, try_divide, try_multiply?

PHILO-HE (approved):
One minor comment. Thanks!

PHILO-HE (dismissed):
Looks good!

cpp/velox/operators/functions/RegistrationAllFunctions.cc
```diff
@@ -67,6 +68,10 @@ void registerFunctionOverwrite() {
       kRowConstructorWithAllNull,
       std::make_unique<RowConstructorWithNullCallToSpecialForm>(kRowConstructorWithAllNull));
   velox::functions::sparksql::registerBitwiseFunctions("spark_");
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedPlusFunction>({"try_add"});
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedMinusFunction>({"try_subtract"});
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedMultiplyFunction>({"try_multiply"});
+  velox::functions::registerBinaryIntegral<velox::functions::CheckedDivideFunction>({"try_divide"});
```
rui-mo, line 74: I notice they are registered in Velox Presto functions. Can we also register them in Velox Spark functions?

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -563,18 +563,68 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
         )
       case a: Add =>
-        BackendsApiManager.getSparkPlanExecApiInstance.genAddTransformer(
+        BackendsApiManager.getSparkPlanExecApiInstance.genArithmeticTransformer(
           substraitExprName,
           replaceWithExpressionTransformerInternal(a.left, attributeSeq, expressionsMap),
           replaceWithExpressionTransformerInternal(a.right, attributeSeq, expressionsMap),
-          a
+          a,
+          ExpressionNames.TRY_ADD
```
rui-mo, line 603: I wonder why we map Add to try_add. Is it to solve some semantic difference?

rui-mo (commented):
Thanks!

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
zhli1142015: After spark 3.4 `try_add` is converted to to `add(EvalMode.TRY)`, here may generate wrong results as we don't map it to correct velox functions. `genArithmeticTransformer` would map add to correct functions based on the `EvalMode`.
Thanks.
https://github.com/apache/incubator-gluten/pull/5356#issue-2234987326

https://github.com/apache/spark/blob/c4f720dfb41919dade7002b49462b3ff6b91eb22/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/TryEval.scala#L82C6-L82C78

cpp/velox/operators/functions/RegistrationAllFunctions.cc
rui-mo: Yes, if we can support them in Velox Spark functions, we could remove the custom registration in Gluten.

PHILO-HE: We can firstly merge this pr. Then remove the registering after velox pr is merged.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -123,42 +123,44 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
   }
 
-  override def genTryAddTransformer(
+  override def genTryArithmeticTransformer(
       substraitExprName: String,
       left: ExpressionTransformer,
       right: ExpressionTransformer,
-      original: TryEval): ExpressionTransformer = {
+      original: TryEval,
+      funcName: String): ExpressionTransformer = {
```
rui-mo, line 131: It is better to add a 'try_' prefix on substraitExprName if the try-function is needed to avoid introducing two functions names?
If we want to keep this implementation, could we rename funcName maybe like tryFunctionName and also document it to avoid misunderstanding? Thanks.

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
rui-mo: I understand. Thanks for the explain.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
zhli1142015: Updated, thanks.

cpp/velox/operators/functions/RegistrationAllFunctions.cc
zhli1142015: I think it's better to use names: `check_xx`. As we translate `try_xx` to `try(check_xx)`. if we use name try_xx, then they may be confusing.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -123,42 +123,50 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
+      Seq(GenericExpressionTransformer(checkArithmeticExprName, Seq(left, right), original)),
       original)
   }
 
-  override def genAddTransformer(
+  /**
+   * Map arithmetic expr to different functions: substraitExprName or try(checkArithmeticExprName)
+   * based on EvalMode.
+   */
+  override def genArithmeticTransformer(
```
rui-mo, line 151: Could you explain the bit the difference of `genTryArithmeticTransformer` and `genArithmeticTransformer`? I notice genTryArithmeticTransformer does not gen a try function but genArithmeticTransformer does.

rui-mo (commented):
Thanks! Added two questions. Please kindly correct me if I'm misunderstanding.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
zhli1142015: Yes, All the try_xx expressions are `RuntimeReplaceable `, they are converted to some other expressions, Before spark 3.4, they are converted to `TryEval(XX(failOnError = true))`, https://github.com/apache/spark/blob/branch-3.3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/TryEval.scala#L81, `genTryArithmeticTransformer ` is added to handle this case.

After spark 3.4,  for numeric types, they are converted to `XX(EvalMode.TRY)`, `genArithmeticTransformer` is added to handle this case, for velox we still convert it to `TryEval(XX(failOnError = true))`.

Both are converted to `try(check_xx)` in velox.

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -563,18 +563,68 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
       case tryEval @ TryEval(a: Add) =>
-        BackendsApiManager.getSparkPlanExecApiInstance.genTryAddTransformer(
+        BackendsApiManager.getSparkPlanExecApiInstance.genTryArithmeticTransformer(
           substraitExprName,
           replaceWithExpressionTransformerInternal(a.left, attributeSeq, expressionsMap),
           replaceWithExpressionTransformerInternal(a.right, attributeSeq, expressionsMap),
-          tryEval
+          tryEval,
+          ExpressionNames.CHECK_ADD
+        )
```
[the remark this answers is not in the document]
zhli1142015: Yes, here is `try`, I think it has same behavior in spark and velox: return null when error happens.
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/TryEval.scala#L40C1-L46C6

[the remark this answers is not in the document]
rui-mo: We notice failOnError is set true for EvalMode.TRY, so I'm a bit confused here.
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L201-L206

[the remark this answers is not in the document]
zhli1142015: Before spark 3.4 try_add is convert to `TryEval(Add(failOnError = true))`, after 3.4 it is converted to Add(EvalMode.TRY), and try-catch logic is executed in https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/arithmetic.scala#L365. I think both have same semantic: return null when error happens.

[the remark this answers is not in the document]
rui-mo: I see. Thanks for the explain!

rui-mo (approved):

Issue #5981, opened 2024-06-04
[CH] DIff result for get_json_object to get null value


exmy:
### Backend

CH (ClickHouse)

### Bug description

`select get_json_object('{"t": null}', '$.t') is null;`

spark result: true

gluten result: false

### Spark version

None

lgbo-ustc:
the result from `JSON_QUERY` in `CH` is a string value

lgbo-ustc:
I will fix it

PR #6001, opened 2024-06-06
[GLUTEN-5981][CH] Make the result be null when the queried field in `get_json_object` is `null`

lgbo-ustc:

Fixes: #5981

## How was this patch tested?

unit tests

### DIFF
--- a/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenClickhouseFunctionSuite.scala
+++ b/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenClickhouseFunctionSuite.scala
@@ -179,4 +179,15 @@ class GlutenClickhouseFunctionSuite extends GlutenClickHouseTPCHAbstractSuite {
         |""".stripMargin
     )(df => checkFallbackOperators(df, 0))
   }
+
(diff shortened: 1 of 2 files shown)


backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenClickhouseFunctionSuite.scala
```diff
@@ -66,6 +66,11 @@ class GlutenClickhouseFunctionSuite extends GlutenClickHouseTPCHAbstractSuite {
       // TODO: support default ANSI policy
       .set("spark.sql.storeAssignmentPolicy", "legacy")
       .set("spark.sql.warehouse.dir", warehouse)
+      .set(
```
zzcclp, line 69: set this config in the following ut case ? otherwise it can impact the other cases.

lgbo-ustc: This suite is for test functions, disable unfold constant functions should be OK

zzcclp (approved):
LGTM

PR #6004, opened 2024-06-06
[CORE] Move JoinSelectionOverrides logic to ColumnarOverrides

zml1206:

1. Drop custom strategy and move `JoinSelectionOverrides` logic to `ColumnarOverrides`
2. Clickhouse backend already support AQE, delete relevant code.


zml1206:
After this pr, buildSide of q3 changes, I looked at the AQE code, AQE is not in effect. After reOptimize,  only when the cost becomes smaller or the physical plan changes, then AQE will take effect, and then the sizeByte of the logical plan will be updated(https://github.com/apache/spark/blob/53d65fd12dd9231139188227ef9040d40d759021/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L383-L392).
So buildSide selection need in custom strategy.
I will implement creates "more vanilla" plan when the join operators are falling back in another way in the next PR. 
cc @zhztheplayer

zml1206:
> Aha this indeed looks to be an issue. Thank you for update.
> 
> > I will implement creates "more vanilla" plan when the join operators are falling back in another way in the next PR.
> 
> Do you have any detailed plan on this? Do we finally have to write a custom strategy?


I think writing a custom strategy is necessary. My initial idea is to generate a new ShuffledHashJoinExecBase(sparkJoinExec, buildSide) in the custom strategy.

zhztheplayer:
Do you think we can append the required changes to this patch? It sounds a little bit risky to fail some plan checks then wait for another PR to land.

> I think writing a custom strategy is necessary. My initial idea is to generate a new ShuffledHashJoinExecTemp(sparkJoinExec, buildSide) in the custom strategy, transform ShuffledHashJoinExecTemp in ColumnarOverrides.

OK. And you can compare the overall code complexity with current main code. If we can finally simply the code although a custom strategy is still needed, then it sounds worth it to do the refactor.

zml1206:
> Do you think we can append the required changes to this patch? It sounds a little bit risky to fail some plan checks then wait for another PR to land.

I will close this PR, new another PR.

zml1206:
Write a custom strategy

PR #6008, opened 2024-06-07
[VL] Add gluten iceberg jar to bundle package

leoluan2009:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/docs/get-started/Velox.md
+++ b/docs/get-started/Velox.md
@@ -303,8 +303,7 @@ First of all, compile gluten-iceberg module by a `iceberg` profile, as follows:
 mvn clean package -Pbackends-velox -Pspark-3.3 -Piceberg -DskipTests
 ```
 
-Then, put the additional `gluten-iceberg-XX-SNAPSHOT.jar` to the class path (usually it's `$SPARK_HOME/jars`).
-The gluten-iceberg jar is in `gluten-iceberg/target` directory.
+Once building successfully, the fat Jar file will be generated in the directory: package/target/\<gluten-jar\> for Spark 3.2.x/Spark 3.3.x/Spark 3.4.x/Spark 3.5.x, put it to the class path (usually it's `$SPARK_HOME/jars`).
 
 After the two steps, you can query iceberg table by gluten/velox without scan's fallback.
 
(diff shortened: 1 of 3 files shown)


zhouyuan:
CC @yma11

yma11:
@leoluan2009 Thanks for contribution. For data lake support like iceberg and delta lake, we don't include them in Gluten jar as it's not used by all customers and will make the package fatter. Any reasons for your changes here?

leoluan2009:
@yma11 Iceberg profile is not active by default, if we manually enable it when building gluten, it means we need it to read/write iceberg table, so we should put it in Gluten jar for easy deployment？

yma11:
> @yma11 Iceberg profile is not active by default, if we manually enable it when building gluten, it means we need it to read/write iceberg table, so we should put it in Gluten jar for easy deployment？

I get your point. In this way, we don't need extra include of the gluten-iceberg-* jar file. So I think you need update this [doc](https://github.com/apache/incubator-gluten/blob/main/docs/get-started/Velox.md#iceberg-support) as well.

leoluan2009:
@yma11  updated doc, can you help to review again? thanks

docs/get-started/Velox.md
```diff
@@ -303,8 +303,7 @@ First of all, compile gluten-iceberg module by a `iceberg` profile, as follows:
 mvn clean package -Pbackends-velox -Pspark-3.3 -Piceberg -DskipTests
 ```
 
-Then, put the additional `gluten-iceberg-XX-SNAPSHOT.jar` to the class path (usually it's `$SPARK_HOME/jars`).
-The gluten-iceberg jar is in `gluten-iceberg/target` directory.
+Once building successfully, the fat Jar file will be generated in the directory: package/target/\<gluten-jar\> for Spark 3.2.x/Spark 3.3.x/Spark 3.4.x/Spark 3.5.x, put it to the class path (usually it's `$SPARK_HOME/jars`).
```
yma11, line 306: "Once built successfully, iceberg features will be included in `gluten-velox-bundle-X` jar. Then you can query iceberg table by gluten/velox without scan's fallback. "

leoluan2009:
@yma11 Thanks for your review, I have updated doc by your comment

yma11 (approved):

PR #6009, opened 2024-06-07
[VL] Provide options to combine small batches before sending to shuffle

zhztheplayer:
It's observed that Velox hash-based shuffle is slowed down by small input batches.

The patch:

1. Adds two options:
   - `spark.gluten.sql.columnar.backend.velox.coalesceBatchesBeforeShuffle`
      (Default: false) Set to true to combine small batches with minimal batch size determined by `spark.gluten.sql.columnar.maxBatchSize`. (Note the misnaming of `maxBatchSize` in Gluten, it might tend to be `minBatchSize` or just `batchSize`)
   - `spark.gluten.sql.columnar.backend.velox.minBatchSizeForShuffle`
      (Optional) Set to override the minimal batch size used by `coalesceBatchesBeforeShuffle`.
2. Does essential code refactors and cleanups. 

### Comparisons

(spark.gluten.sql.columnar.backend.velox.coalesceBatchesBeforeShuffle=false/true)

Q31 total time, before and after (SF1000 partitioned table, scan partitions 112, shuffle partitions 112):

image

Closer look at exchange, before and after:

image

image

### DIFF
--- a/cpp/core/jni/JniCommon.cc
+++ b/cpp/core/jni/JniCommon.cc
@@ -65,3 +65,61 @@ gluten::Runtime* gluten::getRuntime(JNIEnv* env, jobject runtimeAware) {
   GLUTEN_CHECK(ctx != nullptr, "FATAL: resource instance should not be null.");
   return ctx;
 }
+
+std::unique_ptr<gluten::JniColumnarBatchIterator> gluten::makeJniColumnarBatchIterator(
+    JNIEnv* env,
+    jobject jColumnarBatchItr,
+    gluten::Runtime* runtime,
+    std::shared_ptr<ArrowWriter> writer) {
+  return std::make_unique<JniColumnarBatchIterator>(env, jColumnarBatchItr, runtime, writer);
+}
+
+gluten::JniColumnarBatchIterator::JniColumnarBatchIterator(
+    JNIEnv* env,
+    jobject jColumnarBatchItr,
+    gluten::Runtime* runtime,
+    std::shared_ptr<ArrowWriter> writer)
+    : runtime_(runtime), writer_(writer) {
+  // IMPORTANT: DO NOT USE LOCAL REF IN DIFFERENT THREAD
+  if (env->GetJavaVM(&vm_) != JNI_OK) {
+    std::string errorMessage = "Unable to get JavaVM instance";
+    throw gluten::GlutenException(errorMessage);
+  }
+  serializedColumnarBatchIteratorClass_ =
+      createGlobalClassReferenceOrError(env, "Lorg/apache/gluten/vectorized/ColumnarBatchInIterator;");
+  serializedColumnarBatchIteratorHasNext_ =
+      getMethodIdOrError(env, serializedColumnarBatchIteratorClass_, "hasNext", "()Z");
+  serializedColumnarBatchIteratorNext_ = getMethodIdOrError(env, serializedColumnarBatchIteratorClass_, "next", "()J");
+  jColumnarBatchItr_ = env->NewGlobalRef(jColumnarBatchItr);
+}
+
+gluten::JniColumnarBatchIterator::~JniColumnarBatchIterator() {
+  JNIEnv* env;
+  attachCurrentThreadAsDaemonOrThrow(vm_, &env);
+  env->DeleteGlobalRef(jColumnarBatchItr_);
+  env->DeleteGlobalRef(serializedColumnarBatchIteratorClass_);
+  vm_->DetachCurrentThread();
+}
+
+std::shared_ptr<gluten::ColumnarBatch> gluten::JniColumnarBatchIterator::next() {
+  JNIEnv* env;
+  attachCurrentThreadAsDaemonOrThrow(vm_, &env);
+  if (!env->CallBooleanMethod(jColumnarBatchItr_, serializedColumnarBatchIteratorHasNext_)) {
+    checkException(env);
+    return nullptr; // stream ended
+  }
+
+  checkException(env);
+  jlong handle = env->CallLongMethod(jColumnarBatchItr_, serializedColumnarBatchIteratorNext_);
+  checkException(env);
+  auto batch = runtime_->objectStore()->retrieve<ColumnarBatch>(handle);
+  if (writer_ != nullptr) {
+    // save snapshot of the batch to file
+    std::shared_ptr<ArrowSchema> schema = batch->exportArrowSchema();
+    std::shared_ptr<ArrowArray> array = batch->exportArrowArray();
+    auto rb = gluten::arrowGetOrThrow(arrow::ImportRecordBatch(array.get(), schema.get()));
+    GLUTEN_THROW_NOT_OK(writer_->initWriter(*(rb->schema().get())));
+    GLUTEN_THROW_NOT_OK(writer_->writeInBatches(rb));
+  }
+  return batch;
+}
--- a/cpp/core/jni/JniCommon.h
+++ b/cpp/core/jni/JniCommon.h
@@ -28,6 +28,7 @@
 #include "memory/AllocationListener.h"
 #include "shuffle/rss/RssClient.h"
 #include "utils/Compression.h"
+#include "utils/ResourceMap.h"
 #include "utils/exception.h"
 
 static jint jniVersion = JNI_VERSION_1_8;
@@ -119,6 +120,12 @@ static inline void attachCurrentThreadAsDaemonOrThrow(JavaVM* vm, JNIEnv** out) 
   }
 }
 
+template <typename T>
+static T* jniCastOrThrow(gluten::ResourceHandle handle) {
+  auto instance = reinterpret_cast<T*>(handle);
+  GLUTEN_CHECK(instance != nullptr, "FATAL: resource instance should not be null.");
+  return instance;
+}
 namespace gluten {
 
 class JniCommonState {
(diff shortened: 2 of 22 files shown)


zhztheplayer:
@marin-ma There might be some batch-wise overhead around shuffle split processing. We may want to figure it out later to avoid doing such batch coalesce operations that introduce extra copies.

marin-ma:
cc: @WangGuangxin @FelixYBW

Yohahaha:
https://github.com/apache/incubator-gluten/pull/5951#issuecomment-2153866781
I see above comments try add a new operator to collect small batch to a bigger batch, is will be included in this PR?

zhztheplayer:
@XinShuoWang 

Do you have any thoughts on moving to this approach since https://github.com/apache/incubator-gluten/commit/ec3e92ec841ef71490c8c41c64647c60888e3885 has been merged? I don't have strong preference except that we need a configuration and metrics for batch-appending.

I suggest merging this to make things configurable and displayable at first, then if we want to continue on https://github.com/apache/incubator-gluten/pull/5951 's approach, you can open another PR to bring the code back and re-use the conf code added in this patch. And remove calls to `maybeAddAppendBatchesExec` to disable AppendBatchesExec for shuffle.

zhztheplayer:
> [#5951 (comment)](https://github.com/apache/incubator-gluten/pull/5951#issuecomment-2153866781) I see above comments try add a new operator to collect small batch to a bigger batch, is will be included in this PR?

It's included, namely `VeloxAppendBatchesExec`.

Code https://github.com/apache/incubator-gluten/pull/6009/files#diff-e058841b0f54a15144e01a1089457b7e5a28a953f85df3a1e656fcf6eddfe203R38

cpp/core/jni/JniCommon.cc
```diff
@@ -65,3 +65,61 @@ gluten::Runtime* gluten::getRuntime(JNIEnv* env, jobject runtimeAware) {
...
+  if (writer_ != nullptr) {
+    // save snapshot of the batch to file
+    std::shared_ptr<ArrowSchema> schema = batch->exportArrowSchema();
+    std::shared_ptr<ArrowArray> array = batch->exportArrowArray();
+    auto rb = gluten::arrowGetOrThrow(arrow::ImportRecordBatch(array.get(), schema.get()));
+    GLUTEN_THROW_NOT_OK(writer_->initWriter(*(rb->schema().get())));
+    GLUTEN_THROW_NOT_OK(writer_->writeInBatches(rb));
+  }
+  return batch;
+}
```
zhztheplayer, line 125: code movement

cpp/core/jni/JniCommon.h
```diff
@@ -251,6 +258,40 @@ DEFINE_SAFE_GET_PRIMITIVE_ARRAY_FUNCTIONS(kLong, jlongArray, Long)
...
+  jclass serializedColumnarBatchIteratorClass_;
+  jmethodID serializedColumnarBatchIteratorHasNext_;
+  jmethodID serializedColumnarBatchIteratorNext_;
+};
+
+std::unique_ptr<JniColumnarBatchIterator> makeJniColumnarBatchIterator(
+    JNIEnv* env,
+    jobject jColumnarBatchItr,
+    Runtime* runtime,
+    std::shared_ptr<ArrowWriter> writer);
```
zhztheplayer, line 294: code movement

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -101,23 +101,17 @@ case class OffloadAggregate() extends OffloadSingleNode with LogLevelUtil {
...
+      p
+    case s: ShuffleExchangeExec
+        if (s.child.supportsColumnar || GlutenConfig.getConf.enablePreferColumnar) &&
+          BackendsApiManager.getSettings.supportColumnarShuffleExec() =>
+      logDebug(s"Columnar Processing for ${s.getClass} is currently supported.")
+      BackendsApiManager.getSparkPlanExecApiInstance.genColumnarShuffleExchange(s)
+    case b: BroadcastExchangeExec =>
+      val child = b.child
+      logDebug(s"Columnar Processing for ${b.getClass} is currently supported.")
+      ColumnarBroadcastExchangeExec(b.mode, child)
```
zhztheplayer, line 114: code simplification

gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -101,7 +101,7 @@ trait SparkPlanExecApi {
       aggregateExpressions: Seq[AggregateExpression],
       aggregateAttributes: Seq[Attribute]): HashAggregateExecPullOutBaseHelper
 
-  def genColumnarShuffleExchange(shuffle: ShuffleExchangeExec, newChild: SparkPlan): SparkPlan
+  def genColumnarShuffleExchange(shuffle: ShuffleExchangeExec): SparkPlan
```
zhztheplayer, line 104: API simplification

Yohahaha:
I think the design is same as https://github.com/facebookincubator/velox/issues/7801 and would be great to implement in Gluten side.

my suggestion is we could implement `VeloxAppendBatchesExec` first, then provide config for shuffle write and table scan to add this new operator.

zhli1142015:
> I think the design is same as [facebookincubator/velox#7801](https://github.com/facebookincubator/velox/issues/7801) and would be great to implement in Gluten side.
> 
> my suggestion is we could implement `VeloxAppendBatchesExec` first, then provide config for shuffle write and table scan to add this new operator.

I think filter and HashJoin(with filter) may also need to this as well. Thanks.

Yohahaha:
> > I think the design is same as [facebookincubator/velox#7801](https://github.com/facebookincubator/velox/issues/7801) and would be great to implement in Gluten side.
> > my suggestion is we could implement `VeloxAppendBatchesExec` first, then provide config for shuffle write and table scan to add this new operator.
> 
> I think filter and HashJoin(with filter) may also need this as well. Thanks.

yeah, above velox issue has list out all operators that may need apply this optimization, we could provide a config for these operators, which value is a list of operator name string.

FelixYBW:
> `spark.gluten.sql.columnar.maxBatchSize`. (Note the misnaming of `maxBatchSize` in Gluten, it might tend to be `minBatchSize`)

Just comment this: `minBatchSize` is either not accurate. The accurate description of the batch size is "Velox will try best to limit the row numbers per rowVector to the maxBatchSize config, however, it's not guaranteed.  @zhztheplayer can you update the document to highlight this?

FelixYBW:
@zhztheplayer  Let's remove the option and set it as default behavior as long as it can benefit in all cases.

FelixYBW:
> around shuffle split processing. We may want to figure it out later to avoid doing such batch coalesce operations that intro

It's because the initialization of current split function. Currently we use 3 loops (per column, per reducer, per row) to do the split, if the column data is cached then the solution is the best way to scale to reducer numbers. However to achieve this, we need much initialization work to create several vectors. If the input batch is small, we will suffer from the initialization overhead. Even bigger than the copy to bigger batches.

Another issue is if the data size is too large and exceeds the cache size, then performance will be very poor.

zhztheplayer:
> Just comment this: `minBatchSize` is either not accurate. The accurate description of the batch size is "Velox will try best to limit the row numbers per rowVector to the maxBatchSize config, however, it's not guaranteed. @zhztheplayer can you update the document to highlight this?

It's planned but I'll use another PR to do that. Setting it true by default may fail some plan checks in UTs.

FelixYBW:
> It's planned but I'll use another PR to do that. Setting it true by default may fail some plan checks in UTs.

Why UT fails? we should fix that.

zhztheplayer:
> row numbers per rowVector to the maxBatchSize config, however, it's not guaranteed. @zhztheplayer can you update the document to highlight this?

The PR adds a new operator `VeloxAppendBatchesExec`. Plan checks may fail because of this new operator. It's benign and to fix this we just update the UT code.

FelixYBW:
> checks may fail because of this new operator. It's benign and to fix this we just update the UT code.

Let's update the UT code then.

zhztheplayer:
> Just comment this: `minBatchSize` is either not accurate. The accurate description of the batch size is "Velox will try best to limit the row numbers per rowVector to the maxBatchSize config, however, it's not guaranteed. @zhztheplayer can you update the document to highlight this?

The tricky part is that I don't see we always follow either `min` or `max` criteria when using the option. (correct me if I am wrong) It's used as `min` in shuffle reader, but may be used as `max` in shuffle writer. I don't go through scan's code so not sure about that part.

Maybe we can change the option name to `targetBatchSize` to clarify. Actually it's not that important whether a batch is slightly larger than or smaller than this size.

FelixYBW:
One misunderstanding is that "we should avoid memcpy as much as possible", but in fact Gluten isn't memory throughput bound yet. The sequential data read and write is a cheaper operation if the block size isn't too small (like several Bytes but not predictable, then the overhead is branch misprediction) or too large (like GB level).

FelixYBW:
> e we can change the option name

In long term Velox should limit the batch size to maxBatchSize we configured, so as in Gluten. For operators like Combine, we may limit the batch size to 2 x maxBatchSize so we needn't to cache the second batch.

FelixYBW:
> 
> It's included, namely `VeloxAppendBatchesExec`.
> 
> Code https://github.com/apache/incubator-gluten/pull/6009/files#diff-e058841b0f54a15144e01a1089457b7e5a28a953f85df3a1e656fcf6eddfe203R38

We may need to submit this Exec to Velox

marin-ma (approved):
LGTM. Thanks!

zhztheplayer:
@XinShuoWang I'll merge this then do some follow-ups. Let me know if any comments, thanks.

Also thank you guys for reviewing. :)

zhztheplayer:
> /Benchmark Velox TPCDS

The PR disabled the feature by default. Let's benchmark the subsequent PRs.

PR #6029, opened 2024-06-09
[VL] Pass file size and modification time in split 

acvictor:

Support was made in Velox to pass file properties while creating a file handle for reading. Previously a separate call was made to fetch file properties from remote storage. By passing this information down to Velox in the split we save on one network call per file open.

https://github.com/facebookincubator/velox/issues/9314

## How was this patch tested?

Existing UTs


acvictor:
@FelixYBW can you please review this? Thank you!

shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
```diff
@@ -207,6 +207,11 @@ trait SparkShims {
 
   def attributesFromStruct(structType: StructType): Seq[Attribute]
 
+  // Spark 3.3 and later only have file size and modification time in PartitionedFile
+  def getFileSizeAndModificationTime(file: PartitionedFile): (Option[Long], Option[Long]) = {
+    (None, None)
+  }
+
```
Yohahaha, line 214: we'd better keep pure abstract method in this file, just like other declaration.

Yohahaha (dismissed):
LGTM, leave one comments.

shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
acvictor: Updated, thank you!

acvictor:
> LGTM, leave one comments.

Thank you for the review!

Yohahaha (approved):
thank you!

FelixYBW:
The PR broken the S3 reader

Issue #6040, opened 2024-06-11
[CH] Can't not load part after restart spark session


loneylee:
### Backend

CH (ClickHouse)

### Bug description

Error message as follow
<Error> s3_mergetree.part: Part /tmp/ch_metadata2/tpch10/mergetree/part/00006/556fee1c-2a92-48c9-9aeb-fbc3a8e4f9cf_0_001/ is broken and need manual correction

### Spark version

None

PR #6041, opened 2024-06-11
[GLUTEN-6040][CH] Fix can't not load part after restart spark session

loneylee:

Fix can't not load part after restart spark session after enable enabled_driver_filter_mergetree_index

(Fixes: \#6040)

## How was this patch tested?

Test by ut


zzcclp (approved):
LGTM

PR #6044, opened 2024-06-11
[VL] Change VLOG to DLOG in shuffle to fix performance issue in corner cases

marin-ma:
Even if FLAGS_v is set to a small value, the VLOG with large verbosity is still initialized and taking lots of time.
image

After:
image


marin-ma:
Just found this issue only happens when I set `spark.gluten.sql.debug=true` and `spark.gluten.sql.columnar.backend.velox.glogSeverityLevel=1`. Setting `spark.gluten.sql.columnar.backend.velox.glogSeverityLevel` > 0 (FLAGS_minloglevel) will mute all VLOG, but the logged strings are still constructed. This behavior of `VLOG` seems to be very tricky.

@zhztheplayer Can we merge this patch? Those logs are for debugging purpose.

zhztheplayer (approved):

PR #6046, opened 2024-06-11
[CH] add throttler to GlutenHDFSDisk

liuneng1994:

Add a write speed limit to HDFS to avoid TCP port exhaustion due to too many small files and too fast write rate.

## How was this patch tested?

unit tests， manual test


zzcclp (approved):
manually test passed, LGTM

PR #6054, opened 2024-06-12
[GLUTEN-5720][VL][FOLLOWUP][MIRROR] Fix invalid adding int to string

jackylee-ch:
Fix the invalid adding int to string in log.

## How was this patch tested?
GA

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -925,7 +925,7 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::JoinRel& joinRel
     case ::substrait::JoinRel_JoinType_JOIN_TYPE_ANTI:
       break;
     default:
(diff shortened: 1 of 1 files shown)


cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -925,7 +925,7 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::JoinRel& joinRel
     case ::substrait::JoinRel_JoinType_JOIN_TYPE_ANTI:
       break;
     default:
-      LOG_VALIDATION_MSG("Join type is not supported: {}" + joinRel.type());
+      LOG_VALIDATION_MSG("Join type is not supported: {}" + std::to_string(joinRel.type()));
```
zhztheplayer, line 928: Was "{}" there by mistake?

```suggestion
      LOG_VALIDATION_MSG("Join type is not supported: " + std::to_string(joinRel.type()));
```

jackylee-ch: Good catch. Fixed.

PR #6056, opened 2024-06-12
[VL] Set s.g.s.c.b.v.coalesceBatchesBeforeShuffle=true by default

zhztheplayer:
Related to https://github.com/apache/incubator-gluten/pull/6009

Set spark.gluten.sql.columnar.backend.velox.coalesceBatchesBeforeShuffle=true to make sure shuffle's performance is always not impacted by the issue.

The default batch size is changed to `0.8 * GLUTEN_MAX_BATCH_SIZE` to add some flexibilities to avoid unexpected combinations from buffers that are already large enough.


FelixYBW:
@zhztheplayer @marin-ma Similarly if the batch size is too large, each column exceeds the L2 cache size, the performance should be very bad as well. Can you submit a similar PR to fix this? Split the large batchs into the small one during Split.

marin-ma:
> @zhztheplayer @marin-ma Similarly if the batch size is too large, each column exceeds the L2 cache size, the performance should be very bad as well. Can you submit a similar PR to fix this? Split the large batchs into the small one during Split.

@FelixYBW Do we need a new Operator for it? If it's only for split, we already have it https://github.com/apache/incubator-gluten/pull/5536

zhztheplayer:
Perhaps we can tweak the code of VeloxAppendBatches including renaming it to make it be able to do both appending and slicing by giving it a target batch size range. So the happening of these two operations can be made more consistent.

zzcclp:
Run Gluten Clickhouse CI

marin-ma (approved):
LGTM. Thanks!

PR #6061, opened 2024-06-12
[VL][BUILD] Improve compilation speed for Arrow

jackylee-ch:
Most of arrow modules are not needed in gluten, thus we can remove them in mvn install. Besides, we don't need the `arrow-with-dependencies.jar`, thus we remove it too.

## How was this patch tested?
GA


jackylee-ch:
cc @jinchengchenghh

Yohahaha (approved):
thank you!

PR #6079, opened 2024-06-13
[CH][UT] Fix UT due to https://github.com/ClickHouse/ClickHouse/pull/64427

baibaichen:

Don't know the reason, but this [PR](https://github.com/ClickHouse/ClickHouse/pull/64427) changed the `input_format_parquet_max_block_size` to `DEFAULT_block_size`, which is 65409, causing the MergeTree related tests to fail.  Let's set `input_format_parquet_max_block_size` to 8192, which reverts 7445f02c9414c98956d1ad4d9318f6e545e1740c (this fix looks unstable)

Also I use `assertResult` instead of `assert`, so we can know the actual result once failed.

## How was this patch tested?

Existed UTs


zzcclp (approved):
LGTM

PR #6081, opened 2024-06-13
[VL] Fix undefined symbol with qat

marin-ma:
Below linking error is found with latest QAT driver. Use "-Wl,--no-as-needed" to force link libudev.so
image


weiting-chen (approved):
looks good to me.

PR #682, opened 2022-12-07
[OPPRO-354] Fallback to vanilla spark when exist multi joins

JkSelf:
## What changes were proposed in this pull request?
Fallback TPC-DS Q72 to vanilla spark when smj optimization not available.

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #6093, opened 2024-06-14
[CORE] Creates vanilla plan when the join operators fall back

zml1206:
1. drop custom strategy for create vanilla plan when the join operators fall back
2. clickhouse backend already support AQE, delete relevant code.


gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
```diff
@@ -416,16 +387,7 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
           val transformer = ColumnarShuffleExchangeExec(plan, plan.child, plan.child.output)
           transformer.doValidate().tagOnFallback(plan)
         case plan: ShuffledHashJoinExec =>
-          val transformer = BackendsApiManager.getSparkPlanExecApiInstance
-            .genShuffledHashJoinExecTransformer(
-              plan.leftKeys,
-              plan.rightKeys,
-              plan.joinType,
-              TransformHints.getShuffleHashJoinBuildSide(plan),
```
ulysses-you, line 424: It seems we have already chosen a small table before. The only thing different with before is, we convert smj to shj from strategy to columnar, so is GlutenCostEvaluator required ? I think the cost evaluator is useful when we add tag in strategy.

zml1206: The size obtained here may not be accurate. If `currentPhysicalPlan == newPhysicalPlan`, logicalLink will not be updated. GlutenCostEvaluator is to update logicalplan when `currentPhysicalPlan == newPhysicalPlan`, and also update sizeByte.

backends-velox/src/test/resources/tpch-approved-plan/v1-ras/spark32/3.txt
```diff
@@ -1,56 +1,64 @@
...
-                                 +- ^ NoopFilter (25)
-                                    +- ^ Scan parquet (24)
+   VeloxColumnarToRowExec (43)
+   +- TakeOrderedAndProjectExecTransformer (42)
+      +- ^ ProjectExecTransformer (40)
+         +- ^ RegularHashAggregateExecTransformer (39)
+            +- ^ RegularHashAggregateExecTransformer (38)
+               +- ^ ProjectExecTransformer (37)
+                  +- ^ ShuffledHashJoinExecTransformer Inner BuildLeft (36)
+                     :- ^ SortExecTransformer (26)
```
zml1206, line 11: Can't drop `SortExecTransformer` when RAS enabled is a known issue, #6011 @zhztheplayer

zml1206:
@ulysses-you @zhztheplayer If you have time, can you help to take a look again? Thank you.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
```diff
@@ -0,0 +1,46 @@
...
+case class GlutenCostEvaluator() extends CostEvaluator {
+  private var isReOptimize = false
+  override def evaluateCost(plan: SparkPlan): Cost = {
+    val simpleCost =
+      SimpleCostEvaluator.evaluateCost(plan).asInstanceOf[SimpleCost]
+    var cost = simpleCost.value * 2
+    if (isReOptimize && existsShuffledJoin(plan)) {
+      cost -= 1
+    }
+    isReOptimize = !isReOptimize
```
zhztheplayer, line 35: I could understand how this statement works but it's like something too caller-sensitive to rely on.

I see `Cost` is a trait that can be extended. Could we store the plan's id which is incremental by instantiation order in a customized Cost structure, then use the ID during cost comparison to find out the one that was created later?

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -143,6 +143,11 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
     }
     conf.set(SPARK_SESSION_EXTS_KEY, extensions)
 
+    // sql adaptive customCostEvaluatorClass
+    conf.set(
+      SQLConf.ADAPTIVE_CUSTOM_COST_EVALUATOR_CLASS.key,
+      "org.apache.spark.sql.execution.adaptive.GlutenCostEvaluator")
+
```
zhztheplayer, line 150: Let's have a check to raise an error when there was already another cost evaluator set by user.

shims/spark35/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
```diff
@@ -0,0 +1,41 @@
...
+ * contains [[ShuffledJoin]] and cost is the same.
+ */
+case class GlutenCostEvaluator() extends CostEvaluator with SQLConfHelper {
+  private var isReOptimize = false
+  override def evaluateCost(plan: SparkPlan): Cost = {
+    val forceOptimizeSkewedJoin = conf.getConf(SQLConf.ADAPTIVE_FORCE_OPTIMIZE_SKEWED_JOIN)
+    val simpleCost =
+      SimpleCostEvaluator(forceOptimizeSkewedJoin).evaluateCost(plan).asInstanceOf[SimpleCost]
+    var cost = simpleCost.value * 2
+    if (isReOptimize && plan.exists(_.isInstanceOf[ShuffledJoin])) {
```
ulysses-you, line 35: We should only change the plan when there is an actual better build side.

zml1206: I have considered it before, but doing this requires a lot of cost. First, need to add extra rule to add tags. Secondly, when the plan contains multiple joins, it is not easy to judge which join is about to be executed. I just modified cost equal. If the costs are equal, it is forced to use new physics plan, which doesn't seem to hurt.

zhztheplayer: > If the costs are equal, it is forced to use new physics plan

Would you like to confirm why the newer plan is always better than the older one? E.g., The newer one carries the latest up-to-date statistics so the hash-to-sort conversion in columnar rule could be more reliable?

If we can make sure the newer is better, then it sounds reasonable to use the newer one.

zhztheplayer: Additionally, I assume `reoptimizePlan == plan` happens frequently speaking of AQE execution. If so we should measure on the planer performance too.

zml1206: The default condition for spark to use the new physics plan is `newCost < origCost || (newCost == origCost && currentPhysicalPlan != newPhysicalPlan)` , this pr just adds a situation, `newCost == origCost && currentPhysicalPlan == newPhysicalPlan`, so I think the performance of newPhysicalPlan has not changed.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
zml1206: The planId is increasing, but it may be removed by the parent node, causing the id to become smaller, such as `OptimizeOneRowPlan`,
I'm going to do it this way, what do you think?
```
case class GlutenCost(value: Long, planId: Int) extends Cost {
  override def compare(that: Cost): Int = that match {
    case GlutenCost(thatValue, thatId) =>
      if (value < thatValue || (value == thatValue && planId != thatId)) -1
      else if (value > thatValue) 1
      else 0
    case _ =>
      throw QueryExecutionErrors.cannotCompareCostWithTargetCostError(that.toString)
  }
}
```

shims/spark32/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.apache.spark.sql.execution.adaptive
+
+import org.apache.spark.sql.errors.QueryExecutionErrors
+import org.apache.spark.sql.execution.SparkPlan
+
+case class GlutenCost(value: Long, planId: Int) extends Cost {
+  override def compare(that: Cost): Int = that match {
+    case GlutenCost(thatValue, thatId) =>
+      if (value < thatValue || (value == thatValue && planId != thatId)) -1
```
zml1206, line 25: The planId is increasing, but the parent node may be removed, causing the id to become smaller, such as `OptimizeOneRowPlan`, so use `!=`.

gluten-core/src/main/scala/org/apache/gluten/execution/JoinExecTransformer.scala
```diff
@@ -370,6 +371,66 @@ abstract class ShuffledHashJoinExecTransformerBase(
...
+
+  def from(smj: SortMergeJoinExec): ShuffledHashJoinExecTransformerBase = {
+    BackendsApiManager.getSparkPlanExecApiInstance
+      .genShuffledHashJoinExecTransformer(
+        smj.leftKeys,
+        smj.rightKeys,
+        smj.joinType,
+        getBuildSide(smj),
+        smj.condition,
+        dropPartialSort(smj.left),
```
ulysses-you, line 427: Strip local sort is a bit conflict with pull out project, there is a pre-project leak if we pull out expr from sort first but the sort is stripped later. Before the smj can never happen since we plan join at strategy side.

shims/spark35/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/** This [[CostEvaluator]] is to force use the new physical plan when cost is equal. */
+case class GlutenCostEvaluator() extends CostEvaluator with SQLConfHelper {
+  override def evaluateCost(plan: SparkPlan): Cost = {
+    val forceOptimizeSkewedJoin = conf.getConf(SQLConf.ADAPTIVE_FORCE_OPTIMIZE_SKEWED_JOIN)
+    val simpleCost = SimpleCostEvaluator(forceOptimizeSkewedJoin)
+      .evaluateCost(plan)
+      .asInstanceOf[SimpleCost]
+    GlutenCost(simpleCost.value, plan.id)
```
ulysses-you, line 42: Can we add one more config to crontrol this feature ? if disable then do nothing.

gluten-core/src/main/scala/org/apache/gluten/execution/JoinExecTransformer.scala
zml1206: I don’t understand. Which rule is it?

shims/spark35/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
zml1206: How about add GlutenConfig `spark.gluten.sql.execution.adaptive.costEvaluator.enabled`, if true, use `GlutenCostEvaluator`, else follow the configuration `org.apache.spark.sql.execution.adaptive.GlutenCostEvaluator`?

gluten-core/src/main/scala/org/apache/gluten/execution/JoinExecTransformer.scala
ulysses-you: see PullOutPreProject

shims/spark35/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
ulysses-you: Sounds good, besides, we'd better to respect spark.gluten.enabled even costEvaluator is enabled. People may tune spark.gluten.enabled off at runtime.

zhztheplayer:
@zml1206 Would you like to rebase onto newest main code since https://github.com/apache/incubator-gluten/pull/6107 was merged? Thank you very much.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
zhztheplayer: This part is comparatively tricky and important in this PR so let's code it carefully. `planId != thatId` is still a condition relying on caller code.

Would you check if we can use something like this:

```scala
class GlutenCost(val eval: CostEvaluator, val plan: SparkPlan) extends Cost {
  override def compare(that: Cost): Int = that match {
    case that: GlutenCost if plan eq that.plan =>
      0
    case that: GlutenCost if plan == that.plan =>
      // Plans are identical. Considers the newer one as having lower cost.
      -(plan.id - that.plan.id)
    case that: GlutenCost =>
      // Plans are different. Use the delegated cost evaluator.
      assert(eval == that.eval)
      eval.evaluateCost(plan).compare(eval.evaluateCost(that.plan))
    case _ =>
      throw QueryExecutionErrors.cannotCompareCostWithTargetCostError(that.toString)
  }

  override def hashCode(): Int = throw new UnsupportedOperationException()

  override def equals(obj: Any): Boolean = obj match {
    case that: Cost => compare(that) == 0
    case _ => false
  }
}
```

zml1206: > ```scala
>       eval.evaluateCost(plan).compare(eval.evaluateCost(that.plan))
> ```


There seems to be a codependency issue here. 
How about this：

```
class GlutenCost(value: Long, plan: SparkPlan) extends SimpleCost(value) {
  override def compare(that: Cost): Int = that match {
    case that: GlutenCost if plan eq that.plan =>
      0
    case that: GlutenCost if plan == that.plan =>
      // Plans are identical. Considers the newer one as having lower cost.
      -(plan.id - that.plan.id)
    case _ =>
      // Plans are different. Use the default cost compare.
      super.compare(that)
  }

  override def hashCode(): Int = throw new UnsupportedOperationException()

  override def equals(obj: Any): Boolean = obj match {
    case that: Cost => compare(that) == 0
    case _ => false
  }
}

/** This [[CostEvaluator]] is to force use the new physical plan when cost is equal. */
case class GlutenCostEvaluator() extends CostEvaluator with SQLConfHelper {
  override def evaluateCost(plan: SparkPlan): Cost = {
    val forceOptimizeSkewedJoin = conf.getConf(SQLConf.ADAPTIVE_FORCE_OPTIMIZE_SKEWED_JOIN)
    val simpleCost = SimpleCostEvaluator(forceOptimizeSkewedJoin)
      .evaluateCost(plan)
      .asInstanceOf[SimpleCost]
    new GlutenCost(simpleCost.value, plan)
  }
}
```

zhztheplayer:
```
class GlutenCost(value: Long, plan: SparkPlan) extends SimpleCost(value)
```

I will not be into this approach. Class inherency (despite that it's a case-class inheritance) always make things more complicated.


> There seems to be a codependency issue here.

There should not be codependency. `val eval: CostEvaluator` is the one that used by Spark (SimpleCostEvaluator) rather than GlutenCostEvaluator. E.g.,


```scala
case class GlutenCostEvaluator() extends CostEvaluator with SQLConfHelper {
  override def evaluateCost(plan: SparkPlan): Cost = {
    val forceOptimizeSkewedJoin = conf.getConf(SQLConf.ADAPTIVE_FORCE_OPTIMIZE_SKEWED_JOIN)
    new GlutenCost(SimpleCostEvaluator(forceOptimizeSkewedJoin), plan)
  }
}
```

zml1206: SimpleCostEvaluator.evaluateCost return SimpleCost, not GlutenCost

zhztheplayer: > SimpleCostEvaluator.evaluateCost return SimpleCost, not GlutenCost

It's intentional. Gluten only chooses the newest plan when plans are equal. Otherwise we just compare against SimpleCost.

zml1206:
```
case class GlutenCostEvaluator() extends CostEvaluator with SQLConfHelper {
  override def evaluateCost(plan: SparkPlan): Cost = {
    val forceOptimizeSkewedJoin = conf.getConf(SQLConf.ADAPTIVE_FORCE_OPTIMIZE_SKEWED_JOIN)
    new GlutenCost(SimpleCostEvaluator(forceOptimizeSkewedJoin), plan)
  }
}
```

zhztheplayer (approved):
% a rebase. Thanks.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -124,32 +124,10 @@ case class OffloadJoin() extends OffloadSingleNode with LogLevelUtil {
...
-            plan.leftKeys,
-            plan.rightKeys,
-            plan.joinType,
-            TransformHints.getShuffleHashJoinBuildSide(plan),
-            plan.condition,
-            left,
-            right,
-            plan.isSkewJoin)
+        ShuffledHashJoinExecTransformerBase.from(plan)
       case plan: SortMergeJoinExec =>
```
ulysses-you, line 129: We should also change to use shuffled hash join transformer to do validate when add sort merge join hint. Otherwise, the shj may fail because it only pass the smj validation. But it's a bit confused to add a shj hint to smj, we may want a smj transformer at the end.

gluten-core/src/main/scala/org/apache/gluten/extension/StrategyOverrides.scala
```diff
@@ -1,213 +0,0 @@
...
-                  planLater(left),
-                  planLater(right)))
-          }
-          .getOrElse(Nil)
-      }
-      Nil
-    }
-  }
-
-  def existsMultiJoins(plan: LogicalPlan, count: Int = 0): Boolean = {
```
ulysses-you, line 152: The code in strategy have a feature to prevent converting multi-joins to shj. Is it still exited in the new code ?

gluten-core/src/main/scala/org/apache/gluten/extension/StrategyOverrides.scala
```diff
@@ -1,213 +0,0 @@
...
-import org.apache.spark.sql.execution.adaptive.{BroadcastQueryStageExec, LogicalQueryStage}
-import org.apache.spark.sql.execution.joins.BroadcastHashJoinExec
-
-object StrategyOverrides extends GlutenSparkExtensionsInjector {
-  override def inject(extensions: SparkSessionExtensions): Unit = {
-    extensions.injectPlannerStrategy(JoinSelectionOverrides)
-  }
-}
-
-case class JoinSelectionOverrides(session: SparkSession)
```
ulysses-you, line 41: Do we really to refactor strategy to columnar ? Since we have a cost evalutor and the plan tag can always pass to columnar rules, can we add a tag to specify the shj is converted by gluten and then we can fallback it to smj at columnar rule side if it can not be offloaded.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
zml1206: Do validate is in `AddTransformHintRule`
```
val transformer = OffloadJoin.transformSortMergeJoinExec(plan)
transformer.doValidate().tagOnFallback(plan)
```

gluten-core/src/main/scala/org/apache/gluten/extension/StrategyOverrides.scala
zml1206: My mistake, I will add it to the column rule later.

gluten-core/src/main/scala/org/apache/gluten/extension/StrategyOverrides.scala
zml1206: Adding a tag only puts the buildSide selection in front, which seems to make little difference.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
ulysses-you: I think the expected behavior for the smj is:
1. convert smj to shj if can offload shj and enable the config, otherwise 2
2. offload smj, otherwise 3
3. fallback smj

So, I think we miss the part 2: if we can not offload shj but support offload smj.

zml1206: OffloadJoin.transformSortMergeJoinExec(smj) Will determine whether to offload smj or shj based on conditions.

ulysses-you: I think it does not help at least in these two cases:
1. Do we support offload smj if the shj validation failed ?
2. What happens if we disable offload smj, `spark.gluten.sql.columnar.sortMergeJoin=false`? before we can offload shj.

zml1206: There is indeed a problem, thank you, I will fix it right away.

gluten-core/src/main/scala/org/apache/gluten/extension/StrategyOverrides.scala
zml1206: One more question, why did fallback multi joins need to be used before?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
zml1206: Regarding the first point, how about we keep the logic before PR, when force shj, if validation fails, fallback?

gluten-core/src/main/scala/org/apache/gluten/extension/StrategyOverrides.scala
zhztheplayer: If possible, we'd try our best to offload a computation in one single place in code. If we use tag in strategy and read that tag in columnar rule for fallback, that would mean the e.g. shj plan can be barely allowed to change between the two operations. Such code would dramatically increase the planner/rule code complexity over time.

So I am inclined to doing join selection either in columnar rule or in strategy. Don't do it in both places as long as we have other options.

gluten-core/src/main/scala/org/apache/gluten/extension/StrategyOverrides.scala
zml1206: Q72 has several joins in the same stage, it shows very bad performance if we use hash join. Fallback multi joins and fallback multi codegens is a temp solution to fallback to Vanilla Spark on such stages. #682 

Now that joinSelectionOverride has been removed, fallback multi joins seem no longer needed, FallbackMultiCodegens is enough. cc @zhztheplayer

zhztheplayer: If I remember correctly that two switches are cooperated for falling back to the original vanilla plan. If we could get that plan in columnar rule after this change, then I think we will no longer need that code.

cc @PHILO-HE

zml1206: cc @JkSelf @FelixYBW

zml1206: tpcds Q72 performance is normal.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
zml1206: fixed.

zml1206:
cc @zhztheplayer @ulysses-you If you have time, can you help take a look again? Thank you very much.

gluten-core/src/main/scala/org/apache/gluten/execution/SortUtils.scala
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package org.apache.gluten.execution
+
+import org.apache.spark.sql.execution.{ProjectExec, SortExec, SparkPlan}
+
+object SortUtils {
+  def dropPartialSort(plan: SparkPlan): SparkPlan = plan match {
+    case PartialSortChild(p) => p
+    case ProjectChild(PartialSortChild(ProjectChild(p))) if plan.outputSet == p.outputSet => p
```
zhztheplayer, line 24: nit:

```
case PartialSortLike(child) => child
case ProjectLike(PartialSortLike(ProjectLike(child))) if plan.outputSet == child.outputSet => child
```

For code readability. Also, let's add some comments to demostrate where the projects came from (e.g., from pre/post project-pulling)

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
```diff
@@ -158,7 +161,10 @@ object Validators {
     override def validate(plan: SparkPlan): Validator.OutCome = plan match {
       case p: SortExec if !conf.enableColumnarSort => fail(p)
       case p: WindowExec if !conf.enableColumnarWindow => fail(p)
-      case p: SortMergeJoinExec if !conf.enableColumnarSortMergeJoin => fail(p)
+      case p: SortMergeJoinExec
+          if !conf.enableColumnarSortMergeJoin &&
+            !GlutenConfig.getConf.forceShuffledHashJoin =>
```
zhztheplayer, line 166: Use `conf` insead of `GlutenConfig.getConf`

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -181,7 +159,24 @@ case class OffloadJoin() extends OffloadSingleNode with LogLevelUtil {
       case other => other
     }
   }
+}
 
+object OffloadJoin {
+  def transformSortMergeJoinExec(smj: SortMergeJoinExec): TransformSupport = {
+    if (GlutenConfig.getConf.forceShuffledHashJoin) {
```
ulysses-you, line 166: Can we add a new rule to convert smj to shj first ? There are many code place check `forceShuffledHashJoin`. I'm afraid there may be some leaked and other developers also may miss it in future.

ulysses-you:
thank you @zml1206 , can we add a test case to verify we can use vanilla Spark join plan if fallback ?

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/execution/GlutenReplaceHashWithSortAggSuite.scala
```diff
@@ -85,7 +85,7 @@ class GlutenReplaceHashWithSortAggSuite
     withTempView("t1", "t2") {
       spark.range(100).selectExpr("id as key").createOrReplaceTempView("t1")
       spark.range(50).selectExpr("id as key").createOrReplaceTempView("t2")
-      Seq("COUNT", "COLLECT_LIST").foreach {
+      Seq("COLLECT_LIST").foreach {
```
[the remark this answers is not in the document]
zml1206: Since `ShuffledHashJoinExec` becomes `SortMergeJoinExec`, `COUNT` remains consistent with the vanilla spark suite. For more clarity, I added the `COUNT` suite.

zml1206:
> thank you @zml1206 , can we add a test case to verify we can use vanilla Spark join plan if fallback ?

Added.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -181,7 +157,38 @@ case class OffloadJoin() extends OffloadSingleNode with LogLevelUtil {
...
+
+case class OffloadSortMergeJoin() extends OffloadSingleNode with LogLevelUtil {
+  override def offload(plan: SparkPlan): SparkPlan = {
+    if (TransformHints.isNotTransformable(plan)) {
+      logDebug(s"Columnar Processing for ${plan.getClass} is under row guard.")
+      return plan
+    }
+    plan match {
+      case plan: SortMergeJoinExec =>
+        OffloadSortMergeJoin.transformSortMergeJoinExec(plan)
```
ulysses-you, line 170: I mean we can add a rule to convert `SortMergeJoinExec` to `ShuffledHashJoinExec`. So that we do not need to check  `forceShuffledHashJoin` here and there.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteSparkPlanRulesManager.scala
```diff
@@ -67,12 +67,16 @@ class RewriteSparkPlanRulesManager private (rewriteRules: Seq[RewriteSingleNode]
     }
   }
 
+  private val nodeNameMap = Map(
+    "SortMergeJoin" -> "ShuffledHashJoin",
+    "SortMergeJoin(skew=true)" -> "ShuffledHashJoin(skew=true)")
+
```
zhztheplayer, line 73: Why is this mapping needed? This is like an exception comparing to other rewrites.

zml1206: year, other rules are to add nodes, and rewriteJoin is to replace nodes.

zhztheplayer:
Run Gluten Clickhouse CI

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteSparkPlanRulesManager.scala
zhztheplayer: cc @ulysses-you

ulysses-you: Use nodeName to get original plan is bit strict. It seems the only added node is ProjectExec. We can skip all ProjectExec and make sure there is only one target rewritten plan. e.g.:
```diff
-  private def getTransformHintBack(
-      origin: SparkPlan,
-      rewrittenPlan: SparkPlan): Option[TransformHint] = {
-    // The rewritten plan may contain more nodes than origin, here use the node name to get it back
+  private def getTransformHintBack(rewrittenPlan: SparkPlan): Option[TransformHint] = {
+    // The rewritten plan may contain more nodes than origin, for now it should only be
+    // `ProjectExec`.
     val target = rewrittenPlan.collect {
-      case p if p.nodeName == origin.nodeName => p
+      case p if !p.isInstanceOf[ProjectExec] => p
```

zml1206: Updated, also excludes `RewrittenNodeWall`.

ulysses-you:
@zml1206  It seems clickhouse failed test is related, jenkins account:
https://github.com/apache/incubator-gluten/blob/main/docs/get-started/ClickHouse.md#new-ci-system

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteJoin.scala
```diff
@@ -0,0 +1,61 @@
...
+    } else {
+      None
+    }
+  }
+
+  override def rewrite(plan: SparkPlan): SparkPlan = plan match {
+    case smj: SortMergeJoinExec if GlutenConfig.getConf.forceShuffledHashJoin =>
+      getBuildSide(smj.joinType) match {
+        case Some(buildSide) =>
+          ShuffledHashJoinExec(
```
ulysses-you, line 48: Can we move drop local sort code into this rule ? The sort is a part pf smj, we should handle them together. We do not need introduce SortUtils, just move code into this rule.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteJoin.scala
```diff
@@ -0,0 +1,61 @@
...
+      Some(BuildRight)
+    } else if (leftBuildable) {
+      Some(BuildLeft)
+    } else {
+      None
+    }
+  }
+
+  override def rewrite(plan: SparkPlan): SparkPlan = plan match {
+    case smj: SortMergeJoinExec if GlutenConfig.getConf.forceShuffledHashJoin =>
```
ulysses-you, line 45: nit: better to define `val forceShuffledHashJoin = GlutenConfig.getConf.forceShuffledHashJoin` as the class member variables

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
```diff
@@ -421,7 +395,7 @@ case class AddTransformHintRule() extends Rule[SparkPlan] {
               plan.leftKeys,
               plan.rightKeys,
               plan.joinType,
-              TransformHints.getShuffleHashJoinBuildSide(plan),
```
ulysses-you, line 424: Do we need change it? The build side may be changed when we offload shj, and we should use the target build side to do validate.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteSparkPlanRulesManager.scala
ulysses-you, line 1: We should match `SortMergeJoinExec` in `mayNeedRewrite`. The CI did not fail due to the `SortMergeJoinExec` always have `SortExec`.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteJoin.scala
zml1206: RewriteJoin is oject, this cannot be modified forceShuffledHashJoin dynamically.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteSparkPlanRulesManager.scala
zml1206: I don't quite understand, mayNeedRewrite match `BaseJoinExec`, it contains `SortMergeJoinExec`.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
zml1206: There is no problem with the current logic, but there may be hidden dangers in future logic changes. I will modify it immediately.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteJoin.scala
zml1206: `RewriteSparkPlanRulesManager` currently handles a single node, and simple move code cannot be achieved. 
Is it possible to consider whether it can be optimized in future pr?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
zml1206: I thought about it carefully. The buildSide itself is the legal buildSide in the corresponding backends obtained. The buildSide will not be verified in doValidate, and it should not be done in the future. If there are changes in the future, the corresponding supportHashBuildJoinTypeOnLeft and supportHashBuildJoinTypeOnRight should be changed.

ulysses-you: The buildSide here is from vanilla Spark rather than Gluten backends.

zml1206: yea, but I think there is no need to convert here, because do validate does not require buildside.

zml1206: We should still stay consistent to avoid risks, updated.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/RewriteJoin.scala
zml1206: Done, keep `SortUtils` for use it later for sortAgg.

zml1206:
There seems to be a problem with the benchmark workflow and it keeps being queued. @ulysses-you cc @zhztheplayer

zhouyuan:
CC @zzcclp @baibaichen

zml1206:
Resolve conflicts

ulysses-you (approved):
lgtm, thank you @zml1206

zml1206:
Thank you very much for your patiently review. @ulysses-you @zhztheplayer

Issue #6101, opened 2024-06-16
[VL] Enable map_from_arrays function


WangGuangxin:
### Description

Enable map_from_arrays function

ArnavBalyan:
Hi @WangGuangxin I came across this, we need this support, just wanted to know if this is active. I can help cover any work that is left. Thanks

WangGuangxin:
> Hi [@WangGuangxin](https://github.com/WangGuangxin) I came across this, we need this support, just wanted to know if this is active. I can help cover any work that is left. Thanks

@ArnavBalyan yes, its pending the pr https://github.com/facebookincubator/velox/pull/10280, you can help test and review this if you are convenient

PR #6102, opened 2024-06-16
[GLUTEN-6101][VL] Enable map_from_array function

WangGuangxin:

Enable map_from_array function

(Fixes: \#6101)

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
@@ -2074,6 +2074,28 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
   test("Deduplicate sorting keys") {
     runQueryAndCompare("select * from lineitem order by l_orderkey, l_orderkey") {
       checkGlutenOperatorMatch[SortExecTransformer]
+  }
+
+  test("test map_from_arrays") {
+    withTempView("t") {
+      Seq((Seq(1, 2, 1), Seq("a", "b", "c"))).toDF("k", "v").createOrReplaceTempView("t")
+      withSQLConf(SQLConf.MAP_KEY_DEDUP_POLICY.key -> SQLConf.MapKeyDedupPolicy.LAST_WIN.toString) {
+        runQueryAndCompare(
+          """
+            |select map_from_arrays(k, v) from t
+            |""".stripMargin
+        ) {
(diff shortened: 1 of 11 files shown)


WangGuangxin:
cc @rui-mo @PHILO-HE

rui-mo (commented):
Do we need to fallback the execution if MAP_KEY_DEDUP_POLICY is LAST_WIN? Thanks.

https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L4139-L4148

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -656,6 +656,23 @@ object ExpressionConverter extends SQLConfHelper with Logging {
           Seq(replaceWithExpressionTransformerInternal(c.child, attributeSeq, expressionsMap)),
           c
         )
+      case _: MapFromArrays if BackendsApiManager.getBackendName == "velox" =>
+        // for velox backend, we should mapping map_from_arrays based on the map key dedup policy
+        val mapKeyDedupPolicy = SQLConf.get.getConf(SQLConf.MAP_KEY_DEDUP_POLICY)
+        val newSubstraitExprName =
+          if (mapKeyDedupPolicy == SQLConf.MapKeyDedupPolicy.EXCEPTION.toString) {
+            "map_from_arrays_exception_on_duplicates"
```
PHILO-HE, line 664: Is this function name really used?

PHILO-HE (commented):
@WangGuangxin, could you do a rebase? Thanks!

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
WangGuangxin: yes, it's related to https://github.com/facebookincubator/velox/pull/10280.

But I think it's more reasonable to check the `mapKeyDedupPolicy` config in velox instead of register two functions. WDYT @PHILO-HE @rui-mo @jinchengchenghh

PHILO-HE: Using a config looks better to me. But it requires some extra code change for that presto function. You can check with Meta guys.

WangGuangxin:
pending https://github.com/facebookincubator/velox/pull/10280

PR #6106, opened 2024-06-17
[VL] Fix RowToColumn metric convert time

jinchengchenghh:
next() expires whole stage time

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/RowToVeloxColumnarExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/RowToVeloxColumnarExec.scala
@@ -144,11 +144,21 @@ object RowToVeloxColumnarExec {
           it.hasNext
         }
       }
+      
+      def convertToUnsafeRow(row: InternalRow): UnsafeRow = {
+        row match {
+          case unsafeRow: UnsafeRow => unsafeRow
+          case _ =>
+            converter.apply(row)
+        }
+      }
 
-      def nativeConvert(row: UnsafeRow): ColumnarBatch = {
+      override def next(): ColumnarBatch = {
+        val firstRow = it.next()
+        val start = System.currentTimeMillis()
+        val row = convertToUnsafeRow(firstRow)
         var arrowBuf: ArrowBuf = null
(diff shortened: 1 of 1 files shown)


jinchengchenghh:
image

jinchengchenghh:
@FelixYBW  Can you help take a look? Thanks!

zhztheplayer:
Would suggest using `Iterators.wrap.collectReadMillis` to calculate self time with `out time - in time` basis to simplify code.

See example

https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/execution/VeloxAppendBatchesExec.scala#L69
https://github.com/apache/incubator-gluten/blob/e2f90e96146b97c5fef1fad5fe704ce37faa9894/backends-velox/src/main/scala/org/apache/gluten/execution/VeloxAppendBatchesExec.scala#L82

jinchengchenghh:
Without string in project
image

jinchengchenghh:
When I implement as you suggest, the bug occur again. 
image

zhztheplayer:
> When I implement as you suggest, the bug occur again.

"out - in" should be accurate enough in most cases. There might be something wrong or hidden? I'll help to check once I got some time.

zhztheplayer:
@jinchengchenghh Can you share the entire stage's DAG? Thanks.

backends-velox/src/main/scala/org/apache/gluten/execution/RowToVeloxColumnarExec.scala
```diff
@@ -145,10 +145,20 @@ object RowToVeloxColumnarExec {
...
+      def convertToUnsafeRow(row: InternalRow): UnsafeRow = {
+        row match {
+          case unsafeRow: UnsafeRow => unsafeRow
+          case _ =>
+            converter.apply(row)
+        }
+      }
+
+      override def next(): ColumnarBatch = {
+        val firstRow = it.next()
```
FelixYBW, line 157: why the first row needs to be handled seperately?

jinchengchenghh: Because we need to get estimate row size by the first row

zhztheplayer (approved):

PR #6112, opened 2024-06-17
[CORE] Support JDK 11

surnaik:

* Add jdk-11 maven profile
* Add jdk11 tpcds/tpch github action workflow
* Update docs
* Upgrades Caffeine cache version for Java 11 and 17

## How was this patch tested?

Pass CI


surnaik:
@ulysses-you Requesting your review since you upgraded to Java 17. Thanks!

surnaik:
Please take a look - @zhouyuan @yaooqinn Thanks!!

zhouyuan (commented):
Looks good to me, is it ok to limit the tests to Spark 3.5 first, so we could save some CI resource?

ulysses-you:
+1 for @zhouyuan to reduce the CI resource

.github/workflows/velox_docker.yml
```diff
@@ -510,6 +533,8 @@ jobs:
       - name: Setup java and maven
         run: |
           apt-get update && apt-get install -y openjdk-8-jdk maven wget
+          apt remove openjdk-11* -y
+          export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
```
zhztheplayer, line 537: GHA creates new shell for each step so this export command may not work. You can try using `>> $GITHUB_ENV` to store the variables.

surnaik: Thanks, fixed!!

surnaik:
@zhouyuan @ulysses-you I have now enabled it only for Spark 3.5, Please take a look at this again. Thanks!

PR #6117, opened 2024-06-18
[VL] Gluten-it: Reuse Spark sessions that share same configuration

zhztheplayer:
Reuse sessions with same configuration for better support of large scale / long run integration testing.

Reusing the session would avoid reloading the libraries if Gluten is enabled, and would avoid recreating table catalogs again and again.

Add `--no-session-reuse` which is enabled in `random kill task` CI job. Otherwise the job would fail by `no space on disk` since too many temporary files are created in the same session.


zhztheplayer:
cc @baibaichen @zzcclp another unrelated CH CI error

PHILO-HE (approved):
Thanks!

PR #6123, opened 2024-06-18
[VL] Prefer to use `path.getFileSystem` instead of `FileSystem.get` to create `FileSystem`

yikf:
Prefer to use `path.getFileSystem` instead of `FileSystem.get` to create `FileSystem`, this can always default to the path's schema to get fs which handle the path.

## How was this patch tested?
exist tests and manual tests


JkSelf (approved):
Thanks.

PR #6129, opened 2024-06-18
[VL] Support linking system libprotobuf.a when building arrow

jackylee-ch:
Fix the [bug](https://github.com/apache/arrow/issues/39919) in arrow-15.0.0, which has already been fixed in 16.0.0 or higher version.

See https://github.com/apache/arrow/pull/40015.

## How was this patch tested?
GA


PHILO-HE (approved):
Thanks!

PR #6021, opened 2024-06-07
[VL] [Core] Spark Input_file_name Support

gaoyangxiaozhu:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: https://github.com/facebookincubator/velox/issues/9957)

## How was this patch tested?

UT and manually test

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #6139, opened 2024-06-19
[VL][Core] Add InputFileReplaceFallback Rule

gaoyangxiaozhu:

The is a following up PR for `input_file_name` native support, it 
1. Add a `inputfileNameReplaceFallbackRule` to handle the case if scan fallback due to complex data type
2. Add a feature flag to control if enable the rules in case any corner issues happen we can disable the rule

(Fixes: https://github.com/facebookincubator/velox/issues/9957)

## How was this patch tested?

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/InputFileNameReplaceFallbackRule.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/InputFileNameReplaceFallbackRule.scala
@@ -22,6 +22,31 @@ import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.{ColumnarToRowExec, FileSourceScanExec, ProjectExec, SparkPlan}
 import org.apache.spark.sql.execution.datasources.v2.BatchScanExec
 
+/**
+ * Velox doesn't have input_file_name function support yet and it's also hard to add one, checking
+ * this discussion - https://github.com/facebookincubator/velox/issues/9957
+ *
+ * For workaround, Gluten currently use `InputFileNameReplaceRule` to convert `project` with
+ * `input_file_name` to a new project with a new projected metadata column named `$input_file_name$`
+ * which store the file name.
+ *
+ * While if scan is fallback, the replacement would fail the job due to the metadata column
+ * `$input_file_name$` not available.
+ *
+ * This final rule use to address the scan fallback cases. If scan is fallback and has replaced
+ * attrs, the rule would inject a project after the scan to convert the replaced attrs back to
+ * `input_file_name`.
+ *
+ * For example: + ProjectTransformer [$input_file_name$, name] +
+ * InputIteratorTransformer[$input_file_name$, name] + RowToVeloxColumnar + ColumnarToRow + FileScan
+ * [$input_file_name$, name]
+ *
+ * would be convert to
+ *
+ * + ProjectTransformer [$input_file_name$, name] + InputIteratorTransformer[$input_file_name$,
+ * name] + project [input_file_name() as $input_file_name$, name] + RowToVeloxColumnar +
+ * ColumnarToRow + FileScan [name]
+ */
 case class InputFileNameReplaceFallbackRule(spark: SparkSession) extends Rule[SparkPlan] {
   import InputFileNameReplaceRule._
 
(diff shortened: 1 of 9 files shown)


backends-velox/src/main/scala/org/apache/gluten/extension/InputFileNameReplaceFallbackRule.scala
```diff
@@ -0,0 +1,85 @@
...
+ */
+package org.apache.gluten.extension
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, InputFileBlockLength, InputFileBlockStart, InputFileName, NamedExpression}
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.{ColumnarToRowExec, FileSourceScanExec, ProjectExec, SparkPlan}
+import org.apache.spark.sql.execution.datasources.v2.BatchScanExec
+
+case class InputFileNameReplaceFallbackRule(spark: SparkSession) extends Rule[SparkPlan] {
```
zhztheplayer, line 25: Would you please add comments for this rule as well as `InputFileNameReplaceRule`? 

I wouldn't easily figure out the purpose of the rules only by looking into the class name.

zhztheplayer:
I think we'd follow a criteria when rewriting an expression / operator **before** entering actual offloading phase (e.g., using genExtendedColumnarValidationRules to do the replacement): to make it compatible with Spark as long as possible and to avoid converting it back when the computation doesn't offload.

Can you take a look at some current rules that already follow the criteria, e.g., the ones to offload collet_list, collect_set, hll, bloom-filter, dpp subquery, to see if we can do the same on the expressions used by `input_file_name` ?

zhztheplayer:
A basic question for https://github.com/apache/incubator-gluten/pull/6021 : Why didn't we do the replacement during filter-pushdown?

backends-velox/src/main/scala/org/apache/gluten/extension/InputFileNameReplaceFallbackRule.scala
gaoyangxiaozhu: done

gaoyangxiaozhu:
> I think we'd follow a criteria when rewriting an expression / operator **before** entering actual offloading phase (e.g., using genExtendedColumnarValidationRules to do the replacement): to make it compatible with both native lib and vanilla Spark as long as possible and to avoid converting it back when the computation doesn't offload.
> 
> Can you take a look at some current rules that already follow the criteria, e.g., the ones to offload collet_list, collect_set, hll, bloom-filter, dpp subquery, to see if we can do the same on the expressions used by `input_file_name` ?

@zhztheplayer ,  thank you! actually i have checked this part but didn't figure out a similar way for `input_file_name` support.  The current implement for `collect_list`, `collect_set` etc is based on truth that velox already has a related function available only need some data type compatible change. But velox right now doesn't have a `input_file_name` available.

zhztheplayer:
> I'd like do a following PR after this PR to refactor this part to merge 2 rules together. :)

Which works for me but let's turn off the option by default and mark it as experimental? Until we got a more robust design. Thanks.

zhztheplayer:
BTW I'll take a closer look at the code asap.

gaoyangxiaozhu:
> BTW I'll take a closer look at the code asap.

thanks, since without this change, user may encounter the failure issue if scan fallback with complex data type and project has `input_file_name`.

gaoyangxiaozhu:
don't think the fail related to my change and it is randomly, meanwhile the CH CI fail also looks a CI issue ?  @zhouyuan / @PHILO-HE could you help double check ? 
image

PHILO-HE:
Run Gluten Clickhouse CI

PHILO-HE:
@gaoyangxiaozhu, Spark 3.4 tests passed after re-triggered. Let's wait for CH CI.

zhztheplayer:
@gaoyangxiaozhu I opened https://github.com/apache/incubator-gluten/issues/6157 for further work to improve this. I am assign that issue to you.

gaoyangxiaozhu:
> @gaoyangxiaozhu I opened #6157 for further work to improve this. I am assign that issue to you. And I suggest to have a clear plan for the new implementation before merging this.


> @gaoyangxiaozhu I opened #6157 for further work to improve this. I am assign that issue to you. And I suggest to have a clear plan for the new implementation before merging this.

got!  So I send seperate PR to turn off `inputFileNameReplaceRule` with feature flag https://github.com/apache/incubator-gluten/pull/6161, let make sure that PR merged in case any issue caused due to scan fallback.

@jinchengchenghh / @zhztheplayer / @FelixYBW please help review.

lwz9103:
Run Gluten Clickhouse CI

lwz9103:
Run Gluten Clickhouse CI

Issue #6145, opened 2024-06-19
Gluten may core dump when running on JVM 17


yangzhg:
### Backend

VL (Velox)

### Bug description

![img_v3_02c0_ea1f2083-0df2-48d2-a1e2-07ebd9bd96fg](https://github.com/apache/incubator-gluten/assets/9098473/f90fb50c-5834-4909-9c5f-11d733395f6c)

spark will core dump when load gluten

### Spark version

Spark-3.2.x

PR #6146, opened 2024-06-19
[VL] Fix potential double unload of JNI libraries on Java 11+

yangzhg:

Previously, the addHookForLibUnloading method added a shutdown hook unconditionally to unload JNI libraries. This could potentially lead to double unload issues on Java 11 and later versions where the JVM automatically handles JNI library cleanup.

This commit modifies addHookForLibUnloading to conditionally add the unload hook based on the JVM version:
- For Java 11 and later, the method adds a no-op hook since the JVM handles the cleanup automatically.
- For Java versions prior to 11, the method adds the provided hook to ensure proper cleanup.

This change ensures compatibility and prevents potential double unload issues on Java 11 and later.

Fixes: #6145

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
@@ -16,6 +16,10 @@
  */
 package org.apache.spark.util
 
+import org.apache.gluten.backendsapi.BackendsApiManager
+
+import org.apache.commons.lang3.{JavaVersion, SystemUtils}
+
 object GlutenShutdownManager {
 
   def addHook(hook: () => Unit): AnyRef = {
@@ -23,7 +27,17 @@ object GlutenShutdownManager {
   }
 
   def addHookForLibUnloading(hook: () => Unit): AnyRef = {
-    ShutdownHookManager.addShutdownHook(ShutdownHookManager.SPARK_CONTEXT_SHUTDOWN_PRIORITY)(hook)
+    ShutdownHookManager.addShutdownHook(ShutdownHookManager.SPARK_CONTEXT_SHUTDOWN_PRIORITY) {
+      if (
+        BackendsApiManager.getBackendName == "velox" && SystemUtils.isJavaVersionAtLeast(
+          JavaVersion.JAVA_11)
+      ) {
+        // In Java 11 and later, the JVM automatically handles the cleanup of JNI libraries
+        () => Unit
+      } else {
+        hook
+      }
+    }
   }
 
   def addHookForTempDirRemoval(hook: () => Unit): AnyRef = {


zhztheplayer:
Which library Velox's lib loader tries to unload? I remember that option (`boolean requireUnload`) was only used by CH backend.

yangzhg:
> Which library Velox's lib loader tries to unload? I remember that option (`boolean requireUnload`) was only used by CH backend.

I'm not sure about which libs tries to unload, but I found after Java 11 there is no need to unload JNI libs, and  the unload may cause core dump or UBs.

zhztheplayer:
> > Which library Velox's lib loader tries to unload? I remember that option (`boolean requireUnload`) was only used by CH backend.
> 
> I'm not sure about which libs tries to unload, but I found after Java 11 there is no need to unload JNI libs, and the unload may cause core dump or UBs.

Then I suppose this is not related to https://github.com/apache/incubator-gluten/issues/6145 ? I see `Fixes: #6145` in PR description.

@zzcclp Regarding lib-unloading, why CH backend needs to unload a library?

zzcclp:
> requireUnload

There are some resources need to be released when shutting jvm, otherwise ch will core dump

yangzhg:
> Then I suppose this is not related to #6145 ? I see `Fixes: #6145` in PR description.

apply this patch  the core dump is fixed

yangzhg:
> resources

I will try to fix only in veloxbackend

yangzhg:
@zzcclp @zhztheplayer I have made it only worked on Velox backend

lwz9103:
Run Gluten Clickhouse CI

gluten-core/src/main/scala/org/apache/spark/util/GlutenShutdownManager.scala
```diff
@@ -16,14 +16,28 @@
...
 
   def addHook(hook: () => Unit): AnyRef = {
     ShutdownHookManager.addShutdownHook(ShutdownHookManager.DEFAULT_SHUTDOWN_PRIORITY)(hook)
   }
 
   def addHookForLibUnloading(hook: () => Unit): AnyRef = {
-    ShutdownHookManager.addShutdownHook(ShutdownHookManager.SPARK_CONTEXT_SHUTDOWN_PRIORITY)(hook)
+    ShutdownHookManager.addShutdownHook(ShutdownHookManager.SPARK_CONTEXT_SHUTDOWN_PRIORITY) {
+      if (
+        BackendsApiManager.getBackendName == "velox" && SystemUtils.isJavaVersionAtLeast(
```
zhztheplayer, line 32: `if (BackendsApiManager.getBackendName == "...")`

Seems like a writing we should avoid in core module's code. Let's figure out if there is a better way.

zhztheplayer (commented):
Thanks. But let's find another way rather than checking backend type in core module.

zhztheplayer:
@yangzhg I could not reproduce the unloading error since the static storage `REQUIRE_UNLOAD_LIBRARY_PATHS` looks to be always empty in Velox backend thus the shutdown hook should ideally do nothing.

image

For different build configurations I assume this list should either be empty since we always pass `requireUnload=false` when using JniLibLoader in Velox backend. Maybe I miss something?

Would you like to collect some log or debug info to see which library is being unloaded in your case? Let's find out the root cause prior to doing any code changes. Thank you very much.

PR #6160, opened 2024-06-20
[VL] Performs partial sort before global sort in SortShuffleWriter

marin-ma:


marin-ma:
@GlutenPerfBot benchmark

PR #6163, opened 2024-06-20
[GLUTEN-6151] Reset  local property after finishing write operator

JkSelf:

In parquet write, we use several local properties to pass the status in session. We need reset them after finishing write operator.

## How was this patch tested?

Existing unit tests.


JkSelf:
@FelixYBW

FelixYBW:
Are here the best place to reset? are they the per session config, right? what happens if two tasks from different queries are scheduled to the same spark session? is it possible?

@zhztheplayer

FelixYBW:
What happens if we have two SQLs and both can offload to native? Does the second SQL set the property again?

FelixYBW (approved):

FelixYBW:
COnfirmed bug is fixed. Thank you, Ke

zhztheplayer:
> Are here the best place to reset? are they the per session config, right? what happens if two tasks from different queries are scheduled to the same spark session? is it possible?
> 
> @zhztheplayer

I recall that this part of code relies on thread local storages which should make the invocations thread safe. Having said that we can still look for better choices, e.g., passing through arguments, to improve portability of the code.

Issue #4451, opened 2024-01-19
[CH] filter may change the schema in clickhouse


shuai-xu:
### Backend

CH (ClickHouse)

### Bug description

when run sql 14695, it throw exception "Illegal column Int64 of argument of function empty: while executing 'FUNCTION empty(day#609 : 105) -> empty(day#609) Nullable(UInt8) : 107'", We found that it's due to mismatch schema in Filter and its following Project. In clickhouse, FilterTransform may change the input schema, e.g, it may remove duplicated columns, and then in its following ProjectTransform, the schema from FilterTransfrom is different from schama passed from spark driver.

### Spark version

Spark-3.3.x

PR #6166, opened 2024-06-20
[GLUTEN-4451] [CH] fix header maybe changed by FilterTransform

shuai-xu:

Rollback header if changed in FilterTransform

(Fixes: \#4451)

## How was this patch tested?

This patch was tested by integration tests.


liuneng1994 (approved):
LGTM

PR #6167, opened 2024-06-20
[GLUTEN-5643] Fix the failure when the pre-project of GenerateExec falls back

marin-ma:
Currently, if GenerateExec needs a pre-project, only ProjectNode is allowed in the native substrait -> Velox PlanNode transformation. Query would fail if the pre-project of GenerateExec falls back due to native validation failure. We should consider `ValueStreamNode` as valid in case of pre-project fallback.

Introduce `VeloxDummyExpression` for test purpose. An expression wrapped by `VeloxDummyExpression` can mock the case when a certain expression is supported by Gluten, but it fails on native validation.

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -26,6 +26,7 @@
 #include "utils/ConfigExtractor.h"
 
 #include "config/GlutenConfig.h"
+#include "operators/plannodes/RowVectorStream.h"
 
 namespace gluten {
 namespace {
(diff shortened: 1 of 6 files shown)


marin-ma:
Run Gluten Clickhouse CI

marin-ma:
@xumingming Could you help to review? Thanks!

xumingming:
> @xumingming Could you help to review? Thanks!

Sure, will do the first round of review in the recent days.

xumingming:
> Query would fail if the pre-project of GenerateExec falls back due to native validation failure. We should consider ValueStreamNode as valid in case of pre-project fallback.

@marin-ma I understand Project might fallback(some function not supported?), but I don't understand why we should consider ValueStreamNode as valid,  can you elaborate?

marin-ma:
@xumingming `ValueStreamNode` represents the input iterator is not from the pipeline. With pre-project fallback, it represents the input is from RowToColumnar conversion.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -752,9 +761,10 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
       SubstraitParser::configSetInOptimization(generateRel.advanced_extension(), "injectedProject=");
 
   if (injectedProject) {
-    auto projNode = std::dynamic_pointer_cast<const core::ProjectNode>(childNode);
+    // Child should be either ProjectNode or ValueStreamNode in case of project fallback.
     VELOX_CHECK(
-        projNode != nullptr && projNode->names().size() > requiredChildOutput.size(),
+        std::dynamic_pointer_cast<const core::ProjectNode>(childNode) != nullptr ||
+            std::dynamic_pointer_cast<const ValueStreamNode>(childNode) != nullptr,
         "injectedProject is true, but the Project is missing or does not have the corresponding projection field")
```
xumingming, line 768: The error message need to be updated to reflect the code change.

zhztheplayer (approved):

PR #6179, opened 2024-06-21
[VL] Enable ShuffleSuite with ColumnarShuffleManager

acvictor:

Enable ShuffleSuite against ColumnarShuffleManager.

## How was this patch tested?

Enabled a new suite.


acvictor:
@marin-ma can you please review?

marin-ma:
Seems like we don't need to enable this suite. `ShuffleSuite` is an abstract class in spark, and we have enabled `SortShuffleSuite` to run all test cases.

PR #6190, opened 2024-06-24
[GLUTEN-3806][VL] Add Arm64 libhdfs3 support for velox build

guyuqi:
`SSE4.2` is always enabled in Project oap-project/libhdfs3 [Options.cmake](https://github.com/oap-project/libhdfs3/blob/master/CMake/Options.cmake#L3).
When building Gluten-Velox with enabling HDFS(`--enable_hdfs=ON`) on Arm64 Ubuntu, build issues occurred:

```
c++: error: unrecognized command-line option ‘-msse4.2’
ninja: build stopped: subcommand failed.
```

## How was this patch tested?
Build Gluten-Velox with --enable_hdfs=ON
./dev/builddeps-veloxbe.sh --build_tests=ON --build_benchmarks=ON --enable_hdfs=ON

libhdfs.so could be properly installed.

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -96,6 +96,8 @@ function process_setup_ubuntu {
   sed -i '/libgmock-dev/d' scripts/setup-ubuntu.sh # resolved by ep/build-velox/build/velox_ep/CMake/resolve_dependency_modules/gtest.cmake
   sed -i 's/github_checkout boostorg\/boost \"\${BOOST_VERSION}\" --recursive/wget_and_untar https:\/\/github.com\/boostorg\/boost\/releases\/download\/boost-1.84.0\/boost-1.84.0.tar.gz boost \&\& cd boost/g' scripts/setup-ubuntu.sh
   if [ $ENABLE_HDFS == "ON" ]; then
+    # TODO: The 'modify_libhdfs3.patch' will be removed once the following PR is merged into the oap-project/libhdfs.
+    # Track: https://github.com/oap-project/libhdfs3/pull/2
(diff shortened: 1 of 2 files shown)


ep/build-velox/src/get_velox.sh
```diff
@@ -137,7 +137,7 @@ function process_setup_centos8 {
   sed -i '/^  run_and_time install_fbthrift/a \  run_and_time install_openssl' scripts/setup-centos8.sh
 
   if [ $ENABLE_HDFS == "ON" ]; then
-    sed -i '/^function install_gflags.*/i function install_libhdfs3 {\n cd "\${DEPENDENCY_DIR}"\n github_checkout oap-project/libhdfs3 master\n cmake_install\n}\n' scripts/setup-centos8.sh
+    sed -i '/^function install_gflags.*/i function install_libhdfs3 {\n cd "\${DEPENDENCY_DIR}"\n github_checkout oap-project/libhdfs3 master\n git apply ../../../src/modify_libhdfs3.patch\n cd ..\n cmake_install\n}\n' scripts/setup-centos8.sh
```
ulysses-you, line 140: can we use `CURRENT_DIR` ?

guyuqi: Yes, it seems` CURRENT_DIR` is deprecated.
Removed it. Thanks.

ep/build-velox/src/modify_libhdfs3.patch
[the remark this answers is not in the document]
guyuqi: Hi @ulysses-you  Thanks for your comments. 

It makes sense to push it to libhdfs repo directly.
I've submitted the related [PR](https://github.com/oap-project/libhdfs3/pull/2) in `oap-project/libhdfs3`. 
However, it appears that `oap-project/libhdfs3` is an inactive project. I'm unsure about the time it will take for the PR process to conclude.
I've added my comments inline. I'll continue to track this issue and will remove this libhdfs patch as soon as `oap-project/libhdfs3` accepts the related PR.

Would it be possible to accept this PR in the meantime?

Thank you. :-)

Issue #6189, opened 2024-06-24
[CORE] Shouldn't pushdown undeterministic filter to scan node


WangGuangxin:
### Backend

VL (Velox)

### Bug description

We shouldn't pushdown undeterministic filter to scan node

### Spark version

None

PR #6191, opened 2024-06-24
[GLUTEN-6189][CORE] Shouldn't pushdown undeterministic filter to scan node

WangGuangxin:

We shouldn't pushdown undeterministic filter to scan node, for example, if we push down filter with `rand() < 0.5` to scan, the result will be around 1/4 instead of 1/2.

(Fixes: \#6189)

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
@@ -1873,7 +1873,7 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
     }
   }
 
-  test("Shouldn't pushdown deterministic filter to scan node") {
+  test("Shouldn't pushdown undeterministic filter to scan node") {
     // there are total 60175 rows in lineitem, after rand filter,
     // there should be around 30000 rows
     val resultLength = spark
@@ -1883,5 +1883,6 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
       .collect()
       .length
     assert(resultLength >= 25000)
+    assert(resultLength >= 40000)
   }
 }
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
@@ -1872,4 +1872,16 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
         }
     }
   }
+
+  test("Shouldn't pushdown deterministic filter to scan node") {
+    // there are total 60175 rows in lineitem, after rand filter,
+    // there should be around 30000 rows
+    val resultLength = spark
+      .sql("""
+             |select * from lineitem where rand() <= 0.5
+             |""".stripMargin)
+      .collect()
+      .length
+    assert(resultLength >= 25000)
+  }
 }
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
@@ -1883,6 +1883,6 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
       .collect()
       .length
     assert(resultLength >= 25000)
-    assert(resultLength >= 40000)
+    assert(resultLength <= 40000)
   }
 }
--- a/gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
@@ -672,9 +672,11 @@ trait SparkPlanExecApi {
     def getPushedFilter(dataFilters: Seq[Expression]): Seq[Expression] = {
       val pushedFilters =
         dataFilters ++ FilterHandler.getRemainingFilters(dataFilters, extraFilters)
-      pushedFilters.filterNot(_.references.exists {
-        attr => SparkShimLoader.getSparkShims.isRowIndexMetadataColumn(attr.name)
-      })
+      pushedFilters
+        .filterNot(_.references.exists {
+          attr => SparkShimLoader.getSparkShims.isRowIndexMetadataColumn(attr.name)
+        })
+        .filter(_.deterministic)
     }
     sparkExecNode match {
       case fileSourceScan: FileSourceScanExec =>


WangGuangxin:
cc @PHILO-HE @rui-mo @ulysses-you

backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
```diff
@@ -1872,4 +1872,16 @@ class TestOperator extends VeloxWholeStageTransformerSuite with AdaptiveSparkPla
...
+  test("Shouldn't pushdown deterministic filter to scan node") {
+    // there are total 60175 rows in lineitem, after rand filter,
+    // there should be around 30000 rows
+    val resultLength = spark
+      .sql("""
+             |select * from lineitem where rand() <= 0.5
+             |""".stripMargin)
+      .collect()
+      .length
+    assert(resultLength >= 25000)
```
zhztheplayer, line 1885: Could give `resultLength` a upper bound to improve stability of the test? Say if `rand()` yielded a constant `0.1` then the test could still pass as false-positive.

zhztheplayer (commented):
Is there any root cause from Velox's code for this issue? It doesn't make sense if `rand()` results in a constant vector or involves in constant folding procedure.

backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
WangGuangxin: It's not velox's rand bug, bug `rand() < 0.5` was evaluate twice,  first is in `ScanNode`, then in `FilterNode`, so that the filtered result is 0.5 * 0.5.

rui-mo (commented):
> rand() < 0.5 was evaluate twice, first is in ScanNode, then in FilterNode

If one filter exists in scan node, we do not add it to the filter node again. `getRemainingFilters` is for that purpose. Is there any bug in it? Thanks.

backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
zhztheplayer: > it was because rand() < 0.5 was evaluate twice

In which place it gets evaluated twice? It should be either in subfield filters or in remaining filters already by design. Am I missing something? cc @rui-mo 

Would you mean it's placed both in subfield filters and remaining filters?

rui-mo: The design is all filters are pushed down to Spark's Scan operator on Scala side, and in Velox's TableScanNode there are subfield filters and remaining filter to handle two kinds filters according to if their push-down is supported. I'm not clear if the duplication occurs on Scala side between Scan and Filter operator, or on native side between subfield filters and remaining filter. @WangGuangxin Could you clarify?

WangGuangxin: let me double check

WangGuangxin: Take the UT as an example, the `rand() < 0.5` both exists in 
1. `getPushedFilter`  https://github.com/apache/incubator-gluten/blob/9cceba6812d4de9883695f20bc506648083715f9/gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala#L672
which is pushed to `ScanNode` ( it has nothing to do with the logic of `remainingFilter` and `subFieldFilter`)

2. and also, the `FilterExecTransformer` also has `rand() < 0.5` in `getRemainingCondition` https://github.com/apache/incubator-gluten/blob/9cceba6812d4de9883695f20bc506648083715f9/backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala#L39

zhztheplayer: > Take the UT as an example, the `rand() < 0.5` both exists in
> 
> 1. `getPushedFilter`  https://github.com/apache/incubator-gluten/blob/9cceba6812d4de9883695f20bc506648083715f9/gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala#L672
>    
>    which is pushed to `ScanNode` ( it has nothing to do with the logic of `remainingFilter` and `subFieldFilter`)
> 2. and also, the `FilterExecTransformer` also has `rand() < 0.5` in `getRemainingCondition` https://github.com/apache/incubator-gluten/blob/9cceba6812d4de9883695f20bc506648083715f9/backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala#L39

It looks like a bug. By looking into the code, I believe the author didn't want a filter expression appearing in `FilterHandler.getRemainingFilters` once it was pushed into scan. So perhaps we need a bugfix for that.

WangGuangxin:
> > rand() < 0.5 was evaluate twice, first is in ScanNode, then in FilterNode
> 
> If one filter exists in scan node, we do not add it to the filter node again. `getRemainingFilters` is for that purpose. Is there any bug in it? Thanks.

@rui-mo I think the assumption was broken since `rand() < 0.5` doesn't exists in `FileSourceScanExec`'s dataFilters, that was because Spark will not pushdown undeterministic filter to scan https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceUtils.scala#L291

backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
zhztheplayer: However I will not be against this change if it doesn't bring performance issues. Given that it's a more robust way to make sure non-deterministic filter expressions be only evaluated once from JVM side. 

But I think we still need to find out reason of the bug.

zhztheplayer: > if it doesn't bring performance issues.

But probably it's hardly true in Velox backend if the backend does support pushing non-deterministic exprs to file reader. So my assumption could be too ideal, I am not sure.

rui-mo:
@WangGuangxin The purpose of postProcessPushDownFilter is to push down all the filters to Scan even if only Filter operator contains it. After that, Filter will only process those do not exist in Scan. It looks like a bug if both Scan and Filter contain the same expression.
https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala#L38-L40

gluten-core/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -672,9 +672,11 @@ trait SparkPlanExecApi {
...
       val pushedFilters =
         dataFilters ++ FilterHandler.getRemainingFilters(dataFilters, extraFilters)
-      pushedFilters.filterNot(_.references.exists {
-        attr => SparkShimLoader.getSparkShims.isRowIndexMetadataColumn(attr.name)
-      })
+      pushedFilters
+        .filterNot(_.references.exists {
+          attr => SparkShimLoader.getSparkShims.isRowIndexMetadataColumn(attr.name)
+        })
+        .filter(_.deterministic)
```
ulysses-you, line 679: Vanilla Spark would not push down non-deterministic filer exprs. But as Gluten have already supported it, can we add a new config to control if push down non-deterministic filer exprs ?

zml1206:
Bug of getRemainingFilters. ExpressionSet can only remove deterministic expression.
```
def getRemainingFilters(scanFilters: Seq[Expression], filters: Seq[Expression]): Seq[Expression] =
    (ExpressionSet(filters) -- ExpressionSet(scanFilters)).toSeq
```
to 
```
def getRemainingFilters(scanFilters: Seq[Expression], filters: Seq[Expression]): Seq[Expression] =
    (filters.toSet -- scanFilters.toSet).toSeq
```
This seems to work, if the filter reference pushed down has not changed.

zml1206:
I submitted a PR to resolve it. #6296

FelixYBW:
Is this error in Velox or Gluten? Velox table scan supports undeterministic filter pushdown so it shouldn't be an issue to pushdown to Velox. 

@rui-mo I remember initially we have to distinguish subfiled filter and remaining filter in oap/velox. Is the code moved to Gluten?

rui-mo:
@FelixYBW This bug is in Gluten Scala code, seeing https://github.com/apache/incubator-gluten/pull/6296.

> I remember initially we have to distinguish subfiled filter and remaining filter in oap/velox.

This code is in Gluten cpp, taking effect during the conversion from Substrait to Velox. The bug in this PR is not related to that.

Issue #6067, opened 2024-06-13
[CH] Support CH backend with Spark3.5


zzcclp:
### Description

Support CH backend with Spark3.5:
- [x] https://github.com/apache/incubator-gluten/pull/6068
   - [Upgrade Spark version to 3.5 and compile passed](https://github.com/apache/incubator-gluten/pull/6068);
   - [Upgrade Delta version to 3.2 and compile passed](https://github.com/apache/incubator-gluten/pull/6068);
- [x] [Support Scala 2.13 #6311](https://github.com/apache/incubator-gluten/pull/6311) 
- [ ]  Support native write
   - https://github.com/apache/incubator-gluten/pull/6197
   - https://github.com/apache/incubator-gluten/pull/6403
   - https://github.com/apache/incubator-gluten/pull/6586
   - [ ] Collect file bytes that write pipeline produces.
   - [ ] Support `InsertIntoHiveTable`
   - [ ] Support Bucket table
   - [ ] Support escape partition value. e.g. incluing space
- [ ] CH backend UT passed: (now only the MergeTree + Delta UT passed); 
   -  https://github.com/apache/incubator-gluten/pull/6114
   - https://github.com/apache/incubator-gluten/pull/6623
      - [ ] https://github.com/apache/incubator-gluten/pull/6891
      - [ ] Failed
         - [ ] test posexplode issue: https://github.com/oap-project/gluten/issues/ 1767 - failed
         - [ ] GlutenParquetFilterSuite q2 q9
      - [ ] Fallack
        - [ ] TPCDS 44, 67, 70 - fallback
        - [ ] GLUTEN-1874 not null in one stream - fallback
        - [ ] GLUTEN-1874 not null in both streams - fallback
        - [ ] GLUTEN-4032: fix shuffle read coredump after union - fallback
        - [ ] GlutenClickHouseHiveTableSuite `test 'hive udf'`  => fallback      
      - [ ] GlutenClickhouseCountDistinctSuite `check count distinct with expr fallback` => why **no** fallback
      - [ ] GlutenClickHouseDecimalSuite TPCH Decimal(38,19) Q1 result is not matched if allowPrecisionLoss = true
- [ ] Spark 35 UT passed

zzcclp:
there are some tasks need to be completed.

zzcclp:
DON'T CLOSE THIS ISSUE, there are some tasks need to be completed.

baibaichen:
[...]

PR #6197, opened 2024-06-24
[GLUTEN-6067][CH] [Part 2] Support CH backend with Spark3.5 - Prepare for supporting sink transform

baibaichen:

We refactor the codes for the following purpose:
1. Calling `SparkShim::genExtendedColumnarPostRules`, so we can fallback write to vanilla spark in case of spark 3.5
2. Add `NativeWriteCheck`，Make UT failed in spark 3.5
3. Refactor `LocalExecutor`, moving pipeline building to `SerializedPlanParser`, because we can not add sink transform in this class. 
4. Reducing duplcate codes, i.e. `CHIteratorApi` and  `SubstraitPlanPrinterUtil`

(Fixes: \#6067)

## How was this patch tested?

Existed UTs


zzcclp (approved):
LGTM

Issue #6202, opened 2024-06-25
[VL] No extra SortExec needed before WindowExec if window type is `sort`


WangGuangxin:
### Backend

VL (Velox)

### Bug description

When `spark.gluten.sql.columnar.backend.velox.window.type` is set to `sort`, there should be no extra `SortExec` before `WindowExec` in plan.

### Spark version

None

FelixYBW:
Can you list such a query as example?

JkSelf:
In order to remove the duplicated sort operator, we implement the streaming window for spark in velox. Why need to set the sort based window in your use case? Can you share more about your use cases? Thanks.

WangGuangxin:
> Can you list such a query as example?

Any window query with following conf set to `sort`
--conf spark.gluten.sql.columnar.backend.velox.window.type=sort

for example (lineitem from tpch)
```
select row_number() over (partition by l_suppkey order by l_orderkey) from lineitem 
```


When we set `spark.gluten.sql.columnar.backend.velox.window.type=sort`, velox will use sort based window builder, so there is no need to sort outside `WindowExec` operator, right? 

![image](https://github.com/apache/incubator-gluten/assets/1312321/614544e0-0bd4-4e28-8bcb-785dcae3f8d9)

cc @JkSelf

WangGuangxin:
> In order to remove the duplicated sort operator, we implement the streaming window for spark in velox. Why need to set the sort based window in your use case? Can you share more about your use cases? Thanks.

@JkSelf We are just doing some investigating.  Both `SortExec`  and `WindowExec` need lots of C2R and R2C within the implemention, especially spill occured.  We are hard to do something for `Sort` outside `Window` , but for `Sort` within `Window` ( aka sort based window), maybe we can eliminate some C2R and C2R.  
Do you have any suggestion?

FelixYBW:
> > In order to remove the duplicated sort operator, we implement the streaming window for spark in velox. Why need to set the sort based window in your use case? Can you share more about your use cases? Thanks.
> 
> @JkSelf We are just doing some investigating. Both `SortExec` and `WindowExec` need lots of C2R and R2C within the implemention, especially spill occured. It is hard to do something for `Sort` outside `Window` , but for `Sort` within `Window` ( aka sort based window), maybe we can eliminate some C2R and C2R. Do you have any suggestion?

Why do you need C2R, R2C?

PR #6203, opened 2024-06-25
[GLUTEN-6202][VL] No extra SortExec needed before WindowExec if window type is `sort`

WangGuangxin:

When `spark.gluten.sql.columnar.backend.velox.window.type` is set to `sort`, there should be no extra `SortExec` before `WindowExec` in plan.

(Fixes: \#6202)

## How was this patch tested?

Mannual Test and UT

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
@@ -26,7 +26,7 @@ import org.apache.gluten.utils.{LogLevelUtil, PlanUtil}
 
 import org.apache.spark.api.python.EvalPythonExecTransformer
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, Expression, InputFileBlockLength, InputFileBlockStart, InputFileName, NamedExpression}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, Expression, InputFileBlockLength, InputFileBlockStart, InputFileName, NamedExpression, SortOrder}
 import org.apache.spark.sql.catalyst.optimizer.{BuildLeft, BuildRight, BuildSide}
 import org.apache.spark.sql.catalyst.plans.logical.Join
 import org.apache.spark.sql.execution._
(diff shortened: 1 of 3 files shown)


WangGuangxin:
cc @JkSelf

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -232,6 +233,38 @@ case class OffloadFilter() extends OffloadSingleNode with LogLevelUtil {
...
+      }
+
+      val transformer = replace.doReplace(window)
+      val newChild = transformer.children.head match {
+        case SortExec(_, false, child, _)
+            if outputOrderSatisfied(child, transformer.requiredChildOrdering) =>
+          child
+        case p @ ProjectExec(_, SortExec(_, false, child, _))
+            if outputOrderSatisfied(child, transformer.requiredChildOrdering) =>
+          p.copy(child = child)
```
zhztheplayer, line 253: Any reason not using `p.withNewChildren` to keep tags?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -232,6 +233,38 @@ case class OffloadFilter() extends OffloadSingleNode with LogLevelUtil {
...
+      transformer.withNewChildren(Seq(newChild))
+    case other => other
+  }
+
+  private def outputOrderSatisfied(child: SparkPlan, required: Seq[Seq[SortOrder]]): Boolean = {
+    Seq(child).zip(required).forall {
+      case (child, requiredOrdering) =>
+        SortOrder.orderingSatisfies(child.outputOrdering, requiredOrdering)
+    }
+  }
```
zhztheplayer, line 265: Could just assume an unary node is returned from `replace.doReplace(window)` then we can avoid `zip` to simplify code.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -232,6 +233,38 @@ case class OffloadFilter() extends OffloadSingleNode with LogLevelUtil {
   }
 }
 
+case class OffloadWindow() extends OffloadSingleNode with LogLevelUtil {
```
zhztheplayer, line 236: The optimization should only be done for sort window. Are we missing a condition somewhere?

zhztheplayer (commented):
Thanks.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
```diff
@@ -232,6 +233,35 @@ case class OffloadFilter() extends OffloadSingleNode with LogLevelUtil {
...
+  private val replace = new ReplaceSingleNode()
+
+  override def offload(plan: SparkPlan): SparkPlan = plan match {
+    case window: WindowExec =>
+      if (TransformHints.isNotTransformable(window)) {
+        return window
+      }
+
+      val transformer = replace.doReplace(window)
+      val newChild = transformer.children.head match {
```
zhztheplayer, line 247: Oops, let's either add an assertion like `assert(transformer.children.size == 1)`, or cast it to a unary node.

zhztheplayer (commented):
Just one nit left, thank you for the update.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/OffloadSingleNode.scala
WangGuangxin: ok，got it

zhztheplayer:
And CH CI failed https://opencicd.kyligence.com/job/gluten/job/gluten-ci/10302/, please also take a look. Thanks.

PR #6211, opened 2024-06-25
[VL] Enable log function

zhli1142015:

Enable log function.

## How was this patch tested?

UT.


cpp/velox/substrait/SubstraitParser.cc
```diff
@@ -407,7 +407,8 @@ std::unordered_map<std::string, std::string> SubstraitParser::substraitVeloxFunc
     {"exists", "any_match"},
     {"negative", "unaryminus"},
     {"get_array_item", "get"},
-    {"arrays_zip", "zip"}};
+    {"arrays_zip", "zip"},
+    {"logarithm", "log"}};
```
PHILO-HE, line 411: It looks strange in existing Gluten code that this function is not named to "log", which is Spark's function name. Maybe, just corrected here: 
https://github.com/apache/incubator-gluten/blob/main/shims/common/src/main/scala/org/apache/gluten/expression/ExpressionNames.scala#L178
This may also need the change for CH code. @exmy, does it make sense to you?
https://github.com/apache/incubator-gluten/blob/main/cpp-ch/local-engine/Parser/scalar_function_parser/log.cpp#L41

exmy: It's ok for me, and the `LOG` may need change to `ln` https://github.com/apache/incubator-gluten/blob/de26ed2dad41d2d1e893c8d1b3ae806385d9972f/shims/common/src/main/scala/org/apache/gluten/expression/ExpressionNames.scala#L177

PHILO-HE (approved):
Thanks!

PR #6212, opened 2024-06-25
[VL][Minor] Fix udf jni signature mismatch

marin-ma:
In the previous implementation, Yarn AM container only gets the relative filename. e.g. "libmyudf.so", not a full path. The full path should be constructed using the current working directory.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
@@ -278,6 +278,7 @@ object UDFResolver extends Logging {
       .split(",")
       .map {
         f =>
+          val file = new File(f)
           // Relative paths should be uploaded via --files or --archives
           if (isRelativePath(f)) {
             logInfo(s"resolve relative path: $f")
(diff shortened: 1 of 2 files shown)


marin-ma:
@kecookier Please kindly review. Thanks!

backends-velox/src/main/scala/org/apache/spark/sql/expression/UDFResolver.scala
```diff
@@ -278,21 +278,25 @@ object UDFResolver extends Logging {
...
             logInfo(s"resolve relative path: $f")
             if (isDriver && isYarnClient) {
               throw new IllegalArgumentException(
                 "On yarn-client mode, driver only accepts absolute paths, but got " + f)
             }
-            if (isYarnCluster || isYarnClient) {
-              file
+            if (isDriver && isYarnCluster) {
+              // Yarn AM container only gets the filename in the working directory.
+              new File(System.getProperty("user.dir"), f)
```
kecookier, line 290: I'm having a bit of trouble understanding this piece of code. In yarn-cluster mode, it seems that both the driver and executor can access the UDF library in the working directory. Could you please explain why it is necessary to distinguish the driver separately here? Thank you!

marin-ma: > it seems that both the driver and executor can access the UDF library in the working directory.
That's true on yarn-cluster mode. But the AM container, i.e. driver only gets the filename, while the executor containers gets the absolute file path.

If we only pass the filename to the configuration map, it fails on reading this file on native side.

marin-ma: @kecookier Thanks for your information. Seems like the error I encountered was due to misconfiguration. I will revert this change.

Do you agree with changes in https://github.com/apache/incubator-gluten/pull/6212/commits/6f47f5df71af5f4b88308b4b99d60ee2cc2d0bdc ? Just a minor coding bug fix.

kecookier: > Do you agree with changes in https://github.com/apache/incubator-gluten/commit/6f47f5df71af5f4b88308b4b99d60ee2cc2d0bdc ?

kecookier: > Do you agree with changes in https://github.com/apache/incubator-gluten/commit/6f47f5df71af5f4b88308b4b99d60ee2cc2d0bdc ?

kecookier: > Do you agree with changes in https://github.com/apache/incubator-gluten/commit/6f47f5df71af5f4b88308b4b99d60ee2cc2d0bdc ?  

The change looks good to me. Thanks.

kecookier (approved):
LGTM. Thanks!

PR #6228, opened 2024-06-26
[VL] Make jni debug workspace configurable

Yohahaha:

Add new config `spark.gluten.sql.debug.keepJniWorkspaceDir` to store libraries, and decouple the functionality from `spark.gluten.sql.debug`

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/vectorized/JniWorkspace.java
+++ b/gluten-core/src/main/java/org/apache/gluten/vectorized/JniWorkspace.java
@@ -75,14 +75,14 @@ public class JniWorkspace {
     }
   }
 
-  public static void enableDebug() {
+  public static void enableDebug(String debugDir) {
     // Preserve the JNI libraries even after process exits.
     // This is useful for debugging native code if the debug symbols were embedded in
     // the libraries.
     synchronized (DEFAULT_INSTANCE_INIT_LOCK) {
       if (DEBUG_INSTANCE == null) {
         final File tempRoot =
-            Paths.get("/tmp").resolve("gluten-jni-debug-" + UUID.randomUUID()).toFile();
+            Paths.get(debugDir).resolve("gluten-jni-debug-" + UUID.randomUUID()).toFile();
         try {
           FileUtils.forceMkdir(tempRoot);
         } catch (IOException e) {
--- a/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
@@ -391,8 +391,7 @@ class GlutenConfig(conf: SQLConf) extends Logging {
     conf.getConf(COLUMNAR_VELOX_MEMORY_USE_HUGE_PAGES)
 
   def debug: Boolean = conf.getConf(DEBUG_ENABLED)
-  def debugKeepJniWorkspace: Boolean =
-    conf.getConf(DEBUG_ENABLED) && conf.getConf(DEBUG_KEEP_JNI_WORKSPACE)
+  def debugKeepJniWorkspace: Boolean = conf.getConf(DEBUG_KEEP_JNI_WORKSPACE)
   def taskStageId: Int = conf.getConf(BENCHMARK_TASK_STAGEID)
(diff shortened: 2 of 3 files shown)


gluten-core/src/main/java/org/apache/gluten/vectorized/JniWorkspace.java
```diff
@@ -75,14 +75,14 @@ private static JniWorkspace createDefault() {
     }
   }
 
-  public static void enableDebug() {
+  public static void enableDebug(String debugDir) {
```
zhztheplayer, line 78: What happens if `debugDir` changed by user in session?

We either handle the case or just raise an error (by making the option static or adding assertion in code).

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1585,6 +1585,12 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(false)
 
+  val DEBUG_KEEP_JNI_WORKSPACE_DIR =
+    buildConf(GLUTEN_DEBUG_KEEP_JNI_WORKSPACE_DIR)
+      .internal()
+      .stringConf
+      .createWithDefault("/tmp")
```
zhztheplayer, line 1592: I think we can just make the conf static, as well as `DEBUG_KEEP_JNI_WORKSPACE`. Thanks.

gluten-core/src/main/java/org/apache/gluten/vectorized/JniWorkspace.java
zhztheplayer: @Yohahaha Let's still add an assertion to make sure the function doesn't get call twice with different `debugDir` values. In case the method is used in test code or somewhere. Thanks.

That may involve adding another global variable but it could be worth it I guess.

Yohahaha: so you think `DEBUG_INSTANCE` will be initialized twice in one process? if yes, we should refactor jni initialization first.

zhztheplayer: > so you think DEBUG_INSTANCE will be initialized twice in one process

Noop. Just a protection in this method assuming unpredictable caller code. E.g., One could call `enableDebug(...)` for test purpose. We'd throw something otherwise the second call will be silently swalloed.

Yohahaha: this functionality should only be called once by design from my understanding, if someone want call it twice, I think they dont know that they did. We dont really need protect anything.
Related conf already be static and `enableDebug` has ensure only initialize `DEBUG_INSTANCE` once.

zhztheplayer: > I think they dont know that they did. We dont really need protect anything.

I think anti cases are appearing easier than we think it could be.

For example, if someone in test code, calling `JniWorkspace.enableDebug("folder1") for debug purpose`, however in somewhere a Spark config "spark.gluten.sql.debug.keepJniWorkspaceDir="folder2" was already set for some reason.

Then he could go mad because the files was not found in folder1 and the console doesn't tell him anything about that.

Yohahaha:
Run Gluten Clickhouse CI

gluten-core/src/main/java/org/apache/gluten/vectorized/JniWorkspace.java
liujiayi771: As a debug conf, I feel that we don't need to provide excessive protection for the user.

liujiayi771 (approved):
LGTM. Thank you for adding this configuration, it is very helpful for troubleshooting.

gluten-core/src/main/java/org/apache/gluten/vectorized/JniWorkspace.java
zhztheplayer: `enableDebug("folder1")` should just enable debug in folder `folder1`. If there's potential that it doesn't use `folder1`, please give user/caller a note.

I don't think it's excessive protection. Next time please close comments like this one before merging a PR.

PR #6229, opened 2024-06-26
[VL] Support building arrow CPP and finding installed arrow libs from system

PHILO-HE:

This is a part of work to let Gluten/Velox use installed arrow libraries instead of always
building from source, which is for simplifying Gluten/Velox build and saving CI time.

We also have a velox PR to let velox find and use the installed arrow libraries: https://github.com/facebookincubator/velox/pull/9992.

## How was this patch tested?

CI.


zhouyuan (approved):

jinchengchenghh:
Please also update the document, thanks!

PHILO-HE:
> Please also update the document, thanks!

@jinchengchenghh, I will do that. Thanks!

jinchengchenghh:
And we don't need to add arrow patch here https://github.com/apache/incubator-gluten/blob/main/ep/build-velox/src/get_velox.sh#L249

PHILO-HE:
> And we don't need to add arrow patch here https://github.com/apache/incubator-gluten/blob/main/ep/build-velox/src/get_velox.sh#L249

@jinchengchenghh, you mean we don't need this patch here or don't need to use git to add it? The git add may be still useful for checking ep cache I guess.

PR #6233, opened 2024-06-26
[Core] Rename isTransformable API to maybeTransformable

gaoyangxiaozhu:

[(Fixes: \#6234)

## How was this patch tested?

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
@@ -78,20 +78,6 @@ object TransformHints {
     }
   }
 
-  /**
-   * NOTE: To be deprecated. Do not create new usages of this method.
-   *
-   * Since it's usually not safe to consider a plan "transformable" during validation phase. Another
-   * validation rule could turn "transformable" to "non-transformable" before implementing the plan
-   * within Gluten transformers.
-   */
-  def isTransformable(plan: SparkPlan): Boolean = {
-    getHintOption(plan) match {
-      case None => true
-      case _ => false
-    }
-  }
-
   def tag(plan: SparkPlan, hint: TransformHint): Unit = {
     val mergedHint = getHintOption(plan)
       .map {
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
@@ -78,6 +78,14 @@ object TransformHints {
     }
   }
 
+  /**
+   * If true, it implies the plan maybe transformable during validation phase but not guaranteed,
+   * since another validation rule could turn it to "non-transformable" before implementing the plan
+   * within Gluten transformers. If false, the plan node will be guaranteed fallback to Vanilla plan
+   * node while being implemented.
+   */
+  def maybeTransformable(plan: SparkPlan): Boolean = !isNotTransformable(plan)
+
   def tag(plan: SparkPlan, hint: TransformHint): Unit = {
     val mergedHint = getHintOption(plan)
       .map {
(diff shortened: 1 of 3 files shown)


gaoyangxiaozhu:
@rui-mo / @PHILO-HE / @zhztheplayer please help review

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
```diff
@@ -78,20 +78,6 @@ object TransformHints {
     }
   }
 
-  /**
-   * NOTE: To be deprecated. Do not create new usages of this method.
-   *
-   * Since it's usually not safe to consider a plan "transformable" during validation phase. Another
-   * validation rule could turn "transformable" to "non-transformable" before implementing the plan
-   * within Gluten transformers.
-   */
```
zhztheplayer, line 87: My comment might be misleading but, changing code from `if (isTransformable)` to `if (!isNotTransformable)` doesn't seem to be the right way to deal with the problem...

Because neither `isTransformable` nor `!isNotTransformable` ensures the plan node is able to offload. This kind of assertions are unreliable. Only `isNotTransformable` ensures that the plan node is not offload-able.

zhztheplayer (changes requested):
Let's see if we can find a better way than this solution.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
PHILO-HE: I have the same feeling.

gaoyangxiaozhu: hey @zhztheplayer / @PHILO-HE  I actually got your point before do the change.

The purpose of this PR is not to introduce any logical improvements or refactoring but solely to clean up the deprecated `isTransformable` API.

As you've mentioned, both `!isNotTransformable` and `isTransformable` do not necessarily guarantee that the plan node is able to offload; rather, they imply that the plan *may be offloadable*.

For the changes in this PR involving converting `isTransformable` to `!isNotTransformable`, this makes sense as it conveys that "the plan may still be offloadable." Thus, we can attempt the code path to `may rewrite` or `may offload`.
And isNotTransformable is more suitable for conveying that "the plan may still be offloadable."

Overall, this PR focuses on:
1. Cleaning up the isTransformable API.
2. Changing statements where the check means "the plan may be offloadable" to `!isNotTransformable`, which should be appropriate.

And actually I knew that we can for example construct a new hit instance to apply and check for the plan ensures the plan is offloadable. However, that is not the case for this PR change.

gaoyangxiaozhu:
> Let's see if we can find a better way than this solution.

hey @zhztheplayer thanks, could you check my above comment ?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/TransformHintRule.scala
zhztheplayer: Sounds reasonable to me.

However I will hardly get the intention of using `!isNotTransformable` without reading your above comments. 

Do you think we can just rename `isTransformable` to `maybeTransformable` or something? Would it sound more straightforward for you?

gaoyangxiaozhu: sounds a good suggestion @zhztheplayer Hongze, let me do the change. thank you.

gaoyangxiaozhu: done

zhztheplayer (approved):
Thanks!

PHILO-HE (approved):
Thanks!

gaoyangxiaozhu:
Run Gluten Clickhouse CI

FelixYBW:
Looks good

PR #6243, opened 2024-06-27
[CORE] Execution runtime / native memory manager refactor

zhztheplayer:
The patch introduces a new design for (Velox backend) execution runtime and native memory manager. The new design will make the relevant components much clearer and simpler than before both in concept and code. 

The following figure will demonstrate the new design in detail:

New Runtime Arch

In the new design, native memory manager instances will be 1:1 bound to execution runtime instances. Objects stores are still in runtime instances to make sure all the allocated contextual objects don't escape their resident runtimes' lifecycle scope.

It's no longer needed to create memory manager instances from JVM side as they will be managed internally in runtime instances. Spillers will be accounted in a list in runtime and will no longer need to be created during instantiating memory managers.

JVM side `MemoryManager` and `MemoryAllocator` APIs are removed as no longer used.


zhztheplayer:
Issues should be followed up:

1. OOM CI steps should be untangled, some of them were marked `to be fixed` but have been passing for sometime
2. Should use a global instance pool used by `ColumnarBatches#create` to avoid accountant ownerships on different runtime

JkSelf (approved):

Issue #6253, opened 2024-06-27
[VL] Using Velox udf will fail with yarn-client mode when driver and executors are not on the same host


marin-ma:
### Backend

VL (Velox)

### Bug description

Command to run:

bin/spark-submit --master yarn --files /home/sparkuser/libmyudf.so  --jars /home/sparkuser/gluten-velox-bundle-spark3.5_2.12-centos_7_x86_64-1.2.0-SNAPSHOT.jar --deploy-mode client --num-executors 1 --conf spark.gluten.sql.columnar.backend.velox.driver.udfLibraryPaths=file:///home/sparkuser/libmyudf.so /home/sparkuser/udf.py

udf.py:
```
from pyspark.sql import SparkSession, SQLContext
from pyspark import SparkConf, SparkContext
import time

if __name__ == "__main__":

    conf = SparkConf()\
        .set("spark.driver.extraClassPath", "gluten-velox-bundle-spark3.5_2.12-centos_7_x86_64-1.2.0-SNAPSHOT.jar")\
        .set("spark.executor.extraClassPath", "gluten-velox-bundle-spark3.5_2.12-centos_7_x86_64-1.2.0-SNAPSHOT.jar")\
        .set("spark.plugins", "org.apache.gluten.GlutenPlugin")\
        .set("spark.memory.offHeap.size", "20g")\
        .set("spark.memory.offHeap.enabled", "true")\
        .set("spark.gluten.sql.columnar.backend.velox.udfLibraryPaths", "libmyudf.so")

    sc = SparkContext(conf = conf)

    spark = SQLContext(sc)
    print(spark.sql("select myudf1(1L)").collect())
    time.sleep(10)

```

Exception in executor's log

```
24/06/27 08:57:43 ERROR NativeBackendInitializer: Failed to call native backend's initialize method
java.lang.RuntimeException: File path not exists: /mnt/DP_disk5/sparkuser/hdfs/yarn/local/gluten-b6794efd-41d3-4033-b057-fea34344e204/jni/60db35df-1086-4034-bc73-b46f88a0b7e8/gluten-837934689327669466/libmyudf.so
	at org.apache.gluten.init.NativeBackendInitializer.initialize(Native Method)
	at org.apache.gluten.init.NativeBackendInitializer.initialize0(NativeBackendInitializer.java:61)
	at org.apache.gluten.init.NativeBackendInitializer.initializeBackend(NativeBackendInitializer.java:48)
[...]

PR #6255, opened 2024-06-27
[GLUTEN-6253] Use internal udf config to avoid modify the original one

marin-ma:
The current implementation sets `spark.gluten.sql.columnar.backend.velox.udfLibraryPaths` on driver side after resolving the library paths. This approach can overwrite the original settings with a local file path on the driver node before sending the SparkConf to all executors, and the executors on different nodes will fail while accessing that path.

This PR sets the resolved library paths to an internal config to avoid the conflicts.

Manually verified on a multi-node cluster.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -65,6 +65,7 @@ object VeloxBackendSettings extends BackendSettingsApi {
 
   val GLUTEN_VELOX_UDF_LIB_PATHS = getBackendConfigPrefix() + ".udfLibraryPaths"
   val GLUTEN_VELOX_DRIVER_UDF_LIB_PATHS = getBackendConfigPrefix() + ".driver.udfLibraryPaths"
+  val GLUTEN_VELOX_INTERNAL_UDF_LIB_PATHS = getBackendConfigPrefix() + ".internal.udfLibraryPaths"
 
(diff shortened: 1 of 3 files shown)


kecookier:
Hi @marin-ma,  I have some concerns. In yarn-client mode, we may set an extra configuration `--conf spark.gluten.sql.columnar.backend.velox.driver.udfLibraryPaths=file:///path/to/libmyudf.so`  to let the driver read the local UDF library. The complete configuration is as follows:
```
# Needed for Yarn client mode
--conf spark.gluten.sql.columnar.backend.velox.driver.udfLibraryPaths=file:///path/to/libmyudf.so
```
It seems like spark.gluten.sql.columnar.backend.velox.internal.udfLibraryPaths has the same meaning. Could you please clarify this?

marin-ma:
@kecookier That's the exact configuration I used in https://github.com/apache/incubator-gluten/issues/6253

`spark.gluten.sql.columnar.backend.velox.udfLibraryPaths` and `spark.gluten.sql.columnar.backend.velox.driver.udfLibraryPaths` are the configurations exposed to users. `spark.gluten.sql.columnar.backend.velox.internal.udfLibraryPaths` is used as an internal configuration to save the resolved udf library path and is used on native side for loading the libraries. https://github.com/apache/incubator-gluten/pull/6255/files#diff-a633836e086157189ba6a590a79c4823e73937f896ecab711a324586e13aa73aL102-R102

kecookier:
Thanks for your clarification, it looks good to me.

kecookier (approved):

PR #6256, opened 2024-06-27
[CH] Fix array distinct core dump 

liuneng1994:

fix core dump when arrayDistinct(Array(Nullable(Decimal)))

## How was this patch tested?

unit tests


zzcclp (approved):
LGTM

PR #6259, opened 2024-06-27
[CH] Support bit_length/octet_length function

exmy:

* support `bit_length/octet_length` function

* move `length` impl to `funciton parser`

These three functions may receive `int` type argument in spark ut, add a `cast to string`.

## How was this patch tested?

PASS CI


gluten-ut/spark32/src/test/scala/org/apache/gluten/utils/clickhouse/ClickHouseTestSettings.scala
```diff
@@ -438,6 +438,9 @@ class ClickHouseTestSettings extends BackendTestSettings {
     .exclude("string overlay function")
     .exclude("binary overlay function")
     .exclude("string parse_url function")
+    .exclude("string / binary length function")
+    .exclude("SPARK-36751: add octet length api for scala")
+    .exclude("SPARK-36751: add bit length api for scala")
```
exmy, line 443: exlcude these three uts due to diff for `cast(2.0f to string)`

liuneng1994 (approved):
LGTM

PR #6266, opened 2024-06-28
[CORE] Remap the name of LOG/LOGARITHM

exmy:
## How was this patch tested?

PASS CI


PHILO-HE (approved):
Thanks so much for your fix! cc @zhli1142015

zhli1142015 (approved):
Thanks.

PR #6294, opened 2024-07-01
[CORE] Drop redundant partial sort which has pre-project when offload sortAgg

zml1206:
As titile.

## How was this patch tested?
UT


zml1206:
Run Gluten Clickhouse CI

gluten-core/src/main/scala/org/apache/gluten/execution/SortUtils.scala
```diff
@@ -23,11 +23,27 @@ import org.apache.spark.sql.execution.{ProjectExec, SortExec, SparkPlan}
 object SortUtils {
   def dropPartialSort(plan: SparkPlan): SparkPlan = plan match {
     case RewrittenNodeWall(p) => RewrittenNodeWall(dropPartialSort(p))
-    case sort: SortExec if !sort.global => sort.child
+    case PartialSortChild(p) => p
```
zhztheplayer, line 26: Why we only drop partial sorts? I know it's previous code but I am still curious about that.

Also I'd suggest renaming:

`PartialSortChild(p)` -> `PartialSortLike(child)`
`ProjectChild(p)` -> `ProjectLike(child)`

zml1206: Sort generated by `EnsureRequirements`.
https://github.com/apache/spark/blob/1bfc9c3cbdc5d017fef68b0e84dd7fd22d2fef0f/sql/core/src/main/scala/org/apache/spark/sql/execution/exchange/EnsureRequirements.scala#L204

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxAggregateFunctionsSuite.scala
```diff
@@ -50,6 +50,32 @@ abstract class VeloxAggregateFunctionsSuite extends VeloxWholeStageTransformerSu
       .set("spark.sql.sources.useV1SourceList", "avro")
   }
 
+  test("drop redundant partial sort which has pre-project when offload sortAgg") {
```
zhztheplayer, line 53: Would you like to move this test case down, to the bottom of this file, probably? Thanks.

zhztheplayer (approved):
Just some nits. Thank you.

zhztheplayer (approved):

PR #6295, opened 2024-07-01
[VL] CI: Fix CPP tests are not running

zhztheplayer:
This is a follow up to https://github.com/apache/incubator-gluten/pull/5768 in which we may missed enabling CPP tests.


zhztheplayer:
@PHILO-HE

PHILO-HE (approved):
Thanks for your fix!

PR #6297, opened 2024-07-01
[VL] Disable protobuf build by default

PHILO-HE:

Disable protobuf by default as it is not required for each build.
Also removed the build protobuf option from get_velox.sh since it is confusing and unnecessary. This option just controls whether to change velox's setup script. So it depends on the enabling of --run_setup_script.

## How was this patch tested?

CI.

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -13,7 +13,7 @@ BUILD_TESTS=OFF
 BUILD_EXAMPLES=OFF
 BUILD_BENCHMARKS=OFF
 BUILD_JEMALLOC=OFF
-BUILD_PROTOBUF=ON
+BUILD_PROTOBUF=OFF
 BUILD_VELOX_TESTS=OFF
 BUILD_VELOX_BENCHMARKS=OFF
 ENABLE_QAT=OFF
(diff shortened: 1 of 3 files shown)


FelixYBW:
when do we need run_setup_script to be enabled? is it used for dynamic link only, right?

PHILO-HE:
> when do we need run_setup_script to be enabled? is it used for dynamic link only, right?

@FelixYBW, yes, vcpkg build doesn't require this script.

rui-mo (approved):

FelixYBW:
with ep_cache=on, the protobuf is build every time when I build Velox. The PR can solve the issue, right?

PHILO-HE:
> with ep_cache=on, the protobuf is build every time when I build Velox. The PR can solve the issue, right?

This pr addresses the repeated build of protobuf by Gluten. If it's Velox that builds protobuf every time, it's because the required protobuf is not installed into system. Does it happen in vcpkg build?

FelixYBW:
> Does it happen in vcpkg build?

Yes

PHILO-HE:
> > Does it happen in vcpkg build?
> 
> Yes

@FelixYBW, as enable_ep_cache=ON, I guess it's Gluten that triggers protobuf build every time. This pr should have fixed it already. If protobuf is still built every time, please let me know.

PR #6307, opened 2024-07-02
[VL] Add isStreamingAgg info to HashAggregateTransformer

liujiayi771:

Add isStreamingAgg info to the toString method of HashAggregateBaseTransformer, so that we can easily determine the current type of the Aggregate.

## How was this patch tested?
N/A.


liujiayi771:
cc @ulysses-you, thanks.

liujiayi771:
By adding this info, we can determine if #3828 is effective.

image

ulysses-you (approved):
lgtm,  thank you

Issue #6317, opened 2024-07-03
[CORE] Condition store the actual executed filter expressions


zml1206:
### Description

Currently, condition still holds the filter expression pushed down to scan, and the real condition is obtained through getRemainingCondition.
Condition store the actual executed filter expressions will make the plan clearer, and after this we can easier to remove filter whose all conditions have been pushed down to scan.

PR #6319, opened 2024-07-03
[GLUTEN-6317][CORE] Condition store the actual executed filter expressions

zml1206:

Fixes: \#6317

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala
@@ -16,30 +16,11 @@
  */
 package org.apache.gluten.execution
 
-import org.apache.spark.sql.catalyst.expressions.{And, Expression}
+import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.execution.SparkPlan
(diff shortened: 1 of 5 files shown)


zml1206:
There are some problems if condition store the actual executed filter expressions. Filter may have different output nullability,  It will affect DPP anti join. Spark also retains the condition that is pushed down to scan.

backends-velox/src/main/scala/org/apache/gluten/execution/FilterExecTransformer.scala
```diff
@@ -16,30 +16,11 @@
...
 package org.apache.gluten.execution
 
-import org.apache.spark.sql.catalyst.expressions.{And, Expression}
+import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.execution.SparkPlan
 
 case class FilterExecTransformer(condition: Expression, child: SparkPlan)
   extends FilterExecTransformerBase(condition, child) {
-  // FIXME: Should use field "condition" to store the actual executed filter expressions.
-  //  To make optimization easier (like to remove filter when it actually does nothing)
```
zml1206, line 25: Should we remove this FIXME? @zhztheplayer

PR #6324, opened 2024-07-03
[VL] RAS: Remove NoopFilter if has same output with child

zml1206:

As title


gluten-core/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RemoveFilter.scala
```diff
@@ -41,9 +41,13 @@ object RemoveFilter extends RasRule[SparkPlan] {
   override def shift(node: SparkPlan): Iterable[SparkPlan] = {
     val filter = node.asInstanceOf[FilterExecTransformerBase]
     if (filter.isNoop()) {
-      val out = NoopFilter(filter.child, filter.output)
-      out.copyTagsFrom(filter)
-      return List(out)
+      if (filter.output == filter.child.output) {
+        return List(filter.child)
```
zhztheplayer, line 45: Let’s invert the if condition. Also avoid adding ‘else’ clause as ‘return’ is already used.

zhztheplayer (approved):

PR #6341, opened 2024-07-05
[VL] Make velox writer queue size configurable

yikf:
This PR aims to make velox writer queue size is configurable to avoid oom from being caused by too many batches.

## How was this patch tested?
GA

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/VeloxColumnarBatchIterator.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/VeloxColumnarBatchIterator.scala
@@ -26,10 +26,10 @@ import org.apache.arrow.vector.types.pojo.Schema
 
 import java.util.concurrent.{ArrayBlockingQueue, TimeUnit}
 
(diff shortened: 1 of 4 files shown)


yikf:
cc @PHILO-HE

PHILO-HE:
@yikf, it looks good to me. Did you observe that this config can help relieve the OOM issue on your side?

BTW, please fix code style issue, see reference:
https://github.com/apache/incubator-gluten/blob/main/docs/developers/NewToGluten.md#javascala-code-style

yikf:
> BTW, please fix code style issue, see reference:
> https://github.com/apache/incubator-gluten/blob/main/docs/developers/NewToGluten.md#javascala-code-style

thanks for the reminding of codestyle.



> it looks good to me. Did you observe that this config can help relieve the OOM issue on your side?

yes, we have a scenario which trigger an oom is due to a large number of batches being held in a queue. making queue configurable may reduce oom occurrence, but it will also block the query thread, which is not optimal, but it is the fastest way to prevent oom. then we can implement push-based writing to refactor the logic of this piece

PHILO-HE (approved):

PR #6363, opened 2024-07-07
[VL] Fix process_setup_alinux3 arrow CMakeLists.txt path

liujiayi771:

Velox moved `third_party/CMakeLists.txt` to `CMake/resolve_dependency_modules/arrow/CMakeLists.txt`. The original `sed` command will give an error. Upon verification, it is now confirmed that there is no need to modify `${CMAKE_INSTALL_LIBDIR}`, as its value is already set to `lib64`.

## How was this patch tested?

N/A


liujiayi771:
@Yohahaha Could you help to check this change?

Yohahaha (approved):

PR #6367, opened 2024-07-08
[DNM] test parquet complex datatype

zhouyuan:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
@@ -1979,7 +1979,7 @@ object GlutenConfig {
       .internal()
       .doc("Force fallback for complex type scan, including struct, map, array.")
       .booleanConf
-      .createWithDefault(true)
+      .createWithDefault(false)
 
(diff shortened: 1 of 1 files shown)


zhouyuan:
note 4 unit tests failed
```
char type values should be padded: nested in map value *** FAILED ***
char type values should be padded: nested in both map key and value *** FAILED ***
char type values should be padded: nested in map key *** FAILED ***
 SPARK-18053: ARRAY equality is broken *** FAILED ***
```

FelixYBW:
@yma11 Can you report the issue to Velox thread?

yma11:
> @yma11 Can you report the issue to Velox thread?

Sure. Let me take a look at the root cause first.

FelixYBW:
@yma11 if all spark UT of complex parquet scan passed, let's remove the flag: spark.gluten.sql.complexType.scan.fallback.enabled

From customer side, we already verified the complex type works

yma11:
The previous issue is valid and I proposed a [fix](https://github.com/facebookincubator/velox/issues/10490) in upstream. I will remove the flag once the fix accepted.

PR #6380, opened 2024-07-09
[CELEBORN] Add compression for row-based shuffle

kerwin-zk:
When using ```CelebornShuffleManager```, columnar shuffle data is typically compressed on the Gluten side and sent directly to Celeborn, so compression is not needed on the Celeborn side. Therefore, ```spark.celeborn.client.shuffle.compression.codec``` is set to none. However, when row-based shuffle occurs, this configuration causes Celeborn to skip compression, leading to data bloat.

## How was this patch tested?

CI

### DIFF
--- a/gluten-celeborn/common/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
+++ b/gluten-celeborn/common/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
@@ -84,6 +84,8 @@ public class CelebornShuffleManager implements ShuffleManager {
 
   private final SparkConf conf;
   private final CelebornConf celebornConf;
+  private final SparkConf rowBasedConf;
+  private final CelebornConf rowBasedCelebornConf;
   // either be "{appId}_{appAttemptId}" or "{appId}"
   private String appUniqueId;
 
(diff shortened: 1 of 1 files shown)


gluten-celeborn/common/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
```diff
@@ -132,6 +142,11 @@ private SparkShuffleManager vanillaCelebornShuffleManager() {
     if (_vanillaCelebornShuffleManager == null) {
       synchronized (this) {
         if (_vanillaCelebornShuffleManager == null) {
+          if ("none"
+              .equalsIgnoreCase(
+                  conf.get(SPARK_CELEBORN_COMPRESSION_CODEC_KEY, celebornDefaultCodec))) {
+            conf.set(SPARK_CELEBORN_COMPRESSION_CODEC_KEY, celebornDefaultCodec);
```
ulysses-you, line 148: Shall we use a new spark conf copy to hold the configs for row-based celeborn shuffle manager ? I think it's risk to share the configs for both row-based and columnar shuffle manager, for example, the first shuffle is row-based then the conf will overwrite, and later the columnar shuffle manager would use the overwritten conf.

gluten-celeborn/common/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
```diff
@@ -330,14 +351,18 @@ public <K, C> ShuffleReader<K, C> getReader(
     if (handle instanceof CelebornShuffleHandle) {
       @SuppressWarnings("unchecked")
       CelebornShuffleHandle<K, ?, C> h = (CelebornShuffleHandle<K, ?, C>) handle;
+      CelebornConf readerConf = celebornConf;
+      if (_vanillaCelebornShuffleManager != null) {
+        readerConf = rowBasedCelebornConf;
```
ulysses-you, line 356: When the `_vanillaCelebornShuffleManager` is null if we get row-based, can it happen ?

kerwin-zk: Yes, when gettingReader, ```vanillaCelebornShuffleManager``` is usually null. I am looking to see if there is any way to pass this message.

kerwin-zk: @ulysses-you cc

ulysses-you (approved):
thank you @kerwin-zk

PR #6381, opened 2024-07-09
[CORE] Remove local sort for TopNRowNumber

ulysses-you:

This pr adds a new rule `EliminateLocalSort` to unify the existed drop local sort code, and also correct `TopNRowNumber` operator require child ordering to help remove unneceesary local sort.

For now, the unnecessary local sort would happen if:
- Convert sort merge join to shuffled hash join
- Offload SortAggregate to native hash aggregate
- Offload WindowGroupLimit to native TopNRowNumber
- The columnar window type is `sort`

## How was this patch tested?

add test & pass tests


ulysses-you:
@GlutenPerfBot benchmark

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/EliminateLocalSort.scala
```diff
@@ -0,0 +1,80 @@
...
+  def unapply(plan: SparkPlan): Option[(SparkPlan, SparkPlan)] = {
+    plan match {
+      case p1 @ ProjectExec(_, SortExecTransformer(_, false, p2: ProjectExec, _))
+          if p1.outputSet == p2.child.outputSet =>
+        Some((p1, p2.child))
+      case p1 @ ProjectExecTransformer(
+            _,
+            SortExecTransformer(_, false, p2: ProjectExecTransformer, _))
+          if p1.outputSet == p2.child.outputSet =>
+        Some((p1, p2.child))
```
zml1206, line 72: Is there any more
```
      case p1 @ ProjectExec(_, SortExec(_, false, p2: ProjectExec, _))
          if p1.outputSet == p2.child.outputSet =>
        Some((p1, p2.child))
      case p1 @ProjectExecTransformer(
            _,
            SortExec(_, false, p2: ProjectExecTransformer, _))
          if p1.outputSet == p2.child.outputSet =>
        Some((p1, p2.child))
```

ulysses-you: I think it won't happen. We only pull out project from sort if the sort is offloaded, so if there the pre/post project should only happen with SortExecTransformer.

gluten-core/src/main/scala/org/apache/gluten/execution/SortUtils.scala
```diff
@@ -1,49 +0,0 @@
...
-import org.apache.spark.sql.execution.{ProjectExec, SortExec, SparkPlan}
-
-object SortUtils {
-  def dropPartialSort(plan: SparkPlan): SparkPlan = plan match {
-    case RewrittenNodeWall(p) => RewrittenNodeWall(dropPartialSort(p))
-    case PartialSortLike(child) => child
-    // from pre/post project-pulling
-    case ProjectLike(PartialSortLike(ProjectLike(child))) if plan.outputSet == child.outputSet =>
-      child
-    case ProjectLike(PartialSortLike(child)) => plan.withNewChildren(Seq(child))
```
zml1206, line 30: This logic is lost.

ulysses-you: thank you @zml1206 for the reminder

ulysses-you:
@GlutenPerfBot benchmark

ulysses-you:
@GlutenPerfBot benchmark

ulysses-you:
cc @zhztheplayer @JkSelf @zml1206 thank you

gluten-core/src/main/scala/org/apache/gluten/execution/WindowExecTransformer.scala
```diff
@@ -67,14 +67,10 @@ case class WindowExecTransformer(
...
-      BackendsApiManager.getSettings.requiredChildOrderingForWindow()
-      && GlutenConfig.getConf.veloxColumnarWindowType.equals("streaming")
-    ) {
-      // Velox StreamingWindow need to require child order.
-      Seq(partitionSpec.map(SortOrder(_, Ascending)) ++ orderSpec)
+    if (GlutenConfig.getConf.veloxColumnarWindowType.equals("sort")) {
+      super.requiredChildOrdering
     } else {
-      Seq(Nil)
+      Seq(partitionSpec.map(SortOrder(_, Ascending)) ++ orderSpec)
```
zml1206, line 73: There seems have two problems, first, super.requiredChildOrdering = Seq(partitionSpec.map(SortOrder(_, Ascending)) ++ orderSpec), in addition, according to the original logic, ck backend should not need child order.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/EliminateLocalSort.scala
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * This rule is used to eliminate unnecessary local sort.
+ *
+ * This could happen if:
+ *   - Convert sort merge join to shuffled hash join
+ *   - Offload SortAggregate to native hash aggregate
+ *   - Offload WindowGroupLimit to native TopNRowNumber
+ *   - The columnar window type is `sort`
+ */
+object EliminateLocalSort extends Rule[SparkPlan] {
```
JkSelf, line 34: @ulysses-you Do we have change to eliminate the local sort for VeloxColumnarWriteFilesExec here?

gluten-core/src/main/scala/org/apache/gluten/execution/WindowExecTransformer.scala
ulysses-you: for 1, it's a good point, it seems the `WindowExecBase` has implemented these methods since 3.3
for 2, the ck backend requires child ordering and before we did not remove local sort before window

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/EliminateLocalSort.scala
ulysses-you: There are some limitation to remove sort before write files, e.g., we can not remove the sort which is created by user (to improve compression ratio). We can only remove the sort which is added from Spark, but it's hard to find that sort in this new rule only using require child ordering framework, we need to check if all sort fields are partition columns.

gluten-core/src/main/scala/org/apache/gluten/execution/WindowExecTransformer.scala
zml1206: I think we should retain requiredChildOrderingForWindow instead of using velox configuration to control ck backend  
although it has no effect by default value.

ulysses-you:
The ck failure is irrelevant. @zml1206 @JkSelf any more comments ?

zml1206:
LGTM

yaooqinn (approved):

PR #6403, opened 2024-07-11
[GLUTEN-6067][VL] [Part 3-1] Refactor: Rename VeloxColumnarWriteFilesExec to ColumnarWriteFilesExec

baibaichen:

(Fixes: \#6067)

This PR Refactors Velox side code,  rename `VeloxColumnarWriteFilesExec` to `GlutenColumnarWriteFilesExec`, move it to gluten-core, so that Clickhouse backend can use the same SparkPlan in the followup PR.

By supporting spark 3.4, Velox supports whole stage native write pipeline which is better than old implementation, clickhouse backend also adopt such implementation.

### Major change 1
The only major difference between velox and clichouse is how to parse native metrics. which I introduce a new trait called `BackendWrite`, it only has one member now. Once native write pipeline is compeleted,  we get it by `BackendsApiManager.getSparkPlanExecApiInstance.createBackendWrite`, Please see `VeloxBackendWrite` for details

``` scala
trait BackendWrite {
  def collectNativeWriteFilesMetrics(batch: ColumnarBatch): Option[WriteTaskResult]
}
```
### Minor change 2
The other minor diffierence is clickhose backend doesn't generate filename. To compute filename per task, it uses  `HadoopMapReduceCommitProtocol::getFilename`, and then injects them to backend. This is ok because Velox doesn't support `maxRecordsPerFile`, see https://github.com/apache/incubator-gluten/pull/4329 and clickhouse backend also follow this,  which means one task only produce one file, no need more injections.

### Improve
I also pass File Format to backed.

## How was this patch tested?

Uisng Existed UTs.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/sql/execution/GlutenColumnarWriteFilesExec.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/execution/GlutenColumnarWriteFilesExec.scala
@@ -17,149 +17,46 @@
 package org.apache.spark.sql.execution
 
 import org.apache.gluten.backendsapi.BackendsApiManager
-import org.apache.gluten.columnarbatch.ColumnarBatches
 import org.apache.gluten.exception.GlutenException
 import org.apache.gluten.extension.GlutenPlan
-import org.apache.gluten.memory.arrow.alloc.ArrowBufferAllocators
 import org.apache.gluten.sql.shims.SparkShimLoader
 
 import org.apache.spark.{Partition, SparkException, TaskContext, TaskOutputFileAlreadyExistException}
 import org.apache.spark.internal.io.{FileCommitProtocol, SparkHadoopWriterUtils}
-import org.apache.spark.internal.io.FileCommitProtocol.TaskCommitMessage
 import org.apache.spark.rdd.RDD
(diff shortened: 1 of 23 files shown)


gluten-core/src/main/scala/org/apache/spark/sql/execution/GlutenColumnarWriteFilesExec.scala
```diff
@@ -17,150 +17,47 @@
...
+trait BackendWrite {
+  def collectNativeWriteFilesMetrics(batch: ColumnarBatch): Option[WriteTaskResult]
+}
 
 /**
  * This RDD is used to make sure we have injected staging write path before initializing the native
  * plan, and support Spark file commit protocol.
  */
-class VeloxColumnarWriteFilesRDD(
+class GlutenColumnarWriteFilesRDD(
```
JkSelf, line 54: After moving VeloxColumnarWriteFilesExec from backend-velox to gluten-core, can we update the class names by renaming GlutenColumnarWriteFilesExec to ColumnarWriteFilesExec and GlutenColumnarWriteFilesRDD to ColumnarWriteFilesRDD?

zzcclp (approved):

PR #6406, opened 2024-07-11
[VL] Add schema validation for all operators

zhli1142015:
This PR fixed the failure below.
```
val data = "2019-09-09 01:02:03.456789"
val df = Seq(data).toDF("strTs").selectExpr(s"CAST(strTs AS TIMESTAMP_NTZ) AS ts")
df.coalesce(1).write.format("parquet").save("/tmp/test")
```

```
 org.apache.gluten.exception.GlutenException: Input schema contains unsupported type when convert row to columnar for StructType(StructField(ts,TimestampNTZType,true)) due to Schema / data type not supported: TimestampNTZType
  at org.apache.gluten.execution.RowToVeloxColumnarExec.$anonfun$doExecuteColumnarInternal$1(RowToVeloxColumnarExec.scala:54)
  at scala.Option.foreach(Option.scala:407)
  at org.apache.gluten.execution.RowToVeloxColumnarExec.doExecuteColumnarInternal(RowToVeloxColumnarExec.scala:51)
  at org.apache.gluten.execution.RowToColumnarExecBase.doExecuteColumnar(RowToColumnarExecBase.scala:62)
  at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeColumnar$1(SparkPlan.scala:222)
  at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:246)
  at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:243)
  at org.apache.spark.sql.execution.SparkPlan.executeColumnar(SparkPlan.scala:218)
  at org.apache.gluten.execution.ColumnarCoalesceExec.doExecuteColumnar(ColumnarCoalesceExec.scala:46)
```
## How was this patch tested?

UT.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/GlutenPlan.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/GlutenPlan.scala
@@ -65,7 +65,13 @@ trait GlutenPlan extends SparkPlan with Convention.KnownBatchType with LogLevelU
   final def doValidate(): ValidationResult = {
     try {
       TransformerState.enterValidation
-      val res = doValidateInternal()
+      val res = BackendsApiManager.getValidatorApiInstance
+        .doSchemaValidate(schema)
+        .map {
+          reason =>
+            ValidationResult.notOk(s"Found schema check failure for $schema, due to: $reason")
+        }
+        .getOrElse(doValidateInternal())
       if (!res.isValid) {
         TestStats.addFallBackClassName(this.getClass.toString)
       }
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/GlutenPlan.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/GlutenPlan.scala
@@ -63,15 +63,19 @@ trait GlutenPlan extends SparkPlan with Convention.KnownBatchType with LogLevelU
    * Validate whether this SparkPlan supports to be transformed into substrait node in Native Code.
    */
   final def doValidate(): ValidationResult = {
+    val schemaVaidationResult = BackendsApiManager.getValidatorApiInstance
+      .doSchemaValidate(schema)
+      .map {
+        reason => ValidationResult.notOk(s"Found schema check failure for $schema, due to: $reason")
+      }
+      .getOrElse(ValidationResult.ok)
+    if (!schemaVaidationResult.isValid) {
+      TestStats.addFallBackClassName(this.getClass.toString)
+      return schemaVaidationResult
+    }
     try {
       TransformerState.enterValidation
-      val res = BackendsApiManager.getValidatorApiInstance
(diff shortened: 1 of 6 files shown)


zhli1142015:
Aslo create a separate task to track the support for `TimestampNTZType`
#6407

zhli1142015:
cc @zhztheplayer  and @PHILO-HE , thanks

zhli1142015:
Run Gluten Clickhouse CI

PHILO-HE (commented):
I note this exception happens in RowToColumnar. Looks it will also happen for other Gluten operators following RowToColumnar, other than for Gluten coalesce operator. Right? If so, can we just do the check in `doValidate()` of GlutenPlan.scala to cover all?

ulysses-you:
+1 with @PHILO-HE

zhli1142015:
I thought this is just the missed case, as I saw this check is already added for broadcastExchange, union. But sure, let me move this to GlutenPlan.

PHILO-HE:
@zhli1142015, then, could you help remove the redundant check for `Union`, `BroadcastExchange`, `RowToColumnar`? With RowToColumnar added prior to GlutenPlan whose schema check passes, we can infer RowToColumnar's same schema check also passes.

gluten-core/src/main/scala/org/apache/gluten/extension/GlutenPlan.scala
```diff
@@ -65,7 +65,13 @@ trait GlutenPlan extends SparkPlan with Convention.KnownBatchType with LogLevelU
...
     try {
       TransformerState.enterValidation
-      val res = doValidateInternal()
+      val res = BackendsApiManager.getValidatorApiInstance
+        .doSchemaValidate(schema)
+        .map {
+          reason =>
+            ValidationResult.notOk(s"Found schema check failure for $schema, due to: $reason")
+        }
+        .getOrElse(doValidateInternal())
```
PHILO-HE, line 74: It would be better to move schema check before `TransformerState.enterValidation`. The transformer state is required by `tranform` method called by `doValidateInternal`. So before `doValidateInternal` is called, we don't need to set that state.

zhli1142015: Updated, thanks.

zhli1142015:
> @zhli1142015, then, could you help remove the redundant check for `Union`, `BroadcastExchange`, `RowToColumnar`? With RowToColumnar added prior to GlutenPlan whose schema check passes, we can infer RowToColumnar's same schema check also passes.

Done, thanks.

PHILO-HE (approved):
Looks nice! Thanks!

ulysses-you (approved):

PR #6413, opened 2024-07-11
[VL] Try to find arrow libs from velox bundled path firstly

PHILO-HE:

We expect cmake's find_library function always finds lib from specified PATHS firstly, then system paths. But local build shows it isn't true. This pr is proposed to make sure that.


zhztheplayer (approved):
Thanks!

PR #6420, opened 2024-07-12
[CORE] Move CountDistinctWithoutExpand,RewriteDateTimestampComparisonRule to ch backend

ulysses-you:

Avoid compile unnecessary class.

## How was this patch tested?

PASS CI


lgbo-ustc:
Why

ulysses-you:
> Why. These are enable in `CH`, but doesn't mean they only could be used in `CH`

Thank you @lgbo-ustc for review. This pr only mvoes rule which is dedicated for ch backend...  If a rule can be reused for other backends, it should check `BackendsApiManager.getSettings.supportxxx`, e.g., `MergeTwoPhasesHashBaseAggregate` which is also only used for ch backend but live in core.

lgbo-ustc:
> > Why. These are enable in `CH`, but doesn't mean they only could be used in `CH`
> 
> Thank you @lgbo-ustc for review. This pr only mvoes rule which is dedicated for ch backend... If a rule can be reused for other backends, it should check `BackendsApiManager.getSettings.supportxxx`, e.g., `MergeTwoPhasesHashBaseAggregate` which is also only used for ch backend but live in core.

I think there have been settings to enable/disable there rules,

lgbo-ustc:
It may be a problem to add rules for some common issues that could exists in backends

ulysses-you:
I think it's not a good story to say a rule is created for one backend but say it is not designed for that backend... and at least the tests should be added into gluten-ut module for all backends and skip backend which is not enabled by default.

Do you mean `RewriteDateTimestampComparisonRule` and `CommonSubexpressionEliminateRule` are generic for velox backend ? If so ,i can leave them to core for now and see if we can help to make it work.

BTW, the `RewriteToDateExpresstionRule` and `CountDistinctWithoutExpand` are bound to ch backend code and we can move them.

PR #6426, opened 2024-07-12
[VL] Accelerate Arrow compile

jinchengchenghh:


zhouyuan (approved):

dev/build_arrow.sh
```diff
@@ -70,7 +70,8 @@ function build_arrow_java() {
 
     mvn clean install -pl bom,maven/module-info-compiler-maven-plugin,vector -am \
           -DskipTests -Drat.skip -Dmaven.gitcommitid.skip -Dcheckstyle.skip -Dassembly.skipAssembly
-
+    
+    export CMAKE_BUILD_PARALLEL_LEVEL=112
```
zhouyuan, line 74: better to use `nproc`

zhztheplayer, line 74: Can we use `nproc` or some command to obtain a proper parallelism number? `112` is a magic number.

PHILO-HE: I guess this variable can impact velox/gluten-cpp build also. If this is the purpose, maybe, just move the `export` to builddeps-velox.sh.

jinchengchenghh: velox and gluten-cpp build use their own computed parallel level(also is NPROC), put it here because user may build step by step.

zhouyuan (approved):

PR #6433, opened 2024-07-12
[CORE] Fix fallback for spark sequence function with literal array data as input

gaoyangxiaozhu:

Before this PR the below query would fail and fallback
```
spark.sql("SELECT sequence(1, 5), string_column from t").collect
```
with error

image

(Fixes: \#ISSUE-ID)

## How was this patch tested?

UT


gaoyangxiaozhu:
@PHILO-HE / @rui-mo help review

backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
```diff
@@ -664,6 +665,15 @@ class ScalarFunctionsValidateSuite extends FunctionsValidateTest {
     }
   }
 
+  test("Test sequence function") {
+    withSQLConf(("spark.sql.optimizer.excludedRules", NullPropagation.ruleName)) {
```
PHILO-HE, line 669: Curious why this rule matters, instead of constant folding rule. Could you clarify?

gaoyangxiaozhu: hey good question @PHILO-HE , actually the matter is `constant folding rule` which should be removed from the `excludedRules`, so here is only use to override and remove it from default test sql conf https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/test/scala/org/apache/gluten/execution/FunctionsValidateTest.scala#L44

PHILO-HE: With constant folding rule applied, will the physical plan delivered to Gluten become an `Alias` plan (the result is actually computed by Spark, and Gluten just offloads the `Alias` plan)? Could you please verify?

gaoyangxiaozhu: hey @PHILO-HE, maybe you mis-understand here,  I just removed `constant folding rule` from `excludedRules` to make sure the rule can take effect (the default spark behavior), so the result of the UT is computed actually by spark and been put into `unsafeArray` with Alias Literal type to offload to gluten.

if put `constant folding rule` in `excludedRules`, it will try offload `sequence` to velox and fallback due to unsupporetd function name.

PHILO-HE: @gaoyangxiaozhu, I see. I think it would be better to revise test name to the one like: "sequence function optimized by Spark constant folding"

PHILO-HE (dismissed):

gaoyangxiaozhu:
@PHILO-HE could you help merge ?

PHILO-HE:
> @PHILO-HE could you help merge ?

@gaoyangxiaozhu, do you think this small change suggestion makes sense?
https://github.com/apache/incubator-gluten/pull/6433#discussion_r1677727658

gluten-core/src/main/java/org/apache/gluten/substrait/expression/ExpressionBuilder.java
```diff
@@ -215,6 +217,19 @@ public static LiteralNode makeLiteral(Object obj, TypeNode typeNode) {
 
   public static LiteralNode makeLiteral(Object obj, DataType dataType, Boolean nullable) {
     TypeNode typeNode = ConverterUtils.getTypeNode(dataType, nullable);
+    if (obj instanceof UnsafeArrayData) {
```
PHILO-HE, line 220: @gaoyangxiaozhu, can we just overload `makeLiteral` with `UnsafeArrayData` type as input?

gaoyangxiaozhu:
> > @PHILO-HE could you help merge ?
> 
> @gaoyangxiaozhu, do you think this small change suggestion makes sense? [#6433 (comment)](https://github.com/apache/incubator-gluten/pull/6433#discussion_r1677727658)

updated

PHILO-HE:
@gaoyangxiaozhu, please also check this comment: https://github.com/apache/incubator-gluten/pull/6433/files#r1678722614

gluten-core/src/main/java/org/apache/gluten/substrait/expression/ExpressionBuilder.java
gaoyangxiaozhu: em.. actually not sure, unless we sure the literal list type always be `UnSafeArrayData`, but i am not sure right now.
Let's keep current change , and let me do follow up to check spark code to see if can always move to `UnSafeArrayData`

gaoyangxiaozhu:
> @gaoyangxiaozhu, please also check this comment: https://github.com/apache/incubator-gluten/pull/6433/files#r1678722614

hey @PHILO-HE , please check my above reply. :)

PHILO-HE (approved):
Looks nice!

PR #6449, opened 2024-07-15
[VL] Enable right and anti join in smj

JkSelf:

Enable right and anti join in smj

## How was this patch tested?

Adding new unit tests.


rui-mo (approved):
Thanks.

PR #6455, opened 2024-07-15
[CH][CELEBORN] CHCelebornColumnarBatchSerializer uses AtomicBoolean to identify whether to call close() to avoid calling close() twice situation

SteNicholas:

`CHCelebornColumnarBatchSerializer` should use `AtomicBoolean` to identify whether to call `close()` to avoid calling `close()` twice situation which causes the replication of metrics like `readBatchNumRows`, `numOutputRows`.

## How was this patch tested?

CI.


SteNicholas:
Ping @taiyang-li, @zzcclp, @zhztheplayer.

taiyang-li:
Let's wait for ci

taiyang-li (approved):
LGTM

PR #6458, opened 2024-07-15
[VL] Add helper function ColumnarBatches.toString and InternalRow toString

jinchengchenghh:
For test purpose, add this helper function.
Add refactor columnarToRow and rowToColumnar functions to support used in otherwhere.


zhztheplayer:
What's the purpose of the API? Is it for testing use?

Also please fill in the PR description. Thanks.

zhztheplayer:
Hi @jinchengchenghh,

If it's for testing purpose from Java side, my suggestion is not to propagate the call down to C++ code. We can add a Java API `ColumnarBatches#toString` which converts the input batch to Arrow batch (via #ensureLoaded), then use some regular ways to stringify it. This could be the simplest way to make the API support both Velox and Arrow data.

jinchengchenghh:
ArrowWritableVector does not have print function in java side.
We can also change it to InternalRow Iterator, but we don't have the print function too.

zhztheplayer:
> ArrowWritableVector does not have print function in java side. We can also change it to InternalRow Iterator, but we don't have the print function too.

If we can get the rows, there must be a way to stringify them since Spark requires for this to implement `df.show`

jinchengchenghh:
> > ArrowWritableVector does not have print function in java side. We can also change it to InternalRow Iterator, but we don't have the print function too.
> 
> If we can get the rows, there must be a way to stringify them since Spark requires for this to implement `df.show`

Yes, it uses ToPrettyString to show the result in dataframe, we can only use some of code, I have implemented our version.

jinchengchenghh:
Can you help review this PR again? Thanks! @zhztheplayer

jinchengchenghh:
Can you help review? Thanks!

zhztheplayer:
@jinchengchenghh Will take a look as soon as possible. Thanks.

jinchengchenghh:
Can I take your some time? @zhztheplayer

zhztheplayer:
Reviewing now. Thank you for noting.

zhztheplayer:
Hi @jinchengchenghh , I have been thinking if the code can be simplified to ease maintenance. 

Would you please have a check about the following example code to pretty print a row iterator with much less code than `ToStringUtil`:

```scala
test("UnsafeRow to string 2") {
  val util = ToStringUtil(Option.apply(SQLConf.get.sessionLocalTimeZone))
  val row1 =
    InternalRow.apply(UTF8String.fromString("hello"), UTF8String.fromString("world"), 123)
  val rowWithNull = InternalRow.apply(null, null, 4)
  val row2 = UnsafeProjection
    .create(Array[DataType](StringType, StringType, IntegerType))
    .apply(rowWithNull)
  val it = List(row1, row2, row1, row1, row2).toIterator
  val struct = new StructType().add("a", StringType).add("b", StringType).add("c", IntegerType)

  val encoder = RowEncoder(struct).resolveAndBind()
  val deserializer = encoder.createDeserializer()
  it.map(deserializer).foreach(r => println(r.mkString("|")))
}
```

jinchengchenghh:
For Iterator[UnsafeRow], it is OK.
Then the problem occurs again, loaded Arrow ColumnarBatch cannot invoke rowIterator because ColumnarBatchRow column is IndicatorVector @zhztheplayer

zhztheplayer:
> loaded Arrow ColumnarBatch cannot invoke rowIterator because ColumnarBatchRow column is IndicatorVector

Then we should fix this at first... Which sounds like a bug. I'll take another look on the issue.

jinchengchenghh:
```
Executing SQL query from resource path /tpcds-queries/q24b.sql...
24/08/16 09:15:21 WARN Runtime: WholeStageIterator Reservation listener org.apache.gluten.memory.listener.ManagedReservationListener@4523caff still reserved non-zero bytes, which may cause memory leak, size: 2.0 MiB. 
24/08/16 09:15:21 ERROR Executor: Exception in task 0.0 in stage 477.0 (TID 1749)
org.apache.spark.SparkException: Managed memory leak detected; size = 2097152 bytes, task 0.0 in stage 477.0 (TID 1749)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$4(Executor.scala:516)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1500)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:509)
#
#
#  SIGSEGV (0xb) at pc=0x00007f372fa0c9a6, pid=11701, tid=0x00007f372e207640
#
# JRE version: OpenJDK Runtime Environment (8.0_422-b05) (build 1.8.0_422-8u422-b05-1~22.04-b05)
# Java VM: OpenJDK 64-Bit Server VM (25.422-b05 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libvelox.so+0x18049a6]  facebook::velox::memory::ScopedMemoryPoolArbitrationCtx::~ScopedMemoryPoolArbitrationCtx()+0x6
#
# Core dump written. Default location: /__w/incubator-gluten/incubator-gluten/tools/gluten-it/core or core.11701
#
# /__w/incubator-gluten/incubator-gluten/tools/gluten-it/hs_err_pid11701.log
24/08/16 09:15:21 WARN TaskSetManager: Lost task 0.0 in stage 477.0 (TID 1749) (051b84fb203f executor driver): org.apache.spark.SparkException: Managed memory leak detected; size = 2097152 bytes, task 0.0 in stage 477.0 (TID 1749)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$4(Executor.scala:516)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:1500)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:509)
```

jinchengchenghh:
@zhztheplayer  Thanks for your suggestion, It really make code clean and easy to maintain. Can you help to review again?

backends-velox/src/test/java/org/apache/gluten/columnarbatch/ColumnarBatchTest.java
```diff
@@ -134,4 +151,23 @@ private static ColumnarBatch newArrowBatch(String schema, int numRows) {
...
+      col2.putInt(j, 15 - j);
+    }
+    col2.putNull(numRows - 1);
+    for (ArrowWritableColumnVector col : columns) {
+      col.setValueCount(numRows);
+    }
+    final ColumnarBatch batch = new ColumnarBatch(columns);
+    batch.setNumRows(numRows);
+    return batch;
+  }
```
zhztheplayer, line 172: Can we remove this method? Could fill in the vectors in test case's code.

E.g.,

``` scala
          final int numRows = 100;
          final ColumnarBatch batch = newArrowBatch("a boolean, b int", numRows);
          final ArrowWritableColumnVector col0 = (ArrowWritableColumnVector) batch.column(0);
          final ArrowWritableColumnVector col1 = (ArrowWritableColumnVector) batch.column(1);
          for (int j = 0; j < numRows; j++) {
            col0.putBoolean(j, j % 2 == 0);
            col1.putInt(j, 15 - j);
          }
          col1.putNull(numRows - 1);
```

zhztheplayer (approved):
Thanks!

Issue #6487, opened 2024-07-17
ColumnarBatchOutIterator should be interruptible iterator


WangGuangxin:
### Backend

VL (Velox)

### Bug description

The `ColumnarBatchOutIterator` should check whether the task is interrupted, so that speculation tasks can end as soon as possible. 
Since when `Thread.interrupt()` is called and it's in a JNI call, the `InterruptedException` is not thrown, then we need to check the interrupt status.

A case in our env is the speculation task run for 2 hours, but in fact, the task with attempt id = 0 has already fininshed 1.5 hours ago.
![image](https://github.com/user-attachments/assets/bbc80a7e-04f3-4407-ad98-57f9d43ec536)

### Spark version

None

PR #6488, opened 2024-07-17
[GLUTEN-6487][VL] ColumnarBatchOutIterator should be interruptible iterator

WangGuangxin:

The `ColumnarBatchOutIterator` should check whether the task is interrupted, so that speculation tasks can end as soon as possible. 
Since when `Thread.interrupt()` is called and it's in a JNI call, the `InterruptedException` is not thrown, then we need to check the interrupt status.

A case in our env is the speculation task run for 2 hours, but in fact, the task with attempt id = 0 has already fininshed 1.5 hours ago.
image

(Fixes: \#6487)

## How was this patch tested?
Mannually

### DIFF
--- a/backends-clickhouse/src/main/java/org/apache/gluten/vectorized/BatchIterator.java
+++ b/backends-clickhouse/src/main/java/org/apache/gluten/vectorized/BatchIterator.java
@@ -19,7 +19,7 @@ package org.apache.gluten.vectorized;
 import org.apache.gluten.metrics.IMetrics;
 import org.apache.gluten.metrics.NativeMetrics;
 
-import org.apache.spark.TaskContext
+import org.apache.spark.TaskContext;
 import org.apache.spark.sql.execution.utils.CHExecUtil;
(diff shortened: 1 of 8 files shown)


zhztheplayer:
Would you please help check if [this API](https://github.com/apache/incubator-gluten/blob/178430f4ab8ece6966d367a4e0292e5d1381ff7a/gluten-core/src/main/scala/org/apache/gluten/utils/iterator/Iterators.scala#L38) can be used to acheive the same? Thanks.

WangGuangxin:
> Would you please help check if [this API](https://github.com/apache/incubator-gluten/blob/178430f4ab8ece6966d367a4e0292e5d1381ff7a/gluten-core/src/main/scala/org/apache/gluten/utils/iterator/Iterators.scala#L38) can be used to acheive the same? Thanks.

@zhztheplayer I double checked that it's works now, the root cause for my case is because a single `next()` call didn't return for more than 1 hour. Seems no there is no elegant way to interrupt a long running JNI call

PR #6493, opened 2024-07-18
[VL] RAS: Make default rough cost model exhaustively offload computations

zhztheplayer:
This is to align the default cost model in RAS as much as possible with the heuristic query planner. To make RAS be able to pass more UTs with the minimal impact from default cost model.

Also do some minor code improvements.


PHILO-HE (approved):
Looks nice!

Issue #6501, opened 2024-07-18
[VL] When constructing a LocalFilesNode in Velox, the fileReadProperties is missing


kecookier:
### Backend

VL (Velox)

### Bug description

[Expected behavior] and [actual behavior].

### Spark version

None

PR #6503, opened 2024-07-18
[GLUTEN-6501][VL] Fix the missing fileReadProperties when constructing a LocalFilesNode

kecookier:

(Fixes: #6501)

## How was this patch tested?
Exist CI.

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
@@ -123,7 +123,7 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
       partitionSchema: StructType,
       fileFormat: ReadFileFormat,
       metadataColumnNames: Seq[String],
-      scan: BasicScanExecTransformer): SplitInfo = {
+      properties: Map[String, String]): SplitInfo = {
     partition match {
       case p: GlutenMergeTreePartition =>
         val partLists = new JArrayList[String]()
@@ -185,7 +185,7 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
           new JArrayList[JMap[String, String]](),
           fileFormat,
           preferredLocations.toList.asJava,
-          mapAsJavaMap(scan.getProperties)
+          mapAsJavaMap(properties)
         )
       case _ =>
         throw new UnsupportedOperationException(s"Unsupported input partition: $partition.")
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
@@ -122,7 +122,8 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
       partition: InputPartition,
       partitionSchema: StructType,
       fileFormat: ReadFileFormat,
-      metadataColumnNames: Seq[String]): SplitInfo = {
+      metadataColumnNames: Seq[String],
+      scan: BasicScanExecTransformer): SplitInfo = {
     partition match {
       case p: GlutenMergeTreePartition =>
         val partLists = new JArrayList[String]()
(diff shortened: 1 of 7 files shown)


gluten-core/src/main/java/org/apache/gluten/substrait/rel/LocalFilesNode.java
```diff
@@ -78,6 +80,7 @@ public enum ReadFileFormat {
     this.partitionColumns.addAll(partitionColumns);
     this.metadataColumns.addAll(metadataColumns);
     this.preferredLocations.addAll(preferredLocations);
+    this.fileReadProperties= properties;
```
liujiayi771, line 83: Do we use this variable in `LocalFilesNode`?

backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
```diff
@@ -122,7 +122,8 @@ class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
       partition: InputPartition,
       partitionSchema: StructType,
       fileFormat: ReadFileFormat,
-      metadataColumnNames: Seq[String]): SplitInfo = {
+      metadataColumnNames: Seq[String],
+      scan: BasicScanExecTransformer): SplitInfo = {
```
liujiayi771, line 126: It is better to only pass the properties that are needed.

gluten-core/src/main/java/org/apache/gluten/substrait/rel/LocalFilesNode.java
kecookier: Yes, in my situation, this member would be used as the `tableParameters` of `velox::connector::hive::HiveTableHandle`.

liujiayi771: I see, previously we were setting the fileReadProperties through the setFileReadProperties method. If setFileReadProperties is no longer in use, we can go ahead and remove it.

backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
liujiayi771: The other parameters of this method are also from the scan transformer. It seems a bit strange to directly pass BasicScanExecTransformer here. I think it would be better to add additional parameters for other fixes.

Yohahaha (dismissed):
LGTM!

PHILO-HE:
@kecookier, CI failed. Please fix it.
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.12.1:compile (default-compile) on project gluten-iceberg: Compilation failure
Error:  /__w/incubator-gluten/incubator-gluten/gluten-iceberg/src/main/java/org/apache/gluten/substrait/rel/IcebergLocalFilesNode.java:[40,4] error: no suitable constructor found for LocalFilesNode(Integer,List<String>,List<Long>,List<Long>,ArrayList<Object>,ArrayList<Object>,List<Map<String,String>>,ArrayList<Object>,ReadFileFormat,List<String>)
Error:      constructor LocalFilesNode.LocalFilesNode(Integer,List<String>,List<Long>,List<Long>,List<Long>,List<Long>,List<Map<String,String>>,List<Map<String,String>>,ReadFileFormat,List<String>,Map<String,String>) is not applicable
Error:        (actual and formal argument lists differ in length)
Error:      constructor LocalFilesNode.LocalFilesNode(String) is not applicable
Error:        (actual and formal argument lists differ in length)
```

backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
kecookier: Let me fix it.

PHILO-HE (approved):
Looks good to me. Thanks!

liujiayi771:
@kecookier Could you remove the `setFileReadProperties` method in `LocalFilesNode` if it is unused.

kecookier:
> @kecookier Could you remove the `setFileReadProperties` method in `LocalFilesNode` if it is unused.

@liujiayi771  Thanks! At first, I missed `IcebergLocalFilesNode` in my situation. I mistakenly thought all constructors of LocalFilesNode required properties; actually, IcebergLocalFilesNode did not need them. Now, keeping `setFileReadProperties` is a better choice.  I will update the code.

liujiayi771:
@kecookier You can pass an empty map for iceberg.

kecookier:
> @kecookier You can pass an empty map for iceberg. There are other parameters in the constructor that not be used by iceberg.

Let's do as your suggested.

liujiayi771 (approved):

PR #6505, opened 2024-07-18
[VL] Ensure sed pattern can be matched when modifying velox setup scripts

PHILO-HE:

sed doesn't report error if given pattern is not matched. It's possible our expected changes are not applied to velox setup script due to upstream's code change. Here, simply run grep command to ensure pattern matched.

Only do the check for setup-ubuntu.sh, since other scripts like setup-centos7.sh, setup-centos8.sh are not maintained in upstream velox now.

## How was this patch tested?

CI/local.

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -69,6 +69,19 @@ for arg in "$@"; do
   esac
 done
 
+function ensure_pattern_matched {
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@yma11, could you take a look?

rui-mo (approved):

PHILO-HE:
Let's merge this one firstly. If some issue is found, please let me know.

Issue #6529, opened 2024-07-20
Build failed on macOS: 'arrow/status.h' file not found


xumingming:
### Backend

VL (Velox)

### Bug description

```
In file included from /tmp/gluten/cpp/core/memory/MemoryAllocator.cc:20:
/tmp/gluten/cpp/core/utils/macros.h:22:10: fatal error: 'arrow/status.h' file not found
#include <arrow/status.h>
         ^~~~~~~~~~~~~~~~
In file included from /tmp/gluten/cpp/core/compute/ResultIterator.cc:18:
In file included from /tmp/gluten/cpp/core/compute/ResultIterator.h:20:
/tmp/gluten/cpp/core/memory/ColumnarBatch.h:22:10: fatal error: 'arrow/c/bridge.h' file not found
#include "arrow/c/bridge.h"
         ^~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [core/CMakeFiles/gluten.dir/memory/MemoryAllocator.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from /tmp/gluten/cpp/core/compute/Runtime.cc:18:
In file included from /tmp/gluten/cpp/core/compute/Runtime.h:23:
In file included from /tmp/gluten/cpp/core/compute/ResultIterator.h:20:
/tmp/gluten/cpp/core/memory/ColumnarBatch.h:22:10: fatal error: 'arrow/c/bridge.h' file not found
#include "arrow/c/bridge.h"
         ^~~~~~~~~~~~~~~~~~
In file included from /tmp/gluten/cpp/core/compute/ProtobufUtils.cc:25:
/tmp/gluten/cpp/core/utils/exception.h:20:10: fatal error: 'arrow/status.h' file not found
#include <arrow/status.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [core/CMakeFiles/gluten.dir/compute/ProtobufUtils.cc.o] Error 1
In file included from /tmp/gluten/cpp/core/config/GlutenConfig.cc:22:
In file included from /tmp/gluten/cpp/core/jni/JniError.h:22:
/tmp/gluten/cpp/core/jni/JniCommon.h:20:10: fatal error: 'arrow/ipc/reader.h' file not found
#include <arrow/ipc/reader.h>
         ^~~~~~~~~~~~~~~~~~~~
In file included from /tmp/gluten/cpp/core/jni/JniWrapper.cc:21:
[...]

PR #6530, opened 2024-07-20
[GLUTEN-6529] Fix build error on macOS caused by ConfigArrow.cmake

xumingming:

(Fixes: \#6529)


PHILO-HE (approved):
@xumingming, thanks for your fix! It's strange that this issue is not reported on other OS. Could you also create a pr targets for branch-1.2? Thanks!

Issue #6535, opened 2024-07-20
Shell scripts in dev directory should have execute privilege


xumingming:
### Description

![image](https://github.com/user-attachments/assets/91f415a1-e8dc-4ea9-9a0a-24d1ef68c060)

PR #6536, opened 2024-07-20
[GLUTEN-6535] Make helper scripts executable

xumingming:
(Fixes: \#6535)


FelixYBW:
@zhouyuan any specific reason we didn't make them executable?

PHILO-HE (approved):
I think we should make them executable. Thanks!

zhouyuan:
@FelixYBW missed this patch - Some of us are not using the standard "bash" but `zsh` so I usually prefer to directly call `bash script` to ensure the behavior is correct, but for this patch it should be no difference. 

thanks, -yuan

FelixYBW:
Thank you. Let's keep it executable since it's merged.

Issue #6534, opened 2024-07-20
Simple query fails on macOS: Module not found in resource map: 4


xumingming:
### Backend

VL (Velox)

### Bug description

```
spark-sql> select * from person;
24/07/20 23:48:04 INFO HiveMetaStore: 0: get_database: default
24/07/20 23:48:04 INFO audit: ugi=admin	ip=unknown-ip-addr	cmd=get_database: default
24/07/20 23:48:04 INFO HiveMetaStore: 0: get_table : db=default tbl=person
24/07/20 23:48:04 INFO audit: ugi=admin	ip=unknown-ip-addr	cmd=get_table : db=default tbl=person
24/07/20 23:48:04 INFO HiveMetaStore: 0: get_table : db=default tbl=person
24/07/20 23:48:04 INFO audit: ugi=admin	ip=unknown-ip-addr	cmd=get_table : db=default tbl=person
24/07/20 23:48:04 INFO DataSourceStrategy: Pruning directories with:
24/07/20 23:48:04 INFO FileSourceStrategy: Pushed Filters:
24/07/20 23:48:04 INFO FileSourceStrategy: Post-Scan Filters:
24/07/20 23:48:04 INFO FileSourceStrategy: Output Data Schema: struct<id: int, name: string, age: int ... 1 more fields>
libc++abi: terminating due to uncaught exception of type gluten::GlutenException: Module not found in resource map: 4
```

### Spark version

Spark-3.3.x

### System information

Velox System Info v0.0.2
Commit: ae6a6c3c300465adf93754dc59694eece910bfce
CMake Version: 3.29.6
System: Darwin-23.3.0
Arch: arm64
CPU Name: 
C++ Compiler: /Library/Developer/CommandLineTools/usr/bin/c++
C++ Compiler Version: 15.0.0.15000309
C Compiler: /Library/Developer/CommandLineTools/usr/bin/cc
C Compiler Version: 15.0.0.15000309
CMake Prefix Path: /Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr;/opt/homebrew;/usr/local;/usr;/;/opt/homebrew/Cellar/cmake/3.29.6;/usr/local;/usr/X11R6;/usr/pkg;/opt;/sw;/opt/local

xumingming:
After some digging, it seems that `ObjectStore::stores` are initialized twice:

[...]

PR #6549, opened 2024-07-22
[GLUTEN-6534] [VL] Fix ObjectStore::stores initialized twice issue

xumingming:
(Fixes: \#6534)


zhztheplayer (approved):
Thanks for catching.

PR #6554, opened 2024-07-23
[VL] Following #6526, minor fixes and improvements

zhztheplayer:
A follow-up to #6526

1. Use the old way before https://github.com/apache/incubator-gluten/pull/5770 for block reservation because
   * It uses same algorithm for `diff < 0` and `diff > 0` so simplifies code
   * It reduces code branches
 2. Code `peakBytes_.store(std::max(peakBytes_.load(), usedBytes_.load()))` is still not atomic so needs to be locked

### DIFF
--- a/cpp/core/jni/JniCommon.h
+++ b/cpp/core/jni/JniCommon.h
@@ -403,9 +403,17 @@ class SparkAllocationListener final : public gluten::AllocationListener {
       env->CallLongMethod(jListenerGlobalRef_, jReserveMethod_, size);
       checkException(env);
     }
-    std::lock_guard<std::mutex> lock(mutex_);
     usedBytes_ += size;
-    peakBytes_ = std::max(usedBytes_, peakBytes_);
+    while (true) {
+      int64_t savedPeakBytes = peakBytes_;
+      if (usedBytes_ <= savedPeakBytes) {
+        break;
+      }
+      // usedBytes_ > savedPeakBytes, update peak
+      if (peakBytes_.compare_exchange_weak(savedPeakBytes, usedBytes_)) {
+        break;
+      }
+    }
   }
 
   int64_t currentBytes() override {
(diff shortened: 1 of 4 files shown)


cpp/core/memory/AllocationListener.h
```diff
@@ -87,11 +73,28 @@ class BlockAllocationListener final : public AllocationListener {
   }
 
  private:
+  int64_t reserve(int64_t diff) {
```
FelixYBW, line 76: add inline

cpp/core/jni/JniCommon.h
```diff
@@ -403,26 +403,34 @@ class SparkAllocationListener final : public gluten::AllocationListener {
...
+    while (true) {
+      int64_t savedPeakBytes = peakBytes_;
+      if (usedBytes_ <= savedPeakBytes) {
+        break;
+      }
+      // usedBytes_ > savedPeakBytes, update peak
+      if (peakBytes_.compare_exchange_weak(savedPeakBytes, usedBytes_)) {
+        break;
+      }
+    }
```
FelixYBW, line 416: With C++26 in future, we can use fetch_max directly. https://en.cppreference.com/w/cpp/atomic/atomic/fetch_max

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -92,8 +92,17 @@ int64_t ListenableMemoryAllocator::peakBytes() const {
...
+  while (true) {
+    int64_t savedPeakBytes = peakBytes_;
+    if (usedBytes_ <= savedPeakBytes) {
+      break;
+    }
+    // usedBytes_ > savedPeakBytes, update peak
+    if (peakBytes_.compare_exchange_weak(savedPeakBytes, usedBytes_)) {
+      break;
+    }
+  }
```
FelixYBW, line 105: With C++26 in future, we can use fetch_max directly. https://en.cppreference.com/w/cpp/atomic/atomic/fetch_max

FelixYBW (approved):

PR #6560, opened 2024-07-23
[MINOR][VL] Remove duplicate mvn packages

wForget:

The mvn package commands have been executed in `buildbundle-veloxbe.sh`, so we need to remove the duplicate mvn package commands in `package-vcpkg.sh`.

## How was this patch tested?

manual test


PHILO-HE (approved):
Thanks!

PR #6575, opened 2024-07-24
[GLUTEN-6574][VL]Validate GenerateRel should attempt to compile the generator.

lyy-pineapple:

Validate GenerateRel should attempt to compile the generator to avoid job failure when the generator is not supported and cannot fallback. For example, version 1.1.0 has the following error:
sql:
```
SELECT explode(slice(a, 2, 2)) from t
```

```
Job aborted due to stage failure: Task 1 in stage 26.0 failed 4 times, most recent failure: Lost task 1.3 in stage 26.0 (TID 360) (xx executor 11): io.glutenproject.exception.GlutenException: java.lang.RuntimeException: Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Scalar function slice not registered with arguments: (ARRAY<VARCHAR>, INTEGER, INTEGER). Found function registered with the following signatures:
((array(T),bigint,bigint) -> array(T))
Retriable: False
Function: compileRewrittenExpression
File: /root/package/gluten/ep/build-velox/build/velox_ep/velox/expression/ExprCompiler.cpp
Line: 468
```

(Fixes: \#ISSUE-6574)

## How was this patch tested?

through exist uts

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -477,7 +477,7 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::GenerateRel& gen
   }
   auto rowType = std::make_shared<RowType>(std::move(names), std::move(types));
   std::vector<core::TypedExprPtr> expressions;
-  if (generateRel.has_generator() && !validateExpression(generateRel.generator(), rowType)) {
+  if (generateRel.has_generator()) {
     if (!validateExpression(generateRel.generator(), rowType) {
       LOG_VALIDATION_MSG("Input validation fails in GenerateRel.");
       return false;
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -482,7 +482,12 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::GenerateRel& gen
       LOG_VALIDATION_MSG("Input validation fails in GenerateRel.");
       return false;
     }
-    expressions.emplace_back(exprConverter_->toVeloxExpr(generateRel.generator(), rowType));
+    auto generator = generateRel.generator().scalar_function();
+    for (const auto& argument : generator.arguments()) {
+      if (argument.has_value()) {
+        expressions.emplace_back(exprConverter_->toVeloxExpr(argument.value(), rowType));
+      }
+    }
     // Try to compile the expressions. If there is any unregistred funciton or
     // mismatched type, exception will be thrown.
     exec::ExprSet exprSet(std::move(expressions), execCtx_);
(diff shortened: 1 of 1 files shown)


cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -476,9 +476,16 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::GenerateRel& gen
     names.emplace_back(SubstraitParser::makeNodeName(inputPlanNodeId, colIdx));
   }
   auto rowType = std::make_shared<RowType>(std::move(names), std::move(types));
+  std::vector<core::TypedExprPtr> expressions;
   if (generateRel.has_generator() && !validateExpression(generateRel.generator(), rowType)) {
-    LOG_VALIDATION_MSG("Input validation fails in GenerateRel.");
-    return false;
+    if (!validateExpression(generateRel.generator(), rowType) {
```
jackylee-ch, line 481: Why should we check the `!validateExpression(generateRel.generator(), rowType)` again?

lyy-pineapple: Sorry ,cherry pick error.

lyy-pineapple: The outer layer judgment should not be verified

jinchengchenghh:
This exception because for function `slice`, it register `((array(T),bigint,bigint) -> array(T))`, but not `ARRAY<VARCHAR>, INTEGER, INTEGER`, only the second and third arguments type are different.
There is two solutions
1. move the slice from gluten expressionMap to only CH backend https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionMappings.scala#L221
2. Implement the function with `ARRAY<VARCHAR>, INTEGER, INTEGER`. https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/Slice.cpp#L210

I would prefer the second approach, datatype change is very easy to implement

lyy-pineapple:
> 这个异常是因为对于函数来说`slice`，它注册了`((array(T),bigint,bigint) -> array(T))`，但没有`ARRAY<VARCHAR>, INTEGER, INTEGER`，只有第二个和第三个参数类型不同。 有两种解决方案
> 
> 1. 将切片从 gluten expressionMap 移动到仅 CH 后端https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionMappings.scala#L221
> 2. 使用 实现该功能`ARRAY<VARCHAR>, INTEGER, INTEGER`。https [://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/Slice.cpp#L210](https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/Slice.cpp#L210)
> 
> 我更喜欢第二种方法，数据类型的改变非常容易实现

This is not just an issue with slice; similar problems can occur with other array functions when there is a mismatch.

jinchengchenghh:
Yes, We could compile the expression for all the scalar functions.
https://github.com/facebookincubator/velox/blob/main/velox/expression/ExprCompiler.cpp#L423
But it will take some expense, and our final destination is to support all the functions, so now we just list the functions unsupported type in validateScalarFunctions https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc#L195

PR #6579, opened 2024-07-24
[VL] Row based sort follow-up

marin-ma:
1. Enable radix sort. TPCH sf1T 1342s -> 1300s
2. Fix spill coredump. (using velox StlAllocator will cause coredump, replace the std::vector with raw buffer).
3. Refine metrics. Add "shuffle wall time" to calculate the overall shuffle write time.
4. Cut the memory overhead from sort buffer allocation in half.

Unresolved issues:
1. Perf regression for useRadixSort=false. This is likely caused by using stdlib qsort to replace the std::sort, as the std::vector is replaced with raw buffer to avoid coredump.
2. Less perf gain on TPCH sf3T because of spill.


cpp/velox/shuffle/RadixSort.h
```diff
@@ -0,0 +1,157 @@
...
+ */
+#include <algorithm>
+#include <cassert>
+#include <iostream>
+#include <vector>
+
+namespace gluten {
+
+template <typename Element>
+class RadixSort {
```
zhztheplayer, line 25: Is the file a 1:1 porting from vanilla Spark's Java code? If so can we add some comments somewhere in code to note that?

zhztheplayer (approved):

PR #6585, opened 2024-07-25
[Minor] Move a test from spark-3.2 module to a common test module

PHILO-HE:

This is a test for gluten core code which makes correct transition nodes are added when vanilla spark columnar scan is used. Should be placed in a common test module to ensure the relevant core code is still tested after spark-3.2 test is removed in the future.

## How was this patch tested?

CI.


PHILO-HE:
@jackylee-ch, could you take a look?

rui-mo (approved):

PR #6587, opened 2024-07-25
[VL] Gluten-it: --data-gen-strategy=once to skip generating data when it already exists

zhztheplayer:
As title, also replace previous option `--skip-data-gen` with `--data-gen-strategy=skip`.


PHILO-HE (approved):
Thanks!

PR #6618, opened 2024-07-29
[VL] Support row type and fix subfield in filter push-down

rui-mo:

Supports the push-down of IsNull and IsNotNull filter for row type.
Filter on subfield cannot be pushed-down, and we put them in the remaining filter.

## How was this patch tested?

With unit test.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/TestOperator.scala
@@ -29,7 +29,7 @@ import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
 import org.apache.spark.sql.execution.window.WindowExec
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.types.{DecimalType, IntegerType, StringType, StructField, StructType}
+import org.apache.spark.sql.types.{ArrayType, DecimalType, IntegerType, StringType, StructField, StructType}
 
 import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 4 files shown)


rui-mo (commented):
@FelixYBW @PHILO-HE Could you help take a review? Thanks.

PHILO-HE (approved):
Looks nice! cc @FelixYBW

zhli1142015:
One question here about the lgoic here.
I thought there is already logic to split filters to subfield filters and remaining filters in Velox: https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/HiveConnectorUtil.cpp#L790
I wonder what's the reason we maintain similar logic here. Thanks.

rui-mo:
> One question about the lgoic here. I thought there is already logic to split filters to subfield filters and remaining filters in Velox: https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/HiveConnectorUtil.cpp#L790 I wonder what's the reason we maintain similar logic here or what's the difference between them. Thanks.

@zhli1142015 Thanks for bringing this up. We were using our implementation from the start since Velox added the function `extractFiltersFromRemainingFilter` after we implemented it in Gluten. One distinction I see is that Gluten eliminates the filters to be pushed down from the call expression, whereas Velox does not. Before taking use of the Velox implementation, we might need to verify Gluten supported cases and fix the met issue if there is. We can then seamlessly transition to the Velox implementation. @FelixYBW Do you have any suggestion? Thanks.

FelixYBW:
> 
> @zhli1142015 Thanks for bringing this up. We were using our implementation from the start since Velox added the function `extractFiltersFromRemainingFilter` after we implemented it in Gluten. One distinction I see is that Gluten eliminates the filters to be pushed down from the call expression, whereas Velox does not. Before taking use of the Velox implementation, we might need to verify Gluten supported cases and fix the met issue if there is. We can then seamlessly transition to the Velox implementation. @FelixYBW Do you have any suggestion? Thanks.

Compared to Velox's, Is there any logic we missed?

FelixYBW (approved):

FelixYBW:
Thank you, Rui for your fix.

rui-mo:
> Compared to Velox's, Is there any logic we missed?

@FelixYBW I notice Velox supports the pushdown of `between` and `neq`, while Gluten does not. While a previous quick trial shows Velox's implementation could create subfield filters that are supported, which leads to runtime exception. We might need to do a full validation before using it.

FelixYBW:
> @FelixYBW I notice Velox supports the pushdown of `between` and `neq`, while Gluten does not. But a previous quick trial shows Velox's implementation could create subfield filters that are not supported, which leads to runtime exception. We might need to do a full validation before using it.

Thank you, Rui. Can you create an enhancement issue to track?

rui-mo:
@FelixYBW Opened an issue to track https://github.com/apache/incubator-gluten/issues/6666.

Issue #6612, opened 2024-07-27
[VL] table scan fallback gets wrong query plan


FelixYBW:
### Backend

VL (Velox)

### Bug description

A query has parquet scan with complex data type fallbacked. 
If I set spark.sql.parquet.enableVectorizedReader = False and run the query twice, the first one is correct, but the second one wrongly add ColumnartoRow, caused error `UnsafeRow cannot be cast to org.apache.spark.sql.vectorized.ColumnarBatch`

First query plan:
```
+- ^ FilterExecTransformer (5)
   +- ^ InputIteratorTransformer (4)
      +- RowToVeloxColumnar (2)
         +- Scan parquet  (1)
```

The second query plan:
```
+- ^ InputIteratorTransformer (5)
   +- RowToVeloxColumnar (3)
      +- * ColumnarToRow (2)
          +- Scan parquet  (1)
```

Similarly if I set spark.sql.parquet.enableVectorizedReader = True, the first run reports error and the second run successed.

### Spark version

Spark-3.2.x

FelixYBW:
So there are two issues:

1. wrongly recognized output of vectorized parquet scan as row format
2. some variables isn't reset after a query

FelixYBW:
1. because customer backport the vectorized parquet scan from spark3.3. ParquetFileFormat.scala in shimlayer needs to be ported.
2. The same issue as #6131. isNativeAppliable is set during first query. the local property is used by parquet scan as well. 6131 only reset parquet write.

PR #6627, opened 2024-07-30
[GLUTEN-6612] Fix ParquetFileFormat issue caused by the setting of local property isNativeApplicable

PHILO-HE:

* Fix typo: isNativeAppliable -> isNativeApplicable.
* Use == for null-safe comparing the value of local properties.
* Remove the check of this local property in `supportBatch` method. `supportBatch` is only called in read path, which shouldn't be impacted by that property used for native write.
* Reset local properties used for native write before next setting.

Fix https://github.com/apache/incubator-gluten/issues/6612, which shows inconsistent query behavior caused by historical property setting.

## How was this patch tested?

CI/local test.

### DIFF
--- a/shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/orc/OrcFileFormat.scala
+++ b/shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/orc/OrcFileFormat.scala
@@ -83,7 +83,7 @@ class OrcFileFormat extends FileFormat with DataSourceRegister with Serializable
       options: Map[String, String],
       files: Seq[FileStatus]): Option[StructType] = {
     // Why if (false)? Such code requires comments when being written.
-    if ("true".equals(sparkSession.sparkContext.getLocalProperty("isNativeApplicable")) && false) {
+    if ("true" == sparkSession.sparkContext.getLocalProperty("isNativeApplicable") && false) {
       GlutenOrcWriterInjects
         .getInstance()
         .inferSchema(sparkSession, Map.empty[String, String], files)
@@ -109,7 +109,7 @@ class OrcFileFormat extends FileFormat with DataSourceRegister with Serializable
       .asInstanceOf[JobConf]
       .setOutputFormat(classOf[org.apache.orc.mapred.OrcOutputFormat[OrcStruct]])
 
-    if ("true".equals(sparkSession.sparkContext.getLocalProperty("isNativeApplicable"))) {
+    if ("true" == sparkSession.sparkContext.getLocalProperty("isNativeApplicable")) {
       // pass compression to job conf so that the file extension can be aware of it.
       val nativeConf =
         GlutenOrcWriterInjects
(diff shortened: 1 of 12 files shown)


shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala
```diff
@@ -210,14 +210,10 @@ class ParquetFileFormat extends FileFormat with DataSourceRegister with Logging
 
   /** Returns whether the reader will return the rows as batch or not. */
   override def supportBatch(sparkSession: SparkSession, schema: StructType): Boolean = {
-    if ("true".equals(sparkSession.sparkContext.getLocalProperty("isNativeAppliable"))) {
```
PHILO-HE, line 213: @JkSelf, do you know why this check is added in the main code?

JkSelf: @PHILO-HE No need to add this check here. It is only used in parquet reader not writer.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/orc/OrcFileFormat.scala
```diff
@@ -83,7 +83,7 @@ class OrcFileFormat extends FileFormat with DataSourceRegister with Serializable
       options: Map[String, String],
       files: Seq[FileStatus]): Option[StructType] = {
     // Why if (false)? Such code requires comments when being written.
-    if ("true".equals(sparkSession.sparkContext.getLocalProperty("isNativeAppliable")) && false) {
+    if ("true" == sparkSession.sparkContext.getLocalProperty("isNativeApplicable") && false) {
```
zhztheplayer, line 86: Would it be more straight forward to use `sparkSession.sparkContext.getLocalProperty("isNativeApplicable").toBoolean`?

zhztheplayer (approved):
I am still a little bit concerned about the use of these local properties. Perhaps we can manage to remove them later some time.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/orc/OrcFileFormat.scala
PHILO-HE: @zhztheplayer, the property can be null and needs extra check if we use that. Maybe, just keep the current change.

> I am still a little bit concerned about the use of these local properties. Perhaps we can manage to remove them later some time.

Yes, it's not a good approach to use local property.

FelixYBW: Can we use true == xxx? why do we use string "true", is it from gluten config?

PHILO-HE: @FelixYBW, it's the raw value we obtained for this property.
The property value is kept by spark context and it is set by Gluten in `GlutenWriterColumnarRules`.

FelixYBW (approved):

PHILO-HE:
The implementation for `inferSchema` has code like `if (xxx && false)`. I will confirm with this code's author to see whether we can remove it in another pr.

FelixYBW:
> The implementation for `inferSchema` has code like `if (xxx && false)`. I will confirm with this code's author to see whether we can remove it in another pr.

Let's add comments in code once you get it. We should avoid such use.

PR #3069, opened 2023-09-07
[VL] Add backtrace for memory allocation tracking

zuochunwei:
## What changes were proposed in this pull request?

We designed a new class `BacktraceAllocationListener` derived from `gluten::AllocationListener`, It may be useful when we encounter a memory issue.

We use `SparkAllocationListener` by default, but we can use a wrapper class `BacktraceAllocationListener` to dump the allocation call stack by config `spark.gluten.backtrace.allocation = true`.

(Please fill in changes proposed in this fix)

(Fixes: \#ISSUE-ID)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #6635, opened 2024-07-30
[VL] Recover broken memory-trace option spark.gluten.backtrace.allocation

zhztheplayer:
`spark.gluten.backtrace.allocation` was added in https://github.com/apache/incubator-gluten/pull/3069 to print backtrace of large memory allocations into log.

The option was broken for a while and the PR is to recover the logic of it.

Additionally, option key `spark.gluten.backtrace.allocation` will be changed to `spark.gluten.memory.backtrace.allocation`.


zhztheplayer:
@FelixYBW

FelixYBW (approved):

PR #6637, opened 2024-07-30
[GLUTEN-6600][VL] Disable date type in window range frame support

JkSelf:

To support range frames, we need to add a [pre-project](https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala#L197) before the window to calculate the range frame's value. If the key in the order by clause is of date type, using [Add(orderSpec.child, bound)](https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/utils/PullOutProjectHelper.scala#L167) will result in the column's attribute being Unresolved, because the data types on both sides of `Add` are inconsistent. We try to use `DateAdd `for date types here to perform the calculation, and the pre-project can work. However, the post-project will throw same error because we have rewritten the [frame ](https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/gluten/utils/PullOutProjectHelper.scala#L214) in the Window to be of date type, but Spark's [frame type](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L155) does not support DateType. 

## How was this patch tested?

Adding new unit tests.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -350,7 +350,7 @@ object VeloxBackendSettings extends BackendSettingsApi {
                         case ByteType | ShortType | IntegerType | LongType =>
                         case _ =>
                           throw new GlutenNotSupportException(
-                            "Only integral type & date type are" +
+                            "Only integral type is" +
                               " supported for sort key when literal bound type is used!")
                       })
                 case _ =>
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -347,7 +347,7 @@ object VeloxBackendSettings extends BackendSettingsApi {
                   orderSpec.foreach(
                     order =>
                       order.dataType match {
-                        case ByteType | ShortType | IntegerType | LongType | DateType =>
+                        case ByteType | ShortType | IntegerType | LongType =>
                         case _ =>
                           throw new GlutenNotSupportException(
                             "Only integral type & date type are" +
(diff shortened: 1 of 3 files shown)


JkSelf:
@WangGuangxin @ulysses-you

WangGuangxin:
LGTM

ulysses-you:
Seems failed tests are related

zml1206:
Spark's frame type support DateType,https://github.com/apache/spark/blob/08943104fad9f5ea274e022d75135aedc55a566f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L109, I'm trying to support date type for gluten.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -347,7 +347,7 @@ object VeloxBackendSettings extends BackendSettingsApi {
                   orderSpec.foreach(
                     order =>
                       order.dataType match {
-                        case ByteType | ShortType | IntegerType | LongType | DateType =>
+                        case ByteType | ShortType | IntegerType | LongType =>
                         case _ =>
                           throw new GlutenNotSupportException(
                             "Only integral type & date type are" +
```
zml1206, line 353: Update exception message.

zml1206 (approved):
LGTM, only one comment.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
JkSelf: @zml1206 Good catch. Updated.

JkSelf:
> Seems failed tests are related

Yes. Fixed.

ulysses-you (approved):

Issue #6645, opened 2024-07-31
[VL] Remove VeloxWriteQueue which may introduce deadlock


WangGuangxin:
### Backend

VL (Velox)

### Bug description

Remove VeloxWriteQueue which may introduce deadlock

### Spark version

None

Issue #6622, opened 2024-07-29
[VL] Deadlock by race between main task thread and background io threads


zhztheplayer:
### Backend

VL (Velox)

### Bug description

Error:

Main thread:

```
org.apache.spark.memory.TaskMemoryManager.acquireExecutionMemory(TaskMemoryManager.java:146)
org.apache.spark.memory.MemoryConsumer.acquireMemory(MemoryConsumer.java:137)
org.apache.gluten.memory.memtarget.spark.TreeMemoryConsumer.borrow(TreeMemoryConsumer.java:65)
org.apache.gluten.memory.memtarget.TreeMemoryTargets$Node.borrow0(TreeMemoryTargets.java:126)
org.apache.gluten.memory.memtarget.TreeMemoryTargets$Node.borrow(TreeMemoryTargets.java:118)
org.apache.gluten.memory.memtarget.TreeMemoryTargets$Node.borrow0(TreeMemoryTargets.java:126)
org.apache.gluten.memory.memtarget.TreeMemoryTargets$Node.borrow(TreeMemoryTargets.java:118)
org.apache.gluten.memory.memtarget.OverAcquire.borrow(OverAcquire.java:56)
org.apache.gluten.memory.memtarget.ThrowOnOomMemoryTarget.borrow(ThrowOnOomMemoryTarget.java:35)
org.apache.gluten.memory.listener.ManagedReservationListener.reserve(ManagedReservationListener.java:43) => holding Monitor(org.apache.gluten.memory.listener.ManagedReservationListener@1272868201)
org.apache.gluten.vectorized.ShuffleWriterJniWrapper.write(Native Method)
org.apache.spark.shuffle.ColumnarShuffleWriter.internalWrite(ColumnarShuffleWriter.scala:180)
org.apache.spark.shuffle.ColumnarShuffleWriter.write(ColumnarShuffleWriter.scala:237)
org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:104)
org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:54)
org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:166)
org.apache.spark.scheduler.Task.run(Task.scala:141)
[...]

PR #6646, opened 2024-07-31
[GLUTEN-6645][VL] Remove VeloxWriteQueue which may introduce deadlock

WangGuangxin:

Since we are using Spark 3.2, and when we using parquet writer, a deadlock encountered. I think the root cause is the same with https://github.com/apache/incubator-gluten/issues/6622
image
image

Current code path makes write thread different with main task thread.
In fact, the `VeloxWriteQueue` is uselss here I think, and since it will buffer input batch (max up to 64 column batch), it will cause offheap oom in some case.

(Fixes: \#6645)

## How was this patch tested?
Mannually


zhztheplayer:
+1 to the change. I think VeloxWriteQueue is anyway not needed as per Velox's execution model regardless whether it causes deadlock or not.

cpp/core/jni/JniWrapper.cc
```diff
@@ -1093,22 +1093,16 @@ JNIEXPORT void JNICALL Java_org_apache_gluten_datasource_DatasourceJniWrapper_cl
   JNI_METHOD_END()
 }
 
-JNIEXPORT void JNICALL Java_org_apache_gluten_datasource_DatasourceJniWrapper_write( // NOLINT
+JNIEXPORT void JNICALL Java_org_apache_gluten_datasource_DatasourceJniWrapper_writeBatch( // NOLINT
     JNIEnv* env,
     jobject wrapper,
     jlong dsHandle,
-    jobject jIter) {
+    jlong batchHandler) {
```
zhztheplayer, line 1100: nit: `batchHandle` rather than `batchHandler`

backends-velox/src/main/scala/org/apache/spark/sql/execution/datasources/velox/VeloxFormatWriterInjects.scala
```diff
@@ -74,29 +72,26 @@ trait VeloxFormatWriterInjects extends GlutenFormatWriterInjectsBase {
...
+          if (batch.numCols == 0) {
+            // the operation will find a zero column batch from a task-local pool
+            ColumnarBatchJniWrapper.create(runtime).getForEmptySchema(batch.numRows)
+          } else {
+            val offloaded =
+              ColumnarBatches.ensureOffloaded(ArrowBufferAllocators.contextInstance, batch)
+            ColumnarBatches.getNativeHandle(offloaded)
+          }
+        }
+        datasourceJniWrapper.writeBatch(dsHandle, batchHandler)
```
zhztheplayer, line 90: ditto

gluten-data/src/main/java/org/apache/gluten/datasource/DatasourceJniWrapper.java
```diff
@@ -53,7 +52,7 @@ public long nativeInitDatasource(String filePath, long cSchema, Map<String, Stri
 
   public native void close(long dsHandle);
 
-  public native void write(long dsHandle, ColumnarBatchInIterator iterator);
+  public native void writeBatch(long dsHandle, long batchHandler);
```
zhztheplayer, line 55: ditto

FelixYBW:
> d since it will buffer input batch (max up to 64 colum

@JkSelf

gluten-data/src/main/java/org/apache/gluten/datasource/DatasourceJniWrapper.java
WangGuangxin: done

JkSelf:
@WangGuangxin We utilize the WriteQueue to minimize the number of JNI calls, which can enhance performance. Without it, we would invoke JNI operations for each batch. Could you conduct a benchmark to compare performance with and without the WriteQueue? If there's no significant difference in performance, I'm open to removing the queue.

FelixYBW:
> @WangGuangxin We utilize the WriteQueue to minimize the number of JNI calls, which can enhance performance. Without it, we would invoke JNI operations for each batch. Could you conduct a benchmark to compare performance with and without the WriteQueue? If there's no significant difference in performance, I'm open to removing the queue.

JNI call per batch should be OK. We have multiple JNI call per batch during C2R/R2C and shuffle writter.

FelixYBW (approved):

zhztheplayer:
`WriteQueue` was initially added in Gazelle project for working with Arrow writer's pulling execution model. Which is, writing is driven by Arrow that reads data actively from an iterator we passed to native code. That's the major reason why a queue was needed.

Because Velox applies pushing execution model for writing, it's definitely not needed to keep the queue in Gluten code.

JkSelf (approved):
LGTM.

PR #6651, opened 2024-07-31
[VL] Support date type in window range frame

JkSelf:

Follow up https://github.com/apache/incubator-gluten/pull/6637. In order to enable support for the date type within the window range frame, it is necessary to incorporate the date type into the Spark codebase, specifically within the AbstractDataType.scala file located at [here](https://github.com/apache/spark/blob/master/sql/api/src/main/scala/org/apache/spark/sql/types/AbstractDataType.scala#L82). To achieve this, we have added the AbstractDataType file to the shim folder. 

## How was this patch tested?

Adding new tests.

### DIFF
--- a/shims/spark32/src/main/scala/org/apache/spark/sql/types/AbstractDataType.scala
+++ b/shims/spark32/src/main/scala/org/apache/spark/sql/types/AbstractDataType.scala
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


shims/spark32/src/main/scala/org/apache/spark/sql/types/AbstractDataType.scala
```diff
@@ -0,0 +1,86 @@
...
+  override private[sql] def simpleString: String = {
+    types.map(_.simpleString).mkString("(", " or ", ")")
+  }
+}
+
+private[sql] object TypeCollection {
+
+  /** Types that include numeric types and ANSI interval types. */
+  val NumericAndAnsiInterval =
+    TypeCollection(NumericType, DayTimeIntervalType, YearMonthIntervalType, DateType)
```
JkSelf, line 72: Add the `DateType ` in `NumericAndAnsiInterval`.

zml1206: I think rewrite `AbstractDataType` is not a good idea. The first point is unstable because there may be class loading order problems. The second point may affect the results of other uses of this method. I used the method of rewrite orderSpec in #6653. What do you think？

JkSelf:
@zml1206 Thanks for your fix in https://github.com/apache/incubator-gluten/pull/6653. I will close this pr.

Issue #6650, opened 2024-07-31
[CORE] Remove bloom filter from partition filter since the BloomFilter  result format is different with spark


WangGuangxin:
### Backend

VL (Velox)

### Bug description

If `BloomFilterMightContains` is pushed down to partition filter, it's executed in Spark driver using codegen. But since the bloomfilter contructed by native are different with spark, a NPE will throw in this case

### Spark version

None

wForget:
Thanks, fixed in #9850

PR #6652, opened 2024-07-31
[GLUTEN-6650] Remove bloomfilter from partition filter

WangGuangxin:

If `BloomFilterMightContains` is pushed down to partition filter, it's executed in Spark driver using codegen. But since the bloom filter contructed by native are different with Spark, a NPE will throw in this case

(Fixes: \#6650)

## How was this patch tested?
UT

### DIFF
--- a/shims/spark34/src/main/scala/org/apache/spark/sql/execution/FileSourceScanExecShim.scala
+++ b/shims/spark34/src/main/scala/org/apache/spark/sql/execution/FileSourceScanExecShim.scala
@@ -21,7 +21,7 @@ import org.apache.gluten.sql.shims.SparkShimLoader
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.{InternalRow, TableIdentifier}
-import org.apache.spark.sql.catalyst.expressions.{And, Attribute, AttributeReference, BoundReference, Expression, FileSourceConstantMetadataAttribute, FileSourceGeneratedMetadataAttribute, FileSourceMetadataAttribute, PlanExpression, Predicate}
+import org.apache.spark.sql.catalyst.expressions.{And, Attribute, AttributeReference, BloomFilterMightContain, BoundReference, Expression, FileSourceConstantMetadataAttribute, FileSourceGeneratedMetadataAttribute, FileSourceMetadataAttribute, PlanExpression, Predicate}
 import org.apache.spark.sql.execution.datasources.{FileFormat, HadoopFsRelation, PartitionDirectory}
 import org.apache.spark.sql.execution.datasources.parquet.ParquetUtils
 import org.apache.spark.sql.execution.metric.SQLMetric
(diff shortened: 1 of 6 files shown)


shims/spark34/src/main/scala/org/apache/spark/sql/execution/FileSourceScanExecShim.scala
```diff
@@ -98,7 +105,9 @@ abstract class FileSourceScanExecShim(
 
   @transient override lazy val dynamicallySelectedPartitions: Array[PartitionDirectory] = {
     val dynamicPartitionFilters =
-      partitionFilters.filter(isDynamicPruningFilter)
+      partitionFilters
+        .filter(isDynamicPruningFilter)
+        .filterNot(isBloomFilterMightContain)
```
zhztheplayer, line 110: Do bloom filters still take effect in Gluten if we do this?

WangGuangxin: It doesn't task effect in this case, and partition filter will not re-evalute in following `FilterExec` or other operator.

zhztheplayer: So the PR completely disables runtime bloom-filter in Gluten? Am I missing something?

WangGuangxin: > completely

@zhztheplayer Only disable the case when bloomfilter is applied on hive table's partition column ( join key is hive table's partition column). 

In most case, bloom filter is applied on data column.

zhztheplayer:
> But since the bloom filter contructed by native are different with Spark, a NPE will throw in this case

I remember we had make Velox bloom-filter functions runnable in vanilla Spark operators so there shouldn't have compatibility issues. https://github.com/apache/incubator-gluten/pull/5435. Am I missing something? Or there is a bug?

WangGuangxin:
> > But since the bloom filter contructed by native are different with Spark, a NPE will throw in this case
> 
> I remember we had made Velox bloom-filter functions runnable in vanilla Spark operators so there shouldn't have compatibility issues. #5435. Am I missing something? Or there is a bug?

@zhztheplayer Got it. Let me check.

WangGuangxin:
> > > But since the bloom filter contructed by native are different with Spark, a NPE will throw in this case
> > 
> > 
> > I remember we had made Velox bloom-filter functions runnable in vanilla Spark operators so there shouldn't have compatibility issues. #5435. Am I missing something? Or there is a bug?
> 
> @zhztheplayer Got it. Let me check.

@zhztheplayer I digged into it, the main reason is that in this case, its a partition filter and the bloom filter is evaluted in driver side, so we need to change VeloxBloomFilterMightContain's codegen from 

```
val bf = ctx.addMutableState(className, "bloomFilter")
    ctx.addPartitionInitializationStatement(s"$bf = $className.readFrom($bfData);")
```

to 

```
val bf = ctx.addMutableState(className, "bloomFilter", bf => s"$bf = $className.readFrom($bfData);")
```

Otherwise, the `bf` object is not initialized when executed in driver and a NPE is thrown.

But after this change, the `VeloxBloomFilter` still cannot evaluted on driver side, since most of JNI and Resource Management will check whether we are in a spark task or not.
such as 

```
Runtimes.contextInstance
```
and 
```
Spillable reservation listener must be used in a Spark task.
```

zhztheplayer:
Sorry for missing the latest message and I am now revisiting. Thanks! @WangGuangxin

Issue #6600, opened 2024-07-26
[VL] window function error


FelixYBW:
### Backend

VL (Velox)

### Bug description

```
24/07/26 00:58:09 ERROR [Driver] datasources.FileFormatWriter: Aborting job c89ecb16-71d7-49b6-8963-cd58dd9873e1.
org.apache.spark.sql.catalyst.analysis.UnresolvedException: Invalid call to dataType on unresolved object
	at org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute.dataType(unresolved.scala:160)
	at org.apache.spark.sql.types.StructType$.$anonfun$fromAttributes$1(StructType.scala:548)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:238)
	at scala.collection.Iterator.foreach(Iterator.scala:941)
	at scala.collection.Iterator.foreach$(Iterator.scala:941)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1429)
	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	at scala.collection.TraversableLike.map(TraversableLike.scala:238)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:231)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
```

the write columns are all string and bigint. except one date column which is used as partition column.

@JkSelf is the date columun the reason?

### Spark version

Spark-3.2.x

LoseYSelf:
what is the sql and the table metadata?

FelixYBW:
The error is caused by window function:

```
sum(a) OVER (
        PARTITION BY b,
        c
        ORDER BY date RANGE BETWEEN 6 preceding AND CURRENT ROW
      ) 
```

zml1206:
Can you assign this issue to me, I want to try to solve it, thank you. @FelixYBW

JkSelf:
I have disabled the date type in https://github.com/apache/incubator-gluten/pull/6637. @zml1206, if you have the time, you are welcome to continue supporting the date type in window range frame. Thank you for your contributions.

FelixYBW:
@JkSelf Is there dependency of your current streaming window PR?

FelixYBW:
@JkSelf Looks the issue isn't fixed by 6600

[...]

PR #6653, opened 2024-07-31
[GLUTEN-6600][VL] Support date type in window range frame

zml1206:
Support date type in window range frame by rewrite orderSpecs to IntType.
(Fixes: #6600 )
## How was this patch tested?
UT


FelixYBW:
With the PR, we still can't offload below SQL, right?
```
sum(a) OVER (
        PARTITION BY b,
        c
        ORDER BY date RANGE BETWEEN 6 preceding AND CURRENT ROW
      ) 
```

zml1206:
> With the PR, we still can't offload below SQL, right?
> 
> ```
> sum(a) OVER (
>         PARTITION BY b,
>         c
>         ORDER BY date RANGE BETWEEN 6 preceding AND CURRENT ROW
>       ) 
> ```

No, this PR make it can  be offloaded.

FelixYBW:
iss6600 isn't fixed. @JkSelf

PR #6655, opened 2024-07-31
[VL] Gluten-it: --auto-cluster-resource to automatically set up CPU cores and memory sizes for local cluster

zhztheplayer:
When `--local-cluster` and `--auto-cluster-resource` are both set, cluster CPU cores / heap memory sizes / off-heap memory sizes will be automatically configured using the maximum available resource information fetched from OS API.

Basically, up to all CPU cores and 80% total RAM memory will be used to run benchmarks when the options are turned on.


FelixYBW:
Is it a spark feature?

zhztheplayer:
> Is it a spark feature?

No, it's `gluten-it`'s feature for better testing.

zhouyuan (approved):

Issue #6656, opened 2024-07-31
[CH] CHCelebornColumnarShuffleWriter assertion failed


exmy:
### Backend

CH (ClickHouse)

### Bug description

![image](https://github.com/user-attachments/assets/c7a7a2b9-b782-41b8-80c4-f5bd656a8f1b)

### Spark version

None

PR #6432, opened 2024-07-12
[CELEBORN] CHCelebornColumnarShuffleWriter supports celeborn.client.spark.shuffle.writer to use memory sort shuffle in ClickHouse backend

SteNicholas:
## What changes were proposed in this pull request?

1. `CHCelebornColumnarShuffleWriter` supports `celeborn.client.spark.shuffle.writer` to use memory sort shuffle in ClickHouse backend.
2. `CHCelebornColumnarShuffleWriter` initializes `nativeShuffleWriter` without skipping `ColumnarBatch` in `internalWrite` to avoid unnecessary creation of `nativeShuffleWriter` for skipping `ColumnarBatch`.
3. Extracts the same writing records code of `CHCelebornColumnarShuffleWriter` and `VeloxCelebornColumnarShuffleWriter` into `CelebornColumnarShuffleWriter`.

## How was this patch tested?

CI.

PR #6657, opened 2024-07-31
[GLUTEN-6656][CELEBORN] Fix CelebornColumnarShuffleWriter assertion failed

exmy:

After https://github.com/apache/incubator-gluten/pull/6432, if all of the ColumnarBatch have empty rows in `CHCelebornColumnarShuffleWriter`, the `nativeShuffleWriter` won't be initialized and causes `assertion failed`. The `VeloxCelebornColumnarShuffleWriter ` also has the same issue even without https://github.com/apache/incubator-gluten/pull/6432.

(Fixes: \#6656)

## How was this patch tested?

PASS CI

### DIFF
--- a/gluten-celeborn/clickhouse/src/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
+++ b/gluten-celeborn/clickhouse/src/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
@@ -71,7 +71,12 @@ class CHCelebornColumnarShuffleWriter[K, V](
       }
     }
 
-    assert(nativeShuffleWriter != -1L)
+    // If all of the ColumnarBatch have empty rows, the nativeShuffleWriter still equals -1
+    if (nativeShuffleWriter == -1L) {
+      handleEmptyIterator()
+      return
+    }
+
     splitResult = jniWrapper.stop(nativeShuffleWriter)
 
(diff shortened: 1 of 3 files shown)


taiyang-li (approved):
LGTM

gluten-celeborn/clickhouse/src/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
```diff
@@ -71,7 +71,12 @@ class CHCelebornColumnarShuffleWriter[K, V](
       }
     }
 
-    assert(nativeShuffleWriter != -1L)
+    // If all of the ColumnarBatch have empty rows, the nativeShuffleWriter still equals -1
```
SteNicholas, line 74: When `!records.hasNext=true`, does this also need to invoke `handleEmptyIterator`?

exmy: Yes，the logic of invoking `handleEmptyIterator` when `!records.hasNext=true` is in the `wtite` method of super class `CelebornColumnarShuffleWriter`.

exmy:
> @exmy, `VeloxCelebornColumnarShuffleWriter` has the same problem without #6432, right? If yes, please update the description of this pull request.

Done.

SteNicholas (approved):

PR #6660, opened 2024-07-31
[VL] Allow udf type conversion

marin-ma:
1. set `allowTypeConversion=true` when creating the UDF function to inject possible `Cast` if the sql datatype doesn't match any of the registered ones.
2. Or set `spark.gluten.sql.columnar.backend.velox.udfAllowTypeConversion=true` and the implicit type conversion will take effect to all udf functions. 

UDF developers can decide whether to allow type conversion for a specific udf with the first approach, but need to modify/update existing udf libraries. The second approach doesn't modify existing UDF libraries, but user need to set this option for the application.


marin-ma:
@kecookier Not sure if this approach can fit your use case. Could you help to take a look first? Thanks!

FelixYBW:
> set `spark.gluten.sql.columnar.backend.velox.udfAllowTypeConversion=true` to inject possible `Cast` if the sql datatype doesn't match any of the registered ones.

can we remove the config and always set it true? what's the side effect?

kecookier:
@marin-ma It seems to meet my needs, thank you.

marin-ma:
@kecookier Could you help to review again? Thanks!

Currently there are 2 ways to set udf type conversion, as described in the PR description.

FelixYBW (approved):
Looks good

kecookier (commented):
LGTM, thanks, rong!

PR #6670, opened 2024-08-01
[VL] Allow specifying maximum batch size for batch resizing

zhztheplayer:
This is a rework of https://github.com/apache/incubator-gluten/pull/6009.

1. Rename operator `VeloxAppendBatchesExec` to `VeloxResizeBatchesExec`, and have it supporting batch-splitting as well as batch-appending.
2. Remove old config options `spark.gluten.sql.columnar.backend.velox.coalesceBatchesBeforeShuffle` and `spark.gluten.sql.columnar.backend.velox.minBatchSizeForShuffle`
3. Add new config options:
   - `spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInput=true`
     Enable batch resizing for shuffle input
   - `spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInput.range=500~1000`
     Specify batch resizing target size range: `[500, 1000]`
4. Default range of shuffle batch resizing: `[0.25 * 4096, 4 * 4096]` (changed from `[0.8 * 4096,]`)


Yohahaha:
would you consider implement such logic into a Velox operator like ValueStream? then it can be reused in these place: before shuffle write, before agg, after scan. cc @zhli1142015 . 

And lots java/scala changes can be removed.

Yohahaha:
> > would you consider implement such logic into a Velox operator like ValueStream?
> 
> It will be the next step of the topic. See [a related comment](https://github.com/apache/incubator-gluten/pull/5952#issuecomment-2254795661) from another PR.
> 
> The major pros of doing it in Velox is to avoid breaking one Velox task into several.
> 
> > And lots java/scala changes can be removed.
> 
> Regarding the code, it's expected most of C++ code can be removed then. However it might still needed to have a JVM side plan node to map to Velox's corresponding plan node.

yeah, indeed, I have done a POC for above ideas, just one rule/transformer/velox operator added.

zhztheplayer:
> > > would you consider implement such logic into a Velox operator like ValueStream?
> > 
> > 
> > It will be the next step of the topic. See [a related comment](https://github.com/apache/incubator-gluten/pull/5952#issuecomment-2254795661) from another PR.
> > The major pros of doing it in Velox is to avoid breaking one Velox task into several.
> > > And lots java/scala changes can be removed.
> > 
> > 
> > Regarding the code, it's expected most of C++ code can be removed then. However it might still needed to have a JVM side plan node to map to Velox's corresponding plan node.
> 
> yeah, indeed, I have done a POC for above ideas, just one rule/transformer/velox operator added.

That sounds great. Do you tend to contribute to upstream Velox? If so please feel free to open a PR to refresh the feature in Gluten once Velox change has landed.

Yohahaha:
> > > > would you consider implement such logic into a Velox operator like ValueStream?
> > > 
> > > 
> > > It will be the next step of the topic. See [a related comment](https://github.com/apache/incubator-gluten/pull/5952#issuecomment-2254795661) from another PR.
> > > The major pros of doing it in Velox is to avoid breaking one Velox task into several.
> > > > And lots java/scala changes can be removed.
> > > 
> > > 
> > > Regarding the code, it's expected most of C++ code can be removed then. However it might still needed to have a JVM side plan node to map to Velox's corresponding plan node.
> > 
> > 
> > yeah, indeed, I have done a POC for above ideas, just one rule/transformer/velox operator added.
> 
> That sounds great. Do you tend to contribute to upstream Velox? If so please feel free to open a PR to refresh the feature in Gluten once Velox change has landed.

yes, I'm preparing the codes.

cpp/velox/utils/VeloxBatchResizer.cc
```diff
@@ -0,0 +1,111 @@
...
+    GLUTEN_CHECK(in->size() > maxOutputBatchSize, "Invalid state");
+  }
+
+  std::shared_ptr<ColumnarBatch> next() override {
+    int32_t remainingLength = in_->size() - cursor_;
+    GLUTEN_CHECK(remainingLength >= 0, "Invalid state");
+    if (remainingLength == 0) {
+      return nullptr;
+    }
+    int32_t sliceLength = std::min(maxOutputBatchSize_, remainingLength);
```
marin-ma, line 36: Why use `maxOutputBatchSize_` but not default batch size (4096)?

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -305,14 +305,32 @@ class GlutenConfig(conf: SQLConf) extends Logging {
...
+    assert(min > 0, "Min batch size should be larger than 0")
+    assert(max > 0, "Max batch size should be larger than 0")
+  }
+
+  private object ResizeRange {
+    def parse(pattern: String): ResizeRange = {
+      assert(pattern.count(_ == '~') == 1, s"Invalid range pattern for batch resizing: $pattern")
+      val splits = pattern.split('~')
+      assert(splits.length == 2)
+      ResizeRange(splits(0).toInt, splits(1).toInt)
```
marin-ma, line 319: Should we check the min size always < `COLUMNAR_MAX_BATCH_SIZE` and max size > `COLUMNAR_MAX_BATCH_SIZE` ?

zhztheplayer: Any reason doing such check?

To me it's not necessary to create tight coupling between this range and `COLUMNAR_MAX_BATCH_SIZE`, the former's value was just derived from the latter.

cpp/velox/utils/VeloxBatchResizer.cc
zhztheplayer: This could be a good point. I was worrying if batch slicing brings extra overhead when size is just slightly larger than 4096, say 5000.

I could raise a PR later to change the value to a more reasonable one. Before that I think it's just OK to use a large value since it aligns with current code (which doesn't split at all) as well.

marin-ma (approved):
LGTM. Thanks!

FelixYBW:
Sorry, just noted this PR. 
The config of spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInput.range=500~1000 is too complex. The maximal batch size should honor the configued batch size always. Slice has little overhead. we just need to define the threshold to merge batches.

FelixYBW:
> 
> yes, I'm preparing the codes.

We can merge small batches into large one. But the slice shouldn't happen, instead we should fix the operator which doesn't honor the maximal batch size define. @jinchengchenghh will fix the generator operator. The other operator I know is the hashjoin.

Issue #6620, opened 2024-07-29
[VL] Fallback scan when reading from unsupported file system


zhli1142015:
### Description

Fallback scan when reading from unsupported file system.
This issue is to track the continue work form https://github.com/apache/incubator-gluten/pull/4086

PR #6672, opened 2024-08-01
[VL] Fall back scan if file scheme is not supported by registered file systems

zhli1142015:

Fallback scan if root path is not supported by registered file systems

Fixes: #6620

## How was this patch tested?
UT.


zhli1142015:
cc @PHILO-HE , thanks.

PHILO-HE:
@wForget, could you take a look if you have time?

wForget:
@zhli1142015 Thanks for your work, as comment in https://github.com/apache/incubator-gluten/pull/4086#issuecomment-1871937485  suggested putting the validation into `SubstraitToVeloxPlanValidator`, I also tried it in the original PR, can you confirm which way is better?

zhli1142015:
Yes. There are lots of change sinse your PR. `setSplitInfos` method is removed from the `SubstraitContext`. If we want to still follow that way, we need to bing it back or add separate property for this.
And `getSplitInfos` is an expensive call, we should avoid to use it in validation. Instead we retrieve the first root path for validation here.
And looks all file related validation is kept in the method `supportFileFormatRead`, file format, schema, etc. So we put the check here also.
Last, `FakeParentPathFileSystem` doesn't work as expected in my local UT runs, mockfs:// is not supported by local file system. so I think a separate JNI wrapper API is needd for UT.
Hee are the differences in my mind, please le me know is this makes sense to you, @PHILO-HE  and @wForget , thanks.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -73,7 +74,14 @@ object VeloxBackendSettings extends BackendSettingsApi {
       format: ReadFileFormat,
       fields: Array[StructField],
       partTable: Boolean,
+      rootPaths: Seq[String],
       paths: Seq[String]): ValidationResult = {
+    if (
+      !rootPaths.isEmpty && !VeloxFileSystemValidationJniWrapper.supportedPaths(rootPaths.toArray)
```
PHILO-HE, line 80: Can we just extract the scheme to validate? For example, if `rootPaths consists of multiple HDFS files, we only need to let scheme "hdfs://" be validated on native side, not one file by one file validation.

In addition, as local file system is always supported by Gluten + Velox. Maybe, we can have a fast path to directly let validation pass on scala side for such case.

PHILO-HE (commented):
@zhli1142015, just posted some comments. Thanks!

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
```diff
@@ -263,4 +264,31 @@ class FallbackSuite extends VeloxWholeStageTransformerSuite with AdaptiveSparkPl
...
+            .format("parquet")
+            .save(path.getCanonicalPath)
+          runQueryAndCompare(s"SELECT count(*) FROM `parquet`.`${path.getCanonicalPath}`") {
+            df =>
+              val plan = df.queryExecution.executedPlan
+              val fileScan = collect(plan) { case s: FileSourceScanExecTransformer => s }
+              assert(fileScan.size == 1)
+              val rootPaths = fileScan(0).getRootPathsInternal
+              assert(rootPaths.length == 1)
+              assert(rootPaths(0).startsWith("file:/"))
```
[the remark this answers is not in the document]
zhli1142015: Looks the path has only single slash in my local.
`  "file:/tmp/spark-49887bc5-c568-4566-81d5-405a269787b1" did not start with "file://" (FallbackSuite.scala:285)`

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
zhli1142015: > Can we just extract the scheme to validate?

Velox API is designed to accept file path, now looks all file systems validate the schema only. But I think it's better to align this with the velox api to pass the file path. Now only one file path is validated.

> In addition, as local file system is always supported by Gluten + Velox. Maybe, we can have a fast path to directly let validation pass on scala side for such case.

Sure.

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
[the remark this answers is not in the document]
PHILO-HE: @zhli1142015, OK, let's keep it unchanged.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
PHILO-HE: @zhli1142015, I note Spark accepts data from two or more kinds of file systems, e.g.,

`df = spark.read.format("parquet").load("file:///path/to/local/data", "s3a://bucket/path/to/s3/data")`.
 
Not sure whether we need to also consider other files other than the first one, which definitely brings some cost. Does your production env. have such usage? If it is rarely used, we can just put some comments to clarify. cc @wForget

zhli1142015: I see, we don't see such usage, but let's only validate one path fpr each schema.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -73,7 +74,16 @@ object VeloxBackendSettings extends BackendSettingsApi {
       format: ReadFileFormat,
       fields: Array[StructField],
       partTable: Boolean,
+      rootPaths: Seq[String],
       paths: Seq[String]): ValidationResult = {
```
PHILO-HE, line 78: What's the differences between `rootPaths` and `paths`?

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
```diff
@@ -263,4 +264,43 @@ class FallbackSuite extends VeloxWholeStageTransformerSuite with AdaptiveSparkPl
       }
     }
   }
+
+  test("fallback reader with unsupported filesystem") {
```
PHILO-HE, line 268: Nit:
"Fall back scan operator for unsupported file scheme"

gluten-core/src/main/scala/org/apache/gluten/execution/FileSourceScanExecTransformer.scala
```diff
@@ -126,6 +127,15 @@ abstract class FileSourceScanExecTransformerBase(
     relation.location.inputFiles.toSeq
   }
 
+  override def getRootPathsInternal: Seq[String] = {
+    relation.location match {
+      case c: CatalogFileIndex =>
+        val index = c.filterPartitions(Nil)
```
wForget, line 133: Why not use `org.apache.spark.sql.execution.datasources.CatalogFileIndex#rootPaths`? `c.filterPartitions(Nil)` seems to be a dangerous operation, it may list all hive partitions.

gluten-core/src/main/scala/org/apache/gluten/utils/FileIndexUtil.scala
```diff
@@ -0,0 +1,41 @@
...
+object FileIndexUtil {
+  def getRootPath(index: FileIndex): Seq[String] = {
+    index.rootPaths
+      .filter(_.isAbsolute)
+      .map(_.toString)
+      .toSeq
+  }
+
+  def distinctRootPaths(paths: Seq[String]): Seq[String] = {
+    // Skip native validation for local path, as local file system is always registered.
```
wForget, line 32: I think this logic should not be in the util method, we should put it in the method that does the validation

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1968,6 +1971,13 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(false)
 
+  val VELOX_SCAN_ROOT_PATH_SCHEMA_VALIDATION_ENABLED =
+    buildConf("spark.gluten.sql.scan.root.path.schema.validation.enabled")
```
wForget, line 1975: This configuration name is too complicated, can we simplify it?

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
wForget, line 268: This test case does not seem to see the fallback logic

gluten-core/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
```diff
@@ -92,7 +101,12 @@ trait BasicScanExecTransformer extends LeafTransformSupport with BaseDataSource
     }
 
     val validationResult = BackendsApiManager.getSettings
-      .supportFileFormatRead(fileFormat, fields, getPartitionSchema.nonEmpty, getInputFilePaths)
+      .supportFileFormatRead(
```
PHILO-HE, line 104: `supportFileFormatRead` is not only checking file format now. Better to change it to other name, like "validateScan" or some other a bit common name.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
wForget: There is a discussion about this: https://github.com/apache/incubator-gluten/pull/4086#discussion_r1432313011

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
wForget: We don't have this usage, but I'd favor limiting it, perhaps by simply using `rootPaths.take(10)`

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
zhli1142015: paths is emty for velox, as intializing it is time consuming op.
https://github.com/apache/spark/blob/b54b3006f94604c41f230cb94fa1ef86da8ff6c9/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/PartitioningAwareFileIndex.scala#L114C16-L114C26

gluten-core/src/main/scala/org/apache/gluten/execution/FileSourceScanExecTransformer.scala
zhli1142015: I saw this function appear twice in the CatalogFileIndex, so I think this may not be an expensive call. Thanks for pointing this out, make sense to me.

backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
zhli1142015: Moved to VeloxScanSuite, thanks.

gluten-core/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
zhli1142015: Updated, thanks

gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
```diff
@@ -76,6 +76,8 @@ case class HiveTableScanExecTransformer(
     Seq.empty
   }
 
+  override def getRootPathsInternal: Seq[String] = Seq.empty
```
wForget, line 79: Can we use table location as root path? Or add a TODO comment,

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergScanTransformer.scala
```diff
@@ -54,6 +54,8 @@ case class IcebergScanTransformer(
 
   override def getInputFilePathsInternal: Seq[String] = Seq.empty
 
+  override def getRootPathsInternal: Seq[String] = Seq.empty
```
wForget, line 57: ditto

wForget:
Thanks, overall LGTM.

PHILO-HE:
@zhli1142015, could you respond to @wForget's comments?

gluten-core/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
zhli1142015: Added TODO comment

zhli1142015:
Just updated, thanks.

PHILO-HE (approved):
Thanks!

PR #6679, opened 2024-08-01
[Core] Extract a separate FallbackSummary class from GlutenImplicits

### LINKED ISSUE #6678: [Core] Provide a public collectQueryExecutionFallbackSummary method
### Description

Currently, `GlutenImplicits` provides a method to explain gluten `FallbackSummary` from a dataset, which is useful for checking whether the sql is suitable for gluten. I want to extract a separate `FallbackSummary` class that can directly construct `FallbackSummary` from `SparkPlan`. In that way, we can support getting gluten `FallbackSummary` of sql in kyuubi plan only mode.

wForget:

Extract a separate FallbackSummary class from GlutenImplicits

closes #6678

## How was this patch tested?

Existing unit tests, just code refactor


wForget:
cc @ulysses-you

ulysses-you (approved):

PR #6682, opened 2024-08-01
[VL] Add reader process to shuffle benchmark

marin-ma:
- Add `--run-shuffle-read` command line flag to add shuffle read process after write. Only takes effect when `--run-shuffle` is set.
- Add `--debug-mode` command line flag to enable debug mode (same as `spark.gluten.sql.debug`)
- Refined log printing. Fixed coredump when using `--help`.


FelixYBW:
Thank you! Rong. Is there new parameter, can you add to document?
Also can you add an example of commandline as well as the output to the threads and document if not yet.

FelixYBW (approved):

FelixYBW:
Thank you, @marin-ma

Issue #6688, opened 2024-08-02
[VL] Add config can disable rewrite collect in objectHashAggregate


zml1206:
### Description

Spark collect is `TypedImperativeAggregate` which can use objectHashAggregate, but velox collect is `DeclarativeAggregate`. `CollectRewriteRule` is a logical plan rule, it will replace spark collect to velox collect. If a collect aggregate is `ObjectHashAggregate` in vanilla spark, it will change to `SortAggregate` in gluten. The performance of `ObjectHashAggregate` is much greater than `SortAggregate`, if fallback it will cause performance regression.
Add config `spark.gluten.sql.columnar.backend.velox.objectHashAggregate.collect.rewrite.enabled`, it can disable rewrite collect in objectHashAggregate.

zml1206:
cc @zhztheplayer what do you think?

zhztheplayer:
+1 Let's add the option. Thanks.

FelixYBW:
If we need the option to be set true always, let's remove the option and just implement as it's true. We can document it in code or in the PR

zml1206:
> If we need the option to be set true always, let's remove the option and just implement as it's true. We can document it in code or in the PR

Add this option is to allow users to selectively disable when regression occurs.

FelixYBW:
> 
> Add this option is to allow users to selectively disable when regression occurs.

Did you see regression in some cases?

zml1206:
> Did you see regression in some cases?

Yes, vanilla spark is ObjectHashAggregate, gluten fallback to SortAggregate.

PHILO-HE:
@zml1206, I remember Gluten replaces ObjectHashAggregate with its HashAggregate, not SortAggregate. Right?

zml1206:
> @zml1206, I remember Gluten replaces ObjectHashAggregate with its HashAggregate, not SortAggregate. Right?

Yea, but `CollectRewriteRule` rewrite collect to velox_collect, velox_collect is not `TypedImperativeAggregate`.
It affects the aggregate type selection when converting spark logical plan to physical plan, only  `TypedImperativeAggregate` can create `ObjectHashAggregateExec`. For example, `select collect_set(string_field), velox_unsupport_udaf() from t`, in vanilla spark is `ObjectHashAggregateExec`, in gluten is `SortAggregateExec`.

zhztheplayer:
> If we need the option to be set true always, let's remove the option and just implement as it's true. We can document it in code or in the PR

[...]

PR #6689, opened 2024-08-02
[GLUTEN-6688][VL] Add config to disable rewrite collect in objectHashAggregate

zml1206:

Fixes: \#6688

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
@@ -51,14 +51,15 @@ case class CollectRewriteRule(spark: SparkSession) extends Rule[LogicalPlan] {
     out
   }
 
+  /**
+   * Do not replace collect in objectHashAggregate if disable objectHashAgg collectRewrite. When
+   * AggregateExpressions does not support HashAgg, spark_collect aggregate will be converted to
+   * objectHashAggregate, and velox_collect will be converted to sortAggregate in AggUtils, because
+   * velox_collect is not TypedImperativeAggregate. If collect aggregate fallback, disable
+   * objectHashAgg collectRewrite it will fallback to objectHashAggregate, otherwise fallback to
+   * sortAggregate. The performance of objectHashAggregate is much greater than sortAggregate.
+   */
   private def needReplace(plan: LogicalPlan): Boolean = plan match {
-    // Do not replace collect in objectHashAggregate if disable objectHashAgg collectRewrite. When
-    // AggregateExpressions does not support HashAgg, spark_collect aggregate will be converted
-    // to objectHashAggregate, and velox_collect will be converted to sortAggregate in AggUtils,
-    // because velox_collect is not TypedImperativeAggregate. If collect aggregate fallback,
-    // disable objectHashAgg collectRewrite it will fallback to objectHashAggregate, otherwise
-    // fallback to sortAggregate. The performance of objectHashAggregate is much greater than
-    // sortAggregate.
     case PhysicalAggregation(_, aggregateExpr, _, _)
         if !GlutenConfig.getConf.veloxObjectHashAggCollectRewriteEnabled =>
       val aggregateExpressions = aggregateExpr.map(expr => expr.asInstanceOf[AggregateExpression])
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
@@ -16,6 +16,7 @@
  */
 package org.apache.gluten.extension
 
+import org.apache.gluten.GlutenConfig
 import org.apache.gluten.expression.ExpressionMappings
 import org.apache.gluten.expression.aggregate.{VeloxCollectList, VeloxCollectSet}
 import org.apache.gluten.utils.LogicalPlanSelector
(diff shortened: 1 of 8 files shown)


backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
```diff
@@ -48,6 +51,19 @@ case class CollectRewriteRule(spark: SparkSession) extends Rule[LogicalPlan] {
...
+    case PhysicalAggregation(_, aggregateExpr, _, _)
+        if !GlutenConfig.getConf.veloxObjectHashAggCollectRewriteEnabled =>
+      val aggregateExpressions = aggregateExpr.map(expr => expr.asInstanceOf[AggregateExpression])
+      val useHash = SparkShimLoader.getSparkShims.supportsHashAggregate(
+        aggregateExpressions.flatMap(_.aggregateFunction.aggBufferAttributes))
+      val useObjectHash = plan.conf.useObjectHashAggregation &&
+        SparkShimLoader.getSparkShims.supportsObjectHashAggregate(aggregateExpressions)
+      useHash || !useObjectHash
+    case _ => true
+  }
```
zhztheplayer, line 65: Would you like to explain a bit about this code? Thanks.

zml1206: Updated

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
```diff
@@ -48,6 +51,25 @@ case class CollectRewriteRule(spark: SparkSession) extends Rule[LogicalPlan] {
     out
   }
 
+  private def needReplace(plan: LogicalPlan): Boolean = plan match {
+    // Do not replace collect in objectHashAggregate if disable objectHashAgg collectRewrite. When
```
xumingming, line 55: The comment is more appropriate to be the comment of the method.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1940,6 +1943,14 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val VELOX_OBJECTHASHAGG_COLLECT_REWRITE_ENABLED =
+    buildConf("spark.gluten.sql.columnar.backend.velox.objectHashAggregate.collect.rewrite.enabled")
```
xumingming, line 1947: umm...the config name seems too long?

zml1206: It is indeed a bit long, but it seems difficult to shorten. How do you think it should be changed?

xumingming: how about spark.gluten.sql.columnar.backend.velox.collectRewriteEnabled?

zml1206: > how about spark.gluten.sql.columnar.backend.velox.collectRewriteEnabled?

This is not global, it just disables objectHashAggregate's collect, although currently there is only objectHashAggregate, there will be ambiguity in naming this way. cc @zhztheplayer

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
zml1206: Updated, thanks.

zml1206:
CI failure is unrelated.

PHILO-HE:
@zml1206, can we add a more general config to allow disable any logical rule? It's like Spark's `spark.sql.optimizer.excludedRules`. cc @zhztheplayer

zml1206:
> @zml1206, can we add a more general config to allow disabling any logical rule? It's like Spark's `spark.sql.optimizer.excludedRules`. cc @zhztheplayer

There is a difference between these two configurations. If exclude `CollectRewriteRule`, not only ObjectHashAggregate is offloaded, but HashAggregate is also disabled when collect is included.

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
zhztheplayer: Sorry I still don't quite understand (thought it will be a simple switch to turn the rule on/off). If all you wanted is to fallback to vanilla Spark's object hash aggregate, why not just disable this rule?

If it's hard to explain, would you like to share the comparison of query plans in text before and after the change?

zhztheplayer:
> @zml1206, can we add a more general config to allow disabling any logical rule? It's like Spark's `spark.sql.optimizer.excludedRules`. cc @zhztheplayer

We have a `spark.gluten.expression.blacklist` to control expression fallbacks which could be used for avoid offloading collect_list / collect_set either. If all we wanted is to disable the rule, Perhaps we can just set that option to `spark.gluten.expression.blacklist=collect_list,collect_set` or something. I may be wrong that we should have an individual option for this.

zml1206:
> We have a `spark.gluten.expression.blacklist` to control expression fallbacks which could be used for avoid offloading collect_list / collect_set either. If all we wanted is to disable the rule, perhaps we can just set that option to `spark.gluten.expression.blacklist=collect_list,collect_set` or something. I may be wrong that we should have an individual option for this.

`spark.gluten.expression.blacklist=collect_list,collect_set` and `spark.sql.optimizer.excludedRules=***.CollectRewriteRule`  have the same effect, no matter what kind of aggregation they will fallback collect, but this new conf only fallback collect of objectHashAggregate.

PHILO-HE:
> > @zml1206, can we add a more general config to allow disabling any logical rule? It's like Spark's `spark.sql.optimizer.excludedRules`. cc @zhztheplayer
> 
> We have a `spark.gluten.expression.blacklist` to control expression fallbacks which could be used for avoid offloading collect_list / collect_set either. If all we wanted is to disable the rule, perhaps we can just set that option to `spark.gluten.expression.blacklist=collect_list,collect_set` or something. I may be wrong that we should have an individual option for this.

@zhztheplayer, makes sense to me. Thanks!

zml1206:
Sorry for my mistake, I just discovered that if an aggregate contains collect, it must be ObjectHashAggregate in vanilla spark, and cannot be HashAggregate, so this new configuration is redundant, thank you everyone for your review. @PHILO-HE @zhztheplayer @xumingming @FelixYBW

PR #6690, opened 2024-08-02
[VL] Remove redundant hash function in substrait function validation

jinchengchenghh:
Here we check the corresbonding Velox function name, which is `hash_with_seed` now, and support the complex data type


FelixYBW:
So it's actually an enhancement to support complex datatype in hash, right?

marin-ma:
> So it's actually an enhancement to support complex datatype in hash, right?

@FelixYBW Complex type hash has been supported. This is a cleanup for unused code.

marin-ma (approved):
LGTM. Thanks!

PR #6693, opened 2024-08-02
[VL] Minor follow-ups for PRs

zhztheplayer:
1. Following PR https://github.com/apache/incubator-gluten/pull/6670, change min batch size to 0.25 * default size
   - Remove option `spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInput.range`
   - Add option `spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInput.minSize`
3. Following PR https://github.com/apache/incubator-gluten/pull/6658, change gluten-it's JVM options to match up with [vanilla Spark](https://github.com/apache/spark/blob/080e7eb0de06d49959f2f2b05a1f446adf12fadb/launcher/src/main/java/org/apache/spark/launcher/JavaModuleOptions.java), since options in `JavaModuleOptions` are passed to executors in local cluster mode, we'd align executors' and driver's options in that mode.


FelixYBW:
As I documented, since the slice has little overhead, let's just honor the max batch size as 1x.

zhztheplayer:
> As I documented, since the slice has little overhead, let's just honor the max batch size as 1x.

If I remember correctly in our code there were some places producing columnar batches that use default batch size `4096` as a minimal threshold. Which means we have chance to frequently get batches with size say, 4100 or something. Let's at least relax a little bit to allow that kind of batches to pass? Otherwise very small batches could be produced.

FelixYBW (approved):

PR #6698, opened 2024-08-02
[GLUTEN-6656][UNIFFLE] VeloxUniffleColumnarShuffleWriter should send commit for all ColumnBatch with empty rows

SteNicholas:

When all of the ColumnarBatch have empty rows in `VeloxUniffleColumnarShuffleWriter`, the `nativeShuffleWriter` is not initialized and causes `IllegalStateException`. `VeloxUniffleColumnarShuffleWriter` should send commit for all `ColumnBatch` with empty rows.

Follow up #6657.

(Fixes: \#6656)

## How was this patch tested?

CI.

### DIFF
--- a/gluten-uniffle/velox/src/main/java/org/apache/spark/shuffle/writer/VeloxUniffleColumnarShuffleWriter.java
+++ b/gluten-uniffle/velox/src/main/java/org/apache/spark/shuffle/writer/VeloxUniffleColumnarShuffleWriter.java
@@ -126,8 +126,8 @@ public class VeloxUniffleColumnarShuffleWriter<K, V> extends RssShuffleWriter<K,
 
   @Override
   protected void writeImpl(Iterator<Product2<K, V>> records) {
-    if (!records.hasNext() && !isMemoryShuffleEnabled) {
-      super.sendCommit();
+    if (!records.hasNext()) {
+      sendCommit();
       return;
     }
     // writer already init
@@ -189,11 +189,13 @@ public class VeloxUniffleColumnarShuffleWriter<K, V> extends RssShuffleWriter<K,
       }
     }
 
-    long startTime = System.nanoTime();
     LOG.info("nativeShuffleWriter value {}", nativeShuffleWriter);
+    // If all of the ColumnarBatch have empty rows, the nativeShuffleWriter still equals -1
     if (nativeShuffleWriter == -1L) {
-      throw new IllegalStateException("nativeShuffleWriter should not be -1L");
+      sendCommit();
+      return;
     }
+    long startTime = System.nanoTime();
     SplitResult splitResult;
     try {
       splitResult = jniWrapper.stop(nativeShuffleWriter);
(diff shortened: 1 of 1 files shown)


SteNicholas:
Ping @exmy, @jackylee-ch, @summaryzb.

gluten-uniffle/velox/src/main/java/org/apache/spark/shuffle/writer/VeloxUniffleColumnarShuffleWriter.java
```diff
@@ -189,11 +189,13 @@ public long spill(MemoryTarget self, Spiller.Phase phase, long size) {
...
 
-    long startTime = System.nanoTime();
     LOG.info("nativeShuffleWriter value {}", nativeShuffleWriter);
-    if (nativeShuffleWriter == -1L) {
-      throw new IllegalStateException("nativeShuffleWriter should not be -1L");
+    // If all of the ColumnarBatch have empty rows, the nativeShuffleWriter still equals -1
+    if (nativeShuffleWriter == -1L && !isMemoryShuffleEnabled) {
+      super.sendCommit();
+      return;
     }
```
xianjingfeng, line 197:
```suggestion
    if (nativeShuffleWriter == -1L) {
      if (!isMemoryShuffleEnabled) {
        super.sendCommit();
      }
      return;
    }
```

SteNicholas: @xianjingfeng, when `!records.hasNext()` returns true but `isMemoryShuffleEnabled` is true, the `VeloxUniffleColumnarShuffleWriter#writeImpl` does not also return. This change keeps the same behavitor with above.

summaryzb:
I'm ok with this fix

gluten-uniffle/velox/src/main/java/org/apache/spark/shuffle/writer/VeloxUniffleColumnarShuffleWriter.java
xianjingfeng: > @xianjingfeng, when `!records.hasNext()` returns true but `isMemoryShuffleEnabled` is true, the `VeloxUniffleColumnarShuffleWriter#writeImpl` does not also return. This change keeps the same behavior with above.

I don't think the above logic makes sense either.

xianjingfeng: https://github.com/apache/incubator-uniffle/blob/master/client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java#L341-L343

xianjingfeng (approved):
LGTM

SteNicholas:
@jackylee-ch, PTAL.

jackylee-ch (approved):
LGTM

PR #6720, opened 2024-08-06
[VL] Minor class name / package name clean-ups

zhztheplayer:
1. `exec.Runtime` -> `runtime.Runtime` as `exec` collides with `execution`
2. Scala test suites with name `...Test` changed to use `Suite` as suffix
3. Simplify class names `...FuzzerTest` to `...Fuzzer`


PHILO-HE (approved):
Thanks!

PR #6729, opened 2024-08-06
[VL] Use the scripts dir in the current path as SCRIPTDIR

liujiayi771:

The user may customize `VELOX_HOME`, or they may use the default directory `ep/build-velox/build/velox_ep`. However, in the current `setup-centos7.sh` and `setup-centos8.sh`, only the default `VELOX_HOME` path will be used.

In fact, before calling `setup-centos7.sh` or `setup-centos8.sh`, `builddeps-veloxbe.sh` has already changed the directory to `VELOX_HOME`, so we can directly use the `scripts` dir in the current path.

## How was this patch tested?

N/A

### DIFF
--- a/ep/build-velox/src/setup-centos7.sh
+++ b/ep/build-velox/src/setup-centos7.sh
@@ -17,7 +17,7 @@
 set -efx -o pipefail
 # Some of the packages must be build with the same compiler flags
 # so that some low level types are the same size. Also, disable warnings.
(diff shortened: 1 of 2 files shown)


liujiayi771:
@PHILO-HE Could you help to review?

PHILO-HE (approved):
Thanks for your fix!

PR #6740, opened 2024-08-07
[VL] Fix shuffle spill not reported to spark metric

marin-ma:
Will have `Spill (Memory)` and `Spill (Disk)` shown on UI
image


FelixYBW (dismissed):

FelixYBW (dismissed):

marin-ma:
cc: @zhztheplayer Seems like spill from Velox pipeline is not reported to the TaskMetrics either.

zhztheplayer:
> cc: @zhztheplayer Seems like spill from Velox pipeline is not reported to the TaskMetrics either.

Hi @marin-ma, would you like to file a issue for that? It was a known issue on which we are lack of tracking. Thanks.

marin-ma:
> Hi @marin-ma, would you like to help file a issue for that? It was a known issue on which we are lack of tracking. Thanks.

@zhztheplayer Filed issue here https://github.com/apache/incubator-gluten/issues/6763

zhztheplayer (approved):

PR #6741, opened 2024-08-07
[GLUTEN-6736] Phase 1: Use task-shared lock in ManagedReservationListener

zhztheplayer:
1. Use task-shared lock in ManagedReservationListener
2. Fix a bug that over-acquire target still accountants bytes in task memory manager when task is ending
3. Add UTs

Using a shared lock forces any of the listeners in the same task - if they are running in parallel - to wait for other allocations to end before itself initializes a new allocation. Otherwise ABBA deadlock might occur, for example:

```
(The steps take place ordinally)

1. Thread A acquires memory
2. Thread A locks `listener A`
3. Thread A locks `task memory manager`
4. Thread B acquires memory
5. Thread B locks `listener B`
6. Thread B locks `task memory manager` -> WAIT (locked by Thread A)
7. Thread A about to spill `listener B`
8. Thread A locks `listener B` -> WAIT (locked by Thread B)
```

This change would correct the call order to:

```
1. Thread A acquires memory
2. Thread A locks `shared lock`
3. Thread A locks `task memory manager`
4. Thread B acquires memory
5. Thread B locks `shared lock` -> WAIT (locked by Thread A)
7. Thread A about to spill `listener B`
8. Thread A locks `shared lock`
9. Thread A spills `listener B`
11. Thread A finishes allocation, releases `shared lock`
12. Thread B locks `shared lock`
13. ...
14. Thread B finishes allocation, releases `shared lock`
```

As a result the Java side ABBA dead lock is fixed.

Note this is phase 1 of the topic. C++ code still has similar locks so requires to change in the next PR.


FelixYBW:
So the root cause is one task thread is going to spill the other task thread's data, right?

If shared lock is acquired always before task memory manager lock, looks the task memory manager lock is unnecessary

zhztheplayer:
> looks the task memory manager lock is unnecessary

We can barely touch that lock's code since it's vanilla Spark. And it is still needed since vanilla Spark's operators may rely on that lock.

marin-ma (approved):

PR #6743, opened 2024-08-07
[CH] Fix memory config spill_mem_ratio always zero

liuneng1994:

fix memory config spill_mem_ratio always zero

## How was this patch tested?

unit tests


liuneng1994:
image

PR #6744, opened 2024-08-07
[DNM][CORE] Fallback cheap project if child is fallbacked

zml1206:

Fallback cheap project if child is fallbacked, to reduce unnecessary conversions of row and columnar for avoid performance regressions.

## How was this patch tested?

UT


zhztheplayer:
Thanks for working on this.

Since the new rule looks like a manually written "coster", actually I recommend to use RAS planner to achieve the same.

RAS planner is so far mature enough to handle cases like this. All we need is to change the cost model. 

I know it's a safer choice to make the change in legacy planner but I am afraid using single rules to do such fallback is a way that increases planner code's complexity silently until being unmaintainable.

zml1206:
How about move it to `OffloadProject`? cc @zhztheplayer

zhztheplayer:
> How about move it to `OffloadProject`? cc @zhztheplayer

It might be an option to make behaviors like this customizable via exposing some kind of API, which may need some designs though. I won't suggest putting relevant rules in Gluten core code.

zhztheplayer:
> What are the current shortcomings of RAS? When is it expected to be enabled by default?

I was planning to make it default for Velox backend in several months but I am not pretty sure at the moment. It's mature enough but the only thing I am worried about is out developers may not be that familiar about writing CBO rules. There could be some threshold playing with RAS planner.

Issue #6705, opened 2024-08-04
[CH] Basic Support Write Paruet in Delta


baibaichen:
### Description

Clichouse Backend needs support Delta write 

Preconditions
- [x] #6745 
- [x] #6761

PR #6745, opened 2024-08-07
[GLUTEN-6705] [CORE] [Part 1] Avoid adding c2r for ColumnarWriteFilesExec, since it neither output Columnar batch data nor InternalRow

baibaichen:

This PR avoids c2r and r2c transition by returning `Convention` which ouputs both `VanillaRow` and `BatchType` for `ColumnarWriteFilesExec`.

```scala
  /**
   * ColumnarWriteFilesExec neither output Row nor columnar data. We output both row and columnar to
   * avoid c2r and r2c transitions. Please note, [[GlutenPlan]] already implement batchType()
   */
  sealed trait ExecuteWriteCompatible extends KnownChildrenConventions with KnownRowType {
   // ...
    override def requiredChildrenConventions(): Seq[ConventionReq] = {
      List(ConventionReq.backendBatch)
    }

    override def rowType(): RowType = {
      RowType.VanillaRow
    }
  }
```
### Why need

`ColumnarWriteFilesExec` neither outputs Columnar batch data nor `InternalRow`, and hence we can't add c2r or r2c on top of it, otherwise it would be am internal error.

1. In the normal code path, `DataWritingCommandExec` is on top of `ColumnarWriteFilesExec` and `DataWritingCommandExec`'s Convention is `VanillaRow` which avoid c2r transition. 
2. In ClickHouse Backend, we are preparing to support native delta write in Spark 3.5 and delta 3.2, `WriteFile` is directly added on top of delta logical plan and hence the root `SparkPlan` is `ColumnarWriteFilesExec` which casue internal error without this change.

### Other refactor
1. Using `org.apache.spark.sql.catalyst.util.sideBySide` to ouptput changed plan, which is same as spark `RuleExecutor`.
2. Improving some readability,  `case _ if !plan.supportsColumnar`  =>  `case _ if plan.supportsColumnar`  
3. Adding comments for `CHBatch.fromRow`

(Fixes: \#6705)

## How was this patch tested?

Existed UTs

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition/ConventionFunc.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition/ConventionFunc.scala
@@ -101,14 +101,14 @@ object ConventionFunc {
       val out = plan match {
         case k: Convention.KnownRowType =>
           k.rowType()
-        case _ if !SparkShimLoader.getSparkShims.supportsRowBased(plan) =>
-          Convention.RowType.None
-        case _ =>
+        case _ if SparkShimLoader.getSparkShims.supportsRowBased(plan) =>
           Convention.RowType.VanillaRow
+        case _ =>
+          Convention.RowType.None
       }
-      if (out != Convention.RowType.None) {
-        assert(SparkShimLoader.getSparkShims.supportsRowBased(plan))
-      }
+      assert(
+        out == Convention.RowType.None || plan.isInstanceOf[Convention.KnownRowType] ||
+          SparkShimLoader.getSparkShims.supportsRowBased(plan))
       out
     }
 
@@ -119,15 +119,13 @@ object ConventionFunc {
           p match {
             case k: Convention.KnownBatchType =>
               k.batchType()
-            case _ if !plan.supportsColumnar =>
-              Convention.BatchType.None
-            case _ =>
+            case _ if plan.supportsColumnar =>
               Convention.BatchType.VanillaBatch
+            case _ =>
+              Convention.BatchType.None
           }
       )
-      if (out != Convention.BatchType.None) {
-        assert(plan.supportsColumnar)
-      }
+      assert(out == Convention.BatchType.None || plan.supportsColumnar)
       out
     }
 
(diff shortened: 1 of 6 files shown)


PHILO-HE:
cc @zhztheplayer

FelixYBW:
@JkSelf

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition/ConventionFunc.scala
```diff
@@ -92,6 +92,12 @@ object ConventionFunc {
         }
         val conv = Convention.of(rowType, batchType)
         conv
+      case _: ColumnarWriteFilesExec.ExecuteWriteCompatible =>
+        // ColumnarWriteFilesExec neither output Row nor columnar data.
+        // We output both row and columnar to avoid c2r and r2c transitions.
+        Convention.of(
+          Convention.RowType.VanillaRow,
+          BackendsApiManager.getSparkPlanExecApiInstance.batchType)
```
zhztheplayer, line 100: Basically this list is for overriding conventions of vanilla Spark's operators. Would you use interface `KnowBatchType` / `KnownRowType` instead?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition/ConventionFunc.scala
```diff
@@ -156,6 +158,10 @@ object ConventionFunc {
       case write: DataWritingCommandExec if SparkShimLoader.getSparkShims.isPlannedV1Write(write) =>
         // To align with ApplyColumnarRulesAndInsertTransitions#insertTransitions
         ConventionReq.any
+      case _: ColumnarWriteFilesExec.ExecuteWriteCompatible =>
+        ConventionReq.of(
+          ConventionReq.RowType.Any,
+          ConventionReq.BatchType.Is(BackendsApiManager.getSparkPlanExecApiInstance.batchType))
```
zhztheplayer, line 164: The same. Use interface `KnownChildrenConventions` instead.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition/Transitions.scala
```diff
@@ -107,8 +100,13 @@ object Transitions {
   }
 
   private def toBatchPlan(plan: SparkPlan, toBatchType: Convention.BatchType): SparkPlan = {
+    convert(
+      plan,
+      ConventionReq.of(ConventionReq.RowType.Any, ConventionReq.BatchType.Is(toBatchType)))
+  }
+
+  private def convert(plan: SparkPlan, req: ConventionReq): SparkPlan = {
```
zhztheplayer, line 108: I'd suggest rename `convert` which sounds too general. How about ``enforceReq`?

zhztheplayer (commented):
Just some minors. Thanks.

zhztheplayer (approved):

PR #6746, opened 2024-08-07
[CORE] Fix schema mismatch between ReadRelNode and LocalFilesNode

jiangjiangtian:
The field names of baseSchema in the ReadRelNode and the field names of schema in the LocalFilesNode may mismatch, which we have met. The reason is that gluten converts the field names of the baseSchema in the ReadRelNode to lower case but does not convert the field names of the schema in the LocalFilesNode.
The bug will result the reader can't deserialize the column data correctly.


jiangjiangtian:
@kecookier

PHILO-HE (approved):
Looks good!

PR #6756, opened 2024-08-08
[VL] RAS: Add a new built-in cost model that avoids offloading trivial projects if its neighbor nodes fell back

zml1206:

1. Rename `RoughCostModel` to `LegacyCostModel`, set default cost model as legacy.
2. Add new `RoughCostModel`, make projectList all cheap `ProjectExec` has the same cost as `ProjectExecTransform`, to reduce unnecessary c2r and r2c for avoid performance regressions.

## How was this patch tested?
UT


zhztheplayer:
Thanks. I've retitled the PR to make things clearer.

gluten-core/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
```diff
@@ -0,0 +1,50 @@
...
+package org.apache.gluten.planner.cost
+
+import org.apache.gluten.extension.columnar.enumerated.RemoveFilter
+import org.apache.gluten.extension.columnar.transition.{ColumnarToRowLike, RowToColumnarLike}
+import org.apache.gluten.utils.PlanUtil
+
+import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, NamedExpression}
+import org.apache.spark.sql.execution.{ColumnarToRowExec, ProjectExec, RowToColumnarExec, SparkPlan}
+
+class RoughCostModel extends LegacyCostModel {
```
zhztheplayer, line 26: Let's add a base class `abstract class LongCostModel` then make the two implementations inherit from it.

gluten-core/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  override def selfLongCostOf(node: SparkPlan): Long = {
+    node match {
+      case _: RemoveFilter.NoopFilter =>
+        // To make planner choose the tree that has applied rule PushFilterToScan.
+        0L
+      case ProjectExec(projectList, _) if projectList.forall(isCheapExpression) => 10L
```
zhztheplayer, line 33: Could add a short comment here to explain about the coster.

zhztheplayer (commented):
And would you like to add a test suite for `RoughCostModel`? Could be placed as `org.apache.gluten.execution.VeloxRoughCostModelSuite`

zml1206:
This UT is very strange. My local verification was normal, but it failed on github. I added logs later in #6760 , output the plan, and found that the scan node was still offloaded, but I added the configuration GlutenConfig.COLUMNAR_FILESCAN_ENABLED.key -> "false" .
```
+- ^(4230) ProjectExecTransformer [c1#80388 AS c3#80394]
   +- ^(4230) NativeFileScan parquet spark_catalog.default.tmp1[c1#80388] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/__w/incubator-gluten/incubator- gluten/backends-velox/spark-wareh..., PartitionFilters: [], PushedFilters: [], ReadSchema: struct<c1:int>
```
cc @zhztheplayer Do you know the reason? Thank you.

zhztheplayer:
@zml1206 

It's because of a bug that makes configuration changes to RAS planner not take effect. https://github.com/apache/incubator-gluten/pull/6766 should fix the issue. Thanks.

zhztheplayer (approved):
Thanks!

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1320,7 +1320,7 @@ object GlutenConfig {
         "Experimental: The class name of user-defined cost model that will be used by RAS. " +
           "If not specified, a rough built-in cost model will be used.")
```
zhztheplayer, line 1321: nit: "If not specified, a legacy built-in cost model that exhaustively offloads computations will be used"

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRoughCostModelSuite.scala
```diff
@@ -0,0 +1,54 @@
...
+    super.afterAll()
+  }
+
+  override protected def sparkConf: SparkConf = super.sparkConf
+    .set(GlutenConfig.RAS_ENABLED.key, "true")
+    .set(GlutenConfig.RAS_COST_MODEL.key, "rough")
+
+  test("fallback trivial project if its neighbor nodes fell back") {
+    withSQLConf(GlutenConfig.COLUMNAR_FILESCAN_ENABLED.key -> "false") {
+      runQueryAndCompare("select c1 as c3 from tmp1") {
```
zhztheplayer, line 49: can we simply use `select c1 from tmp1`?

zhztheplayer (commented):
Just some minors left. Thanks.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRoughCostModelSuite.scala
zml1206: `select c1 from tmp1` will remove project.

zhztheplayer (approved):

PR #6766, opened 2024-08-09
[VL] RAS: Renew validator instance for each rule applier call

zhztheplayer:
`RasOffload.validators` is now initialized only once in the same process when the first time it's accessed. This causes bug that RAS planner never respond to validation-related session configuration changes.

Issue is found in PR https://github.com/apache/incubator-gluten/pull/6756


zml1206:
I don’t understand why the local test is normal, can you explain it? Thank you.

zhztheplayer:
> I don’t understand why the local test is normal, can you explain it? Thank you.

Because CI's test ran other RAS tests first and initialized the validator by default configuration. Then when it ran `VeloxRoughCostModelSuite`, `GlutenConfig.COLUMNAR_FILESCAN_ENABLED.key -> "false"` was actually not used since the validator instance was already created.

JkSelf (approved):

Issue #6768, opened 2024-08-09
[CH] A bad case for joining with mixed join conditions


lgbo-ustc:
### Backend

CH (ClickHouse)

### Bug description

[Expected behavior] and [actual behavior].

We met a query which has  real bad performace on join in production environment . The query looks like below
```sql
select * from t1 left join t2 on 
t1.uid = t2.uid and (t1.id1 = t2.id1 or t1.id2 = t2.id2 or t1.id3 = t2.id3)
```

There are two main problems.

First, The right table is very large, over 5,000,000,000 rows. Using it to build the join hash table is very resource intensive

Second, when only apply join condition `t1.uid = t2.uid`, it could bring a very large matching results, >  5,000,000,000 * 100. But after apply filter condition `(t1.id1 = t2.id1 or t1.id2 = t2.id2 or t1.id3 = t2.id3)` on this matching result, less then 10000000 rows left.

### Spark version

None

lgbo-ustc:
There will be server PRs to solve this problem

PR #6770, opened 2024-08-09
[GLUTEN-6768][CH] Try to reorder hash join tables based on AQE statistics

lgbo-ustc:

Fixes: #6768

If the gap of two tables‘ sizes is too large, try to reorder the tables, make sure the smaller table is used to build the hash table.

With AQE, we could get the row number from `ShuffleQueryStageExec`

## How was this patch tested?

unit tests


backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenClickHouseTPCHSuite.scala
```diff
@@ -531,5 +531,33 @@ class GlutenClickHouseTPCHSuite extends GlutenClickHouseTPCHAbstractSuite {
     spark.sql("drop table t1")
     spark.sql("drop table t2")
   }
+
+  test("GLUTEN-6768 rerorder hash join") {
```
zzcclp, line 535: maybe check whether the join order is right?

zzcclp (approved):
LGTM

PR #6783, opened 2024-08-12
[GLUTEN-6736][VL] Phase 2: Minimize lock scope in ListenableArbitrator

### LINKED ISSUE #6736: [VL] ABBA deadlock by race between task memory manager's object lock and client locks
Incorporates https://github.com/apache/incubator-gluten/issues/6622.

A possible fix will include two parts:

1. In Java code, make all client threads under the same Spark task thread acquire for one single lock in `ManagedReservationListener`
2. In C++ code, remove the mutex in `SparkAllocationListener`

zhztheplayer:
Phase 2 of the task fixing #6736 . The new code avoids the client side lock in `ListenableArbitrator` from being acquired when the arbitrator is calling / is called by the delegated listener. This together with https://github.com/apache/incubator-gluten/pull/6741 should fix #6736 .

### DIFF
--- a/cpp/velox/memory/VeloxMemoryManager.cc
+++ b/cpp/velox/memory/VeloxMemoryManager.cc
@@ -91,33 +91,33 @@ class ListenableArbitrator : public velox::memory::MemoryArbitrator {
 
   bool growCapacity(velox::memory::MemoryPool* pool, uint64_t targetBytes) override {
     velox::memory::ScopedMemoryArbitrationContext ctx(pool);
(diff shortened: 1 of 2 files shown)


marin-ma (approved):
Thanks!

zhouyuan:
Will this impact the memory allocation performance?

zhztheplayer:
> Will this impact the memory allocation performance?

Benchmark doesn't show any difference. Which is reasonable as the lock stuffs shouldn't put visible overhead before.

PR #6787, opened 2024-08-12
[GLUTEN-6768][CH] Try to use multi join on clauses instead of inequal join condition

lgbo-ustc:

Fixes: #6768

Transform a join with inequal condition  into multi join on clauses as possible, it could be more efficient. For example convert
```sql
on t1.key = t2.key and (t1.a1 = t2.a1 or t1.a2 = t1.a2 or t1.a3 = t2.a3)
```
to 
```sql
on (t1.key = t2.key and t1.a1 = t2.a1) or (t1.key = t2.key and t1.a2 = t1.a2) or (t1.key = t2.key and t1.a3 = t2.a3)
```

We need to limit the right table size to avoid OOM, because we can only use `hash` join algorithm on multi join on clauses.

## How was this patch tested?

unit tests


lgbo-ustc:
before(run timeout)
image

after
image

liuneng1994 (approved):
LGTM

PR #6788, opened 2024-08-12
[CORE] Following #6745, append some minor code cleanups

zhztheplayer:
The patch mainly moves the `KnownChildrenConventions` / `KnownRowType` up to `ColumnarWriteFilesExec` as its direct dependency to keep trait class `ExecuteWriteCompatible` minimal.

And some minors.

Link #6745


JkSelf (approved):
LGTM. Thanks for your fix.

Issue #6148, opened 2024-06-19
[VL] Simplify JniLibLoader


zhztheplayer:
### Description

Overtime we find that `JniLibLoader` might be over-designed. Given that when a shared library is unable to be loaded, Gluten usually just refuses to do anything (e.g., just crashes). Thus ideally we load all the libraries needed, or we just raise fatal error.

`JniLoader` implemented a transactional loading mechanism that supports loading all the specified libraries together and with a rarely used "roll back" API. These stuffs may be the first priority to remove in the topic to simplify code.

ArnavBalyan:
Hi @zhztheplayer I would like to work on this. Can you give some more clarity, we want to get rid of the transactional loading?

zhztheplayer:
> we want to get rid of the transactional loading?

I think it doesn't make any actual difference with loading the libs one by one. Probably we can test and prove. The issue is somewhat benign as nevertheless we don't have much code in `JniLibLoader.java`. Anyway, feel free to take if you want. Thanks.

ArnavBalyan:
I see thanks! I can do some testing and raise changes

PR #6791, opened 2024-08-12
[GLUTEN-6148] Simplify JniLibLoader loading mechanism for native libraries

ArnavBalyan:
 - Currently Gluten uses an overdesigned library loading mechanism consisting of rollbacks and transactional loading of native libraries.
 - However during failures, an exception is thrown and app simply crashes without need for "rollbacks".
 - Simplified this design, now we simply load libraries sequentially and maintain the existing behaviour of throwing error when there is an issue with loading the libraries. (This intentionally does not change signature of existing used methods of `JniLibLoader` to maintain simplicity in the change)

Fixes: #6148 

## How was this patch tested?
 - Local Spark shell and ran some SparkSQL queries with gluten.
 - Unit tests should cover the rest, needed some feedback on how we can test it better


gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
```diff
@@ -175,160 +150,80 @@ public static synchronized void unloadFromPath(String libPath) {
...
-    public JniLoadTransaction mapAndLoad(String unmappedLibName, boolean requireUnload) {
-      try {
-        final String mappedLibName = System.mapLibraryName(unmappedLibName);
-        load(mappedLibName, requireUnload);
-        return this;
-      } catch (Exception e) {
-        abort();
-        throw new GlutenException(e);
+  public void loadAndCreateLink(String libName, String linkName, boolean requireUnload) {
+    sync.lock();
```
zhztheplayer, line 184: Do we still need this lock as class member? Perhaps we can minimize the lock scope by only making operations on `loadedLibraries` atomic. Say `synchronized (loadedLibraries) {...}` may be enough.

gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
```diff
@@ -72,7 +62,7 @@ public class JniLibLoader {
   private final Set<String> loadedLibraries = new HashSet<>();
   private final Lock sync = new ReentrantLock();
 
-  JniLibLoader(String workDir) {
+  public JniLibLoader(String workDir) {
```
zhztheplayer, line 65: Can we keep the ctor package-private? If no other places are calling it.

zhztheplayer (approved):
Just a little minors. Thanks.

gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
ArnavBalyan: That's a good catch, updated thanks

gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
ArnavBalyan: Done thanks!

zhztheplayer (approved):
I see there are still some unnecessary synchronizations but should be benign overall. So let's merge this first. Thank you for the effort.

ArnavBalyan:
> I see there are still some unnecessary synchronizations but should be benign overall. So let's merge this first. Thank you for the effort.

Ah missed the inner ones, I'll remove in a follow up change

Issue #6368, opened 2024-07-08
Velox backend does not apply/enforce spark.redaction.regex config property when logging out spark config


xushichangdesmond:
### Description

Ideally, spark.redaction.regex should be applied as well

Yohahaha:
would you submit a PR to support it?

ArnavBalyan:
Hi I can add support for this. I see two ways to do this:

1. Handle all logs centrally via glog. Glog supports custom sinks where users can redirect logs to addition sinks. But this logic needs to be hacked to disable the default glog behaviour, introduce a custom sink, and redact information there. Currently there is no simple way to do it since glog does not support it. (I have raised a request in glog)
2. Until #1 can be supported, a custom macro can be introduced for sensitive logs. Something like `SENSITIVE_LOG(INFO)`. Which can be called by callers. This would be hard to enforce since LOG(severity) is used commonly.

I feel #2 may be simpler until glog provides an easy way to do this. Would love to get your thoughts cc @FelixYBW @PHILO-HE

PHILO-HE:
@xushichangdesmond, @ArnavBalyan, if we only care about sensitive information in Spark config, maybe, we can just locate the code places where Spark config properties are printed and redact the log there. 

See the below Gluten's log when I run a test.
```
VeloxRuntime.cc:142] VeloxRuntime session config:
 [spark.sql.session.timeZone, Asia/Shanghai]
 [spark.gluten.ugi.tokens, ]
 [spark.sql.files.ignoreMissingFiles, false]
 [spark.sql.caseSensitive, false]
 [spark.gluten.memory.task.offHeap.size.in.bytes, 17895697066]
 [spark.gluten.sql.columnar.backend.velox.window.type, sort]
 [spark.gluten.sql.session.timeZone.default, Etc/UTC]
 [spark.gluten.ugi.username, root]
 [spark.gluten.sql.debug, true]
```

ArnavBalyan:
Yes, this makes sense, maybe we can keep a util function anyone in the future can use if logging sensitive configs etc. Will wait for @xushichangdesmond, but I think the ask should be fulfilled with this thanks

PR #6793, opened 2024-08-12
[GLUTEN-6368] Redact sensitive configs when calling `gluten::printConfig`

ArnavBalyan:
- Currently Spark supports `spark.redaction.regex` which allows redacting sensitive info like passwords, tokens etc.
- However gluten does not enforce this and logs in plaintext when debug mode is enabled.
- Added support to redact in printconfig which is being used in the codebase to log the config at various places.
- User can provide a regex which will be matched and a redacted string will be visitble.

Fixes: #6368

## How was this patch tested?
 - Local Spark shell
 Before:
image

After: (Additional regex was supplied to match `spark.gluten.ugi.tokens`)
image

### DIFF
--- a/cpp/core/config/GlutenConfig.cc
+++ b/cpp/core/config/GlutenConfig.cc
@@ -16,12 +16,11 @@
  */
 
 #include <jni.h>
-
+#include <optional>
+#include <regex>
 #include "compute/ProtobufUtils.h"
 #include "config.pb.h"
 #include "jni/JniError.h"
-#include <regex>
-#include <optional>
 
 namespace gluten {
 
@@ -29,35 +28,36 @@ const std::string REDACTED_VALUE = "*********(redacted)";
 const std::string REGEX_REDACT_KEY = "spark.gluten.redaction.regex";
 
 std::unordered_map<std::string, std::string>
-  parseConfMap(JNIEnv* env, const uint8_t* planData, const int32_t planDataLength) {
-    std::unordered_map<std::string, std::string> sparkConfs;
-    ConfigMap pConfigMap;
-    gluten::parseProtobuf(planData, planDataLength, &pConfigMap);
-    for (const auto& pair : pConfigMap.configs()) {
-      sparkConfs.emplace(pair.first, pair.second);
+parseConfMap(JNIEnv* env, const uint8_t* planData, const int32_t planDataLength) {
+  std::unordered_map<std::string, std::string> sparkConfs;
+  ConfigMap pConfigMap;
+  gluten::parseProtobuf(planData, planDataLength, &pConfigMap);
+  for (const auto& pair : pConfigMap.configs()) {
+    sparkConfs.emplace(pair.first, pair.second);
   }
+
   return sparkConfs;
 }
 
 std::optional<std::regex> getRedactionRegex(const std::unordered_map<std::string, std::string>& conf) {
-    auto it = conf.find(REGEX_REDACT_KEY);
-    if (it != conf.end()) {
-        return std::regex(it->second);
-    }
-    return std::nullopt;
+  auto it = conf.find(REGEX_REDACT_KEY);
+  if (it != conf.end()) {
+    return std::regex(it->second);
+  }
+  return std::nullopt;
 }
 
 std::string printConfig(const std::unordered_map<std::string, std::string>& conf) {
   std::ostringstream oss;
   oss << std::endl;
 
-    auto redactionRegex = getRedactionRegex(conf);
+  auto redactionRegex = getRedactionRegex(conf);
 
   for (const auto& [k, v] : conf) {
     if (redactionRegex && std::regex_match(k, *redactionRegex)) {
-        oss << " [" << k << ", " << REDACTED_VALUE << "]\n";
+      oss << " [" << k << ", " << REDACTED_VALUE << "]\n";
     } else {
-        oss << " [" << k << ", " << v << "]\n";
+      oss << " [" << k << ", " << v << "]\n";
     }
   }
   return oss.str();
--- a/cpp/core/config/GlutenConfig.cc
+++ b/cpp/core/config/GlutenConfig.cc
@@ -22,10 +22,21 @@
 #include "config.pb.h"
 #include "jni/JniError.h"
 
+namespace {
+
+const std::string REGEX_REDACT_KEY = "spark.gluten.redaction.regex";
+std::optional<std::regex> getRedactionRegex(const std::unordered_map<std::string, std::string>& conf) {
+  auto it = conf.find(REGEX_REDACT_KEY);
+  if (it != conf.end()) {
+    return std::regex(it->second);
+  }
+  return std::nullopt;
+}
+} // namespace anonymous
+
 namespace gluten {
 
 const std::string REDACTED_VALUE = "*********(redacted)";
-const std::string REGEX_REDACT_KEY = "spark.gluten.redaction.regex";
 
 std::unordered_map<std::string, std::string>
 parseConfMap(JNIEnv* env, const uint8_t* planData, const int32_t planDataLength) {
@@ -37,14 +48,6 @@ parseConfMap(JNIEnv* env, const uint8_t* planData, const int32_t planDataLength)
   }
 
   return sparkConfs;
-}
-
-std::optional<std::regex> getRedactionRegex(const std::unordered_map<std::string, std::string>& conf) {
-  auto it = conf.find(REGEX_REDACT_KEY);
-  if (it != conf.end()) {
-    return std::regex(it->second);
-  }
-  return std::nullopt;
 }
 
 std::string printConfig(const std::unordered_map<std::string, std::string>& conf) {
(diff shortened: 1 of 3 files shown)


shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -673,7 +674,8 @@ object GlutenConfig {
       // gcs config
       SPARK_GCS_STORAGE_ROOT_URL,
       SPARK_GCS_AUTH_TYPE,
-      SPARK_GCS_AUTH_SERVICE_ACCOUNT_JSON_KEYFILE
+      SPARK_GCS_AUTH_SERVICE_ACCOUNT_JSON_KEYFILE,
+      GLUTEN_REGEX_LOG_REDACTION
```
PHILO-HE, line 678: @ArnavBalyan, I think we should just use the existing redaction config key in Spark. Thus, this Spark config will be respected in Gluten.

See the below example to pass Spark's session timezone:
https://github.com/apache/incubator-gluten/pull/6793/files#diff-21453c657b1ccfba45172821a434e5e7cdf007a49b440155514e6641b837d77cL651

PHILO-HE (commented):
@ArnavBalyan, thanks for your work! One comment is posted.
@xushichangdesmond, could you take a review?

cpp/core/config/GlutenConfig.cc
```diff
@@ -16,12 +16,17 @@
...
+#include <optional>
+#include <regex>
 #include "compute/ProtobufUtils.h"
 #include "config.pb.h"
 #include "jni/JniError.h"
 
 namespace gluten {
+
+const std::string REDACTED_VALUE = "*********(redacted)";
+const std::string REGEX_REDACT_KEY = "spark.gluten.redaction.regex";
```
jinchengchenghh, line 28: Duplicated https://github.com/apache/incubator-gluten/pull/6793/files#diff-738908e381e317bdb343354905fd64c5af6c4a58e035376d5dbd2df5313b4a06R67

cpp/core/config/GlutenConfig.cc
```diff
@@ -34,12 +39,28 @@ parseConfMap(JNIEnv* env, const uint8_t* planData, const int32_t planDataLength)
   return sparkConfs;
 }
 
+std::optional<std::regex> getRedactionRegex(const std::unordered_map<std::string, std::string>& conf) {
+  auto it = conf.find(REGEX_REDACT_KEY);
```
jinchengchenghh, line 43: Please move this one to anonymous namespace.

jinchengchenghh:
Thanks. Just several nits.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
ArnavBalyan: Thanks, I was trying to do this, then realized some compatibility problems with C++ `regex` library vs Scala `matching` library used by Spark. 
For example: Spark typically uses something like `(?i)secret\|password\|token\|access[.]key`, but this fails in C++ regex library because `(?i)` is not supported and users need to represent the regex in a different manner. To prevent such conflicts, for now I had created a different config specific to gluten. I agree, this is not the most elegant way, can switch to single config as well, let me know thanks!

cpp/core/config/GlutenConfig.cc
ArnavBalyan: Fixed

cpp/core/config/GlutenConfig.cc
ArnavBalyan: Done thanks!

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
PHILO-HE: @ArnavBalyan, can we use boost::regex instead? Such pattern should be supported by boost.

ArnavBalyan: Yes we can use boost, there are still be parity differences, but these would be less common. Also just noticed the regex redact property is present in spark `core` config, it seems currently the `SQL` configs flow in, we might have to wire up the core configs to flow into gluten.

PHILO-HE: @ArnavBalyan, yes, we cannot guarantee the compatibility for all regex pattern. But at least, this default regex should be supported in Gluten.

xushichangdesmond: yes theres differences in the c++ regex library, we could probably use a fallback to something sensible as a compromise.

xushichangdesmond:
thanks for this; think it looks good though I dont have approval rights for this repo.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
ArnavBalyan: Thanks for the review @PHILO-HE @xushichangdesmond, I'll wire up the spark core config and use that by default, maybe we can improve this to fallback to gluten specific configs if needed in the future

ArnavBalyan:
> thanks for this; think it looks good though I dont have approval rights for this repo.

Ack thanks, I will incorporate @PHILO-HE feedback and use the gluten specific configs, should not be much effort once the Spark core-configs can be exposed to the GlutenConfig, I'll update this shortly thanks!

cpp/core/config/GlutenConfig.cc
```diff
@@ -16,12 +16,28 @@
...
-
+#include <optional>
+#include <regex>
 #include "compute/ProtobufUtils.h"
 #include "config.pb.h"
 #include "jni/JniError.h"
 
+namespace {
+
+const std::string REGEX_REDACT_KEY = "spark.gluten.redaction.regex";
```
Yohahaha, line 27: move config key definition to GlutenConfig.h

Yohahaha: and follow others naming.

cpp/core/config/GlutenConfig.cc
```diff
@@ -16,12 +16,28 @@
...
+  if (it != conf.end()) {
+    return std::regex(it->second);
+  }
+  return std::nullopt;
+}
+} // namespace
+
 namespace gluten {
+
+const std::string REDACTED_VALUE = "*********(redacted)";
```
Yohahaha, line 39: ditto.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
Yohahaha: > For example: Spark typically uses something like (?i)secret\|password\|token\|access[.]key, but this fails in C++ regex library because (?i) is not supported and users need to represent the regex in a different manner.

could we overwrite Spark's default pattern with a valid regex patten in native?

Yohahaha (commented):
thank you, left some comments.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
ArnavBalyan: There are alot of incompatibilities, it's just one of the things where c++ and java don't have complete parity, I would rather, we provide limited support for the same config, or allow users to provide regex supported by c++ using a different properties. Option 1 is better for now and later we can add support

cpp/core/config/GlutenConfig.cc
ArnavBalyan: This will be changed, we will no longer rely on this config

Yohahaha: All configs from GlutenConfig.scala should mapping to GlutenConfig.h/VeloxConfig.h, let's keep config consistent.

ArnavBalyan:
cc @PHILO-HE, could you please review, wired up to existing spark confs thanks!

PHILO-HE:
@ArnavBalyan, thanks for your updating! Did you verify that Spark's this config property takes effect in Gluten in a test?
BTW, let's use boost regex, as mentioned before. The boost lib has been included in some cmake files. So I guess you just need to add its header and create boost regex parser in the code.

Yohahaha (approved):
thank you!

ArnavBalyan:
> @ArnavBalyan, thanks for your updating! Did you verify that Spark's this config property takes effect in Gluten in a test? BTW, let's use boost regex, as mentioned before. The boost lib has been included in some cmake files. So I guess you just need to add its header and create boost regex parser in the code.

Yes this was tested, have switched to regex boost, we are able to see redaction thanks!
image

zhztheplayer:
> Yes this was tested, have switched to regex boost, we are able to see redaction with more flexible regex expressions thanks!

That sounds great. And does that mean we can use the same config key with vanilla Spark ?

ArnavBalyan:
> > Yes this was tested, have switched to regex boost, we are able to see redaction with more flexible regex expressions thanks!
> 
> That sounds great. And does that mean we can use the same config key with vanilla Spark ?

Yes, the same config key (that vanilla spark supports) will now also work with Gluten, no additional changes needed for Gluten thanks

PHILO-HE (approved):
Looks good!

PR #6797, opened 2024-08-13
 [VL] Fix Arrow ColumnarBatch cannnot revoke rowIterator correctly

jinchengchenghh:
`ColumnarBatchRow` first column is IndicatorVector, so we cannot revoke `rowIterator` function correctly, meet this exception
```
java.lang.UnsupportedOperationException
	at org.apache.gluten.columnarbatch.IndicatorVectorBase.getBoolean(IndicatorVectorBase.java:79)
	at org.apache.spark.sql.vectorized.ColumnarBatchRow.getBoolean(ColumnarBatchRow.java:97)
	at org.apache.spark.sql.vectorized.ColumnarBatchRow.get(ColumnarBatchRow.java:155)
	at org.apache.gluten.columnarbatch.ColumnarBatches.toString(ColumnarBatches.java:406)
```
```
public class ColumnarBatch implements AutoCloseable {
  protected int numRows;
  protected final ColumnVector[] columns;

  // Staging row returned from `getRow`.
  protected final ColumnarBatchRow row;

public final class ColumnarBatchRow extends InternalRow {
  public int rowId;
  private final ColumnVector[] columns;
```
transferVector to the `ColumnarBatch` fields columns and ColumnarBatchRow row.

### DIFF
--- a/gluten-data/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
+++ b/gluten-data/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
@@ -34,6 +34,7 @@ import org.apache.spark.sql.catalyst.InternalRow;
 import org.apache.spark.sql.catalyst.expressions.UnsafeRow;
 import org.apache.spark.sql.vectorized.ColumnVector;
 import org.apache.spark.sql.vectorized.ColumnarBatch;
+import org.apache.spark.sql.vectorized.ColumnarBatchRowUtil;
 
 import java.lang.reflect.Field;
 import java.util.Arrays;
@@ -100,6 +101,10 @@ public class ColumnarBatches {
         newVectors[i] = from.column(i);
       }
       FIELD_COLUMNS.set(target, newVectors);
+      // Light batch does not need the row.
+      if (isHeavyBatch(target)) {
+        ColumnarBatchRowUtil.setColumnarBatchRow(newVectors, target);
+      }
     } catch (IllegalAccessException e) {
       throw new GlutenException(e);
     }
@@ -202,7 +207,8 @@ public class ColumnarBatches {
 
       // populate new vectors to input
       transferVectors(output, input);
-      return input;
+
+      return output;
     }
   }
 
--- a/gluten-data/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
+++ b/gluten-data/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
@@ -16,7 +16,6 @@
  */
 package org.apache.gluten.columnarbatch;
 
-import org.apache.gluten.exception.GlutenException;
 import org.apache.gluten.runtime.Runtime;
 import org.apache.gluten.runtime.Runtimes;
 import org.apache.gluten.utils.ArrowAbiUtil;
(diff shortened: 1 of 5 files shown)


zhztheplayer:
Could you please describe the intention in PR description?

zhztheplayer:
Hi @jinchengchenghh. thanks for the update. Would you like to help add a test case in `ColumnarBatchTest`? Then we can avoid regression after fixing the issue.

gluten-data/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
```diff
@@ -186,23 +185,14 @@ private static ColumnarBatch load(BufferAllocator allocator, ColumnarBatch input
...
-
       // close the input one
       for (long i = 0; i < giv.refCnt(); i++) {
         input.close();
       }
 
-      // populate new vectors to input
-      transferVectors(output, input);
-      return input;
+      return output;
```
zhztheplayer, line 195: Thank you for writing the test and I finally got what happened here.

`transferVectors` was used for a reason. Once `ensureLoaded` is called, we should prevent caller code from using the input batch again after the method returns to do further operations (e.g., load it again, etc.) so the code behavior will be more predictable . So it's likely better not to remove it in the patch.

The root cause of the issue is actually we didn't update the column vectors used by `ColumnarBatchRow` in `transferVectors`. A proper way is to update them as well as `ColumnarBatch#columns`.

And I push a commit to the PR to untangle test code, please pull that one before next push. Thanks.

jinchengchenghh: `ColumnarBatchRow` is a protected member in `ColumnarBatch`, we cannot access it.
We close the `input` in `loaded` function, so if the caller revoke the input, it will fail.

jinchengchenghh: ok, I know, do you think we should use reflection to invoke `ColumnarBatchRow`?

jinchengchenghh: Further more, if we just need the caller don't use `input` again, we can set `input` to null. What's other concern I missed?

jinchengchenghh: Now the caller release input batch but not close output batch, to avoid change too much, I will try to transfer vector to ColumnarBatchRow by reflection

zhztheplayer: > Further more, if we just need the caller don't use `input` again, we can set `input` to null. What's other concern I missed?

It's impossible... Java/Scala don't allow you to do that.

zhztheplayer: > Now the caller release input batch but not close output batch, to avoid change too much, I will try to transfer vector to ColumnarBatchRow by reflection

Sounds good to me.

gluten-data/src/main/java/org/apache/spark/sql/vectorized/ColumnarBatchRowUtil.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.vectorized;
+
+import org.apache.gluten.exception.GlutenException;
+
+import java.lang.reflect.Field;
+
+public class ColumnarBatchRowUtil {
```
zhztheplayer, line 23: I think we don't need this file... As `ColumnarBatches` is already a utility class of `ColumnarBatch`

We can just do the refection inline in method `transferVectors`

jinchengchenghh: Because `ColumnarBatchRow` is protected class in spark3.2, so we must use it in same package

zhztheplayer: If so, Let's move method `transferVectors` to this file then rename it with `ColumnarBatchUtil` or something. To store the reflection code together.

Additionally, we may need to set `rowId` as well.

jinchengchenghh: Now I only transfer `ColumnarBatchRow` for Heavy ColumnarBatch, do we need this restriction?

jinchengchenghh: I add the rowId but not add test for it because we cannot revoke any ColumnarBatch API to test it. We could only call `ColumnarBatchRow`, but it is protected. Use reflection to test it is too complex.

zhztheplayer (approved):
LGTM, thanks for fixing this!

PR #6799, opened 2024-08-13
[VL] Add shuffle writer type to ColumnarExchange display string

marin-ma:
To distinguish between the hash and sort shuffle writer types more clearly, add "hash" or "sort" to the operator's display string, and remove `[id=#$id]` as it's the same as `[plan_id=#$id]` in `Exchange` 

image
image

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
@@ -32,7 +32,7 @@ import org.apache.spark.sql.catalyst.plans.logical.Statistics
 import org.apache.spark.sql.catalyst.plans.physical._
 import org.apache.spark.sql.execution.CoalesceExec.EmptyPartition
 import org.apache.spark.sql.execution.exchange._
-import org.apache.spark.sql.execution.metric.{SQLMetric, SQLShuffleWriteMetricsReporter}
+import org.apache.spark.sql.execution.metric.SQLShuffleWriteMetricsReporter
 import org.apache.spark.sql.metric.SQLColumnarShuffleReadMetricsReporter
 import org.apache.spark.sql.vectorized.ColumnarBatch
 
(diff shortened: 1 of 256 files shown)


gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
```diff
@@ -141,8 +141,13 @@ case class ColumnarShuffleExchangeExec(
...
   }
 
-  override def stringArgs: Iterator[Any] =
-    super.stringArgs ++ Iterator(s"[id=#$id]")
+  override def stringArgs: Iterator[Any] = {
+    val shuffleWriterType = {
+      if (useSortBasedShuffle) GlutenConfig.GLUTEN_SORT_SHUFFLE_WRITER
+      else GlutenConfig.GLUTEN_HASH_SHUFFLE_WRITER
+    }
+    super.stringArgs ++ Iterator(s"$shuffleWriterType")
```
jinchengchenghh, line 149: Can we use `[shuffle_writer_type=sort]` instead?

jinchengchenghh:
Thanks! Just a little nit.

gluten-core/src/main/scala/org/apache/spark/sql/execution/ColumnarShuffleExchangeExec.scala
```diff
@@ -141,8 +141,13 @@ case class ColumnarShuffleExchangeExec(
...
   }
 
-  override def stringArgs: Iterator[Any] =
-    super.stringArgs ++ Iterator(s"[id=#$id]")
+  override def stringArgs: Iterator[Any] = {
+    val shuffleWriterType = {
+      if (useSortBasedShuffle) GlutenConfig.GLUTEN_SORT_SHUFFLE_WRITER
+      else GlutenConfig.GLUTEN_HASH_SHUFFLE_WRITER
+    }
+    super.stringArgs ++ Iterator(s"[shuffle_writer_type=$shuffleWriterType]")
```
jinchengchenghh, line 149: You may also need to update the `tpch-approved-plan`

jinchengchenghh (approved):

PR #6800, opened 2024-08-13
[VL] Update vcpkg glog cmake option to remove glog change from modify_velox.patch

jinchengchenghh:
The shared gluten build does not need glog change in modify_velox.patch because it build glog with component shared, but vcpkg builds without shared before, we could change vcpkg build cmake option to remove it in modify_velox.patch to avoid rebase conflict

### DIFF
--- a/ep/build-velox/src/modify_velox.patch
+++ b/ep/build-velox/src/modify_velox.patch
@@ -102,30 +102,6 @@ index cc180b86c..7ca6fa727 100644
  if(NOT TARGET gflags::gflags)
    # This is a bit convoluted, but we want to be able to use gflags::gflags as a
    # target even when velox is built as a subproject which uses
-diff --git a/velox/common/process/tests/CMakeLists.txt b/velox/common/process/tests/CMakeLists.txt
-index 23ef279c2..7e4c2f2b2 100644
---- a/velox/common/process/tests/CMakeLists.txt
-+++ b/velox/common/process/tests/CMakeLists.txt
-@@ -24,4 +24,6 @@ target_link_libraries(
-     fmt::fmt
-     velox_time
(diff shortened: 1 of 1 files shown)


FelixYBW:
Can you add some background info? Does any Velox features enable us to remove this?

PHILO-HE:
@jinchengchenghh, please do a rebase.

jinchengchenghh:
vcpkg expects link static glog, so we cannot relax to use glog shared library.

jinchengchenghh:
gluten-it TPCH/DS core dump
```
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.
#
#
#  SIGSEGV (0xb) at pc=0x00007feb7dc2c806, pid=6380, tid=0x00007fec30fff700
#
# JRE version: OpenJDK Runtime Environment (8.0_422-b05) (build 1.8.0_422-8u422-b05-1~20.04-b05)
# Java VM: OpenJDK 64-Bit Server VM (25.422-b05 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libvelox.so+0x50df806]  google::LogMessage::Init(char const*, int, int, void (google::LogMessage::*)())+0x3f6
#
# Core dump written. Default location: /__w/incubator-gluten/incubator-gluten/tools/gluten-it/core or core.6380
#
# /__w/incubator-gluten/incubator-gluten/tools/gluten-it/hs_err_pid6380.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
sbin/gluten-it.sh: line 50:  6380 Aborted                 (core dumped) $JAVA_HOME/bin/java
```

PR #6829, opened 2024-08-14
[VL] Support create temporary function for native hive udf

marin-ma:
1. Create a native Velox udf, name it to the full name of the hive udf (package + classname) e.g. `org.apache.hadoop.hive.ql.udf.UDFLog10`
2. Register the hive udf by `create temporary function` with user-defined function name. e.g. `create temporary function hive_log10 as org.apache.hadoop.hive.ql.udf.UDFLog10`
3. Execute the sql, the execution will be offloaded to the native hive udf if validation passed, or fallback to the original Java Hive udf.


marin-ma:
Run Gluten Clickhouse CI

FelixYBW:
Works perfectly! Thank you.

Can you update the document as well?

FelixYBW (approved):

marin-ma:
> Can you update the document as well?

@FelixYBW Will update in the next pr with removing the native registration.

liujiayi771:
@marin-ma If a UDF has 2 signatures, does getNumUdf return 1 or 2?

marin-ma:
> @marin-ma If a UDF has 2 signatures, does getNumUdf return 1 or 2?

@liujiayi771 It should return 2. The number of udf represents how many signatures will be registered by the UdfRegisterer. For example, in the implementation of `registerSignatures` if we call `velox::registerFucntion` to register function A with 2 signatures, and function B with 1 signature, then getNumUdf should return 3.

PR #6833, opened 2024-08-14
[GLUTEN-6834][CORE] Remove unused DDL plan that doesn't correspond to Substrait spec

EpsilonPrime:

This is part of a larger effort to unfork Substrait.

## How was this patch tested?

Tested utilizing the available unit tests.


EpsilonPrime:
It's worth noting that DDL operations are connected differently in Substrait so that this code would need to be updated before it could be used anyway.

zhztheplayer:
Hi @rui-mo, where did the code come from, do you remember?

rui-mo:
This was introduced by CH backend with https://github.com/apache/incubator-gluten/commit/614543e4f72d0bbf0ab132de181b6f231db63381. cc: @zzcclp

zhztheplayer (approved):
+1 if the code is unused

zhztheplayer:
@EpsilonPrime 

By the way I still doubt we can unfork Substrait with minimal effort, though it sounds good. Different native accelerators / libraries would have different ways to implement the same logical computation. It's may not be easy to have a single protocol to unify them.

FelixYBW:
> @EpsilonPrime
> 
> By the way I still doubt we can unfork Substrait with minimal effort, though it sounds good. Different native accelerators / libraries would have different ways to implement the same logical computation. It's may not be easy to have a single protocol to unify them at physical level.

we should have the same protocol in long term, substrait should be able to fit all requirements from different frameworks, libraries and accelerators. It's substrait's goal. But now for Gluten it's not urgent.

PR #6803, opened 2024-08-13
[GLUTEN-6600]Fix NPE issue when running window sql

JkSelf:
## What changes were proposed in this pull request?

When running the following sql in yarn client mode, we got the following exception:

```
spark.sql("create temporary view d233 using parquet options ( path 'file:///home/jk/d233.parquet')")
spark.sql("select sum(impression_count_last7days ) from (     SELECT sum(impression_count) OVER (         PARTITION BY user_id,         surface         ORDER BY date RANGE BETWEEN 6 preceding AND CURRENT ROW       ) AS impression_count_last7days     from d233 )").show()
```

```
24/08/12 16:13:26 WARN TaskSetManager: Lost task 86.1 in stage 3.0 (TID 263) (sr246 executor 8): TaskKilled (Stage cancelled: Job aborted due to stage failure: Task 90 in stage 3.0 failed 4 times, most recent failure: Lost task 90.3 in stage 3.0 (TID 254) (sr246 executor 6): java.lang.NullPointerException
        at org.apache.gluten.expression.ExpressionMappings$.expressionsMap(ExpressionMappings.scala:341)
        at org.apache.gluten.expression.ExpressionConverter$.replaceWithExpressionTransformer(ExpressionConverter.scala:54)
        at org.apache.gluten.expression.ExpressionConverter.replaceWithExpressionTransformer(ExpressionConverter.scala)
        at org.apache.gluten.substrait.expression.WindowFunctionNode.setBound(WindowFunctionNode.java:102)
        at org.apache.gluten.substrait.expression.WindowFunctionNode.toProtobuf(WindowFunctionNode.java:180)
        at org.apache.gluten.substrait.rel.WindowRelNode.toProtobuf(WindowRelNode.java:77)
        at org.apache.gluten.substrait.rel.ProjectRelNode.toProtobuf(ProjectRelNode.java:69)
        at org.apache.gluten.substrait.rel.AggregateRelNode.toProtobuf(AggregateRelNode.java:89)
        at org.apache.gluten.substrait.plan.PlanNode.toProtobuf(PlanNode.java:74)
        at org.apache.gluten.backendsapi.velox.VeloxIteratorApi.genFinalStageIterator(VeloxIteratorApi.scala:238)
        at org.apache.gluten.execution.WholeStageZippedPartitionsRDD.$anonfun$compute$1(WholeStageZippedPartitionsRDD.scala:59)
        at org.apache.gluten.utils.Arm$.withResource(Arm.scala:25)
[...]

PR #6837, opened 2024-08-14
[GLUTEN-6803]Init ExpressionUtil.extendedExpressionTransformer in executor side

JkSelf:

Follow up https://github.com/apache/incubator-gluten/pull/6803.

Init ExpressionUtil.extendedExpressionTransformer in GlutenExecutorPlugin side

## How was this patch tested?

local verified.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionMappings.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionMappings.scala
@@ -19,7 +19,7 @@ package org.apache.gluten.expression
 import org.apache.gluten.GlutenConfig
 import org.apache.gluten.backendsapi.BackendsApiManager
 import org.apache.gluten.expression.ExpressionNames._
-import org.apache.gluten.extension.ExpressionExtensionTrait
+import org.apache.gluten.extension.{DefaultExpressionExtensionTransformer, ExpressionExtensionTrait}
 import org.apache.gluten.sql.shims.SparkShimLoader
 
 import org.apache.spark.sql.catalyst.expressions._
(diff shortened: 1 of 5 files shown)


JkSelf:
@zhztheplayer

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionMappings.scala
```diff
@@ -354,6 +354,5 @@ object ExpressionMappings {
       .toMap[Class[_], String]
   }
 
-  var expressionExtensionTransformer: ExpressionExtensionTrait =
-    DefaultExpressionExtensionTransformer()
+  var expressionExtensionTransformer: ExpressionExtensionTrait = _
```
zhztheplayer, line 357: I'd suggest to use getter and setters to avoid directly accessing the variable.

In the setter, we can do a check to make sure it was not set.
Similarly in the getter, we can do a check to make sure it was set.

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionMappings.scala
```diff
@@ -354,6 +354,16 @@ object ExpressionMappings {
       .toMap[Class[_], String]
   }
 
-  var expressionExtensionTransformer: ExpressionExtensionTrait =
-    DefaultExpressionExtensionTransformer()
+  private var expressionExtensionTransformer: Option[ExpressionExtensionTrait] = None
+
+  def getExpressionExtensionTransformer: ExpressionExtensionTrait = {
+    expressionExtensionTransformer.getOrElse(new DefaultExpressionExtensionTransformer)
```
zhztheplayer, line 360: Can we throw when it's none? Otherwise we don't make it sure whether the change (setting it in executor side) takes effect. Say if we throw when none, then we must set it in both driver and executor sides otherwise the application will fail by error.

JkSelf: @zhztheplayer Yes. Updated.

gluten-core/src/main/scala/org/apache/gluten/expression/ExpressionMappings.scala
JkSelf: @zhztheplayer Added the get and set method.

zhztheplayer (approved):
Thanks!

PR #6847, opened 2024-08-14
[VL] No need to obtain old shrunken memory

boneanxs:

`mm->arbitrator()->shrinkCapacity` should return the shrunken bytes


zhztheplayer (approved):
Thanks!

PR #6855, opened 2024-08-15
[GLUTEN-6849][VL] Call static initializers once in Spark local mode / when session is renewed

### LINKED ISSUE #6849: [VL] Move process-wise static initializers out from VeloxListenerApi
### Description

We have some process-wise static initializers defined in `VeloxListenerApi` or `GlutenPlugin` which works in both Driver/Executor's initializations. This doesn't work well since in Spark local mode the initializers are called twice which is unnecessary and needs synchronizations. 

We should move the initializers that is expected to be run once in the same process out from the places and guard them in some kind of static blocks (e.g. Scala object initialization block).

zhztheplayer:
In Spark local mode, Spark driver and executor are both calling ListenerApi to run static initializers. This could cause UBs in some cases since the initializers were not deigned to be called more than once. This patch fixes the issue by adding a check `inLocalMode` then use it to skip executor side initializer when it's true.

The patch also avoids rerunning the static initializers when Spark session is recreated.

And with essential code cleanups.

Closes #6849

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
@@ -18,7 +18,6 @@ package org.apache.gluten.backendsapi.velox
 
 import org.apache.gluten.GlutenConfig
 import org.apache.gluten.backendsapi.ListenerApi
-import org.apache.gluten.exception.GlutenException
 import org.apache.gluten.execution.datasource.{GlutenOrcWriterInjects, GlutenParquetWriterInjects, GlutenRowSplitter}
 import org.apache.gluten.expression.UDFMappings
 import org.apache.gluten.init.NativeBackendInitializer
@@ -27,6 +26,7 @@ import org.apache.gluten.vectorized.{JniLibLoader, JniWorkspace}
 
 import org.apache.spark.{SparkConf, SparkContext}
 import org.apache.spark.api.plugin.PluginContext
+import org.apache.spark.internal.Logging
 import org.apache.spark.sql.execution.datasources.velox.{VeloxOrcWriterInjects, VeloxParquetWriterInjects, VeloxRowSplitter}
 import org.apache.spark.sql.expression.UDFResolver
 import org.apache.spark.sql.internal.{GlutenConfigUtil, StaticSQLConf}
@@ -34,10 +34,8 @@ import org.apache.spark.util.SparkDirectoryUtil
 
 import org.apache.commons.lang3.StringUtils
 
-import scala.sys.process._
-
-class VeloxListenerApi extends ListenerApi {
-  private val ARROW_VERSION = "1500"
+class VeloxListenerApi extends ListenerApi with Logging {
+  import VeloxListenerApi._
 
   override def onDriverStart(sc: SparkContext, pc: PluginContext): Unit = {
     val conf = pc.conf()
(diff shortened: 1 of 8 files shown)


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -191,4 +137,42 @@ class VeloxListenerApi extends ListenerApi {
...
+  // See org.apache.spark.SparkMasterRegex
+  private val LOCAL_N_REGEX = """local\[([0-9]+|\*)\]""".r
+  private val LOCAL_N_FAILURES_REGEX = """local\[([0-9]+|\*)\s*,\s*([0-9]+)\]""".r
+
+  // TODO: Implement graceful shutdown and remove these flags.
+  //  As spark conf may change when active Spark session is recreated.
+  private val driverInitialized: AtomicBoolean = new AtomicBoolean(false)
+  private val executorInitialized: AtomicBoolean = new AtomicBoolean(false)
+
+  private def inLocalMode(conf: SparkConf): Boolean = {
```
jackylee-ch, line 150: nit: This can be can changed to the same code as `Utils.isLocalMaster`.
```
  def isLocalMaster(conf: SparkConf): Boolean = {
    val master = conf.get("spark.master", "")
    master == "local" || master.startsWith("local[")
  }
```

zhztheplayer: Super helpful. Thanks.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -27,138 +26,67 @@ import org.apache.gluten.vectorized.{JniLibLoader, JniWorkspace}
...
-      val systemNamePattern(systemName) =
-        systemInfoLines.find(_.startsWith("NAME=")).getOrElse("")
-      val systemVersionPattern(systemVersion) =
-        systemInfoLines.find(_.startsWith("VERSION=")).getOrElse("")
-      if (systemName.isEmpty || systemVersion.isEmpty) {
-        throw new GlutenException("Failed to get OS name and version info.")
-      }
-      getLibraryLoaderForOS(systemName, systemVersion, system)
+    val conf = pc.conf
+    if (inLocalMode(conf)) {
```
jackylee-ch, line 69: Maybe we can directly call `SparkResourceUtil.isLocalMaster`?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -27,138 +26,67 @@ import org.apache.gluten.vectorized.{JniLibLoader, JniWorkspace}
...
-    // sql table cache serializer
-    if (conf.getBoolean(GlutenConfig.COLUMNAR_TABLE_CACHE_ENABLED.key, defaultValue = false)) {
-      conf.set(
-        StaticSQLConf.SPARK_CACHE_SERIALIZER.key,
-        "org.apache.spark.sql.execution.ColumnarCachedBatchSerializer")
+    if (!driverInitialized.compareAndSet(false, true)) {
+      // Make sure we call the static initializers only once.
+      logInfo(
+        "Skip rerunning static initializers since they are only supposed to run once." +
+          " You see this message probably because you are creating a new SparkSession.")
```
zml1206, line 47: I don’t quite understand that `onDriverStart` will be called multiple times. Can you explain?

zhztheplayer: It happens in Spark local mode. Spark creates one driver and one executor in that mode, in the current process.

zml1206: I understand this. It should be once for `onDriverStart` and once for `onExecutorStart`, not `onDriverStart` is called twice.

zhztheplayer: Oops, got it wrong.

`onDriverStart` will be called twice when spark session is recreated.
`onExecutorStart` may be called twice when dynamic allocation is enabled, I am not sure about this one.

zml1206: Can you give an example of how spark session is recreated, cloneSession or other?

zml1206: About `onExecutorStart`, I think it will not called twice because dynamic allocation add new executor is a new jvm.

zhztheplayer: > Can you give an example of how spark session is recreated, cloneSession or other?

Please refer to `SparkSession.stop` or `SparkContext.stop`

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
zhztheplayer: I am OK to both... Keeping it would shorten the calling code a little bit

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
zml1206: If just re-create sparkSession, it will not restart the driver. Re-creating sparkContext will restart the driver, but a new sparkConf may be set, so is it better to re-initialize it once?

zhztheplayer: Yes, finally we should remove the flags and do re-initializations. See [my comment](https://github.com/apache/incubator-gluten/blob/4096a53e146981ff092c43b56f456137258632b9/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala#L130-L133) and the issue https://github.com/apache/incubator-gluten/issues/6862

PHILO-HE (approved):
Thanks!

PHILO-HE (approved):

PR #6905, opened 2024-08-17
[GLUTEN-6904] Remove additional synchronizations in JNILibLoader

### LINKED ISSUE #6904: Improve JNILibLoader by removing redundant synchronizations
### Description

Currently JNILibLoader is responsible for loading native libraries, which uses a thread safe synchronizedSet.
We have some redundant synchronised blocks which can be removed to improve performance. These should be removed

ArnavBalyan:
- Since we already moved to synchronizedSet for thread safety, there are some additional synchronizations which are no longer needed, these are benign but can be removed for improved performance
- Removed such synchronizations where we are already thread safe.

Fixes #6904

## How was this patch tested?
 - Local build, and sample SQL queries.
 - UTs should cover the rest

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
+++ b/gluten-core/src/main/java/org/apache/gluten/vectorized/JniLibLoader.java
@@ -45,10 +45,8 @@ import scala.runtime.BoxedUnit;
 public class JniLibLoader {
   private static final Logger LOG = LoggerFactory.getLogger(JniLibLoader.class);
 
-  private static final Set<String> LOADED_LIBRARY_PATHS =
-      Collections.synchronizedSet(new HashSet<>());
-  private static final Set<String> REQUIRE_UNLOAD_LIBRARY_PATHS =
-      Collections.synchronizedSet(new LinkedHashSet<>());
+  private static final Set<String> LOADED_LIBRARY_PATHS = new HashSet<>();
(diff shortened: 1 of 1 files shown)


ArnavBalyan:
Thanks @zhztheplayer for pointing it out, if could please take a look

zhztheplayer (commented):
I see we use synchronized collections and `synchronized` keywords at the same time. Were they for similar purpose? If yes, can we only keep one of them? So the code will be easier to understand and maintain. Say if synchronized collections are not enough for our case, we can just use `synchronized` + normal collections.

In addition, if we decide to stay on `synchronized` keywords, I assume its usage in method `mapAndLoad` is not needed now. Since method `load` is already atomic (because `mapLibraryName` should be stateless?).

ArnavBalyan:
> I see we use synchronized collections and `synchronized` keywords at the same time. Were they for similar purpose? If yes, can we only keep one of them? So the code will be easier to understand and maintain. Say if synchronized collections are not enough for our case, we can just use `synchronized` + normal collections.
> 
> In addition, if we decide to stay on `synchronized` keywords, I assume its usage in method `mapAndLoad` is not needed now. Since method `load` is already atomic (because `mapLibraryName` should be stateless?).

Updated thanks, we need synchronized since there were thread unsafe operations being done outside the collections. I have removed the synchronized collection and kept the keyword, there were not much changes so the code flow is more simple now. Also removed it from mapAndLoad, since we have load protected anyway thanks!

Issue #6900, opened 2024-08-17
Use newer version of clang-format(clang-format-18)


xumingming:
### Problem description

There are more recent clang-format-18:

```
root@test:~/Code/gluten# apt search clang-format
Sorting... Done
Full Text Search... Done
clang-format/noble 1:18.0-59~exp2 arm64
  Tool to format C/C++/Obj-C code

clang-format-14/noble 1:14.0.6-19build4 arm64
  Tool to format C/C++/Obj-C code

clang-format-15/noble,now 1:15.0.7-14build3 arm64 [installed]
  Tool to format C/C++/Obj-C code

clang-format-16/noble 1:16.0.6-23ubuntu4 arm64
  Tool to format C/C++/Obj-C code

clang-format-17/noble 1:17.0.6-9ubuntu1 arm64
  Tool to format C/C++/Obj-C code

clang-format-18/noble-updates,now 1:18.1.3-1ubuntu1 arm64 [installed]
  Tool to format C/C++/Obj-C code

libcode-tidyall-plugin-clangformat-perl/noble 0.0.2-2 all
  module to run clang-format using Code::TidyAll
```

### System information

N/A

xumingming:
Velox is using clang-format-18: https://github.com/facebookincubator/velox/blob/54b16e0416ada85dd2904b2d2bee7650a1d27b36/scripts/setup-check.sh#L21

PR #6907, opened 2024-08-18
[GLUTEN-6900] Use clang-format-18

xumingming:

(Fixes: \#6900)

### DIFF
--- a/dev/formatcppcode.sh
+++ b/dev/formatcppcode.sh
@@ -1,13 +1,25 @@
 cd `dirname $0`
 
-# Check if clang-format-15 is installed
-if ! command -v clang-format-15 &> /dev/null
-then
-    echo "clang-format-15 could not be found"
-    echo "Installing clang-format-15..."
-    sudo apt update
-    sudo apt install clang-format-15
+CLANG_FORMAT_CMD=""
+if [ "$(uname)" == "Darwin" ]; then
+    CLANG_FORMAT_CMD="clang-format"
+    if ! command -v $CLANG_FORMAT_CMD &> /dev/null
+    then
+        echo "$CLANG_FORMAT_CMD could not be found"
+        echo "Installing $CLANG_FORMAT_CMD..."
+        sudo brew install $CLANG_FORMAT_CMD
+    fi
+else
+    CLANG_FORMAT_CMD="clang-format-18"
+    if ! command -v $CLANG_FORMAT_CMD &> /dev/null
+    then
+        echo "$CLANG_FORMAT_CMD could not be found"
+        echo "Installing $CLANG_FORMAT_CMD..."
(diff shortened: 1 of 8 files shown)


FelixYBW:
Why do we need to upgrade clang-format?

xumingming:
@FelixYBW The reason is as follows:

1. When I change some c++ code, and submit the PR, the PR will fail at code style check.
2. I noticed there is `dev/formatcppcode.sh`, but it only support Linux
3. After make `dev/formatcppcode.sh` support macOS, I found that clang-format on macOS(which is of version 18) behaves differently from clang-format-15 on Linux.

Do you have suggestions?

dev/formatcppcode.sh
```diff
@@ -1,13 +1,25 @@
...
+else
+    CLANG_FORMAT_CMD="clang-format-18"
+    if ! command -v $CLANG_FORMAT_CMD &> /dev/null
+    then
+        echo "$CLANG_FORMAT_CMD could not be found"
+        echo "Installing $CLANG_FORMAT_CMD..."
+        sudo apt update
+        sudo apt install $CLANG_FORMAT_CMD
+    fi
 fi
```
FelixYBW, line 21: Let's remove the install commands here. It's only for darwin and ubuntu.

FelixYBW:
> @FelixYBW The reason is as follows:
> 
> 1. When I change some c++ code, and submit the PR, the PR will fail at code style check.
> 2. I noticed there is `dev/formatcppcode.sh`, but it only support Linux
> 3. After make `dev/formatcppcode.sh` support macOS, I found that clang-format on macOS(which is of version 18) behaves differently from clang-format-15 on Linux.
> 4. And there seems no easy way to install clang-format-15 on macOS.
> 
> Do you have suggestions?

Different OS has different default clang-format version. e.g. I'm using Ubuntu20.04 which has clang-format-12. Do 12 and 18 have any difference on format? format-12 works well on my side so far.

xumingming:
> > @FelixYBW The reason is as follows:
> > 
> > 1. When I change some c++ code, and submit the PR, the PR will fail at code style check.
> > 2. I noticed there is `dev/formatcppcode.sh`, but it only support Linux
> > 3. After make `dev/formatcppcode.sh` support macOS, I found that clang-format on macOS(which is of version 18) behaves differently from clang-format-15 on Linux.
> > 4. And there seems no easy way to install clang-format-15 on macOS.
> > 
> > Do you have suggestions?
> 
> Different OS has different default clang-format version. e.g. I'm using Ubuntu20.04 which has clang-format-12. Do 12 and 18 have any difference on format? format-12 works well on my side so far.

Yes, they have many difference. Seems not easy to get a same output on macOS, closing it.

Issue #6908, opened 2024-08-18
[VL] stop == StopReason::kBlock || stop == StopReason::kAtEnd || stop == StopReason::kAlreadyTerminated || stop == StopReason::kTerminate


FelixYBW:
### Backend

VL (Velox)

### Bug description

The error raised again, it's usually caused by background thread exit but there is no coredump this time.

```
File: /home/binweiyang/gluten/ep/build-velox/build/velox_ep/velox/exec/Driver.cpp
Line: 370
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorENS1_22CompileTimeEmptyStringEEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox4exec6Driver4nextERSt10shared_ptrINS1_13BlockingStateEE
# 4  _ZN8facebook5velox4exec4Task4nextEPN5folly10SemiFutureINS3_4UnitEEE
# 5  _ZN6gluten24WholeStageResultIterator4nextEv
# 6  Java_org_apache_gluten_vectorized_ColumnarBatchOutIterator_nativeHasNext
# 7  0x00007f3f2c3bf427

	at org.apache.gluten.vectorized.ColumnarBatchOutIterator.nativeHasNext(Native Method)
	at org.apache.gluten.vectorized.ColumnarBatchOutIterator.hasNextInternal(ColumnarBatchOutIterator.java:61)
	at org.apache.gluten.vectorized.GeneralOutIterator.hasNext(GeneralOutIterator.java:37)
	... 31 more

	at org.apache.gluten.vectorized.ColumnarBatchOutIterator.nativeHasNext(Native Method)
	at org.apache.gluten.vectorized.ColumnarBatchOutIterator.hasNextInternal(ColumnarBatchOutIterator.java:61)
	at org.apache.gluten.vectorized.GeneralOutIterator.hasNext(GeneralOutIterator.java:37)
	... 18 more

```
@zhztheplayer 

### Spark version

None

FelixYBW:
OK, bad news, it's caused by io threads in scan. The issue is gone once we set iothread=0.

[...]

PR #6934, opened 2024-08-20
[GLUTEN-6908][VL] Fix error when getting output from a Velox task that is under spilling by background thread

zhztheplayer:
Tends to fix https://github.com/apache/incubator-gluten/issues/6908.

There is a state check to abort Velox / Spark task when Task::next is called on a task that is being paused and under arbitration. The patch fixes the issue by letting Spark wait on the future object returned by Velox task, and continue processing once spill is finished.

Rely on Velox change https://github.com/facebookincubator/velox/pull/10600.


zhztheplayer:
cc @FelixYBW

FelixYBW:
rely on https://github.com/facebookincubator/velox/pull/10600

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -195,7 +195,22 @@ std::shared_ptr<ColumnarBatch> WholeStageResultIterator::next() {
...
+    auto future = velox::ContinueFuture::makeEmpty();
+    auto out = task_->next(&future);
+    if (!future.valid()) {
+      // Not need to wait. Break.
+      vector = std::move(out);
+      break;
+    }
+    // Velox suggested to wait. This might be because another thread (e.g., background io thread) is spilling the task.
+    GLUTEN_CHECK(out == nullptr, "Expected to wait but still got non-null output from Velox task");
+    VLOG(2) << "Velox task " << task_->taskId()
```
FelixYBW, line 209: Can we use Gluten's LOG(INFO) etc. here? Looks VLOG is used several other places in Gluten. We may always use LOG

PR #6962, opened 2024-08-21
[GLUTEN-6961][VL][feat] Add decimal write support for ArrowWritableColumnVector

jinchengchenghh:


gluten-data/src/main/java/org/apache/gluten/vectorized/ArrowWritableColumnVector.java
```diff
@@ -1255,9 +1266,8 @@ void setNull(int rowId) {
       throw new UnsupportedOperationException();
     }
 
-    void setNotNull(int rowId) {
-      throw new UnsupportedOperationException();
-    }
+    // Arrow not need to setNotNull, set the valus is enough.
+    void setNotNull(int rowId) {}
```
zhztheplayer, line 1270: Any reason about the change?

gluten-data/src/main/scala/org/apache/spark/sql/utils/SparkSchemaUtil.scala
```diff
@@ -37,6 +37,16 @@ object SparkSchemaUtil {
...
 
+  def checkSchema(schema: StructType): Boolean = {
+    try {
+      SparkSchemaUtil.toArrowSchema(schema)
+      true
+    } catch {
+      case _: Exception =>
+        false
+    }
+  }
```
zhztheplayer, line 48: What's the error you have seen then decide to catch?

jinchengchenghh: I find here will throw exception, https://github.com/apache/incubator-gluten/blob/main/gluten-data/src/main/scala/org/apache/spark/sql/utils/SparkArrowUtil.scala#L54

But it will not run to here actually, I will drop the change.

gluten-data/src/main/java/org/apache/gluten/vectorized/ArrowWritableColumnVector.java
jinchengchenghh: If we use `MutableProjection`, it will call `MutableColumnarBatchRow`, we can construct `MutableColumnarBatchRow` by `ArrowWritableColumnVector`. https://github.com/apache/spark/blob/master/sql/core/src/main/java/org/apache/spark/sql/execution/vectorized/MutableColumnarRow.java#L286

zhztheplayer: The class is abstract and the method should be overridable by impl classes.

See https://github.com/apache/incubator-gluten/blob/ed3d333c7c357e95f6736f378c972b714bdfb420/gluten-data/src/main/java/org/apache/gluten/vectorized/ArrowWritableColumnVector.java#L1913 as example.

And abstract class `ArrowVectorWriter` doesn't likely have to be abstract. Could make it an interface as it almost doesn't override anything from the hierarchy.

jinchengchenghh: Because it has member `ValueVector`, https://github.com/apache/incubator-gluten/blob/ed3d333c7c357e95f6736f378c972b714bdfb420/gluten-data/src/main/java/org/apache/gluten/vectorized/ArrowWritableColumnVector.java#L1244

jinchengchenghh: The class should not be abstract since its subclass will not implement the functions.

zhztheplayer: Can we route `putDecimal` calls to `accessor.setDecimal`? I am not that familiar with the design here but looks to be a manner we should follow

jinchengchenghh: writer does not have the `setDecimal` function, I copied most code from `https://github.com/apache/spark/blob/master/sql/core/src/main/java/org/apache/spark/sql/execution/vectorized/WritableColumnVector.java#L429` but is different in putByteArray, because Arrow DecimalVector does not support put byte array range.
Put the all bytes is final function in Spark.
```
public abstract int putByteArray(int rowId, byte[] value, int offset, int count);
  public final int putByteArray(int rowId, byte[] value) {
    return putByteArray(rowId, value, 0, value.length);
  }
```
I will add a test for it.

jinchengchenghh: I will override putByteArray but check if it put all bytes, I will refactor it.

jinchengchenghh:
Anymore comments? @zhztheplayer

zhztheplayer (approved):
@jinchengchenghh Would you help fill in the PR description? Thanks.

zhztheplayer:
> Anymore comments? @zhztheplayer

Just some nits, I updated the PR to save some time. Thanks.

PR #6969, opened 2024-08-22
[CH] Ignore cache file with hdfs suite

loneylee:

Because the HDFS service of CI is not very stable, this UT has a small probability.


baibaichen (approved):
LGTM

PR #6974, opened 2024-08-22
[GLUTEN-6957][VL] Fix mvn not found in cache job

PHILO-HE:

In GitHub Actions, when we set an environment variable using echo "VARNAME=value" >> $GITHUB_ENV, the variable is not immediately available in the current step. It will be available in subsequent steps.


rui-mo (approved):
Thanks for the quick fix.

PR #6988, opened 2024-08-23
[GLUTEN-6960][VL] Limit Velox untracked global memory manager's usage 

zhztheplayer:
Limit Velox's global memory manager's usage to 0.75 * Spark overhead memory by default. The overhead memory is calculated by same manner with vanilla Spark.

Velox's global memory manager is used for some global allocations that don't belong to a specific query or task. These allocations are not tracked by Spark task memory manager.

After the patch, `spark.memory.offHeap.size` and `spark.executor.memoryOverhead` (with `spark.executor.memoryOverheadFactor` / `spark.executor.minMemoryOverhead` for higher version Spark, probably) will co-work for Velox backend's memory management.

1. `spark.memory.offHeap.size` limits major of the memory allocations happened during Velox query execution, e.g., hash tables, sort buffers, shuffle buffers, etc.
2. `spark.executor.memoryOverhead` limits the memory allocations that are hardly tracked by Spark task memory manager, or that don't belong to a specific query. E.g., allocations happen during spilling, or possibly the global cache size (unimplemented).

Edit: As the change fails a couple of CI tests, we are setting the internal Velox global poll size limit to `Long.MaxValue` by default in the PR, to bypass the check for sometime, until the relevant bugs from Velox get fixed.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
@@ -48,7 +48,6 @@ abstract class VeloxTPCHTableSupport extends VeloxWholeStageTransformerSuite {
       .set("spark.memory.offHeap.size", "2g")
       .set("spark.unsafe.exceptionOnMemoryLeak", "true")
       .set("spark.sql.autoBroadcastJoinThreshold", "-1")
-      .set("spark.gluten.sql.columnar.backend.velox.glogSeverityLevel", "0")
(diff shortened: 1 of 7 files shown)


zhztheplayer:
Failure https://github.com/apache/incubator-gluten/actions/runs/10519562828/job/29149158021

zhztheplayer:
CH failure doesn't seem to be related cc @zzcclp

zhztheplayer:
Run Gluten Clickhouse CI

marin-ma (approved):

Issue #6995, opened 2024-08-23
[VL] Driver memory leak with soft affinity duplicate reading


jackylee-ch:
### Backend

VL (Velox)

### Bug description

Currently, we monitor TaskEnd/StageEnd events to update and clean up FilePartition and Executor data for duplicate reading. However, in actual execution, we found that the current monitoring logic is slow, and it would cause spark event accumulation. When the events eventually exceed the capacity defined by Spark and they will be discarded. As a result, a large amount of information remains in rddPartitionInfoMap and stageInfoMap and cannot be discarded, which leads to frequent FGC in driver and finnaly OOM.

Bellow is the log for Event discard.
![image](https://github.com/user-attachments/assets/09fed7b4-9af4-4d19-93db-3e9de8c4c879)

Bellow is the memory analyze for our online driver. The driver memory is 2.2G and the memory used by SoftAffinityManager is 1.5G.
<img width="687" alt="image" src="https://github.com/user-attachments/assets/a1f364d7-e49a-4802-a83b-840f650879b8">

### Spark version

Spark-3.5.x

jackylee-ch:
@zhli1142015 We may need to refact the code for duplicate reading or close the duplicated reading by default.

zhli1142015:
Thanks for reporting this, may I ask some questions about this?

> Currently, we monitor TaskEnd/StageEnd events to update and clean up FilePartition and Executor data for duplicate reading.

Does this mean you use partial logic for SA duplicate reading detection? Or you don't use SA at all, SA just impacts you.

> When the events eventually exceed the capacity defined by Spark and they will be discarded.

Yes, if some key events(SparkListenerStageCompleted) are discarded, the middle states would not be cleared, this should be a bug. I can take a look.

> @zhli1142015 We may need to refact the code for duplicate reading or close the duplicated reading by default.

Here is the config to disable the duplicate reading: spark.gluten.soft-affinity.duplicateReadingDetect.enabled

jackylee-ch:
> Does this mean you use partial logic for SA duplicate reading detection? Or you don't use SA at all, SA just impacts you.

We are using velox cache and SA in our environment and found driver memory OOM recentlly. After analyze the heap.dump, we found the `SA duplicate reading detection` is the reason that memory is raising continually.

> Yes, if some key events(SparkListenerStageCompleted) are discarded, the middle states would not be cleared, this should be a bug. I can take a look.

Great, thanks.

[...]

PR #7003, opened 2024-08-26
[GLUTEN-6995][Core] Limit soft affinity duplicate reading detection max cache items

zhli1142015:

Fixes: #6995

## How was this patch tested?
UT.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
@@ -28,7 +28,6 @@ import org.apache.spark.sql.execution.datasources.FilePartition
 
 import com.google.common.cache.{CacheBuilder, CacheLoader, LoadingCache}
 
-import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.locks.ReentrantReadWriteLock
 
@@ -56,18 +55,34 @@ abstract class AffinityManager extends LogLevelUtil with Logging {
 
   lazy val detectDuplicateReading = true
 
-  lazy val maxDuplicateReadingRecords =
-    GlutenConfig.GLUTEN_SOFT_AFFINITY_MAX_DUPLICATE_READING_RECORDS_DEFAULT_VALUE
+  lazy val duplicateReadingMaxCacheItems =
+    GlutenConfig.GLUTEN_SOFT_AFFINITY_DUPLICATE_READING_MAX_CACHE_ITEMS_DEFAULT_VALUE
 
   // rdd id -> patition id, file path, start, length
-  val rddPartitionInfoMap = new ConcurrentHashMap[Int, Array[(Int, String, Long, Long)]]()
+  val rddPartitionInfoMap: LoadingCache[Integer, Array[(Int, String, Long, Long)]] =
+    CacheBuilder
+      .newBuilder()
+      .maximumSize(duplicateReadingMaxCacheItems)
+      .build(new CacheLoader[Integer, Array[(Int, String, Long, Long)]] {
+        override def load(id: Integer): Array[(Int, String, Long, Long)] = {
+          Array.empty[(Int, String, Long, Long)]
+        }
+      })
   // stage id -> execution id + rdd ids: job start / execution end
-  val stageInfoMap = new ConcurrentHashMap[Int, Array[Int]]()
+  val stageInfoMap: LoadingCache[Integer, Array[Int]] =
+    CacheBuilder
+      .newBuilder()
+      .maximumSize(duplicateReadingMaxCacheItems)
+      .build(new CacheLoader[Integer, Array[Int]] {
+        override def load(id: Integer): Array[Int] = {
+          Array.empty[Int]
+        }
+      })
   // final result: partition composed key("path1_start_length,path2_start_length") --> array_host
   val duplicateReadingInfos: LoadingCache[String, Array[(String, String)]] =
     CacheBuilder
       .newBuilder()
-      .maximumSize(maxDuplicateReadingRecords)
+      .maximumSize(duplicateReadingMaxCacheItems)
       .build(new CacheLoader[String, Array[(String, String)]] {
         override def load(name: String): Array[(String, String)] = {
           Array.empty[(String, String)]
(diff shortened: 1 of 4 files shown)


gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
```diff
@@ -56,18 +55,34 @@ abstract class AffinityManager extends LogLevelUtil with Logging {
...
-    GlutenConfig.GLUTEN_SOFT_AFFINITY_MAX_DUPLICATE_READING_RECORDS_DEFAULT_VALUE
+  lazy val duplicateReadingMaxCacheItems =
+    GlutenConfig.GLUTEN_SOFT_AFFINITY_DUPLICATE_READING_MAX_CACHE_ITEMS_DEFAULT_VALUE
 
   // rdd id -> patition id, file path, start, length
-  val rddPartitionInfoMap = new ConcurrentHashMap[Int, Array[(Int, String, Long, Long)]]()
+  val rddPartitionInfoMap: LoadingCache[Integer, Array[(Int, String, Long, Long)]] =
+    CacheBuilder
+      .newBuilder()
+      .maximumSize(duplicateReadingMaxCacheItems)
```
jackylee-ch, line 65: We also need a `expireAfterAccess` to expire old values, as the old value may not be removed if the related events are discarded.

jackylee-ch:
This patch doesn't fix the performance problem for `SoftAffinityListener`, the Spark Events still can be discarded, which may cause user problem if they also use `Listeners`. We need a comment to explain this problem so that users can know the possible problems with this configuration when using it. BTW, maybe it's a better choice to change it to false by default, and users can turn it on as needed.

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
zhli1142015: I thinkt this setting is not suitable here, as it assume the task need to finish in the duration we set here.

zhli1142015:
> This patch doesn't fix the performance problem for `SoftAffinityListener`, the Spark Events still can be discarded, which may cause user problem if they also use `Listeners`. We need a comment to explain this problem so that users can know the possible problems with this configuration when using it. BTW, maybe it's a better choice to change it to false by default, and users can turn it on as needed.

Do you mean is the `SoftAffinityListener` cause of the slowness, I don't observe this, can you help tp share how this is repro in your env?

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
jackylee-ch: I get. However, I still recommend adding a configuration here so that users can adjust it as needed, because for long running jobs, such as Spark Connect, old values ​​may fill up the cache, and eventually the cache will not work.

jackylee-ch:
> Do you mean is the `SoftAffinityListener` cause of the slowness, I don't observe this, can you help tp share how this is repro in your env?

This is how I reproduced the problem locally.
1. Start `spark-sql` with `--conf spark.hadoop.parquet.page.size=1024 --conf spark.hadoop.parquet.block.size=2048 --conf spark.sql.files.maxPartitionBytes=2048`
2. Then, run below sqls.
```
create table test(a string) using parquet;
create table test1(a string) using parquet;
insert into test values(0); // make sure there is 10000 values in it
insert into test1 select /*+ REPARTITION(10000) */ * from test;
```
3. finally, restart `spark-sql` with `SoftAffinity` and run bellow sql
```
select count(*) from test1;
```

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
zhli1142015: I think the default eviction behavior is least recently used eviction, which match the expected behavior, we don't need additional settings to ensure it.
image

zhli1142015:
> > Do you mean is the `SoftAffinityListener` cause of the slowness, I don't observe this, can you help tp share how this is repro in your env?
> 
> This is how I reproduced the problem locally.
> 
> 1. Start `spark-sql` with `--conf spark.hadoop.parquet.page.size=1024 --conf spark.hadoop.parquet.block.size=2048 --conf spark.sql.files.maxPartitionBytes=2048`
> 2. Then, run below sqls.
> 
> ```
> create table test(a string) using parquet;
> create table test1(a string) using parquet;
> insert into test values(0); // make sure there is 10000 values in it
> insert into test1 select /*+ REPARTITION(10000) */ * from test;
> ```
> 
> 3. finally, restart `spark-sql` with `SoftAffinity` and run bellow sql
> 
> ```
> select count(*) from test1;
> ```

From above I actually can't repro the issue. I think maybe this is because of the hardware differences, but with more partitions, we actually observe the latency increasing. From the code, there are only the cache get / put operations in the event handling logic. I thought the memory pressure (GC) is more like the cause.
10K values
image
120k values
image

jackylee-ch:
> From above I actually can't repro the issue. I think maybe this is because of the hardware differences, but with more partitions, we actually observe the latency increasing.

Did you final get 10000+ FilePartitions for sql or only have one partition?

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
jackylee-ch: Great, thanks for explain it.

zhli1142015:
> > From above I actually can't repro the issue. I think maybe this is because of the hardware differences, but with more partitions, we actually observe the latency increasing.
> 
> Did you final get 10000+ FilePartitions for sql or only have one partition?

6k+ partitions for 10K values, and 60k+ partitions for 120k values.
image

image

jackylee-ch:
> > > From above I actually can't repro the issue. I think maybe this is because of the hardware differences, but with more partitions, we actually observe the latency increasing.
> > 
> > 
> > Did you final get 10000+ FilePartitions for sql or only have one partition?
> 
> 6k+ partitions for 10K values, and 60k+ partitions for 120k values. image
> 
> image

Greate, could you try 30 sqls in one Application to check if you can reproduce?

zhli1142015:
> > > > From above I actually can't repro the issue. I think maybe this is because of the hardware differences, but with more partitions, we actually observe the latency increasing.
> > > 
> > > 
> > > Did you final get 10000+ FilePartitions for sql or only have one partition?
> > 
> > 
> > 6k+ partitions for 10K values, and 60k+ partitions for 120k values. image
> > image
> 
> Greate, could you try 30 sqls in one Application to check if you can reproduce?

OOM happens with 120K values, but there is no event dropped logs.
image

zhli1142015:
OOM is not related, even SA is disable, it still happens.

jackylee-ch:
> OOM happens with 120K values, but there is no event dropped logs.

Hm, perhaps it is related to hardware performance. I tried `./bin/spark-sql --master local[192]` with the default the gluten configs and above sqls, the problem can be stably reproduced in our environment.
image

> BTW, maybe it's a better choice to change it to false by default, and users can turn it on as needed.

Since this pr has already closed the duplicate reading detection by default, It's ok to me now.

jackylee-ch (approved):
Thanks for your fix.

PR #7002, opened 2024-08-26
[GLUTEN-6887][VL] Daily Update Velox Version (2024_08_26)

GlutenPerfBot:
Upstream Velox's New Commits:

```txt
d33cdb251 by Jia Ke, Add RowsStreamingWindowBuild to avoid OOM in Window operator (9025)
3d10ccfc1 by Ke, Make processConfigs function pure virtual (10827)
8d45b3c78 by Deepak Majeti, Set CRC32 checksum algorithm for S3 multipart upload (10801)
52c1daaa1 by snadukudy, Add CI migration blog post (10532)
d46459c32 by Yang Zhang, Add write IO metrics and print physical written bytes in PlanNodeStatstoString (10806)
b5cd1e896 by Krishna Pai, Refactor RowTranslationUtil::toElementRows (10820)
eaec1d349 by Zuyu ZHANG, Fix build errors in Presto due to misconfigured generated proto file paths (10807)
2111fb59a by Jialiang Tan, Add disable pool management option to memory manager (10817)
b81dcaa9b by yingsu00, Allow column names to contain space (10784)
b228e0921 by Deepak Majeti, Set ARROW_DEPENDENCY_SOURCE=AUTO for arrow build (10819)
```

PR #7011, opened 2024-08-26
[VL] Add write IO metrics for WriteFiles

Yohahaha:


Yohahaha:
depends on #7002

gluten-data/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
```diff
@@ -182,6 +183,7 @@ object MetricsUtil extends Logging {
       ioWaitTime,
       preloadSplits,
       physicalWrittenBytes,
+      writeIOTime,
```
zhztheplayer, line 186: Perhaps you missed the merge code?

Yohahaha:
cc @PHILO-HE @liujiayi771 thank you!

Yohahaha:
@JkSelf

PHILO-HE (approved):

zhztheplayer:
@Yohahaha Can you confirm, there is a list for merging metric values https://github.com/Yohahaha/gluten/blob/f58163433b83453b772d76a158503bb1a77e28ad/gluten-data/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala#L129-L153 and whether we should append some code into that list.

cc @PHILO-HE

Yohahaha:
> @Yohahaha Can you confirm, there is a list for merging metric values https://github.com/Yohahaha/gluten/blob/f58163433b83453b772d76a158503bb1a77e28ad/gluten-data/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala#L129-L153 and whether we should append some code into that list.

I see, TableWriter count `physicalWrittenBytes`, `writeIONanos`, `numWrittenFiles`, but only `numWrittenFiles` is merged.

why we need merge those metrics?

zhztheplayer:
Sorry, I am not aside a PC now, so I can't actually reason about the code, which looks too magical to me by the way. 

But what's clear is only input and output metrics were being handled separately when the code was [firstly written](https://github.com/Yohahaha/gluten/commit/7996326ddf6e1c8441d3184a8e39036d501ea2ee) .

>  TableWriter count physicalWrittenBytes, writeIONanos, numWrittenFiles, but only numWrittenFiles is merged

After all, I think we should make these 4 metrics consistent. Either merge them all, or none of them. And seems to be safe to merge them all by taking one look. cc @JkSelf

PR #7025, opened 2024-08-27
[VL] Fix collectMetrics error when multi gluten rdd exist in one stage

kecookier:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -301,16 +301,22 @@ void WholeStageResultIterator::collectMetrics() {
     return;
   }
 
+  const auto& taskStats = task_->taskStats();
+  if (taskStats.executionStartTimeMs == 0) {
+    LOG(INFO) << "collectMetrics failed, taskStats is zero, maybe task never call next().";
+    return;
+  }
+
   if (veloxCfg_->get<bool>(kDebugModeEnabled, false) ||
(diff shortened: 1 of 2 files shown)


ulysses-you:
I think the issue is due to one side of join is an empty relation which cause Velox does not actually execute join operator. The pr description is confused.. we collect metrics when each task completion, so one task collect metrics would not affect another.

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -301,16 +301,22 @@ void WholeStageResultIterator::collectMetrics() {
     return;
   }
 
+  const auto& taskStats = task_->taskStats();
+  if (taskStats.executionStartTimeMs == 0) {
+    LOG(INFO) << "collectMetrics failed, taskStats is zero, maybe task never call next().";
```
ulysses-you, line 306: `Skip collect task metrics since task did not call next()`

ulysses-you (approved):
the fix lgtm, thank you @kecookier

kecookier:
> I think the issue is due to one side of join is an empty relation which cause Velox does not actually execute join operator. The pr description is confused.. we collect metrics when each task completion, so one task collect metrics would not affect another.

Yes, you are right.

cpp/velox/compute/WholeStageResultIterator.cc
kecookier: Fixed.

kecookier: @ulysses-you Can you review again?

ulysses-you (approved):

Issue #7035, opened 2024-08-27
[VL] Get wrong TARGET_BUILD_COMMIT when using internal velox repository


wForget:
### Problem description

As title.

build velox backend with internal velox repository: `./dev/package-vcpkg.sh`

got error:
```
+ git clone XXXX/velox.git -b branch-1.2.0 /data/git/gluten/ep/build-velox/src/../build/velox_ep
+ cd /data/git/gluten/ep/build-velox/src/../build/velox_ep
+ git checkout 88856e6b139c761e7876b1cd3b29e8dad236d8c7 88856e6b139c761e7876b1cd3b29e8dad236d8c7
error: pathspec '88856e6b139c761e7876b1cd3b29e8dad236d8c7' did not match any file(s) known to git
```

`git ls-remote` matches multiple commits:

![dfa33f4e-1e9d-4c23-a311-395206d7c551](https://github.com/user-attachments/assets/ad328f7c-1aad-49aa-a797-c63f2c393230)

### System information

Velox System Info v0.0.2
Commit: 05a1c0098a1161f4b43ac569fdafb4b9f5b698a3
CMake Version: 3.17.5
System: Linux-3.10.0-957.el7.x86_64
Arch: x86_64
CPU Name: Model name:            Intel Core Processor (Skylake, IBRS)
C++ Compiler: /opt/rh/devtoolset-9/root/usr/bin/c++
C++ Compiler Version: 9.3.1
C Compiler: /opt/rh/devtoolset-9/root/usr/bin/cc
C Compiler Version: 9.3.1
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

PR #7036, opened 2024-08-27
[GLUTEN-7035][VL] Use first line of `ls-remote`'s output as build's target commit

wForget:

Get head for target_build_commit

Fixes: #7035

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -151,7 +151,7 @@ fi
 VELOX_SOURCE_DIR="${VELOX_HOME}"
 
 # checkout code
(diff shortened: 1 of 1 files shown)


zhztheplayer:
I feel a better fix is to use `git ls-remote $VELOX_REPO refs/heads/$VELOX_BRANCH` which looks up a unique ref.

wForget:
> I feel a better fix is to use `git ls-remote $VELOX_REPO refs/heads/$VELOX_BRANCH` which looks up a unique ref.

Thanks, this looks like a better way.

image

zhztheplayer:
> But it doesn't work for a tag
> 
> image

Yes neither tag nor commit hash works.  Perhaps we have to keep the original solution.

zhztheplayer (approved):

Issue #7055, opened 2024-08-28
[VL][VCPKG] Support configure ssh keys for gluten vcpkg builder image


wForget:
### Description

When building gluten using `gluten-vcpkg-builder`, `git clone velox repo` is executed. It is necessary to configure ssh keys for `gluten-vcpkg-builder` image when using the internal fork repo. So I want to make `gluten-vcpkg-builder` image's ssh keys configurable.

PR #7056, opened 2024-08-28
[GLUTEN-7055][VL] Make ssh keys of gluten-vcpkg-builder configurable

wForget:

Make ssh keys of gluten-vcpkg-builder configurable

Fixes: #7055

## How was this patch tested?

After this, we can configure ssh keys for `gluten-vcpkg-builder`:

```
docker build --file docker/Dockerfile \
    --build-arg BUILDER_UID=`id -u` \
    --build-arg BUILDER_GID=`id -g` \
    --build-arg SSH_PRV_KEY="$(cat ~/.ssh/id_rsa)" \
    --build-arg SSH_PUB_KEY="$(cat ~/.ssh/id_rsa.pub)" \
    --tag "apache/gluten:gluten-vcpkg-builder_2024_08_28" \
    .
```

### DIFF
--- a/dev/vcpkg/docker/Dockerfile
+++ b/dev/vcpkg/docker/Dockerfile
@@ -19,6 +19,21 @@ RUN if ["$BUILDER_UID" != "0"]; then \
     fi
 USER ${BUILDER_UID}:${BUILDER_GID}
 
+ARG SSH_PRV_KEY
+ARG SSH_PUB_KEY
+RUN mkdir -p ~/.ssh && \
+    chmod 0700 ~/.ssh
+RUN if [[ -n "$SSH_PRV_KEY" ]] ; then \
+    echo "$SSH_PRV_KEY" > ~/.ssh/id_rsa && \
+    chmod 600 ~/.ssh/id_rsa && \
+    echo -e "Host *\n    StrictHostKeyChecking no" > ~/.ssh/config && \
+    chmod 600 ~/.ssh/config; \
+    fi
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@wForget, thanks for your effort!
As the purpose is for making `git clone` workable in docker, can we simply use repo's https url? Assume write permission is not needed.

wForget:
> As the purpose is for making git clone workable in docker, can we simply use repo's https url? Assume write permission is not needed.

When using internal gitlab repo https url, we also need to enter the username and password. In addition, since the ssh key is bound to the gitlab user in our internal repository, we can also grant read-only permissions to the repo. 
By the way, we use the `gluten-vcpkg-builder` image to integrate with Jenkins and use it on the Jenkins build machine to automatically build gluten.

I'm not sure if this is a common usage, if not, please ignore this change, thank you.

PHILO-HE:
> > As the purpose is for making git clone workable in docker, can we simply use repo's https url? Assume write permission is not needed.
> 
> When using internal gitlab repo https url, we also need to enter the username and password. In addition, since the ssh key is bound to the gitlab user in our internal repository, we can also grant read-only permissions to the repo. By the way, we use the `gluten-vcpkg-builder` image to integrate with Jenkins and use it on the Jenkins build machine to automatically build gluten.
> 
> I'm not sure if this is a common usage, if not, please ignore this change, thank you.

@wForget, thanks for your clarification! Sounds not a common usage to me, but not quite sure. Let's see whether there are any other feedback from the community. Thanks!

Issue #7078, opened 2024-09-02
[CORE] The fallback check for Scan should not be skipped when DPP is present


wang-zhun:
### Backend

VL (Velox)

### Exception information
ERROR
```
Context: Split [Hive: hdfs://xxx/xxx.parquet 4 - 133633304] Task Gluten_Stage_2_TID_2_VTID_2
Additional Context: Operator: TableScan[0] 0
Function: readInt
File: /incubator-gluten/ep/build-velox/build/velox_ep/./velox/dwio/common/IntDecoder.h
Line: 448
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorENS1_22CompileTimeEmptyStringEEEvRKNS1_18VeloxCheckFailArgsET0_
```

Plan
```
== Physical Plan ==
AdaptiveSparkPlan (44)
+- == Current Plan ==
   SortMergeJoin Inner (33)
   :- Sort (14)
   :  +- ShuffleQueryStage (13)
   :     +- ColumnarExchange (12)
   :        +- VeloxAppendBatches (11)
   :           +- ^ ProjectExecTransformer (9)
   :              +- ^ InputIteratorTransformer (8)
   :                 +- RowToVeloxColumnar (6)
   :                    +- * Project (5)
   :                       +- * Filter (4)
   :                          +- VeloxColumnarToRowExec (3)
   :                             +- ^ IcebergScanTransformer (1)
   +- Sort (32)
      +- ShuffleQueryStage (31)
         +- ColumnarExchange (30)
            +- VeloxAppendBatches (29)
               +- ^ ProjectExecTransformer (27)
                  +- ^ InputIteratorTransformer (26)
                     +- RowToVeloxColumnar (24)
                        +- * Project (23)
                           +- VeloxColumnarToRowExec (22)
                              +- ^ FilterExecTransformer (20)
                                 +- ^ InputIteratorTransformer (19)
                                    +- RowToVeloxColumnar (17)
                                       +- * ColumnarToRow (16)
                                          +- BatchScan (15)
+- == Initial Plan ==
   SortMergeJoin Inner (43)
   :- Sort (38)
[...]

PR #6035, opened 2024-06-11
[CORE] Rework Gluten + DPP compatibility

zhztheplayer:
Goals:
1. Fix hash relation type incompatible errors because of `ColumnarSubqueryBroadcastExec` or its resident filter/scan's fallback (by ras or by expanded-fallback[1]).
2. Simplify DPP code

Changes:
1. Remove DPP (`ColumnarSubqueryBroadcastExec`) handling code from ExpressionConverter
2. Remove all invocations on`ColumnarSubqueryBroadcastExec#transformDynamicPruningExpr`
3. Add rule `RewriteSubqueryBroadcast` which runs before regular columnar rules to replace `SubqueryBroadcastExec` with `ColumnarSubqueryBroadcastExec`

Tested by existing UTs (legacy plannner) and locally (RAS).

[1] Not confirmed. The issue was found when enabling RAS.

PR #7080, opened 2024-09-02
[GLUTEN-7078][CORE] The fallback check for Scan should not be skipped when DPP is present

wang-zhun:
(Fixes: \#7078)
performing fallback detection for Scan in advance is good and does not introduce any side effects.

## How was this patch tested?
manual tests


zzcclp:
it better to add some ut for this scenario.

wang-zhun:
> it better to add some ut for this scenario.

ok, thank you for the response.

zhztheplayer:
Do we have any updates on this PR?

zhztheplayer:
I think the if conditions are no longer needed after clean up https://github.com/apache/incubator-gluten/pull/6035? Let's merge to see if everything goes fine. cc @zzcclp

zhztheplayer (approved):

Issue #7096, opened 2024-09-03
[CH] throw exception when agg grouping has same attributes


shuai-xu:
### Backend

CH (ClickHouse)

### Bug description

when run sql `select
days, rtime, guest_uid,owner_uid, day1
from
(
 select
day1 as days, rtime, guest_uid, owner_uid, day1
from
(
 select distinct coalesce(day, "today") as day1,rtime,uid as guest_uid
 ,if(event.event_info['showeruid'] is null or event.event_info['showeruid']='',0,cast(event.event_info['showeruid'] as bigint) ) as owner_uid
 from xxx
 where day = '${day1}'
) t1 ) t2 group by days, rtime, guest_uid, owner_uid, day1 limit 10`, it throws Missmatch output schema exception

### Spark version

Spark-3.3.x

PR #7101, opened 2024-09-03
[GLUTEN-7096] [CH] fix exception when same names in group by

shuai-xu:

This pr fix two problems, one is that when group by has same names, the duplicated names will be remove by distinct in genHashAggregateExecTransformer, and cause Missmatch result columns size exception. The other is that when group by has same names but with different qualifier due to alias table name, the same names are passed to clickhouse, and cause Mismatch output schema exception.
The solution is that we distinct the groupings in Agg by ignoring qualifier, and insert a Project if has same names.

(Fixes: \#7096)

## How was this patch tested?

This patch was tested by unit tests.


zzcclp (approved):
LGTM

zzcclp:
please fix conflicts first

zzcclp (approved):

PR #7105, opened 2024-09-03
[VL] Remove a limit for BHJ in stage fallback policy

PHILO-HE:

Currently, Gluten allows that BroadcastExchange is offloaded, but BHJ falls back. So we can remove the limit.

## How was this patch tested?

An added test.


PHILO-HE:
@weiting-chen

weiting-chen (approved):
Looks good to me.

PR #7124, opened 2024-09-05
[VL] Fix function `input_file_name()` outputs empty string in certain query plan patterns

zml1206:

The Spark implementations of input_file_name/input_file_block_start/input_file_block_length uses a thread local to stash the file name and retrieve it from the function. If there is a transformer node between project input_file_function and scan, the result of input_file_name is an empty string. So we should push down input_file_function to transformer scan or add fallback project of input_file_function before fallback scan.

The processing logic is as follows:
1.Before offload, add new project before leaf node and push down input file expression to the new project
2.Normal offload project and scan
3.After offload, if scan be offloaded, push down input file expression into scan and remove project

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
@@ -52,7 +52,7 @@ private object VeloxRuleApi {
   def injectLegacy(injector: LegacyInjector): Unit = {
     // Gluten columnar: Transform rules.
     injector.injectTransform(_ => RemoveTransitions)
-    injector.injectTransform(_ => PushDownInputFileExpressionBeforeLeaf)
+    injector.injectTransform(_ => PushDownInputFileExpression.PreOffload)
     injector.injectTransform(c => FallbackOnANSIMode.apply(c.session))
     injector.injectTransform(c => FallbackMultiCodegens.apply(c.session))
     injector.injectTransform(c => PlanOneRowRelation.apply(c.session))
@@ -65,7 +65,7 @@ private object VeloxRuleApi {
     injector.injectTransform(_ => TransformPreOverrides())
     injector.injectTransform(_ => RemoveNativeWriteFilesSortAndProject())
     injector.injectTransform(c => RewriteTransformer.apply(c.session))
-    injector.injectTransform(_ => PushDownInputFileExpressionToScan)
+    injector.injectTransform(_ => PushDownInputFileExpression.PostOffload)
     injector.injectTransform(_ => EnsureLocalSortRequirements)
     injector.injectTransform(_ => EliminateLocalSort)
     injector.injectTransform(_ => CollapseProjectExecTransformer)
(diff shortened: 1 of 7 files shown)


zml1206:
cc @zhztheplayer Thank you.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/PushDownInputFileExpression.scala
```diff
@@ -0,0 +1,116 @@
...
+        if projectList.exists(PushDownInputFileExpression.containsInputFileRelatedExpr) =>
+      val replacedExprs = mutable.Map[String, Alias]()
+      val newProjectList = projectList.map {
+        expr => rewriteExpr(expr, replacedExprs).asInstanceOf[NamedExpression]
+      }
+      val newChild = addMetadataCol(child, replacedExprs)
+      ProjectExec(newProjectList, newChild)
+  }
+
+  private def rewriteExpr(expr: Expression, replacedExprs: Map[String, Alias]): Expression = {
```
zhztheplayer, line 62: Could use `replacedExprs: mutable.Map[String, Alias]`

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/PushDownInputFileExpression.scala
```diff
@@ -0,0 +1,116 @@
...
+      case _: InputFileBlockLength =>
+        replacedExprs
+          .getOrElseUpdate(expr.prettyName, Alias(InputFileBlockLength(), expr.prettyName)())
+          .toAttribute
+      case other =>
+        other.withNewChildren(other.children.map(child => rewriteExpr(child, replacedExprs)))
+    }
+  }
+
+  def addMetadataCol(plan: SparkPlan, replacedExprs: Map[String, Alias]): SparkPlan = plan match {
```
zhztheplayer, line 81: Could use replacedExprs: mutable.Map[String, Alias]

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
```diff
@@ -52,6 +52,7 @@ private object VeloxRuleApi {
   def injectLegacy(injector: LegacyInjector): Unit = {
     // Gluten columnar: Transform rules.
     injector.injectTransform(_ => RemoveTransitions)
+    injector.injectTransform(_ => PushDownInputFileExpressionBeforeLeaf)
```
zhztheplayer, line 55: Let's do a rename for the two rules. Since they seem to always be used together as a rule pair.

```suggestion
    injector.injectTransform(_ => PushDownInputFileExpression.PreOffload)
```

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
```diff
@@ -64,6 +65,7 @@ private object VeloxRuleApi {
     injector.injectTransform(_ => TransformPreOverrides())
     injector.injectTransform(_ => RemoveNativeWriteFilesSortAndProject())
     injector.injectTransform(c => RewriteTransformer.apply(c.session))
+    injector.injectTransform(_ => PushDownInputFileExpressionToScan)
```
zhztheplayer, line 68: Similarly,

```suggestion
    injector.injectTransform(_ => PushDownInputFileExpression.PostOffload)
```

zhztheplayer (commented):
Overall looking good to me. Thank you @zml1206.

cc @liuneng1994, if you are considering switching to this way, perhaps we need a PR for CH individually to add the rules to CH rule list after this PR gets landed.

zhztheplayer (approved):

PR #6942, opened 2024-08-20
[GLUTEN-6571][VL] Add platform and arch subdirectory for base lib package

wForget:
## What changes were proposed in this pull request?

Add platform and arch subdirectory for base lib package.

closes #6571

## How was this patch tested?

Manual package and test.

Check gluten bundle jar:

```
jar -tf gluten-velox-bundle-spark3.5_2.12-centos_7_x86_64-1.3.0-SNAPSHOT.jar | grep -E "*.so$"

# output
org/apache/gluten/linux/amd64/libgluten.so
org/apache/gluten/linux/amd64/libvelox.so
......
```

Successfully load native libs:

```
24/08/21 12:44:53 INFO JniLibLoader: Successfully loaded library org/apache/gluten/linux/amd64/libgluten.so
24/08/21 12:44:54 INFO JniLibLoader: Successfully loaded library org/apache/gluten/linux/amd64/libvelox.so
```

Successfully execute tpcds q3.

PR #7129, opened 2024-09-05
[VL] Minor follow-ups for #6942

zhztheplayer:
Following #6942,

1. Remove prefix `org/apache/gluten` for library pathes in Jar
2. ~~A test to assert on the runtime error message when Gluten build is used on unsupported platform.~~


zhztheplayer:
@wForget

wForget (approved):
Thanks, LGTM

wForget:
Related failures

```
VeloxListenerApiSuite:
- Unsupported arch *** FAILED ***
  Expected exception java.lang.RuntimeException to be thrown, but no exception was thrown (VeloxListenerApiSuite.scala:27)

- csv scan *** FAILED ***
  org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 730.0 failed 1 times, most recent failure: Lost task 0.0 in stage 730.0 (TID 540) (f98bd9d27030 executor driver): java.lang.IllegalStateException: error loading native libraries: java.io.FileNotFoundException: unknown-arch/libarrow_dataset_jni.so
	at org.apache.arrow.dataset.jni.JniLoader.load(JniLoader.java:95)
```

backends-velox/src/test/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApiSuite.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.apache.gluten.backendsapi.velox
+
+import org.apache.gluten.test.VeloxBackendTestBase
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class VeloxListenerApiSuite extends AnyFunSuite {
+  test("Unsupported arch") {
+    System.setProperty("os.arch", "unknown-arch")
```
wForget, line 25: This property needs to be reset

zhztheplayer: Actually there is no easy way to run this test correctly with other tests in same process as we only load the libraries once.

I'll remove the test. It's not that important to place it in code base. I already verified the error locally.

wForget (approved):

PHILO-HE (approved):
Thanks!

PR #7115, opened 2024-09-04
[GLUTEN-7004][CORE] Bump Spark version to 3.4.3

Yohahaha:
## What changes were proposed in this pull request?

Spark 3.4.3 is the latest release version of the 3.4 branch.

(Fixes: \#7004)

PR #7131, opened 2024-09-05
[VL] enable VCPKG build by default

zhouyuan:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


PHILO-HE:
@zhouyuan, just tested on my side. It works on ubuntu-20.04.

Maybe, we can keep one CI job using dynamic link by disabling vcpkg here.
https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_backend.yml#L1012

dev/builddeps-veloxbe.sh
```diff
@@ -173,6 +173,8 @@ if [ "$ENABLE_VCPKG" = "ON" ]; then
     # vcpkg will install static depends and init build environment
     envs="$("$GLUTEN_DIR/dev/vcpkg/init.sh")"
     eval "$envs"
+else
+    echo "Dynamic linked libs based building is deprecated, please enabled VCPKG based building."
```
Yohahaha, line 177: can we reserve dynamic link? I found static build is unstable in CI, it's better keep both way.

zhouyuan: The dynamic link will be kept but we need to print one message. The static packaging has been used starts from gluten 1.0 and seems working well
The issue on dynamic packaging is on the dependencies update case(especially for the minor version update on libraries) , we need to update the related loader as well

https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoaderUbuntu2204.scala

Yohahaha: https://github.com/apache/incubator-gluten/pull/7115#issuecomment-2330591059
https://github.com/apache/incubator-gluten/actions/runs/10769801803/job/29861861312

I found such errors, are these known issues?

weiting-chen: In fact, most of our customers have already use static link in their production ready environment.
One of the major reason is our 3rd party package script is no longer to be updated.
Only a few customers are still using dynamic link since they maintain their own 3rd party libraries instead to use the script in Gluten.

dev/buildbundle-veloxbe.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
 BASEDIR=$(dirname $0)
+
+if [ `uname -m` = "x86_64" ]; then
+  sudo -E ${BASEDIR}/setup-build-depends.sh
```
leoluan2009, line 6: This path may be `sudo -E ${BASEDIR}/vcpkg/setup-build-depends.sh` ?

dev/builddeps-veloxbe.sh
FelixYBW: > #7115 (comment)

Is the issue caused by static link?

FelixYBW:
@zhouyuan Let's close this since we can't get an agreement.

@Yohahaha currently one CI uses the dynamic link. You may run some jekins job internally to track the dynamic one.

PR #7133, opened 2024-09-05
[VL] New option to follow vanilla Spark's build side in shuffled hash join

zhztheplayer:
It's observed the new way to choose smaller side of shuffled hash join as build side, introduced in https://github.com/apache/incubator-gluten/pull/6093, could cause performance regression comparing to elder versions of Gluten because of inaccurate statistic calculations. 

The PR provides an option `spark.gluten.sql.columnar.shuffledHashJoin.optimizeBuildSide` which can be set to `false` to keep using vanilla Spark's build side in Gluten's shuffled hash join.


zhztheplayer:
cc @kecookier @weiting-chen

kecookier:
It looks good to me. Thanks!

kecookier (approved):

Issue #7079, opened 2024-09-02
[VL] Output metrics are counted multiple times for InputIterator of broadcasted exchange


ivoson:
### Backend

VL (Velox)

### Bug description

Below screenshot shows the metrics from tpcds q6. We can see that the output rows was counted multiple times(same as the task number of the join stage). From task execution's perspective, the task indeed read the data multiple times. While from a sql plan node's perspective, this is confusing since the output row counts of the plan node is not right.

I am wondering that does this make sense and shall we count only once for inputIterator of broadcasted exchange.

![image](https://github.com/user-attachments/assets/5f557af5-7439-486e-b54c-69966c061043)

### Spark version

None

zhztheplayer:
@ivoson Would you like to help check how vanilla Spark handles this case? Thanks.

ivoson:
> @ivoson Would you like to help check how vanilla Spark handles this case? Thanks.

Sure, checking on this.

ivoson:
cc @ulysses-you I am wondering do you have any thoughts about this? Thanks.

ulysses-you:
For now the output row number is `broadcasted rows * numPartitions`. I agree it may make user who know Spark a lot a bit confused. I'm fine to use the original broadcasted rows if there is a easy way to resolve it.

ivoson:
> For now the output row number is `broadcasted rows * numPartitions`. I agree it may make user who know Spark a lot a bit confused. I'm fine to use the original broadcasted rows if there is a easy way to resolve it.

I am thinking that maybe we can only collect the metrics from one of the tasks(task 0) for a wrapper of broadcast exchange which is easy to implement but a little bit tricky...
Do you think the proposal make sense? cc @zhztheplayer @ulysses-you 
If it sounds good to you, I'll create a PR for this.

FelixYBW:
Vanilla spark create the hashtable on driver and broadcast to worker. It's one optimization pending in Velox.

I agree your proposal.

Surbhi-Vijay:
Duplicate issue: [[VL] Incorrect Metrics in InputIteratorTransformer](https://github.com/apache/incubator-gluten/issues/4672)

FelixYBW:
> Duplicate issue: [[VL] Incorrect Metrics in InputIteratorTransformer](https://github.com/apache/incubator-gluten/issues/4672)

Can you verify PR #7167?

Surbhi-Vijay:
[...]

PR #7167, opened 2024-09-09
[GLUTEN-7079][VL] Fix metrics for InputIteratorTransformer of broadcast exchange

ivoson:

Fix issue https://github.com/apache/incubator-gluten/issues/7079 collecting metrics for `InputIteratorTransformer` of broadcast exchange.

Here is an example showing the difference before and after the change:
Execute below query on tpch test dataset, the table `lineitem` has 2 partitions:
```
select /*+ BROADCAST(part) */ * from part join lineitem on l_partkey = p_partkey
```
Before the change, the metrics showing in UI is:
image

After the change, the metrics showing in UI is:
image

## How was this patch tested?
UT added.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
@@ -22,7 +22,9 @@ import org.apache.gluten.substrait.{AggregationParams, JoinParams}
 
 import org.apache.spark.SparkContext
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.{ColumnarInputAdapter, SparkPlan}
+import org.apache.spark.sql.execution.adaptive.BroadcastQueryStageExec
+import org.apache.spark.sql.execution.exchange.BroadcastExchangeLike
 import org.apache.spark.sql.execution.metric.{SQLMetric, SQLMetrics}
 
 import java.lang.{Long => JLong}
@@ -48,8 +50,14 @@ class VeloxMetricsApi extends MetricsApi with Logging {
   }
 
   override def genInputIteratorTransformerMetricsUpdater(
+      child: SparkPlan,
       metrics: Map[String, SQLMetric]): MetricsUpdater = {
-    InputIteratorMetricsUpdater(metrics)
+    val forBroadcast = child match {
+      case ColumnarInputAdapter(c) if c.isInstanceOf[BroadcastQueryStageExec] => true
+      case ColumnarInputAdapter(c) if c.isInstanceOf[BroadcastExchangeLike] => true
+      case _ => false
+    }
+    InputIteratorMetricsUpdater(metrics, forBroadcast = forBroadcast)
   }
 
   override def genBatchScanTransformerMetrics(sparkContext: SparkContext): Map[String, SQLMetric] =
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
@@ -23,8 +23,7 @@ import org.apache.gluten.substrait.{AggregationParams, JoinParams}
 import org.apache.spark.SparkContext
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.execution.{ColumnarInputAdapter, SparkPlan}
-import org.apache.spark.sql.execution.adaptive.BroadcastQueryStageExec
-import org.apache.spark.sql.execution.exchange.BroadcastExchangeLike
+import org.apache.spark.sql.execution.adaptive.QueryStageExec
 import org.apache.spark.sql.execution.metric.{SQLMetric, SQLMetrics}
 
 import java.lang.{Long => JLong}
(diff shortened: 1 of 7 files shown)


ivoson:
cc @zhztheplayer @ulysses-you can you please help review this PR? Thanks.

gluten-data/src/main/scala/org/apache/gluten/metrics/InputIteratorMetricsUpdater.scala
```diff
@@ -15,22 +15,29 @@
...
+      if (!forBroadcast || TaskContext.getPartitionId() == 0) {
+        if (operatorMetrics.outputRows == 0 && operatorMetrics.outputVectors == 0) {
+          // Sometimes, velox does not update metrics for intermediate operator,
+          // here we try to use the input metrics
+          metrics("numOutputRows") += operatorMetrics.inputRows
+          metrics("outputVectors") += operatorMetrics.inputVectors
+        } else {
+          metrics("numOutputRows") += operatorMetrics.outputRows
+          metrics("outputVectors") += operatorMetrics.outputVectors
+        }
```
zhztheplayer, line 40: Thank you very much for working on this though I feel this solution is a little bit tricky. Could we just add a metric like `numPartitions` to `InputIteratorTransformer` which may help user understand the duplication of data?

ivoson: Hi @zhztheplayer any suggestions about how to proceed with this?

zhztheplayer: @ivoson Sorry for the late reply. I just went through the relevant code, one of my questions is do we actually need to insert a `InputIteratorTransformer` between BHJ and BE? As BE RDD is handled differently with regular RDDs.

Can we just remove `InputIteratorTransfomer` in this case?

ivoson: > Can we just remove InputIteratorTransfomer in this case?

Thanks @zhztheplayer , let me check on this.

zhztheplayer: Thank you cc @ulysses-you

ulysses-you: Remove InputIteratorTransfomer would break the assuming of operators when doing transform. It's not worthy to me to do this big change but only fix a minor issue.. Besides, even we remove InputIteratorTransfomer, I think the join itself would count input rows like InputIteratorTransfomer.

zhztheplayer: > Remove InputIteratorTransfomer would break the assuming of operators when doing transform

@ulysses-you I was thinking about only removing the ones between BHJs and BEs. Normal `InputIteratorTransfomer`s that converts Spark iterators to Velox inputs are good and should be kept. 

Once we included broadcast RDDs into normal Substrait transformation procedure (if we can manage to do that), we can add `InputIteratorTransfomer` back then restart the discussion about #7079 again. Before that, I think `InputIteratorTransfomer`s between BHJs and BEs are misleading on UT as they don't actually do Spark-to-Substrait transformation.

> Besides, even we remove InputIteratorTransfomer, I think the join itself would count input rows like InputIteratorTransfomer.

Thought it could be a bit more acceptable since it's literally a "broadcast" join on UI.

ulysses-you: For velox backend, the code path of BHJ is same with SHJ, so we would go to substrait transformation code path for the BHJ InputIteratorTransfomers. The only difference thing from SHJ is, BHJ using a fake RDD which just hold a iterator.

zhztheplayer: > Once we included broadcast RDDs into normal Substrait transformation procedure

It's my mistake here. You are right and the build side iterator is already included in Substrait plan.

zhztheplayer: What I was worried about is when something like `TaskContext.getPartitionId() == 0)` is coded, then there might be some corner cases that the broadcasted data are not identically consumed by different join partitions. Say if probe side of one partition is an empty input, backend could do some kind of optimization that prevents the build iterator from being read. Then wrong metrics will be displayed. @ivoson @ulysses-you If you guys could help confirm.

If that's possible, we may still need a better solution to make sure we don't introduce new hidden issues.

ulysses-you: Yeah, I agree this may be a potential issue. But it should not only about broadcast iterator. I think we should make metrics framework work with all empty iterator case, e.g., union an empty partition and do shuffle ? Not sure if it would cause some unexpected behavior.

zhztheplayer: > e.g., union an empty partition and do shuffle 

I assume this case is not that similar to the current one? Say we are reporting only the first partition of input iterator transformer on BE side by condition `if (TaskContext.getPartitionId() == 0)`. Thing is, partition 0 could report zero input rows while partition 1 may report non-zero rows in the corner case. cc @ivoson

ulysses-you: > TaskContext.getPartitionId() == 0

I see, you are reference this pr's code. I did not take deep look in this pr before... Can we pass the broadcasted row count to metrics directly ? I think we do not need to depend on the velox metrics to get the row count.

zhztheplayer: > Can we pass the broadcasted row count to metrics directly ?

Yes, this could be a feasible solution to me too.

ivoson: Thanks @zhztheplayer @ulysses-you for the feedback, you are right about the concerns. Checking more details about other potential solutions...

ivoson:
Hi @zhztheplayer, can you please take another look? I tried to make it share the same output row counts metrics from child plan and modified the previous output metrics as data being consumed which might sound more reasonable when the data got read multiple times.

ivoson:
gentle ping @zhztheplayer , can you please take another look? Thanks.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
```diff
@@ -38,13 +39,29 @@ class VeloxMetricsApi extends MetricsApi with Logging {
...
   override def genInputIteratorTransformerMetrics(
+      child: SparkPlan,
       sparkContext: SparkContext): Map[String, SQLMetric] = {
+    def metricsPlan(plan: SparkPlan): SparkPlan = {
+      plan match {
+        case ColumnarInputAdapter(child) => metricsPlan(child)
+        case q: QueryStageExec => metricsPlan(q.plan)
+        case _ => plan
+      }
+    }
```
zhztheplayer, line 50: Can we only collect child's metrics when it's a broadcast exchange? Thanks.

zhztheplayer (commented):
Thank you for the reminder @ivoson, and sorry for the late look.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
ivoson: Hi @zhztheplayer , can you please clarify about the intention a little bit? If we only collect child's metrics for broadcast exchange, there'll be no output metrics(`numOutputRows` and `outputVectors`) for other operators then.

zhztheplayer: I meant we can still keep the current code path in main stream code for the case of `child != broadcast exchange`. When `child == broadcast exchange`, we can grab the metrics from `child.metrics` to avoid duplicating the numbers. Does that sound feasible to you?

ivoson: > I meant we can still keep the current code path in main stream code for the case of `child != broadcast exchange`. When `child == broadcast exchange`, we can grab the metrics from `child.metrics` to avoid duplicating the numbers. Does that sound feasible to you?

Sure, I am fine with it. Will update.

zhztheplayer (approved):
LGTM. Thanks!

zhztheplayer:
@ivoson Do you want to show some screenshots on Spark UI about the difference made by this PR? Thanks.

ivoson:
> @ivoson Do you want to show some screenshots on Spark UI about the difference made by this PR? Thanks.

Sure, let me add some screenshots.

PR #7168, opened 2024-09-09
[VL][MINOR] Fix typo error when convert lag() to lead()

kecookier:

If the offset is negative, convert lag() to lead().

## How was this patch tested?

Exist UT.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/expression/WindowFunctionsBuilder.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/expression/WindowFunctionsBuilder.scala
@@ -31,7 +31,7 @@ object WindowFunctionsBuilder {
     val substraitFunc = windowFunc match {
       // Handle lag with negative inputOffset, e.g., converts lag(c1, -1) to lead(c1, 1).
       // Spark uses `-inputOffset` as `offset` for Lag function.
-      case lag: Lag if lag.offset.eval(EmptyRow).asInstanceOf[Int] > 0 =>
+      case lag: Lag if lag.offset.eval(EmptyRow).asInstanceOf[Int] < 0 =>
         Some(LEAD)
       // Handle lead with negative offset, e.g., converts lead(c1, -1) to lag(c1, 1).
       case lead: Lead if lead.offset.eval(EmptyRow).asInstanceOf[Int] < 0 =>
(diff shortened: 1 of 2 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/expression/WindowFunctionsBuilder.scala
```diff
@@ -31,7 +31,7 @@ object WindowFunctionsBuilder {
     val substraitFunc = windowFunc match {
       // Handle lag with negative inputOffset, e.g., converts lag(c1, -1) to lead(c1, 1).
       // Spark uses `-inputOffset` as `offset` for Lag function.
-      case lag: Lag if lag.offset.eval(EmptyRow).asInstanceOf[Int] > 0 =>
```
PHILO-HE, line 34: @kecookier, seems offset > 0 implies inputOffset < 0, so the original handling is expected, right?

kecookier: @PHILO-HE  This PR addresses the lag result mismatch issue in my current version of Spark. Thank you for your input. I have reviewed the Spark code, and now I understand the differences.

In Spark 3.0, the LAG function calculates the bound using both the offset and the direction. However, in versions post Spark 3.1, the function does not consider the direction. Instead, it uses a single literal expression that includes the offset of the current row to calculate the bound. For example, in lag(), the offset will be wrapped with UnaryMinus(offset).

lag in spark 3.0 https://github.com/apache/spark/blob/branch-3.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L474

lag in spark 3.2 https://github.com/apache/spark/blob/branch-3.2/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/windowExpressions.scala#L538

PHILO-HE: @kecookier, I see. Thanks for your clarification! Seems there is no way to let gluten adapt to the two implementations. Should you fix it in a forked gluten on your side?

kecookier: Ok, I will close this PR.

kecookier:
I will make it compatible with Spark 3.0 in the internal repo.

Issue #7087, opened 2024-09-03
[CH] Support `WindowGroupLimit` for `row_number`, `rank` and `dense_rank`


lgbo-ustc:
### Description

Part of #6067.

For ds q44, the phyiscal plan is
```
CHNativeColumnarToRow
+- TakeOrderedAndProjectExecTransformer (limit=100, orderBy=[rnk#74 ASC NULLS FIRST], output=[rnk#74,best_performing#80,worst_performing#81])
   +- ^(16) ProjectExecTransformer [rnk#74, i_product_name#135 AS best_performing#80, i_product_name#180 AS worst_performing#81]
      +- ^(16) CHSortMergeJoinExecTransformer [item_sk#75L], [i_item_sk#159L], Inner, false
         :- ^(16) SortExecTransformer [item_sk#75L ASC NULLS FIRST], false, 0
         :  +- ^(16) InputIteratorTransformer[rnk#74, item_sk#75L, i_product_name#135]
         :     +- ColumnarExchange hashpartitioning(item_sk#75L, 5), ENSURE_REQUIREMENTS, [plan_id=1465], [shuffle_writer_type=hash], [OUTPUT] List(rnk:IntegerType, item_sk:LongType, i_product_name:StringType)
         :        +- ^(14) ProjectExecTransformer [rnk#74, item_sk#75L, i_product_name#135]
         :           +- ^(14) CHSortMergeJoinExecTransformer [item_sk#70L], [i_item_sk#114L], Inner, false
         :              :- ^(14) SortExecTransformer [item_sk#70L ASC NULLS FIRST], false, 0
         :              :  +- ^(14) InputIteratorTransformer[item_sk#70L, rnk#74, item_sk#75L]
         :              :     +- ColumnarExchange hashpartitioning(item_sk#70L, 5), ENSURE_REQUIREMENTS, [plan_id=1407], [shuffle_writer_type=hash], [OUTPUT] List(item_sk:LongType, rnk:IntegerType, item_sk:LongType)
         :              :        +- ^(12) ProjectExecTransformer [item_sk#70L, rnk#74, item_sk#75L]
         :              :           +- ^(12) CHSortMergeJoinExecTransformer [rnk#74], [rnk#79], Inner, false
         :              :              :- ^(12) SortExecTransformer [rnk#74 ASC NULLS FIRST], false, 0
         :              :              :  +- ^(12) ProjectExecTransformer [item_sk#70L, rnk#74]
[...]

PR #7176, opened 2024-09-10
[GLUTEN-7087][CH] Support `WindowGroupLimitExec`

lgbo-ustc:

Fixes: #7087 

`WindowGroupLimitExec` is introduced in spark-3.5. This PR makes it supported

## How was this patch tested?

unit tests


cpp-ch/local-engine/local_engine_jni.cpp
```diff
@@ -166,6 +167,8 @@ JNIEXPORT jint JNI_OnLoad(JavaVM * vm, void * /*reserved*/)
     local_engine::BroadCastJoinBuilder::init(env);
     local_engine::CacheManager::initJNI(env);
 
+    DB::registerFormats();
```
baibaichen, line 170: Why does this PR need to call `DB:: tradeFormats()`?

cpp-ch/local-engine/Common/CHUtil.cpp
```diff
@@ -315,6 +316,16 @@ DB::Block BlockUtil::concatenateBlocksMemoryEfficiently(std::vector<DB::Block> &
     return out;
 }
 
+String BlockUtil::dumpBlock(const DB::Block & block)
```
baibaichen, line 319: How are about moving to `BlockTypeUtils.h/.cpp` ?

cpp-ch/local-engine/local_engine_jni.cpp
lgbo-ustc: dump block needs formats.

lgbo-ustc: adding a simple method to dump block contents is very useful for debug

cpp-ch/local-engine/Common/CHUtil.cpp
lgbo-ustc: `BlockTypeUtils.h/.cpp` is just about types?

baibaichen: Actually, it's block and type utils :)

cpp-ch/local-engine/local_engine_jni.cpp
baibaichen: You mean calling `DB:: registerFormats()` is for later dumping?

cpp-ch/local-engine/Common/CHUtil.cpp
baibaichen: If dumping is just for debug aid, I think we should move it to `DebugUtils.h/cpp`, and put it to `debug` namespace. In product code, we should put it into` #ifdef NDEBUG ..  #endif`. e.g.

``` C++
#ifdef NDEBUG
 debug::dumpxxx();
#endif
```

baibaichen (approved):

PR #7186, opened 2024-09-10
[VL] Make conf option `s.g.s.c.shuffledHashJoin.optimizeBuildSide` work correctly with option `s.g.s.c.forceShuffledHashJoin`

zhztheplayer:
Currently when `s.g.s.c.shuffledHashJoin.optimizeBuildSide==false` and `s.g.s.c.forceShuffledHashJoin=true`, SMJ will be converted to SHJ with a randomly chosen build side. This PR fixes the issue by choosing a smaller one in `RewriteJoin` rule.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/FallbackRules.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/FallbackRules.scala
@@ -104,9 +104,9 @@ object FallbackTags {
 
   /**
    * If true, it implies the plan maybe transformable during validation phase but not guaranteed,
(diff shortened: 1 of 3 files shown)


rui-mo (approved):
Thanks. Perhaps remove the prefix 's.g.s.c.' from the title, as it might bring confusion.

zhztheplayer:
> Thanks. Perhaps remove the prefix 's.g.s.c.' from the title, as it might bring confusion.

Thanks although I found it's an easy way to represent a config option in title. Or if you have better suggestions? Like `shuffledHashJoin.optimizeBuildSide` / `forceShuffledHashJoin`?

PR #7188, opened 2024-09-10
[VL] Collapse project if scan is fallback and the outer project is cheap or fallback

zml1206:
follow #7124
In `PushDownInputFileExpression`, we will add new project before leaf node and push down input file expression to the
new project, if scan is fallback and the outer project is cheap or fallback, we can collapse project to simplify physical plan.

## How was this patch tested?

UT.


zml1206:
cc @zhztheplayer Thank you.

zhztheplayer:
Hi @zml1206, why this is done in rule `PushDownInputFileExpression`? Given that we already had individual rules like `CollapseProjectExecTransformer` to collapse projects. Should we have an individual rule for this feature as well?

zml1206:
> Hi @zml1206, why this is done in rule `PushDownInputFileExpression`? 

Because this `Project` is produced by `PushDownInputFileExpression`.

zhztheplayer (approved):

baibaichen:
@zml1206  @zhztheplayer 

IIUC, the following codes colud be removed, after this pr, right?

```
GlutenWholeStageColumnarRDD::compute()
    // To support input_file_name(). According to semantic we should return
    // the exact file name a row belongs to. However in columnar engine it's
    // not easy to accomplish this. so we return a list of file(part) names
    split match {
      case FirstZippedPartitionsPartition(_, g: GlutenPartition, _) =>
        InputFileBlockHolderProxy.set(g.files.mkString(","))
      case _ =>
        InputFileBlockHolderProxy.unset()
    }
```

PR #7189, opened 2024-09-10
[MINOR][BUILD] Extract gcc version from libgluten.so

wecharyu:

Read the gcc version from generated `libgluten.so` file instead of use `gcc --version` directly.

We may compile cpp codes and scala codes in different bash, if multiple versions of GCC are installed on the system and the GCC version used by C++ is not correctly sourced during Scala code compilation, the generated GCC version by `gcc --version` may be inaccurate in `gluten-build-info.properties`.

(Fixes: \#ISSUE-ID)

## How was this patch tested?
Manually test on local by comparing the generated `gluten-build-info.properties` files.

### DIFF
--- a/dev/gluten-build-info.sh
+++ b/dev/gluten-build-info.sh
@@ -39,7 +39,7 @@ echo_build_properties() {
   echo url=$(git config --get remote.origin.url)
 
   if [ "$BACKEND_TYPE" = "velox" ]; then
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@wecharyu, please rebase the code to fix CI failure. BTW, seems this is not workable for macOS. @xumingming, do you have any comment?

wecharyu:
@PHILO-HE how about using `strings` command? It should be available in both linux and macOS.

PHILO-HE:
> @PHILO-HE how about using `strings` command? It should be available in both linux and macOS.

@wecharyu, I guess it should be more appropriate.

PHILO-HE (approved):
Looks good! cc @xumingming

Issue #7208, opened 2024-09-12
Using the static glog library will cause the loading of libvelox.so to fail


JinHelin404:
### Backend

VL (Velox)

### Bug description

Using the local static glog library to compile velox and gluten will result in the following error:

> INFO JniLibLoader: Trying to load library libvelox.so
> ERROR: something wrong with flag 'timestamp_in_logfile_name' in file '/gluten/ep/build-velox/velox-cicd/thirdparty/src/glog-0.6.0/src/logging.cc'.  One possibility: file '/gluten/ep/build-velox/velox-cicd/thirdparty/src/glog-0.6.0/src/logging.cc' is being linked both statically and dynamically into this executable.

### Spark version

Spark-3.2.x

zhztheplayer:
As this doesn't likely happen on all platforms. Can you attach your system information? Thanks.

JinHelin404:
> Why this doesn't happen in CI? Any preconditions to repeat the issue?

@zhztheplayer This will happen when we compile velox and gluten with static glog lib and without ENABLE_GLUTEN_VCPKG=ON:

https://github.com/apache/incubator-gluten/blob/main/cpp/core/CMakeLists.txt#L216.

PR #7209, opened 2024-09-12
[GLUTEN-7208][VL]fix: loading libvelox.so failed when using static glog

JinHelin404:

(Fixes: \#7208)

### DIFF
--- a/cpp/core/CMakeLists.txt
+++ b/cpp/core/CMakeLists.txt
@@ -213,9 +213,9 @@ file(MAKE_DIRECTORY ${root_directory}/releases)
 add_library(gluten SHARED ${SPARK_COLUMNAR_PLUGIN_SRCS})
 add_dependencies(gluten jni_proto)
 
-# Hide symbols of some static dependencies. Otherwise, if such dependencies
-# are already statically linked to libvelox.so, a runtime error will be
-# reported: xxx is being linked both statically and dynamically.
+# Hide symbols of some static dependencies. Otherwise, if such dependencies are
+# already statically linked to libvelox.so, a runtime error will be reported:
+# xxx is being linked both statically and dynamically.
 target_link_options(
   gluten PRIVATE -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/symbols.map)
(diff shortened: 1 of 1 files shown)


JinHelin404:
@PHILO-HE Hi. Could you please help to review this? [4434]( https://github.com/apache/incubator-gluten/pull/4434) Introduced gluten link glog, which will cause [7208](https://github.com/apache/incubator-gluten/issues/7208).

PHILO-HE:
@JinHelin404, thanks for your pr! According to your change, linking glog is only required when `ENABLE_GLUTEN_VCPKG` is ON. Could you clarify a bit?

PHILO-HE:
> @PHILO-HE thanks for your reviewing! Actually I also want to figure out why glog is linked in your pr [4434](https://github.com/apache/incubator-gluten/pull/4434) . This pr fix issue when vcpkg is on so I put linking glog in the `if(ENABLE_GLUTEN_VCPKG) `conditional statement.

@JinHelin404, I remember linking glog is necessary when debug build is enabled (at least it's true at that code point).

The issue of both statically and dynamically linking can happen when static glog lib is used, regardless of vcpkg is enabled or not (enabling vcpkg definitely makes static glog lib be used). I think we can remove the below check to also hide those conflict symbols when vcpkg is not enabled:
https://github.com/apache/incubator-gluten/blob/main/cpp/core/CMakeLists.txt#L216.

JinHelin404:
> The issue of both statically and dynamically linking can happen when static glog lib is used, regardless of vcpkg is enabled or not (enabling vcpkg definitely makes static glog lib be used). I think we can remove the below check to also hide those conflict symbols when vcpkg is not enabled:

@PHILO-HE Thanks for your advice. I agree that we can always hide these symbols and I have fixed this pr.

PHILO-HE:
@JinHelin404, please fix cmake format issue (you can just apply the patch generated in that failed CI log). BTW, please verify  this patch in your local to ensure it can really fix your encountered issue .

PHILO-HE (approved):
Thanks for your fix!

Issue #7213, opened 2024-09-12
[CORE] Avoid fallback caused by `CheckOverflowInTableInsert`


wForget:
### Description

Spark introduces the `CheckOverflowInTableInsert` expression to show proper error messages on the overflow errors of table insert, see: https://issues.apache.org/jira/browse/SPARK-39865.

In gluten, we can directly transform the child expr of `CheckOverflowInTableInsert` to avoid fallback.

Reproduce:

```
create table t1 (a float) stored as parquet;
create table t2 (b decimal(10,4)) stored as parquet;
insert into t1 values(1.1);
insert overwrite t2 select * from t1;
```

gluten fallback detail:

```
== Fallback Summary ==
(4) Project: Not supported to map spark function name to substrait function name: cast(a#2000 as decimal(10,4)), class name: CheckOverflowInTableInsert.

== Physical Plan ==
Execute InsertIntoHadoopFsRelationCommand (14)
+- VeloxColumnarWriteFiles (13)
   :- ^ WriteFilesExecTransformer (8)
   :  +- ^ InputIteratorTransformer (7)
   :     +- RowToVeloxColumnar (5)
   :        +- Project (4)
   :           +- VeloxColumnarToRowExec (3)
   :              +- ^ Scan parquet spark_catalog.pilot_temp.t1 (1)
   +- RowToVeloxColumnar (12)
      +- WriteFiles (11)
         +- NoopLeaf (10)
```

wForget:
Gluten doesn't have good support for `ANSI` mode, so this fallback is expected. I will remove this first.

wForget:
We may be able to make this fallback reason clearer

PR #7214, opened 2024-09-12
[GLUTEN-7213][CORE] Avoid fallback caused by CheckOverflowInTableInsert

wForget:

Using `ChildTransformer` to transform `CheckOverflowInTableInsert` to avoid fallback.

closes: #7213

## How was this patch tested?

added unit test


wForget:
Close with https://github.com/apache/incubator-gluten/issues/7213#issuecomment-2350869006

Issue #7223, opened 2024-09-13
[VL] row2VeloxColumnar bad performance


FelixYBW:
### Description

The query has hive table scan which output complex datatype. Gluten elapsed time is >2x worse than Vanilla Spark. 93% percent of time is spending on RowToVeloxColumnar

FelixYBW:
postpone the r2c until complex type is not presented

zml1206:
It seems that not only complex types, too many rows will also be slow.

PR #7229, opened 2024-09-13
[VL] RAS: Avoid adding R2C whose schema contains complex data types

zhztheplayer:
It's observed Velox R2C becomes expensive when containing complex data types as input. Add a coster in rough cost model to address the issue.

To enable rough cost model:

```
spark.gluten.ras.enabled=true
spark.gluten.ras.costModel=rough
```

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
@@ -36,7 +36,7 @@ class RoughCostModel extends LongCostModel {
         // Make trivial ProjectExec has the same cost as ProjectExecTransform to reduce unnecessary
         // c2r and r2c.
         10L
-      case r2c: RowToColumnarExecBase if hasComplexType(r2c.schema) =>
+      case r2c: RowToColumnarExecBase if hasComplexTypes(r2c.schema) =>
         // Avoid moving computation back to native when transition has complex types in schema.
         // such transitions are observed to be extreme expensive as of now.
         Long.MaxValue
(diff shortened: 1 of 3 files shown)


zhztheplayer:
cc @FelixYBW @jinchengchenghh

gluten-substrait/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
```diff
@@ -34,6 +36,10 @@ class RoughCostModel extends LongCostModel {
         // Make trivial ProjectExec has the same cost as ProjectExecTransform to reduce unnecessary
         // c2r and r2c.
         10L
+      case r2c: RowToColumnarExecBase if hasComplexTypes(r2c.schema) =>
+        // Avoid moving computation back to native when transition has complex types in schema.
+        // Such transitions are observed to be extreme expensive as of now.
+        Long.MaxValue
```
jinchengchenghh, line 42: Can we add a more reasonable value to the cost? How can we measure the operator cost now?

jinchengchenghh: We can use it as a temp solution. The r2c is very expensive.

jinchengchenghh (dismissed):

gluten-substrait/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
zhztheplayer: Yes. It's just a heuristic as a temp solution.

FelixYBW:
temp fix of #7223

JkSelf (approved):

FelixYBW:
Looks it doesn't work. Let me check what happens

Issue #7028, opened 2024-08-27
[CH] Fully Support writing parquet and mergetree in spark 3.5.x with delta protocol


baibaichen:
### Description

This is umbrella issue.  

Previously,  https://github.com/apache/incubator-gluten/issues/6705 is just a POC to prove that we can implemtent Delta Write based on `CumnarWriteFilesExec`.

1. https://github.com/apache/incubator-gluten/pull/7029
2. https://github.com/apache/incubator-gluten/pull/7279
3. https://github.com/apache/incubator-gluten/pull/7234
4. https://github.com/apache/incubator-gluten/pull/7170
5. https://github.com/apache/incubator-gluten/pull/7395
6. https://github.com/apache/incubator-gluten/pull/7506
7. https://github.com/apache/incubator-gluten/pull/7788
8. https://github.com/apache/incubator-gluten/pull/7924
9. https://github.com/apache/incubator-gluten/pull/7993
10. https://github.com/apache/incubator-gluten/pull/8029
11. https://github.com/apache/incubator-gluten/pull/8052
12. https://github.com/apache/incubator-gluten/pull/8237
13. https://github.com/apache/incubator-gluten/pull/8158
14. https://github.com/apache/incubator-gluten/pull/8346
15. https://github.com/apache/incubator-gluten/pull/8364

## backlog
   - [x] Collect file bytes that write pipeline produces.
   - [ ] https://github.com/apache/incubator-gluten/issues/8053
   - [ ] Support `InsertIntoHiveTable`
   - [ ] Support Bucket table? or drop support
       - [x] ~`HDFS, HDFS with rockdb, S3` :: **test mergetree write with the path based bucket table**~ by Design => (#8364)
   - [ ] Support escape partition value. e.g. incluing space #8158
   - [x]  we **need redesign** how to support case (in)sensitive (#8346).
   - [x] Bug Fix (#8364)
      - [x] test mergetree with partition with whitespace
      - [x] `GlutenClickHouseMergeTreeCacheDataSuite::test cache mergetree data no partition columns`
      - [x] `GlutenClickHouseMergeTreePathBasedWriteSuite::test mergetree path based table update` and `GlutenClickHouseMergeTreePathBasedWriteSuite::test mergetree path based table delete`

PR #7234, opened 2024-09-13
[GLUTEN-7028][CH][Part-3] Refactor: Move mergetree related codes to backends-clickhouse

baibaichen:
This is second refacor PR for moving mergetree related codes to backends-clickhouse:
1. Move JniUtils from gluten-arrow to gluten-core and using config.proto, so we can use ConfigMap to pass configuration between java and c++.
2. Move ExtensionTableXXX from gluten-substrait to backends-clickhouse
3. Move `genWriteParameters` to `TransformerApi`, so we can pass mergetree related parameters.

(Fixes: \#7028)

## How was this patch tested?

Using Existed UTs


gluten-core/src/main/java/org/apache/gluten/jni/JniUtils.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.gluten.init;
+package org.apache.gluten.jni;
```
zhztheplayer, line 17: If we move it to `gluten-core`, could we do some minor refactors on this class at the same time? My suggestion would be:

1. `org.apache.gluten.jni.JniUtils` -> `org.apache.gluten.utils.ConfigUtil`
1. `JniUtils.toNativeConf` -> `ConfigUtil.serialize`

Thanks.

zhztheplayer (approved):

PR #7248, opened 2024-09-14
[GLUTEN-7213][CORE] Make fallback reason for CheckOverflowInTableInsert clearer

wForget:

Make fallback reason for CheckOverflowInTableInsert clearer

## How was this patch tested?

added unit test

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
@@ -482,7 +482,7 @@ object ExpressionConverter extends SQLConfHelper with Logging {
           replaceWithExpressionTransformer0(c.child, attributeSeq, expressionsMap),
           c)
       case c if c.getClass.getSimpleName.equals("CheckOverflowInTableInsert") =>
-        throw new UnsupportedOperationException(
+        throw new GlutenNotSupportException(
           "CheckOverflowInTableInsert is used in ansi mode, but gluten does not support ANSI mode."
         )
       case b: BinaryArithmetic if DecimalArithmeticUtil.isDecimalArithmetic(b) =>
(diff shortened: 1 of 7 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -481,6 +481,10 @@ object ExpressionConverter extends SQLConfHelper with Logging {
           substraitExprName,
           replaceWithExpressionTransformer0(c.child, attributeSeq, expressionsMap),
           c)
+      case c if c.getClass.getSimpleName.equals("CheckOverflowInTableInsert") =>
+        throw new UnsupportedOperationException(
```
PHILO-HE, line 485: -> GlutenNotSupportException

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/sources/GlutenInsertSuite.scala
```diff
@@ -623,6 +624,24 @@ class GlutenInsertSuite
       }
     }
   }
+
+  testGluten("GLUTEN-7213: Check fallback reason with CheckOverflowInTableInsert") {
```
PHILO-HE, line 628: Can we put the test into a common place? E.g.,
https://github.com/apache/incubator-gluten/tree/a7c9567bc7acdd14c8474c3d2dc99d1cfd062b56/gluten-ut/test/src/test/scala/org/apache/gluten/expressions

You can use `testWithSpecifiedSparkVersion` to let it be tested with spark-3.5.

wForget: thanks, changed

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
wForget: fixed

PHILO-HE (approved):
Thanks!

PR #7274, opened 2024-09-19
[VL] Fix columns added to `outNames` twice when building Substrait plan

zml1206:


gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -223,9 +223,6 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
         outputSchema,
         null)
     } else {
-      for (attr <- childCtx.outputAttributes) {
-        outNames.add(ConverterUtils.genColumnNameWithExprId(attr))
-      }
```
zhztheplayer, line 228: What's the impact of the issue so far?

zhztheplayer (approved):

PR #7286, opened 2024-09-20
[VL] Gluten-it: In auto cluster mode, add option `--off-heap-ratio` for adjusting memory shares of off-heap and on-heap

zhztheplayer:
In gluten-it's auto cluster mode, the ratio between on-heap memory and off-heap memory is 1:2. The patch makes this ratio configurable.


PHILO-HE (approved):
Looks nice!

PR #7304, opened 2024-09-21
[GLUTEN-7178][VL] Fix field not found error when struct field name contains upper case

zml1206:

Fixes: \#7178

## How was this patch tested?
UT

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHBackend.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHBackend.scala
@@ -326,6 +326,8 @@ object CHBackendSettings extends BackendSettingsApi with Logging {
 
   override def supportStructType(): Boolean = true
 
+  override def structFieldToLowerCase(): Boolean = false
+
   override def supportExpandExec(): Boolean = true
 
   override def excludeScanExecFromCollapsedStage(): Boolean =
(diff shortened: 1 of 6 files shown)


backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHBackend.scala
```diff
@@ -326,6 +326,8 @@ object CHBackendSettings extends BackendSettingsApi with Logging {
 
   override def supportStructType(): Boolean = true
 
+  override def structFieldToLowerCase(): Boolean = false
```
PHILO-HE, line 329: @taiyang-li, curious whether this issue exists in ch backend. If the fix is applicable to both backends, we can remove this separated handling.

zml1206: ch backend seems to be different. I didn’t add `structFieldToLowerCase` at first, but ck ci reported an error.

taiyang-li: @PHILO-HE  it seems that CH haven't met this kind of issues. Let's keep it to be false for CH backend.

PHILO-HE: @taiyang-li, thanks for your verification!

PHILO-HE (approved):

PR #7052, opened 2024-08-28
[VL] Customize VCPKG build features according to user's build options

PHILO-HE:
## What changes were proposed in this pull request?

When vcpkg is enabled, installing native dependencies related to test, hdfs, s3, gcs and abfs should be optional, which depends on user-specified build options. We can put always-required dependencies as vcpkg default feature and enable other feature according to build option. 

## How was this patch tested?
CI.

PR #7331, opened 2024-09-24
[VL] Fix vcpkg binary caching in docker image

PHILO-HE:

Since #7052, we need to enable all vcpkg features to generate binary cache in docker image, which can save CI build time.

## How was this patch tested?

Local test.


PHILO-HE:
@zhztheplayer, this is another follow-up fix. Thanks!

zhztheplayer (approved):

Issue #7323, opened 2024-09-24
[VL] round result it not correct


liuyongvs:
### Backend

VL (Velox)

### Bug description

spark-sql (default)> select round(44, -1);
40

while velox and gluten both return 44

### Spark version

None

surnaik:
@liuyongvs Could you please provide the configs used, in my case it falls back to Spark

24/09/24 14:13:04 WARN ProjectExecTransformer: Validation failed with exception for plan: ProjectExecTransformer, due to: Not supported to map spark function name to substrait function name: toprettystring(round(44, -1), Some(America/Los_Angeles)), class name: ToPrettyString.
24/09/24 14:13:04 WARN GlutenFallbackReporter: Validation failed for plan: Project[QueryId=10], due to: Not supported to map spark function name to substrait function name: toprettystring(round(44, -1), Some(America/Los_Angeles)), class name: ToPrettyString..

surnaik:
I investigated this issue, Spark casts Integral type to Decimal when scale < 0, hence it's output is 40, whereas we treat it as an integral type and return the same number - 44

surnaik:
In Spark, we should always round negative decimals, which is different from Presto semantics. 

https://github.com/prestodb/presto/blob/7d9a0b0ae10a9e78b38d458cb0fafa04088b70d1/presto-main/src/main/java/com/facebook/presto/operator/scalar/MathFunctions.java#L1065

PR #7337, opened 2024-09-24
[GLUTEN-7323][VL] Always round negative decimals for integral types

surnaik:

Spark Round Semantics are different from Presto Round Semantics.
Presto Round Semantics doesn't actually round negative decimals in case of integral types - https://github.com/prestodb/presto/blob/7d9a0b0ae10a9e78b38d458cb0fafa04088b70d1/presto-main/src/main/java/com/facebook/presto/operator/scalar/MathFunctions.java#L1065

Whereas Spark always rounds up Integral Types when Scale < 0 - https://github.com/apache/spark/blob/55d0233d19cc52bee91a9619057d9b6f33165a0a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/mathExpressions.scala#L1546

This PR introduces a change in the RoundFunction to always round negative decimals.

Fixes: #7323 

## How was this patch tested?

Locally Tested on a Spark 3.5.1 cluster as well as CI


zhouyuan:
CC @rui-mo

Issue #7264, opened 2024-09-18
[VL] Rename and refactor `gluten-data` into module `gluten-arrow`


zhztheplayer:
### Description

Part of https://github.com/apache/incubator-gluten/issues/6920 which is the final goal.

1. Rename `gluten-data` to `gluten-arrow`
2. Make `gluten-arrow` depend on `gluten-core` than on `gluten-substrait` (prerequisite: #7031)

PR #7340, opened 2024-09-25
[VL] Follow-ups for #7304

zhztheplayer:
Follow-ups for #7304

To keep the new module dependency graph defined in https://github.com/apache/incubator-gluten/issues/7264.

cc @zml1206 @PHILO-HE

### DIFF
--- a/gluten-arrow/src/main/scala/org/apache/spark/sql/utils/SparkArrowUtil.scala
+++ b/gluten-arrow/src/main/scala/org/apache/spark/sql/utils/SparkArrowUtil.scala
@@ -16,13 +16,14 @@
  */
 package org.apache.spark.sql.utils
 
-import org.apache.gluten.expression.ConverterUtils
-
+import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
 
 import org.apache.arrow.vector.complex.MapVector
 import org.apache.arrow.vector.types.{DateUnit, FloatingPointPrecision, IntervalUnit, TimeUnit}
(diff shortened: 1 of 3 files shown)


gluten-arrow/src/main/scala/org/apache/spark/sql/utils/SparkArrowUtil.scala
```diff
@@ -97,7 +98,7 @@ object SparkArrowUtil {
             .map(
               field =>
                 toArrowField(
-                  ConverterUtils.normalizeStructFieldName(field.name),
+                  normalizeColName(field.name),
```
PHILO-HE, line 101: @zhztheplayer, looks `normalizeColName` is not equivalent to `normalizeStructFieldName` which makes the conversion disabled for CH backend.

zml1206: It seems gluten-arrow only used for velox backend.

PHILO-HE: I see. Thanks!

PHILO-HE (approved):

gluten-arrow/src/main/scala/org/apache/spark/sql/utils/SparkArrowUtil.scala
zhztheplayer: Yes. Let's keep `gluten-arrow` an individual module so future refactors can be made easier. Minor code duplications like this one will be acceptable.

PR #7353, opened 2024-09-25
[VL] Minor: Fix documentation of flushable partial aggregate

surnaik:

The documentation of min partial aggregate rows and min partial aggregate pct to abandon was swapped. Minor fix to correct this.


zhouyuan (approved):
👍

Issue #7313, opened 2024-09-23
[VL] Explicit Arrow transitions that are viewable in query plan and Spark SQL UI


zhztheplayer:
### Description

In Gluten's query processing framework, we have applied the following Arrow-related data transitions that are actually implicit to users:

1. `Spark-Arrow` to `Native-Arrow` through Java API `ColumnarBatches#ensureOffloaded`
2. `Native-Arrow` / `Native-Velox` to `Spark-Arrow` through Java API `ColumnarBatches#ensureLoaded`
3. `Native-Arrow` to `Native-Velox` though C++ API `VeloxColumnarBatch::from`

These implicit transitions are developer-friendly but make performance tunning hard. As we have adopted [transition API](https://github.com/apache/incubator-gluten/issues/5764) in Gluten's query planner, now it becomes possible to pull up the work of adding these transitions to query planning phase.

PR #7360, opened 2024-09-26
[GLUTEN-7359][VL] feat: Support columnar partial project for UDF

jinchengchenghh:
The UDF may cause the total project input col to row converter and the output row to column.

For example, hash(udf(col0)), col1, col2, col3, col4, with partial project, we can only convert col0 as column to row or column to arrow as input, and convert udf(col0) as Alias partitalProject1_, then row to column.
ProjectExecTransformer will be hash(partitalProject1_), col1, col2, col3, col4,partitalProject1_.
Partial project will supply two path, velox columnar to row as input and row to velox columnar as output by UnsafeProjection.
velox columnar to arrow as input and arrow to velox columnar as output by MutableProjection.

The test results indicates it can help accelerate some queries 30%.

### DIFF
--- a/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
+++ b/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
@@ -46,6 +46,9 @@ public final class VeloxColumnarBatches {
   }
 
   public static Boolean isVeloxBatch(ColumnarBatch batch) {
+    if (!ColumnarBatches.isLightBatch(batch)) {
+      return false;
+    }
     final String comprehensiveType = ColumnarBatches.getComprehensiveLightBatchType(batch);
     return Objects.equals(comprehensiveType, COMPREHENSIVE_TYPE_VELOX);
   }
--- a/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
+++ b/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
@@ -45,14 +45,6 @@ public final class VeloxColumnarBatches {
         String.format("Comprehensive batch type is already %s", COMPREHENSIVE_TYPE_VELOX));
   }
 
-  public static Boolean isVeloxBatch(ColumnarBatch batch) {
-    if (!ColumnarBatches.isLightBatch(batch)) {
-      return false;
-    }
-    final String comprehensiveType = ColumnarBatches.getComprehensiveLightBatchType(batch);
-    return Objects.equals(comprehensiveType, COMPREHENSIVE_TYPE_VELOX);
-  }
-
   public static ColumnarBatch toVeloxBatch(ColumnarBatch input) {
     checkNonVeloxBatch(input);
     final Runtime runtime = Runtimes.contextInstance("VeloxColumnarBatches#toVeloxBatch");
--- a/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
+++ b/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
@@ -45,6 +45,11 @@ public final class VeloxColumnarBatches {
         String.format("Comprehensive batch type is already %s", COMPREHENSIVE_TYPE_VELOX));
   }
 
+  public static Boolean isVeloxBatch(ColumnarBatch batch) {
+    final String comprehensiveType = ColumnarBatches.getComprehensiveLightBatchType(batch);
+    return Objects.equals(comprehensiveType, COMPREHENSIVE_TYPE_VELOX);
+  }
+
   public static ColumnarBatch toVeloxBatch(ColumnarBatch input) {
     checkNonVeloxBatch(input);
     final Runtime runtime = Runtimes.contextInstance("VeloxColumnarBatches#toVeloxBatch");
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
@@ -75,6 +75,7 @@ private object VeloxRuleApi {
     injector.injectTransform(_ => CollapseProjectExecTransformer)
     injector.injectTransform(c => FlushableHashAggregateRule.apply(c.session))
     injector.injectTransform(c => InsertTransitions(c.outputsColumnar))
+    injector.injectTransform(c => PartialProjectRule.apply(c.session))
 
     // Gluten columnar: Fallback policies.
     injector.injectFallbackPolicy(
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
@@ -66,6 +66,7 @@ private object VeloxRuleApi {
     injector.injectTransform(_ => RewriteSparkPlanRulesManager())
     injector.injectTransform(_ => AddFallbackTagRule())
     injector.injectTransform(_ => TransformPreOverrides())
+    injector.injectTransform(c => PartialProjectRule.apply(c.session))
     injector.injectTransform(_ => RemoveNativeWriteFilesSortAndProject())
     injector.injectTransform(c => RewriteTransformer.apply(c.session))
     injector.injectTransform(_ => PushDownFilterToScan)
@@ -75,7 +76,6 @@ private object VeloxRuleApi {
     injector.injectTransform(_ => CollapseProjectExecTransformer)
     injector.injectTransform(c => FlushableHashAggregateRule.apply(c.session))
     injector.injectTransform(c => InsertTransitions(c.outputsColumnar))
-    injector.injectTransform(c => PartialProjectRule.apply(c.session))
 
     // Gluten columnar: Fallback policies.
     injector.injectFallbackPolicy(
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
@@ -405,7 +405,8 @@ object ColumnarPartialProjectExec {
     val newProjectList = original.projectList.map {
       p => replaceExpressionUDF(p, replacedAliasUdf).asInstanceOf[NamedExpression]
     }
-    val partialProject = ColumnarPartialProjectExec(original, original.child)(replacedAliasUdf.toSeq)
+    val partialProject =
+      ColumnarPartialProjectExec(original, original.child)(replacedAliasUdf.toSeq)
     ProjectExecTransformer(newProjectList, partialProject)
   }
 }
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
@@ -20,7 +20,6 @@ import org.apache.gluten.GlutenConfig
 import org.apache.gluten.columnarbatch.{ColumnarBatches, VeloxColumnarBatches}
 import org.apache.gluten.expression.ExpressionUtils
 import org.apache.gluten.extension.{GlutenPlan, ValidationResult}
-import org.apache.gluten.extension.columnar.validator.Validator.Passed
 import org.apache.gluten.iterator.Iterators
 import org.apache.gluten.memory.arrow.alloc.ArrowBufferAllocators
 import org.apache.gluten.sql.shims.SparkShimLoader
@@ -157,33 +156,14 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
     if (!original.projectList.forall(validateExpression(_))) {
       return ValidationResult.failed("Contains expression not supported")
     }
-    if (isComplexExpression()) {
+    if (
+      ExpressionUtils.isComplexExpression(
+        original,
+        GlutenConfig.getConf.fallbackExpressionsThreshold)
+    ) {
       return ValidationResult.failed("Fallback by complex expression")
     }
     ValidationResult.succeeded
-  }
-
-  private def isComplexExpression(): Boolean = {
-    new ExpressionUtils.FallbackComplexExpressions(
-      GlutenConfig.getConf.fallbackExpressionsThreshold)
-      .validate(original) match {
-      case Passed => false
-      case _ => true
-    }
-  }
-
-  private def toVeloxBatch(input: ColumnarBatch): ColumnarBatch = {
-    if (VeloxColumnarBatches.isVeloxBatch(input)) {
-      return input
-    }
-    if (ColumnarBatches.isHeavyBatch(input)) {
-      if (input.numCols() == 0) {
-        return input
-      }
-      return VeloxColumnarBatches.toVeloxBatch(
-        ColumnarBatches.offload(ArrowBufferAllocators.contextInstance(), input))
-    }
-    VeloxColumnarBatches.toVeloxBatch(input)
   }
 
   override protected def doExecuteColumnar(): RDD[ColumnarBatch] = {
@@ -209,7 +189,9 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
               val batchIterator = projectedBatch.map {
                 b =>
                   if (b.numCols() != 0) {
-                    val veloxBatch = toVeloxBatch(batch)
+                    val veloxBatch =
+                      if (batch.numCols() == 0) batch
+                      else VeloxColumnarBatches.toVeloxBatch(batch)
                     val compositeBatch = VeloxColumnarBatches.compose(veloxBatch, b)
                     b.close()
                     compositeBatch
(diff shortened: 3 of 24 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -72,6 +73,8 @@ trait SparkPlanExecApi {
       child: SparkPlan): ProjectExecTransformer =
     ProjectExecTransformer.createUnsafe(projectList, child)
 
+  def genSparkPartialProjectColumnarExec(original: ProjectExec): GlutenPlan = null
```
zhztheplayer, line 76: Instead of adding code to `OffloadSingleNode`, can we just add a new rule to [Velox's rule list](https://github.com/apache/incubator-gluten/blob/5790c30959dad59284a5e50015d94912b79f1b23/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala#L47), then remove this API from backend API?

backends-velox/src/main/scala/org/apache/gluten/execution/SparkPartialProjectColumnarExec.scala
```diff
@@ -0,0 +1,408 @@
...
+ * + b + hash(c)), child is (a, b,c ) SparkPartialProjectColumnarExec (a, b, c, myudf(a)),
+ * ProjectExecTransformer(myudf(a) + b + hash(c))
+ *
+ * @param original
+ *   extract the ScalaUDF from original project list as Alias in UnsafeProjection and
+ *   AttributeReference in SparkPartialProjectColumnarExec output
+ * @param child
+ *   child plan
+ */
+case class SparkPartialProjectColumnarExec(original: ProjectExec, child: SparkPlan)(
```
zhztheplayer, line 52: Could put `Columnar` on LHS for plan naming. Like `ColumnarSparkPartialProjectExec`. 

Though it's still weird to insert `Spark` into a plan name. Do we have better choice here?

backends-velox/src/main/scala/org/apache/gluten/execution/SparkPartialProjectColumnarExec.scala
```diff
@@ -0,0 +1,408 @@
...
+import org.apache.spark.sql.execution.metric.{SQLMetric, SQLMetrics}
+import org.apache.spark.sql.execution.vectorized.{MutableColumnarRow, WritableColumnVector}
+import org.apache.spark.sql.hive.HiveUdfUtil
+import org.apache.spark.sql.types.{BinaryType, BooleanType, ByteType, DataType, DateType, DecimalType, DoubleType, FloatType, IntegerType, LongType, NullType, ShortType, StringType, TimestampType, YearMonthIntervalType}
+import org.apache.spark.sql.vectorized.{ColumnarBatch, ColumnVector}
+
+import scala.collection.mutable.ListBuffer
+
+/**
+ * Change the Project to ProjectExecTransformer + SparkPartialProjectColumnarExec e.g. sum(myudf(a)
```
zhztheplayer, line 42:
```
By rule <rule name>, the project that is not offload-able is changed to ....
```

backends-velox/src/main/scala/org/apache/gluten/execution/SparkPartialProjectColumnarExec.scala
```diff
@@ -0,0 +1,408 @@
...
+ * + b + hash(c)), child is (a, b,c ) SparkPartialProjectColumnarExec (a, b, c, myudf(a)),
```
zhztheplayer, line 43: s/`(a, b,c )`/`(a, b, c)`

backends-velox/src/main/scala/org/apache/gluten/execution/SparkPartialProjectColumnarExec.scala
```diff
@@ -0,0 +1,408 @@
...
+  final override def doExecute(): RDD[InternalRow] = {
+    throw new UnsupportedOperationException(
+      s"${this.getClass.getSimpleName} doesn't support doExecute")
+  }
+
+  final override protected def otherCopyArgs: Seq[AnyRef] = {
+    replacedAliasUdf :: Nil
+  }
+
+  final override lazy val supportsColumnar: Boolean = true
```
zhztheplayer, line 88: Could remove `lazy`

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
```diff
@@ -116,7 +116,7 @@ object Validators {
     }
   }
 
-  private class FallbackComplexExpressions(threshold: Int) extends Validator {
+  class FallbackComplexExpressions(threshold: Int) extends Validator {
```
zhztheplayer, line 119: Instead of making this public, could move the shared code into `ExpressionUtils`

gluten-substrait/src/test/scala/org/apache/gluten/execution/WholeStageTransformerSuite.scala
```diff
@@ -316,7 +316,6 @@ abstract class WholeStageTransformerSuite
       }
     }
     checkDataFrame(noFallBack, customCheck, df)
-    df.explain(true)
```
zhztheplayer, line 319: My mistake. Thanks for fixing.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1864,6 +1866,12 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val ENABLE_COLUMNAR_PARTIAL_PROJECT =
+    buildConf("spark.gluten.sql.columnar.partial.project")
+      .doc("Execute partial project which is not supported in backend in Spark")
```
zhztheplayer, line 1871:
```
Break up one project node into 2 phases when some of the expressions are non offload-able. Stage one is a regular offloaded project transformer that evaluates the offload-able expressions in native, stage two preserves the output from stage one and evaluates the remaining non-offload-able expressions using vanilla Spark projections.
```

backends-velox/src/main/scala/org/apache/gluten/execution/SparkPartialProjectColumnarExec.scala
```diff
@@ -0,0 +1,408 @@
...
+    } else if (replacedAliasUdf.size > original.output.size) {
+      // e.g. udf1(col) + udf2(col), it will introduce 2 cols for r2c
+      ValidationResult.failed("Number of RowToColumn columns is more than ProjectExec")
+    } else if (!original.projectList.forall(validateExpression(_))) {
+      ValidationResult.failed("Contains expression not supported")
+    } else if (isComplexExpression()) {
+      ValidationResult.failed("Fallback by complex expression")
+    } else {
+      ValidationResult.succeeded
+    }
```
zhztheplayer, line 158: Can we break this long `if-else` sequence into individual `if-return` blocks?

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -0,0 +1,406 @@
...
+      .create()
+    // TODO: should check the size <= 1, but now it has bug, will change iterator to empty
+  }
+
+  override def verboseStringWithOperatorId(): String = {
+    s"""
+       |$formattedNodeName
+       |${ExplainUtils.generateFieldString("Output", output)}
+       |${ExplainUtils.generateFieldString("Input", child.output)}
+       |${ExplainUtils.generateFieldString("ScalaUDF", replacedAliasUdf)}
```
waitinfuture, line 319: nit: ScalaUDF -> UDF

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
```diff
@@ -75,6 +75,7 @@ private object VeloxRuleApi {
     injector.injectTransform(_ => CollapseProjectExecTransformer)
     injector.injectTransform(c => FlushableHashAggregateRule.apply(c.session))
     injector.injectTransform(c => InsertTransitions(c.outputsColumnar))
+    injector.injectTransform(c => PartialProjectRule.apply(c.session))
```
zhztheplayer, line 78: minor: Can we move this in between `TransformPreOverrides` and `RemoveNativeWriteFilesSortAndProject`? Thanks!

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -0,0 +1,413 @@
...
+    "time" -> SQLMetrics.createTimingMetric(sparkContext, "total time of partial project"),
+    "column_to_row_time" -> SQLMetrics.createTimingMetric(
+      sparkContext,
+      "time of velox to Arrow ColumnarBatch or UnsafeRow"),
+    "row_to_column_time" -> SQLMetrics.createTimingMetric(
+      sparkContext,
+      "time of Arrow ColumnarBatch or UnsafeRow to velox")
+  )
+
+  override def output: Seq[Attribute] = child.output ++ replacedAliasUdf.map(_.toAttribute)
```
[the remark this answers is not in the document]
jinchengchenghh: We will compose the original output and replacedAliasUDF output batch, the child.output will be used in parent operator `ProjectExecTransformer`

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -0,0 +1,413 @@
...
+      case _: FloatType => true
+      case _: DoubleType => true
+      case _: StringType => true
+      case _: TimestampType => true
+      case _: DateType => true
+      case _: BinaryType => true
+      case _: DecimalType => true
+      case YearMonthIntervalType.DEFAULT => true
+      case _: NullType => true
+      case _ => false
```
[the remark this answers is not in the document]
jinchengchenghh: We extract the column from child output by `column index` for compiling expression UDF input, we can also extend complex data type support if the API ColumnarBatches.select() can support complex data type.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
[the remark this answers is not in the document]
waitinfuture:
```
* ProjectExecTransformer + ColumnarPartialProjectExec e.g. sum(myudf(a) + b + hash(c)), child is
 * (a, b, c) ColumnarPartialProjectExec (a, b, c, myudf(a)), ProjectExecTransformer(myudf(a) + b +
 * hash(c))
```
As the example you give, `a` is not needed by `ProjectExecTransformer`

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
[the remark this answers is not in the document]
waitinfuture: IIUC in cpp code `VeloxColumnarBatch::select` supports complex types, it merely selects some columns and wrap to a new RowVector. In java code `ColumnarBatches.select()` just returns a light column batch. Also, c2r and r2c already supports complex types, so I guess it's already supported?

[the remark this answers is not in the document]
jinchengchenghh: No, `VeloxColumnarBatch::select` input argument is the column index, it supports part of complex data type, supports udf(strcut), but if the udf is udf(struct.int), we cannot select the column. In other words, we cannot resolve the table schema
```
struct
   int
   bitint
int
map
```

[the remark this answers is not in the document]
waitinfuture: IIUC in this PR the whole expression tree rooted at udf will be evaluated in java, in the case of `udf(struct.int)`, which is `udf(GetStructField(struct, 0))`, the input to the expr tree is still struct.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -172,6 +172,20 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
...
+    if (VeloxColumnarBatches.isVeloxBatch(input)) {
+      return input
+    }
+    if (ColumnarBatches.isHeavyBatch(input)) {
+      if (input.numCols() == 0) {
+        return input
+      }
+      return VeloxColumnarBatches.toVeloxBatch(
+        ColumnarBatches.offload(ArrowBufferAllocators.contextInstance(), input))
+    }
```
zhztheplayer, line 185:
```
If (batch is not in velox format) {
  convert batch to velox
}
```

Such logics are considered implicit transitions and were removed in effort of https://github.com/apache/incubator-gluten/issues/7313. 

Operators now have certain input formats so we can make clear assumption that which batch is in which format. So let's remove these implicit transitions. Thanks.

jinchengchenghh: The ArrowCsvReader may produce Arrow format heavy ColumnarBatch, does it work?

jinchengchenghh: For the plan ProjectExec, child plan is csv scan.

zhztheplayer: > The ArrowCsvReader may produce Arrow format heavy ColumnarBatch, does it work?

It should work. Planner should take the work to insert Arrow-to-Velox transitions now, you can run some tests to see the final plan Gluten gives.

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
```diff
@@ -45,6 +45,14 @@ public static void checkNonVeloxBatch(ColumnarBatch batch) {
         String.format("Comprehensive batch type is already %s", COMPREHENSIVE_TYPE_VELOX));
   }
 
+  public static Boolean isVeloxBatch(ColumnarBatch batch) {
+    if (!ColumnarBatches.isLightBatch(batch)) {
+      return false;
+    }
+    final String comprehensiveType = ColumnarBatches.getComprehensiveLightBatchType(batch);
+    return Objects.equals(comprehensiveType, COMPREHENSIVE_TYPE_VELOX);
+  }
```
zhztheplayer, line 54: Let's remove this API as it's for implicit transitions which were deprecated.

zhztheplayer (commented):
Some minors remaining. Thank you for your effort on this!

zhztheplayer (approved):
👍

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
zhztheplayer: nit: Remove this if unused. Thanks.

zhztheplayer:
CI is failing at the known CSV issue
```
- csv scan with option string as null
24/10/17 03:19:28 WARN GlutenFallbackReporter: Validation failed for plan: Scan text [QueryId=3356], due to: Unsupported file format for UnknownFormat..
E20241017 03:19:28.696393 29864 Exceptions.h:66] Line: /__w/incubator-gluten/incubator-gluten/ep/build-velox/build/velox_ep/velox/exec/Task.cpp:1955, Function:terminate, Expression:  Cancelled, Source: RUNTIME, ErrorCode: INVALID_STATE
24/10/17 03:19:28 WARN GlutenFallbackReporter: Validation failed for plan: Scan text , due to: Unsupported file format for UnknownFormat..
24/10/17 03:19:28 WARN ArrowNativeMemoryPool: Arrow pool still reserved non-zero bytes, which may cause memory leak, size: 128.0 B. 
#
#
#  SIGSEGV (0xb) at pc=0x00007fa14049fcd0, pid=1158, tid=0x00007fa091b7e700
#
# JRE version: OpenJDK Runtime Environment (8.0_312-b07) (build 1.8.0_312-b07)
# Java VM: OpenJDK 64-Bit Server VM (25.312-b07 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  0x00007fa14049fcd0
#
# Core dump written. Default location: /__w/incubator-gluten/incubator-gluten/backends-velox/core or core.1158
#
# /__w/incubator-gluten/incubator-gluten/backends-velox/hs_err_pid1158.log
- csv scan with option delimiter
#
# If you would like to submit a bug report, please visit:
#   https://bugzilla.redhat.com/enter_bug.cgi?product=Red%20Hat%20Enterprise%20Linux%208&component=java-1.8.0-openjdk
#
```

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/hive/execution/GlutenHiveUDFSuite.scala
```diff
@@ -0,0 +1,139 @@
...
+        // Disable ConvertToLocalRelation for better test coverage. Test cases built on
+        // LocalRelation will exercise the optimization rules better by disabling it as
+        // this rule may potentially block testing of other optimization rules such as
+        // ConstantPropagation etc.
+        .set(SQLConf.OPTIMIZER_EXCLUDED_RULES.key, ConvertToLocalRelation.ruleName)
+    ),
+    false
+  ) {}
+
+class GlutenHiveUDFSuite
```
zhztheplayer, line 79: @jinchengchenghh Kindly reminder that this suite doesn't get run as it's not added into `VeloxTestSettings`

jinchengchenghh: Yes, I know. Because it is different with other suite. It could run the single suite successfully but failed in the total mvn test in case of stopped SparkContext.

Issue #7362, opened 2024-09-26
[VL] Diff of in_or_and


zml1206:
### Backend

VL (Velox)

### Bug description

```
select count(1) from lineitem
where (l_shipmode in ('TRUCK', 'MAIL') or l_shipmode in ('AIR', 'FOB')) 
and l_shipmode in ('RAIL','SHIP')
```
vanilla spark is 0, gluten have the same results with sql
```
select count(1) from lineitem
where l_shipmode in ('TRUCK', 'MAIL') or l_shipmode in ('AIR', 'FOB')
or l_shipmode in ('RAIL','SHIP')
```

### Spark version

None

rui-mo:
@zml1206 I will take a look. Thanks.

zml1206:
I have submitted a PR #7363  to resolve it. @rui-mo

PR #6754, opened 2024-08-08
[VL] Use extractFiltersFromRemainingFilter provided by Velox

rui-mo:
## What changes were proposed in this pull request?

https://github.com/apache/incubator-gluten/issues/6666

## How was this patch tested?

under test

PR #7363, opened 2024-09-26
[GLUTEN-7362][VL] Add test for 'IN' and 'OR' filter in Scan

zml1206:

Fixes: \#7362
Resolved by #6754, add test for it.

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
@@ -208,6 +208,13 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
       "select l_orderkey from lineitem " +
         "where l_partkey in (1552, 674) or l_partkey in (1552) and l_orderkey > 1") { _ => }
     checkLengthAndPlan(df, 73)
+
+    runQueryAndCompare(
+      "select count(1) from lineitem " +
+        "where (l_shipmode in ('TRUCK', 'MAIL') or l_shipmode in ('AIR', 'FOB')) " +
+        "and l_shipmode in ('RAIL','SHIP')") {
+      checkGlutenOperatorMatch[FileSourceScanExecTransformer]
+    }
   }
 
   test("in_not") {


FelixYBW:
Is this a Gluten issue?

zml1206:
> Is this a Gluten issue?

Yes, the first singularOrList in `OR` needs to check in range.

zml1206:
cc @rui-mo Thank you.

backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
```diff
@@ -239,6 +239,13 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
       "select l_orderkey from lineitem " +
         "where l_partkey in (1552, 674) or l_partkey in (1552) and l_orderkey > 1") { _ => }
     checkLengthAndPlan(df, 73)
+
+    runQueryAndCompare(
+      "select count(1) from lineitem " +
+        "where (l_shipmode in ('TRUCK', 'MAIL') or l_shipmode in ('AIR', 'FOB')) " +
+        "and l_shipmode in ('RAIL','SHIP')") {
```
rui-mo, line 246: I think this issue is in the condition `(A or B) and C`, when C is already pushed down, the `(A or B)` cannot be pushed down.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1669,7 +1669,8 @@ bool SubstraitToVeloxPlanConverter::canPushdownOr(
   }
```
rui-mo, line 1669: And I suggest we add a `setOrRange` function in RangeRecorder which returns false when other conditions already exist for a field.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1686,14 +1687,15 @@ bool SubstraitToVeloxPlanConverter::canPushdownOr(
...
     } else if (arg.value().has_singular_or_list()) {
       const auto& singularOrList = arg.value().singular_or_list();
+      // Disable IN pushdown for int-like types.
       if (!canPushdownSingularOrList(singularOrList, true)) {
         return false;
       }
       uint32_t fieldIdx = getColumnIndexFromSingularOrList(singularOrList);
-      // Disable IN pushdown for int-like types.
-      if (!rangeRecorders.at(fieldIdx).setInRange(true /*forOrRelation*/)) {
+      if (!rangeRecorders.at(fieldIdx).setInRange(sign /*forOrRelation*/)) {
```
rui-mo, line 1695: We don't need this change as this logic is checking the nested expressions in an `or` condition, while here the issue is `or` cannot be pushed down totally.

rui-mo (commented):
Thanks. These are my thoughts.

backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
zml1206: Yes, but only for `in`, which is `singularOrList`.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
zml1206: `setInRange` itself retains a parameter `forOrRelation`, which is used for verification. We only need to judge the first singularOrList in or.

zml1206: Thanks for your suggestion, updated.

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -286,6 +286,16 @@ class SubstraitToVeloxPlanConverter {
       return true;
     }
 
+    /// Set the existence of or-range and returns whether it can coexist with
+    /// existing conditions for this field.
+    bool setOrRange() {
```
rui-mo, line 291: Sorry but I just notice 'setMultiRange' is for the 'or' condition, so we can just use it.

rui-mo (commented):
Thanks.

cpp/velox/substrait/SubstraitToVeloxPlan.h
zml1206: `setMultiRange` seems to conflict with `setCertainRangeForFunction` of 1681L.

backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
```diff
@@ -239,6 +239,13 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
       "select l_orderkey from lineitem " +
         "where l_partkey in (1552, 674) or l_partkey in (1552) and l_orderkey > 1") { _ => }
     checkLengthAndPlan(df, 73)
+
+    runQueryAndCompare(
```
rui-mo, line 243: We have merged https://github.com/apache/incubator-gluten/commit/d6326f06c8158de2eb08bb9ba7e3e4eb6e3964ef. Would you like to try if this issue has been fixed? Perhaps we can merge this test to ensure the functionality. Thanks for your patience.

zml1206: Thank you, I will try it locally later.

zml1206:
Local test passed, I modified pr to only add ut. @rui-mo

PR #7365, opened 2024-09-26
[GLUTEN-7364][CORE] Simplify the RuleInjector

beliefer:

This PR proposes to simplify the `RuleInjector`.
The change increases the readability. There are two change.

1. Avoid copy code from `SparkSessionExtensions` and simplify the `SparkInjector`.  The related `RuleApi` also updated. In fact, we can remove the `SparkInjector`, but I suspect it already became the public API.
2. Simplify the `GlutenInjector`.

## How was this patch tested?

integration tests

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
@@ -46,7 +46,7 @@ class VeloxRuleApi extends RuleApi {
 
 private object VeloxRuleApi {
   def injectSpark(injector: SparkInjector): Unit = {
-    // Regular Spark rules.
+    // Inject the regular Spark rules directly.
     injector.injectOptimizerRule(CollectRewriteRule.apply)
     injector.injectOptimizerRule(HLLRewriteRule.apply)
     injector.injectPostHocResolutionRule(ArrowConvertorRule.apply)
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/injector/RuleInjector.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/injector/RuleInjector.scala
@@ -19,12 +19,13 @@ package org.apache.gluten.extension.injector
 import org.apache.spark.sql.SparkSessionExtensions
 
 /** Injector used to inject query planner rules into Spark and Gluten. */
-class RuleInjector {
-  val spark: SparkInjector = new SparkInjector()
+class RuleInjector(extensions: SparkSessionExtensions) {
+  val spark: SparkInjector = new SparkInjector(extensions)
   val gluten: GlutenInjector = new GlutenInjector()
 
-  private[extension] def inject(extensions: SparkSessionExtensions): Unit = {
-    spark.inject(extensions)
+  private[extension] def inject(): Unit = {
+    // The regular Spark rules already injected with the `injectRules` of `RuleApi` directly.
+    // Only inject the Spark columnar rule here.
     gluten.inject(extensions)
   }
 }
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/injector/SparkInjector.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/injector/SparkInjector.scala
@@ -25,59 +25,35 @@ import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution.SparkPlan
 
-import scala.collection.mutable
-
 /** Injector used to inject query planner rules into Spark. */
-class SparkInjector private[injector] {
-  private type RuleBuilder = SparkSession => Rule[LogicalPlan]
-  private type StrategyBuilder = SparkSession => Strategy
-  private type ParserBuilder = (SparkSession, ParserInterface) => ParserInterface
-  private type FunctionDescription = (FunctionIdentifier, ExpressionInfo, FunctionBuilder)
-  private type QueryStagePrepRuleBuilder = SparkSession => Rule[SparkPlan]
+class SparkInjector private[injector] (extensions: SparkSessionExtensions) {
 
-  private val queryStagePrepRuleBuilders = mutable.Buffer.empty[QueryStagePrepRuleBuilder]
-  private val parserBuilders = mutable.Buffer.empty[ParserBuilder]
-  private val resolutionRuleBuilders = mutable.Buffer.empty[RuleBuilder]
-  private val optimizerRules = mutable.Buffer.empty[RuleBuilder]
-  private val plannerStrategyBuilders = mutable.Buffer.empty[StrategyBuilder]
-  private val injectedFunctions = mutable.Buffer.empty[FunctionDescription]
-  private val postHocResolutionRuleBuilders = mutable.Buffer.empty[RuleBuilder]
-
-  def injectQueryStagePrepRule(builder: QueryStagePrepRuleBuilder): Unit = {
-    queryStagePrepRuleBuilders += builder
+  def injectQueryStagePrepRule(builder: SparkSession => Rule[SparkPlan]): Unit = {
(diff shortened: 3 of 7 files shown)


beliefer:
ping @zhouyuan @zzcclp @liuneng1994

beliefer:
cc @zhztheplayer @FelixYBW

zhouyuan (dismissed):
👍

gluten-core/src/main/scala/org/apache/gluten/extension/injector/SparkInjector.scala
```diff
@@ -16,68 +16,8 @@
...
-    parserBuilders.foreach(extensions.injectParser)
-    resolutionRuleBuilders.foreach(extensions.injectResolutionRule)
-    optimizerRules.foreach(extensions.injectOptimizerRule)
-    plannerStrategyBuilders.foreach(extensions.injectPlannerStrategy)
-    injectedFunctions.foreach(extensions.injectFunction)
-    postHocResolutionRuleBuilders.foreach(extensions.injectPostHocResolutionRule)
-  }
-}
+@deprecated("This class is deprecated and will be removed in future versions.", since = "1.3.0")
+class SparkInjector private[injector] (val extensions: SparkSessionExtensions) {}
```
zhztheplayer, line 23: I think we can just remove minor unused APIs like this one, as Gluten doesn't officially guarantee backward compatibility yet.

gluten-core/src/main/scala/org/apache/gluten/extension/injector/RuleInjector.scala
```diff
@@ -19,12 +19,13 @@ package org.apache.gluten.extension.injector
...
+class RuleInjector(extensions: SparkSessionExtensions) {
+  val spark: SparkInjector = new SparkInjector(extensions)
   val gluten: GlutenInjector = new GlutenInjector()
 
   private[extension] def inject(extensions: SparkSessionExtensions): Unit = {
-    spark.inject(extensions)
+    // The regular Spark rules already injected with the `injectRules` of `RuleApi` directly.
+    // Only inject the Spark columnar rule here.
     gluten.inject(extensions)
   }
```
zhztheplayer, line 30: Both constructor and method `#inject` accepts a `extensions: SparkSessionExtensions`. Can we remove the one in `#inject`?

zhztheplayer (commented):
@beliefer Overall lgtm except for some minors. Thanks,

gluten-core/src/main/scala/org/apache/gluten/extension/injector/SparkInjector.scala
beliefer: As expected.

gluten-core/src/main/scala/org/apache/gluten/extension/injector/RuleInjector.scala
beliefer: updated.

zhztheplayer (dismissed):

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
```diff
@@ -38,18 +38,18 @@ class VeloxRuleApi extends RuleApi {
...
   }
 }
 
 private object VeloxRuleApi {
-  def injectSpark(injector: SparkInjector): Unit = {
-    // Regular Spark rules.
-    injector.injectOptimizerRule(CollectRewriteRule.apply)
-    injector.injectOptimizerRule(HLLRewriteRule.apply)
-    injector.injectPostHocResolutionRule(ArrowConvertorRule.apply)
+  def injectSpark(injector: RuleInjector): Unit = {
```
zhztheplayer, line 48: Can we make it 

```suggestion
  def injectSpark(exts: SparkSessionExtensions): Unit = {
```

? Thanks.

gluten-core/src/main/scala/org/apache/gluten/extension/injector/RuleInjector.scala
```diff
@@ -19,12 +19,12 @@ package org.apache.gluten.extension.injector
 import org.apache.spark.sql.SparkSessionExtensions
 
 /** Injector used to inject query planner rules into Spark and Gluten. */
-class RuleInjector {
-  val spark: SparkInjector = new SparkInjector()
+class RuleInjector(val extensions: SparkSessionExtensions) {
```
zhztheplayer, line 22:
```suggestion
class RuleInjector(val spark: SparkSessionExtensions) {
```

zhztheplayer (commented):
Thanks @beliefer, just some suggestions about naming left.

Though overall I will be +0 to this. Do you have some specific reasons to make the change?

The reason why we split the planning phase into `spark` and `gluten` is that we wanted to isolate Gluten's query planer with Spark, to the maximum extent. And we wanted Gluten to inject only one single built-in Spark `ColumnarRule` through `GlutenInjector`, so we hided `#injectColumnar` method by wrapping `SparkSessionExtensions` with `SparkInjector`. If we expose that API, we may need to pay more attention on Gluten PRs that add rules to make sure they do things right.

beliefer:
@zhztheplayer Please review again! Thank you.

gluten-core/src/main/scala/org/apache/gluten/extension/injector/RuleInjector.scala
beliefer: In the world of Spark, Spark usually means to SparkSession.
For consistent, let's use `exts`.

beliefer:
> The reason why we split the planning phase into `spark` and `gluten` is that we wanted to isolate Gluten's query planer with Spark, to the maximum extent. And we wanted Gluten to inject only one single built-in Spark `ColumnarRule` through `GlutenInjector`, so we hided `#injectColumnar` method by wrapping `SparkSessionExtensions` with `SparkInjector`. If we expose that API, we may need to pay more attention on Gluten PRs that add rules to make sure they do things right.


The rules for isolating Spark and Gluten are not a problem. As you said, Gluten to inject only one single built-in Spark `ColumnarRule` through `GlutenInjector`, so we hided `#injectColumnar` method by wrapping `SparkSessionExtensions` with `GlutenInjector`. But `SparkInjector` only copy the API from `SparkSessionExtensions`, so this PR doesn't change the `GlutenInjector` and only remove the `SparkInjector`. This PR doesn't break what the Gluten want.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
beliefer: How about `extensions`?

zhztheplayer:
> But `SparkInjector` only copy the API from `SparkSessionExtensions`, so this PR doesn't change the `GlutenInjector` and only remove the `SparkInjector`. This PR doesn't break what the Gluten want.

I meant after the PR, developer could use API `injector.extensions.injectColumnar` to inject another `ColumnarRule` in `VeloxRuleApi` / `CHRuleAPI`, since we expose `SparkSessionExtensions` to developer. Before this PR, one cannot access `SparkSessionExtensions` in `VeloxRuleApi` / `CHRuleAPI`. Just to make sure we are on the same page here.

beliefer:
> > But `SparkInjector` only copy the API from `SparkSessionExtensions`, so this PR doesn't change the `GlutenInjector` and only remove the `SparkInjector`. This PR doesn't break what the Gluten want.
> 
> I meant after the PR, developer could use API `injector.extensions.injectColumnar` to inject another `ColumnarRule` in `VeloxRuleApi` / `CHRuleAPI`, since we expose `SparkSessionExtensions` to developer. Before this PR, one cannot access `SparkSessionExtensions` in `VeloxRuleApi` / `CHRuleAPI`. Just to make sure we are on the same page here.

Got it. We can reserve the `SparkInjector`. Let me refactor it.

zhztheplayer (approved):

zhztheplayer (approved):

beliefer:
@zhztheplayer @zhouyuan Thank you!

PR #7369, opened 2024-09-26
[VL] Link shared jemalloc lib to work with LD_PRELOAD

PHILO-HE:

Provides a utility to profile jemalloc memory allocation. Recommended way to use jemalloc: set jemalloc with LD_PRELOAD at runtime to override standard memory allocator. Then, user can optionally enable this build option for debugging or profiling. The linked jemalloc lib should be as same as the one set for LD_PRELOAD. Otherwise, the printed statistic figures will not correctly reflect the usage of jemalloc.

## How was this patch tested?

Local test.


zhztheplayer:
```
[100%] Linking CXX shared library ../releases/libvelox.so
Checking ld result of libvelox.so
ld: warning: cannot find entry symbol _start; not setting start address
Checking ldd result of libvelox.so
	linux-vdso.so.1 =>  (0x00007ffef44ea000)
	libgluten.so => /__w/incubator-gluten/incubator-gluten/cpp/build/releases/libgluten.so (0x00007f4fb1a8a000)
	libresolv.so.2 => /lib64/libresolv.so.2 (0x00007f4fb1870000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f4fb166c000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f4fb1464000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f4fb1248000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f4fb0f46000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f4fb0b78000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4fb9bbf000)
```

Log doesn't show libjemalloc.so as libvelox.so's dependency, which is good to me but is it expected?

PHILO-HE:
@zhztheplayer, there is an extra option --build_jemalloc (OFF by default) to control whether to link this lib. It's still OFF in CI build.

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -378,7 +378,7 @@ VeloxMemoryManager::~VeloxMemoryManager() {
                << "ms as there are still outstanding memory resources. ";
   }
 #ifdef ENABLE_JEMALLOC
-  je_gluten_malloc_stats_print(NULL, NULL, NULL);
+  malloc_stats_print(NULL, NULL, NULL);
 #endif
```
zhztheplayer, line 382: Does this feature actually have users? I see this is the only usage of `jemalloc.h` in Gluten.

zhztheplayer (approved):

cpp/velox/memory/VeloxMemoryManager.cc
PHILO-HE: @zhztheplayer, just has internal users for debugging purpose. Yes, it's just used here currently.

PHILO-HE:
>@PHILO-HE I have a doubt about the PR title. What does this PR have to do with LD_PRELOAD? All I see is that we're building jemalloc as a shared library in /usr/local/lib

@surnaik, this pr simply allows some profiling log printed when jemalloc is used with LD_PRELOAD (set by user in runtime env.), e.g., shows memory size allocated by jemalloc. It requires linked jemalloc lib is as same as the one set by LD_PRELOAD.
 
>Enabling this now, will add libjemalloc as a dependency of libvelox. Is this to force users to provide jemalloc using LD_PRELOAD?

Not exactly. With the related build option turned on, it does not require LD_PRELOAD must be set by user. But without the use of LD_PRELOAD, turning on the option is useless since the profiling log has no helpful info (e.g., 0 byte allocated by jemalloc).

Issue #7373, opened 2024-09-27
[VL] Add document for profiling gluten with velox


wForget:
### Description

As title

PR #7374, opened 2024-09-27
[GLUTEN-7373][DOC][VL] Add document for profiling gluten with velox

wForget:

Fixes: #7373

## How was this patch tested?

no


FelixYBW:
Thank you for your input. Can you put an example chart in the doc?

FelixYBW:
Can you change the name to "profile memory consumption of Gluten".

Also you need to build gluten with -g. Like build as relwithdebinfo.

wForget:
> Thank you for your input. Can you put an example chart in the doc?

Sure, I will add it later

> Can you change the name to "profile memory consumption of Gluten".

`gperftools` can also be used as a CPU profiler. Do we still need to change it?

> Also you need to build gluten with -g. Like build as relwithdebinfo.

I'm not sure if this is required. I will update the docs after I try some more and provide a suitable build script example.

FelixYBW:
> `gperftools` can also be used as a CPU profiler. Do we still need to change it?
> 
Yes, let's add another doc about CPU hotspot profile on necessary.

wForget:
> Also you need to build gluten with -g. Like build as relwithdebinfo.

Is `-g` required? Build with release seems to be able to analyze dump files normally.

FelixYBW (approved):

PR #7388, opened 2024-09-29
[VL] pass phase in recusive invocation of spillTree

waitinfuture:
Currently recursive invocation of spillTree calls ` spillTree(TreeMemoryTarget node, final long bytes)`, which always process `SPILL` phase.


zhztheplayer (approved):
Nice catch. Thanks!

PR #7396, opened 2024-09-30
[CORE] Update AllocationListener usage after successfully releasing memory

wForget:

We should update AllocationListener usage after successfully releasing memory to avoid exceeding the usage limit.

## How was this patch tested?

code refactor

### DIFF
--- a/cpp/core/memory/MemoryAllocator.cc
+++ b/cpp/core/memory/MemoryAllocator.cc
@@ -84,7 +84,7 @@ bool ListenableMemoryAllocator::reallocateAligned(
 bool ListenableMemoryAllocator::free(void* p, int64_t size) {
   bool succeed = delegated_->free(p, size);
   if (succeed) {
-    updateUsage(-diff);
+    updateUsage(-size);
   }
   return succeed;
 }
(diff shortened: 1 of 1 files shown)


wForget:
@zhztheplayer This is just a minor improvement I found while reading the relevant codes. Could you please take a quick look?

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -50,9 +50,13 @@ bool ListenableMemoryAllocator::allocateAligned(uint64_t alignment, int64_t size
...
+    updateUsage(diff);
+  }
   bool succeed = delegated_->reallocate(p, size, newSize, out);
-  if (!succeed) {
+  if (succeed && diff < 0) {
+    updateUsage(diff);
+  } else if (!succeed && diff > 0) {
     updateUsage(-diff);
   }
   return succeed;
```
zhztheplayer, line 62: Can we use two different code branches for `diff > 0` and `diff < 0` ? E.g.,

```suggestion
  int64_t diff = newSize - size;
  if (diff >= 0) {
    updateUsage(diff);
    bool succeeded = delegated_->reallocate(p, size, newSize, out);
	if (!succeeded) {
      updateUsage(-diff);
	}
	return succeeded;
  }
  // Code path for diff < 0.
  bool succeeded = delegated_->reallocate(p, size, newSize, out);
  if (succeeded) {
    updateUsage(diff);
  }
  return succeed;
```

zhztheplayer (commented):
Just one minor comment. Thanks!

cpp/core/memory/MemoryAllocator.cc
wForget: thanks, changed

zhztheplayer (approved):
Thanks!

Yohahaha (approved):
👍

Issue #7410, opened 2024-10-04
Need to manually specify javaArgs for the Spark UTs to run when using `mvn test` with Java 17


CodenameGHOST007:
### Problem description

Several Spark UTs need javaArgs like `-Dio.netty.tryReflectionSetAccessible=true` and `--add-opens=java.base/java.nio=ALL-UNNAMED` to run and these configs are present in spark pom which makes it easier to run the tests through CLI.

An Example of the error:
```
An exception or error caused a run to abort: class org.apache.spark.storage.StorageUtils$ (in unnamed module @0x3eb738bb) cannot access class sun.nio.ch.DirectBuffer (in module java.base) because module java.base does not export sun.nio.ch to unnamed module @0x3eb738bb 
java.lang.IllegalAccessError: class org.apache.spark.storage.StorageUtils$ (in unnamed module @0x3eb738bb) cannot access class sun.nio.ch.DirectBuffer (in module java.base) because module java.base does not export sun.nio.ch to unnamed module @0x3eb738bb
	at org.apache.spark.storage.StorageUtils$.<init>(StorageUtils.scala:213)
	at org.apache.spark.storage.StorageUtils$.<clinit>(StorageUtils.scala)
	at org.apache.spark.storage.BlockManagerMasterEndpoint.<init>(BlockManagerMasterEndpoint.scala:114)
	at org.apache.spark.SparkEnv$.$anonfun$create$9(SparkEnv.scala:353)
	at org.apache.spark.SparkEnv$.registerOrLookupEndpoint$1(SparkEnv.scala:290)
	at org.apache.spark.SparkEnv$.create(SparkEnv.scala:339)
	at org.apache.spark.SparkEnv$.createDriverEnv(SparkEnv.scala:194)
	at org.apache.spark.SparkContext.createSparkEnv(SparkContext.scala:279)
	at org.apache.spark.SparkContext.<init>(SparkContext.scala:464)
	at org.apache.spark.SparkContext.<init>(SparkContext.scala:139)
	at org.apache.spark.sql.test.TestSparkSession.<init>(TestSQLContext.scala:30)
	at org.apache.spark.sql.test.SharedSparkSessionBase.createSparkSession(SharedSparkSession.scala:102)
	at org.apache.spark.sql.test.SharedSparkSessionBase.createSparkSession$(SharedSparkSession.scala:100)
[...]

PR #7411, opened 2024-10-04
[GLUTEN-7410][CORE] Add test args to run spark-ut with Java 17

CodenameGHOST007:

Spark UTs need the above args to run with Java 17 and had to be supplied seperately while running the tests. 

(Fixes: #7410)

## How was this patch tested?

Ran Spark UTs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 
 # logs
 *.log
+**/core
 
 # cache files
 *.cache
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 
 # logs
 *.log
-**/core
+**/core$
 
 # cache files
 *.cache
--- a/dev/build-thirdparty.sh
+++ b/dev/build-thirdparty.sh
@@ -82,4 +82,4 @@ elif [ "$LINUX_OS" == "debian" ]; then
   fi
 fi
 cd $THIRDPARTY_LIB/
-jar cvf gluten-thirdparty-lib-$LINUX_OS-$VERSION-$ARCH.jar ./
+"${JAVA_HOME}/bin/jar" cvf gluten-thirdparty-lib-$LINUX_OS-$VERSION-$ARCH.jar ./
--- a/dev/build-thirdparty.sh
+++ b/dev/build-thirdparty.sh
@@ -82,4 +82,4 @@ elif [ "$LINUX_OS" == "debian" ]; then
   fi
 fi
 cd $THIRDPARTY_LIB/
-"${JAVA_HOME}/bin/jar" cvf gluten-thirdparty-lib-$LINUX_OS-$VERSION-$ARCH.jar ./
+jar cvf gluten-thirdparty-lib-$LINUX_OS-$VERSION-$ARCH.jar ./
--- a/dev/package.sh
+++ b/dev/package.sh
@@ -10,6 +10,7 @@ ARCH=`uname -m`
 
 if [[ -n "$JAVA_HOME" ]]; then
   echo "JAVA_HOME needs to be set for Gluten Compilation"
+  exit 1
 fi
 
 cd "$GLUTEN_DIR"
--- a/dev/package.sh
+++ b/dev/package.sh
@@ -8,10 +8,6 @@ LINUX_OS=$(. /etc/os-release && echo ${ID})
 VERSION=$(. /etc/os-release && echo ${VERSION_ID})
 ARCH=`uname -m`
 
-if [[ -z "$JAVA_HOME" ]]; then
-  echo "JAVA_HOME needs to be set for Gluten compilation"
-  exit 1
-fi
 
 cd "$GLUTEN_DIR"
 
--- a/dev/package.sh
+++ b/dev/package.sh
@@ -8,7 +8,6 @@ LINUX_OS=$(. /etc/os-release && echo ${ID})
 VERSION=$(. /etc/os-release && echo ${VERSION_ID})
 ARCH=`uname -m`
 
-
 cd "$GLUTEN_DIR"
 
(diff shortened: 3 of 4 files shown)


CodenameGHOST007:
I checked the workflows that run spark-uts for velox backend in `velox_backend.yml` . All of them just use jdk 8 as the base where this is not a problem. I couldn't find any jdk 17 env that is running spark-uts. There are though jdk 17 envs that are just used to build and run tpc-h etc.

dev/build-thirdparty.sh
```diff
@@ -82,4 +82,4 @@ elif [ "$LINUX_OS" == "debian" ]; then
   fi
 fi
 cd $THIRDPARTY_LIB/
-jar cvf gluten-thirdparty-lib-$LINUX_OS-$VERSION-$ARCH.jar ./
+"${JAVA_HOME}/bin/jar" cvf gluten-thirdparty-lib-$LINUX_OS-$VERSION-$ARCH.jar ./
```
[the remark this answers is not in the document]
CodenameGHOST007: In a dev env its a quite common usecase to have many jdks installed and just modify JAVA_HOME to change the jdk used. This should prevent any accidental issues.

.gitignore
```diff
@@ -5,6 +5,7 @@
 
 # logs
 *.log
+**/core
```
zhztheplayer, line 8: The wildcard is too broad? Will it affect `cpp/core`?

CodenameGHOST007: > The wildcard is too broad? 
Thats as expected since we wouldn't ever want a core file to be added to git as they can be upwards of a GB. 
> Will it affect cpp/core?
If you mean the folder, then no. To effect folders the regex needs to have a `/` at the end.

CodenameGHOST007: @zhztheplayer Updated it to end at core. Thanks for catching.

CodenameGHOST007:
Looks like the ClickHouse CI failed with 
```
[2024-10-04T13:15:16.346Z] GlutenParquetProtobufCompatibilitySuite:
[2024-10-04T13:15:16.346Z] - unannotated array of primitive type *** FAILED ***
[2024-10-04T13:15:16.346Z]   spark.test.home or SPARK_HOME is not set. (SparkFunSuite.scala:128)
[2024-10-04T13:15:16.346Z] - unannotated array of struct *** FAILED ***
[2024-10-04T13:15:16.346Z]   spark.test.home or SPARK_HOME is not set. (SparkFunSuite.scala:128)
[2024-10-04T13:15:16.346Z] - struct with unannotated array !!! IGNORED !!!
[2024-10-04T13:15:16.346Z] - unannotated array of struct with unannotated array *** FAILED ***
[2024-10-04T13:15:16.346Z]   spark.test.home or SPARK_HOME is not set. (SparkFunSuite.scala:128)
[2024-10-04T13:15:16.346Z] - unannotated array of string *** FAILED ***
[2024-10-04T13:15:16.346Z]   spark.test.home or SPARK_HOME is not set. (SparkFunSuite.scala:128)
```
SPARK_HOME not set.

CodenameGHOST007:
Run Gluten Clickhouse CI

CodenameGHOST007:
It seems to be because of `-DargLine="-Dspark.test.home=/tmp/spark32" ` set in jenkins CI. I cant find the file to edit this in since its not ideal. Would need some help here.

dev/package.sh
```diff
@@ -8,6 +8,11 @@ LINUX_OS=$(. /etc/os-release && echo ${ID})
 VERSION=$(. /etc/os-release && echo ${VERSION_ID})
 ARCH=`uname -m`
 
+if [[ -n "$JAVA_HOME" ]]; then
+  echo "JAVA_HOME needs to be set for Gluten Compilation"
+  exit 1
+fi
```
zhztheplayer, line 14: s/Compilation/compilation

What's the major consideration to make JAVA_HOME required?

CodenameGHOST007: General best practice and mainly to avoid inconsistency since if the system has JAVA_HOME set gluten compilation will happen on that but third party libs will be generated with the Java on PATH of the system.

CodenameGHOST007:
@zhztheplayer Could you help rerun the spark35 scala 213 test. Seems to be some flaky issue. Nothing in reports.

dev/package.sh
zhztheplayer: @CodenameGHOST007 

I am worried whether this change break user's build processes much. Can we separate this change out to future PRs? Thanks.

CodenameGHOST007: Removed these changes. Will create a separate issue to track it. Ideally we should find and set JAVA_HOME itself before starting the build like spark and all do.

zhztheplayer (approved):
LGTM. Thanks!

CodenameGHOST007:
@zhztheplayer Thanks for the review. The failing test passed for the last two runs and failed because of `transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/enforcer/enforcer-api/3.3.0/enforcer-api-3.3.0.jar: Connection timed out (Read failed)`. I think we should be good to merge.

zhouyuan (approved):
👍

zhztheplayer:
@CodenameGHOST007 Have you tested the change locally? If yes we can merge in advance. The next move will be adding JDK17 GHA UT job then.

FelixYBW:
> tested the change locally? If yes we can merge in advance. The next mov

@CodenameGHOST007 we can't add jdk17 in CI now. Will you run some jenkins job to track the UT on jdk17?

zhouyuan:
Note: I did a test with OpenJDK-17 locally and seems ok to run the unit tests. As we are migrating to Spark-400 JDK & scala-213 should be a must in CI/CD.

thanks, -yuan

Issue #7420, opened 2024-10-07
Fix kGCSCredentialsPath configuration


majetideepak:
### Backend

VL (Velox)

### Bug description

Recent Velox update did not fix the config change.
https://github.com/apache/incubator-gluten/pull/7404

### Spark version

Spark-3.4.x

PR #7421, opened 2024-10-07
[GLUTEN-7420][VL] Fix GCS configuration

majetideepak:

We must directly set the gcs credential path in the config.
Add GCS configs to the backend config if present.

Fixes: #7420

## How was this patch tested?

Tested locally.

### DIFF
--- a/cpp/velox/utils/ConfigExtractor.cc
+++ b/cpp/velox/utils/ConfigExtractor.cc
@@ -19,9 +19,6 @@
 
 #include "ConfigExtractor.h"
 #include <stdexcept>
-#ifdef ENABLE_GCS
-#include <fstream>
-#endif
 
 #include "utils/Exception.h"
 #include "velox/connectors/hive/HiveConfig.h"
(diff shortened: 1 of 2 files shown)


PHILO-HE:
@pratham76

majetideepak:
@PHILO-HE, @zhouyuan I need your help to plumb the GCS configs. Velox is not seeing the GCS configs.
I see the GCS configs appear in the `VeloxRuntime session config`. Should these appear in the `VeloxBackend config` instead?
Should I be adding them in `GlutenConfig.scala`?

zhouyuan:
@majetideepak can you please have a quick try following S3 configuration? I assume it may need to be fetched from scala side
https://github.com/apache/incubator-gluten/blob/main/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala#L739

majetideepak:
@pratham76 Can you verify this fix again? It worked for me locally.

pratham76:
> @pratham76 Can you verify this fix again? It worked for me locally.

Sure will verify and update

zhouyuan (approved):
👍

pratham76:
Have verified this change, GCS reads working. Thanks

PHILO-HE (approved):
@majetideepak, thanks for your fix!

PR #7442, opened 2024-10-09
[VL] Adapt setup-centos8.sh to latest velox helper functions

liujiayi771:

https://github.com/facebookincubator/velox/pull/10920 modified the `wget_and_untar` to download the source code to the `deps-download` directory by default, and introduced a new method `cmake_install_dir` to adapt this modification. `setup-centos8.sh` will source the `scripts/setup-helper-functions.sh` in `VELOX_HOME`, so we need to adapt this modification. 

## How was this patch tested?

Test offline.


liujiayi771:
`setup-centos7.sh` has its own `wget_and_untar`.

PHILO-HE (approved):
@liujiayi771, thanks for your fix! This pr can fix the failure of a scheduled job on cento-8: https://github.com/apache/incubator-gluten/actions/runs/11204762734/job/31143531480. We also need some fix on other OSes. I will create another pr.

Yohahaha (approved):
👍

PR #7444, opened 2024-10-09
Minor fix for info.sh

PHILO-HE:

This script is used by issue reporter to present environment info.
Just some minor fix here: use Gluten version instead of velox version; remove conda info as conda is not supported by Gluten. 

## How was this patch tested?

N/A.


PHILO-HE:
@JkSelf, could you take a look?

PR #7451, opened 2024-10-09
[GLUTEN-7450][VL] Improve CollectRewriteRule for Velox

beliefer:

This PR proposes to improve `CollectRewriteRule` for Velox. These optimal points show below.
1. Short circuiting of `CollectRewriteRule` in advance.
2. Use tree traversal with pruning instead of full traversal.
3. Remove ensureNonNull since `CollectSet` override the `defaultResult`.
4. Unify the replacement for `CollectSet` and `CollectList`.
5. Use more meaningful variable names. e.g. `coalesce`.
6. Use correct variable names. e.g. `aggExpr`.
7. Other simplifications.

## How was this patch tested?

integration tests

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
@@ -21,12 +21,11 @@ import org.apache.gluten.expression.aggregate.{VeloxCollectList, VeloxCollectSet
 import org.apache.gluten.utils.LogicalPlanSelector
 
 import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.catalyst.expressions.{And, Coalesce, Expression, IsNotNull, Literal, WindowExpression}
+import org.apache.spark.sql.catalyst.expressions.{And, Expression, IsNotNull, WindowExpression}
 import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, LogicalPlan, Window}
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.catalyst.trees.TreePattern.{AGGREGATE_EXPRESSION, WINDOW_EXPRESSION}
-import org.apache.spark.sql.types.ArrayType
 
 import scala.reflect.{classTag, ClassTag}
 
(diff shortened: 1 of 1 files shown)


beliefer:
ping @zhztheplayer @zhouyuan @Yohahaha

zhztheplayer (approved):
LGTM

backends-velox/src/main/scala/org/apache/gluten/extension/CollectRewriteRule.scala
```diff
@@ -37,70 +37,58 @@ import scala.reflect.{classTag, ClassTag}
...
 }
 
 object CollectRewriteRule {
-  private def ensureNonNull(expr: Expression): Expression = {
-    val out =
-      Coalesce(List(expr, Literal.create(Seq.empty, expr.dataType)))
-    assert(!out.nullable)
-    assert(!out.dataType.asInstanceOf[ArrayType].containsNull)
-    out
-  }
```
zhztheplayer, line 89: Curious what's the reason of removing this? I remember it was for the compatibility with vanilla Spark.

beliefer: Because `VeloxCollect` overrides the defaultResult with the default value. The Spark optimizer will ensure that it is not null.

zhztheplayer: > Because VeloxCollect overrides the `defaultResult` with the default value. The Spark optimizer will ensure that it is not null.

Didn't see this method is used much by Spark in regular aggregation planning routine: https://github.com/search?q=repo%3Aapache%2Fspark%20defaultResult&type=code. Do you see the `Coalesce` in plan after removing this logic anyway?

BTW as Rui's Velox PR was merged recently https://github.com/facebookincubator/velox/pull/10737, @rui-mo did you make any corresponding changes in Gluten at that time?

beliefer: The definition of defaultResult in AggregateFunction show below.

```
  /**
   * Result of the aggregate function when the input is empty.
   */
  def defaultResult: Option[Literal] = None
```

The below code is the key logic path.
```
        // Some aggregate functions (COUNT) have the special property that they can return a
        // non-null result without any input. We need to make sure we return a result in this case.
        val resultWithDefault = af.defaultResult match {
          case Some(lit) => Coalesce(Seq(result, lit))
          case None => result
        }
```

beliefer:
@zhztheplayer Thank you!

PR #7457, opened 2024-10-10
[Core] Fix duplicated column names in DS test

FelixYBW:
with the fix, we can write the result to parquet file in the test, like add SQL before each query:

```
INSERT OVERWRITE DIRECTORY '{write_path}/{qid}' using parquet
select ...
```


marin-ma (approved):
Thanks!

PR #7474, opened 2024-10-11
[VL][CI] Use cached Velox build binary to accelerate GHA workflow

PHILO-HE:

In the current main code, `libgluten.so` and `libvelox.so` are cached. If the cache is hit (depending on the hashing result on some directories including `gluten/cpp`), they can be re-used to accelerate workflow. However, any code change for `gluten/cpp` in a pr will make the cache missed. Then, needs to re-build Velox code, which is quite time-consuming.

This pr makes Velox build binary cached. If cache is hit, directly let the workflow build Gluten CPP, which only requires few minutes. Otherwise, re-build Velox code.

For Gluten CI's static build that build_test/benchmark are not enabled, we only need to cache `libvelox.a`. For Gluten CI's dynamic build that build_test/benchamrk are enabled, building Gluten cpp needs some other velox libs, like `libdbgen.a`, and one extra generated header, so we just cache the whole velox build path.

## How was this patch tested?

CI.

### DIFF
--- a/.github/workflows/velox_backend.yml
+++ b/.github/workflows/velox_backend.yml
@@ -82,7 +82,7 @@ jobs:
         if: ${{ steps.cache.outputs.cache-hit != 'true' }}
         run: |
           df -a
-          cd $GITHUB_WORKSPACE/
+          rm -rf ./ep/build-velox/build/velox_ep
           bash dev/ci-velox-buildstatic-centos-7.sh
       - uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/velox_backend.yml
```diff
@@ -56,19 +56,29 @@ jobs:
...
     runs-on: ubuntu-20.04
     container: apache/gluten:vcpkg-centos-7
+    env:
+      VELOX_BUILD_PATH: "./ep/build-velox/build/velox_ep/_build/release/"
     steps:
       - uses: actions/checkout@v2
       - name: Generate cache key
         run: |
-          echo ${{ hashFiles('./ep/build-velox/src/**', './dev/**', './cpp/*', './.github/workflows/*') }} > cache-key
+          echo ${{ hashFiles('./ep/build-velox/src/**', './dev/**', './.github/workflows/velox_backend.yml') }} > cache-key
```
zhztheplayer, line 65: > ./ep/build-velox/src/**

I think we could change our OAP Velox dependency from branches to tags while landing this PR. Branches can be updated without invalidating the new cache.

PHILO-HE: @zhztheplayer, yes, this is a potential issue. But before a velox rebase branch is finalized, we may still need oap/velox to hold it as a branch that allows updating (e.g., needs some fixes due to Gluten CI failure). So maybe, we should just have a internal convention to forbid updating oap/velox branch once it is already referenced by Gluten code.

Issue #7475, opened 2024-10-11
[VL] remove unnecessary trim node in CAST, cuz velox does it


Henry2SS:
### Description

I noticed that gluten added a trim node in CAST, when input type is VARCHAR.
But it is unnecessary, cuz velox does it in `SparkCastHooks::removeWhiteSpaces`.

After remove the trim node, the performance is 1.x times faster.

PR #7476, opened 2024-10-11
[GLUTEN-7475][VL] Add a config to control whether add trim node when CAST from varchar

Henry2SS:

Add a config to control whether a trim node is added in CAST. Because we know that removing the trim node can improve the performance by 1.x times. But we are not sure now whether the results will be consistent with spark after removing it.

(Fixes: \#7475)

## How was this patch tested?

Integration tests passed locally. The performance is 1.x times faster.


PHILO-HE:
@Henry2SS, thanks for your pr! This piece of scala code is for making Gluten/Velox consistent with Spark. Seems the current Velox code doesn't have the consistent white spaces definition? Could you check further?

Henry2SS:
> @Henry2SS, thanks for your pr! This piece of scala code is for making Gluten/Velox consistent with Spark. Seems the current Velox code doesn't have the consistent white spaces definition? Could you check further?

There is a method called `isUnicodeWhiteSpace` in `velox/functions/lib/string/StringImpl.h`. 

And I noticed that, from https://github.com/facebookincubator/velox/pull/7377 this PR, the author proposed different corner cases for Presto and Spark on CAST, including removeWhiteSpaces.

Unfortunately, I don't have much knowledge about this, please confirm.

Henry2SS:
> @Henry2SS, thanks for your pr! This piece of scala code is for making Gluten/Velox consistent with Spark. Seems the current Velox code doesn't have the consistent white spaces definition? Could you check further?

And do we have some unit-tests to test CAST from VARCHAR in CI?
Is it possible to confirm it by all the tests in CI?

xumingming:
> The performance is 1.x times faster.

@Henry2SS Do you mind show the concrete performance improvement?

Henry2SS:
> > The performance is 1.x times faster.
> 
> @Henry2SS Do you mind share the concrete performance improvement?

My pleasure.

There is a benchmark testing CAST VARCHAR to BIGINT. The sql is like `INSERT INTO tmp_table as select CAST(string_col as BIGINT) from table`. And the table has 2,879,966,589 rows.

We used gperftool to find that there is a function that takes up the most of the execution time in benchmark.

Before removing trim node, the total time of `ProjectExecTransformer` is 17.3 min ,dividing into 515 tasks.
After removing trim node, the total time of `ProjectExecTransformer` is 7.5 min ,dividing into 515 tasks.

So it is more than 1.x times faster.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -706,50 +706,6 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
-   */
-  override def genCastWithNewChild(c: Cast): Cast = {
-    // scalastyle:off nonascii
-    // Common whitespace to be trimmed, including: ' ', '\n', '\r', '\f', etc.
-    val trimWhitespaceStr = " \t\n\u000B\u000C\u000D\u001C\u001D\u001E\u001F"
-    // Space separator.
-    val trimSpaceSepStr = "\u1680\u2008\u2009\u200A\u205F\u3000" +
-      ('\u2000' to '\u2006').toList.mkString
-    // Line separator.
-    val trimLineSepStr = "\u2028"
```
jackylee-ch, line 721: Have you checked if they have the same sementic between vanilla Spark and velox?

Yohahaha:
```
val expect = Seq(Row(123))
checkAnswer(sql("SELECT cast('\\t\\n\\u001F 123\\u000B' as int);"), expect)
```

this UT works.

Yohahaha:
We can merge this PR when GHA passed and no blocking comments.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
jackylee-ch: Maybe add a config, default false, for it? A little worried about the result diff if we haven't tested the total case between vanilla spark and velox.

Yohahaha: > Maybe add a config, default false, for it? A little worried about the result diff if we haven't tested the total case between vanilla spark and velox.

@Henry2SS would you?

Henry2SS: > > Maybe add a config, default false, for it? A little worried about the result diff if we haven't tested the total case between vanilla spark and velox.
> 
> @Henry2SS would you?

I'll do it later.

jackylee-ch: After looking at the velox code, the conversion from string to int/long is basically similar between [vanilla Spark](https://github.com/apache/spark/blob/master/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java#L200) and [velox](https://github.com/facebookincubator/velox/blob/main/velox/functions/lib/string/StringImpl.h#L354). But for the conversion from string to float/double, [vanilla Spark](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L390) uses `String.trim` to remove characters smaller than spaces, which is not same as velox, so we still need this extra trim.
A new config, which is default false, is good to me. Thanks @Henry2SS for working on it.

PHILO-HE: @Henry2SS, Spark UT doesn't cover many corner cases used in production workload. I'm afraid users would complain result mismatching issues due to this removement. Agree with @jackylee-ch, it's good to use a config to give the control to users. And maybe at some future time point, we can remove it.

Henry2SS: > @Henry2SS, Spark UT doesn't cover many corner cases used in production workload. I'm afraid users would complain result mismatching issues due to this removement. Agree with @jackylee-ch, it's good to use a config to give the control to users. And maybe at some future time point, we can remove it.

Done.
And thanks for your advice. @PHILO-HE @jackylee-ch @Yohahaha

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -730,20 +730,28 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
+              c.withNewChildren(Seq(trimNode)).asInstanceOf[Cast]
+            } else {
+              c
+            }
           case _ =>
             c
         }
       case _ =>
         c.child.dataType match {
           case StringType =>
```
jackylee-ch, line 744:
```suggestion
          case StringType if GlutenConfig.getConf.castFromVarcharAddTrimNode =>
              val trimNode = StringTrim(
                c.child,
                Some(
                  Literal(trimWhitespaceStr +
                    trimSpaceSepStr + trimLineSepStr + trimParaSepStr)))
              c.withNewChildren(Seq(trimNode)).asInstanceOf[Cast]
```

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -730,20 +730,28 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
       case FloatType | DoubleType | _: DecimalType =>
         c.child.dataType match {
           case StringType =>
```
jackylee-ch, line 732:
```suggestion
          case StringType if GlutenConfig.getConf.castFromVarcharAddTrimNode =>
              val trimNode = StringTrim(c.child, Some(Literal(trimSpaceStr)))
              c.withNewChildren(Seq(trimNode)).asInstanceOf[Cast]
```

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
Henry2SS: done

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
Henry2SS: done

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -2125,4 +2127,12 @@ object GlutenConfig {
...
       .booleanConf
       .createWithDefault(true)
+
+  val CAST_FROM_VARCHAR_ADD_TRIM_NODE =
+    buildConf("spark.gluten.velox.castFromVarcharAddTrimNode")
+      .internal()
+      .doc("If true, will add a trim node which has the same sementic as vanilla Spark." +
+        "Otherwise, do nothing.")
+      .booleanConf
+      .createWithDefault(true)
```
Yohahaha, line 2137: I think we can set default value to false, because all GHA has passed and performance boost can not be ignored. If incorrect result issue occurs, we can fix it or make this config more fine-grained. 

Just like scan parquet complex types, we remove force fallback config after all UT passed.

@PHILO-HE @jackylee-ch

jackylee-ch: It's ok to me. And if it's set default to false, we need add this config to `Configuration.md` to make sure users know about it.

Henry2SS: > need add this config to

I'll do this.

Henry2SS: > It's ok to me. And if it's set default to false, we need add this config to `Configuration.md` to make sure users know about it.

I'll do this.

Yohahaha (approved):
👍

jackylee-ch (approved):

PHILO-HE (approved):
Looks good. CI once passed with trim node removed, will merge the pr soon.

PHILO-HE:
@Henry2SS, one code style violation.

```
11:10:48  error file=/home/jenkins/agent/workspace/gluten/gluten-ci/ut-stage-5/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala message=File line length exceeds 100 characters line=2134
```

Yohahaha:
```
12:35:41  [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:check (spotless-check) on project spark-sql-columnar-shims-common: The following files had format violations:
12:35:41  [ERROR]     src/main/scala/org/apache/gluten/GlutenConfig.scala
12:35:41  [ERROR]         @@ -311,7 +311,7 @@
12:35:41  [ERROR]          
12:35:41  [ERROR]          ··def·veloxBloomFilterMaxNumBits:·Long·=·conf.getConf(COLUMNAR_VELOX_BLOOM_FILTER_MAX_NUM_BITS)
12:35:41  [ERROR]          
12:35:41  [ERROR]         -··def·castFromVarcharAddTrimNode·:·Boolean·=·conf.getConf(CAST_FROM_VARCHAR_ADD_TRIM_NODE)
12:35:41  [ERROR]         +··def·castFromVarcharAddTrimNode:·Boolean·=·conf.getConf(CAST_FROM_VARCHAR_ADD_TRIM_NODE)
12:35:41  [ERROR]          
12:35:41  [ERROR]          ··case·class·ResizeRange(min:·Int,·max:·Int)·{
12:35:41  [ERROR]          ····assert(max·>=·min)
12:35:41  [ERROR]         @@ -2131,9 +2131,10 @@
12:35:41  [ERROR]          ··val·CAST_FROM_VARCHAR_ADD_TRIM_NODE·=
12:35:41  [ERROR]          ····buildConf("spark.gluten.velox.castFromVarcharAddTrimNode")
12:35:41  [ERROR]          ······.internal()
12:35:41  [ERROR]         -······.doc("If·true,·will·add·a·trim·node·"·+
12:35:41  [ERROR]         -········"which·has·the·same·sementic·as·vanilla·Spark·to·CAST-from-varchar."·+
12:35:41  [ERROR]         -········"Otherwise,·do·nothing.")
12:35:41  [ERROR]         +······.doc(
12:35:41  [ERROR]         +········"If·true,·will·add·a·trim·node·"·+
12:35:41  [ERROR]         +··········"which·has·the·same·sementic·as·vanilla·Spark·to·CAST-from-varchar."·+
12:35:41  [ERROR]         +··········"Otherwise,·do·nothing.")
12:35:41  [ERROR]          ······.booleanConf
12:35:41  [ERROR]          ······.createWithDefault(false)
12:35:41  [ERROR]          }
12:35:41  [ERROR] Run 'mvn spotless:apply' to fix these violations.
```

CH CI failed due to style check, try use `mvn spotless:apply -P backends-clickhouse -DskipTests` to fix.

Henry2SS:
> ```
> 12:35:41  [ERROR] Run 'mvn spotless:apply' to fix these violations.
> ```
> 
> CH CI failed due to style check, try use `mvn spotless:apply -P backends-clickhouse -DskipTests` to fix.

done

Issue #7291, opened 2024-09-20
[VL][DELTA] MergeInto Command Failed with `DELTA_FILE_TO_OVERWRITE_NOT_FOUND`


dcoliversun:
### Backend

VL (Velox)

### Bug description

Spark SQL:  
```sql
CREATE DATABASE delta_db_0920;

USE delta_db_0920;

CREATE TABLE source_table(id INT, name STRING, age INT) using delta;

    
CREATE TABLE target_table(id INT, name STRING, age INT) using delta;
    
insert into source_table values(1, 'a', 10),(2, 'b', 20);

insert into target_table values(1, 'c', 10),(3, 'c', 30);
    
MERGE INTO target_table AS target
USING source_table AS source
ON target.id = source.id
WHEN MATCHED THEN
UPDATE SET
  target.name = source.name,
  target.age = source.age
WHEN NOT MATCHED THEN
INSERT (id, name, age) VALUES (source.id, source.name, source.age);
```

### Spark version

Spark-3.5.x

### Spark configurations

```shell
bin/spark-shell \
--conf spark.plugins=org.apache.gluten.GlutenPlugin \
--conf spark.memory.offHeap.enabled=true \
--conf spark.memory.offHeap.size=20g \
--conf spark.driver.extraClassPath=/root/gluten/package/target/gluten-velox-bundle-spark3.5_2.12-ubuntu_22.04_x86_64-1.3.0-SNAPSHOT.jar:/root/delta/antlr4-runtime-4.9.3.jar:/root/delta/delta-spark_2.12-3.2.0.jar:/root/delta/delta-storage-3.2.0.jar \
--conf spark.executor.extraClassPath=/root/gluten/package/target/gluten-velox-bundle-spark3.5_2.12-ubuntu_22.04_x86_64-1.3.0-SNAPSHOT.jar:/root/delta/antlr4-runtime-4.9.3.jar:/root/delta/delta-spark_2.12-3.2.0.jar:/root/delta/delta-storage-3.2.0.jar \
--conf spark.shuffle.manager=org.apache.spark.shuffle.sort.ColumnarShuffleManager \
--conf spark.eventLog.enabled=true \
--conf spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension \
--conf spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog \
--conf spark.sql.warehouse.dir=/tmp/warehouse \
[...]

PR #7483, opened 2024-10-11
[GLUTEN-7291][DELTA] fix: push down input_file_name expression to transformer scan in delta

dcoliversun:

This PR aims to push down input_file_name expression to transformer scan in delta.

(Fixes: \#7291)

## How was this patch tested?

Add new unit test.

### DIFF
--- a/gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaRewriteTransformerRules.scala
+++ b/gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaRewriteTransformerRules.scala
@@ -17,21 +17,21 @@
 package org.apache.gluten.extension
 
 import org.apache.gluten.execution.{DeltaScanTransformer, ProjectExecTransformer}
-import org.apache.gluten.extension.DeltaRewriteTransformerRules.columnMappingRule
+import org.apache.gluten.extension.DeltaRewriteTransformerRules.{columnMappingRule, pushDownInputFileExprRule}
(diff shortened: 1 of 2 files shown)


dcoliversun:
@yma11 could you please have a review? thanks

yma11 (approved):

felipepessoto:
@yma11, I didn't get what the unit test is doing, it is not using the input_file_name. Could you help me understand?

yma11:
> @yma11, I didn't get what the unit test is doing, it is not using the input_file_name. Could you help me understand?

@dcoliversun can you help explain? I guess `input_file_name` is kind of from analyzed plan.

dcoliversun:
@yma11 Thanks for ping me and you are right.

@felipepessoto Test sql fails without this commit, you could confirm it with relative issue. Delta needs file name when overwrite and Apache Spark use input_file_name to stash it. But transformer node between project and scan will return an empty string.
https://github.com/apache/incubator-gluten/blob/a07a5de3ae45a5e942824623919a49cb33d1b825/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/PushDownInputFileExpression.scala#L28-L40

felipepessoto:
Got it. So it means input_file_name is now supported by Gluten? I was using a custom Delta version where I replaced input_file_name by _metadata.file_path, I think I can revert my internal change.

dcoliversun:
@felipepessoto yes, `input_file_name` is supported by Apache Gluten. I recommend using this feature. If there are any issues, feel free to open an issue for discussion.

Issue #7397, opened 2024-09-30
[VL] RSS(Resident Set Size) memory of spark executor keep rising


wForget:
### Backend

VL (Velox)

### Bug description

When I was analyzing spark executor being killed due to exceeding yarn memory limit, I found that rss memory of spark executor keeps rising even if spill occurs.

`OffHeapExecutionMemory` and `currently in use` have both decreased, but `ProcessTreeJVMRSSMemory` keeps rising.

![image](https://github.com/user-attachments/assets/d861525e-627e-44eb-8329-7ad01deed263)
![image](https://github.com/user-attachments/assets/1d2e7e78-c373-4545-a72d-a4094ac60098)
![image](https://github.com/user-attachments/assets/2fe46632-1727-4a5f-ae8a-5c02e91eada7)

### Spark version

Spark-3.5.x

FelixYBW:
You may run the same query 50 times and check if RSS keep growing. If yes then it's a new mem leak.

wForget:
> You may run the same query 50 times and check if RSS keep growing. If yes then it's a new mem leak.

Thanks, I will try this

wForget:
After switching to jemalloc (support a new JemallocMemoryAllocator), it seems to be normal. And the sql that failed before can run successfully.

![image](https://github.com/user-attachments/assets/808c4dde-f34c-4b05-b6f7-68c8a6606744)
![image](https://github.com/user-attachments/assets/0d4fc199-b162-4751-95ca-52eb48c0c0d3)
![image](https://github.com/user-attachments/assets/169307e3-d3ab-4933-a6ce-8077428832ad)

FelixYBW:
what allocator do you use previously?

wForget:
> what allocator do you use previously?

I didn't configure it additionally before, is it the default `StdMemoryAllocator`?

https://github.com/apache/incubator-gluten/blob/4b6e929cccd31d0b6312dc4263746583321f022c/cpp/core/memory/MemoryAllocator.cc#L199

FelixYBW:
Yes, by default it's libc. We always preload jemalloc, the test in last year shows jemalloc can boost perf by 10%.

You may also tune the jemalloc parameters and see if it can benefit your real workload. Main tuning is the decay value there, which controls how much memory jemalloc can hold before it returns the memory to system.

wForget:
> We always preload jemalloc, the test in last year shows jemalloc can boost perf by 10%.

[...]

Issue #6571, opened 2024-07-24
[VL] Add an arch subdirectory for native libs package


wForget:
### Description

There are machines with different CPU architectures in our internal hybrid cloud cluster, so I want to add an arch subdirectory for native libs package.

zhztheplayer:
Do you mean to include libraries for different archs into the same bundled Gluten Jar?

wForget:
> Do you mean to include libraries for different archs into the same bundled Gluten Jar?

Yeah, is this reasonable?

zhztheplayer:
Sounds great to me. It's something missing from Gluten for long time.

wForget:
> Sounds great to me. It's something missing from Gluten for long time.

Thank you, I will try to implement this.

PR #7503, opened 2024-10-12
[VL] Support jemalloc memory allocator

wForget:

As commented in https://github.com/apache/incubator-gluten/issues/7397#issuecomment-2408427286, I experienced RSS memory keep growing and it became normal after replacing jemalloc.

By the way, switching to jemalloc in preload way (https://github.com/apache/incubator-gluten/issues/7397#issuecomment-2408452437) also works well, so I am not sure whether this change is necessary.


FelixYBW:
we should use LD_PRELOAD, in this way all memory allocation including the one not tracked by memory pool can be allocated by jemalloc as well.

wForget:
> we should use LD_PRELOAD, in this way all memory allocation including the one not tracked by memory pool can be allocated by jemalloc as well.

Got it, thank you. As described in https://github.com/apache/incubator-gluten/issues/6571#issue-2426589391, we have hybrid cluster with different CPU architectures, so we prefer to package the native libs of multiple environments into one package and select the appropriate lib directory at runtime.

FelixYBW:
> Got it, thank you. As described in [#6571 (comment)](https://github.com/apache/incubator-gluten/issues/6571#issue-2426589391), we have hybrid cluster with different CPU architectures, so we prefer to package the native libs of multiple environments into one package and select the appropriate lib directory at runtime.

It's better to package the libjemalloc.so into jar, then set LD_PRELOAD path to relative address. It can also solve the potential conflicts between the one linked into Gluten and system. @PHILO-HE is it possible?

FelixYBW:
FYI, some more info about jemalloc:
Jemalloc:

- 5.0 changed a lot. Most document online is for 4.0 and before
- Free process: dirty (physical mapped) -> muzzy(MADV_FREE) -> clean (MADV_DONTNEED)
- doesn't release virtual memory, just free physical memory
- dirty_decay_ms and muzzy_decay_ms control when to release page to OS. Default is 10s. 0 means no delay, -1 means never release
- Default 10s is good enough, may increase to 30s
- Default background thread is not useful
- Tcache size is 32k bydefault, no help when increase to 8M
- Madvise(willneed) looks useless to jemalloc allocated memory
- Huge allocation memory isn't hold by jemalloc, it will be released (retained) once free is called. oversize_threshold controls the threshold, default is 8M
- To cache all memory in pool, use param: muzzy_decay_ms:-1,dirty_decay_ms:-1,background_thread:true,oversize_threshold:5368709121
- dirty_decay_ms=10s doesn't mean the memory block wait until 10s, but keep release in the 10s and release all list by end of 10s. So to keep all page in memory pool, you need to config it as -1

image

wForget:
@FelixYBW Thank you very much for your help.

> It's better to package the libjemalloc.so into jar, then set LD_PRELOAD path to relative address. It can also solve the potential conflicts between the one linked into Gluten and system.

Sounds good to me, I will try that.

PHILO-HE:
> @FelixYBW Thank you very much for your help.
> 
> > It's better to package the libjemalloc.so into jar, then set LD_PRELOAD path to relative address. It can also solve the potential conflicts between the one linked into Gluten and system.
> 
> Sounds good to me, I will try that.

@wForget, @FelixYBW, maybe, one way is to pack libjemalloc.so into gluten third-party jar. And let Gluten put this lib in some fixed path (not temp path like other shared lib). Then, set env LD_PRELOAD for spark executor with that fixed lib jemalloc path.

wForget:
Maybe we can package them into a zip file with directories like this:

```
./
└── Linux
    └── x86_64
        └── libjemalloc.so
...
```

Then specify preload jemalloc lib like this:

```
spark.yarn.dist.archives=/path/to/jemallocs.zip
spark.executorEnv.LD_PRELOAD=./jemallocs.zip/$(uname)/$(uname -m)/libjemalloc.so
```

FelixYBW:
> 
> @wForget, @FelixYBW, maybe, one way is to pack libjemalloc.so into gluten third-party jar. And let Gluten put this lib in some fixed path (not temp path like other shared lib) on executor node. Then, configure env LD_PRELOAD for spark executor with that fixed lib jemalloc path.

For static link, we don't have a thirt-party jar.

PR #7516, opened 2024-10-14
[VL] Move unsafe delete operations closer to JNI wrapper

zhztheplayer:
Requires release hooks to be registered when registering factories for `Runtime` and `MemoryManager` instances. So the unsafe `new` and `delete` operations will be organized in the same manner in code.

### DIFF
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -62,16 +62,25 @@ using namespace facebook;
 namespace gluten {
 
 namespace {
-MemoryManager* veloxMemoryManagerFactory(std::unique_ptr<AllocationListener> listener) {
-  return new VeloxMemoryManager(std::move(listener));
+MemoryManager* veloxMemoryManagerFactory(const std::string& kind, std::unique_ptr<AllocationListener> listener) {
+  return new VeloxMemoryManager(kind, std::move(listener));
+}
+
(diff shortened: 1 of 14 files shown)


JkSelf (approved):

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -62,16 +62,25 @@ using namespace facebook;
...
 namespace {
-MemoryManager* veloxMemoryManagerFactory(std::unique_ptr<AllocationListener> listener) {
-  return new VeloxMemoryManager(std::move(listener));
+MemoryManager* veloxMemoryManagerFactory(const std::string& kind, std::unique_ptr<AllocationListener> listener) {
+  return new VeloxMemoryManager(kind, std::move(listener));
+}
+
+void veloxMemoryManagerReleaser(MemoryManager* memoryManager) {
+  delete memoryManager;
 }
```
zhouyuan, line 71: My original thought is we can use some scope ptr for this, since memory manager is `static`

zhztheplayer: > My original thought is we can use some scope ptr for this, since memory manager is `static`

Auto ptr management is not possible for this part of code at least at the moment, since the lifecycle is managed over JNI bridge.

What we can do is to centralize the `new` and `delete` operations to one place as a better design.

PR #7521, opened 2024-10-14
[GLUTEN-6887][VL] switch to upstream Velox

zhouyuan:

This patch changed from OAP/Velox to Meta/Velox + several Gluten patches
 

## How was this patch tested?

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -16,8 +16,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=2024_10_13
+VELOX_REPO=https://github.com/facebookincubator/velox.git
+VELOX_BRANCH=main
 VELOX_HOME=""
 
 OS=`uname -s`
@@ -170,6 +170,7 @@ function apply_compilation_fixes {
   current_dir=$1
   velox_home=$2
   sudo cp ${current_dir}/modify_velox.patch ${velox_home}/
+  sudo cp ${current_dir}/gluten_patch ${velox_home}/ -r
   sudo cp ${current_dir}/modify_arrow.patch ${velox_home}/CMake/resolve_dependency_modules/arrow/
   sudo cp ${current_dir}/modify_arrow_dataset_scan_option.patch ${velox_home}/CMake/resolve_dependency_modules/arrow/
   git add ${velox_home}/modify_velox.patch # to avoid the file from being deleted by git clean -dffx :/
@@ -178,6 +179,8 @@ function apply_compilation_fixes {
   cd ${velox_home}
   echo "Applying patch to Velox source code..."
   git apply modify_velox.patch
+  git apply gluten_patch/*patch
+  git add .
   if [ $? -ne 0 ]; then
     echo "Failed to apply compilation fixes to Velox: $?."
     exit 1
--- a/ep/build-velox/src/gluten_patch/0001-5962-Support-struct-column-reading-with-different-sc.patch
+++ b/ep/build-velox/src/gluten_patch/0001-5962-Support-struct-column-reading-with-different-sc.patch
@@ -1,0 +1,488 @@
+From fe6f7b72e24841c6c9f3fc33b51dfcb03016a543 Mon Sep 17 00:00:00 2001
+From: Jia Ke <ke.a.jia@intel.com>
+Date: Mon, 17 Jun 2024 18:10:48 +0800
+Subject: [PATCH 01/11] [5962 ] Support struct column reading with different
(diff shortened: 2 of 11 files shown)


ep/build-velox/src/get_velox.sh
```diff
@@ -16,8 +16,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=2024_10_13
+VELOX_REPO=https://github.com/facebookincubator/velox.git
+VELOX_BRANCH=main
```
zhouyuan, line 20: should set to a specific commit id

ep/build-velox/src/get_velox.sh
```diff
@@ -178,6 +179,8 @@ function apply_compilation_fixes {
   cd ${velox_home}
   echo "Applying patch to Velox source code..."
   git apply modify_velox.patch
+  git apply gluten_patch/*patch
+  git add .
```
zhztheplayer, line 183: Can we consider using [`git am`](https://git-scm.com/docs/git-am) rather than `git apply` here to help a developer regenerate the patches after a manual rebase?

FelixYBW:
There are several things to do before we switch to upstream:
1. Upstream our oap/velox commit to Velox as much as possible. there are still 11 commits now:
https://github.com/facebookincubator/velox/compare/main...oap-project:velox:update. See if we can ping the author of those PRs.
2. We still need some where to easily track our changes, like the link above. `git apply` isn't good because the batch is hard to maintain.
3. Let's switch to git submodule
4. there are some `sed -i` command in our build-velox.sh. Let's take this opportunity to put them together as other commits.

ep/build-velox/src/get_velox.sh
zhztheplayer: Will not be feeling strongly between the two ways except we can have the patch list synced to a forked Velox repo daily to create a view of repo for the change code. That way we can retain all the required changed into one repo but does add complexity.

zhztheplayer:
> there are some sed -i command in our build-velox.sh. Let's take this opportunity to put them together as other commits.

+1

> Let's switch to git submodule

I heard some voices from other open source projects that git submodule has comparatively steep curve to work with. Perhaps we don't have to make the decision right now and lower the priority of doing that.

zhouyuan:
Hi @FelixYBW
thank you for the inputs, indeed this is just a draft to check how far before can we move to the direction - will keep looking
on this

ep/build-velox/src/get_velox.sh
gaoyangxiaozhu: Agree, unless we found a way to auto sync and generate the patch files, otherwise I can image once this PR merged, developers would suffer painful rebase / patch generation work.

zhouyuan:
> @zhouyuan should we replace the 'sed -i xxx' in https://github.com/apache/incubator-gluten/blob/main/ep/build-velox/src/get_velox.sh as a commit of OAP velox?
> 
> Also the changes in https://github.com/apache/incubator-gluten/blob/main/ep/build-velox/src/modify_velox.patch
> 
> Then let's try to upstream the commits as much as possible. Let's see what's the left patches and define the next step.

Yes, in general the private changes are reduced to decimal avg, parquet struct reader, timestamp support - which requires more time on align on the design/code with Velox community 

The changes in get_velox.sh is not maintained well, some of them are not used any more. BTW it's not used if you are using vcpkg based static build. 

The two changes in "modify_velox.patch" is reduced since last month and the rest changes are not supported by Velox community. We may need more time to discuss with them to align.

-yuan

FelixYBW:
> 
> Yes, in general the private changes are reduced to decimal avg, parquet struct reader, timestamp support - which requires more time on align on the design/code with Velox community
> 

Let's follow up the PRs.

> The changes in get_velox.sh is not maintained well, some of them are not used any more. BTW it's not used if you are using vcpkg based static build.
> 

Good to know. Let's add comments there and don't maintain it. 


> The two changes in "modify_velox.patch" is reduced since last month and the rest changes are not supported by Velox community. We may need more time to discuss with them to align.
> 

If Velox community doesn't accept it, let's put it as a commit in oap/velox instead of a patch to apply.

Yohahaha:
> 3. Let's switch to git submodule

+1 on git submodule, and move velox under cpp folder

Yohahaha:
> > The changes in get_velox.sh is not maintained well, some of them are not used any more. BTW it's not used if you are using vcpkg based static build.
> 
> Good to know. Let's add comments there. @Yohahaha we are going to stop maintaining the dynamic build script. Currently we still have 1 CI to track the dynamic one.
> 

thanks for the reminder, I will keep tracking dynamic build CI.

PR #7527, opened 2024-10-14
[GLUTEN-7526][VL] Scala code style for VeloxCollect

beliefer:

This PR proposes to make `VeloxCollect` follows up the Scala code style.

## How was this patch tested?

integration tests

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
@@ -21,50 +21,55 @@ import org.apache.spark.sql.catalyst.expressions.aggregate.DeclarativeAggregate
 import org.apache.spark.sql.catalyst.trees.UnaryLike
 import org.apache.spark.sql.types.{ArrayType, DataType}
 
-abstract class VeloxCollect extends DeclarativeAggregate with UnaryLike[Expression] {
+abstract class VeloxCollect(child: Expression)
+  extends DeclarativeAggregate
+  with UnaryLike[Expression] {
+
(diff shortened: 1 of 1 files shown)


FelixYBW (approved):
Thank you for your fix!

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
```diff
@@ -21,50 +21,55 @@ import org.apache.spark.sql.catalyst.expressions.aggregate.DeclarativeAggregate
...
+
+  override def prettyName: String = "velox_collect_set"
+
+  override protected def withNewChildInternal(newChild: Expression): Expression =
+    copy(child = newChild)
 }
 
-case class VeloxCollectList(override val child: Expression) extends VeloxCollect {
-  override def prettyName: String = "velox_collect_list"
+case class VeloxCollectList(child: Expression) extends VeloxCollect(child) {
```
zhztheplayer, line 65: Can we keep `override` keyword for `child` in this file? Otherwise the overriding will look more implicit than before. Thanks.

zhztheplayer (commented):
Other parts lgtm. Thanks!

backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
zhztheplayer: Just went through Spark's code and as it's a normal case in Spark, we can follow the style anyway.

beliefer:
@zhztheplayer @FelixYBW Thank you!

PR #7533, opened 2024-10-15
[GLUTEN-7532][CORE] Rewrite TakeOrderedAndProject as actually executed operators when do offload

ivoson:

Rewrite the `TakeOrderedAndProjectExecTransformer` to the operators which are actually executed when do offload. Currently, the logic is wrapped in the `RDD` of the operator. Which means that in the RDD computation we play with the sql operators to generate new `RDD`s which is not a common practice processing sql queries. And the final executed plan and metrics information won't be shown correctly in the Spark UI page.

In this PR, rewrite the node `TakeOrderedAndProjectExecTransformer` when do offload so that the operators can be shown correctly. 

## How was this patch tested?
UT


zhztheplayer:
+1 to the idea. I remember there was some vanilla Spark limitation as blocker of an similar effort. But let's see if it's doable at this time.

ivoson:
cc @zhztheplayer can you please review this PR? Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/planner/metadata/OutputPartitioning.scala
```diff
@@ -0,0 +1,44 @@
...
+
+case class OutputPartitioning(partitioning: Partitioning) {
+  override def hashCode(): Int = System.identityHashCode(partitioning)
+  override def equals(obj: Any): Boolean = obj match {
+    case OutputPartitioning(otherPartitioning) => partitioning.equals(otherPartitioning)
+    case _ => false
+  }
+
+  override def toString: String = s"$partitioning"
+}
```
zhztheplayer, line 32: Speaking of query optimizer, ideally `OutputPartitioning` should be categorized into a physical [`Property`](https://github.com/apache/incubator-gluten/blob/main/gluten-ras/common/src/main/scala/org/apache/gluten/ras/PropertyModel.scala) than into [`Metadata`](https://github.com/apache/incubator-gluten/blob/main/gluten-ras/common/src/main/scala/org/apache/gluten/ras/MetadataModel.scala).

But the implementation can be trickier. Is this inevitable to add this code?

ivoson: > Is this inevitable to add 'OutputPartitioning' for implementing this feature?

I think so. We need the partitioning information to decide whether a shuffle is needed. Otherwise, we can only add the shuffle blindly and then add another rule to remove unnecessary ones afterwards somewhere else.

zhztheplayer: Thank you for the explanation. And I'll do some local tests with the change until this weekend then get back to you.

zhztheplayer: @ivoson There was an old attempt from my side for incorporating `OutputPartitioning` / `OutputOrdering` into RAS planner.

[1] https://github.com/zhztheplayer/gluten/commit/d1575cce3626eaf4efe8c06924d108008e6e2c1e#diff-305f3a48d262389cc311d17f15cd27f6a1efc0d6da2dcd0ebe75975ccf2e9447 

This PR's solution is OK to me as of now, but I think we can do a refactor in future to use properties rather than metadata.

Or I think we can keep the PR open and take some time to look back on `[1]` to see if there are still blockers against the effort. I feel we should make sure the refactor will be viable then it's safe to merge the current PR's code as a temporary solution.

ivoson: > Or I think we can keep the PR open and take some time to look back on [1] to see if there are still blockers against the effort.

Hi @zhztheplayer , do you mean any blockers to incorporate  OutputPartitioning / OutputOrdering into RAS planner? Anything I can help here?

Issue #7499, opened 2024-10-12
[VL][CI] improve CI speed by providing more cache


zhouyuan:
### Description

GHA CI is slow on building Velox code if code changes, we could improve this by providing 
`ccache` or `sscache`

PR #7546, opened 2024-10-15
[GLUTEN-7499][VL][CI] Enable ccache in GHA job

zhouyuan:

This patch enabled ccache in GHA job
- on fully hit, the native build is shorten to ~2 min from 48 min on Centos 7
- on partial hit(velox rebase), the native build is ~30 min and unit tests are passing

By default it will download the `ccache` of latest commit

fixes: #7499

## How was this patch tested?

pass GHA


.github/workflows/velox_backend.yml
```diff
@@ -68,12 +70,24 @@ jobs:
...
+      - name: Get Ccache
+        uses: actions/cache/restore@v3
+        with:
+          path: '${{ env.CCACHE_DIR }}'
+          key: ccache-centos7-release-default
       - name: Build Gluten native libraries
         if: ${{ steps.cache.outputs.cache-hit != 'true' }}
         run: |
           df -a
+          yum install ccache -y
```
zhouyuan, line 82: ccache not installed on centos7 
CC @PHILO-HE

zhouyuan:
@zhztheplayer @PHILO-HE

zhztheplayer (approved):

.github/workflows/velox_backend.yml
PHILO-HE: Yes, will make it installed in another pr. Thanks!

.github/workflows/velox_backend.yml
```diff
@@ -68,12 +70,24 @@ jobs:
...
           cd $GITHUB_WORKSPACE/
           bash dev/ci-velox-buildstatic-centos-7.sh
+      - name: "Save ccache"
+        uses: actions/cache/save@v3
+        id: ccache
+        with:
+          path: '${{ env.CCACHE_DIR }}'
+          key: ccache-centos7-release-default
```
PHILO-HE, line 90: @zhouyuan, do we need to save ccache here? If so, I guess every build triggered for different pr will save and overwrite existing ccache. Maybe, only save ccache in velox_backend_cache.yml on push commit event?

zhztheplayer: > @zhouyuan, do we need to save ccache here? If so, I guess every build triggered for different pr will save and overwrite existing ccache. Maybe, only save ccache in velox_backend_cache.yml on push commit event?

Per GitHub's design PR doesn't access caches created by other PRs. Perhaps storing caches in PR might accelerate subsequent build in the same PR, we may need to test. But I don't feel strongly of doing this too since it might affect build stability among commits in the same PR.

zhouyuan:
it seems it's not updating the ccache, if uploaded with same name
```
/usr/bin/docker exec  a1bb467fffb229f3009d5eec47c48473b0dadac82fa801780e8141db2abf9ebb sh -c "cat /etc/*release | grep ^ID"
/usr/bin/tar --posix -cf cache.tgz --exclude cache.tgz -P -C /__w/incubator-gluten/incubator-gluten --files-from manifest.txt -z
Failed to save: Unable to reserve cache with key ccache-centos8-release-default, another job may be creating this cache. More details: Cache already exists. Scope: refs/pull/7546/merge, Key: ccache-centos8-release-default, Version: 6b8d00d82c8f69c7dcde172f22429edadfab05ef3d01c77520d3df20b620247d
Warning: Cache save failed.
```

PHILO-HE (approved):
I see. Thanks!

zhouyuan:
Notes on GH cache management:

`The cache action first searches for cache hits for key and the cache version in the branch containing the workflow run. If there is no hit, it searches for restore-keys and the version. If there are still no hits in the current branch, the cache action retries same steps on the default branch. Please note that the scope restrictions apply during the search. `


https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows

PHILO-HE (approved):

Issue #6896, opened 2024-08-16
[VL] Umbrella issue for sort-based shuffle


marin-ma:
### Description

- Initial patch: https://github.com/apache/incubator-gluten/pull/6475 (TPCDS SF3T 40000 partitions 1.13x vs hash)
  - follow up: https://github.com/apache/incubator-gluten/pull/6579
  - follow up: https://github.com/apache/incubator-gluten/pull/6628
- [Performance] Reduce shuffle spill: https://github.com/apache/incubator-gluten/pull/6639 (1.25x)
- [Performance] Reduce shuffle writer memory waste: https://github.com/apache/incubator-gluten/pull/6727 (1.3x)
- [UI] display shuffle_writer_type: https://github.com/apache/incubator-gluten/pull/6799
- [Performance] Optimize c2r in shuffle. https://github.com/apache/incubator-gluten/pull/6782 (Pending on Velox PR)
- [Bug fix] #6822 https://github.com/apache/incubator-gluten/pull/6832
- [Bug fix] #6863 https://github.com/apache/incubator-gluten/pull/6869

FelixYBW:
Good summary! thank you

PR #7553, opened 2024-10-16
[VL] Fix sort based shuffle oom in spill when compress was disabled

clay4megtr:

Fix oom when compress was disabled in spill.

image

### DIFF
--- a/cpp/core/shuffle/rss/RssPartitionWriter.cc
+++ b/cpp/core/shuffle/rss/RssPartitionWriter.cc
@@ -85,7 +85,7 @@ arrow::Status RssPartitionWriter::doEvict(
       inMemoryPayload->toBlockPayload(
           payloadType, payloadPool_.get(), codec_ ? codec_.get() : nullptr, std::move(compressed)));
   // Copy payload to arrow buffered os.
-  ARROW_ASSIGN_OR_RAISE(auto rssBufferOs, arrow::io::BufferOutputStream::Create(options_.pushBufferMaxSize, pool_));
+  ARROW_ASSIGN_OR_RAISE(auto rssBufferOs, arrow::io::BufferOutputStream::Create(options_.pushBufferMaxSize));
   RETURN_NOT_OK(payload->serialize(rssBufferOs.get()));
   payload = nullptr; // Invalidate payload immediately.
 


cpp/core/shuffle/rss/RssPartitionWriter.cc
```diff
@@ -85,7 +85,7 @@ arrow::Status RssPartitionWriter::doEvict(
       inMemoryPayload->toBlockPayload(
           payloadType, payloadPool_.get(), codec_ ? codec_.get() : nullptr, std::move(compressed)));
   // Copy payload to arrow buffered os.
-  ARROW_ASSIGN_OR_RAISE(auto rssBufferOs, arrow::io::BufferOutputStream::Create(options_.pushBufferMaxSize, pool_));
+  ARROW_ASSIGN_OR_RAISE(auto rssBufferOs, arrow::io::BufferOutputStream::Create(options_.pushBufferMaxSize));
```
zhztheplayer, line 88: Thanks for working on a fix. Though this could turn the memory buffer into untracked? 

cc @marin-ma

marin-ma:
Thanks. Can we pre-allocate a buffer of size `options_.pushBufferMaxSize`, and make rssBufferOs reuse the memory each time?

cpp/core/shuffle/rss/RssPartitionWriter.cc
clay4megtr: It is, actually I'm a little confused about delegate the timing of spill trigger to spark framework, which means we can't make any memory allocation when spilling.

clay4megtr: Maybe we should have a threshold to indicate whether trigger spill actively or not?

clay4megtr:
> Thanks. Can we pre-allocate a buffer of size `options_.pushBufferMaxSize`, and make rssBufferOs reuse the memory each time?

That's a better solution, I'll update later. btw, I have another question, is it possible that `sortTime` which defined [here](https://github.com/apache/incubator-gluten/blob/66bb774f33e7419996bd052ae2ba9cbafcf9a489/gluten-celeborn/velox/src/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala#L99) include cost time of other spill type? because the trigger root is `nativeShuffleWriter.write(...)`  ?

zhouyuan:
related 
https://github.com/apache/incubator-gluten/issues/6896

marin-ma:
> Thanks. Can we pre-allocate a buffer of size `options_.pushBufferMaxSize`, and make rssBufferOs reuse the memory each time?

@clay4megtr Let's merge this one first. Perhaps consider this change in the future. Thanks!

marin-ma (approved):

PR #7574, opened 2024-10-17
[GLUTEN-7359][VL] Enable partial project in RAS

zhztheplayer:
Following #7360, enable the feature in RAS.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
@@ -157,7 +157,7 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
       return ValidationResult.failed("Contains expression not supported")
     }
     if (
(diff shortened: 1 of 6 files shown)


zhztheplayer:
cc @jinchengchenghh

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -157,7 +157,7 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
       return ValidationResult.failed("Contains expression not supported")
     }
     if (
-      ExpressionUtils.isComplexExpression(
+      ExpressionUtils.hasComplextExpressions(
```
jinchengchenghh, line 160: a typo hasComplextExpressions -> hasComplexExpressions

zhztheplayer: Thanks for catching.

jinchengchenghh (approved):
Thanks!

PR #7578, opened 2024-10-17
[GLUTEN-7458][VL] Upgrade to GCC-11 for centos-7/8 and ubuntu-20.04

PHILO-HE:

Since the recently upgraded folly in Velox requires GCC >=10, we need to upgrade it for OSes whose default gcc package doesn't meet this requirement, such as centos-7/8 and ubuntu-20.04.

Also remove build tool setup for ubuntu-18.04 from `setup-build-depends.sh`, as it is not officially supported now.

## How was this patch tested?

CI.

### DIFF
--- a/dev/vcpkg/setup-build-depends.sh
+++ b/dev/vcpkg/setup-build-depends.sh
@@ -160,11 +160,10 @@ install_alinux_3() {
 }
 
 install_tencentos_3.2() {
-    # FIXME: upgrade to gcc-11.
     yum -y install \
         wget curl tar zip unzip git which \
         cmake ninja-build perl-IPC-Cmd autoconf autoconf-archive automake libtool \
-        gcc-toolset-9-gcc gcc-toolset-9-gcc-c++ \
+        gcc-toolset-11 \
         flex bison python3 \
         java-8-konajdk
 
--- a/dev/vcpkg/setup-build-depends.sh
+++ b/dev/vcpkg/setup-build-depends.sh
@@ -26,10 +26,10 @@ install_maven_from_source() {
     fi
 }
 
-install_gcc9_from_source() {
+install_gcc11_from_source() {
      cur_gcc_version=$(gcc -dumpversion)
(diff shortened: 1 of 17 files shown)


PHILO-HE:
> @PHILO-HE should we upgrade the docker image in a separate PR, and tag it with gcc-11? otherwise it may break some old tests

@zhouyuan, maybe, we can directly install upgraded GCC in this pr with the old docker images used. After docker images are updated with new gcc installed, we can remove unnecessary installation from yml.

Regarding updating docker image, I think we should keep both gcc-9/11 installed. Then, use required gcc version by sourcing corresponding SCL devtoolset. This is for considering branch-1.2 GHA workflow, where gcc-9 is still used.

zhouyuan:
> > @PHILO-HE should we upgrade the docker image in a separate PR, and tag it with gcc-11? otherwise it may break some old tests
> 
> @zhouyuan, maybe, we can directly install upgraded GCC in this pr with the old docker images used. After docker images are updated with new gcc installed, we can remove unnecessary installation from yml.
> 
> Regarding updating docker image, I think we should keep both gcc-9/11 installed. Then, use required gcc version by sourcing corresponding SCL devtoolset. This is for considering branch-1.2 GHA workflow, where gcc-9 is still used.

Yes, that's the issue. We will need to modify the branch-1.2 also if image is updated with same name

dev/vcpkg/setup-build-depends.sh
```diff
@@ -135,60 +134,18 @@ install_centos_8() {
     install_maven_from_source
 }
 
-install_ubuntu_18.04() {
```
zhouyuan, line 138: CC @zhli1142015

PHILO-HE:
@Yohahaha, do you have any comment?

Yohahaha:
> @Yohahaha, do you have any comment?

no comments, feel free to merge this PR.

rui-mo (approved):

wForget:
The CI https://github.com/apache/incubator-gluten/actions/runs/11499412404/job/32007324721?pr=7648 for branch 1.2 seems to be failing due to this change, do we have a solution?

PR #7590, opened 2024-10-18
[VL] Remove unnecessary vanilla Spark compatibility code for VeloxCollectSet function

zhztheplayer:
Following https://github.com/apache/incubator-gluten/pull/7451

Since https://github.com/facebookincubator/velox/pull/10737 was merged to Velox, the previous code for aligning Velox's `collect_set` with vanilla Spark's semantic in query planner is not needed now. Removing them.


zhztheplayer:
@rui-mo @beliefer

PR #7592, opened 2024-10-18
[GLUTEN-7359][VL] Optimize string in partial project

jinchengchenghh:
Remove velox to row and then UnsafeProjection, then row to velox.
Introduce ArrowProjection to support write string and binary to Arrow directly.
MutableProjection writes result to buffer and then write by writer for thread safe for Aggregation, change to write to Arrow writer directly and remove setNotNull before write a value.


gluten-arrow/src/main/scala/org/apache/gluten/expression/InterpretedColumnarProjection.scala
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.spark.sql.types.{BinaryType, DataType, StringType}
+import org.apache.spark.unsafe.types.UTF8String
+
+/**
+ * A [[ArrowProjection]] that is calculated by calling `eval` on each of the specified expressions.
+ *
+ * @param expressions
+ *   a sequence of expressions that determine the value of each column of the output row.
+ */
+class InterpretedColumnarProjection(expressions: Seq[Expression]) extends ArrowProjection {
```
zhztheplayer, line 35: `InterpretedColumnarProjection` vs `InterpretedArrowProjection` ?

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -214,60 +211,12 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
...
   private def getProjectedBatchArrow(
       childData: ColumnarBatch,
       c2a: SQLMetric,
       a2c: SQLMetric): Iterator[ColumnarBatch] = {
     // select part of child output and child data
-    val proj = MutableProjection.create(replacedAliasUdf, projectAttributes.toSeq)
+    val proj = ArrowProjection.create(replacedAliasUdf, projectAttributes.toSeq)
     val numRows = childData.numRows()
     val start = System.currentTimeMillis()
     val arrowBatch = if (childData.numCols() == 0 || ColumnarBatches.isHeavyBatch(childData)) {
```
zhztheplayer, line 222: Guessing dong a `checkOffloaded` check is enough, as the input of this operator is guaranteed to be in Velox format

gluten-arrow/src/main/scala/org/apache/gluten/expression/ArrowProjection.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def create(expr: Expression): ArrowProjection = create(Seq(expr))
+
+  /**
+   * Returns an ArrowProjection for given sequence of Expressions, which will be bound to
+   * `inputSchema`.
+   */
+  def create(exprs: Seq[Expression], inputSchema: Seq[Attribute]): ArrowProjection = {
+    create(bindReferences(exprs, inputSchema))
+  }
```
zhztheplayer, line 66: Didn't see all the APIs are used so far. Shall we have chance to use them in future?

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -183,9 +181,8 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
             } else {
               val start = System.currentTimeMillis()
               val childData = ColumnarBatches.select(batch, projectIndexInChild.toArray)
```
zhztheplayer, line 183: Is there possibility that `projectIndexInChild` includes duplicated indexes? If yes can we have a test case for that? Thanks.

zhztheplayer (commented):
Most of the comments are minors. Thanks.

BTW I see there is a `return` at line 120 here:

https://github.com/apache/incubator-gluten/blob/0f82fd0b9d0189534b8fa5b006e56d0e964114e2/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala#L117-L127

Would you like to reason about it a bit? I think we should avoid this kind of return statements in lambda. Let's use `Expression.exists` for fast-failing if needed.

zhztheplayer:
Would you rebase also? Thanks.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
jinchengchenghh: No, it won't be duplicated, I have checked exists status when I append to it.

gluten-arrow/src/main/scala/org/apache/gluten/expression/ArrowProjection.scala
jinchengchenghh: Yes, I copied from MutableProjection, maybe it will be used in future.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
jinchengchenghh: Thanks for catching it.

gluten-arrow/src/main/scala/org/apache/gluten/expression/InterpretedColumnarProjection.scala
jinchengchenghh: Yes, ArrowProjection corresponds to  `InterpretedArrowProjection`, we may do a small refactor to add `ColumnarRow` interface to support write to string and binary and rename to `ColumnarProjection` and `InterpretedColumnarProjection` after we have Velox Vector Java API.

shims/spark32/src/main/scala/org/apache/spark/sql/catalyst/expressions/ExpressionsEvaluatory.scala
```diff
@@ -0,0 +1,50 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.catalyst.expressions
+
+import org.apache.spark.sql.internal.SQLConf
+
+// A helper class to evaluate expressions.
+trait ExpressionsEvaluator {
```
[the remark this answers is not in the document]
jinchengchenghh: yes

jinchengchenghh:
> Most of the comments are minors. Thanks.
> 
> BTW I see there is a `return` at line 120 here:
> 
> https://github.com/apache/incubator-gluten/blob/0f82fd0b9d0189534b8fa5b006e56d0e964114e2/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala#L117-L127
> 
> Would you like to reason about it a bit? I think we should avoid this kind of return statements in lambda. Let's use `Expression.exists` for fast-failing if needed.

Thanks for catching it, I change to `forall` to end up fast.

zhztheplayer (approved):
:+1

Yohahaha:
@jinchengchenghh would you support array type in ArrayColumnarRow?

jinchengchenghh:
No, we have no requirement for it now. @Yohahaha

Issue #7589, opened 2024-10-18
[VL] Support date_trunc function


zml1206:
### Description

Velox already supports date_trunc of presto, which seems to be reusable for spark.

PR #7611, opened 2024-10-19
[GLUTEN-7589][VL] Support date_trunc function

zml1206:

Fixes: \#7589

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
@@ -1727,4 +1727,32 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
         }
     }
   }
+
+  test("date_trunc") {
+    withTable("t") {
+      sql("create table t (c0 TIMESTAMP) using parquet")
+      sql("insert into t values(Timestamp('2015-07-22 10:01:40.123456'))")
+      runQueryAndCompare("""
+                           |SELECT
+                           |  date_trunc('yy',c0) as t1,
+                           |  date_trunc('yyyy', c0) as t2,
+                           |  date_trunc('year', c0) as t3,
+                           |  date_trunc('quarter', c0) as t4,
+                           |  date_trunc('mon', c0) as t5,
+                           |  date_trunc('month', c0) as t6,
+                           |  date_trunc('mm', c0) as t7,
+                           |  date_trunc('week', c0) as t8,
+                           |  date_trunc('dd', c0) as t9,
+                           |  date_trunc('day', c0) as t10,
+                           |  date_trunc('hour', c0) as t11,
+                           |  date_trunc('minute', c0) as t12,
+                           |  date_trunc('second', c0) as t13,
+                           |  date_trunc('millisecond', c0) as t14,
+                           |  date_trunc('microsecond', c0) as t15
+                           |FROM t
+                           |""".stripMargin) {
+        checkGlutenOperatorMatch[ProjectExecTransformer]
+      }
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


ayushi-agarwal:
@zml1206 Do you plan to reopen this to support date_trunc function in gluten?

zml1206:
> @zml1206 Do you plan to reopen this to support date_trunc function in gluten?


After this PR(https://github.com/facebookincubator/velox/pull/11340) is merged, I will open a new PR to support it.

backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
```diff
@@ -1438,4 +1438,37 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
         checkGlutenOperatorMatch[FilterExecTransformer](df)
     }
   }
+
+  test("date_trunc") {
```
ayushi-agarwal, line 1442: @zml1206 I was running part of this test by using your velox changes and it was giving incorrect result for year
Output for this 2015-07-22 10:01:40.123456 was coming to be 2014-12-31 16:00:00.0 instead of 2015-01-01 00:00:00.0
It may be due to timezone conversion, could you suggest if I am missing something?

zml1206: withSQLConf(SQLConf.SESSION_LOCAL_TIMEZONE.key -> "UTC")

ayushi-agarwal: I had tried setting this and it gives correct results but I was confused that for a real workload also do we need to set it? Is setting of this config not needed if we pass timezone also to this function? 
From velox side I see that in call method of this function, timeZone_ comes out to be a nullptr, but gluten passes spark.sql.session.timeZone to velox as query config so I am confused as to why it comes to be a nullptr? Could you please share light on it as I am new to the timestamp part.

zml1206: https://github.com/zml1206/velox/blob/5fb41e85b66eaac7063c92ebc0643d8dc48722aa/velox/functions/sparksql/DateTimeFunctions.h#L547

ayushi-agarwal: Here it comes to be nullptr when I added a logline to prine timezone, that's what caused the confusion.

zml1206: image

ayushi-agarwal: Thanks, then it seems some issue in my environment due to which it comes out to be null.

ayushi-agarwal: For me initialize function was never getting called when I was running test from gluten, after changing the definition of initialize to this it is working fine.
FOLLY_ALWAYS_INLINE void initialize(
      const std::vector<TypePtr>& /*inputTypes*/,
      const core::QueryConfig& config,
      const arg_type<Varchar>* /*unitString*/,
      const arg_type<Timestamp>* /*timestamp*/) {
    std::cout << "Initializing DateTruncFunction" << std::endl;
    timeZone_ = getTimeZoneFromConfig(config);
  }

zml1206: Oh, my mistake, thank you.

zhouyuan:
hi @zml1206 
`date_trunc` is merged in velox, would you re-open this patch? 
https://github.com/apache/incubator-gluten/pull/9213

thanks, -yuan

zhouyuan (approved):
Thanks.

Issue #7621, opened 2024-10-21
[CH]The repeat function reports an error when times is a negative number


loneylee:
### Backend

CH (ClickHouse)

### Bug description

```
select repeat('1', -1);
```

### Spark version

None

PR #7622, opened 2024-10-21
[GLUTEN-7621][CH] Fix repeat function reports an error when times is a negative number

loneylee:

(Fixes: \#7621)

## How was this patch tested?

Test by ut


zzcclp (approved):
LGTM

Issue #7623, opened 2024-10-21
[CH] Running cache command with error when executor add and removed.


loneylee:
### Backend

CH (ClickHouse)

### Bug description

No error. 

### Spark version

None

PR #7625, opened 2024-10-21
[GLUTEN-7623][CH] Fix running cache command with error when executor add and removed.

loneylee:

(Fixes: \#7623)

## How was this patch tested?

Test by Manual


baibaichen (approved):
LGTM

PR #7629, opened 2024-10-22
[VL] Fix missing VELOX_HOME in builddeps-veloxbe.sh

liujiayi771:

`VELOX_HOME` should be passed as a parameter to the `build_velox.sh` script in builddeps-veloxbe.sh.

## How was this patch tested?

N/A


liujiayi771:
cc @PHILO-HE, a minor fix, thanks.

zhztheplayer (approved):

guixiaowen:
@liujiayi771  May i ask you a question？
In build_velox.sh, VELOX_HOME will be initialized if it is not passed in. And it is the same as the path passed in builddepts-veloxbe.sh. Is there anything I misunderstood? Can you give me some guidance？

In build_velox.sh, line 208-210

if [ "$VELOX_HOME" == "" ]; then
  VELOX_HOME="$CURRENT_DIR/../build/velox_ep"
fi

liujiayi771:
@guixiaowen If we want to specify `VELOX_HOME` to a custom path, we need to pass this parameter to `build_velox.sh`. It's not an issue for the default path.

guixiaowen:
> @guixiaowen If we want to specify `VELOX_HOME` to a custom path, we need to pass this parameter to `build_velox.sh`. It's not an issue for the default path.

@liujiayi771  Thank you for your reply. I understand now.

PR #7650, opened 2024-10-23
[BUILD][DOCS][VL] Change BUILD_JEMALLOC to ENABLE_JEMALLOC_STATS

surnaik:

This PR changes the Cmake option BUILD_JEMALLOC to ENABLE_JEMALLOC_STATS

Currently, the option is confusing, when enabled, it builds jemalloc and then sets the `ENABLE_JEMALLOC` definition which prints the stats. The only reason to build jemalloc here is to print the memory stats for debugging.

This PR changes the name to signify the intent of that option better. It's imperatively understood that enabling this stats will require building of jemalloc, hence we don't need an explicit BUILD_JEMALLOC and ENABLE_JEMALLOC option separately.

## How was this patch tested?

Manually built and tested that the stats flow into logs.


PHILO-HE:
@surnaik, this change looks good to me. Could you update this doc: https://github.com/apache/incubator-gluten/blob/main/docs/get-started/build-guide.md?

PHILO-HE (approved):
In the future, if we extend to use more jemalloc-specific functions, we can change to use some other option name. For now, this change is good.

FelixYBW:
It's related to this PR actually:
https://github.com/apache/incubator-gluten/pull/7369

In future we should have a way to always build jemalloc.so, then set to LD_PRELOAD so we can add jemalloc stat info in Velox and Gluten.

Let's rename it temporarily.

PR #7651, opened 2024-10-23
[VL] Code cleanup for Arrow CSV UTs

zhztheplayer:
1. Add individual suite `ArrowCsvScanSuite` for Arrow CSV tests
2. Centralize table creation
3. Cover more cases with DS V2
4. Other essential refactors


zhztheplayer:
cc @jinchengchenghh

marin-ma (approved):
Thanks!

jinchengchenghh:
Thanks!
The ORC scan is also in TestOperator, do we also need to refactor it? @zhztheplayer

PR #7672, opened 2024-10-24
[VL] Rework the implementation of spark.gluten.enabled

zhztheplayer:
1. Remove `QueryPlanSelector` and the explicit calls to this utility in code
2. Remove `SkipCondition`
3. Add `InjectorControl` for building wrappers for the injected entities that should be dynamically controlled by `spark.gluten.enabled` / `gluten.enabledForCurrentThread`

After the change is applied, all injections to `SparkSessionExtensions` in Gluten through `RuleInjector`, including rules and parsers will automatically obey option `spark.gluten.enabled` and local property `gluten.enabledForCurrentThread` without extra efforts. This will ease the further development work a bit.


rui-mo (approved):

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -469,7 +469,7 @@ object GlutenConfig {
   import SQLConf._
 
   var GLUTEN_ENABLE_BY_DEFAULT = true
```
zhztheplayer, line 471: s/GLUTEN_ENABLE_BY_DEFAULT/GLUTEN_ENABLED_BY_DEFAULT/

gluten-core/src/main/scala/org/apache/gluten/extension/injector/InjectorControl.scala
```diff
@@ -0,0 +1,160 @@
...
+            new InvocationHandler {
+              override def invoke(proxy: Any, method: Method, args: Array[AnyRef]): AnyRef = {
+                try {
+                  if (disabler.disabled(session)) {
+                    return method.invoke(before, args: _*)
+                  }
+                  method.invoke(after, args: _*)
+                } catch {
+                  case e: InvocationTargetException =>
+                    // Unwrap the UTE.
```
zhztheplayer, line 101: s/UTE/ITE/

PR #7677, opened 2024-10-24
[GLUTEN-7458][VL][FOLLOWUP] Install devtoolset-9 for centos7 static build image

wForget:

Install devtoolset-9 for centos7 static build image to support branch 1.2 GHA.

### DIFF
--- a/dev/docker/Dockerfile.centos7-static-build
+++ b/dev/docker/Dockerfile.centos7-static-build
@@ -13,6 +13,9 @@ RUN sed -i \
 
 RUN yum install -y git patch wget sudo java-1.8.0-openjdk-devel ccache
 
+# TODO: Remove this after branch 1.2 EOL, just for branch 1.2 GHA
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@wForget, thanks for your pr!
Another thought is we can install gcc-9 in [velox_backend.yml](https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_backend.yml) only for branch-1.2. Thus, we don't need to maintain such code in main branch when branch-1.2 is EOL in the future.

wForget:
> @wForget, thanks for your pr! Another thought is we can install gcc-9 in [velox_backend.yml](https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_backend.yml) only for branch-1.2. Thus, we don't need to maintain such code in main branch when branch-1.2 is EOL in the future.

Makes sense, thank you, I will try that.

Issue #7685, opened 2024-10-25
[VL][RAS] Optimize rough model to solve the regression caused by r2c large table


zml1206:
### Description

In production, we encountered a situation where the consumption of r2c large table was far greater than the performance improvement brought by native, so I considered using RAS to solve this problem.

zhztheplayer:
I assume this is closed by https://github.com/apache/incubator-gluten/pull/7686. Please reopen if needed.

PR #7686, opened 2024-10-25
[GLUTEN-7685][VL][RAS] Add new cost model to avoid costly r2c

zml1206:

(Fixes: \#7685)
In production, we encountered a situation where the consumption of r2c large table was far greater than the performance improvement brought by native.
for example:
```
sql("set spark.gluten.sql.columnar.filescan=false")
spark.range(100000000).toDF("id").selectExpr("concat('id_', round(id/1000000)) as k", "id % 10 as v")
      .write.mode("overwrite").parquet("tmp/t1")
spark.read.parquet("tmp/t1").createOrReplaceTempView("t1")
sql("select  k,sum(v) as v from t1 group by k").collect()
```

The local test takes 18 seconds to enble gluten, and only 6 seconds to disablegluten.Therefore, I hope to fallback this through RAS.

The optimization points are as follows:
1. Increase bytesSize factor, cost = bytesSizeFactor * opCost
2. r2c cost can be configured separately, and the default is 100. If sizeBytes is less than the threshold, the cost of RowToColumnarLike is ignored.
3. Remove r2c fallback containing complex types
4. Vanilla op cost is configurable, the default is 20, gluten op cost is 1

## How was this patch tested?

UT and local test，the picture below shows the tasks that have significantly improved our environment.

image


zml1206:
CI failed unrelated.

zhztheplayer:
Run Gluten Clickhouse CI

gluten-substrait/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
```diff
@@ -16,50 +16,62 @@
...
+import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec
 
 /** A rough cost model with some empirical heuristics. */
 class RoughCostModel extends LongCostModel {
 
+  private def getSizeFactor(plan: SparkPlan): Long = {
+    // Get the bytes size that the plan needs to consume
+    val sizeBytes = plan match {
+      case scan: DataSourceScanExec => getStatSizeBytes(scan)
+      case _: LeafExecNode => 0L
```
zhztheplayer, line 34: Why this line is needed?

gluten-substrait/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
```diff
@@ -16,50 +16,62 @@
...
+import org.apache.spark.sql.execution.{ColumnarToRowExec, DataSourceScanExec, LeafExecNode, ProjectExec, RowToColumnarExec, SparkPlan}
+import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanExec
 
 /** A rough cost model with some empirical heuristics. */
 class RoughCostModel extends LongCostModel {
 
+  private def getSizeFactor(plan: SparkPlan): Long = {
+    // Get the bytes size that the plan needs to consume
+    val sizeBytes = plan match {
+      case scan: DataSourceScanExec => getStatSizeBytes(scan)
```
zhztheplayer, line 33: Do we need to consider DSV2?

zhztheplayer (commented):
Is it possible for you to share some test results about this change? Thanks!

gluten-substrait/src/main/scala/org/apache/gluten/planner/cost/RoughCostModel.scala
zml1206: Some LeafExecNode bytesSize is particularly large and inaccurate, such as ExecutedCommandExec, which is ignored here. In addition, the AdaptiveSparkPlanExec match is lost here. I will fix it immediately.

zml1206:
> Is it possible for you to share some test results about this change? Thanks!

I ran it in our grayscale adhoc query environment yesterday, and the overall performance increased by 10%. @zhztheplayer

zhztheplayer:
> > Is it possible for you to share some test results about this change? Thanks!
> 
> I ran it in our grayscale adhoc query environment yesterday, and executorCpuTime overall reduced by 10%. @zhztheplayer

Sounds great.

BTW, based on my impression RoughCostModel has been used by a few of users. If you want to continue working on the new costers, would you consider creating a new one like `RoughCostModel2`, with alias `rough2`, probably? So we don't break the current usages of `rough` model by accident. Once it gets mature rough, we can move the logics to a more standardized cost model then.

zhztheplayer (approved):
👍

zhztheplayer:
@zml1206 Can you add one line in PR description to demonstrate the way to enable this cost model? Thanks.

zml1206:
> @zml1206 Can you add one line in PR description to demonstrate the way to enable this cost model? Thanks.

Added.

Issue #7664, opened 2024-10-24
[VL] Build gluten on aarch64 with VCPKG


z-anderson:
### Problem description

I'm trying to build gluten and velox on aarch64 architecture. I'm running the script `./dev/builddeps-veloxbe.sh --enable_vcpkg=ON --spark_version=3.5` . I get the error below. I see that these options `-mavx2 -mfma -mavx -mf16c -mlzcnt -mbmi2` are only for x86_64, so they won't work on aarch64. Do you know how I can build gluten for aarch64 please? (In version 1.1.x.x., we were able to build for both aarch64 and x86_64). 

I see vcpkg `init.sh` has `VCPKG_TRIPLET=x64-linux-avx` https://github.com/apache/incubator-gluten/blob/c82af60a635fb52bf1314cd831f5915dd37d910d/dev/vcpkg/init.sh#L10. It makes sense that the `x64` and the `avx` parts of this triplet will cause errors running on aarch64. I see it got changed in https://github.com/apache/incubator-gluten/commit/f00e4536bc7ccc56a2eee5800084319130cbf806#diff-4a8a09cebf7ed92cb7321136b9ef223c4653757270e77e942fd6fe2dca30793e -  do you recommend that we should cherry-pick that commit (and a few more commits that it depends on)? 

```
CMake Error at /usr/local/share/cmake-3.28/Modules/CMakeTestCCompiler.cmake:67 (message):
  The C compiler

    "/usr/bin/cc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: '/home/circleci/project/dev/vcpkg/.vcpkg/buildtrees/detect_compiler/x64-linux-avx-rel/CMakeFiles/CMakeScratch/TryCompile-9obttD'

    Run Build Command(s): /usr/bin/ninja -v cmTC_32844
    [1/2] /usr/bin/cc   -fPIC -mavx2 -mfma -mavx -mf16c -mlzcnt -mbmi2 -o CMakeFiles/cmTC_32844.dir/testCCompiler.c.o -c /home/circleci/project/dev/vcpkg/.vcpkg/buildtrees/detect_compiler/x64-linux-avx-rel/CMakeFiles/CMakeScratch/TryCompile-9obttD/testCCompiler.c
    FAILED: CMakeFiles/cmTC_32844.dir/testCCompiler.c.o
[...]

PR #7687, opened 2024-10-25
[GLUTEN-7664][VL][ARM] use arm64 triplet with vcpkg

z-anderson:

Uses the `arm64-linux` triplet when building on arm64 with vcpkg. It uses the community triplet. Currently, vcpkg uses the `x64-linux-avx` triplet, with avx-specific instructions, which gives an error.

(Fixes: https://github.com/apache/incubator-gluten/issues/7664)

## How was this patch tested?

### DIFF
--- a/dev/vcpkg/env.sh
+++ b/dev/vcpkg/env.sh
@@ -10,7 +10,15 @@ SCRIPT_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")")"
 
 export VCPKG_ROOT="$SCRIPT_ROOT/.vcpkg"
 export VCPKG="$SCRIPT_ROOT/.vcpkg/vcpkg"
-export VCPKG_TRIPLET=x64-linux-avx
+ARCH=`uname -m`
+if [ $ARCH == 'x86_64' ]; then
+  VCPKG_TRIPLET=x64-linux-avx
+elif [[ "$ARCH" == 'arm64' || "$ARCH" == 'aarch64' ]]; then
+  VCPKG_TRIPLET=arm64-linux
+else
+  echo "Unsupported arch: $ARCH"
+  exit 1
+fi
 export VCPKG_TRIPLET_INSTALL_DIR=${SCRIPT_ROOT}/vcpkg_installed/${VCPKG_TRIPLET}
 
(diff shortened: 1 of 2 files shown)


PHILO-HE:
@z-anderson, we may need to create a file named `arm64-linux.cmake` like the below `x64-linux-avx.cmake`.
https://github.com/apache/incubator-gluten/blob/main/dev/vcpkg/triplets/x64-linux-avx.cmake

FelixYBW:
@guyuqi would you have a try?

zhouyuan:
@z-anderson thanks for the improvement, could you please share which platform are you running? `aarch64` `arm64` seems different architectures

-yuan

dev/vcpkg/env.sh
```diff
@@ -10,7 +10,15 @@ SCRIPT_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")")"
 
 export VCPKG_ROOT="$SCRIPT_ROOT/.vcpkg"
 export VCPKG="$SCRIPT_ROOT/.vcpkg/vcpkg"
-export VCPKG_TRIPLET=x64-linux-avx
+ARCH=`uname -m`
+if [ $ARCH == 'x86_64' ]; then
+  VCPKG_TRIPLET=x64-linux-avx
```
PHILO-HE, line 15: We should keep `export` used.

z-anderson:
@zhouyuan I'm using aarch64, but I see triplets defined for arm64 (not for aarch64), so I think the triplet is supposed to be for arm64 https://github.com/microsoft/vcpkg/tree/master/triplets

FelixYBW:
Looks like arm64 is used by apple and aarch64 is by others. They are the same. @guyuqi

guyuqi:
> Looks like arm64 is used by apple and aarch64 is by others. They are the same. @guyuqi

`Arm64` and `aarch64` are often used interchangeably and they just have slight differences.

`aarch64` is used more precisely in the context of low-level programming, compiler settings, and technical specifications.
`Arm64` is used more in the context of discussing ARM's architecture in general, especially in marketing and broader technical discussions.

And on most Linux distributions, the "arch" command returns "aarch64", while on Apple M series systems, the "arch" command returns "arm64".

Issue #7690, opened 2024-10-26
GlutenConfig should using SQLConf provided by SparkSession first


beliefer:
### Description

GlutenConfig wrapped the immutable SQLConf instance, this lead to the dynamically modified configurations in interactive scenarios can't apply to runtime.
Take for an example.
```
scala > SET spark.gluten.sql.enable.native.validation = false;
...
scala > SELECT ...
...
SET spark.gluten.sql.enable.native.validation = true;
...
scala > SELECT ...
...
```

PR #7691, opened 2024-10-26
[GLUTEN-7690][CORE][CH][VL] GlutenConfig should support runtime configuration changes

beliefer:

This PR proposes to improve `GlutenConfig` so as it supports runtime configuration changes.
This PR also makes below improvements.

1. Unify the variable name of `SparkSession` with `spark` is usually used by Spark.
2. Avoid declaring configuration value as a constant. e.g. `val scanOnly: Boolean = glutenConf.enableScanOnly`.
3. Other changes.

(Fixes: https://github.com/apache/incubator-gluten/issues/7690)

## How was this patch tested?

integration tests

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/extension/RewriteDateTimestampComparisonRule.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/extension/RewriteDateTimestampComparisonRule.scala
@@ -53,11 +53,11 @@ class RewriteDateTimestampComparisonRule(spark: SparkSession)
     "yyyy"
   )
 
+  private val glutenConf = new GlutenConfig(spark)
+
   override def apply(plan: LogicalPlan): LogicalPlan = {
     if (
-      plan.resolved &&
-      GlutenConfig.getConf.enableGluten &&
-      GlutenConfig.getConf.enableRewriteDateTimestampComparison
+      plan.resolved && glutenConf.enableGluten && glutenConf.enableRewriteDateTimestampComparison
     ) {
       visitPlan(plan)
     } else {
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHRuleApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHRuleApi.scala
@@ -68,29 +68,29 @@ private object CHRuleApi {
     // Gluten columnar: Transform rules.
     injector.injectTransform(_ => RemoveTransitions)
     injector.injectTransform(_ => PushDownInputFileExpression.PreOffload)
-    injector.injectTransform(c => FallbackOnANSIMode.apply(c.session))
-    injector.injectTransform(c => FallbackMultiCodegens.apply(c.session))
+    injector.injectTransform(c => FallbackOnANSIMode.apply(c.spark))
+    injector.injectTransform(c => FallbackMultiCodegens.apply(c.spark))
     injector.injectTransform(_ => RewriteSubqueryBroadcast())
-    injector.injectTransform(c => FallbackBroadcastHashJoin.apply(c.session))
-    injector.injectTransform(c => MergeTwoPhasesHashBaseAggregate.apply(c.session))
+    injector.injectTransform(c => FallbackBroadcastHashJoin.apply(c.spark))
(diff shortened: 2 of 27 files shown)


zhztheplayer:
> GlutenConfig prevents the dynamically modified configurations in interactive scenarios.

@beliefer Would you like to post an code example of the interactive case you mentioned? Or probably to add a test case to guard against it?

backends-clickhouse/src/main/scala/org/apache/gluten/extension/RewriteDateTimestampComparisonRule.scala
```diff
@@ -37,7 +37,7 @@ import org.apache.spark.unsafe.types.UTF8String
 // This rule try to make the filter condition into integer comparison, which is more efficient.
 // The above example will be rewritten into
 //    select * from table where to_unixtime('2023-11-02', 'yyyy-MM-dd') >= unix_timestamp
-class RewriteDateTimestampComparisonRule(session: SparkSession, conf: SQLConf)
```
zhztheplayer, line 40: Thanks, this kind of simplification is needed.

zhztheplayer: Though, do we really need `session` as a parameter to create `GlutenConfig`? What's the issue of using `GlutenConfig.getConf`?

beliefer: If there is an active `SparkSession`, the proper `SQLConf` associated with the thread's active session is used. So we should use `spark.sessionState.conf` first.
`Glutenconfig.getConf` actually proxies `SQLConf.get`, which will prioritize obtaining SQLConf from the active session, otherwise the default SQLConf will be used.

beliefer:
> > GlutenConfig prevents the dynamically modified configurations in interactive scenarios.
> 
> @beliefer Would you like to post a code example of the interactive case you mentioned? Or probably to add a test case to guard against it?

Spark already have a lot of test cases. So I added the code example of the interactive case into the description of [issue](https://github.com/apache/incubator-gluten/issues/7690)

beliefer:
@zhztheplayer @zhouyuan @FelixYBW The failure GC seems unrelated.

backends-clickhouse/src/main/scala/org/apache/gluten/extension/RewriteDateTimestampComparisonRule.scala
zhztheplayer: > Glutenconfig.getConf actually proxies SQLConf.get

Spark has `SQLConfHelper` that a lot of components (including `Rule`) rely on, using `SQLConf.get` to access session config. Is there any special consideration to avoid using it in Gluten?

zhztheplayer:
@beliefer Aside of the constructor changes for `GlutenConfig`, I think 2 parts of code cleanups can be split and merged first. Would you want to split out the following parts from the PR?

1. Rule constructor simplifications: E.g., `RewriteDateTimestampComparisonRule(session: SparkSession, conf: SQLConf)` to `RewriteDateTimestampComparisonRule(conf: SQLConf)` or `RewriteDateTimestampComparisonRule(session: SparkSession)`
2. Naming improvements: E.g., `columnarConf` to `glutenConf`

Thanks!

backends-clickhouse/src/main/scala/org/apache/gluten/extension/RewriteDateTimestampComparisonRule.scala
beliefer: No. This PR doesn't avoid using `SQLConf.get` in Gluten.
In fact, `SQLConf.get` and `spark.sessionState.conf` are functionally equivalent. But `spark.sessionState.conf` is faster than `SQLConf.get` because the latter requires visit thread local. So we can see the usage scenarios:

1. The code rely on `SQLConfHelper` in Spark we can't find `SparkSession` in context.
2. All the places in Spark rely on `spark.sessionState.conf` if exists `SparkSession`.

beliefer:
Run Gluten Clickhouse CI

beliefer:
Run Gluten Clickhouse CI

PR #7697, opened 2024-10-28
[WIP] Support libhdfs3 in velox

JkSelf:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


FelixYBW:
Just talked with Ke. Here is a summary:

1. Previously we link to libhdfs3 at compile time, it can be static or dynamic link.
2. Velox currently loads libhdfs library (can be jvm libhdfs or libhdfs3) at run time, record the pointer to libhdfs functions, then indirectly call the function when accessing hdfd.
3. libhdfs shows much worse performance than libhdfs3
4. libhdfs3 has a ~7% regression compared to link way in 1.

The PR is to confirm 4. Also we need to make clear the root cause of 3.

Thank you! Ke.

zhouyuan:
@GlutenPerfBot benchmark

PHILO-HE (commented):
@JkSelf, we need change to use `ENABLE_HDFS3` for the below code, then static libhdfs3 from vcpkg's build can be available to use. 
https://github.com/apache/incubator-gluten/blob/c55e3d5bce2e0e6faf38e80b233f2ec3088e14fb/dev/vcpkg/init.sh#L78

Also note CMAKE_FIND_LIBRARY_SUFFIXES is set to ".so" when finding `libhdfs3`, but this patch's purpose is to use static linkage. Is it simply for some test?

ep/build-velox/src/build_velox.sh
```diff
@@ -104,6 +110,9 @@ function compile {
   if [ $ENABLE_HDFS == "ON" ]; then
     COMPILE_OPTION="$COMPILE_OPTION -DVELOX_ENABLE_HDFS=ON"
   fi
+  if [ $ENABLE_HDFS3 == "ON" ]; then
+    COMPILE_OPTION="$COMPILE_OPTION -DVELOX_ENABLE_HDFS3=ON"
```
leoluan2009, line 114: After the pr https://github.com/facebookincubator/velox/commit/10cdf6fd8ca372b11779431ef77a177551c01700 , velox have removed VELOX_ENABLE_HDFS3

dev/vcpkg/init.sh
```diff
@@ -26,6 +27,10 @@ for arg in "$@"; do
     ENABLE_HDFS=("${arg#*=}")
     shift # Remove argument name from processing
```
gaoyangxiaozhu, line 28: does allow people enable flag both `ENABLE_HDFS` and `ENABLE_HDFS3` ?

FelixYBW: If we can't see performance gain from compile time link, we will close the PR.

Issue #7703, opened 2024-10-28
[VL] ColumnarBuildSideRelation transform support multiple key columns


yikf:
### Backend

VL (Velox)

### Bug description

ColumnarBuildSideRelation transform support multiple key columns. 

Currently, ColumnarBuildSideRelation#transform will fail with IllegalArgumentException. In fact, we can calculate the value of the expression directly from `UnsafeProjection.create(Seq(key), output)`

### Spark version

None

Issue #1141, opened 2023-03-15
TPC-DS q72 failed cause more than one key


Yohahaha:
**Describe the bug**
TPC- DS q72 throw exceptions
```
Caused by: java.lang.IllegalStateException: More than one key d_date_sk#166 found from build side relation output: List(d_date_sk#92, d_date#94, d_week_seq#96, d_date_sk#166)
	at org.apache.spark.sql.execution.GlutenBuildSideRelation.$anonfun$transform$1(GlutenBuildSideRelation.scala:100)
```

zhouyuan:
@Yohahaha 

The issue is due to "keys with same name" is marked as unsupported when building hash map. I also run into this issue on Q72 + CBO enabled. 

Can you please help to have a check if below quick patch could work? In my testing env it works.
 
```
diff --git a/gluten-data/src/main/scala/org/apache/spark/sql/execution/GlutenBuildSideRelation.scala b/gluten-data/src/main/scala/org/apache/spark/sql/execution/GlutenBuildSideRelation.scala
index 40ea5b95..f926a0d1 100644
--- a/gluten-data/src/main/scala/org/apache/spark/sql/execution/GlutenBuildSideRelation.scala
+++ b/gluten-data/src/main/scala/org/apache/spark/sql/execution/GlutenBuildSideRelation.scala
@@ -89,7 +89,7 @@ case class GlutenBuildSideRelation(mode: BroadcastMode,

         val columnInOutput = output.zipWithIndex.filter {
           p: (Attribute, Int) =>
-            p._1.name == columnExpr.name
+            p._1.name == columnExpr.name && p._1.exprId == columnExpr.exprId
         }
         if (columnInOutput.isEmpty) {
           throw new IllegalStateException(

```

-yuan

Yohahaha:
@zhouyuan 

It works.

I think use `exprId` and `name` to check `AttributeReference`'s equality is better, why we dont use this way or apply this patch for now?

jinchengchenghh:
Can we apply this patch? @zhouyuan

FelixYBW:
> Can we apply this patch? @zhouyuan

Is this fixed? If not who can submit the PR？

zhouyuan:
@jinchengchenghh  The patch is a quick fix and it seems introducing more bugs on other cases.  Created a pull request(#1288 ) on fixing these issues 

-yuan

jinchengchenghh:
It has passed CI. Does it have some hidden bugs?

[...]

PR #7704, opened 2024-10-28
ColumnarBuildSideRelation transform support multiple key columns

yikf:

This PR aims to fix https://github.com/apache/incubator-gluten/issues/7703.

Currently, ColumnarBuildSideRelation#transform will fail with IllegalArgumentException. In fact, we can calculate the value of the expression directly from `UnsafeProjection.create(Seq(key), output)`

## How was this patch tested?
unit tests

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarBuildSideRelation.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarBuildSideRelation.scala
@@ -25,7 +25,7 @@ import org.apache.gluten.utils.ArrowAbiUtil
 import org.apache.gluten.vectorized.{ColumnarBatchSerializerJniWrapper, NativeColumnarToRowJniWrapper}
 
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, BoundReference, Expression, UnsafeProjection, UnsafeRow}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression, UnsafeProjection, UnsafeRow}
 import org.apache.spark.sql.execution.joins.BuildSideRelation
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.utils.SparkArrowUtil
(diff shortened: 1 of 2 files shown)


yikf:
@PHILO-HE Please take a look if you find a time, thanks

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxHashJoinSuite.scala
```diff
@@ -144,4 +147,38 @@ class VeloxHashJoinSuite extends VeloxWholeStageTransformerSuite {
...
+                     |SELECT t1.date, t1.t1_c1, t2.t2_c2
+                     |FROM t1
+                     |JOIN t2 ON t1.date = t2.date
+                     |WHERE t1.date=if(3 <= t2.t2_c2, if(3 < t2.t2_c1, 3, t2.t2_c1), t2.t2_c2)
+                     |ORDER BY t1.date DESC, t1.t1_c1 DESC, t2.t2_c2 DESC
+                     |LIMIT 1
+                     |""".stripMargin)
+
+      checkAnswer(df, Row("3", 1, 4) :: Nil)
+      // collect the DDP plan.
```
PHILO-HE, line 172: typo: DDP -> DPP

PHILO-HE (approved):
It's a clear way to bind reference. Looks good. Thanks!

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarBuildSideRelation.scala
```diff
@@ -141,48 +141,7 @@ case class ColumnarBuildSideRelation(output: Seq[Attribute], batches: Array[Arra
...
-            val oneColumnWithSameName = output.count(_.name == columnExpr.name) == 1
-            val columnInOutput = output.zipWithIndex.filter {
-              p: (Attribute, Int) =>
-                if (oneColumnWithSameName) {
-                  // The comparison of exprId can be ignored when
-                  // only one attribute name match is found.
-                  p._1.name == columnExpr.name
-                } else {
-                  // A case where output has multiple columns with same name
-                  p._1.name == columnExpr.name && p._1.exprId == columnExpr.exprId
```
zhouyuan, line 166: I recall this is related one special AQE reuse when CBO enabled
https://github.com/apache/incubator-gluten/issues/1141

yikf: I suspect this has something to do with the previous implementation of bound references. the purpose of the transform is to retrieve the values computed by expr from the relation, I think we can directly simply use the clear way to retrieve them.

I ran q72 in my test environment and it works normally. Can you help double check if you find a time?

PHILO-HE: I'm doing the verification. Thanks!

yikf: @PHILO-HE , Hi, is there anything wrong with the verification?

PHILO-HE: @yikf, sorry for the delayed response. I tested this patch on all TPCDS queries with Spark CBO turned on. There is no functionality issue.

yikf: Thank you @PHILO-HE @zhouyuan

PR #7701, opened 2024-10-28
[GLUTEN-7143][VL] Fix several UTs for RAS

zhztheplayer:
Part of #7143

Fix following UTs when RAS=ON:
```
	Line  3111: 2024-10-28T03:45:33.5270060Z - fallback final aggregate of collect_list *** FAILED ***
	Line  3272: 2024-10-28T03:45:44.0086490Z - count(1) on csv scan *** FAILED ***
	Line  4403: 2024-10-28T03:47:23.3298481Z - Date type *** FAILED ***
	Line  4794: 2024-10-28T03:47:41.5650328Z - Date type *** FAILED ***
	Line  5030: 2024-10-28T03:47:50.0627363Z - count(1) on csv scan *** FAILED ***
	Line  5129: 2024-10-28T03:47:50.4155755Z - Struct Literal *** FAILED ***
	Line  5131: 2024-10-28T03:47:50.4905146Z - Array Literal *** FAILED ***
	Line  5133: 2024-10-28T03:47:50.5682606Z - Map Literal *** FAILED ***
	Line  5135: 2024-10-28T03:47:50.6383717Z - Null Literal *** FAILED ***
	Line  5137: 2024-10-28T03:47:50.7036012Z - Scalar Type Literal *** FAILED ***
	Line  5156: 2024-10-28T03:47:52.2401153Z - arrow_udf test: without projection *** FAILED ***
	Line  5161: 2024-10-28T03:47:52.2709234Z - arrow_udf test: with unrelated projection *** FAILED ***
	Line  6657: 2024-10-28T03:52:33.4765722Z - test hive static partition write table *** FAILED ***
	Line  7025: 2024-10-28T03:52:34.9769802Z - test hive write table *** FAILED ***
	Line 31956: 2024-10-28T03:55:26.4473034Z - test OneRowRelation *** FAILED ***
	Line 32130: 2024-10-28T03:55:37.9358997Z - Test sample op *** FAILED ***
	Line 32372: 2024-10-28T03:56:02.3208216Z - cast date to timestamp with timezone *** FAILED ***
	Line 33543: 2024-10-28T03:56:57.4949591Z *** 18 TESTS FAILED ***
```

Remaining unfixed:
```
	Line  3149: 2024-10-28T03:45:35.3305691Z - test ignore row to columnar *** FAILED ***
        ... (more)
```

Issue #7143, opened 2024-09-06
[VL] Enable RAS planner by default in Velox backend


zhztheplayer:
### Description

RAS planner has been added into code base for 6 months, and it's time to move it forward.

In out company, a internal fallback benchmark, which consists of queries with considerable number of operators fell back to vanilla Spark, has been showing averagely ~20% overall speedup (for certain queries, up to ~300%) when RAS planner is turned on and a well-tuned RAS cost model is set.

Note the cost model we are testing on internally is still based on hardcoded estimated cost numbers for operators, without reading logical statistics from vanilla Spark. So we still have space to enhance the cost model work in future.

And given that we now able to align query plans generated between RAS planner and legacy planner with the new default cost model https://github.com/apache/incubator-gluten/pull/6493, it's feasible now to turn on RAS planner for Velox by default.

Note, one of the biggest difference between RAS planner and legacy planner is that RAS planner doesn't support [whole stage fallback](https://github.com/apache/incubator-gluten/blob/37d09c1092a3916a28f23da9b27cc799a8885a5e/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala#L1206-L1212). Since RAS planner achieves similar effect via cost model. If you are using that feature and are not prepared moving to RAS, then in your case RAS should be turned off after the change is made. (TODO: raise a warning or error message to user when whole stage fallback and RAS are both enabled)

And note, switching to RAS planner doesn't mean deprecation of legacy columnar rules. In a way RAS is only in charge of the computation offloading part in the entire columnar planning procedure. Regular Spark columnar rules can still be inserted to RAS's columnar rule list. Meanwhile we will soon add a API for backends for inserting RAS rules as well.

Code references:
1. RAS rule applier: https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/EnumeratedApplier.scala
2. RAS rule list: https://github.com/apache/incubator-gluten/blob/37d09c1092a3916a28f23da9b27cc799a8885a5e/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala#L34

DamonZhao-sfu:
Are there any details on the internal fallback benchmark? I'm doing research on the fallback policy with cost model as well. Thanks! @zhztheplayer

zhztheplayer:
[...]

PR #7725, opened 2024-10-30
[GLUTEN-7143][VL] RAS: Fix test case "test ignore row to columnar" when RAS=ON

zhztheplayer:
Part of #7143 

Following https://github.com/apache/incubator-gluten/pull/7701, fix a test case "test ignore row to columnar" when RAS=ON 

And with other follow-up cleanups.


PHILO-HE (approved):
Thanks!

PR #7731, opened 2024-10-30
[VL] RAS: Fix fallen back plan nodes are not tagged with meaningful fallback reasons

zhztheplayer:
Part of #7143

This fixes UT `GlutenExpressionMappingSuite.GLUTEN-7213: Check fallback reason with CheckOverflowInTableInsert` in RAS.


PHILO-HE (approved):
Make sense. Thanks!

Issue #7446, opened 2024-10-09
Gluten should build thirdparty lib using `jar` from JAVA_HOME and not from default classpath


CodenameGHOST007:
### Description

Currently we just reference `jar` while building thirdparty lib which would pick up the jdk from classpath of the system and that is not ideal since a dev env can have multiple java configured. Ideally gluten should verify JAVA_HOME and use that for all subsequent builds.

PR #7736, opened 2024-10-30
[GLUTEN-7446][BUILD] build third party libs using jar from JAVA_HOME

Zand100:

Builds third party libs jar using the `jar` binary from `JAVA_HOME`. Just using `jar` would pick up the JDK from the classpath of the system, and that is not ideal since a dev env can have multiple JDK configured.

(Fixes: \#7446)

## How was this patch tested?

Existing unit and integration tests

### DIFF
--- a/dev/build-thirdparty.sh
+++ b/dev/build-thirdparty.sh
@@ -82,4 +82,4 @@ elif [ "$LINUX_OS" == "debian" ]; then
   fi
 fi
 cd $THIRDPARTY_LIB/
-jar cvf gluten-thirdparty-lib-$LINUX_OS-$VERSION-$ARCH.jar ./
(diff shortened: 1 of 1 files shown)


Yohahaha:
what will happened if we use `jar` but `$JAVA_HOME/bin/jar` ?

Zand100:
@Yohahaha I'm confused; can you give more details please?

Yohahaha:
> @Yohahaha I'm confused; can you give more details please?

this PR LGTM, please add the reason in PR description for why this changes is needed, or what's difference between `jar` and `$JAVA_HOME/bin/jar` when building Gluten.

Zand100:
@Yohahaha can you review please?

Yohahaha (approved):
👍

Issue #7751, opened 2024-10-31
[VL] Velox backend support merge two phase hash-based aggregate into one aggregate


yikf:
### Description

In issue #4668, support for the CH backend's merge agg capability was added. We can also support this feature in the velox backend.

Issue #4668, opened 2024-02-07
[CH] Merge two phase hash-based aggregate into one aggregate in the spark plan when there is no shuffle


zzcclp:
### Description

Merge two phase hash-based aggregate into one aggregate in the spark plan when there is no shuffle:
Examples:
```
 HashAggregate(t1.i, SUM, final)
                |                                 =>    HashAggregate(t1.i, SUM, complete)
 HashAggregate(t1.i, SUM, partial)
```
now this feature only support for CH backend.

PR #7752, opened 2024-10-31
[GLUTEN-7751][VL] Merge two consecutive aggregates to one in complete mode

yikf:

Fix https://github.com/apache/incubator-gluten/issues/7751.

In issue #4668, support for the CH backend's merge agg capability was added. We can also support this feature in the velox backend.

## How was this patch tested?
Through existing unit tests and GA verification, and additional new unit tests have been added.


yikf:
@zzcclp @PHILO-HE Could you please take a look if you find a time, thanks

zhouyuan:
@yikf can you please also check if `COMPLETE` agg is offload to Velox native? I think we need more changes to support this feature

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
```diff
@@ -356,7 +356,7 @@ abstract class HashAggregateExecTransformer(
           // The process of handling the inconsistency in column types and order between
           // Spark and Velox is exactly the opposite of applyExtractStruct.
           aggregateExpression.mode match {
-            case PartialMerge | Final =>
+            case PartialMerge | Final | Complete =>
```
jackylee-ch, line 359: Have you tested the complete Agg? This may cause failure on rewrite. I have tested the #6198 , and it has run through all TPCDS queries with `Partial Agg Pushed down`, which would lead many Complete Agg Nodes.

FelixYBW:
Glad to see we can learn each other from different backends. It's one of our goal to support different backends in Gluten.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/MergeTwoPhasesHashBaseAggregate.scala
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.gluten.extension
+package org.apache.gluten.extension.columnar
```
zzcclp, line 17: It needs to modify the import path for this class in CHRuleApi.injectLegacy

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
yikf: I added [unit tests](https://github.com/apache/incubator-gluten/pull/7752/files#diff-742ce9027f0d774d324125a130d9993bad09d4ea5eb6d35d791b39bb57138c77R1260) to check data consistency, that the operator is executed as HashAggregateExecBaseTransformer, that substrait plan is in the COMPLETE mode, and that there are existing TPC-H test execution plans and unit tests. cc @zhouyuan 
image


In velox backend, the mode of the aggregate function depends on the core::AggregationNode::Step, I understand the right aggregation patterns are pushed down on it. And I'm sorry that I don't quite understand the problem `which would lead many Complete Agg Nodes`. Could you explain it in more detail?

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/MergeTwoPhasesHashBaseAggregate.scala
yikf: The [import of this package](https://github.com/apache/incubator-gluten/blob/main/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHRuleApi.scala#L21) was previously included.

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
jackylee-ch: I used #6198 to support COMPLETE Mode for velox before. If ok, I can do double check for your pr to see if there is some prolem with it.

yikf: It seems to be push down in the optimizer phase, where partial and final aggs are not connected aggregates, and the merge agg theoretically won't handle this pattern. I'm guessing you mean the modification in the HashAggregateExecTransformer that supports the COMPLETE part, okay, you can help me check that, thank you.

jackylee-ch: Current code works with our environment, thanks.

docs/Configuration.md
```diff
@@ -58,6 +58,7 @@ You can add these configurations into spark-defaults.conf to enable or disable t
...
+| spark.gluten.sql.mergeTwoPhasesAggregate.enabled             | Whether to merge two phases aggregate if if there are no other operators between two aggregates.                                                                                                                                                                                                                                                                                                                                                                                                                                          | true                                                 |
```
weixiuli, line 61: May be : `Whether to merge two phases aggregate if there are no other operators between two aggregates.` ,right ?

jackylee-ch:
> @yikf can you please also check if `COMPLETE` agg is offload to Velox native? I think we need more changes to support this feature

This pr works well in our environment with `COMPLETE` mode Agg, I think we can move on now.

docs/Configuration.md
yikf: addressed, thanks

jackylee-ch (approved):

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxMergeTwoPhasesHashBaseAggregateSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+        2)
+    }
+  }
+}
+
+class VeloxMergeTwoPhasesAggregateSuiteOn extends BaseVeloxMergeTwoPhasesHashBaseAggregateSuite {
+  override val adaptiveExecutionOn: Boolean = true
+}
+
+class VeloxMergeTwoPhasesAggregateSuiteOff extends BaseVeloxMergeTwoPhasesHashBaseAggregateSuite {
```
PHILO-HE, line 196: ditto.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxMergeTwoPhasesHashBaseAggregateSuite.scala
```diff
@@ -0,0 +1,198 @@
...
+        spark.sql("""
+                    |SELECT sum(if(id<0,0,id)) FROM t1
+                    |""".stripMargin),
+        2)
+    }
+  }
+}
+
+class VeloxMergeTwoPhasesAggregateSuiteOn extends BaseVeloxMergeTwoPhasesHashBaseAggregateSuite {
+  override val adaptiveExecutionOn: Boolean = true
```
PHILO-HE, line 193: Can we simply let it extend Spark's `EnableAdaptiveExecutionSuite`?

PHILO-HE (commented):
Thanks for your work!

gluten-ut/spark33/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
```diff
@@ -886,6 +886,7 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude("replace partial and final hash aggregate together with sort aggregate")
     .exclude("do not replace hash aggregate if child does not have sort order")
     .exclude("do not replace hash aggregate if there is no group-by column")
+    .excludeGlutenTest("replace partial and final hash aggregate together with sort aggregate")
```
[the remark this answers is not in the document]
yikf: no, it conflicts with Spark's ReplaceHashWithSortAgg, gluten prefers to MergeTwoPhasesHashBaseAggregate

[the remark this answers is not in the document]
PHILO-HE: When Spark's `ReplaceHashWithSortAgg` is enabled, Gluten still transforms sort agg to its hash agg by default. This test was actually ported from Spark and the expected number of hash/sort agg nodes were modified to adapt to Gluten's planning. Can we keep this test? Maybe, you need modify the test or simply disable `MergeTwoPhasesHashBaseAggregate` here?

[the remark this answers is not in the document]
jackylee-ch: I have tried this suite before. To retain this test, we must set `mergeTwoPhasesAggregate` to false and eliminate the `COUNT` test, since it is no longer relevant after the current PR's conversion to HashAggTransformer.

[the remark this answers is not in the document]
yikf: ok, will keep this test

zhztheplayer (approved):
Would you like to add `MergeTwoPhasesHashBaseAggregate` to `injectRas` list the remove this line to make the feature covered by RAS UT? Which could be done in a separate PR. Thanks.

zhouyuan (approved):
👍 
verified locally with a patched spark with pushing agg thru join enabled

Issue #7753, opened 2024-10-31
[CORE] `PullOutPreProject` should not replace the `literal`s in `expand`'s projections


lgbo-ustc:
### Description

The `literal`s in `expand`'s projection list is used as grouping id, they should not be replaced by pre-projection attributes

lgbo-ustc:
For query
```sql
select n_name, count(distinct n_regionkey) as col1, count(distinct concat(n_regionkey, n_nationkey)) as col2 from tpch_pq.nation group by n_name;
```
The `literal`s in `expand` are all replaced by pre-projection attributes
```
CHNativeColumnarToRow
+- ^(5) HashAggregateTransformer(keys=[n_name#6], functions=[count(spark_catalog.tpch_pq.nation.n_regionkey#35L), count(concat(CAST(spark_catalog.tpch_pq.nation.n_regionkey AS STRING), CAST(spark_catalog.tpch_pq.nation.n_nationkey AS STRING))#36)], isStreamingAgg=false)
   +- ^(5) InputIteratorTransformer[n_name#6, count#39L, count#40L]
      +- ColumnarExchange hashpartitioning(n_name#6, 5), ENSURE_REQUIREMENTS, [plan_id=343], [shuffle_writer_type=hash], [OUTPUT] ArrayBuffer(n_name:StringType, count:LongType, count:LongType)
         +- ^(4) HashAggregateTransformer(keys=[n_name#6], functions=[partial_count(spark_catalog.tpch_pq.nation.n_regionkey#35L) FILTER (WHERE _pre_6#46), partial_count(concat(CAST(spark_catalog.tpch_pq.nation.n_regionkey AS STRING), CAST(spark_catalog.tpch_pq.nation.n_nationkey AS STRING))#36) FILTER (WHERE _pre_7#47)], isStreamingAgg=false)
            +- ^(4) ProjectExecTransformer [n_name#6, spark_catalog.tpch_pq.nation.n_regionkey#35L, concat(CAST(spark_catalog.tpch_pq.nation.n_regionkey AS STRING), CAST(spark_catalog.tpch_pq.nation.n_nationkey AS STRING))#36, gid#34, (gid#34 = 1) AS _pre_6#46, (gid#34 = 2) AS _pre_7#47]
[...]

PR #7756, opened 2024-10-31
[GLUTEN-7753][CORE] Do not replace literals of expand's projects in `PullOutPreProject`

lgbo-ustc:

Fixes: #7753

## How was this patch tested?

unit tests

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
@@ -222,7 +222,13 @@ object PullOutPreProject extends RewriteSingleNode with PullOutProjectHelper {
     case expand: ExpandExec if needsPreProject(expand) =>
       val expressionMap = new mutable.HashMap[Expression, NamedExpression]()
       val newProjections =
-        expand.projections.map(_.map(replaceExpressionWithAttribute(_, expressionMap)))
+        expand.projections.map(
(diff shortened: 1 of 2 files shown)


marin-ma:
Thanks!
@liujiayi771 Could you PTAL? I just wonder if we can apply this change to other operators.

marin-ma (approved):

zzcclp (approved):

liujiayi771:
@marin-ma Many operators do not support `Literal`, such as agg. Therefore, for other operators, we should pull out the `Literal`. If expand supports `Literal`, we can modify it separately in this way.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
```diff
@@ -222,7 +222,8 @@ object PullOutPreProject extends RewriteSingleNode with PullOutProjectHelper {
     case expand: ExpandExec if needsPreProject(expand) =>
       val expressionMap = new mutable.HashMap[Expression, NamedExpression]()
       val newProjections =
-        expand.projections.map(_.map(replaceExpressionWithAttribute(_, expressionMap)))
+        expand.projections.map(
+          _.map(replaceExpressionWithAttribute(_, expressionMap, false, false)))
```
liujiayi771, line 226: It is better to use name parameter here.
```
replaceBoundReference = false,
replaceLiteral = false
```

liujiayi771 (approved):

PR #7770, opened 2024-11-01
[GLUTEN-7143][VL] RAS: Enable the RAS UT jobs in GHA CI

### LINKED ISSUE #7143: [VL] Enable RAS planner by default in Velox backend

zhztheplayer:
Part of #7143

After a series of efforts, the failed UTs when RAS=ON should be all fixed. Enable the GHA jobs introduced in https://github.com/apache/incubator-gluten/pull/7702 so they can guard against RAS from regressions.

After the change, every PR should pass the following UT jobs otherwise CI will fail:

* run-spark-test-spark35-ras
* run-spark-test-spark35-slow-ras

We would keep observing on the jobs. When the time is ripe, we can set RAS planner as the default planner then finish #7143. And before that, a comprehensive community discussion will be needed.


marin-ma (approved):
Thanks!

PR #7764, opened 2024-11-01
[GLUTEN-7143][VL] RAS: Revert #7731, disable the relevant test cases since RAS doesn't report fallback details due to #7763

zhztheplayer:
Part of #7143 

This reverts #7731

See #7763, RAS doesn't support detailed fallback report yet. Leave the feature as TODO.

This also fixes two UTs:

GlutenFallbackSuite | Gluten - test fallback logging
GlutenFallbackSuite | Gluten - Add logical link to rewritten spark plan

PR #7773, opened 2024-11-01
[VL] Remove a duplicated Maven dependency, and some follow-ups for #7764

zhztheplayer:
`gluten-substrait` is referred twice in `backends-velox`'s pom.xml and causes a compile time warning. Fix it.

And with a minor follow-up change for https://github.com/apache/incubator-gluten/pull/7764.


PHILO-HE (approved):
Looks good!

PR #7799, opened 2024-11-04
[GLUTEN-7800][VL] Add config for max reclaim wait time to avoid dead lock when memory arbitration

Yohahaha:

Add config `spark.gluten.sql.columnar.backend.velox.reclaimMaxWaitMs` to avoid dead lock when memory arbitration has bug, default timeout is 60min.

### DIFF
--- a/cpp/velox/memory/VeloxMemoryManager.cc
+++ b/cpp/velox/memory/VeloxMemoryManager.cc
@@ -22,6 +22,7 @@
 
 #include "velox/common/memory/MallocAllocator.h"
 #include "velox/common/memory/MemoryPool.h"
+#include "velox/common/memory/SharedArbitrator.h"
 #include "velox/exec/MemoryReclaimer.h"
 
 #include "compute/VeloxBackend.h"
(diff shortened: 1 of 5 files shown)


Yohahaha:
@zhztheplayer would you help take a look?

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -22,6 +22,7 @@
 
 #include "velox/common/memory/MallocAllocator.h"
 #include "velox/common/memory/MemoryPool.h"
+#include "velox/common/memory/SharedArbitrator.h"
 #include "velox/exec/MemoryReclaimer.h"
```
zhztheplayer, line 26: Why we need to include `SharedArbitrator.h`? Thought we should avoid this kind of dependency.

zhztheplayer (commented):
Only one question about the header dependency. Thanks.

cpp/velox/memory/VeloxMemoryManager.cc
Yohahaha: I just want reuse `velox::memory::SharedArbitrator::ExtraConfig`, would you give some input?

zhztheplayer (approved):

zhztheplayer:
@Yohahaha 

A bit dramatical that upstream Velox really updated the function name in `SharedArbitrator.h`...

See https://github.com/facebookincubator/velox/commit/62b0a128702f4092b1680c4387d892d1d6829d20#diff-00e1979427f8bb6a5bda4f62c02e98dfd2cc0bdadb9219772e667946ffe1421dR111

It causes [main CI build failure](https://github.com/apache/incubator-gluten/commits/761f1c58a3adb6d768680523431ccb67b459f6c5) and let's revert this PR

Issue #7802, opened 2024-11-04
[VL] Use original total memory size if enable dynamic offheap


yikf:
### Backend

VL (Velox)

### Bug description

[Expected behavior] and [actual behavior].

### Spark version

None

Issue #7605, opened 2024-10-18
[VL] Failing allocation in DynamicOffHeapSizingMemoryTarget


wenwj0:
### Backend

VL (Velox)

### Bug description

I am trying to set `spark.gluten.memory.dynamic.offHeap.sizing.enabled=true"`, but OOM exception occurs.
spark configuration :
```
spark.executor.memory=4g;
spark.executor.memoryOverhead=1G;
spark.gluten.memory.dynamic.offHeap.sizing.enabled=true
spark.memory.offHeap.enabled=true
```
and web ui shows:
```
spark.gluten.memory.conservative.task.offHeap.size.in.bytes=597059174
spark.gluten.memory.offHeap.size.in.bytes=2388236697
spark.gluten.memory.task.offHeap.size.in.bytes=597059174
spark.memory.offHeap.size=2388236697
```

And I got the OOM exception:
```
Reason: Operator::getOutput failed for [operator: ValueStream, plan node ID: 0]: Error during calling Java code from native code: org.apache.gluten.memory.memtarget.ThrowOnOomMemoryTarget$OutOfMemoryException: Not enough spark off-heap execution memory. Acquired: 8.0 MiB, granted: 0.0 B. Try tweaking config option spark.memory.offHeap.size to get larger space to run this application (if spark.gluten.memory.dynamic.offHeap.sizing.enabled is not enabled). 
Current config settings: 
    spark.gluten.memory.offHeap.size.in.bytes=3.4 GiB
    spark.gluten.memory.task.offHeap.size.in.bytes=876.6 MiB
    spark.gluten.memory.conservative.task.offHeap.size.in.bytes=876.6 MiB
    spark.memory.offHeap.enabled=true
    spark.gluten.memory.dynamic.offHeap.sizing.enabled=true
Memory consumer stats: 
    Task.52:                                             Current used bytes: 104.0 MiB, peak bytes:        N/A
    \- Gluten.Tree.0:                                    Current used bytes: 104.0 MiB, peak bytes:  112.0 MiB
       \- root.0:                                        Current used bytes: 104.0 MiB, peak bytes:  112.0 MiB
          +- CelebornShuffleWriter.0:                    Current used bytes:  48.0 MiB, peak bytes:   48.0 MiB
          |  \- single:                                  Current used bytes:  48.0 MiB, peak bytes:   48.0 MiB
[...]

PR #7803, opened 2024-11-04
[GLUTEN-7802][CORE] Use original total memory size if enable dynamic offheap

yikf:

This PR aims to fix https://github.com/apache/incubator-gluten/issues/7802.

## How was this patch tested?
manual tests

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -148,8 +148,9 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
           minOffHeapSize))
     ) {
       throw new GlutenException(
-        s"Must set '$SPARK_OFFHEAP_ENABLED' to true " +
-          s"and set '$SPARK_OFFHEAP_SIZE_KEY' to be greater than $minOffHeapSize")
+        s"Must set '$SPARK_OFFHEAP_ENABLED' to true and set " +
+          s"'$SPARK_OFFHEAP_SIZE_KEY' to be greater than $minOffHeapSize, or set " +
+          s"'$GLUTEN_DYNAMIC_OFFHEAP_SIZING_ENABLED' to true to enable dynamic offheap feature.")
     }
 
     // Session's local time zone must be set. If not explicitly set by user, its default
@@ -159,8 +160,6 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
     // Task slots.
     val taskSlots = SparkResourceUtil.getTaskSlots(conf)
     conf.set(GLUTEN_NUM_TASK_SLOTS_PER_EXECUTOR_KEY, taskSlots.toString)
-
-    val onHeapSize: Long = conf.getSizeAsBytes(SPARK_ONHEAP_SIZE_KEY, 1024 * 1024 * 1024)
 
     // If dynamic off-heap sizing is enabled, the off-heap size is calculated based on the on-heap
     // size. Otherwise, the off-heap size is set to the value specified by the user (if any).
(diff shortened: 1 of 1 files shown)


yikf:
@supermem613 @zhztheplayer @ulysses-you  @PHILO-HE Could you please take a look?

FelixYBW:
Be careful, the dynamic offheap feature is experimental.

yikf:
> Be careful, the dynamic offheap feature is experimental.

Thank you for reminding me, yea, I was just testing the feature in a test environment, and I found that this feature ignores the user-configured offheap memory.

zhztheplayer:
> > Be careful, the dynamic offheap feature is experimental.
> 
> Thank you for reminding me, yea, I was just testing the feature in a test environment, and I found that this feature ignores the user-configured offheap memory.

Noted. So the work is related to https://github.com/apache/incubator-gluten/issues/7605? Since I didn't see the tickets linked.

yikf:
> > > Be careful, the dynamic offheap feature is experimental.
> > 
> > 
> > Thank you for reminding me, yea, I was just testing the feature in a test environment, and I found that this feature ignores the user-configured offheap memory.
> 
> Noted. So the work is related to #7605? Since I didn't see the tickets linked.

I didn't notice it before, but I understand the magic of https://github.com/apache/incubator-gluten/pull/5439 lies in using an on-heap memory quota to control on-heap and off-heap memory. Off-heap memory is floating and reserved for OFFHEAP_SIZING_MEMORY_FRACTION to prevent overlapping behavior when on-heap and off-heap memory are allocated simultaneously. 

https://github.com/apache/incubator-gluten/issues/7605 may have been intentionally designed by https://github.com/apache/incubator-gluten/pull/5439, but the current quota ignores the specified off-heap memory. This PR is used to make the quota's memory include the user-specified off-heap+on-heap.

supermem613:
Right - we purposefully ignore the user-configured offheap memory because we are going to size it based on the executor memory set. I don't think we should consider a manually considered off-heap setting as additional memory space we should use, especially if there may be use of off-heap memory outside of Gluten - which the user would be setting it up for... since this feature would preclude manually setting that, no?

supermem613:
Putting it differently - if what we are looking for is the capability to reduce the dynamic memory allocated with an offset, not just a factor, we should add a setting that does that, not use the off-heap setting.

yikf:
> Right - we purposefully ignore the user-configured offheap memory because we are going to size it based on the executor memory set. I don't think we should consider a manually considered off-heap setting as additional memory space we should use, especially if there may be use of off-heap memory outside of Gluten - which the user would be setting it up for... since this feature would preclude manually setting that, no?

Thank you for your explanation. If a user specifies "onheap 10G offheap 5G", was the available memory for the app 15G before, and now only 10G? If I missed something, please correct me.

supermem613:
Fundamentally, if the dynamic sizing off-heap memory feature is on, the user manually setting the off-heap size - which will be overwritten - could only be interpreted in two ways:

(1) The user wants to manually control the overall off-heap memory usage (Gluten + any other uses).
(2) The user wants to use off-heap memory outside of Gluten, say, for a custom JAR. 

For #1, it directly conflicts with the feature. Either one manually sets the off-heap or one lets the feature do its thing.

For #2, the current implementation of the feature is counter-productive as it will size the off-heap based solely on Gluten calculation and will use it ignoring any other user off-heap usage. This scenario is essentially not well addressed right now. I'd argue that really, we should separate the off-heap tracking in Gluten vs. Spark off-heap (maybe only when the dynamic sizing feature is on), so that customers can choose to set aside off-heap for their custom JARs, etc, and separately set the memory for Gluten OR use the dynamic sizing feature, but without conflicting with the off-heap setting.

Now, going back to your example of "onheap 10G offheap 5G". Fundamentally the question is what the use is attempting to do. If the user wants to use 10GB for JVM and 5GB for Gluten + custom JARs, it would do so with the dynamic sizing feature off (scenario #1 above). If they want 10GB for JVM or Gluten and 5GB for custom JARs, we hit the #2 scenario above - so right now they would turn off the dynamic sizing feature.

Overall, IMO, this is a strong argument for pursuing next what I described above - separating off-heap usage, perhaps only when the dynamic sizing feature is on, for Gluten vs. everything else. This would cover the user scenario of setting aside memory for things other than Gluten nicely.

yikf:
@supermem613 Thank you for your very detailed explanation. you are right. thanks!

PR #7818, opened 2024-11-05
[VL] Remove load shared libhdfs

Yohahaha:

velox does not depend on libhdfs.


JkSelf (approved):
LGTM. Thanks for your fix.

PR #7497, opened 2024-10-12
[GLUTEN-7495][BUILD] Fix macOS does not support version-script

zml1206:
## What changes were proposed in this pull request?

Fixes: \#7495

## How was this patch tested?

Local build test.

PR #7820, opened 2024-11-05
[VL] Do not use --version-script link option on Darwin

PHILO-HE:

This is a follow-up pr for [this comment](https://github.com/apache/incubator-gluten/pull/7497#issuecomment-2408417284).

If not hidden, the below error can be reported when running gluten CPP test.
```
something wrong with flag 'velox_memory_num_shared_leaf_pools' in file '/__w/incubator-gluten/incubator-gluten/ep/build-velox/build/velox_ep/velox/flag_definitions/flags.cpp'.  One possibility: file '/__w/incubator-gluten/incubator-gluten/ep/build-velox/build/velox_ep/velox/flag_definitions/flags.cpp' is being linked both statically and dynamically into this executable.
CMake Error at /usr/local/lib64/python3.9/site-packages/cmake/data/share/cmake-3.28/Modules/GoogleTestAddTests.cmake:112 (message):
Error running test executable.

  Path: '/__w/incubator-gluten/incubator-gluten/cpp/build/velox/tests/velox_shuffle_writer_test'
```
`velox_memory_num_shared_leaf_pools` is declared in Velox under `google` namespace. See `DEFINE_int32` used by [this velox code](https://github.com/facebookincubator/velox/blob/main/velox/flag_definitions/flags.cpp#L21). Even though `local: *;` is defined for version script, we have to explicitly make all symbols under `*google::*` namespace hidden. The possible reason is that the specified global `*facebook::velox::*` symbols cover those `*google::*` symbols.

## How was this patch tested?

CI.

### DIFF
--- a/cpp/velox/CMakeLists.txt
+++ b/cpp/velox/CMakeLists.txt
@@ -197,7 +197,7 @@ endif()
 
 add_library(velox SHARED ${VELOX_SRCS})
 
-if(ENABLE_GLUTEN_VCPKG)
+if(NOT CMAKE_SYSTEM_NAME MATCHES "Darwin")
   # Hide symbols of static dependencies
   target_link_options(
     velox PRIVATE -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/symbols.map)
--- a/cpp/velox/CMakeLists.txt
+++ b/cpp/velox/CMakeLists.txt
@@ -197,8 +197,8 @@ endif()
 
 add_library(velox SHARED ${VELOX_SRCS})
 
-if(NOT CMAKE_SYSTEM_NAME MATCHES "Darwin")
-  # Hide symbols of static dependencies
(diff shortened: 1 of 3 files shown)


PHILO-HE:
@zml1206, could you take a look? This is a follow-up for a comment in your merged pr.

zml1206 (approved):
LGTM, thank you.

PHILO-HE:
Just found some other issue. Let's hold on.

PHILO-HE:
This pr once intended to use one same symbols.map set for version-script in both static build and dynamic build. But it seems we cannot achieve this. For static build (with vcpkg), the below symbol patterns need to be put inside `extern "C++"` block. But for dynamic build, they should be moved out from this block. Otherwise, symbol conflict issue will be reported when running gluten cpp tests. One possible root cause is the use of different linking tool or linking options.
```
local:
    extern "C++" {
      *fL*::*;
      *google::*;
    };
```
As currently dynamic build has no symbol conflict so far, let's simply exclude the use of symbols.map in this build path.

rui-mo (approved):

PR #7836, opened 2024-11-07
Revert "[GLUTEN-7800][VL] Add config for max reclaim wait time to avoid dead lock when memory arbitration"

zhztheplayer:
Reverts apache/incubator-gluten#7799

Due to https://github.com/apache/incubator-gluten/pull/7799#issuecomment-2461091637


Yohahaha (approved):
I will try re-apply this PR, thank you!

Issue #7850, opened 2024-11-07
[VL] native writer support `CreateHiveTableAsSelectCommand`


yikf:
### Description

native writer support `CreateHiveTableAsSelectCommand`

PR #7851, opened 2024-11-07
[GLUTEN-7850][VL] Native writer support CreateHiveTableAsSelectCommand

yikf:
https://github.com/apache/incubator-gluten/issues/7850

## How was this patch tested?
unit tests

### DIFF
--- a/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
@@ -176,4 +176,13 @@ class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
       checkAnswer(sql("SELECT * FROM t"), Row(1))
     }
   }
+
+  test("native writer support CreateHiveTableAsSelectCommand") {
+    withTable("t") {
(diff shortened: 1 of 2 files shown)


yikf:
cc @jackylee-ch

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
```diff
@@ -176,4 +176,13 @@ class VeloxParquetWriteForHiveSuite extends GlutenQueryTest with SQLTestUtils {
       checkAnswer(sql("SELECT * FROM t"), Row(1))
     }
   }
+
+  test("native writer support CreateHiveTableAsSelectCommand") {
+    withTable("t", "t2") {
```
jackylee-ch, line 181: `t2` is not used in the test, it should be removed.
BTW, `create table stored as orc` should also be needed in this test.

yikf: removed, this `VeloxParquetWriteForHiveSuite` suite is specifically for Parquet, and the code paths for ORC and Parquet are consistent.

jinchengchenghh (approved):

jackylee-ch (approved):

Issue #7837, opened 2024-11-07
[VL] VeloxBackend should know it run in executor or driver


leoluan2009:
### Description

VeloxBackend show know where it run, executor or driver? for example if if run driver ,it should not init velox cache. There are two methods to this enhancement:
1. pass a config spark.gluten.isDriver when creating VeloxBackend instance, this is simple.
2. Add a member variables in VeloxBackend class, need to change jni code.

leoluan2009:
@zhztheplayer   @zhouyuan  can you give some thoughts? thanks!

zhztheplayer:
Driver and executor do have different plugin entrypoints, https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala, are you suggesting a new approach?

leoluan2009:
> Driver and executor do have different plugin entrypoints, https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala, are you suggesting a new approach?

But in VeloxBackend.cc, we can not know where it run

zhztheplayer:
> > Driver and executor do have different plugin entrypoints, https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala, are you suggesting a new approach?
> 
> But in VeloxBackend.cc, we can not know where it run. The info do not pass from java code to cpp code

I see. Do you know which part of C++ code requires for this information?

leoluan2009:
> > > Driver and executor do have different plugin entrypoints, https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala, are you suggesting a new approach?
> > 
> > 
> > But in VeloxBackend.cc, we can not know where it run. The info do not pass from java code to cpp code
> 
> I see. Do you know which part of C++ code requires for this information?

if it run in driver, it should not init velox cache. https://github.com/apache/incubator-gluten/blob/main/cpp/velox/compute/VeloxBackend.cc#L197

zhztheplayer:
I am curious why it matters to initialize the cache in driver or not. Do you already see some issues or errors in your circumstance? 

[...]

PR #7853, opened 2024-11-07
[GLUTEN-7837][VL] Spark driver should not initialize cache if not in local mode

leoluan2009:

Fixes: https://github.com/apache/incubator-gluten/issues/7837

Spark driver should not initialize cache if not in local mode

## How was this patch tested?

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
@@ -89,7 +89,7 @@ class VeloxListenerApi extends ListenerApi with Logging {
 
     SparkDirectoryUtil.init(conf)
     UDFResolver.resolveUdfConf(conf, isDriver = true)
-    initialize(conf)
+    initialize(conf, isDriver = true)
     UdfJniWrapper.registerFunctionSignatures()
   }
 
@@ -116,12 +116,12 @@ class VeloxListenerApi extends ListenerApi with Logging {
 
     SparkDirectoryUtil.init(conf)
     UDFResolver.resolveUdfConf(conf, isDriver = false)
-    initialize(conf)
+    initialize(conf, isDriver = false)
   }
 
   override def onExecutorShutdown(): Unit = shutdown()
 
-  private def initialize(conf: SparkConf): Unit = {
+  private def initialize(conf: SparkConf, isDriver: Boolean): Unit = {
     // Force batch type initializations.
     VeloxBatch.getClass
     ArrowJavaBatch.getClass
@@ -157,7 +157,11 @@ class VeloxListenerApi extends ListenerApi with Logging {
     }
 
     // Initial native backend with configurations.
-    val parsed = GlutenConfigUtil.parseConfig(conf.getAll.toMap)
+    var parsed = GlutenConfigUtil.parseConfig(conf.getAll.toMap)
+    // Workaround for https://github.com/apache/incubator-gluten/issues/7837
+    if (isDriver && !inLocalMode(conf)) {
+      parsed += (GlutenConfig.COLUMNAR_VELOX_CACHE_ENABLED.key -> "false")
+    }
     NativeBackendInitializer.initializeBackend(parsed)
 
     // Inject backend-specific implementations to override spark classes.


leoluan2009:
@zhztheplayer  can you help review?

zhztheplayer:
@leoluan2009 Does this solve the cache issue mentioned in https://github.com/apache/incubator-gluten/issues/7837?

leoluan2009:
> @leoluan2009 Does this solve the cache issue mentioned in #7837?

Yes, if VeloxBackend is running in driver, it will not init cache.

yikf (approved):

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -195,7 +195,7 @@ void VeloxBackend::initJolFilesystem() {
 }
 
 void VeloxBackend::initCache() {
-  if (backendConf_->get<bool>(kVeloxCacheEnabled, false)) {
+  if (backendConf_->get<bool>(kVeloxCacheEnabled, false) && !isDriver_) {
```
zhztheplayer, line 198: Could just change from the caller code, something like:

```
if (!isDriver_) {
  initCache();
}
...
```

Would be enough and more readable.

zhztheplayer (commented):
How does the change work with Spark local mode? I mean we may need to turn the cache on in local mode if user doesn't explicitly disable it.

cpp/velox/compute/VeloxBackend.cc
leoluan2009: Done

zhztheplayer:
Hi @leoluan2009 , as `spark.gluten.sql.columnar.backend.velox.cacheEnabled` is now a static SQL conf, can we just modify the config map passed to native to set it to false? It's a workaround and we can remove once a optimal design is out.

leoluan2009:
> Hi @leoluan2009 , as `spark.gluten.sql.columnar.backend.velox.cacheEnabled` is now a static SQL conf, can we just modify the config map passed to native to set it to false from non-local driver side? It's a workaround and we can remove once a optimal design is out.

it's ok and I will update the pr

leoluan2009:
@zhztheplayer  I have updated this PR, please help to review, thanks!

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -157,7 +155,10 @@ class VeloxListenerApi extends ListenerApi with Logging {
     }
 
     // Initial native backend with configurations.
-    val parsed = GlutenConfigUtil.parseConfig(conf.getAll.toMap)
+    var parsed = GlutenConfigUtil.parseConfig(conf.getAll.toMap)
+    if (isDriver && !inLocalMode(conf)) {
+      parsed += (GlutenConfig.COLUMNAR_VELOX_CACHE_ENABLED.key -> "false")
```
zhztheplayer, line 160: Would you like to add some comments in code? Or just leave a link to the relevant issue/PR. Thanks.

zhztheplayer (approved):
:+1

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -26,7 +26,6 @@ import org.apache.gluten.init.NativeBackendInitializer
 import org.apache.gluten.jni.{JniLibLoader, JniWorkspace}
 import org.apache.gluten.udf.UdfJniWrapper
 import org.apache.gluten.utils._
-
```
jackylee-ch, line 29: This line should not be removed.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -36,7 +35,6 @@ import org.apache.spark.sql.execution.datasources.velox.{VeloxOrcWriterInjects,
 import org.apache.spark.sql.expression.UDFResolver
 import org.apache.spark.sql.internal.{GlutenConfigUtil, StaticSQLConf}
 import org.apache.spark.util.{SparkDirectoryUtil, SparkResourceUtil}
-
```
jackylee-ch, line 39: ditto

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
leoluan2009: Done, thanks!

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
leoluan2009: Done, thanks!

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
leoluan2009: Done, thanks

leoluan2009:
@zhztheplayer  I resolved all comments

FelixYBW:
@zhztheplayer @leoluan2009
can you go through all the initializations there? Is there any other things we shouldn't run on driver? Is the local cache only exception? I'd expect the initialization should be much different.

If we have more than one, we should considered to seperate the initialization routine for driver and worker.

zhztheplayer:
> can you go through all the initializations there? Is there any other things we shouldn't run on driver? Is the local cache only exception? I'd expect the initialization should be much different.

Not much difference so far and it's also about our C++ API's encapsulation. We can make trade-offs once it's really needed.

zhztheplayer:
> If we have more than one, we should considered to seperate the initialization routine for driver and worker.

In future there will be different ways to implement differentiated initialization, but exposing Spark arch conceptions "driver" /  "executor" / "task" to C++ code could be the last ones to consider. We can use configurations, or compose finer-grained JNI initialization APIs, etc.

FelixYBW:
> In future there will be different ways to implement differentiated initialization, but exposing Spark arch conceptions "driver" / "executor" / "task" to C++ code could be the last ones to consider. We can use configurations, or compose finer-grained JNI initialization APIs, etc.

Let's re-evaluate the mythology once we have another exception.

Issue #7856, opened 2024-11-08
[CORE] To ensure the correct use of the GlutenCostEvaluator


weixiuli:
### Backend

VL (Velox)

### Bug description

The GlutenConfig is not initialized by sparkConf when calling setPredefinedConfigs. As a result, GlutenConfig.getConf.enableGluten and GlutenConfig.getConf.enableGlutenCostEvaluator in setPredefinedConfigs only retrieve their default values. To ensure correct configuration, we should retrieve these settings directly from sparkConf.

### Spark version

None

### System information

Gluten Version: 1.3.0-SNAPSHOT
Commit: e204dc5f06eef147d88b5ae6e8532b3902b19153
CMake Version: 3.20.2
System: Linux-5.4.119-19-0009.11
Arch: x86_64
CPU Name: Model name:          Intel(R) Xeon(R) Platinum 8372HC CPU @ 3.40GHz
BIOS Model name:     3.0
C++ Compiler: /usr/local/bin/g++
C++ Compiler Version: 10.3.1
C Compiler: /usr/bin/cc
C Compiler Version: 8.5.0
CMake Prefix Path: /usr/local;/usr;/;/usr;/usr/local;/usr/X11R6;/usr/pkg;/opt

PR #7857, opened 2024-11-08
[GLUTEN-7856][CORE] Ensure correct enabling of GlutenCostEvaluator

weixiuli:

The GlutenConfig is not initialized by sparkConf when calling setPredefinedConfigs. As a result, GlutenConfig.getConf.enableGluten and GlutenConfig.getConf.enableGlutenCostEvaluator in setPredefinedConfigs only retrieve their default values. To ensure correct configuration, we should retrieve these settings directly from sparkConf.

(Fixes: https://github.com/apache/incubator-gluten/issues/7856)

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -134,13 +134,8 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
     conf.set(SPARK_SESSION_EXTENSIONS.key, extensions)
 
     // adaptive custom cost evaluator class
-    val enableGlutenCostEvaluator = conf.getBoolean(
-      GlutenConfig.GLUTEN_COST_EVALUATOR_ENABLED,
-      GLUTEN_COST_EVALUATOR_DEFAULT_VALUE)
-    if (enableGlutenCostEvaluator) {
-      val costEvaluator = "org.apache.spark.sql.execution.adaptive.GlutenCostEvaluator"
-      conf.set(SQLConf.ADAPTIVE_CUSTOM_COST_EVALUATOR_CLASS.key, costEvaluator)
-    }
+    val costEvaluator = "org.apache.spark.sql.execution.adaptive.GlutenCostEvaluator"
+    conf.set(SQLConf.ADAPTIVE_CUSTOM_COST_EVALUATOR_CLASS.key, costEvaluator)
 
     // check memory off-heap enabled and size
     val minOffHeapSize = "1MB"
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -134,11 +134,10 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
     conf.set(SPARK_SESSION_EXTENSIONS.key, extensions)
 
     // adaptive custom cost evaluator class
-    val glutenEnabled = conf.getBoolean(GlutenConfig.GLUTEN_ENABLED_KEY, GLUTEN_ENABLED_BY_DEFAULT)
-    lazy val enableGlutenCostEvaluator = conf.getBoolean(
+    val enableGlutenCostEvaluator = conf.getBoolean(
       GlutenConfig.GLUTEN_COST_EVALUATOR_ENABLED,
       GLUTEN_COST_EVALUATOR_DEFAULT_VALUE)
-    if (glutenEnabled && enableGlutenCostEvaluator) {
+    if (enableGlutenCostEvaluator) {
       val costEvaluator = "org.apache.spark.sql.execution.adaptive.GlutenCostEvaluator"
       conf.set(SQLConf.ADAPTIVE_CUSTOM_COST_EVALUATOR_CLASS.key, costEvaluator)
     }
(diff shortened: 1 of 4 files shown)


gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -134,7 +134,11 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
...
     // adaptive custom cost evaluator class
-    if (GlutenConfig.getConf.enableGluten && GlutenConfig.getConf.enableGlutenCostEvaluator) {
+    val glutenEnabled = conf.getBoolean(GlutenConfig.GLUTEN_ENABLED_KEY, GLUTEN_ENABLED_BY_DEFAULT)
+    lazy val enableGlutenCostEvaluator = conf.getBoolean(
+      GlutenConfig.GLUTEN_COST_EVALUATOR_ENABLED,
+      GLUTEN_COST_EVALUATOR_DEFAULT_VALUE)
+    if (glutenEnabled && enableGlutenCostEvaluator) {
       val costEvaluator = "org.apache.spark.sql.execution.adaptive.GlutenCostEvaluator"
       conf.set(SQLConf.ADAPTIVE_CUSTOM_COST_EVALUATOR_CLASS.key, costEvaluator)
     }
```
zhztheplayer, line 144: It turns out that this part of code is problematical.

The entrypoint `GlutenPlugin` is called only once in the same Spark session. So say user could set `spark.gluten.enabled=false` then set it to true using `SET` prompt after a while.

Will there be any issue if we just make `GlutenCostEvaluator` always load? Which means removing the conditions, namely `if (glutenEnabled && enableGlutenCostEvaluator)` here.

cc @zml1206

zml1206: There seems to be no problem with always loading GlutenCostEvaluator, because GlutenCostEvaluator will determine whether gluten is enabled.

weixiuli: should we only retain the `enableGlutenCostEvaluator`  condition whether to enable gluten?  @zml1206 @zhztheplayer

zhztheplayer: > Should we only retain the `enableGlutenCostEvaluator` condition, no matter whether to enable gluten? @zml1206 @zhztheplayer

Yes, let's try remove the if conditions.

shims/spark33/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
```diff
@@ -26,7 +26,7 @@ import org.apache.spark.sql.internal.SQLConf
 case class GlutenCostEvaluator() extends CostEvaluator with SQLConfHelper {
   override def evaluateCost(plan: SparkPlan): Cost = {
     val forceOptimizeSkewedJoin = conf.getConf(SQLConf.ADAPTIVE_FORCE_OPTIMIZE_SKEWED_JOIN)
-    if (GlutenConfig.getConf.enableGluten) {
+    if (GlutenConfig.getConf.enableGlutenCostEvaluator) {
```
zml1206, line 29: We need to use `GlutenConfig.getConf.enableGluten` to determine whether to use vanilla spark cost or gluten cost.

shims/spark34/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
```diff
@@ -26,7 +26,7 @@ import org.apache.spark.sql.internal.SQLConf
 case class GlutenCostEvaluator() extends CostEvaluator with SQLConfHelper {
   override def evaluateCost(plan: SparkPlan): Cost = {
     val forceOptimizeSkewedJoin = conf.getConf(SQLConf.ADAPTIVE_FORCE_OPTIMIZE_SKEWED_JOIN)
-    if (GlutenConfig.getConf.enableGluten) {
+    if (GlutenConfig.getConf.enableGlutenCostEvaluator) {
```
zml1206, line 29: ditto.

shims/spark33/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
weixiuli:
```
    if (GlutenConfig.getConf.enableGluten && GlutenConfig.getConf.enableGlutenCostEvaluator) {
      new GlutenCost(SimpleCostEvaluator(forceOptimizeSkewedJoin), plan)
    } else {
      SimpleCostEvaluator(forceOptimizeSkewedJoin).evaluateCost(plan)
    }
```
 right ?

zml1206: I understand there is no need to change here, `enableGlutenCostEvaluator` should be a static config.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -2107,17 +2103,6 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
-  val COST_EVALUATOR_ENABLED =
-    buildConf(GlutenConfig.GLUTEN_COST_EVALUATOR_ENABLED)
```
zml1206, line 2111: I mean update `buildConf` to `buildStaticConf`.

zml1206 (approved):
LGTM except for some small comments, thank you.
cc @zhztheplayer

weixiuli:
> @weixiuli Can we have a better PR title for this? Thanks!

Done.

Issue #7866, opened 2024-11-08
[CH] fallback for unsupported regex in re2


exmy:
### Backend

CH (ClickHouse)

### Bug description

OptimizedRegularExpression: cannot compile re2: d(?!d), error: invalid perl operator: (?!. Look at https://github.com/google/re2/wiki/Syntax for reference.

### Spark version

None

FelixYBW:
Does CH also use re2?

@PHILO-HE we may reuse the preprocessing script as well as the skip list.

PHILO-HE:
> @PHILO-HE we may reuse the preprocessing script as well as the skip list.

@FelixYBW, currently, we have no script, but have a pre-validation function on Gluten native side. It validate all listed regrex functions by letting RE2 try to compile the pattern. If the compilation fails, we definitely need to make it fall back. This pre-validation can be reused by both backend.

FelixYBW:
> @FelixYBW, currently, we have no script, but have a pre-validation function on Gluten native side. It validate all listed regrex functions by letting RE2 try to compile the pattern. If the compilation fails, we definitely need to make it fall back. This pre-validation can be reused by both backend.

I remember we have some preprocessing of the pattern before we send to re2, which can make more pattern workable on re2. Is the code still there?

PHILO-HE:
> I remember we have some preprocessing of the pattern before we send to re2, which can make more pattern workable on re2. Is the code still there?

@FelixYBW, I just found Velox has such preprocessing for presto: see [code](https://github.com/facebookincubator/velox/blob/main/velox/functions/prestosql/RegexpReplace.h#L33). And Meituan also proposed a pr to re-use this code for Spark and also make some improvement. See https://github.com/facebookincubator/velox/pull/10981.

PR #7867, opened 2024-11-08
[GLUTEN-7866][CH] Fallback for unsupported regex in re2

exmy:

(Fixes: \#7866)

## How was this patch tested?

add ut

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/utils/CHExpressionUtil.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/utils/CHExpressionUtil.scala
@@ -32,6 +32,14 @@ object FunctionValidator {
     case DateType | TimestampType => true
     case _ => false
   }
+
+  // https://github.com/google/re2/wiki/Syntax
+  private val unsupportedRe2Patterns = Set("(?=", "(?!", "(?<=", "(?<!")
+
+  def validateRe2(pattern: Expression): Boolean = {
+    val p = pattern.eval().asInstanceOf[UTF8String].toString
+    !unsupportedRe2Patterns.exists(p.contains(_))
+  }
 }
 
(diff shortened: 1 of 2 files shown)


PHILO-HE:
@exmy, I'm not familiar with CH backend code. In velox backend, we have a validation function on native side (see below link). In its implementation, RE2 tries to compile literal regex pattern, and then some known unsupported cases are filter out  even though compiling can pass. Can this handling be a reference to you to fix your issues?

https://github.com/apache/incubator-gluten/blob/main/cpp/velox/utils/Common.cc#L55

exmy:
> @exmy, I'm not familiar with CH backend code. In velox backend, we have a validation function on native side (see below link). In its implementation, RE2 tries to compile literal regex pattern, and then some known unsupported or incompatible cases are filter out even though compiling can pass. Can this handling be a reference to you to fix your issues?
> 
> https://github.com/apache/incubator-gluten/blob/main/cpp/velox/utils/Common.cc#L55

Thanks for advice! It's a indeed better way to validate on native side, but ch backend currently does not support it. Maybe this way wil be adopted later as well.

BTW, I noticed it seems missing validate for `split` expression in https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc#L56. The second parameter for `split` is also a regex string.

PHILO-HE:
> BTW, I noticed it seems missing validate for `split` expression in https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc#L56. The second parameter for `split` is also a regex string.

@exmy, thanks for letting me know this! Currently, this function's implementation is not fully ready in Velox. I have created an issue to track this: https://github.com/apache/incubator-gluten/issues/7872.

Issue #7890, opened 2024-11-11
[UI] Optimize cleanup gluten sql executions ui data


zml1206:
### Description

There is a phenomenon in our production environment. After running for a period of time, only a few gluten sql executions ui data will be retained. We are resident in spark, and there are some queries that disable gluten.
The reason is that the current cleanup of gluten sql executions ui data depends on SQLExecutionUIData. There is a problem here. When sql execution disables gluten, gluten sql executions ui data will not be generated, but cleanup will still be triggered.

gaoyangxiaozhu:
curious if we disable gluten， why the event of `GlutenSQLAppStatusListener` still be triggerred ?

zml1206:
> curious if we disable gluten， why the event of `GlutenSQLAppStatusListener` still be triggerred instead of been disabled also?


`GlutenSQLAppStatusListener` is registered by `GlutenPlugin`, this is done by driver initialization.

gaoyangxiaozhu:
> > curious if we disable gluten， why the event of `GlutenSQLAppStatusListener` still be triggerred instead of been disabled also?
> 
> `GlutenSQLAppStatusListener` is registered by `GlutenPlugin`, this is done by driver initialization.

thanks @zml1206 , so i am just wondering does it possbile to let glutenPlugin do nothing (at least not register gluten listener) if gluten is been disabled by `spark.gluten.enabled` conf ?

zml1206:
> so i am just wondering does it possbile to let glutenPlugin do nothing (at least not register gluten listener) if gluten is been disabled by `spark.gluten.enabled` conf ?

AFAIK, this is inappropriate because `spark.gluten.enabled` is a dynamic configuration, while listener is static and cannot be changed after initialization.

PR #7891, opened 2024-11-11
[GLUTEN-7890][UI] Optimize cleanup gluten sql executions ui data

zml1206:

(Fixes: \#7890)

## How was this patch tested?

local test.


PHILO-HE:
I'm not sure on this. @ulysses-you, could you take a look?

gluten-ui/src/main/scala/org/apache/spark/sql/execution/ui/GlutenSQLAppStatusListener.scala
```diff
@@ -52,7 +52,7 @@ private class GlutenSQLAppStatusListener(conf: SparkConf, kvstore: ElementTracki
         event.physicalPlanDescription,
         event.fallbackNodeToReason.toSeq.sortBy(_._1)
       )
-      kvstore.write(uiData)
+      kvstore.write(uiData, checkTriggers = true)
```
ulysses-you, line 55: I'm afraid it would introduce overhead for driver since each aqe stage would post one event. I think fire the trigger at  `onSQLExecutionStart` is enough.

zml1206: That's a good suggestion for me, thank you, fixed.

ulysses-you (approved):

PR #7897, opened 2024-11-11
[GLUTEN-6896] Add buffered read for hash/sort shuffle reader

marin-ma:
The deserialization process in native shuffle reader contains lots of small reads. E.g. read only 4 or 8 bytes each time to get the buffer metadata. We should use buffered read to avoid small I/O.

  | main | PR 7897
-- | -- | --
runtime | 1,029.37 | 1,004.93
avg %user | 61.2 | 63.1
avg %system | 17.2 | 16.6
avg %iowait | 5.5 | 4.8
total read (G) | 3,093.0 | 2,720.3
avg read bw (MB/s) | 3,249.7 | 2,704.4
pathlength_sum | 1,201,191 | 1,182,288


cpp/velox/shuffle/VeloxShuffleReader.h
```diff
@@ -85,7 +86,50 @@ class VeloxSortShuffleReaderDeserializer final : public ColumnarBatchIterator {
...
+      ScopedTimer timer(&readStreamTime_);
+      return in_->Read(nbytes);
+    }
+
+    int64_t readStreamTime() const {
+      return readStreamTime_;
+    }
+
+   private:
+    std::shared_ptr<arrow::io::InputStream> in_;
```
jinchengchenghh, line 127: const

FelixYBW:
@marin-ma can you add some desc what's the issue the PR is going to solve?

zhouyuan (commented):
+1 on performance

FelixYBW:
Does vanilla spark have the same issue?

marin-ma:
> Does vanilla spark have the same issue?

@FelixYBW Spark used `BufferedInputStream` with a fixed size of 8192. https://github.com/apache/spark/blob/branch-3.2/sql/core/src/main/scala/org/apache/spark/sql/execution/UnsafeRowSerializer.scala#L102

I've tested with 16k buffer size for this patch, but there's not much perf gain. The performance in the PR description uses 1M buffer size.

FelixYBW (approved):

Issue #7900, opened 2024-11-12
[VL] Enable prefix sort in velox spill by config


jinchengchenghh:
### Description

velox use config `spill_enable_prefix_sort` to control whether use prefix sort or timsort in spill.
Prefix sort requires extra memory for prefix data but gets 2x times acceleration. It does not support string and complex data type now, fallback to std::sort in that case.

FelixYBW:
how much impact to Gluten spill performance?

PR #7904, opened 2024-11-12
[GLUTEN-7900][VL] Enable prefix sort config in spill

jinchengchenghh:
Prefix sort can reduce the spill sort time by 3x with the sampled Meta production query but timsort increase the sort time by 20%.  So timsort performance seems depends on the actual data pattern.
After pick this PR  https://github.com/facebookincubator/velox/pull/11527, a query in jenkins spill with string is 41s prefixsort vs 37s timsort vs 63s stdsort.

Config `spark.gluten.sql.columnar.backend.velox.spillPrefixsortEnabled` disable in this PR, enable it as default after test.

Relevant Velox PR: https://github.com/facebookincubator/velox/pull/11384
Resolves https://github.com/apache/incubator-gluten/issues/7900

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -487,7 +487,7 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
     configs[velox::core::QueryConfig::kSpillCompressionKind] =
         veloxCfg_->get<std::string>(kSpillCompressionKind, "lz4");
     configs[velox::core::QueryConfig::kSpillPrefixSortEnabled] =
-        veloxCfg_->get<std::string>(kSpillPrefixSortEnabled, "true");
+        veloxCfg_->get<std::string>(kSpillPrefixSortEnabled, "false");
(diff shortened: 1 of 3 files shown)


zhztheplayer:
Would you like to refer the related Velox PRs for this feature in PR description? Which will help users to track. Thanks.

I'm also thinking about setting up a more comprehensive integration benchmark for spill performance. 

The following is by the existing GHA oom tests:

Before:

image

After:

TBD (not completed)

jinchengchenghh:
We have a spill performance in internal Jenkins, I have trigger it. It will run after machine is ready since serves down today.

FelixYBW:
Let's put the jenkins performance here.

jinchengchenghh:
Can you help review again? Now it only enables the config.

zhztheplayer (approved):

boneanxs:
> a query in jenkins spill with string is 41s prefixsort vs 37s timsort vs 63s stdsort.

@jinchengchenghh Hi, is there any practice under what kind of data, timsort is better? Is there any reason we don't set prefix sort by default? Thanks!

jinchengchenghh:
If the data is approximately sorted, the timsort will performa better, we have a user case than can confirm that. @boneanxs

Issue #7750, opened 2024-10-31
[VL] Move `ColumnarBuildSideRelation`'s memory occupation to Spark off-heap


zhztheplayer:
### Description

So far `ColumnarBuildSideRelation` is allocated on Spark JVM heap memory.

```
case class ColumnarBuildSideRelation(output: Seq[Attribute], batches: Array[Array[Byte]]) {
  ...
}
```

It appears that we can replace `batches: Array[Array[Byte]]` with an off-heap allocated container to move the memory usage to off-heap. This could avoid some of the heap OOM issues.

zhouyuan:
CC @kecookier @NEUpanning

z-anderson:
Hi, I would like to do this ticket to learn more about Gluten. Could you please point me to where an off-heap allocated container is used in the code? Thank you.

Zand100:
Hi, I would like to do this ticket to learn more about Gluten. Could you please point me to where an off-heap allocated container is used in the code? Thank you.

zhztheplayer:
Thank you in advance for helping @Zand100 .

You can refer to vanilla Spark's code [UnsafeHashedRelation](https://github.com/apache/spark/blob/4771638b3a6873e329bb3bdde5abeeeac059acb8/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashedRelation.scala#L207) where `BytesToBytesMap` can be allocated from off-heap. Probably we can adopt a similar approach in Gluten.

So I can assign this ticket to you, I suppose?

FelixYBW:
@Zand100 You may consider to upstream the PR to upstream spark as well. It's another solution about offheap/onheap conflict, that we move all Spark's large memory allocation to offheap once offheap is enabled.

Zand100:
Thank you! 

Just to check I'm on the right track, I'm trying to use the `BytesToBytesMap` instead of `Array[Array[Bytes]]` in https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarBuildSideRelation.scala#L39. Is that right?

Yes, you can assign the ticket to me.

Zand100:
Hi, how should I handle constructing a `ColumnarBuildSideRelation` for the `HashedRelationBroadcastMode`  case? 

```
val fromBroadcast = from.asInstanceOf[Broadcast[HashedRelation]]
val fromRelation = fromBroadcast.value.asReadOnlyCopy()
```
[...]

PR #7902, opened 2024-11-12
[GLUTEN-7750][VL]:  store unsafe batches data

Zand100:
## What changes were proposed in this pull request?

Adds a binary container implementing a Spark `MemoryConsumer` to be used instead of `Array[Array[Bytes]]`.

(Fixes: \#7750)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #7944, opened 2024-11-14
[GLUTEN-7750][VL]: store unsafe batches data

Zand100:

Adds a binary container implementing a Spark MemoryConsumer to be used instead of Array[Array[Bytes]]. Uses that binary container with the existing object handles in `ColumnarBuildSideRelation`.

(Fixes: https://github.com/apache/incubator-gluten/issues/7750) 

Duplicate of https://github.com/apache/incubator-gluten/pull/7902

## How was this patch tested?

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/UnsafeArray.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/UnsafeArray.scala
@@ -94,8 +94,12 @@ class UnsafeArray(taskMemoryManager: TaskMemoryManager)
     var spillOffset = page.getBaseOffset
     var numRecords: Int = UnsafeAlignedOffset.getSize(spillBase, spillOffset) // probably wrong
     val writeMetrics = new ShuffleWriteMetrics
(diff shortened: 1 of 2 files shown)


backends-velox/src/main/scala/org/apache/spark/sql/execution/UnsafeArray.scala
```diff
@@ -0,0 +1,110 @@
...
+
+import org.apache.spark.SparkEnv
+import org.apache.spark.executor.ShuffleWriteMetrics
+import org.apache.spark.memory.{MemoryConsumer, MemoryMode, SparkOutOfMemoryError, TaskMemoryManager}
+import org.apache.spark.sql.catalyst.expressions.UnsafeArrayData
+import org.apache.spark.unsafe.{Platform, UnsafeAlignedOffset}
+import org.apache.spark.unsafe.memory.MemoryBlock
+import org.apache.spark.util.collection.unsafe.sort.UnsafeSorterSpillWriter
+
+class UnsafeArray(taskMemoryManager: TaskMemoryManager)
```
zhouyuan, line 27: @zhztheplayer 
should we use the mem pool from Gluten?

Issue #6666, opened 2024-08-01
[VL] Switch to use extractFiltersFromRemainingFilter in Velox


rui-mo:
### Description

There is some logic in Velox to extract subfield filters from call expression:
https://github.com/facebookincubator/velox/blob/f8e076ecaaffae0d7dc0e702bb6ad8aa767471b3/velox/connectors/hive/HiveConnectorUtil.cpp#L819

To switch to it from Gluten's implementation, we might need to do a full validation, then we can seamlessly transition to the Velox implementation.

PR #7946, opened 2024-11-14
[VL] Enable filter push-down on nested field

rui-mo:

We add a fix in https://github.com/oap-project/velox/commit/fc0e7ceda3b8a749bdffd7e30b6561ec0afb6405 for filter push-down on nested field to incorporate with schema pruning in table scan, and thus the nested subfield could be enabled in Gluten.

## How was this patch tested?

Spark `SchemaPruningSuite`.


zhouyuan:
related
https://github.com/apache/incubator-gluten/issues/6666

PR #7970, opened 2024-11-18
[GLUTEN-7969][VL] Enable spill to multiple directories for micro benchmark

marin-ma:


FelixYBW:
Current microbenchmark reports errors when the spill folder and file already exists. Can the PR fix it?

FelixYBW (approved):

PR #7996, opened 2024-11-19
[VL] fallback unsupported orc write for spark32 and spark33

jackylee-ch:
I have run bellow sqls failed with footer parse while reading orc file. After look inside, I found the file is written with parquet, not orc, and we haven't support the orc write now.
```
CREATE TABLE t (c int) USING ORC;
INSERT OVERWRITE TABLE t SELECT 1 as c;
select * from t;
```


jackylee-ch:
cc @PHILO-HE

PHILO-HE:
I'm not sure about this. Could you add some description? It has no impact on other Spark versions?

jackylee-ch:
> I'm not sure about this. Could you add some description? 

Sure, I would update the pr description.

> It has no impact on other Spark versions?

For spark32 and spark33, we use `GlutenWriterColumnarRules` to check the supported fileformats. But for spark34 and spark35, we use `BackendsApiManager.getSettings.supportWriteFilesExec` to check the supported file formats. Hense this won't impact other spark versions.

PHILO-HE (commented):
@JkSelf, please confirm. the changes.

JkSelf (approved):
Thanks for your fix. LGTM.

PR #725, opened 2022-12-16
Save stage output to parquet file

jinchengchenghh:

PR #7998, opened 2024-11-20
[GLUTEN-7953][VL] Fetch and dump all inputs for micro benchmark on middle stage begin

marin-ma:
Collect all input data and save it into a Parquet file. Then, read the data from the Parquet file to feed it into the pipeline.

Update `spark.gluten.sql.benchmark_task.partitionId` and `spark.gluten.sql.benchmark_task.taskId` to accept a comma-separated string of multiple partition ids/task ids

Currently, the input files are being dumped during the middle stage of execution. This PR will fetch all the data and save it into file before the pipeline starts. The stage input will then be read from this dumped file. This will help to save a complete input file even if the task fails. In this way, we can have the full input data of the failed task, and reproduce the failure with microbenchmark.


Yohahaha:
does this PR only works for middle stage? or dump the data comes from shuffle?

marin-ma:
@Yohahaha Yes. It only works for middle stage. Updated PR titile.

zhztheplayer:
@marin-ma Curious what's the improvements here, comparing to https://github.com/apache/incubator-gluten/pull/725? Thanks.

marin-ma:
> @marin-ma Curious what're the improvements here, comparing to #725? Thanks.

@zhztheplayer Currently, the input files are being dumped during the middle stage of execution. This PR will fetch all the data and save it into file before the pipeline starts. The stage input will then be read from this dumped file. This will help to save a complete input file even if the task fails. In this way, we can have the full input data of the failed task, and reproduce the failure with microbenchmark.

cpp/core/jni/JniCommon.cc
```diff
@@ -109,17 +109,30 @@ std::shared_ptr<gluten::ColumnarBatch> gluten::JniColumnarBatchIterator::next()
...
+        checkException(env);
+        auto batch = ObjectStore::retrieve<ColumnarBatch>(handle);
+
+        // Save the snapshot of the batch to file.
+        std::shared_ptr<ArrowSchema> schema = batch->exportArrowSchema();
+        std::shared_ptr<ArrowArray> array = batch->exportArrowArray();
+        auto rb = gluten::arrowGetOrThrow(arrow::ImportRecordBatch(array.get(), schema.get()));
+        GLUTEN_THROW_NOT_OK(writer_->initWriter(*(rb->schema().get())));
+        GLUTEN_THROW_NOT_OK(writer_->writeInBatches(rb));
+      } while (env->CallBooleanMethod(jColumnarBatchItr_, serializedColumnarBatchIteratorHasNext_));
```
jinchengchenghh, line 127: checkEnv after CallBooleanMethod

jinchengchenghh: We get all the batch, so the batch Iterator cannot be reused, will it affect the following stage

cpp/velox/operators/writer/VeloxArrowWriter.cc
```diff
@@ -0,0 +1,34 @@
...
+    facebook::velox::memory::MemoryPool* pool)
+    : ArrowWriter(path), batchSize_(batchSize), pool_(pool) {}
+
+std::shared_ptr<ColumnarBatch> VeloxArrowWriter::retrieveColumnarBatch() {
+  if (!reader_) {
+    reader_ = std::make_unique<ParquetStreamReaderIterator>(path_, batchSize_, pool_);
+  }
+  return reader_->next();
+}
+} // namespace gluten
```
jinchengchenghh, line 34: end empty line

cpp/core/jni/JniCommon.cc
marin-ma: @jinchengchenghh In what cases would the batch iterator be reused? Could you elaborate more on this?

FelixYBW (approved):
works. Thank you Rong.

PR #8001, opened 2024-11-20
[CORE] Minor: Use lower case for Maven profile names

zhztheplayer:
A minor cleanup. We'd use lower case on the first alphabet for Maven profile names.


zhztheplayer:
@PHILO-HE thanks

PHILO-HE (approved):

PR #8004, opened 2024-11-20
[CORE] Remove TPC-H test dataset from module backends-velox, use the one from gluten-core uniformly

zhztheplayer:
Use gluten-core/test/resources/tpch-data-sf100 uniformly in tests where tpch dataset is needed.


zhztheplayer:
@zzcclp This is the attempt to use Parquet data in `gluten-core` from `backend-velox`'s test code. But it seems that the resource files in core jar cannot be accessed from other modules.

I am thinking to introduce an env var like `GLUTEN_HOME` to pass to test suites to avoid specifying relevant paths in other modules. otherwise we'll continue breaking the module encapsulations.

backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenClickHouseDecimalSuite.scala
```diff
@@ -38,7 +38,7 @@ class GlutenClickHouseDecimalSuite
 
   override protected val needCopyParquetToTablePath = true
 
-  override protected val tablesPath: String = basePath + "/tpch-data"
+  override protected val tablesPath: String = basePath + "/tpch-data-sf100"
```
zzcclp, line 41: the sf of this tpch data is 1G, no 100G

zhztheplayer: Yes this was my mistake.

It's likely SF 0.1 but I am not sure, will confirm before merging.

PR #7268, opened 2024-09-18
[GLUTEN-7267][CH]Support nested column pruning for `HiveTableScan`  json/parquet/orc format

KevinyhZou:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#7267)

## How was this patch tested?

BY UT

PR #8008, opened 2024-11-20
[GLUTEN-7267][CORE][CH] Move schema pruning optimization of HiveTableScan to an individual post-transform rule

zhztheplayer:
Following https://github.com/apache/incubator-gluten/pull/7268, move the new code to an individual rule `PruneNestedColumnsInHiveTableScan`.

The reason is `HiveTableScanNestedColumnPruning` actually matches on a `project + hive scan` pattern so is not simply offloading a project. Move the code out from `OffloadSingleNode` which targets to handle one single query plan node.


zhztheplayer:
Hi @KevinyhZou @rui-mo Would you help review this, thanks.

KevinyhZou:
LGTM

rui-mo (approved):

Issue #8010, opened 2024-11-20
Metrics of  FilterExecTransformer are incorrect when remainingCondition is null


zml1206:
### Backend

VL (Velox)

### Bug description

<img width="469" alt="image" src="https://github.com/user-attachments/assets/0e5aff12-1173-45b3-b1b0-69a7621131d4">

When the remainingCondition is empty, the metrics of FilterExecTransformer are all 0.

### Spark version

None

PR #8011, opened 2024-11-20
[GLUTEN-8010][CORE] Don't generate native metrics if transformer don't generate relNode

zml1206:

(Fixes: \#8010)

## How was this patch tested?

Local test.
Before PR:
image
After PR:
image

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
@@ -242,15 +242,6 @@ object MetricsUtil extends Logging {
           val opMetrics: OperatorMetrics = mergeMetrics(operatorMetrics)
           lu.updateNativeMetrics(opMetrics)
         }
-      case fu: FilterMetricsUpdater if operatorMetrics.size() == 0 =>
-        val childOperatorMetrics = new JArrayList[OperatorMetrics]()
-        relMap
-          .get(operatorIdx - 1)
-          .forEach(
-            _ => {
-              childOperatorMetrics.add(metrics.getOperatorMetrics(curMetricsIdx))
-            })
-        fu.inheritChildMetrics(mergeMetrics(childOperatorMetrics))
       case u =>
         val opMetrics: OperatorMetrics = mergeMetrics(operatorMetrics)
         u.updateNativeMetrics(opMetrics)
--- a/backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
@@ -54,9 +54,6 @@ object MetricsUtil extends Logging {
           MetricsUpdaterTree(
             smj.metricsUpdater(),
             Seq(treeifyMetricsUpdaters(smj.bufferedPlan), treeifyMetricsUpdaters(smj.streamedPlan)))
-        case t: TransformSupport if t.metricsUpdater() == MetricsUpdater.None =>
-          assert(t.children.size == 1, "MetricsUpdater.None can only be used on unary operator")
-          treeifyMetricsUpdaters(t.children.head)
         case t: TransformSupport =>
           MetricsUpdaterTree(t.metricsUpdater(), t.children.map(treeifyMetricsUpdaters))
         case _ =>
@@ -219,6 +216,7 @@ object MetricsUtil extends Logging {
         })
 
     mutNode.updater match {
+      case MetricsUpdater.None =>
       case ju: HashJoinMetricsUpdater =>
         // JoinRel outputs two suites of metrics respectively for hash build and hash probe.
         // Therefore, fetch one more suite of metrics here.
--- a/backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
@@ -242,6 +242,15 @@ object MetricsUtil extends Logging {
           val opMetrics: OperatorMetrics = mergeMetrics(operatorMetrics)
           lu.updateNativeMetrics(opMetrics)
         }
+      case fu: FilterMetricsUpdater if operatorMetrics.size() == 0 =>
+        val childOperatorMetrics = new JArrayList[OperatorMetrics]()
+        relMap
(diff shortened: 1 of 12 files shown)


zhztheplayer:
Thanks!

Though it feels like a workaround since there is no real filter node in Velox query plan? Do we have other better options?

zml1206:
> Though it feels like a workaround since there is no real filter node in Velox query plan? Do we have other better options?

How about set metricsUpdater to `MetricsUpdater.None` when getRemainingCondition? 
In this way, there is no metrics information.
image

zml1206:
> BTW I think one of the promising ways is to exclude the filter node from Spark query plan as well. Though I remember there were some issues with an relevant attempt.

We can't exclude the filter node, because it's output may differ from child's output.

backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
```diff
@@ -54,9 +54,6 @@ object MetricsUtil extends Logging {
           MetricsUpdaterTree(
             smj.metricsUpdater(),
             Seq(treeifyMetricsUpdaters(smj.bufferedPlan), treeifyMetricsUpdaters(smj.streamedPlan)))
-        case t: TransformSupport if t.metricsUpdater() == MetricsUpdater.None =>
-          assert(t.children.size == 1, "MetricsUpdater.None can only be used on unary operator")
-          treeifyMetricsUpdaters(t.children.head)
```
zml1206, line 59: It will cause operatorId inconsistencies.

zml1206:
@zhztheplayer Can you help take a look again? Thanks.

backends-velox/src/main/scala/org/apache/gluten/metrics/MetricsUtil.scala
```diff
@@ -219,6 +216,7 @@ object MetricsUtil extends Logging {
         })
 
     mutNode.updater match {
+      case MetricsUpdater.None =>
```
zhztheplayer, line 219: Are we sure about this?

The code is likely something done by `MetricsUpdater.Todo`. I remember `MetricsUpdater.Todo` and `MetricsUpdater.None` don't share the same semantic. 

Did you check the UI? Are all the metrics still normal with this change?

zml1206: It doesn't seem to make any difference whether to add it or not.
Theoretically, the size of None's operatorMetrics is 0, and a judgment safety point is added.
Also, I didn't find any UI anomalies.

zml1206: I thought about it carefully later, and I should remove it, which can help us distinguish whether None is set incorrectly. If an UnsupportedOperationException is thrown later, it means that it should not be None

gluten-substrait/src/main/scala/org/apache/gluten/execution/ExpandExecTransformer.scala
```diff
@@ -112,13 +115,12 @@ case class ExpandExecTransformer(
 
   override protected def doTransform(context: SubstraitContext): TransformContext = {
     val childCtx = child.asInstanceOf[TransformSupport].transform(context)
-    val operatorId = context.nextOperatorId(this.nodeName)
-    if (projections == null || projections.isEmpty) {
+    if (metricsUpdater == MetricsUpdater.None) {
```
zhztheplayer, line 118: Hi @zml1206

Can we create a new utility method, e.g, `isNoop` that can be called both by `doTransform` and `metricUpdater` in each operator? Thanks.

zhztheplayer (commented):
Other code looks good to me.

gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -61,7 +61,9 @@ abstract class FilterExecTransformerBase(val cond: Expression, val input: SparkP
     case _ => false
   }
 
-  override def metricsUpdater(): MetricsUpdater = if (getRemainingCondition == null) {
+  override def isLoop: Boolean = getRemainingCondition == null
```
zhztheplayer: s/isLoop/isNoop/

gluten-substrait/src/main/scala/org/apache/gluten/execution/ExpandExecTransformer.scala
```diff
@@ -48,7 +48,9 @@ case class ExpandExecTransformer(
     AttributeSet.fromAttributeSets(projections.flatten.map(_.references))
   }
 
-  override def metricsUpdater(): MetricsUpdater = if (projections == null || projections.isEmpty) {
+  override def isLoop: Boolean = projections == null || projections.isEmpty
```
zhztheplayer: s/isLoop/isNoop/

gluten-substrait/src/main/scala/org/apache/gluten/execution/SortExecTransformer.scala
```diff
@@ -44,7 +44,9 @@ case class SortExecTransformer(
   @transient override lazy val metrics =
     BackendsApiManager.getMetricsApiInstance.genSortTransformerMetrics(sparkContext)
 
-  override def metricsUpdater(): MetricsUpdater = if (sortOrder == null || sortOrder.isEmpty) {
+  override def isLoop: Boolean = sortOrder == null || sortOrder.isEmpty
```
zhztheplayer: s/isLoop/isNoop/

gluten-substrait/src/main/scala/org/apache/gluten/execution/WindowExecTransformer.scala
```diff
@@ -51,9 +51,9 @@ case class WindowExecTransformer(
   @transient override lazy val metrics =
     BackendsApiManager.getMetricsApiInstance.genWindowTransformerMetrics(sparkContext)
 
-  override def metricsUpdater(): MetricsUpdater = if (
-    windowExpression == null || windowExpression.isEmpty
-  ) {
+  override def isLoop: Boolean = windowExpression == null || windowExpression.isEmpty
```
zhztheplayer: s/isLoop/isNoop/

zhztheplayer (approved):
One nit. Thanks!

PR #8015, opened 2024-11-21
[VL] Align the timezone of decimal partition value with Velox

rui-mo:

In this way we don't need additional change for Velox.

## How was this patch tested?

under test


rui-mo:
Hi @surnaik, this issue is about the timestamp partition value, which is converted from string in Velox. During this conversion, the default timezone (which is set as "America/Los Angeles" in Velox) is used. This PR tries to adopt the same timezone as Velox to see if we can remove this additional change: https://github.com/oap-project/velox/commit/567f7df756325dacb37b33efecd0a31ba95a7ff6.

https://github.com/facebookincubator/velox/blob/c395c553bd9627b4a3211e34f16bb1949bbcd43a/velox/connectors/hive/SplitReader.cpp#L58-L60 

https://github.com/facebookincubator/velox/blob/c395c553bd9627b4a3211e34f16bb1949bbcd43a/velox/type/Timestamp.cpp#L96-L101

rui-mo:
Replacing with a configuration in Velox to decide UTC or local time zone.

Issue #8025, opened 2024-11-22
[VL] Distinct aggregation OOM when getOutput


ccat3z:
### Backend

VL (Velox)

### Bug description

Distinct aggregation will merge all sorted spill file in `getOutput()` (`SpillPartition::createOrderedReader`). If there are too many spill files, reading the first batch of each file into memory will consume a significant amount of memory. In one of our internal cases, one task generated 300 spill files, which requires close to 3G of memory.

![image](https://github.com/user-attachments/assets/23dd540e-a4b7-448e-84e0-caae00aa5147)

### Spark version

None

ccat3z:
A possible workaround is to reduce `kSpillWriteBufferSize`. Why it is set to 4M by default? Is there any experience in performance tuning?

FelixYBW:
I can only find the configuration: spark.shuffle.spill.diskWriteBufferSize.

No spill merge one.

FelixYBW:
Thank you, @ccat3z . I encounted the same issue in order by operator and debugged several days!

jinchengchenghh:
The kSpillReadBufferSize controls the size to read each file, the ordered reader will create FileOutputStream for each file and allocate kSpillReadBufferSize (default 1MB) for one file. Can you try to adjust this value? 
Maybe we should add a new config to control the read buffer size for all the files. numFiles * bufferSize < threshold.
```
for (auto& fileInfo : files_) {
    streams.push_back(FileSpillMergeStream::create(
        SpillReadFile::create(fileInfo, bufferSize, pool, spillStats)));
  }

input_ = std::make_unique<common::FileInputStream>(
      std::move(file), bufferSize, pool_);

FileInputStream::FileInputStream(
    std::unique_ptr<ReadFile>&& file,
    uint64_t bufferSize,
    memory::MemoryPool* pool)
    : file_(std::move(file)),
      fileSize_(file_->size()),
      bufferSize_(std::min(fileSize_, bufferSize)),
      pool_(pool),
      readAheadEnabled_((bufferSize_ < fileSize_) && file_->hasPreadvAsync()) {
  VELOX_CHECK_NOT_NULL(pool_);
  VELOX_CHECK_GT(fileSize_, 0, "Empty FileInputStream");

  buffers_.push_back(AlignedBuffer::allocate<char>(bufferSize_, pool_)); // allocate buffer cause OOM
  if (readAheadEnabled_) {
[...]

PR #8026, opened 2024-11-23
[GLUTEN-8025][VL] set default kSpillWriteBufferSize to the same as velox

FelixYBW:
The kSpillWriteBufferSize is set to 4M by typo. correct it to 1M. The value control the buffer size when spill write data to disk. 

Add config kSpillReadBufferSize. it controls the read buffer size when spill data is fetch back. The memory will be allocated as offheap memory. So if there are to many spill files, the memory will exceed offheap size and trigger the OOM issue by operator::getOutput.

The spill files number is impacted by MaxSpillFileSize and MaxSpillRunRows, both are the smaller the more spill files.


FelixYBW:
#8025

zhztheplayer:
Shall we update [the code](https://github.com/apache/incubator-gluten/blob/58e7d831e7a18d707576a983f2b49d91b0b3d67f/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala#L1588-L1593) for consistency? Thanks!

jinchengchenghh:
Please update the PR to respect `kSpillReadBufferSize` and map it to Spark config `UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE`

FelixYBW:
close. merge to #8045

PR #7916, opened 2024-11-12
[DNM][VL] test march=native

zhouyuan:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#ISSUE-ID)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #8027, opened 2024-11-23
[GLUTEN-7950][VL] Keep Core module's build flag consistent with Velox

surnaik:
1. Keep build flags consistent with Velox Module in cpp/velox/CMakeList.txt
2. Keep velox build flag to default to increase coverage of targeted CPU Platforms

## How was this patch tested?

Local Build and tested on multiple GCE VMs - N1, N2, C2, N2D, C2D etc

### DIFF
--- a/cpp/core/CMakeLists.txt
+++ b/cpp/core/CMakeLists.txt
@@ -44,7 +44,7 @@ if(COMMAND_STATUS EQUAL "1")
   message(FATAL_ERROR "Unable to determine compiler flags!")
 endif()
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${SCRIPT_CXX_FLAGS}")
-message("Core module final CMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}"
+message("Core module final CMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}")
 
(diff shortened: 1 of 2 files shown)


surnaik:
@PHILO-HE / @FelixYBW please review. Thanks!

PHILO-HE:
@surnaik, thanks so much for your pr and test feedback! We already have a draft pr to fix this issue: https://github.com/apache/incubator-gluten/pull/7916.
And I think we should set flags in root CMakeLists.txt to make them applied consistently across all submodules.
cc @zhouyuan

surnaik:
> @surnaik, thanks so much for your pr and test feedback! We already have a draft pr to fix this issue: #7916. And I think we should set flags in root CMakeLists.txt to make them applied consistently across all submodules. cc @zhouyuan

Thanks @PHILO-HE , You may please close this PR if Yuan is already fixing it, didn't notice it.

zhouyuan (approved):
👍 
thanks for the fix, looks good to me. We are discussing if we could make another profile so some AVX512 instructions can be used inside Gluten. I will try to improve more on this part

PHILO-HE (approved):
Let's land this pr firstly. @surnaik, thanks so much for your work!

surnaik:
Thanks @PHILO-HE , as you suggested, let me actually clean it up and add it to the top level CMakeList.txt

PR #8029, opened 2024-11-24
[GLUTEN-7028][CH][Part-10] Collecting Delta stats for MergeTree

baibaichen:
(Fixes: #7028)

Similar with #7993, this PR supports collecting stats for mergetree.  I **remove** the old compaction algorithm and replaced it with delta's implementaion for two reasons:
1.  After https://delta.io/blog/delta-lake-3-2/, delta optimazie command. supports  **Liquid clustering**.
2. It's impossible to collect stats with old compaction.

> Notes
> Since we conside droping support `Bucket` table, we can use delta's `OptimizeTableCommand `instead of gluen's version

There are some tests failed with one pipeline write, recored in #7028, we can fix it in the next PRs

## How was this patch tested?

Uisng existed UTs


loneylee:
LGTM

loneylee (approved):

Issue #6920, opened 2024-08-19
[CORE] Enhancement Proposal: Plug-able backends


zhztheplayer:
### Description

Currently only only single backend can be used in Gluten. This topic suggests a plug-able system to allow enabling multiple backends at compile time. For example, eventually one could build Gluten with `-P backends-velox,backends-another` to enable both Velox backend and `another` backend that could possibly be added to Gluten code base in future. In the case when a query is being run, Gluten's query planner could choose the best path to implement the physical plan that may be a mixing of Velox sub-plans and `another` sub-plans.

The feature could drive Gluten to evolve from a hand-written middle layer to a development platform to offload Spark SQL with different data formats, acceleration libraries, or hardware architectures.

Prior to doing this, a bunch of preparation work should be done including essential code refactors.

PR #8036, opened 2024-11-25
[GLUTEN-6920][CORE] Redesign and move trait `GlutenPlan` to `gluten-core`

zhztheplayer:
Part of #6920

This work factors out `GlutenPlan` to separate the APIs to 2 places:

1. For APIs related to native validation, e.g., `doValidate`, move to new API `ValidatablePlan` and `TransformSupport`
2. Other convention-related APIs, move to gluten-core, and still be in `GlutenPlan`

After the change, `GlutenPlan` will become a thin and base interface for query plans of all Gluten backends. APIs in `GlutenPlan` will be used by query planner to decide whether pre-transition or post-transition is needed for this plan node. Backends could still define their own validation logic if needed, by extending the trait by themselves.

`GlutenPlan` will hide `supportsColumnar` and `supportsRowBased` (for Spark version > 3.2) up and expose Gluten APIs `rowType` and `batchType` to sub-classes instead. The latter two are more flexible since Gluten's query planner could distinguish between different batch types and add essential transitions among them and vanilla rows.


zhztheplayer:

rui-mo (approved):

Issue #7631, opened 2024-10-22
[VL] Bad optional access with lag function in Window


zhztheplayer:
Test name:

`SPARK-16633: lead/lag should return the default value if the offset row does not exist`

Error:

```
2024-10-22T06:07:14.2572740Z - SPARK-16633: lead/lag should return the default value if the offset row does not exist *** FAILED ***
2024-10-22T06:07:14.2573967Z   org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 152.0 failed 1 times, most recent failure: Lost task 0.0 in stage 152.0 (TID 109) (da070ca603ce executor driver): org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: bad optional access
2024-10-22T06:07:14.2574400Z 	at org.apache.gluten.iterator.ClosableIterator.hasNext(ClosableIterator.java:41)
2024-10-22T06:07:14.2574708Z 	at scala.collection.convert.Wrappers$JIteratorWrapper.hasNext(Wrappers.scala:45)
2024-10-22T06:07:14.2575092Z 	at org.apache.gluten.iterator.IteratorsV1$InvocationFlowProtection.hasNext(IteratorsV1.scala:159)
2024-10-22T06:07:14.2575433Z 	at org.apache.gluten.iterator.IteratorsV1$IteratorCompleter.hasNext(IteratorsV1.scala:71)
2024-10-22T06:07:14.2575751Z 	at org.apache.gluten.iterator.IteratorsV1$PayloadCloser.hasNext(IteratorsV1.scala:37)
2024-10-22T06:07:14.2576103Z 	at org.apache.gluten.iterator.IteratorsV1$LifeTimeAccumulator.hasNext(IteratorsV1.scala:100)
2024-10-22T06:07:14.2576371Z 	at scala.collection.Iterator.isEmpty(Iterator.scala:387)
2024-10-22T06:07:14.2576631Z 	at scala.collection.Iterator.isEmpty$(Iterator.scala:387)
2024-10-22T06:07:14.2576979Z 	at org.apache.gluten.iterator.IteratorsV1$LifeTimeAccumulator.isEmpty(IteratorsV1.scala:90)
2024-10-22T06:07:14.2577467Z 	at org.apache.gluten.execution.VeloxColumnarToRowExec$.toRowIterator(VeloxColumnarToRowExec.scala:121)
[...]

PR #8038, opened 2024-11-25
[GLUTEN-7631][VL] Fall back lead/lag if input is foldable

zml1206:

Velox does not support lead/lag when its input is foldable.

(Fixes: \#7631)

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -400,7 +400,9 @@ object VeloxBackendSettings extends BackendSettingsApi {
           }
           windowExpression.windowFunction match {
             case _: RowNumber | _: Rank | _: CumeDist | _: DenseRank | _: PercentRank |
-                _: NthValue | _: NTile | _: Lag | _: Lead =>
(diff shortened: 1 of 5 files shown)


PHILO-HE:
@zml1206, thanks for your pr! What's Velox's current behavior for this case? Will Velox interrupt with some error?

zml1206:
> @zml1206, thanks for your pr! What's Velox's current behavior for this case? Will Velox interrupt with some error?

The error reported by velox is the same as the error reported by gluten in the issue, "bad optional access".
image

image

zml1206:
https://github.com/facebookincubator/velox/blob/f33b40da09441d542f32ee9ed9fb2e340d3c2a75/velox/functions/prestosql/window/LeadLag.cpp#L33
The error should be reported here. @PHILO-HE

PHILO-HE (approved):
Looks good to me! Assume the foldable input is rarely used. So it has low priority. It's ok to firstly let Gluten fall back for that.

zml1206:
> Looks good to me! Assume the foldable input is rarely used. So it has low priority. It's ok to firstly let Gluten fall back for that.

Thank you @PHILO-HE 
Yes, although vanilla Spark supports foldable input of lag/lead, there is almost no usage scenario in this case, and the values ​​obtained are all null.

Issue #8039, opened 2024-11-25
[VL] Native writer should respect table properties for spark 3.2/3.3


yikf:
### Backend

VL (Velox)

### Bug description

Native writer should respect table properties for spark 3.2/3.3, like compress codec.

Let's say have a test suite, 
```
Seq(true, false).foreach { enableNativeWrite =>
      withSQLConf("spark.gluten.sql.native.writer.enabled" -> enableNativeWrite.toString) {
        withTable("t") {
          withSQLConf(
            "spark.sql.hive.convertMetastoreParquet" -> "false",
            "spark.sql.parquet.compression.codec" -> "gzip") {
            checkNativeWrite(
              "CREATE TABLE t STORED AS PARQUET TBLPROPERTIES ('parquet.compression'='zstd') AS SELECT 1 as c",
              checkNative = enableNativeWrite)
            val warehouse = conf.getConf(StaticSQLConf.WAREHOUSE_PATH)
          }
        }
      }
    }
```
vanilla spark will write parquet file with ZSTD compress codec, but the current native writer use GZIP

### Spark version

Spark-3.2.x

PR #8040, opened 2024-11-25
[GLUTEN-8039][VL] Native writer should respect table properties

yikf:
Fix https://github.com/apache/incubator-gluten/issues/8039.

Native writer should respect table properties for spark 3.2/3.3, like compress codec.

Let's say have a test suite, 
```
Seq(true, false).foreach { enableNativeWrite =>
      withSQLConf("spark.gluten.sql.native.writer.enabled" -> enableNativeWrite.toString) {
        withTable("t") {
          withSQLConf(
            "spark.sql.hive.convertMetastoreParquet" -> "false",
            "spark.sql.parquet.compression.codec" -> "gzip") {
            checkNativeWrite(
              "CREATE TABLE t STORED AS PARQUET TBLPROPERTIES ('parquet.compression'='zstd') AS SELECT 1 as c",
              checkNative = enableNativeWrite)
            val warehouse = conf.getConf(StaticSQLConf.WAREHOUSE_PATH)
          }
        }
      }
    }
```
vanilla spark will write parquet file with ZSTD compress codec, but the current native writer use GZIP
## How was this patch tested?
manual tests

**vanilla spark**
image

**before PR, native writer**
image

**with this PR, native writer**
image


yikf:
cc @jackylee-ch

shims/spark32/src/main/scala/org/apache/spark/sql/hive/execution/HiveFileFormat.scala
```diff
@@ -101,19 +102,26 @@ class HiveFileFormat(fileSinkConf: FileSinkDesc)
     val outputFormat = fileSinkConf.tableInfo.getOutputFileFormatClassName
     if ("true" == sparkSession.sparkContext.getLocalProperty("isNativeApplicable")) {
       val nativeFormat = sparkSession.sparkContext.getLocalProperty("nativeFormat")
+      val tableProperties = mutable.Map.empty[String, String]
+      tableDesc.getProperties.asScala.foreach {
```
jackylee-ch, line 106: It seems `val tableOptions = tableDesc.getProperties.asScala.toMap` is enough.

Besides, please add some suites to test about this.

zhouyuan:
CC @JkSelf

FelixYBW:
Another issue related to #8043 

Can you help to check all table write configures?

FelixYBW (approved):

JkSelf:
@yikf Thanks for your fix. Can you help to check the whether [ParquetFileFormat ](https://github.com/apache/incubator-gluten/blob/main/shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala#L89) in spark 32 and 33 also have this issue? Thanks.

yikf:
> @yikf Thanks for your fix. Can you help to check the whether [ParquetFileFormat ](https://github.com/apache/incubator-gluten/blob/main/shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala#L89) in spark 32 and 33 also have this issue? Thanks.

I checked the ParquetFileFormat in spark 32 and 33, and both Vanilla Spark and Gluten use GZIP instead of configuration in table properties, with consistent behavior, which I understand is as expected.

JkSelf (approved):

jackylee-ch:
> @yikf Thanks for your fix. Can you help to check the whether [ParquetFileFormat ](https://github.com/apache/incubator-gluten/blob/main/shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/ParquetFileFormat.scala#L89) in spark 32 and 33 also have this issue? Thanks.

@JkSelf AFAIK, this only occurs when using HiveFileFormat to read Hive Table. In that way Spark passes properties through configuration, we cannot directly get the compression codec from options.

yikf:
thanks @jackylee-ch @FelixYBW @JkSelf @zhouyuan

PR #8045, opened 2024-11-26
[GLUTEN-8025][VL] Respect config kSpillReadBufferSize and add spill compression codec

jinchengchenghh:
Respect Spark config `UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE`. as  Velox `kSpillReadBufferSize`.
Add the config `spark.gluten.sql.columnar.backend.velox.spillCompressionCodec`, if not set, use Spark config `spark.io.compression.codec` instead.
Respect Spark  `spark.shuffle.spill.diskWriteBufferSize` as Velox `kSpillWriteBufferSize`.

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
@@ -321,8 +321,6 @@ class GlutenConfig(conf: SQLConf) extends Logging {
 
   def veloxMaxSpillBytes: Long = conf.getConf(COLUMNAR_VELOX_MAX_SPILL_BYTES)
 
-  def veloxMaxWriteBufferSize: Long = conf.getConf(COLUMNAR_VELOX_MAX_SPILL_WRITE_BUFFER_SIZE)
-
   def veloxBloomFilterExpectedNumItems: Long =
     conf.getConf(COLUMNAR_VELOX_BLOOM_FILTER_EXPECTED_NUM_ITEMS)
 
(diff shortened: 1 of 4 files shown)


shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -1585,13 +1590,6 @@ object GlutenConfig {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefaultString("100G")
 
-  val COLUMNAR_VELOX_MAX_SPILL_WRITE_BUFFER_SIZE =
```
FelixYBW, line 1588: Let's put all configurations in GlutenConfig.scala even it's not documented in Gluten doc.

Can you add spillreadbuffersize as well?

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -732,7 +734,10 @@ object GlutenConfig {
         COLUMNAR_MEMORY_BACKTRACE_ALLOCATION.defaultValueString),
       (
         GLUTEN_COLUMNAR_TO_ROW_MEM_THRESHOLD.key,
-        GLUTEN_COLUMNAR_TO_ROW_MEM_THRESHOLD.defaultValue.get.toString)
+        GLUTEN_COLUMNAR_TO_ROW_MEM_THRESHOLD.defaultValue.get.toString),
```
FelixYBW, line 737: Let's check if spark.unsafe.sorter.spill.read.ahead.enabled is enabled or not. If not, close the spill read adhead buffer.

FelixYBW (changes requested):

FelixYBW:
some spark tunning parameters about spill, FYI

https://mageswaran1989.medium.com/spark-optimizations-for-advanced-users-spark-cheat-sheet-d74464618c20

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
jinchengchenghh: FileInputStream implement decides it must read some buffer ahead even if spark.unsafe.sorter.spill.read.ahead.enabled is disabled. We could add a new class FileInputReader to read the file as need.

FelixYBW: what happens if we set the buffer size to 0?

jinchengchenghh: It does not check the config value reasonable.
It may cause dead cycle because it read file 0 byte one time.
 https://github.com/facebookincubator/velox/blob/main/velox/common/file/FileInputStream.cpp#L207

FelixYBW: Then let's ignore it for now and open an enhancement issue on it.

jinchengchenghh: Yes, added, https://github.com/facebookincubator/velox/issues/11673.
It won't affect us now, because Spark config requires the value is more than default value.
```
private[spark] val UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE =
    ConfigBuilder("spark.unsafe.sorter.spill.reader.buffer.size")
      .internal()
      .version("2.1.0")
      .bytesConf(ByteUnit.BYTE)
      .checkValue(v => 1024 * 1024 <= v && v <= MAX_BUFFER_SIZE_BYTES,
        s"The value must be in allowed range [1,048,576, ${MAX_BUFFER_SIZE_BYTES}].")
      .createWithDefault(1024 * 1024)
```

I will add the FileReadStream to Velox to support disable read ahead.

FelixYBW:
@jinchengchenghh can you rebase. Let's merge it

jinchengchenghh:
Can you help approve and merge this PR? Thanks! @FelixYBW

FelixYBW (approved):

PR #8047, opened 2024-11-26
[VL] vcpkg: Broken libelf mirror

zhztheplayer:
The link we use for downloading libelf no longer works and is causing failures in static build:

https://ftp.osuosl.org/pub/blfs/conglomeration/libelf/libelf-0.8.13.tar.gz

Change to another mirror.


zhztheplayer:
@PHILO-HE

PHILO-HE (approved):

zhztheplayer:
perhaps should be ports back to 1.2

PR #8059, opened 2024-11-27
[VL] Branch 1.2: Port #8047 to fix libelf link

zhztheplayer:
#8047


PHILO-HE:
Will merge the patch soon. I will create another pr to port some other patches, including the fix for CI.

Issue #8061, opened 2024-11-27
nth_value(1) throw VeloxRuntimeError


zml1206:
### Backend

VL (Velox)

### Bug description

```
select nth_value(1, 2) over (order by a) from values (1), (2) as tbl(a)
```

### Spark version

None

### Relevant logs

```bash
org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Retriable: False
Expression: args[0].constantValue == nullptr
Function: NthValueFunction
File: /Users/zml/Desktop/git_hub/incubator-gluten1/ep/build-velox/build/velox_ep/velox/functions/lib/window/NthValue.cpp
Line: 34
```

PR #8062, opened 2024-11-27
[GLUTEN-8061][VL] Fall back nth_value if input is foldable

zml1206:

Velox does not support nth_value when its input is foldable.

Although vanilla Spark supports foldable input of nth_value, there is almost no usage scenario in this case, and the values ​​obtained are all null, so simply fall back for it.

(Fixes: \#8061)

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -399,8 +399,8 @@ object VeloxBackendSettings extends BackendSettingsApi {
             case _ =>
           }
           windowExpression.windowFunction match {
-            case _: RowNumber | _: Rank | _: CumeDist | _: DenseRank | _: PercentRank |
-                _: NthValue | _: NTile =>
+            case _: RowNumber | _: Rank | _: CumeDist | _: DenseRank | _: PercentRank | _: NTile =>
+            case nv: NthValue if !nv.input.foldable =>
             case l: Lag if !l.input.foldable =>
             case l: Lead if !l.input.foldable =>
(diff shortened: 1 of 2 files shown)


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -399,8 +399,8 @@ object VeloxBackendSettings extends BackendSettingsApi {
             case _ =>
           }
           windowExpression.windowFunction match {
-            case _: RowNumber | _: Rank | _: CumeDist | _: DenseRank | _: PercentRank |
-                _: NthValue | _: NTile =>
+            case _: RowNumber | _: Rank | _: CumeDist | _: DenseRank | _: PercentRank | _: NTile =>
+            case nv: NthValue if !nv.input.foldable =>
```
PHILO-HE, line 403: Just for brevity. Maybe, we can use:
```
case _ @NthValue(input, _, _) if !input.foldable =>
```

Ditto for Lead/Lag.

PHILO-HE (commented):
A minor comment. Thanks!

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
zml1206: This method is more friendly to changes in the number of parameters that may exist in multi-version support. Generally, we will not change the parameter names. What do you think?

PHILO-HE: @zml1206, it makes sense. Thanks!

PHILO-HE (approved):

Issue #8070, opened 2024-11-28
[CORE] Support generate configurations golden file and check it 


yikf:
### Description

Add generate configurations golden file tools to auto generate Configuration.md doc and check it if PR change configs.

Issue #6970, opened 2024-08-22
[CORE] Configuration improvements


zhztheplayer:
### Description

Gluten's configuration utilities, especially `GlutenConfig.scala`, is getting messed up over time. To keep code clean and easy to maintain, will propose the below changes:

### Simplify and unify option key formats

1. Could simplify config key prefix from `spark.gluten.sql.columnar` tp `spark.gluten`.
2. Use camel case uniformly, e.g, `spark.gluten.sql.columnar.backend.velox.IOThreads` to `spark.gluten.sql.columnar.backend.velox.ioThreads`

### Move specific configurations to backend module

For better module-wise isolation, we should move configuration definitions specifically for backends to their own module. E.g., `spark.gluten.sql.columnar.backend.velox.window.type` should be defined in `backends-velox` than `gluten-core`.

### (Velox) Remove the list of configurations passing to native

In Velox we currently have a list to decide whether a config option should be passed to native. Rather than using this way, we can just set a flag when define the config. Or just using Spark's built-in static / non-static config flag to distinguish.

### Backward compatibility

Since configuration keys may be modified during doing this, we should maintain a list of configuration keys for backward compatibility. E.g., `spark.gluten.sql.columnar.backend.velox.IOThreads` can be the alternative of `spark.gluten.backend.velox.ioThreads`. But we should raise a warning when the former is set by user as it's considered deprecated. After several development iterations, support of the old one will be removed completely.

zhztheplayer:
This is only a draft plan. Should indeed revisit before starting to work on it.

Yohahaha:
that's great!
I suppose `spark.gluten.sql.columnar.backend.velox.ioThreads` will be changed to `spark.gluten.velox.ioThreads`, right?

zhztheplayer:
> .backend. can be removed.

Yes that's one possible solution. If we drop `.backend`, then for the option keys that are common to all backends, say `spark.gluten.sql.columnar.broadcastJoin`, I think we could insert a `common.` or something to distinguish the namespace with backend configs, e.g., rename it to `spark.gluten.common.broadcastJoin`.


Basically we have two ideas:

- Idea 1, `backend.` for backend configs
   - `spark.gluten.sql.columnar.backend.velox.ioThread` renamed to `spark.gluten.backend.velox.ioThread`
   - `spark.gluten.sql.columnar.broadcastJoin` renamed to `spark.gluten.broadcastJoin`
- Idea 2, `common.` for common configs
[...]

PR #8071, opened 2024-11-28
[GLUTEN-8070][CORE] Generate gluten configuration doc and check it

yikf:
https://github.com/apache/incubator-gluten/issues/8070, Add generate configurations golden file tools to auto generate Configuration.md doc and check it if PR change configs.

## How was this patch tested?
manual tests


yikf:
@zhouyuan @PHILO-HE @jackylee-ch do you like the feature and Is this feature necessary?

gluten-core/src/test/scala/org/apache/gluten/config/AllGlutenConfiguration.scala
```diff
@@ -0,0 +1,260 @@
...
+      """There are many configurations could impact the Gluten Plugin performance
+        | and can be fine-tuned in Spark.""" +=
+      """You can add these configurations into spark-defaults.conf to enable or
+        | disable the setting."""
+
+    builder ++=
+      s"""
+         |## Gluten configurations
+         |
+         | Key | Default | Meaning
```
PHILO-HE, line 88: Suggestion: 
Meaning -> Description

gluten-core/src/test/scala/org/apache/gluten/MarkdownBuilder.scala
```diff
@@ -0,0 +1,127 @@
...
+import com.vladsch.flexmark.util.data.{MutableDataHolder, MutableDataSet}
+import com.vladsch.flexmark.util.sequence.SequenceUtils.EOL
+
+import scala.collection.mutable.ListBuffer
+
+class MarkdownBuilder {
+  private val buffer = new ListBuffer[String]
+
+  /**
+   * append a single line with replacing EOL to empty string
```
PHILO-HE, line 31: Nit: 
comment style, capitalize the initial letter and end up with period mark. Please also check other comments.

PHILO-HE (commented):
Thanks for your work!
Can we move the code to `tools` module? Maybe, create another pom module inside it, like  `tools/gluten-it`. I assume this tool doesn't depend on the test framework. So it may be not necessary to implement it in test folder.

docs/Configuration.md
```diff
@@ -1,107 +1,218 @@
...
-| spark.driver.extraClassPath                                  | Gluten Plugin jar file to prepend to the classpath of the driver.                                                                                                                                                                                                                                                                                                                                                                                                                                                                         | /path/to/jar_file                                    |
```
jackylee-ch, line 16: We need something to do before regenrating the `Configuration.md`.
1. A way to define the spark parameters that used for gluten;
2. Synchronize the configuration information in `Configuration.md` and `GlutenConfig`, because some configurations are not well defined in GlutenConfig;
3. Use a unified interface to define the gluten configs, such as internal, backends, and so on, so we can define how to show each config in `Configuration.md`

jackylee-ch: There is a issue #6970 that discussed `GlutenConfig` before.

jackylee-ch:
also cc @zhztheplayer @zzcclp

yikf:
thanks @jackylee-ch and @PHILO-HE , After learning more about your suggestions and relevant background, I recommend dividing the optimization efforts for config into two parts:

1. Optimize the relevant config definitions, just like what was mentioned in #6970.
2. Support generate gluten configuration doc.

Regarding `#1`, I will be putting together a draft in the next few days and initiating discussion in the community.

We can continue with this RP after `#1` is complete.

zhztheplayer:
> I recommend dividing the optimization efforts for config into two parts

+1, works for me. Thanks.

Issue #7261, opened 2024-09-18
[VL] offload table scan when filter need fallback


FelixYBW:
### Description

currently when a filter has UDF and fallback, the table scan is also fallback. We should offload table scan since it can benefit from native and much more expensive than filter and new C2R.

In this query, parquet scan output int and string only.

![image](https://github.com/user-attachments/assets/3b5d5263-2ebf-4b11-bdb3-2d3d76d10cd4)

zhztheplayer:
There is a pending PR possibly related to this topic https://github.com/apache/incubator-gluten/pull/7215

@zml1206 Would we still be able to optimize against such issues after that PR get landed?

FelixYBW:
Should be. Let me do a quick test once @zml1206 resolve conflicts

zml1206:
Conflicts resolved. But I'm sorry I don't understand why scan will fallback. Is it because RAS is enabled?

zml1206:
Can we determine whether expression supports offload? @zhztheplayer There are three layers of filters here. The first layer is pushedDownFilters of FileSourceScanExec, which is also used by vanilla spark scan. The second layer is the dataFilters of FileSourceScanExec, which is used to generate the pushedDownFilters of FileSourceScanExec, as well as the filters used after the current offload. The third layer is the condition of the filter. In theory, the filter condition contains dataFilters, and dataFilters contains pushedDownFilters.
The current implementation will push down dataFilters or condition filters. Simply rewrite dataFilters to pushedDownFilters may cause dataFilters to be pushed down before, but only pushedDownFilters can be pushed down.

zhztheplayer:
@zml1206 Thanks for explanation. 

> Can we determine whether expression supports offload?
Let's see if https://github.com/apache/incubator-gluten/pull/6754 on which @rui-mo is working could finally help here.

PR #8082, opened 2024-11-28
[GLUTEN-7261][CORE] Support offloading partial filters to native scan

zml1206:

Scan used to fallback if there was an unsupported filter. This PR filters out the supported expressions and offloads scan as much as possible to improve the performance. Before this change, the plan was "vanilla vectorized scan + c2r + vanilla filter" when scan contained an unsupported filter, and now the plan becomes "native scan + c2r + vanilla filter".

(Fixes: \#7261)

## How was this patch tested?

UT

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
@@ -109,7 +109,9 @@ abstract class BatchScanExecTransformerBase(
             ExpressionConverter.replaceAttributeReference(expr),
             output)
       }
-    case _ => Seq.empty
+    case _ =>
+      logInfo(s"${scan.getClass.toString} does not support push down filters")
+      Seq.empty
   }
 
   def setPushDownFilters(filters: Seq[Expression]): Unit = {
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
@@ -17,7 +17,7 @@
 package org.apache.gluten.execution
 
 import org.apache.gluten.backendsapi.BackendsApiManager
-import org.apache.gluten.exception.GlutenNotSupportException
+import org.apache.gluten.expression.ExpressionConverter
 import org.apache.gluten.extension.ValidationResult
 import org.apache.gluten.metrics.MetricsUpdater
 import org.apache.gluten.sql.shims.SparkShimLoader
@@ -101,18 +101,22 @@ abstract class BatchScanExecTransformerBase(
   // class. Otherwise, we will encounter an issue where makeCopy cannot find a constructor
   // with the corresponding number of parameters.
   // The workaround is to add a mutable list to pass in pushdownFilters.
-  protected var pushdownFilters: Option[Seq[Expression]] = None
+  protected var pushdownFilters: Seq[Expression] = scan match {
+    case fileScan: FileScan =>
+      fileScan.dataFilters.filter {
+        expr =>
+          ExpressionConverter.canReplaceWithExpressionTransformer(
+            ExpressionConverter.replaceAttributeReference(expr),
+            output)
+      }
+    case _ => Seq.empty
+  }
 
   def setPushDownFilters(filters: Seq[Expression]): Unit = {
-    pushdownFilters = Some(filters)
+    pushdownFilters = filters
   }
 
-  override def filterExprs(): Seq[Expression] = scan match {
-    case fileScan: FileScan =>
-      pushdownFilters.getOrElse(fileScan.dataFilters)
-    case _ =>
-      throw new GlutenNotSupportException(s"${scan.getClass.toString} is not supported")
-  }
+  override def filterExprs(): Seq[Expression] = pushdownFilters
 
   override def getMetadataColumns(): Seq[AttributeReference] = Seq.empty
 
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/ScanTransformerFactory.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/ScanTransformerFactory.scala
@@ -16,6 +16,7 @@
  */
 package org.apache.gluten.execution
 
+import org.apache.gluten.expression.ExpressionConverter
 import org.apache.gluten.sql.shims.SparkShimLoader
 
 import org.apache.spark.sql.execution.FileSourceScanExec
@@ -41,6 +42,12 @@ object ScanTransformerFactory {
           .asInstanceOf[DataSourceScanTransformerRegister]
           .createDataSourceTransformer(scanExec)
       case _ =>
+        val dataFilters = scanExec.dataFilters.filter {
+          expr =>
+            ExpressionConverter.canReplaceWithExpressionTransformer(
+              ExpressionConverter.replaceAttributeReference(expr),
+              scanExec.output)
+        }
         FileSourceScanExecTransformer(
           scanExec.relation,
           scanExec.output,
@@ -48,7 +55,7 @@ object ScanTransformerFactory {
           scanExec.partitionFilters,
           scanExec.optionalBucketSet,
           scanExec.optionalNumCoalescedBuckets,
-          scanExec.dataFilters,
+          dataFilters,
           scanExec.tableIdentifier,
           scanExec.disableBucketedScan
         )
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/ScanTransformerFactory.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/ScanTransformerFactory.scala
@@ -16,7 +16,6 @@
  */
 package org.apache.gluten.execution
 
-import org.apache.gluten.expression.ExpressionConverter
 import org.apache.gluten.sql.shims.SparkShimLoader
 
 import org.apache.spark.sql.execution.FileSourceScanExec
@@ -42,12 +41,6 @@ object ScanTransformerFactory {
           .asInstanceOf[DataSourceScanTransformerRegister]
           .createDataSourceTransformer(scanExec)
       case _ =>
-        val dataFilters = scanExec.dataFilters.filter {
-          expr =>
-            ExpressionConverter.canReplaceWithExpressionTransformer(
-              ExpressionConverter.replaceAttributeReference(expr),
-              scanExec.output)
-        }
         FileSourceScanExecTransformer(
           scanExec.relation,
           scanExec.output,
@@ -55,7 +48,7 @@ object ScanTransformerFactory {
           scanExec.partitionFilters,
           scanExec.optionalBucketSet,
           scanExec.optionalNumCoalescedBuckets,
-          dataFilters,
+          scanExec.dataFilters,
           scanExec.tableIdentifier,
           scanExec.disableBucketedScan
         )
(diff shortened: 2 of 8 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
```diff
@@ -165,4 +166,11 @@ abstract class BatchScanExecTransformerBase(
     case "ClickHouseScan" => ReadFileFormat.MergeTreeReadFormat
     case _ => ReadFileFormat.UnknownFormat
   }
+
+  override def simpleString(maxFields: Int): String = {
+    val truncatedOutputString = truncatedString(output, "[", ", ", "]", maxFields)
+    val runtimeFiltersString = s"RuntimeFilters: ${filterExprs().mkString("[", ",", "]")}"
+    val result = s"$nodeName$truncatedOutputString ${scan.description()} $runtimeFiltersString"
+    redact(result)
+  }
```
zml1206, line 175: `filterExprs` is the real `RuntimeFilters`.

zml1206:
cc @FelixYBW it can resolve #7261

zhztheplayer:
cc @rui-mo

gluten-substrait/src/main/scala/org/apache/gluten/execution/ScanTransformerFactory.scala
```diff
@@ -55,6 +56,11 @@ object ScanTransformerFactory {
           scanExec.tableIdentifier,
           scanExec.disableBucketedScan
         )
+        if (!transform.doValidate().ok()) {
+          transform.copy(dataFilters = PushDownUtil.pushFilters(scanExec.dataFilters))
+        } else {
+          transform
+        }
```
zhztheplayer, line 63: The code in `ScanTransformerFactory` is used by validator and offload rules. It feels a little weird to do validation in it? Do we have better choices? If not it's also OK to me to keep code like this.

zml1206: How about use only `pushedFilter` here and rely on `PushDownFilterToScan` for subsequent pushdown?

zhztheplayer: Sounds feasible to me. Thanks.

zml1206:
Test failure seems unrelated.

gluten-substrait/src/main/scala/org/apache/spark/sql/utils/PushDownUtil.scala
```diff
@@ -0,0 +1,44 @@
...
+
+object PushDownUtil {
+
+  // Refer to vanilla spark to get pushedFilters, and then convert Filter to Expression.
+  def pushFilters(dataFilters: Seq[Expression]): Seq[Expression] = {
+    val translatedFilterToExpr = mutable.HashMap.empty[sources.Filter, Expression]
+    val translatedFilters = mutable.ArrayBuffer.empty[sources.Filter]
+    for (filterExpr <- dataFilters) {
+      val translated =
+        DataSourceStrategy.translateFilterWithMapping(
```
rui-mo, line 33: Would you elaborate on how this translation happens and how the pushed filters differ from Spark in most cases? If it is based on Spark rules, we cannot control the expressions that are to be pushed down. Is it more reasonable to adopt specific rules according to the backend status?

rui-mo (commented):
Thanks. Added some questions.

gluten-substrait/src/main/scala/org/apache/spark/sql/utils/PushDownUtil.scala
zml1206: This logic is the same as that of vanilla spark to generate pushedFilter, and then convert Seq[Filter] to Seq[Expression]. dataFilter does not contain non-deterministic expressions, but contains expensive expressions, such as udf. pushedFilter only contains cheap expressions, such as `a>1`, `a in (1,2)`.
Refer https://github.com/apache/spark/blob/1eb558c3a6fbdd59e5a305bc3ab12ce748f6511f/sql/core/src/main/scala/org/apache/spark/sql/execution/DataSourceScanExec.scala#L374

rui-mo: > This logic is the same as that of vanilla spark to generate pushedFilters from dataFilters, and then convert Seq[Filter] to Seq[Expression]. dataFilter does not contain non-deterministic expressions, but contains expensive expressions, such as udf. pushedFilter only contains cheap expressions, such as `a>1`, `a in (1,2)`. Refer https://github.com/apache/spark/blob/1eb558c3a6fbdd59e5a305bc3ab12ce748f6511f/sql/core/src/main/scala/org/apache/spark/sql/execution/DataSourceScanExec.scala#L374 https://github.com/apache/spark/blob/1eb558c3a6fbdd59e5a305bc3ab12ce748f6511f/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/v2/FileScanBuilder.scala#L72

Perhaps the above information could be documented. Also, I have an idea: can we try if the expressions can be converted as Gluten expression transformers? The unsupported ones will not be in the pushed filters.

https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala

zml1206: > Also, I have an idea: can we try if the expressions can be converted as Gluten expression transformers? The unsupported ones will not be in the pushed filters.

Good idea! 
Vanilla spark's pushedFilters do not contain expensive expressions, so here not needed? We can try this idea later to optimize PushDownFilterToScan.

zml1206: Good idea! I have already done, please help take a look again, thanks @rui-mo

rui-mo: Thanks for the update! Can the issue https://github.com/apache/incubator-gluten/issues/7261 for a filter has UDF be resolved with current fix?

zml1206: Yes, I verified it locally, but the method used in UT is to disable `add`.

zml1206:
```
withSQLConf(
      GlutenConfig.EXPRESSION_BLACK_LIST.key -> "add",
      SQLConf.USE_V1_SOURCE_LIST.key -> "") {
      val plusOne = udf((x: Long) => x + 1)
      spark.udf.register("plus_one", plusOne)
      createTPCHNotNullTables()
      val query = "select l_partkey from lineitem where plus_one(l_partkey) > 5 and l_partkey - 1 < 8"
      runQueryAndCompare(query) {
        df =>
          {
            val executedPlan = getExecutedPlan(df)
            val scans = executedPlan.collect { case p: BatchScanExecTransformer => p }
            assert(scans.size == 1)
            // isnotnull(l_partkey) and l_partkey - 1 < 8
            assert(scans.head.filterExprs().size == 2)
          }
      }
    }
```

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -55,6 +55,24 @@ object ExpressionConverter extends SQLConfHelper with Logging {
     replaceWithExpressionTransformer0(expr, attributeSeq, expressionsMap)
   }
 
+  def canReplaceWithExpressionTransformer(
+      expr: Expression,
+      attributeSeq: Seq[Attribute]): Boolean = {
+    val expressionsMap = ExpressionMappings.expressionsMap
+    try {
+      replaceWithExpressionTransformer0(expr, attributeSeq, expressionsMap)
```
rui-mo, line 63: Can we just use 'replaceWithExpressionTransformer'?

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -55,6 +55,24 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+      true
+    } catch {
+      case _: Exception => false
```
rui-mo, line 66: nit: better to log this message for easy debugging.

gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
```diff
@@ -96,18 +96,22 @@ abstract class BatchScanExecTransformerBase(
...
-  protected var pushdownFilters: Option[Seq[Expression]] = None
+  protected var pushdownFilters: Seq[Expression] = scan match {
+    case fileScan: FileScan =>
+      fileScan.dataFilters.filter {
+        expr =>
+          ExpressionConverter.canReplaceWithExpressionTransformer(
+            ExpressionConverter.replaceAttributeReference(expr),
+            output)
+      }
+    case _ => Seq.empty
```
rui-mo, line 107: Perhaps add some warnings for easy debugging.

rui-mo (commented):
Regarding the plan change, we used to get 'vanilla scan + vanilla filter' for unsupported filter, and now we can get 'native scan + c2r + vanilla filter'. The new plan offloads 'scan' to native while I assume when the c2r is large and time-consuming we might not get performance improvement. Can our RAS strategy cover this case? cc: @zhztheplayer Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
zml1206: I don’t quite understand what the warning is here. Can you explain it? Thank you. @rui-mo

zml1206:
> Regarding the plan change, we used to get 'vanilla scan + vanilla filter' for unsupported filter, and now we can get 'native scan + c2r + vanilla filter'. The new plan offloads 'scan' to native while I assume when the c2r is large and time-consuming we might not get performance improvement. Can our RAS strategy cover this case? cc: @zhztheplayer Thanks.


RAS currently cannot solve this problem, but from our production point of view, the cost of c2r is relatively small. @zhztheplayer What do you think?

gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
rui-mo: Previously we throw an exception if it is not file scan. Do we need to log the information for unsupported cases?
https://github.com/apache/incubator-gluten/pull/8082/files#diff-0637bf279d5bc9ae2791c4e8919675e1674426634f187dfc55596f7c1eb4c5afL109

zml1206:

rui-mo:
> from our production point of view, the cost of c2r is relatively small

With the previous plan, C2R is only needed for rows after the filter, whose number might be largely reduced, while in the new plan, all rows need to be converted as rows. Perhaps in some cases the speedup of native scan cannot compensate for this overhead, and we might get performance regression.

zml1206:
> > from our production point of view, the cost of c2r is relatively small
> 
> With the previous plan, C2R is only needed for rows after the filter, whose number might be largely reduced, while in the new plan, all rows need to be converted as rows. Perhaps in some cases the speedup of native scan cannot compensate for this overhead, and we might get performance regression.

Before PR is Scan + ColumnToRow + Filter, After PR is nativeScan + VeloxColumnToRow + Filter, the number of rows after nativeScan should be less than or equal to that after scan.

rui-mo:
> Before PR is Scan + ColumnToRow + Filter

Why do we have the C2R between, because for unsupported filters they should be vanilla Spark computed?

zml1206:
> Why do we have the C2R between, because for unsupported filters they should be vanilla Spark computed?

The default for vanilla Spark reading parquet is vectorization.

rui-mo:
@zml1206 I see your point. Thanks for explaning!

rui-mo (approved):

Issue #8060, opened 2024-11-27
[CORE] Register native shuffles through backend API


zhztheplayer:
Part of https://github.com/apache/incubator-gluten/issues/6920.

So far Gluten's native shuffle implementations, including VL, CH, VL + RSS are activated by specifying corresponding shuffle manager types in Spark configuration. 

However if we want to mix different native shuffles in the same Spark session even in the same query plan, as Spark only allows one shuffle manager to be registered (see also a related [SPIP](https://issues.apache.org/jira/browse/SPARK-45792)) yet, we should enhance out current approach by providing a registry of native shuffles internally under an umbrella shuffle manager. Then query planner could choose an optimal one among all the candidates.

PR #8084, opened 2024-11-28
[GLUTEN-8060][CORE] GlutenShuffleManager as a registry of shuffle managers

zhztheplayer:
#8060

Prepare `GlutenShuffleManager` as the new shuffle manager implementation of Gluten.

The shuffle manager is simply a router for all shuffle managers registered through API `ShuffleManagerRegistry.get().register`.

About backward compatibility:

1. Use of `ColumnarShuffleManager` / `CelebornShuffleManager` / `UniffleShuffleManager` will be deprecated in future (perhaps, in next few of PRs). Then once Spark configuration like `spark.shuffle.manager=....ColumnarShuffleManager` is specified explicitly, Gluten could still run but raise a warning indicating that the option is deprecated.
2. Instead, use of `GlutenShuffleManager` will be recommended. `GlutenShuffleManager` will maintain a mechanism (either automatically or by configuration) to route shuffle manager calls to the underlying registered shuffle managers. Hence, it will become recommended to use `spark.shuffle.manager=....GlutenShuffleManager` instead, no matter RSS is required or not.

This PR is only an API preparation with essential UTs. No existing production code is affected.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
@@ -9,13 +9,22 @@ class ShuffleManagerRegistry private[ShuffleManagerRegistry] {
   import ShuffleManagerRegistry._
   private val all: mutable.Buffer[(LookupKey, String)] = mutable.Buffer()
   private val routerBuilders: mutable.Buffer[RouterBuilder] = mutable.Buffer()
+  private val classDeDup: mutable.Set[String] = mutable.Set()
 
   def register(lookupKey: LookupKey, shuffleManagerClass: String): Unit = {
     val clazz = Utils.classForName(shuffleManagerClass)
-    assert(
+    require(
+      !clazz.isAssignableFrom(classOf[GlutenShuffleManager]),
+      "It's not allowed to register GlutenShuffleManager recursively")
+    require(
       classOf[ShuffleManager].isAssignableFrom(clazz),
-      s"Shuffle manager class to register is not an implementation of Spark ShuffleManager")
+      s"Shuffle manager class to register is not an implementation of Spark ShuffleManager: $shuffleManagerClass"
+    )
+    require(
+      !classDeDup.contains(shuffleManagerClass),
+      s"Shuffle manager class already registered: $shuffleManagerClass")
     this.synchronized {
+      classDeDup += shuffleManagerClass
       all += lookupKey -> shuffleManagerClass
       // Invalidate all shuffle managers cached in each alive router builder instances.
       // Then, once the router builder is accessed, a new router will be forced to create.
@@ -23,11 +32,21 @@ class ShuffleManagerRegistry private[ShuffleManagerRegistry] {
     }
   }
 
-  private[shuffle] def newRouterBuilder(conf: SparkConf, isDriver: Boolean): RouterBuilder = this.synchronized {
-    val out = new RouterBuilder(this, conf, isDriver)
-    routerBuilders += out
-    out
+  // Visible for testing
+  private[shuffle] def clear(): Unit = {
+    this.synchronized {
+      classDeDup.clear()
+      all.clear()
+      routerBuilders.foreach(_.invalidateCache())
+    }
   }
+
+  private[shuffle] def newRouterBuilder(conf: SparkConf, isDriver: Boolean): RouterBuilder =
+    this.synchronized {
+      val out = new RouterBuilder(this, conf, isDriver)
+      routerBuilders += out
+      out
+    }
 }
 
 object ShuffleManagerRegistry {
@@ -45,7 +64,7 @@ object ShuffleManagerRegistry {
     private[shuffle] def getOrBuild(): ShuffleManagerRouter = synchronized {
       if (router.isEmpty) {
         val instances = registry.all.map(key => key._1 -> instantiate(key._2, conf, isDriver))
-        router = Some(new ShuffleManagerRouter(new ShuffleManagerLookup(instances)))
+        router = Some(new ShuffleManagerRouter(new ShuffleManagerLookup(instances.toSeq)))
       }
       router.get
     }
--- a/gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
@@ -23,7 +23,7 @@ class ShuffleManagerRegistry private[ShuffleManagerRegistry] {
     }
   }
 
-  def newRouterBuilder(conf: SparkConf, isDriver: Boolean): RouterBuilder = this.synchronized {
+  private[shuffle] def newRouterBuilder(conf: SparkConf, isDriver: Boolean): RouterBuilder = this.synchronized {
     val out = new RouterBuilder(this, conf, isDriver)
     routerBuilders += out
     out
--- a/gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
@@ -1,0 +1,58 @@
+package org.apache.spark.shuffle
+
+import org.apache.spark.SparkConf
+import org.apache.spark.util.Utils
+
+import scala.collection.mutable
+
+class ShuffleManagerRegistry private[ShuffleManagerRegistry] {
+  import ShuffleManagerRegistry._
+  private val all: mutable.Buffer[(LookupKey, String)] = mutable.Buffer()
+  private val routerBuilders: mutable.Buffer[RouterBuilder] = mutable.Buffer()
+
+  def register(lookupKey: LookupKey, shuffleManagerClass: String): Unit = {
+    val clazz = Utils.classForName(shuffleManagerClass)
+    assert(
+      classOf[ShuffleManager].isAssignableFrom(clazz),
+      s"Shuffle manager class to register is not an implementation of Spark ShuffleManager")
+    this.synchronized {
+      all += lookupKey -> shuffleManagerClass
+      // Invalidate all shuffle managers cached in each alive router builder instances.
+      // Then, once the router builder is accessed, a new router will be forced to create.
+      routerBuilders.foreach(_.invalidateCache())
+    }
+  }
+
+  def newRouterBuilder(conf: SparkConf, isDriver: Boolean): RouterBuilder = this.synchronized {
+    val out = new RouterBuilder(this, conf, isDriver)
+    routerBuilders += out
+    out
+  }
+}
+
+object ShuffleManagerRegistry {
+  private val instance = new ShuffleManagerRegistry()
+
+  def get(): ShuffleManagerRegistry = instance
+
+  class RouterBuilder(registry: ShuffleManagerRegistry, conf: SparkConf, isDriver: Boolean) {
+    private var router: Option[ShuffleManagerRouter] = None
+
+    private[ShuffleManagerRegistry] def invalidateCache(): Unit = synchronized {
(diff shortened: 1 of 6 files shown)


zhztheplayer:
cc @kerwin-zk @summaryzb @marin-ma @wForget

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
```diff
@@ -0,0 +1,94 @@
...
+  def get(): ShuffleManagerRegistry = instance
+
+  class RouterBuilder(registry: ShuffleManagerRegistry, conf: SparkConf, isDriver: Boolean) {
+    private var router: Option[ShuffleManagerRouter] = None
+
+    private[ShuffleManagerRegistry] def invalidateCache(): Unit = synchronized {
+      router = None
+    }
+
+    private[shuffle] def getOrBuild(): ShuffleManagerRouter = synchronized {
```
wForget, line 81:
```
if (router.isEmpty) {
  synchronized {
    if (router.isEmpty) {
       ....
    }
  }
}
router.get
```

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRouter.scala
```diff
@@ -0,0 +1,137 @@
...
+package org.apache.spark.shuffle
+import org.apache.spark.{ShuffleDependency, TaskContext}
+import org.apache.spark.network.buffer.ManagedBuffer
+import org.apache.spark.network.shuffle.MergedBlockMeta
+import org.apache.spark.storage.{BlockId, ShuffleBlockBatchId, ShuffleBlockId, ShuffleMergedBlockId}
+
+/** The internal shuffle manager instance used by GlutenShuffleManager. */
+private class ShuffleManagerRouter(lookup: ShuffleManagerLookup) extends ShuffleManager {
+  import ShuffleManagerRouter._
+  private val cache = new Cache()
```
wForget, line 26: Will this status be sensed by spark executor?

kerwin-zk:
@zhztheplayer  Although spark.shuffle.manager can be uniformly configured as GlutenShuffleManager, additional configurations are still required to choose between ColumnarShuffleManager, CelebornShuffleManager, or UniffleShuffleManager. Essentially, this is no different from the current setup where spark.shuffle.manager is configured to different managers. Moreover, from a certain perspective, the introduction of GlutenShuffleManager couples the various ShuffleManagers together, whereas the current fully decoupled setup seems clearer.

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRouter.scala
```diff
@@ -0,0 +1,137 @@
...
+    val manager = lookup.findShuffleManager(dependency)
+    cache.store(shuffleId, manager).registerShuffle(shuffleId, dependency)
+  }
+
+  override def getWriter[K, V](
+      handle: ShuffleHandle,
+      mapId: Long,
+      context: TaskContext,
+      metrics: ShuffleWriteMetricsReporter): ShuffleWriter[K, V] = {
+    cache.get(handle.shuffleId).getWriter(handle, mapId, context, metrics)
```
wForget, line 41: It may be better to determine the appropriate ShuffleManager based on ShuffleHandle

wForget:
Do you originally want to implement a mixture of ess and rss like https://github.com/apache/incubator-uniffle/blob/master/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java and https://github.com/apache/celeborn/blob/3bf91929b6bd02974b5d15b4d6804c9b2b01cfc0/client-spark/spark-3/src/main/java/org/apache/spark/shuffle/celeborn/SparkShuffleManager.java#L164?

zhztheplayer:
@kerwin-zk 

> Moreover, from a certain perspective, the introduction of GlutenShuffleManager couples the various ShuffleManagers together, whereas the current fully decoupled setup seems clearer.

I haven't clearly gotten what's the new code coupling come from. Say ideally we can remove the proxied calls from RSS shuffle managers to default shuffle manager because of this work (each shuffle manager can be registered for certain shuffle dependency type it can handle). So I see this a decoupling of code rather than coupling of code.

> Essentially, this is no different from the current setup where spark.shuffle.manager is configured to different managers.

But I may understand your concern totally especially this one. The reason I am proposing this is not simply because of code refactor. The solution's kind of a trade-off for letting user enable different backends at the same time. See the umbrella https://github.com/apache/incubator-gluten/issues/6920. For that purpose, we have to develop a way to make the shuffle managers be registered together so Gluten could choose the right one to use. For example, if we introduce GPU backend, we should give back the flexibility to developers when they decide not to let GPU backend rely on Velox backend, but they may still want query planner use Velox shuffle when the query plan is not offloaded to GPU.

May be we can put off deprecating the current use of `CelebornShuffleManager` / `UniffleShuffleManager`? We can keep the support for the legacy configurations during a considerable long period. I agree this is very normal that a user want's to specify a shuffle manager via Spark config explicitly. Does that work for you?

zhztheplayer:
> Do you originally want to implement a mixture of ess and rss like https://github.com/apache/incubator-uniffle/blob/master/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java and https://github.com/apache/celeborn/blob/3bf91929b6bd02974b5d15b4d6804c9b2b01cfc0/client-spark/spark-3/src/main/java/org/apache/spark/shuffle/celeborn/SparkShuffleManager.java#L165?

No, it's not intentionally a refactor against current shuffles. It's for preparing the base support of mixed backend(s): #6920 .

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRouter.scala
zhztheplayer: `ShuffleHandle` is not passing to `ShuffleBlockResolver` so it could be tricky if using it as key. Would you expect any potential issues if we use shuffle ID as the key? Maybe we need another solution to avoid them.

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
zhztheplayer: I realized it's not the case for double-checked locking... `router` will be set to `None` again when new shuffle manager is registered.

Perhaps we can simplify the logic in future by enclosing the registry once all the backends' initializations are done, I can take care of this part.

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRouter.scala
wForget: Will the mapping saved in `cache.store` also be synchronized to the executor? I am worried that calling `cache.get(shuffleId)` on the executor will not be able to find the ShuffleManager

zhztheplayer: I see. It's likely a problem. Let me test, thanks.

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
wForget: Makes sense

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRouter.scala
zhztheplayer: Perhaps a simple solution could be register `ShuffleHandle -> ShuffleManager` mappings as well which can be used for lookup on executor side.. Would you think that is feasible?

kerwin-zk:
> @kerwin-zk
> 
> > Moreover, from a certain perspective, the introduction of GlutenShuffleManager couples the various ShuffleManagers together, whereas the current fully decoupled setup seems clearer.
> 
> I haven't clearly gotten what's the new code coupling come from. Say ideally we can remove the proxied calls from RSS shuffle managers to default shuffle manager because of this work (each shuffle manager can be registered for certain shuffle dependency type it can handle). So I see this a decoupling of code rather than coupling of code.
> 
> > Essentially, this is no different from the current setup where spark.shuffle.manager is configured to different managers.
> 
> But I may understand your concern totally especially this one. The reason I am proposing this is not simply because of code refactor. The solution's kind of a trade-off for letting user enable different backends at the same time. See the umbrella #6920. For that purpose, we have to develop a way to make the shuffle managers be registered together so Gluten could choose the right one to use. For example, if we introduce GPU backend, we should give back the flexibility to developers when they decide not to let GPU backend rely on Velox backend, but they may still want query planner use Velox shuffle when the query plan is not offloaded to GPU.
> 
> May be we can put off deprecating the current use of `CelebornShuffleManager` / `UniffleShuffleManager`? We can keep the support for the legacy configurations during a considerable long period. I agree this is very normal that a user wants to specify a shuffle manager via Spark config explicitly. Does that work for you?

@zhztheplayer  Thank you very much for your detailed explanation. I believe that #6920  indeed requires this feature. At the same time, retaining the current option that allows users to specify the ShuffleManager through configuration is also necessary. I think this is acceptable.

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import org.apache.spark.SparkConf
+import org.apache.spark.util.Utils
+
+import scala.collection.mutable
+
+class ShuffleManagerRegistry private[ShuffleManagerRegistry] {
+  import ShuffleManagerRegistry._
+  private val all: mutable.Buffer[(LookupKey, String)] = mutable.Buffer()
+  private val routerBuilders: mutable.Buffer[RouterBuilder] = mutable.Buffer()
```
marin-ma, line 27: Could you explain in what cases there would be multiple RouterBuilders? Thanks!

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRouter.scala
zhztheplayer: @wForget I assume the current code is fine, given that usually shuffle dependency is carried by shuffle RDD and called from task code. Will run some tests later to prove.

https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/ShuffledRDD.scala

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRegistry.scala
zhztheplayer: I think we usually have one single router builder at the same time. So this practice can be simplified in future with some kind of assertions (something like `if (router created) throw error`). Though it's benign at the moment.

zhztheplayer: The reason we don't simply adopt a lazy initialization here is because the instantiation of `RouterBuilder` is parameterized: It takes `SparkConf` and `isDriver` in ctor. So we'd guard our code from the twice invocation with different parameters, no matter it's practically possible or not.

gluten-core/src/main/scala/org/apache/spark/shuffle/ShuffleManagerRouter.scala
wForget: > Perhaps a simple solution could be registering `ShuffleHandle -> ShuffleManager` mappings as well which can be used for lookup on executor side. Would you think that is feasible?

Sounds feasible

zhztheplayer:
cc @zzcclp @baibaichen

marin-ma (approved):

PR #8095, opened 2024-11-29
[GLUTEN-8094][CH][Part-1] Support reading data from the iceberg with CH backend

zzcclp:
Support reading data from the iceberg with CH backend

- basic iceberg scan transformer
- read from the iceberg with the copy-on-write mode

## How was this patch tested?

### DIFF
--- a/backends-clickhouse/src/test-iceberg/scala/org/apache/gluten/execution/iceberg/ClickHouseIcebergSuite.scala
+++ b/backends-clickhouse/src/test-iceberg/scala/org/apache/gluten/execution/iceberg/ClickHouseIcebergSuite.scala
@@ -1,0 +1,640 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gluten.execution.iceberg
+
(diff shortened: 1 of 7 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/TransformerApi.scala
```diff
@@ -76,4 +76,7 @@ trait TransformerApi {
   def invalidateSQLExecutionResource(executionId: String): Unit = {}
 
   def genWriteParameters(fileFormat: FileFormat, writeOptions: Map[String, String]): Any
+
+  /** use Hadoop Path class to encode the file path */
+  def encodeFilePathIfNeed(filePath: String): String = filePath
```
zhztheplayer, line 81: It turn out that a similar difference exists on regular scan

VL https://github.com/apache/incubator-gluten/blob/e56da333dc33de90540efa6dc63efa70c3c71ce8/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxIteratorApi.scala#L124-L126
CH https://github.com/apache/incubator-gluten/blob/e56da333dc33de90540efa6dc63efa70c3c71ce8/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala#L166

Maybe the logics should be somehow consolidated in future, I am not sure.

zhztheplayer (approved):
+1 on the common change, thanks.

backends-clickhouse/src/test-iceberg/scala/org/apache/gluten/execution/iceberg/ClickHouseIcebergSuite.scala
```diff
@@ -0,0 +1,640 @@
...
+ * limitations under the License.
+ */
+package org.apache.gluten.execution.iceberg
+
+import org.apache.gluten.execution.{GlutenClickHouseWholeStageTransformerSuite, IcebergScanTransformer}
+import org.apache.gluten.GlutenConfig
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.Row
+
+class ClickHouseIcebergSuite extends GlutenClickHouseWholeStageTransformerSuite {
```
liujiayi771, line 24: Is it possible to allow CH and Velox to share this part of the test cases?

zzcclp: will share this part after the CH backend support the merge on read mode for the iceberg, and there is also a bug when using the timestamp type as partition column for the CH backend.

PR #8107, opened 2024-11-29
[GLUTEN-8108][VL] Correct the logic of null on failure behavior for try cast

acvictor:
There are 3 failure behaviors defined for cast in Substrait
```
enum FailureBehavior { 
    FAILURE_BEHAVIOR_UNSPECIFIED = 0; 
    FAILURE_BEHAVIOR_RETURN_NULL = 1; 
    FAILURE_BEHAVIOR_THROW_EXCEPTION = 2; 
} 
```
Failure behavior should be set to 2 only when evaluation mode is ANSI. The current logic in Spark 3.4 and 3.5 uses the evaluation of `ansiEnabled = evalMode == EvalMode.ANSI || (evalMode == EvalMode.TRY && !canUseLegacyCastForTryCast)` to set the failure behavior and is incorrectly evaluated in some cases depending on the value of `canUseLegacyCastForTryCast`. `canUseLegacyCastForTryCast` should not matter in Velox/Gluten since it is used in Spark to optimize execution by avoiding adding a try-catch block around the cast.

A minimal repro shows that 
select **try_cast**(value as bigint) from tbl translates to Velox plan **cast**((trim( :VARCHAR, n0_0)) as BIGINT), implying that failure behavior is incorrectly set to 2 and so `SparkCastExpr` in Velox is created with `nullOnFailure_` set to false instead of true https://github.com/facebookincubator/velox/blob/main/velox/expression/ExprCompiler.cpp#L411.

(Fixes: \#ISSUE-ID https://github.com/apache/incubator-gluten/issues/8108)

## How was this patch tested?
Unit tests.


acvictor:
@zhli1142015 @rui-mo can you please review?

gluten-substrait/src/main/scala/org/apache/gluten/expression/UnaryExpressionTransformer.scala
```diff
@@ -43,7 +43,10 @@ case class CastTransformer(substraitExprName: String, child: ExpressionTransform
   extends UnaryExpressionTransformer {
   override def doTransform(args: java.lang.Object): ExpressionNode = {
     val typeNode = ConverterUtils.getTypeNode(dataType, original.nullable)
-    ExpressionBuilder.makeCast(typeNode, child.doTransform(args), original.ansiEnabled)
+    ExpressionBuilder.makeCast(
+      typeNode,
+      child.doTransform(args),
+      original.evalMode == EvalMode.ANSI)
```
zhli1142015, line 49: This is added in spark 3.4. I think you only need to fix the behavior in spark 3.4 and after.

backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
```diff
@@ -1457,4 +1457,29 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
         checkGlutenOperatorMatch[FilterExecTransformer](df)
     }
   }
+
+  test("Test cast and try_cast") {
```
zhli1142015, line 1461: Please add separate test case for each function

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenCastWithAnsiOnSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+package org.apache.spark.sql.catalyst.expressions
+
+import org.apache.spark.sql.GlutenTestsTrait
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.DataType
+
+class GlutenCastWithAnsiOnSuite extends CastWithAnsiOnSuite with GlutenTestsTrait {
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    SQLConf.get.setConf(SQLConf.ANSI_ENABLED, true)
```
zhli1142015, line 26: If ANSI is on, every thing should fallback, can you clarify what's this test suite for?

gluten-substrait/src/main/scala/org/apache/gluten/expression/UnaryExpressionTransformer.scala
zhli1142015: https://github.com/apache/incubator-gluten/blob/main/shims/spark35/src/main/scala/org/apache/gluten/sql/shims/spark35/Spark35Shims.scala#L485

backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
acvictor: Done.

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenCastWithAnsiOnSuite.scala
acvictor: I have removed this, thanks.

gluten-substrait/src/main/scala/org/apache/gluten/expression/UnaryExpressionTransformer.scala
acvictor: Thank you! Used this shim function.

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
```diff
@@ -92,6 +92,16 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude(
       "Process Infinity, -Infinity, NaN in case insensitive manner" // +inf not supported in folly.
     )
+  enableSuite[GlutenTryCastSuite]
+    .exclude(
+      "Process Infinity, -Infinity, NaN in case insensitive manner" // +inf not supported in folly.
+    )
+    .exclude("ANSI mode: Throw exception on casting out-of-range value to byte type")
```
rui-mo, line 99: Would you add some comments on why these tests are excluded? I remember ANSI ON causes execution fallback, can the tests produce Spark expected result?

rui-mo (commented):
Thanks.

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
acvictor: These tests come from https://github.com/apache/spark/blob/master/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/CastWithAnsiOnSuite.scala and when the suite is run separately with `SQLConf.get.setConf(SQLConf.ANSI_ENABLED, true)` falls back but when run as part of `GlutenTryCastSuite` where the conf is not set does not give expected results. (Also, the tests fail without the code changes in this PR.)

rui-mo: > the conf is not set does not give expected results

Thanks for providing the details. These tests are for ANSI OFF cases, right?

rui-mo (approved):

zhli1142015 (approved):
LGTM, thanks

zhli1142015:

gluten-substrait/src/main/scala/org/apache/gluten/expression/UnaryExpressionTransformer.scala
```diff
@@ -43,7 +44,10 @@ case class CastTransformer(substraitExprName: String, child: ExpressionTransform
   extends UnaryExpressionTransformer {
   override def doTransform(args: java.lang.Object): ExpressionNode = {
     val typeNode = ConverterUtils.getTypeNode(dataType, original.nullable)
-    ExpressionBuilder.makeCast(typeNode, child.doTransform(args), original.ansiEnabled)
+    ExpressionBuilder.makeCast(
+      typeNode,
+      child.doTransform(args),
+      SparkShimLoader.getSparkShims.withAnsiEvalMode(original))
```
PHILO-HE, line 50: Yes, I note Spark also sets `ansiEnabled = true` for `EvalMode.TRY`, which is for re-using the code logic of `EvalMode.ANSI`. The difference is, in TRY mode any exception is caught and then NULL is returned.
Velox has different implementation, which requires us to set this flag simply according to whether EvalMode is ANSI or not.

@acvictor, please leave some comments here to clarify, which should be helpful for future code maintenance.

https://github.com/apache/spark/blob/v3.5.3/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L471

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
PHILO-HE: These tests are for ANSI ON.
@acvictor, for these tests in TryCastSuite, exceptions are not thrown, but null results are returned? It should be an expected behavior for Velox cast.
I cannot figure out how these exception checks can pass in vanilla Spark, assuming exception is handled internally in TRY mode.

acvictor: @PHILO-HE I don't fully understand what you mean. Are you saying Velox behaviour is incorrect?

PHILO-HE: @acvictor, please rebase the code. The core change looks good to me.

acvictor: Updated, thank you!

PHILO-HE (approved):
Thanks!

PR #8112, opened 2024-11-30
[GLUTEN-1632][CH]Daily Update Clickhouse Version (20241130)

kyligence-git:
Auto commit by gluten daily build, please check the build status and merge it if it's green.


lgbo-ustc:
It may be safer to  let `max_bytes_ratio_before_external_group_by` be 0.0. We cannot let the spill happen inside  `Aggregator`,  it is conflict with `GraceAggregation`

baibaichen (approved):

Issue #8115, opened 2024-12-01
[VL] BuildSideRelation supports transforms for all scenarios


yikf:
### Backend

VL (Velox)

### Bug description

If exchange reuse, the `buildKeys` of `ColumnarSubqueryBroadcastExec` and the `buildKeys` of `BroadcastExchangeExec` may not be the same, making it difficult to construct the projection of the transform. therefore, the `BuildSideRelation` should retain the `BroadcastMode` of `BroadcastExchangeExec` to simplify the projection of the transform.

### Spark version

None

PHILO-HE:
Assume this is relevant to this discussion: https://github.com/apache/incubator-gluten/issues/7807.

Issue #7807, opened 2024-11-04
[VL] TPCDS q05 failed due to exception in ColumnarBuildSideRelation


leoluan2009:
### Backend

VL (Velox)

### Bug description

error message:
2024-11-04 22:41:56,466 [ERROR] [dynamicpruning-0] TaskResources: Task -1 failed by error: 
java.lang.IllegalStateException: Couldn't find d_date_sk#376 in [d_date_sk#103]
        at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1.applyOrElse(BoundAttribute.scala:80)
        at org.apache.spark.sql.catalyst.expressions.BindReferences$$anonfun$bindReference$1.applyOrElse(BoundAttribute.scala:73)
        at org.apache.spark.sql.catalyst.trees.TreeNode.$anonfun$transformDownWithPruning$1(TreeNode.scala:461)
        at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(origin.scala:76)
        at org.apache.spark.sql.catalyst.trees.TreeNode.transformDownWithPruning(TreeNode.scala:461)
        at org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:437)
        at org.apache.spark.sql.catalyst.trees.TreeNode.transform(TreeNode.scala:405)
        at org.apache.spark.sql.catalyst.expressions.BindReferences$.bindReference(BoundAttribute.scala:73)
        at org.apache.spark.sql.catalyst.expressions.BindReferences$.$anonfun$bindReferences$1(BoundAttribute.scala:94)
        at scala.collection.immutable.List.map(List.scala:293)
        at org.apache.spark.sql.catalyst.expressions.BindReferences$.bindReferences(BoundAttribute.scala:94)
        at org.apache.spark.sql.catalyst.expressions.UnsafeProjection$.create(Projection.scala:161)
        at org.apache.spark.sql.execution.ColumnarBuildSideRelation$$anon$2.next(ColumnarBuildSideRelation.scala:144)
        at org.apache.spark.sql.execution.ColumnarBuildSideRelation$$anon$2.next(ColumnarBuildSideRelation.scala:111)
        at scala.collection.Iterator$$anon$10.next(Iterator.scala:461)
        at scala.collection.TraversableOnce$FlattenOps$$anon$2.hasNext(TraversableOnce.scala:521)
[...]

PR #8116, opened 2024-12-01
[GLUTEN-8115][CORE] Refine the BuildSideRelation transform to support all scenarios

yikf:
Fix https://github.com/apache/incubator-gluten/issues/8115.

If exchange reuse, the `buildKeys` of `ColumnarSubqueryBroadcastExec` and the `buildKeys` of `BroadcastExchangeExec` may not be the same, making it difficult to construct the projection of the transform. therefore, the `BuildSideRelation` should retain the `BroadcastMode` of `BroadcastExchangeExec` to simplify the projection of the transform.

## How was this patch tested?
exists unit tests, manual tests


yikf:
@PHILO-HE @zhztheplayer @jackylee-ch @zzcclp Could you please take a look if you find a time, thanks

FelixYBW:
Do you have a specific query example?

yikf:
> Do you have a specific query example?

TPC-DS Q5 is a reuse-exchange example, see: https://github.com/apache/incubator-gluten/issues/7807

After reuse, ColumnarSubqueryBroadcastExec's buildKeys and exchange's buildKeys may not be the same reference, and in our internal case, even the names referenced by alias are different because of alias. 

In Spark, the transform of relation is [always carried out in BroadcastExchangeExec](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/exchange/BroadcastExchangeExec.scala#L193), so this problem does not exist. we can save the buildKeys of BroadcastExchangeExec. this way, in the subsequent transform, we can do the relation transform like spark would have done in BroadcastExchangeExec, in which case the references are always the same, and then we can take the desired index value from the buildKeys output.

yikf:
@FelixYBW @zhouyuan @PHILO-HE Hi, Could you help please take a look, thanks.

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/ColumnarSubqueryBroadcastExec.scala
```diff
@@ -64,6 +65,14 @@ case class ColumnarSubqueryBroadcastExec(
     copy(name = "native-dpp", buildKeys = keys, child = child.canonicalized)
   }
 
+  // Copy from org.apache.spark.sql.execution.joins.HashJoin#canRewriteAsLongType
+  // we should keep consistent with it to identify the LongHashRelation.
+  private def canRewriteAsLongType(keys: Seq[Expression]): Boolean = {
```
PHILO-HE, line 70: @yikf, if spark changes this logic, e.g., for adding support for other types, unexpected failure can happen in Gluten? Seems there is no other way.

PHILO-HE (approved):
Verified with internal TPC-DS 3TB test.

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/ColumnarSubqueryBroadcastExec.scala
yikf: There doesn't seem to be any other way at the moment.

PHILO-HE: @yikf, ok, let's merge this pr firstly. Thanks for your efforts!

yikf: thanks @PHILO-HE

PR #8121, opened 2024-12-02
Revert "[GLUTEN-8080][CH]Support function transform_keys/transform_values"

taiyang-li:
Reverts apache/incubator-gluten#8085 because of 
image


zzcclp (approved):
+1

PR #8126, opened 2024-12-03
[CORE][MIRROR] Fix performance issue when allScanPartitions are very large

WangGuangxin:

When the scanned partitions are very large, the random access of seq maybe a performance bottleneck when generating scan rdd, which will cause driver stuck for a long time

img_v3_02h6_8e8b3890-60e0-4e8f-a509-7ecb7e1ae1cg

img_v3_02h6_20d195db-23eb-429a-864a-460f8b8881dg

## How was this patch tested?

Mannually


PHILO-HE:
@WangGuangxin, have you verified that this patch can help alleviate the issue you mentioned?

zhztheplayer (approved):
> @WangGuangxin, have you verified that this patch can help alleviate the issue you mentioned?

Curious on this too. Would you share some? Thanks.

Code change looks reasonable to me. 👍

WangGuangxin:
> > @WangGuangxin, have you verified that this patch can help alleviate the issue you mentioned?
> 
> Curious on this too. Would you share some? Thanks.
> 
> Code change looks reasonable to me. 👍

We have a job stuck on this stage for more than 20min, and resolved with this patch. This is the flame

flame

PHILO-HE:
@WangGuangxin, thanks so much for your feedback!

Issue #6947, opened 2024-08-20
[VL] task killed by Yarn


FelixYBW:
### Backend

VL (Velox)

### Bug description

one common issue of Gluten is that the task is killed by yarn. Currently Gluten has some memory allocation like std::vector isn't tracked by spark's memory management. It's counted into executor.overhead memory now. But there should be some such allocation with large data size. We should create a proxy allocator for such memory allocation.

### Spark version

None

Yohahaha:
We may need provide the ability to proxy malloc/free in Gluten, and inject this proxy to Velox MemoryAllocator.

FelixYBW:
Related to #6960. The golbal memory allocation isn't tracked either, which doesn't have a limit now. It's mostly the root cause.

Thank @zhztheplayer and @marin-ma

FelixYBW:
PR6988 added the limitation of global memory allocation, 0.75xoverhead memory. But Velox has no control of the global memory allocation. So with PR6988, we can workaround to set a large overhead memory, otherwise you will see the `Exceeded memory allocator limit` exception. Though not kill by yarn anymore.

FelixYBW:
https://github.com/FelixYBW/gluten/commit/fbfe0f30b99d8fb759c9eb447968af4ab7a3f4dd

Here is a simple way to detect this. It output the memory allocation bigger than 1M but not from memory pool. disable jemalloc, set executor.core to 1, then you can get output like 
```
allocated 67108864
libvelox.so(+0x15b4ff5) [0x7f3301cf2ff5]
libvelox.so(+0x15b51eb) [0x7f3301cf31eb]
libgluten.so(_Znwm+0x19) [0x7f3307be9ea9]
libvelox.so(_ZNSt6vectorISt10shared_ptrIN8facebook5velox4exec15WindowPartitionEESaIS5_EE17_M_realloc_insertIJS5_EEEvN9__gnu_cxx17__normal_iteratorIPS5_S7_EEDpOT_+0x80) [0x7f3305130180]
libvelox.so(_ZN8facebook5velox4exec24RowsStreamingWindowBuild18addPartitionInputsEb+0x1cf) [0x7f330512f18f]
libvelox.so(_ZN8facebook5velox4exec24RowsStreamingWindowBuild8addInputESt10shared_ptrINS0_9RowVectorEE+0x419) [0x7f330512f799]
[...]

PR #8127, opened 2024-12-03
[GLUTEN-7750][VL] Move ColumnarBuildSideRelation's memory occupation to Spark off-heap

zjuwangg:
Implement feature: #7750 

## How was this patch tested?

Add unit test and manual run tpcds in my dev box.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -45,12 +45,14 @@ import org.apache.spark.sql.execution.exchange.{BroadcastExchangeExec, ShuffleEx
 import org.apache.spark.sql.execution.joins.{BuildSideRelation, HashedRelationBroadcastMode}
 import org.apache.spark.sql.execution.metric.SQLMetric
 import org.apache.spark.sql.execution.python.ArrowEvalPythonExec
+import org.apache.spark.sql.execution.unsafe.UnsafeColumnarBuildSideRelation
 import org.apache.spark.sql.execution.utils.ExecUtil
 import org.apache.spark.sql.expression.{UDFExpression, UserDefinedAggregateFunction}
 import org.apache.spark.sql.hive.VeloxHiveUDFTransformer
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.vectorized.ColumnarBatch
+import org.apache.spark.task.TaskResources
 
 import org.apache.commons.lang3.ClassUtils
 
@@ -621,6 +623,8 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
       child: SparkPlan,
       numOutputRows: SQLMetric,
       dataSize: SQLMetric): BuildSideRelation = {
+    val useOffheapBroadcastBuildRelation =
+      GlutenConfig.getConf.enableBroadcastBuildRelationInOffheap
     val serialized: Array[ColumnarBatchSerializeResult] = child
       .executeColumnar()
       .mapPartitions(itr => Iterator(BroadcastUtils.serializeStream(itr)))
@@ -633,7 +637,13 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
     }
     numOutputRows += serialized.map(_.getNumRows).sum
     dataSize += rawSize
-    ColumnarBuildSideRelation(child.output, serialized.map(_.getSerialized), mode)
+    if (useOffheapBroadcastBuildRelation) {
+      TaskResources.runUnsafe {
+        new UnsafeColumnarBuildSideRelation(child.output, serialized.map(_.getSerialized), mode)
+      }
+    } else {
+      ColumnarBuildSideRelation(child.output, serialized.map(_.getSerialized), mode)
+    }
   }
 
   override def doCanonicalizeForBroadcastMode(mode: BroadcastMode): BroadcastMode = {
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.execution.unsafe
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.memory.{MemoryConsumer, MemoryMode, TaskMemoryManager}
+import org.apache.spark.unsafe.Platform
+import org.apache.spark.unsafe.array.LongArray
+
+/**
+ * Used to store broadcast variable off-heap memory for broadcast variable. The underlying data
+ * structure is a LongArray allocated in off-heap memory.
+ *
+ * @param arraySize
+ *   underlying array[array[byte]]'s length
+ * @param bytesBufferLengths
+ *   underlying array[array[byte]] per bytesBuffer length
+ * @param totalBytes
+ *   all bytesBuffer's length plus together
+ */
+// scalastyle:off no.finalize
+case class UnsafeBytesBufferArray(
+    arraySize: Int,
+    bytesBufferLengths: Array[Int],
+    totalBytes: Long,
+    tmm: TaskMemoryManager)
+  extends MemoryConsumer(tmm, MemoryMode.OFF_HEAP)
+  with Logging {
+
+  {
+    assert(
+      arraySize == bytesBufferLengths.length,
+      "Unsafe buffer array size " +
+        "not equal to buffer lengths!")
+    assert(totalBytes >= 0, "Unsafe buffer array total bytes can't be negative!")
+  }
+
+  /**
+   * A single array to store all bytesBufferArray's value, it's inited once when first time get
+   * accessed.
+   */
+  private var longArray: LongArray = _
+
+  /** Index the start of each byteBuffer's offset to underlying LongArray's initial position. */
+  private val bytesBufferOffset = if (bytesBufferLengths.isEmpty) {
+    new Array(0)
+  } else {
+    bytesBufferLengths.init.scanLeft(0)(_ + _)
+  }
+
+  override def spill(l: Long, memoryConsumer: MemoryConsumer): Long = 0L
+
+  /**
+   * Put bytesBuffer at specified array index.
+   *
+   * @param index
+   *   index of the array.
+   * @param bytesBuffer
+   *   bytesBuffer to put.
+   */
+  def putBytesBuffer(index: Int, bytesBuffer: Array[Byte]): Unit = this.synchronized {
+    assert(index < arraySize)
+    assert(bytesBuffer.length == bytesBufferLengths(index))
+    // first to allocate underlying long array
+    if (null == longArray && index == 0) {
+      longArray = allocateArray((totalBytes + 7) / 8)
+    }
+
+    Platform.copyMemory(
+      bytesBuffer,
+      Platform.BYTE_ARRAY_OFFSET,
+      longArray.getBaseObject,
+      longArray.getBaseOffset + bytesBufferOffset(index),
+      bytesBufferLengths(index))
+  }
+
+  /**
+   * Get bytesBuffer at specified index.
+   * @param index
+   * @return
+   */
+  def getBytesBuffer(index: Int): Array[Byte] = {
+    assert(index < arraySize)
+    if (null == longArray) {
+      return new Array[Byte](0)
+    }
+    val bytes = new Array[Byte](bytesBufferLengths(index))
+    log.warn(s"get bytesBuffer at index $index bytesBuffer length ${bytes.length}")
+    Platform.copyMemory(
+      longArray.getBaseObject,
+      longArray.getBaseOffset + bytesBufferOffset(index),
+      bytes,
+      Platform.BYTE_ARRAY_OFFSET,
+      bytesBufferLengths(index))
+    bytes
+  }
+
+  /**
+   * Get the bytesBuffer memory address and length at specified index, usually used when read memory
+   * direct from offheap.
+   *
+   * @param index
+   * @return
+   */
+  def getBytesBufferOffsetAndLength(index: Int): (Long, Int) = {
+    assert(index < arraySize)
+    assert(longArray != null, "The broadcast data in offheap should not be null!")
+    val offset = longArray.getBaseOffset + bytesBufferOffset(index)
+    val length = bytesBufferLengths(index)
+    (offset, length)
+  }
+
+  /**
+   * It's needed once the broadcast variable is garbage collected. Since now, we don't have an
+   * elegant way to free the underlying memory in offheap.
+   */
+  override def finalize(): Unit = {
+    try {
+      if (longArray != null) {
+        freeArray(longArray)
+        longArray = null
+      }
+    } finally {
+      super.finalize()
+    }
+  }
+}
+// scalastyle:on no.finalize
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
@@ -16,6 +16,7 @@
  */
 package org.apache.spark.sql.execution.unsafe
 
+import org.apache.spark.annotation.Experimental
 import org.apache.spark.internal.Logging
 import org.apache.spark.memory.{MemoryConsumer, MemoryMode, TaskMemoryManager}
 import org.apache.spark.unsafe.Platform
@@ -33,6 +34,7 @@ import org.apache.spark.unsafe.array.LongArray
  *   all bytesBuffer's length plus together
  */
 // scalastyle:off no.finalize
+@Experimental
 case class UnsafeBytesBufferArray(
     arraySize: Int,
     bytesBufferLengths: Array[Int],
@@ -99,7 +101,6 @@ case class UnsafeBytesBufferArray(
       return new Array[Byte](0)
     }
     val bytes = new Array[Byte](bytesBufferLengths(index))
-    log.warn(s"get bytesBuffer at index $index bytesBuffer length ${bytes.length}")
     Platform.copyMemory(
       longArray.getBaseObject,
       longArray.getBaseOffset + bytesBufferOffset(index),
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
@@ -26,6 +26,7 @@ import org.apache.gluten.utils.ArrowAbiUtil
 import org.apache.gluten.vectorized.{ColumnarBatchSerializerJniWrapper, NativeColumnarToRowJniWrapper}
 
 import org.apache.spark.{SparkEnv, TaskContext}
+import org.apache.spark.annotation.Experimental
 import org.apache.spark.internal.Logging
 import org.apache.spark.memory.TaskMemoryManager
 import org.apache.spark.sql.catalyst.InternalRow
@@ -56,6 +57,7 @@ import scala.collection.JavaConverters.asScalaIteratorConverter
  * @param mode
  *   the broadcast mode.
  */
+@Experimental
 case class UnsafeColumnarBuildSideRelation(
     private var output: Seq[Attribute],
     private var batches: UnsafeBytesBufferArray,
@@ -135,7 +137,6 @@ case class UnsafeColumnarBuildSideRelation(
       val length = bytesBufferLengths(i)
       val tmpBuffer = new Array[Byte](length)
       in.read(tmpBuffer)
-      log.warn(s"readExternal $i--- $length")
       batches.putBytesBuffer(i, tmpBuffer)
     }
   }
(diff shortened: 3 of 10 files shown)


shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -2246,4 +2249,12 @@ object GlutenConfig {
...
       .booleanConf
       .createWithDefault(false)
+
+  val VELOX_BROADCAST_BUILD_RELATION_USE_OFFHEAP =
+    buildConf("spark.gluten.velox.BroadcastBuildRelationUseOffheap.enabled")
+      .internal()
+      .doc("If enabled, broadcast build relation will use offheap memory. " +
+        "Otherwise, broadcast build relation will use onheap memory.")
+      .booleanConf
+      .createWithDefault(false)
```
zhouyuan, line 2259: it's better to make this optimization as default on

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
```diff
@@ -0,0 +1,312 @@
...
+  override def readExternal(in: ObjectInput): Unit = Utils.tryOrIOException {
+    output = in.readObject().asInstanceOf[Seq[Attribute]]
+    mode = in.readObject().asInstanceOf[BroadcastMode]
+    val totalArraySize = in.readInt()
+    val bytesBufferLengths = in.readObject().asInstanceOf[Array[Int]]
+    val totalBytes = in.readLong()
+
+    val taskMemoryManager = new TaskMemoryManager(
+      new UnifiedMemoryManager(SparkEnv.get.conf, Long.MaxValue, Long.MaxValue / 2, 1),
+      0)
```
zhztheplayer, line 129: Would you like to explain a bit against this code? Thanks!

gluten-arrow/src/main/java/org/apache/gluten/vectorized/ColumnarBatchSerializerJniWrapper.java
```diff
@@ -42,5 +42,8 @@ public long rtHandle() {
 
   public native long deserialize(long serializerHandle, byte[] data);
 
+  // Return the native ColumnarBatch handle using memory address and length
+  public native long deserializeDirectAddress(long serializerHandle, long offset, int len);
```
zhztheplayer, line 46: nit: Can we rename the method with `deserializeDirect`?

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
```diff
@@ -0,0 +1,312 @@
...
+      case HashedRelationBroadcastMode(k, _) => UnsafeProjection.create(k)
+      case IdentityBroadcastMode => UnsafeProjection.create(output, output)
+    }
+  }
+
+  override def deserialized: Iterator[ColumnarBatch] = {
+    val runtime =
+      Runtimes.contextInstance(BackendsApiManager.getBackendName, "BuildSideRelation#transform")
+    val jniWrapper = ColumnarBatchSerializerJniWrapper.create(runtime)
+    val serializeHandle: Long = {
```
zhztheplayer, line 177: nit: serializeHandle -> serializerHandle

zhouyuan:
CC @surnaik

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -2246,4 +2249,12 @@ object GlutenConfig {
       .doc("If enabled, gluten will convert the viewfs path to hdfs path in scala side")
       .booleanConf
       .createWithDefault(false)
+
+  val VELOX_BROADCAST_BUILD_RELATION_USE_OFFHEAP =
+    buildConf("spark.gluten.velox.BroadcastBuildRelationUseOffheap.enabled")
```
zhztheplayer, line 2254: nit: Let's use camel case.

Also could rename with `spark.gluten.velox.offHeapBroadcastBuildRelation`? Note `offHeap` is also camel case to match vanilla Spark's naming convention.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
```diff
@@ -305,6 +305,18 @@ class VeloxTPCHV1BhjSuite extends VeloxTPCHSuite {
...
+class VeloxTPCHV1BhjOffheapSuite extends VeloxTPCHSuite {
+  override def subType(): String = "v1-bhj-offheap"
+
+  override protected def sparkConf: SparkConf = {
+    super.sparkConf
+      .set("spark.sql.sources.useV1SourceList", "parquet")
+      .set("spark.sql.autoBroadcastJoinThreshold", "30M")
+      .set("spark.gluten.velox.BroadcastBuildRelationUseOffheap.enabled", "true")
+  }
+}
```
zhztheplayer, line 318: Is the test still needed? Given off-heap relation is enabled by default. Perhaps add a suite for on-heap relation?

zhztheplayer (commented):
Just roughly went though the PR. Seems in good shape already. 

Saw CI is failing BTW.

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
```diff
@@ -0,0 +1,167 @@
...
+  /** Index the start of each byteBuffer's offset to underlying LongArray's initial position. */
+  private val bytesBufferOffset = new Array[Int](arraySize)
+
+  {
+    assert(arraySize == bytesBufferLengths.length)
+    assert(totalBytes >= 0)
+    if (arraySize > 0) {
+      for (i <- 0 until arraySize) {
+        bytesBufferOffset(i) =
+          if (i > 0) bytesBufferLengths(i - 1) + bytesBufferOffset(i - 1) else 0
```
liujiayi771, line 61: Perhaps using `scanLeft` can simplify the code.
```scala
val bytesBufferOffset = bytesBufferLengths.init.scanLeft(0)( _ + _)
```

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
```diff
@@ -0,0 +1,167 @@
...
+  with Logging {
+
+  /**
+   * A single array to store all bytesBufferArray's value, it's inited once when first time get
+   * accessed.
+   */
+  private var longArray: LongArray = _
+
+  /** Index the start of each byteBuffer's offset to underlying LongArray's initial position. */
+  private val bytesBufferOffset = new Array[Int](arraySize)
```
liujiayi771, line 53: Can we use `scanLeft` to simplify the code, and only assert that `bytesBufferLengths` is not empty.
```scala
val bytesBufferOffset = bytesBufferLengths.init.scanLeft(0)( _ + _)
```

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
zjuwangg: This is used in Broadcast, shared by multiple tasks, so here introduce a new TaskMemoryManager instead of use task-related TaskMemoryManager.
Similar to `UnsafeHashRelation` in https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/joins/HashedRelation.scala#L389-L410

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
zjuwangg: addressed

zjuwangg:
Thanks for your guys detailed review. @zhztheplayer @liujiayi771 @zhouyuan 
Just addressed the comments and will work on fixing the failed ut later.

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
```diff
@@ -0,0 +1,165 @@
...
+    arraySize: Int,
+    bytesBufferLengths: Array[Int],
+    totalBytes: Long,
+    tmm: TaskMemoryManager)
+  extends MemoryConsumer(tmm, MemoryMode.OFF_HEAP)
+  with Logging {
+
+  {
+    assert(arraySize == bytesBufferLengths.length)
+    assert(totalBytes >= 0)
```
jackylee-ch, line 48: nit：add some message when assert failed, so we can find the reason from logs

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
```diff
@@ -0,0 +1,165 @@
...
+      longArray.getBaseOffset + bytesBufferOffset(index),
+      bytesBufferLengths(index))
+  }
+
+  /**
+   * Get bytesBuffer at specified index.
+   * @param index
+   * @return
+   */
+  def getBytesBuffer(index: Int): Array[Byte] = {
```
jackylee-ch, line 98: do we need a `ReadWriteLock` here?

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
```diff
@@ -0,0 +1,328 @@
...
+        bytesBufferArray.map(_.length),
+        bytesBufferArray.map(_.length.toLong).sum,
+        TaskContext.get().taskMemoryManager
+      ),
+      mode
+    )
+    val batchesSize = bytesBufferArray.length
+    for (i <- 0 until batchesSize) {
+      val length = bytesBufferArray(i).length
+      log.debug(s"this $i--- $length")
```
jackylee-ch, line 87: this seems like not a good log, can we delete it or refine it?

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -2250,4 +2253,12 @@ object GlutenConfig {
...
       .booleanConf
       .createWithDefault(false)
+
+  val VELOX_BROADCAST_BUILD_RELATION_USE_OFFHEAP =
+    buildConf("spark.gluten.velox.offHeapBroadcastBuildRelation.enabled")
+      .internal()
+      .doc("If enabled, broadcast build relation will use offheap memory. " +
+        "Otherwise, broadcast build relation will use onheap memory.")
+      .booleanConf
+      .createWithDefault(true)
```
yikf, line 2263: Let's default it to false, and need a smooth transition period?

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
```diff
@@ -305,6 +305,18 @@ class VeloxTPCHV1BhjSuite extends VeloxTPCHSuite {
   }
 }
 
+/** BroadcastBuildSideRelation use onheap. */
+class VeloxTPCHV1BhjOnheapSuite extends VeloxTPCHSuite {
+  override def subType(): String = "v1-bhj-offheap"
```
yikf, line 310: v1-bhj-onheap?

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxHashJoinSuite.scala
```diff
@@ -114,71 +115,84 @@ class VeloxHashJoinSuite extends VeloxWholeStageTransformerSuite {
...
-                     |WHERE t1.date=if(3 <= t2.t2_c2, if(3 < t2.t2_c1, 3, t2.t2_c1), t2.t2_c2)
-                     |ORDER BY t1.date DESC, t1.t1_c1 DESC, t2.t2_c2 DESC
-                     |LIMIT 1
-                     |""".stripMargin)
-
-      checkAnswer(df, Row("3", 1, 4) :: Nil)
-      // collect the DPP plan.
-      val subqueryBroadcastExecs = collectWithSubqueries(df.queryExecution.executedPlan) {
-        case subqueryBroadcast: ColumnarSubqueryBroadcastExec => subqueryBroadcast
+    for (enabledOffheapBroadcast <- Seq("true", "false")) {
```
yikf, line 158:
```scala
Seq(true, false).foreach { enabledOffheapBroadcast =>
    xxx
}
```

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
```diff
@@ -0,0 +1,328 @@
...
+import java.io.{Externalizable, ObjectInput, ObjectOutput}
+
+import scala.collection.JavaConverters.asScalaIteratorConverter
+
+/**
+ * UnsafeColumnarBuildSideRelation should backed by offheap to avoid on-heap oom. Almost the same as
+ * ColumnarBuildSideRelation, we should remove ColumnarBuildSideRelation when
+ * UnsafeColumnarBuildSideRelation get matured.
+ *
+ * @param output
```
yikf, line 55: nit, need comments.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -633,7 +637,13 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
     }
     numOutputRows += serialized.map(_.getNumRows).sum
     dataSize += rawSize
-    ColumnarBuildSideRelation(child.output, serialized.map(_.getSerialized), mode)
+    if (useOffheapBroadcastBuildRelation) {
+      TaskResources.runUnsafe {
```
yikf, line 641: Do we still need `TaskResources.runUnsafe`?

yikf (approved):
LGTM overall, with a few minor comments. I think it has a good shape to merging if CI passed.

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
zjuwangg: Don't need here since when broadcast variable get constructed only once and there be will a write lock in Spark framework.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
zjuwangg: good catch!

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
zjuwangg: Yes. Since this method will be called in driver side and there will be no TaskContext aviable.

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
zhztheplayer: Thanks for the explanation. And I need to apologize that my previous comment https://github.com/apache/incubator-gluten/issues/7750#issuecomment-2451909071 was wrong, because we are apparently facing the similar issue with Gazelle project here again.

I see `finalize()` is used and I assume it's working in your test, right? And given that invocations on `finalize()` relies on JVM GC, how do we make sure the off-heap memory can be released to avoid off-heap overhead, even JVM GC is never triggered?

This code (line :127 - :129) looks feasible but the allocation is still not managed by the default memory manager. Do we have a better approach? Otherwise [Yarn kill](https://github.com/apache/incubator-gluten/issues/6947) issue will be led.

We can revisit the topic and the code a little bit and hopefully we can come up with an optimal solution. My description in https://github.com/apache/incubator-gluten/issues/7750 was indeed a bit misleading, this issue doesn't look to have some kind of `simple` solution so far.

yikf: @zhztheplayer Spark uses the GC mechanism of Cleaner to clean up broadcasts. `finalize()` should also be a feasible solution, but it cannot be actively cleaned up when offheap is insufficient. As for the need for a unified memory manager to manage this memory, we can directly use SparkEnv.get.memoryManager. see https://github.com/apache/incubator-gluten/pull/8349

yikf: Edit: 

Broadcast variables can usually be assumed to be small enough, and the way of `finalize()` is roughly acceptable (maybe I'm wrong). Later, we can look for a more suitable solution, such as implementing the spill of UnsafeBytesBufferArray?

yikf: @zhztheplayer Hi, terry is on vacation recently. What's your opinion on this idea? I can continue to follow up.

lwz9103:

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
zhztheplayer: @yikf Thank you for continuing the work.  I was on vacation either today so sorry for the late reply.

> @zhztheplayer Hi, terry is on vacation recently. What's your opinion on this idea? I can continue to follow up.

I was thinking about using storage memory but the implementation is complicated? If so let's put that idea on hold.

> we can directly use SparkEnv.get.memoryManager.

👍 Sounds good to me if it can utilize Spark's off-heap memory, since at least the memory is tracked. Though the release could still be a problem? Because Spark doesn't do `System.gc()` on off-heap OOM. 

If the issue does apply, maybe we can consider one of the following:

Option 1. Hack into Spark to make sure `System.gc()` is triggered on OOM. (if there's an apporach without modifications on vanilla Spark's code)
Option 2. Set a fixed capacity (e.g., 15% of off-heap memory) for the outstanding task memory manager used in the PR. When this part of memory is run out, trigger `System.gc()`, then if still unsatisified, throw an OOM.

The above are just based on my assumption. Let me know if any other possible solutions.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
zhztheplayer: Yes, let's mark the doc with `(experimental)` prefix as well. Also mark `UnsafeBytesBufferArray` / `UnsafeColumnarBuildSideRelation` with `@Experimental` annotations.

zhztheplayer (commented):
Basically only the releasing issue remaining, from my perspective. Perhaps we can find a temporary solution then merge this next week. Thanks everyone.

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
yikf: @zhztheplayer There is another idea whose feasibility is uncertain. 

Spark uses the GC mechanism on the Driver side to [clean broadcast variables](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/ContextCleaner.scala#L205)through ContextCleaner and supports the [attachListener](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/ContextCleaner.scala#L123). We can fully utilize this mechanism.

Prerequisite,
1. Implement our own set of endpoints for the off-heap broadcast cleanup.
2. Setup the endpoint if use off-heap broadcast.
3. `UnsafeColumnarBuildSideRelation` record the broadcast id.
4. Implement ContextCleaner listener.
5. deserialize the `UnsafeColumnarBuildSideRelation` at executor side, use a singleton class record broadcast id and the `UnsafeColumnarBuildSideRelation` (read external). 

Then,
When the Driver triggers broadcast clearing, the listener catchs the event and sends the clearing request to the executor side using endpoint. The endpoint of the executor uses broadcastId to find the corresponding relation. Actively penalizes the release method to free offheap memory. don't use `finalize()`.

The idea is similar to cleaning the broadcast block on the executor side.

Not sure if it will work, but if it does, this idea seems to be a final solution that fits the current Spark mechanics.

zhztheplayer: Thank you for the thoughtful insights.

More or less, the idea looks very to the solution CH backend is using. In case you happend to miss that one, could refer to code starting from [here](https://github.com/apache/incubator-gluten/blob/main/backends-clickhouse/src/main/scala/org/apache/spark/rpc/GlutenDriverEndpoint.scala).

Looks like feasible to me though there are 2 limitations coming out from my mind with the solution:

1. The GC timing of driver-side variable is still uncertain
2. What if the driver-side variable is GC-ed, while executor-side variable is still in use? (Perhaps, set driver-side variable to null manually, immediately after broadcasted to testify)

yikf: #2 does not seem to happen, at least until the end of the current stage, functions on the driver side will refer to broadcast variables, not sure if this is the case.

The gc timing on the driver side is also not controllable, but let's revisit the topic, we need to free up offheap memory when the broadcast variable is not in use, but there doesn't seem to be a certain thing on the executor side to recognize that the broadcast is no longer in use, so we need to trigger it on the executor side via gc, which is also somewhat strange in nature... Whether it's actively triggered gc or triggered by the jvm, and active gc is actually not a good solution in my opinion, it affects a lot of things.

zjuwangg:
Just make the `offHeapBroadcastBuildRelation` default behavior to false and adress the comments!
Thanks for your review @zhztheplayer @yikf

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
zjuwangg: > I was thinking about using storage memory but the implementation is complicated? If so let's put that idea on hold.
Using storage memory is a bit more complex and Spark broadcast mechanism basic implements is based on Spark's storage memory in nature. We can try this way later to see if things is ok.

zjuwangg:
ok，will open the issue later





------------------ Original ------------------
From: Hongze Zhang ***@***.***&gt;
Date: Wed,Jan 1,2025 2:13 AM
To: apache/incubator-gluten ***@***.***&gt;
Cc: Terry Wang ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [apache/incubator-gluten] [GLUTEN-7750][VL] MoveColumnarBuildSideRelation's memory occupation to Spark off-heap (PR #8127)





 
@zhztheplayer approved this pull request.
 
Looks like feasible to merge since the feature is marked with experimental flags.
 
Thank you for working on this @zjuwangg. Also would you like to open an issue ticket in advance to track on the OOM risk as we mentioned? Thanks.

Issue #8128, opened 2024-12-03
[VL] Gluten OOM with multi-slot executor configuration due to the Vanilla Spark memory acquisition strategy


kecookier:
### Backend

VL (Velox)

### Bug description

After spilled large memory, Task OOM, the detail log
```
24/12/03 09:46:23 INFO Executor task launch worker for task 1925 ColumnarShuffleWriter: Gluten shuffle writer: Trying to spill 2033398579 bytes of data
24/12/03 09:46:28 INFO Executor task launch worker for task 1925 ColumnarShuffleWriter: Gluten shuffle writer: Spilled 5991913856 / 2033398579 bytes of data
24/12/03 09:46:28 WARN Executor task launch worker for task 1925 TreeMemoryConsumer: TreeMemoryConsumer.spill end spilled:5991913856
24/12/03 09:46:29 ERROR Executor task launch worker for task 1925 ThrowOnOomMemoryTarget: Not enough spark off-heap execution memory. Acquired: 8.0 MiB, granted: 0.0 B. Try tweaking config option spark.memory.offHeap.size to get larger space to run this application (if spark.gluten.memory.dynamic.offHeap.sizing.enabled is not enabled). 
Current config settings: 
	spark.gluten.memory.offHeap.size.in.bytes=12.8 GiB
	spark.gluten.memory.task.offHeap.size.in.bytes=6.4 GiB
	spark.gluten.memory.conservative.task.offHeap.size.in.bytes=3.2 GiB
	spark.memory.offHeap.enabled=true
	spark.gluten.memory.dynamic.offHeap.sizing.enabled=false
Memory consumer stats: 
	Task.1925:                                             Current used bytes: 1144.0 MiB, peak bytes:        N/A
	\- Gluten.Tree.3:                                      Current used bytes: 1144.0 MiB, peak bytes:    8.8 GiB
	   \- root.3:                                          Current used bytes: 1144.0 MiB, peak bytes:    8.8 GiB
	      +- ShuffleWriter.3:                              Current used bytes:  744.0 MiB, peak bytes:    6.5 GiB
	      |  \- single:                                    Current used bytes:  744.0 MiB, peak bytes:    6.5 GiB
	      |     +- root:                                   Current used bytes:  719.1 MiB, peak bytes:  864.0 MiB
	      |     |  \- default_leaf:                        Current used bytes:  719.1 MiB, peak bytes:  859.2 MiB
[...]

PR #8132, opened 2024-12-03
[GLUTEN-8128][VL] Retry borrowing when granted size is less than requested in multi-slot and shared mode

kecookier:

Retry memory borrow when granted less than requested size in multi-slot and Gluten shared mode. If ThrowOnOomMemoryTarget cannot acquire memory, spill more memory and then retry.

(Fixes: \#8128)

If the case in issue #8128 occurs, the first borrow(sizeA) will spill sizeA memory. After spill, if curMemory is greater than the maxPerTaskSize, we'll get 0 returned. So we retry spill(long.max), then re-borrow.
- Add argument `isDynamicCapacity` when construct `TreeMemoryTargets#Node`, which means multi-task-slot and shared mode.

## How was this patch tested?
Test with my production ETL.

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargetVisitor.java
+++ b/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargetVisitor.java
@@ -35,4 +35,6 @@ public interface MemoryTargetVisitor<T> {
   T visit(NoopMemoryTarget noopMemoryTarget);
 
   T visit(DynamicOffHeapSizingMemoryTarget dynamicOffHeapSizingMemoryTarget);
+
+  T visit(RetryOnOomMemoryTarget retryOnOomMemoryTarget);
 }
--- a/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargets.java
+++ b/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargets.java
@@ -20,8 +20,10 @@ import org.apache.gluten.GlutenConfig;
 import org.apache.gluten.memory.MemoryUsageStatsBuilder;
 import org.apache.gluten.memory.memtarget.spark.TreeMemoryConsumers;
 
+import org.apache.spark.SparkEnv;
 import org.apache.spark.annotation.Experimental;
 import org.apache.spark.memory.TaskMemoryManager;
+import org.apache.spark.util.SparkResourceUtil;
 
 import java.util.Map;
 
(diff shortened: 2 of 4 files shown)


zhztheplayer:
Hi @kecookier thanks! Is this ready for review or still in draft?

kecookier:
> Hi @kecookier thanks! Is this ready for review or still in draft?

@zhztheplayer Thanks, still in draft, more tests are needed for this PR.

kecookier:
@zhztheplayer This PR is ready. Can you help review it?

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/TreeMemoryTargets.java
```diff
@@ -118,7 +125,29 @@ public long borrow(long size) {
...
+            granted);
+        long spilled = TreeMemoryTargets.spillTree(this, Long.MAX_VALUE);
+        long remain = requiredSize - granted;
+        if (spilled > remain) {
+          granted += borrow0(remain);
+        } else {
+          // OOM
+        }
+      }
+      return granted;
```
zhztheplayer, line 150: hi @kecookier 

Can we add a individual memory target for doing theretries?

Something like

```java
  public static class RetryOnOomMemoryTarget extends MemoryTarget {
    private final TreeMemoryTarget target;

    @Override
    public long borrow(long size) {
      final long granted = target.borrow(size);
      if (granted >= size) {
        return granted;
      }
      // Granted < size. Spill the underlying memory target then retry borrowing.
      final long remaining = size - granted;
      TreeMemoryTargets.spillTree(target, Long.MAX_VALUE);
      final long granted2 = target.borrow(remaining);
      return granted + granted2;
    }
```

I can help do a further benchmark too see if chaining another memory target hurts the overall performance. Thought we can make a try like this anyway.

kecookier: @zhztheplayer  I see, something like ThrowOnOomMemoryTarget( RetryOnOomMemoryTarget( Overacquired(...) ) )?

zhztheplayer: Something like that but I assume we tended to add it to [this line](https://github.com/apache/incubator-gluten/blob/05b1e7a28fa8b2915b573496466cbe94c7baab9b/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargets.java#L64)? Since shared mode is only interested.

zhztheplayer: We can add a comment like "Retry spilling is needed in shared mode because ... " or something

```java
    if (GlutenConfig.getConf().memoryIsolation()) {
      return TreeMemoryConsumers.isolated()
          .newConsumer(tmm, name, spiller, virtualChildren);
    } else {
      // Retry spilling is needed in shared mode because...
      return MemoryTargets.retrySpillOnOom(
          TreeMemoryConsumers.shared()
              .newConsumer(tmm, name, spiller, virtualChildren));
    }
```

kecookier: @zhztheplayer  Thanks, I'll update the code soon.

kecookier:
> Can we just spill maxPerTaskSize memory in the first try? Since task1 can't allocate any "new" memory any more. The retry has high possibility to be triggered.

@FelixYBW  Good question. In gluten shared mode, the first spill is triggered by [Vanilla Spark TaskMemoryManager](https://github.com/apache/spark/blob/branch-3.5/core/src/main/java/org/apache/spark/memory/TaskMemoryManager.java#L228), which only spills the requested memory. The spill size might be difficult to control easily within gluten.

zhztheplayer (approved):
The code itself looks fine to me % I may do some minor follow-ups with another PR. Thanks.

Feel free to continue your discussion in this thread or through a new issue if needed. @kecookier @FelixYBW

PR #8144, opened 2024-12-04
[CORE] Simplify code of offload scan

zml1206:

1. Offload scan does not need to doValidate again.
2. Remove catch of FallbackByNativeValidation.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
@@ -121,6 +121,8 @@ trait BackendSettingsApi {
 
   def supportCartesianProductExec(): Boolean = false
 
+  def supportCartesianProductExecWithCondition(): Boolean = true
+
   def supportBroadcastNestedLoopJoinExec(): Boolean = true
 
   def supportSampleExec(): Boolean = false
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
@@ -16,7 +16,6 @@
  */
 package org.apache.gluten.extension.columnar.enumerated
 
-import org.apache.gluten.exception.GlutenNotSupportException
 import org.apache.gluten.execution.{GlutenPlan, ValidatablePlan}
 import org.apache.gluten.extension.columnar.offload.OffloadSingleNode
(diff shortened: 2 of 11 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
```diff
@@ -118,21 +119,29 @@ object RasOffload {
...
+                    //
+                    // TODO: Tag the original plan with fallback reason. This is a non-trivial work
+                    //  in RAS as the query plan we got here may be a copy so may not propagate tags
+                    //  to original plan.
+                    from
+                  } else {
+                    offloadedPlan
+                  }
+                } catch {
+                  case _: GlutenNotSupportException =>
```
zml1206, line 140: Keep and disable RAS consistent.

zhztheplayer: Would you help check the corresponding logic in legacy planner to see whether we should only catch exceptions from `doValidate()` or catch more code like this way? Thanks!

zhztheplayer (commented):
% one comment. Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
zml1206: I didn't find the place where GlutenNotSupportException is thrown in the offload rule. `doValidate()` contains `doValidateInternal()` and `doSchemaValidate()`, `doValidateInternal()` has been caught, `doSchemaValidate()` will not trigger an exception, so restore it？

zhztheplayer: I see. So in this PR we can shrink the range of the catch block to only catch `doValidate()` to match legacy planner. Perhaps in future we should make all exceptions be caught in `doValidate` internally to make `doValidate` exception-free.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
```diff
@@ -121,6 +121,8 @@ trait BackendSettingsApi {
 
   def supportCartesianProductExec(): Boolean = false
 
+  def supportCartesianProductExecWithCondition(): Boolean = true
```
zhztheplayer, line 124: 👍

zhztheplayer (approved):
LGTM. Thank you very much for talking this code cleanup.

PR #8149, opened 2024-12-05
[VL] Add back RAII style Velox driver suspension into RowVectorStream

zhztheplayer:
SuspendedSection was removed from Velox code base from commit https://github.com/facebookincubator/velox/commit/fc5aa37fe44ca5bee9c91f9bee69069f0c804b11.

The PR introduce the code back to Gluten's code base to replace the temporary manual resource suspension code to avoid UBs when exception throws from `iterator_->hasNext()` or `iterator_->next()`.

### DIFF
--- a/cpp/velox/operators/plannodes/RowVectorStream.h
+++ b/cpp/velox/operators/plannodes/RowVectorStream.h
@@ -23,6 +23,27 @@
 #include "velox/exec/Operator.h"
 #include "velox/exec/Task.h"
 
+namespace {
+class SuspendedSection {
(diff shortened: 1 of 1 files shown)


PHILO-HE (approved):
Do we have any feedback from the Velox community that why they discard the use of SuspendedSection?

zhztheplayer:
> Do we have any feedback from the Velox community that why they discard the use of SuspendedSection?

I think Velox removed this only because no production code relies on it. I had sent a reminder to the Velox PR.

PR #8154, opened 2024-12-05
[GLUTEN-7143][VL] RAS: Remove experimental flags for RAS

zhztheplayer:
Part of #7143 (not to close it)

RAS is considered production-ready now as all Spark / Gluten UTs, TPC-H, TPC-DS ITs have passed for a period of time. More documentations needed but we can remove the experimental flags to state it's ready for use.

The patch removes the flags from doc and code.


zhouyuan (approved):
👍

PR #7842, opened 2024-11-07
[GLUTEN-7745][VL] Incorporate SQL Union operator into Velox execution pipeline

zhztheplayer:
#7745

Edit: The single-threaded execution issue will be fixed in Velox https://github.com/facebookincubator/velox/commit/e80bf12e3f5db0af81793986b0707477027c4e80.

Introduce new config `spark.gluten.sql.native.union` (by default, `false`) which works for Velox backend only at the moment.

Native union will not be used when children of union operator don't share the same known partition number. (perhaps can be fixed in future)

There is a Spark UT failing when enabling native union, it's related to Velox LocalParition output order and requires fixing in future:

```
2024-11-29T12:48:14.0003314Z - SPARK-14393: values generated by non-deterministic functions shouldn't change after coalesce or union *** FAILED ***
2024-11-29T12:48:14.0004432Z   Array([0,0], [2,1], [1,8589934592], [3,8589934593]) did not equal Array([0,0], [1,8589934592], [2,0], [3,8589934592]) Values changed after union when codegenFallback=true and wholeStage=false. (DataFrameFunctionsSuite.scala:3780)
```

So far the **failure is benign** since in SQL we don't often guarantee output ordering of union operator.

~~This is currently a PoC and not yet workable because of a blocker from Velox's single-threaded execution design:~~

<details>

```
I20241107 15:06:23.310680 1849235 VeloxPlanConverter.cc:128] Plan Node: 
-- Aggregation[7][PARTIAL n7_0 := count_partial("n0_0")] -> n7_0:BIGINT
  -- LocalPartition[6][GATHER] -> n0_0:BIGINT
    -- Project[4][expressions: (n0_0:BIGINT, "n1_1")] -> n0_0:BIGINT
      -- Project[1][expressions: (n1_1:BIGINT, "n0_0")] -> n1_1:BIGINT
        -- TableScan[0][table: hive_table] -> n0_0:BIGINT
    -- Project[5][expressions: (n0_0:BIGINT, "n3_1")] -> n0_0:BIGINT
      -- Project[3][expressions: (n3_1:BIGINT, "n2_0")] -> n3_1:BIGINT
        -- TableScan[2][table: hive_table] -> n2_0:BIGINT
24/11/07 15:06:23 ERROR TaskResources: Task 11 failed by error: 
org.apache.gluten.exception.GlutenException: Task doesn't support single thread execution: -- Aggregation[7]

[...]

PR #8188, opened 2024-12-09
[GLUTEN-8187][VL] Support velox cache metrics

yikf:
Velox backend supports cache functionality controlled by the config spark.gluten.sql.columnar.backend.velox.cacheEnabled, When the cache is enabled, we want to measure its effectiveness through a series of metrics.

This PR intends to add metrics for cache read operations.
## How was this patch tested?
unit tests


backends-velox/src/test/scala/org/apache/gluten/execution/VeloxCacheSuite.scala
```diff
@@ -0,0 +1,93 @@
...
+        } else {
+          assert(metrics2("storageReadBytes").value > 0)
+          assert(metrics2("ramReadBytes").value == 0)
+        }
+      } finally {
+        spark.sql(s"DROP TABLE IF EXISTS $table")
+      }
+    }
+
+    Seq("true", "false").foreach(
```
yikf, line 90: this UT will fail due to the current lack of a shutdown, which will be fixed in another PR.

yikf:
@PHILO-HE @zhztheplayer Could you please take a look, thanks!

cpp/core/jni/JniWrapper.cc
```diff
@@ -229,7 +229,10 @@ jint JNI_OnLoad(JavaVM* vm, void* reserved) {
   metricsBuilderClass = createGlobalClassReferenceOrError(env, "Lorg/apache/gluten/metrics/Metrics;");
 
   metricsBuilderConstructor = getMethodIdOrError(
-      env, metricsBuilderClass, "<init>", "([J[J[J[J[J[J[J[J[J[JJ[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J)V");
+      env,
+      metricsBuilderClass,
+      "<init>",
+      "([J[J[J[J[J[J[J[J[J[JJ[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J[J)V");
```
FelixYBW, line 235: We eventually need a better way to pass metrics.

FelixYBW (approved):

cpp/core/jni/JniWrapper.cc
zhztheplayer: Agreed. 

I have similar concern during working on a previous [patch](https://github.com/apache/incubator-gluten/pull/7842#discussion_r1865418672).

The code requires for a reworking.

yikf: thanks @FelixYBW , agree with you.

WangGuangxin: > We eventually need a better way to pass metrics.

@FelixYBW do you think "Foreign Function and Memory (FFM) API" from JDK 17 is a better way?

yikf: @WangGuangxin I investigated the FFM API[1][2] and it seems that there is no way to create and call Java objects from the native side. Can you provide some information on this?

[1] https://docs.oracle.com/en/java/javase/22/core/foreign-function-and-memory-api.html
[2] https://openjdk.org/jeps/454

yikf: FFM API/JNI's native side returns two-dimensional arrays, how about parsing mertrics on the java side?

FelixYBW: we may also use a protobuf or json string to pass the metrics.

PR #8198, opened 2024-12-10
[CORE] A new unified approach of source folder isolation for iceberg / hudi / delta

zhztheplayer:
This patch refactors Maven settings to add a unified method for managing source / resource / test-source folders' profile-based activations.

After the change, All the required folders in every Maven module will be toggled on if the relevant Maven profile is enabled. No further settings needed in each Maven module.

The new approach will give all the flexibilities to user since each data lake extension will have its own immediate source folder under each Maven module. Which means, there will be `src-iceberg` / `src-delta` / `src-hudi` on the same level with `src` under `backend-velox` and any other modules. Each of them is a full-featured maven source folder that consists of Java sources, Scala sources, resources, test sources, etc.


zhztheplayer:
cc @baibaichen thanks

backends-velox/pom.xml
```diff
@@ -318,6 +268,27 @@
...
+            <id>add-cpp-libraries</id>
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>add-resource</goal>
+            </goals>
+            <configuration>
+              <resources>
+                <resource>
+                  <directory>${cpp.releases.dir}</directory>
+                  <targetPath>${platform}/${arch}</targetPath>
```
zhztheplayer, line 285: Weird. It seems that Intellij IDEA doesn't pick up this line and put the libraries directly under the root resource dir. Investigating.

zhouyuan:
@zhztheplayer thanks, could you please create one issue to track these changes? 

Thanks,
-yuan

zhouyuan:
Hi @zzcclp @baibaichen  could you please prioritize review on this patch? since this patch introduced a big change on code footprint and may easily get conflict

Thanks,
-yuan

zhztheplayer:
For the `-P delta` issue we discussed, I managed to find a temp solution to make Maven build fail fast when `-P clickhouse` is set without `-P delta`, which is like

image

cc @zzcclp @baibaichen

lwz9103:

zhouyuan (approved):
👍

zhztheplayer:
I'll merge this after CI passed. Thank you guys for helping

@zhouyuan @baibaichen @zzcclp @lwz9103

lwz9103:

baibaichen (approved):
LGTM

zhztheplayer:
@lwz9103 @baibaichen Thank you for helping verifying this.

PR #8199, opened 2024-12-10
[GLUTEN-8202][ICEBERG] Fix get iceberg index error

lyy-pineapple:
When parsing Iceberg partition column fields, you should build the index first and then apply the filter. Otherwise, the field cannot correspond correctly when retrieving data using `partition.get`
for error

```
Error operating EXECUTE_STATEMENT: Wrong class, expected java.lang.Integer, but was java.lang.String, for object: xxx
java.lang.IllegalArgumentException: Wrong class, expected java.lang.Integer, but was java.lang.String, for object: xxxx
	at org.apache.iceberg.PartitionData.get(PartitionData.java:119)
	at org.apache.iceberg.spark.source.GlutenIcebergSourceUtil$.$anonfun$getPartitionColumns$3(GlutenIcebergSourceUtil.scala:180)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
```
[fix-8202](https://github.com/apache/incubator-gluten/issues/8202)

## How was this patch tested?


yma11 (approved):
Thanks for your fix.

PR #8203, opened 2024-12-11
[GLUTEN-8043] Use spark.shuffle.spill.diskWriteBufferSize in sort-based shuffle

marin-ma:


jinchengchenghh (approved):

jinchengchenghh:
Does hash shuffle writer or rss writer need to respect this config?

marin-ma:
@jinchengchenghh The implementation detail of hash shuffle writer is quite different from Vanilla spark's, and we do not have an entry to support this config. 

RSS writer doesn't spill to disks. It sends data to remote service when spill happens.

PR #8221, opened 2024-12-12
[GLUTEN-8222][VL] Support Factorial function

ArnavBalyan:
 - Supports offloading factorial to Velox.
 - Upstream PR in Velox https://github.com/facebookincubator/velox/pull/11844
 - Removes factorial from the blacklist in Gluten
 - https://github.com/apache/incubator-gluten/pull/9485 is merged which carries the upstream change

## How was this patch tested?
 - Unit tests in Velox


ArnavBalyan:
Being worked upon

rui-mo (commented):
Would you please add a test in the below suite to make sure this function could be offloaded?

https://github.com/apache/incubator-gluten/blob/main/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala

backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
```diff
@@ -2092,4 +2092,38 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
         }
       })
   }
+
+  test("Test factorial function correctness") {
```
PHILO-HE, line 2096: Better to move this new test to ScalarFunctionsValidateSuite.scala. Please also rename the test to "factorial function" for brevity. No need to redundantly add "Test" in the name.

PHILO-HE (commented):
Thanks for your pr! Just one trivial comment.

PHILO-HE (approved):
Thanks!

PR #8231, opened 2024-12-13
[MINOR] [VL] Enhance the gluten timer to support seconds, milliseconds, and microseconds

kecookier:
Sometimes we need a timer in seconds, e.g., for measuring file write time. Enhance the gluten timer to support seconds, milliseconds, microseconds and nanoseconds.

## How was this patch tested?
Exist CI.

### DIFF
--- a/cpp/core/shuffle/Payload.cc
+++ b/cpp/core/shuffle/Payload.cc
@@ -300,7 +300,7 @@ arrow::Result<std::vector<std::shared_ptr<arrow::Buffer>>> BlockPayload::deseria
     uint32_t& numRows,
     int64_t& deserializeTime,
     int64_t& decompressTime) {
-  Timer timer;
+  auto timer = std::make_unique<ScopedTimer>(&deserializeTime);
   static const std::vector<std::shared_ptr<arrow::Buffer>> kEmptyBuffers{};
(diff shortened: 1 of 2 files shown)


kecookier:
@marin-ma Can you help review this?

cpp/core/shuffle/Payload.cc
```diff
@@ -310,7 +310,7 @@ arrow::Result<std::vector<std::shared_ptr<arrow::Buffer>>> BlockPayload::deseria
   RETURN_NOT_OK(inputStream->Read(sizeof(uint32_t), &numRows));
   uint32_t numBuffers;
   RETURN_NOT_OK(inputStream->Read(sizeof(uint32_t), &numBuffers));
-  timer.reset();
+  deserializeTime += timer.realTimeUsed();
```
marin-ma, line 313: Why do we need this change? timer.start() and timer.stop() is not called. the realTimeUsed would be zero

kecookier: @marin-ma  Thanks, Um.. I've added this to fix the compile error with `std::unique_ptr<ScopedTimer>`. It was a typo. Now, I have fixed it in another way.

kecookier:
@marin-ma The comments have been fixed. Could you please review them again?

FelixYBW (approved):

Issue #8244, opened 2024-12-16
[CORE] Softaffinity use consistent hash schedule


yikf:
### Description

Softaffinity use consistent hash schedule

PR #8245, opened 2024-12-16
[GLUTEN-8244][CORE] Softaffinity use consistent hash schedule

yikf:
https://github.com/apache/incubator-gluten/issues/8244, This PR aims to make softaffinity use consistent hash schedule.

## How was this patch tested?
unit tests and GA.


yikf:
@zzcclp @jackylee-ch @zhztheplayer @PHILO-HE Could you please take a look, thanks!

yikf:
> Naive question: why we need consistent hash here? Is it for avoiding cache misses to the max extent when cluster gets changed?

Yes, especially when using the spark dynamic executor.

FelixYBW:
What does vanilla spark do here? Is it an enhancement of Spark scheduling?

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/clickhouse/utils/MergeTreePartsPartitionsUtil.scala
```diff
@@ -377,6 +377,9 @@ object MergeTreePartsPartitionsUtil extends Logging {
       partNameWithLocation: util.HashMap[String, String],
       locationDistinct: util.HashSet[String]): Unit = {
 
+    logWarning(
+      s"[DEBUG] PR-8245 " +
+        s"partNameWithLocation = $partNameWithLocation,  locationDistinct = $locationDistinct")
```
jackylee-ch, line 382: Looks like this is a debug log? Can we remove it?

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/clickhouse/utils/MergeTreePartsPartitionsUtil.scala
```diff
@@ -434,6 +437,7 @@ object MergeTreePartsPartitionsUtil extends Logging {
     }
 
     locationDistinct.forEach(closePartition)
+    logWarning(s"[DEBUG] PR-8245, partitions = $partitions")
```
jackylee-ch, line 440: ditto

jackylee-ch:
AFAIK, the consistent hashing cannot solve the cache miss problem for velox backend, but can bring a small improvement. For the ch backend, can consistent hashing significantly improve the performance and resove the cache miss? @zzcclp @loneylee

zhztheplayer:
> AFAIK, the consistent hashing cannot solve the cache miss problem for velox backend, but can bring a small improvement. 

Thank you for helping confirm. Also, is it possible that the [local cache](https://gluten.apache.org/docs/getting-started/localcache/) can be benefited?

> Yes, especially when using the spark dynamic executor.

Yes I assume this is another key point. `Dynamic Allocation ON + Local Cache ON`, could this be a typical target scenario of the change? @yikf are you using local cache so far?

jackylee-ch:
> Also, is it possible that the [local cache](https://gluten.apache.org/docs/getting-started/localcache/) can be benefited?

Yes. With consistent hashing, we can schedule the cache location in advance, which is better than knowing the cache location after execution. And consistent hashing is useful for cluster changes, especially when executors are rescheduled due to executor deaths.

yikf:
> > AFAIK, the consistent hashing cannot solve the cache miss problem for velox backend, but can bring a small improvement.
> 
> Thank you for helping confirm. Also, is it possible that the [local cache](https://gluten.apache.org/docs/getting-started/localcache/) can be benefited?
> 
> > Yes, especially when using the spark dynamic executor.
> 
> Yes I assume this is another key point. `Dynamic Allocation ON + Local Cache ON`, could this be a typical target scenario of the change? @yikf are you using local cache so far?

As @jackylee-ch said, pure soft affinity scheduling using consistent hashing would be better than the current logic in scenarios where executors change, which is the charm of consistent hashing. for local cache, I assume that there will be some benefits from the scheduling optimization as well.

We have used it in the TPC-DS, and we are also looking for scenarios in our online environment. Additionally, we would like to use soft affinity scheduling to optimize other scheduling within our organization.

backends-clickhouse/src/main/scala/org/apache/spark/sql/execution/datasources/clickhouse/utils/MergeTreePartsPartitionsUtil.scala
yikf: thanks.

FelixYBW:
> We have used it in the TPC-DS, and we are also looking for scenarios in our online environment. Additionally, we would like to use soft affinity scheduling to optimize other scheduling within our organization.

Is it an enhancement of vanilla Spark as well? or Gluten only?

jackylee-ch:
> Is it an enhancement of vanilla Spark as well? or Gluten only?

AFAIK, it is only used for Gluten, vanilla Spark doesn't need this.

yikf:
> > We have used it in the TPC-DS, and we are also looking for scenarios in our online environment. Additionally, we would like to use soft affinity scheduling to optimize other scheduling within our organization.
> 
> Is it an enhancement of vanilla Spark as well? or Gluten only?

It is currently only using gluten, but the soft affinity scheduling is relatively independent, and we can optimize it internally with the internal Spark version.

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
```diff
@@ -88,6 +90,13 @@ abstract class AffinityManager extends LogLevelUtil with Logging {
         }
       })
 
+  private var hashRing: Option[ConsistentHash[ExecutorNode]] = _
```
jackylee-ch, line 93: Why we need an Option here? Since we must init the hashRing, why not directly defined here?

gluten-core/src/main/java/org/apache/gluten/hash/ConsistentHash.java
```diff
@@ -0,0 +1,277 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.util.*;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * A consistent hash ring implementation. the class is thread-safe.
```
jackylee-ch, line 31: nit: maybe need alittle more details here. Such as the relation between node, virtual node, partition and slot.

yikf:
@jackylee-ch @zhztheplayer @FelixYBW Hi, Could you please take a look another?

jackylee-ch:
Basically look good to me, left a few comments.
BTW, have you test this pr with `duplicateReadingDetect`?

jackylee-ch:
Also cc @zzcclp @zhli1142015 , any more question about this pr?

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
yikf: Because the strategy is fundamentally scalable, and there may be other strategies in the future besides the consistentHash strategy, I think we should keep a scalable implementation.

yikf:
> Basically look good to me, left a few comments. BTW, have you test this pr with `duplicateReadingDetect`?

The UTs contains the `duplicateReadingDetect ` suite, and it passed the UTs validation.

zhli1142015:
LGTM.
BTW, I saw you also use SA with local cache, which cache do you use?

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
jackylee-ch: Okey. Then maybe move the `hashRing` to ConsistentHashSoftAffinityStrategy, which I think will be better expanded

yikf:
> LGTM. BTW, I saw you also use SA with local cache, which cache do you use?

We use Velox cache on TPC-DS, mainly with SSDs and RAM, and we are currently exploring scenarios for the online environment.

gluten-core/src/main/java/org/apache/gluten/hash/ConsistentHash.java
yikf: addressed

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
yikf: Thanks for you suggestion, only the manager is aware of the changes in the executor now, moving to the strategy requires that the strategy has the relevant interfaces for the executor's changes. the current soft affinity code is not very friendly in terms of extensibility, and we can refactor it in subsequent PRs to achieve better extensibility.

zzcclp:
> AFAIK, the consistent hashing cannot solve the cache miss problem for velox backend, but can bring a small improvement. For the ch backend, can consistent hashing significantly improve the performance and resove the cache miss? @zzcclp @loneylee

For the ch backend, the consistent hashing will reduce the cache missing, I don't know what about the performance.

jackylee-ch (approved):
LGTM.

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
```diff
@@ -88,6 +90,13 @@ abstract class AffinityManager extends LogLevelUtil with Logging {
         }
       })
 
+  private var hashRing: Option[ConsistentHash[ExecutorNode]] = None
```
zhztheplayer, line 93: Can we make it a `val`? Likely it's only set once.

zhztheplayer (approved):
👍

gluten-core/src/main/scala/org/apache/gluten/softaffinity/SoftAffinityManager.scala
yikf: thanks, addressed.

PR #8247, opened 2024-12-16
[CORE] Minor: OverTarget is required only with sufficient memory and doesn't spill due to zero used bytes post-borrow

kecookier:
Two minor updates:
1. Require overTarget only when sufficient target memory is granted.
2. overTarget doesn't need to really spill because overTarget.usedBytes() is always 0 after OverAcquire.borrow().

## How was this patch tested?
Exist CI.


zhztheplayer (approved):
👍 thanks

Issue #8257, opened 2024-12-17
[Core] Make IDEA support IssueNavigationLink


merrily01:
### Description

Make IDEA support IssueNavigationLink

Before:

![image](https://github.com/user-attachments/assets/3d8f4151-ba82-4100-a03e-93658b48b72a)

After:

![image](https://github.com/user-attachments/assets/261c79ad-8b76-4bdd-8388-84194653f000)

PR #8258, opened 2024-12-17
[GLUTEN-8257][CORE] Make IDEA support IssueNavigationLink

merrily01:

Make IDEA support IssueNavigationLink.

(Fixes: \#8257)

## How was this patch tested?

Manual test.

Before:

image

After:

image

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -20,7 +20,7 @@
     <option name="links">
       <list>
         <IssueNavigationLink>
-          <option name="issueRegexp" value="#(\d+)" />
+          <option name="issueRegexp" value="#(\d+)|GLUTEN-(\d+)" />
           <!--
             GitHub share the sequence number of issues and pull requests, and it will redirect to
             the right place when the the sequence number not match kind.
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@merrily01, could you clarify how this window is opened? It's just for checking commit history? Seems `Annotate` is more useful to see historical commits for some file. And the IDE already supports jumping to the commit link.

Yohahaha:
> @merrily01, could you clarify how this window is opened? It's just for checking commit history? Seems `Annotate` is more useful to see historical commits for some file. And the IDE already supports jumping to the commit link.

sadly, IDE does not support jumping to the commit link for now.

Yohahaha (approved):
👍 very helpful improvement

zhztheplayer:
@merrily01

The feature looks great. Though will this change create conflict between repo's and user's `vcx.xml` files?

Yohahaha:
maybe we should use a new doc to let user integrate to their IDEA.

zhztheplayer (approved):
Thank you for the contribution!

PR #8254, opened 2024-12-17
[CORE] Minor code cleanups for TreeMemoryConsumer

zhztheplayer:
Includes:

1. Code cleanups for TreeMemoryConsumer
   - Rephrase memory `shared mode` to `legacy mode` as `shared` is in a way misleading (what are shared, shared what, etc.). `legacy` means the behavior will be aligned with vanilla Spark. The new word is less confusing.
   - Untangle the weak caching code of memory consumer and the factories.
2. Minor follow-ups for https://github.com/apache/incubator-gluten/pull/8132

PR #8262, opened 2024-12-18
[VL] Fix `RetryOnOomMemoryTarget` only spills one single consumer on retrying

zhztheplayer:
Another follow-up for https://github.com/apache/incubator-gluten/pull/8254 and https://github.com/apache/incubator-gluten/pull/8132.

There is a issue introduced in #8254 unexpectedly, that `RetryOnOomMemoryTarget` only spills one single memory consumer when retrying happens. It should spill all the consumers in the same task instead. The patch fixes the issue with some other refactors in passing.

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargets.java
+++ b/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargets.java
@@ -64,9 +64,11 @@ public final class MemoryTargets {
       Map<String, MemoryUsageStatsBuilder> virtualChildren) {
     final TreeMemoryConsumers.Factory factory = TreeMemoryConsumers.factory(tmm);
     if (GlutenConfig.getConf().memoryIsolation()) {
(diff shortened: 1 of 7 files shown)


gluten-core/src/main/java/org/apache/gluten/memory/memtarget/MemoryTargets.java
```diff
@@ -86,7 +88,8 @@ public static TreeMemoryTarget newConsumer(
         consumer,
         () -> {
           LOGGER.info("Request for spilling on consumer {}...", consumer.name());
-          long spilled = TreeMemoryTargets.spillTree(consumer, Long.MAX_VALUE);
+          // Note: Spill from root node so other consumers also get spilled.
+          long spilled = TreeMemoryTargets.spillTree(root, Long.MAX_VALUE);
```
zhztheplayer, line 92: It's the core fix. @kecookier

PHILO-HE (approved):

Issue #8266, opened 2024-12-18
[VL] GHA is not stable when downloading Spark package


zhouyuan:
### Backend

VL (Velox)

### Bug description

```
bash .github/workflows/util/install_spark_resources.sh 3.5
  shell: sh -e {0}
  env:
    ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
    MVN_CMD: mvn -ntp
    WGET_CMD: wget -nv
    SETUP: bash .github/workflows/util/setup_helper.sh
    CCACHE_DIR: /home/runner/work/incubator-gluten/incubator-gluten/.ccache
failed: Connection timed out.
failed: Network is unreachable.
Error: Process completed with exit code [4](https://github.com/apache/incubator-gluten/actions/runs/12389480827/job/34583002711?pr=8265#step:6:4).
```

### Spark version

None

zhouyuan:
archive.apache.org is not stable
https://github.com/apache/incubator-gluten/blob/main/.github/workflows/util/install_spark_resources.sh#L27

zjuwangg:
I also just opened a discussion https://github.com/apache/incubator-gluten/discussions/8275. Glad to see the issue and solution!

zhouyuan:
Notes on several tests:
- retry on wget doesn't help
- the issue is due to we launch these jobs concurrently, so apache.org banned some IP
- Hosting the package to other place or move into docker can fix it

PHILO-HE:
I will try to pre-install these resources into docker image today. Assume the docker size limit will not be hit. cc @zjuwangg

zhouyuan:
fixed by #8294

PR #8267, opened 2024-12-18
[GLUTEN-8266][CI] retry more times on download spark release

zhouyuan:

This patch increase the wget retry times 

(Fixes: #8266 )

## How was this patch tested?

pass GHA

### DIFF
--- a/.github/workflows/util/install_spark_resources.sh
+++ b/.github/workflows/util/install_spark_resources.sh
@@ -19,17 +19,18 @@
 
 set -e
 
+WGET_CMD="wget --tries=100 --waitretry=30 --wait=20 --random-wait  -nv"
 INSTALL_DIR=$GITHUB_WORKSPACE
 case "$1" in
 3.2)
     # Spark-3.2
     cd ${INSTALL_DIR} && \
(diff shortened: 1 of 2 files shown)


zjuwangg:
I also noticed that https://archive.apache.org/ has a limit on download archive
`Do note that heavy use of this service will result in immediate throttling of your download speeds to either 12 or 6 mbps for the remainder of the day, depending on severity. Continuous abuse (to the tune of more than 40 GB downloaded per week) will cause an [automatic ban](https://infra.apache.org/infra-ban.html), so please tune your services to this fact.`


I don't know whether our GHA will be affected by this restriction. If so, we may need install spark in docker image or download from other dist mirror?

zhztheplayer (approved):
Found some mirrors may help as well:

https://github.com/fink/fink-mirrors/blob/master/apache (a list of apache mirrors)
https://ftp.unicamp.br/pub/apache/spark (spark)

Would also suggest doing checksum after downloading the tarball.

zhouyuan:
Note: 
- retry on wget doesn't help 
- the issue is due to we launch these jobs concurrently, so apache.org baned some IP 
- Hosting the package to other place or move into docker can fix it

Issue #8268, opened 2024-12-18
[VL] Preconditions fail in OverAcquire.repay() when overTarget.borrow() triggers a retrying spill


kecookier:
### Backend

VL (Velox)

### Bug description

Backtrace of Exception.
```
24/12/18 17:07:41 INFO Executor task launch worker for task 78584 ColumnarShuffleWriter: Gluten shuffle writer: Spilled 4250790592 / 8388608 bytes of data
24/12/18 17:07:41 INFO Executor task launch worker for task 78584 RetryOnOomMemoryTarget: Retrying spill require:1436968550 got:1431306240
24/12/18 17:07:41 INFO Executor task launch worker for task 78584 ColumnarShuffleWriter: Gluten shuffle writer: Trying to spill 9223372036854775807 bytes of data
24/12/18 17:07:41 ERROR Executor task launch worker for task 78584 ManagedReservationListener: Error unreserving memory from target
java.lang.IllegalStateException
	at org.apache.gluten.shaded.com.google.common.base.Preconditions.checkState(Preconditions.java:133)
	at org.apache.gluten.memory.memtarget.OverAcquire.repay(OverAcquire.java:77)
	at org.apache.gluten.memory.memtarget.ThrowOnOomMemoryTarget.repay(ThrowOnOomMemoryTarget.java:124)
	at org.apache.gluten.memory.listener.ManagedReservationListener.unreserve(ManagedReservationListener.java:63)
	at org.apache.gluten.vectorized.ShuffleWriterJniWrapper.nativeEvict(Native Method)
	at org.apache.spark.shuffle.ColumnarShuffleWriter$$anon$1.spill(ColumnarShuffleWriter.scala:170)
	at org.apache.gluten.memory.memtarget.Spillers$AppendableSpillerList.spill(Spillers.java:86)
	at org.apache.gluten.memory.memtarget.Spillers$WithMinSpillSize.spill(Spillers.java:66)
	at org.apache.gluten.memory.memtarget.TreeMemoryTargets.spillTree(TreeMemoryTargets.java:80)
	at org.apache.gluten.memory.memtarget.TreeMemoryTargets.spillTree(TreeMemoryTargets.java:55)
	at org.apache.gluten.memory.memtarget.TreeMemoryTargets.spillTree(TreeMemoryTargets.java:73)
	at org.apache.gluten.memory.memtarget.TreeMemoryTargets.spillTree(TreeMemoryTargets.java:55)
[...]

PR #8269, opened 2024-12-18
[GLUTEN-8268][CORE] Remove preconditions in OverAcquire.repay

kecookier:

After https://github.com/apache/incubator-gluten/pull/8132, overTarget.borrow(overSize) may trigger a retrying spill. If so, it calls OverAcquire.repay() during the spill procedure, which checks Preconditions.checkState(overTarget.usedBytes() == 0);. However, currently, 0 < overTarget.usedBytes() < overSize. We can remove this precondition in repay(), and only keep the precondition in borrow().

(Fixes: \#8268)

## How was this patch tested?

integration tests


zhztheplayer (approved):
👍

PR #8273, opened 2024-12-19
[VL] Simplify code for PartialProjectRule

zml1206:

as title.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/PartialProjectRule.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/PartialProjectRule.scala
@@ -27,12 +27,12 @@ case class PartialProjectRule(spark: SparkSession) extends Rule[SparkPlan] {
     plan.transformUp {
       case plan: ProjectExec =>
         val transformer = ColumnarPartialProjectExec.create(plan)
-        if (transformer.doValidate().ok()) {
-          if (transformer.child.asInstanceOf[ColumnarPartialProjectExec].doValidate().ok()) {
-            transformer
-          } else plan
+        if (
+          transformer.doValidate().ok() &&
+          transformer.child.asInstanceOf[ColumnarPartialProjectExec].doValidate().ok()
+        ) {
+          transformer
         } else plan
-      case p => p
     }
   }
 }


backends-velox/src/main/scala/org/apache/gluten/extension/PartialProjectRule.scala
```diff
@@ -27,12 +27,12 @@ case class PartialProjectRule(spark: SparkSession) extends Rule[SparkPlan] {
...
         val transformer = ColumnarPartialProjectExec.create(plan)
-        if (transformer.doValidate().ok()) {
-          if (transformer.child.asInstanceOf[ColumnarPartialProjectExec].doValidate().ok()) {
-            transformer
-          } else plan
+        if (
+          transformer.doValidate().ok() &&
+          transformer.child.asInstanceOf[ColumnarPartialProjectExec].doValidate().ok()
+        ) {
+          transformer
```
Yohahaha, line 34: we should log the reason of validation failed.

zml1206: debug level?

Yohahaha: I think should be warning, same as default value of `spark.gluten.sql.validation.logLevel`

zml1206: `CollapseProjectExecTransformer` used debug level, then unify it into warning level?

FelixYBW:
Have you tested the feature? In my last test it doesn't work well yet.

zml1206:
> Have you tested the feature? In my last test it doesn't work well yet.

I haven't tested this feature yet. I'll try to compile it locally and test now.

zml1206:
I tested it works, can you give examples of failures? @FelixYBW

FelixYBW:
> I tested it works, can you give examples of failures? @FelixYBW

no failure，perf is not good, consumes much onheap memory

@jinchengchenghh

jinchengchenghh:
Your test key is String before, it use Velox Column to row and row to Velo column, and I have optimized to Column to Arrow and Arrow to column, the performance for string should be improved.

And this case may go to regression, suppose udf(a) returns string,  hash(udf(a)) + hash(b), change to execute udf(a) in Spark, this will introduce extra arrow to column of udf(a) string result type as _partialProjectAlias column, if totally fallback, the row to column type is int.

backends-velox/src/main/scala/org/apache/gluten/extension/PartialProjectRule.scala
```diff
@@ -27,12 +27,20 @@ case class PartialProjectRule(spark: SparkSession) extends Rule[SparkPlan] {
     plan.transformUp {
       case plan: ProjectExec =>
         val transformer = ColumnarPartialProjectExec.create(plan)
-        if (transformer.doValidate().ok()) {
-          if (transformer.child.asInstanceOf[ColumnarPartialProjectExec].doValidate().ok()) {
+        val validationResult = transformer.doValidate()
+        if (validationResult.ok()) {
+          val childValidationResult =
+            transformer.child.asInstanceOf[ColumnarPartialProjectExec].doValidate()
```
jinchengchenghh, line 33: Maybe we should respect GlutenConfig `spark.gluten.sql.validation.logLevel`, add some code in https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala#L86, If the result is not OK, log as this level. I think some refactors may break log, we don't expect catch exception here.

zml1206: OK, then I will roll back. This PR will only simplify the code, and the log will be done later.

zhztheplayer (approved):
The PR itelf looks like a nit change. I think we can merge in advance. Feel free to continue your discussions, thanks.

PR #8276, opened 2024-12-19
[GLUTEN-8228][VL] Remove CSV option support

jinchengchenghh:

Remove CSV option because no customer uses it, we does not apply Arrow csv option to Spark csv option map patch, so it can be dropped. Change the arrow version to 15.0.0

(Fixes: \#ISSUE-ID)

## How was this patch tested?


jinchengchenghh:
For protobuf issue, can not drop

PR #8309, opened 2024-12-23
[GLUTEN-7641][VL] Fix security issue for perf analysis scripts

marin-ma:
And some minor fixes to the scripts


marin-ma:
@zhouyuan Could you please help to review? I can see the security issue are fixed in my repo https://github.com/marin-ma/gluten/security/dependabot

zhouyuan (approved):
👍

PR #8320, opened 2024-12-24
[GLUTEN-8340][VL] Enable from_json function

zhli1142015:

Fixes: #8340

## How was this patch tested?

UT.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -718,8 +718,9 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
         "'from_json' with 'spark.sql.caseSensitive = true' is not supported in Velox")
     }
 
-    val hasCaseInsensitiveDuplicateKey = expr.schema match {
+    val hasDuplicateKey = expr.schema match {
       case s: StructType =>
+        s.names.distinct.size != s.names.size ||
         !s.filter(
           f =>
             !s.names
@@ -730,9 +731,9 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
       case other =>
         false
     }
-    if (hasCaseInsensitiveDuplicateKey) {
+    if (hasDuplicateKey) {
       throw new GlutenNotSupportException(
-        "'from_json' with case-insensitive duplicate keys is not supported in Velox")
+        "'from_json' with duplicate keys is not supported in Velox")
     }
     val hasCorruptRecord = expr.schema match {
       case s: StructType =>
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -58,6 +58,8 @@ import org.apache.commons.lang3.ClassUtils
 
 import javax.ws.rs.core.UriBuilder
 
+import java.util.Locale
+
 class VeloxSparkPlanExecApi extends SparkPlanExecApi {
 
   /** Transform GetArrayItem to Substrait. */
@@ -714,6 +716,23 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
     if (SQLConf.get.caseSensitiveAnalysis) {
       throw new GlutenNotSupportException(
         "'from_json' with 'spark.sql.caseSensitive = true' is not supported in Velox")
+    }
+
+    val hasCaseInsensitiveDuplicateKey = expr.schema match {
+      case s: StructType =>
+        !s.filter(
+          f =>
+            !s.names
+              .filter(
+                n => n != f.name && n.toLowerCase(Locale.ROOT) == f.name.toLowerCase(Locale.ROOT))
+              .isEmpty)
+          .isEmpty
+      case other =>
+        false
+    }
+    if (hasCaseInsensitiveDuplicateKey) {
+      throw new GlutenNotSupportException(
+        "'from_json' with case-insensitive duplicate keys is not supported in Velox")
     }
     val hasCorruptRecord = expr.schema match {
       case s: StructType =>
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -711,6 +711,10 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
     if (!expr.options.isEmpty) {
       throw new GlutenNotSupportException("'from_json' with options is not supported in Velox")
     }
+    if (SQLConf.get.caseSensitiveAnalysis) {
+      throw new GlutenNotSupportException(
+        "'from_json' with 'spark.sql.caseSensitive = true' is not supported in Velox")
+    }
     val hasCorruptRecord = expr.schema match {
       case s: StructType =>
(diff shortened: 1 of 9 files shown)


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -690,6 +690,33 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
 
+  /** Generate an expression transformer to transform JsonToStructs to Substrait. */
+  override def genFromJsonTransformer(
+      substraitExprName: String,
+      children: Seq[ExpressionTransformer],
+      expr: JsonToStructs): ExpressionTransformer = {
+    if (!GlutenConfig.getConf.enableVeloxFromJsonFunction) {
+      throw new GlutenNotSupportException("'from_json' is not supported in Velox")
+    }
+    if (!SparkShimLoader.getSparkShims.fromJsonSupportPartialResults) {
```
ayushi-agarwal, line 701: So will it not be offloaded to velox for spark version < 3.4 ? Are there some cases where we can still offload if this is false for lower versions?

zhli1142015: > So will it not be offloaded to velox for spark version < 3.4 ?

Yes, as in Velox we only implement the code path of "enablePartialResults" now.

> Are there some cases where we can still offload if this is false for lower versions?

You can change `fromJsonSupportPartialResults` to true and run test suites "GlutenJsonExpressionsSuite" and "GlutenJsonFunctionsSuite".

zhli1142015:
cc @PHILO-HE , thanks.

shims/common/src/main/scala/org/apache/gluten/GlutenConfig.scala
```diff
@@ -2250,4 +2252,11 @@ object GlutenConfig {
       .doc("If enabled, gluten will convert the viewfs path to hdfs path in scala side")
       .booleanConf
       .createWithDefault(false)
+
+  val VELOX_FROM_JSON_FUNC_ENABLED =
+    buildConf("spark.gluten.sql.velox.fromJson.enabled")
+      .internal()
+      .doc("When true, offload from_json function to Velox.")
+      .booleanConf
+      .createWithDefault(true)
```
PHILO-HE, line 2261: The existing `spark.gluten.expression.blacklist` can be used to disable some functions. Seems no need to introduce a dedicated one.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxJsonFunctionsSuite.scala
PHILO-HE, line 1: Suggest to name it like `JsonFunctionsValidateSuite.scala` for consistency. The purpose is for validating the offloading behavior by using some simple cases.

PHILO-HE (commented):
Looks good! Some minor comments.

zhouyuan:
@zhli1142015 thanks a lot for the patch, would you please help to create one issue to track this? I assume there will be some follow up patches required

zhli1142015:
> @zhli1142015 thanks a lot for the patch, would you please help to create one issue to track this? I assume there will be some follow up patches required

Yes, https://github.com/apache/incubator-gluten/issues/8340. This patch targets phase 1 only.

zhouyuan:
pending: https://github.com/facebookincubator/velox/pull/11709

ayushi-agarwal:
val jsonData = Seq(
      """{"platformId": "IPHONE", "userId": "123", "sessionId": "abc"}""",
      """{"platformId": "ANDROID", "userId": "456", "sessionId": "def"}""",
      """{"platformId": "IPHONE", "userId": "789", "sessionId": "ghi"}"""
    )
    val df = spark.createDataFrame(jsonData.map(Tuple1(_))).toDF("json_column")
    df.printSchema()

    df.write.mode("overwrite").parquet("output/json_parquet_data")
    val parquetDF = spark.read.parquet("output/json_parquet_data")

    val r2 = parquetDF.collect()
    r2.foreach(println)

    val schema = new StructType().add("platformId", StringType).add("userId", StringType).add("sessionId", StringType)
    val filteredDF = parquetDF.withColumn("parsed_json", from_json(col("json_column"), schema)).select("parsed_json")
    val result = filteredDF.collect()
This prints null result when offloaded.
@zhli1142015 Are structs not supported? Shall we add a check to not offload for struct types?

ayushi-agarwal:
> val jsonData = Seq( """{"platformId": "IPHONE", "userId": "123", "sessionId": "abc"}""", """{"platformId": "ANDROID", "userId": "456", "sessionId": "def"}""", """{"platformId": "IPHONE", "userId": "789", "sessionId": "ghi"}""" ) val df = spark.createDataFrame(jsonData.map(Tuple1(_))).toDF("json_column") df.printSchema()
> 
> ```
> df.write.mode("overwrite").parquet("output/json_parquet_data")
> val parquetDF = spark.read.parquet("output/json_parquet_data")
> 
> val r2 = parquetDF.collect()
> r2.foreach(println)
> 
> val schema = new StructType().add("platformId", StringType).add("userId", StringType).add("sessionId", StringType)
> val filteredDF = parquetDF.withColumn("parsed_json", from_json(col("json_column"), schema)).select("parsed_json")
> val result = filteredDF.collect()
> ```
> 
> This prints null result when offloaded. @zhli1142015 Are structs not supported? Shall we add a check to not offload for struct types?

@zhli1142015 Will this be fixed by the last change you made, shall we add this as a test which check for results matching with and without offload?

zhli1142015:
Yes, this issue has been resolved. The null value occurred because the schema's case did not match that of the input field name. I did the test mostly on 1.2 which doesn't have such problem. I'm not sure which commit caused this difference.

zhli1142015:
cc @rui-mo and @PHILO-HE , this PR is ready, thanks.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
PHILO-HE: @zhli1142015, can we simply depend on the below value returned?

`conf.getBoolean("spark.sql.json.enablePartialResults", false)`

Assume if this key is not defined in lower Spark version, the default value (2nd arg) will be used.

Moreover, I note false is the default value in your added API, does this mean partial result is NOT allowed in lower Spark version (< 3.4)? Better to add one comment to clarify. Thank you!

zhli1142015: Updated.
Before version 3.4, this configuration did not exist, and partial results were NOT allowed. After version 3.4, partial results are supported by default.

PHILO-HE (approved):
Looks good! Thank you!

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -700,6 +702,64 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
   }
 
+  /** Generate an expression transformer to transform JsonToStructs to Substrait. */
+  override def genFromJsonTransformer(
+      substraitExprName: String,
+      children: Seq[ExpressionTransformer],
+      expr: JsonToStructs): ExpressionTransformer = {
+    val enablePartialResults =
+      try {
+        SQLConf.get.getConfString(s"spark.sql.json.enablePartialResults").toBoolean
```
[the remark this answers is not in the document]
zhli1142015: I tried this also, it would always use the provided default value "false" not the default value from Config Entry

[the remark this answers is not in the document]
PHILO-HE: I see. It seems under that usage the given default will always be returned if the config key is not explicitly set by user. Let's keep using your code.

Issue #8327, opened 2024-12-24
[CORE] Introduce the ConfigEntry to make the config definition more flexible


yikf:
### Description

parent: #8326.

Issue #8326, opened 2024-12-24
[CORE][umbrella] Configuration enhancements


yikf:
### Description

As discussed in https://github.com/apache/incubator-gluten/issues/6970, we plan to enhance the gluten configuration, and the specific enhancement solutions can be found in section https://github.com/apache/incubator-gluten/issues/6970#issuecomment-2511452471.

This issue serves as an umbrella to record the task list that we need to include to achieve our goals, because enhance config involves a large number of files, splitting multiple tasks facilitates code review and reduces the risk of introducing other problems.

- [ ] Introduce the ConfigEntry to make the config definition more flexible.
- [ ] Split the configuration into individual modules that belong to each other.
- [ ] Split configuration into two separate files: one static and one non-static.
- [ ] Standardize configuration naming, log the deprecated config name if used.
- [ ] Support generate gluten configuration doc and check it. https://github.com/apache/incubator-gluten/issues/8070

baibaichen:
@yikf as discussed in the offline, there are 2  requirement

1. Pass spark config to backend. For example, ch backend need `spark.sql.caseSensitive` to choose compare behavior.
2. Pass value type to backend. currently we use `ConfigMap` which use assum value is string type.

FelixYBW:
@yikf Is the enhancement finished? Is there a readme about how to add a configuration?

yikf:
> [@yikf](https://github.com/yikf) Is the enhancement finished? Is there a readme about how to add a configuration?

It's not completed yet. I will keep following up in the next few days. 

Eventually, there will be a configuration.md file to describe the configuration. Regarding the description of how to add the configuration, I can add it in this file.

PR #8328, opened 2024-12-24
[GLUTEN-8327][CORE] Introduce the ConfigEntry to make the config definition more flexible

yikf:
#8327 . subtask of #8326 

## How was this patch tested?
unit tests and GA

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
@@ -1,0 +1,229 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gluten.config
+
+import org.apache.spark.network.util.ByteUnit
+import org.apache.spark.network.util.JavaUtils
+
+import java.util.concurrent.TimeUnit
+import java.util.regex.Pattern;
+
+object BackendType extends Enumeration {
+  type BackendType = Value
+  val COMMON, VELOX, CLICKHOUSE = Value
+}
+
+private[gluten] case class ConfigBuilder(key: String) {
+  import ConfigHelpers._
+
+  private[config] var _doc = ""
+  private[config] var _version = ""
+  private[config] var _backend = BackendType.COMMON
+  private[config] var _public = true
+  private[config] var _alternatives = List.empty[String]
+  private[config] var _onCreate: Option[ConfigEntry[_] => Unit] = None
+
+  def doc(s: String): ConfigBuilder = {
+    _doc = s
+    this
+  }
+
+  def version(s: String): ConfigBuilder = {
+    _version = s
+    this
+  }
+
+  def backend(backend: BackendType.BackendType): ConfigBuilder = {
+    _backend = backend
+    this
+  }
+
+  def internal(): ConfigBuilder = {
+    _public = false
+    this
+  }
+
+  def onCreate(callback: ConfigEntry[_] => Unit): ConfigBuilder = {
+    _onCreate = Option(callback)
+    this
+  }
+
+  def withAlternative(key: String): ConfigBuilder = {
+    _alternatives = _alternatives :+ key
+    this
+  }
+
+  def intConf: TypedConfigBuilder[Int] = {
+    new TypedConfigBuilder(this, toNumber(_, _.toInt, key, "int"))
+  }
+
+  def longConf: TypedConfigBuilder[Long] = {
+    new TypedConfigBuilder(this, toNumber(_, _.toLong, key, "long"))
+  }
+
+  def doubleConf: TypedConfigBuilder[Double] = {
+    new TypedConfigBuilder(this, toNumber(_, _.toDouble, key, "double"))
+  }
+
+  def booleanConf: TypedConfigBuilder[Boolean] = {
+    new TypedConfigBuilder(this, toBoolean(_, key))
+  }
+
+  def stringConf: TypedConfigBuilder[String] = {
+    new TypedConfigBuilder(this, identity)
+  }
+
+  def timeConf(unit: TimeUnit): TypedConfigBuilder[Long] = {
+    new TypedConfigBuilder(this, timeFromString(_, unit), timeToString(_, unit))
+  }
+
+  def bytesConf(unit: ByteUnit): TypedConfigBuilder[Long] = {
+    new TypedConfigBuilder(this, byteFromString(_, unit), byteToString(_, unit))
+  }
+
+  def fallbackConf[T](fallback: ConfigEntry[T]): ConfigEntry[T] = {
+    val entry =
+      new ConfigEntryFallback[T](key, _doc, _version, _backend, _public, _alternatives, fallback)
+    _onCreate.foreach(_(entry))
+    entry
+  }
+}
+
+private object ConfigHelpers {
+  def toNumber[T](s: String, converter: String => T, key: String, configType: String): T = {
+    try {
+      converter(s.trim)
+    } catch {
+      case _: NumberFormatException =>
+        throw new IllegalArgumentException(s"$key should be $configType, but was $s")
+    }
+  }
+
+  def toBoolean(s: String, key: String): Boolean = {
+    try {
+      s.trim.toBoolean
+    } catch {
+      case _: IllegalArgumentException =>
+        throw new IllegalArgumentException(s"$key should be boolean, but was $s")
+    }
+  }
+
+  private val TIME_STRING_PATTERN = Pattern.compile("(-?[0-9]+)([a-z]+)?")
+
+  def timeFromString(str: String, unit: TimeUnit): Long = JavaUtils.timeStringAs(str, unit)
+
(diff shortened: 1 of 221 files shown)


yikf:
@zhztheplayer @jackylee-ch Could you please take a look, thanks!

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
```diff
@@ -0,0 +1,262 @@
...
+    } catch {
+      case _: IllegalArgumentException =>
+        throw new IllegalArgumentException(s"$key should be boolean, but was $s")
+    }
+  }
+
+  private val TIME_STRING_PATTERN = Pattern.compile("(-?[0-9]+)([a-z]+)?")
+
+  def timeFromString(str: String, unit: TimeUnit): Long = {
+    val lower = str.toLowerCase(Locale.ROOT).trim
```
jackylee-ch, line 129: Can we just use `JavaUtils.timeStringAs`?

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
```diff
@@ -0,0 +1,262 @@
...
+
+  def timeConf(unit: TimeUnit): TypedConfigBuilder[Long] = {
+    new TypedConfigBuilder(this, timeFromString(_, unit), timeToString(_, unit))
+  }
+
+  def bytesConf(unit: ByteUnit): TypedConfigBuilder[Long] = {
+    new TypedConfigBuilder(this, byteFromString(_, unit), byteToString(_, unit))
+  }
+
+  def fallbackConf[T](fallback: ConfigEntry[T]): ConfigEntry[T] = {
```
jackylee-ch, line 99: any different between `withAlternatives` and `fallbackConf`? maybe only use one of them?

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -37,97 +38,90 @@ case class GlutenNumaBindingInfo(
...
+  def veloxColumnarWindowType: String = getConf(COLUMNAR_VELOX_WINDOW_TYPE)
 
-  def enableColumnarShuffledHashJoin: Boolean = conf.getConf(COLUMNAR_SHUFFLED_HASH_JOIN_ENABLED)
+  def enableColumnarShuffledHashJoin: Boolean = getConf(COLUMNAR_SHUFFLED_HASH_JOIN_ENABLED)
 
   def shuffledHashJoinOptimizeBuildSide: Boolean =
-    conf.getConf(COLUMNAR_SHUFFLED_HASH_JOIN_OPTIMIZE_BUILD_SIDE)
+    getConf(COLUMNAR_SHUFFLED_HASH_JOIN_OPTIMIZE_BUILD_SIDE)
 
-  def enableNativeColumnarToRow: Boolean = conf.getConf(COLUMNAR_COLUMNAR_TO_ROW_ENABLED)
```
jackylee-ch, line 81: ok, we doesn't use this config, we should remove all of them in sperated pr.

yikf: yes.

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
yikf: The `fallbackConf` is a complete ConfigEntry, including doc, version information, etc. `withAlternatives` is just a key.

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
yikf: sure.

shims/common/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
```diff
@@ -0,0 +1,203 @@
...
+    _valueConverter: String => T,
+    _stringConverter: T => String)
+  extends ConfigEntry[Option[T]] {
+  override def key: String = _key
+
+  override def doc: String = _doc
+
+  override def version: String = _version
+
+  override def backend: BackendType = _backend
```
zhztheplayer, line 62: It seems that this API is not yet called. This is designed for future use, right?

shims/common/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
```diff
@@ -0,0 +1,203 @@
...
+  override def readFrom(conf: ConfigProvider): T = {
+    readString(conf).map(valueConverter).getOrElse(_defaultVal)
+  }
+
+  override def defaultValue: Option[T] = Option(_defaultVal)
+
+  override def defaultValueString: String = stringConverter(_defaultVal)
+}
+
+private class ConfigEntryWithDefaultString[T](
```
zhztheplayer, line 115: This one is `private` (actually equivalent to `private[config]`) while others are `private[gluten]`. Is there any particular reason?

zhztheplayer (commented):
Thanks for taking on the configuration enhancements. This series of work could be a big millstone for Gluten's overall code quality.

Also, would you like to add some comments on the newly added config-related classes (if needed), mentioning that the code is similar to Spark's relevant code but extended for Gluten's use? So users will know the difference here more quickly.

shims/common/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
yikf: no, will keep consistent.

shims/common/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
yikf: yes.

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
jackylee-ch: `withAlternative` should be enough if we have some config name changed.

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
```diff
@@ -0,0 +1,229 @@
...
+
+private[gluten] case class ConfigBuilder(key: String) {
+  import ConfigHelpers._
+
+  private[config] var _doc = ""
+  private[config] var _version = ""
+  private[config] var _backend = BackendType.COMMON
+  private[config] var _public = true
+  private[config] var _alternatives = List.empty[String]
+  private[config] var _onCreate: Option[ConfigEntry[_] => Unit] = None
```
jackylee-ch, line 38: Maybe we can add `deprecated` here, so we can deprecated some configs and warning user while user setting it.

yikf: It should not be the case, maybe. `onCreate` occurs when the entry is created and it always happens. However, deprecation should take place when the user actively sets a deprecated key. Checking for deprecated keys from `sparkConf` during plugin initialization is more appropriate and will only issue a warning once.

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
yikf: The meanings expressed are still different. "Alternative" indicates a choice, while for "rename", what we aim to do is to eventually deprecated it, and it should contain more warning information.

zhztheplayer:
@yikf Would you rebase? Thanks!
@jackylee-ch If you have further review comments? Thanks!

shims/common/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
jackylee-ch: Oh, vanilla Spark already has some cases for this kind of definition.
```
  val ARROW_PYSPARK_FALLBACK_ENABLED =
    buildConf("spark.sql.execution.arrow.pyspark.fallback.enabled")
      .doc(s"When true, optimizations enabled by '${ARROW_PYSPARK_EXECUTION_ENABLED.key}' will " +
        "fallback automatically to non-optimized implementations if an error occurs.")
      .version("3.0.0")
      .fallbackConf(ARROW_FALLBACK_ENABLED)

  val ARROW_FALLBACK_ENABLED =
    buildConf("spark.sql.execution.arrow.fallback.enabled")
      .doc("(Deprecated since Spark 3.0, please set " +
        "'spark.sql.execution.arrow.pyspark.fallback.enabled'.)")
      .version("2.4.0")
      .booleanConf
      .createWithDefault(true)
```

zhztheplayer: @jackylee-ch 

Thank you for sharing ideas!

I see Spark's `SQLConf.scala` uses both `fallbackConf` (as mentioned) and `withAlternative` for this kind of key routiing functionality. So would you suggest adding another `deprecated` API to emphasis on the deprecation of the configuration options?

jackylee-ch: > So would you suggest adding another `deprecated` API to emphasis on the deprecation of the configuration options?

Yeap.

yikf: SGTM, will follow this approach at the subsequent PR to change the configuration name.

zhztheplayer: See you guys' points. Temporarily, I think it's reasonable to keep the first PR aligned with vanilla Spark's manner. Let's see if we could use our own way to handle the deprecated options in subsequent efforts.

jackylee-ch (approved):
LGTM

yikf:
@jackylee-ch @zhztheplayer Kindly ping, Could you take a look another? and hele re-run the failed pipeline : )

yikf:
@jackylee-ch @zhztheplayer Thanks all guys!

baibaichen:
@yikf would we directly use `org.apache.spark.ConfigEntry` without define a new `ConfigEntry` trait

yikf:
> @yikf would we directly use `org.apache.spark.ConfigEntry` without defining a new `ConfigEntry` trait

Hi @baibaichen , Thank you for your well-considered insights.  We hope to use a custom `ConfigEntry` to achieve more flexible configuration. You can refer to the related discussion at https://github.com/apache/incubator-gluten/issues/6970#issuecomment-2511452471.

Issue #8330, opened 2024-12-24
[VL] Improve the performance of converting from viewfs to hdfs


wangyum:
### Description

1. Avoid RPC calls.
2. Add a cache.

PR #8331, opened 2024-12-24
[GLUTEN-8330][VL] Improve convert the viewfs path to hdfs path

wangyum:

1. Avoid RPC calls.
2. Add a cache.

(Fixes: \#8330)

## How was this patch tested?

Manual tests, before: 225007 ms, after: 1330 ms:
```scala
import org.apache.hadoop.fs.viewfs.ViewFileSystemUtils
import org.apache.hadoop.conf.Configuration
import org.apache.hadoop.fs.{FileSystem, Path}

val fs = FileSystem.get(spark.sparkContext.hadoopConfiguration)
val paths = fs.listStatus(new Path("viewfs://my-cluster/path/to/table")).take(50000).map(_.getPath.toString)

val start1 = System.currentTimeMillis
paths.map(p => FileSystem.get(new Path(p).toUri, spark.sparkContext.hadoopConfiguration).resolvePath(new Path(p)))
println(s"Convert viewfs to HDFS using resolvePath took: ${System.currentTimeMillis - start1} ms")

val start2 = System.currentTimeMillis
paths.map(p => ViewFileSystemUtils.convertViewfsToHdfs(p, spark.sparkContext.hadoopConfiguration))
println(s"Convert viewfs to HDFS using convertViewfsToHdfs took: ${System.currentTimeMillis - start2} ms")
```

Output:
```
Convert viewfs to HDFS using resolvePath took: 225007 ms
Convert viewfs to HDFS using convertViewfsToHdfs took: 1330 ms
```

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -49,8 +49,10 @@ import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
 import org.apache.spark.util.SerializableConfiguration
 
-import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.hadoop.fs.Path
+import org.apache.hadoop.fs.viewfs.ViewFileSystemUtils
 
+import scala.collection.mutable
 import scala.util.control.Breaks.breakable
 
 class VeloxBackend extends SubstraitBackend {
@@ -107,16 +109,24 @@ object VeloxBackendSettings extends BackendSettingsApi {
       val filteredRootPaths = distinctRootPaths(rootPaths)
       if (filteredRootPaths.nonEmpty) {
         val resolvedPaths =
-          if (
-            GlutenConfig.getConf.enableHdfsViewfs && filteredRootPaths.head.startsWith("viewfs")
-          ) {
+          if (GlutenConfig.getConf.enableHdfsViewfs) {
+            val viewfsToHdfsCache: mutable.Map[String, String] = mutable.Map.empty
             // Convert the viewfs path to hdfs path.
             filteredRootPaths.map {
-              viewfsPath =>
-                val viewPath = new Path(viewfsPath)
-                val viewFileSystem =
-                  FileSystem.get(viewPath.toUri, serializableHadoopConf.get.value)
-                viewFileSystem.resolvePath(viewPath).toString
+              path =>
+                if (path.startsWith("viewfs")) {
+                  val pathSplit = path.split(Path.SEPARATOR)
+                  val prefixIndex = pathSplit.size - 1
+                  val pathPrefix = pathSplit.take(prefixIndex).mkString(Path.SEPARATOR)
+                  val hdfsPathPrefix = viewfsToHdfsCache.getOrElseUpdate(
+                    pathPrefix,
+                    ViewFileSystemUtils
+                      .convertViewfsToHdfs(pathPrefix, serializableHadoopConf.get.value))
+                  hdfsPathPrefix + Path.SEPARATOR +
+                    pathSplit.drop(prefixIndex).mkString(Path.SEPARATOR)
+                } else {
+                  path
+                }
             }
           } else {
             filteredRootPaths
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -111,23 +111,10 @@ object VeloxBackendSettings extends BackendSettingsApi {
         val resolvedPaths =
           if (GlutenConfig.getConf.enableHdfsViewfs) {
             val viewfsToHdfsCache: mutable.Map[String, String] = mutable.Map.empty
-            // Convert the viewfs path to hdfs path.
-            filteredRootPaths.map {
-              path =>
-                if (path.startsWith("viewfs")) {
-                  val pathSplit = path.split(Path.SEPARATOR)
-                  val prefixIndex = pathSplit.size - 1
-                  val pathPrefix = pathSplit.take(prefixIndex).mkString(Path.SEPARATOR)
-                  val hdfsPathPrefix = viewfsToHdfsCache.getOrElseUpdate(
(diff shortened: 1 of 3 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -377,26 +378,28 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
       val allScanSplitInfos =
         getSplitInfosFromPartitions(basicScanExecTransformers, allScanPartitions)
       if (GlutenConfig.getConf.enableHdfsViewfs) {
+        val viewfsToHdfsCache: mutable.Map[String, String] = mutable.Map.empty
```
zhouyuan, line 381: Would this cache take too much memory?

wangyum: No, the size of this map is at most equal to the total number of partitions of these tables.

gluten-substrait/src/main/scala/org/apache/hadoop/fs/viewfs/ViewFileSystemUtils.scala
```diff
@@ -0,0 +1,45 @@
...
+      case vfs: ViewFileSystem =>
+        val fsStateField = vfs.getClass.getDeclaredField("fsState")
+        fsStateField.setAccessible(true)
+        val fsState = fsStateField.get(vfs).asInstanceOf[InodeTree[FileSystem]]
+        val res = fsState.resolve(f, true)
+        if (res.isInternalDir) {
+          f
+        } else {
+          Path.mergePaths(new Path(res.targetFileSystem.getUri), res.remainingPath).toString
+        }
```
wangyum, line 40: If we don't use Java reflection, we need to add this method in ViewFileSystem:
```java
  public String convertViewfsToHdfs(String f) throws IOException {
    final InodeTree.ResolveResult<FileSystem> res;
    res = fsState.resolve(f, true);
    if (res.isInternalDir()) {
      return f;
    } else {
      return Path.mergePaths(new Path(res.targetFileSystem.getUri()), res.remainingPath).toString();
    }
  }
```

zhouyuan: CC @JkSelf for her comments

zhouyuan (approved):
👍

gluten-substrait/src/main/scala/org/apache/hadoop/fs/viewfs/ViewFileSystemUtils.scala
JkSelf: @wangyum It make sense to use Java reflection instead of modifying the code in Hadoop ViewFileSystem.

> return Path.mergePaths(new Path(res.targetFileSystem.getUri()), res.remainingPath).toString();\

Can we directly return the `new Path(res.targetFileSystem.getUri())`? It seems the code here https://github.com/apache/incubator-gluten/pull/8331/files#diff-68925b6a5ee67a942a976971ae6727a45a0b166d200d8d983454169e7c3bd1f6R124 only retrieves the hdfs prefix not the full hdfs path? Correct me if wrong understand. Thanks.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -107,16 +109,24 @@ object VeloxBackendSettings extends BackendSettingsApi {
...
             // Convert the viewfs path to hdfs path.
             filteredRootPaths.map {
-              viewfsPath =>
-                val viewPath = new Path(viewfsPath)
-                val viewFileSystem =
-                  FileSystem.get(viewPath.toUri, serializableHadoopConf.get.value)
-                viewFileSystem.resolvePath(viewPath).toString
+              path =>
+                if (path.startsWith("viewfs")) {
+                  val pathSplit = path.split(Path.SEPARATOR)
```
JkSelf, line 118: Can we move this code into a new method in the ViewFileSystemUtils object? This would help in keeping the code reusable.

JkSelf (commented):
LGTM. Thanks for your work. Except two small comments.

gluten-substrait/src/main/scala/org/apache/hadoop/fs/viewfs/ViewFileSystemUtils.scala
wangyum: The origin path is:
```
viewfs://my-cluser/path/to/table/dt=20241209/hour=03/ns=ns7/part-01996-12a9e188-d81f-4b6b-9054-cef61acb8b44.c000.snappy.parquet
```
The `f` is:
```
viewfs://my-cluser/path/to/table/dt=20241209/hour=03/ns=ns7
```
the `res.targetFileSystem.getUri()` is:
```
hdfs://my-cluser-router
```
So we need to append remainingPath:
```
hdfs://my-cluser-router/path/to/table/dt=20241209/hour=03/ns=ns7
```

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
wangyum: done

gluten-substrait/src/main/scala/org/apache/hadoop/fs/viewfs/ViewFileSystemUtils.scala
JkSelf: @wangyum I see. Thanks for your explanation.

JkSelf (approved):
LGTM. Thanks.

JkSelf:
@wangyum Can you help to resolve the following compile issue? Thanks.

```
Error:  /__w/incubator-gluten/incubator-gluten/gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala:386: error: type mismatch;
Error:   found   : scala.collection.mutable.Buffer[String]
Error:   required: Seq[String]
Error:                    splitInfo.getPaths.asScala,
Error:                                       ^
Warning: [WARNING] 1 warning
Error: [ERROR] 1 error
Error:  exception compilation error occurred!!!
```

Issue #8352, opened 2024-12-26
[VL] read column error when iceberg v1 table partition was drop


lyy-pineapple:
### Backend

VL (Velox)

### Bug description

When the p2 partition of Iceberg v1 table  is deleted, the column is not physically removed in the code but marked as a `VoidTransform`. As a result, the file path changes to xxx/p2=null/xxx. During data reading, when encountering a `VoidTransform`, the data should be retrieved directly from the file instead of relying on the partition value. However, in this scenario, Gluten reads the p2 values as null for all records.

### Spark version

None

PR #8353, opened 2024-12-26
[GLUTEN-8352][ICEBERG] Fix read error when partition column was drop

lyy-pineapple:

When the p2 partition of Iceberg  v1 table  is deleted, the column is not physically removed in the code but marked as a VoidTransform. As a result, the file path changes to xxx/p2=null/xxx. During data reading, when encountering a VoidTransform, the data should be retrieved directly from the file instead of relying on the partition value. However, in this scenario, Gluten reads the p2 values as null for all records

(Fixes: \#8352)


lyy-pineapple:
@yma11 could you review this,thanks～～

lyy-pineapple:
iceberg [code](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L91)
image

yma11:
> @yma11 could you review this,thanks～～

Possible to add a UT for this scenario?

lyy-pineapple:
> > @yma11 could you review this,thanks～～
> 
> Possible to add a UT for this scenario?

Done ~

yma11 (approved):

PR #8346, opened 2024-12-25
[GLUTEN-7028][CH][Part-14] Refactor Case Sensitive Support for MergeTree

baibaichen:
### problem
This PR refactor how clickhosue backend supports wrting merge tree.  When `spark.sql.caseSensitive` is `false`,

Before this PRs, We convert column name to lower case , and save to disc, for example, given the following DDL

```sql
CREATE TABLE IF NOT EXISTS LINEITEM_MERGETREE_CASE_SENSITIVE
(
 L_ORDERKEY      bigint,
 L_PARTKEY       bigint,
 L_SUPPKEY       bigint,
 L_LINENUMBER    bigint,
 L_QUANTITY      double,
 L_EXTENDEDPRICE double,
 L_DISCOUNT      double,
 L_TAX           double,
 L_RETURNFLAG    string,
 L_LINESTATUS    string,
 L_SHIPDATE      date,
 L_COMMITDATE    date,
 L_RECEIPTDATE   date,
 L_SHIPINSTRUCT  string,
 L_SHIPMODE      string,
 L_COMMENT       string
)
USING clickhouse
PARTITIONED BY (L_SHIPDATE)
TBLPROPERTIES (orderByKey='L_DISCOUNT')
LOCATION '$basePath/LINEITEM_MERGETREE_CASE_SENSITIVE
```

Here are columns.txt 

| before                                                       | after                                                        |
| ------------------------------------------------------------ | ------------------------------------------------------------ |
[...]

PR #8364, opened 2024-12-27
[GLUTEN-7028][CH][Part-15] [MINOR] Fix UTs

baibaichen:

This PR fixed 3 UTs
1. **`GlutenClickHouseMergeTreeCacheDataSuite::test cache mergetree data no partition columns`**, it's fixed by #8346, just reopen test
2. **`GlutenClickHouseMergeTreePathBasedWriteSuite::test mergetree path based table update`** and **`GlutenClickHouseMergeTreePathBasedWriteSuite::test mergetree path based table delete`**, since one pipeline write will collect stats, so that pruning will be more accurate in point query, adding  `withSQLConf(("spark.databricks.delta.stats.skipping", "false"))` to make test suucess.
3.  **`test mergetree with partition with whitespace`**,  whitespace will be saved as `%20` in Delta commit protocol, so that we need follow this, and hence decode it to normal string in backend.
4. other ignored cases are bucket table case, let's ignore it.

(Fixes: #7028)

## How was this patch tested?

Existed ignore uts


zzcclp (approved):
+1

PR #8355, opened 2024-12-26
[GLUTEN-8060][CORE][VL] Various of fixes for the experimental `GlutenShuffleManager`

zhztheplayer:
Note: `GlutenShuffleManager` is still experimental. The change shouldn't impact production behavior.

1. Register `SortShuffleManager` through  `GlutenShuffleManager` by default
2. (Velox backend) Register `ColumnarShuffleManager` through  `GlutenShuffleManager` by default
3. Make `GlutenShuffleManager` work correctly in cluster mode
4. Add some simple test code

This will remove some blockers against the mixed backend work.

PR #8373, opened 2024-12-30
[VL] Add some fixes following #8355

zhztheplayer:
Following #8355

1. Fix "shuffle handle not found" when GlutenShuffleManager is set
2. Fix utility method `isUseGlutenShuffleManager`

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
@@ -20,7 +20,7 @@ import org.apache.gluten.config.GlutenConfig
 
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{DataFrame, Row, TestUtils}
-import org.apache.spark.sql.execution.FormattedMode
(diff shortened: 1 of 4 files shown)


backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
```diff
@@ -304,6 +304,20 @@ class VeloxTPCHV1GlutenShuffleManagerSuite extends VeloxTPCHSuite {
...
+  override protected def runQueryAndCompare(
+      sqlStr: String,
+      compareResult: Boolean,
+      noFallBack: Boolean,
+      cache: Boolean)(customCheck: DataFrame => Unit): DataFrame = {
+    super.runQueryAndCompare(sqlStr, compareResult, noFallBack, cache) {
+      df =>
+        assert(df.queryExecution.executedPlan.collect {
+          case p if p.isInstanceOf[ColumnarShuffleExchangeExec] => p
+        }.nonEmpty)
```
PHILO-HE, line 317: Do we need to respect `noFallBack`?

zhztheplayer: Addressed.

JkSelf (approved):

PR #8380, opened 2024-12-31
[GLUTEN-8379][VL] Support query trace

jinchengchenghh:
Run the MicroBenchmark to generate stage level plan and then enable query trace in benchmark to profile node level query. Benchmark with query trace enabled replaces ValueStreamNode which is hard to serialize to ValuesNode. This issue may be fixed by plan serialization optimization that only serializes the plan node to  profile in velox query trace. https://github.com/facebookincubator/velox/issues/12084

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -178,7 +178,7 @@ WholeStageResultIterator::WholeStageResultIterator(
 std::shared_ptr<velox::core::QueryCtx> WholeStageResultIterator::createNewVeloxQueryCtx() {
   std::unordered_map<std::string, std::shared_ptr<velox::config::ConfigBase>> connectorConfigs;
   connectorConfigs[kHiveConnectorId] = createConnectorConfig();
-
+  static std::atomic<uint32_t> vqId{0}; // Velox query ID, same with taskId.
   std::shared_ptr<velox::core::QueryCtx> ctx = velox::core::QueryCtx::create(
       nullptr,
       facebook::velox::core::QueryConfig{getQueryContextConf()},
@@ -186,7 +186,11 @@ std::shared_ptr<velox::core::QueryCtx> WholeStageResultIterator::createNewVeloxQ
       gluten::VeloxBackend::get()->getAsyncDataCache(),
       memoryManager_->getAggregateMemoryPool(),
       spillExecutor_.get(),
-      "");
+      fmt::format(
+          "Gluten_Stage_{}_TID_{}_VTID_{}",
+          std::to_string(taskInfo_.stageId),
+          std::to_string(taskInfo_.taskId),
+          std::to_string(vqId++)));
   return ctx;
 }
 
@@ -558,6 +562,18 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
       configs[velox::core::QueryConfig::kSparkLegacyDateFormatter] = "false";
     }
 
+    const auto setIfExists = [&](const std::string& glutenKey, const std::string& veloxKey) {
+      const auto valueOptional = veloxCfg_->get<std::string>(glutenKey);
+      if (valueOptional.hasValue()) {
+        configs[veloxKey] = valueOptional.value();
+      }
+    };
+    setIfExists(kQueryTraceEnabled, velox::core::QueryConfig::kQueryTraceEnabled);
+    setIfExists(kQueryTraceDir, velox::core::QueryConfig::kQueryTraceDir);
+    setIfExists(kQueryTraceNodeIds, velox::core::QueryConfig::kQueryTraceNodeIds);
+    setIfExists(kQueryTraceMaxBytes, velox::core::QueryConfig::kQueryTraceMaxBytes);
+    setIfExists(kQueryTraceTaskRegExp, velox::core::QueryConfig::kQueryTraceTaskRegExp);
+    setIfExists(kOpTraceDirectoryCreateConfig, velox::core::QueryConfig::kOpTraceDirectoryCreateConfig);
   } catch (const std::invalid_argument& err) {
     std::string errDetails = err.what();
     throw std::runtime_error("Invalid conf arg: " + errDetails);
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -27,6 +27,7 @@
 #include "utils/ConfigExtractor.h"
 
 #include "config/GlutenConfig.h"
+#include "config/VeloxConfig.h"
 #include "operators/plannodes/RowVectorStream.h"
 
 namespace gluten {
@@ -1206,6 +1207,24 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::constructValueStreamNode(
   return node;
 }
 
+core::PlanNodePtr SubstraitToVeloxPlanConverter::constructValuesNode(
+    const ::substrait::ReadRel& readRel,
+    int32_t streamIdx) {
+  std::vector<RowVectorPtr> values;
+  VELOX_CHECK_LT(streamIdx, inputIters_.size(), "Could not find stream index {} in input iterator list.", streamIdx);
+  const auto iterator = inputIters_[streamIdx];
+  while (iterator->hasNext()) {
+    auto cb = VeloxColumnarBatch::from(defaultLeafVeloxMemoryPool().get(), iterator->next());
+    values.emplace_back(cb->getRowVector());
+  }
+  auto node = std::make_shared<facebook::velox::core::ValuesNode>(nextPlanNodeId(), std::move(values));
+
+  auto splitInfo = std::make_shared<SplitInfo>();
+  splitInfo->isStream = true;
+  splitInfoMap_[node->id()] = splitInfo;
+  return node;
+}
+
 core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::ReadRel& readRel) {
   // emit is not allowed in TableScanNode and ValuesNode related
   // outputs
@@ -1217,7 +1236,12 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
   // Check if the ReadRel specifies an input of stream. If yes, build ValueStreamNode as the data source.
   auto streamIdx = getStreamIndex(readRel);
   if (streamIdx >= 0) {
-    return constructValueStreamNode(readRel, streamIdx);
+    // Only used in benchmark enable query trace, replace ValueStreamNode to ValuesNode to support serialization.
+    if (LIKELY(confMap_[kQueryTraceEnabled] != "true")) {
+      return constructValueStreamNode(readRel, streamIdx);
+    } else {
+      return constructValuesNode(readRel, streamIdx);
+    }
   }
 
   // Otherwise, will create TableScan node for ReadRel.
--- a/docs/developers/QueryTrace.md
+++ b/docs/developers/QueryTrace.md
@@ -1,0 +1,174 @@
+---
+layout: page
+title: QueryTrace
+nav_order: 14
+parent: Developer Overview
+---
+
+# Background
+Currently, we have [MicroBenchmarks](https://github.com/apache/incubator-gluten/blob/main/docs/developers/MicroBenchmarks.md) to profile the Velox plan execution in stage level, now we have query trace replayer to specify plan node id to profile in operator level.
+
+We cannot profile operator level directly because ValueStreamNode cannot serialize and deserialize, so we should generate benchmark first, and then enable query trace in benchmark which replaces the ValueStreamNode to ValuesNode.
+
+Relevant link:
+https://facebookincubator.github.io/velox/develop/debugging/tracing.html
+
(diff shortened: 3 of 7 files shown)


zhli1142015:
We are also looking into this. I tried it locally but encountered the error below. Would you fix it in your PR?
```
Caused by: org.apache.gluten.exception.GlutenException: Exception: VeloxUserError
Error Source: USER
Error Code: UNSUPPORTED
Reason: ValueStream plan node is not serializable
Retriable: False
Function: serialize
File: /var/git/incubator-gluten/cpp/velox/operators/plannodes/RowVectorStream.h
Line: 136
Stack trace:
# 0  facebook::velox::VeloxException::VeloxException(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, facebook::velox::VeloxException::Type, std::basic_string_view<char, std::char_traits<char> >)
# 1  void facebook::velox::detail::veloxCheckFail<facebook::velox::VeloxUserError, char const*>(facebook::velox::detail::VeloxCheckFailArgs const&, char const*)
# 2  gluten::ValueStreamNode::serialize() const
# 3  facebook::velox::core::PlanNode::serialize() const
# 4  facebook::velox::core::TopNNode::serialize() const
# 5  facebook::velox::core::PlanNode::serialize() const
# 6  facebook::velox::core::ProjectNode::serialize() const
# 7  facebook::velox::exec::trace::TaskTraceMetadataWriter::write(std::shared_ptr<facebook::velox::core::QueryCtx> const&, std::shared_ptr<facebook::velox::core::PlanNode const> const&)
# 8  facebook::velox::exec::Task::maybeInitTrace()
```

jinchengchenghh:
No, I just add the config. The exception is because we don't support to serialize ValueStream Node to json, we could support it, it is not hard.

jinchengchenghh:
QueryTrace only supports some operators, we need to extend it.

zhli1142015:
I think we might need to convert the exception below into a warning log to tolerate the issue you mentioned, where only some operators are supported.
https://github.com/facebookincubator/velox/blob/9da5fa7ee803540e330bc626a57388c9cbcad1b3/velox/exec/Operator.cpp#L130

zhli1142015:
And we also need to register spark fucntions in `velox_query_replayer`.

jinchengchenghh:
If you don't set the trace config, we won't run into this exception. The exception indicates you could not enable query trace in that case.

Looks like we still have further more work to do to enable query trace in Gluten

FelixYBW:
@jinchengchenghh what's the difference/advantage from our microbenchmark tool?

jinchengchenghh:
Our benchmark accepts a total velox plan, but query trace can specify the plan node id, it can save the input of any plan node @FelixYBW

zhztheplayer:
> Our benchmark accepts a total velox plan, but query trace can specify the plan node id, it can save the input of any plan node @FelixYBW

Can we consolidate the two functionalities? They sound overlap more or less?

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -558,6 +558,18 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
       configs[velox::core::QueryConfig::kSparkLegacyDateFormatter] = "false";
     }
 
+    const auto setIfExists = [&](const std::string& glutenKey, const std::string& veloxKey) {
```
Yohahaha, line 561: if this func is common, should we declare it as static member function?

jinchengchenghh: Yes, good suggestion, I think the config needs to do the refactor, now if the config not set by java side, we will set a default value of velox config to velox query config, we should change to not set the config.

If velox config default value is changed, we can change automatic with them if we don't have different default value with velox.

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -51,6 +51,14 @@ const std::string kWriteIOTime = "writeIOTime";
...
 const std::string kHiveDefaultPartition = "__HIVE_DEFAULT_PARTITION__";
 
+std::string getQueryId(const std::unordered_map<std::string, std::string>& confMap) {
+  auto it = confMap.find(kQueryTraceQueryId);
+  if (it != confMap.end()) {
+    return it->second;
+  }
+  return "";
+}
+
```
zhztheplayer, line 61: A little bit confused with the empty query ID. Can we always give a meaningful ID to Velox?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1206,6 +1207,24 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::constructValueStreamNode(
...
 }
 
+core::PlanNodePtr SubstraitToVeloxPlanConverter::constructValuesNode(
+    const ::substrait::ReadRel& readRel,
+    int32_t streamIdx) {
+  std::vector<RowVectorPtr> values;
+  VELOX_CHECK_LT(streamIdx, inputIters_.size(), "Could not find stream index {} in input iterator list.", streamIdx);
+  const auto iterator = inputIters_[streamIdx];
+  while (iterator->hasNext()) {
+    auto cb = VeloxColumnarBatch::from(defaultLeafVeloxMemoryPool().get(), iterator->next());
```
zhztheplayer, line 1217: Is the memory managed by Spark? Given that `defaultLeafVeloxMemoryPool` is global.

zhztheplayer (commented):
Is it possible we finally get the generic benchmark and this feature merged in Gluten? Since both are invasive to the core execution code path. E.g., Generic benchmark requires for a individual parameter passed through JNI bridge and this feature requires for changing the `ValueStreamNode` to `ValueNode`.

docs/developers/QueryTrace.md
```diff
@@ -0,0 +1,175 @@
+---
+layout: page
+title: How To Use Gluten
+nav_order: 1
+parent: Developer Overview
+---
```
zhztheplayer, line 6: The header needs to update.

duanmeng:
@jinchengchenghh  It is awesome to support query tracing in gluten. By the way, I am planning to support `PlanNode` serialization without leaf nodes to avoid serializing potential `ValueStreamNode` so that we can support more general query tracing in gluten.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
jinchengchenghh: This is only used in benchmark, so we don't need to track it.

jinchengchenghh:
The GenericBenchmark is frequently used by us CC @JkSelf , it is really usefully for debug in cpp side, we can use vscode to do step-by-step debug. And query trace is widely used in Velox to do correctness verification and performance profiling, so I think we need to enable it in Gluten. and with this optimization https://github.com/apache/incubator-gluten/pull/8380#issuecomment-2594652831, we could drop the ValueStreamNode to ValusNode replace. @zhztheplayer

cpp/velox/compute/WholeStageResultIterator.cc
jinchengchenghh: Can we uses the applicationId_stageId as the queryId?

duanmeng:
> The GenericBenchmark is frequently used by us CC @JkSelf , it is really usefully for debug in cpp side, we can use vscode to do step-by-step debug. And query trace is widely used in Velox to do correctness verification and performance profiling, so I think we need to enable it in Gluten. and with this optimization [#8380 (comment)](https://github.com/apache/incubator-gluten/pull/8380#issuecomment-2594652831), we could drop the ValueStreamNode to ValusNode replace. @zhztheplayer

As discussed with @jinchengchenghh offline, we could add simple `ValueStreamNode` serde methods that serialize only `id` and `output type` and create a `ValueStreamNode` with empty `ResultIterator`. Because we do not need to trace the `ValueStreamNode`.

jinchengchenghh:
I tried the serialize ValueStreamNode as empty serialization, but failed by deserialization.

The plan node deserialization is registered in velox, but ValueStreamNode exists in gluten, so we cannot do that.
https://github.com/facebookincubator/velox/blob/main/velox/core/PlanNode.cpp#L2716
```
root@sr249:/mnt/DP_disk1/code/velox/build/velox/tool/trace# ./velox_query_replayer  --root_dir /tmp/query_trace --task_id Gluten_Stage_0_TID_0_VTID_0 --query_id=query_1 --node_id=7 --summary
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0117 20:08:46.359496 2923199 HiveConnector.cpp:56] Hive connector test-hive created with maximum of 20000 cached file handles.
terminate called after throwing an instance of 'facebook::velox::VeloxUserError'
  what():  Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Deserialization function for class: ValueStreamNode is not registered
Retriable: False
Expression: registry.Has(name)
Function: deserialize
File: /mnt/DP_disk1/code/velox/./velox/common/serialization/Serializable.h
Line: 196
Stack trace:
Stack trace has been disabled. Use --velox_exception_user_stacktrace_enabled=true to enable it
```

jinchengchenghh:
Can you help review again? Thanks! @zhztheplayer

docs/developers/QueryTrace.md
zhztheplayer: @jinchengchenghh 

nit: Would you update `title` (and perhaps other fields as well) in the header? Thanks.

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -2288,4 +2294,48 @@ object GlutenConfig {
...
       .createWithDefault(false)
+
+  val QUERY_TRACE_ENABLED = buildConf("spark.gluten.sql.columnar.backend.velox.queryTraceEnabled")
+    .doc("Enable query tracing flag.")
+    .internal()
+    .booleanConf
+    .createWithDefault(false)
+
+  val QUERY_TRACE_DIR = buildConf("spark.gluten.sql.columnar.backend.velox.queryTraceDir")
+    .doc("Base dir of a query to store tracing data.")
```
zhztheplayer, line 2305: What happens if one leaves this empty?

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: > Can we uses the applicationId_stageId as the queryId?

Do we have applicationId passed through JNI? Or could just align it with the [task name](https://github.com/apache/incubator-gluten/blob/b29aa3b9973d28ed03e9840772734dd1f47b3cdd/cpp/velox/compute/WholeStageResultIterator.cc#L87-L91) somehow.

zhztheplayer (commented):
Some nits. Thank you!

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: If `applicationId` is important for the feature and not yet passed through JNI, could also add it and use it for query context and task names / IDs in another PR. Thank you.

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
jinchengchenghh: If the config is not set correctly, Velox will throw exception, so as other config.
```
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
W0120 15:09:59.768762 3040717 GenericBenchmark.cc:253] Setting CPU for thread 0 to 0
terminate called after throwing an instance of 'facebook::velox::VeloxUserError'
  what():  Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Query trace enabled but the trace dir is not set
Retriable: False
Expression: !queryConfig.queryTraceDir().empty()
Function: maybeMakeTraceConfig
File: /mnt/DP_disk1/code/incubator-gluten/ep/build-velox/build/velox_ep/velox/exec/Task.cpp
Line: 3027
Stack trace:

```

cpp/velox/compute/WholeStageResultIterator.cc
jinchengchenghh: The queryId concept is not really query id, it is velox query id, not Spark queryId, now one query for one task, so I think we can use the same name with task name.

jinchengchenghh: Now we don't pass the applicationId

zhztheplayer (approved):

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
jackylee-ch: Can we set it with default `/tmp/query_trace`?
BTW, if we want expose the config to user, we need remove the `internal()`

jinchengchenghh: This config is only used in internal, user will not set it by this config name, because we need to enable query trace in benchmark, so user set it by FLAGS_XX in benchmark

FelixYBW:
Thank you, Chengcheng for your effort!

PR #8394, opened 2025-01-02
[GLUTEN-8393][VL] Fix the smj result mismatch issue 

JkSelf:

1. Enable full outer join in smj.
2. Fix the inner join result mismatch caused by gluten's incorrect formulation of the join condition. Specifically, gluten incorrectly reorders the condition, such as changing "A = B and C = D" to "C = D and A = B", which leads to a mismatched final join result.
3. Fix the TPC-DS and TPC-H result mismatch issue in anti, semi and full outer join in https://github.com/facebookincubator/velox/pull/11772 and https://github.com/facebookincubator/velox/pull/11771
4. Set the ci job to cover the smj unit test.

## How was this patch tested?

Existing unit tests and locally run TPC-H and TPC-DS.

### DIFF
--- a/.github/workflows/velox_backend.yml
+++ b/.github/workflows/velox_backend.yml
@@ -973,6 +973,70 @@ jobs:
           name: test-report-spark35-slow-ras
           path: '**/surefire-reports/TEST-*.xml'
 
+  run-spark-test-spark35-smj:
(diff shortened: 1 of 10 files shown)


rui-mo (approved):
Thanks. Looks great!

PR #8251, opened 2024-12-17
[VL] Allow shared dependencies for lib GCS

PHILO-HE:
## What changes were proposed in this pull request?

This pr simply removes the setting for CMAKE_FIND_LIBRARY_SUFFIXES before finding lib GCS.

Issue #8232, opened 2024-12-13
[VL] Link openssl dynamicly for FIPS


FelixYBW:
### Description

In some use cases we need to enable FIPS for openssl and use certain version of system installed libcrypto. It's related to security requirement. Let's add the exception of openssl for static link.

FelixYBW:
more details of FIPS can be found here:
https://access.redhat.com/compliance/fips
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/security_hardening/switching-rhel-to-fips-mode_security-hardening#federal-information-processing-standards-140-and-fips-mode_switching-rhel-to-fips-mode

PHILO-HE:
@FelixYBW, can static link still be used for openssl if required openssl & libcrypto version are installed by vcpkg?

FelixYBW:
No, it's said the libcrypto.so is certified. vcpkg does have FIPS enabled option but looks it's not enough. To avoid any potential issue related to security, the safe way is to use system's libcrypto.so always.

So we need to add an option to remove libssl from vcpkg, use system installed one instead. Is it easy?

PHILO-HE:
@FelixYBW, it seems a bit complex to handle it. Openssl is also introduced as a transitive dependency. I note grpc, folly and aws-sdk-cpp are depending on it. If we want to use shared openssl thoroughly, we may have to also remove these libs from vcpkg since static openssl lib is linked to these libs in vcpkg.

FelixYBW:
> @FelixYBW, it seems a bit complex to handle it. Openssl is also introduced as a transitive dependency. I note grpc, folly and aws-sdk-cpp are depending on it. If we want to use shared openssl thoroughly, we may have to also remove these libs from vcpkg since static openssl lib is linked to these libs in vcpkg.

Update here, vcpkg can build openssl as dynamic library and linked to grpc, folly etc, though grpc, folly etc. still are static link.

majetideepak:
> Update here, vcpkg can build openssl as dynamic library and linked to grpc, folly etc, though grpc, folly etc. still are static link.

Won't this cause ODR violations if the dynamic OpenSSL library version differs from the static link? How will users know which library version was statically linked?

majetideepak:
> Update here, vcpkg can build openssl as dynamic library and linked to grpc, folly etc, though grpc, folly etc. still are static link.

To clarify, I am reading this as OpenSSL is always dynamically linked even for transitive dependencies.

[...]

PR #8402, opened 2025-01-02
[VL] Allow shared dependencies for s3 and abfs libs

majetideepak:

This pr removes the setting for CMAKE_FIND_LIBRARY_SUFFIXES 
before finding libs for s3 and abfs.
This is similar to https://github.com/apache/incubator-gluten/pull/8251
Without this change, users are forced to link statically with libcrypto, which is not feasible
with fips. https://github.com/apache/incubator-gluten/issues/8232


PHILO-HE (approved):
@majetideepak, thanks for your patch!

Issue #8414, opened 2025-01-03
ColumnarPartialProjectExec has special filed but not implement doCanonicalize method


lifulong:
### Backend

VL (Velox)

### Bug description

![image](https://github.com/user-attachments/assets/2deba384-9672-4240-a174-013ea545e961)

sample plan log which contains ColumnarPartialProject node, we can see ColumnarPartialProject node not do canonicalized as we hope, so if this stage happend multi times in the whole spark plan, it can not be reused.

### Spark version

None

Issue #3154, opened 2023-09-14
[CORE] BatchScanExecTransformer not override doCanonicalize


liujiayi771:
### Backend

VL (Velox)

### Bug description

BatchScanExecTransformer not override doCanonicalize. We found some cases in our production jobs. In AQE createQueryStages, the exchange cannot be reused because BatchScanExecTransformer cannot return the correct canonicalized.
<img width="1465" alt="image" src="https://github.com/oap-project/gluten/assets/13622031/f7076906-07f1-4d6d-8b60-cfd398f60f2b">

### Spark version

None

PR #8415, opened 2025-01-03
[GLUTEN-8414][VL] Override doCanonicalize in ColumnarPartialProjectEx…

lifulong:
…ec node

implement doCanonicalize in ColumnarPartialProjectExec node, for reuse exchange contains ColumnarPartialProjectExec
(same kind problem as https://github.com/apache/incubator-gluten/issues/3154)
(Fixes: \#8414)

## How was this patch tested?

manual test in our product env

image

the biggest table read four times before implement doCanonicalize in ColumnarPartialProjectExec

image

the biggest table read one times before implement doCanonicalize in ColumnarPartialProjectExec

image

we can also see the stage reuse multi times after implement doCanonicalize in ColumnarPartialProjectExec in spark plan

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
@@ -30,6 +30,7 @@ import org.apache.gluten.vectorized.{ArrowColumnarRow, ArrowWritableColumnVector
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, AttributeReference, CaseWhen, Coalesce, Expression, If, LambdaFunction, NamedExpression, NaNvl, ScalaUDF}
+import org.apache.spark.sql.catalyst.plans.QueryPlan
 import org.apache.spark.sql.execution.{ExplainUtils, ProjectExec, SparkPlan, UnaryExecNode}
 import org.apache.spark.sql.execution.metric.{SQLMetric, SQLMetrics}
 import org.apache.spark.sql.hive.HiveUdfUtil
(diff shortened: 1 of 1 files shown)


FelixYBW:
Thank you for your first PR!

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -75,6 +76,14 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
 
   override def output: Seq[Attribute] = child.output ++ replacedAliasUdf.map(_.toAttribute)
 
+  override def doCanonicalize(): ColumnarPartialProjectExec = {
+    val canonicalized = original.canonicalized.asInstanceOf[ProjectExec]
+    this.copy(
+      original = canonicalized,
+      child = child.canonicalized
+    )(replacedAliasUdf.map(QueryPlan.normalizeExpressions(_, child.output)))
+  }gggg
```
jinchengchenghh, line 85: gggg?

zhztheplayer:
Thanks for the contribution in advance.

I am also curious why `replacedAliasUdf` was placed into 2nd parameter list of the class? Which might be related to the issue per my understanding? cc @jinchengchenghh 

https://github.com/apache/incubator-gluten/blob/4867d60bcde6627abb6f6a0988772b069f1229be/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala#L53-L54

jinchengchenghh:
Because the class `Alias` ExprId is in the 2rd parameters, I would not like to change it when copy. @zhztheplayer 
```
case class Alias(child: Expression, name: String)(
    val exprId: ExprId = NamedExpression.newExprId,
    val qualifier: Seq[String] = Seq.empty,
    val explicitMetadata: Option[Metadata] = None,
    val nonInheritableMetadataKeys: Seq[String] = Seq.empty)
```

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
lifulong: resubmited, copy code from our inner project error

lifulong:
@zhztheplayer @jinchengchenghh hello, is this pr ready for merging ?

zhztheplayer (approved):
Thanks!

PR #8445, opened 2025-01-07
[VL] Gluten-it: New option `--collect-sql-metrics=execution-time` to collect slowest plan nodes into benchmark report

zhztheplayer:
New option `--collect-sql-metrics=execution-time` to make user be able to find out the top 10 query plan nodes that took longest time to execute in the benchmark report.

The option is only available with subcommand `gluten-it queries` at the moment.

For example, run:
```
<gluten-it> queries --local --data-gen=once --enable-history --collect-sql-metrics=execution-time
```

The test report will include:
```
Top 10 plan nodes that took longest time to execute: 

|  Query  | Node ID |         Node Name         |     Execution Time (ns)     |
|---------|---------|---------------------------|-----------------------------|
|   q1.sql|      573|     ProjectExecTransformer|  [time of project] 776031999|
|   q2.sql|     1299|  ColumnarBroadcastExchange|  [time to collect] 769000000|
|   q2.sql|     1241|  ColumnarBroadcastExchange|  [time to collect] 457000000|
|   q3.sql|     3327|  ColumnarBroadcastExchange|  [time to collect] 401000000|
|  q21.sql|    20047|  ColumnarBroadcastExchange|  [time to collect] 396000000|
|   q3.sql|     3272|  ColumnarBroadcastExchange|  [time to collect] 373000000|
|  q21.sql|    19991|  ColumnarBroadcastExchange|  [time to collect] 357000000|
|   q2.sql|     1001|  ColumnarBroadcastExchange|  [time to collect] 328000000|
|  q18.sql|    17275|  ColumnarBroadcastExchange|  [time to collect] 326000000|
|   q9.sql|     9866|  ColumnarBroadcastExchange|  [time to collect] 322000000|
```

PR #7845, opened 2024-11-07
[VL] Re-enable background IO threads by default

zhztheplayer:
Derived from discussion in https://github.com/apache/incubator-gluten/issues/7810.

We disabled Velox's background IO threads by default since https://github.com/apache/incubator-gluten/pull/7165. As we had added shared memory allocation lock and fixed [cross-task spilling issues](https://github.com/apache/incubator-gluten/issues/7243), could explore enabling the feature by default again.

PR #8447, opened 2025-01-07
[VL] Remove an out-of-date warning message

zhztheplayer:
This is a follow-up minor fix for https://github.com/apache/incubator-gluten/pull/7845.

Enabling Velox background IO thread now no longer causes hanging or dead lock issues. Remove the related warning message which could be misleading for users.

Also, simplify gluten-it option `--collect-sql-metrics` which is introduced in https://github.com/apache/incubator-gluten/pull/8445 to `--sql-metrics`.


zhli1142015 (approved):
LGTM, thanks

Issue #8432, opened 2025-01-06
[CH] org.apache.gluten.exception.GlutenException: Not found column 8Ð'ï


lgbo-ustc:
### Backend

CH (ClickHouse)

### Bug description

Got an exception on aggregate.
```
org.apache.gluten.exception.GlutenException: Not found column 8Ð'ï
	at org.apache.gluten.vectorized.ExpressionEvaluatorJniWrapper.nativeCreateKernelWithIterator(Native Method)
	at org.apache.gluten.vectorized.CHNativeExpressionEvaluator.createKernelWithBatchIterator(CHNativeExpressionEvaluator.java:70)
	at org.apache.gluten.backendsapi.clickhouse.CHIteratorApi.createNativeIterator(CHIteratorApi.scala:91)
	at org.apache.gluten.backendsapi.clickhouse.CHIteratorApi.genFinalStageIterator(CHIteratorApi.scala:322)
	at org.apache.gluten.execution.WholeStageZippedPartitionsRDD.$anonfun$compute$1(WholeStageZippedPartitionsRDD.scala:59)
	at org.apache.gluten.utils.Arm$.withResource(Arm.scala:25)
	at org.apache.gluten.metrics.GlutenTimeMetric$.millis(GlutenTimeMetric.scala:37)
	at org.apache.gluten.execution.WholeStageZippedPartitionsRDD.compute(WholeStageZippedPartitionsRDD.scala:46)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:365)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:329)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:365)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:329)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:60)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
	at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:52)
	at org.apache.spark.scheduler.Task.run(Task.scala:136)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$3(Executor.scala:548)
[...]

PR #8450, opened 2025-01-07
[GLUTEN-8432][CH]Remove duplicate output attributes of aggregate's child

lgbo-ustc:

Fixes: #8432 

CH doesn' support duplicate columns in a block. Normally, there is no duplicate attributes in the output of an aggregate's child. There is one exception, when AQE is enabled, an shuffle resutl may be reused. If one of the aggregate function is `count(distinct)` and its arguments contain the grouping keys, duplicate attributes would be introduced. And we cannot erase the duplicate attributes by #8164, since the rule cannot be applied.

## How was this patch tested?

manual tests


zzcclp:
It better to add some test cases for this scenario?

zzcclp (approved):
LGTM

Issue #8455, opened 2025-01-07
[VL] Velox Runtime Error on Encrypted Parquet Files


ArnavBalyan:
### Backend

VL (Velox)

### Bug description

Currently Velox does not support encypted Parquet files. Gluten offloads the scan operator leading to runtime error from Velox.
It leads to job failure, we can prevevnt offloading to Velox if the data is encrypted and gracefully fallback to Java Scan operator. 
```
exception while sleepingorg.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 0.0 failed 4 times, most recent failure: Lost task 0.3 in stage 0.0 (TID 9) 
org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: Exception: VeloxUserError
Error Source: USER
Error Code: UNSUPPORTED
Reason: Encrypted Parquet files are not supported
Retriable: False
Context: Split [Hive: ] Task Gluten_Stage_0_TID_9_VTID_0
Function: initializeSchema
/home/user/gluten/ep/build-velox/build/velox_ep/velox/dwio/parquet/reader/ParquetReader.cpp
Line: 251
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_14VeloxUserErrorEPKcEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox7parquet10ReaderBase16initializeSchemaEv
# 4  _ZN8facebook5velox7parquet10ReaderBaseC1ESt10unique_ptrINS0_4dwio6common13BufferedInputESt14default_deleteIS6_EERKNS5_13ReaderOptionsE
# 5  _ZN8facebook5velox7parquet13ParquetReaderC2ESt10unique_ptrINS0_4dwio6common13BufferedInputESt14default_deleteIS6_EERKNS5_13ReaderOptionsE
# 6  _ZN8facebook5velox7parquet20ParquetReaderFactory12createReaderESt10unique_ptrINS0_4dwio6common13BufferedInputESt14default_deleteIS6_EERKNS5_13ReaderOptionsE
# 7  _ZN8facebook5velox9connector4hive11SplitReader12createReaderEv
[...]

PR #8456, opened 2025-01-07
[GLUTEN-8455][VL] Fallback Scan for Encrypted Parquet Files

ArnavBalyan:
- Currently Scan is offloaded to Velox, however Velox does not support encrypted parquet files.
 - This leads to a Velox side error and runtime failure. Added support to fallback in such cases.
 - We attempt to read parquet footer for the root paths in scan operator, if decryption is detected, the scan fallsback
 - This is behind a config and will not be enabled by default, users can enable scan fallback by passing `spark.gluten.sql.fallbackEncryptedParquet` true.

Fixes: #8455


shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -2278,4 +2281,11 @@ object GlutenConfig {
         "Otherwise, broadcast build relation will use onheap memory.")
       .booleanConf
       .createWithDefault(false)
+
+  val ENCRYPTED_PARQUET_FALLBACK_ENABLED =
+    buildStaticConf("spark.gluten.sql.fallbackEncryptedParquet")
```
Yohahaha, line 2286: why use static conf?

Yohahaha (approved):
LGTM! just one comments

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -191,10 +191,35 @@ object VeloxBackendSettings extends BackendSettingsApi {
       }
     }
 
-    validateScheme().orElse(validateFormat()) match {
-      case Some(reason) => ValidationResult.failed(reason)
-      case _ => ValidationResult.succeeded
+    def validateEncryption(): Option[String] = {
+
+      val encryptionValidationEnabled = GlutenConfig.getConf.enableEncryptedParquetFallback
```
jackylee-ch, line 196: The config name is a little wird to me. maybe `parquetEncryptionValidationEnabled`?

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
```diff
@@ -0,0 +1,137 @@
...
+      fileLimit: Int
+  ): Boolean = {
+
+    val filesIterator: RemoteIterator[LocatedFileStatus] = fs.listFiles(path, true)
+    var checkedFileCount = 0
+    while (filesIterator.hasNext && checkedFileCount < fileLimit) {
+      val fileStatus = filesIterator.next()
+      checkedFileCount += 1
+      try {
+        ParquetFileReader.readFooter(conf, fileStatus.getPath).toString
```
jackylee-ch, line 102: Is it a better way to check encrypted metadata than use the Exception?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
ArnavBalyan: Updated it

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
ArnavBalyan: This is done to keep it backward compatible, spark 33 uses parquet 1.12, support is not there yet

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
ArnavBalyan: updated, this config changing at runtime is an unlikely event, kept removed static to allow any such usecases

ArnavBalyan:
Thanks for the review @Yohahaha addressed comments, could you please take a look

FelixYBW:
CI fails. can you fix format?

Yohahaha:
```
[ERROR] /__w/incubator-gluten/incubator-gluten/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala:196: error: value getConf is not a member of object org.apache.gluten.config.GlutenConfig
[ERROR]       val encryptionValidationEnabled = GlutenConfig.getConf.parquetEncryptionValidationEnabled
[ERROR]                                                      ^
[ERROR] one error found
[ERROR] exception compilation error occurred!!!
```

please rebase and fix conflict.

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
jackylee-ch: Spark 3.5 uses parquet 1.13, and I found it defined in thrift. This could work for Spark 3.5?

jackylee-ch:
@Yohahaha The way to check encrypted file haven't been confirmed, I'm afraid this is not work on Spark 3.5

ArnavBalyan:
> @Yohahaha The way to check encrypted file haven't been confirmed, I'm afraid this is not work on Spark 3.5

Hi @jackylee-ch, spark 3.5 uses parquet 1.15. After parquet 1.13, there is a new field added to check for encryption, which can provide if the file is encrypted. However if we try to read the encrypted file footer, it throws ParquetCryptoRuntimeException. Could you please elaborate on why it may not work on 3.5? Thanks

Alternatively, I was planning to shade parquet 1.14 and bring the shaded parquet as a packaged dependency in Gluten (to check the encryption), which will be a more elegant solution and work with multiple spark version. If there are no concerns with bringing in shaded parquet inside gluten, I'm happy to work on that implementation as well. I could not see such implementations inside Gluten so was a bit hesitant, althought parquet encryption seems to be a special case and could benefit from such a solution. Let me know what you think thanks!

Yohahaha:
> @Yohahaha The way to check encrypted file haven't been confirmed, I'm afraid this is not work on Spark 3.5

@jackylee-ch sorry, I missed your comments. 

the check is disabled by default, it's safe to merge to main branch. if we verified it's not work in Spark 3.5, we can fix with a followup PR.

Yohahaha:
@ArnavBalyan thanks for the update!

would you add UT in different spark shim module with a minimal encrypted parquet file?

we do not need package parquet deps in Gluten I think, just do verification in shim module.

Yohahaha:
@jackylee-ch I'm ok if you want to revert this PR with Spark 3.5 compatibility concern.

jackylee-ch:
> Hi @jackylee-ch, spark 3.5 uses parquet 1.13. After parquet 1.13, there is a new field added to check for encryption, which can provide if the file is encrypted. However if we try to read the encrypted file footer, it throws ParquetCryptoRuntimeException. Could you please elaborate on why it may not work on 3.5? Thanks

You mean that no matter which Spark version we use, we can get `ParquetCryptoRuntimeException` if I try to read the encrypted footer?

BTW what would happen if the footer is not encrypted but the column is encrypted?

ArnavBalyan:
> > Hi @jackylee-ch, spark 3.5 uses parquet 1.13. After parquet 1.13, there is a new field added to check for encryption, which can provide if the file is encrypted. However if we try to read the encrypted file footer, it throws ParquetCryptoRuntimeException. Could you please elaborate on why it may not work on 3.5? Thanks
> 
> @ArnavBalyan You mean that no matter which Spark version we use, we can get `ParquetCryptoRuntimeException` if I try to read the encrypted footer?
> 
> BTW what would happen if the footer is not encrypted but the column is encrypted?

Sure let me add a follow up UT for 3.5, this feature is behind feature flag and verified for parquet 1.13. @jackylee-ch, it would depend how you are doing encryption in your setup. Typically the footer metadata will indicate encryption for newer versions of parquet.

> we do not need package parquet deps in Gluten I think, just do verification in shim module.
@Yohahaha, if we want to keep it backward compatible for parquet 1.12, and not use exception, then we will need a newer parquet inside Gluten regardless of the spark version, in that case, the check can hold true if there are future parquet upgrades inside spark. Using the shim layer we can probably do validation but still will use exception checking for 1.12.

jackylee-ch:
We can use different checker for different Spark version in shims, and for parquet 1.12, the exception is good to me if there is no other better idea.

PR #8461, opened 2025-01-08
[CH] Add tools to dump ActionsDAG into tree graph

taiyang-li:
## How was this patch tested?

actions


cpp-ch/local-engine/Parser/AggregateFunctionParser.cpp
```diff
@@ -25,6 +25,7 @@
 #include <Parser/TypeParser.h>
 #include <Common/CHUtil.h>
 #include <Common/Exception.h>
+#include <Common/logger_useful.h>
```
baibaichen, line 28: should we chanage this CPP?

taiyang-li: Yes it is needed. otherwise LOG_INFO is unknown symbol.

Issue #8462, opened 2025-01-08
[CH] Gluten failed to load META-INF, causing abnormal execution


gleonSun:
### Backend

CH (ClickHouse)

### Bug description

[Expected behavior]
Gluten can load the required Components and Backend, and function normally.

[actual behavior]
Gluten failed to load the required Components and Backend during initialization, resulting in abnormal execution.

### RootCause
The `ResourceUtil#getResources(java.util.regex.Pattern)` method loads META-INF from the `java.class.path`, but it might not be the META-INF from the project's corresponding resource directory. For example, when other projects reference Gluten, it retrieves the META-INF from their resource directories instead, which does not align with expectations.

### Spark version

None

### Relevant logs

```bash
java.lang.UnsupportedOperationException: empty.reduceLeft
        at scala.collection.mutable.ArrayBuffer.scala$collection$IndexedSeqOptimized$$super$reduceLeft(ArrayBuffer.scala:49) ~[scala-library-2.12.15.jar:?]
        at scala.collection.IndexedSeqOptimized.reduceLeft(IndexedSeqOptimized.scala:77) ~[scala-library-2.12.15.jar:?]
        at scala.collection.IndexedSeqOptimized.reduceLeft$(IndexedSeqOptimized.scala:76) ~[scala-library-2.12.15.jar:?]
        at scala.collection.mutable.ArrayBuffer.reduceLeft(ArrayBuffer.scala:49) ~[scala-library-2.12.15.jar:?]
        at scala.collection.TraversableOnce.reduce(TraversableOnce.scala:254) ~[scala-library-2.12.15.jar:?]
        at scala.collection.TraversableOnce.reduce$(TraversableOnce.scala:254) ~[scala-library-2.12.15.jar:?]
        at scala.collection.AbstractTraversable.reduce(Traversable.scala:108) ~[scala-library-2.12.15.jar:?]
        at org.apache.gluten.extension.columnar.transition.ConventionFunc$$anon$1.<init>(ConventionFunc.scala:78) ~[gluten.jar:?]
        at org.apache.gluten.extension.columnar.transition.ConventionFunc$.newOverride(ConventionFunc.scala:76) ~[gluten.jar:?]
        at org.apache.gluten.extension.columnar.transition.ConventionFunc$.create(ConventionFunc.scala:62) ~[gluten.jar:?]
        at org.apache.gluten.extension.columnar.transition.InsertTransitions.<init>(Transitions.scala:27) ~[gluten.jar:?]
[...]

PR #8464, opened 2025-01-08
[GLUTEN-8462][CH] Fixed the loading of Components and Backend

gleonSun:

#8271 Introduced Issue: Gluten failed to load the required Components and Backend during initialization, resulting in abnormal execution.

(Fixes: #8462 )

## How was this patch tested?

TEST ARM and x86

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/utils/ResourceUtil.java
+++ b/gluten-core/src/main/java/org/apache/gluten/utils/ResourceUtil.java
@@ -22,6 +22,7 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


gluten-core/src/main/java/org/apache/gluten/utils/ResourceUtil.java
```diff
@@ -47,12 +47,24 @@ public class ResourceUtil {
    */
   public static List<String> getResources(final Pattern pattern) {
     final List<String> buffer = new ArrayList<>();
-    final String classPath = System.getProperty("java.class.path");
+    String classPath = System.getProperty("java.class.path");
+    processClassPathElements(classPath, pattern, buffer);
+    if (buffer.isEmpty()) {
```
lwz9103, line 52: move this to processClassPathElements method

lwz9103, line 52: please desc in which case buffer will be empty

gleonSun: done

gleonSun: Added comments.

gluten-core/src/test/scala/org/apache/gluten/component/ComponentSuite.scala
```diff
@@ -53,6 +56,18 @@ class ComponentSuite extends AnyFunSuite with BeforeAndAfterAll {
...
+
+  test("Discover without component class") {
+    System.setProperty("java.class.path", classPath)
+    assertThrows[GlutenException](Discovery.discoverAll())
+  }
+
+  // Gluten attempts to load from the "classes" directory instead of "test-classes,"
+  // but the META-INF is not located in the "classes" directory.
+  test("Discover without component class again") {
+    System.setProperty("java.class.path", "")
```
lwz9103, line 68: please use real classpath in your case

zhztheplayer:
@gleonSun Let's remove the test code and keep the core change in this PR to fix your issue first. I could work on another approach rather than "java.class.path" to make sure the component discovery could cover all of the cases.

gleonSun:
> @gleonSun Let's remove the test code and keep the core change in this PR to fix your issue first. I could work on another approach rather than "java.class.path" to make sure the component discovery cover all cases.

okok

gleonSun:

gleonSun:

gluten-core/src/test/scala/org/apache/gluten/component/ComponentSuite.scala
gleonSun: The unit tests are too tricky, and this part of the code will be refactored later.

zhztheplayer (approved):

PR #8468, opened 2025-01-08
[GLUTEN-8462][CORE] Raise a meaningful error when no component is found from classpath

zhztheplayer:
When no component files (with name `META-INF/gluten-components/<classname>`) found from classpath, raise an error instead of running into UB.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/component/package.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/component/package.scala
@@ -39,7 +39,7 @@ package object component extends Logging {
     val components = Component.sortedUnsafe()
     require(
       components.nonEmpty,
-      s"No component found in container directories named with " +
+      s"No component files found in container directories named with " +
         s"'META-INF/gluten-components' from classpath. JVM classpath value " +
         s"is: ${System.getProperty("java.class.path")}"
     )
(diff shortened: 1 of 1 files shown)


gluten-core/src/main/scala/org/apache/gluten/component/package.scala
```diff
@@ -37,6 +37,12 @@ package object component extends Logging {
     // Output log so user could view the component loading order.
     // Call #sortedUnsafe than on #sorted to avoid unnecessary recursion.
     val components = Component.sortedUnsafe()
+    require(
+      components.nonEmpty,
+      s"No component files found in container directories named with " +
+        s"'META-INF/gluten-components' from classpath. JVM classpath value " +
+        s"is: ${System.getProperty("java.class.path")}"
```
PHILO-HE, line 44: @zhztheplayer, is user configured `extraClassPath` covered by `java.class.path`?

zhztheplayer: Yes, usually it is included. I am investigating on issue https://github.com/apache/incubator-gluten/issues/8462 so perhaps we shouldn't rely on `java.class.path` in future. But at his moment this error message is needed.

PHILO-HE (approved):

Issue #7544, opened 2024-10-15
[Core] Qualification Tool for Gluten


srinivasst:
### Description

The purpose of this enhancement is to create a qualification tool that analyzes customer event files to determine which workloads are suitable for execution with Gluten. This is crucial when onboarding new customers, as not all workloads benefit from Gluten's native acceleration—especially workloads with RDD operations, unsupported SQL operators or UDF workloads.

# Proposed Solution:

Develop a Java program to analyze the event files, given a Hadoop file path as input. The program will generate two reports:

- Application Report:
  - Percentage of RDD usage
  - Percentage of Unsupported SQL operations
  - Percentage of supported SQL operations
  - Cumulative task time for each application
  - Recommendation to use Gluten Acceleration (Recommended if Percentage of supported SQL operations >= 70%)
- Unsupported Operator Report:
  - Unsupported SQL operators
  - Impact on cumulative CPU time

## Requirements:

- Compatibility with Hadoop file paths that point to:
  - Single event files
  - Event directories with rolling event files
  - Deeply nested directories containing event files
  - Compressed event files

srinivasst:
WIP - will upload approach and PR in a few days.

xumingming:
Interesting idea, Looking forward to it.

LoseYSelf:
([spark-rapids-tools](https://github.com/NVIDIA/spark-rapids-tools)) is NVIDIA RAPIDS tool, has similar functions

surnaik:
You may give it a try here while @srinivasst works on open sourcing this - https://cloud.google.com/dataproc-serverless/docs/guides/native-query-execution#native_query_execution_qualification_tool

PR #8484, opened 2025-01-09
[GLUTEN-7544][CORE] Add Qualification Tool

srinivasst:
Adding a qualification tool that can go through Spark event files and output percentage of the application that can be run using Gluten.

Fixes: #7544


Yohahaha:
could you help confirm which class/rule do the work that decide whether a operator is supported or not?

srinivasst:
@Yohahaha tools/qualification-tool/src/main/java/org/apache/gluten/qt/support/NodeSupportVisitor.java marks nodes as supported / not supported

FelixYBW:
Good feature! It's something we want to have for long time. Would you add a section and a link to main page (https://github.com/apache/incubator-gluten/blob/main/README.md)

FelixYBW:
Also can you add an example output file?

tools/qualification-tool/pom.xml
```diff
@@ -0,0 +1,205 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.gluten</groupId>
+  <artifactId>qualification-tool</artifactId>
+  <version>1.3.0-SNAPSHOT</version>
+  <name>Qualification Tool</name>
+
+  <properties>
+    <maven.compiler.source>11</maven.compiler.source>
```
surnaik, line 9: Can we rely on the top level pom's java version here?

tools/qualification-tool/pom.xml
```diff
@@ -0,0 +1,205 @@
...
+    <maven.compiler.target>11</maven.compiler.target>
+    <encoding>UTF-8</encoding>
+    <scala.version>2.12.15</scala.version>
```
surnaik, line 12: For all the configs which can take the value from top level pom, can we do that?

tools/qualification-tool/pom.xml
```diff
@@ -0,0 +1,205 @@
...
+        <version>3.8.1</version>
+        <executions>
+          <execution>
+            <id>default-compile</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+            <configuration>
+              <source>11</source>
```
surnaik, line 99: Read value from previous config?

srinivasst:
@FelixYBW I have addressed your comments and added sample output.
I also have added an UT that compares output against expected results.

tools/qualification-tool/pom.xml
srinivasst: Inheriting values from the parent POM resulted in some classes not being included in the tool. This issue occurred because the parent POM excludes certain classes from the Spark JARs that are required by this tool. Overriding the parent Spark dependencies to include those classes proved to be challenging. As a result, I decided to exclude the parent POM.

surnaik (approved):

FelixYBW:
> @FelixYBW I have addressed your comments and added sample output. I also have added an UT that compares output against expected results.

Thank you. Let me have a quick try.

FelixYBW (approved):
Looks good to me. Thank you to add the big feature!

zhixingheyi-tian:
Hi @srinivasst 
I have one question:
Why it is necessary to use JDK 11+ version? Does JDK 8 not meet the requirements?

Thanks

FelixYBW:
> Hi @srinivasst I have one question: Why it is necessary to use JDK 11+ version? Does JDK 8 not meet the requirements?


You may try but I don't think jdk8 works.

tools/qualification-tool/src/main/java/org/apache/gluten/qt/support/NodeSupportVisitor.java
```diff
@@ -0,0 +1,225 @@
...
+                          executionDescription.getMetrics(metric.getAccumulatorId()))
+                          .stream())
+              .flatMap(
+                  metricStr ->
+                      Arrays.stream(metricStr.split("\n", -1))
+                          .reduce((first, second) -> second)
+                          .stream())
+              .flatMap(metricStr -> extractSize(metricStr).stream())
+              .flatMap(metricStr -> stringToBytes(metricStr).stream())
+              .anyMatch(size -> size.compareTo(BigInteger.valueOf(100L << 20)) <= 0);
```
ayushi-agarwal, line 191: Could you help understand why there is special case for BroadcastHashJoin?

Issue #8493, opened 2025-01-10
[VL] Security Voliation NULL Pointer Dereference


jkhaliqi:
### Backend

VL (Velox)

### Bug description

We scanned couple of security violations in the code and will add fix. It helps the code to be more reliable.
NULL Pointer Dereference@cpp/core/benchmarks/[CompressionBenchmark.cc:45](http://compressionbenchmark.cc:45/)

### Spark version

None

jkhaliqi:
Closing as this was a false positive since in Benchmark

PR #8494, opened 2025-01-10
[GLUTEN-8493][VL] Fix null pointer dereference

jkhaliqi:

Initialize all values to nullptr

(Fixes: \#8493)

### DIFF
--- a/cpp/core/benchmarks/CompressionBenchmark.cc
+++ b/cpp/core/benchmarks/CompressionBenchmark.cc
@@ -42,7 +42,7 @@ void printTrace(void) {
   char** strings;
   size_t i, size;
   enum Constexpr { kMaxSize = 1024 };
-  void* array[kMaxSize];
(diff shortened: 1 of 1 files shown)


zhztheplayer (approved):

cpp/core/benchmarks/CompressionBenchmark.cc
```diff
@@ -42,7 +42,7 @@ void printTrace(void) {
   char** strings;
   size_t i, size;
   enum Constexpr { kMaxSize = 1024 };
-  void* array[kMaxSize];
+  void* array[kMaxSize] = {};
```
majetideepak, line 45: This change does not resolve any potential null pointer dereference. This can be a false positive since this is a Benchmark.

jkhaliqi: Gotcha, closing the issue and PR with false positive, thank you!

Issue #8497, opened 2025-01-10
[VL] Use of columnar table cache causes error when data emitted from the cached plan is in vanilla columnar format


zhztheplayer:
Edit: Plan to fix the issue:

1. Add test case https://github.com/apache/incubator-gluten/pull/8498
2. Add a query planer hook similar to AdaptiveContext.isAdaptiveContext, to determine whether the Gluten columnar rule is called for columnar table cache's inner query plan's optimization https://github.com/apache/incubator-gluten/pull/8551
3. Unify the hook API for columnar table cache and AQE https://github.com/apache/incubator-gluten/pull/8551
4. Fix test cases using the new hook API. Add a rule to make sure the output convention of the query plan conform to vanilla Spark's table cache optimization code. Specifically: (https://github.com/apache/incubator-gluten/pull/9230)
   - For query plan without a topmost ColumnarToRow, keep it as is;
   - For query plan with a topmost ColumnarToRow (which will be removed by Spark' code `convertToColumnarIfPossible`) and the second node is not in Velox columnar fomat, compare the costs between the following 2 query plan candidates, then choose the cheaper one:
      - The input plan unchanged, with a top fake unary node that prevents columnar cache to be used;
      - Velox columnar plan + a fake C2R to be removed;

The [current code](https://github.com/apache/incubator-gluten/blob/6d71d27884dd6bda38848cb012a63fb37ee2d298/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala#L103-L117) for determining whether to use columnar-based cache doesn't cover all possible cases. For example, the case when a vanilla Spark columnar scan fell back by unsupported data type.

If the case is met, the following error will be thrown:

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 10.0 failed 1 times, most recent failure: Lost task 0.0 in stage 10.0 (TID 10) (55f8ad28c275 executor driver): java.lang.IllegalStateException: Heavy batch should consist of arrow vectors
	at org.apache.gluten.columnarbatch.ColumnarBatches.identifyBatchType(ColumnarBatches.java:81)
	at org.apache.gluten.columnarbatch.ColumnarBatches.isLightBatch(ColumnarBatches.java:99)
	at org.apache.gluten.columnarbatch.ColumnarBatches.ensureOffloaded(ColumnarBatches.java:134)
[...]

PR #8498, opened 2025-01-10
[GLUTEN-8497][VL] A bad test case that fails columnar table cache query

zhztheplayer:
https://github.com/apache/incubator-gluten/issues/8497

Add a bad case which is ignored at the moment to reproduce the bug described in https://github.com/apache/incubator-gluten/issues/8497.

I am still looking for a proper solution of the bug. Could merge the test code in advance so one could try with it if interested.


FelixYBW:
Good catch. We need to have a way to automatically convert any columnar format to Velox format. Do we need to hack spark code to do the conversion?

zhztheplayer:
> Good catch. We need to have a way to automatically convert any columnar format to Velox format.

Yes, we already have the [transition system](https://github.com/apache/incubator-gluten/tree/main/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition) to insert the convertions automaically in query planning phase. This bug spcifically applies to columnar table cache feature since the conversions are expected to be done at execution time.

> Do we need to hack spark code to do the conversion?

For fixing this issue, I am not sure yet. Will keep looking for solutions.

PHILO-HE (approved):

PR #8500, opened 2025-01-10
[GLUTEN-8499][VL] Fix the logic of cast expression

### LINKED ISSUE #8499: [VL] Result mismatch of `try_cast`
### Backend

VL (Velox)

### Bug description

reproduce:

```
select try_cast('1.1' as int)
```

result of vanilla spark:
```
NULL
```

result of gluten:

```
1
```

### Spark version

Spark-3.5.x

### Spark configurations

_No response_

### System information

_No response_

### Relevant logs

_No response_

wForget:

The cast expression in ansi mode not only returns null on error, but also has different calculation logic from non-ansi mode, so we should fall back the cast expression in ansi mode.

Like `cast string to long`, they define different `allowDecimal`:

https://github.com/apache/spark/blob/3c831ebf7cd81c2ed774c692b2da7a35ce0fa252/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java#L1612

https://github.com/apache/spark/blob/3c831ebf7cd81c2ed774c692b2da7a35ce0fa252/common/unsafe/src/main/java/org/apache/spark/unsafe/types/UTF8String.java#L1813

Closes #8499

## How was this patch tested?

unit test

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
@@ -1482,7 +1482,7 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
     withTempView("try_cast_table") {
       withTempPath {
         path =>
-          Seq[(String)](("123456"), ("000A1234"), ("1.1"))
+          Seq[(String)](("123456"), ("000A1234"), ("1.1"), ("1a.1"))
             .toDF("str")
             .write
             .parquet(path.getCanonicalPath)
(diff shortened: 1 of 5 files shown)


FelixYBW:
Velox doesn't support ANSI mode yet, we should fallback to Vanilla spark once user set ANSI mode. It will be big issue on Spark4.0

wForget:
> Velox doesn't support ANSI mode yet, we should fallback to Vanilla spark once user set ANSI mode. It will be big issue on Spark4.0

Yes, there are still a lot of works to support ansi mode, until then we have to fallback it to avoid data consistency issues.

zhli1142015:
> Velox doesn't support ANSI mode yet, we should fallback to Vanilla spark once user set ANSI mode. It will be big issue on Spark4.0

Should we start the work of ANSI mode support? 
cc @zhouyuan, @rui-mo and @weiting-chen .
Thanks.

FelixYBW:
> Should we start the work of ANSI mode support? cc @zhouyuan, @rui-mo and @weiting-chen . Thanks.

It depends on customer requirement. Not sure its importance yet. Do you have request to support this year?

Issue #8471, opened 2025-01-08
[VL] Security Voliations


FelixYBW:
### Backend

VL (Velox)

### Bug description

We scaned couple of security voliations in the code and will add fix. It helps the code to be more reliable.

@zhli1142015 @surnaik Not sure if it has overlap with the ones you have done.

### Spark version

None

PR #8506, opened 2025-01-10
[VL] fix security volation 

FelixYBW:
Avoid Public Native Methods

#8471


FelixYBW:
Close it since new tool doesn't report issues

Issue #8524, opened 2025-01-14
[VL] Security Voliations Detect and handle input/output errors


jkhaliqi:
### Backend

VL (Velox)

### Bug description

We scanned a couple of security violations in the code and will add the fix. It helps the code to be more reliable.

### Spark version

None

### Relevant logs

```bash

```

PR #8525, opened 2025-01-14
[GLUTEN-8524][VL] Fix input output errors

jkhaliqi:

Detect and handle input/output errors

(Fixes: \#8524)

### DIFF
--- a/cpp/core/benchmarks/CompressionBenchmark.cc
+++ b/cpp/core/benchmarks/CompressionBenchmark.cc
@@ -45,8 +45,11 @@ void printTrace(void) {
   void* array[kMaxSize];
   size = backtrace(array, kMaxSize);
   strings = backtrace_symbols(array, size);
-  for (i = 0; i < size; i++)
-    printf("    %s\n", strings[i]);
+  for (i = 0; i < size; i++) {
+    if (strings[i] != nullptr) {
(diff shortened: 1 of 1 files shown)


cpp/core/benchmarks/CompressionBenchmark.cc
```diff
@@ -45,8 +45,13 @@ void printTrace(void) {
   void* array[kMaxSize];
   size = backtrace(array, kMaxSize);
   strings = backtrace_symbols(array, size);
-  for (i = 0; i < size; i++)
-    printf("    %s\n", strings[i]);
+  for (i = 0; i < size; i++) {
+    if (strings[i] != nullptr) {
+      printf("    %s\n", strings[i]);
+    } else {
+      printf("It was null at i = %lu\n", i);
```
FelixYBW, line 52: It prints the symbols. You may either add an assert before print or just ignore null pointer.  Don't print the this.

FelixYBW: @jkhaliqi can you update?

jkhaliqi: updated, ignoring the else case!

FelixYBW (approved):

PR #8527, opened 2025-01-14
[CORE] Use RAS's cost model for legacy transition planner to evaluate cost of transitions

zhztheplayer:
This combines RAS cost model into transition graph's cost model to unify the cost evaluation code of transitions.

For example, a transition from Arrow Java columnar batch (referred to as Arrow Java batch) to Velox columnar batch (referred to as Velox batch) could be selected from the following candidates:

1. Arrow Java Batch ->(C2R) Spark Row ->(Velox R2C) Velox Batch
2. Arrow Java Batch ->(Offload) Arrow Native Batch ->(Import) Velox Batch

The transition planner should evaluate the candidates' costs then select a cheaper one (for the case, usually the second one) based on the cost model. After the change, it will reuse RAS's cost model instead of the hard coded transition cost model which will be removed in this PR.

Also, remove the `rough2` RAS cost model as unused.


zhztheplayer:

JkSelf (approved):

PR #8541, opened 2025-01-15
[CORE][VL] Cost model code refactors

zhztheplayer:
Add new independent trait `GlutenCostModel` so legacy transition planner's cost model code doesn't have to depend on RAS cost model code.

After the change, both legacy transition planner and RAS optimizer will rely on `GlutenCostModel` through their own API adapters.

Also, make costers be registered directly through a individual API of `Component`. This will enable the code to create cost model from anywhere without having a Spark session started in advance.


zhouyuan (approved):
a big refactor! 
+1 assuming CI green

PR #8542, opened 2025-01-15
[GLUTEN-8538][VL] Fix incorrect calculation of buffer size

jkhaliqi:

 Fix incorrect calculation of buffer size
False positives:
cpp/velox/udf/UdfLoader.cc:79
cpp/velox/udf/UdfLoader.cc:106

(Fixes: \#8538)

### DIFF
--- a/cpp/core/memory/MemoryAllocator.cc
+++ b/cpp/core/memory/MemoryAllocator.cc
@@ -121,6 +121,7 @@ void ListenableMemoryAllocator::updateUsage(int64_t size) {
 }
 
 bool StdMemoryAllocator::allocate(int64_t size, void** out) {
+  GLUTEN_CHECK(size >= 0, "size is less than 0");
   *out = std::malloc(size);
   if (*out == nullptr) {
     return false;
@@ -130,6 +131,8 @@ bool StdMemoryAllocator::allocate(int64_t size, void** out) {
 }
 
 bool StdMemoryAllocator::allocateZeroFilled(int64_t nmemb, int64_t size, void** out) {
+  GLUTEN_CHECK(nmemb >= 0, "nmemb is less than 0");
+  GLUTEN_CHECK(size >= 0, "size is less than 0");
   *out = std::calloc(nmemb, size);
   if (*out == nullptr) {
     return false;
@@ -139,6 +142,7 @@ bool StdMemoryAllocator::allocateZeroFilled(int64_t nmemb, int64_t size, void** 
 }
 
 bool StdMemoryAllocator::allocateAligned(uint64_t alignment, int64_t size, void** out) {
+  GLUTEN_CHECK(size >= 0, "size is less than 0");
   *out = aligned_alloc(alignment, size);
   if (*out == nullptr) {
     return false;


cpp/core/memory/MemoryAllocator.cc
```diff
@@ -121,6 +121,9 @@ void ListenableMemoryAllocator::updateUsage(int64_t size) {
 }
 
 bool StdMemoryAllocator::allocate(int64_t size, void** out) {
+  if (size < 0) {
+    return false;
+  }
```
FelixYBW, line 126: Let's use assert

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -130,6 +133,10 @@ bool StdMemoryAllocator::allocate(int64_t size, void** out) {
 }
 
 bool StdMemoryAllocator::allocateZeroFilled(int64_t nmemb, int64_t size, void** out) {
```
FelixYBW, line 135: @zhztheplayer can we change to uint64\_t here?

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -130,6 +133,10 @@ bool StdMemoryAllocator::allocate(int64_t size, void** out) {
...
+  size_t totalSize = nmemb * size;
+  if (totalSize / size != nmemb || nmemb < 0 || size < 0) {
+    return false;
+  }
```
FelixYBW, line 139: isn't `nmemb<0 || size <0` enough?  
  
use assert instead

jkhaliqi: Right I was thinking we had to check to make sure there was no overflow but if calloc fails then it will give nullptr anyways and return false so the sizing check being bigger than 0 should be enough!

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -121,6 +121,7 @@ void ListenableMemoryAllocator::updateUsage(int64_t size) {
 }
 
 bool StdMemoryAllocator::allocate(int64_t size, void** out) {
+  assert(size < 0);
```
FelixYBW, line 124: GLUTEN\_CHECK

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -130,6 +131,7 @@ bool StdMemoryAllocator::allocate(int64_t size, void** out) {
 }
 
 bool StdMemoryAllocator::allocateZeroFilled(int64_t nmemb, int64_t size, void** out) {
+  assert(nmemb < 0 && size < 0);
```
FelixYBW, line 134: GLUTEN\_CHECK

cpp/core/memory/MemoryAllocator.cc
```diff
@@ -139,6 +141,7 @@ bool StdMemoryAllocator::allocateZeroFilled(int64_t nmemb, int64_t size, void**
 }
 
 bool StdMemoryAllocator::allocateAligned(uint64_t alignment, int64_t size, void** out) {
+  assert(size < 0);
```
FelixYBW, line 144: GLUTEN\_CHECK

FelixYBW (approved):

PR #8548, opened 2025-01-16
[CORE] Optimize duplicated code for create rel node

zml1206:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


gluten-substrait/src/main/scala/org/apache/gluten/substrait/rel/RelBuilderUtil.scala
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.gluten.substrait.expression.ExpressionNode
+import org.apache.gluten.substrait.extensions.{AdvancedExtensionNode, ExtensionBuilder}
+
+import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression}
+
+import java.util.{List => JList}
+
+import scala.collection.JavaConverters._
+
+object RelBuilderUtil {
```
zhztheplayer, line 32: Any consideration not adding the APIs to `RelBuilder` directly?

zhztheplayer (commented):
Just a minor. Thanks

gluten-substrait/src/main/scala/org/apache/gluten/substrait/rel/RelBuilderUtil.scala
zml1206: RelBuilder is a java class, add it in requires additional type conversion, such as scala `Seq`.

zhztheplayer: There is also Java classes like `JList[ExpressionNode]` in this file. Can we just put them to RelBuilder? Then make caller call "asJava" when passing the  Scala collections which should be trivial. Otherwise developer will have to know 2 different helper classes for similar purposes.

zml1206: Mainly it seems inappropriate to call replaceWithExpressionTransformer in RelBuild. Here I extracted part of the logic, stay up replaceWithExpressionTransformer, CI failure unrelated, please take a look again, thank you.

zhztheplayer (approved):

PR #8551, opened 2025-01-16
[GLUTEN-8497][CORE] A unified CallInfo API to replace AdaptiveContext

zhztheplayer:
Background:

The overall purpose is to add an API to determine whether a columnar rule call is for columnar table cache's cached plan's optimization, which is a subtask of #8497.

Solution:

Since we already have `AdaptiveContext` for similar purpose but for deciding whether the call is for AQE incremental optimization,  the patch adds a new API `CallInfo` and incorporates  both `isAqe()` (to replace `AdaptiveContext`) and `isCache` (for #8497) into it.

`AdaptiveContext` is removed in this PR.


JkSelf (approved):
LGTM.

PR #8555, opened 2025-01-17
[GLUTEN-6067] Open spark 35 ut

baibaichen:

(Fixes: \#6067)

## How was this patch tested?


baibaichen:
@CodiumAI-Agent /review

gluten-ut/common/src/test/scala/org/apache/gluten/utils/BackendTestSettings.scala
```diff
@@ -98,6 +98,13 @@ abstract class BackendTestSettings {
       exclusion.add(Exclude(testNames: _*))
       this
     }
+    def includeCH(testNames: String*): SuiteSettings = {
+      this
+    }
```
lgbo-ustc, line 103: Nothing happen?

baibaichen: yes，it's used to indicate this case is supported by ch backend

baibaichen: yes，it's used to indicate this case is supported by ch backend

lgbo-ustc: I just consider the method names are not general for backends, They are only used in CH test settings. How about this ?
```scala
def includeBackend(backendName, String, testNames: String*)
def excludeBackend(backendName, String, testNames: String*)
```

lgbo-ustc (approved):

baibaichen:
we now prepare to enable spark35 and disable spark 32
image

PR #8560, opened 2025-01-17
[GLUTEN-8455][VL] Support encrypted parquet fallback for 3.5

ArnavBalyan:
- Adds support for encrypted parquet fallback for Spark 3.5.
 - Previous Spark versions use an older parquet dependency which does not provide direct metadata to check encryption.
 - Using `EncryptionType` provided by Parquet 1.13 for Spark 3.5.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/utils/ParquetEncryptionDetectionSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/utils/ParquetEncryptionDetectionSuite.scala
@@ -112,7 +112,7 @@ class ParquetEncryptionDetectionSuite extends GlutenQueryTest {
 
   testWithSpecifiedSparkVersion(
     "Detect encrypted Parquet with encrypted footer",
-    Array("3.2", "3.3", "3.4")) {
+    Array("3.2", "3.3", "3.4", "3.5")) {
     withTempDir {
       tempDir =>
         val filePath = s"${tempDir.getAbsolutePath}/encrypted_footer.parquet"
@@ -136,7 +136,7 @@ class ParquetEncryptionDetectionSuite extends GlutenQueryTest {
 
   testWithSpecifiedSparkVersion(
     "Detect encrypted Parquet without encrypted footer (plaintext footer)",
-    Array("3.2", "3.3", "3.4")) {
+    Array("3.2", "3.3", "3.4", "3.5")) {
     withTempDir {
       tempDir =>
         val filePath = s"${tempDir.getAbsolutePath}/plaintext_footer.parquet"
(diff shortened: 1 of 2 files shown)


shims/spark35/src/main/scala/org/apache/gluten/sql/shims/spark35/Spark35Shims.scala
```diff
@@ -554,7 +559,22 @@ class Spark35Shims extends SparkShims {
...
+        ParquetFileReader.readFooter(conf, fileStatus.getPath, ParquetMetadataConverter.NO_FILTER)
+      val fileMetaData = footer.getFileMetaData
+      fileMetaData.getEncryptionType match {
+        case EncryptionType.UNENCRYPTED =>
+          false
+        case EncryptionType.PLAINTEXT_FOOTER =>
+          true
+        case _ =>
+          false
+      }
```
Yohahaha, line 573: why return file is encrypted when `EncryptionType.PLAINTEXT_FOOTER`?

would you post official parquet encrypt type doc?

ArnavBalyan: Sorry accidentally deleted the comment, here is the code ref - https://github.com/apache/parquet-java/blob/master/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L1766. We still rely on the exception to check the footer + file encryption. The `EncryptionType.PLAINTEXT_FOOTER` checks for encrypted file but plain footer. It seems there is no doc on this, but hope the code ref helps. thanks!

jackylee-ch: Could you add some comments to it? It should have some `EncryptionType` description and how we deal with it.

ArnavBalyan: Added comments

backends-velox/src/test/scala/org/apache/gluten/utils/ParquetEncryptionDetectionSuite.scala
```diff
@@ -112,7 +112,7 @@ class ParquetEncryptionDetectionSuite extends GlutenQueryTest {
 
   testWithSpecifiedSparkVersion(
     "Detect encrypted Parquet with encrypted footer",
-    Array("3.2", "3.3", "3.4")) {
+    Array("3.2", "3.3", "3.4", "3.5")) {
```
Yohahaha, line 115: it's time to replace `testWithSpecifiedSparkVersion` to `test`, right?

Yohahaha (approved):
LGTM! only one comments

backends-velox/src/test/scala/org/apache/gluten/utils/ParquetEncryptionDetectionSuite.scala
ArnavBalyan: Sure let me update

jackylee-ch (approved):

PR #8566, opened 2025-01-19
[GLUTEN-8565][VL] Support CollectLimit Operator

ArnavBalyan:
- Currently CollectLimit falls back, as a result all partitions must undergo VeloxColumnarToRow
 - Adds support for CollectLimit, we prune ColumnarBatches locally within each partition before shuffling.
 - Next reduce to single partition, and apply another local limit on the final partition, data remains light batch throughout


backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
```diff
@@ -0,0 +1,164 @@
...
+        val prunedBatch = pruneBatch(currentBatch, remaining)
+        rowsCollected += remaining
+        nextBatch = Some(prunedBatch)
+      }
+      true
+    }
+  }
+
+  private def pruneBatch(batch: ColumnarBatch, neededRows: Int): ColumnarBatch = {
+    val loadedBatch = ColumnarBatches.load(ArrowBufferAllocators.contextInstance, batch)
```
jinchengchenghh, line 95: This will convert the input velox batch to Arrow ColumnarBatch, which is a heavy load. And then convert the arrow batch to velox. Can we have a JNI call to prune velox batch? BaseVector::slice may help you. https://github.com/facebookincubator/velox/blob/main/velox/vector/BaseVector.h#L541

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
```diff
@@ -0,0 +1,164 @@
...
+      ),
+      readMetrics
+    )
+  }
+
+  override protected def doExecute(): RDD[org.apache.spark.sql.catalyst.InternalRow] =
+    throw new UnsupportedOperationException("doExecute is not supported for this operator")
+
+  override def rowType0(): Convention.RowType = Convention.RowType.None
+
```
jinchengchenghh, line 159: Does this operator also requires https://github.com/apache/incubator-gluten/issues/3154?

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
```diff
@@ -0,0 +1,164 @@
...
+        serializer,
+        writeMetrics,
+        metrics,
+        useSortBasedShuffle
+      ),
+      readMetrics
+    )
+  }
+
+  override protected def doExecute(): RDD[org.apache.spark.sql.catalyst.InternalRow] =
```
jinchengchenghh, line 155: Do we need to override? The super class `ColumnarCollectLimitExecBaseTransformer` implement is same

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExecBaseTransformer.scala
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.apache.gluten.execution
+
+import org.apache.gluten.backendsapi.BackendsApiManager
+import org.apache.gluten.extension.ValidationResult
+
+import org.apache.spark.sql.catalyst.plans.physical.{Partitioning, SinglePartition}
+import org.apache.spark.sql.execution.{CollectLimitExec, LimitExec, SparkPlan}
+
+abstract class ColumnarCollectLimitExecBaseTransformer(
```
jinchengchenghh, line 25: We don't need the xxTransformer, this is only used in some plan node will be changed to substrait plan, and then change to velox plan.

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/execution/GlutenSQLCollectLimitExecSuite.scala
```diff
@@ -0,0 +1,117 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.execution
+
+import org.apache.gluten.execution.ColumnarCollectLimitExecBaseTransformer
+
+import org.apache.spark.sql.{DataFrame, GlutenSQLTestsTrait, Row}
+
+class GlutenSQLCollectLimitExecSuite extends GlutenSQLTestsTrait {
```
jinchengchenghh, line 23: Please move it to module backends_velox as https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/test/scala/org/apache/gluten/execution/ArrowCsvScanSuite.scala

Class in this module usually extends Spark class test suite and exists in all Spark version.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
ArnavBalyan: That's a good point, since this only happens with the last batch, had implemented in JVM, moved to a JNI call

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
ArnavBalyan: No, this operator does not provide doCanonicalize

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
ArnavBalyan: Updated it

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExecBaseTransformer.scala
ArnavBalyan: Updated this

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/execution/GlutenSQLCollectLimitExecSuite.scala
ArnavBalyan: Moved

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
```diff
@@ -0,0 +1,160 @@
...
+        val currentBatch = partitionIter.next()
+        val currentBatchRowCount = currentBatch.numRows()
+        val remaining = limit - rowsCollected
+
+        if (currentBatchRowCount <= remaining) {
+          ColumnarBatches.retain(currentBatch)
+          rowsCollected += currentBatchRowCount
+          nextBatch = Some(currentBatch)
+        } else {
+          val prunedBatch = ColumnarBatchUtils.pruneBatch(currentBatch, remaining)
```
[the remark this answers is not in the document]
ArnavBalyan: Yes, moving to JNI will require retain, the previous logic was copying batches. Added the retain

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
```diff
@@ -0,0 +1,160 @@
...
+      /**
+       * Attempt to fetch the next batch from the underlying iterator if we haven't yet hit the
+       * limit. Returns true if we found a new batch, false otherwise.
+       */
+      private def fetchNext(): Boolean = {
+        if (rowsCollected >= limit || !partitionIter.hasNext) {
+          return false
+        }
+
+        val currentBatch = partitionIter.next()
```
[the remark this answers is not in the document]
ArnavBalyan: The nextBatch will not get populated if we early return here, added a look ahead to keep things simple and elegant. this probably can be implemented in other ways

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
ArnavBalyan: Ah probably missed it, should be updated now

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitBaseExec.scala
```diff
@@ -0,0 +1,66 @@
...
+  override protected def doExecute()
+      : org.apache.spark.rdd.RDD[org.apache.spark.sql.catalyst.InternalRow] = {
+    throw new UnsupportedOperationException(s"This operator doesn't support doExecute().")
+  }
+
+}
+object ColumnarCollectLimitBaseExec {
+  def from(collectLimitExec: CollectLimitExec): ColumnarCollectLimitBaseExec = {
+    BackendsApiManager.getSparkPlanExecApiInstance
+      .genColumnarCollectLimitExec(
```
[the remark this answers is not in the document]
ArnavBalyan: Good point, currently offset is only supported on 3.4 and 3.5. Moved this to 3.2, 3.3 temporarily, I'll add a followup to add support for 33+ in the shim layer thanks!

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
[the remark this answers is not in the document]
jinchengchenghh: For the iterator definition, we should call hasNext first and call next, so we don;t need to call `hasNext` in the `next` function, and hasNext should be a simple function, it can even be called once, but `next` should be called once. So if we use the structure I suggest, we can drop `private var nextBatch: Option[ColumnarBatch] = None`, I think it will be more precise.

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitBaseExec.scala
```diff
@@ -0,0 +1,68 @@
...
+
+    if (!isSupported) {
+      return ValidationResult.failed(
+        s"CollectLimitExec is not supported by the current backend."
+      )
+    }
+
+    if (
+      (childPlan.supportsColumnar && GlutenConfig.get.enablePreferColumnar) &&
+      BackendsApiManager.getSettings.supportColumnarShuffleExec() && (getSparkVersion.startsWith(
```
jinchengchenghh, line 46: Please don't directly check the SparkVersion, use the shim layer function like that https://github.com/apache/incubator-gluten/blob/main/shims/spark32/src/main/scala/org/apache/gluten/sql/shims/spark32/Spark32Shims.scala

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatchJniWrapper.java
```diff
@@ -34,6 +34,8 @@ public static VeloxColumnarBatchJniWrapper create(Runtime runtime) {
 
   public native long compose(long[] batches);
 
+  public native long pruneBatch(long batchHandle, int limit);
```
jinchengchenghh, line 37: Please rename it to slice, this is a common function, maybe used by others in the future. And also change the argument name to align with velox.

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitBaseExec.scala
ArnavBalyan: done

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatchJniWrapper.java
ArnavBalyan: done

backends-velox/src/main/scala/org/apache/gluten/utils/ColumnarBatchUtils.scala
```diff
@@ -0,0 +1,56 @@
...
+ */
+package org.apache.gluten.utils
+
+import org.apache.gluten.backendsapi.BackendsApiManager
+import org.apache.gluten.columnarbatch.{ColumnarBatches, VeloxColumnarBatchJniWrapper}
+import org.apache.gluten.runtime.{Runtime, Runtimes}
+
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+object ColumnarBatchUtils {
```
zhztheplayer, line 25: Any particular reason not putting the method into `VeloxColumnarBatches`?

ArnavBalyan: This was not initially supposed to be generic since it was loading the light batch, pruning and offloading back. I can do the refactor and merge in the upcoming PR for 34/35 support

ArnavBalyan: Just saw the CI failed, let me make this change here

ArnavBalyan: Updated this

ArnavBalyan:
cc @jinchengchenghh @zhztheplayer could you please take a look, have addressed all comments and UTs are now green thanks

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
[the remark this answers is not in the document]
jinchengchenghh: nextBatch make it a bit hard to understand for me.
Could you try this while `fetchNext` returns ColumnarBatch?
```
override def hasNext: Boolean = {
        rowsCollected < limit || partitionIter.hasNext
      }

      override def next(): ColumnarBatch = {
        if (!hasNext) {
          throw new NoSuchElementException("No more batches available.")
        }
        fetchNext()
      }
```

ArnavBalyan:
cc @zhztheplayer @jinchengchenghh, gentle reminder could you please take a look. All comments are addressed, UTs are now green

zhztheplayer:

ArnavBalyan:
> Run Gluten Clickhouse CI on x86

Thanks for re-run, investigating a failure here

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
```diff
@@ -123,4 +123,31 @@ public static ColumnarBatch compose(ColumnarBatch... batches) {
...
+   *
+   * <p>If `limit >= batch.numRows()`, returns the original batch. Otherwise, copies up to `limit`
+   * rows into new column vectors.
+   *
+   * @param batch the original batch
+   * @param limit the maximum number of rows to include
+   * @return a new pruned [[ColumnarBatch]] with row count = `limit`, or the original batch if no
+   *     pruning is required
+   */
+  public static ColumnarBatch pruneBatch(ColumnarBatch batch, int limit) {
```
jinchengchenghh, line 138: Also rename to slice, thanks!

shims/spark34/src/main/scala/org/apache/gluten/sql/shims/spark34/Spark34Shims.scala
```diff
@@ -530,4 +530,7 @@ class Spark34Shims extends SparkShims {
         false
     }
   }
+
+  override def isColumnarLimitExecSupported(): Boolean = false
```
jinchengchenghh, line 534: Can we change to getColumnarLimitExecOffset and fallback when offset is not 0 or limit < 0, then we would not need to change these codes when you support offset and we could also support part of CollectLimitExec operator in higher Spark version.

jinchengchenghh: Moreover, I would suggest to support offset in this PR, it would not expect too much efforts.

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
ArnavBalyan: updated this!

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatchJniWrapper.java
```diff
@@ -34,6 +34,8 @@ public static VeloxColumnarBatchJniWrapper create(Runtime runtime) {
 
   public native long compose(long[] batches);
 
+  public native long sliceBatch(long veloxBatchHandle, int limit);
```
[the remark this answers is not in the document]
ArnavBalyan: Is the `vector_size_t` also accessible in Java? We may have to continue using long here, and can probably cast to vector_size_t in the cc code. Updated to slice thanks

shims/spark34/src/main/scala/org/apache/gluten/sql/shims/spark34/Spark34Shims.scala
ArnavBalyan: Had intentionally not introduced the notion of offset, since Spark 32/33 don't support it. Else it maybe misleading to accept offset but not support it. Just wanted to confirm before updating this. Please let me know
We need this for our Spark 33 workloads falling back, there is some effort needed for offset support which I'll be able to add in the upcoming PR and can support Spark 34 and 35. Would that work? thanks! cc @FelixYBW

ArnavBalyan:
cc @jinchengchenghh @zhztheplayer gentle reminder can you please take a look

ArnavBalyan:
@zhztheplayer @jinchengchenghh just a gentle reminder, can you please review the PR

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatchJniWrapper.java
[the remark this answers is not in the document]
jinchengchenghh: vector_size_t is a alias in Velox, `using vector_size_t = int32_t;`, we could use int in Java instead.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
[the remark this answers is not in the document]
jinchengchenghh: Please resolve this comment, thanks!

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatchJniWrapper.java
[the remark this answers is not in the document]
jinchengchenghh: I mean you miss the argument `int offset, int length`, current implementation is not same with slice concept, other one may need to slice from not zero.

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
```diff
@@ -123,4 +123,31 @@ public static ColumnarBatch compose(ColumnarBatch... batches) {
...
+   *
+   * <p>If `limit >= batch.numRows()`, returns the original batch. Otherwise, copies up to `limit`
+   * rows into new column vectors.
+   *
+   * @param batch the original batch
+   * @param limit the maximum number of rows to include
+   * @return a new pruned [[ColumnarBatch]] with row count = `limit`, or the original batch if no
+   *     pruning is required
+   */
+  public static ColumnarBatch sliceBatch(ColumnarBatch batch, int limit) {
```
jinchengchenghh, line 138: Rename to `slice` and also add offset argument

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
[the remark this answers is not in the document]
ArnavBalyan: I had updated the PR with it, but there were some UT failures, there seems to be some case where hasnext is called repeatedly causing some failure. Current code pre-fetches it and keeps it ready for the next call

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
ArnavBalyan: Updated thanks

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatchJniWrapper.java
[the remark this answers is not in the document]
ArnavBalyan: Ah I see, updated it

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
```diff
@@ -0,0 +1,153 @@
...
+        val currentBatchRowCount = currentBatch.numRows()
+        val remaining = limit - rowsCollected
+
+        ColumnarBatches.retain(currentBatch)
```
jinchengchenghh, line 83: When we retain currentBatch, the last operator will not close it, but the next operator only closes the sliced batch, the original parent batch is not closed in time, I suppose it is closed after the query ends up. Do we need to add a `SliceBatch` like `ComposeBatch` to maintain the same lifecycle of batches? Then the next operator closes both of them. @zhztheplayer

jinchengchenghh: The sliced batch is a view of input batch without no copy.

zhztheplayer: > The sliced batch is a view of input batch without no copy.

Doesn't Velox handle the referencing internally? I assume it's easy in C++ with shared pointers.

https://github.com/facebookincubator/velox/blob/48d21bfd74be313d54b6cd33530525575ac44717/velox/vector/FlatVector-inl.h#L417-L438

What happens if we don't retain the batch?

jinchengchenghh: It returns a BufferView, https://github.com/facebookincubator/velox/blob/48d21bfd74be313d54b6cd33530525575ac44717/velox/buffer/Buffer.h#L181
So the original Buffer value should be keepped.

jinchengchenghh: Thanks for your investigation, the sliced BufferView will hold the shared pointer of value Buffer, so we don't need to retain the original Batch.

jinchengchenghh: Could you please update the `retain` and rebase? Then we can merge it. @ArnavBalyan

jinchengchenghh (approved):

PR #8575, opened 2025-01-20
[VL] Remove override of test in GlutenDynamicPartitionPruningSuite

acvictor:

This PR removes the override of `SPARK-32659: Fix the data issue when pruning DPP on non-atomic type` in `GlutenDynamicPartitionPruningSuite` since it succeeds as is.

## How was this patch tested?
CI run.


zhli1142015 (approved):
LGTM, thanks

PR #8579, opened 2025-01-21
[VL] CMake configuration cleanup to remove variable VELOX_COMPONENTS_PATH

zhztheplayer:
`VELOX_COMPONENTS_PATH` should be removed and the usages should be replaced by `$VELOX_BUILD_PATH/velox` already. This line (for `facebook::velox::exec_test_lib`) was likely missing from a former cleanup. The path does a follow-up on this line.


Yohahaha (approved):

zhztheplayer:
Thanks for the review!

Issue #8580, opened 2025-01-21
[Core] Don't report  'Not supported to map spark function name to substrait function name: input_file_name(), class name: InputFileName.'


baibaichen:
### Problem description

Both ch and velox support `InputFileName`, `InputFileBlockStart` and `InputFileBlockLength`, 

#### Issue 1 We still get the following warning: 

```shell
WARN org.apache.gluten.execution.ProjectExecTransformer: Validation failed with exception for plan: ProjectExecTransformer, due to: Not supported to map spark function name to substrait function name: input_file_name(), class name: InputFileName.
```

####  Issue 2 Unnessary codes:

spark supports `InputXXX` through `ThreadLocal`, while Gluten supports them through push down. IIUC, the following codes colud be removed, after this [pr](https://github.com/apache/incubator-gluten/pull/7188) and https://github.com/apache/incubator-gluten/pull/7124.

```
GlutenWholeStageColumnarRDD::compute()
    // To support input_file_name(). According to semantic we should return
    // the exact file name a row belongs to. However in columnar engine it's
    // not easy to accomplish this. so we return a list of file(part) names
    split match {
      case FirstZippedPartitionsPartition(_, g: GlutenPartition, _) =>
        InputFileBlockHolderProxy.set(g.files.mkString(","))
      case _ =>
        InputFileBlockHolderProxy.unset()
    }
```
####  Issue 3  Impreove Code quality:
Can PushDownInputFileExpression be optimized without going to CollapseProject later?

### System information

no

### CMake log

```bash

```

zml1206:
CollapseProject in advance needs to rely on whether the child is offloaded. I don't recommend doing this, as it will make offload complicated. Offload single is easier to maintain. @baibaichen

PR #8584, opened 2025-01-21
[GLUTEN-8580][CORE][Part-1] Clean up unnecessary code related to input file expression

zml1206:

(Fixes: \#8580)


FelixYBW (approved):

zhztheplayer:
The change looks good to me. @baibaichen Can you also take a look?

gluten-substrait/src/main/scala/org/apache/gluten/execution/GlutenWholeStageColumnarRDD.scala
```diff
@@ -62,17 +59,6 @@ class GlutenWholeStageColumnarRDD(
...
-    // To support input_file_name(). According to semantic we should return
-    // the exact file name a row belongs to. However in columnar engine it's
-    // not easy to accomplish this. so we return a list of file(part) names
-    split match {
-      case FirstZippedPartitionsPartition(_, g: GlutenPartition, _) =>
-        InputFileBlockHolderProxy.set(g.files.mkString(","))
-      case _ =>
-        InputFileBlockHolderProxy.unset()
-    }
-
```
Yohahaha, line 75: not sure if we can delete these codes. @gaoyangxiaozhu

zml1206: There are problems with the previous input file expression implementation. #7124 optimizes the solution and pushes the input file expression down to scanTransform or the project before scan. The results come from native scan or spark thread local, so there is no need to retain the information in InputFileBlockHolder.

zml1206:
Thanks for review, are there any other comments? If not, it will be merged later. @Yohahaha @baibaichen

Issue #8479, opened 2025-01-09
 [CORE] Split the configuration into individual modules that belong to each other.


yikf:
### Description

subtask of https://github.com/apache/incubator-gluten/issues/8326

PR #8586, opened 2025-01-21
[GLUTEN-8479][CORE][Part-3] Split backend configs to its corresponding modules

yikf:

Fix https://github.com/apache/incubator-gluten/issues/8479, This PR aims to split the backend configurations into individual modules that belong to each other.

This PR mainly accomplishes two things: 
1. It splits the configurations defined in GlutenConfig, moving Velox-related configurations to VeloxConfig and CH-related configurations to CHConf. 
3. It deletes the configuration retrieval methods that are not in use.

## How was this patch tested?

GA.

### DIFF
--- a/backends-clickhouse/src-celeborn/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
+++ b/backends-clickhouse/src-celeborn/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
@@ -16,7 +16,7 @@
  */
 package org.apache.spark.shuffle
 
-import org.apache.gluten.backendsapi.clickhouse.CHBackendSettings
+import org.apache.gluten.backendsapi.clickhouse.{CHBackendSettings, CHConfig}
 import org.apache.gluten.execution.ColumnarNativeIterator
 import org.apache.gluten.memory.CHThreadGroup
 import org.apache.gluten.vectorized._
@@ -29,7 +29,6 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
 import org.apache.celeborn.client.ShuffleClient
 import org.apache.celeborn.common.CelebornConf
 import org.apache.celeborn.common.protocol.ShuffleMode
-import org.apache.gluten.config.GlutenConfig
 
 import java.io.IOException
 import java.util.Locale
@@ -80,10 +79,10 @@ class CHCelebornColumnarShuffleWriter[K, V](
       nativeBufferSize,
       capitalizedCompressionCodec,
       compressionLevel,
-      GlutenConfig.get.chColumnarShuffleSpillThreshold,
+      CHConfig.get.chColumnarShuffleSpillThreshold,
       CHBackendSettings.shuffleHashAlgorithm,
       celebornPartitionPusher,
-      GlutenConfig.get.chColumnarForceMemorySortShuffle
+      CHConfig.get.chColumnarForceMemorySortShuffle
         || ShuffleMode.SORT.name.equalsIgnoreCase(shuffleWriterType)
     )
 
--- a/backends-clickhouse/src-delta-20/main/scala/org/apache/spark/sql/delta/ClickhouseOptimisticTransaction.scala
+++ b/backends-clickhouse/src-delta-20/main/scala/org/apache/spark/sql/delta/ClickhouseOptimisticTransaction.scala
@@ -16,7 +16,7 @@
  */
 package org.apache.spark.sql.delta
 
-import org.apache.gluten.backendsapi.clickhouse.CHConf
+import org.apache.gluten.backendsapi.clickhouse.CHConfig
 
 import org.apache.spark.SparkException
 import org.apache.spark.sql.Dataset
@@ -110,7 +110,7 @@ class ClickhouseOptimisticTransaction(
         spark.conf.getAll.foreach(
           entry => {
             if (
-              CHConf.startWithSettingsPrefix(entry._1)
+              CHConfig.startWithSettingsPrefix(entry._1)
               || entry._1.equalsIgnoreCase(DeltaSQLConf.DELTA_OPTIMIZE_MIN_FILE_SIZE.key)
             ) {
               options += (entry._1 -> entry._2)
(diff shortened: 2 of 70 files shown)


yikf:
@jackylee-ch @baibaichen Would you please take a look, thanks!

jackylee-ch:
It seems that the current PR is about splitting config with `backend` rather than `module`. For this pr, `.backend()` should handle this case?

yikf:
@jackylee-ch Move the configuration of a specific backend to its own module. The backendType is used to identify the configuration type and has nothing to do with where the configuration is located. It's not certain whether dividing into multiple configuration classes makes sense.

jackylee-ch:
> @jackylee-ch Move the configuration of a specific backend to its own module. The backendType is used to identify the configuration type and has nothing to do with where the configuration is located. It's not certain whether dividing into multiple configuration classes makes sense.

Prefer putting the configs in GlutenConfigs and using `backend()` or `module()` to seperate the configs, it is not easy to maintain these configurations separately. cc @zhztheplayer @PHILO-HE

zhztheplayer:
I am inclined to placing them in separated modules. But also would like to hear more views here.

yikf:
@jackylee-ch @zhztheplayer If separated, the final state seems quite good. The backend-related configurations will only be used by something like VeloxBackendApi, making it clearer. I tend to separate and see if anyone else has any ideas.

jackylee-ch:
BTW, Is our goal to split the configuration of all modules or just the backends? I only see backend config splitting in this PR.

backends-clickhouse/src-celeborn/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
```diff
@@ -16,16 +16,14 @@
  */
 package org.apache.spark.shuffle
 
-import org.apache.gluten.backendsapi.clickhouse.CHBackendSettings
+import org.apache.gluten.backendsapi.clickhouse.{CHBackendSettings, CHConf}
 import org.apache.gluten.execution.ColumnarNativeIterator
 import org.apache.gluten.memory.CHThreadGroup
 import org.apache.gluten.vectorized._
-
```
jackylee-ch, line 23: this line should not be removed

backends-clickhouse/src-celeborn/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
```diff
@@ -16,16 +16,14 @@
...
 import org.apache.spark._
 import org.apache.spark.scheduler.MapStatus
 import org.apache.spark.shuffle.celeborn.CelebornShuffleHandle
 import org.apache.spark.sql.vectorized.ColumnarBatch
-
```
jackylee-ch, line 28: ditto

backends-clickhouse/src-celeborn/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
```diff
@@ -80,10 +79,10 @@ class CHCelebornColumnarShuffleWriter[K, V](
       nativeBufferSize,
       capitalizedCompressionCodec,
       compressionLevel,
-      GlutenConfig.get.chColumnarShuffleSpillThreshold,
```
PHILO-HE, line 83: Can we keep the use of `GlutenConfig.get` after the refactor? I prefer a uniform API. It requires either velox conf or CH conf to be covered by GlutenConfig at runtime somehow.

PHILO-HE:
@baibaichen, please take a look.

backends-clickhouse/src-celeborn/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
yikf: So that is to say, the definition of the configuration is independent, but all the acquisitions are from GlutenConfig?

zhztheplayer: Perhaps we can make `VeloxConfig` inherit from `GlutenConfig`? Then `VeloxConfig` can be used uniformly in Velox backend for both Velox / Common configurations.

yikf: The current implementation is an inheritance relationship.

zhztheplayer: Aha, then I feel it's already convenient to developers.

jackylee-ch:
@yikf There are some code error for ch backend, could you fix it?

backends-clickhouse/src-celeborn/main/scala/org/apache/spark/shuffle/CHCelebornColumnarShuffleWriter.scala
PHILO-HE: > So that is to say, the definition of the configuration is independent, but all the acquisitions are from GlutenConfig?

@yikf, yes, it's not a mature idea. Maybe, firstly keep using your current changes.

baibaichen:
@CodiumAI-Agent /review

yikf:
@baibaichen @zhztheplayer The potential issues reported by CodiumAI-Agent have been confirmed. These are existing validation and can be ignored. If the manual review finds no performance or functional problems, we can merge it.

yikf:
> @yikf, one minor comment. Better to rename CHConf to CHConfig for consistency with GlutenConfig, VeloxConfig. cc @baibaichen
> 
> Will merge the pr today if there is no other comment from the community. Thank you!

+1 for rename.

PHILO-HE (approved):
@yikf, thank you so much for your continuous efforts! Will merge the patch soon.

PR #8593, opened 2025-01-22
[VL] Support casting integral type to timestamp type

PHILO-HE:

There is a velox pr merged: https://github.com/facebookincubator/velox/pull/11089. This pr fixes the fallback even though velox has supported such cast.

## How was this patch tested?

Existing Spark tests.


cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -269,12 +269,21 @@ bool SubstraitToVeloxPlanValidator::validateCast(
...
 
-  // Only support cast from date to timestamp
-  if (toType->kind() == TypeKind::TIMESTAMP && !input->type()->isDate()) {
-    LOG_VALIDATION_MSG(
-        "Casting from " + input->type()->toString() + " to " + toType->toString() + " is not supported.");
-    return false;
-  }
+  // Only support cast from date or integral type to timestamp.
+  if (toType->kind() == TypeKind::TIMESTAMP)
+    if (!input->type()->isDate()) {
```
marin-ma, line 274: We can still use one `if` here, same as the original one `if (toType->kind() == TypeKind::TIMESTAMP && !input->type()->isDate())`

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetColumnIndexSuite.scala
```diff
@@ -38,7 +38,25 @@ class GlutenParquetColumnIndexSuite extends ParquetColumnIndexSuite with GlutenS
         "cast(id as float) as _5",
         "cast(id as double) as _6",
         "cast(id as decimal(20,0)) as _7",
-        "cast(cast(1618161925000 + id * 1000 * 60 * 60 * 24 as timestamp) as date) as _9"
+        "cast(cast(1618161925 + id * 60 * 60 * 24 as timestamp) as date) as _9"
```
marin-ma, line 41: Why do we need this change here? Is it because larger timestamp will overflow in Velox? Could add some comments like "// Change 1618161925000 to 1618161925 because ... " for explanations?

marin-ma:
Should we also update the cast functions support status? Seems like the current document is also outdated. Casting from date to timestamp is already supported. https://github.com/apache/incubator-gluten/blob/main/docs/velox-backend-support-progress.md#cast-functions-support-status

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetColumnIndexSuite.scala
PHILO-HE: @marin-ma, it's for avoiding reaching the limitation of Velox. Any timestamp past 2037-11-01 09:00:00 is not supported by Velox.

PHILO-HE:
> Should we also update the cast functions support status? Seems like the current document is also outdated. Casting from date to timestamp is already supported. https://github.com/apache/incubator-gluten/blob/main/docs/velox-backend-support-progress.md#cast-functions-support-status

@marin-ma, seems detailed support state for cast function has been removed. Do we need to add it back?

PHILO-HE:
@marin-ma, sorry for delayed update. Could you take a look again? Thanks!

JkSelf (approved):
LGTM. Thanks.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -263,8 +263,16 @@ bool SubstraitToVeloxPlanValidator::isAllowedCast(const TypePtr& fromType, const
 
   // Limited support for X to Timestamp.
   if (toType->isTimestamp() && !(fromType->isDate() || fromType->isVarchar())) {
-    LOG_VALIDATION_MSG("Casting from " + fromType->toString() + " to TIMESTAMP is not supported.");
-    return false;
+    switch (fromType->kind()) {
+      case TypeKind::TINYINT:
```
jackylee-ch, line 267: can we move `fromType->isDate() || fromType->isVarchar()` to here?

PHILO-HE: @jackylee-ch, just moved and also use isXXX() to replace TypeKind check.

PR #8602, opened 2025-01-23
[GLUTEN-8565][VL] Remove unused code in velox batches

ArnavBalyan:
 - Removing unused code in VeloxColumnarBatch.
 - checkNonVeloxBatch is not called anywhere, removing the same

## How was this patch tested?

Unit Tests

### DIFF
--- a/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
+++ b/backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
@@ -47,15 +47,6 @@ public final class VeloxColumnarBatches {
             COMPREHENSIVE_TYPE_VELOX, ColumnarBatches.getComprehensiveLightBatchType(batch)));
   }
 
-  public static void checkNonVeloxBatch(ColumnarBatch batch) {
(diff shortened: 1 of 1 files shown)


ArnavBalyan:
cc @zhztheplayer could you ptal thanks!

backends-velox/src/main/java/org/apache/gluten/columnarbatch/VeloxColumnarBatches.java
```diff
@@ -46,16 +46,7 @@ public static void checkVeloxBatch(ColumnarBatch batch) {
...
-  public static void checkNonVeloxBatch(ColumnarBatch batch) {
-    if (ColumnarBatches.isZeroColumnBatch(batch)) {
-      return;
-    }
-    Preconditions.checkArgument(
-        !isVeloxBatch(batch),
-        String.format("Comprehensive batch type is already %s", COMPREHENSIVE_TYPE_VELOX));
-  }
-
+
```
jackylee-ch, line 49: Why we have this change for this new line? it has spaces?

ArnavBalyan: Yes it seems to have crept in, updated

FelixYBW (approved):

PR #7789, opened 2024-11-03
[GLUTEN-7706][CORE] Support Spark-344 + JDK17

zhouyuan:
## What changes were proposed in this pull request?

Support Spark-344 + JDK17

## How was this patch tested?

pass GHA tests

PR #8605, opened 2025-01-24
[VL][CI] Use a unified docker image for Spark tests

PHILO-HE:

In #7789, Spark-3.4.4 support is added and JDK-17 docker is used to avoid some issues caused by arrow. This pr changes to use the unified docker image and switches JDK version before running the test, which can reduce the efforts a bit for maintaining docker images.

I cannot figure out the reason why this way didn't pass in previous test. Maybe, we can land this pr firstly and continue observing the CI feedback while still keeping the JDK-17 docker image available in docker hub for some time.

## How was this patch tested?

CI.

### DIFF
--- a/.github/workflows/velox_backend.yml
+++ b/.github/workflows/velox_backend.yml
@@ -750,7 +750,7 @@ jobs:
   run-spark-test-spark34:
     needs: build-native-lib-centos-7
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


zhouyuan (approved):
+1
I remember there is some in-compatible arrow class reported previously. If all tests are passing now so should be all good

Issue #8306, opened 2024-12-23
[VL] Fallback happens if child node of GetStructField is Scalar Function Node


ayushi-agarwal:
### Description

Enhance https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala#L54 to add support for child node type as Scalar Function Node:

Example:
24/12/23 10:00:23 WARN GlutenFallbackReporter: Validation failed for plan: Filter[QueryId=4], due to: Unsupported child expression of GetStructField: from_json(StructField(platformId,StringType,true), json_column#7, Some(America/Los_Angeles)).platformId

More reference here:
https://github.com/apache/incubator-gluten/issues/4039#issuecomment-2558892038

rui-mo:
@ayushi-agarwal Thanks for reporting this issue! I understand the case now and will spare some time to investigate this or next week.

WangGuangxin:
@rui-mo we need to add a GetStructField scalar function in velox side. I would like to work on this if you have not started yet

rui-mo:
@WangGuangxin We probably don't need a scalar function implementation in Velox as it already provides 'FieldAccessTypedExpr' and 'DereferenceTypedExpr' for this purpose. I assume we need to investigate how to create such expression when the child is scalar function.

https://github.com/facebookincubator/velox/blob/d97bae061265e78194cde11f4132714f7c01e44d/velox/core/Expressions.h#L292-L298

https://github.com/facebookincubator/velox/blob/d97bae061265e78194cde11f4132714f7c01e44d/velox/core/Expressions.h#L393-L395

WangGuangxin:
> when

@rui-mo We also encountered this fallback in our production env, what we do here at first is `PreProject` if there is `GetStructFiled` to make sure the input is always an attribute.

rui-mo:
> what we do here at first is PreProject if there is GetStructFiled to make sure the input is always an attribute.

Hi @WangGuangxin, thanks for sharing this information! I would like to know if you have encountered any problems with this approach that have made you consider replacing it with a scalar function.

WangGuangxin:
> > what we do here at first is PreProject if there is GetStructFiled to make sure the input is always an attribute.
> 
> Hi @WangGuangxin, thanks for sharing this information! I would like to know if you have encountered any problems with this approach that have made you consider replacing it with a scalar function.

[...]

PR #8606, opened 2025-01-24
[GLUTEN-8306][VL] Support GetStructField with scalar function as input

WangGuangxin:

Support GetStructField with scalar function as input

(Fixes: \#8306)

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala
@@ -16,7 +16,6 @@
  */
 package org.apache.gluten.expression
 
-import org.apache.gluten.exception.GlutenNotSupportException
 import org.apache.gluten.expression.ConverterUtils.FunctionConfig
 import org.apache.gluten.expression.ExpressionConverter.replaceWithExpressionTransformer
 import org.apache.gluten.substrait.`type`.StructNode
@@ -56,7 +55,9 @@ case class VeloxGetStructFieldTransformer(
     child: ExpressionTransformer,
     ordinal: Int,
     original: GetStructField)
-  extends UnaryExpressionTransformer {
+  extends BinaryExpressionTransformer {
+  override def left: ExpressionTransformer = child
+  override def right: ExpressionTransformer = LiteralTransformer(ordinal)
   override def doTransform(args: Object): ExpressionNode = {
     val childNode = child.doTransform(args)
     childNode match {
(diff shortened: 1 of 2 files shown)


WangGuangxin:
@rui-mo @jinchengchenghh @PHILO-HE Can you please help review this?

jackylee-ch:
There is a UT failed, have you checked the failed reason?

WangGuangxin:
> There is a UT failed, have you checked the failed reason?

I'v checked it's not related to my change. But seems I cann't re-trigger it without code change

backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
```diff
@@ -1990,4 +1990,33 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
...
+                | select
+                |  from_json(json_str, 'a INT, b ARRAY<INT>').a,
+                |  from_json(json_str, 'a INT, b ARRAY<INT>').b
+                | from t
+                |""".stripMargin
+
+            runQueryAndCompare(query)(
+              df => {
+                val executedPlan = getExecutedPlan(df)
+                assert(executedPlan.count(_.isInstanceOf[ProjectExec]) == 0)
```
rui-mo, line 2016: nit: perhaps also assert the ProjectExecTransformer.

jackylee-ch (approved):
👍

backends-velox/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala
```diff
@@ -56,7 +55,9 @@ case class VeloxGetStructFieldTransformer(
     child: ExpressionTransformer,
     ordinal: Int,
     original: GetStructField)
-  extends UnaryExpressionTransformer {
+  extends BinaryExpressionTransformer {
```
PHILO-HE, line 58: Could you explain why we need this change?

WangGuangxin: because we need to transform both `child` and `ordinal`, so the children size should be two.  Before this change, it doesn't need to transform `ordinal`.

PHILO-HE (approved):

backends-velox/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala
```diff
@@ -70,8 +71,7 @@ case class VeloxGetStructFieldTransformer(
           node.getTypeNode.asInstanceOf[StructNode].getFieldTypes.get(ordinal)
         ExpressionBuilder.makeNullLiteral(nodeType)
       case _ =>
-        throw new GlutenNotSupportException(
-          s"Unsupported child expression of GetStructField: $original.")
+        super.doTransform(args)
```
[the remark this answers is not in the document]
rui-mo: https://github.com/facebookincubator/velox/pull/12166

PR #8610, opened 2025-01-24
[GLUTEN-8609][VL] Remove force cleanup in build_velox.sh

marin-ma:


PHILO-HE (approved):
As far as I can recall, this line was for cleaning up code changes made by modify_velox.patch. Since this patch has been removed, I think we can land this pr.

PR #8615, opened 2025-01-25
[VL] Update document, remove the experimental word for spark.gluten.enabled

FelixYBW:
spark.gluten.enabled is stable for long time. It can enable/disable gluten at run time. With it disable, all query plans fallback to vanilla spark. Compare to plugin setting, the config may have a little overhead when generating query plan.


PHILO-HE (approved):

docs/Configuration.md
```diff
@@ -21,7 +21,7 @@ You can add these configurations into spark-defaults.conf to enable or disable t
...
+| spark.gluten.enabled                                         | Enable Gluten at runtime, default is true. It fallbacks to vanilla Spark for all query plans. Recommend to enable/disable Gluten through the setting for `spark.plugins`.                                                                                                                                                                                                                                                                                                                                                                                                | true                                                 |
```
marin-ma, line 24: To be clearer "It fallbacks to vanilla Spark for all query plans if set to false."

liujp:
I think the config `spark.gluten.enabled` is useful in runtime, for example, in the same sparkSession, we can disable gluten dynamically in runtime

Issue #8616, opened 2025-01-25
[VL] Support Existence Join for BroadcastNestedLoopJoin


ArnavBalyan:
### Description

 - Existence join is not supported for BNLJ
 - Can be mapped to left semi project cross join

PR #8617, opened 2025-01-25
[GLUTEN-8616] [VL] Add support for Existence Join in broadcast nested loop join

ArnavBalyan:

 - Existence join falls back for broadcast nested loop join.
 - Added support for the same in Velox as leftSemiProject in the cross join/nested loop join.
 - Enabled existence join for conversion.

(Fixes: \#8616)

## How was this patch tested?
 - Unit Tests


ArnavBalyan:
Will re-open once upstream is available in Gluten

PR #8621, opened 2025-01-26
[GLUTEN-8616] [VL] Make filescan limit for encrypted fallback as configurable

ArnavBalyan:
 - Making parquet file scan limit as configurable.
 - Currently it's hardcoded to default 10, now available as a user config can be changed externally.
 - Default remains the same as before of 10.

## How was this patch tested?
 - Existing UTs

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
@@ -503,6 +503,8 @@ class GlutenConfig(conf: SQLConf) extends Logging {
 
   def autoAdjustStageFallenNodeThreshold: Double =
     getConf(AUTO_ADJUST_STAGE_RESOURCES_FALLEN_NODE_RATIO_THRESHOLD)
+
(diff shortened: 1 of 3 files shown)


ArnavBalyan:
cc @zhztheplayer could you please review thanks!

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -2310,4 +2312,13 @@ object GlutenConfig {
...
       .doubleConf
       .createWithDefault(0.5d)
+
+  val ENCRYPTED_PARQUET_FALLBACK_FILE_LIMIT =
+    buildConf("spark.gluten.sql.fallbackEncryptedParquet.limit")
+      .internal()
+      .doc("If supplied, `limit` number of files will be checked to determine encryption " +
+        "and falling back java scan")
+      .intConf
+      .createWithDefault(10)
```
jackylee-ch, line 2322: nit: `checkValue` is required here to ensure the data is non-negative.

ArnavBalyan: done

ArnavBalyan:
cc @jackylee-ch fixed, can you please take a look

jackylee-ch (approved):

PR #8628, opened 2025-01-27
[GLUTEN-8627][VL] Fix cpp build and build script on MacOS

marin-ma:
Fix the cpp compilation error and make builddeps-veloxbe.sh work on MacOS.

Manually tested with `./dev/builddeps-veloxbe.sh --run_setup_script=ON --build_arrow=ON --build_type=Debug --build_tests=ON --build_examples=ON` and also tested with switching OFF the build options and Release build type.

Note `--build_benchmarks=ON` still fails because the function to set CPU affinity on macOS is not supported. I will fix that in a separate patch.


marin-ma:
@FelixYBW @zhouyuan Could you help to review? Thanks!

FelixYBW (approved):

zhouyuan (approved):
👍 
All CI green

Issue #8627, opened 2025-01-27
[VL] Fix cpp build and build script on MacOS


marin-ma:
### Backend

VL (Velox)

### Bug description

Fix the cpp compilation error and make `builddeps-veloxbe.sh` work on MacOS.

### Spark version

None

### Relevant logs

```bash

```

PR #8654, opened 2025-02-03
[VL][MIRROR] Fix build faile don Macos with INSTALL_PREFIX not set

jackylee-ch:
After #8627 , I met the `dev/builddeps-veloxbe.sh: line 218: INSTALL_PREFIX: unbound variable`.

## How was this patch tested?
manual test on macos


marin-ma (approved):
Thanks for the fix!

jackylee-ch:
Thanks for your review @marin-ma

PR #8674, opened 2025-02-06
[VL] Enable cast when input type equal to dest type.

j7nhai:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -269,6 +269,12 @@ bool SubstraitToVeloxPlanValidator::validateCast(
   const auto& toType = SubstraitParser::parseType(castExpr.type());
   core::TypedExprPtr input = exprConverter_->toVeloxExpr(castExpr.input(), inputType);
 
+  if (toType->kind() == input->type()->kind()) {
+    if (toType->isTimestamp()) {
+      return true;
+    }
+  }
+
   // Only support cast from date to timestamp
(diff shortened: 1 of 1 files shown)


j7nhai:
> This case shouldn't happen. See https://github.com/apache/incubator-gluten/blob/main/docs/velox-backend-support-progress.md#cast-functions-support-status
> 
> > N/A: not applicable case, e.g., from type is as same as to type, where cast will not actually happen.
> 
> Could you provide a sample to demonstrate when this code path will be called?

```scala
sql("use test")
sql("drop table if exists t1")
sql("drop table if exists t2")

sql("create table t1 (a TIMESTAMP) USING ICEBERG")
sql("create table t2 (b TIMESTAMP) USING ICEBERG")

sql("INSERT OVERWRITE t1 VALUES (CAST('2025-01-21 00:00:00' AS TIMESTAMP))")
sql("INSERT OVERWRITE t1 SELECT * FROM t2")  // Not supported to map spark function name to substrait function name: cast(b#7 as timestamp), class name: AnsiCast.

```

Then I try to convert AnsiCast to cast when origin type equal to dest type. And the native cast not support ts to ts.

PHILO-HE:
@j7nhai, does this issue only happen for iceberg? I found using parquet has no issue.

Spark has an optimizer that removes the same type cast. Not sure why it is not applied in your case.
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/optimizer/expressions.scala#L1091

PR #8677, opened 2025-02-06
[VL] nit: Remove shadowed variables in SubstraitToVeloxPlan.cc

zhztheplayer:
`gluten::kHiveConnectorId` was defined as const so no need to define `kHiveConnectorId` locally in the function bodies.


marin-ma (approved):
Thanks

Issue #8678, opened 2025-02-06
[Core] Gluten jar contains unresolved variable name when building on macOS


marin-ma:
### Backend

VL (Velox)

### Bug description

After running maven command `mvn clean package -Pbackends-velox -Pceleborn -Puniffle -Pspark-3.4 -DskipTests`, I got a jar with unresolved variable in its name under `package/target`:
gluten-velox-bundle-spark3.4_2.12-${os.detected.release}_${os.detected.release.version}_aarch_64-1.4.0-SNAPSHOT.jar

`os.detected.*` are variables from os-maven-plugin which are only valid on linux platform: https://github.com/trustin/os-maven-plugin/?tab=readme-ov-file#property-osdetectedrelease-linux-only

### Spark version

None

### Relevant logs

```bash

```

PR #4529, opened 2024-01-26
[CORE] Add arch suffix for package jar

Yohahaha:
## What changes were proposed in this pull request?

Add arch suffix for package jar and thridparty-lib.jar

See https://github.com/trustin/os-maven-plugin/blob/master/README.md#property-osdetectedarch

PR #8679, opened 2025-02-06
[GLUTEN-8678] Fix jar name on macos

marin-ma:
After this fix, the jar name on macOS will be like:
gluten-velox-bundle-spark3.5_2.12-darwin_aarch64-1.4.0-SNAPSHOT.jar
gluten-velox-bundle-spark3.5_2.12-darwin_x86_64-1.4.0-SNAPSHOT.jar

Fixes: #8678

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,11 @@
     <project.prefix>spark-sql-columnar</project.prefix>
     <jar.assembly.name.prefix>gluten</jar.assembly.name.prefix>
     <os-maven-plugin.version>1.5.0.Final</os-maven-plugin.version>
+    <!--
+    Will be overwritten by the profile on different platforms.
+    Otherwise, it is built from an unsupported platform.
+    -->
+    <os.full.name>unknown</os.full.name>
     <!-- To build built-in backend c++ codes -->
     <scala.recompile.mode>all</scala.recompile.mode>
     <!-- For unit tests -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -978,6 +985,7 @@
       <properties>
         <platform>linux</platform>
         <arch>amd64</arch>
+        <os.full.name>${os.detected.release}_${os.detected.release.version}_${os.detected.arch}</os.full.name>
```
zhztheplayer, line 988: Shall we use `<os.full.name>${platform}_${arch}</os.full.name>` for linux as well?

Given https://github.com/apache/incubator-gluten/pull/6942 was merged later than https://github.com/apache/incubator-gluten/pull/4529

cc @Yohahaha

zhztheplayer:
(off-topic perhaps) I assume one purpose of https://github.com/apache/incubator-gluten/pull/6942 was to make one fat jar support multiple archs and platforms. So in future we might completely remove the arch / platform marks in the jar name? cc @wForget

pom.xml
Yohahaha: both ok to me.

zhztheplayer: @marin-ma Would you help make the change? Thanks!

marin-ma: @zhztheplayer Updated. Thanks!

PR #8682, opened 2025-02-07
[VL] Skip the velox download if velox_repo not accessable

FelixYBW:
When velox_home is specified and velox_repo:velox_branch doesn't exists, it means user prepared the customized Velox. We can skip the velox download and continue the build. It's user's responsibility to make sure the velox has correct version.

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -145,26 +145,36 @@ if [ "$VELOX_HOME" == "" ]; then
 fi
 VELOX_SOURCE_DIR="${VELOX_HOME}"
 
-# checkout code
-TARGET_BUILD_COMMIT="$(git ls-remote $VELOX_REPO $VELOX_BRANCH | awk '{print $1;}' | head -n 1)"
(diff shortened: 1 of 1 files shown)


ep/build-velox/src/get_velox.sh
```diff
@@ -150,18 +150,23 @@ TARGET_BUILD_COMMIT="$(git ls-remote $VELOX_REPO $VELOX_BRANCH | awk '{print $1;
...
+    git reset --hard HEAD
+    git checkout refs/tags/build_$TARGET_BUILD_COMMIT
   fi
-  git reset --hard HEAD
-  git checkout refs/tags/build_$TARGET_BUILD_COMMIT
 else
   git clone $VELOX_REPO -b $VELOX_BRANCH $VELOX_SOURCE_DIR
   cd $VELOX_SOURCE_DIR
   git checkout $TARGET_BUILD_COMMIT
 fi
```
zhouyuan, line 168: nit: better to print some warning message here?

zhouyuan (approved):
+1
works for me

PR #8683, opened 2025-02-07
[VL] Enable make_date function

zhli1142015:

Enable make_date function.

## How was this patch tested?
UT.


marin-ma (approved):
Thanks for enabling this function!

PR #8684, opened 2025-02-07
[VL] Should convert kSpillReadBufferSize and kShuffleSpillDiskWriteBufferSize to number

boneanxs:

Using 1.3.0, we could meet the error if manually set `spark.unsafe.sorter.spill.reader.buffer.size` to value like `2m`

```java
org.apache.gluten.exception.GlutenException: Non-whitespace character found after end of conversion: "m"
	at org.apache.gluten.vectorized.PlanEvaluatorJniWrapper.nativeCreateKernelWithIterator(Native Method)
	at org.apache.gluten.vectorized.NativePlanEvaluator.createKernelWithBatchIterator(NativePlanEvaluator.java:68)
	at org.apache.gluten.backendsapi.velox.VeloxIteratorApi.genFirstStageIterator(VeloxIteratorApi.scala:204)
	at org.apache.gluten.execution.GlutenWholeStageColumnarRDD.$anonfun$compute$1(GlutenWholeStageColumnarRDD.scala:88)
	at org.apache.gluten.utils.Arm$.withResource(Arm.scala:25)
	at org.apache.gluten.metrics.GlutenTimeMetric$.millis(GlutenTimeMetric.scala:37)
	at org.apache.gluten.execution.GlutenWholeStageColumnarRDD.compute(GlutenWholeStageColumnarRDD.scala:77)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:380)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:344)
	at org.apache.spark.rdd.UnionRDD.compute(UnionRDD.scala:106)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:380)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:344)
```
We should parse value to number in bytes before put to nativeConf

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
@@ -576,9 +576,7 @@ object GlutenConfig {
   val SPARK_REDACTION_REGEX = "spark.redaction.regex"
   val SPARK_SHUFFLE_FILE_BUFFER = "spark.shuffle.file.buffer"
   val SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE = "spark.unsafe.sorter.spill.reader.buffer.size"
-  val SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE_DEFAULT: Int = 1024 * 1024
   val SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE = "spark.shuffle.spill.diskWriteBufferSize"
-  val SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE_DEFAULT: Int = 1024 * 1024
   val SPARK_SHUFFLE_SPILL_COMPRESS = "spark.shuffle.spill.compress"
   val SPARK_SHUFFLE_SPILL_COMPRESS_DEFAULT: Boolean = true
 
@@ -650,16 +648,26 @@ object GlutenConfig {
       (
         GLUTEN_COLUMNAR_TO_ROW_MEM_THRESHOLD.key,
         GLUTEN_COLUMNAR_TO_ROW_MEM_THRESHOLD.defaultValue.get.toString),
-      (
-        SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE,
-        SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE_DEFAULT.toString),
-      (
-        SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE,
-        SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE_DEFAULT.toString),
       (SPARK_SHUFFLE_SPILL_COMPRESS, SPARK_SHUFFLE_SPILL_COMPRESS_DEFAULT.toString)
     )
     keyWithDefault.forEach(e => nativeConfMap.put(e._1, conf.getOrElse(e._1, e._2)))
 
+    conf
+      .get(SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE)
+      .foreach(
+        v =>
+          nativeConfMap
+            .put(
+              SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE,
+              JavaUtils.byteStringAs(v, ByteUnit.BYTE).toString))
+    conf
+      .get(SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE)
+      .foreach(
+        v =>
+          nativeConfMap
+            .put(
+              SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE,
+              JavaUtils.byteStringAs(v, ByteUnit.BYTE).toString))
     conf
       .get(SPARK_SHUFFLE_FILE_BUFFER)
       .foreach(


Yohahaha:
https://github.com/apache/incubator-gluten/blob/e0092089453286b6822447a875666b41ab6f0b5c/shims/common/src/main/scala/org/apache/spark/sql/internal/GlutenConfigUtil.scala#L21-L45

`spark.unsafe.sorter.spill.reader.buffer.size=2m` should be converted by above codes, could you investigate why it not works?

Yohahaha:
> https://github.com/apache/incubator-gluten/blob/e0092089453286b6822447a875666b41ab6f0b5c/shims/common/src/main/scala/org/apache/spark/sql/internal/GlutenConfigUtil.scala#L21-L45
> 
> `spark.unsafe.sorter.spill.reader.buffer.size=2m` should be converted by above codes, could you investigate why it not works?

oh, GlutenConfigUtil only process the config which prefix is 'spark.gluten'.

boneanxs:
> > https://github.com/apache/incubator-gluten/blob/e0092089453286b6822447a875666b41ab6f0b5c/shims/common/src/main/scala/org/apache/spark/sql/internal/GlutenConfigUtil.scala#L21-L45
> > 
> > `spark.unsafe.sorter.spill.reader.buffer.size=2m` should be converted by above codes, could you investigate why it not works?
> 
> oh, GlutenConfigUtil only process the config which prefix is 'spark.gluten'.

Do we need to extend this function to allow all configures?

Yohahaha:
> 
> Do we need to extend this function to allow all configures?

yeah, we may need add new method `GlutenConfigUtil#get(ConfigEntry)` to process non-sql configs.

boneanxs:
> > Do we need to extend this function to allow all configures?
> 
> yeah, we may need add new method `GlutenConfigUtil#get(ConfigEntry)` to process non-sql configs.

Hey @Yohahaha We cannot use `ConfigEntry` since they are private in Spark package only

https://github.com/apache/spark/blob/cea79dc1918b7f03870fe1cb189da9a152e3bbaf/core/src/main/scala/org/apache/spark/internal/config/package.scala#L1892-L1899

How about only extract a specific method that handle bytes value only to reduce duplicates?

Yohahaha:
> > > Do we need to extend this function to allow all configures?
> > 
> > 
> > yeah, we may need add new method `GlutenConfigUtil#get(ConfigEntry)` to process non-sql configs.
> 
> Hey @Yohahaha We cannot use `ConfigEntry` since they are private in Spark package only
> 
> https://github.com/apache/spark/blob/cea79dc1918b7f03870fe1cb189da9a152e3bbaf/core/src/main/scala/org/apache/spark/internal/config/package.scala#L1892-L1899
> 
> How about only extract a specific method that handle bytes value only to reduce duplicates?

sounds good to me.

boneanxs:
@Yohahaha Hey, comments addressed, could you review this again?

shims/common/src/test/scala/org/apache/spark/sql/internal/GlutenConfigUtilSuite.scala
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.internal
+
+import org.apache.spark.network.util.ByteUnit
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class GlutenConfigUtilSuite extends AnyFunSuite {
```
Yohahaha, line 23: maybe move this suite to `gluten-core` is better.

Yohahaha (approved):
LGTM! only one comments.

Yohahaha:
```
- csv scan with option string as null
/opt/gluten/ep/_ep/arrow_ep/java/dataset/src/main/cpp/jni_util.cc:79: Failed to update reservation while freeing bytes: Java Exception: java.lang.IncompatibleClassChangeError

/tmp/jnilib-1738162292075553983.tmp(+0x11b6338)[0x7f4bbada5338]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow4util8ArrowLogD1Ev+0xed)[0x7f4bbada576d]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow7dataset3jni31ReservationListenableMemoryPool4FreeEPhll+0x47d)[0x7f4bba41a58d]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow10PoolBufferD0Ev+0x47)[0x7f4bbb04a6e7]
/tmp/jnilib-1738162292075553983.tmp(_ZNSt23_Sp_counted_ptr_inplaceIN5arrow9ArrayDataESaIS1_ELN9__gnu_cxx12_Lock_policyE2EE10_M_disposeEv+0x10e)[0x7f4bba61b58e]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow17SimpleRecordBatchD1Ev+0x10e)[0x7f4bbaed394e]
/tmp/jnilib-1738162292075553983.tmp(_ZNSt16_Sp_counted_baseILN9__gnu_cxx12_Lock_policyE2EE10_M_releaseEv+0x3a)[0x7f4bba41146a]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow8internal6FnOnceIFvRKNS_10FutureImplEEE6FnImplINS_6FutureISt10shared_ptrINS_11RecordBatchEEE20WrapResultOnComplete8CallbackINSC_14ThenOnCompleteIZNS_23DefaultIfEmptyGeneratorISB_EclEvEUt_NSC_17PassthruOnFailureISI_EEEEEEED0Ev+0x2e)[0x7f4bba465ffe]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow18ConcreteFutureImpl22DoMarkFinishedOrFailedENS_11FutureStateE+0x1a4)[0x7f4bbadbf494]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow6FutureISt10shared_ptrINS_11RecordBatchEEE12MarkFinishedENS_6ResultIS3_EE+0x96)[0x7f4bba486526]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow8internal6FnOnceIFvRKNS_10FutureImplEEE6FnImplINS_6FutureISt10shared_ptrINS_11RecordBatchEEE20WrapResultOnComplete8CallbackINS_6detail16MarkNextFinishedISC_SC_Lb0ELb0EEEEEE6invokeES4_+0x55)[0x7f4bba486895]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow18ConcreteFutureImpl21RunOrScheduleCallbackERKSt10shared_ptrINS_10FutureImplEEONS2_14CallbackRecordEb+0x40)[0x7f4bbadbf1c0]
/tmp/jnilib-1738162292075553983.tmp(_ZN5arrow10FutureImpl11AddCallbackENS_8internal6FnOnceIFvRKS0_EEENS_15CallbackOptionsE+0xe5)[0x7f4bbadb5b35]
```

just ignore this failure, it's a known issue.

shims/common/src/test/scala/org/apache/spark/sql/internal/GlutenConfigUtilSuite.scala
boneanxs: done

jinchengchenghh:
> > https://github.com/apache/incubator-gluten/blob/e0092089453286b6822447a875666b41ab6f0b5c/shims/common/src/main/scala/org/apache/spark/sql/internal/GlutenConfigUtil.scala#L21-L45
> > 
> > `spark.unsafe.sorter.spill.reader.buffer.size=2m` should be converted by above codes, could you investigate why it not works?
> 
> oh, GlutenConfigUtil only process the config which prefix is 'spark.gluten'.

Do we need to update it from `spark.gluten` to `spark.`? We may map other Spark config with Velox config.
We could also keep this implementation because it may take effect for the Spark downstream project such as Iceberg config in the future. How about it? @Yohahaha

jackylee-ch:
> Do we need to update it from `spark.gluten` to `spark.`? We may map other Spark config with Velox config. We could also keep this implementation because it may take effect for the Spark downstream project such as Iceberg config in the future. How about it? @Yohahaha

The current code only gets the configuration of gluten, as `ConfigEntry` has been refactored. we need another method to handle the `ConfigEntry` of vanilla spark.
BTW, fully scanning all spark conf will be time-consuming. Maybe we can add a spark config registration list so that only the registered conf will be passed.

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -504,24 +502,20 @@ object GlutenConfig {
...
-
-    conf
-      .get(SPARK_SHUFFLE_FILE_BUFFER)
-      .foreach(
-        v =>
-          nativeConfMap
-            .put(
-              SPARK_SHUFFLE_FILE_BUFFER,
-              (JavaUtils.byteStringAs(v, ByteUnit.KiB) * 1024).toString))
+    GlutenConfigUtil.mapByteConfValue(
```
jinchengchenghh, line 508: Maybe we should add a new list like `keyWithDefault`, move the bytes config together, it will be more easier for users to add byte config.

jinchengchenghh: This is byteKeys

jinchengchenghh:
> > Do we need to update it from `spark.gluten` to `spark.`? We may map other Spark config with Velox config. We could also keep this implementation because it may take effect for the Spark downstream project such as Iceberg config in the future. How about it? @Yohahaha
> 
> The current code only gets the configuration of gluten, as `ConfigEntry` has been refactored. we need another method to handle the `ConfigEntry` of vanilla spark. BTW, fully scanning all spark conf will be time-consuming. Maybe we can add a spark config registration list so that only the registered conf will be passed.

Yes, maybe add a list to `parseConfig` is more reasonable.

Yohahaha:
thanks for the comments, we could treat these comments as https://github.com/apache/incubator-gluten/issues/8326 's subtask, or would you keep iterating in current PR? @boneanxs 

I think we can merge current PR if above comments does not blocking.

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -504,24 +502,20 @@ object GlutenConfig {
...
+      conf,
+      SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE,
+      ByteUnit.BYTE)(
+      v => nativeConfMap.put(SPARK_UNSAFE_SORTER_SPILL_READER_BUFFER_SIZE, v.toString))
+    GlutenConfigUtil.mapByteConfValue(
+      conf,
+      SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE,
+      ByteUnit.BYTE)(v => nativeConfMap.put(SPARK_SHUFFLE_SPILL_DISK_WRITE_BUFFER_SIZE, v.toString))
+    GlutenConfigUtil.mapByteConfValue(conf, SPARK_SHUFFLE_FILE_BUFFER, ByteUnit.KiB)(
```
jinchengchenghh, line 517: Is JavaUtils.byteStringAs(v, ByteUnit.KiB) * 1024) equal to JavaUtils.byteStringAs(v, ByteUnit.BYTE)?

boneanxs: Nope, if `v` doesn't provide unit, then this unit provided will be used. E.g, when v is 1024, then `JavaUtils.byteStringAs(v, ByteUnit.KiB) * 1024)` will be 1024 * 1024, while `JavaUtils.byteStringAs(v, ByteUnit.BYTE)` is 1024.

jinchengchenghh (approved):

Issue #8685, opened 2025-02-07
[VL] Gluten + rss shuffle(celeborn) encounter core dump in some situations


zjuwangg:
### Backend

VL (Velox)

### Bug description

When running gluten + celeborn(rss shuffle), I noticed one of the task consistent failed and the executor also failed.

```
# JRE version: OpenJDK Runtime Environment (ByteOpenJDK) (17.0.9+10) (build 17.0.9+10)
# Java VM: OpenJDK 64-Bit Server VM (ByteOpenJDK) (17.0.9+10, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0x28c550]  AccessInternal::PostRuntimeDispatch<G1BarrierSet::AccessBarrier<548964ul, G1BarrierSet>, (AccessInternal::BarrierType)2, 548964ul>::oop_access_barrier(void*)+0x0
---------------  T H R E A D  ---------------

Current thread (0x00007fb2713f1a00):  JavaThread "Executor task launch worker for task 125.1 in stage 4.0 (TID 330)" daemon [_thread_in_vm, id=76, stack(0x00007fb26cfff000,0x00007fb26d400000)]

Stack: [0x00007fb26cfff000,0x00007fb26d400000],  sp=0x00007fb26d3fd038,  free space=4088k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [libjvm.so+0x28c550]  AccessInternal::PostRuntimeDispatch<G1BarrierSet::AccessBarrier<548964ul, G1BarrierSet>, (AccessInternal::BarrierType)2, 548964ul>::oop_access_barrier(void*)+0x0

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
j  io.glutenproject.vectorized.ShuffleWriterJniWrapper.split(JIJJ)V+0
j  org.apache.spark.shuffle.VeloxCelebornHashBasedColumnarShuffleWriter.internalWrite(Lscala/collection/Iterator;)V+265
j  org.apache.spark.shuffle.CelebornHashBasedColumnarShuffleWriter.write(Lscala/collection/Iterator;)V+2
[...]

PR #8686, opened 2025-02-07
[GLUTEN-8685][VL]Fix coredump when rss push partition data size once exceeds int MAX_VALUE

zjuwangg:

When JavaRssClient push partition data once exceeds Int.MAX_VALUE or when executor memory is in a pressure situation, It's possible that [https://github.com/apache/incubator-gluten/blob/main/cpp/core/jni/JniCommon.h#L500](https://github.com/apache/incubator-gluten/blob/main/cpp/core/jni/JniCommon.h#L500)
`array_ = env->NewByteArray(size);` will return a null ptr, due to int64_t cast to a int32_t or there no enough memory, and following logic will cause a unexpected core dump as described in #8685 .

This PR split the bytes into serval chunk if the bytes length is very large and add the check logic to make sure throwing exception when get a null ptr array!

(Fixes: \#8685)

## How was this patch tested?

manual tests in my own local environment.


cpp/core/jni/JniCommon.h
```diff
@@ -488,22 +488,36 @@ class JavaRssClient : public RssClient {
   }
 
   int32_t pushPartitionData(int32_t partitionId, const char* bytes, int64_t size) override {
+    const int64_t maxBufferSize = 128 * 1024 * 1024; // 128 MB
```
jackylee-ch, line 491: nit: a config for this buffer size.

zjuwangg:
Thanks @jackylee-ch review. I just get a information that maybe the bytes passed to pushPartitionData can not be split into multiple blocks, which depends on the underlying rss client implementation.

So I planed here just add a null check here to avoid unexpected core dump!

zhouyuan:
CC: @kerwin-zk

Issue #8693, opened 2025-02-10
[VL] Build velox failed for machines without AVX-512 instruction set


xinghuayu007:
### Problem description

When compiling simdjson for Velox using dynamic compile method, it failed. The detail is as follow:

<img width="1917" alt="Image" src="https://github.com/user-attachments/assets/d7dfcf23-37cf-40ae-9bef-faba4422ead2" />

'vpcompressb' is an instruction of AVX-512 instruction set, see: https://en.wikipedia.org/wiki/AVX-512

But my machine does not have AVX-512, see:

<img width="1912" alt="Image" src="https://github.com/user-attachments/assets/f0f90166-7ebc-4781-bdef-d8e25c099ebc" />

### System information

<img width="1912" alt="Image" src="https://github.com/user-attachments/assets/159a2a2e-a556-4e87-92da-7190f00f477b" />

### CMake log

```bash

```

PR #8694, opened 2025-02-10
[GLUTEN-8693][VL] Disallow AVX-512 for simdjson by default

xinghuayu007:

Disallow AVX-512 for simdjson by default, when the machine doesn't have AVX-512

https://github.com/apache/incubator-gluten/issues/8693

## How was this patch tested?

Compile velox dynamicly

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -15,6 +15,8 @@ BUILD_BENCHMARKS=OFF
 ENABLE_JEMALLOC_STATS=OFF
 BUILD_VELOX_TESTS=OFF
 BUILD_VELOX_BENCHMARKS=OFF
+# Set to ON for machines have AVX-512 instruction.
+SIMDJSON_AVX512_ALLOWED=OFF
 ENABLE_QAT=OFF
 ENABLE_IAA=OFF
 ENABLE_HBM=OFF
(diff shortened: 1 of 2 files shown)


FelixYBW:
We shouldn't add a new flag for AVX512 of simdjson. We should sue the same config. Can we use the same way as #8027?

@zhouyuan @PHILO-HE

PHILO-HE:
> We shouldn't add a new flag for AVX512 of simdjson. We should sue the same config. Can we use the same way as #8027?
> 

Yes, it would be better to set the option according to the detected CPU architecture, avx, avx512, etc. One gap is velox's helper function `get_cxx_flags` doesn't consider avx512.

xinghuayu007:
> > We shouldn't add a new flag for AVX512 of simdjson. We should sue the same config. Can we use the same way as #8027?
> 
> Yes, it would be better to set the option according to the detected CPU architecture, avx, avx512, etc. One gap is velox's helper function `get_cxx_flags` doesn't consider avx512.

I reviewed the code of Velox, 'SIMDJSON_AVX512_ALLOWED' is a cmake option for simd-json and it's default value is 'ON'. Velox detects the CPU architecture, like arm64, avx, sse, aarch64 and chooses suitable instruction set. For a CPU which owns AVX instruction set but not AVX512, Velox still uses SIMDJSON_AVX512_ALLOWED=ON to compile, it is not correct.


I think it is better to detect the CPU architecture and set the value of SIMDJSON_AVX512_ALLOWED in Gluten then passes it to Velox. Although modifying the code of Velox is more direct, but not convenient.

If I have any mistake, please tell me, thanks.

Detect CPU architecture in Velox:


The default value of SIMDJSON_AVX512_ALLOWED in simd-json

FelixYBW:
Velox doesn't support AVX512 yet. Let's set SIMDJSON\_AVX512\_ALLOWED to false as default always. When Velox add the AVX512 support, we may pass it to simdjson.

Issue #8709, opened 2025-02-11
[VL] Build on openEuler 24.03 LTS with Velox backend


kevinw66:
### Description

As the usage of openEuler in China is rapidly increasing, I believe adding support for openEuler will benefit our project in multiple aspects.

PR #8710, opened 2025-02-11
[GLUTEN-8709][VL] Support build on openEuler 24.03 LTS with Velox backend

kevinw66:

(Fixes: #8709)

## How was this patch tested?

Run build scripts on openEuler 24.03 LTS.


zhztheplayer:
Should we maintain scripts for each linux distributions? E.g., does a static build of Gluten with centos / ubuntu docker work directly for openEuler?

kevinw66:
> Should we maintain scripts for every linux distributions? E.g., does a static build of Gluten with centos / ubuntu docker work directly for openEuler?

@zhztheplayer I think we can support these as long as we have contributors who are willing to maintain the scripts, as it doesn't affect the main functions and will also extend our user base.

PHILO-HE:
@kevinw66, could you share how many differences between this script and the centos-8 script?

kevinw66:
@PHILO-HE There isn't a significant difference at present. However, considering the current development trend, the difference will gradually become more prominent. Therefore, I provided an independent script file. But I can update the PR if we prefer using centos8 script.

ep/build-velox/src/setup-openeuler24.sh
```diff
@@ -0,0 +1,199 @@
...
+      run_and_time install_build_prerequisites
+    else
+      echo "Skipping installation of build dependencies since INSTALL_PREREQUISITES is not set"
+    fi
+
+    install_velox_deps
+    echo "All dependencies for Velox installed!"
+    dnf clean all
+  fi
+)
```
PHILO-HE, line 199: Please leave an empty line at the end.

PHILO-HE:
It's ok to keep a separate script. Thanks!

ep/build-velox/src/setup-openeuler24.sh
kevinw66: Done

kevinw66:
> @kevinw66, I heard this pr is for making Gluten run on arm platform. If some more popular OS like centos-8 is workable, do you still need this OS support?

@PHILO-HE Yes, I provided arm docker images of CentOS8 to be compatible with the current amd64 build(also it's more clear and maintainable in this way), and it will be more convenient for the community try to upgrade vcpkg from CentOS7 to CentOS8 in the future. the compatibility of the openEuler is still needed and I'll keep working on this.

PHILO-HE (approved):
Thanks!

Issue #8557, opened 2025-01-17
[CH]Optimize nested function calls


KevinyhZou:
### Description

In our online sqls, there are many nested functions calls, such as 
- `select a.b.c from test_tbl` => `select get_struct_field((get_struct_field(a, b),c) from test_tbl`
- `select count(1) from test_tbl where a = 1 and b = 2 and c = 3` => `select count(1) from test_tbl where and(and(a = 1, b = 2), c = 3 )`
- `select count(1) from test_tbl where a = 1 or b = 2 or c = 3` => `select count(1) from test_tbl where or(or(a = 1, b = 2), c = 3 )`
.....

and these functions can be optimized by make the nested arguments collapsed , which means
`get_struct_field((get_struct_field(a, b),c)` => get_struct_field(a, b, c)
`and(and(a = 1, b = 2), c = 3 )` => and(a =1, b=2, c=3)
`or(or(a=1, b= 2), c=3)` => or(a=1, b=2, c=3)
`get_json_object(get_json_object(a, '$.b'), '$.c')` => `get_json_object(a, '$.b', '$.c')`
.....

which can reduce the function calls and improve performance.

lgbo-ustc:
```
get_json_object(get_json_object(a, '$.b'), '$.c') => get_json_object(a, '$.b', '$.c')

```
印象之前是用spark rule改写成`get_json_object(a, '$.b.c')`, 这里为何改写成这样，这是spark function支持的形式？

KevinyhZou:
改写成get_json_object(a, '$.b.c') , 有些case 兼容不了。 get_json_object(a, '$.b', '$.c') 不是spark `get_json_object` 兼容的形式，这里是通过ScalaUDF 的方式将get_json_object 转为 ScalaUDF，来兼容get_json_object(a, '$.b', '$.c') 这种形式。 @lgbo-ustc

lgbo-ustc:
什么case

KevinyhZou:
1.  get_json_object(get_json_object({"a":"{\\\"x\\\":5}"}', '$.a'), '$.x'), the json string has  backslashes to escape quotes ;
 2. get_json_object(get_json_object('{"a.b": 0}', '$.a), '$.b'), the json key contains dot character(.) and it's same as the collapsed json path;

you can also look at the comments at `CollapseGetJsonObjectExpressionRule`

lgbo-ustc:
Let's make things clear.
[...]

PR #8733, opened 2025-02-14
[GLUTEN-8557][CH]Optimize nested function calls for `GetJsonObject`

KevinyhZou:

(Fixes: \#8557)

## How was this patch tested?

test by ut


lgbo-ustc:
Since `JSON` type has been support in `CH`, could you see if following works
- Define a new function `get_json_object_ex`, its inputs are
   - 1st argument is dynamic type, it is a string or a json object. when the input is string, parse it into json object at first
   - 2nd argument is string type, it's the json path
   and its return is a json object
- At the final output, add a cast call to cast the json object into string

By using `get_json_object_ex`, we could reuse the intermedia json object of the same json string by sharing common subexpression.

KevinyhZou:
cc @lgbo-ustc

PR #8753, opened 2025-02-17
[VL] nit: gluten-it: Fix non-POSIX shell warnings in centos-7-deps.sh

zhztheplayer:
The following is reported by `shellcheck`:

```
In POSIX sh, [[ ]] is undefined.
See SC2039.
```

Fix by changing to `#/bin/bash`

### DIFF
--- a/tools/gluten-te/centos/centos-7-deps.sh
+++ b/tools/gluten-te/centos/centos-7-deps.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#!/bin/bash
 # Licensed to the Apache Software Foundation (ASF) under one or more
 # contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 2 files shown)


zhztheplayer:
cc @PHILO-HE thank you

PHILO-HE:
@zhztheplayer, curious can we make it a bash script by adding `#!/bin/bash`? I note many of our scripts are actually using bash-specific features.

zhztheplayer:
> @zhztheplayer, curious can we make it a bash script by adding `#!/bin/bash`? I note many of our scripts are actually using bash-specific features.

Thank you for pointing out. I didn't notice that there are `#/bin/sh`s. Let's modify the shell kind directly.

PHILO-HE (approved):
Thanks!

PR #8780, opened 2025-02-19
[GLUTEN-8779][VL][Minor] Code cleanup for native validation

marin-ma:
`SubstraitToVeloxPlanValidator::validateLiteral` allows all literals to pass, so it should be removed


rui-mo (approved):
Thanks!

PR #8793, opened 2025-02-20
[WIP][VL] Fix inconsistency issue of PartitionFile path unescaping & GPL issue

yaooqinn:

GlutenURLDecoder.java is copied from OpenJDK and it's under GPL v2 which belongs to **Category X**, we can't have it in Apache Releases.

URLDecoder decode/encode is not fully compatible with the Hive catalog path escaping/unescaping, which Spark also follows. 

Besides, https://github.com/apache/spark/pull/46938 has improved `unescapePathName`'s speed at the Spark side by ~10x. So This PR also helps gluten gain perf which handles datasets w/ large partition numbers. 

## How was this patch tested?

The current tests shall be fine.


FelixYBW (approved):

FelixYBW:
Thank you for the catch.

PR #8798, opened 2025-02-20
[GLUTEN-8794] Support logging as many fallback reasons as possible

marin-ma:
1. Add configuration `spark.gluten.sql.validation.failFast` to control whether to fail validation on the first failure or to continue verifying as much as possible. More validations can provide additional fallback reasons for a given query. `true` by default to follow the current fail-fast behaviour.

For example, with setting `spark.gluten.sql.validation.failFast=false` the log for query `SELECT convert_timezone('Europe/Moscow', 'America/Los_Angeles', timestamp_ntz'2022-01-01 00:00:00')` shows like
Before:
```
15:26:54.204 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=38], due to: Found schema check failure for StructType(StructField(convert_timezone(Europe/Moscow, America/Los_Angeles, TIMESTAMP_NTZ '2022-01-01 00:00:00'),TimestampNTZType,false)), due to: Schema / data type not supported: TimestampNTZType.
```

After:
```
15:09:35.179 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=38], due to: 
 - Found schema check failure for StructType(StructField(convert_timezone(Europe/Moscow, America/Los_Angeles, TIMESTAMP_NTZ '2022-01-01 00:00:00'),TimestampNTZType,false)), due to: Schema / data type not supported: TimestampNTZType
 - Validation failed with exception from: ProjectExecTransformer, reason: Not supported to map spark function name to substrait function name: convert_timezone(Europe/Moscow, America/Los_Angeles, 2022-01-01 00:00:00), class name: ConvertTimezone.
```

2. Remove duplicated log.

For example, the log for query `select timestamp_ntz'2022-01-01 00:00:00' = date'2022-01-01'`

Before:
```
15:26:53.895 WARN org.apache.gluten.execution.ProjectExecTransformer: Validation failed with exception for plan: ProjectExecTransformer, due to: Type TimestampNTZType not supported.
15:26:53.896 WARN org.apache.gluten.execution.ProjectExecTransformer: Validation failed with exception for plan: ProjectExecTransformer, due to: Type TimestampNTZType not supported.
15:26:53.896 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=25], due to: Type TimestampNTZType not supported..
```

After:
```
15:31:58.753 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=44], due to: 
 - Validation failed with exception from: ProjectExecTransformer, reason: Type TimestampNTZType not supported.
```

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/ValidationResult.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/ValidationResult.scala
@@ -26,7 +26,6 @@ import org.apache.spark.sql.catalyst.trees.TreeNode
 sealed trait ValidationResult {
   def ok(): Boolean
   def reason(): String
-  def merge(other: ValidationResult): ValidationResult
 }
 
 object ValidationResult {
@@ -43,28 +42,23 @@ object ValidationResult {
     override def ok(): Boolean = true
     override def reason(): String = throw new UnsupportedOperationException(
       "Succeeded validation doesn't have failure details")
-
-    override def merge(other: ValidationResult): ValidationResult = {
-      if (!other.ok()) {
-        return other
-      }
-      this
-    }
   }
 
   private case class Failed(override val reason: String) extends ValidationResult {
     override def ok(): Boolean = false
-
-    override def merge(other: ValidationResult): ValidationResult = {
-      if (!other.ok()) {
-        return ValidationResult.failed(reason + other.reason(), prefix = "")
-      }
-      this
-    }
   }
 
   def succeeded: ValidationResult = Succeeded
   def failed(reason: String, prefix: String = "\n - "): ValidationResult = Failed(prefix + reason)
+  def merge(left: ValidationResult, right: ValidationResult): ValidationResult =
+    (left.ok(), right.ok()) match {
+      case (_, true) =>
+        left
+      case (true, false) =>
+        right
+      case (false, false) =>
+        failed(left.reason() + right.reason(), prefix = "")
+    }
 
   implicit class EncodeFallbackTagImplicits(result: ValidationResult) {
     def tagOnFallback(plan: TreeNode[_]): Unit = {
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/ValidationResult.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/ValidationResult.scala
@@ -26,6 +26,7 @@ import org.apache.spark.sql.catalyst.trees.TreeNode
 sealed trait ValidationResult {
   def ok(): Boolean
   def reason(): String
+  def merge(other: ValidationResult): ValidationResult
 }
 
 object ValidationResult {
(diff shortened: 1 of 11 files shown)


marin-ma:

gluten-substrait/src/main/scala/org/apache/gluten/extension/GlutenPlan.scala
```diff
@@ -26,6 +26,7 @@ import org.apache.spark.sql.catalyst.trees.TreeNode
 sealed trait ValidationResult {
   def ok(): Boolean
   def reason(): String
+  def merge(other: ValidationResult): ValidationResult
```
zhztheplayer, line 29: I see fallback tags already support appending: https://github.com/apache/incubator-gluten/blob/453cbebc26f5dfef76669f8abcda578820420352/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/FallbackTag.scala#L37C14-L37C24

Can we do merge either only on tags or only on validation results? Say, either to make tags mergeable and then convert the merged tag to a big validation result, or to make tags non-mergeable then convert the tags to validation results that can be merged afterwards.

marin-ma: If we want to consolidate the two, the validation failure should match the tagged operator. However, the validation result failures are not strictly bound to the tagged operator. For example in ShuffleExchangeExec, we add a sort before exchange for round-robin partitioning and the validation failure of the Sort will add a fallback tag on ColumnarShuffleExchange.

Merging the validation reason is more for the debugging purpose to give the developer more information by not returning the validation result immediately. As it's merged in a single validation call, it should be considered as one source of fallback. It seems that appending FallbackTags is intended to add multiple fallback reasons to an operator when fallback is triggered by different sources.

gluten-substrait/src/main/scala/org/apache/gluten/extension/ValidationResult.scala
```diff
@@ -26,6 +26,7 @@ import org.apache.spark.sql.catalyst.trees.TreeNode
 sealed trait ValidationResult {
   def ok(): Boolean
   def reason(): String
+  def merge(other: ValidationResult): ValidationResult
```
zhztheplayer, line 29: Just my 2 cents, but would suggest just adding a static merge method, say:

```scala
def merge(left: ValidationResult, right: ValidationResult): ValidationResult =
  (left.ok(), right.ok()) match {
    case (true, true) =>
      succeeded
    case (false, false) =>
      failed(s"${left.reason()}\n - ${right.reason()}")
    case (true, false) =>
      right
    case (false, true) =>
      left
  }
```

zhztheplayer (approved):
Just a minor, and could remove the option if fast-failing is not important here.

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1310,6 +1312,12 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(false)
 
+  val VALIDATION_FAIL_FAST =
+    buildConf("spark.gluten.sql.validation.failFast")
+      .internal()
+      .booleanConf
+      .createWithDefault(true)
```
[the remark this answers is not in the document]
marin-ma: Non-fast-failing may impact performance, as there are more validation process to run. It should be preferred to still keep the fast-failing by default.

PR #8829, opened 2025-02-26
[VL] Remove resolving ViewFs file path from scan validation

PHILO-HE:

I note Velox's HDFS filesystem allows ViewFs scheme. 
https://github.com/facebookincubator/velox/blob/main/velox/connectors/hive/storage_adapters/hdfs/HdfsFileSystem.cpp#L112.

Then, we can just rely on the uniform JNI call through `allSupportedByRegisteredFileSystems`, instead of resolving the ViewFS file path in Gluten scan validation.

## How was this patch tested?

Customer validation.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -51,9 +51,7 @@ import org.apache.spark.sql.types._
 import org.apache.spark.util.SerializableConfiguration
 
 import org.apache.hadoop.fs.Path
-import org.apache.hadoop.fs.viewfs.ViewFileSystemUtils
 
-import scala.collection.mutable
 import scala.util.control.Breaks.breakable
 
 class VeloxBackend extends SubstraitBackend {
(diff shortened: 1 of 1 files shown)


PHILO-HE:
@wangyum, @JkSelf, could you take a look?

JkSelf:
@PHILO-HE We still require this support on the Gluten side. Velox's ViewFS is only supported on a single node and does not extend to cross-node clusters, making it unsuitable for @wangyum  environment.

PHILO-HE:
> @PHILO-HE We still require this support on the Gluten side. Velox's ViewFS is only supported on a single node and does not extend to cross-node clusters, making it unsuitable for @wangyum environment.

@JkSelf, I understand Velox doesn't fully support ViewFS, so path resolving is required before plan execution. But during scan validation, as ViewFS path will always be supported by HDFS file system if it's registered (it's guaranteed by `allSupportedByRegisteredFileSystems`), it seems unnecessary to do a path resolving here.

JkSelf:
@PHILO-HE Oh, I understand. It makes sense to remove this conversion during the validation phase. Thanks.

JkSelf (approved):
LGTM.

wangyum:
lgmt.

PR #8834, opened 2025-02-26
[GLUTEN-8811][VL]Fix bucket scan when some partitionValue is empty

jinchengchenghh:
With storage partition join, suppose the left leaf partition is [1, 2], the right partition is [1, 2, 10], the number of partitions is not equal, use empty partition for those partition values that are not present.
```
index | left | right
0 | Seq(1) | Seq(1)
1 | Seq(2) | Seq(2)
2 |  Seq() | Seq(10)
```
Relevant Spark PR: 
https://github.com/apache/spark/pull/35657


FelixYBW:
@z123 Is bucket join used in your side?

FelixYBW (approved):

PR #8845, opened 2025-02-27
Revert "[CORE] Change DISCLAIMER to DISCLAIMER-WIP"

yaooqinn:
Reverts apache/incubator-gluten#6442

Since we have already published 3-ASF releases w/ DISCLAIMER-WIP, it's OK for us to use the standard DISCLAIMER


PHILO-HE (approved):
@weiting-chen

PR #8866, opened 2025-03-02
[GLUTEN-8865][VL] Support year month interval to integral casts

ArnavBalyan:
 - Add support for creating yearmonthinterval expressions.
 - Support offload cast of INTERVAL_YEAR_MONTH to integral type in velox.

## How was this patch tested?
 - Unit tests


gluten-ut/spark33/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenCastSuite.scala
```diff
@@ -152,4 +152,38 @@ class GlutenCastSuite extends CastSuite with GlutenTestsTrait {
 
     checkEvaluation(cast(Literal.create(null, IntegerType), ShortType), null)
   }
+
+  test("casting from year-month interval to integer (Spark 3.2 / 3.3)") {
```
jinchengchenghh, line 156: Why only Spark 3.2 / 3.3?

ArnavBalyan: still working on the PR, I'll update for all versions thanks

PR #8842, opened 2025-02-27
[VL] Enable spark function map_filter

j7nhai:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#ISSUE-ID)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #8869, opened 2025-03-03
Revert "[VL] Enable spark function map_filter"

baibaichen:
Reverts apache/incubator-gluten#8842, it cause ch failed


zhanglistar (approved):
LGTM

PR #8878, opened 2025-03-03
[CORE] Make scalatest.testFailureIgnore configurable for convenience

ccat3z:

Make testFailureIgnore configurable for local tests.

## How was this patch tested?

`mvn ... -Dmaven.test.failure.ignore=true`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1354,7 +1354,7 @@
           <version>${scalatest-maven-plugin.version}</version>
           <configuration>
             <reportsDirectory>${project.build.directory}/surefire-reports</reportsDirectory>
-            <testFailureIgnore>false</testFailureIgnore>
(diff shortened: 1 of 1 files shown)


Yohahaha (approved):

pom.xml
```diff
@@ -1354,7 +1354,7 @@
           <version>${scalatest-maven-plugin.version}</version>
           <configuration>
             <reportsDirectory>${project.build.directory}/surefire-reports</reportsDirectory>
-            <testFailureIgnore>false</testFailureIgnore>
+            <testFailureIgnore>${maven.test.failure.ignore}</testFailureIgnore>
```
Yohahaha, line 1357: doest `-DtestFailureIgnore=true` not work?

what would the default value be if we dont specify `maven.test.failure.ignore`?

Yohahaha: if the default value is false, should we just remove this line?

PR #8186, opened 2024-12-09
[VL] Change loadQuantum default value to 8MB from 258MB

yikf:
## What changes were proposed in this pull request?

This PR aims to change loadQuantum default value to 8MB from 258MB.

Velox currently only [support up to 8MB load quantum size on SSD](https://github.com/facebookincubator/velox/blob/main/velox/dwio/common/CachedBufferedInput.h#L182) and add a limit on loadQuantum by https://github.com/facebookincubator/velox/commit/c6d7390db644e0d0769f248e5b3c271dea8f58b1. 

We should modify the default values to make them user-friendly out of the box and unnecessary large values.

```c++
// We only support up to 8MB load quantum size on SSD and there is no need for
  // larger SSD read size performance wise.
  void checkLoadQuantum() {
    if (cache_->ssdCache() != nullptr) {
      VELOX_CHECK_LE(
          options_.loadQuantum(),
          1 << cache::SsdRun::kSizeBits,
          "Load quantum exceeded SSD cache entry size limit.");
    }
  }
```

## How was this patch tested?
doc only.

PR #8880, opened 2025-03-03
[GLUTEN-5884][VL] change default load quantum to 8M for local SSD cache

zhouyuan:

Velox is using 23 bits fro cache entry size
https://github.com/facebookincubator/velox/pull/10242

## How was this patch tested?

manually tested with local SSD cache enabled

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
@@ -406,10 +406,10 @@ object VeloxConfig {
   val LOAD_QUANTUM =
     buildStaticConf("spark.gluten.sql.columnar.backend.velox.loadQuantum")
       .internal()
-      .doc("Set the load quantum for velox file scan, recommend to use the default value (256MB) " +
+      .doc("Set the load quantum for velox file scan, recommend to use the default value (8MB) " +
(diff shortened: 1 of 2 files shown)


jackylee-ch:
We had another pr #8186 that talk abouk this before. Given the [suggestion](https://github.com/apache/incubator-gluten/pull/8186#issuecomment-2532694505) by @FelixYBW , we finally use check for loadQuantum config in #8197

zhouyuan:
> We had another pr #8186 that talk abouk this before. Given the [suggestion](https://github.com/apache/incubator-gluten/pull/8186#issuecomment-2532694505) by @FelixYBW , we finally use check for loadQuantum config in #8197

Thanks! it looks like the default `256M` is not working now, we either need to modify Velox to support bigger entry size or we need to set it to lower in Gluten.

FelixYBW (approved):

Yohahaha:
@zhouyuan may I ask why we still need change the default value after #8197 merged?

FelixYBW:
#8197 makes sense. Let's revert this PR since we have the docs.  
  
Can we add a warning to check if local ssd cache is enabled, override the config to 8MB?  
  
"Set the load quantum for velox file scan, recommend to use the default value (256MB) for performance consideration. If Velox cache is enabled, it can be 8MB at most."

zhouyuan:
Yes I agree. Let me try to refine the logic and override the config when local cache is enabled.

PR #8881, opened 2025-03-03
[INFRA] Switch archive.a.o to closer.lua to avoid abuse

yaooqinn:

The archive.a.o has a threshold for daily traffic and a threshold for weekly total download, a continuous abuse might get automic ban.


yaooqinn:
The related docker-image.yml test record can be found here  https://github.com/yaooqinn/gluten/actions/runs/13630345892/workflow

FelixYBW (approved):

Issue #8891, opened 2025-03-04
[VL] Allow to reuse local SSD cache on Spark context restart


zhouyuan:
### Description

Currently local SSD cache will be discarded after Spark context shutdown. It would better to make it reusable with a new optional config

zhouyuan:
@jackylee-ch Thanks, I was told the soft cache affinity can help to alleviate this issue, but I'm trying to find more resource verify 
https://github.com/apache/incubator-gluten/blob/main/shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala#L691

jackylee-ch:
> I was told the soft cache affinity can help to alleviate this issue, but I'm trying to find more resource verify 

AFAIK, the cache involved in duplicateReading will become invalid after Spark restarts, unless the cache can be reused and the executor rescheduling issue is resolved. Nevertheless, the pr is still helpful for cache reuse within the same application.

zhouyuan:
For single instance based tests it should be still useful.

PR #8892, opened 2025-03-04
[GLUTEN-8891][VL] Allow to reuse local SSD cache on Spark context restart

zhouyuan:

This patch adds one config to allow reuse of local SSD cache
Before this patch, the ssd cache will be discarded on spark context shutdown. After this with the new config(ssdReuse) the cache will be kept there, and the users need to manually delete it

Fixes: #8891 

## How was this patch tested?

manual tests

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
@@ -147,13 +147,12 @@ object VeloxConfig {
       .booleanConf
       .createWithDefault(false)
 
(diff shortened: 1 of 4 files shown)


jinchengchenghh (approved):

FelixYBW:
when the data on local ssd is deleted with and without the PR?

zhouyuan:
> when the data on local ssd is deleted with and without the PR?

Before this patch, the ssd cache will be discarded on spark context shutdown. After this with the new config(ssdReuse) the cache will be kept there, and the users need to manually delete it

Issue #8894, opened 2025-03-04
[VL] Buffer overflow in jStringToCString causes SIGSEGV on arm


kevinw66:
### Backend

VL (Velox)

### Bug description

Test fails on arm when running [VeloxScanSuite](https://github.com/apache/incubator-gluten/blob/ca2ab6ad7d9c461b7ca1eb7b032f460ce4d567ca/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala#L111) with this assertion 
```scala
assert(
  VeloxFileSystemValidationJniWrapper.allSupportedByRegisteredFileSystems(
    Array("file:/test_path/")))
```

#### Possible reason
When calling `jStringToCString` with JNI string conversion, a buffer overflow occurs because the allocated buffer is too small to accommodate the null terminator. This causes SIGSEGV (segmentation fault) specifically on ARM64 platforms.

#### Steps to Reproduce
1. Create a `jstring` with content that when converted to UTF-8 has a length that's a multiple of 8 bytes (e.g., "file:/test_path/")
2. The function allocates a buffer with size `clen`
3. `GetStringUTFRegion` attempts to write `clen + 1` bytes (including null terminator)
4. On ARM64, this causes SIGSEGV due to buffer overflow

#### Current Code
```cpp
static inline std::string jStringToCString(JNIEnv* env, jstring string) {
  int32_t jlen, clen;
  clen = env->GetStringUTFLength(string);
  jlen = env->GetStringLength(string);
  char buffer[clen];  // Buffer too small
  env->GetStringUTFRegion(string, 0, jlen, buffer);
  return std::string(buffer, clen);
}
```

#### Expected Behavior
The buffer should be large enough to accommodate the null terminator that GetStringUTFRegion writes.  

#### Actual Behavior
On x86_64: Works due to stack alignment padding (but still undefined behavior)
On ARM64: Crashes with SIGSEGV due to stricter memory access requirements

### Spark version

None

### Relevant logs

```bash
- tpch q22 subquery filter pushdown - v1
- tpch q22 subquery filter pushdown - v2
25/02/24 16:40:00 WARN GlutenFallbackReporter: Validation failed for plan: Exchange, due to: [FallbackByBackendSettings] Validation failed on node Exchange.
25/02/24 16:40:00 WARN GlutenFallbackReporter: Validation failed for plan: Exchange[QueryId=3636], due to: [FallbackByBackendSettings] Validation failed on node Exchange.
#
[...]

PR #350, opened 2022-09-01
[Gluten-349] Bug fix for checkException memory double free error

KevinyhZou:
## What changes were proposed in this pull request?
bug fix for convert parquet to dwrf memory double free error as describe in #349 

## How was this patch tested?
manual tests
1. run tpch_convert_parquet_dwrf.sh it cause the error describe in #349
2. modfiy the jni_common.h as in pr, and re-run the script, no errors, and the dwrf file was produced.

PR #8895, opened 2025-03-04
[GLUTEN-8894][VL] Fix buffer overflow in jStringToCString on arm

kevinw66:

(Fixes: #8894)

## How was this patch tested?

### DIFF
--- a/cpp/core/jni/JniCommon.h
+++ b/cpp/core/jni/JniCommon.h
@@ -39,7 +39,7 @@ static inline std::string jStringToCString(JNIEnv* env, jstring string) {
   int32_t jlen, clen;
   clen = env->GetStringUTFLength(string);
   jlen = env->GetStringLength(string);
-  char buffer[clen];
+  char buffer[clen + 1];
   env->GetStringUTFRegion(string, 0, jlen, buffer);
   return std::string(buffer, clen);
 }


jinchengchenghh (approved):

jinchengchenghh (approved):
Thanks for your fix. We noticed here was core dump, but didn't find the root cause.
This function origins from arrow https://github.com/apache/arrow-java/blob/main/adapter/orc/src/main/cpp/jni_wrapper.cpp#L83, maybe you also are interested on fixing it on arrow.

kevinw66:
> Thanks for your fix. We noticed here was core dump, but didn't find the root cause. This function origins from arrow https://github.com/apache/arrow-java/blob/main/adapter/orc/src/main/cpp/jni_wrapper.cpp#L83, maybe you also are interested on fixing it on arrow.

@jinchengchenghh Thanks. However, I think the code in Apache Arrow should be fine. Since it allocates memory using vector, which is more intelligent in this regard. But considering that we don't use vector perhaps to improve performance, so I made the changes in our own way.

jinchengchenghh:
No, if use std::vector<>, it writes to vector.data(), cause coredump at all the platforms, https://github.com/apache/incubator-gluten/pull/350

kevinw66:
@jinchengchenghh Thanks for the information. I'll take a close look at these.

marin-ma:
I got below compilation error with clang
```
In file included from /Users/rong/workspace/github/apache/gluten/cpp/core/jni/JniError.cc:17:
In file included from /Users/rong/workspace/github/apache/gluten/cpp/core/jni/JniError.h:22:
/Users/rong/workspace/github/apache/gluten/cpp/core/jni/JniCommon.h:39:15: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
   39 |   char buffer[clen + 1];
      |               ^~~~~~~~
/Users/rong/workspace/github/apache/gluten/cpp/core/jni/JniCommon.h:39:15: note: read of non-const variable 'clen' is not allowed in a constant expression
/Users/rong/workspace/github/apache/gluten/cpp/core/jni/JniCommon.h:36:17: note: declared here
   36 |   int32_t jlen, clen;
      |                 ^
1 error generated.
```

Perhaps it's not a good practice to use variable lengths arrays in the code. VLA is not part of the C++ standard.

Here's another implementation in arrow https://github.com/apache/arrow-java/blame/main/dataset/src/main/cpp/jni_util.cc#L318-L326 , which is more clear to use `GetStringUTFChars` and `ReleaseStringUTFChars` @jinchengchenghh Do you know if there are any reason we didn't implement it in this way?

jinchengchenghh:
We use https://github.com/apache/arrow-java/blob/main/adapter/orc/src/main/cpp/jni_wrapper.cpp#L83 before, but core dump, we don't try this one https://github.com/apache/arrow-java/blame/main/dataset/src/main/cpp/jni_util.cc#L318-L326, maybe you can have a try, or just change char[clen + 1] to std::vector<char> buffer(clen + 1); @marin-ma

PR #8904, opened 2025-03-05
[CORE] Enlarge defaultRecursionLimit by pre-loading the protobuf class

PHILO-HE:

We previously introduced custom protobuf dependency with only defaultRecursionLimit enlarged to avoid the limit is hit for deeply nested plans (one user reported this issue). It is not easy to upgrade this dependency since it requires local build and then publishing the jar to maven repo.

With this change, we can remove the custom protobuf dependency for easily upgrading protobuf when needed.

## How was this patch tested?

Verified by local remote debug.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -23,6 +23,7 @@ import org.apache.gluten.config.GlutenConfig._
 import org.apache.gluten.events.GlutenBuildInfoEvent
 import org.apache.gluten.exception.GlutenException
 import org.apache.gluten.extension.GlutenSessionExtensions
+import org.apache.gluten.initializer.Initializer
 import org.apache.gluten.task.TaskListener
 
 import org.apache.spark.{SparkConf, SparkContext, TaskFailedReason}
@@ -55,6 +56,7 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
   private var _sc: Option[SparkContext] = None
 
   override def init(sc: SparkContext, pluginContext: PluginContext): util.Map[String, String] = {
+    Initializer
     _sc = Some(sc)
     val conf = pluginContext.conf()
 
@@ -267,6 +269,7 @@ private[gluten] class GlutenExecutorPlugin extends ExecutorPlugin {
 
   /** Initialize the executor plugin. */
   override def init(ctx: PluginContext, extraConf: util.Map[String, String]): Unit = {
+    Initializer
     // Initialize Backend.
     Component.sorted().foreach(_.onExecutorStart(ctx))
   }
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -23,7 +23,7 @@ import org.apache.gluten.config.GlutenConfig._
 import org.apache.gluten.events.GlutenBuildInfoEvent
 import org.apache.gluten.exception.GlutenException
 import org.apache.gluten.extension.GlutenSessionExtensions
-import org.apache.gluten.initializer.Initializer
+import org.apache.gluten.initializer.CodedInputStreamClassInitializer
 import org.apache.gluten.task.TaskListener
 
 import org.apache.spark.{SparkConf, SparkContext, TaskFailedReason}
@@ -56,7 +56,6 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
   private var _sc: Option[SparkContext] = None
 
   override def init(sc: SparkContext, pluginContext: PluginContext): util.Map[String, String] = {
-    Initializer
     _sc = Some(sc)
     val conf = pluginContext.conf()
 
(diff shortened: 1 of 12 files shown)


PHILO-HE:
@FelixYBW, @zhztheplayer

zhouyuan:
related
https://github.com/apache/incubator-gluten/pull/3814
https://github.com/protocolbuffers/protobuf/issues/13301

FelixYBW:
[https://github.com/apache/incubator-gluten/blob/18f151017f4442b53340802166f7a222b0731723/gluten-iceberg/pom.xml#L84](https://github.com/apache/incubator-gluten/blob/18f151017f4442b53340802166f7a222b0731723/gluten-iceberg/pom.xml#L84)

set to "provide"  
several other places also need to be updated

FelixYBW:
[https://github.com/apache/incubator-gluten/blob/18f151017f4442b53340802166f7a222b0731723/package/pom.xml#L356](https://github.com/apache/incubator-gluten/blob/18f151017f4442b53340802166f7a222b0731723/package/pom.xml#L356)  
  
Remove?

PHILO-HE:
> https://github.com/apache/incubator-gluten/blob/18f151017f4442b53340802166f7a222b0731723/gluten-iceberg/pom.xml#L84
> 
> set to "provide" several other places also need to be updated

@FelixYBW, I just removed protobuf from iceberg & hudi module as it is not really used, and changed the scope to provided for hudi module.

PHILO-HE:
> https://github.com/apache/incubator-gluten/blob/18f151017f4442b53340802166f7a222b0731723/package/pom.xml#L356
> 
> Remove?

Yes, seems we can remove it. Let's see CI feedback.

FelixYBW (approved):

FelixYBW:
Where is the code to modify protobuf and build customized version? Is it manually upload to mvn repo?

PHILO-HE:
> Where is the code to modify protobuf and build customized version? Is it manually upload to mvn repo?

@FelixYBW , the code is maintained in my personal repo: https://github.com/PHILO-HE/protobuf/tree/v3.23.4-0.

Yes, we manually did that.
We firstly obtained a jar by building protobuf from source at local, then go through maven's jar publish process to make it available to download when building gluten.

FelixYBW:
> @FelixYBW , the code is maintained in my personal repo: [PHILO-HE/protobuf@`v3.23.4-0`](https://github.com/PHILO-HE/protobuf/tree/v3.23.4-0).
> 
> Yes, we manually did that. We firstly obtained a jar by building protobuf from source at local, then go through maven's jar publish process to make it available to download when building gluten.

Thank you for the PR! manually doing this definitely isn't a right way. Glad we fixed it.

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -267,6 +268,7 @@ private[gluten] class GlutenExecutorPlugin extends ExecutorPlugin {
 
   /** Initialize the executor plugin. */
   override def init(ctx: PluginContext, extraConf: util.Map[String, String]): Unit = {
+    CodedInputStreamClassInitializer
```
zhztheplayer, line 271: Can we give an API for this? E.g.,

```scala
CodedInputStreamClassInitializer.modifyDefaultRecursionLimitUnsafe(100000)
```

gluten-core/src/main/scala/org/apache/gluten/initializer/CodedInputStreamClassInitializer.scala
```diff
@@ -0,0 +1,42 @@
...
+    try {
+      // scalastyle:off classforname
+      // Use the shaded class name.
+      val clazz: Class[_] =
+        Class.forName("org.apache.gluten.shaded.com.google.protobuf.CodedInputStream")
+      // scalastyle:on classforname
+      val field: Field = clazz.getDeclaredField("defaultRecursionLimit")
+      field.setAccessible(true)
+      // Enlarge defaultRecursionLimit whose original value is 100.
+      field.setInt(null, 100000)
```
zhztheplayer, line 37: nit: We can add a log for such settings.

```
logInfo("... has been modified to ...")
```

zhztheplayer (commented):
Thanks!

gluten-core/src/main/scala/org/apache/gluten/initializer/CodedInputStreamClassInitializer.scala
PHILO-HE: @zhztheplayer, fixed by https://github.com/apache/incubator-gluten/pull/8996. Thanks!

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
PHILO-HE: @zhztheplayer, I feel it may be not necessary to expose the setting API since the value we set is large enough. I prefer hiding the detail in this piece of code. Thanks!

zhztheplayer: Thank you for the explanation. It's OK if we don't pass a number. Though I would suggest having an method for this purpose than relying on class initialization, e.g.,

```
CodedInputStreamClassInitializer.modifyDefaultRecursionLimitUnsafe()
```

Code like `CodedInputStreamClassInitializer` can be less intuitive and perhaps could cause IDEs to complain about the writing or so.

PR #8906, opened 2025-03-05
[GLUTEN-8905][VL]Ignore some CSV flaky tests

jinchengchenghh:


zhouyuan (approved):
👍 
Will try to check if static vcpkg build of arrow can help

PR #8911, opened 2025-03-05
[CORE] Decrease offheap memory size in resource profile for whole stage fallback case

PHILO-HE:

For whole stage fallback case, we should not request the same offheap memory size as offload case.

## How was this patch tested?

Fix a feature that is still experimental.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
+++ b/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
@@ -88,6 +88,11 @@ case class GlutenAutoAdjustStageResourceProfile(glutenConf: GlutenConfig, spark:
       val newExecutorMemory =
         new ExecutorResourceRequest(ResourceProfile.MEMORY, newMemoryAmount.toLong)
       executorResource.put(ResourceProfile.MEMORY, newExecutorMemory)
+
+      val newExecutorOffheap =
+        new ExecutorResourceRequest(ResourceProfile.OFFHEAP_MEM, offheapRequest.get.amount / 10)
+      executorResource.put(ResourceProfile.OFFHEAP_MEM, newExecutorOffheap)
+
       val newRP = new ResourceProfile(executorResource.toMap, taskResource.toMap)
       return applyNewResourceProfileIfPossible(plan, newRP, rpManager)
     }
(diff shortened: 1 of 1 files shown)


gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
```diff
@@ -88,6 +88,11 @@ case class GlutenAutoAdjustStageResourceProfile(glutenConf: GlutenConfig, spark:
       val newExecutorMemory =
         new ExecutorResourceRequest(ResourceProfile.MEMORY, newMemoryAmount.toLong)
       executorResource.put(ResourceProfile.MEMORY, newExecutorMemory)
+
+      val newExecutorOffheap =
+        new ExecutorResourceRequest(ResourceProfile.OFFHEAP_MEM, offheapRequest.get.amount / 10)
```
PHILO-HE, line 93: @zjuwangg, could you take a look? This is just an empirical setting. And I am not sure which setting is better.

zjuwangg: LGTM in whole stage fallback case!

zjuwangg (approved):

zjuwangg (approved):

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
jackylee-ch: This would work only for Spark 3.5.4 or higher as https://github.com/apache/spark/pull/48963

JkSelf: @PHILO-HE Should we consider making this setting configurable?

PHILO-HE: @JkSelf, let's first use this empirical fixed setting and wait for feedback from user side. In theory, user don't need to configure too much offheap memory in stage fallback case. Thanks!

jackylee-ch (approved):

JkSelf (approved):
LGTM. Thanks

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
zhouyuan: @PHILO-HE 
Thanks for improving on this!
just to confirm, I can see the code logic is to reduce the offheap memory size to 1/10, but on-heap memory size is not increased, is this intended? 

Thanks, -yuan

PR #8913, opened 2025-03-05
[GLUTEN-3620][VL] RangeExec support for fallback by user options

ArnavBalyan:
 - Supports enabling or disabling range exec via user options.
 - spark.gluten.sql.columnar.range can be used to control the replacement.

## How was this patch tested?
 - Existing UTs already in place


shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1691,4 +1693,10 @@ object GlutenConfig {
       .checkValue(_ > 0, s"must be positive.")
       .createWithDefault(10)
 
+  val COLUMNAR_RANGE_ENABLED =
+    buildConf("spark.gluten.sql.columnar.range")
```
Yohahaha, line 1697: what the difference between `spark.gluten.sql.columnar.expand` and here.

ArnavBalyan: spark.gluten.sql.columnar.expand belongs to the expand operator. `spark.gluten.sql.columnar.range` is meant for the newly added columnar range

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
```diff
@@ -186,6 +186,7 @@ object Validators {
           if !(glutenConf.enableColumnarSample && BackendsApiManager.getSettings
             .supportSampleExec()) =>
         fail(p)
+      case p: ExpandExec if !glutenConf.enableColumnarRange => fail(p)
```
Yohahaha, line 189: seems this same as gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala#line 154

Yohahaha: you are trying to match ExpandExec again with glutenConf.enableColumnarRange, why

ArnavBalyan: Ah that's a good catch, it seems the ExpandExec got copied, updated it thanks!

Yohahaha (approved):

Yohahaha:
@ArnavBalyan would you rebase latest commit?

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
Yohahaha: do you fix the wrong case match? I still see `case p: ExpandExec`

Yohahaha (approved):

PR #8950, opened 2025-03-10
[GLUTEN-8949][Core] Simplify synchronization from JniLibLoader

ArnavBalyan:
 - JniLibLoader some synchronizations which are no longer needed, these are benign but can be removed for improved performance.
 - Removed such synchronizations where we are already thread safe.

## How was this patch tested?
 - Unit Tests


zhztheplayer:
As the APIs are mostly `public static`, I think the synchronisations are not completely meaningless. But there seems to be a problem that the synchronisations code is a little bit messy. Can we use a simple way to synchronize the code? E.g., by simply adding `synchronzized` keyword to all the public methods. Performance will not be an issue here, the APIs are expected only to be called a few times.

zhztheplayer (approved):
👍

PR #8957, opened 2025-03-11
[GLUTEN-8956][VL] Add support for casting binary to string

ArnavBalyan:
 - Binary to string currently falls back.
 - Offload it to Gluten.

## How was this patch tested?
 - Existing Spark UTs
 - New Unit Tests added


backends-velox/src/test/scala/org/apache/spark/sql/catalyst/expressions/VeloxCastSuite.scala
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package org.apache.spark.sql.catalyst.expressions
+
+import org.apache.gluten.execution.VeloxWholeStageTransformerSuite
+
+import org.apache.spark.sql.types._
+
+class VeloxCastSuite extends VeloxWholeStageTransformerSuite with ExpressionEvalHelper {
+  def cast(v: Any, targetType: DataType, timeZoneId: Option[String] = None): Cast = {
```
xinghuayu007, line 24: timeZoneId is not used. Maybe you can pass a const value to function 'Cast()'.

backends-velox/src/test/scala/org/apache/spark/sql/catalyst/expressions/VeloxCastSuite.scala
```diff
@@ -0,0 +1,53 @@
...
+    v match {
+      case lit: Expression =>
+        logDebug(s"Cast from: ${lit.dataType.typeName}, to: ${targetType.typeName}")
```
xinghuayu007, line 27: It is not better to add logs in a unit test.

FelixYBW (approved):

PR #8961, opened 2025-03-11
[GLUTEN-8960][VL] Move spark32 tests to nightly run

zhouyuan:

This patch moved Spark-3.2 related tests to nightly run to save the CI budget

## How was this patch tested?

pass GHA


PHILO-HE (commented):
Do we need to run Spark-3.2 test for PR by need? I.e., let GHA monitor `shims/spark32`, `gluten-ut/spark32`. If there are any code changes under these folders, trigger spark32 tests.

I am afraid we will have to do follow-up work for PRs with spark-32 code changes if we only do the test nightly.

Issue #8964, opened 2025-03-11
BroadcastNestedLoopJoin supports more join type when there are no join conditions


WangGuangxin:
### Description

BroadcastNestedLoopJoin can support full outer join when there are no join conditions

PR #8965, opened 2025-03-11
[GLUTEN-8964][VL] Support BNLJ full outer join without condition

WangGuangxin:
Currently, BroadcastNestedLoopJoin doesn't support full outer join. That's because in velox/presto, it supposed that the probe side and build side(broadcast side) are all in one worker, so that it can output mismatch records from build side.

But for spark, the probe side are distributed across different executors, so that velox cann't recognize the mismatch records from build side.
But if the join condition is empty (which is usually the case where BroadcastedNestedLoopJoin take effect), there is no need to output mismatch records. 

(Fixes: \#8964)

## How was this patch tested?

UT and manually

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/BroadcastNestedLoopJoinExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/BroadcastNestedLoopJoinExecTransformer.scala
@@ -169,6 +169,13 @@ abstract class BroadcastNestedLoopJoinExecTransformer(
   def validateJoinTypeAndBuildSide(): ValidationResult = {
     val result = joinType match {
       case _: InnerLike | LeftOuter | RightOuter => ValidationResult.succeeded
+      case FullOuter =>
+        if (condition.isEmpty) {
+          ValidationResult.succeeded
+        } else {
+          ValidationResult.failed(
+            s"FullOuter join with join condition is not supported with BroadcastNestedLoopJoin")
+        }
       case _ =>
         ValidationResult.failed(s"$joinType join is not supported with BroadcastNestedLoopJoin")
     }
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/BroadcastNestedLoopJoinExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/BroadcastNestedLoopJoinExecTransformer.scala
@@ -169,7 +169,8 @@ abstract class BroadcastNestedLoopJoinExecTransformer(
   def validateJoinTypeAndBuildSide(): ValidationResult = {
     val result = joinType match {
       case _: InnerLike | LeftOuter | RightOuter => ValidationResult.succeeded
-      case FullOuter =>
+      case FullOuter
+          if BackendsApiManager.getSettings.broadcastNestedLoopJoinSupportsFullOuterJoin() =>
         if (condition.isEmpty) {
           ValidationResult.succeeded
         } else {
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -556,8 +556,6 @@ object VeloxBackendSettings extends BackendSettingsApi {
 
   override def supportCartesianProductExec(): Boolean = true
 
-  override def supportBroadcastNestedLoopJoinExec(): Boolean = true
-
   override def supportSampleExec(): Boolean = true
 
   override def supportColumnarArrowUdf(): Boolean = true
@@ -565,5 +563,7 @@ object VeloxBackendSettings extends BackendSettingsApi {
   override def needPreComputeRangeFrameBoundary(): Boolean = true
 
   override def supportCollectLimitExec(): Boolean = true
+
+  override def broadcastNestedLoopJoinSupportsFullOuterJoin(): Boolean = true
 
 }
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
@@ -24,6 +24,7 @@ import org.apache.spark.SparkConf
 import org.apache.spark.sql.{AnalysisException, DataFrame, Row}
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.adaptive.AdaptiveSparkPlanHelper
+import org.apache.spark.sql.execution.joins.BroadcastNestedLoopJoinExec
 import org.apache.spark.sql.execution.window.WindowExec
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.internal.SQLConf
@@ -1988,6 +1989,25 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
   test("support null type in aggregate") {
     runQueryAndCompare("SELECT max(null), min(null) from range(10)".stripMargin) {
       checkGlutenOperatorMatch[HashAggregateExecTransformer]
+    }
+  }
+
+  test("FullOuter in BroadcastNestLoopJoin") {
+    withTable("t1", "t2") {
+      spark.range(10).write.format("parquet").saveAsTable("t1")
+      spark.range(10).write.format("parquet").saveAsTable("t2")
+
+      // with join condition should fallback.
+      withSQLConf("spark.sql.autoBroadcastJoinThreshold" -> "1MB") {
+        runQueryAndCompare("SELECT * FROM t1 FULL OUTER JOIN t2 ON t1.id < t2.id") {
+          checkSparkOperatorMatch[BroadcastNestedLoopJoinExec]
+        }
+
+        // without join condition should offload to gluten operator.
+        runQueryAndCompare("SELECT * FROM t1 FULL OUTER JOIN t2") {
+          checkGlutenOperatorMatch[BroadcastNestedLoopJoinExecTransformer]
+        }
+      }
     }
   }
 }
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -343,6 +343,9 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
     case ::substrait::CrossRel_JoinType::CrossRel_JoinType_JOIN_TYPE_LEFT:
       joinType = core::JoinType::kLeft;
       break;
+    case ::substrait::CrossRel_JoinType::CrossRel_JoinType_JOIN_TYPE_OUTER:
+      joinType = core::JoinType::kFull;
+      break;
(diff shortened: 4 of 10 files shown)


WangGuangxin:
@PHILO-HE @jinchengchenghh @rui-mo Can you please help review this?

gluten-substrait/src/main/scala/org/apache/gluten/execution/BroadcastNestedLoopJoinExecTransformer.scala
```diff
@@ -169,6 +169,13 @@ abstract class BroadcastNestedLoopJoinExecTransformer(
   def validateJoinTypeAndBuildSide(): ValidationResult = {
     val result = joinType match {
       case _: InnerLike | LeftOuter | RightOuter => ValidationResult.succeeded
+      case FullOuter =>
```
JkSelf, line 172: It seems there exist failed unit test in ck backend. Maybe need to only support in velox backend.

JkSelf (commented):
Thanks for your work. Can you help to fix the failed unit tests?

WangGuangxin:
The clickhouse ci seems success? but finally failed because of unknow reason

jinchengchenghh:
No, it contains 3 workflows, please review all the log to find the error information.
```
-1-0' YEAR TO MONTH), class name: IntegralDivide.
[2025-03-12T09:02:59.225Z] - Gluten - test fallback event *** FAILED ***
[2025-03-12T09:02:59.229Z]   2 did not equal 0 (GlutenFallbackSuite.scala:126)
[2025-03-12T09:02:59.278Z] 09:02:59.093 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=4398], due to: fallback input file expression
[2025-03-12T09:02:59.302Z] - TPCH Q7
[2025-03-12T09:02:59.378Z] 2025-03-12 09:02:59.172 [1][ScalaTest-main-running-DiscoverySuite] WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=4571], due to: fallback input file expression
[2025-03-12T09:02:59.380Z] 2025-03-12 09:02:59.273 [1][ScalaTest-main-running-DiscoverySuite] WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=4572], due to: fallback input file expression
[2025-03-12T09:02:59.419Z] 09:02:59.150 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=28451], due to: 
[2025-03-12T09:02:59.422Z]  - Validation failed with exception from: ProjectExecTransformer, reason: Not supported to map spark function name to substrait function name: (INTERVAL '2-0' YEAR TO MONTH div INTERVAL '1-0' YEAR TO MONTH), class name: IntegralDivide.
[2025-03-12T09:02:59.422Z] 09:02:59.169 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=28452], due to: 
[2025-03-12T09:02:59.422Z]  - Validation failed with exception from: ProjectExecTransformer, reason: Not supported to map spark function name to substrait function name: (INTERVAL '2-0' YEAR TO MONTH div INTERVAL '-1-0' YEAR TO MONTH), class name: IntegralDivide.
[2025-03-12T09:02:59.422Z] 09:02:59.187 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=28453], due to: 
[2025-03-12T09:02:59.422Z]  - Validation failed with exception from: ProjectExecTransformer, reason: Not supported to map spark function name to substrait function name: (INTERVAL '1-0' YEAR TO MONTH div INTERVAL '0-3' YEAR TO MONTH), class name: IntegralDivide.
[2025-03-12T09:02:59.422Z] 09:02:59.204 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: Project[QueryId=28454], due to: 
[2025-03-12T09:02:59.422Z]  - Validation failed with exception from: ProjectExecTransformer, reason: Not
```

@baibaichen Can we split it to separate workflow? Now it is really confused and hard to find the error message, several persons have reported to me.

JkSelf (approved):
LGTM. Thanks.

PHILO-HE (approved):
Thanks!

xinyuezg:
We’ve been investigating correctness issues in `BroadcastNestedLoopJoin` with FULL OUTER JOIN, and traced the root cause back to this PR (#8965).

This change assumes that when the join condition is empty, Velox can skip emitting build-side mismatches because there’s “no need.” That assumption works in Presto/Velox where probe and build run in a single worker, but it does not hold in Spark:
	•	In Spark, the probe side is distributed across multiple executors.
	•	**Velox cannot tell globally whether the probe side is empty or not, since each driver only sees its own partition.**
	•	As a result, with FULL OUTER JOIN and empty join condition, build-side mismatches may be suppressed incorrectly.
	•	This leads to wrong results compared to Spark’s native behavior (where a repartition(1) is used to guarantee correct global build-mismatch handling).

We validated this both in our environment and by comparing against Spark’s implementation. The conclusion is that Velox cannot currently produce correct results for Spark in this case.

Proposed resolution:
We believe the correct fix is to fall back to Spark for all FULL OUTER JOIN cases in BNLJ, regardless of whether a join condition is specified. This guarantees correctness.

jinchengchenghh:
Could you help take a look? Thanks! @WangGuangxin

WangGuangxin:
> mismatches

@xinyuezg Hi, If it is FULL OUTER JOIN and empty join condition, there should be NO mismatched rows in build side, right?

WangGuangxin:
@xinyuezg So what't the data diff looks like in your env? Are there any duplicate records

xinyuezg:
> > mismatches
> 
> @xinyuezg Hi, If it is FULL OUTER JOIN and empty join condition, there should be NO mismatched rows in build side, right?

Hi @WangGuangxin - that's correct: with FULL OUTER and an empty join condition, there should be no build-side mismatches **when the probe side is non-empty**.

The corner case we're hitting is about global vs per-executor emptiness on Spark:
* If the probe side is globally empty (i.e. no executor receives any probe rows), the correct result is all build rows padded with NULLs on the probe side.
* If the probe side is non-empty overall, but some executors happen to get empty probe partitions, those executors **must not** emit build-mismatch rows. Only the cross product rows from executors that have probe data should be produced; no mismatches at all in this case.

In current Velox code, a single Task/driver can decide this correctly because probe+build are co-located. On Spark, the probe side is distributed across executors and Velox lacks a **global "did any executor see probe rows?"** signal. With PR #8965, a driver that saw an empty probe partition may incorrectly emit build-mismatch rows even though other executors did see probe data, causing extra rows.

Concretely, consider two executors:
* Executor X: probe partition non-empty -> emits cross-join rows (correct).
* Executor Y: probe partition empty -> should emit nothing in this scenario (because probe is not globally empty). Today Y can still emit build-side NULL rows, which is wrong.

Because Velox doesn't have a task-global aggregation for this today, we recommend falling back to Spark for all FULL OUTER BNLJ cases (condition or not). To support Velox with no condition BNLJ, we probably need to add a global coordination/OR of a `sawProbeRows` flag so that build-mismatch rows are emitted only when probe is globally empty, and only emit once.

xinyuezg:
> @xinyuezg So what't the data diff looks like in your env? Are there any duplicate records

Example we encountered from `GlutenOuterJoinSuiteForceShjOff`

* no condition full outer join using BroadcastNestedLoopJoin build left (whole-stage-codegen off)
* no condition full outer join using BroadcastNestedLoopJoin build left (whole-stage-codegen on)
* no condition full outer join using BroadcastNestedLoopJoin build right (whole-stage-codegen off)
* no condition full outer join using BroadcastNestedLoopJoin build right (whole-stage-codegen on)

Zooming into no condition full outer join using BroadcastNesteLoopJoin build left:

Initial input data:
```
left(a, b)  
(1, 2.0)  
(2, 100.0)  
(2, 1.0)  
(2, 1.0)   ← duplicated
(3, 3.0)  
(5, 1.0)  
(6, 6.0)  
(null, null)  
  
right(c, d)  
(0, 0.0)  
(2, 3.0)  
(2, -1.0)  
(2, -1.0)  ← duplicated  
(2, 3.0)   ← duplicated  
(3, 2.0)  
(4, 1.0)  
(5, 3.0)  
(7, 7.0)  
(null, null)
```

Spark plan:
```
 BroadcastNestedLoopJoin BuildLeft, FullOuter
:- Filter (isnotnull(a#220) AND (a#220 = 2))
:  +- Scan ExistingRDD[a#220,b#221]
+- Filter (isnotnull(c#226) AND (c#226 = 2))
   +- Scan ExistingRDD[c#226,d#227]
```

So the effective inputs to joins are:  
* filteredLeft (build side) = {(2, 100.0), (2, 1.0), (2, 1.0)}  → 3 rows  
* filteredRight (probe side) = {(2, 3.0), (2, -1.0), (2, -1.0), (2, 3.0)}  → 4 rows

Results:
```
 == Results ==
 !== Expected Answer - 12 ==   == Actual Answer - 15 ==
 [2,1.0,2,-1.0]               [2,1.0,2,-1.0]
 [2,1.0,2,-1.0]               [2,1.0,2,-1.0]
 [2,1.0,2,-1.0]               [2,1.0,2,-1.0]
 [2,1.0,2,-1.0]               [2,1.0,2,-1.0]
 [2,1.0,2,3.0]                [2,1.0,2,3.0]
 [2,1.0,2,3.0]                [2,1.0,2,3.0]
 [2,1.0,2,3.0]                [2,1.0,2,3.0]
 [2,1.0,2,3.0]                [2,1.0,2,3.0]
![2,100.0,2,-1.0]             [2,1.0,null,null]
![2,100.0,2,-1.0]             [2,1.0,null,null]
![2,100.0,2,3.0]              [2,100.0,2,-1.0]
![2,100.0,2,3.0]              [2,100.0,2,-1.0]
!                             [2,100.0,2,3.0]
!                             [2,100.0,2,3.0]
!                             [2,100.0,null,null]
```

@xinyuezg Hi, thanks for your detailed explanation. Yes, it will produce wrong result if the prob side is empty, this situation is quite unique and was not something I initially anticipated.
Currently, there seems to be no other workarounds available. I think we can revert this change.

xinyuezg:
@WangGuangxin Thanks! We validated on our side that reverting this change restores correct results by falling back to Spark for **all FULL OUTER BNLJ** cases (conditioned or not). Glad we reached consensus. We can follow up later if we add a task-global coordination to enable this safely in Velox.

One thing I'm curious about -- in your validation, the existing test suite (`GlutenOuterJoinSuiteForceShjOff`) seems didn't fail? Was that expected? Could it be that the inputs in the tests didn't cover the empty-partition scenario in your testing environment? Along with the revert, would it make sense to add explicit test cases for FULL OUTER BNLJ with both globally empty probe side and partition-empty but non-globally empty probe side? That would help ensure we don't regress again in the future.

jinchengchenghh:
Could you draft a PR to revert this one? @WangGuangxin

zml1206:
I understand that `NestedLoopJoinProbe::beginBuildMismatch` is designed to avoid this situation. Could you provide a complete unit test to reproduce it? @xinyuezg Thanks.

zml1206:
image
image

I've reproduced the issue. I'll create a pull request to remove it.

PR #8972, opened 2025-03-11
[GLUTEN-6887][VL] Daily Update Velox Version (2025_03_12)

GlutenPerfBot:
Upstream Velox's New Commits:

```txt
4b1740dfd by Richard Barnes, Fix deprecated use of 0/NULL in velox/tpch/gen/dbgen/bm_utils.cpp + 1 (12612)
c81bb1b3e by Bradley Dice, build: Add Docker images for CUDA development (12413)
7f1e4ef08 by Yuan, fix(fuzz): Fix ssd cache o_direct flag in cache fuzz test (12523)
35a124705 by Bikramjeet Vig, fix: Recursive selector in json path used with scalar json (12595)
c2eb47b59 by Natasha Sehgal, Velox Array Join Unknown support (12584)
47aa2b533 by Karteek Murthy, Add new stats to to track decompression and compression times (12593)
89afec85e by Wei He, misc(fuzzer): Add custom verification for approx_set in AggregationFuzzerTest and WindowFuzzerTest (12425)
a254ad762 by Wei He, fix: Fix approx_set to use murmur3 hash function to match Presto (12374)
b12d69f47 by Kevin Wilfong, feat: Throw invalid_time_zone exception from tzdb [4/4] (12475)
21f618051 by Kevin Wilfong, test: The new tzdb library can handle DST into the distant future [3/4] (12471)
915d304cf by Kevin Wilfong, feat: Support writing zoned_time to stream [2/4] (12469)
16e446060 by Kevin Wilfong, feat: Replace tz implementation from date with LLVM implementation [1/4] (12422)
```

PR #8988, opened 2025-03-13
[VL] Update centos setup scripts to install new tzdata

zhouyuan:

followup #8972 

## How was this patch tested?

pass GHA


ep/build-velox/src/setup-centos8.sh
```diff
@@ -59,6 +59,7 @@ function install_build_prerequisites {
   dnf_install ninja-build curl ccache gcc-toolset-11 git wget which
   dnf_install yasm
   dnf_install autoconf automake python39 python39-devel python39-pip libtool
```
zhouyuan, line 61: centos8 does not provide the updates for tzdata, so we used the package for almalinux

PHILO-HE (approved):

PR #8996, opened 2025-03-14
[CORE] Avoid ClassNotFoundException when loading a shaded protobuf class

PHILO-HE:

Follow-up fix for #8904. This exception can happen in maven test due to the class we are loading is shaded in final package phase.

## How was this patch tested?

No such exception printed in CI test.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/initializer/CodedInputStreamClassInitializer.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/initializer/CodedInputStreamClassInitializer.scala
@@ -28,12 +28,19 @@ import java.lang.reflect.Field
 object CodedInputStreamClassInitializer extends Logging {
   private val newDefaultRecursionLimit = 100000
 
-  {
+  def modifyDefaultRecursionLimitUnsafe: Unit = {
     try {
       // scalastyle:off classforname
-      // Use the shaded class name.
       val clazz: Class[_] =
-        Class.forName("org.apache.gluten.shaded.com.google.protobuf.CodedInputStream")
+        try {
+          // Use the shaded class name.
+          Class.forName("org.apache.gluten.shaded.com.google.protobuf.CodedInputStream")
+        } catch {
+          // The above class is shaded in final package phase (see package/pom.xml).
+          // If ClassNotFoundException happens, e.g., in mvn test, load the original class instead.
+          case _: ClassNotFoundException =>
+            Class.forName("com.google.protobuf.CodedInputStream")
+        }
       // scalastyle:on classforname
       val field: Field = clazz.getDeclaredField("defaultRecursionLimit")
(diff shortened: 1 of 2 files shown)


PHILO-HE:
@zhouyuan, please take a look.

gluten-core/src/main/scala/org/apache/gluten/initializer/CodedInputStreamClassInitializer.scala
```diff
@@ -34,8 +38,14 @@ object CodedInputStreamClassInitializer {
...
       // Enlarge defaultRecursionLimit whose original value is 100.
-      field.setInt(null, 100000)
+      field.setInt(null, newDefaultRecursionLimit)
+      logInfo(
+        s"The defaultRecursionLimit in protobuf has been increased to $newDefaultRecursionLimit")
     } catch {
+      // The class is shaded during package phase (see package/pom.xml). We tolerate
+      // this ClassNotFoundException in mvn test.
+      case _: ClassNotFoundException =>
+        logWarning("The defaultRecursionLimit in protobuf has not been modified.")
```
zhztheplayer, line 48: Curious why don't we support both cases? Say we either find `com.google.protobuf.CodedInputStream`, or find `org.apache.gluten.shaded.com.google.protobuf.CodedInputStream`, or raise an error.

PHILO-HE: @zhztheplayer, just updated the pr to allow loading the original class. Thanks!

zhztheplayer (approved):

gluten-core/src/main/scala/org/apache/gluten/initializer/CodedInputStreamClassInitializer.scala
```diff
@@ -16,27 +16,41 @@
...
       field.setAccessible(true)
       // Enlarge defaultRecursionLimit whose original value is 100.
-      field.setInt(null, 100000)
+      field.setInt(null, newDefaultRecursionLimit)
+      logInfo(
+        s"The defaultRecursionLimit in protobuf has been increased to $newDefaultRecursionLimit")
     } catch {
-      case e: Exception => e.printStackTrace()
+      case e: Exception =>
+        log.error("Failed to modify the DefaultRecursionLimit in protobuf: " + e.getMessage)
```
zhztheplayer, line 53: Is Gluten still reliable if the limitation is failed to modify? If not we can just throw here.

PHILO-HE: @zhztheplayer, if it's not modified, Gluten can still work well for most cases. This modification is just for avoiding the protobuf representation failure for deeply nested plans. So we can tolerate the exception.

PR #9011, opened 2025-03-15
[GLUTEN-9010][VL] Fix GlutenCastSuite for Spark 34 and 35

ArnavBalyan:
 - Spark 34 and 35 runs CastWithAnsiON suite by default. Certain tests fail since Gluten does not support ANSI on. And we don't run CastWithAnsiOff.
 - Move the GlutenCastSuite to CastWithAnsiOff rather than CastSuiteBase.
 - Certain tests with ANSI off are failing before this PR excluding them and will fix in upcoming changes.

## How was this patch tested?
 - Unit tests


jackylee-ch:
It would be better to have `GlutenCastSuite` extend `CastWithAnsiOffSuite` instead of `CastSuiteBase`. Then, we can keep `GlutenTryCastSuite` for ANSI mode.

jackylee-ch:
> Thanks, for Gluten do you think we want to run tests for ANSI mode enabled? Since it's not supported and unit tests may not work as expected. It seems we correctly run it with ANSI off for Spark 32,33 however behaviour changed due to Spark 34/35. Pls let me know your thoughts thanks

We have included `ANSI support` in `Gluten's 2025 Roadmap` and we need this test to keep the behavior with vanilla Spark.

jackylee-ch:
Basically look good to me, thanks. Please update the PR description.

jackylee-ch (approved):

Issue #9019, opened 2025-03-17
[VL] Cast from DECIMAL(10, 2) to TIMESTAMP is not supported


wForget:
### Backend

VL (Velox)

### Bug description

I found an error through fuzz-testing: https://github.com/wForget/fuzz-test-spark-native/issues/160

error message:

```
[ERROR] Query failed in native engine gluten: Job aborted due to stage failure: Task 0 in stage 2304.0 failed 1 times, most recent failure: Lost task 0.0 in stage 2304.0 (TID 1606) (fv-az1758-193.vo1scjuza2hupagbdr52ddpx1h.cx.internal.cloudapp.net executor driver): org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: Error during calling Java code from native code: org.apache.gluten.exception.GlutenException: org.apache.gluten.exception.GlutenException: Exception: VeloxUserError
Error Source: USER
Error Code: UNSUPPORTED
Reason: Cast from DECIMAL(10, 2) to TIMESTAMP is not supported
Retriable: False
Context: Top-level Expression: try_cast((n0_0) as TIMESTAMP)
Function: applyDecimalToPrimitiveCast
/work/ep/build-velox/build/velox_ep/./velox/expression/CastExpr-inl.h
Line: 673
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_14VeloxUserErrorERKSsEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox4exec8CastExpr27applyDecimalToPrimitiveCastIlEESt10shared_ptrINS0_10BaseVectorEERKNS0_17SelectivityVectorERKS5_RNS1_7EvalCtxERKS4_IKNS0_4TypeEESI_
# 4  _ZN8facebook5velox4exec8CastExpr11applyPeeledERKNS0_17SelectivityVectorERKNS0_10BaseVectorERNS1_7EvalCtxERKSt10shared_ptrIKNS0_4TypeEESG_RSB_IS6_E
[...]

PR #9022, opened 2025-03-17
[GLUTEN-9019][VL] Add a check to fall back "cast decimal to timestamp"

wForget:

velox does not support cast decimal to timestamp.

The `isInteger()` check in velox returns true for decimal type.

Fixes: #9019

## How was this patch tested?

added unit test

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -263,6 +263,11 @@ bool SubstraitToVeloxPlanValidator::isAllowedCast(const TypePtr& fromType, const
 
   // Limited support for X to Timestamp.
   if (toType->isTimestamp()) {
+    if (fromType->isDecimal()) {
+      // Decimal to Timestamp is not supported.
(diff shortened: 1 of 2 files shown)


wForget:
The newly added test case can reproduce this issue:

image

PHILO-HE (approved):
Looks good!

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -263,6 +263,11 @@ bool SubstraitToVeloxPlanValidator::isAllowedCast(const TypePtr& fromType, const
 
   // Limited support for X to Timestamp.
   if (toType->isTimestamp()) {
+    if (fromType->isDecimal()) {
```
jackylee-ch, line 266: Since we already defined the default false, why we need a extra check here?

PHILO-HE (commented):
I feel Velox's isXXX api should correctly reflect the logical type. I will try to submit a pr in Velox to see the community's feedback. Anyway, we can firstly merge this pr if CI passes.

PR #8870, opened 2025-03-03
[CORE] Format file at src-*/

j7nhai:
## What changes were proposed in this pull request?

Currently files at src-* cannot be formatted by spotless

(Fixes: \#ISSUE-ID)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #9024, opened 2025-03-17
[VL][CI] Dump & upload logs for unit test to GitHub artifact

yaooqinn:

This PR adds changes to test log4j configurations and related GA steps to dump and upload logs for unit tests to GitHub artifacts.

- Reduce the noise of the Console appendder
- Reduce the chances for the console log being clipped in some build environment like inside a Docker
- Add them to GitHub artifacts to let them be downloaded and viewed locally

## How was this patch tested?

Files like `gluten-substrait/target/unit-tests.log` being generated locally


.github/workflows/velox_backend.yml
```diff
@@ -1223,3 +1319,15 @@ jobs:
           # Depends on --build_example=ON.
           $MVN_CMD test -Pspark-3.5 -Pbackends-velox -Piceberg -Pdelta -DtagsToExclude=None \
           -DtagsToInclude=org.apache.gluten.tags.UDFTest
+            - name: Upload test report
```
yaooqinn, line 1322:
```suggestion
```

zhouyuan (approved):
+1
I saw the info level logs are gone from the GHA

zhouyuan:
Hi @yaooqinn thanks for adding this feature, it looks like the log file are not uploaded after run?
image

JkSelf:
cc @PHILO-HE

yaooqinn:
@zhouyuan The logs are only necessary for GA job failures, but skipping for those in canceled and ok status.

FYI, here is a GA for mocking the uploaded case
https://github.com/apache/incubator-gluten/actions/runs/13894317684/job/38871553505

pom.xml
```diff
@@ -1371,6 +1371,9 @@
             <testFailureIgnore>${maven.test.failure.ignore}</testFailureIgnore>
             <junitxml>.</junitxml>
             <argLine>${argLine} ${extraJavaTestArgs}</argLine>
+            <systemProperties>
+              <log4j.configurationFile>file:src/test/resources/log4j2.properties</log4j.configurationFile>
```
PHILO-HE, line 1375: not all modules have `src/test/xxx` structure, see module `gluten-iceberg/`.

Can we avoid duplicating this configuration file by just setting it like `${project.basedir}/xxx` in root pom?

yaooqinn: I haven't tested w/ a single root `log4j2.properties`, so I'm not sure whether it will work or not. But copying it module by module is also a good practice that follows the behavior of the Spark project. It allows us to customize each copy for different logging behaviors of different modules, or w/ different log4j2 dependencies.

.github/workflows/velox_backend.yml
```diff
@@ -638,6 +638,12 @@ jobs:
         with:
           name: test-report-spark32
           path: '**/surefire-reports/TEST-*.xml'
+      - name: Upload unit tests log files
+        if: ${{ !success() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: unit-tests-log-spark32
```
PHILO-HE, line 645: Suggest to simply reference the job name, e.g., `name: "${{ github.job }}-test-log"`. Ditto for other places.

pom.xml
PHILO-HE: ok. Please note some modules' structure is like `src-iceberg/test/resources/...`.

PHILO-HE: Maybe, do some refactor on the directory name? See previous discussion: https://github.com/apache/incubator-gluten/pull/8870#issuecomment-2696899361

.github/workflows/velox_backend.yml
yaooqinn: This name aligns w/ the test-report* and golden-files*, I don't think we need to prefix it w/ job names?

PHILO-HE: I would recommend to change golden file name also.
It' just for well distinguishing which job produces the file. It's also for better maintenance, especially when we add more test job derived from an existing one. I note `golden-files-spark35` was unexpected both used as golden file name for JDK 8 and 17 test job, which is a small mistake easy to make.

yaooqinn: OK, as we're not leveraging the matrix feature of GA, it looks like a good idea for avoiding copy-paste issues.

zhztheplayer (commented):
Thank you @yaooqinn .

I am curious do we have to add one `log4j2.properties` for every Maven module? Can we just set to the same file from parent `pom.xml`?

yaooqinn:
@zhztheplayer https://github.com/apache/incubator-gluten/pull/9024#discussion_r2004872166

zhztheplayer (approved):

PR #8890, opened 2025-03-04
[GLUTEN-8889][CORE] Bump Spark version from 3.5.2 to 3.5.5

jackylee-ch:
## What changes were proposed in this pull request?
Support Spark 3.5.5 and upgrade iceberg version to 1.8.1 to fix the [compatibility issue](https://github.com/apache/iceberg/pull/11731#issuecomment-2530114142) with Spark 3.5.4.

## How was this patch tested?
GA

PR #9031, opened 2025-03-17
[GLUTEN-8889][CORE] Bump Spark version from 3.5.2 to 3.5.5 for CH backend

jlfsdtc:

Support Spark 3.5.5 for CH backend

## How was this patch tested?

GA

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -278,6 +278,7 @@
       </activation>
       <properties>
         <java.version>1.8</java.version>
+        <iceberg.version>1.5.0</iceberg.version>
       </properties>
     </profile>
     <profile>
@@ -288,6 +289,7 @@
       <properties>
         <java.version>11</java.version>
         <caffeine.version>3.1.8</caffeine.version>
+        <iceberg.version>1.8.0</iceberg.version>
       </properties>
     </profile>
     <profile>
@@ -298,6 +300,7 @@
       <properties>
         <java.version>17</java.version>
         <caffeine.version>3.1.8</caffeine.version>
+        <iceberg.version>1.8.0</iceberg.version>
       </properties>
     </profile>
     <profile>
@@ -347,8 +350,7 @@
       <properties>
         <sparkbundle.version>3.5</sparkbundle.version>
         <sparkshim.artifactId>spark-sql-columnar-shims-spark35</sparkshim.artifactId>
-        <spark.version>3.5.2</spark.version>
-        <iceberg.version>1.5.0</iceberg.version>
+        <spark.version>3.5.5</spark.version>
         <delta.package.name>delta-spark</delta.package.name>
         <delta.version>3.2.0</delta.version>
         <delta.binary.version>32</delta.binary.version>
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,6 @@ dist/
 metastore_db/
 
 .ipynb_checkpoints
+
+# For Spark warehouse
+gluten-ut/*/spark-warehouse/
--- a/backends-clickhouse/pom.xml
+++ b/backends-clickhouse/pom.xml
@@ -273,6 +273,10 @@
             <artifactId>protobuf-java</artifactId>
             <groupId>com.google.protobuf</groupId>
           </exclusion>
+          <exclusion>
+            <groupId>jdk.tools</groupId>
+            <artifactId>jdk.tools</artifactId>
+          </exclusion>
         </exclusions>
     </dependency>
     <dependency>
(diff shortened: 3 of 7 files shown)


pom.xml
```diff
@@ -278,6 +278,7 @@
       </activation>
       <properties>
         <java.version>1.8</java.version>
+        <iceberg.version>1.5.0</iceberg.version>
```
jackylee-ch, line 281: does it works well on ch backends? For velox backends, the iceberg version must use 1.8.0

zhouyuan: For Velox backend tests, we already upgraded to JDK17. I think the intention here is to enable JDK8 + Spark-355

jackylee-ch: If we try to build with JD8 and Spark 355, iceberg should not be included. The changes here don't seem necessary to me.
Besides, we need to define `iceberg.version` in spark config file instead of jdk config file so that we can use different iceberg versions for different sparks.

jlfsdtc: If Iceberg and Spark versions are bound together, it means that after upgrading Spark to version 3.5.5, it will no longer be possible to use JDK 8. However, Spark 3.5.5 itself still supports the use of JDK 8.
Use different versions of JDK and Spark to support different versions of Iceberg.

jackylee-ch: Not get your mind. I have runed Iceberg 1.5.0 with Spark 3.5.5 before, and it can't work that way. Thus If we want use Spark 3.5.5 with JDK8, then iceberg can not be involved.

zhouyuan: @jackylee-ch 
I made a quick patch to test iceberg 1.5 + Spark-355 + JDK8, 
https://github.com/apache/incubator-gluten/pull/9001/

and it looks like all the unit tests passed:
https://github.com/apache/incubator-gluten/actions/runs/13859404046/job/38785108628

The JDK17 based tests failed as expected.
https://github.com/apache/incubator-gluten/actions/runs/13859404046/job/38785108162

zhouyuan: > If we try to build with JD8 and Spark 355, iceberg should not be included. The changes here don't seem necessary to me. Besides, we need to define `iceberg.version` in spark config file instead of jdk config file so that we can use different iceberg versions for different sparks.

+1 
it seems we should not bind JDK version with iceberg version as this will impact with Spark-344 also.

jackylee-ch: > @jackylee-ch I made a quick patch to test iceberg 1.5 + Spark-355 + JDK8, #9001
> 
> and it looks like all the unit tests passed: https://github.com/apache/incubator-gluten/actions/runs/13859404046/job/38785108628

It seems you were using https://github.com/apache/incubator-gluten/pull/8890 to test it, sorry for that as I have removed the `-Piceberg` from spark-3.5.5 and jdk8 GA.

BTW, current PR has already met the [problem](https://github.com/apache/iceberg/pull/11731#issuecomment-2530114142) I described before, the vanilla spark would core dump in `VeloxTPCHIcebergSuite`
https://github.com/apache/incubator-gluten/actions/runs/13916861429/job/38941504564?pr=9031

zhouyuan: @jackylee-ch Thanks for point out, I didnt realize this change - so it looks like Velox backend does not support this combination `JDK8 + Spark355 + Iceberg1.5`

Not sure if this applies to the CK backend also. In case CK backend can work with this, could we add a new special profile so it wont impact Velox backend?

Cc: @baibaichen

jackylee-ch:
@ jlfsdtc 
BTW, It seems that you meet the failed problem with `show create_table.sql`, you need reinstall sql tests for Spark 3.5.5, just like I did in [another PR](https://github.com/apache/incubator-gluten/pull/8890/files#diff-8dbcf212005134fd84bfbfc125fe323196c75a9bc87f81b3484d52ab1517d390R1047).

pom.xml
jackylee-ch: Unfortunately, this core dump issue occurs in the vanilla Spark 3.5.5 environment with Iceberg 1.5.0, and it is unrelated to the Velox backend. Similar problems can also be encountered when using the CH backend.

jlfsdtc: > > If we try to build with JD8 and Spark 355, iceberg should not be included. The changes here don't seem necessary to me. Besides, we need to define `iceberg.version` in spark config file instead of jdk config file so that we can use different iceberg versions for different sparks.
> 
> +1 it seems we should not bind JDK version with iceberg version as this will impact with Spark-344 also.

en, i lose of that.  Spark-344 is another version of Iceberg

Issue #9034, opened 2025-03-18
[VL] Add VeloxResizeBatchesExec right after ShuffleRead


WangGuangxin:
### Description

Add VeloxResizeBatchesExec right after ShuffleRead

PR #9035, opened 2025-03-18
[GLUTEN-9034][VL] Add VeloxResizeBatchesExec for Shuffle

WangGuangxin:
Shuffle read may generate small batch with few rows, which may hurt performance a lot. 

A example in our production case is
image

So in this PR proposed to add `VeloxResizeBatchesExec` right after shuffle read, the plan is changed to 

image

As we can see, the average batch size increased from 9 to 1000, and the total hour reduced from 2066h to 766h.

(Fixes: \#9034)

## How was this patch tested?

manually and UT


WangGuangxin:
cc @jinchengchenghh @jackylee-ch

jackylee-ch:
Can you provide the complete dag diagram? Maybe it can be solved by adjusting the number of input partitions, such as maxPartitionSize?

WangGuangxin:
## without VeloxResizeBatchesExec
image

## with VeloxResizeBatchesExec
image

WangGuangxin:
> Can you provide the complete dag diagram? Maybe it can be solved by adjusting the number of input partitions, such as maxPartitionSize?

@jackylee-ch In some cases, it can be controled by `maxPartitionSize`. But if it's under Join/Filter etc, it's not easy to make sure the batch size in a reasonable range

jackylee-ch:
image
It seems that the anomal number of vectors occurs after the shuffle read. Maybe consider adding a resize operation right after the shuffle read so that all subsequent operators can benefit?

WangGuangxin:
> image It seems that the anomal number of vectors occurs after the shuffle read. Maybe consider adding a resize operation right after the shuffle read so that all subsequent operators can benefit?

@jackylee-ch yeah, agree with that. I'll update it later

WangGuangxin:
> > image It seems that the anomal number of vectors occurs after the shuffle read. Maybe consider adding a resize operation right after the shuffle read so that all subsequent operators can benefit?
> 
> @jackylee-ch yeah, agree with that. I'll update it later

@jackylee-ch @jinchengchenghh @zhztheplayer updated, and also update the title and description. please take a look when you are convenient

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -914,4 +915,25 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
       child: Seq[SparkPlan]): ColumnarRangeBaseExec =
     ColumnarRangeExec(start, end, step, numSlices, numElements, outputAttributes, child)
 
+  override def genShuffleRead(plan: SparkPlan): SparkPlan = {
+    if (!VeloxConfig.get.veloxResizeBatchesShuffleInput) return plan
```
jackylee-ch, line 919: can we use another config for this?
Besides, have you checked TPCDS performance on this PR? It seems suitable only for specific scenarios.

WangGuangxin: seems there are performance downgrade. I'll investigate later

marin-ma:
@WangGuangxin Any update on this patch? Could you do a rebase? Thanks!

WangGuangxin:
> @WangGuangxin Any update on this patch? Could you do a rebase? Thanks!

thinks, will rebase today

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -306,6 +317,21 @@ object VeloxConfig {
       .intConf
       .createOptional
 
+  val COLUMNAR_VELOX_RESIZE_BATCHES_SHUFFLE_INPUT_OUTPUT_MIN_SIZE =
```
zhztheplayer, line 320: `s/SHUFFLE_INPUT_OUTPUT/SHUFFLE_OUTPUT` ?

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -34,16 +34,19 @@ class VeloxConfig(conf: SQLConf) extends GlutenConfig(conf) {
...
+    getConf(COLUMNAR_VELOX_RESIZE_BATCHES_SHUFFLE_OUTPUT)
+
   case class ResizeRange(min: Int, max: Int) {
     assert(max >= min)
     assert(min > 0, "Min batch size should be larger than 0")
     assert(max > 0, "Max batch size should be larger than 0")
   }
 
-  def veloxResizeBatchesShuffleInputRange: ResizeRange = {
+  def veloxResizeBatchesShuffleInputOutputRange: ResizeRange = {
```
zhztheplayer, line 46: ditto

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -294,6 +297,14 @@ object VeloxConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val COLUMNAR_VELOX_RESIZE_BATCHES_SHUFFLE_OUTPUT =
+    buildConf("spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleOutput")
+      .internal()
+      .doc(s"If true, combine small columnar batches together right after shuffle read. " +
+        s"The default minimum output batch size is equal to 0.8 * ${COLUMNAR_MAX_BATCH_SIZE.key}")
```
zhztheplayer, line 304: Is 0.8 out-of-date? I see it's now 0.25. Can you update it in passing?

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeAfterShuffleRead.scala
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.gluten.extension
+
+import org.apache.gluten.config.VeloxConfig
+import org.apache.gluten.execution.VeloxResizeBatchesExec
+
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.{ColumnarShuffleExchangeExec, SparkPlan}
+import org.apache.spark.sql.execution.adaptive.{AQEShuffleReadExec, ShuffleQueryStageExec}
+
+case class AppendBatchResizeAfterShuffleRead() extends Rule[SparkPlan] {
```
zhztheplayer, line 26: We don't have a rule for the resizing on the input side of shuffle. Can we make the ways of optimizations more consistent? Either both via rules, or both not?

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
WangGuangxin: This is used both for the BatchResizing before shuffle input and after shuffle output, so that we can reduce some config. Usually there is need to do too much customized min size config for these two scenario. What do you think?

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
WangGuangxin: ok

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeAfterShuffleRead.scala
WangGuangxin: At first,  resizing on the output side of shuffle followings the way for shuffle read, that's do it when converting to transformer. But after the `DummpyLeafExec` is introduced, it doesn't work.
So I'll refactor the way to add resizing on the input side of shuffle, to make it enabled by rule

zhztheplayer: Fair enough. Let's make it an individual rule.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
WangGuangxin: done

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeAfterShuffleRead.scala
WangGuangxin: update

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -345,19 +345,6 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
-            if !shuffle.useSortBasedShuffle &&
-              VeloxConfig.get.veloxResizeBatchesShuffleInput =>
-          val range = VeloxConfig.get.veloxResizeBatchesShuffleInputRange
-          val appendBatches =
-            VeloxResizeBatchesExec(shuffle.child, range.min, range.max)
-          shuffle.withNewChildren(Seq(appendBatches))
-        case _ => plan
-      }
-    }
-
```
zhztheplayer, line 360: Thanks for factoring this out!

zhztheplayer (approved):

WangGuangxin:
@jackylee-ch @zhztheplayer Can we merge this?

jackylee-ch (approved):
Sorry for late response. Great work!

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Try to append [[VeloxResizeBatchesExec]] for shuffle input and ouput to make the batch sizes in
+ * good shape.
+ */
+case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
+  override def apply(plan: SparkPlan): SparkPlan = {
+    val range = VeloxConfig.get.veloxResizeBatchesShuffleInputOutputRange
+    plan.transformUp {
+      case shuffle: ColumnarShuffleExchangeExec
+          if !shuffle.useSortBasedShuffle &&
```
zhouyuan, line 35: it looks like this will be only enabled on hash based shuffle?
Cc @marin-ma

jinchengchenghh:
Do you notice in some queries, the plan cannot be fully replaced? Like TPCDS Q95, there are 4 AQEShuffleRead, but only one add the VeloxResizeBatch node. @WangGuangxin 
Image 11-11-2025 at 14 46

jinchengchenghh:
Fixed in https://github.com/apache/incubator-gluten/pull/11069 @WangGuangxin @jackylee-ch

Issue #9050, opened 2025-03-18
[CH] Fix bad uri caused by duplicated uri decode


taiyang-li:
### Backend

CH (ClickHouse)

### Bug description

```
org.apache.gluten.exception.GlutenException: Bad URI syntax
0. ./contrib/llvm-project/libcxx/include/__exception/exception.h:106: Poco::Exception::Exception(String const&, String const&, int) @ 0x0000000016b416f9
1. ./build_gcc/./base/poco/Foundation/src/Exception.cpp:145: Poco::URISyntaxException::URISyntaxException(String const&, String const&, int) @ 0x0000000016b4f009
2. ./build_gcc/./base/poco/Foundation/src/URI.cpp:666: Poco::URI::decode(String const&, String&, bool) @ 0x0000000016b95cba
3. ./build_gcc/./utils/extern-local-engine/Storages/SubstraitSource/FormatFile.cpp:119: local_engine::FormatFile::FormatFile(std::shared_ptr<DB::Context const>, substrait::ReadRel_LocalFiles_FileOrFiles const&, std::shared_ptr<local_engine::ReadBufferBuilder> const&) @ 0x000000000d9e0612
4. ./build_gcc/./utils/extern-local-engine/Storages/SubstraitSource/ORCFormatFile.cpp:35: local_engine::ORCFormatFile::ORCFormatFile(std::shared_ptr<DB::Context const>, substrait::ReadRel_LocalFiles_FileOrFiles const&, std::shared_ptr<local_engine::ReadBufferBuilder>) @ 0x000000000da50df5
5. ./contrib/llvm-project/libcxx/include/__memory/construct_at.h:41: local_engine::FormatFileUtil::createFile(std::shared_ptr<DB::Context const>, std::shared_ptr<local_engine::ReadBufferBuilder>, substrait::ReadRel_LocalFiles_FileOrFiles const&) @ 0x000000000d9e1bd9
6. ./build_gcc/./utils/extern-local-engine/Storages/SubstraitSource/SubstraitFileSource.cpp:68: local_engine::SubstraitFileSource::SubstraitFileSource(std::shared_ptr<DB::Context const> const&, DB::Block const&, substrait::ReadRel_LocalFiles const&) @ 0x000000000d9d980f
7. ./contrib/llvm-project/libcxx/include/__memory/construct_at.h:41: local_engine::ReadRelParser::parseReadRelWithLocalFile(substrait::ReadRel const&) @ 0x000000000d75fe6d
[...]

PR #9051, opened 2025-03-18
[GLUTEN-9050][CH] Remove duplicated uri decode which causes runtime exceptions

taiyang-li:

(Fixes: \#9050)

## How was this patch tested?

newly add uts


zzcclp:
it better add a ut to test this case

PR #9056, opened 2025-03-19
[GLUTEN-8437][VL] Fix the exception when verifying the PrestoPage header during the Presto deserialization process

kerwin-zk:

Since the total size of the stream cannot be directly obtained while reading the RSS stream, the current remainingSize represents the remaining size of the current buffer, not the actual remaining size of the entire stream. However, in Velox, there is a validation process for remainingSize. Therefore, this PR is necessary to address the issue.

## How was this patch tested?

CI


kerwin-zk:
@marin-ma  cc

boneanxs (approved):

marin-ma (approved):

marin-ma:
@kerwin-zk Thanks!

Issue #9057, opened 2025-03-19
[VL] Avoid flatten unnecessary vector in ColumnarBatch.select


WangGuangxin:
### Description

Avoid flatten unnecessary vector in ColumnarBatch.select

PR #9058, opened 2025-03-19
[GLUTEN-9057][VL] Avoid flatten unnecessary vector in ColumnarBatch.select

WangGuangxin:

In some cases, `ensureFlattened` is a bit heavy. 

image

While in `ColumnarBatch.select`, not all children of input row vector are needed. 
In some of our test case, this can reduce ~5% vcore times.

(Fixes: \#9057)

## How was this patch tested?

manually

### DIFF
--- a/cpp/velox/memory/VeloxColumnarBatch.cc
+++ b/cpp/velox/memory/VeloxColumnarBatch.cc
@@ -147,17 +147,16 @@ std::shared_ptr<VeloxColumnarBatch> VeloxColumnarBatch::select(
   std::vector<VectorPtr> childVectors;
   childNames.reserve(columnIndices.size());
   childVectors.reserve(columnIndices.size());
-  auto vector = getFlattenedRowVector();
-  auto type = facebook::velox::asRowType(vector->type());
+  auto type = facebook::velox::asRowType(rowVector_->type());
 
   for (uint32_t i = 0; i < columnIndices.size(); i++) {
     auto index = columnIndices[i];
-    auto child = vector->childAt(index);
+    auto child = rowVector_->childAt(index);
     childNames.push_back(type->nameOf(index));
     childVectors.push_back(child);
   }
 
-  auto rowVector = makeRowVector(pool, numRows(), std::move(childNames), vector->nulls(), std::move(childVectors));
+  auto rowVector = makeRowVector(pool, numRows(), std::move(childNames), rowVector_->nulls(), std::move(childVectors));
   return std::make_shared<VeloxColumnarBatch>(rowVector);
 }
 
--- a/cpp/velox/memory/VeloxColumnarBatch.cc
+++ b/cpp/velox/memory/VeloxColumnarBatch.cc
@@ -45,7 +45,7 @@ RowVectorPtr makeRowVector(
 }
 } // namespace
 
-void VeloxColumnarBatch::doFlatten(std::shared_ptr<BaseVector>& child) {
+void VeloxColumnarBatch::doFlatten(VectorPtr& child) {
   facebook::velox::BaseVector::flattenVector(child);
   if (child->isLazy()) {
     child = child->as<facebook::velox::LazyVector>()->loadedVectorShared();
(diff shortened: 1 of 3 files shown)


cpp/velox/memory/VeloxColumnarBatch.cc
```diff
@@ -147,7 +168,7 @@ std::shared_ptr<VeloxColumnarBatch> VeloxColumnarBatch::select(
   std::vector<VectorPtr> childVectors;
   childNames.reserve(columnIndices.size());
   childVectors.reserve(columnIndices.size());
-  auto vector = getFlattenedRowVector();
+  auto vector = getPartialFlattenedRowVector(columnIndices);
```
jinchengchenghh, line 171: I don't know why we need to flatten it, can we remove it?

WangGuangxin: I'm afraid not. `ColumnarBatch.select` are used in several scenario, including `ColumnarPartialProject` and `ColumnarCachedBatchSerializer`. 
The former will do `ensureFlatten` again before converting to arrow by `VeloxColumnarBatch::exportArrowArray`. But since at that time, the input batch is already flattend, so there is no cost.
The latter will access the bytes of input vector directly, so we cannot remove this.

jinchengchenghh: For the former, we can do the flatten for essential columns when converting to arrow.
For the latter, it serializes the batch to bytes, does the serializer only supports flatten batch? I remember PrestoVectorSerde supports all the RowVector.

WangGuangxin: For the latter, it use `UnsafeRowSerializer`, and internally use `DecodedVector` to decode input row vector, while the `DecodedVector` requires that the input should be flat, constant or dictionary, but don't support lazy vector.

jinchengchenghh: So we only need to load lazy vector here.

WangGuangxin: udpate. remove flatten in `ColumnarBatch.select`
1. for the former case, `exportArrowArray` will make sure it's flattend.
2. for the latter case, flatten it before `UnsafeRowSerializer`.

jinchengchenghh: You notice me something, maybe we can support the dictionary vector in columnar to row in the future

jinchengchenghh (approved):

cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -70,7 +70,7 @@ void VeloxColumnarToRowConverter::refreshStates(facebook::velox::RowVectorPtr ro
 
 void VeloxColumnarToRowConverter::convert(std::shared_ptr<ColumnarBatch> cb, int64_t startRow) {
   auto veloxBatch = VeloxColumnarBatch::from(veloxPool_.get(), cb);
-  refreshStates(veloxBatch->getRowVector(), startRow);
```
zhli1142015, line 73: Hello @WangGuangxin 
Could you point out why we need to replace getRowVector with getFlattenedRowVector here? My understanding is that UnsafeRowFast will wrap the passed-in vector into a DecodedVector which can handle all encodings, so there should be no need to call getFlattenedRowVector here — and getFlattenedRowVector may introduce extra vector copy operations when flatting.

zhli1142015: cc @jinchengchenghh 
Do you have any comments for this question?

WangGuangxin: @zhli1142015 at first, this is one codepath of `ColumnarBatch.select`, and it's flattend in `ColumnarBatch.select`. Since `ColumnarBatch.select` does not do flatten, so I move it here. 
But I just checked that `UnsafeRowFast` doesn't require input vector is flatten. So I think it's save to revert change here.

zhli1142015: Get it, thx

PR #9062, opened 2025-03-19
[GLUTEN-9065][VL] Read dynamic offheap sizing configs from SparkEnv

acvictor:
(Fixes: \#9065)

## How was this patch tested?
Manual test, does not repro via UT.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/DynamicOffHeapSizingSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/DynamicOffHeapSizingSuite.scala
@@ -64,7 +64,8 @@ class DynamicOffHeapSizingSuite extends VeloxWholeStageTransformerSuite {
     val data = Seq(1, 2, 3, 4).toDF("col1")
     val result = data.select("col1").distinct().rdd.flatMap(row => Seq(row.getInt(0))).collect()
     val expected = Array(1, 2, 3, 4)
-    assert(result.sorted.sameElements(expected),
+    assert(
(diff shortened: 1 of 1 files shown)


gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
```diff
@@ -30,7 +31,19 @@ public class DynamicOffHeapSizingMemoryTarget implements MemoryTarget {
   private final MemoryTarget delegated;
   // When dynamic off-heap sizing is enabled, the off-heap should be sized for the total usable
   // memory, so we can use it as the max memory we will use.
-  private static final long MAX_MEMORY_IN_BYTES = GlutenConfig.get().offHeapMemorySize();
+  private static final long MAX_MEMORY_IN_BYTES;
+
+  static {
+    SparkEnv sparkEnv = SparkEnv.get();
+    if (sparkEnv != null && sparkEnv.conf() != null) {
```
zhli1142015, line 38: Do we need to check if SparkEnv.get is null?

zhli1142015:
Please add UTs, thanks.

gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
acvictor: Removed thanks.

PR #9066, opened 2025-03-19
[VL] Add API for reserving global off-heap memory from Spark

zhztheplayer:
Add the following common APIs for acquiring / releasing global off-heap memory from Spark:

- GlobalOffHeapMemory.acquire()
- GlobalOffHeapMemory.free()

The API could be used to replace various of current practices to apply for global off-heap memory. E.g., the fake task memory manager used by off-heap broadcasting: https://github.com/apache/incubator-gluten/blob/bf980dcbb8bc912995381bed2ade6dd082ec324a/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala#L151.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.memory
+
+import org.apache.gluten.exception.GlutenException
+
+import org.apache.spark.{SparkEnv, TaskContext}
+import org.apache.spark.storage.BlockId
+
+import java.lang.reflect.Field
+import java.util.UUID
+
+object GlobalOffHeapMemory {
+  private val FIELD_MEMORY_MANAGER: Field = {
+    val f = classOf[TaskMemoryManager].getDeclaredField("memoryManager")
+    f.setAccessible(true)
+    f
+  }
+
+  def acquire(numBytes: Long): Boolean = {
+    memoryManager().acquireStorageMemory(
+      BlockId(s"test_${UUID.randomUUID()}"),
+      numBytes,
+      MemoryMode.OFF_HEAP)
+  }
+
+  def free(numBytes: Long): Unit = {
+    memoryManager().releaseStorageMemory(numBytes, MemoryMode.OFF_HEAP)
+  }
+
+  private def memoryManager(): MemoryManager = {
+    val env = SparkEnv.get
+    if (env != null) {
+      return env.memoryManager
+    }
+    val tc = TaskContext.get()
+    if (tc != null) {
+      return FIELD_MEMORY_MANAGER.get(tc.taskMemoryManager()).asInstanceOf[MemoryManager]
+    }
+    throw new GlutenException(
+      "Memory manager not found because the code is unlikely be run in a Spark application")
+  }
+}
--- a/backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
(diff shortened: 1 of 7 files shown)


backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
```diff
@@ -0,0 +1,57 @@
...
+object GlobalOffHeapMemory {
+  private val FIELD_MEMORY_MANAGER: Field = {
+    val f = classOf[TaskMemoryManager].getDeclaredField("memoryManager")
+    f.setAccessible(true)
+    f
+  }
+
+  def acquire(numBytes: Long): Boolean = {
+    memoryManager().acquireStorageMemory(
+      BlockId(s"test_${UUID.randomUUID()}"),
```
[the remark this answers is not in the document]
zhztheplayer: It intentionally matches RegEx `test_(.*)` to generate a `TestBlockId`. We have to pick an existing `BlockId` from Spark code to use since `BlockId` is a sealed trait and we cannot extend it from Gluten code. 

image

shims/common/src/main/scala/org/apache/spark/storage/BlockManagerUtils.scala
```diff
@@ -0,0 +1,49 @@
...
+      new BlockInfoManager,
+      new SerializerManager(
+        Utils.instantiateSerializerFromConf[Serializer](SERIALIZER, conf, isDriver),
+        conf),
+      memoryManager,
+      new BlockEvictionHandler {
+        override private[storage] def dropFromMemory[T: ClassTag](
+            blockId: BlockId,
+            data: () => Either[Array[T], ChunkedByteBuffer]): StorageLevel = {
+          throw new UnsupportedOperationException(
```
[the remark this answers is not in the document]
zhztheplayer: It's just to set a dummy `MemoryStore` to avoid [this error](https://github.com/apache/spark/blob/aaa94c4e40dfe3dd4f7d05c2388903bad93a2907/core/src/main/scala/org/apache/spark/memory/StorageMemoryPool.scala#L55) during acquiring.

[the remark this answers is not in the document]
zhztheplayer: Or were you asking for the use case of the whole feature?

backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
```diff
@@ -0,0 +1,57 @@
...
+      numBytes,
+      MemoryMode.OFF_HEAP)
```
zhouyuan, line 38: it looks like we are using OFF_HEAP only, not sure if this will impact on the fallback case

backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.spark.{SparkEnv, TaskContext}
+import org.apache.spark.storage.BlockId
+
+import java.lang.reflect.Field
+import java.util.UUID
+
+object GlobalOffHeapMemory {
+  private val FIELD_MEMORY_MANAGER: Field = {
+    val f = classOf[TaskMemoryManager].getDeclaredField("memoryManager")
```
zhouyuan, line 29: should we guard on this, if this call on getField() may fail?

backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.gluten.exception.GlutenException
+
+import org.apache.spark.{SparkEnv, TaskContext}
+import org.apache.spark.storage.BlockId
+
+import java.lang.reflect.Field
+import java.util.UUID
+
+object GlobalOffHeapMemory {
```
zhouyuan, line 27: it seems we are allocating/free on the storage memory in this class, shall we rename with storagememory related?

zhouyuan (commented):
Sorry for the late review - I have some questions.

backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  def free(numBytes: Long): Unit = {
+    memoryManager().releaseStorageMemory(numBytes, MemoryMode.OFF_HEAP)
+  }
+
+  private def memoryManager(): MemoryManager = {
+    val env = SparkEnv.get
+    if (env != null) {
+      return env.memoryManager
```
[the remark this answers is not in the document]
zhztheplayer: > If SparkEnv is not ready, then the following TaskContext() should not be ready too?

The case that `SparkEnv = null && TaskContext != null` is only for test code that may create a mock `TaskContext` as of now.

backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
zhztheplayer: This API is supposed to be called only from Gluten's operators. So we don't need to consider the fallback case yet.

backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
zhztheplayer: Sounds reasonable. I'll update the code to deliver more precise message in the error thrown here in a future PR.

Issue #9073, opened 2025-03-20
[VL] Add support for tail operator


ArnavBalyan:
### Description

CollectTailExec falls back due to lack of support
Implement the same for columnar light batch

PR #9074, opened 2025-03-20
[GLUTEN-9073][VL] Add support for CollectTail Operator

ArnavBalyan:
Fixes: #9073, implementation follows similar to CollectLimitExec

## How was this patch tested?
Unit Tests


gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailBaseExec.scala
```diff
@@ -0,0 +1,56 @@
...
+abstract class ColumnarCollectTailBaseExec(
+    limit: Int,
+    childPlan: SparkPlan
+) extends LimitExec
+  with ValidatablePlan {
+
+  override def outputPartitioning: Partitioning = SinglePartition
+
+  override protected def doValidateInternal(): ValidationResult = {
+    val isSupported = BackendsApiManager.getSettings.supportCollectTailExec()
```
jinchengchenghh, line 35: Don't need to validate again, duplicated with`FallbackByBackendSettings` , may also refactor for ColumnarCollectLimitBaseExec

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
```diff
@@ -0,0 +1,136 @@
...
+
+      override def next(): ColumnarBatch = {
+        if (!hasNext) {
+          throw new NoSuchElementException("No more batches available.")
+        }
+
+        val currentBatch = partitionIter.next()
+        val currentBatchRowCount = currentBatch.numRows()
+        val remaining = limit - rowsCollected
+
```
jinchengchenghh, line 60: What if first batch has 10 rows, second batch has 5 rows, we need tail 6 rows?

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
```diff
@@ -0,0 +1,140 @@
...
+      return Iterator.empty
+    }
+
+    val tailQueue = new mutable.ListBuffer[ColumnarBatch]()
+    var totalRowsInTail = 0L
+
+    while (partitionIter.hasNext) {
+      val batch = partitionIter.next()
+      val batchRows = batch.numRows()
+      ColumnarBatches.retain(batch)
```
jinchengchenghh, line 57: If use the sliced batch, we don't need to retain. Only reuse input batch, we need to retain

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+import scala.collection.mutable
+
+case class ColumnarCollectTailExec(
+    limit: Int,
+    child: SparkPlan
+) extends ColumnarCollectTailBaseExec(limit, child) {
+
+  override def batchType(): Convention.BatchType =
```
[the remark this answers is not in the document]
jinchengchenghh: Please refer to this PR https://github.com/apache/incubator-gluten/pull/9139 to do the refactor

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
jinchengchenghh: Can we use something like this one?
```
Iterators
      .wrap(relation.deserialized)
      .recyclePayload(batch => batch.close())
      .create()
```

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/CollectTailTransformerRule.scala
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.apache.gluten.extension.columnar
+
+import org.apache.gluten.backendsapi.BackendsApiManager
+import org.apache.gluten.config.GlutenConfig
+
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.{CollectTailExec, SparkPlan}
+
+case class CollectTailTransformerRule() extends Rule[SparkPlan] {
```
zhztheplayer, line 25: What's the relationship between this rule and the code added in OffloadSingleNodeRule.scala?

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/validator/Validators.scala
```diff
@@ -137,6 +137,7 @@ object Validators {
       case p: CartesianProductExec if !settings.supportCartesianProductExec() => fail(p)
       case p: TakeOrderedAndProjectExec if !settings.supportColumnarShuffleExec() => fail(p)
       case p: CollectLimitExec if !settings.supportCollectLimitExec() => fail(p)
+      case p: CollectTailExec if !settings.supportCollectTailExec() => fail(p)
```
zhztheplayer, line 140: Is this still needed?

zhztheplayer: Validators are for tagging the node with fallback tags. Does the rule `CollectTailTransformerRule` rely on the fallback tags?

zhztheplayer: > this validation would fail for CH and node should be tagged with false.

I see. So you wanted to make it still add a tag but that tag is not actually used for determining whether the node is offload-able. 

This sounds also reasonable. Would be better to have some comments for this on line :140?

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  private lazy val writeMetrics =
+    SQLShuffleWriteMetricsReporter.createShuffleWriteMetrics(sparkContext)
+
+  private lazy val readMetrics =
+    SQLColumnarShuffleReadMetricsReporter.createShuffleReadMetrics(sparkContext)
+
+  private lazy val useSortBasedShuffle: Boolean =
+    BackendsApiManager.getSparkPlanExecApiInstance
+      .useSortBasedShuffle(outputPartitioning, child.output)
```
jinchengchenghh, line 90: Also here https://github.com/apache/incubator-gluten/pull/9074#discussion_r2055703829 
https://github.com/apache/incubator-gluten/pull/9139/files#diff-25e08fa25b94cc139cb036eac3dd5c958d0aeb7b8570ffa6b020bcae27191a7d

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
```diff
@@ -0,0 +1,142 @@
...
+          val partial = VeloxColumnarBatches.slice(front, overflow.toInt, keep.toInt)
+          tailQueue.prepend(partial)
+          totalRowsInTail -= overflow
+        }
+        front.close()
+      }
+    }
+    Iterators
+      .wrap(tailQueue.iterator)
+      .recyclePayload(batch => batch.close())
```
jinchengchenghh, line 78: Can you try with `recyclePayload` can we remove `ColumnarBatches.retain(batch)`? Can you give some suggestions? @zhztheplayer

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
```diff
@@ -0,0 +1,142 @@
...
+      while (totalRowsInTail > limit && tailQueue.nonEmpty) {
+        val front = tailQueue.remove(0)
+        val frontRows = front.numRows().toLong
+        val overflow = totalRowsInTail - limit
+
+        if (frontRows <= overflow) {
+          totalRowsInTail -= frontRows
+        } else {
+          val keep = frontRows - overflow
+          val partial = VeloxColumnarBatches.slice(front, overflow.toInt, keep.toInt)
```
jinchengchenghh, line 69: Is here possible that some batches should be discarded but slice and then discarded? The strategy should be reconsidered.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
jinchengchenghh: So we don't need Iterators to wrap it if it cannot remove retain input batch

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
jinchengchenghh: I mean can we remove the extra slice cost? Maybe we can delay the slice, make sure the number of batch rows is always greater than the limit, after all the batches are processed, slice the first batch.

jinchengchenghh: No, we don't keep all the batches in the queue, just the first batch with extra rows. Only in the corner case, the first batch is a very very big batch, and we only need less rows.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectTailExec.scala
jinchengchenghh: Just move it to `ColumnarCollectTailExecBase` to let CH backend use. The code is same for CH backend.

jinchengchenghh (approved):

PR #8762, opened 2025-02-18
[GLUTEN-8761][VL] Fix mode in UnsafeColumnarBuildSideRelation not get properly serialize

zjuwangg:
## What changes were proposed in this pull request?

Fix mode field in UnsafeColumnarBuildSideRelation can not get properly serialized and deserialized, also fix when using Java default serializer potential data mismatch by change  `read` to `readFully`.

(Fixes: \#8761)

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)
unit test

PR #9075, opened 2025-03-20
[VL] Acquire off-heap global memory via the new API for off-heap broadcast exchange

zhztheplayer:
A follow-up for #8762 to avoid creating a dummy TaskMemoryManager (which was hack) for reserving off-heap memory for off-heap broadcast exchange. 

The code now calls new API introduced in https://github.com/apache/incubator-gluten/pull/9066 to acquire for off-heap storage memory.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
@@ -41,6 +41,7 @@ import org.apache.spark.unsafe.memory.MemoryAllocator
 @Experimental
 case class UnsafeBytesBufferArray(arraySize: Int, bytesBufferLengths: Array[Int], totalBytes: Long)
   extends Logging {
+  private val allocatedBytes = totalBytes + 7;
   {
     assert(
       arraySize == bytesBufferLengths.length,
@@ -75,8 +76,7 @@ case class UnsafeBytesBufferArray(arraySize: Int, bytesBufferLengths: Array[Int]
     assert(bytesBuffer.length == bytesBufferLengths(index))
     // first to allocate underlying long array
     if (null == longArray && index == 0) {
-      val numBytes = totalBytes + 7
-      if (!GlobalOffHeapMemory.acquire(numBytes)) {
+      if (!GlobalOffHeapMemory.acquire(allocatedBytes)) {
         val memoryManager = SparkEnv.get.memoryManager
         val offHeapMemoryTotal =
           memoryManager.maxOffHeapStorageMemory + memoryManager.offHeapExecutionMemoryUsed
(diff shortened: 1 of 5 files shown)


zhztheplayer:
cc @JkSelf ptal

The new API tends to help with broadcast's off-heap memory management.

zhztheplayer:
cc @zjuwangg

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
```diff
@@ -79,7 +75,17 @@ case class UnsafeBytesBufferArray(
     assert(bytesBuffer.length == bytesBufferLengths(index))
     // first to allocate underlying long array
     if (null == longArray && index == 0) {
-      longArray = allocateArray((totalBytes + 7) / 8)
+      val numBytes = totalBytes + 7
```
jinchengchenghh, line 78: Why the numBytes is not (totalBytes + 7) / 8

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
```diff
@@ -132,8 +138,8 @@ case class UnsafeBytesBufferArray(
   override def finalize(): Unit = {
     try {
       if (longArray != null) {
-        freeArray(longArray)
         longArray = null
+        GlobalOffHeapMemory.release(longArray.size())
```
jinchengchenghh, line 142: call longArray.size() after set it to null

zhztheplayer: Good catch. Thanks.

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
zhztheplayer: `allocateArray` accepts number of long elements so number of bytes will be multiplied by 8

jinchengchenghh (approved):

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
```diff
@@ -48,6 +48,7 @@ case class UnsafeBytesBufferArray(arraySize: Int, bytesBufferLengths: Array[Int]
         "not equal to buffer lengths!")
     assert(totalBytes >= 0, "Unsafe buffer array total bytes can't be negative!")
   }
+  private val allocatedBytes = totalBytes + 7;
```
jinchengchenghh, line 51: I think it is not totalBytes + 7, it is the mutiples of 8

zhztheplayer: I am not sure I am following? `allocatedBytes` is the number of bytes, not the number of long integers.

jinchengchenghh: If better change it in a follow up PR to `totalBytes + 7）/8 * 8`

jinchengchenghh (approved):

backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeBytesBufferArray.scala
zhztheplayer: No problem. I am updating this PR.

zjuwangg:
> cc @zjuwangg

Sorry for the missed message. Glad to see such improvement. I'll investigate this later.

Issue #9078, opened 2025-03-20
Simplify code of SoftAffinity


WangGuangxin:
### Description

Simplify code of SoftAffinity

PR #9079, opened 2025-03-20
[GLUTEN-9078][CORE] Simplify code of SoftAffinity

WangGuangxin:

There are two minor changes.
1. Simple the code when iterate over `allInputPartitions`, which is a bit confusion.

```
    allInputPartitions.head.indices.foreach(
      i => {
        val currentPartitions = allInputPartitions.map(_(i))
        currentPartitions.indices.foreach(
          i =>
            currentPartitions(i) match {
              case f: FilePartition =>
                SoftAffinity.updateFilePartitionLocations(f, rdd.id)
              case _ =>
            })
      })
```

2.  Don't need to iterate and check whether `SoftAffinityManager.usingSoftAffinity` are enabled every time inside loop

(Fixes: \#9078)

## How was this patch tested?

Existing UT

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
@@ -39,7 +39,6 @@ import org.apache.spark.sql.catalyst.expressions.{Attribute, SortOrder}
 import org.apache.spark.sql.catalyst.plans.physical.Partitioning
 import org.apache.spark.sql.connector.read.InputPartition
 import org.apache.spark.sql.execution._
-import org.apache.spark.sql.execution.datasources.FilePartition
 import org.apache.spark.sql.execution.metric.SQLMetric
 import org.apache.spark.sql.utils.SparkInputMetricsUtil.InputMetricsWrapper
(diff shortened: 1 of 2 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -421,17 +420,7 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
...
-        val currentPartitions = allInputPartitions.map(_(i))
-        currentPartitions.indices.foreach(
-          i =>
-            currentPartitions(i) match {
-              case f: FilePartition =>
-                SoftAffinity.updateFilePartitionLocations(f, rdd.id)
-              case _ =>
-            })
-      })
+    SoftAffinity.updateFilePartitionLocations(Seq(allInputPartitions), rdd.id)
```
jackylee-ch, line 423: Since we don't need IndexSeq, could you remove the `.toIndexedSeq` in https://github.com/apache/incubator-gluten/pull/9079/files#diff-5690472af86b8a7836d5b2e167d22ce95953758c29f9889d0a25515fc7f07aa4R385

Issue #9087, opened 2025-03-21
Make substrait node not depend on spark


shuai-xu:
### Description

No some substrait nodes depend on spark's class, such as StructLiteralNode and so on. To make substrait module can be used in both gluten spark and flink, we need to make these classes  not depend on spark. This is part of work for [issue-9046](https://github.com/apache/incubator-gluten/issues/9046).

PR #9088, opened 2025-03-21
[GLUTEN-9087] [Core] Make substrait node not depend on spark

shuai-xu:

This pr make the substrait related class not depend on spark. So it can be reused by flink.

(Fixes: \#9087)

## How was this patch tested?

This patch was tested by unit tests and existing integration tests

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
@@ -25,7 +25,7 @@ import org.apache.gluten.metrics.IMetrics
 import org.apache.gluten.sql.shims.{DeltaShimLoader, SparkShimLoader}
 import org.apache.gluten.substrait.plan.PlanNode
 import org.apache.gluten.substrait.rel._
-import org.apache.gluten.substrait.rel.LocalFilesNode.{ReadFileFormat, SchemaField}
+import org.apache.gluten.substrait.rel.LocalFilesNode.{FileSchema, ReadFileFormat, SchemaField}
 import org.apache.gluten.vectorized.{BatchIterator, CHNativeExpressionEvaluator, CloseableCHColumnBatchIterator}
 
 import org.apache.spark.{InterruptibleIterator, SparkConf, TaskContext}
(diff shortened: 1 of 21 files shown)


baibaichen:
@CodiumAI-Agent /review

gluten-substrait/src/main/java/org/apache/gluten/utils/SparkToJavaConverter.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class SparkToJavaConverter {
+
+  private static Object toJava(Object obj, TypeNode nodeType) {
+    return toJava(obj, nodeType, null);
+  }
+
+  public static Object toJava(Object obj, TypeNode nodeType, DataType dataType) {
```
zhztheplayer, line 57: Signature of the API seems like too general? `Object` used in both parameter list and return type. Is there a chance we can use strong types? Can we leverage visitor pattern?

zhztheplayer: @shuai-xu Could check this https://github.com/apache/spark/blob/e428fe902bb1f12cea973de7fe4b885ae69fd6ca/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/CatalystTypeConverters.scala#L489-L500

shuai-xu: This class also has a same signature `def convertToScala(catalystValue: Any, dataType: DataType): Any = `, it can not be used in gluten, or else will throw exceptions. And it return a spark Row for StructType.

backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHIteratorApi.scala
```diff
@@ -55,18 +55,24 @@ import scala.collection.mutable.ArrayBuffer
 
 class CHIteratorApi extends IteratorApi with Logging with LogLevelUtil {
 
-  private def getFileSchema(schema: StructType, names: Seq[String]): StructType = {
-    val dataSchema = ArrayBuffer[StructField]()
+  private def getFileSchema(schema: StructType, names: Seq[String]): java.util.List[SchemaField] = {
```
PHILO-HE, line 58: Can we just create a class type that holds multiple field elements? Then, directly replace StructType with the new class type.

gluten-substrait/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java
PHILO-HE, line 1: I am afraid that these code for constructing window node cannot be reused by Flink. If true, do we need these code changes for removing the dependency on Spark? I am thinking whether it is reasonable to move it to a submodule also.

PHILO-HE (commented):
Post some thoughts for discussion. Thanks!

PR #9096, opened 2025-03-21
[GLUTEN-9095][UT] Remove Vanilla Spark InternalRow based checkEvaluation

ArnavBalyan:
 - GlutenCheckExpression currently falls back to internal row for certain complex/unsupported types.
 - As discussed in the PR, this is running Vanilla Spark and does not mean anything for Gluten.
 - Removing the same
## How was this patch tested?
 - Unit Tests

### DIFF
--- a/gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
+++ b/gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
@@ -26,7 +26,6 @@ import org.apache.spark.sql.GlutenQueryTestUtil.isNaNOrInf
 import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
 import org.apache.spark.sql.catalyst.analysis.ResolveTimeZone
 import org.apache.spark.sql.catalyst.expressions._
-import org.apache.spark.sql.catalyst.expressions.codegen.GenerateUnsafeProjection
 import org.apache.spark.sql.catalyst.optimizer.{ConstantFolding, ConvertToLocalRelation, NullPropagation}
 import org.apache.spark.sql.catalyst.util.{ArrayData, GenericArrayData, MapData, TypeUtils}
 import org.apache.spark.sql.internal.SQLConf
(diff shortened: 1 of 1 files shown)


gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
```diff
@@ -138,13 +138,7 @@ trait GlutenTestsTrait extends GlutenTestsCommonTrait {
...
     } else {
       logWarning(s"The status of this unit test is not guaranteed.")
-      val catalystValue = CatalystTypeConverters.convertToCatalyst(expected)
-      checkEvaluationWithoutCodegen(expr, catalystValue, inputRow)
-      checkEvaluationWithMutableProjection(expr, catalystValue, inputRow)
-      if (GenerateUnsafeProjection.canSupport(expr.dataType)) {
-        checkEvaluationWithUnsafeProjection(expr, catalystValue, inputRow)
-      }
-      checkEvaluationWithOptimization(expr, catalystValue, inputRow)
+      checkEvaluationWithFallback(expression, expected, inputRow)
```
jackylee-ch, line 141: Why do we need this extra functions? They looks same as before.

gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
```diff
@@ -361,4 +355,20 @@ trait GlutenTestsTrait extends GlutenTestsCommonTrait {
...
   }
+
+  protected def checkEvaluationWithFallback(
+      expression: => Expression,
+      expected: Any,
+      inputRow: InternalRow = EmptyRow): Unit = {
+    val resolver = ResolveTimeZone
+    val expr = resolver.resolveTimeZones(expression)
+    val catalystValue = CatalystTypeConverters.convertToCatalyst(expected)
+    checkEvaluationWithoutCodegen(expr, catalystValue, inputRow)
```
jackylee-ch, line 366: It seems these function are testing vaniila Cast, not related to gluten. Why we need it here?

jackylee-ch: I'm trying to reproduce the problem with `checkExecution`, but the local compilation environment has some problems. The tests added to gluten should be related to testing gluten trasformer. If a case is just for testing vanilla, there is really no need to add it.
cc @PHILO-HE

jackylee-ch: For now, It seems useless. We may consider adding it later once we have a better reason.

jackylee-ch (approved):
Thanks.

gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
```diff
@@ -136,15 +135,6 @@ trait GlutenTestsTrait extends GlutenTestsCommonTrait {
 
     if (canConvertToDataFrame(inputRow)) {
       glutenCheckExpression(expr, expected, inputRow)
-    } else {
```
jackylee-ch, line 139: Could you add a comment to "checkEvaluation" to indicate that the function only verifies the native result and skips the vanilla check.

PHILO-HE: Maybe, it would be better to add a log to indicate the evaluation is skipped.

PHILO-HE:
This patch may also have an impact on CH backend test. @zzcclp, could you also take a look?

gluten-ut/common/src/test/scala/org/apache/spark/sql/GlutenTestsTrait.scala
```diff
@@ -130,21 +129,17 @@ trait GlutenTestsTrait extends GlutenTestsCommonTrait {
...
-      logWarning(s"The status of this unit test is not guaranteed.")
-      val catalystValue = CatalystTypeConverters.convertToCatalyst(expected)
-      checkEvaluationWithoutCodegen(expr, catalystValue, inputRow)
-      checkEvaluationWithMutableProjection(expr, catalystValue, inputRow)
-      if (GenerateUnsafeProjection.canSupport(expr.dataType)) {
-        checkEvaluationWithUnsafeProjection(expr, catalystValue, inputRow)
-      }
-      checkEvaluationWithOptimization(expr, catalystValue, inputRow)
+      logWarning(
+        "Skipping evaluation - Expression could not be converted " +
```
PHILO-HE, line 141: Suggestion:

Skipping evaluation - Nonempty inputRow cannot be converted to DataFrame due to **

PHILO-HE (approved):
Thanks!

PR #9107, opened 2025-03-24
[GLUTEN-9106][VL] Add support for staticInvoke CharVarcharCodegenUtils. #9107

Yifeng-Wang:

To add support for static function callls in Spark's CharVarcharCodegenUtils.java, we add expr matching pattern in Gluten ExpressionConverter.

(Fixes: \#9106)

## How was this patch tested?

Added test cases in Gluten's ScalarFunctionsValidateSuite.
Built Gluten Velox with Spark 3.4 (Spark3.4.4 in pom.xml default) and test sqls.

for test_varcharWriteSideCheck: 
```sql
create table srcvarchar(id string) stored as parquet;
create table tgt_vanilla(id varchar(3)) stored as parquet;
create table tgt(id varchar(3)) stored as parquet;

create table srcchar(id string) stored as parquet;
create table tgt_vanilla(id char(3)) stored as parquet;
create table tgt(id char(3)) stored as parquet;

insert into srcvarchar values ('a');
insert into srcvarchar values ('ab');
insert into srcvarchar values ('t  '); -- trail space but should retain
insert into srcvarchar values ('tra '); -- trail space but should trim
insert into srcvarchar values ('中  '); -- trailing space but should retain
insert into srcvarchar values ('中   '); -- trailing space but should trim to 3
insert into srcvarchar values ('中文中国'); -- will fail

insert into tgt select id from srcvarchar
```

for charWriteSideCheck: 
```sql
insert into srcchar values ('a');
insert into srcchar values ('ab');
insert into srcchar values ('快');
insert into srcchar values ('捷  ');
insert into tgt_vanilla select id from srcchar where id='abcd';
```

check plan to validate such operations offload to native and do not cause operator fallback:
3668dcde999255823c0bfcfc9f55970

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
@@ -164,6 +164,32 @@ object ExpressionConverter extends SQLConfHelper with Logging {
                 child.map(replaceWithExpressionTransformer0(_, attributeSeq, expressionsMap)),
                 i)
           }
+        } else if (objectName.endsWith("CharVarcharCodegenUtils")) {
+          val children = i.arguments
+          i.functionName match {
+            case "varcharTypeWriteSideCheck" =>
+              return GenericExpressionTransformer(
+                ExpressionNames.VARCHAR_TYPE_WRITE_SIDE_CHECK,
+                children
+                  .map(replaceWithExpressionTransformer0(_, attributeSeq, expressionsMap))
+                  .toSeq,
+                i)
+            case "charTypeWriteSideCheck" =>
+              return GenericExpressionTransformer(
+                ExpressionNames.CHAR_TYPE_WRITE_SIDE_CHECK,
+                children
+                  .map(replaceWithExpressionTransformer0(_, attributeSeq, expressionsMap))
+                  .toSeq,
+                i)
+            case "readSidePadding" =>
+              // println(s"enters readSidePadding")
+              return GenericExpressionTransformer(
+                ExpressionNames.READ_SIDE_PADDING,
+                children
+                  .map(replaceWithExpressionTransformer0(_, attributeSeq, expressionsMap))
+                  .toSeq,
+                i)
+          }
         }
       case _ =>
     }
--- a/shims/common/src/main/scala/org/apache/gluten/expression/ExpressionNames.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/expression/ExpressionNames.scala
@@ -356,4 +356,9 @@ object ExpressionNames {
   // A placeholder for native UDF functions
   final val UDF_PLACEHOLDER = "udf_placeholder"
   final val UDAF_PLACEHOLDER = "udaf_placeholder"
+
+  // Spark Catalyst util functions
+  final val VARCHAR_TYPE_WRITE_SIDE_CHECK = "varchar_type_write_side_check"
+  final val CHAR_TYPE_WRITE_SIDE_CHECK = "char_type_write_side_check"
+  final val READ_SIDE_PADDING = "read_side_padding"
 }
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/ScalarFunctionsValidateSuite.scala
@@ -854,6 +854,47 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
     }
   }
 
+  // Add test suite for CharVarcharCodegenUtils functions.
+  // A ProjectExecTransformer is expected to be constructed after expr support.
+  // We currently test below functions with Spark v3.4
+  testWithSpecifiedSparkVersion("Test charTypeWriteSideCheck function", Some("3.4")) {
+    withTable("src", "dest") {
+
+      sql("create table src(id string) USING PARQUET")
+      sql("insert into src values('s')")
+      sql("create table dest(id char(3)) USING PARQUET")
+      // check whether the executed plan of a dataframe contains the expected plan.
+      runQueryAndCompare("insert into dest select id from src") {
+        checkGlutenOperatorMatch[ProjectExecTransformer]
+      }
+    }
+  }
+
+  testWithSpecifiedSparkVersion("Test varcharTypeWriteSideCheck function", Some("3.4")) {
+    withTable("src", "dest") {
+
+      sql("create table src(id string) USING PARQUET")
+      sql("insert into src values('abc')")
+      sql("create table dest(id varchar(10)) USING PARQUET")
+      // check whether the executed plan of a dataframe contains the expected plan.
+      runQueryAndCompare("insert into dest select id from src") {
+        checkGlutenOperatorMatch[ProjectExecTransformer]
+      }
+    }
+  }
+
+  testWithSpecifiedSparkVersion("Test readSidePadding function", Some("3.4")) {
+    withTable("src", "dest") {
+
+      sql("create table tgt(id char(3)) USING PARQUET")
+      sql("insert into tgt values('p')")
+      // check whether the executed plan of a dataframe contains the expected plan.
+      runQueryAndCompare("select id from tgt") {
+        checkGlutenOperatorMatch[ProjectExecTransformer]
+      }
+    }
+  }
+
   test("Test hex function") {
     runQueryAndCompare("SELECT hex(l_partkey), hex(l_shipmode) FROM lineitem limit 1") {
       checkGlutenOperatorMatch[ProjectExecTransformer]


Yifeng-Wang:
@zzcclp @jinchengchenghh Hi, dear Zhichao and Jincheng, could you please review this PR? It addresses issue #9106 .
We are now creating a relative issue & PR at velox side, and we look forward to your feedbacks as well.
Thanks in advance, best regards,

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -164,6 +164,32 @@ object ExpressionConverter extends SQLConfHelper with Logging {
                 child.map(replaceWithExpressionTransformer0(_, attributeSeq, expressionsMap)),
                 i)
           }
+        } else if (objectName.endsWith("CharVarcharCodegenUtils")) {
```
jinchengchenghh, line 167: Please extract the logic for case i: StaticInvoke => to a new function, here will be much code

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -109,6 +109,66 @@ object ExpressionConverter extends SQLConfHelper with Logging {
     }
   }
 
+  private def replaceStaticInvokeWithExpressionTransformer(
+      i: StaticInvoke,
+      attributeSeq: Seq[Attribute],
+      expressionsMap: Map[Class[_], String]): ExpressionTransformer = {
+    val objectName = i.staticObject.getName.stripSuffix("$")
+
+    if (objectName.endsWith("UrlCodec")) {
```
jinchengchenghh, line 118: Please use this format to simplify the code, and easy to implement others
```
val staticInvokeMap = Map("UrlCodec"-> Map("decode" -> ExpressionNames.URL_DECODE, "encode"-> ExpressionNames.URL_ENCODE ))
 val transformer = staticInvokeMap.filter(o => objectName.endsWith(o._1))
if (transformer.isEmpty) {
      throw
    }
```

shims/common/src/main/scala/org/apache/gluten/expression/ExpressionNames.scala
```diff
@@ -356,4 +356,9 @@ object ExpressionNames {
   // A placeholder for native UDF functions
   final val UDF_PLACEHOLDER = "udf_placeholder"
   final val UDAF_PLACEHOLDER = "udaf_placeholder"
+
+  // Spark Catalyst util functions
```
jinchengchenghh, line 360: Spark StaticInvoke Catalyst util functions

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
jinchengchenghh: I think most of the cases we need all the argument, UrlCodec is a special case, maybe because the backend only supports "UTF-8" , we should throw exception if 2nd argument is not "UTF-8"  but not.

So other common cases should use the `staticInvokeMap`, that can also help developer figure out the special case.

jinchengchenghh:
Yes, you don't need to support it, just make sure it is fallback in CH backend.
The CH CI user and password is gluten/hN2xX3uQ4m, please resolve it.
Like the ExpressionMapping function `extraExpressionMappings` for each backend, we may need the  `extraStaticInvokeMappings`, I suppose the failed CH CI is cause by this, native validation does not fallback.

jinchengchenghh:
The password is in the CH backend document, but hard to find

jinchengchenghh:
Just add the new functions name to CH_BLACKLIST_SCALAR_FUNCTION like URL_DECODE

jinchengchenghh:
Yes, I think so. Add the substrait name check in replaceStaticInvokeWithExpressionTransformer@taiyang-li Do you have some suggestions that make CH backend fallback these unsupported functions

jinchengchenghh:
These function are not merged in velox, we cannot add a unit test to it, can we merge these function native implement first?

Yifeng-Wang:
> This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.

Current PR awaits the PR merge at Velox side and is not stale.

Yifeng-Wang:
> This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.

This work will be done this week.

Yifeng-Wang:
resolving conflicts in progress.

Yifeng-Wang:
**Summary** (2025-08-25 SITREP)

##### **What has been done**

**Ⅰ. Rebased previous proposal onto current main branch progress**

Rebased code onto current main branch. This PR still extractS logics for `case i: StaticInvoke` branch into a separated `replaceStaticInvokeWithExpressionTransformer`, but now an extra `validateandTransform` is introduced to do the backend validation. The logic remain the same which ensures a Transformer is instantiated only when the underlying backend supports this expression.



**Ⅱ.  Tests for the new round**

 - **Gluten Test Suite**

run `mvn clean install -Pbackends-velox -Pspark-3.5 -DwildcardSuites="org.apache.gluten.functions.ScalarFunctionsValidateSuite"`

(-Pspark-3.4 for Spark v3.4) . 

**Validation**:

Got BUILD SUCCESS and checked surefire test reports "TEST-org.apache.gluten.functions.ScalarFunctionsValidateSuite*" all pass.





- **Spark + Gluten + Velox tests** 

Test for all current supported StaticInvoke Expressions.

```SQL
-- url_decode
create table url_tbl(id string) stored as parquet;
insert into url_tbl values ('https%3A%2F%2Fspark.apache.org');
select url_decode(id) from url_tbl; -- https://spark.apache.org
-- url_encode
create table url_tbl2(id string) stored as parquet;
insert into url_tbl2 values ('https://spark.apache.org');
select url_encode(id) from url_tbl2;   -- https%3A%2F%2Fspark.apache.org

-- luhn_check, min Spark>=3.5
select luhn_check(id) from url_tbl;

-- base64 and unbase64
select unbase64(base64(id)) from url_tbl;

-- CharVarcharCodegenUtils functions (this PR)
create table srcvarchar(id string) stored as parquet;
insert into srcvarchar values ('ab');
create table tgtvarchar(id varchar(3)) stored as parquet;
create table tgtchar(id char(3)) stored as parquet;
-- varcharTypeWriteSideCheck
insert into tgtvarchar select id from srcvarchar;
-- charTypeWriteSideCheck
insert into tgtchar select id from srcvarchar;
-- readSidePadding
SELECT id FROM tgtchar;
```

- **Validation:** 

Verify Transformers within SQL execution plans without fallbacks and results correct.

Yifeng-Wang:
Hi @jinchengchenghh and @rui-mo!

I've just pushed the latest updates to this PR and have included a summary above to summarize the current status.

I understand this PR has been through a lengthy process with a pause in the middle, and I appreciate your patience and valuable feedback thus far. It's now ready for a second round of review.

When you have a moment, could you please take a look at the recent changes? Your feedback would be greatly appreciated.

Thanks! :)

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,6 +144,67 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+        exprName: String,
+        childTransformers: => Seq[ExpressionTransformer]): ExpressionTransformer = {
+      if (!BackendsApiManager.getValidatorApiInstance.doExprValidate(exprName, i)) {
+        throw new GlutenNotSupportException(
+          s"Not supported to map current ${i.getClass} call on function: ${i.functionName}.")
+      }
+      GenericExpressionTransformer(exprName, childTransformers, i)
+    }
+
+    if (Seq("encode", "decode").contains(i.functionName) && i.objectName.endsWith("UrlCodec")) {
```
rui-mo: Are the comparisons case-sensitive?

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,6 +144,67 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+      return validateAndTransform(
+        "url_" + i.functionName,
+        Seq(replaceWithExpressionTransformer0(i.arguments.head, attributeSeq, expressionsMap))
+      )
+    }
+
+    if (i.functionName == "isLuhnNumber") {
```
rui-mo: ditto

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,6 +144,67 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+      return validateAndTransform(
+        ExpressionNames.LUHN_CHECK,
+        Seq(replaceWithExpressionTransformer0(i.arguments.head, attributeSeq, expressionsMap))
+      )
+    }
+
+    if (Seq("encode", "decode").contains(i.functionName) && i.objectName.endsWith("Base64")) {
```
rui-mo: ditto

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,6 +144,67 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+      }
+      return BackendsApiManager.getSparkPlanExecApiInstance.genBase64StaticInvokeTransformer(
+        ExpressionNames.BASE64,
+        replaceWithExpressionTransformer0(i.arguments.head, attributeSeq, expressionsMap),
+        i
+      )
+    }
+
+    if (
+      Set("varcharTypeWriteSideCheck", "charTypeWriteSideCheck", "readSidePadding").contains(
```
rui-mo: ditto

backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
```diff
@@ -622,6 +622,47 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
     }
   }
 
+  // Add test suite for CharVarcharCodegenUtils functions.
+  // A ProjectExecTransformer is expected to be constructed after expr support.
+  // We currently test below functions with Spark v3.4
+  testWithMinSparkVersion("Test charTypeWriteSideCheck function", "3.4") {
```
rui-mo: nit: could be rename as "charTypeWriteSideCheck" to be consistent in style with other cases. And are these functions only supported after Spark 3.4?

Yifeng-Wang:
Hi @rui-mo !  I saw spark-tests fail and test fails were due to Exception type/message mismatches, not because of incorrect functionality, e.g.: the test expect "did not contain "Exceeds char/varchar type length limitation: 5", but actually got: "(6 vs. 5) Exceeds allowed length limitation: 5", 

image
image

Since we should not touch Spark's tests and should remain the error msg from Velox side, I wonder the correct way we handle current situation is to exclude the vanilla spark tests in the `VeloxTestSettings.scala`, but enable and implement the test suite at Gluten's test suite.
Since only some cases have the error msg mismatch, there should be both err msgs, I wonder if could add 
```
  private val ERROR_MESSAGE =
    "Exceeds char/varchar type length limitation: 5"
  private val VELOX_ERROR_MESSAGE = "Exceeds allowed length limitation: 5"
```
and then assert `VELOX_ERROR_MESSAGE` ?
Thanks for your help! 🤝

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: Yes, both the codes in origin main branch and this PR remain case-sensitive because they're comparing against actual Java method and class names.
image

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: Yes, case-sensitive is intentional. Same as above.
image

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: Yes, case-sensitive is intentional. Same as above.
image

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: Yes, case-sensitive is intentional. Same as above.
image

backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
Yifeng-Wang: Yes, all these functions should be tested with min Spark version 3.4, 
1. `readSidePadding`
was introduced at `branch-3.4` in `sql/catalyst/src/main/java/org/apache/spark/sql/catalyst/util/CharVarcharCodegenUtils.java`.
2. `charTypeWriteSideCheck` & `varcharTypeWriteSideCheck`
prior to 3.4, Spark's analyzer introduces an extra redundant cast(... as string) expression. The current transformer for StaticInvoke does not handle this specific pattern with the extra cast, which causes a fallback to the vanilla Spark ProjectExec.
image

rui-mo:
@Yifeng-Wang Yes, for the error msg mismatch we usually exclude Spark's test, and rewrite it in the Gluten suite.

Yifeng-Wang:
Hi @rui-mo , the tests fails are now solved for each of the Spark's version. The test suites are updated to make correct message assertions. Would you please help me take a look at the changes when you have time? Thanks!

backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
```diff
@@ -622,6 +622,47 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
     }
   }
 
+  // Add test suite for CharVarcharCodegenUtils functions.
+  // A ProjectExecTransformer is expected to be constructed after expr support.
+  // We currently test below functions with Spark v3.4
+  testWithMinSparkVersion("charTypeWriteSideCheck", "3.4") {
```
jinchengchenghh, line 628: Why only test with Spark3,4? Does Spark 3.3 not have this function?

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,29 +144,89 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+        exprName: String,
+        childTransformers: => Seq[ExpressionTransformer]): ExpressionTransformer = {
+      if (!BackendsApiManager.getValidatorApiInstance.doExprValidate(exprName, i)) {
+        throw new GlutenNotSupportException(
+          s"Not supported to map current ${i.getClass} call on function: ${i.functionName}.")
+      }
+      GenericExpressionTransformer(exprName, childTransformers, i)
+    }
+
+    if (Seq("encode", "decode").contains(i.functionName) && i.objectName.endsWith("UrlCodec")) {
```
jinchengchenghh: Could you use case match for the function name? This will be more clean

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,29 +144,89 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+      }
+      return validateAndTransform(
+        exprName,
+        i.arguments.map(replaceWithExpressionTransformer0(_, attributeSeq, expressionsMap))
+      )
     }
-    if (transformer.isEmpty) {
-      throw new GlutenNotSupportException(s"Not supported staticInvoke call object: $objName")
+
+    if (
```
jinchengchenghh: Please do not expand the replaceIcebergStaticInvoke here, this function is a bit long and will be longger

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
```diff
@@ -674,7 +674,20 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude("length check for input string values: nested in both map key and value")
     .exclude("length check for input string values: nested in array of struct")
     .exclude("length check for input string values: nested in array of array")
+    .exclude("length check for input string values: top-level columns")
+    .exclude("length check for input string values: partitioned columns")
+    .exclude("length check for input string values: nested in struct of array")
+    .exclude("length check for input string values: with implicit cast")
+    .exclude("char/varchar type values length check: partitioned columns of other types")
   enableSuite[GlutenDSV2CharVarcharTestSuite]
+    .exclude("length check for input string values: top-level columns")
```
jinchengchenghh, line 691: Is there only the error message difference? Is there any change to align the error message?

jinchengchenghh: Looks like even if we align the error message, we still need to copy the tests, so let us keep current solution.

backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
Yifeng-Wang: Hi, thanks for pointing out. From my obervation turns out that although Spark 3.3 does have this function, the execution plan is different, not triggering `staticInvoke`, thus tested with min=Spark 3.4

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
Yifeng-Wang: Ok. Keep current solution.

jinchengchenghh:
Please resolve the conflict, thanks!

Yifeng-Wang:
> Please resolve the conflict, thanks!

Yes I am doing it right now 🤝

Yifeng-Wang:
The latest commit added `excludeGlutenTest` to ClickHouseTestSettings.scala across different versions. The added expression support feature in this PR is only supported by Velox backend, while for ClickHouse backend, it fallbacks and thus should not match the expected err msg at gluten test suite.

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/clickhouse/ClickHouseTestSettings.scala
```diff
@@ -848,6 +856,14 @@ class ClickHouseTestSettings extends BackendTestSettings {
     .includeCH("length check for input string values: nested in both map key and value")
     .includeCH("length check for input string values: nested in array of struct")
     .includeCH("length check for input string values: nested in array of array")
+    .excludeGlutenTest("length check for input string values: top-level columns")
```
rui-mo, line 865: Would you please add comments above to illustrate the reason to exclude these tests (because they are rewritten due to ...)?

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,6 +144,66 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+        validateAndTransform(
+          ExpressionNames.LUHN_CHECK,
+          Seq(replaceWithExpressionTransformer0(i.arguments.head, attributeSeq, expressionsMap))
+        )
+
+      case "encode" | "decode" if i.objectName.endsWith("Base64") =>
+        if (!BackendsApiManager.getValidatorApiInstance.doExprValidate(ExpressionNames.BASE64, i)) {
+          throw new GlutenNotSupportException(
+            s"Not supported to map current ${i.getClass} call on function: ${i.functionName}.")
+        }
```
rui-mo, line 178: Thanks for the refactor. This validations seem to be newly introduced, would you please clarify a bit?

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -144,6 +144,66 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+
+      case fn
+          if i.objectName.endsWith("CharVarcharCodegenUtils") && Set(
+            "varcharTypeWriteSideCheck",
+            "charTypeWriteSideCheck",
+            "readSidePadding").contains(fn) =>
+        val exprName = fn match {
+          case "varcharTypeWriteSideCheck" => ExpressionNames.VARCHAR_TYPE_WRITE_SIDE_CHECK
+          case "charTypeWriteSideCheck" => ExpressionNames.CHAR_TYPE_WRITE_SIDE_CHECK
+          case "readSidePadding" => ExpressionNames.READ_SIDE_PADDING
```
rui-mo, line 193: The mappings from Spark's name to Velox's name are more commonly to be put at https://github.com/apache/incubator-gluten/blob/main/cpp/velox/substrait/SubstraitParser.cc#L387 to allow compatibility among different backends.

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: Hi rui-mo, thanks for mentioning this. Apologize for such a long context within this PR.
This `doExprValidate` call on validator instance was added deliberately to ensure backend actually supports.
Technically this was from my earlier findings (if you scroll up a bit, on Mar 26) , and currently it ensures that:
- An explicit call on `doExprValidate`.
- `CH_BLACKLIST_SCALAR_FUNCTION` is looked through during e.g., `CHValidatorApi::doExprValidate` so that the functions which are ONLY supported at Velox side, won't cause  "Unknown function parser" for clickhouse BE(in CI tests).
- Intercept at Gluten with `GlutenNotSupportException`, instead of sending to actual unsupported (ClickHouse) backend.

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: @rui-mo Hi rui-mo, could you help give me some guidance on how I should update the codes?
Currently the gluten-velox run of my expression convert does not seems like relying on this map?

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/clickhouse/ClickHouseTestSettings.scala
Yifeng-Wang: will do now.

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: To add some context, 
at Gluten side in `ExpressionNames.scala`. I had e.g.:
`final val VARCHAR_TYPE_WRITE_SIDE_CHECK = "varchar_type_write_side_check"` .
And at velox side, there is
```
registerFunction<
VarcharTypeWriteSideCheckFunction,
Varchar,
Varchar,
int32_t>
({prefix + "varchar_type_write_side_check"});
```
so I guess substrait function name (from `ExpressionNames.VARCHAR_TYPE_WRITE_SIDE_CHECK = "varchar_type_write_side_check"`) already matches the Velox registered function name (prefix + "varchar_type_write_side_check").

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
rui-mo: Thanks, I understand.

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
rui-mo: Typically, in Gluten’s ExpressionNames.scala, the Spark function names are defined, while in the C++ map, the corresponding function names for a specific backend are provided. For instance, the Spark function add might be represented as plus in certain backends. To maintain consistency, Gluten’s Scala side keeps Spark’s original naming (e.g., add), and it is the responsibility of each backend to map it to the appropriate equivalent name. This is just a minor suggestion—please feel free to decide whether you’d like to make this change.

rui-mo (approved):

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
Yifeng-Wang: Hi rui-mo! ​Thanks for clarifying the reasoning behind this design. Maybe I'll keep the current state for this PR, but I'll be sure to follow this mapping mechanism in subsequent PR. Really appreciate the guidance!

Yifeng-Wang:
Hi @rui-mo , @jinchengchenghh !

Thank you so much for your thorough review and patience on this long-running PR. Your feedback was invaluable. 
Greatly appreciated!

Issue #9113, opened 2025-03-24
[VL] remove unused function mapping for "not_equal" to "notequalto"


kevincmchen:
### Description

The function mapping {"not_equal", "notequalto"} should be removed because:

1. Spark already converts not_equal(!=) to  `Not` combined with `EqualTo` (effectively NOT (a = b)). This means Gluten will never receive a `not_equal` function call requiring translation to `notequalto`.  Here's a example:

- sql:
   `
SELECT i_item_id, i_item_sk, MAX(i_wholesale_cost)
FROM item  
GROUP BY GROUPING SETS((i_item_id, i_item_sk))  
HAVING UPPER(i_item_id) != 'AAAAAAAACAAAAAAA' 
ORDER BY 1,2
LIMIT 10;
`

- plan:
     <img width="1716" alt="Image" src="https://github.com/user-attachments/assets/91efef11-c161-4dec-8d96-0e36bc62b230" />

2. Velox doesn't implement the `notequalto` function, making this mapping non-functional even if invoked.

PR #9114, opened 2025-03-24
[GLUTEN-9113][VL]  Remove Unused not_equal Function Mapping

kevincmchen:

Spark internally rewrites (a != b) as NOT (a = b), this meas spark never use the mapping(not_equal -> notequalto). velox does not implement a native notequalto function. so this maping should be removed.

(Fixes: \#9113)

### DIFF
--- a/cpp/velox/substrait/SubstraitParser.cc
+++ b/cpp/velox/substrait/SubstraitParser.cc
@@ -393,7 +393,6 @@ std::unordered_map<std::string, std::string> SubstraitParser::substraitVeloxFunc
     {"lte", "lessthanorequal"},
     {"gt", "greaterthan"},
     {"gte", "greaterthanorequal"},
-    {"not_equal", "notequalto"},
     {"char_length", "length"},
     {"strpos", "instr"},
     {"ends_with", "endswith"},


jinchengchenghh:
I would prefer keep notequalto, because we should so some optimization in Velox to support this expression instead of using 2 operators. Is it that? @rui-mo

rui-mo:
> I would prefer keep notequalto, because we should so some optimization in Velox to support this expression instead of using 2 operators. Is it that? @rui-mo

Currently we are implementation Velox functions as a one-one mapping way, which means one Spark function corresponds to a Velox function for most cases. Perhaps we will need a benchmark to show noticeable performance improvement before we implementing a combined function.

PHILO-HE:
@rui-mo, seems this mapping `{"not_equal", "notequalto"}` is legacy code that has never been used now.

CH backend engineer is doing some optimization to collapse nested expression into one, e.g., And((a And b), c) into And(a, b, c). The collapse optimization can generally bring perf. benefit. But for this legacy involved in this pr, I think we firstly can remove it.

PHILO-HE:
FYI. The registering for "notequalto" (a presto function) was removed by https://github.com/facebookincubator/velox/pull/5569, which seems for fixing semantic inconsistency issue.

rui-mo:
> And((a And b), c) into And(a, b, c)

@PHILO-HE It appears that Velox contains this optimisation https://facebookincubator.github.io/velox/develop/expression-evaluation.html#flatten-ands-and-ors. I agree we remove the legacy code first.

PHILO-HE (approved):

PR #9115, opened 2025-03-24
[VL] Account some C++ untracked memory allocations into Spark global off-heap memory

zhztheplayer:
1. Pass an allocation listener that reserves memory from `GlobalOffHeapmemory` from Java to C++ for `VeloxMemoryManager::getDefaultMemoryManager()`
2. Memory allocations during the following operations will be tracked with Spark global off-heap memory now (previously untracked):
   - VeloxRuntime::planString
   - nativeValidateWithFailureReason
   - SubstraitToVeloxPlanConverter::constructValuesNode
   - gluten::recordBatch2VeloxColumnarBatch (for testing)

Future TODOs:

1. leak-detection
2. evict / spill from global off-heap memory

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
@@ -52,28 +52,32 @@ object GlobalOffHeapMemory {
     f
   }
 
-  def acquire(numBytes: Long): Unit = {
-    val mm = memoryManager()
-    val succeeded =
-      mm.acquireStorageMemory(BlockId(s"test_${UUID.randomUUID()}"), numBytes, MemoryMode.OFF_HEAP)
+  def acquire(numBytes: Long): Unit = memoryManagerOption().foreach {
+    mm =>
+      val succeeded =
+        mm.acquireStorageMemory(
+          BlockId(s"test_${UUID.randomUUID()}"),
+          numBytes,
+          MemoryMode.OFF_HEAP)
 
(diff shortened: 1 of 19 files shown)


backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
```diff
@@ -48,28 +53,72 @@ object GlobalOffHeapMemory {
...
+      return Some(env.memoryManager)
     }
     val tc = TaskContext.get()
     if (tc != null) {
       // This may happen in test code that mocks the task context without booting up SparkEnv.
-      return FIELD_MEMORY_MANAGER.get(tc.taskMemoryManager()).asInstanceOf[MemoryManager]
+      return Some(FIELD_MEMORY_MANAGER.get(tc.taskMemoryManager()).asInstanceOf[MemoryManager])
     }
-    throw new GlutenException(
+    logWarning(
```
JkSelf, line 120: Does it throw unexpected exception if only print the log warning info?

JkSelf (commented):
LGTM except one small comment.

backends-velox/src/main/scala/org/apache/spark/memory/GlobalOffHeapMemory.scala
zhztheplayer: I prefer throwing here but it will lead to crash issues so far when the executor process is shutting down. So temporarily we can rely on warning messages.

JkSelf (approved):

PR #9119, opened 2025-03-24
[GLUTEN-9116] Allow passing user-specific configurations to native backend

marin-ma:
Support passing custom native backend configurations through `--conf spark.gluten.native.additionalConfList=key1,key2`. Then configurations `--conf key1=value1 --conf key2=value2` will be passed to the native backend.


docs/Configuration.md
```diff
@@ -74,6 +74,8 @@ You can add these configurations into spark-defaults.conf to enable or disable t
...
+| spark.gluten.native.additionalConfList                       | Comma-separated additional configurations to be passed to the native backend.                                                                                                                                                                                                                                                                                                                                                                                                                                                             |                                                      |
```
jinchengchenghh, line 77: What' s the format, `k1=v1,k2=v2`?

marin-ma: Only the keys. Updated in the pr description.

jinchengchenghh: Can you also describe it in the document, and update to spark.gluten.native.additionalConfNameList

marin-ma: Refined the doc. Thanks!

FelixYBW:
Can we use `spark.gluten.native.customized.cfg.key1=value1`, `spark.gluten.native.customized.cfg.key2=value2`? All config with pattern "spark.gluten.native.customized.cfg" are passed to native library without any check.

marin-ma:
> Can we use `spark.gluten.native.customized.cfg.key1=value1`, `spark.gluten.native.customized.cfg.key2=value2`? All config with pattern "spark.gluten.native.customized.cfg" are passed to native library without any check.

This requires user to rename the existing configurations. With the current approach, they only need to add one extra configuration for Gluten to recognize them. cc: @majetideepak

jinchengchenghh:
Now all configuration starts with `spark.gluten.sql.columnar.backend.velox` will be send to native backend.

zhztheplayer:
> Now all configuration starts with `spark.gluten.sql.columnar.backend.velox` will be send to native backend.

@marin-ma Have you tried with this approach?

jinchengchenghh:
This feature is for the user modified native backend, so I think if they need it, they should cherry-pick this PR to their user modified gluten backend.

PR #9134, opened 2025-03-25
[VL] Rename some `src-*/` source folders to `src/`

zhztheplayer:
Some modules don't have a `src` source folder. This patch renames some specific source folders to `src` to make sure each module has at least one `src` folder. This will make it easier to add configurations related to source folders in the root `pom.xml`.

With some code style fixes that were not detected because of the source folder names.


PHILO-HE (approved):

yaooqinn (approved):
LGTM

Issue #9055, opened 2025-03-19
[CH] input_file_name() returns wrong results


taiyang-li:
### Backend

CH (ClickHouse)

### Bug description

``` sql 
CREATE TABLE sales (
  product_id STRING,
  quantity INT
) PARTITIONED BY (year STRING)
ROW FORMAT DELIMITED FIELDS TERMINATED BY '\t'
STORED AS TEXTFILE;

INSERT INTO sales PARTITION(year='%s')
SELECT 'prod1', 100;

INSERT INTO sales PARTITION(year='%25s')
SELECT 'prod1', 100;
```

gluten 
``` sql 
0: jdbc:hive2://localhost:10000/> select input_file_name() from sales;   
+--------------------+
| input_file_name()  |
+--------------------+
|                    |
|                    |
+--------------------+
2 rows selected (0.228 seconds)

``` 

vanilla
``` sql 
0: jdbc:hive2://localhost:10000/> select input_file_name() from sales;  
+----------------------------------------------------+
|                 input_file_name()                  |
+----------------------------------------------------+
| file:/data1/liyang/cppproject/spark/spark-3.3.2-bin-hadoop3/spark-warehouse/sales/year=%25s/part-00000-fc66b1b8-3d72-4fb4-8e69-d6b4c8ce094f-c000 |
| file:/data1/liyang/cppproject/spark/spark-3.3.2-bin-hadoop3/spark-warehouse/sales/year=%2525s/part-00000-c55620e3-25c1-49cf-ad9b-2dd6d947686d-c000 |
+----------------------------------------------------+
2 rows selected (0.469 seconds)
```

### Spark version

None

### Relevant logs

```bash

```

PR #9142, opened 2025-03-26
[GLUTEN-9055][CH] Fix input_file_name diff from hive text table

taiyang-li:

Fixes: \#9055 by making PushDownInputFileExpression effective for hive text table.  

After fixing, there is still tiny diff between vanilla and gluten: when format is textfile, input_file_name() in vanilla returns paths like 'file:/xxx', ut in gluten it returns paths like 'file:///xxx', users who use input_file_name() in gluten should be aware of that.

## How was this patch tested?


baibaichen:
@CodiumAI-Agent /review

baibaichen (approved):

Issue #9149, opened 2025-03-27
Make JniWorkspace not depend on spark


shuai-xu:
### Description

JniWorkspace should not depend on spark, it may be used for flink. This is part of work for https://github.com/apache/incubator-gluten/issues/9046

PR #1143, opened 2023-03-16
[CH-364] Explicitly invoke JNI_ONUnload when spark driver or executor are shuting down

taiyang-li:
## What changes were proposed in this pull request?

Fix coredump existed in libch.so: https://github.com/Kyligence/ClickHouse/issues/364

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

test by CH[[377]]

PR #9150, opened 2025-03-27
[GLUTEN-9149][CORE] Remove Spark-specific code from JniLibLoader & JniWorkspace

shuai-xu:

Make JniWorkspace not depend on spark.

(Fixes: \#9149)

## How was this patch tested?

This patch was tested by existing unit tests and integration tests.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
@@ -149,11 +149,19 @@ class VeloxListenerApi extends ListenerApi with Logging {
   override def onExecutorShutdown(): Unit = shutdown()
 
   private def initialize(conf: SparkConf, isDriver: Boolean): Unit = {
+    // Sets this configuration only once, since not undoable.
+    // DebugInstance should be created first.
+    if (conf.getBoolean(GlutenConfig.DEBUG_KEEP_JNI_WORKSPACE.key, defaultValue = false)) {
+      val debugDir = conf.get(GlutenConfig.DEBUG_KEEP_JNI_WORKSPACE_DIR.key)
+      JniWorkspace.enableDebug(debugDir)
+    }
     JniWorkspace.initializeDefault(
-      SparkDirectoryUtil.get
-        .namespace("jni")
-        .mkChildDirRandomly(UUID.randomUUID.toString)
-        .getAbsolutePath)
+      () =>
+        SparkDirectoryUtil.get
+          .namespace("jni")
+          .mkChildDirRandomly(UUID.randomUUID.toString)
+          .getAbsolutePath)
+
     UDFResolver.resolveUdfConf(conf, isDriver)
 
     // Do row / batch type initializations.
@@ -174,12 +182,6 @@ class VeloxListenerApi extends ListenerApi with Logging {
         },
         classOf[ColumnarShuffleManager].getName
       )
-
-    // Sets this configuration only once, since not undoable.
-    if (conf.getBoolean(GlutenConfig.DEBUG_KEEP_JNI_WORKSPACE.key, defaultValue = false)) {
-      val debugDir = conf.get(GlutenConfig.DEBUG_KEEP_JNI_WORKSPACE_DIR.key)
-      JniWorkspace.enableDebug(debugDir)
-    }
 
     // Set the system properties.
     // Use appending policy for children with the same name in a arrow struct vector.
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
@@ -41,7 +41,7 @@ import org.apache.spark.sql.execution.datasources.GlutenWriterColumnarRules
 import org.apache.spark.sql.execution.datasources.velox.{VeloxParquetWriterInjects, VeloxRowSplitter}
 import org.apache.spark.sql.expression.UDFResolver
 import org.apache.spark.sql.internal.{GlutenConfigUtil, StaticSQLConf}
-import org.apache.spark.util.{SparkDirectoryUtil, SparkResourceUtil}
+import org.apache.spark.util.{SparkDirectoryUtil, SparkResourceUtil, SparkShutdownManagerUtil}
 
 import org.apache.commons.lang3.StringUtils
 
@@ -149,6 +149,7 @@ class VeloxListenerApi extends ListenerApi with Logging {
   override def onExecutorShutdown(): Unit = shutdown()
 
   private def initialize(conf: SparkConf, isDriver: Boolean): Unit = {
+    addShutdownHook
     // Sets this configuration only once, since not undoable.
     // DebugInstance should be created first.
     if (conf.getBoolean(GlutenConfig.DEBUG_KEEP_JNI_WORKSPACE.key, defaultValue = false)) {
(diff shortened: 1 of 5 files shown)


shuai-xu:

zzcclp (approved):
LGTM

zhztheplayer (approved):

gluten-core/src/main/java/org/apache/gluten/jni/JniLibLoader.java
```diff
@@ -50,14 +47,6 @@ public class JniLibLoader {
   private static final Set<String> REQUIRE_UNLOAD_LIBRARY_PATHS =
       Collections.synchronizedSet(new LinkedHashSet<>());
 
-  static {
-    SparkShutdownManagerUtil.addHookForLibUnloading(
-        () -> {
-          forceUnloadAll();
```
jackylee-ch, line 56: It seems the `forceUnloadAll` is used to fix [the coredump in ch backends](https://github.com/apache/incubator-gluten/pull/1143). Can we remove this now? @taiyang-li @baibaichen

gluten-core/src/main/java/org/apache/gluten/jni/JniWorkspace.java
```diff
@@ -99,12 +85,17 @@ public static void enableDebug(String debugDir) {
     }
   }
 
-  public static JniWorkspace getDefault() {
+  public static void initializeDefault(String rootDir) {
     synchronized (DEFAULT_INSTANCE_INIT_LOCK) {
```
PHILO-HE, line 89: Do we still need `synchronized`? Seems it can only be executed by one thread in `initialize` context after this refactoring.

gluten-core/src/main/java/org/apache/gluten/jni/JniWorkspace.java
```diff
@@ -99,12 +85,17 @@ public static void enableDebug(String debugDir) {
...
       if (DEFAULT_INSTANCE == null) {
-        DEFAULT_INSTANCE = createDefault();
+        DEFAULT_INSTANCE = createOrGet(rootDir);
       }
-      Preconditions.checkNotNull(DEFAULT_INSTANCE);
+    }
+  }
+
+  public static JniWorkspace getDefault() {
+    synchronized (DEFAULT_INSTANCE_INIT_LOCK) {
```
PHILO-HE, line 97: Remove `synchronized`.

PHILO-HE (commented):
Thanks for your work! Two comments.

gluten-core/src/main/java/org/apache/gluten/jni/JniLibLoader.java
taiyang-li: Yes, it is still needed.

shuai-xu: It seems can not be removed, or will still coredump

gluten-core/src/main/java/org/apache/gluten/jni/JniWorkspace.java
shuai-xu: There is also `synchronized` used in other function, I will remove them all in another pr.

gluten-core/src/main/java/org/apache/gluten/jni/JniWorkspace.java
shuai-xu: ditto

gluten-core/src/main/java/org/apache/gluten/jni/JniWorkspace.java
shuai-xu: In tests, driver and executors are in the same process, so `initialize` may be call by driver and executor at the same time.

zzcclp:
@PHILO-HE @jackylee-ch is it ok to merge this pr ?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -142,13 +142,19 @@ class VeloxListenerApi extends ListenerApi with Logging {
...
   }
 
   override def onExecutorShutdown(): Unit = shutdown()
 
   private def initialize(conf: SparkConf, isDriver: Boolean): Unit = {
+    JniWorkspace.initializeDefault(
+      SparkDirectoryUtil.get
+        .namespace("jni")
+        .mkChildDirRandomly(UUID.randomUUID.toString)
+        .getAbsolutePath)
```
PHILO-HE, line 155: Compared with the original code, one difference is making dir is outside the synchronized block, which can cause extra directory unnecessarily created. Maybe, we can pass the making dir function as argument to initializeDefault and let the dir be created inside initializeDefault.

gluten-core/src/main/java/org/apache/gluten/jni/JniWorkspace.java
PHILO-HE: @shuai-xu, seems no need to add "synchronized" in getting the instance.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -142,13 +142,27 @@ class VeloxListenerApi extends ListenerApi with Logging {
...
   override def onExecutorShutdown(): Unit = shutdown()
 
   private def initialize(conf: SparkConf, isDriver: Boolean): Unit = {
+    // Sets this configuration only once, since not undoable.
+    // DebugInstance should be created first.
+    if (conf.getBoolean(GlutenConfig.DEBUG_KEEP_JNI_WORKSPACE.key, defaultValue = false)) {
+      val debugDir = conf.get(GlutenConfig.DEBUG_KEEP_JNI_WORKSPACE_DIR.key)
+      JniWorkspace.enableDebug(debugDir)
+    }
+    JniWorkspace.initializeDefault(
```
PHILO-HE, line 157: If debug is enabled, no need to `initializeDefault`. Should we add an `else`?

PHILO-HE (approved):
Looks good! One trivial comment only.

PHILO-HE (approved):

jackylee-ch (approved):

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -215,6 +224,7 @@ class VeloxListenerApi extends ListenerApi with Logging {
 
   private def shutdown(): Unit = {
     // TODO shutdown implementation in velox to release resources
+    JniLibLoader.forceUnloadAll
```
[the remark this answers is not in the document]
shuai-xu: `Loadlib` is called in `initialize` for each session. Can we open another pr to remove it if not necessary.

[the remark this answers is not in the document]
zhztheplayer: > `Loadlib` is called in `initialize` for each session.

I don't think so. Please check the usage of `driverInitialized` and `executorInitialized`. They were to guarantee we only initialize once across sessions.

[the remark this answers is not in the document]
zhztheplayer: The usage of `unload` is also questionable. JVM doesn't have a certain timing for such operation because it's related to GC / JVM implementations. We'd avoid unloading the libs unless really necessary.

zzcclp:
Run Gluten ClickHouse CI

Issue #9152, opened 2025-03-27
[VL] Avoid unnecessary serialization of hadoop conf


zml1206:
### Description

After upgrading from 1.2 to 1.3, I found that simple queries were several hundred milliseconds slower because of closure serialization.

PR #9153, opened 2025-03-27
[GLUTEN-9152][CORE] Avoid unnecessary serialization of hadoop conf

zml1206:

Significantly reduce the closure serialization time of `BasicScanExecTransformer` and `WholeStageTransformer`.
(Fixes: \#9152)


backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHBackend.scala
```diff
@@ -175,7 +176,7 @@ object CHBackendSettings extends BackendSettingsApi with Logging {
       fields: Array[StructField],
       rootPaths: Seq[String],
       properties: Map[String, String],
-      serializableHadoopConf: Option[SerializableConfiguration] = None): ValidationResult = {
+      hadoopConf: Configuration): ValidationResult = {
```
zhouyuan, line 179: The serialize code is added for viewfs support as we need to pass some info to Velox. Not sure if the new code can be ser/de-ser and passed down. 
The code itself is correct, but there is no viewfs related tests in GHA now so we may need to check this manually

Cc @wangyum @JkSelf

zml1206: But from the code logic, I did not find that velox uses the hadoop conf here.

JkSelf: @wangyum Could you please help in verifying this PR in your ViewFS environment? My ViewFS test setup is currently not functioning properly.

wangyum: OK. will test it soon.

gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -236,9 +235,6 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
 
   val sparkConf: SparkConf = sparkContext.getConf
 
-  val serializableHadoopConf: SerializableConfiguration = new SerializableConfiguration(
```
Yohahaha, line 239: maybe add `@transient lazy` could fix the issue

zml1206: `@transient lazy` is a solution, but if serialization is not needed here, it would be simpler to use `sparkContext.hadoopConfiguration` directly.

backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHBackend.scala
wangyum: I have tested this patch on our cluster and it still works.

JkSelf:
@zml1206 can you help to take a look the failed unit tests?

zml1206:
> @zml1206 can you help to take a look the failed unit tests?

It's strange that PR seems to cause Java heap oom, but JDK 17 passes. I haven't found the reason yet. I'm not sure whether it is related to PR or the test itself is unstable.

zml1206:
I can't reproduce it locally. I tried many ways, including `lazy`, `@transient`, local variables, etc. `WholeStageTransformer` in spark 3.5 and jdk8 will cause OOM in HashExpressionSuite. I guess it is related to jdk8's GC. Now there is no spark 3.5-jdk8 workflow. Can it be merged or the changes of `WholeStageTransformer` can be rolled back? @JkSelf

wangyum:
@zml1206 Can we increase JVM heap [here](https://github.com/apache/incubator-gluten/blob/0c4854fea82942e991515edef9cd2dbd49be6386/pom.xml#L1377), e.g.:
```
<argLine>${argLine} -Xmx4g ${extraJavaTestArgs}</argLine>
```

zml1206:
> @zml1206 Can we increase JVM heap [here](https://github.com/apache/incubator-gluten/blob/0c4854fea82942e991515edef9cd2dbd49be6386/pom.xml#L1377), e.g.:
> 
> ```
> <argLine>${argLine} -Xmx4g ${extraJavaTestArgs}</argLine>
> ```


Thank you, after #9209, there is no CI for spark3.5-jdk8.

zml1206:
> > I can't reproduce it locally. I tried many ways, including `lazy`, `@transient`, local variables, etc. `WholeStageTransformer` in spark 3.5 and jdk8 will cause OOM in HashExpressionSuite. I guess it is related to jdk8's GC. Now there is no spark 3.5-jdk8 workflow. Can it be merged or roll back the changes of `WholeStageTransformer`? @JkSelf
> 
> @zml1206, do you mean reverting the changes for `WholeStageTransformer` can make that test pass in the previous CI test? If yes, it's strange since that code path is for viewFS and viewFS is not verified in CI.

Yes, revert  the changes for `WholeStageTransformer` can make that test pass in the previous CI test.

PHILO-HE (approved):
I think we can merge this patch first. If there is some issue found at user side, we can make a follow-up pr.

zml1206:
Thanks for everyone's review, if no one else has any comments, I'll merge it later.

PR #9155, opened 2025-03-27
[GLUTEN-9196][CH] Use wide-table aggregation to eliminate multi-table joins

lgbo-ustc:

Fixes: #9196

## How was this patch tested?


lgbo-ustc:
When querying only two tables, the optimization does not significantly improve execution speed. However, when querying three tables, the optimized query speed shows a noticeable improvement. This is partly because as the number of tables involved in the query increases, the join method requires building more hash tables. Additionally, the more join operations performed within a single node, the more likely it is to trigger memory spill operations, which further slows down execution.
[three_tables_aggregate_union.pdf](https://github.com/user-attachments/files/19561181/three_tables_aggregate_union.pdf)
[three_tables_join_aggregate.pdf](https://github.com/user-attachments/files/19561182/three_tables_join_aggregate.pdf)
[two_tables_by_aggregate_union.pdf](https://github.com/user-attachments/files/19561183/two_tables_by_aggregate_union.pdf)
[two_tables_by_join_aggregate.pdf](https://github.com/user-attachments/files/19561184/two_tables_by_join_aggregate.pdf)

zzcclp (approved):
LGTM

PR #9167, opened 2025-03-28
[CORE] LocalPartitionWriter should discard the evict since it use hash/sort evict

yikf:
LocalPartitionWriter should discard the evict since it use hash/sort evict

## How was this patch tested?
GA


yikf:
cc @marin-ma Could you please take a look, thanks!

jackylee-ch (approved):

PR #9168, opened 2025-03-28
[VL] Fix weekly build job

PHILO-HE:

1. Since https://github.com/facebookincubator/velox/commit/e8c986c9b2edbcd32d6855959b957a6b906eec42, /usr/local/lib/libprotobuf.so.32 is not installed on Ubuntu but a static lib is installed, which causes packing third-party lib fails. 

2. The `tzdata-2025a-1.el9.noarch.rpm` rpm package cannot be installed on centos-7. Like Yuan's fix in velox_backend.yml, use pip to install and update zoneinfo.
```
Error: transaction check vs depsolve:
rpmlib(PayloadIsZstd) <= 5.4.18-1 is needed by tzdata-2025a-1.el9.noarch
```

## How was this patch tested?

Local test.

### DIFF
--- a/dev/build-thirdparty.sh
+++ b/dev/build-thirdparty.sh
@@ -27,12 +27,12 @@ ARCH=`uname -m`
 mkdir -p $THIRDPARTY_LIB
 function process_setup_ubuntu_2004 {
   cp /usr/lib/${ARCH}-linux-gnu/{libroken.so.18,libasn1.so.8,libcrypto.so.1.1,libnghttp2.so.14,libnettle.so.7,libhogweed.so.5,librtmp.so.1,libssh.so.4,libssl.so.1.1,liblber-2.4.so.2,libsasl2.so.2,libwind.so.0,libheimbase.so.1,libhcrypto.so.4,libhx509.so.5,libkrb5.so.26,libheimntlm.so.0,libgssapi.so.3,libldap_r-2.4.so.2,libcurl.so.4,libdouble-conversion.so.3,libevent-2.1.so.7,libgflags.so.2.2,libunwind.so.8,libglog.so.0,libidn.so.11,libntlm.so.0,libgsasl.so.7,libicudata.so.66,libicuuc.so.66,libxml2.so.2,libre2.so.5,libsnappy.so.1,libpsl.so.5,libbrotlidec.so.1,libbrotlicommon.so.1,libthrift-0.13.0.so} $THIRDPARTY_LIB/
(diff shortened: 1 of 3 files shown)


PHILO-HE:

ep/build-velox/src/setup-centos7.sh
```diff
@@ -233,7 +233,10 @@ dnf_install ccache wget which libevent-devel \
   openssl-devel libzstd-devel lz4-devel double-conversion-devel \
   curl-devel libxml2-devel libgsasl-devel libuuid-devel patch libicu-devel tzdata
 
-dnf_install https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/tzdata-2025a-1.el9.noarch.rpm
+# Update tzdata, required by Velox at runtime.
+dnf_install python3-pip
+pip3 install tzdata
```
jinchengchenghh, line 238: What do you think about ubuntu? I test in ubuntu, it also failed

dev/build-thirdparty.sh
```diff
@@ -27,12 +27,12 @@ ARCH=`uname -m`
 mkdir -p $THIRDPARTY_LIB
 function process_setup_ubuntu_2004 {
   cp /usr/lib/${ARCH}-linux-gnu/{libroken.so.18,libasn1.so.8,libcrypto.so.1.1,libnghttp2.so.14,libnettle.so.7,libhogweed.so.5,librtmp.so.1,libssh.so.4,libssl.so.1.1,liblber-2.4.so.2,libsasl2.so.2,libwind.so.0,libheimbase.so.1,libhcrypto.so.4,libhx509.so.5,libkrb5.so.26,libheimntlm.so.0,libgssapi.so.3,libldap_r-2.4.so.2,libcurl.so.4,libdouble-conversion.so.3,libevent-2.1.so.7,libgflags.so.2.2,libunwind.so.8,libglog.so.0,libidn.so.11,libntlm.so.0,libgsasl.so.7,libicudata.so.66,libicuuc.so.66,libxml2.so.2,libre2.so.5,libsnappy.so.1,libpsl.so.5,libbrotlidec.so.1,libbrotlicommon.so.1,libthrift-0.13.0.so} $THIRDPARTY_LIB/
-  cp /usr/local/lib/{libprotobuf.so.32,libboost_context.so.1.84.0,libboost_regex.so.1.84.0} $THIRDPARTY_LIB/
```
zhouyuan, line 30: it seems libprotobuf is removed?

zhouyuan (approved):
Thanks!

dev/build-thirdparty.sh
PHILO-HE: With upstream's changes, only static protobuf is installed through setup script. So we remove this shared lib from packing list.

Issue #9177, opened 2025-03-31
[CH] parse_url diff compare to valina


KevinyhZou:
### Backend

CH (ClickHouse)

### Bug description

`parse_url('http://user:pass@localhost', 'HOST')`

```
valina: localhost
gluten:
```

`parse_url('http://user:pass@localhost/a/b/c', 'HOST')`
```
valina:localhost
gluten:
```

`parse_url('http://user:pass@localhost:10010/a/b/c', 'HOST')`
```
valina:localhost
gluten:
```

### Spark version

None

### Relevant logs

```bash

```

PR #9179, opened 2025-03-31
[GLUTEN-9177][CH]Fix diff on parse host of url and refactor `SparkParseURL`

KevinyhZou:

(Fixes: \#9177)

## How was this patch tested?
test by ut

### DIFF
--- a/backends-clickhouse/src/test/scala/org/apache/gluten/execution/tpch/GlutenClickHouseTPCHSaltNullParquetSuite.scala
+++ b/backends-clickhouse/src/test/scala/org/apache/gluten/execution/tpch/GlutenClickHouseTPCHSaltNullParquetSuite.scala
@@ -3405,6 +3405,7 @@ class GlutenClickHouseTPCHSaltNullParquetSuite extends GlutenClickHouseTPCHAbstr
         }
         assert(aggregates.size == 1)
       }
+
       val sql1 =
         """
           |select t1.*, t2.* from nation as t1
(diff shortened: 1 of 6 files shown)


backends-clickhouse/src/test/scala/org/apache/gluten/execution/tpch/GlutenClickHouseTPCHSaltNullParquetSuite.scala
```diff
@@ -3430,5 +3431,12 @@ class GlutenClickHouseTPCHSaltNullParquetSuite extends GlutenClickHouseTPCHAbstr
       compareResultsAgainstVanillaSpark(sql3, true, checkOnlyOneAggregate)
     }
   }
+
+  test("GLUTEN-9177: Fix diff of parse_url") {
+    val select_sql = "select id, parse_url('http://user:pass@locahost', 'HOST'), " +
```
taiyang-li, line 3436: const expression is still evaluated in vanilla spark although gluten is enabled.

cpp-ch/local-engine/Functions/SparkParseURL.cpp
```diff
@@ -380,6 +380,21 @@ struct SparkExtractURLHost
...
+            {
+                DB::Pos first_path_delim_pos = find_first_symbols<'/'>(userinfo_delim_pos + 1, end);
+                if (first_path_delim_pos && first_path_delim_pos - userinfo_delim_pos - 1 > 0)
+                {
+                    DB::Pos host_port_delim_pos = find_first_symbols<':'>(userinfo_delim_pos + 1, first_path_delim_pos);
+                    host = std::string_view(userinfo_delim_pos + 1, host_port_delim_pos - userinfo_delim_pos - 1);
+                }
+                else
+                {
+                    host = std::string_view{};
```
taiyang-li, line 395: let host as it be. It looks redundant here

cpp-ch/local-engine/Functions/SparkParseURL.cpp
```diff
@@ -380,6 +380,21 @@ struct SparkExtractURLHost
         if (userinfo_delim_pos && userinfo_delim_pos < end)
```
taiyang-li, line 380: instead of parsing url manually, why not use `Poco::URI`?  It should be much easier than current implementation.

KevinyhZou: The implementation of `SparkParseURL.cpp` is seems try to use the url parsing function of clickhouse,  and I think it is may be better than `Poco::URI`.

taiyang-li: But the existing code looks a bit messy and will be difficult to maintain in the future. Would you mind investigating the feasibility of Poco::URI?

taiyang-li: Besides, make sure that all related spark uts are enabled.

taiyang-li (approved):
LGTM

PR #9195, opened 2025-04-01
[VL] Enable to_timestamp test

rui-mo:
## How was this patch tested?

under test


gluten-ut/spark35/src/test/resources/backends-velox/sql-tests/inputs/datetime-parsing-invalid.sql
```diff
@@ -2,10 +2,10 @@
 
 -- parsing invalid value with pattern 'y'
 -- disable since https://github.com/facebookincubator/velox/pull/12694
--- select to_timestamp('294248', 'y'); -- out of year value range [0, 294247]
+select to_timestamp('294248', 'y'); -- out of year value range [0, 294247]
```
rui-mo, line 5:
```
Expected "[java.lang.ArithmeticException long overflow]", but got "[org.apache.gluten.exception.GlutenException
Could not convert Timestamp(9223402809600, 0) to microseconds]" Result did not match for query select to_timestamp('294248', 'y') (GlutenSQLQueryTestSuite.scala:987)
```

rui-mo: @zhouyuan Two two exceptions both on the overflow of int64_t, so it might be good to ignore this difference.

rui-mo: I suppose this is not a limitation, but different exceptions between Spark and Velox, and we usually ignore this kind of difference. Thanks.

PR #9207, opened 2025-04-02
[VL] Change to use Velox's wget_and_untar in setup-centos7.sh

yaooqinn:

This PR makes the dependency installation look up the local download source tar tall first.

We can reduce some network issues for developers while they build their own CI or just daily dev.
## How was this patch tested?

tested locally

```
+ set -u
+ '[' -d /tmp/velox-deps ']'
+ run_and_time install_cmake
+ install_cmake
+ cd /tmp/velox-deps
+ wget_and_untar https://cmake.org/files/v3.28/ cmake-3.28.3.tar.gz cmake-3
+ local URL=https://cmake.org/files/v3.28/
+ local BIN=cmake-3.28.3.tar.gz
+ local DIR=cmake-3
+ mkdir -p cmake-3
+ [[ -f cmake-3.28.3.tar.gz ]]
+ tar -tzf cmake-3.28.3.tar.gz
+ echo 'Using cached cmake-3.28.3.tar.gz'
Using cached cmake-3.28.3.tar.gz
+ tar -xzf cmake-3.28.3.tar.gz -C cmake-3 --strip-components=1
+ return 0
```

### DIFF
--- a/ep/build-velox/src/setup-centos7.sh
+++ b/ep/build-velox/src/setup-centos7.sh
@@ -46,13 +46,6 @@ function yum_install {
   $SUDO yum install -y "$@"
 }
 
-function wget_and_untar {
-  local URL=$1
-  local DIR=$2
-  mkdir -p "${DIR}"
-  wget -q --max-redirect 3 -O - "${URL}" | tar -xz -C "${DIR}" --strip-components=1
-}
-
 function install_cmake {
   cd "${DEPENDENCY_DIR}"
   wget_and_untar https://cmake.org/files/v3.28/cmake-3.28.3.tar.gz cmake-3


ep/build-velox/src/setup-centos7.sh
```diff
@@ -48,14 +48,29 @@ function yum_install {
 
 function wget_and_untar {
```
PHILO-HE, line 49: @yaooqinn, thanks for work!

I note upstream velox's wget_and_untar has a check for dir to let user decide whether to re-download the tar. Does it meet our needs?

Actually, by removing this `wget_and_untar` function, Velox's `wget_and_untar` (placed in `setup-helper-functions.sh`) will be used in this script. I am wondering whether this improvement can be put in upstream velox. Then, both `setup-centos7.sh` and `setup-centos8.sh` (maintained in Gluten) can use that, as well as `setup-centos9.sh` and `setup-ubuntu.sh` (maintained in Velox).

yaooqinn: Hi, @PHILO-HE. Thanks for the suggestion.

I had considered using velox's wget_and_untar, but I didn't apply it here. 

- Some of the dependency names might conflict as these names are in the form of `vx.y.z`. 
- velox's wget_and_untar does not check whether the tar ball is valid or not.

> I am wondering whether this improvement can be put in upstream velox.

This is a good suggestion, I will try to send a PR to the upstream. But how about we focus the dependency management here first

yaooqinn: On second thought, I agree w/ you about removing the wget_and_untar and use the upstream one

philo-he (approved):
Thanks for your continuous work!

Issue #9188, opened 2025-03-31
[VL][CI] Spark-3.5 test needs 3.5 hours with Java-8


philo-he:
### Description

One finding is, running "cast string to timestamp" test (added recently) with Java-8 is ~20min slower than that with Java-17, where `ThreadUtils.parmap` is used. There are two such tests in Spark-3.5.

![Image](https://github.com/user-attachments/assets/1e2bba76-fa36-4ad9-9e14-40dc33590d7f)

Issue #8574, opened 2025-01-20
Support Spark 4.0 and Deprecate JDK 8


weiting-chen:
### Description

Current Spark and JDK support status are in below
Spark 3.4.4 support JDK8(8u362) or above version
Spark 3.5.4 support JDK8(8u371) or above version
Spark 4.0 can only support JDK 11 and 17
For better CI resource usage, we can consider to deprecate JDK 8 support when adding Spark 4.0 support.
The plan to support Spark 4.0 will be initialized in Gluten 1.4 and done the work in Gluten 1.5.

Yohahaha:
would you send a email to gluten dev mail list to notify all developers?

weiting-chen:
Yes, will send a mail to dev mail list.

FelixYBW:
@weiting-chen we may move our CI to JDK17 now but still leave the legacy jdk8 support code there, we still accept PRs for JDK8. Customer who are using jdk8 needs to run the CI from their internal env.

Once we release Gluten1.4, we may remove the JDK8 support completely.

jackylee-ch:
> Once we release Gluten1.4, we may remove the JDK8 support completely.

This means we would drop Spark 3.2 and Spark 3.3 support after Gluten 1.4, right?

zhztheplayer:
Is it possible to build Gluten with higher JDK version only for `-P spark-4.0`?

PR #9209, opened 2025-04-02
[VL][CI] Change to use JDK-17 for Spark 3.3/3.4/3.5 tests

philo-he:

According to CI usage feedback, Spark 3.3/3.4/3.5 tests requires around 3 hours (see #9188). By upgrading to JDK-17, the test time can be reduced a lot.

We will continue the support for JDK-8. In CI,  all Spark-3.2 UTs and some TPC tests are still verified on JDK-8, which should be sufficient to uncover code issues on JDK-8, if any.

(Fixes: #9188)

### DIFF
--- a/docs/developers/NewToGluten.md
+++ b/docs/developers/NewToGluten.md
@@ -12,10 +12,11 @@ Now gluten supports Ubuntu20.04, Ubuntu22.04, centos8, centos7 and macOS.
 
 ## JDK
 
-Currently, Gluten supports JDK-8 for Spark 3.2/3.3/3.4/3.5. And for Spark 3.3 and higher versions, Gluten
-supports JDK 11/17. Please note Spark-4.0 only supports JDK 11 and 17. So we recommend Velox backend users
-to use higher JDK version to ease the migration for deploying Gluten with Spark-4.0 in the future. And we
-may probably upgrade Arrow from 15.0.0 to some higher version, which also requires using JDK-11 or JDK-17.
+Currently, Gluten supports JDK 8 for Spark 3.2/3.3/3.4/3.5. For Spark 3.3 and higher versions, Gluten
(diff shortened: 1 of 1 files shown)


zhouyuan:
We need to document the JDK support policy in the index page.
related: https://github.com/apache/incubator-gluten/issues/8574

.github/workflows/velox_backend.yml
```diff
@@ -908,116 +863,29 @@ jobs:
...
-        uses: actions/upload-artifact@v4
-        with:
-          name: golden-files-spark35
-          path: /tmp/tpch-approved-plan/**
 
-  run-spark-test-spark35-jdk17:
+  spark-test-spark35:
     needs: build-native-lib-centos-7
     runs-on: ubuntu-20.04
     container: apache/gluten:centos-8-jdk8
```
zhouyuan, line 891: should change with jdk17 image for this job also

PHILO-HE: JDK-17 is installed and set to use for each JDK-17 jobs.
I note arrow jars generated by build-native-lib job are actually compiled by JDK-8 and they are reused by other JDK-17 jobs. It works well. So currently we don't need to use JDK-17 docker inside which arrow jars are compiled by JDK-17.

Maybe, we can just keep one docker where both JDK-11 & 17 installed and switch the JDK to the required one for different CI jobs. For now, we can keep the two dockers and observe the CI for some time.

zhouyuan: thanks, I see the point now, it seems for now we need to use a combination of JDK8(arrow jar) + JDK17 spark jars which are little odd. will try to check this on https://github.com/apache/incubator-gluten/pull/9218

zhouyuan (approved):
👍

PR #9216, opened 2025-04-03
[GLUTEN-9215][VL] Release buffered input stream in native ShuffleReader on EOS to avoid OOM

marin-ma:


marin-ma:
Close this one as it cannot fix the OOM issue. Opened another possible fix: https://github.com/apache/incubator-gluten/pull/9221

PR #9228, opened 2025-04-04
[GLUTEN-8891][VL] Refine local ssd cache feature 

zhouyuan:

This patch adds serveral new configurations for local SSD cache.
Also aded one basic parquet read test with local cache 

## How was this patch tested?

pass GHA


zhouyuan:
@zhli1142015 would you please help to take a look on this? Thanks, -yuan

PR #9229, opened 2025-04-04
[GLUTEN-8851][VL] feat: Support cudf

jinchengchenghh:
Enable this feature by compiling with `--enable_gpu`.
Now only support OrderBy operator.
Use this branch https://github.com/apache/incubator-gluten/tree/cudf

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -315,12 +315,13 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        digests: [ vcpkg-centos-8, centos-8-jdk8, centos-8-jdk11, centos-8-jdk17 ]
+        digests: [ vcpkg-centos-8, centos-8-jdk8, centos-8-jdk11, centos-8-jdk17, centos-9-jdk8-cudf ]
     needs:
       - build-vcpkg-centos-8
       - build-centos-8-jdk8
       - build-centos-8-jdk11
       - build-centos-8-jdk17
+      - build-centos-9-jdk8-cudf
     steps:
       - name: Download digests
         uses: actions/download-artifact@v4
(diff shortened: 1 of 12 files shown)


jinchengchenghh:
Can you help review again? Thanks!@zhouyuan

.github/workflows/docker_image.yml
```diff
@@ -259,6 +260,56 @@ jobs:
...
+          digest="${{ steps.build.outputs.digest }}"
+          touch "${{ runner.temp }}/digests/${digest#sha256:}"
+
+      - name: Upload digest
+        uses: actions/upload-artifact@v4
+        with:
+          name: digests-centos-9-jdk8-cudf-${{ matrix.os }}
+          path: ${{ runner.temp }}/digests/*
+          if-no-files-found: error
+          retention-days: 1
```
zhouyuan, line 311: we may also need to add this new image in the merge section
https://github.com/apache/incubator-gluten/pull/9229/files#diff-e0ed06ed167d42ded61bb8913c0f02d8a691900a916546329d70b7ae1a14e994R318

jinchengchenghh: The docker image uses the gluten branch cudf, not the main branch, https://github.com/apache/incubator-gluten/pull/9229/files#diff-395a263b6f69bd7ef4991face3666e100acdd60e47f4b28e0737adbbf5fb945aR6, so we set the ENABLE_CUDF default OFF.

zhouyuan (approved):
👍

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1716,4 +1717,11 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val COLUMNAR_CUDF_ENABLED =
+    buildConf("spark.gluten.sql.columnar.cudf")
+      .internal()
+      .doc("Enable or disable cudf support.")
+      .booleanConf
+      .createWithDefault(true)
```
zhztheplayer, line 1725: Several questions:

1. Could we add an "experimental" mark in the option `doc()`? 
2. Should we turn it off by default?
3. What happens from user side if CPP code is not built with `--enable_gpu` but this option is turned on?

PR #9230, opened 2025-04-04
[GLUTEN-8497][VL] Fix columnar batch type mismatch in table cache

zhztheplayer:
This fixes the test case added in https://github.com/apache/incubator-gluten/pull/8498.

The patch conditionally adds a `ColumnarToRowRemovalGuard` node that does nothing on top of a 

```
+- ColumnarToRow
   +- FileScan parquet
```

which is to be cached to avoid [this Spark code](https://github.com/apache/spark/blob/9d3f937c555ccab7777c976b66da7c7229582f26/sql/core/src/main/scala/org/apache/spark/sql/execution/columnar/InMemoryRelation.scala#L335-L351) from removing the C2R, the plan will become:

```
ColumnarToRowRemovalGuard
+- ColumnarToRow
   +- FileScan parquet [l_orderkey_read#128L] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/tmp/spark-e732391d-d3f4-45e7-ae2e-d521d7658b01], PartitionFilters: [], PushedFilters: [], ReadSchema: struct<l_orderkey_read:bigint>
```

which will be treated as regular row-based plan by `ColumnarCachedBatchSerializer` then be handled with vanilla Spark batch serializer.

Previously, there was an columnar batch type mismatch error because Spark's cache planner removes the top `ColumnarToRow` but treats the remaining of the plan as vanilla Spark columnar plan.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
@@ -21,7 +21,7 @@ import org.apache.gluten.columnarbatch.VeloxBatch
 import org.apache.gluten.config.GlutenConfig
 import org.apache.gluten.extension._
 import org.apache.gluten.extension.columnar._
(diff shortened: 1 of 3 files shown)


zhouyuan (approved):
👍  LGTM

Issue #9240, opened 2025-04-07
Write NULL value into relation in gluten unit test


dcoliversun:
We need read null from relation, otherwise it won't offload to native side.

_Originally posted by @jackylee-ch in https://github.com/apache/incubator-gluten/pull/9009#discussion_r2030833984_

PR #9241, opened 2025-04-07
[GLUTEN-9240][VL] Write NULL value into relation in gluten unit tests

dcoliversun:

This PR aims to write NULL value into relation in gluten unit tests, otherwise it won't offload to native side.

(Fixes: \#9240)

## How was this patch tested?

Pass original unit tests.


dcoliversun:
cc @jackylee-ch

backends-velox/src/test/scala/org/apache/gluten/execution/JsonFunctionsValidateSuite.scala
```diff
@@ -54,9 +54,21 @@ class JsonFunctionsValidateSuite extends FunctionsValidateSuite {
     runQueryAndCompare(
       s"select l_orderkey, json_array_length('[1,2,3,4]') " +
```
jackylee-ch, line 55: This case should also read from relation, could you rewrite it as well? 
BTW, ther are same problems with `get_json_object`.

dcoliversun: ok, I check other tests also.

jackylee-ch (approved):

dcoliversun:
@jackylee-ch gentle ping 😄 it's ready to merge

Issue #9244, opened 2025-04-07
Native runtime keeps the same timezone as JVM when SESSION_LOCAL_TIMEZONE is not set


zml1206:
### Description

Consistent with vanilla Spark, simplify Spark UT to avoid rewriting.

PR #9249, opened 2025-04-07
[GLUTEN-9244][CORE] Change the way of passing default timezone to native config

zml1206:

(Fixes: \#9244)

### DIFF
--- a/gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtime.scala
+++ b/gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtime.scala
@@ -47,7 +47,7 @@ object Runtime {
       nmm.getHandle(),
       ConfigUtil.serialize(
         GlutenConfig
-          .getNativeSessionConf(backendName, GlutenConfigUtil.parseConfig(getConf)))
+          .getNativeSessionConf(backendName, GlutenConfigUtil.parseConfig(SQLConf.get.getAllConfs)))
     )
 
     private val released: AtomicBoolean = new AtomicBoolean(false)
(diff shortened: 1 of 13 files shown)


gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtime.scala
```diff
@@ -47,7 +47,7 @@ object Runtime {
       nmm.getHandle(),
       ConfigUtil.serialize(
         GlutenConfig
-          .getNativeSessionConf(backendName, GlutenConfigUtil.parseConfig(SQLConf.get.getAllConfs)))
+          .getNativeSessionConf(backendName, GlutenConfigUtil.parseConfig(getConf)))
```
zml1206, line 50: @taiyang-li Is the way clickhouse backend passes SQLConf different from velox? Can you help take a look? Thank you.

zhztheplayer:
@PHILO-HE

PHILO-HE (commented):
Maybe, with this fix, we can remove `GLUTEN_DEFAULT_SESSION_TIMEZONE`, `kDefaultSessionTimezone` and the relevant code.

I suggest to change the below code to throw exception if `valueExists` returns false or  `veloxCfg_->get<std::string>(kSessionTimezone, "")` returns empty string, since we always expect the session timezone is set by user or set with the default value.

https://github.com/apache/incubator-gluten/blob/main/cpp/velox/compute/WholeStageResultIterator.cc#L484

PHILO-HE (approved):
Looks good! The construction of QueryConfig will check the validity of configured timezone.

PR #9092, opened 2025-03-21
[VL] Improve native plan validation code

PHILO-HE:
## What changes were proposed in this pull request?

Make global class reference for NativePlanValidationInfo. It's inefficient to find the class and creat a local reference in each JNI call.
Refactor SubstraitToVeloxPlanValidator.

## How was this patch tested?

Existing tests.

PR #9268, opened 2025-04-09
[VL] Delete global reference to a class object in JNI unload

PHILO-HE:

The global reference to class object of `NativePlanValidationInfo` was introduced in #9092. It should be explicitly deleted in JNI unload.


PHILO-HE:
@jackylee-ch, could you please take a look?

jackylee-ch (approved):

zhztheplayer:
@PHILO-HE Have you met any issue without the patch?

zhztheplayer:
> @zhztheplayer, no, seems this reference cannot be auto deleted, right? 

For now, actually the unloading is invoked only when JVM is shutting down, so it's not a big deal whether we explicitly delete them at that time. But it's good to keep a consistent form of code. I was just curious whether there was an error or so.

PR #9269, opened 2025-04-09
[VL] Minor cleanups

zml1206:


backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -100,28 +100,24 @@ case class ColumnarPartialProjectExec(original: ProjectExec, child: SparkPlan)(
   }
 
   private def getProjectIndexInChildOutput(exprs: Seq[Expression]): Unit = {
-    exprs.forall {
+    exprs.foreach {
```
jinchengchenghh, line 103: forall can exit the condition early when not matched

zml1206: Got it, thank you.

PR #9272, opened 2025-04-09
[CORE] Post events until both spark ui and gluten ui are enable

yikf:
Post events until both spark ui and gluten ui are enable. Prevents invalid gluten event post when the spark ui is not turned on

## How was this patch tested?
GA

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenFallbackReporter.scala
+++ b/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenFallbackReporter.scala
@@ -25,7 +25,7 @@ import org.apache.gluten.logging.LogLevelUtil
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.catalyst.util.StringUtils.PlanStringConcat
-import org.apache.spark.sql.execution.ui.GlutenEventUtils
+import org.apache.spark.sql.execution.ui.GlutenUIUtils
 
 /**
  * This rule is used to collect all fallback reason.
@@ -40,7 +40,7 @@ case class GlutenFallbackReporter(glutenConf: GlutenConfig, spark: SparkSession)
       return plan
     }
     printFallbackReason(plan)
-    if (glutenConf.glutenUiEnabled) {
+    if (GlutenUIUtils.uiEnabled(spark.sparkContext, Some(glutenConf))) {
       postFallbackReason(plan)
     }
     plan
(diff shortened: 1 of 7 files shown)


yikf:
@jackylee-ch Could you please take a look.

gluten-ui/src/main/scala/org/apache/spark/sql/execution/ui/GlutenUIUtils.scala
```diff
@@ -16,13 +16,21 @@
...
 
 import org.apache.spark.SparkContext
 import org.apache.spark.status.ElementTrackingStore
 
-object GlutenEventUtils {
-  def post(sc: SparkContext, event: GlutenEvent): Unit = {
+object GlutenUIUtils {
+
+  def uiEnabled(sc: SparkContext): Boolean = {
+    sc.ui.isDefined && sc.getConf.getBoolean(
```
jackylee-ch, line 28: If gluten_ui is set via the set command, will it still take effect?

yikf: gluten_ui should be a static config

jackylee-ch: oh, right.

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenFallbackReporter.scala
```diff
@@ -40,7 +40,7 @@ case class GlutenFallbackReporter(glutenConf: GlutenConfig, spark: SparkSession)
       return plan
     }
     printFallbackReason(plan)
-    if (glutenConf.glutenUiEnabled) {
+    if (GlutenUIUtils.uiEnabled(spark.sparkContext)) {
```
jackylee-ch, line 43: Can we use another method to check if the UI is enabled? We need to use `glutenConf.glutenUiEnabled` to prevent user configuration at runtime.

yikf: the gluten ui config cannot be modified at runtime. I'm not quite sure what you mean. Please correct me if I'm miss something.

yikf: the most accurate approach would be to check if the tabs in the webUI include GlutenSQLTab instance, but perhaps it's a bit heavy? I'm fine with both.

jackylee-ch: `spark.sparkContext.ui.isDefined && glutenConf.glutenUiEnabled` should be enough

yikf: this is not enough. `uiEnabled` might be called anywhere, and when it is called, `sqlConf` might not have been initialized yet, such as during plugin initialization.

jackylee-ch (approved):

jackylee-ch:
Thanks for your work! @yikf

PR #9277, opened 2025-04-09
[CORE] Remove unload API from JniLibLoader as unused

zhztheplayer:
The API was originally added for CH backend library unloading. Now the C++ side JNI_OnUnload implmentation in CH backend was removed in https://github.com/apache/incubator-gluten/pull/9242. So we don't need this API in Java-side as well ideally.

The invocation code to Java side unload API relies on reflections and some uncertain promises from JVM. Its portability is not verified and maintenance will also be an issue. It's the main reason we do the removal and avoid introduce a similar mechanism in future also.


zhztheplayer:
cc @baibaichen @PHILO-HE @zzcclp

zzcclp (approved):
LGTM

PR #9285, opened 2025-04-10
[GLUTEN-6101][VL] Enable map_from_array function

WangGuangxin:

Enable map_from_array function

(Fixes: \#6101)

## How was this patch tested?

more ut

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -16,8 +16,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=2025_04_22
+VELOX_REPO=https://github.com/wangguangxin/velox.git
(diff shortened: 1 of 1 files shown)


cpp/core/config/GlutenConfig.h
```diff
@@ -37,6 +37,10 @@ const std::string kAllowPrecisionLoss = "spark.sql.decimalOperations.allowPrecis
 
 const std::string kIgnoreMissingFiles = "spark.sql.files.ignoreMissingFiles";
 
+const std::string kMapKeyDedupPolicy = "spark.sql.mapKeyDedupPolicy";
+
+const std::string kDefaultSessionTimezone = "spark.gluten.sql.session.timeZone.default";
```
Yohahaha, line 42: why this change is needed?

WangGuangxin: It's a mistake. reverted

PR #9290, opened 2025-04-10
[VL] Remove param original of ColumnarPartialProjectExec

zml1206:

Simplify plan.

## How was this patch tested?

Existing ut.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
@@ -44,9 +44,6 @@ import scala.collection.mutable.ListBuffer
  * (a, b, c) ColumnarPartialProjectExec (a, b, c, myudf(a) as _SparkPartialProject1),
  * ProjectExecTransformer(_SparkPartialProject1 + b + hash(c))
  *
- * @param original
- *   extract the ScalaUDF from original project list as Alias in UnsafeProjection and
- *   AttributeReference in ColumnarPartialProjectExec output
  * @param projectList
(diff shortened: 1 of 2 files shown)


backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -141,21 +136,20 @@ case class ColumnarPartialProjectExec(
     if (projectAttributes.size == child.output.size) {
       return ValidationResult.failed("UDF need all the columns in child output")
     }
-    if (original.output.isEmpty) {
-      return ValidationResult.failed("Project fallback because output is empty")
-    }
```
zml1206, line 146: If `original.output` is empty, then `replacedAliasUdf` must be also empty, repeat the following judgment.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -80,11 +75,11 @@ case class ColumnarPartialProjectExec(
   override def output: Seq[Attribute] = child.output ++ replacedAliasUdf.map(_.toAttribute)
 
   override def doCanonicalize(): ColumnarPartialProjectExec = {
-    val canonicalized = original.canonicalized.asInstanceOf[ProjectExec]
-    this.copy(
-      original = canonicalized,
-      child = child.canonicalized
-    )(replacedAliasUdf.map(QueryPlan.normalizeExpressions(_, child.output)))
+    super
```
jinchengchenghh, line 78: Can you help review here? @lifulong

PR #9312, opened 2025-04-14
[CORE] Correctly handle driver configurations when `spark.sql.extensions` is explicitly set for GlutenSessionExtensions

zhztheplayer:
`spark.sql.extensions=org.apache.gluten.extension.GlutenSessionExtensions` should be set automatically when Spark driver started with Gluten. However if user explicitly sets it, unexpected errors will be caused. For example:

```
java.lang.UnsupportedOperationException: This operator doesn't support doTransform with SubstraitContext.
	at org.apache.gluten.execution.TransformSupport.doTransform(WholeStageTransformer.scala:192)
	at org.apache.gluten.execution.TransformSupport.doTransform$(WholeStageTransformer.scala:190)
	at org.apache.gluten.execution.WholeStageTransformer.doTransform(WholeStageTransformer.scala:222)
	at org.apache.gluten.execution.TransformSupport.$anonfun$transform$1(WholeStageTransformer.scala:185)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$executeQuery$1(SparkPlan.scala:246)
	at org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:243)
```

The patch enhances relevant logic to handle this case correctly.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -142,7 +142,7 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
     if (!extensionSeq.toSet.contains(GlutenSessionExtensions.GLUTEN_SESSION_EXTENSION_NAME)) {
       conf.set(
         SPARK_SESSION_EXTENSIONS.key,
-        (extensionSeq ++ GlutenSessionExtensions.GLUTEN_SESSION_EXTENSION_NAME).mkString(","))
+        (extensionSeq :+ GlutenSessionExtensions.GLUTEN_SESSION_EXTENSION_NAME).mkString(","))
     }
 
     // adaptive custom cost evaluator class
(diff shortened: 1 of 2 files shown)


yaooqinn:
You can use SPI to load this extension instead of grabbing it with end-users

FYI, https://github.com/apache/spark/commit/51815430b289dc43eaf2d24b98a1597b73c7c163

zhztheplayer:
@yaooqinn Thank you for the inputs. 

Though the static service file doesn't feel straightforward to work with `spark.plugins`? Which is the recommended way for enabling Gluten for Spark.

yaooqinn:
If we need to load to `GlutenSessionExtensions` by default and we've provided a dynamic config to disable it, using SPI might be the best way.  Besides, AFAIK, some cloud vendor platforms that provide Spark services are not that easy to load jars/extentions with configs.

BTW, `spark.plugins` hasn't provided SPI registration yet

zhztheplayer:
> If we need to load to GlutenSessionExtensions by default and we've provided a dynamic config to disable it

We load `GlutenSessionExtensions` only when `spark.plugins=org.apache.gluten.GlutenPlugin`. We have a dynamic config `spark.gluten.enabled` to disable Gluten but the option can take effect even after the extensions are injected. So it's implemented in the different way.

The PR is just to address a corner case that user accidentally specifies `spark.sql.extensions=org.apache.gluten.extension.GlutenSessionExtensions` manually which is not recommended. We'd make Gluten work normally in that case.

Issue #9313, opened 2025-04-14
ColumnarPartialProject supports build-in blacklist expression


WangGuangxin:
### Description

ColumnarPartialProject supports build-in blacklist expression

PR #9315, opened 2025-04-14
[GLUTEN-9313][VL] ColumnarPartialProject supports built-in but blacklisted function

WangGuangxin:

The `ColumnaPartialProject` can also supports build-in functions, especially the blacklist expressions.

One typical scenario is regexp. The native regexp lib `re2` is much slower than Java regexp lib, and also has some semantic difference with Java lib.

Take a simple sql in our production as an example
```
SELECT  p_date,
        cast(id AS BIGINT) AS creative_id,
        regexp_replace(DATA, '||    ', '') AS tmp
FROM    test_table
WHERE   p_date='20250227'
```

| Test  | Cost |
|--------|--------|
| Partial Project Fallback | 1284h |
| Whole Project Fallback | 1389h |
| Native (No Fallback) | 3384h |

In this PR, try to use `ColumnarPartialProject` to handle the blacklist expressions.

(Fixes: \#9313)

## How was this patch tested?

more UT

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
@@ -19,7 +19,7 @@ package org.apache.gluten.execution
 import org.apache.gluten.backendsapi.BackendsApiManager
 import org.apache.gluten.columnarbatch.{ColumnarBatches, VeloxColumnarBatches}
 import org.apache.gluten.config.GlutenConfig
-import org.apache.gluten.expression.{ArrowProjection, ExpressionUtils}
+import org.apache.gluten.expression.{ArrowProjection, ExpressionMappings, ExpressionUtils}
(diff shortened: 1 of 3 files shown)


WangGuangxin:
cc @jinchengchenghh @jackylee-ch

jinchengchenghh:
Why it happens, do we need to optimize re2?

> The native regexp lib re2 is much slower than Java regexp lib

jinchengchenghh:
We may extend to all native backend not supported expressions, blacklist expression is one of it.

jackylee-ch (approved):
+1. We need a better way to find all the unsupported/blacklisted functions.

jinchengchenghh:
Can you rebase? Thanks!

WangGuangxin:
> Why it happens, do we need to optimize re2?
> 
> > The native regexp lib re2 is much slower than Java regexp lib

@jinchengchenghh I think so. We also tried another native regexp lib named `ICU`(https://unicode-org.github.io/icu/userguide/strings/regexp.html), but the performance is very pool as well.

Issue #9276, opened 2025-04-09
[CORE] Investigate: Heap shrinking to share on-heap memory for off-heap use


zhztheplayer:
### Description

(ONLY A PATHFINDING NOW)

Based on accessible documents, there might be a way to do real memory sharing between on-heap and off-heap in Java so probably in Gluten, by shrinking on-heap memory dynamically and then increase the off-heap memory accordingly.

Preconditions:

* (Required) `-XX:MaxHeapFreeRatio` / `-XX:MinHeapFreeRatio` in G1 GC + HotSpot JVM  (https://www.oracle.com/java/technologies/javase/vmoptions-jsp.html)
* (Optional?) \>= Java 12? ([JEP-346](https://openjdk.org/jeps/346))
* (Optional?) `-Xmaxf` / `-Xminf` in HotSpot VM (https://docs.oracle.com/cd/E19683-01/806-7930/vmoptions-chapter/index.html)

PR #9321, opened 2025-04-14
[VL] Count total JVM memory as the on-heap portion for the off-heap sizing feature

zhztheplayer:
Previously the feature uses `total memory - free memory` as the on-heap occupied size, which was wrong. The free memory is unused by JVM but not yet returned to the OS. The heap occupation from operating system is always the total JVM memory.

The patch corrects the calculation by using the total JVM memory instead.

Related:

https://github.com/apache/incubator-gluten/pull/5439
https://github.com/apache/incubator-gluten/issues/9276
https://github.com/apache/incubator-gluten/discussions/8600

With some minor code cleanups in passing.

### DIFF
--- a/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
+++ b/gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
@@ -43,12 +43,16 @@ public class DynamicOffHeapSizingMemoryTarget implements MemoryTarget {
       return 0;
     }
 
-    long totalMemory = Runtime.getRuntime().totalMemory();
-    long freeMemory = Runtime.getRuntime().freeMemory();
-    long usedOnHeapBytes = (totalMemory - freeMemory);
+    // Only JVM shrinking can reclaim space from the total JVM memory.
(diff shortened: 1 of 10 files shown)


gluten-core/src/main/java/org/apache/gluten/memory/memtarget/DynamicOffHeapSizingMemoryTarget.java
```diff
@@ -43,22 +43,26 @@ public long borrow(long size) {
...
+    // See https://github.com/apache/incubator-gluten/issues/9276.
+    long totalHeapMemory = Runtime.getRuntime().totalMemory();
+    long freeHeapMemory = Runtime.getRuntime().freeMemory();
+
     long usedOffHeapBytesNow = USED_OFFHEAP_BYTES.get();
 
-    if (size + usedOffHeapBytesNow + usedOnHeapBytes > MAX_MEMORY_IN_BYTES) {
+    // Adds the total JVM memory which is the actual memory the JVM occupied from the operating
+    // system into the counter.
+    if (size + usedOffHeapBytesNow + totalHeapMemory > MAX_MEMORY_IN_BYTES) {
```
zhouyuan, line 55: I got the idea and this fix looks correct to me
Cc: @zhli1142015

zhouyuan (approved):
👍

Issue #7652, opened 2024-10-23
[VL][1.2] Type mismatch error in VectorHasher


wForget:
### Backend

VL (Velox)

### Bug description

Type mismatch error in VectorHasher.

### Spark version

Spark-3.5.x

### Relevant logs

```bash
org.apache.gluten.exception.GlutenException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Type mismatch: VARCHAR vs. VARBINARY
Retriable: False
Expression: type_->kindEquals(vector.type())
Context: Operator: PartialAggregation[1] 1
Function: decode
/data/workspace/gluten-deploy-dist/ep/build-velox/build/velox_ep/./velox/exec/VectorHasher.h
Line: 168
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorERKSsEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox4exec13BaseHashTable20prepareForGroupProbeERNS1_10HashLookupERKSt10shared_ptrINS0_9RowVectorEERNS0_17SelectivityVectorEba
# 4  _ZN8facebook5velox4exec11GroupingSet21addInputForActiveRowsERKSt10shared_ptrINS0_9RowVectorEEb
# 5  _ZN8facebook5velox4exec11GroupingSet8addInputERKSt10shared_ptrINS0_9RowVectorEEb
# 6  _ZN8facebook5velox4exec15HashAggregation8addInputESt10shared_ptrINS0_9RowVectorEE
# 7  _ZN8facebook5velox4exec6Driver11runInternalERSt10shared_ptrIS2_ERS3_INS1_13BlockingStateEERS3_INS0_9RowVectorEE
# 8  _ZN8facebook5velox4exec6Driver4nextERSt10shared_ptrINS1_13BlockingStateEE
# 9  _ZN8facebook5velox4exec4Task4nextEPN5folly10SemiFutureINS3_4UnitEEE
# 10 _ZN6gluten24WholeStageResultIterator4nextEv
[...]

PR #9325, opened 2025-04-15
[GLUTEN-7652][VL] Support binary as string

wForget:

Velox supports binary as string of parquet after https://github.com/facebookincubator/velox/pull/10399, we need to pass hive columns type to HiveTableHandle

Fixes: #7652

## How was this patch tested?

added unit test

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -1292,18 +1292,19 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
 
   // Velox requires Filter Pushdown must being enabled.
   bool filterPushdownEnabled = true;
+  auto names = colNameList;
+  auto types = veloxTypeList;
+  auto dataColumns = ROW(std::move(names), std::move(types));
   std::shared_ptr<connector::hive::HiveTableHandle> tableHandle;
   if (!readRel.has_filter()) {
     tableHandle = std::make_shared<connector::hive::HiveTableHandle>(
-        kHiveConnectorId, "hive_table", filterPushdownEnabled, common::SubfieldFilters{}, nullptr);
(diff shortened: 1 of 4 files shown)


cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1292,18 +1292,20 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
 
   // Velox requires Filter Pushdown must being enabled.
   bool filterPushdownEnabled = true;
+  auto names = colNameList;
```
wForget, line 1295: `readRel.base_schema` are output schema and are not consistent with parquet file schema, so it seems that `readRel.base_schema` cannot be set as hive fileSchema.

wForget: Since we map by filed name instead of indices, this change is acceptable.

https://github.com/apache/incubator-gluten/blob/e8b723791086fd520e1c97f6da4f969783da2215/cpp/velox/compute/VeloxBackend.cc#L313-L315

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/hive/execution/GlutenHiveSQLQuerySuite.scala
```diff
@@ -48,4 +48,23 @@ class GlutenHiveSQLQuerySuite extends GlutenHiveSQLQuerySuiteBase {
       purge = false)
   }
 
+  testGluten("test binary as string") {
```
Yohahaha, line 51: could we move this ut to backend-velox module?

wForget: > could we move this ut to backend-velox module?

Thank you, I have moved it

Yohahaha (approved):
👍

wForget:
@Yohahaha CI has passed, could you please take another look?

Yohahaha (approved):
LGTM!

PR #9327, opened 2025-04-15
[VL] Add a config to enable or disable check usage leak

j7nhai:
We found that the memory leak problem still exists. In order to make the task run normally, we add a configuration to enable or disable check memory leak.


shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1265,6 +1266,13 @@ object GlutenConfig {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefaultString("8MB")
 
+  val COLUMNAR_MEMORY_CHECK_USAGE_LEAK =
+    buildConf("spark.gluten.memory.checkUsageLeak")
+      .internal()
+      .doc("Enable check memory usage leak.")
+      .booleanConf
+      .createWithDefault(true)
```
zhztheplayer, line 1274: Could move it to `VeloxConfig.scala` with the prefixed option name.

j7nhai: DONE.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -594,4 +594,11 @@ object VeloxConfig {
       .stringConf
       .checkValues(Set("loose", "strict"))
       .createWithDefault("loose")
+
+  private val COLUMNAR_VELOX_MEMORY_CHECK_USAGE_LEAK =
+    buildConf("spark.gluten.sql.columnar.backend.velox.checkUsageLeak")
```
jackylee-ch, line 599: It seem this config is a static conf, we need use `buildStaticConf` to avoid runtime change

PR #9333, opened 2025-04-15
[GLUTEN-9243][VL] Fix cuda docker image

zhouyuan:

The image `ghcr.io/facebookincubator/velox-dev:adapters` is not avaiable on ARM
followup on https://github.com/apache/incubator-gluten/pull/9229

Signed-off-by: Yuan Zhou <yuan.zhou@ibm.com>

## How was this patch tested?
pass GHA


PHILO-HE (approved):
Looks good!

ep/build-velox/src/build_velox.sh
```diff
@@ -128,7 +128,8 @@ function compile {
     COMPILE_OPTION="$COMPILE_OPTION -DVELOX_ENABLE_BENCHMARKS=ON"
   fi
   if [ $ENABLE_GPU == "ON" ]; then
-    COMPILE_OPTION="$COMPILE_OPTION -DVELOX_ENABLE_GPU=ON -DVELOX_ENABLE_CUDF=ON"
+    # the cuda default options are for Centos9 image from Meta
+    COMPILE_OPTION="$COMPILE_OPTION -DVELOX_ENABLE_GPU=ON -DVELOX_ENABLE_CUDF=ON -DCMAKE_CUDA_ARCHITECTURES=70 -DCMAKE_CUDA_COMPILER=/usr/local/cuda-12.8/bin/nvcc"
```
jinchengchenghh, line 132: I use the ENV here but without CUDA_COMPILER https://github.com/apache/incubator-gluten/pull/9333/files#diff-395a263b6f69bd7ef4991face3666e100acdd60e47f4b28e0737adbbf5fb945aR12,is all the environment CUDA COMPILER this path? Do we need to make it an environment variable in docker image?

jinchengchenghh:
Thanks for your fix.

.github/workflows/docker_image.yml
```diff
@@ -126,7 +101,7 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: ${{ env.DOCKERHUB_REPO }}
-          tags: centos-8-jdk8
+          tags: vcpkg-centos-8
```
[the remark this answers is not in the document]
zhouyuan: it's not changed actually - it's diff issue from github. here's the full file:
https://github.com/apache/incubator-gluten/blob/245ea1e22cf81e158fa101aa6b6d970ac17381f9/.github/workflows/docker_image.yml#L63

ep/build-velox/src/build_velox.sh
zhouyuan: In my tests it will report `cuda identifier is undefined in device code` if not setting these two variables. Note the GHA runner does not have NV GPU installed.

`CUDA_ARCHITECTURES`is able to read from `CMAKE_CUDA_ARCHITECTURES`
https://cmake.org/cmake/help/v3.28/prop_tgt/CUDA_ARCHITECTURES.html
however it seems there's no similar variable like `CUDA_COMPILERS`

jinchengchenghh (approved):

zhouyuan:
The fix does not work on GHA due to: 
- meta/velox missed one commit to allow skip building tests on GPU enabled 
- GHA runner disk space is not enough on building with meta centos9 adaptors
Will try to make a fix again.

PR #9336, opened 2025-04-15
[VL] Move pre-configuration code of dynamic off-heap sizing to its own place

zhztheplayer:
Dynamic off-heap sizing was introduced in https://github.com/apache/incubator-gluten/pull/5439. The PR does a few of improvements against the implementation.

1. Move pre-configuration code from `GlutenConfig.scala` to `DynamicOffHeapSizingMemoryTarget`.
2. Avoid chaining `DynamicOffHeapSizingMemoryTarget` with Spark consumer since it basically never triggers spills because of its design. The target will manage OOMs by itself.
3. Do not set off-heap size automatically when dynamic off-heap sizing is enabled. All the memory allocation is from a counter that indicates the remaining unused size from JVM on-heap memory cap.
4. Other minor improvements.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -188,74 +188,25 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
     // check memory off-heap enabled and size.
     checkOffHeapSettings(conf)
 
-    // Task slots.
+    // Get the off-heap size set by user.
+    val offHeapSize = conf.getSizeAsBytes(SPARK_OFFHEAP_SIZE_KEY)
+
+    // Set off-heap size in bytes.
+    conf.set(COLUMNAR_OFFHEAP_SIZE_IN_BYTES.key, offHeapSize.toString)
+
+    // Set off-heap size in bytes per task.
     val taskSlots = SparkResourceUtil.getTaskSlots(conf)
     conf.set(NUM_TASK_SLOTS_PER_EXECUTOR.key, taskSlots.toString)
-
-    val onHeapSize: Long = conf.getSizeAsBytes(SPARK_ONHEAP_SIZE_KEY, 1024 * 1024 * 1024)
-
-    // If dynamic off-heap sizing is enabled, the off-heap size is calculated based on the on-heap
-    // size. Otherwise, the off-heap size is set to the value specified by the user (if any).
-    // Note that this means that we will IGNORE the off-heap size specified by the user if the
-    // dynamic off-heap feature is enabled.
-    val offHeapSize: Long =
-      if (
-        conf.getBoolean(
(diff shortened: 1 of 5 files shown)


shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1634,7 +1637,7 @@ object GlutenConfig {
       .createWithDefault(true)
 
   val DYNAMIC_OFFHEAP_SIZING_ENABLED =
-    buildConf("spark.gluten.memory.dynamic.offHeap.sizing.enabled")
+    buildStaticConf("spark.gluten.memory.dynamic.offHeap.sizing.enabled")
```
zhouyuan, line 1640: is this a behavior change by changing this to static config?

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -188,74 +188,25 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
...
-      val fraction = 1.0d - conf.getDouble("spark.memory.storageFraction", 0.5d)
-      val conservativeOffHeapPerTask = (offHeapSize * fraction).toLong / taskSlots
-      conf.set(
-        COLUMNAR_CONSERVATIVE_TASK_OFFHEAP_SIZE_IN_BYTES.key,
-        conservativeOffHeapPerTask.toString)
-    }
+    // Pessimistic off-heap sizes, with the assumption that all non-borrowable storage memory
+    // determined by spark.memory.storageFraction was used.
+    val fraction = 1.0d - conf.getDouble("spark.memory.storageFraction", 0.5d)
+    val conservativeOffHeapPerTask = (offHeapSize * fraction).toLong / taskSlots
```
zhouyuan, line 206: it looks like changed the scope: `conservativeOffHeapPerTask` is effective on `DYNAMIC_OFFHEAP_SIZING_ENABLED=false` on previous code

zhouyuan (approved):
Thanks, the logic on dynamic off-heap sizing is cleaner.

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
zhztheplayer: Not really, it's just an improvement for configuration definitions.

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
zhztheplayer: We don't have to deal with off-heap size in the future anymore since we don't rely on off-heap memory manager.

We are just recovering the old code here before off-heap sizing was landed: https://github.com/apache/incubator-gluten/blame/022c208564dbb6dc59b7d64fbc4ad002c166e012/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala

PR #9341, opened 2025-04-16
[CORE] Provide an configuration option to completely turn off off-heap memory tracking with Spark memory manager

zhztheplayer:
We noticed some users unexpectedly rely on dynamic off-heap sizing to emulate a case that all the memory allocations are not tracked by Spark for testing or PoC purpose. As the feature dynamic off-heap sizing is not reliable itself (with wrong free on-heap memory calculations), we are providing a new option in this patch, `spark.gluten.memory.untracked` which will completely make all native allocations untracked by Spark when being set to true.

Note the new option is only be used for similar testing or PoC purpose as well. The previous usages on `spark.gluten.memory.dynamic.offHeap.sizing.enabled` can be changed to this new option because we are fixing the existing issues on the dynamic off-heap sizing feature which may cause more OOMs reported when that feature is on.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -153,6 +153,7 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
       // enabled. Skip the check.
       return
     }
+
     val minOffHeapSize = "1MB"
     if (
       !conf.getBoolean(GlutenConfig.SPARK_OFFHEAP_ENABLED, false) ||
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -135,6 +135,36 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
     }
   }
 
+  private def checkOffHeapSettings(conf: SparkConf): Unit = {
+    if (
+      conf.getBoolean(
+        DYNAMIC_OFFHEAP_SIZING_ENABLED.key,
+        DYNAMIC_OFFHEAP_SIZING_ENABLED.defaultValue.get)
+    ) {
+      // When dynamic off-heap sizing is enabled, off-heap mode is not strictly required to be
+      // enabled. Skip the check.
+      return
+    }
+
+    if (
+      conf.getBoolean(COLUMNAR_MEMORY_UNTRACKED.key, COLUMNAR_MEMORY_UNTRACKED.defaultValue.get)
+    ) {
+      // When untracked memory mode is enabled, off-heap mode is not strictly required to be
+      // enabled. Skip the check.
+      return
+    }
+    val minOffHeapSize = "1MB"
+    if (
+      !conf.getBoolean(GlutenConfig.SPARK_OFFHEAP_ENABLED, false) ||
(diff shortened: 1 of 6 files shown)


shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1241,6 +1243,16 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(false)
 
+  val COLUMNAR_MEMORY_UNTRACKED =
+    buildConf("spark.gluten.memory.untracked")
+      .internal()
+      .doc(
+        "When enabled, turn all native memory allocations in Gluten into untracked. Spark " +
```
zhouyuan, line 1250: based on the description, this feature should be effect only with `DYNAMIC_OFFHEAP_SIZING_ENABLED` case? or do you intend to introduce this feature in case with static off-heap also?

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -135,6 +135,37 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
...
+        DYNAMIC_OFFHEAP_SIZING_ENABLED.key,
+        DYNAMIC_OFFHEAP_SIZING_ENABLED.defaultValue.get)
+    ) {
+      // When dynamic off-heap sizing is enabled, off-heap mode is not strictly required to be
+      // enabled. Skip the check.
+      return
+    }
+
+    if (
+      conf.getBoolean(COLUMNAR_MEMORY_UNTRACKED.key, COLUMNAR_MEMORY_UNTRACKED.defaultValue.get)
```
zhouyuan, line 150: looking at the logic, if `DYNAMIC_OFFHEAP_SIZING_ENABLED=false`, and `COLUMNAR_MEMORY_UNTRACKED=true` then it will also skip the check of off-heap settings, is this inteded?

zhouyuan (commented):
Thanks. I have one directional question: is this new memory untrack feature depends on the dynamic off-heap sizing feature?

shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
zhztheplayer: It doesn't relate to the off-heap sizing feature, the idea use case is to allow user set 

```
spark.memory.offHeap.enabled=false
spark.gluten.memory.untracked=true
```

to bypass allocation tracking from Spark memory manager.

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
zhztheplayer: Yes, as mentioned in the other comment, it's allowed for user to set

```
spark.memory.offHeap.enabled=false
spark.gluten.memory.untracked=true
```

at the same time.

zhouyuan (approved):
+1 Works for me
My review notes in case someone is also looking: 
- originally gluten has built two ways of memory management 1) static off-heap 2) dynamic off-heap sizing. In both case Spark will track the memory allocations and report OOM issue  
- this patch adds a new feature to "ignore" the memory allocation so Spark will not introduce "OOM" error(but the OS may kill the application due to its big memory usage)

zhztheplayer:
> 2) dynamic off-heap sizing.

There are still some messy code need to sort out for off-heap sizing. I'll have another PR for that.

After this series of work I hope we can either continue on or immediately remove the off-heap sizing feature in future based on our decision. Because the code is made more independent by the effort.

PR #9364, opened 2025-04-18
[VL] Change isDistinct of distinct aggregateExpression to false

zml1206:

Vanilla spark will add an aggregate to remove duplicates for the distinct aggregation function, so velox does not need to process distinct. Moreover, currently velox distinct hash aggregation does not support spill.

## How was this patch tested?

Already existing UT.

### DIFF
--- a/backends-velox/src/test/resources/tpch-approved-plan/v1-bhj-ras/spark32/16.txt
+++ b/backends-velox/src/test/resources/tpch-approved-plan/v1-bhj-ras/spark32/16.txt
@@ -145,7 +145,7 @@ Results [4]: [p_brand#X, p_type#X, p_size#X, ps_suppkey#X]
 (21) FlushableHashAggregateExecTransformer
 Input [4]: [p_brand#X, p_type#X, p_size#X, ps_suppkey#X]
 Keys [3]: [p_brand#X, p_type#X, p_size#X]
-Functions [1]: [partial_count(distinct ps_suppkey#X)]
+Functions [1]: [partial_count(ps_suppkey#X)]
 Aggregate Attributes [1]: [count(ps_suppkey#X)#X]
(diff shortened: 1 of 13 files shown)


zhztheplayer:
Perhaps this change can speed up distinct aggregation processing?

zhztheplayer:
> Vanilla spark will add an aggregate to remove duplicates for the distinct aggregation function, so velox does not need to process distinct.

I am curious why Spark keeps the distinct flag after planing the distinct aggregation. Generally it's not needed by vanilla Spark as well. Have you investigated?

zml1206:
> I am curious why Spark keeps the distinct flag after planing the distinct aggregation. Generally it's not needed by vanilla Spark as well. Have you investigated?


Maybe Spark just didn't process it anymore, isDistinct is only used to create aggregate.

zml1206:
> Perhaps this change can speed up distinct aggregation processing?

Yes, it should also avoid the memory problem caused by velox distinct hash aggregation does not support spill.

zhztheplayer:
> > Perhaps this change can speed up distinct aggregation processing?
> 
> Yes, it should also avoid the memory problem caused by velox distinct hash aggregation does not support spill.

I am checking the code. Do we actually pass the flag to the substrait plan? I didn't find the relevant code. Would you help highlight the code here? Or else we never passed this flag to native? cc @rui-mo

zml1206:
> I am checking the code. Do we actually pass the flag to the substrait plan? I didn't find the relevant code. Would you help highlight the code here? Or else we never passed this flag to native? cc @rui-mo


Oh, I overlooked it. Yes, we did not pass `isDistinct` to the substrait plan. This change is unnecessary. Thank you. @zhztheplayer

zhztheplayer:
Got it. Thank you for the attempt anyway.

PR #9374, opened 2025-04-19
[VL] Ensure config.guess & config.sub files are executable while installing libelf

pkgajulapalli:

Docker image build fails sometimes as the config.guess/config.sub files are not executables.

## How was this patch tested?

Manually tried building docker image for centos8.


jinchengchenghh:
Do you know this issue? @zhouyuan

PR #5697, opened 2024-05-11
[GLUTEN-5696] Add preprojection support for ArrowEvalPythonExec

yma11:
## What changes were proposed in this pull request?

Add preprojection support for ArrowEvalPythonExec
Fixes: \#5696)

## How was this patch tested?

New UTs added

PR #9385, opened 2025-04-21
[VL] Fallback pandas udf when input is not an instance of AttributeReference

Surbhi-Vijay:
Fallback Pandas UDF when the input is not an instance of `AttributeReference`.
Example => `udf(1)`  `udf(count(null))`

We had encountered this error while trying to run Pandas testcases in `GlutenSQLQueryTestSuite`.
Error: `ColumnarArrowEvalPythonExec should only have [AttributeReference] inputs.`

The testcase was failing due to the commit: [[GLUTEN-5696] Add preprojection support for ArrowEvalPythonExec by yma11 · Pull Request #5697 · apache/incubator-gluten · GitHub](https://github.com/apache/incubator-gluten/pull/5697) where it expects each input to be `AttributeReference`

## How was this patch tested?
Tested by running Pandas testcases in our internal setup.

It is difficult to add the testcase for it as of now here. Once the support for running PySpark Pandas testcases is available, this scenario would be covered.


Surbhi-Vijay:
cc: @zhouyuan @yma11

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
```diff
@@ -211,12 +212,27 @@ case class ColumnarArrowEvalPythonExec(
...
 
   override def rowType0(): Convention.RowType = Convention.RowType.None
 
+  override protected def doValidateInternal(): ValidationResult = {
+    val (_, inputs) = udfs.map(collectFunctions).unzip
+    inputs.map {
+      input =>
+        input.map {
+          e =>
+            if (!e.isInstanceOf[AttributeReference]) {
```
jinchengchenghh, line 228: Can you also check `ColumnarArrowEvalPythonExec can't find referred input col.` in `doExecuteColumnar`? And remove the checks from `doExecuteColumnar`

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
```diff
@@ -211,12 +212,27 @@ case class ColumnarArrowEvalPythonExec(
     child: SparkPlan,
     evalType: Int)
   extends EvalPythonExec
-  with GlutenPlan {
+  with GlutenPlan
+  with ValidatablePlan {
```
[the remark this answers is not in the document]
Surbhi-Vijay: Resolved, thanks.

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
```diff
@@ -211,12 +211,26 @@ case class ColumnarArrowEvalPythonExec(
...
 
   override def rowType0(): Convention.RowType = Convention.RowType.None
 
+  override protected def doValidateInternal(): ValidationResult = {
+    val (_, inputs) = udfs.map(collectFunctions).unzip
+    inputs.foreach {
+      input =>
+        input.foreach {
+          e =>
+            if (!e.isInstanceOf[AttributeReference]) {
```
jinchengchenghh, line 226: Please use match case here

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
Surbhi-Vijay: Removed the check and verified that all pyspark testcases are passing in internal setup.

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
Surbhi-Vijay: Changed as suggested, thanks

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
```diff
@@ -311,19 +323,11 @@ case class ColumnarArrowEvalPythonExec(
...
-                    "ColumnarArrowEvalPythonExec should only have [AttributeReference] inputs.")
-                } else if (allInputs.exists(_.semanticEquals(e))) {
+                if (allInputs.exists(_.semanticEquals(e))) {
                   allInputs.indexWhere(_.semanticEquals(e))
                 } else {
-                  var offset: Int = -1
-                  offset = child.output.indexWhere(
+                  val offset = child.output.indexWhere(
                     _.exprId.equals(e.asInstanceOf[AttributeReference].exprId))
-                  if (offset == -1) {
```
jinchengchenghh, line 323: Also check if the exprId exists in `doValidateInternal`.

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
```diff
@@ -211,12 +211,28 @@ case class ColumnarArrowEvalPythonExec(
...
   override def batchType(): Convention.BatchType = ArrowJavaBatch
 
   override def rowType0(): Convention.RowType = Convention.RowType.None
 
+  override protected def doValidateInternal(): ValidationResult = {
+    val (_, inputs) = udfs.map(collectFunctions).unzip
+    inputs.foreach {
+      input =>
+        input.foreach {
+          case e: AttributeReference if e.exprId != null =>
```
jinchengchenghh, line 225: Not check the exprId is null, check the exprId exists in child output
if child.output.exists(o => o.exprId == e.exprId)

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
Surbhi-Vijay: `AttributeReference` will always have `exprId`. We do not need to check for it explicitly.
image

Surbhi-Vijay: Added as suggested.

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
Surbhi-Vijay: Got confused previously, added the condition now in doValidateInternal.

PR #9384, opened 2025-04-21
Fix ColumnarToRowRemovalGuard not able to be copied

yaooqinn:
## What changes were proposed in this pull request?

Fix a wrong constructor bug 

```
  java.lang.IllegalStateException: Failed to copy node.
Is otherCopyArgs specified correctly for ColumnarToRowRemovalGuard.
Exception message: wrong number of arguments
ctor: public org.apache.gluten.extension.columnar.MiscColumnarRules$PreventBatchTypeMismatchInTableCache$PreventColumnarTypeMismatchInTableCache$ColumnarToRowRemovalGuard(org.apache.gluten.extension.columnar.MiscColumnarRules$PreventBatchTypeMismatchInTableCache$PreventColumnarTypeMismatchInTableCache$,org.apache.spark.sql.execution.SparkPlan)?
types: class org.apache.gluten.extension.columnar.transition.TransitionSuite$BatchToRow
args: BatchToRow VanillaBatch$
+- Range (0, 1, step=1, splits=2)
  at org.apache.spark.sql.catalyst.trees.TreeNode.makeCopy(TreeNode.scala:854)
  at org.apache.spark.sql.catalyst.trees.TreeNode.mapChildren(TreeNode.scala:781)
  at org.apache.spark.sql.catalyst.trees.TreeNode.clone(TreeNode.scala:860)
  at org.apache.gluten.extension.columnar.MiscColumnarRulesSuite.$anonfun$new$1(MiscColumnarRulesSuite.scala:32)
  at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
  at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
  at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
  at org.scalatest.Transformer.apply(Transformer.scala:22)
  at org.scalatest.Transformer.apply(Transformer.scala:20)
  at org.scalatest.funsuite.AnyFunSuiteLike$$anon$1.apply(AnyFunSuiteLike.scala:226)
```

## How was this patch tested?

new ut

PR #9393, opened 2025-04-22
[VL] Follow up on #9384 to avoid swallowing exceptions in UT

zhztheplayer:
Following #9364, a minor refactor on the test code to avoid using "try-catch" as control statement and to avoid swallowing exceptions.


zhztheplayer:
@yaooqinn A minor follow-up for the previous PR. Thanks.

marin-ma (approved):

PR #9397, opened 2025-04-22
[GLUTEN-9335][VL] Support iceberg write unpartitioned table

jinchengchenghh:
Base on this PR https://github.com/facebookincubator/velox/pull/10996, which is merged to ibm/velox, and lacks for the metadata, so the read performance is not performed as expected. Use the flag `--enable_enhanced_features` to enable this feature, default disable. 
Use `org.apache.gluten.tags.EnhancedFeaturesTest` test Tag on the specified enhanced features tests to exclude, exclude the tests default by profile `exclude-tests`, we cannot use the jni call to decide if run the tests because the library is not loaded when listing the tests.

Only supports Spark34, spark35 iceberg version 1.5.0 is not supported.

Supports parquet format because avro and orc write is not supported in Velox

Fallback the complex data type write because the metric does not support.

### DIFF
--- a/gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergAppendDataExec.scala
+++ b/gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergAppendDataExec.scala
@@ -101,7 +101,7 @@ trait IcebergAppendDataExec extends ColumnarAppendDataExec {
     if (Seq("brotli, lzo").contains(codec)) {
       return ValidationResult.failed("Not support this codec " + codec)
     }
-    if (output.exists(a => !AvroSchemaUtil.makeCompatibleName(a.name).equals(a.name))) {
+    if (query.output.exists(a => !AvroSchemaUtil.makeCompatibleName(a.name).equals(a.name))) {
       return ValidationResult.failed("Not support the compatible column name")
     }
 


cpp/CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ option(ENABLE_HDFS "Enable HDFS" OFF)
 option(ENABLE_ORC "Enable ORC" OFF)
 option(ENABLE_ABFS "Enable ABFS" OFF)
 option(ENABLE_GPU "Enable GPU" OFF)
+option(ENABLE_ICEBERG_WRITE "Enable iceberg write" OFF)
```
zhztheplayer, line 60: What happens if a user turns it on right now? Given the Velox PR is not landed

backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/OffloadIcebergWrite.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.gluten.extension.columnar.offload.OffloadSingleNode
+import org.apache.gluten.extension.columnar.validator.Validators
+import org.apache.gluten.extension.injector.Injector
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.datasources.v2.AppendDataExec
+
+case class OffloadIcebergWrite() extends OffloadSingleNode {
+  override def offload(plan: SparkPlan): SparkPlan = plan match {
+    case a: AppendDataExec =>
+      VeloxIcebergAppendDataExec(a)
```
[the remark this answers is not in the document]
jinchengchenghh: VeloxIcebergAppendDataExec uses the jni wrapper to write iceberg, but for delta, the jni wrapper might be different, for example, delta lake may call the rust jni call.

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarAppendDataExec.scala
```diff
@@ -0,0 +1,119 @@
...
+        }
+      )
+
+      logInfo(s"Data source write support $batchWrite is committing.")
+      batchWrite.commit(messages)
+      logInfo(s"Data source write support $batchWrite committed.")
+      commitProgress = Some(
+        StreamWriterCommitProgressUtil.getStreamWriterCommitProgress(totalNumRowsAccumulator.value))
+    } catch {
+      case cause: Throwable =>
```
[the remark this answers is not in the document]
jinchengchenghh: This is copied from Spark

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarAppendDataExec.scala
```diff
@@ -0,0 +1,119 @@
...
+        logError(s"Data source write support $batchWrite is aborting.")
+        try {
+          batchWrite.abort(messages)
+        } catch {
+          case t: Throwable =>
```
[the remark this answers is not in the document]
jinchengchenghh: copied too

backends-velox/src-iceberg/main/scala/org/apache/gluten/connector/write/IcebergDataWriteFactory.scala
```diff
@@ -0,0 +1,63 @@
...
+   * object instance when sending data to the data writer, for better performance. Data writers
+   * are responsible for defensive copies if necessary, e.g. copy the data before buffer it in a
+   * list.
+   * <p>
+   * If this method fails (by throwing an exception), the corresponding Spark write task would fail
+   * and get retried until hitting the maximum retry times.
+   *
+   */
+  override def createWriter(): DataWriter[ColumnarBatch] = {
+    val(writerHandle, jniWrapper) = getJniWrapper(schema, format, directory, codec)
```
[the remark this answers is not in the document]
jinchengchenghh: The code style check make no effect on iceberg code, maybe we should fix it.

cpp/CMakeLists.txt
jinchengchenghh: The compile will fail

backends-velox/src-iceberg/main/scala/org/apache/gluten/connector/write/IcebergDataWriteFactory.scala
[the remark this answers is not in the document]
jinchengchenghh: com.diffplug.spotless:spotless-maven-plugin:2.27.2:check can check the module gluten-iceberg, but cannot check the backends-velox/src/iceberg introduced by `build-helper-maven-plugin`, scala style plugin can work well with src-iceberg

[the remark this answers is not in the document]
zhztheplayer: @jinchengchenghh I had opened this https://github.com/apache/incubator-gluten/pull/9478 either

gluten-substrait/src/main/java/org/apache/gluten/connector/write/ColumnarBatchDataWriterFactory.java
```diff
@@ -0,0 +1,50 @@
...
+ * A factory of {@link DataWriter} returned by {@link
+ * BatchWrite#createBatchWriterFactory(PhysicalWriteInfo)}, which is responsible for creating and
+ * initializing the actual data writer at executor side.
+ *
+ * <p>Note that, the writer factory will be serialized and sent to executors, then the data writer
+ * will be created on executors and do the actual writing. So this interface must be serializable
+ * and {@link DataWriter} doesn't need to be.
+ *
+ * @since 3.0.0
+ */
```
zhztheplayer: Let's rephrase the comments. Maybe we just convey that it's a companion interface with Spark's row-based version.

gluten-iceberg/src/main/java/org/apache/gluten/connector/write/ColumnarBatchWrite.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract class ColumnarBatchWrite implements BatchWrite {
+  @Override
+  public DataWriterFactory createBatchWriterFactory(PhysicalWriteInfo info) {
+    throw new UnsupportedOperationException();
+  }
+
+  public ColumnarDataWriterFactory createColumnarBatchWriterFactory(PhysicalWriteInfo info) {
+    throw new UnsupportedOperationException();
+  }
+}
```
zhztheplayer: Could move the API to `gluten-substrait`?

gluten-iceberg/src/main/java/org/apache/gluten/connector/write/ColumnarDataWriterFactory.java
```diff
@@ -0,0 +1,57 @@
...
+ * initializing the actual data writer at executor side.
+ *
+ * <p>Note that, the writer factory will be serialized and sent to executors, then the data writer
+ * will be created on executors and do the actual writing. So this interface must be serializable
+ * and {@link DataWriter} doesn't need to be.
+ *
+ * @since 3.0.0
+ */
+@Evolving
+public interface ColumnarDataWriterFactory extends Serializable {
```
zhztheplayer: Ditto, could move to `gluten-substrait` and revive the docs.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
```diff
@@ -158,4 +158,6 @@ trait BackendSettingsApi {
 
   def supportIcebergEqualityDeleteRead(): Boolean = true
 
+  def supportAppendDataExec(): Boolean = false
```
zhztheplayer: Do we have to add the API? Given the plan is offloaded out of the regular `tag-pull-offload` cycle.

Or else we could just decide whether to offload the node in `VeloxIcebergAppendDataExec` based on certain conditions.

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarAppendDataExec.scala
```diff
@@ -0,0 +1,119 @@
...
+  def writingTaskBatch: WritingColumnarBatchSparkTask[_] = DataWritingColumnarBatchSparkTask
+
+  override def doExecute(): RDD[InternalRow] = {
+    result
+    sparkContext.parallelize(Nil, 1)
+  }
+
+  def createFactory(schema: StructType): ColumnarBatchDataWriterFactory
+
+  protected def writeColumnarBatchWithV2(batchWrite: BatchWrite): Unit = {
```
zhztheplayer: If the code could be reused, can we follow Spark's pattern to move it into a trait like `ColumnarV2ExistingTableWriteExec` or something?

gluten-substrait/src/main/scala/org/apache/spark/sql/datasources/v2/AppendColumnarBatchDataExec.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.spark.TaskContext
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.connector.write._
+import org.apache.spark.sql.execution.datasources.v2.StreamWriterCommitProgress
+import org.apache.spark.sql.execution.metric.{CustomMetrics, SQLMetric}
+import org.apache.spark.sql.vectorized.ColumnarBatch
+import org.apache.spark.util.Utils
+
+case class DataWritingColumnarBatchSparkTaskResult(
+    numRows: Long,
```
zhztheplayer: Why the file is named "AppendColumnarBatchDataExec.scala"?

cpp/velox/compute/iceberg/IcebergWriter.h
```diff
@@ -0,0 +1,52 @@
...
+
+#pragma once
+
+#include "compute/iceberg/IcebergFormat.h"
+#include "memory/VeloxColumnarBatch.h"
+#include "velox/connectors/hive/iceberg/IcebergDataSink.h"
+
+namespace gluten {
+
+class IcebergWriter {
```
zhztheplayer, line 25: We had an V1 write API in C++

https://github.com/apache/incubator-gluten/blob/main/cpp/velox/operators/writer/VeloxDataSource.h

How can we possibly unify these writers together?

cpp/velox/jni/VeloxJniWrapper.cc
```diff
@@ -463,6 +463,59 @@ JNIEXPORT jlong JNICALL Java_org_apache_gluten_columnarbatch_VeloxColumnarBatchJ
   JNI_METHOD_END(kInvalidObjectHandle)
 }
 
+#ifdef GLUTEN_ENABLE_ICEBERG_WRITE
```
zhztheplayer: Can we wait a bit until we have the functionality public accessible in either upstream Velox or IBM Velox? I hope that will not block the schedule here but not sure.

zhztheplayer (commented):
Sorry for the delay here.

Let's see if we can generalize some of the APIs like `writeColumnarBatchWithV2`, then I think if the code added is manageable, we could use this approach in contrast to the fake row facilities.

Thanks!

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
jinchengchenghh: I suppose CH backends do not support AppendDataExec operator

gluten-substrait/src/main/scala/org/apache/spark/sql/datasources/v2/AppendColumnarBatchDataExec.scala
jinchengchenghh: In spark, it is WriteToDataSourceV2Exec.scala, we may need to rename `AppendColumnarBatchDataExec.scala` to `ColumnarWriteToDataSourceV2Exec.scala`

cpp/velox/compute/iceberg/IcebergWriter.h
jinchengchenghh: The API is different, for function commit vs close return type, and initialize arguments, Iceberg datasource needs partition spec, now it does not include in this PR but will be supported later. And VeloxDataSource onlys supports parquet, so it does not have the argument format, and the compressionKind gets from iceberg TableProperty not sparkConfs, so much difference makes it a separate class.

jinchengchenghh:
Could you help review again?Thanks! The first velox PR will be merged to ibm/velox after CI ready. After then, I will update the enhanced frature branch to ibm velox main. @zhztheplayer

dev/builddeps-veloxbe.sh
```diff
@@ -7,7 +7,7 @@
 # (the "License"); you may not use this file except in compliance with
 # the License.  You may obtain a copy of the License at
 #
-#    http://www.apache.org/licenses/LICENSE-2.0
+#    http://www.apache.org/licenses/LICENSE-2.0C
```
zhztheplayer: is this intentional?

gluten-core/src/main/resources/org/apache/gluten/proto/IcebergPartitionSpec.proto
```diff
@@ -0,0 +1,28 @@
...
+message IcebergPartitionField {
+  string name = 1;
+  TransformType transform = 2;
+  optional int32 parameter = 3;  // Optional parameter for transform config
+}
+
+message IcebergPartitionSpec {
+  int32 spec_id = 1;  // Field name uses snake_case per protobuf conventions
+  repeated IcebergPartitionField fields = 2;
+}
```
zhztheplayer: Possible to move to gluten-substrait or backend-velox? Thanks!

I understand it doesn't belong to the substrait protos but still it doesn't seem general to place here.

ep/build-velox/src/get_velox.sh
```diff
@@ -20,8 +20,8 @@ VELOX_REPO=https://github.com/oap-project/velox.git
 VELOX_BRANCH=2025_06_25
 VELOX_HOME=""
 RUN_SETUP_SCRIPT=ON
-VELOX_ENHANCED_REPO=https://github.com/oap-project/velox.git
-VELOX_ENHANCED_BRANCH=2025_06_24
+VELOX_ENHANCED_REPO=https://github.com/jinchengchenghh/velox.git
+VELOX_ENHANCED_BRANCH=2025_06_25
```
zhztheplayer: I have no strong opinion on this, but as long as there are no objections from everybody.

Are you going to change the repo to `ibm/velox` right in the next PR?

.github/workflows/velox_backend_enhanced_features.yml
```diff
@@ -0,0 +1,170 @@
...
+          name: velox-native-lib-enhanced-centos-7-${{github.sha}}
+          path: ./cpp/build/releases/
+          if-no-files-found: error
+      - uses: actions/upload-artifact@v4
+        with:
+          name: arrow-jars-enhanced-centos-7-${{github.sha}}
+          path: .m2/repository/org/apache/arrow/
+          if-no-files-found: error
+
+  spark-test-spark34:
```
zhztheplayer, line 114: Can we include Spark 3.5 as well? As long as there's no technical blockers it's better to have the lasted Spark version covered in test.

zhztheplayer (approved):
I know it might be difficult, but I hope we could finally be able to get rid of this two-repos design of the Velox dependency. More community discussions will be needed on that part.

ep/build-velox/src/get_velox.sh
jinchengchenghh: No, I will merge this PR after the velox PR is merged to ibm/velox, it will be soon

.github/workflows/velox_backend_enhanced_features.yml
jinchengchenghh: The spark 34 uses iceberg 1.7.1 version while spark35 uses iceberg 1.5.0 iceberg version which causes version mismatch, the reflection to get writeProperty from SparkWrite will failed in the test.

cpp/velox/compute/iceberg/IcebergWriter.cc
```diff
@@ -0,0 +1,174 @@
...
+            columnTypes.at(i),
+            columnTypes.at(i)));
+    } else {
+      columnHandles.push_back(
+        std::make_shared<connector::hive::HiveColumnHandle>(
+            columnNames.at(i),
+            connector::hive::HiveColumnHandle::ColumnType::kRegular,
+            columnTypes.at(i),
+            columnTypes.at(i)));
+  }
```
avevad, line 63: Looks like the '}' of else-clause is absent

cpp/velox/compute/iceberg/IcebergWriter.cc
```diff
@@ -0,0 +1,174 @@
...
+using namespace facebook::velox;
+using namespace facebook::velox::connector::hive;
+using namespace facebook::velox::connector::hive::iceberg;
+namespace {
+
+std::shared_ptr<connector::hive::LocationHandle> makeLocationHandle(
+    std::string targetDirectory,
+    std::optional<std::string> writeDirectory = std::nullopt,
+    connector::hive::LocationHandle::TableType tableType = connector::hive::LocationHandle::TableType::kExisting) {
+  return;
```
avevad: Return what? Does this PR even compile? I keep seeing syntax errors here and there, but I don't really understand how can one write such amount of code without compiling it

jinchengchenghh: One simple refactor commit breaks the compile, the PR has been tested, and wait for PR to merge to native side

j7nhai:
Originally, when writing with Iceberg, a field id is generated in the final Parquet file, and this field id corresponds to the field id in the Iceberg schema. It appears that the Velox Parquet writer is implemented based on Arrow, but Arrow's Parquet schema does not seem to support creating field ids. How did you solve this problem? @jinchengchenghh

PR #9405, opened 2025-04-23
[GLUTEN][VL] Enable RAS for CI Spark 3.3

ArnavBalyan:
 - RAS enabled is only running for Spark 3.5
 - Certain failed tests were previously only seen with RAS enabled.
 - Enabling it, since Spark 33 is has significant changes from 3.5 and users can introduce changes for lower versions.
 - Only enabling for 33 for now to save resources


zhztheplayer:
@ArnavBalyan 

> was not sure if there was some reason to enable only for 3.5,

It was for saving CI resources. So we just covered the latest version.

If there aren't other specific reasons, I think it's fine to keep it in 3.5 for now. Given the optimizer code is 99% similar among different Spark versions. We can tolerate a little coverage loss to save some Ci resource.

zhztheplayer:
> Let me close this for now, can revisit in future if needed? @zhztheplayer

Yes, thanks!

If you have local test failures, perhaps add a local UT with RAS=ON to cover that specific case :)

PR #9419, opened 2025-04-25
[VL] Fix Hudi scan fallback

xushiyan:

When file format is `NewHoodieParquetFileFormat`, the condition was matched and the plan was passed to velox for execution. But the result wasn't correct - merging wasn't performed, only parquet records were returned.

We should fallback to spark for `NewHoodieParquetFileFormat`

## How was this patch tested?

Manual tests.


zhouyuan (approved):
Thanks for the fix!

xushiyan:
@zhouyuan thanks for the approval. is it good to merge ?

Issue #9383, opened 2025-04-21
[VL][1.2] OrderBy operator borrows large memory


wForget:
### Backend

VL (Velox)

### Bug description

error message:

```
25/04/21 16:16:07 ERROR ManagedReservationListener: Error reserving memory from target
org.apache.gluten.memory.memtarget.ThrowOnOomMemoryTarget$OutOfMemoryException: Not enough spark off-heap execution memory. Acquired: 1976.0 MiB, granted: 1008.0 MiB. Try tweaking config option spark.memory.offHeap.size to get larger space to run this application (if spark.gluten.memory.dynamic.offHeap.sizing.enabled is not enabled). 
Current config settings: 
	spark.gluten.memory.offHeap.size.in.bytes=2.0 GiB
	spark.gluten.memory.task.offHeap.size.in.bytes=2.0 GiB
	spark.gluten.memory.conservative.task.offHeap.size.in.bytes=1024.0 MiB
	spark.memory.offHeap.enabled=true
	spark.gluten.memory.dynamic.offHeap.sizing.enabled=false
Memory consumer stats: 
	Task.15:                                              Current used bytes: 1040.0 MiB, peak bytes:        N/A
	\- Gluten.Tree.0:                                     Current used bytes: 1040.0 MiB, peak bytes:    2.0 GiB
	   \- root.0:                                         Current used bytes: 1040.0 MiB, peak bytes:    2.0 GiB
	      +- WholeStageIterator.0:                        Current used bytes: 1040.0 MiB, peak bytes:    2.0 GiB
	      |  \- single:                                   Current used bytes: 1040.0 MiB, peak bytes:    2.0 GiB
	      |     +- root:                                  Current used bytes:   50.4 MiB, peak bytes: 2041.0 MiB
	      |     |  +- task.Gluten_Stage_1_TID_15_VTID_0:  Current used bytes:   50.4 MiB, peak bytes:   57.0 MiB
	      |     |  |  +- node.0:                          Current used bytes:   44.9 MiB, peak bytes:   52.0 MiB
	      |     |  |  |  +- op.0.0.0.TableScan:           Current used bytes:   44.9 MiB, peak bytes:   48.8 MiB
	      |     |  |  |  \- op.0.0.0.TableScan.test-hive: Current used bytes:      0.0 B, peak bytes:      0.0 B
	      |     |  |  +- node.2:                          Current used bytes:    3.8 MiB, peak bytes:    4.0 MiB
[...]

PR #9424, opened 2025-04-25
[GLUTEN-9383][VL] Fix leak when growing capacity

wForget:

When allocationChanged fails in growCapacity, we should free reclaimed bytes

Fixes: #9383

## How was this patch tested?

After this repair, the failed job of #9383 succeeded

### DIFF
--- a/cpp/velox/memory/VeloxMemoryManager.cc
+++ b/cpp/velox/memory/VeloxMemoryManager.cc
@@ -168,7 +168,13 @@ class ListenableArbitrator : public velox::memory::MemoryArbitrator {
     }
     auto reclaimedFreeBytes = shrinkPool(pool, 0);
     auto neededBytes = velox::bits::roundUp(bytes - reclaimedFreeBytes, memoryPoolTransferCapacity_);
-    listener_->allocationChanged(neededBytes);
+    try {
+      listener_->allocationChanged(neededBytes);
+    } catch (const std::exception& e) {
+      // if allocationChanged failed, we need to free the reclaimed bytes
+      listener_->allocationChanged(-reclaimedFreeBytes);
+      throw e;
+    }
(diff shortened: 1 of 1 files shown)


zhztheplayer (approved):

cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -168,7 +168,13 @@ class ListenableArbitrator : public velox::memory::MemoryArbitrator {
     }
     auto reclaimedFreeBytes = shrinkPool(pool, 0);
     auto neededBytes = velox::bits::roundUp(bytes - reclaimedFreeBytes, memoryPoolTransferCapacity_);
-    listener_->allocationChanged(neededBytes);
+    try {
+      listener_->allocationChanged(neededBytes);
+    } catch (const std::exception& e) {
+      // if allocationChanged failed, we need to free the reclaimed bytes
+      listener_->allocationChanged(-reclaimedFreeBytes);
+      throw e;
```
marin-ma, line 176: We should use `throw;` instead of `throw e;`. Otherwise the original exception type will be cast down to `std::exception`.

wForget: > We should use `throw;` instead of `throw e;`. 

Thank you, I will send a follow up PR to change it

> Otherwise the original exception type will be cast down to `std::exception`.

Sorry, I'm not familiar with c++, could this cause any problems? how can I handle it?

wForget: Is it ok like this?

```
    try {
      listener_->allocationChanged(neededBytes);
    } catch (const std::exception&) {
      // if allocationChanged failed, we need to free the reclaimed bytes
      listener_->allocationChanged(-reclaimedFreeBytes);
      std::rethrow_exception(std::current_exception());
    }
```

PR #9294, opened 2025-04-10
[VL] Fix weekly build job on Centos 7

PHILO-HE:
## What changes were proposed in this pull request?

This issue was related to the recent pr #9207 which changes the script to use upstream Velox wget_and_untar function.

PR #9444, opened 2025-04-28
[VL] Replace wget w/ wget_and_untar to make dependency download more reliable

yaooqinn:

Using wget_and_untar from upstream that accepts options for download tools

### DIFF
--- a/ep/build-velox/src/setup-centos7.sh
+++ b/ep/build-velox/src/setup-centos7.sh
@@ -163,9 +163,8 @@ function install_boost {
 
 function install_protobuf {
   cd "${DEPENDENCY_DIR}"
-  wget https://github.com/protocolbuffers/protobuf/releases/download/v21.4/protobuf-all-21.4.tar.gz
-  tar -xzf protobuf-all-21.4.tar.gz
-  cd protobuf-21.4
+  wget_and_untar https://github.com/protocolbuffers/protobuf/releases/download/v21.4/protobuf-all-21.4.tar.gz protobuf
+  cd protobuf
   ./configure  CXXFLAGS="-fPIC"  --prefix=/usr/local
   make "-j$(nproc)"
   $SUDO make install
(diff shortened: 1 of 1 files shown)


ep/build-velox/src/setup-centos7.sh
```diff
@@ -163,19 +163,17 @@ function install_boost {
 
 function install_protobuf {
   cd "${DEPENDENCY_DIR}"
-  wget https://github.com/protocolbuffers/protobuf/releases/download/v21.4/protobuf-all-21.4.tar.gz
-  tar -xzf protobuf-all-21.4.tar.gz
-  cd protobuf-21.4
+  wget_and_untar https://github.com/protocolbuffers/protobuf/releases/download/v21.4/protobuf-all-21.4.tar.gz protobuf
+  cd protobuf
```
PHILO-HE, line 167: Use ` pushd protobuf`, then add `popd` at the end of this function. Then, we can avoid wrong directory may be possibly used when executing some other functions. See https://github.com/apache/incubator-gluten/pull/9294.

ep/build-velox/src/setup-centos7.sh
```diff
@@ -163,19 +163,17 @@ function install_boost {
...
   $SUDO make install
 }
 
 function install_gtest {
   cd "${DEPENDENCY_DIR}"
-  wget https://github.com/google/googletest/archive/refs/tags/release-1.12.1.tar.gz
-  tar -xzf release-1.12.1.tar.gz
-  cd googletest-release-1.12.1
+  wget_and_untar https://github.com/google/googletest/archive/refs/tags/release-1.12.1.tar.gz googletest
+  cd googletest
```
PHILO-HE, line 176: ditto

PHILO-HE (commented):
Thanks for your work!

ep/build-velox/src/setup-centos7.sh
yaooqinn: 👍

PHILO-HE (approved):
Thanks!

PR #9450, opened 2025-04-28
[CORE] Support batch scan customMetrics

Zouxxyy:

before

image

after

image

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
@@ -95,7 +95,12 @@ abstract class BatchScanExecTransformerBase(
 
   // Note: "metrics" is made transient to avoid sending driver-side metrics to tasks.
   @transient override lazy val metrics: Map[String, SQLMetric] =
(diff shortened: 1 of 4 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
```diff
@@ -95,7 +95,12 @@ abstract class BatchScanExecTransformerBase(
 
   // Note: "metrics" is made transient to avoid sending driver-side metrics to tasks.
   @transient override lazy val metrics: Map[String, SQLMetric] =
-    BackendsApiManager.getMetricsApiInstance.genBatchScanTransformerMetrics(sparkContext)
+    BackendsApiManager.getMetricsApiInstance.genBatchScanTransformerMetrics(
+      sparkContext) ++ customMetrics
+
+  def doPostDriverMetrics(): Unit = {
```
jinchengchenghh, line 101: Why add the function instead of directly call `postDriverMetrics`

Zouxxyy: because Symbol postDriverMetrics is inaccessible from this place

jinchengchenghh (approved):

PR #9469, opened 2025-04-29
[GLUTEN-9468][VL] Remove parquet-arrow reader dependency in benchmarks

marin-ma:
Currently parquet-arrow reader is only used in Velox benchmarks to read input data. Replace with Velox parquet reader and remove the <parquet/arrow/reader.h> headers

Also found that using parquet-arrow reader causes the microbenchmarks to crash on MacOS. The microbenchmarks works fine after switching to the velox parquet reader.


marin-ma:
@Yohahaha @jinchengchenghh Could you help to review? Thanks!

cpp/velox/benchmarks/ColumnarToRowBenchmark.cc
```diff
@@ -15,153 +15,54 @@
  * limitations under the License.
```
jinchengchenghh, line 15: This benchmark can be deleted because Velox has a similar benchmark, and we only call the velox API

jinchengchenghh (commented):
Can we remove the arrow::parquet dependency from core/CMakeLists?
target_link_libraries(gluten PUBLIC Arrow::parquet Arrow::arrow
                                    Arrow::arrow_bundled_dependencies), and does not need to compile arrow with parquet

jinchengchenghh:
If remove the arrow writer, we can remove arrow parquet dependency.

jinchengchenghh (approved):

marin-ma:
> Can we remove the arrow::parquet dependency from core/CMakeLists? target_link_libraries(gluten PUBLIC Arrow::parquet Arrow::arrow Arrow::arrow_bundled_dependencies), and does not need to compile arrow with parquet

That's the final goal. We need to remove the arrow writer used for dumping parquet input. Next I will replace it with the parquet writer in Velox, or it's also an option to directly write the serialized RowVectors as @Yohahaha suggested.

jinchengchenghh:
Both are ok for me.

jinchengchenghh:
If we write the parquet file, we may easily use spark to check the input though no one does that so.

Yohahaha (approved):

PR #9472, opened 2025-04-30
[VL] Support decimal as double in gluten-it

jinchengchenghh:
Cudf does not supports decimal type because cudf table decimal lost the `precision` information, so convert to double to test.

### DIFF
--- a/tools/gluten-it/common/src/main/java/org/apache/gluten/integration/BaseMixin.java
+++ b/tools/gluten-it/common/src/main/java/org/apache/gluten/integration/BaseMixin.java
@@ -80,6 +80,9 @@ public class BaseMixin {
   @CommandLine.Option(names = {"--scan-partitions"}, description = "Scan partition number. This is an approximate value, so the actual scan partition number might vary around this value. -1 for letting Spark choose an appropriate number.", defaultValue = "-1")
   private int scanPartitions;
 
+  @CommandLine.Option(names = {"--decimal-as-double"}, description = "Generate double value for decimal type column", defaultValue = "false")
+  private boolean decimalAsDouble;
+
   @CommandLine.Option(names = {"--extra-conf"}, description = "Extra Spark config entries applying to generated Spark session. E.g. --extra-conf=k1=v1 --extra-conf=k2=v2")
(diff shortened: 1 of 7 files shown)


jinchengchenghh:
Not sure if we need a `no-decimal` suffix for the data generated directory.
Now the --data-gen mode is always as default value, so it won't affect the data, if set it to once, user may need to delete the data folder manually.

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/Suite.scala
```diff
@@ -195,6 +196,10 @@ abstract class Suite(
 
   private[integration] def allQueryIds(): Array[String]
 
+  protected[integration] def typeModifiers(): List[TypeModifier] = {
+    if (decimalAsDouble) List(new DecimalTypeModifier()) else List()
+  }
+
```
zhztheplayer, line 202: nit: Could make it `private[integration]` and move it to line 189. It can still be overridden without `protected` since it's not final.

tools/gluten-it/common/src/main/java/org/apache/gluten/integration/BaseMixin.java
```diff
@@ -80,6 +80,9 @@ public class BaseMixin {
   @CommandLine.Option(names = {"--scan-partitions"}, description = "Scan partition number. This is an approximate value, so the actual scan partition number might vary around this value. -1 for letting Spark choose an appropriate number.", defaultValue = "-1")
   private int scanPartitions;
 
+  @CommandLine.Option(names = {"--decimal-as-double"}, description = "Generate double value for decimal type column", defaultValue = "false")
```
zhztheplayer, line 83: nit: value -> values

zhztheplayer:
@jinchengchenghh I remember such feature required for change on the TPC SQL queries. No?

jinchengchenghh:
I run the queres with TPCH and TPCDS, the queries can pass. I think if we use date as int, we need to change the query

zhztheplayer (approved):

Issue #8227, opened 2024-12-13
[VL] Result mismatch in CollectList when 'sort by' is involved


NEUpanning:
### Backend

VL (Velox)

### Bug description

Reproducing SQL:
```sql
CREATE OR REPLACE TEMP VIEW temp_table AS
SELECT * FROM VALUES
  (1, 'a'), (1, 'b'), (1, 'c'),
  (2, 'd'), (2, 'e'), (2, 'f'),
  (3, 'g'), (3, 'h'), (3, 'i')
AS t(id, value);

SELECT 1-id, collect_list(value) AS values_list
FROM (
  select * from
  (SELECT id, value
  FROM temp_table
  DISTRIBUTE BY rand()) 
  DISTRIBUTE BY id sort by id,value
) t
GROUP BY 1;
```

**Results:**

The vanilla result is deterministic and values_list is sorted by value column:

id   values_list
1    ["a", "b", "c"]
2    ["d", "e", "f"]
3    ["g", "h", "i"]
The gluten result is non-deterministic and values_list is not sorted, e.g. :

id   values_list
1    ["a", "c", "b"]
3    ["g", "i", "h"]
2    ["f", "e", "d"]

**gluten physical plan:**
```
== Physical Plan ==
AdaptiveSparkPlan isFinalPlan=true
+- == Final Plan ==
   VeloxColumnarToRowExec
   +- ^(4) HashAggregateTransformer(keys=[(1 - id#0)#15], functions=[velox_collect_list(value#1)], isStreamingAgg=false, output=[(1 - id)#4, values_list#2])
      +- ^(4) InputIteratorTransformer[(1 - id#0)#15, buffer#11]
         +- CustomShuffleReader coalesced
            +- ShuffleQueryStage 2
               +- ColumnarExchange hashpartitioning((1 - id#0)#15, 20), ENSURE_REQUIREMENTS, [(1 - id#0)#15, buffer#11], [id=#1137], [id=#1137], [OUTPUT] List((1 - id#0):IntegerType, buffer:ArrayType(StringType,false)), [OUTPUT] List((1 - id#0):IntegerType, buffer:ArrayType(StringType,false))
                  +- VeloxAppendBatches 3276
                     +- ^(3) ProjectExecTransformer [hash((1 - id#0)#15, 42) AS hash_partition_key#16, (1 - id#0)#15, buffer#11]
[...]

PR #9473, opened 2025-04-30
[GLUTEN-8227][VL] fix: Update sort elimination rules for Hash Aggregate

acvictor:

Update sort elimination rules for hash aggregate only when the base aggregate was a sort aggregate. Only if a sort aggregate is transformed into a hash aggregate, can we safely assume that the aggregation does not depend on input order 

(Part of fix for: \#8227)

## How was this patch tested?

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
@@ -716,6 +716,46 @@ case class RegularHashAggregateExecTransformer(
     ignoreNullKeys
   ) {
 
+  override def isOffloadedSortExec: Boolean = false
+
+  override protected def allowFlush: Boolean = false
+
+  override def simpleString(maxFields: Int): String =
+    s"${super.simpleString(maxFields)}"
+
+  override def verboseString(maxFields: Int): String =
+    s"${super.verboseString(maxFields)}"
+
+  override protected def withNewChildInternal(newChild: SparkPlan): HashAggregateExecTransformer = {
+    copy(child = newChild)
+  }
+}
+
+// Hash aggregation that is offloaded from sort aggregation.
+// Is identical to RegularHashAggregateExecTransformer but with a 
+// different value of isOffloadedSortExec.
+case class OffloadedSortHashAggregateExecTransformer(
+    requiredChildDistributionExpressions: Option[Seq[Expression]],
+    groupingExpressions: Seq[NamedExpression],
+    aggregateExpressions: Seq[AggregateExpression],
+    aggregateAttributes: Seq[Attribute],
+    initialInputBufferOffset: Int,
+    resultExpressions: Seq[NamedExpression],
+    child: SparkPlan,
+    ignoreNullKeys: Boolean = false)
+  extends HashAggregateExecTransformer(
+    requiredChildDistributionExpressions,
+    groupingExpressions,
+    aggregateExpressions,
+    aggregateAttributes,
+    initialInputBufferOffset,
+    resultExpressions,
+    child,
+    ignoreNullKeys
+  ) {
+
+  override def isOffloadedSortExec: Boolean = true
+
   override protected def allowFlush: Boolean = false
 
   override def simpleString(maxFields: Int): String =
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/FlushableHashAggregateRule.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/FlushableHashAggregateRule.scala
@@ -83,12 +83,25 @@ case class FlushableHashAggregateRule(session: SparkSession) extends Rule[SparkP
   }
 
   private def replaceEligibleAggregates(plan: SparkPlan)(
-      func: RegularHashAggregateExecTransformer => SparkPlan): SparkPlan = {
+      func: HashAggregateExecTransformer => SparkPlan): SparkPlan = {
     def transformDown: SparkPlan => SparkPlan = {
       case agg: RegularHashAggregateExecTransformer
           if !agg.aggregateExpressions.forall(p => p.mode == Partial || p.mode == PartialMerge) =>
         // Not a intermediate agg. Skip.
         agg
+      case agg: RegularHashAggregateExecTransformer
+          if isAggInputAlreadyDistributedWithAggKeys(agg) =>
+        // Data already grouped by aggregate keys, Skip.
+        agg
+      case agg: RegularHashAggregateExecTransformer
+          if aggregatesNotSupportFlush(agg.aggregateExpressions) =>
+        agg
+      case agg: RegularHashAggregateExecTransformer =>
+        func(agg)
+      case agg: RegularHashAggregateExecTransformer
+        if !agg.aggregateExpressions.forall(p => p.mode == Partial || p.mode == PartialMerge) =>
+      // Not a intermediate agg. Skip.
+      agg
       case agg: RegularHashAggregateExecTransformer
           if isAggInputAlreadyDistributedWithAggKeys(agg) =>
         // Data already grouped by aggregate keys, Skip.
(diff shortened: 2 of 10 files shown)


acvictor:
@zhztheplayer @NEUpanning can you please help review this PR?

acvictor:
@ulysses-you can you please take a look at this PR?

zhztheplayer:
Compared to having this flag, can we just add a new operator for offloaded sort aggregation?

acvictor:
> Compared to having this flag, can we just add a new operator for offloaded sort aggregation?

@zhztheplayer can you please review again? Thanks!

acvictor:
@rui-mo can you please review this PR?

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
```diff
@@ -716,6 +716,46 @@ case class RegularHashAggregateExecTransformer(
...
+
+  override protected def withNewChildInternal(newChild: SparkPlan): HashAggregateExecTransformer = {
+    copy(child = newChild)
+  }
+}
+
+// Hash aggregation that is offloaded from sort aggregation.
+// Is identical to RegularHashAggregateExecTransformer but with a
+// different value of isOffloadedSortExec.
+case class OffloadedSortHashAggregateExecTransformer(
```
rui-mo: How about the name `HashFromSortAggregateExecTransformer`, which emphasizes it is a hash aggregate but converted from sort aggregate?

rui-mo (commented):
LGTM. Let's waiting for a final review from @zhztheplayer. Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -78,7 +78,8 @@ trait SparkPlanExecApi {
       aggregateAttributes: Seq[Attribute],
       initialInputBufferOffset: Int,
       resultExpressions: Seq[NamedExpression],
-      child: SparkPlan): HashAggregateExecBaseTransformer
+      child: SparkPlan,
+      offloadedSortExec: Boolean = false): HashAggregateExecBaseTransformer
```
zhztheplayer: minor: Could just remove the default value if we don't need to maintain backward compatibility for the API.

gluten-substrait/src/main/scala/org/apache/gluten/execution/HashAggregateExecBaseTransformer.scala
```diff
@@ -188,7 +190,9 @@ object HashAggregateExecBaseTransformer {
     case a: SortAggregateExec => a.initialInputBufferOffset
   }
 
-  def from(agg: BaseAggregateExec): HashAggregateExecBaseTransformer = {
+  def from(
+      agg: BaseAggregateExec,
+      offloadedSortExec: Boolean = false): HashAggregateExecBaseTransformer = {
```
zhztheplayer: ditto

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
```diff
@@ -716,6 +716,46 @@ case class RegularHashAggregateExecTransformer(
...
+    aggregateExpressions,
+    aggregateAttributes,
+    initialInputBufferOffset,
+    resultExpressions,
+    child,
+    ignoreNullKeys
+  ) {
+
+  override def isOffloadedSortExec: Boolean = true
+
```
zhztheplayer: Let's override the `requiredChildOrdering` and `outputOrdering` so that this API doesn't have to be added.

The implementation may be the same with Spark's `SortAggrtegateExec`:

https://github.com/apache/spark/blob/abecd4affbd9102d73434caf1f1ca00bda9ef6fe/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/SortAggregateExec.scala#L50-L56

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
acvictor: Done.

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
acvictor: Do you mean `requiredChildOrdering` and `orderingExpressions`?

gluten-substrait/src/main/scala/org/apache/gluten/execution/HashAggregateExecBaseTransformer.scala
acvictor: Done.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
acvictor: Done.

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
acvictor: > Let's override the `requiredChildOrdering` and `outputOrdering` so that this API doesn't have to be added.
> 
> The implementation may be the same with Spark's `SortAggrtegateExec`:
> 
> https://github.com/apache/spark/blob/abecd4affbd9102d73434caf1f1ca00bda9ef6fe/sql/core/src/main/scala/org/apache/spark/sql/execution/aggregate/SortAggregateExec.scala#L50-L56

@zhztheplayer I tried this approach but it does eliminate the sort in some cases, for example `drop redundant partial sort which has pre-project when offload sortAgg`. Do you think it is okay to retain the bool API?

acvictor: @zhztheplayer I also realized that we can have flushable aggregate from sort aggregate too. Is it better to just add a boolean to HashAggregateExecTransformer? Else I will have to add another FlushableHashFromSortAggregateExec.

FelixYBW:
@zhztheplayer Any more comments?

zhztheplayer:
Apologize for the delay here! I'll revisit this shortly. Please feel free to jump in and review if anyone likes to.

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
zhztheplayer: Apologize for the relay. 

If we have to add a boolean flag, how about directly adding `SortAggregateExecTransformer`? Then we can decide whether to remove the sort based on the operator class. It was actually suggested in comment https://github.com/apache/incubator-gluten/pull/9473#issuecomment-2863912230 which I haven't seen is addressed. Would you like to help check this? Thanks.

acvictor: > Apologize for the relay.
> 
> If we have to add a boolean flag, how about directly adding `SortAggregateExecTransformer`? Then we can decide whether to remove the sort based on the operator class. It was actually suggested in comment [#9473 (comment)](https://github.com/apache/incubator-gluten/pull/9473#issuecomment-2863912230) which I haven't seen is addressed. Would you like to help check this? Thanks.

@zhztheplayer sorry for dropping the ball on this. I have updated the PR based on your comments. Can you please take a look?

backends-velox/src/main/scala/org/apache/gluten/extension/FlushableHashAggregateRule.scala
```diff
@@ -83,7 +83,7 @@ case class FlushableHashAggregateRule(session: SparkSession) extends Rule[SparkP
   }
 
   private def replaceEligibleAggregates(plan: SparkPlan)(
-      func: RegularHashAggregateExecTransformer => SparkPlan): SparkPlan = {
+      func: HashAggregateExecTransformer => SparkPlan): SparkPlan = {
```
rui-mo, line 92: Could you please add some comments for this function for readability? Thanks.

acvictor:
@zhztheplayer @rui-mo can you please review this again? The main change was in response to the comment "how about directly adding SortAggregateExecTransformer? Then we can decide whether to remove the sort based on the operator class."

PR #9477, opened 2025-04-30
[VL] Support ObjectDebugInfo in ObjectStore during destruction of native runtime

ArnavBalyan:
 - Following up on #9455
 - Added ObjectDebugInfo which can hold other attributes for the object (such as size) and used for logging.

## How was this patch tested?
 - Existing UTs

### DIFF
--- a/cpp/core/utils/ObjectStore.h
+++ b/cpp/core/utils/ObjectStore.h
@@ -34,6 +34,16 @@ using StoreHandle = int32_t;
 using ObjectHandle = int64_t;
 constexpr static ObjectHandle kInvalidObjectHandle = -1;
 
+template <typename T>
+struct SafeSizeOf {
+  static constexpr size_t value = sizeof(T);
+};
+
+template <>
+struct SafeSizeOf<void> {
+  static constexpr size_t value = 0;
+};
+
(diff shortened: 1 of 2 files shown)


cpp/core/utils/ObjectStore.h
```diff
@@ -74,14 +74,20 @@ class ObjectStore {
   ObjectHandle save(std::shared_ptr<T> obj) {
     const std::lock_guard<std::mutex> lock(mtx_);
     const std::string_view description = typeid(T).name();
```
zhztheplayer, line 76: Could rename `description` to `typeName`?

zhztheplayer (commented):
Can we reword all the appearances of "description" with "type name"? Thanks!

cpp/core/utils/ObjectStore.h
```diff
@@ -73,15 +83,21 @@ class ObjectStore {
...
+    aliveObjects_.emplace(handle, ObjectDebugInfo{typeName, size});
     return toObjHandle(handle);
   }
 
  private:
   static ResourceMap<ObjectStore*>& stores();
 
+  struct ObjectDebugInfo {
+    std::string_view typeName;
+    size_t size;
```
zhztheplayer, line 98: nit: Possible to add const modifiers?

zhztheplayer (approved):

PR #9479, opened 2025-04-30
[branch-1.4][VL] Fix docker image name for 1.4 branch

zhouyuan:

Adding "1.4" to docker image name used in branch-1.4, so the changes in main branch does not impact the branch-1.4

A follow-up patch will modify the CI to use the right docker

## How was this patch tested?

no tests required

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -18,7 +18,7 @@ name: Build and Push Docker Image
 on:
   push:
     branches:
-      - main
+      - branch-1.4
     paths:
       - '.github/workflows/docker_image.yml'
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_image.yml
```diff
@@ -18,7 +18,7 @@ name: Build and Push Docker Image
 on:
   push:
     branches:
-      - main
+      - branch-1.4
```
zhouyuan, line 21: Cc @weiting-chen

PHILO-HE (approved):
Looks good! We can remove these images from docker hub once branch-1.4 is not maintained.

Issue #9502, opened 2025-05-03
[VL] Remove unused datatype check code from concat function


dcoliversun:
### Description

The concat function only supports string / binary / array type[1], we could remove unused datatype check code from concat function

[1] https://github.com/apache/spark/blob/v3.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala#L2195

### Gluten version

None

PR #9503, opened 2025-05-03
[GLUTEN-9502][VL] Remove useless datatype check for concat function

dcoliversun:

This PR aims to remove unused datatype check code from concat function, as this function only supports string / binary / array type in apache spark[1].  

[1] https://github.com/apache/spark/blob/v3.2.0/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala#L2195

  

(Fixes: \#9502)

## How was this patch tested?


dcoliversun:
cc @PHILO-HE Could you please have a review for this? Thanks!

PR #9506, opened 2025-05-04
[CORE] RasOffload: Fix false positive discard

li-boxuan:

In RasOffload.scala, we discard a plan if the offload doesn't succeed. We do this check by `offloaded.fastEquals(rewritten)`. This is not the most efficient way, and has false positive discards. For example, in TPC-DS, we have observed that q44 & q49 didn't offload `WindowExec` to `WindowExecTransformer` with legacy RAS, and the reason is `WindowExecTransformer` is deemed as "equal" to its `WindowExec` counterparty. Consequently, step 6 would return empty list, causing `WindowExecTransformer` not to enter RAS coster.

## How was this patch tested?

We have a set of internal tests to compare TPC-DS plans. With this PR, default gluten (without RAS) and gluten with RAS (legacy coster) generate exactly same plan for all TPC-DS queries.


zhztheplayer (approved):
+1

Why a `WindowExecTransformer` could equal to `WindowExec` ? They are of different Scala types so I assume the equals call should fail fast from the beginning.

PR #9483, opened 2025-04-30
[GLUTEN-9468][VL] Remove parquet-arrow dependency

marin-ma:
Replace parquet-arrow writer with velox parquet writer to dump input for microbenchmark.
Refactor the code for dumping input for microbenchmark.

PR #9511, opened 2025-05-06
[GLUTEN-9468][VL] Follow up: Remove parquet-arrow cpp build

marin-ma:
#9483 has removed the dependency on arrow parquet.

### DIFF
--- a/dev/build_arrow.sh
+++ b/dev/build_arrow.sh
@@ -37,6 +37,7 @@ function build_arrow_cpp() {
  pushd $ARROW_PREFIX/cpp
 
  cmake_install \
+       -DARROW_PARQUET=OFF \
        -DARROW_FILESYSTEM=ON \
        -DARROW_PROTOBUF_USE_SHARED=OFF \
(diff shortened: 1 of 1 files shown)


marin-ma:
cc: @zhouyuan

Yohahaha (approved):
LGTM! will test it in our dynamic link build process later.

FelixYBW:
@zhouyuan can we add a doc to list the arrow reference in Gluten? like the csv scan.

Yohahaha (approved):

zhouyuan:
> @zhouyuan can we add a doc to list the arrow reference in Gluten? like the csv scan.

I made some quick check. Let me check where is the best to place this document
On C++ side:
- Arrow memory pool
- Arrow IPC/IPC: io/Schema/Compression 
- Arrow bridge
-Arrow CSV scan
 
On java side:
- Arrow schema
- Arrow record batch
- Arrow memory pool 

Thanks, -yuan

Issue #9475, opened 2025-04-30
[VL][1.2] Multiple copies were created during serialization of broadcast columnar batches, resulting in excessive memory usage


wForget:
### Description

An OOM exception occurred during broadcast, but the input of the task was only 69.3 MiB.

```
25/04/29 16:30:44 WARN ManagedReservationListener: Error reserving memory from target
org.apache.gluten.memory.memtarget.ThrowOnOomMemoryTarget$OutOfMemoryException: Not enough spark off-heap execution memory. Acquired: 1056.0 MiB, granted: 120.0 MiB. Try tweaking config option spark.memory.offHeap.size to get larger space to run this application (if spark.gluten.memory.dynamic.offHeap.sizing.enabled is not enabled). 
Current config settings: 
	spark.gluten.memory.offHeap.size.in.bytes=4.0 GiB
	spark.gluten.memory.task.offHeap.size.in.bytes=4.0 GiB
	spark.gluten.memory.conservative.task.offHeap.size.in.bytes=2.0 GiB
	spark.memory.offHeap.enabled=true
	spark.gluten.memory.dynamic.offHeap.sizing.enabled=false
Memory consumer stats: 
	Task.10:                                              Current used bytes:    3.9 GiB, peak bytes:        N/A
	\- Gluten.Tree.7:                                     Current used bytes:    3.9 GiB, peak bytes:    4.0 GiB
	   \- root.7:                                         Current used bytes:    3.9 GiB, peak bytes:    4.0 GiB
	      +- BroadcastUtils#serializeStream.7:            Current used bytes:    2.1 GiB, peak bytes:    2.2 GiB
	      |  \- single:                                   Current used bytes:    2.1 GiB, peak bytes:    2.1 GiB
	      |     +- gluten::MemoryAllocator:               Current used bytes: 1063.2 MiB, peak bytes: 1063.2 MiB
	      |     \- root:                                  Current used bytes: 1059.2 MiB, peak bytes: 1064.0 MiB
	      |        \- default_leaf:                       Current used bytes: 1059.2 MiB, peak bytes: 1059.2 MiB
	      +- WholeStageIterator.7:                        Current used bytes: 1848.0 MiB, peak bytes: 1864.0 MiB
	      |  \- single:                                   Current used bytes: 1848.0 MiB, peak bytes: 1864.0 MiB
	      |     +- root:                                  Current used bytes: 1841.9 MiB, peak bytes: 1864.0 MiB
[...]

PR #9521, opened 2025-05-06
[GLUTEN-9475][VL] Serialize ColumnarBatch one by one to reduce memory footprint when broadcasting

wForget:

Serialize ColumnarBatch one by one when broadcasting

Fixes: #9475

## How was this patch tested?

After this fix, the original abnormal job will no longer have offheap OOM.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
@@ -173,19 +173,19 @@ object BroadcastUtils {
       return ColumnarBatchSerializeResult.EMPTY
     }
     var rowNums = 0
-    val values = filtered.map(b => {
-      val handle = ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, b)
-      rowNums += b.numRows()
-      try {
-        ColumnarBatchSerializerJniWrapper
-          .create(
-            Runtimes
+    val values = filtered.map(
+      b => {
+        val handle = ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, b)
+        rowNums += b.numRows()
+        try {
+          ColumnarBatchSerializerJniWrapper
+            .create(Runtimes
               .contextInstance(BackendsApiManager.getBackendName, "BroadcastUtils#serializeStream"))
-          .serialize(handle)
-      } finally {
-        ColumnarBatches.release(b)
-      }
-    })
+            .serialize(handle)
+        } finally {
+          ColumnarBatches.release(b)
+        }
+      })
     new ColumnarBatchSerializeResult(rowNums, values)
   }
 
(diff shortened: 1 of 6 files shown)


wForget:
Benchmark testing did not find performance regression: https://github.com/wForget/benchmarks-spark-native/actions/runs/14861256989

First: https://github.com/apache/incubator-gluten/commit/a783515e30ebfc99bb22c5222d8e2a98f92d9e63
Second: https://github.com/wForget/gluten/tree/GLUTEN-9475

tpcds_allqueries

wForget:
@zhztheplayer @PHILO-HE  Could you please take a look?

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
```diff
@@ -172,19 +172,21 @@ object BroadcastUtils {
...
+            .create(Runtimes
               .contextInstance(BackendsApiManager.getBackendName, "BroadcastUtils#serializeStream"))
-          .serialize(handleArray)
-      } finally {
-        filtered.foreach(ColumnarBatches.release)
-      }
-    serializeResult
+            .serialize(handle)
+        } finally {
+          ColumnarBatches.release(b)
```
xinghuayu007, line 186: So here release 'b' every loop can reduce memory consumption?

xinghuayu007, line 186: So release 'b' every loop can reduce memory consumption?

wForget:
> Please add some unit tests to prove this feature's correctness.

Do you mean I need to create a query with broadcast to verify it doesn't result in incorrect results? Currently, the unit test cases should already have many queries with broadcast.

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
xinghuayu007: Thanks for your explains.

xinghuayu007:
> > Please add some unit tests to prove this feature's correctness.
> 
> Do you mean I need to create a query with broadcast to verify it doesn't result in incorrect results? Currently, the unit test cases should already have many queries with broadcast.

ok

PHILO-HE (approved):
Looks good! Will merge the pr if no other comment from the community.

zhztheplayer (approved):
:+1:

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala
zhztheplayer: @wForget 

Did you observe a decrement of peak off-heap usage after this change? As shown in code, `filtered` is still of the type `Array[ColumnarBatch]` which may cause high usage from the beginning of the loop.

wForget: > Did you observe a decrement of peak off-heap usage after this change? 

After this change, job no longer has offheap OOM. Is this enough to indicate that offheap peek memory has decreased?

> As shown in code, filtered is still of the type Array[ColumnarBatch] which may cause high usage from the beginning of the loop.

Makes sense, this is indeed an improvement, I will try to convert it to an iterator

Issue #9526, opened 2025-05-06
[CH] Add an option to not rename to libch.so


baibaichen:
### Description

In the production environment, the library is renamed to libch.so. However, this naming convention creates a development dependency where re-linking is required during each debug run.

### Gluten version

None

PR #9527, opened 2025-05-06
[GLUTEN-9526][CH] Add cmake option to create symbol instead of rename

baibaichen:

This commit introduces a new CMake option `USE_SYMLINK` to the ClickHouse  local engine build system. When enabled, this option creates a symbolic link to the library file instead of renaming it. This approach helps maintain the original file while making it accessible via the expected name.

The default behavior (rename) is preserved when the option is not set.

(Fixes: \#9526)

## How was this patch tested?

Existed UTs


lgbo-ustc (approved):
LGTM

PR #9531, opened 2025-05-06
[VL] Support user fallback option for CollectTail

ArnavBalyan:
 - Supports config driven fallbacks for CollectTail, added to user fallback options.
 - Also updated the collecttail rule to rely on it instead which was previously on an incorrect config.
 
## How was this patch tested?
 - Existing UTs


shims/common/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1759,4 +1760,10 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(false)
 
+  val COLUMNAR_COLLECT_TAIL_ENABLED =
+    buildConf("spark.gluten.sql.columnar.collectTail")
+      .internal()
```
PHILO-HE, line 1765: Seems we should remove `internal` as this config is public to users? cc @yikf

PHILO-HE (approved):
Thanks!

Issue #9518, opened 2025-05-06
[VL][FLINK] Support watermark assigner


shuai-xu:
### Description

Flink has a operator to generate watarmak, need to support it.

### Gluten version

None

PR #9541, opened 2025-05-07
[GLUTEN-9518][FLINK] Support watermark assigner

shuai-xu:

This pr support WatermarkAssginer. 

(Fixes: \#9518)

## How was this patch tested?

### DIFF
--- a/gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/FunctionMappings.java
+++ b/gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/FunctionMappings.java
@@ -36,7 +36,7 @@ public class FunctionMappings {
             put("CASE", new DefaultFunctionConverter("if"));
             put("*", new DefaultFunctionConverter("multiply"));
             put("-", new SubtractFunctionConverter("subtract"));
-            put("MOD", new DefaultFunctionConverter("pmod"));
+            put("MOD", new DefaultFunctionConverter("remainder"));
             put("AND", new DefaultFunctionConverter("and"));
         }
     };
--- a/gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/FunctionMappings.java
+++ b/gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/FunctionMappings.java
@@ -16,25 +16,32 @@
  */
 package org.apache.gluten.rexnode;
 
+import org.apache.gluten.rexnode.functions.DefaultFunctionConverter;
+import org.apache.gluten.rexnode.functions.FunctionConverter;
+import org.apache.gluten.rexnode.functions.SubtractFunctionConverter;
+
 import java.util.HashMap;
 import java.util.Map;
 
 /** Mapping of flink function and velox function. */
 public class FunctionMappings {
     // A map stores the relationship between flink function name and velox function.
-    private static Map<String, String> functionMappings = new HashMap() {
+    private static Map<String, FunctionConverter> functionMappings = new HashMap() {
         {
             // TODO: support more functions.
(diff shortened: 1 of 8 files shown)


gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/SubtractFunctionConverter.java
```diff
@@ -0,0 +1,54 @@
...
+    public SubtractFunctionConverter(String function) {
+        this.function = function;
+    }
+
+    @Override
+    public CallTypedExpr toVeloxFunction(Type nodeType, List<TypedExpr> params) {
+        Preconditions.checkNotNull(params.size() == 2, "Subtract must contain exactly two parameters");
+
+        // TODO: need refine for more type cast
+        if (params.get(0).getReturnType() != params.get(1).getReturnType()) {
```
zhztheplayer, line 40: Is this code reliable as `!=` only compares references?

shuai-xu: Oh, We need to add compare function to Type.

zhztheplayer: Yes we can open a issue for that. So do we still need to perform this check in the PR?

shuai-xu: Use class compare first

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/FunctionMappings.java
```diff
@@ -16,25 +16,32 @@
...
-            put("CASE", "if");
-            put("*", "multiply");
+            put(">", new DefaultFunctionConverter("greaterthan"));
+            put("<", new DefaultFunctionConverter("lessthan"));
+            put("=", new DefaultFunctionConverter("equalto"));
+            put("CAST", new DefaultFunctionConverter("cast"));
+            put("CASE", new DefaultFunctionConverter("if"));
+            put("*", new DefaultFunctionConverter("multiply"));
+            put("-", new SubtractFunctionConverter("subtract"));
+            put("MOD", new DefaultFunctionConverter("pmod"));
```
PHILO-HE, line 39: Should be mapped to "remainder". "pmod" has different meaning, which returns positive remainder.

gluten-flink/runtime/src/main/java/org/apache/gluten/util/LogicalTypeConverter.java
```diff
@@ -58,6 +59,9 @@ public static Type toVLType(LogicalType logicalType) {
             return new io.github.zhztheplayer.velox4j.type.DecimalType(
                     decimalType.getPrecision(),
                     decimalType.getScale());
+        } else if (logicalType instanceof DayTimeIntervalType) {
+            // TODO: it seems interval now can be used as bigint for nexmark.
+            return new io.github.zhztheplayer.velox4j.type.BigIntType();
```
PHILO-HE, line 64: Can we use `IntervalDayTimeType()`? IntervalDayTimeType is implemented based on BigIntType in Velox, but they are logically different.

shuai-xu: IntervalDayTimeType can not be cast to BigIntType now in velox. Change it to BigIntType temporary, I will assgin this issuse to others to fix it.

PR #9545, opened 2025-05-07
[FLINK] Bump Velox4J version

zhztheplayer:
~~This is an example work to rebase `bigo/velox4j` on to the latest `velox4j/velox4j` code.~~

~~Relying on this rebased Velox4J branch: https://github.com/zhztheplayer/velox4j/commits/wip-bigo-rebase-0507.~~

Bump Velox4J to commit [c069ca63aa3a200b49199f7dfa285f011ce6fda5](https://github.com/bigo-sg/velox4j/commits/c069ca63aa3a200b49199f7dfa285f011ce6fda5)

### DIFF
--- a/gluten-flink/docs/Flink.md
+++ b/gluten-flink/docs/Flink.md
@@ -49,7 +49,7 @@ As some features have not been committed to upstream, you have to use the follow
 git clone https://github.com/bigo-sg/velox4j.git
 cd velox4j
 git checkout -b gluten origin/gluten
-git reset --hard 3275df667458a494a7ba5be93907779d1b013979
+git reset --hard c069ca63aa3a200b49199f7dfa285f011ce6fda5
 mvn clean install
 ```
 **Get gluten**
--- a/gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/operators/GlutenSingleInputOperator.java
+++ b/gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/operators/GlutenSingleInputOperator.java
@@ -20,9 +20,10 @@ package org.apache.gluten.table.runtime.operators;
 import io.github.zhztheplayer.velox4j.connector.ExternalStream;
 import io.github.zhztheplayer.velox4j.connector.ExternalStreamConnectorSplit;
 import io.github.zhztheplayer.velox4j.connector.ExternalStreamTableHandle;
+import io.github.zhztheplayer.velox4j.connector.ExternalStreams;
 import io.github.zhztheplayer.velox4j.iterator.DownIterators;
 import io.github.zhztheplayer.velox4j.iterator.UpIterator;
-import io.github.zhztheplayer.velox4j.query.BoundSplit;
+import io.github.zhztheplayer.velox4j.query.SerialTask;
 import io.github.zhztheplayer.velox4j.type.RowType;
 import org.apache.gluten.streaming.api.operators.GlutenOperator;
 import org.apache.gluten.vectorized.FlinkRowToVLVectorConvertor;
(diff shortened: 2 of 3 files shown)


lgbo-ustc:
Should L135 be remove?
```java
inRv.close();
```

zhztheplayer:
> Should L135 be removed?
> 
> ```java
> inRv.close();
> ```

Hi @lgbo-ustc , the close call is still needed. A Java side close call releases Java side reference to that row-vector only. The C++ side reference will still be held by the native blocking queue [here](https://github.com/velox4j/velox4j/blob/1e5cb702b449f7cb5b8817c87fa2487721626a3d/src/main/cpp/main/velox4j/iterator/BlockingQueue.h#L61). Hence, the row-vector will be completely released as soon as it's taken from the native blocking queue and is finished using by the Velox task.

lgbo-ustc:
> > Should L135 be removed?
> > ```java
> > inRv.close();
> > ```
> 
> Hi @lgbo-ustc , the close call is still needed. A Java side close call releases Java side reference to that row-vector only. The C++ side reference will still be held by the native blocking queue [here](https://github.com/velox4j/velox4j/blob/1e5cb702b449f7cb5b8817c87fa2487721626a3d/src/main/cpp/main/velox4j/iterator/BlockingQueue.h#L61). Hence, the row-vector will be completely released as soon as it's taken from the native blocking queue and is finished using by the Velox task.

I encountered some issues when trying to release the input row vector here. When it reaches L135 and release the vector, the row vector is still held by the C++ side, causing a lookup exception.​​ But I can't reproduce the issue now. 

​​For safety, I believe it’s better to let the C++ side take ownership and handle its release.
- After pushing into the queue, the Java side no longer seems to need the row vector.
- The Java side cannot determine whether the C++ side still requires the object.

lgbo-ustc:
> > Should L135 be removed?
> > ```java
> > inRv.close();
> > ```
> 
> Hi @lgbo-ustc , the close call is still needed. A Java side close call releases Java side reference to that row-vector only. The C++ side reference will still be held by the native blocking queue [here](https://github.com/velox4j/velox4j/blob/1e5cb702b449f7cb5b8817c87fa2487721626a3d/src/main/cpp/main/velox4j/iterator/BlockingQueue.h#L61). Hence, the row-vector will be completely released as soon as it's taken from the native blocking queue and is finished using by the Velox task.

I encountered an issue releasing the object here when using the previous version of the blocking queue. The object is still required by C++ side when we release it at line 135, causing an lookup exception in C++ side.

The object is still required by the C++ side when released at line 135, causing a lookup exception in the C++ code.

Does this new version of the blocking queue ensure objects can be safely released after being enqueued, even if they aren't processed immediately or are still needed by other components?

lgbo-ustc:
BTW, I previously tested the new version of the blocking queue without calling close() on the Java side, but I didn’t notice any warnings about resources not being properly closed.

zhztheplayer:
> Does this new version of the blocking queue ensure objects can be safely released after being enqueued in the JAVA side , even if they aren't processed immediately or are still needed by other components in the C++ side?

Yes exactly. This is why we had https://github.com/velox4j/velox4j/pull/173 replaced the previous Java blocking queue solution, because it was difficult to decide when to release the Java row-vector after it's pushed into the Java queue. But now with https://github.com/velox4j/velox4j/pull/173 we can release the row-vector right after it's pushed into the native queue, as long as it's unneeded by Java code.

shuai-xu:
please uppdate velox4j commit to c069ca63aa3a200b49199f7dfa285f011ce6fda5

shuai-xu (approved):
LGTM, thanks.

PR #9546, opened 2025-05-07
[VL] Refactor iceberg scan validation

jinchengchenghh:


FelixYBW:
Can you add desc?

jinchengchenghh:
I try to check the table is `BaseTable` and then use the BaseTable to do the validation, but seems like spark 3.2/spark3.3 is not the BaseTable, I will close this PR

PR #9591, opened 2025-05-11
[CORE] Remove unnecessary conf spark.gluten.sql.columnar.enableVanillaVectorizedReaders

Zouxxyy:

Users can configure Spark's own columnar read configuration, and we do not need to introduce additional gluten configuration to control it. (Especially when the configuration defaults to true)

## How was this patch tested?


backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRoughCostModelSuite.scala
```diff
@@ -43,7 +43,6 @@ class VeloxRoughCostModelSuite extends VeloxWholeStageTransformerSuite {
   override protected def sparkConf: SparkConf = super.sparkConf
     .set(GlutenConfig.RAS_ENABLED.key, "true")
     .set(GlutenConfig.RAS_COST_MODEL.key, "rough")
-    .set(GlutenConfig.VANILLA_VECTORIZED_READERS_ENABLED.key, "false")
```
Zouxxyy, line 46: Checked, adding this configuration has no effect on the results of the following cases. Because no matter
enable: FileScan(Vectorized) -> ColumnarToRow
disabled: FileScan(row)
their costs are the same, and have to add the r2c node later, if don't want fallback

PR #9596, opened 2025-05-12
Enable base64 unbase64

zhli1142015:

Depends on https://github.com/facebookincubator/velox/pull/13238 and https://github.com/facebookincubator/velox/pull/13185

## How was this patch tested?

UT


rui-mo (commented):
Thanks for opening the relevant Gluten PR! Would you resolve the conflict and ensure the CI passes?

zhouyuan:
@zhli1142015 Would you please do a rebase? It seems the two patch are landed Velox already

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -753,6 +764,8 @@ object ExpressionConverter extends SQLConfHelper with Logging {
           substraitExprName,
           expr.children.map(replaceWithExpressionTransformer0(_, attributeSeq, expressionsMap)),
           j)
+      case u: UnBase64 if SparkShimLoader.getSparkShims.unBase64FunctionFailsOnError(u) =>
```
zhouyuan, line 767: does this mean it has some limitation? 
Would better to add some notes here: 
https://github.com/apache/incubator-gluten/blob/main/docs/velox-backend-limitations.md#json-functions

zhouyuan (approved):
Thanks!

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
zhli1142015: This just trigger a fallback, create a follow up task for it.
https://github.com/apache/incubator-gluten/issues/10150

Issue #9567, opened 2025-05-08
[VL][FLINK] Velox4J query plan binary serialization


zhztheplayer:
### Description

Gluten-flink performs Java / Kyro serializations on the Velox4J query plan. This currently requires an [API change](https://github.com/bigo-sg/velox4j/blob/1239b1ea42961cacbe38294a01304f2acc0f10f3/src/main/java/io/github/zhztheplayer/velox4j/serde/NativeBean.java#L5-L6) in `bigo-sg/velox4j` to make `NativeBean` extend `Serializable`.

This change is reasonable but so far there are some bean definitions in Velox4J not compatible with Java `Serializable`. The beans typically have Java optionals which are not serializable by design as class members. We should find a better way to address the issue and make the beans Java-serializable in Velox4J.

### Gluten version

None

PR #9600, opened 2025-05-12
[GLUTEN-9567][FLINK] Add container class `Velox4JBean` for proper JSON-based object serialization

zhztheplayer:
There is a change in bigo-sg/velox4j to mark `NativeBean` as serializable:

https://github.com/bigo-sg/velox4j/commit/6dbe131daa399fa6432e1bd25f7857ee6ca325e8#diff-cff66579a3561ef0f1870c5d5bc1b35761d3eb585ce581bde23a4e964d7c5c36R21

This practice may cause Flink serialization to fail, when the bean implementation class has Java optionals defined as class member (which is common in Velox4J), or when there are  anonymous classes / lambda definitions in the class.

The patch introduces `Velox4JBean` which internally uses JSON serde as Java serde, the API is well-maintained in Velox4J so it's more reliable for Flink's use. 

The `NativeBean extends Serializable` change in Velox4J will not be needed with this change.

### DIFF
--- a/gluten-flink/runtime/src/main/java/org/apache/gluten/util/Velox4JBean.java
+++ b/gluten-flink/runtime/src/main/java/org/apache/gluten/util/Velox4JBean.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.apache.gluten.util;
 
 import com.google.common.base.Preconditions;


gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/operators/GlutenSingleInputOperator.java
```diff
@@ -50,19 +49,17 @@
...
 
     private static final Logger LOG = LoggerFactory.getLogger(GlutenSingleInputOperator.class);
 
-    private final PlanNode glutenPlan;
+    private final Velox4JBean<PlanNode> glutenPlan;
     private final String id;
-    private final RowType inputType;
-    private final RowType outputType;
+    private final Velox4JBean<RowType> inputType;
+    private final Velox4JBean<RowType> outputType;
```
shuai-xu, line 62: If outputType changes to Map<String, RowType>, does it work?

shuai-xu (commented):
It seems to work, but looks a little strange. I am not sure whether it's a general way. Can we remove the Optional in Velox4j, and make it serializable?

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/operators/GlutenSingleInputOperator.java
zhztheplayer: This will work: `Map<String, Velox4JBean<RowType>>`

zhztheplayer:
> It seems to work, but looks a little strange. I am not sure whether it's a general way. Can we remove the Optional in Velox4j, and make it serializable?

Not only the optionals, there are some lambda usages that could fail the serialization. So it's not a trivial work to just make the beans Java-serializable.

Having said that, let's consider the patch a temporary fix as it does complicate the code-writing. But it's a safer solution than just marking the beans as serializable.

shuai-xu (approved):
LGTM, need we create a issue and relate to it?

zhztheplayer:
@shuai-xu will link it to https://github.com/apache/incubator-gluten/issues/9567

jinchengchenghh (approved):

zhztheplayer:
This will cause a serialization error from client side due to the beans are not registered. Reverting...

Issue #9625, opened 2025-05-14
[VL][GCS] Application Hangs when GCS configs are not passed correctly.


pratham76:
### Backend

VL (Velox)

### Bug description

To access data from GCS object stores, the documentation mentions user to use the following authentication configurations.

```
spark.hadoop.fs.gs.auth.type                         SERVICE_ACCOUNT_JSON_KEYFILE
spark.hadoop.fs.gs.auth.service.account.json.keyfile // path to the json file with the credentials.
```

If the config `spark.hadoop.fs.gs.auth.type` is not set nothing is passed to velox 

https://github.com/apache/incubator-gluten/blob/08a71fe6639c40fe6fb44f58f03904148f89fbad/cpp/velox/utils/ConfigExtractor.cc#L189C3-L194C54

This would cause the application to hang as velox does not receive the credentials for the GCS path it is trying to access with the following warning in executor logs.

```
W20250512 07:54:31.934717   187 GcsFileSystem.cpp:323] Config hive.gcs.json-key-file-path is empty or key file path not found
``` 

### Gluten version

main branch

### Spark version

Spark-3.4.x

PR #9627, opened 2025-05-14
[GLUTEN-9625][VL] Add explicit GCS auth type check for service account key file

pratham76:

Have added validation for the case when `spark.hadoop.fs.gs.auth.type` is not set, or invalid. This would fail the application with prompt msg indicating the missing or invalid config.

Fixes: #9625 

## How was this patch tested?
Manual tests

Have tested the fix by not passing the required `spark.hadoop.fs.gs.auth.type`, and the validation does throw following error and the apllication does not gets stuck.

```
W20250514 06:09:42.919829   215 ConfigExtractor.cc:201] STARTUP: conf spark.hadoop.fs.gs.auth.service.account.json.keyfile is set, but conf spark.hadoop.fs.gs.auth.type is not SERVICE_ACCOUNT_JSON_KEYFILE
25/05/14 06:09:42 ERROR NativeBackendInitializer: Failed to call native backend's initialize method
org.apache.gluten.exception.GlutenException: Conf spark.hadoop.fs.gs.auth.type is missing or incorrect

```


zhouyuan (approved):
Thanks!

PR #9634, opened 2025-05-14
[GLUTEN-8851][VL] Validate the plan and use runtime config to enable it

jinchengchenghh:
Construct a Task to get the operators after DriverAdapter adapts, then validate the operators if it satisfies the offload condition, now only check it can be fully offloaded to cudf without extra data format conversion cost.
Further more, we can compare the cost of the plan offload to cudf may with fallback and cpu operators.

The cudf and wave uses operators instead of PlanNode to validate the plan, it benefits Project validation which can be easily validated by FilterProject::exprsAndProjection.

Set the velox QueryConfig `cudf.enabled` to apply cudf adapter or not.
Related PR: https://github.com/facebookincubator/velox/pull/13341

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -575,7 +575,8 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
 
 #ifdef GLUTEN_ENABLE_GPU
     if (veloxCfg_->get<bool>(kCudfEnabled, false)) {
-      configs[cudf_velox::kCudfEnabled] = "false";
+      // TODO: wait for PR https://github.com/facebookincubator/velox/pull/13341
+      // configs[cudf_velox::kCudfEnabled] = "false";
     }
 #endif
 
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -21,6 +21,9 @@
 #include "velox/connectors/hive/HiveConfig.h"
 #include "velox/connectors/hive/HiveConnectorSplit.h"
 #include "velox/exec/PlanNodeStats.h"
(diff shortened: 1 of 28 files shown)


cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -572,6 +576,12 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
     configs[velox::core::QueryConfig::kSparkLegacyStatisticalAggregate] =
         std::to_string(veloxCfg_->get<bool>(kSparkLegacyStatisticalAggregate, false));
 
+#ifdef GLUTEN_ENABLE_GPU
+    if (!enableCudf_) {
```
zhztheplayer, line 580: @jinchengchenghh

Is there any reason for passing `enableCudf` by method calls down to here, instead of reading a configuration option? I am a bit concerned about such many APIs have to be changed for adding the parameter?

gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -198,6 +208,12 @@ trait TransformSupport extends ValidatablePlan {
 
   // When true, it will not generate relNode, nor will it generate native metrics.
   def isNoop: Boolean = false
+
+  // If set the isCudf, the config will send to native backend,
+  // and enable the cudf plan or not in runtime stage
+  def setIsCudf: Unit = {
+    isCudf = true
+  }
```
zhztheplayer, line 216: A flag like this will break Spark's plan immutability / equality? For example, the subquery reuse rule will find a non-cudf node and use it to replace a cudf node since they are actually equal.

zhztheplayer (commented):
Thank you @jinchengchenghh.

zhztheplayer (approved):

PR #9656, opened 2025-05-15
Revert "[GLUTEN-9567][FLINK] Add container class `Velox4JBean` for proper JSON-based object serialization"

zhztheplayer:
Reverts apache/incubator-gluten#9600 which causes a serialization error from client side due to the beans are not registered.


lgbo-ustc:
> @lgbo-ustc Able to approve this? Thanks!

done

PR #9669, opened 2025-05-16
[FLINK] Fix incorrect Flink-to-Velox conversion on varbinary literals

zhztheplayer:
`new VarBinaryValue(..)` the constructor only accepts base64 string as input. The patch fix the usage. That constructor should be enclosed by `private` later on.


shuai-xu (approved):
LGTM, thanks

Issue #9684, opened 2025-05-19
[VL] Add default config.guess and config.sub


zml1206:
### Description

Some environments prohibit access to git.savannah.gnu.org, such as Huawei Cloud's overseas machines. So consider maintaining a default config.guess and config.sub.

### Gluten version

None

PR #9685, opened 2025-05-19
[GLUTEN-9684][VL] Add default config.guess and config.sub

zml1206:

(Fixes: \#9684)

## How was this patch tested?

Local build test.
```
-- Using source at /opt/gluten/dev/vcpkg/.vcpkg/buildtrees/libelf/src/libelf-0-93dd260ed8.clean
CMake Warning at /opt/gluten/dev/vcpkg/ports/libelf/portfile.cmake:20 (message):
  Download config.guess failed！code: 28, message: "Timeout was reached"
Call Stack (most recent call first):
  scripts/ports.cmake:168 (include)


CMake Warning at /opt/gluten/dev/vcpkg/ports/libelf/portfile.cmake:30 (message):
  Download config.sub failed！code: 28, message: "Timeout was reached"
Call Stack (most recent call first):
  scripts/ports.cmake:168 (include)
```


zml1206:
@zhztheplayer what do you think？

zhouyuan:
@zml1206 thanks for the fix
these two files are actually to make configure work on ARM(to detect the latest arch), on X86 it's not necessary 
there's a small conflict, could you please do a rebase?

dev/vcpkg/ports/libelf/portfile.cmake
```diff
@@ -14,10 +14,22 @@ vcpkg_extract_source_archive(
 file(DOWNLOAD "https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD"
      "${SOURCE_PATH}/config.guess"
      STATUS status_guess)
+list(GET status_guess 0 status_code)
+list(GET status_guess 1 status_string)
+if(NOT status_code EQUAL 0)
```
zhouyuan, line 19: we may should skip the download, if local file already exists? 
The other issues on libelf seems due to downloading some empty config.guess/config.sub

kevinw66: > we may should skip the download, if local file already exists?

+1

kevinw66:
> @kevinw66, as this change may be related to arm build, do you have any comment?

thanks @PHILO-HE , I'm ok with it ^^

dev/vcpkg/ports/libelf/portfile.cmake
zml1206: If the download fails, an empty file will be generated.

zhouyuan (approved):
Thanks!

Issue #9691, opened 2025-05-19
[VL]Coredump happend when offHeapBroadcastBuildRelation enabled and BroadcastRelation is large


zjuwangg:
### Backend

VL (Velox)

### Bug description

Gluten Config

```
spark.gluten.velox.offHeapBroadcastBuildRelation.enabled true
```

and coredump happens when broadcast side relation is huge

Coredump Message

Current thread (0x00007effc69f0c00):  JavaThread "Executor task launch worker for task 88.0 in stage 9.0 (TID 3484)" daemon [_thread_in_Java, id=77, stack(0x00007effc2fff000,0x00007effc3400000)]

Stack: [0x00007effc2fff000,0x00007effc3400000],  sp=0x00007effc33fc1b0,  free space=4084k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
v  ~StubRoutines::jbyte_disjoint_arraycopy

siginfo: si_signo: 11 (SIGSEGV), si_code: 1 (SEGV_MAPERR), si_addr: 0x00007ef934ec5027

### Gluten version

Gluten-1.4

### Spark version

None

### System information

NA

### Relevant logs

```bash
NA
```

PR #9692, opened 2025-05-19
[GLUTEN-9691][VL]Fix driver core dump when broadcast too large when UnsafeColumnarBuildSideRelation enabled

zjuwangg:

(Fixes: #9691)

The type of `bytesBufferOffset`  in UnsafeBytesBufferArray.scala is Array[Int], while when `bytesBufferLengths ` are all large number, will cause the element in `bytesBufferOffset`  overflowed the Int.max_value, and finally lead to jvm coredump in other operation like 
``` scala 
Platform.copyMemory(
      bytesBuffer,
      Platform.BYTE_ARRAY_OFFSET,
      longArray.getBaseObject,
      longArray.getBaseOffset + bytesBufferOffset(index),
      bytesBufferLengths(index))
```

## How was this patch tested?

Manually test in our production environment


zjuwangg:
@zhztheplayer Please have a look when you have free time 🙏

zhztheplayer (approved):

PR #9708, opened 2025-05-20
[GLUTEN-9666][VL] Add BatchCarrierRow in preparation for replacement of FakeRow

zhztheplayer:
Difference:

`BatchCarrierRow` is similar to `FakeRow`, but has two implementations:

- `PlaceholderRow`
- `TerminalRow`

Comparing to `FakeRow`, we emit `PlaceholderRow`s along with `TerminalRow` that wraps the actual columnar batch, to make the overall faked rows match the total row numbers of the original columnar batch. This way we don't need to hack into Spark's code to make correction on the row number statistics caused by the wrong numbers of fake rows that were emitted.

`FakeRow` is marked deprecated in the patch. It's to be removed in the next iteration.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/batchcarrier/BatchCarrierRow.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/batchcarrier/BatchCarrierRow.scala
@@ -77,7 +77,8 @@ sealed abstract class BatchCarrierRow extends InternalRow {
   override def get(ordinal: Int, dataType: DataType): AnyRef = throw unsupported()
 
   private def unsupported() = {
-    new UnsupportedOperationException("Underlying columnar data is inaccessible from BatchCarrierRow")
+    new UnsupportedOperationException(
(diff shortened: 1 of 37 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/batchcarrier/BatchCarrierRow.scala
```diff
@@ -0,0 +1,94 @@
...
+  override def getStruct(ordinal: Int, numFields: Int): InternalRow = throw unsupported()
+
+  override def getArray(ordinal: Int): ArrayData = throw unsupported()
+
+  override def getMap(ordinal: Int): MapData = throw unsupported()
+
+  override def get(ordinal: Int, dataType: DataType): AnyRef = throw unsupported()
+
+  private def unsupported() = {
+    new UnsupportedOperationException("Underlying columnar data is inaccessible from CarrierRow")
```
jinchengchenghh, line 80: CarrierRow -> BatchCarrierRow

PR #9710, opened 2025-05-20
Fix the TPCH/DS script cannot throw exception

jinchengchenghh:
Remove try block, and df.show() may be the operator TOPN, not orderby, so changed to collectAsList()

### DIFF
--- a/tools/workload/tpcds/run_tpcds/tpcds_parquet.scala
+++ b/tools/workload/tpcds/run_tpcds/tpcds_parquet.scala
@@ -117,11 +117,7 @@ for (t <- sorted) {
   println(t)
   val fileContents = Source.fromFile(t).getLines.filter(!_.startsWith("--")).mkString(" ")
   println(fileContents)
-  try {
-    time{spark.sql(fileContents).show}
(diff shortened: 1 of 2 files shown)


tools/workload/tpcds/run_tpcds/tpcds_parquet.scala
```diff
@@ -117,11 +117,7 @@ for (t <- sorted) {
...
   val fileContents = Source.fromFile(t).getLines.filter(!_.startsWith("--")).mkString(" ")
   println(fileContents)
-  try {
-    time{spark.sql(fileContents).show}
-    // spark.sql(fileContents).explain
-    Thread.sleep(2000)
-  } catch {
-    case e: Exception => None
-  }
+  time{spark.sql(fileContents)..collectAsList()}
```
marin-ma, line 120: typo `..collectAsList()`

jinchengchenghh: Thanks for your catching

marin-ma (approved):

PR #9720, opened 2025-05-21
[GLUTEN-9719][VL] Unify output definitions for Broadcast/Hash/Columnar shuffle joins

ArnavBalyan:
 - Currently broadcast nested loop join, hash join and columnar shuffle join implement their own output definitions which are all duplicated. Refactor and simplify this code to rely on the function provided through utils.
 - We already have the same definition used in joinutils. Refactored the code to rely on the utils.

## How was this patch tested?
 - Existing unit tests


zhztheplayer (approved):
👍

PR #9721, opened 2025-05-21
[VL] Manually register velox conf on executor start

Zouxxyy:

At present, velox conf has never been used on executor start, resulting in the conf inside `VeloxConfig` not being registered.

For example, if we set `spark.gluten.sql.columnar.backend.velox.filePreloadThreshold=50MB`

Then on the executor side, the following error will be reported

```
Caused by: org.apache.gluten.exception.GlutenException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Operator::getOutput failed for [operator: TableScan, plan node ID: 0]: Non-whitespace character found after end of conversion: "MB"
Retriable: False
Function: operator()
File: /velox/velox/exec/Driver.cpp
Line: 616
Stack trace:
# 0  facebook::velox::VeloxException::VeloxException(char const*, unsigned long, char const*, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >, bool, facebook::velox::VeloxException::Type, std::basic_string_view<char, std::char_traits<char> >)
```

## How was this patch tested?

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
@@ -20,6 +20,7 @@ import org.apache.gluten.backendsapi.ListenerApi
 import org.apache.gluten.columnarbatch.ArrowBatches.{ArrowJavaBatch, ArrowNativeBatch}
 import org.apache.gluten.columnarbatch.VeloxBatch
 import org.apache.gluten.config.GlutenConfig
+import org.apache.gluten.config.VeloxConfig
 import org.apache.gluten.config.VeloxConfig._
 import org.apache.gluten.execution.datasource.GlutenFormatFactory
 import org.apache.gluten.expression.UDFMappings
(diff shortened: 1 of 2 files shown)


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -55,6 +56,8 @@ class VeloxListenerApi extends ListenerApi with Logging {
 
   override def onDriverStart(sc: SparkContext, pc: PluginContext): Unit = {
     val conf = pc.conf()
+    // Register velox conf
+    VeloxConfig.get
```
Yohahaha, line 60: move this logic to `initialize` method.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxListenerApi.scala
```diff
@@ -127,6 +130,8 @@ class VeloxListenerApi extends ListenerApi with Logging {
 
   override def onExecutorStart(pc: PluginContext): Unit = {
     val conf = pc.conf()
+    // Register velox conf
+    VeloxConfig.get
```
Yohahaha, line 134: ditto.

gluten-arrow/src/main/java/org/apache/gluten/init/NativeBackendInitializer.java
```diff
@@ -68,13 +69,18 @@ public void initialize(ReservationListener rl, scala.collection.Map<String, Stri
...
       Map<String, String> nativeConfMap = GlutenConfig.getNativeBackendConf(backendName, conf);
+      this.nativeConfMap = nativeConfMap;
       initialize(rl, ConfigUtil.serialize(nativeConfMap));
     } catch (Exception e) {
       LOG.error("Failed to call native backend's initialize method", e);
       throw e;
     }
   }
 
+  public Map<String, String> nativeConfMap() {
```
Yohahaha, line 80: seems this new method only visible in test? could we try another way to verify the config has parsed? line 40 will increase memory footprint.

Yohahaha (commented):
leave some comments, thank you!

Yohahaha (approved):
👍

Zouxxyy:
@zhztheplayer @yikf Can you help review it, thanks

zhztheplayer (approved):

PR #9278, opened 2025-04-09
[GLUTEN-9163][VL] Use stream de/compressor in sort-based shuffle

marin-ma:
For shuffle writer tasks with heavily spill, streaming compression should provide better compression ratio. Currently, spilled data is compressed before writing to the output stream. This patch modifies the compression logic for sort-based shuffle, removes that compression and uses the compressed output stream to do the work.

PR #9766, opened 2025-05-27
[GLUTEN-9163][VL][FOLLOWUP] Fix segfault triggered by fixed-width inputs

marin-ma:
Fixed the issue reported [here](https://github.com/apache/incubator-gluten/pull/9278#issuecomment-2901970336)


marin-ma:
@wangyum Could you try this patch? Thanks!

zhouyuan (approved):
Thanks!

Issue #9790, opened 2025-05-29
[Flink] Refine: Add a function resolver for complex logic support


lgbo-ustc:
### Description

There are cases where different computation logics correspond to the same function name, but their parameter types differ. Relying solely on the function name cannot meet this requirement.

### Gluten version

None

PR #9792, opened 2025-05-29
[GLUTEN-9790][Flink] Support resolving different converters for Rex calls that share the same operator name

lgbo-ustc:

Fixes: #9790 

## How was this patch tested?

unit tests


lgbo-ustc:
For example, `timestamp +/- interval n second` is also represented by a RexCallNode with the operator `+/-`. We cannot treat it as a simple arithmetic expression."

exmy (approved):

PR #9797, opened 2025-05-29
[FLINK] Fix flink CI

philo-he:
Fix issue:
Invalid workflow file: .github/workflows/flink.yml#L1`test` is not a valid event name.


zhouyuan (approved):
👍

PR #9802, opened 2025-05-29
[CORE] Test trait `WithQueryPlanListener` for ensuring on fallback query plan nodes conveniently

zhztheplayer:
When overwriting a Spark test class in Gluten, add the trait `WithQueryPlanListener` to the class inheritance tree, then call APIs on `planListeners` in the class body to do assertions on all executed query plans during the test execution. E.g.:

```scala
class Gluten...Suite extends ...Suite with ... with WithQueryPlanListener {
  planListeners.assertNoInstanceOf[SortExec]
}
```

In the above code, all the test cases ran from the suite will be ensured not having the fallen back `SortExec` nodes in the query plan. Exceptions will be thrown otherwise.

### DIFF
--- a/gluten-substrait/src/test/scala/org/apache/gluten/test/FallbackUtil.scala
+++ b/gluten-substrait/src/test/scala/org/apache/gluten/test/FallbackUtil.scala
@@ -67,9 +67,13 @@ object FallbackUtil extends Logging with AdaptiveSparkPlanHelper {
   }
 
   def hasFallback(plan: SparkPlan): Boolean = {
-    val fallbackOperator = collectWithSubqueries(plan) { case plan => plan }.filterNot(
-      plan => plan.isInstanceOf[GlutenPlan] || skip(plan))
+    val fallbackOperator =
+      collectWithSubqueries(plan) { case plan => plan }.filter(plan => nodeHasFallback(plan))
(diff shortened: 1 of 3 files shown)


gluten-substrait/src/test/scala/org/apache/gluten/test/FallbackUtil.scala
```diff
@@ -67,9 +67,13 @@ object FallbackUtil extends Logging with AdaptiveSparkPlanHelper {
...
-      plan => plan.isInstanceOf[GlutenPlan] || skip(plan))
+    val fallbackOperator =
+      collectWithSubqueries(plan) { case plan => plan }.filter(plan => nodeHasFallback(plan))
     fallbackOperator.foreach(operator => log.info(s"gluten fallback operator:{$operator}"))
     fallbackOperator.nonEmpty
   }
+
+  def nodeHasFallback(plan: SparkPlan): Boolean = {
+    !plan.isInstanceOf[GlutenPlan] && !skip(plan)
+  }
```
zhztheplayer, line 78: Just to prepare a API for the following potential use in the future:

```scala
planListeners.assertNoNodeExists(nodeHasFallback)
```

marin-ma (approved):
LGTM. Thanks!

Issue #9801, opened 2025-05-29
[VL] Native parquet writer writing a corrupt / invalid parquet file on spark task failure


RushabhK:
### Backend

VL (Velox)

### Bug description

Hello team, I am using Gluten 1.3.0 compiled with JDK 17 on Ubuntu 20.04.
I ingested parquet files with a spark job with gluten which underwent 1 task failure and a successful retry, with the job succeeding.
Although, while reading the parquet folder, I can find one file which says the parquet file is corrupt with the following exception:
`OSError: Could not open parquet input source 'gluten-part-239f94b0-2c93-4b5d-a8a7-3be9a9e05dac.zstd.parquet': Invalid: Parquet magic bytes not found in footer. Either the file is corrupted or this is not a parquet file.`
This is a 300 MB file and I verified from the parquet-tools that it is not a valid parquet file.

Although, reading the path with spark.read.option("ignoreCorruptFiles", "true").parquet(path), I get exactly the data that is expected (I did the checksum validation the dataframe with the expectation, and it passed).

I am suspecting the failed task should not have copied the partial parquet file which it had written to the temp location to the final location as a part of the commit.

I am using the Manifest committer with version 2, please find the following configs:
`--conf spark.hadoop.mapreduce.fileoutputcommitter.algorithm.version=2 \
--conf spark.hadoop.mapreduce.fileoutputcommitter.cleanup.skipped=true \
--conf spark.hadoop.mapreduce.fileoutputcommitter.marksuccessfuljobs=false \
--conf spark.hadoop.mapreduce.outputcommitter.factory.scheme.gs=org.apache.hadoop.mapreduce.lib.output.committer.manifest.ManifestCommitterFactory \
--conf spark.sql.parquet.output.committer.class=org.apache.spark.internal.io.cloud.BindingParquetOutputCommitter`

Can anyone point out if I am missing some gluten configs while ingesting parquet which is not preventing this copy? Also, does Gluten honor the committer specified in the spark conf?
I am not facing any such issues with vanilla spark for the same runs.

### Gluten version

Gluten-1.3

### Spark version

Spark-3.5.x

### Spark configurations

--conf spark.hadoop.mapreduce.outputcommitter.factory.scheme.gs=org.apache.hadoop.mapreduce.lib.output.committer.manifest.ManifestCommitterFactory \
--conf spark.sql.parquet.output.committer.class=org.apache.spark.internal.io.cloud.BindingParquetOutputCommitter \
[...]

PR #9808, opened 2025-05-30
[GLUTEN-9801]Delete the written file if the task failed

JkSelf:

Fix https://github.com/apache/incubator-gluten/issues/9801


JkSelf:
@ulysses-you Can you help to review this PR? Thanks.

backends-velox/src/main/scala/org/apache/spark/sql/execution/SparkWriteFilesCommitProtocol.scala
```diff
@@ -97,8 +97,11 @@ class SparkWriteFilesCommitProtocol(
     }
   }
 
-  def abortTask(): Unit = {
+  def abortTask(writePath: String): Unit = {
     committer.abortTask(taskAttemptContext)
+
+    val tmpPath = new Path(writePath)
+    tmpPath.getFileSystem(taskAttemptContext.getConfiguration).delete(tmpPath, false)
```
ulysses-you, line 104: `delete(tmpPath, true)` ? the boolean param means if delete files recursive when the path is a directory.  IIUC, `tmpPath` is a directory.

JkSelf: @ulysses-you Good catch. Updated.

zhouyuan (approved):

backends-velox/src/main/scala/org/apache/spark/sql/execution/SparkWriteFilesCommitProtocol.scala
```diff
@@ -97,8 +97,11 @@ class SparkWriteFilesCommitProtocol(
     }
   }
 
-  def abortTask(): Unit = {
+  def abortTask(writePath: String): Unit = {
     committer.abortTask(taskAttemptContext)
+
+    val tmpPath = new Path(writePath)
```
RushabhK, line 103: @JkSelf This would delete all the files in the temporary staging directory. This leads to FileNotFoundException, I have seen this in my setup: https://github.com/apache/incubator-gluten/issues/9801#issuecomment-2929038203
We need to selectively delete the files like spark's HadoopMapReduceCommitProtocol selectively deletes the absolute paths that were added: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/internal/io/HadoopMapReduceCommitProtocol.scala#L302-L305

Yohahaha: +1, spark only delete committed `absPath` which means delete customer partition location path.

Spark and Hadoop never delete task committed path.

Issue #9775, opened 2025-05-28
[VL] Add CI for static/dynamic build on openEuler24


kevinw66:
### Description

Since we are now support openEuler24, I'd like to add CI jobs to check the stability(build only), including docker images for x86/arm, like vcpkg-centos-8 and centos-8-jdk17 in [this file](https://github.com/apache/incubator-gluten/blob/main/.github/workflows/docker_image.yml), and nightly check for native-lib-build like [here](https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_nightly.yml#L34)

### Gluten version

main branch

kevinw66:
@PHILO-HE @FelixYBW Do you have any concerns?

zhouyuan:
+1  
I have seen more Gluten issues related with ARM recently. It's better to add CI to check and guard. The docker image for ARM should be added already in the weekly job.

kevinw66:
> +1 I have seen more Gluten issues related with ARM recently. It's better to add CI to check and guard. The docker image for ARM should be added already in the weekly job.

@zhouyuan The ARM-related CI will runs on CentOS 8. Currently, there are bugs need to be fixed first. Perhaps we can continue the discussion in #8181  (sorry I forgot to response in the thread), this issue is about openEuler compatibility validation.

PR #9816, opened 2025-05-30
[GLUTEN-9775][VL] Add CI to build openEuler24 static/dynamic docker images

kevinw66:

Fixes: #9775

### DIFF
--- a/.github/workflows/docker_image_openeuler.yml
+++ b/.github/workflows/docker_image_openeuler.yml
@@ -18,7 +18,7 @@ name: Build and Push Docker Image (openEuler)
 on:
   push:
     branches:
-      - openeuler-docker
+      - main
     paths:
       - '.github/workflows/docker_image_openeuler.yml'
       - '.github/workflows/util/install_spark_resources.sh'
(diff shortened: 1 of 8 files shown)


PHILO-HE (commented):
Hi @kevinw66, thanks for your pr! Currently in docker hub apache/gluten, we only publish popular docker images, most of which are for CI jobs to use. I think we can remove the proposed docker build job for now.
For build verification on openeuler, you can add a weekly job in the file below. https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_weekly.yml

kevinw66:
> Hi @kevinw66, thanks for your pr! Currently in docker hub apache/gluten, we only publish popular docker images, most of which are for CI jobs to use. I think we can remove the proposed docker build job for now. For build verification on openeuler, you can add a weekly job in the file below. https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_weekly.yml

@PHILO-HE Not only for build purpose, I believe that in addition to verifying our primarily supported os, we can also validate secondary supported ones. By using nightly/weekly CI to check the stability of build/ut, we can show gluten's support for a wide range of os while ensuring that nightly/weekly CI doesn't consume too much CI resources. Of course, I will separate these workflows from the primary ones to prevent interference, so the image here is prepared for subsequent ut validation.

Alternatively, do you think it's sufficient to only validate build rather than ut for different os?

philo-he:
@kevinw66, I think weekly build verification from clean OS environment can uncover most issues. If necessary, we can also add a TPC-H or DS test in weekly jobs after the build verification to mainly ensure native libs working well at runtime. As discussed before, we should avoid too much test redundancy in CI. You can see even though Gluten officially supports centos-7/8, ubuntu-20.04/22.04, Spark unit tests are only verified on cento-8 in CI. The combination of UT verification on cento-8 (x86 plus future ARM) and weekly basic verification (build + possible TPC-H/DS tests) on other OSes may be sufficient.

kevinw66:
Close as discussed

PR #9821, opened 2025-05-31
[MINOR] Rename ConfigProvider to GlutenConfigProvider

wangyum:

To avoid class conflict issue if add gluten jar as a dependency of the Spark project. For example:
```xml
    <dependency>
      <groupId>org.apache.gluten</groupId>
      <artifactId>gluten-velox-bundle</artifactId>
      <version>1.5.0-SNAPSHOT</version>
      <scope>system</scope>
      <systemPath>/Users/yumwang/Downloads/gluten-velox-bundle-spark3.5_2.12-linux_amd64-1.5.0-SNAPSHOT.jar</systemPath>
    </dependency>
```
Error message:
```
/Users/yumwang/spark/sql/core/src/main/scala/org/apache/spark/sql/internal/VariableSubstitution.scala:38:24
overloaded method constructor ConfigReader with alternatives:
  (conf: java.util.Map[String,String])org.apache.spark.internal.config.ConfigReader <and>
  (conf: org.apache.spark.internal.config.ConfigProvider)org.apache.spark.internal.config.ConfigReader
 cannot be applied to (org.apache.spark.sql.internal.ConfigProvider)
  private val reader = new ConfigReader(provider)
```

## How was this patch tested?

Manual tests.

### DIFF
--- a/shims/common/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
@@ -18,7 +18,7 @@ package org.apache.gluten.config
 
 import org.apache.gluten.config.BackendType.BackendType
 
-import org.apache.spark.sql.internal.ConfigProvider
+import org.apache.spark.sql.internal.GlutenConfigProvider
 
 /**
  * An entry contains all meta information for a configuration.
@@ -61,8 +61,8 @@ trait ConfigEntry[T] {
   /** How to convert a value to a string that the user can use it as a valid string value. */
   def stringConverter: T => String
 
-  /** Read the configuration from the given ConfigProvider. */
-  def readFrom(conf: ConfigProvider): T
+  /** Read the configuration from the given GlutenConfigProvider. */
+  def readFrom(conf: GlutenConfigProvider): T
 
   /** The default value of the configuration. */
   def defaultValue: Option[T]
@@ -70,7 +70,7 @@ trait ConfigEntry[T] {
   /** The string representation of the default value. */
   def defaultValueString: String
 
-  final protected def readString(provider: ConfigProvider): Option[String] = {
+  final protected def readString(provider: GlutenConfigProvider): Option[String] = {
     alternatives.foldLeft(provider.get(key))((res, nextKey) => res.orElse(provider.get(nextKey)))
   }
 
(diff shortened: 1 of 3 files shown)


FelixYBW:
Do you mean your spark has dependency on Gluten? or a third project depends on spark and gluten at the same time?

wangyum:
> Do you mean your spark has dependency on Gluten? or a third project depends on spark and gluten at the same time?

Yes.

FelixYBW:
> > Do you mean your spark has dependency on Gluten? or a third project depends on spark and gluten at the same time?
> 
> Yes.

It's not a good thing to have Spark dependency on Gluten. In theory we only need Gluten depends on Spark

wangyum:
We added lots of third-party dependencies to `${SPARK_HOME}/jars` and released them together, which resulted we can not run integration tests because of class conflict.

Yohahaha (approved):
👍

shims/common/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
```diff
@@ -18,7 +18,7 @@ package org.apache.gluten.config
 
 import org.apache.gluten.config.BackendType.BackendType
 
-import org.apache.spark.sql.internal.ConfigProvider
+import org.apache.spark.sql.internal.GlutenConfigProvider
```
jackylee-ch, line 21: nit: We can rename it to `org.apache.gluten.config.GlutenConfigProvider`

philo-he:
> We added lots of third-party dependencies to `${SPARK_HOME}/jars` and released them together, which resulted we can not run integration tests because of class conflict.

@wangyum, I am curious why you are introducing Gluten as a dependency in your Spark project. The recommended way is putting gluten jar into Spark class path and correctly setting Gluten as Spark plugin. I'm afraid there are some other conflict classes.

FelixYBW:
> @wangyum, I am curious why you are introducing Gluten as a dependency in your Spark project. The recommended way is putting gluten jar into Spark class path and correctly Gluten as Spark plugin. I'm afraid there are some other conflict classes.

I remember we did meet issues if we put Gluten jar in spark/jars. It's why we recommend to set classpath. You may try to put your spark code which depends on Gluten into Gluten's shimlayer, instead of add Gluten as Spark's dependency.

wForget:
We also put gluten(1.2.0) jar into spark_home/jars and have not encountered class conflict issues yet, but this change makes sense to me.

wangyum:
We add gluten jar though `pom.xml` to run integration tests.
We also won't encounter class conflicts if we put the gluten jar into `${SPARK_HOME}/jars`.

FelixYBW (approved):

PR #9828, opened 2025-06-02
[BUILD][VL] Fix velox test build breaking due to GMock missing in System library

surnaik:
Currently, when we build Velox with tests on

./dev/buildbundle-veloxbe.sh --build_type=Debug --build_tests=ON --enable_gcs=ON --enable_hdfs=ON --build_velox_tests=ON --spark_version=3.5

The tests fail to build with the following error

CMake Error at velox/buffer/tests/CMakeLists.txt:18 (target_link_libraries):
  Target "velox_buffer_test" links to:

    GTest::gmock

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

I reinstalled googletest as well, but it still couldn't resolve gmock. I was able to get it working after using the BUNDLED GTest.

Set GTest_SOURCE=BUNDLED

## How was this patch tested?

Existing UTs and Build pipelines

### DIFF
--- a/.github/workflows/flink.yml
+++ b/.github/workflows/flink.yml
@@ -1,0 +1,41 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Gluten Flink Github Runner
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/flink.yml'
+      - 'gluten-flink/**'
+
+jobs:
+  flink-test:
+    runs-on: ubuntu-22.04
+    container: apache/gluten:centos-8-jdk17
+    steps:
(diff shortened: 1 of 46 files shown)


surnaik:
@zhouyuan Can you please take a look? Thanks!

surnaik:
@PHILO-HE could you please take a look, thanks!

philo-he:
@surnaik, this seems not a fix for the root cause. It would be better to also make pre-installed gtest work. Could you tell me which OS you are using?

surnaik:
I’m using Ubuntu 22.04

FelixYBW:
I just tried using apache/gluten:vcpkg-centos-8 to build tests. Static version works well.

surnaik:
> I just tried using apache/gluten:vcpkg-centos-8 to build tests. Static version works well.

Thanks, let me give a try. I currently only use Ubuntu or a Debian based custom OS, could be a system issue on my side as well

surnaik:
@FelixYBW That worked for me as well, but on Ubuntu 22.04, I rebuilt the googletest library as well, but somehow it's not able to still resolve GMock, I'll give this a try to find the reason later as I'm able to workaround using BUNDLED

philo-he:
@surnaik, I note you are not enabling vcpkg. Then, the dependency libs can be pre-installed by setup scripts. According to the script below, you can try to install libgtest-dev & libgmock-dev to see if it works. Please remove manually installed gtest and gmock to avoid conflict.

https://github.com/facebookincubator/velox/blob/main/scripts/setup-ubuntu.sh#L121-L122

FelixYBW:
@surnaik for dynamic link, you need to run `--run_setup_script=ON` at least once on a fresh system, which install all depencency libraries into your system. In following build you can set `--run_setup_script=OFF` 

https://github.com/apache/incubator-gluten/blob/33ce2edfaba5f562117c9fbec5d94405f6cea334/dev/docker/Dockerfile.centos8-dynamic-build#L55C4-L55C66

PR #9829, opened 2025-06-02
[VL] Fix DPP not working for Arrow Scan

surnaik:

Scan that gets pushed to Spark that return Arrow compatible vectors fails when DPP is present since the filteredPartitions subquery is never scheduled by Spark.

BatchScanExec does support DPP but when we use the ArrowScanReplaceRule, the ArrowBatchScan doesn’t support it since it doesn’t implement the filteredPartitions.

This change makes ArrowBatchScan extend BatchScanExecShim which supports DPP

## How was this patch tested?
Existing UTs
Manual Test with a Datasource that return Arrow Vector


surnaik:
@FelixYBW @PHILO-HE please review. Thanks!

jackylee-ch:
For V2 Source Scan, you can try [this PR](https://github.com/apache/spark/pull/37514) to support DPP.

surnaik:
> For V2 Source Scan, you can try [this PR](https://github.com/apache/spark/pull/37514) to support DPP.

Thanks, currently, BatchScanExec does support DPP but when we use the ArrowScanReplaceRule, the ArrowBatchScan doesn’t support it since it doesn’t implement the filteredPartitions

surnaik:
@jinchengchenghh Could you please take a look!

shims/spark35/src/main/scala/org/apache/spark/sql/execution/datasources/v2/BatchScanExecShim.scala
```diff
@@ -146,10 +146,35 @@ abstract class BatchScanExecShim(
...
+  }
+
+  override def doExecuteColumnar(): RDD[ColumnarBatch] = {
+    val numOutputRows = longMetric("numOutputRows")
+    inputRDD.asInstanceOf[RDD[ColumnarBatch]].map {
+      b =>
+        numOutputRows += b.numRows()
+        b
+    }
+  }
```
zhztheplayer: Any difference of this code among the Spark versions? Or else we can move it to `ArrowBatchScanExec.scala`?

surnaik: This is same, I’ll move it to a new file. Thanks!

surnaik:
@zhztheplayer could you please take a look, test failure seems unrelated, passing locally

jinchengchenghh (approved):
Please also update the PR description including https://github.com/apache/incubator-gluten/pull/9829#issuecomment-2933287774

jinchengchenghh:
Thanks for your fix.

surnaik:
> Please also update the PR description including [#9829 (comment)](https://github.com/apache/incubator-gluten/pull/9829#issuecomment-2933287774)

Done. Thanks for your review!

PR #9843, opened 2025-06-03
[MINOR][VL] Remove dead code in GlutenImplicits

zhouyuan:

this is a minor change to remove the dead code

## How was this patch tested?

pass GHA

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenImplicits.scala
+++ b/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenImplicits.scala
@@ -218,10 +218,4 @@ object GlutenImplicits {
       totalFallbackNodeToReason.toSeq
     )
   }
-
-  implicit class DatasetTransformer[T](dateset: Dataset[T]) {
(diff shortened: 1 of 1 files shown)


gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenImplicits.scala
```diff
@@ -218,10 +218,4 @@ object GlutenImplicits {
       totalFallbackNodeToReason.toSeq
     )
   }
-
-  implicit class DatasetTransformer[T](dateset: Dataset[T]) {
```
wForget, line 222: This does not seem to be dead code but a class for implicit conversion. The usage is as shown in example:

https://github.com/apache/incubator-gluten/blob/b21ea90efccf0804a92e4675a0ed43c266ec6fdd/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenImplicits.scala#L46-L52

cc @ulysses-you

zhouyuan:
@wForget thanks! I mis-understood some of the code

PR #9782, opened 2025-05-28
[GLUTEN-9779][VL] Remove redundant GetTimestamp

zml1206:
## What changes were proposed in this pull request?

Remove redundant GetTimestamp to support to_date(Timestamp, format) and to_timestamp(Timestamp, format).

(Fixes: \#9779)

## How was this patch tested?

Unit test.

PR #9848, opened 2025-06-03
[VL] Minor fix: simplify a few function tests' verbose names

PHILO-HE:

Follow-up fix for this comment:
https://github.com/apache/incubator-gluten/pull/9782#discussion_r2113247329

Remove "Test" from test name if not very necessary. Keep "function" in test name only if it help avoid ambiguity.


philo-he:
@zml1206, could you take a look?

zml1206:
In addition, is it appropriate to remove all `function`, because these three classes are specifically used to test function, so there is no ambiguity.

philo-he:
> In addition, is it appropriate to remove all `function`, because these three classes are specifically used to test function, so there is no ambiguity.

@zml1206, it makes sense. Just fixed. Thank you!

Issue #9849, opened 2025-06-03
[VL] VeloxBloomFilterMightContain should not be applied to FileSourceScan partition filters.


wForget:
### Backend

VL (Velox)

### Bug description

Some queries apply `BloomFilterMightContain` to the parition filter of FileSourceScan, but gluten converts it to `VeloxBloomFilterMightContain` which cannot be executed in driver and causes the task to fail.

### Gluten version

main branch

### Spark version

Spark-3.5.x

### Relevant logs

```bash

```

wForget:
Duplicate of https://github.com/apache/incubator-gluten/issues/6650

PR #9850, opened 2025-06-03
[GLUTEN-9849][VL] Avoid VeloxBloomFilterMightContain being applied to FileSourceScan partition filters

wForget:

Exclude VeloxBloomFilterMightContain from FileSourceScan parition filters

Fixes: #9849

## How was this patch tested?

added unit test

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/expression/VeloxBloomFilterMightContain.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/expression/VeloxBloomFilterMightContain.scala
@@ -16,6 +16,7 @@
  */
 package org.apache.gluten.expression
 
+import org.apache.gluten.execution.GlutenTaskOnlyExpression
 import org.apache.gluten.sql.shims.SparkShimLoader
 import org.apache.gluten.utils.VeloxBloomFilter
 
@@ -35,7 +36,8 @@ import org.apache.spark.task.TaskResources
 case class VeloxBloomFilterMightContain(
     bloomFilterExpression: Expression,
     valueExpression: Expression)
-  extends BinaryExpression {
+  extends BinaryExpression
+  with GlutenTaskOnlyExpression {
 
   private val delegate =
     SparkShimLoader.getSparkShims.newMightContain(bloomFilterExpression, valueExpression)
@@ -89,6 +91,7 @@ case class VeloxBloomFilterMightContain(
     val valueEval = valueExpression.genCode(ctx)
     val code =
       code"""
+      org.apache.gluten.expression.VeloxBloomFilterMightContain.checkInSparkTask();
       ${valueEval.code}
       boolean ${ev.isNull} = ${valueEval.isNull};
       ${CodeGenerator.javaType(dataType)} ${ev.value} = ${CodeGenerator.defaultValue(dataType)};
@@ -96,5 +99,13 @@ case class VeloxBloomFilterMightContain(
         ${ev.value} = $bf.mightContainLong((Long)${valueEval.value});
       }"""
     ev.copy(code = code)
+  }
+}
+
+object VeloxBloomFilterMightContain {
+  def checkInSparkTask(): Unit = {
+    if (!TaskResources.inSparkTask()) {
+      throw new UnsupportedOperationException("velox_might_contain is not evaluable on Driver")
+    }
   }
 }
--- a/backends-velox/src/main/scala/org/apache/gluten/expression/VeloxBloomFilterMightContain.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/expression/VeloxBloomFilterMightContain.scala
@@ -32,6 +32,9 @@ import org.apache.spark.task.TaskResources
  * Velox's bloom-filter implementation uses different algorithms internally comparing to vanilla
  * Spark so produces different intermediate aggregate data. Thus we use different filter function /
  * agg function types for Velox's version to distinguish from vanilla Spark's implementation.
+ *
+ * FIXME: Remove GlutenTaskOnlyExpression after the VeloxBloomFilter expr is made compatible with
+ * spark. See: https://github.com/apache/incubator-gluten/pull/9850#issuecomment-3007448538
  */
 case class VeloxBloomFilterMightContain(
     bloomFilterExpression: Expression,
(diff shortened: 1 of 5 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/execution/ScanTransformerFactory.scala
```diff
@@ -25,11 +25,17 @@ object ScanTransformerFactory {
...
+    val partitionFilters = scanExec.partitionFilters.filter {
+      case _: GlutenTaskOnlyExpression => false
+      case _ => true
+    }
     FileSourceScanExecTransformer(
       scanExec.relation,
       scanExec.output,
       scanExec.requiredSchema,
-      scanExec.partitionFilters,
+      partitionFilters,
```
rui-mo, line 38: Do you need to consider this issue also for the BatchScanTransformer? Thanks.

wForget: > Do you need to consider this issue also for the BatchScanTransformer? Thanks.

~~I will try to add a unit test to cover this case~~

wForget: > Do you need to consider this issue also for the BatchScanTransformer? Thanks.

There are no partition filters in `BatchScanExecTransformer`, and runtime filters will also be converted to source v2 predicates, so there is no similar issue for ``BatchScanExecTransformer`.

wForget:
Duplicate of #6650 , cc @WangGuangxin @zhztheplayer Could you please take a look?

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/GlutenInjectRuntimeFilterSuite.scala
```diff
@@ -16,6 +16,27 @@
...
+      SQLConf.DYNAMIC_PARTITION_PRUNING_ENABLED.key -> "false",
+      "spark.gluten.sql.native.bloomFilter" -> "false",
+      SQLConf.AUTO_BROADCASTJOIN_THRESHOLD.key -> "2000"
+    ) {
+      val query = "select * from bf5part join bf2 on bf5part.f5 = bf2.c2 where bf2.a2 = 67"
+      val df = sql(query)
+      df.collect()
+      val plan = df.queryExecution.executedPlan.toString()
+      logWarning(s"GlutenInjectRuntimeFilterSuite Plan:")
+      plan.split("\n").foreach(logWarning(_))
```
wForget: This test case fails on the CH backend. It is only used to print the execution plan on CH backend, and I will remove it later. https://opencicd.kyligence.com/job/gluten/job/gluten-ci-arm/293/console

image

zhouyuan:
Cc @WangGuangxin as he made a fix on similar issue before https://github.com/apache/incubator-gluten/pull/6652

zhztheplayer (changes requested):
Give me a couple of minutes to check this. Thanks!

zhztheplayer (approved):
I haven't had a simple solution finalized so feel free to proceed.

@wForget If we exclude the might_contain from partition filters, we may lose the advantage brought by Spark bloom filters. No?

wForget:
> @wForget If we exclude the might_contain from partition filters, we may lose the advantage brought by Spark bloom filters. No?

Yes, this will indeed cause performance regression, but the bloomFilterData that has been constructed by VeloxBloomFilter cannot be used in the driver unless we use jni to call velox in the driver.

jinchengchenghh:
I would suggest to implement spark native bloom filter in velox, then we can remove this rule.

zhztheplayer:
> I would suggest to implement spark native bloom filter in velox, then we can remove this rule.

It's a good idea while the challenge is we need to maintain the consistency among Spark versions and Velox. I.e., once Spark updates its bloom filter algorithm, we have to add a copy of that algorithm to Velox.  Maybe we can first leave an abstraction layer in the C++ implementation for different Spark versions.

backends-velox/src/main/scala/org/apache/gluten/expression/VeloxBloomFilterMightContain.scala
```diff
@@ -31,11 +32,15 @@ import org.apache.spark.task.TaskResources
  * Velox's bloom-filter implementation uses different algorithms internally comparing to vanilla
  * Spark so produces different intermediate aggregate data. Thus we use different filter function /
  * agg function types for Velox's version to distinguish from vanilla Spark's implementation.
+ *
+ * FIXME: Remove GlutenTaskOnlyExpression after the VeloxBloomFilter expr is made compatible with
```
wForget, line 36: @zhztheplayer @jinchengchenghh I added a comment to record this improvement, how about we merge this PR first?

zhztheplayer: No problem from my end. cc @jinchengchenghh 

And can we log an issue highlighting the performance gap because of the unavailability of scan + might_contain?

Also I guess vanilla Spark's scan + Velox might_contain will cause the issue as well?

wForget: > Also I guess vanilla Spark's scan + Velox might_contain will cause the issue as well?

In what scenarios does a vanilla spark scan with native expressions occur? Do we have an existing test case?

zhztheplayer: I meant when velox_might_contain is included in a vanilla Spark scan node, which for whatever reason was fallen back by Gluten. I was just guessing whether the same issue will happen in the case.

wForget: > I meant when velox_might_contain is included in a vanilla Spark scan node, which for whatever reason was fallen back by Gluten. I was just guessing whether the same issue will happen in the case.

Oh, I see.

wForget:
By the way, the current scenario may be uncommon. `PartitionPruning` rule is applied before `InjectRuntimeFilter`, and `InjectRuntimeFilter` will check whether a DPP filter exists.

backends-velox/src/main/scala/org/apache/gluten/expression/VeloxBloomFilterMightContain.scala
wForget: > And can we log an issue highlighting the performance gap because of the unavailability of scan + might_contain?

Filed an issue #10071

wForget:
@zhztheplayer @jinchengchenghh @WangGuangxin @zhouyuan Thank you for your review, I will merge this pr later if there is no further discussion.

wForget:
Thanks, merged to main

zhztheplayer:
> I would suggest to implement spark native bloom filter in velox, then we can remove this rule.

@jinchengchenghh The Spark community just accepted a V2 bloom filter implementation https://github.com/apache/spark/pull/50933.

PR #9851, opened 2025-06-03
[VL] Fix link issues found in release process

philo-he:

* Fix outdated links.
* Remove symbolic link which is not valid in codebase, but only valid after build. Just create it in script.

### DIFF
--- a/tools/gluten-it/spark-home/jars
+++ b/tools/gluten-it/spark-home/jars
@@ -1,1 +1,0 @@
-../package/target/lib
--- a/tools/workload/tpch/README.md
+++ b/tools/workload/tpch/README.md
@@ -1,7 +1,7 @@
 # Test on Velox backend with TPC-H workload
 
 ## Test datasets
-Parquet and DWRF(a fork of the ORC file format) format files are both supported. Here are the steps to generate the testing datasets:
+Parquet and DWRF (a fork of the ORC file format) format files are both supported. Here are the steps to generate the testing datasets:
 
 ### Generate the Parquet dataset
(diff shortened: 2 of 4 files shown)


tools/gluten-it/spark-home/jars
philo-he, line 1: This is for fixing: "gluten-it contains an invalid symbolic link (for jars)", which is reported by incubator committee. It is replaced by using a create command in gluten-it.sh. @zhztheplayer, could you take a look?

tools/workload/tpch/README.md
```diff
@@ -26,27 +26,6 @@ val rootDir = "/PATH/TO/TPCH_PARQUET_PATH" // root directory of location to crea
 val dbgenDir = "/PATH/TO/TPCH_DBGEN" // location of dbgen
 ```
 
-Currently, Gluten with Velox can support both Parquet and DWRF file format and three compression codec including snappy, gzip, zstd.
-Below step, to convert Parquet to DWRF, is optional if you are using Parquet format to run the testing.
-
-### Convert the Parquet dataset to DWRF dataset(OPTIONAL)
-And then please refer to the scripts in [dwrf_dataset](./gen_data/dwrf_dataset/) directory to convert the Parquet dataset to DWRF dataset.
```
philo-he, line 33: The dwrf dataset doesn't exist now.

FelixYBW (approved):

tools/gluten-it/spark-home/jars
zhztheplayer: Sorry for the delayed response; The issue does seem to be broken, I am fixing it within https://github.com/apache/incubator-gluten/pull/10232

PR #9863, opened 2025-06-04
[VL] add rhel setup script support

FelixYBW:
Add setup-rhel.sh. Only verified on rhel9.6 now.

@zhouyuan to enable FIPS support

### DIFF
--- a/ep/build-velox/src/setup-rhel.sh
+++ b/ep/build-velox/src/setup-rhel.sh
@@ -1,0 +1,393 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


ep/build-velox/src/setup-rhel.sh
```diff
@@ -0,0 +1,393 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
```
yaooqinn:
```suggestion
```

Is this copied from any Facebook repos? Add ep/build-velox/src/setup-rhel.sh to the NOTICE file to declare the authority, otherwise remove it

zhouyuan (approved):
Thanks!

yaooqinn (approved):

vjanelle:
@zhouyuan was there a reason for needing to download bison (just as an example)?

FelixYBW:
> @zhouyuan was there a reason for needing to download bison (just as an example)?

Do you mean bison isn't used or we should install instead of download?

vjanelle:
> > @zhouyuan was there a reason for needing to download bison (just as an example)?
> 
> 
> 
> Do you mean bison isn't used or we should install instead of download?

Not in front of my computer, but IIRC-

Poking around, the centos9 build process just uses the distribution bison 3.7, and AFAICT(although this is not my forte) the features needed were in 3.3. Only the rhel ones download it.

There was a few other things that were different between the centos9 and setup-rhel scripts that seemed odd.

FelixYBW:
We didn't refine the script. We just use it and pass the dynamic gluten/velox build. Feel free to refine it if you see issues.

PR #9869, opened 2025-06-04
[GLUTEN-9836][VL] Remove ShuffleMemoryPool and support creating arrow pool instances

marin-ma:


cpp/velox/memory/VeloxMemoryManager.h
```diff
@@ -100,20 +103,24 @@ class VeloxMemoryManager final : public MemoryManager {
...
 #endif
 
-  // This is a listenable allocator used for arrow.
-  std::unique_ptr<MemoryAllocator> listenableAlloc_;
   std::unique_ptr<AllocationListener> listener_;
   std::unique_ptr<AllocationListener> blockListener_;
-  std::unique_ptr<arrow::MemoryPool> arrowPool_;
+
+  std::shared_ptr<ArrowMemoryPool> defaultArrowPool_;
+  std::unordered_map<std::string, std::weak_ptr<ArrowMemoryPool>> arrowPools_;
```
zhztheplayer, line 117: It's understandable to use weak pointers, but I am curious how many instance will be added to this map in the worst case?

marin-ma: Not sure. But for long running spark process, a new pool will be created for every shuffle writer, which might be a large number. If we don't use weak pointer here, we still need to delete the pool if it won't be used.

zhztheplayer (approved):

cpp/core/jni/JniWrapper.cc
```diff
@@ -144,10 +144,14 @@ class InternalMemoryManager : public MemoryManager {
  public:
   InternalMemoryManager(const std::string& kind) : MemoryManager(kind) {}
 
-  arrow::MemoryPool* getArrowMemoryPool() override {
+  arrow::MemoryPool* defaultArrowMemoryPool() override {
     throw GlutenException("Not implemented");
   }
 
+  std::shared_ptr<arrow::MemoryPool> addArrowMemoryPool(const std::string& name) override {
```
FelixYBW: rename it as createArrowMemoryPool since it's make_shared in the function?

cpp/velox/shuffle/VeloxShuffleWriter.h
```diff
@@ -136,7 +134,7 @@ class VeloxShuffleWriter : public ShuffleWriter {
 
   // Memory Pool used to track memory usage of partition buffers.
   // The actual allocation is delegated to options_.memoryPool.
-  std::unique_ptr<ShuffleMemoryPool> partitionBufferPool_;
+  std::shared_ptr<arrow::MemoryPool> partitionBufferPool_;
```
FelixYBW, line 137: Why do we change to shared_ptr here? Is the pool shared?

marin-ma: Because `VeloxMemoryManager` also owns a weak pointer to the pool. It can't be unique_ptr here.

FelixYBW: I see. it's used by ptr.lock(). Add a comment here that
VeloxShuffleWritter owns the ownership actually.

FelixYBW (approved):
Good to me

FelixYBW:
@JkSelf You may use this way to pass a arrowmemorypool to parquet writter

FelixYBW:
after the change:
image

cpp/velox/shuffle/VeloxShuffleWriter.h
marin-ma: Added comment for the API
https://github.com/apache/incubator-gluten/pull/9869/files#diff-af0d96b5a965e6d46b56d039f6da80ea25ce3ac90be2be922a92087513b97b4aR42-R48

PR #9876, opened 2025-06-05
Add RichSparkConf to simplify the interoperations with gluten config entries

yaooqinn:

Add RichSparkConf to simplify the interoperations with the gluten config entries, mostly reduce the redefination, reclaiming of default values

## How was this patch tested?


yaooqinn:
> I feel it might not harm if we directly move GlutenConfig to package `org.apache.spark.sql.internal` so get all the access to the internal APIs, though not quite sure.

This only solves half of the issue

Issue #9881, opened 2025-06-05
[CORE] Minimize module dependency set of module `gluten-core`


zhztheplayer:
This follows https://github.com/apache/incubator-gluten/issues/7031.

The enhancement lies in moving the Maven module dependency `spark-sql-columnar-shims-common` / `gluten-ui` of `gluten-core` to `gluten-substrait` to make `gluten-core` more flexible for extension.

For example, once a 3rd user needs to implement a Gluten backend from scratch, they may not want to have the shim layer code of Gluten introduced from the beginning. Plus there are a bunch of copied code from Spark code base in shim layer that may impact the portability of the final Jar.

Issue #9880, opened 2025-06-05
[CORE] Move `GlutenConfig.scala` from shim layer to `gluten-core`


zhztheplayer:
There are not many usages of `GlutenConfig.scala` in the shim layer code. We'd keep the shim layer minimized and only handle essential differences between Spark versions. 

This topic is for moving `GlutenConfig.scala` to `gluten-core`.

This is also part of https://github.com/apache/incubator-gluten/issues/9881.

PR #9883, opened 2025-06-05
[GLUTEN-9880][CORE] Move `GlutenConfig.scala` from shim layer to `gluten-core`

zhztheplayer:
See #9880.

Moving the `GlutenConfig.scala` to `gluten-core` to reduce the shim code, and to prepare for #9881 also.

Actually more files are moved from shim to `gluten-core` to make the change function:

- gluten-core/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
- gluten-core/src/main/scala/org/apache/gluten/config/ConfigEntry.scala
- gluten-core/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
- gluten-core/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
- gluten-core/src/main/scala/org/apache/spark/internal/ConfigProvider.scala
- gluten-core/src/main/scala/org/apache/spark/internal/GlutenConfigUtil.scala
- gluten-core/src/main/scala/org/apache/spark/internal/SparkConfigUtil.scala
- gluten-core/src/main/scala/org/apache/spark/sql/execution/ui/GlutenUiUtils.scala


gluten-core/src/main/scala/org/apache/spark/sql/execution/ui/GlutenUiUtils.scala
```diff
@@ -17,13 +17,12 @@
...
 
 import org.apache.gluten.config.GlutenConfig
-import org.apache.gluten.config.GlutenConfig.GLUTEN_UI_ENABLED
 import org.apache.gluten.events.GlutenEvent
 
 import org.apache.spark.SparkContext
 import org.apache.spark.status.ElementTrackingStore
 
-object GlutenUIUtils {
+object GlutenUiUtils {
```
zml1206: nit: UI is the abbreviation of User Interface. Is it more appropriate to keep the status quo? The same is true for spark, such as `org.apache.spark.ui.UIUtils`.

PHILO-HE (approved):

gluten-core/src/main/scala/org/apache/spark/sql/execution/ui/GlutenUiUtils.scala
zhztheplayer: Nice catch. I was just to fix the non pascal case casually. Let's keep up with Spark though.

zhztheplayer:

zhztheplayer:
CH CI was once passed https://opencicd.kyligence.com/job/gluten/job/gluten-ci/16325, the failure shouldn't be related.

cc @zzcclp @baibaichen

PR #9890, opened 2025-06-06
Loose spark.memory.offHeap.enabled

yaooqinn:

This PR sets `spark.memory.offHeap.enabled` forcefully for the Gluten runtime and picks a suitable off-heap size according to the executor memory setting.

This makes Gluten easy for users to deploy w/ vanilla Spark together.

## How was this patch tested?


FelixYBW:
So if user set offheap enabled and size, everything works as current solution. If user didn't set offheap enabled, Gluten overrite it to true and set the offheap size to min(executor.memory,1G). This will change the overall executor memory size to 2*xexecutor.memory or (executor.memory+1G) Is it OK?

Can we use our stage level resource adjust policy to set it? Ideally user set executor.memory and Gluten automatically adjust offheap/onheap according to stage requirement.

yaooqinn:
It's difficult to deploy Gluten by maintaining two different setting copies for end users to activate. The goal here is to minimize the gap. I'm not sure if this is OK with others. I will keep this feature in-house if there are any objections here.

PR #9630, opened 2025-05-14
[GLUTEN-9540][FLINK] Port flink's unit tests to gluten

yuanhang-dev:
## What changes were proposed in this pull request?

Example of porting unit tests from Flink to Gluten

(Please fill in changes proposed in this fix)

(Fixes: #9540 )

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #9893, opened 2025-06-06
[GLUTEN-9892][FLINK] Add correctness validation tests for Nexmark queries(ddl_gen & Q0)

yuanhang-dev:

(Fixes: \#ISSUE-ID)

## How was this patch tested?


gluten-flink/runtime/src/main/java/org/apache/gluten/vectorized/ArrowVectorAccessor.java
```diff
@@ -21,15 +21,10 @@
...
 import org.apache.flink.table.data.binary.BinaryStringData;
 
-import org.apache.arrow.vector.BigIntVector;
-import org.apache.arrow.vector.BitVector;
-import org.apache.arrow.vector.DateDayVector;
-import org.apache.arrow.vector.FieldVector;
-import org.apache.arrow.vector.Float8Vector;
-import org.apache.arrow.vector.IntVector;
-import org.apache.arrow.vector.VarCharVector;
+import org.apache.arrow.vector.*;
```
shuai-xu: It's better not use * in import.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -0,0 +1,216 @@
...
+    env = StreamExecutionEnvironment.getExecutionEnvironment();
+    env.setParallelism(1);
+
+    EnvironmentSettings settings = EnvironmentSettings.newInstance().inStreamingMode().build();
+    tEnv = StreamTableEnvironment.create(env, settings);
+  }
+
+  @Test
+  void testNexmarkDDLGeneration() {
+    String createDatagenTable =
```
shuai-xu: Can we put the the queries in files, and read them from file like TPC-H tests in gluten?

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -0,0 +1,216 @@
...
+    } catch (Exception e) {
+      LOG.error("Error collecting schema info", e);
+    }
+
+    assertThat(schemaRows).hasSize(4);
+  }
+
+  @Test
+  void testNexmarkQ0Query() {
+    String createSourceTable =
```
shuai-xu: As above, we can copy the nexmark quiries in files and run them. Check  the supported ones can run without fallback and then read the results to  verify correctness.

shuai-xu (commented):
Thank you for your work, a little commets

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -0,0 +1,127 @@
...
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to read SQL file: " + fileName, e);
+    }
+  }
+
+  private List<Row> collectResults(TableResult result, int maxRows) {
+    List<Row> rows = new ArrayList<>();
+    try (CloseableIterator<Row> iterator = result.collect()) {
+      int count = 0;
+      while (iterator.hasNext() && (maxRows == -1 || count < maxRows)) {
```
shuai-xu: Better to pass Integer.max_value instead of -1.

shuai-xu (changes requested):
We'd better make nexmark source configurable, and use the original nexmark queries.

gluten-flink/ut/src/test/resources/nexmark/ddl_views.sql
```diff
@@ -0,0 +1,36 @@
...
+CREATE VIEW bid AS
+SELECT
+    bid.auction,
+    bid.bidder,
+    bid.price,
+    bid.channel,
+    bid.url,
+    `dateTime`,
+    bid.extra
+FROM datagen WHERE event_type = 2;
```
philo-he: Nit: leave one empty line at the end.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkQ0Test.java
```diff
@@ -0,0 +1,98 @@
...
+  public void setup() {
+    env = StreamExecutionEnvironment.getExecutionEnvironment();
+    env.setParallelism(1);
+
+    EnvironmentSettings settings = EnvironmentSettings.newInstance().inStreamingMode().build();
+    tEnv = StreamTableEnvironment.create(env, settings);
+  }
+
+  @Test
+  void testNexmarkQ0Query() {
```
philo-he: It seems this method can be applied for other queries. If true, rename it and this class for future extension.
Then, the below table initialization can be also shared for other queries.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkQ0Test.java
```diff
@@ -0,0 +1,98 @@
...
+    String createNexmarkSource = readSqlFromFile("nexmark/ddl_gen.sql");
+    tEnv.executeSql(createNexmarkSource);
+    String createTableView = readSqlFromFile("nexmark/ddl_views.sql");
+    String[] sqlTableView = createTableView.split(";");
+    for (String sql : sqlTableView) {
+      String trimmedSql = sql.trim();
+      tEnv.executeSql(trimmedSql);
+    }
+
+    String q0Query = readSqlFromFile("nexmark/q0.sql");
```
philo-he: Maybe, better to obtain all files (name pattern q*.sql) from the directory and handle it in loop, though currently there is only one.

gluten-flink/ut/pom.xml
```diff
@@ -90,6 +90,11 @@
             <artifactId>flink-clients</artifactId>
             <version>${flink.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.github.nexmark</groupId>
+            <artifactId>nexmark-flink</artifactId>
+            <version>0.3-SNAPSHOT</version>
```
philo-he, line 96: Better to add test scope for clarity. Please also add it for other dependencies that only required for test.

philo-he: Please clarify which code needs this dependency.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkQ0Test.java
```diff
@@ -0,0 +1,98 @@
...
+    String[] sqlStatements = q0Query.split(";");
+    String createResultTable = sqlStatements[0].trim();
+    TableResult createResult = tEnv.executeSql(createResultTable);
+    assertThat(createResult.getJobClient().isPresent()).isFalse();
+    String insertQuery = sqlStatements[1].trim();
+    TableResult insertResult = tEnv.executeSql(insertQuery);
+    assertThat(insertResult.getJobClient().isPresent()).isTrue();
+    try {
+      Thread.sleep(100000);
+    } catch (InterruptedException e) {
```
philo-he: Seems not necessary to handle this exception since it rarely happens. Maybe, just let the test method throw this exception and fail the test if it happens.

PHILO-HE (commented):
Some comments. Please confirm if it makes sense.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -0,0 +1,158 @@
...
+    LOG.info("NexmarkTest setup");
+    Velox4jEnvironment.initializeOnce();
+
+    StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+    env.setParallelism(1);
+
+    EnvironmentSettings settings = EnvironmentSettings.newInstance().inStreamingMode().build();
+    StreamTableEnvironment tEnv = StreamTableEnvironment.create(env, settings);
+
+    setupNexmarkEnvironment(tEnv);
```
philo-he, line 75: Can we move the above setup to a static method called `setup` with `@BeforeAll` annotation?

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  private Map<String, String> createNexmarkVariables() {
+    Map<String, String> variables = new HashMap<>();
+    variables.put("TPS", "10");
+    variables.put("EVENTS_NUM", "100");
+    variables.put("PERSON_PROPORTION", "1");
+    variables.put("AUCTION_PROPORTION", "3");
+    variables.put("BID_PROPORTION", "46");
```
philo-he: Seems we can directly create a static map variable for this class and initialize it when declaring?

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+  private String replaceVariables(String sql, Map<String, String> variables) {
+    String result = sql;
+    for (Map.Entry<String, String> entry : variables.entrySet()) {
+      result = result.replace("${" + entry.getKey() + "}", entry.getValue());
+    }
+    return result;
+  }
+
+  private void executeNexmarkQuery(StreamTableEnvironment tEnv, String queryFileName) {
```
philo-he: name suggestion for simplicity: executeQuery.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    String insertQuery = sqlStatements[1].trim();
+    if (!insertQuery.isEmpty()) {
+      TableResult insertResult = tEnv.executeSql(insertQuery);
+      assertThat(insertResult.getJobClient().isPresent()).isTrue();
+    }
+  }
+
+  private List<String> discoverQueryFiles() {
```
philo-he: name suggestion for simplicity: getQueries

gluten-flink/ut/pom.xml
philo-he: Should we also add test scope for junit-jupiter-api, etc?

gluten-flink/runtime/src/main/java/org/apache/gluten/vectorized/ArrowVectorAccessor.java
```diff
@@ -244,3 +247,17 @@ protected Object getImpl(int rowIndex) {
...
+class TimeStampMicroVectorAccessor extends BaseArrowVectorAccessor<TimeStampMicroVector> {
+
+  public TimeStampMicroVectorAccessor(FieldVector vector) {
+    super(vector);
+  }
+
+  @Override
+  public Object getImpl(int rowIndex) {
+    long microseconds = typedVector.get(rowIndex);
+    long milliseconds = microseconds / 1000;
```
philo-he: Nit: combine the above two lines.

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/BaseRexCallConverters.java
```diff
@@ -84,3 +86,30 @@ public TypedExpr toTypedExpr(RexCall callNode, RexConversionContext context) {
...
+  @Override
+  public TypedExpr toTypedExpr(RexCall callNode, RexConversionContext context) {
+    List<TypedExpr> params = getParams(callNode, context);
+
+    if (params.size() == 2
+        && params.get(0).getReturnType() instanceof TimestampType
+        && params.get(1).getReturnType() instanceof BigIntType) {
+
+      Type bigIntType = new BigIntType();
+      TypedExpr castedParam0 = new CallTypedExpr(bigIntType, List.of(params.get(0)), "cast");
```
philo-he: name suggestion: castedParam0->castExpr

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/BaseRexCallConverters.java
```diff
@@ -84,3 +86,30 @@ public TypedExpr toTypedExpr(RexCall callNode, RexConversionContext context) {
...
+
+  public SubtractRexCallConverter() {
+    super("subtract");
+  }
+
+  @Override
+  public TypedExpr toTypedExpr(RexCall callNode, RexConversionContext context) {
+    List<TypedExpr> params = getParams(callNode, context);
+
+    if (params.size() == 2
```
philo-he: Seems the `subtract` expression can only have two inputs? Do we need this check?

PHILO-HE (commented):
Some minor comments. Thanks!

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/BaseRexCallConverters.java
```diff
@@ -84,3 +86,30 @@ public TypedExpr toTypedExpr(RexCall callNode, RexConversionContext context) {
...
+        && params.get(0).getReturnType() instanceof TimestampType
+        && params.get(1).getReturnType() instanceof BigIntType) {
+
+      Type bigIntType = new BigIntType();
+      TypedExpr castedParam0 = new CallTypedExpr(bigIntType, List.of(params.get(0)), "cast");
+
+      List<TypedExpr> newParams = List.of(castedParam0, params.get(1));
+      return new CallTypedExpr(bigIntType, newParams, functionName);
```
philo-he, line 107: For spark, the expression's result type is generally as same as left input's type. Not sure Flink's behavior. Such handling with cast added may also be extended for other binary arithmetic expressions whose left type is not as same as right type.

We may first use this case-by-case handling.

yuanhang-dev: Here, the return type required by Flink is bigIntType.

PHILO-HE (approved):
Looks good! cc @KevinyhZou, @shuai-xu

PHILO-HE (commented):
It seems the result is not checked to see if it is consistent with Flink's. Do we need to do this?

yuanhang-dev:
> It seems the result is not checked to see if it is consistent with Flink's. Do we need to do this?

I previously discussed this problem with @shuai-xu . This ut is to ensure that the already supported Nexmark queries do not fallback. For now, it only needs to verify that the job finishes normally.

philo-he:
@yuanhang-dev, I see. Just revised the pr title to avoid misunderstanding.

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/BaseRexCallConverters.java
KevinyhZou: Should we add a ut test for this? as in my own testing, `select to_timestamp('2024-01-01 12:12:12') - cast(134455611 as bigint)` is not supported in valina flink.

yuanhang-dev: > Should we add a ut test for this? as in my own testing, `select to_timestamp('2024-01-01 12:12:12') - cast(134455611 as bigint)` is not supported in valina flink.


Good point! This is indeed an important design consideration.

This is a temporary modification made to pass nexmark q0. This converter supports timestamp and interval day-time operations. Vanilla Flink supports INTERVAL syntax like `SELECT to_timestamp('2024-01-01 12:12:12') - INTERVAL '134455611' SECOND;`

I will ensure behavioral consistency with vanilla Flink through unit tests in https://github.com/apache/incubator-gluten/pull/9630.

KevinyhZou: So here why not use `io.github.zhztheplayer.velox4j.type.IntervalDayTimeType` to replace the `bigint` type?

shuai-xu (approved):
LGTM

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/BaseRexCallConverters.java
yuanhang-dev: The conversion from `INTERVAL SECOND` to `BigIntType` happens in the upstream `RexNodeConverter.toVariant()` method. https://github.com/apache/incubator-gluten/blob/50111a44112b5e0f63d2167bb982a87c9afef0eb/gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/RexNodeConverter.java#L111

This is why `bigint` is used here

philo-he:
@KevinyhZou, do you have any other comment?

KevinyhZou:
No other comments now, thanks. @PHILO-HE

lgbo-ustc:
@yuanhang-dev  @PHILO-HE 
```
f98e465da16037310e517f6_cbc357ccb763df2852fee8c4fc7d55f2_0_0) switched from RUNNING to FAILED with failure cause:
io.github.zhztheplayer.velox4j.exception.VeloxException: Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Cannot cast TIMESTAMP '2025-06-24T04:19:13.000085000' to BIGINT. Conversion from Timestamp to Int is not supported
Retriable: False
Context: cast((dateTime) as BIGINT)
Additional Context: Top-level Expression: subtract(cast((dateTime) as BIGINT), 4000:BIGINT)
Function: toVeloxUserError
File: /home/liangjiabiao/workspace/github/velox4j/src/main/cpp/build/_deps/velox-src/velox/expression/EvalCtx.cpp
Line: 158
Stack trace:
```

yuanhang-dev:
> @yuanhang-dev @PHILO-HE
> 
> ```
> f98e465da16037310e517f6_cbc357ccb763df2852fee8c4fc7d55f2_0_0) switched from RUNNING to FAILED with failure cause:
> io.github.zhztheplayer.velox4j.exception.VeloxException: Exception: VeloxUserError
> Error Source: USER
> Error Code: INVALID_ARGUMENT
> Reason: Cannot cast TIMESTAMP '2025-06-24T04:19:13.000085000' to BIGINT. Conversion from Timestamp to Int is not supported
> Retriable: False
> Context: cast((dateTime) as BIGINT)
> Additional Context: Top-level Expression: subtract(cast((dateTime) as BIGINT), 4000:BIGINT)
> Function: toVeloxUserError
> File: /home/liangjiabiao/workspace/github/velox4j/src/main/cpp/build/_deps/velox-src/velox/expression/EvalCtx.cpp
> Line: 158
> Stack trace:
> ```

Can you tell me the SQL statement executed when encountering this error? @lgbo-ustc

lgbo-ustc:
mvn test failed. 

May ignore this first

yuanhang-dev:
> mvn test failed.

I just ran `mvn test` on the main branch, and it worked. Could you provide me with more context?

lgbo-ustc:
The failure is in `NexmarkTest`. I believe the velox and velox4j has been updated

lgbo-ustc:
btw, time interval is supported in https://github.com/apache/incubator-gluten/pull/9781

yuanhang-dev:
> The failure is in `NexmarkTest`. I believe the velox and velox4j has been updated

I think the problem might be with the version of velox4j; the version used here is in `bigo/gluten 16607ee`

> btw, time interval is supported in https://github.com/apache/incubator-gluten/pull/9781

I saw the pr before, the modification here is only to pass the nexmark q0 test and can be overridden later.

PR #9894, opened 2025-06-06
[VL][CI] Remove unnecessary downloads for arrow jars

PHILO-HE:


philo-he:
According to CI feedback, when running gluten that native code is built on centos-7 together with arrow built on centos-8, core dump happens.
```
37:03.946 WARN org.apache.gluten.columnarbatch.IndicatorVectorPool: There are still unreleased native columnar batches during ending the task. Will close them automatically however the batches should be better released manually to minimize memory pressure.

#
#
#  SIGSEGV (0xb) at pc=0x00007fd66f8191e9, pid=55284, tid=57337
#
# JRE version: OpenJDK Runtime Environment 21.9 (17.0.1+12) (build 17.0.1+12-LTS)
# Java VM: OpenJDK 64-Bit Server VM 21.9 (17.0.1+12-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
# Problematic frame:
# C  [libgluten.so+0x9041e9]  std::vector<arrow::compute::Expression, std::allocator<arrow::compute::Expression> >::~vector()+0x29
#
# Core dump will be written. Default location: Core dumps may be processed with "/lib/systemd/systemd-coredump %P %u %g %s %t 9223372036854775808 %h" (or dumping to /__w/incubator-gluten/incubator-gluten/gluten-ut/spark35/core.55284)
#
# /__w/incubator-gluten/incubator-gluten/gluten-ut/spark35/hs_err_pid55284.log
```

PR #9899, opened 2025-06-06
[GLUTEN-9860] followup: Fix rsync remote path

zhouyuan:

this patch adds the trailing '/' for the remote path. 
example from opendal: 
https://github.com/apache/opendal/blob/main/.github/workflows/docs.yml#L715

## How was this patch tested?

pass GHA

### DIFF
--- a/.github/workflows/nightly_sync.yml
+++ b/.github/workflows/nightly_sync.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           switches: -avzr
           path: docs/*
-          remote_path: ${{ secrets.NIGHTLIES_RSYNC_PATH }}/gluten/docs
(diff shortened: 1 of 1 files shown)


yaooqinn:
The CI is still unrecovered.

Probably, you missed `incubator-` for gluten.

zhouyuan:
@yaooqinn Thanks for the tip! Sorry for introducing serval "noisy" patches on this - I will confirm with the infra team and update back.

zhouyuan:
It turns out the initial directory needs to be created firstly - the apache infra team helped to created the directory and now the docs are already uploaded
https://nightlies.apache.org/gluten/docs/

Issue #9639, opened 2025-05-14
[VL] Merge modules `gluten-arrow` and `backends-velox`


zhztheplayer:
### Description

This will mean merging Velox's CPP code base `cpp/core` and `cpp/velox` altogether at the same time.

We had some discussions like https://github.com/apache/incubator-gluten/issues/4201 which becomes more reasonable over time. At the very beginning, `cpp/core` and `cpp/velox` were split for extensibility purpose, but we don't seem to have chance to leverage this extensibility anymore. ClickHouse backend is unlikely going to rely on `cpp/core` in the future either. Maintaining this extensible design is costly for the community and has no clear reason.

The work will have two parts:

1. Merge Maven module `gluten-arrow` into `backends-velox`.
2. Merge `cpp/core` to `cpp/velox`.

### Gluten version

None

PR #9900, opened 2025-06-06
[GLUTEN-9881] Minimize module dependency set of module gluten-core

zhztheplayer:
#9881 

TBD


gluten-arrow/pom.xml
```diff
@@ -47,7 +47,19 @@
   <dependencies>
     <dependency>
       <groupId>org.apache.gluten</groupId>
-      <artifactId>gluten-core</artifactId>
+      <artifactId>gluten-substrait</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
```
zhztheplayer, line 53: Temporarily make `gluten-arrow` rely on `gluten-substrait` since some classes are moved from `gluten-core` to `gluten-substrait`. In the future we'd merge `backends-velox` and `gluten-arrow` into one.

https://github.com/apache/incubator-gluten/issues/9639

zhztheplayer:
@yaooqinn Thanks for reviewing!

PR #9914, opened 2025-06-09
Add a RichSparkConf to simplify interoperations gluten config entries

yaooqinn:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHListenerApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHListenerApi.scala
@@ -101,7 +101,7 @@ class CHListenerApi extends ListenerApi with Logging {
     }
 
     // add memory limit for external sort
-    if (conf.get(RuntimeSettings.MAX_BYTES_BEFORE_EXTERNAL_SORT) < 0) {
+    if (conf.get(RuntimeSettings.MAX_BYTES_BEFORE_EXTERNAL_SORT) <= 0) {
       if (conf.getBoolean(GlutenConfig.SPARK_OFFHEAP_ENABLED, defaultValue = false)) {
         val memSize = conf.getSizeAsBytes(GlutenConfig.SPARK_OFFHEAP_SIZE_KEY, 0)
         if (memSize > 0L) {
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHListenerApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHListenerApi.scala
@@ -30,10 +30,10 @@ import org.apache.spark.{SPARK_VERSION, SparkConf, SparkContext}
 import org.apache.spark.api.plugin.PluginContext
 import org.apache.spark.internal.Logging
 import org.apache.spark.listener.CHGlutenSQLAppStatusListener
-import org.apache.spark.network.util.JavaUtils
 import org.apache.spark.rpc.{GlutenDriverEndpoint, GlutenExecutorEndpoint}
 import org.apache.spark.sql.execution.datasources.GlutenWriterColumnarRules
 import org.apache.spark.sql.execution.datasources.v1._
+import org.apache.spark.sql.internal.SparkConfigUtil._
 import org.apache.spark.sql.utils.ExpressionUtil
 import org.apache.spark.util.{SparkDirectoryUtil, SparkShutdownManagerUtil}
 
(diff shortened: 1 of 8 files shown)


backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHListenerApi.scala
```diff
@@ -102,14 +101,14 @@ class CHListenerApi extends ListenerApi with Logging {
     }
 
     // add memory limit for external sort
-    if (conf.getLong(RuntimeSettings.MAX_BYTES_BEFORE_EXTERNAL_SORT.key, -1) < 0) {
-      if (conf.getBoolean("spark.memory.offHeap.enabled", defaultValue = false)) {
-        val memSize = JavaUtils.byteStringAsBytes(conf.get("spark.memory.offHeap.size"))
+    if (conf.get(RuntimeSettings.MAX_BYTES_BEFORE_EXTERNAL_SORT) <= 0) {
```
zml1206, line 104: nit:  `== 0` ?  Also add .checkValue(_ >= 0, "must not be negative.")
https://clickhouse.com/docs/operations/settings/settings#max_bytes_before_external_sort
Possible values:
Positive integer.
0 — Memory control is disabled.
cc @baibaichen

yaooqinn: This PR is more about refactoring, and we can revisit this part by the module owners

gluten-core/src/main/scala/org/apache/spark/sql/internal/SparkConfigUtil.scala
```diff
@@ -16,11 +16,56 @@
...
+    entry.valueConverter(conf.getOrDefault(entry.key, entry.defaultValueString))
+  }
+
+  def set[T](conf: SparkConf, entry: SparkConfigEntry[T], value: T): SparkConf = {
+    conf.set(entry, value)
+  }
+
+  def set[T](conf: SparkConf, entry: ConfigEntry[T], value: T): SparkConf = {
+    value match {
+      case Some(v) => conf.set(entry.key, v.toString)
```
zml1206, line 74: Why use `v.toString` instead of `entry.stringConverter(v)`?

yaooqinn: This is a limitation of the native side, which cannot handle values like `1234b`

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -169,22 +160,19 @@ private[gluten] class GlutenDriverPlugin extends DriverPlugin with Logging {
...
     // Spark SQL extensions
-    val extensionSeq =
-      SparkConfigUtil.getEntryValue(conf, SPARK_SESSION_EXTENSIONS).getOrElse(Seq.empty)
+    val extensionSeq = conf.get(SPARK_SESSION_EXTENSIONS).getOrElse(Seq.empty)
     if (!extensionSeq.toSet.contains(GlutenSessionExtensions.GLUTEN_SESSION_EXTENSION_NAME)) {
       conf.set(
-        SPARK_SESSION_EXTENSIONS.key,
-        (extensionSeq :+ GlutenSessionExtensions.GLUTEN_SESSION_EXTENSION_NAME).mkString(","))
+        SPARK_SESSION_EXTENSIONS,
+        Some(extensionSeq :+ GlutenSessionExtensions.GLUTEN_SESSION_EXTENSION_NAME))
```
zml1206: nit: How about add a new function `def set[T](entry: OptionalConfigEntry[T], value: T): SparkConf`? 
This way, there is no need for `Some`.

yaooqinn: 👌

gluten-core/src/main/scala/org/apache/spark/sql/internal/SparkConfigUtil.scala
```diff
@@ -16,11 +16,64 @@
...
+  def set[T](conf: SparkConf, entry: SparkConfigEntry[T], value: T): SparkConf = {
+    conf.set(entry, value)
+  }
+
+  def set[T](conf: SparkConf, entry: OptionalConfigEntry[T], value: T): SparkConf = {
+    conf.set(entry, value)
+  }
+
+  def set[T](conf: SparkConf, entry: ConfigEntry[T], value: T): SparkConf = {
+    value match {
```
zml1206, line 73: It seems that there is no need for match here.

yaooqinn: No

zml1206 (approved):

PR #9915, opened 2025-06-09
[GLUTEN-9901][VL] RAS: DistinguishIdenticalScans rule to distinguish identical scans

### LINKED ISSUE #9901: Gluten RAS miscounts metrics for streaming queries
### Backend

VL (Velox)

### Bug description

I added a unit test in https://github.com/apache/incubator-gluten/blob/b865e5f7eb3c16bd97a3c99d6de98e5cfb3e6084/gluten-ut/spark35/src/test/scala/org/apache/spark/sql/execution/GlutenStreamingQuerySuite.scala

and it failed ONLY WHEN `RAS` is enabled. Error logs:

```
2025-06-06T18:42:55.4541087Z GlutenStreamingQuerySuite:
2025-06-06T18:42:55.4959057Z 18:42:55.495 WARN org.apache.spark.sql.execution.streaming.ResolveWriteToStream: Temporary checkpoint location created which is deleted normally when the query didn't fail: /tmp/temporary-40bbc770-4310-4878-9d7b-6b05c59e5ae0. If it's required to delete it under any circumstances, please set spark.sql.streaming.forceDeleteTempCheckpointLocation to true. Important to know deleting temp checkpoint folder is best effort.
2025-06-06T18:42:55.4961626Z 
2025-06-06T18:42:55.4962614Z 18:42:55.495 WARN org.apache.spark.sql.execution.streaming.ResolveWriteToStream: spark.sql.adaptive.enabled is not supported in streaming DataFrames/Datasets and will be disabled.
2025-06-06T18:42:55.4963674Z 
2025-06-06T18:42:56.7794827Z - input row calculation with same V1 source used twice in self-join *** FAILED ***
2025-06-06T18:42:56.7796438Z   40 did not equal 20 (GlutenStreamingQuerySuite.scala:48)
```

The test checks the `numInputRows` metrics of a self-join. The correct number shall be 20, but gluten (under RAS mode) gives 40.

### Gluten version

_No response_

### Spark version

None

### Spark configurations

_No response_

### System information

_No response_

### Relevant logs

```bash

```

li-boxuan:

Fixes #9901: Gluten RAS miscounts metrics for streaming queries

## How was this patch tested?

Added a unit test suite that extends Spark test suite. A few tests would fail without the fix in this PR. The failure is also demonstrated in #9901.


zhztheplayer (commented):
@li-boxuan Thank you for your work! 

I found this in the code comment:

> The worst possible consequence for an over-strict equality check is acceptable

This does seem possible. I think there is another safer approach to have an individual Spark rule to check for duplicated leaf scan nodes, then to replace the duplicated ones for distinction? This rule could be placed after the RAS core EnumeratedTransform rule.

By all means to make sure each leaf scan node in the query DAG has only one single parent, am I understanding correctly here?

li-boxuan:
> @li-boxuan Thank you for your work!
> 
> I found this in the code comment:
> 
> > The worst possible consequence for an over-strict equality check is acceptable (bigger search space)
> 
> This does seem possible. A bad case could be a rule that simply duplicates an input RDDScanNode makes the search endless. I think there is another safer approach to have an individual Spark rule to check for identical leaf scan nodes, then to renew the identical ones for distinction (i.e., copy the duplicated leaf nodes with nothing changed inside)? This rule could be placed after the RAS core EnumeratedTransform rule.
> 
> By all means to make sure each leaf scan node in the query DAG has only one single parent, am I understanding correctly here?

I gave your suggestion a try;

Core logic looks like this:

```scala
object DistinguishIdenticalScans extends Rule[SparkPlan] {

    // Set to track seen scan objects
    val seenScans = mutable.Set[SparkPlan]()
    
    // Custom traversal to avoid transformUp's fastEquals optimization
    def traverse(node: SparkPlan): (SparkPlan, Boolean) = {
      // First, check if this node is a scan and handle scan distinction logic
      val (nodeToProcess, scanChanged) = node match {
        case scan: FileSourceScanExec =>
          distinguishScan(scan, seenScans)
        case scan: BatchScanExec =>
          distinguishScan(scan, seenScans)
        case scan: DataSourceV2ScanExecBase =>
          distinguishScan(scan, seenScans)
        case scan: RDDScanExec =>
          distinguishScan(scan, seenScans)
        case other => (other, false)
      }
      
      // Then, recursively process children of the (possibly cloned) node
      var childrenChanged = false
      val newChildren = nodeToProcess.children.map { child =>
        val (newChild, changed) = traverse(child)
        if (changed) childrenChanged = true
        newChild
      }
      
      // Create final node with new children if any children changed
      val finalNode = if (childrenChanged) {
        // TROUBLE!!!! Spark internally checks equivalence and then skips this
        nodeToProcess.withNewChildren(newChildren)
      } else {
        nodeToProcess
      }
      
      (finalNode, scanChanged || childrenChanged)
    }
    
    val (newPlan, _) = traverse(plan)
    newPlan
  }

  private def distinguishScan[T <: SparkPlan](
      scan: T,
      seenScans: mutable.Set[SparkPlan]): (SparkPlan, Boolean) = {
    if (seenScans.contains(scan)) {
      // Scan already seen, clone it to distinguish from the original
      val newScan = cloneScan(scan)
      (newScan, true)
    } else {
      // First time seeing this scan
      seenScans.add(scan)
      (scan, false)
    }
  }

  private def cloneScan(scan: SparkPlan): SparkPlan = {
    scan.clone()
  }
}
```

Unfortunately Spark doesn't expose a public method to force replacing an arbitrary operator. It always checks for "equivalence" and ignores our attempt to clone a Scan operator.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/enumerated/planner/plan/GlutenPlanModel.scala
```diff
@@ -76,24 +76,40 @@ object GlutenPlanModel {
     }
 
     private def withEqualityWrapper(node: SparkPlan): AnyRef = node match {
+      case scan: RDDScanExec =>
```
li-boxuan: Need

```
case scan if scan.isInstanceOf[RDDScanExec] || scan.isInstanceOf[FileSourceScanExec] =>
```

to pass `StreamingSelfUnionSuite`

li-boxuan:
I suspect the 2 failing checks are transient... from the raw logs I cannot find any occurrence of `*** FAILED ***`

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/execution/GlutenStreamingQuerySuite.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.gluten.config.GlutenConfig
+
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.GlutenSQLTestsTrait
+import org.apache.spark.sql.streaming._
+
+class GlutenStreamingQuerySuite extends StreamingQuerySuite with GlutenSQLTestsTrait {
+  override def sparkConf: SparkConf =
+    super.sparkConf
+      .set(GlutenConfig.RAS_ENABLED.key, "true")
```
zhztheplayer: We had this CI job https://github.com/apache/incubator-gluten/blob/aa485e53767d7ebac12477314693577ebc683e88/.github/workflows/velox_backend_x86.yml#L1051 running Spark UTs with `RAS=ON` already. So probably we don't have to explicitly set this option to test RAS.

zhztheplayer (commented):
Thanks @li-boxuan.

> Unfortunately Spark doesn't expose a public method to force replacing an arbitrary operator. It always checks for "equivalence" and ignores our attempt to clone a Scan operator.

Yes, may we call the protected `withNewChildrenInternal` by a utility class for package hacking? 

There are some existing practices for reference, e.g.,

https://github.com/apache/incubator-gluten/blob/main/gluten-core/src/main/scala/org/apache/spark/util/SparkShutdownManagerUtil.scala

I am on leave this week and sorry for the delayed reply here.

li-boxuan:
I reverted my fix as of now. Let's see if CI would fail due to the new test (which would be expected).

li-boxuan:
As expected, 3 tests failed:

> 2025-06-21T18:13:54.1051727Z - input row calculation with same V1 source used twice in self-join *** FAILED ***

> 2025-06-21T18:13:54.9965813Z - input row calculation with same V2 source used twice in self-union *** FAILED ***

> 2025-06-21T18:13:56.6916495Z - input row calculation with same V2 source used twice in self-join *** FAILED ***

li-boxuan:
> may we call the protected withNewChildrenInternal by a utility class for package hacking?

That doesn't seem to work. The example you showed works because you need to hack a private object's non-private method. In my case, I need to hack a public object's protected method. Seems the only way is to use reflection.

I modified Spark code locally to make that method public, and this, together with the new DistinguishIdenticalScans rule, works... meaning this is a viable approach as long as we can work around the protected modifier.

UPDATE: I chose the reflection way

li-boxuan:
One test suite is failing but seems transient and unrelated to my PR.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/DistinguishIdenticalScans.scala
```diff
@@ -0,0 +1,79 @@
...
+      // 4. The final 'changed' status is true if EITHER the children changed OR
+      // the current node changed.
+      val overallChanged = childrenChanged || currentNodeChanged
+      (finalNode, overallChanged)
+    }
+
+    def distinguish[T <: SparkPlan](scan: T): (T, Boolean) = {
+      if (seenScans.contains(scan)) {
+        // A clone was made, so return the new object and 'true'.
+        (scan.clone().asInstanceOf[T], true)
```
zhztheplayer, line 68: Do we need to copy tags to the new node?

zhztheplayer (approved):
@li-boxuan LGTM % a minor comment. Thank you for the PR!

gluten-core/src/main/scala/org/apache/spark/sql/execution/SparkPlanReflectionUtil.scala
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.spark.sql.execution
+
+import org.apache.spark.sql.catalyst.trees.TreeNode
+
+import java.lang.reflect.Method
+
+/**
+ * This is a hack to access withNewChildrenInternal method of TreeNode, which is a protected method
+ */
+object SparkPlanReflectionUtil {
```
zhztheplayer: Perhaps we can call this `SparkPlanUtil`? Users don't have to know the internal of this utility.

I know we already have a SparkPlanUtil but that one is in another package so we should be fine.

zhztheplayer (commented):
Just another. Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/DistinguishIdenticalScans.scala
li-boxuan: Fortunately, clone() method calls `makeCopy` method which would copy tags.

PR #8987, opened 2025-03-13
[GLUTEN-8948][VL] Fallback iceberg delete from scan

jinchengchenghh:

PR #9927, opened 2025-06-10
[GLUTEN-8948] [VL] Rename the iceberg test name to starts with Test

jinchengchenghh:


PHILO-HE (commented):
Hi @jinchengchenghh, the naming convention for Spark test class is using "Gluten" as prefix. Should we keep following this convention for iceberg test?

jinchengchenghh:
Please see this comment https://github.com/apache/incubator-gluten/pull/8987#issuecomment-2954889616

philo-he:
I see. According to the given link in pr description, it seems we can also override the name patterns for this plugin by setting `<includes>` field. Which way is preferred?

jinchengchenghh:
I would prefer to rename the files, GlutenTest prefix is not a valid name, we add the pattern only if it is essential.

PHILO-HE (approved):

jinchengchenghh:
Could you help search if `TestGlutenCopyOnWriteDelete` is really running? We network is poor cannot download the log archive, https://github.com/apache/incubator-gluten/actions/runs/15551583869/job/43783159758?pr=9927 @PHILO-HE

philo-he:
> Could you help search if `TestGlutenCopyOnWriteDelete` is really running? We network is poor cannot download the log archive, https://github.com/apache/incubator-gluten/actions/runs/15551583869/job/43783159758?pr=9927 @PHILO-HE

@jinchengchenghh, it's strange. Seems this test is not found from the log which is obtained from "Download log archive" as below.

image

jinchengchenghh:
Thanks, I will try to find out why

zhztheplayer:
Thank you for working on the fix!

PR #9934, opened 2025-06-10
[VL] Download Maven using dlcdn.apache.org

zml1206:

CI often encounters timeout when downloading Maven. `dlcdn.apache.org` is the latest download address officially recommended by Apache. It is more stable and faster than `downloads.apache.org`. Try using it to see if it can solve the timeout problem.


yaooqinn:
We shall not use `dlcdn` as it usually only keeps the latest versions

zml1206:
> We shall not use `dlcdn` as it usually only keeps the latest versions

downloads.apache.org also only keeps the latest versions, archive.apache.org don't, but it is easy to be restricted.

zml1206:
What do you think of using dlcdn instead of downloads and keeping the Maven version updated? @yaooqinn

yaooqinn:
> What do you think of using dlcdn instead of downloads and keeping the Maven version updated?

It's tough to maintain such sync progress, especially in a multi branch develop mode

zml1206:
@yaooqinn Could you help me take a look again? The error seems unrelated, cc @PHILO-HE

yaooqinn (approved):

PHILO-HE (approved):
Looks good!

PR #9944, opened 2025-06-11
[GLUTEN-9942] Refactor the JNI for the native ShuffleWriter and PartitionWriter

marin-ma:
Currently we have three types of native ShuffleWriter: hash, sort and rss_sort
And two types of native PartitionWriter: LocalPartitionWriter, RssPartitionWriter
image

The creation of all types are using the same JNI call, which results in a long argument list but many of them are only used by certain types. The JNI should be refactored to separate different kinds of creation, and make it easier to add new arguments for certain types or add new types of ShuffleWriter or PartitionWriter.

This patch separate the creation of native shuffle writer and partition writer, allowing more kinds of compositions for the shuffle writer and partition writer. For example, we can add another configuration for choosing between SortShuffleWriter or RssSortShuffleWriter for celeborn sort-based shuffle.

Please see below chart for the architecture after the change, and the yellow/white part demonstrates how to add a new shuffle implementation in Gluten.

image

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -35,7 +35,6 @@
 #include "shuffle/ShuffleReader.h"
 #include "shuffle/ShuffleWriter.h"
 #include "shuffle/Utils.h"
-#include "shuffle/rss/RssPartitionWriter.h"
 #include "utils/ArrowStatus.h"
 #include "utils/StringUtil.h"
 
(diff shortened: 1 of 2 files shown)


backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala
```diff
@@ -119,28 +124,42 @@ class VeloxCelebornColumnarShuffleWriter[K, V](
...
       clientPushBufferMaxSize,
       clientPushSortMemoryThreshold,
-      celebornPartitionPusher,
-      ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, columnarBatch),
-      context.taskAttemptId(),
-      GlutenShuffleUtils.getStartPartitionId(dep.nativePartitioning, context.partitionId),
-      "celeborn",
-      shuffleWriterType,
-      GlutenConfig.get.columnarShuffleReallocThreshold
+      celebornPartitionPusher
```
zhouyuan, line 135: Cc: @kerwin-zk for this API change

kerwin-zk: @zhouyuan It looks like the heap space for the test case was insufficient, causing some instability. I reran it and it worked fine.

marin-ma: CI was all green https://github.com/apache/incubator-gluten/actions/runs/15677369131/job/44173006095?pr=9944

I will do the rebase to resolve the new conflict.

FelixYBW:
It's odd the second chart can't be opened

gluten-celeborn/src/main/scala/org/apache/spark/shuffle/CelebornColumnarShuffleWriter.scala
```diff
@@ -106,9 +106,6 @@ abstract class CelebornColumnarShuffleWriter[K, V](
       .getOrElse(0)
   }
 
-  protected val bufferCompressThreshold: Int =
-    GlutenConfig.get.columnarShuffleCompressionThreshold
-
```
zhouyuan, line 111: It seems we removed this config, maybe we should also update here?
https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala#L182

zhouyuan (approved):
Thanks!

gluten-celeborn/src/main/scala/org/apache/spark/shuffle/CelebornColumnarShuffleWriter.scala
marin-ma: It's used by the LocalPartitionWriter.

cpp/core/jni/JniWrapper.cc
```diff
@@ -780,141 +780,223 @@ JNIEXPORT jlong JNICALL Java_org_apache_gluten_columnarbatch_ColumnarBatchJniWra
...
+      partitionWriterOptions,
+      dataFile,
+      std::move(localDirs));
+
+  return ctx->saveObject(partitionWriter);
+  JNI_METHOD_END(kInvalidObjectHandle)
+}
+
+JNIEXPORT jlong JNICALL
+Java_org_apache_gluten_vectorized_CelebornPartitionWriterJniWrapper_createPartitionWriter( // NOLINT
```
zhztheplayer: (minor)

The function could be moved to VeloxJniWrapper.cc given the relevant Java JNI wrapper is moved, ditto for the one for Uniffle.

Once https://github.com/apache/incubator-gluten/issues/9639 is done we will no longer need to manage these module mappings.

Issue #9952, opened 2025-06-12
Vulnerable versions of Jupyter_core and requests in velox benchmark dependencies


ashahba:
### Description

Current versions of `jupyter_core` and `requests` within tools/workload/benchmark_velox/analysis/requirements.txt b/tools/workload/benchmark_velox/analysis/requirements.txt are both flagged as vulnerable.

We need:
```
jupyter_core==5.8.1
requests==2.32.4
```

### Gluten version

main branch

PR #9954, opened 2025-06-12
[GLUTEN-9952][VL] Python Security updates June 2025

ashahba:
## The following file was updated to ensure the desired versions are installed:
tools/workload/benchmark_velox/analysis/requirements.txt b/tools/workload/benchmark_velox/analysis/requirements.txt 

(Fixes: #9952)

## How was this patch tested?
Mostly just creating a fresh virtualenv for 3 different Python versions (3.10, 3.11, 3.12) and making sure that pip can install this requirements file without any issues and using no cached versions.


zhouyuan (approved):
Thanks!

Issue #9953, opened 2025-06-12
Add a weekly job to check compatibility with openEuler


kevinw66:
### Description

Add a weekly job to check compatibility with openEuler

### Gluten version

main branch

PR #9955, opened 2025-06-12
[GLUTEN-9953][VL] Add a weekly job to check compatibility with openEuler

kevinw66:
Fixes: #9953 

## How was this patch tested?

unit tests


.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,102 @@ jobs:
...
+          name: velox-native-lib-openeuler-24-${{github.sha}}
+          path: ./cpp/build/releases/
+          if-no-files-found: error
+      - uses: actions/upload-artifact@v4
+        with:
+          name: arrow-jars-openeuler-24-${{github.sha}}
+          path: .m2/repository/org/apache/arrow/
+          if-no-files-found: error
+
+  tpc-test-openeuler:
```
philo-he: I think it may be better to simply start a docker to run this tpc-h test in the above `build-on-openeuler-vcpkg` job after the build. Then, no need to upload artifacts, but only need to mount some paths to the docker, e.g., mount .m2 to include arrow jars, and mount $GITHUB_WORKSPACE to include native libs at /cpp/build/releases/ and gluten source code.

By doing so, we can reduce the complexity.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,102 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: build
+        run: |
+          dnf update -y && dnf install -y java-1.8.0-openjdk-devel sudo patch wget git perl
+          $SETUP install_maven
+          export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk
+          export MAVEN_HOME=/usr/lib/maven
+          export PATH=${PATH}:${MAVEN_HOME}/bin:${JAVA_HOME}/bin
+          cd $GITHUB_WORKSPACE/ && ./dev/package.sh
```
philo-he: You can start a docker here to run a tpc-h test also. Please set `spark.gluten.loadLibFromJar=true` to make sure all required native libs packed in third-party jar are loaded at runtime.

The code path of loading lib from jar is not verified in CI. We are considering to add this test for other OS in subsequent pr.

PHILO-HE (commented):
Thanks for your work! Some comments for making it simpler a bit.

.github/workflows/velox_weekly.yml
kevinw66: Why use mount? can't we just copy `Build gluten-it` and `Run TPC-H` steps and place them after the `build` step?

philo-he: @kevinw66, I meant let's start a new docker to run the test, `docker run xxx bash -c "xxx"`, which can expose any libs lacking issue in deployment environment. “mount" is just for convenience to use some binaries built by previous step.

kevinw66:
> Thanks for your work! Some comments for making it simpler a bit.

@PHILO-HE Add tpch tests for both jobs, thanks!

backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoader.scala
```diff
@@ -124,7 +126,7 @@ object SharedLibraryLoader {
     } else {
       throw new GlutenException(
         s"Found unsupported OS($systemName, $systemVersion)! Currently, Gluten's Velox backend" +
-          " only supports Ubuntu 20.04/22.04, CentOS 7/8, Oracle 7/8" +
+          " only supports Ubuntu 20.04/22.04, CentOS 7/8, Oracle 7/8, openEuler 24.03 " +
```
philo-he: Nit: putting a comma next to "openEuler 24.03".

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,86 @@ jobs:
...
+          export PATH=${MAVEN_HOME}/bin:${JAVA_HOME}/bin:${PATH}
+          export VCPKG_PATH=/var/cache/vcpkg
+          export VCPKG_BINARY_SOURCES="clear;files,${VCPKG_PATH},readwrite"
+          cd $GITHUB_WORKSPACE/ 
+          ./dev/vcpkg/setup-build-depends.sh
+          mkdir -p ${VCPKG_PATH}
+          sed -i 's/--build_arrow=OFF/--build_arrow=ON/' dev/package-vcpkg.sh && ./dev/package-vcpkg.sh
+      - name: Set environment variables
+        run: |
+          echo "JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk" >> $GITHUB_ENV
```
philo-he, line 150: Nit: Let's move this setting to the above task for brevity. Seems we can directly replace the above `export JAVA_HOME=` with this setting.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,86 @@ jobs:
...
+          dnf update -y && dnf install -y java-1.8.0-openjdk-devel sudo patch wget git perl
+          $SETUP install_maven
+          export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk
+          export MAVEN_HOME=/usr/lib/maven
+          export PATH=${MAVEN_HOME}/bin:${JAVA_HOME}/bin:${PATH}
+          cd $GITHUB_WORKSPACE/ && ./dev/package.sh
+          cp package/target/thirdparty-lib/gluten-thirdparty-lib-*.jar /root/
+      - name: Set environment variables
+        run: |
+          echo "JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk" >> $GITHUB_ENV
```
philo-he, line 120: ditto.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,86 @@ jobs:
...
+    container: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: build
+        run: |
+          dnf update -y && dnf install -y java-1.8.0-openjdk-devel sudo patch wget git perl
+          $SETUP install_maven
+          export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk
+          export MAVEN_HOME=/usr/lib/maven
+          export PATH=${MAVEN_HOME}/bin:${JAVA_HOME}/bin:${PATH}
```
philo-he: Remove setting MAVEN_HOME and PATH.
`$SETUP install_maven` have set it for maven, and setting PATH with `${JAVA_HOME}/bin` is also unnecessary maybe.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,86 @@ jobs:
...
```
philo-he: ditto

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,86 @@ jobs:
...
           export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
           cd $GITHUB_WORKSPACE/ && ./dev/package.sh
+
+  build-on-openeuler:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ "openeuler/openeuler:24.03" ]
+        spark: [ "spark-3.5" ]
+        java: [ "java-8" ]
```
philo-he: Remove unnecessary matrix setting. Directly use the values in the below. What we added are basic tests to verify native libs. So I think we will not extend the tests for other JDKs or Spark versions in the future.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,86 @@ jobs:
...
+            --extra-conf=spark.gluten.loadLibFromJar=true \
+            --extra-conf=spark.gluten.loadLibOS=openEuler --extra-conf=spark.gluten.loadLibOSVersion=24.03
+
+  build-on-openeuler-vcpkg:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ "openeuler/openeuler:24.03" ]
+        spark: [ "spark-3.5" ]
+        java: [ "java-8" ]
```
philo-he: ditto.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,86 @@ jobs:
...
+          $SETUP install_maven
+          export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk
+          export MAVEN_HOME=/usr/lib/maven
+          export PATH=${MAVEN_HOME}/bin:${JAVA_HOME}/bin:${PATH}
+          export VCPKG_PATH=/var/cache/vcpkg
+          export VCPKG_BINARY_SOURCES="clear;files,${VCPKG_PATH},readwrite"
+          cd $GITHUB_WORKSPACE/ 
+          ./dev/vcpkg/setup-build-depends.sh
+          mkdir -p ${VCPKG_PATH}
+          sed -i 's/--build_arrow=OFF/--build_arrow=ON/' dev/package-vcpkg.sh && ./dev/package-vcpkg.sh
```
philo-he: Can we just use `./dev/package-vcpkg.sh --build_arrow=ON`? It may override the internal build arrow setting.

PHILO-HE (commented):
Basically looks good! Some minor comments. Thanks!

kevinw66:
@PHILO-HE All fixed, thanks!

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,68 @@ jobs:
...
+          dnf update -y && dnf install -y java-1.8.0-openjdk-devel sudo patch wget git perl
+          $SETUP install_maven
+          echo "JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk" >> $GITHUB_ENV
+      - name: Build
+        run: |
+          echo "JAVA_HOME: $JAVA_HOME"
+          cd $GITHUB_WORKSPACE/
+          ./dev/package.sh
+          cp package/target/thirdparty-lib/gluten-thirdparty-lib-*.jar /root/
+          $MVN_CMD clean install -Pspark-3.5 -Pjava-8 -Pbackends-velox -DskipTests
```
philo-he: Without spark version specified, package.sh will build gluten for all supported Spark versions, including spark-3.5. To avoid redundant build and unnecessary build, please use `./deve/package.sh  --spark_version=3.5` and remove the above `mvn clean install` command.

If `mvn install` is necessary in gluten-it test, I think we can directly modify `buildbundle-veloxbe.sh` (called by package.sh) to replace `mvn clean pakcage` with `mvn clean install`.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,68 @@ jobs:
...
+          dnf update -y && dnf install -y java-1.8.0-openjdk-devel sudo patch wget git perl
+          $SETUP install_maven
+          echo "JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk" >> $GITHUB_ENV
+      - name: Build
+        run: |
+          echo "JAVA_HOME: $JAVA_HOME"
+          cd $GITHUB_WORKSPACE/
+          ./dev/vcpkg/setup-build-depends.sh
+          ./dev/package-vcpkg.sh --build_arrow=ON
+          $MVN_CMD clean install -Pspark-3.5 -Pjava-8 -Pbackends-velox -DskipTests
```
philo-he: ditto

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,68 @@ jobs:
...
+          cd $GITHUB_WORKSPACE/tools/gluten-it
+          $MVN_CMD clean install -Pspark-3.5 -Pjava-8
+      - name: Run TPC-H
+        run: |
+          echo "JAVA_HOME: $JAVA_HOME"
+          cd $GITHUB_WORKSPACE/tools/gluten-it
+          GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
+            --local --preset=velox --benchmark-type=h --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1
```
philo-he: Nit:
To reduce duplication, please create an env variable, like `TPCH_TEST`, in the `env` section of this file to use it in the two jobs and future new jobs.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,68 @@ jobs:
...
+          $MVN_CMD clean install -Pspark-3.5 -Pjava-8
+      - name: Run TPC-H
+        run: |
+          echo "JAVA_HOME: $JAVA_HOME"
+          cd $GITHUB_WORKSPACE/tools/gluten-it
+          cp /root/gluten-thirdparty-lib-*.jar package/target/lib/
+          GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
+            --local --preset=velox --benchmark-type=h --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1 \
+            --extra-conf=spark.gluten.loadLibFromJar=true \
+            --extra-conf=spark.gluten.loadLibOS=openEuler --extra-conf=spark.gluten.loadLibOSVersion=24.03
```
philo-he, line 134: Use the proposed env variable with extra configs appended.

.github/workflows/velox_weekly.yml
```diff
@@ -99,3 +100,68 @@ jobs:
...
+        run: |
+          dnf update -y && dnf install -y java-1.8.0-openjdk-devel sudo patch wget git perl
+          $SETUP install_maven
+          echo "JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk" >> $GITHUB_ENV
+      - name: Build
+        run: |
+          echo "JAVA_HOME: $JAVA_HOME"
+          cd $GITHUB_WORKSPACE/
+          ./dev/package.sh
+          cp package/target/thirdparty-lib/gluten-thirdparty-lib-*.jar /root/
```
philo-he: Maybe, we can directly copy the jar to package/target/lib/ if we only use `./deve/package.sh  --spark_version=3.5` to build gluten.

kevinw66:
@PHILO-HE All fixed, thanks!

philo-he:
If no other comment from the community, I will merge the pr today. cc @zhouyuan

zhouyuan (approved):
Thanks.

PR #9957, opened 2025-06-12
[GLUTEN-9949][VL] Enable enhanced features compile

jinchengchenghh:
Use the temporary branch now, replace the branch when the enhanced branch is ready

### DIFF
--- a/.github/workflows/velox_backend_cache.yml
+++ b/.github/workflows/velox_backend_cache.yml
@@ -125,7 +125,7 @@ jobs:
         run: |
           df -a
           source /opt/rh/gcc-toolset-11/enable
-          bash dev/buildbundle-veloxbe.sh --run_setup_script=OFF --build_arrow=OFF --spark_version=3.4 --enable_enhanced_feature=ON
+          bash dev/buildbundle-veloxbe.sh --run_setup_script=OFF --build_arrow=OFF --spark_version=3.4 --enable_enhanced_features=ON
           ccache -s
       - name: Save Ccache
         uses: actions/cache/save@v3
--- a/.github/workflows/velox_backend_cache.yml
+++ b/.github/workflows/velox_backend_cache.yml
@@ -106,6 +106,34 @@ jobs:
           path: '${{ env.CCACHE_DIR }}'
           key: ccache-centos8-release-shared-${{runner.arch}}-${{github.sha}}
 
+  cache-enhanced-shared-lib-centos-8:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-22.04 ]
+    container: apache/gluten:centos-8-jdk8
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get Ccache
+        uses: actions/cache/restore@v3
+        with:
+          path: '${{ env.CCACHE_DIR }}'
+          key: ccache-enhanced-centos8-release-shared-${{runner.arch}}-${{github.sha}}
+          restore-keys: |
+            ccache-enhanced-centos8-release-shared-${{runner.arch}}
+      - name: Build Gluten
+        run: |
+          df -a
+          source /opt/rh/gcc-toolset-11/enable
+          bash dev/buildbundle-veloxbe.sh --run_setup_script=OFF --build_arrow=OFF --spark_version=3.4 --enable_enhanced_feature=ON
+          ccache -s
+      - name: Save Ccache
+        uses: actions/cache/save@v3
+        id: ccache
+        with:
+          path: '${{ env.CCACHE_DIR }}'
+          key: ccache-enhanced-centos8-release-shared-${{runner.arch}}-${{github.sha}}
+
   cache-shared-lib-centos-9:
     runs-on: ${{ matrix.os }}
     strategy:
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -40,6 +40,7 @@ ENABLE_HDFS=OFF
 ENABLE_ABFS=OFF
 ENABLE_VCPKG=OFF
 ENABLE_GPU=OFF
+ENABLE_ENHANCED_FEATURES=OFF
 RUN_SETUP_SCRIPT=ON
 VELOX_REPO=""
 VELOX_BRANCH=""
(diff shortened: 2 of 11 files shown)


dev/builddeps-veloxbe.sh
```diff
@@ -121,6 +122,10 @@ do
         ENABLE_GPU=("${arg#*=}")
         shift # Remove argument name from processing
         ;;
+        --enable_enhanced_feature=*)
```
philo-he: Name suggestion: `xxx_features`, to imply multiple features are being covered or will be covered. Ditto for other places.

PHILO-HE (commented):
Just one very minor comment on the naming. Thanks!

gluten-core/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -117,6 +117,9 @@ class GlutenConfig(conf: SQLConf) extends Logging {
   def enableExtendedColumnPruning: Boolean =
     getConf(ENABLE_EXTENDED_COLUMN_PRUNING)
 
+  def enableEnhancedFeature(): Boolean =
+    System.getenv().getOrDefault("GLUTEN_ENABLE_ENHANCED_FEATURE", "false").toBoolean
+
```
zhztheplayer: Should this be defined in `GlutenConfig.scala` or in `VeloxConfig.scala`?

And it feels worth it to add an individual JNI method to avoid relying on the env?

E.g.,

```java
native boolean isEnhancedFeaturesEnabled();
```

More generally in the future, we may have

```java
native String[] getEnhancedFeatures();
```

jinchengchenghh: The VeloxConfig.scala is in backends-velox module, but gluten-iceberg only relies on gluten-substrait, gluten-iceberg should can access this config, so I place it in GlutenConfig.scala
JNI call is also ok, I will change to it.

ep/build-velox/src/get_velox.sh
```diff
@@ -20,17 +20,22 @@ VELOX_REPO=https://github.com/oap-project/velox.git
 VELOX_BRANCH=2025_06_11
 VELOX_HOME=""
 RUN_SETUP_SCRIPT=ON
+VELOX_ENHANCED_REPO=https://github.com/oap-project/velox.git
+VELOX_ENHANCED_BRANCH=2025_06_10
```
zhouyuan, line 24: since new CI is added for this feature and sometimes the gluten:main code is tightly coupled with velox rebase branch. so on each rebase we may need to take care of these two branches

.github/workflows/velox_backend_cache.yml
```diff
@@ -106,6 +106,34 @@ jobs:
...
+      matrix:
+        os: [ ubuntu-22.04 ]
+    container: apache/gluten:centos-8-jdk8
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get Ccache
+        uses: actions/cache/restore@v3
+        with:
+          path: '${{ env.CCACHE_DIR }}'
+          key: ccache-enhanced-centos8-release-shared-${{runner.arch}}-${{github.sha}}
```
zhouyuan, line 121: could we reuse the centos8-shared cache? by default the cache size for each project is relatively small(10G)

zhouyuan (commented):
Thanks!

gluten-core/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
zhztheplayer: @jinchengchenghh Got it. I think we need a backend API for the feature then. E.g., `BackendSettingsApi#isEnhancedFeatureEnabled()`

For CH backend, we can make it false by default.

Then we can move `ConfigJniWrapper.java` which I see is newly added to `gluten-core` to `backends-velox`. This is to follow the rule that we don't put any JNI code to `gluten-core` so it's more flexible for backends.

Yohahaha:
what is `enhanced features`? could you add a link or enrich the PR description?

Yohahaha:
oh I see the `enhanced` means `Some Iceberg features`, why just clarify the compile options with iceberg?

and I wonder the details of `Some Iceberg features`, could you explain more about that?

jinchengchenghh:
We will support to write iceberg partition table with statistics, read equality delete,  other features not only iceberg or not essential fix may also be merged to this enhanced-features branch, (essential fix will be merged to daily updated oap-velox branch), once Meta mainline velox merge the PR, we will remove it from this branch. @Yohahaha

ep/build-velox/src/get_velox.sh
FelixYBW: Can we reuse the VELOX_REPO and make it configurable, instead of adding the new VELOX_ENHANCED_REPO?

jinchengchenghh: Yes, I place it here because it is easy to update daily and understand. If we move it to the logic, it might be hard for user to find the repo and branch. Now it is configurable, user can set --velox_repo and --velox_branch to specify the enhanced feature branch. Do I need to update?

jinchengchenghh:
Could you help review agin? All the comments resolved.

ep/build-velox/src/get_velox.sh
philo-he: @jinchengchenghh, will the current setting for VELOX_ENHANCED_REPO & VELOX_ENHANCED_BRANCH be merged to the code base? Will they soon be substituted with genuine enhanced repo/branch?

jinchengchenghh: The ibm/velox repo exists, we prepare the ci now, it  will be ready soon.

PHILO-HE (approved):

FelixYBW:
@Yohahaha

The enhanced repo is going to solve the issue that some features can't be merged into upstream velox for long time. So we will pick the PR into ibm/velox:main branch, pass the Velox and Gluten CI, rebase the PR for every upstream PR merged. So the Gluten customer can use the feature if they want. It's just for the big features, not every PRs.

If one want their PR to be picked as well, they need to take the responsibility to fix the PR conflict every day. 

@zhli1142015 in case you didn't noted it. We are still trying to find some solution to the Velox governor issue.

Yohahaha:
> ,

thank you binwei! I have realized the whole context from your message in wecaht group.

Issue #9962, opened 2025-06-13
[HUDI] Meta field validation is unnecessary


alexr17:
### Backend

VL (Velox)

### Bug description

Hudi meta fields can be supported, the validation in HudiScanTransformer is unnecessary. This causes gluten to fallback when running `SELECT * FROM ...`

Patch is very simple.

### Gluten version

main branch

### Spark version

Spark-3.5.x

### Relevant logs

```bash

```

PR #9963, opened 2025-06-13
[GLUTEN-9962][HUDI] Remove unnecessary meta field validation

alexr17:

Fixes this issue: https://github.com/apache/incubator-gluten/issues/9962

Which causes fallback whenever running `SELECT * FROM` on hudi tables.

## How was this patch tested?

Manual testing


FelixYBW (approved):

FelixYBW:
We still don't have Hudi UT enabled in Gluten yet. I assumes the test passes

alexr17:
> I assumes the test passes

I have verified locally but I will add a unit test

jinchengchenghh:
I suppose the native engine does not support to query metadata column, what if you query metadata column with this PR?

rmahindra123:
> I suppose the native engine does not support to query metadata column, what if you query metadata column with this PR?

Why do you say that? Hoodie metadata column is like any other parquet column

Issue #9328, opened 2025-04-15
[VL][Doc] GlutenSQLQueryTestSuite aborted in gen-function-support-docs.py


dcoliversun:
### Backend

VL (Velox)

### Bug description

Expected behavior:  
Running `python3 tools/scripts/gen-function-support-docs.py --spark_home=/opt/ --categories=scalar` command can execute the entire scalar suite.

Actual behavior:  
Only GlutenSQLQueryTestSuite aborted.     
Re-running with `mvn clean test -P spark-3.5,spark-ut,backends-velox -DargLine="-Dspark.test.home=/opt/" -DwildcardSuites=org.apache.spark.sql.GlutenSQLQueryTestSuite -Dtest=none -Dsurefire.failIfNoSpecifiedTests=false -DtagsToInclude=org.apache.spark.tags.ExtendedSQLTest` can get same result.  

```
[INFO] --- scalatest-maven-plugin:2.2.0:test (test) @ gluten-ut-spark35 ---
[INFO] ScalaTest report directory: /root/incubator-gluten/gluten-ut/spark35/target/surefire-reports
Discovery starting.
Discovery completed in 5 seconds, 153 milliseconds.
Run starting. Expected test count is: 789
GlutenSQLQueryTestSuite:
org.apache.spark.sql.GlutenSQLQueryTestSuite *** ABORTED ***
  java.lang.RuntimeException: Unable to load a Suite class org.apache.spark.sql.GlutenSQLQueryTestSuite that was discovered in the runpath: null
  at org.scalatest.tools.DiscoverySuite$.getSuiteInstance(DiscoverySuite.scala:80)
  at org.scalatest.tools.DiscoverySuite.$anonfun$nestedSuites$1(DiscoverySuite.scala:38)
  at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
  at scala.collection.Iterator.foreach(Iterator.scala:943)
  at scala.collection.Iterator.foreach$(Iterator.scala:943)
  at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
  at scala.collection.IterableLike.foreach(IterableLike.scala:74)
  at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
  at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
  at scala.collection.TraversableLike.map(TraversableLike.scala:286)
  ...
  Cause: java.lang.NullPointerException:
[...]

PR #9499, opened 2025-05-02
[GLUTEN-9219][VL] Weekly check for spark function support

zhouyuan:
## What changes were proposed in this pull request?

adding weekly job to check the spark function support coverage
fix: #9219

## How was this patch tested?

pass GHA

PR #9969, opened 2025-06-13
[GLUTEN-9328][VL] Update GlutenSQLQueryTestSuite error handling for missing resource files and doc updates

dmsuehir:

This PR:
* Clarifies the documentation regarding how to set the `--spark_home` arg when running the `gen-function-support-docs.py` script to explain that it can either be Spark built from source or an installed Spark release with additional resource files copied to the install directory
* Similarly, clarifies the documentation for running `spark-ut` unit tests to mention the same for the `spark.test.home` arg.
* Updates the `GlutenSQLQueryTestSuite`, where the original error was reported to check for the path existing before the code trying to get files from the directory. This is what was causing the null pointer exception previously. With the updated error handling, running the test run will still fail if the resource files are missing, but the message will be more clear and include the path to the missing directory.
* After running `gen-function-support-docs.py` to check my updates, there was one update to the Gluten support docs, so that is included in this PR.

(Fixes: \#9328)

## How was this patch tested?

Ran `gen-function-support-docs.py` and the `spark-ut` tests.


okhleif-IL (approved):
Verified this works for me with 3.5. Thanks

docs/developers/HowTo.md
```diff
@@ -125,7 +125,27 @@ To run a Scala/Java test class, you can use the below mvn command (take Velox ba
 mvn test -Pspark-3.5 -Pspark-ut -Pbackends-velox -DargLine="-Dspark.test.home=/path/to/spark/source/code/home/" -DwildcardSuites=xxx
 ```
 
-Please set `wildcardSuites` with a fully qualified class name. `spark.test.home` is optional to set. It is only required for some test suites to use Spark resources.
+Please set `wildcardSuites` with a fully qualified class name.
+
```
dcoliversun: I think this is unnecessary to split one line into three lines. Please keep it as it is.

dmsuehir: Thank you for the review, I pushed an update.

dcoliversun (approved):
Thanks for you fix! LGTM.  

cc @marin-ma Feel free to share any additional feedback or suggestions.

PHILO-HE (commented):
@dmsuehir, thanks for your pr! I am wondering whether it would be better to just let developer run `install_spark_resources.sh`. See how it is used below: https://github.com/apache/incubator-gluten/blob/2d44c9b28b306f7bc9bd4c357f16d6d82a9f7d7f/dev/docker/Dockerfile.centos8-dynamic-build#L45

It downloads the Spark binary and Spark source, and copies resources from Spark source into Spark binary directory. This should be the most convenient way that avoids building Spark from source and also considers scala version.

dmsuehir:
@PHILO-HE Thanks, I've tried that out and ran into a couple of issues:
* The install directory for `install_spark_resources.sh` is hardcoded to `/opt`. When trying to run this without sudo, I get a permission denied error
* After running `install_spark_resources.sh`, the directory has the Spark test resource files, but it does not have the Spark binaries. This works for running unit tests, but without binaries the doc generation script fails. You can see from this [PR](https://github.com/apache/incubator-gluten/pull/9499/files#diff-cc9b23a1706d7aac3ffaa94c96692d477c2cfb375b994d23289190d0f581e3c4R125), which is trying to add a GHA weekly run for the doc generation script, that it's needing to download the Spark binaries and then combine that with the test resource files from the `install_spark_resources.sh` directory.

I can mitigate these issues with a couple of changes to the `install_spark_resource.sh` script:
* Adding an optional second positional arg to the script for the spark dir (if it's not provided, it can still default to `/opt`)
* Extracting the bin/python dir from the spark .tgz file (right now the script only grabs `jars`)

PR #9984, opened 2025-06-16
[VL] Add SupportsColumnarShuffle trait

marin-ma:
`supportColumnarShuffleExec()` in Velox backend hard codes the supported ShuffleManager type. Add trait `SupportsColumnarShuffle` to allow plugin custom shuffle manager implementation.


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -468,10 +468,8 @@ object VeloxBackendSettings extends BackendSettingsApi {
 
   override def supportColumnarShuffleExec(): Boolean = {
     val conf = GlutenConfig.get
-    conf.enableColumnarShuffle && (conf.isUseGlutenShuffleManager
-      || conf.isUseColumnarShuffleManager
-      || conf.isUseCelebornShuffleManager
-      || conf.isUseUniffleShuffleManager)
```
zhztheplayer, line 474: If these methods remain unused, may we move them to CHConfig?

Indeed it feels helpful too if we can finally remove the usages from CH backend as well.

PR #9760, opened 2025-05-27
[VL] Add mirror for ftp.gnu.org

zml1206:
## What changes were proposed in this pull request?

Since the normal ftp.gnu.org site was not accessible for all users to acquire several GNU libraries, such as Huawei Cloud overseas machines.
I added mirror https://www.mirrorservice.org/sites to some GNU libraries to fix this issue.
Reference https://github.com/microsoft/vcpkg/pull/9123

PR #9992, opened 2025-06-17
[VL] Fix wget automake and gcc

zml1206:

Follow up #9760, fix conflict with `set -e` and `$?`.


PHILO-HE (commented):
Thanks for the fix! Can we use `wget url1 || wget url2` for simplicity?

PHILO-HE (approved):
Thanks!

PR #10001, opened 2025-06-18
[CORE] Minor refactors on ValidatablePlan

zhztheplayer:
1. Move `ValidatablePlan` class to an individual source file
Reason: It's previously placed in `WholeStageTransformer.scala` which is wrong because `ValidatablePlan` is used more broadly than `WholeStageTransformer`.  
3. Move `ValidationResult` to the same package of `ValidatablePlan`
Reason: It's strongly related to `ValidatablePlan`.


zhztheplayer:
Thanks @marin-ma

PR #10005, opened 2025-06-18
[VL] Fix FallbackTags for Delta ops

zhli1142015:

When a Delta scan falls back, the reported reason is incorrect—please see below.
```
== Fallback Summary ==
(1) Scan parquet XX: Unsupported file format UnknownFormat.; Unsupported file format UnknownFormat.

```
## How was this patch tested?

UT.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
@@ -89,6 +89,8 @@ object RasOffload {
           case Validator.Passed =>
           case Validator.Failed(reason) =>
             if (FallbackTags.maybeOffloadable(node)) {
+              // FIXME: This is a solution that is not completely verified for all cases, however
+              // it's no harm anyway so we temporarily apply this practice.
               FallbackTags.add(node, reason)
             }
             return List.empty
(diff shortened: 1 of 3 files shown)


zhli1142015:
cc @zhztheplayer , thanks.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
```diff
@@ -131,6 +134,9 @@ object RasOffload {
                   // TODO: Tag the original plan with fallback reason. This is a non-trivial work
                   //  in RAS as the query plan we got here may be a copy so may not propagate tags
                   //  to original plan.
+                  if (FallbackTags.maybeOffloadable(from)) {
+                    outComes.foreach(FallbackTags.add(from, _))
```
zhztheplayer, line 142: I thought this practice may not work correctly for RAS for the reason commented above. Have you tested? Could delay on supporting RAS if not.

zhztheplayer (commented):
LGTM % a minor. Thank you.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/enumerated/RasOffload.scala
zhztheplayer: Ahh, I see. the test has covered RAS. Would you add a comment here:

```
// FIXME: This is a solution that is not completely verified for all cases, however it's no harm anyway so we
// temporarily apply this practice.
```

zhztheplayer (approved):

PR #10012, opened 2025-06-19
[VL] Use Velox config in plan converter

Yohahaha:
Use velox conf class in plan converter and plan validator, it has more convenience method than std map.


Yohahaha:
@wForget @liujiayi771 please help take a look, thank you!

Issue #9525, opened 2025-05-06
[VL] [Flink] Support tps metric for nexmark


shuai-xu:
### Description

Nexmark runner need to get tps metric from flink to calculate throughput. Gluten should support it.

### Gluten version

None

PR #10023, opened 2025-06-20
[GLUTEN-9525][FLINK]Support tps metric for source task

KevinyhZou:

(Fixes: \#9525)

## How was this patch tested?

### DIFF
--- a/gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/metrics/SourceTaskMetrics.java
+++ b/gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/metrics/SourceTaskMetrics.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


KevinyhZou:
Testing on both `Kafka` and `Nexmark` source is feasible, please take a review at this pr.  @shuai-xu

shuai-xu:
How does flink update the counter?

KevinyhZou:
> How does flink update the counter?

The `numRecodsOut` & `numBytesOut` counter have already registered in flink, so we just get them from the registry, and update the value of them by velox metrics, and flink will get the updated value.

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/metrics/SourceTaskMetrics.java
```diff
@@ -0,0 +1,69 @@
...
+ * limitations under the License.
+ */
+package org.apache.gluten.table.runtime.metrics;
+
+import io.github.zhztheplayer.velox4j.query.SerialTask;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+public class SourceTaskMetrics {
```
shuai-xu, line 27: Does flink have some class that could be used for source metrics?

KevinyhZou: Yes, such as `KafkaSourceReaderMetrics`. But it is a single class that can not be reused here, so we create this class which is smilar to flink's.

PR #10030, opened 2025-06-23
Fix missing logical link issue

li-boxuan:

This is similar to https://github.com/apache/incubator-gluten/pull/9716. Without the two fixes in this PR, TPCDS q6 (v2.7) sees missing logical links in the final plan, when RAS is enabled (Note that the same issue doesn't happen for non-RAS).

## How was this patch tested?

Manually tested in our private test suite. Writing a unit test to reproduce the issue is very hard.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/ProjectColumnPruning.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/ProjectColumnPruning.scala
@@ -34,6 +34,7 @@ object ProjectColumnPruning extends RewriteSingleNode {
 
       if (unusedAttribute.nonEmpty) {
         val newProject = project.copy(projectList = project.projectList.diff(unusedAttribute.toSeq))
+        newProject.copyTagsFrom(project)
         parent.withNewChildren(Seq(newProject))
       } else {
         parent
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
@@ -165,6 +165,9 @@ object PullOutPreProject extends RewriteSingleNode with PullOutProjectHelper {
       val preProject = ProjectExec(
         eliminateProjectList(agg.child.outputSet, expressionMap.values.toSeq),
         agg.child)
+      if (agg.child.logicalLink.isDefined) {
+        preProject.setLogicalLink(agg.child.logicalLink.get)
+      }
       newAgg.withNewChildren(Seq(preProject))
 
     case window: WindowExec if needsPreProject(window) =>


gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/ProjectColumnPruning.scala
```diff
@@ -34,6 +34,7 @@ object ProjectColumnPruning extends RewriteSingleNode {
 
       if (unusedAttribute.nonEmpty) {
         val newProject = project.copy(projectList = project.projectList.diff(unusedAttribute.toSeq))
+        newProject.copyTagsFrom(project)
```
li-boxuan, line 37: Tbh I don't know why we need this (I am pretty sure we do). I would have thought `project.copy` to handle the tags copying, but it didn't.

li-boxuan:
Also, I am pretty sure there are a few other places where we forget to copy tags and/or set logical links, but these don't cause any test to fail, so not sure if I shall fix them here too....

kapilks:
> Also, I am pretty sure there are a few other places where we forget to copy tags and/or set logical links, but these don't cause any test to fail, so not sure if I shall fix them here too....

We fixed few places internally, not sure if anymore left. There is a UT in Spark which checks this that does pass in our internal build

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/ProjectColumnPruning.scala
kapilks: `.copy` doesn't copy tags as it's auto generated by scala.
There is `makeCopy` method in Spark plan which handles tags copying but it is not used frequently

li-boxuan: I thought so but then I don't know why Scala doesn't copy tags. Anyways, good to know, thanks Kapil.

zhztheplayer: Scala-generated `#copy` only handles the ~~mutable~~ immutable fields of a case class, while tags are [mutable](https://github.com/apache/spark/blob/5e1e4d31d7d9fb2f93ce8c72d32774b560bcd81f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala#L79-L91) in Spark's code so it's not included in copying.

zhztheplayer:
> Also, I am pretty sure there are a few other places where we forget to copy tags and/or set logical links,

Yes I agreed. 

> but these don't cause any test to fail

This may happen, because I think tags are not widely relied by critical code paths either in Spark and Gluten. I can tell Spark CBO may need the logical links for cost evaluation however CBO is by default off. And in Gluten we have the fallback tags but the use of them are only scoped in one rules or two.

zhztheplayer (approved):
Thanks!

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/ProjectColumnPruning.scala
li-boxuan: Do you mean it only handles immutable fields?

zhztheplayer: Ahh, yes, sorry for the typo.

li-boxuan:
@zhztheplayer could you please review again? I just fixed more logical link missing issues that I could find.

li-boxuan:
Hmmm I got some test failure:

> Gluten Unit Test Spark35 ........................... FAILURE [53:25 min]

but I cannot find any concrete test failure in the log

zhztheplayer:
> Hmmm I got some test failure:
> 
> > Gluten Unit Test Spark35 ........................... FAILURE [53:25 min]
> 
> but I cannot find any concrete test failure in the log

It should be unrelated as rerunning fixed it. We have some flaky cases in UT.

zhztheplayer:
@li-boxuan Would you like to rebase? Thanks!

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
```diff
@@ -165,6 +180,9 @@ object PullOutPreProject extends RewriteSingleNode with PullOutProjectHelper {
       val preProject = ProjectExec(
         eliminateProjectList(agg.child.outputSet, expressionMap.values.toSeq),
         agg.child)
+      if (agg.child.logicalLink.isDefined) {
+        preProject.setLogicalLink(agg.child.logicalLink.get)
+      }
```
li-boxuan:
```suggestion
      agg.child.logicalLink.foreach(preProject.setLogicalLink)
```

li-boxuan: and many other places

zhztheplayer (approved):

Issue #10035, opened 2025-06-23
[VL] Duplicate projections in HashProbe when projecting a column multiple times without an alias


NEUpanning:
### Backend

VL (Velox)

### Bug description

Reproducing unit test:
```scala
    withTable("t1", "t2", "t3") {
      Seq((1, 1), (2, 2)).toDF("c1", "c2").write.saveAsTable("t1")
      Seq(1, 2, 3).toDF("c1").write.saveAsTable("t2")

      val q5 =
        """
          |select tt1.* from
          |(select c1, c2, c2 from t1) tt1
          |left join t2
          |on tt1.c1 = t2.c1
          |""".stripMargin

    runQueryAndCompare(q5) { _ => }
  }
```

Error message:
```
Caused by: org.apache.gluten.exception.GlutenException: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: An unloaded lazy vector cannot be wrapped by two different top level vectors.
Retriable: False
Expression: dictionaryValues_->markAsContainingLazyAndWrapped()
Context: Operator: HashProbe[3] 2
Function: setInternalState
/opt/meituan/panning/gluten-dev/velox/./velox/vector/DictionaryVector-inl.h
Line: 35
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_17VeloxRuntimeErrorEPKcEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox16DictionaryVectorIiE16setInternalStateEv
# 4  _ZN8facebook5velox16DictionaryVectorIiEC2EPNS0_6memory10MemoryPoolEN5boost13intrusive_ptrINS0_6BufferEEEmSt10shared_ptrINS0_10BaseVectorEES9_RKNS0_17SimpleVectorStatsIiEESt8optionalIiESI_SH_IbESI_SI_
# 5  _ZN8facebook5velox10BaseVector16wrapInDictionaryEN5boost13intrusive_ptrINS0_6BufferEEES5_iSt10shared_ptrIS1_Eb
[...]

PR #10036, opened 2025-06-23
[GLUTEN-10035][VL] Fix PullOutDuplicateProject rule when projecting a column multiple times without an alias

NEUpanning:

Fixes: \#10035

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxHashJoinSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxHashJoinSuite.scala
@@ -295,6 +295,7 @@ class VeloxHashJoinSuite extends VeloxWholeStageTransformerSuite {
           }
       }
 
+      // Test HashProbe operation when projecting a column multiple times without using an alias.
       val q5 =
         """
           |select tt1.* from


NEUpanning:
@rui-mo @zml1206 Could you help take a look? Thanks!

backends-velox/src/main/scala/org/apache/gluten/extension/PullOutDuplicateProject.scala
```diff
@@ -60,13 +60,13 @@ object PullOutDuplicateProject extends Rule[SparkPlan] with PredicateHelper {
         bhj.streamedPlan.asInstanceOf[ProjectExecTransformer],
         bhj.references,
         pullOutAliases)
+      val newBhj = bhj.joinBuildSide match {
+        case BuildLeft => bhj.copy(right = newStreamedPlan)
+        case BuildRight => bhj.copy(left = newStreamedPlan)
+      }
       if (pullOutAliases.isEmpty) {
```
zml1206: Thanks for the fix, at present, pullOutAliases cannot determine whether rewriting is needed. We can separate the calculation of duplicates from rewriteProject and use duplicates to determine whether rewriting is needed.

backends-velox/src/main/scala/org/apache/gluten/extension/PullOutDuplicateProject.scala
```diff
@@ -32,37 +32,47 @@ import scala.collection.mutable.ArrayBuffer
...
+      if (duplicates.isEmpty) {
         f
       } else {
+        val pullOutAliases = new ArrayBuffer[Alias]()
+        val newChild = rewriteProject(child, f.references, pullOutAliases, duplicates)
         outerProject(f.copy(child = newChild), f.output, pullOutAliases)
       }
     case bhj: BroadcastHashJoinExecTransformer
         if bhj.streamedPlan.isInstanceOf[ProjectExecTransformer] =>
       val pullOutAliases = new ArrayBuffer[Alias]()
```
zml1206: Please move pullOutAliases down.

PR #10045, opened 2025-06-24
[VL] Move SupportsColumnarShuffle.scala to module `gluten-substrait`

zhztheplayer:
`SupportsColumnarShuffle.scala` is strongly related to `ColumnarShuffleExchangeExec` which is placed in `gluten-substrait`. The patch moves the code to the module.


gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -141,7 +141,7 @@ class GlutenConfig(conf: SQLConf) extends GlutenCoreConfig(conf) {
 
   // scalastyle:off classforname
   def shuffleManagerSupportsColumnarShuffle: Boolean = {
-    classOf[SupportsColumnarShuffle].isAssignableFrom(Class.forName(SQLConf.get
+    classOf[SupportsColumnarShuffle].isAssignableFrom(Class.forName(conf
```
zhztheplayer, line 144: a minor nit fix

PR #10051, opened 2025-06-24
[CORE] Shim: Remove unused Spark33Scan / Spark34Scan / Spark35Scan

zhztheplayer:
I am not sure about the reason these scans were added in the first place but it appears they are not in use. Attempting to remove them from the shim layer.


zhztheplayer:
cc @baibaichen @JkSelf

JkSelf (approved):

Issue #10056, opened 2025-06-25
[FLINK]Support function `Split_index`


KevinyhZou:
### Description

while query nexmark q22, we found split_index function not supported.

### Gluten version

None

PR #10057, opened 2025-06-25
[GLUTEN-10056][FLINK] Support function split_index

KevinyhZou:

(Fixes: \#10056)

depends on pr: https://github.com/bigo-sg/velox/pull/8

## How was this patch tested?
UT

### DIFF
--- a/.github/workflows/flink.yml
+++ b/.github/workflows/flink.yml
@@ -59,7 +59,7 @@ jobs:
           source /opt/rh/gcc-toolset-11/enable
           sudo dnf install -y patchelf
           git clone -b gluten-0530 https://github.com/bigo-sg/velox4j.git
-          cd velox4j && git reset --hard ea2ca5755ae91a8703717a85b77f9eb1620899de
+          cd velox4j && git reset --hard 0180528e9b98fad22bc9da8a3864d2929ef73eec
           git apply $GITHUB_WORKSPACE/gluten-flink/patches/fix-velox4j.patch
           mvn clean install -DskipTests -Dgpg.skip -Dspotless.skip=true
           cd ..
--- a/gluten-flink/docs/Flink.md
+++ b/gluten-flink/docs/Flink.md
@@ -48,7 +48,7 @@ As some features have not been committed to upstream, you have to use the follow
 ## fetch velox4j code
 git clone -b gluten-0530 https://github.com/bigo-sg/velox4j.git
 cd velox4j
-git reset --hard ea2ca5755ae91a8703717a85b77f9eb1620899de
+git reset --hard 0180528e9b98fad22bc9da8a3864d2929ef73eec
 mvn clean install -DskipTests -Dgpg.skip -Dspotless.skip=true
 ```
 **Get gluten**


gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/SplitIndexRexCallConverter.java
```diff
@@ -0,0 +1,85 @@
...
+    if (!TypeUtils.isStringType(RexNodeConverter.toType(firstOp.getType()))
+        || !TypeUtils.isStringType(RexNodeConverter.toType(secondOp.getType()))
+        || !TypeUtils.isIntegerType(RexNodeConverter.toType(thirdOp.getType()))) {
+      return false;
+    }
+    if (!(secondOp instanceof RexLiteral) || !(thirdOp instanceof RexLiteral)) {
+      return false;
+    }
+    return true;
+  }
```
lgbo-ustc: I think we could return true here without check, since it's the only convertor for `split_index`. The velox backend will do some validation check.

KevinyhZou: fixed.

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/SplitIndexRexCallConverter.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public SplitIndexRexCallConverter() {
+    super(FUNCTION_NAME);
+  }
+
+  @Override
+  public TypedExpr toTypedExpr(RexCall callNode, RexConversionContext context) {
+    List<TypedExpr> params = getParams(callNode, context);
+    ConstantTypedExpr indexExpr = (ConstantTypedExpr) params.get(params.size() - 1);
+    if (TypeUtils.isIntegerType(indexExpr.getReturnType())) {
```
shuai-xu, line 61: Why need this transform? if it's not int, need to check it valid ?

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -126,4 +126,20 @@ void testEqual() {
     String query4 = "select c = d as x from tblEqual where a > 0";
     runAndCheck(query4, Arrays.asList("+I[false]", "+I[true]", "+I[false]"));
   }
+
+  @Test
+  void testSplitIndex() {
+    List<Row> rows =
+        Arrays.asList(
```
shuai-xu, line 135: Is there any corner case need to test to verify the result same with flink?

KevinyhZou:
> Which nexmark query use this function?

nexmark q22 need this

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/SplitIndexRexCallConverter.java
KevinyhZou: The corresponding function in velox is `split_part`, and the last parameter's type is `int64_t`, but in flink `split_index` function, it is `integer` type, so we need this transform

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
KevinyhZou: we can cover this by use ut in flink `ScalaFunctionsTest#testSplitIndex`

gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/SplitIndexRexCallConverter.java
shuai-xu: please add comments for it.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
shuai-xu: seems there are many null related tests in flink not covered by this test.

shuai-xu: and others such as index is invalid.

KevinyhZou: I have copyed some related flink tests here related to the index invalid, input has nulls, or delimiter is defined by ascii, but others tests e.g. `index` parameter is null or `delimiter` parameter is null, can not run here. as we can not pass the the null constant parameter value to velox. Or we should introduce a Variant to pass `null` constant.

PR #10093, opened 2025-07-01
[VL] Fix unsupported OS(<undefined>, <undefined>) error

liujiayi771:
For configurations that are created as `createOptional` and not set, calling `get(ConfigEntry)` will return the string `<undefined>`, which may cause unintended entry into the `getForOS` logic.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoader.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoader.scala
@@ -53,9 +53,9 @@ object SharedLibraryLoader {
   }
 
   private def find(conf: SparkConf): SharedLibraryLoader = {
-    val systemName = conf.get(GlutenConfig.GLUTEN_LOAD_LIB_OS)
+    val systemName = conf.getOption(GlutenConfig.GLUTEN_LOAD_LIB_OS.key)
     val loader = if (systemName.isDefined) {
-      val systemVersion = conf.get(GlutenConfig.GLUTEN_LOAD_LIB_OS_VERSION)
+      val systemVersion = conf.getOption(GlutenConfig.GLUTEN_LOAD_LIB_OS_VERSION.key)
       if (systemVersion.isEmpty) {
         throw new GlutenException(
           s"${GlutenConfig.GLUTEN_LOAD_LIB_OS_VERSION.key} must be specified when specifies the " +


liujiayi771:
cc @yaooqinn, @zhztheplayer, @zhouyuan, @PHILO-HE, @zml1206.

liujiayi771:
Perhaps returning `None` when encountering `<undefined>` in `RichSparkConf` would be a better choice.

yaooqinn:
> Perhaps returning `None` when encountering `<undefined>` in `RichSparkConf` would be a better choice.

Thank you for catching this, +1 for this approach.

PHILO-HE (approved):
Thanks!

gluten-core/src/main/scala/org/apache/spark/sql/internal/SparkConfigUtil.scala
```diff
@@ -50,15 +50,15 @@ object SparkConfigUtil {
   }
 
   def get[T](conf: SparkConf, entry: ConfigEntry[T]): T = {
-    entry.valueConverter(conf.get(entry.key, entry.defaultValueString))
+    entry.valueConverter(checkUndefined(conf.get(entry.key, entry.defaultValueString)))
```
yaooqinn: Instead of calling `checkUndefined` every time, we can modify like this
```diff
-    entry.valueConverter(conf.get(entry.key, entry.defaultValueString))
+    conf
+      .getOption(entry.key)
+      .map(entry.valueConverter)
+      .getOrElse(entry.defaultValue.getOrElse(None).asInstanceOf[T])
```

liujiayi771: @yaooqinn Should we also apply `entry.valueConverter` to the defaultValue? So it will be
```scala
entry.valueConverter(
      conf
        .getOption(entry.key)
        .getOrElse(entry.stringConverter(entry.defaultValue.getOrElse(None).asInstanceOf[T])))
```
or
```scala
entry.valueConverter(
      conf
        .getOption(entry.key)
        .getOrElse(checkUndefined(entry.defaultValueString)))
```
And for `conf: java.util.Map[String, String]`， it will be
```scala
entry.valueConverter(
      Option(conf.get(entry.key))
        .getOrElse(checkUndefined(entry.defaultValueString)))
```
What do you think? I think calling `checkUndefined` is more simple, and it won't be invoked unnecessarily every time.

yaooqinn: What `valueConverter + checkUndefined + defaultValueString` produces is `defaultValue`. So, how about we use it directly?

liujiayi771: @yaooqinn Thank you for your explanation, it has been revised.

Issue #7628, opened 2024-10-22
[VL] Result mismatch on monotonically_increasing_id


zhztheplayer:
### Backend

VL (Velox)

### Bug description

Error:
```
GlutenColumnExpressionSuite.monotonically_increasing_id

Results do not match for query:
Timezone: sun.util.calendar.ZoneInfo[id="America/Los_Angeles",offset=-28800000,dstSavings=3600000,useDaylight=true,transitions=185,lastRule=java.util.SimpleTimeZone[id=America/Los_Angeles,offset=-28800000,dstSavings=3600000,useDaylight=true,startYear=0,startMode=3,startMonth=2,startDay=8,startDayOfWeek=1,startTime=7200000,startTimeMode=0,endMode=3,endMonth=10,endDay=1,endDayOfWeek=1,endTime=7200000,endTimeMode=0]]
Timezone Env: 

== Parsed Logical Plan ==
'Project [unresolvedalias(monotonically_increasing_id(), Some(org.apache.spark.sql.Column$$Lambda$7598/1906035518@69d7dddf)), unresolvedalias('monotonically_increasing_id(), Some(org.apache.spark.sql.Column$$Lambda$7598/1906035518@69d7dddf))]
+- Project [_1#605907 AS a#605910]
   +- SerializeFromObject [knownnotnull(assertnotnull(input[0, scala.Tuple1, true]))._1 AS _1#605907]
      +- ExternalRDD [obj#605906]

== Analyzed Logical Plan ==
monotonically_increasing_id(): bigint, monotonically_increasing_id(): bigint
Project [monotonically_increasing_id() AS monotonically_increasing_id()#605912L, monotonically_increasing_id() AS monotonically_increasing_id()#605913L]
+- Project [_1#605907 AS a#605910]
   +- SerializeFromObject [knownnotnull(assertnotnull(input[0, scala.Tuple1, true]))._1 AS _1#605907]
      +- ExternalRDD [obj#605906]

== Optimized Logical Plan ==
Project [monotonically_increasing_id() AS monotonically_increasing_id()#605912L, monotonically_increasing_id() AS monotonically_increasing_id()#605913L]
+- SerializeFromObject
   +- ExternalRDD [obj#605906]

== Physical Plan ==
VeloxColumnarToRow
+- ^(18972) ProjectExecTransformer [monotonically_increasing_id() AS monotonically_increasing_id()#605912L, monotonically_increasing_id() AS monotonically_increasing_id()#605913L]
[...]

PR #10097, opened 2025-07-01
[VL] Support monotonically_increasing_id function

liujiayi771:

Support monotonically_increasing_id function in Velox backend. Velox had implemented it in https://github.com/facebookincubator/velox/pull/9007 early on.

## How was this patch tested?

Add a test case in `ScalarFunctionsValidateSuite`.

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
@@ -588,7 +588,6 @@ class CHSparkPlanExecApi extends SparkPlanExecApi with Logging {
     List(
       Sig[CollectList](ExpressionNames.COLLECT_LIST),
       Sig[CollectSet](ExpressionNames.COLLECT_SET),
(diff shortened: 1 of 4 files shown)


liujiayi771:
@zhli1142015 @zhztheplayer Could you help to review?

zhli1142015:
I remembre that this function was disabled because of the issue below.
https://github.com/apache/incubator-gluten/issues/7628 
The discrepancy in results between Velox and Spark was due to the compiled expression being cached, resulting in a single expression instance whose ID kept incrementing.
https://github.com/facebookincubator/velox/blob/main/velox/expression/ExprCompiler.cpp#L376C29-L376C47
I thought we may need a fix in velox to allow marking this expression as “do not cache.”

liujiayi771:
@zhli1142015 OK. I'll close this PR. But isn't this function itself non-deterministic anyway? As long as it generates an increasing non-repeating sequence, it shouldn't cause issues in practice—it just won't be perfectly consistent with Spark's results, right?

liujiayi771:
@zhli1142015 But the issue in #7628 is that calling `monotonically_increasing_id` twice in a single row is expected to yield the same result.

zhli1142015:
> it shouldn't cause issues in practice?

It seems to me that this unit test verifies a non publicly documented behavior; but we re-enabled this function after we had fixed this issue internally.

PR #10111, opened 2025-07-03
[GLUTEN-10112][VL] Fix deadLock caused by objectStore when native writer throws oom exception

zjuwangg:

Fix dead lock founded in #10112

(Fixes: \#10112)

### Problem description 
The task thread on the Executor is blocked by a single runnable task thread. The runnable task thread continues to run in nativeSpill without exiting.
Image

### Cause Analysis

1. Observing the runnable task thread stack, it was found that the task is already in the resource release phase, and theoretically should not trigger spill operations again.

2. Checking the log of the runnable task and observing the task's operation logs, it was found that during the native write phase, insufficient OOM (Out Of Memory) triggered the task's failure cleanup action.

3. Why does the nativeSpill operation get triggered again during the resource release phase? 

 Additional logs were added to record the native stack of the last reserveMemory operation

image

4. Why does triggering the nativeSpill method again during the RuntimeJniWrapper resource release phase lead to a stuck (deadlock) issue?

The core reason is that all operations on the ObjectStore data structure are protected by a mutual exclusion lock.

First, looking at the destructor method, a lock is acquired at the start of the method. When destructing VeloxParquetDatasource, the acquireMemory operation is triggered, which further triggers the spill operation.

image

nativeSpill method:

image

The NativeSpill method then performs a retrieve iterator operation in the objectStore. The retrieve operation waits for the lock acquired by the `~ObjectStore` destructor, leading to a deadlock and the task not exiting.

image

### how this modification would work?
Understanding the root cause, the fix approach is relatively straightforward, with several options:

Avoid acquiring additional memory during the destruction of VeloxParquetDatasource.
Change the locks for ObjectStore operations to reentrant locks.
Do not acquire locks when releasing objects held by ObjectStore.

After comprehensive evaluation, method 3 was chosen for the fix to minimize the impact of modifications on overall stability.

## How was this patch tested?
Tested in our production environment

### DIFF
--- a/cpp/core/utils/ObjectStore.cc
+++ b/cpp/core/utils/ObjectStore.cc
@@ -19,7 +19,7 @@
 #include <glog/logging.h>
 #include <iostream>
 
-//static
+// static
 std::unique_ptr<gluten::ObjectStore> gluten::ObjectStore::create() {
   static std::mutex mtx;
   std::lock_guard<std::mutex> lock(mtx);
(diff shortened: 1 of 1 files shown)


zjuwangg:
@zhztheplayer plz help review this when you have time

cpp/core/utils/ObjectStore.cc
```diff
@@ -37,23 +37,31 @@ gluten::ResourceMap<gluten::ObjectStore*>& gluten::ObjectStore::stores() {
...
-    const size_t size = info.size;
-    VLOG(2) << "Unclosed object ["
-            << "Store ID: " << storeId_ << ", Resource handle ID: " << handle << ", TypeName: " << typeName
-            << ", Size: " << size
-            << "] is found when object store is closing. Gluten will"
-               " destroy it automatically but it's recommended to manually close"
-               " the object through the Java closing API after use,"
-               " to minimize peak memory pressure of the application.";
-    store_.erase(handle);
+  std::vector<std::shared_ptr<void>> objects;
```
zhztheplayer: I searched for usages of this vector but found nothing. Is it intended to be updated in the loop or something?

zhztheplayer:
@zjuwangg Thank you for the comprehensive analysis of the bug. It certainly makes sense.

cpp/core/utils/ObjectStore.cc
zjuwangg: Good catch, it should be updated in the following loop. And keep aliveObjects so that we can destruct related object out of mtx scope.
Will updated soon!

cpp/core/utils/ObjectStore.cc
```diff
@@ -37,23 +37,34 @@ gluten::ResourceMap<gluten::ObjectStore*>& gluten::ObjectStore::stores() {
...
+      // transfer ownership of the object to the temporary vector
+      objects.push_back(store_.lookup(handle));
+      store_.erase(handle);
+    }
+    stores().erase(storeId_);
+  }
+  // destructing objects outside the lock to avoid deadlock
+  for (auto& obj : objects) {
+    if (obj) {
+      obj.reset();
```
zhztheplayer: In the case your mentioned, `store_` may be accessed again when `obj.reset()` is called. However, IIUC, at this moment, the main object container `store_` is already emptied. Wouldn't there be a new `object not found` error being thrown, which is unexpected either?

zjuwangg: You are right but in this situation there will be no problem for ParquetFileWriter, which will ignore all exception during it's destructors as https://github.com/facebookincubator/velox/blob/main/velox/dwio/parquet/writer/arrow/FileWriter.cpp#L646.

`In C++, destructors should generally not throw exceptions that are allowed to escape the destructor. This is a crucial guideline for writing robust and reliable C++ code`.

For other obj  besides `ParquetFileWriter`, if `object not found` being thrown we can fix it's destructors method, it's better to run in endless deadLock loop.

zhztheplayer: Thank you for the explanation. I'll take a closer look tomorrow on this. You are right that at least we can avoid a dead lock by the change.

zhztheplayer: @zjuwangg Would you think the following could be feasible (pseudo)? Thanks!

```cpp
for (;;) {
  lock_.lock();
  if (aliveObjects_.empty()) {
    break;
  }
  obj o = aliveObjects_.pop()
  lock_.unlock()
  o.release();
}
```

zjuwangg: It should be okey, will address it tomorrow.

zjuwangg: Just addressed

zhztheplayer (approved):
Thanks! I assume this version doesn't crash but haven't had a chance to test it. Let me know if any issue is found afterwards.

zjuwangg:
Can anyone help merge this?

Issue #10103, opened 2025-07-02
Invalid call to dataType on unresolved object


wenfang6:
### Backend

VL (Velox)

### Bug description

sql like this has the error. Invalid call to dataType on unresolved object
example：
```
CREATE TABLE map_test_gluten (
  sessionid STRING,
  features MAP<STRING, STRING>)
USING parquet;
INSERT overwrite dap_dev.map_test_gluten VALUES 
(
  'session_001',
  map(
    '118', 'geek_position_1',
    '125', 'boss_position_1',
    'tag', 'abc'
  )
),
(
  'session_002',
  map(
    '118', 'geek_position_2',
    '125', 'boss_position_2',
    'tag', 'xyz'
  )
);
query：
 select 
/*+ REPARTITION(2) */
a.features,
a.geek_position
from 
(select features,features ['118'] as geek_position from dap_dev.map_test_gluten) a
```
error：
```
org.apache.spark.sql.catalyst.analysis.UnresolvedException: Invalid call to dataType on unresolved object
	at org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute.dataType(unresolved.scala:160)
	at org.apache.spark.sql.types.StructType$.$anonfun$fromAttributes$1(StructType.scala:548)
	at scala.collection.immutable.List.map(List.scala:293)
	at org.apache.spark.sql.types.StructType$.fromAttributes(StructType.scala:548)
	at org.apache.spark.sql.catalyst.plans.QueryPlan.schema$lzycompute(QueryPlan.scala:372)
	at org.apache.spark.sql.catalyst.plans.QueryPlan.schema(QueryPlan.scala:372)
	at org.apache.gluten.execution.ValidatablePlan.doValidate(WholeStageTransformer.scala:74)
	at org.apache.gluten.execution.ValidatablePlan.doValidate$(WholeStageTransformer.scala:72)
	at org.apache.gluten.execution.ProjectExecTransformerBase.doValidate(BasicPhysicalOperatorTransformer.scala:179)
	at org.apache.gluten.extension.columnar.CollapseProjectExecTransformer$$anonfun$apply$1.applyOrElse(CollapseProjectExecTransformer.scala:42)
	at org.apache.gluten.extension.columnar.CollapseProjectExecTransformer$$anonfun$apply$1.applyOrElse(CollapseProjectExecTransformer.scala:33)
```

### Gluten version

Gluten-1.3

### Spark version

Spark-3.5.x

### Relevant logs

```bash

[...]

PR #10138, opened 2025-07-08
[GLUTEN-10103][VL] Fall back to vanilla spark when UnresolvedException occurs in the schema validation

JkSelf:

Fallback to vanilla spark rather than raising an `AnalysisException `during the validation stage.

Fix https://github.com/apache/incubator-gluten/issues/10103

## How was this patch tested?

Existing tests

### DIFF
--- a/docs/velox-backend-limitations.md
+++ b/docs/velox-backend-limitations.md
@@ -163,3 +163,10 @@ Gluten's.
 ### CSV Read
 The header option should be true. And now we only support DatasourceV1, i.e., user should set `spark.sql.sources.useV1SourceList=csv`. User defined read option is not supported, which will make CSV read fall back to vanilla Spark in most case.
 CSV read will also fall back to vanilla Spark and log warning when user specifies schema is different with file schema.
+
+### Utilizing Map Type as Hash Keys in ColumnarShuffleExchange
+Spark uses the `spark.sql.legacy.allowHashOnMapType` configuration to support hash map key functions. 
+Gluten enables this configuration during the creation of ColumnarShuffleExchange, as shown in the code [link](https://github.com/apache/incubator-gluten/blob/0dacac84d3bf3d2759a5dd7e0735147852d2845d/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L355-L363). 
+This method bypasses Spark's unresolved checks and creates projects with the hash(mapType) operator before ColumnarShuffleExchange. 
+However, if `spark.sql.legacy.allowHashOnMapType` is disabled in a test environment, projects using the hash(mapType) expression may throw an `Invalid call to dataType on unresolved object` exception during validation, causing them to fallback to vanilla Spark, as referenced in the code [link](https://github.com/apache/spark/blob/de5fa426e23b84fc3c2bddeabcd2e1eda515abd5/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/hash.scala#L291-L296).
+ Enabling this configuration allows the project to be offloaded to Velox.
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/ValidatablePlan.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/ValidatablePlan.scala
@@ -23,7 +23,7 @@ import org.apache.gluten.expression.TransformerState
 import org.apache.gluten.logging.LogLevelUtil
 import org.apache.gluten.test.TestStats
 
-import org.apache.spark.sql.AnalysisException
+import org.apache.spark.sql.catalyst.analysis.UnresolvedException
 
 /**
  * Base interface for a Gluten query plan that is also open to validation calls.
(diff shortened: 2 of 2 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/execution/ValidatablePlan.scala
```diff
@@ -43,7 +43,7 @@ trait ValidatablePlan extends GlutenPlan with LogLevelUtil {
     try {
       f
     } catch {
-      case e @ (_: GlutenNotSupportException | _: UnsupportedOperationException) =>
+      case e @ (_: GlutenNotSupportException | _: UnsupportedOperationException | _: Throwable) =>
```
zhztheplayer: Would prefer catching `Exception` so we still allow fatal errors (e.g., a heap OOM) to crash the application.

JkSelf: Does the OOM issue arise during the validation stage?

zhztheplayer: Generally JVM heap OOM could be raised at any point of the Java / Scala code. It's just an example of fatal errors. There are more of them like illegal access errors, class compatibility errors, etc.

docs/velox-backend-limitations.md
```diff
@@ -163,3 +163,6 @@ Gluten's.
...
 CSV read will also fall back to vanilla Spark and log warning when user specifies schema is different with file schema.
+
+### Utilizing Map Type as Hash Keys in ColumnarShuffleExchange
+Spark uses the `spark.sql.legacy.allowHashOnMapType` configuration to support hash map key functions. Gluten enables this configuration during the creation of ColumnarShuffleExchange, as shown in the code [link](https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L355-L363). This method bypasses Spark's unresolved checks and creates projects with the hash(mapType) operator before ColumnarShuffleExchange. However, if `spark.sql.legacy.allowHashOnMapType` is disabled in a test environment, projects using the hash(mapType) expression may throw an `Invalid call to dataType on unresolved object` exception during validation, causing them to fallback to vanilla Spark, as referenced in the code [link](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/hash.scala#L291-L296). Enabling this configuration allows the project to be offloaded to Velox.
```
philo-he: Break this line into multiple code lines.

Please use github permanent link to avoid unexpected reference once code line is changed.

gluten-substrait/src/main/scala/org/apache/gluten/execution/ValidatablePlan.scala
philo-he: Similarly, I feel it may not be good to catch all exceptions which can hide some critical issues behind the fallback behavior.

Can we detect this exception somehow and throw GlutenNotSupportException instead?

FelixYBW: can we catch the exact exception instead of all Throwable?

docs/velox-backend-limitations.md
JkSelf: @PHILO-HE Yes. Updated. Please help to review again. Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/execution/ValidatablePlan.scala
JkSelf: @zhztheplayer @PHILO-HE Thanks for your detailed explanation. @FelixYBW Updated to only catch the `AnalysisException`. Please help to review again. Thanks.

JkSelf:
@zhztheplayer @PHILO-HE @FelixYBW Do you have any further comment? Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/execution/ValidatablePlan.scala
```diff
@@ -63,13 +65,26 @@ trait ValidatablePlan extends GlutenPlan with LogLevelUtil {
...
+        BackendsApiManager.getValidatorApiInstance
+          .doSchemaValidate(schema)
+          .map {
+            reason =>
+              ValidationResult.failed(
+                s"Found schema check failure for $schema, due to: $reason")
+          }
+          .getOrElse(ValidationResult.succeeded)
+      } catch {
+        case e: AnalysisException =>
```
philo-he: Suggest to use `UnsolvedException`, 1) it's more specific 2) To make this exception catching also works Spark 4.0.

Prior to Spark 4.0, UnsolvedException extends AnalysisException, but since Spark 4.0, it extends SparkException.

JkSelf: @PHILO-HE Yes. Updated.

PHILO-HE (approved):
Thanks!

PR #2553, opened 2023-08-01
[GLUTEN-2552][CH] Support native insert for orc format

taiyang-li:
## What changes were proposed in this pull request?

(Please fill in changes proposed in this fix)

(Fixes: \#2552)

PR #10147, opened 2025-07-09
[CORE] Refactor: Make `GlutenFormatWriterInjects#executeWriterWrappedSparkPlan` return the wrapped query plan rather than the executed RDD

zhztheplayer:
API

```scala
def executeWriterWrappedSparkPlan(plan: SparkPlan): RDD[InternalRow]
```

is being changed to

```scala
def getWriterWrappedSparkPlan(plan: SparkPlan): SparkPlan
```

We don't have to execute the SparkPlan before it's returned to the file format writer. This change will make it easier to debug or operate on the wrapped query plan. The previous use of RDD more or less limited the flexibility of coding and troubleshooting.

Related to #2205, #2553


zhztheplayer:
cc @taiyang-li, @JkSelf

Issue #10155, opened 2025-07-10
[VL] Build failed on macOS 15


zml1206:
### Problem description

There are the following errors：
1. clang 17 and arrow's built-in zlib are incompatible
```
[34/277] Performing build step for 'zlib_ep'
FAILED: [code=1] zlib_ep-prefix/src/zlib_ep-stamp/zlib_ep-build zlib_ep/src/zlib_ep-install/lib/libz.a /Users/zml/Desktop/git_hub/incubator-gluten/ep/_ep/arrow_ep/cpp/_build/zlib_ep-prefix/src/zlib_ep-stamp/zlib_ep-build /Users/zml/Desktop/git_hub/incubator-gluten/ep/_ep/arrow_ep/cpp/_build/zlib_ep/src/zlib_ep-install/lib/libz.a 
cd /Users/zml/Desktop/git_hub/incubator-gluten/ep/_ep/arrow_ep/cpp/_build/zlib_ep-prefix/src/zlib_ep-build && /opt/homebrew/bin/cmake -P /Users/zml/Desktop/git_hub/incubator-gluten/ep/_ep/arrow_ep/cpp/_build/zlib_ep-prefix/src/zlib_ep-stamp/zlib_ep-build-RELEASE.cmake && /opt/homebrew/bin/cmake -E touch /Users/zml/Desktop/git_hub/incubator-gluten/ep/_ep/arrow_ep/cpp/_build/zlib_ep-prefix/src/zlib_ep-stamp/zlib_ep-build
CMake Error at /Users/zml/Desktop/git_hub/incubator-gluten/ep/_ep/arrow_ep/cpp/_build/zlib_ep-prefix/src/zlib_ep-stamp/zlib_ep-build-RELEASE.cmake:37 (message):
  Command failed: 1
   '/opt/homebrew/bin/cmake' '--build' '.'
  See also
    /Users/zml/Desktop/git_hub/incubator-gluten/ep/_ep/arrow_ep/cpp/_build/zlib_ep-prefix/src/zlib_ep-stamp/zlib_ep-build-*.log
-- stdout output is:
...skipping to end...
/src/zlib_ep/gzguts.h:21:
In file included from /Library/Developer/CommandLineTools/SDKs/MacOSX15.4.sdk/usr/include/stdio.h:61:
/Library/Developer/CommandLineTools/SDKs/MacOSX15.4.sdk/usr/include/_stdio.h:318:7: error: expected ')'
[...]

PR #9579, opened 2025-05-09
[GLUTEN-9577][VL] Fix build Gluten on Mac with cmake 4.0 issue

wangyum:
## What changes were proposed in this pull request?

Fix build Gluten on Mac with cmake 4.0 issue.

(Fixes: \#9577)

## How was this patch tested?

manual tests

PR #10158, opened 2025-07-10
[GLUTEN-10155][INFRA] Fix build on macOS

zml1206:

(Fixes: \#10155)

## How was this patch tested?

local mac build

### DIFF
--- a/dev/build_arrow.sh
+++ b/dev/build_arrow.sh
@@ -37,6 +37,8 @@ function prepare_arrow_build() {
 function build_arrow_cpp() {
   pushd $ARROW_PREFIX/cpp
   ARROW_WITH_ZLIB=ON
+  # The zlib version bundled with arrow is not compatible with clang 17.
+  # It can be removed after upgrading the arrow version.
   if [[ "$(uname)" == "Darwin" ]]; then
     clang_major_version=$(echo | clang -dM -E - | grep __clang_major__ | awk '{print $3}')
     if [ "${clang_major_version}" -ge 17 ]; then
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -44,15 +44,10 @@ ENABLE_ENHANCED_FEATURES=OFF
 RUN_SETUP_SCRIPT=ON
 VELOX_REPO=""
 VELOX_BRANCH=""
-VELOX_HOME=""
+VELOX_HOME="$GLUTEN_DIR/ep/build-velox/build/velox_ep"
 VELOX_PARAMETER=""
 BUILD_ARROW=ON
 SPARK_VERSION=ALL
-if [[ "$(uname)" == "Darwin" ]]; then
-    INSTALL_PREFIX=${INSTALL_PREFIX:-}
-else
-    INSTALL_PREFIX=${INSTALL_PREFIX:-"/usr/local"}
-fi
 
 # set default number of threads as cpu cores minus 2
 if [[ "$(uname)" == "Darwin" ]]; then
@@ -173,6 +168,12 @@ do
     esac
 done
 
+if [[ "$(uname)" == "Darwin" ]]; then
+    INSTALL_PREFIX=${INSTALL_PREFIX:-${VELOX_HOME}/deps-install}
+else
+    INSTALL_PREFIX=${INSTALL_PREFIX:-"/usr/local"}
+fi
+
 function concat_velox_param {
     # check velox repo
     if [[ -n $VELOX_REPO ]]; then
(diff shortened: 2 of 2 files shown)


dev/build_arrow.sh
```diff
@@ -35,9 +35,15 @@ function prepare_arrow_build() {
...
 function build_arrow_cpp() {
- pushd $ARROW_PREFIX/cpp
-
- cmake_install \
+  pushd $ARROW_PREFIX/cpp
+  ARROW_WITH_ZLIB=ON
+  if [[ "$(uname)" == "Darwin" ]]; then
+    clang_major_version=$(echo | clang -dM -E - | grep __clang_major__ | awk '{print $3}')
+    if [ "${clang_major_version}" -ge 17 ]; then
+      ARROW_WITH_ZLIB=OFF
```
philo-he, line 45: With ARROW_WITH_ZLIB=OFF, is there any impact on the functionalities?

zml1206: Here we just turn off on mac with clang 17,  no impact on others.

philo-he: I meant its impact when clang 17 is used.

I note there are some fixes in Velox community for clang 17 build, e.g., https://github.com/facebookincubator/velox/pull/13598. So I assume building velox/arrow can pass with clang 17 at that side. Curious if there is any other fix in Velox community instead of turning off this option.

zml1206: Velox can compile successfully probably because ZLIB_SOURCE=AUTO, I will try gluten later.

zml1206: There is no problem compiling arrow/velox using system's zlib, but compiling gluten cpp gives an error, it seems that the version is incompatible.
```
Undefined symbols for architecture arm64:
  "_deflate", referenced from:
      arrow::util::internal::(anonymous namespace)::GZipCodec::Compress(long long, unsigned char const*, long long, unsigned char*) in libarrow.a[120](compression_zlib.cc.o)
      arrow::util::internal::(anonymous namespace)::GZipCompressor::Compress(long long, unsigned char const*, long long, unsigned char*) in libarrow.a[120](compression_zlib.cc.o)
      arrow::util::internal::(anonymous namespace)::GZipCompressor::Flush(long long, unsigned char*) in libarrow.a[120](compression_zlib.cc.o)
```

I ran some UT locally on my Mac and there were no problems. I guess as long as don't use zlib compression, it shouldn't have any effect.

philo-he:
@zhouyuan, @rui-mo, are you using Mac OS clang 17 to build Gluten/Velox? Could you help confirm if this build issue can be reproduced on your side? Thanks!

zml1206:
This is a zlib problem: https://github.com/madler/zlib/issues/856
zlib 1.3.1 or later fixed this.

dev/build_arrow.sh
rui-mo: If zlib could be disabled for mac, can it actually be disabled for all systems?

rui-mo:
> @zhouyuan, @rui-mo, are you using Mac OS clang 17 to build Gluten/Velox? Could you help confirm if this build issue can be reproduced on your side? Thanks!

No, I'm using clang 16 and the compilation works.

dev/build_helper_functions.sh
```diff
@@ -141,6 +141,7 @@ function cmake_install {
 }
 
 function setup_macos {
+  sed -i '' '/run_and_time install_arrow/d' scripts/setup-macos.sh
```
rui-mo, line 144: Thanks for this fix!

dev/builddeps-veloxbe.sh
```diff
@@ -48,7 +48,11 @@ VELOX_HOME=""
 VELOX_PARAMETER=""
 BUILD_ARROW=ON
 SPARK_VERSION=ALL
-INSTALL_PREFIX=${INSTALL_PREFIX:-"/usr/local"}
+if [[ "$(uname)" == "Darwin" ]]; then
+    INSTALL_PREFIX=${INSTALL_PREFIX:-}
+else
+    INSTALL_PREFIX=${INSTALL_PREFIX:-"/usr/local"}
```
rui-mo: The INSTALL_PREFIX could be set manually in `.zshrc` for mac compilation. I wonder if we need this change to differentiate.

zml1206: It seems more appropriate to keep the default value of INSTALL_PREFIX consistent with velox's macOS.

dev/build_arrow.sh
zml1206: This is risky. We just temporarily disable zlib and enable it after arrow is upgraded.

zhouyuan (approved):
👍

zhouyuan:
There is some efforts from @wangyum to add basic CI on mac:
https://github.com/apache/incubator-gluten/pull/9579/files#diff-7b74d73de878e0e53f6768bbf3a3d4b50c3c6f1f16f9b64518690028b8206b1c

dev/build_arrow.sh
rui-mo: Given this, the question https://github.com/apache/incubator-gluten/pull/10158#discussion_r2197067155 comes back, would this change be risky to the mac os as well?

rui-mo: We may have a better track on the mac os build after https://github.com/apache/incubator-gluten/pull/9579, and it's good for me to merge this fix first.

zml1206: velox is preparing to upgrade arrow. We can try to enable zlib after the upgrade. https://github.com/facebookincubator/velox/pull/14102

zml1206:
Thanks @zhouyuan @rui-mo @PHILO-HE ,merge to main.

beliefer (commented):
Thanks for the fix.

PR #10161, opened 2025-07-10
[GLUTEN-10162] [build] fix: GCC build fix

NN---:
Clang 18 fails building.

## How was this patch tested?
Building using clang 18.

### DIFF
--- a/ep/build-velox/src/build_velox.sh
+++ b/ep/build-velox/src/build_velox.sh
@@ -101,7 +101,7 @@ function compile {
   # Maybe there is some set option in velox setup script. Run set command again.
   set -exu
 
-  CXX_FLAGS='-Wno-error=stringop-overflow -Wno-error=cpp -Wno-missing-field-initializers -Wno-unknown-warning-option'
+  CXX_FLAGS='-Wno-error=stringop-overflow -Wno-error=cpp -Wno-missing-field-initializers -Wno-unknown-warning-option -Wno-template-id-cdtor'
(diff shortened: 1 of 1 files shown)


zhouyuan:
Cc @zml1206 as he is also fixing macos builds

ep/build-velox/src/build_velox.sh
```diff
@@ -101,7 +101,7 @@ function compile {
   # Maybe there is some set option in velox setup script. Run set command again.
   set -exu
 
-  CXX_FLAGS='-Wno-error=stringop-overflow -Wno-error=cpp -Wno-missing-field-initializers -Wno-unknown-warning-option'
+  CXX_FLAGS='-Wno-error=stringop-overflow -Wno-error=cpp -Wno-missing-field-initializers -Wno-unknown-warning-option -Wno-template-id-cdtor'
```
zml1206, line 104: -Wno-template-id-cdtor seems to be gcc specific, not supported by clang.
```
clang-18 -Wno-template-id-cdtor a.cpp
warning: unknown warning option '-Wno-template-id-cdtor' [-Wunknown-warning-option]
```

NN---: Sorry for confusion. Yes it is for GCC.

zml1206: What version of gcc? Could you post the relevant error? Thanks

NN---: GCC 14

zml1206: It seems to be caused by C++20. Docker uses gcc11 and C++17 to compile, so this problem will not occur. You can use gcc11 or specify -std=c++17 to see if the problem can be solved. cc @zhouyuan @rui-mo

NN---: I see. Thanks. I am moving gluten build to a separate environment.

Issue #10099, opened 2025-07-02
[VL] Support reading offHeap memory from ResourceProfile


yes-un:
### Description

**Description**
Stage-level adjustment is already supported https://github.com/apache/incubator-gluten/pull/8209. However, Velox is still [reading the off-heap memory settings from SparkConf](https://github.com/apache/incubator-gluten/blob/f1fc5697d086a5b4a518edba318f44b7c7bf0860/cpp/velox/compute/WholeStageResultIterator.cc#L485). Our team tried increasing the off-heap memory size using ResourceProfile in gluten, but Velox continues to use the earlier value from SparkConf instead of recognizing the updated setting.

### Gluten version

main branch

yes-un:
@FelixYBW

zhztheplayer:
> However, Velox is still [reading the off-heap memory settings from SparkConf](https://github.com/apache/incubator-gluten/blob/f1fc5697d086a5b4a518edba318f44b7c7bf0860/cpp/velox/compute/WholeStageResultIterator.cc#L485).

This code should only be effective when there're hash aggregations in the query plan. Have you already faced issues with hash aggregation's memory usage?

yes-un:
> > However, Velox is still [reading the off-heap memory settings from SparkConf](https://github.com/apache/incubator-gluten/blob/f1fc5697d086a5b4a518edba318f44b7c7bf0860/cpp/velox/compute/WholeStageResultIterator.cc#L485).
> 
> This code should only be effective when there're hash aggregations in the query plan. Have you already faced issues with hash aggregation's memory usage?

@zhztheplayer The query itself doesn't include any has aggregations. The error we got is below(we used ResourceProfile to ask offHeap memory to be 4G, but still is restricted by SparkConf offheap size):

```
[...]

PR #10172, opened 2025-07-11
[CORE] Reflect resource changes in configurations when applying new resource profile

PHILO-HE:

See discussion: https://github.com/apache/incubator-gluten/issues/10099#issuecomment-3034219121

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
+++ b/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
@@ -23,11 +23,11 @@ import org.apache.gluten.logging.LogLevelUtil
 import org.apache.spark.SparkConf
 import org.apache.spark.annotation.Experimental
 import org.apache.spark.internal.Logging
-import org.apache.spark.internal.config.{CPUS_PER_TASK, EXECUTOR_CORES}
+import org.apache.spark.internal.config.{CPUS_PER_TASK, EXECUTOR_CORES, MEMORY_OFFHEAP_SIZE}
 import org.apache.spark.resource.{ExecutorResourceRequest, ResourceProfile, ResourceProfileManager, TaskResourceRequest}
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.rules.Rule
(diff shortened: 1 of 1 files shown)


gluten-substrait/src/main/scala/org/apache/spark/sql/execution/GlutenAutoAdjustStageResourceProfile.scala
```diff
@@ -156,17 +158,25 @@ object GlutenAutoAdjustStageResourceProfile extends Logging {
...
+    val coresPerTask = rp.getTaskCpus.getOrElse(sparkConf.get(CPUS_PER_TASK))
+    val taskSlots = coresPerExecutor / coresPerTask
+    SQLConf.get.setConfString(GlutenCoreConfig.NUM_TASK_SLOTS_PER_EXECUTOR.key, taskSlots.toString)
+    val offHeapSize = rp.executorResources.get(ResourceProfile.OFFHEAP_MEM).map(_.amount).get
+    SQLConf.get.setConfString(
+      GlutenCoreConfig.COLUMNAR_OFFHEAP_SIZE_IN_BYTES.key,
+      offHeapSize.toString)
+    SQLConf.get.setConfString(
+      GlutenCoreConfig.COLUMNAR_TASK_OFFHEAP_SIZE_IN_BYTES.key,
+      (offHeapSize / taskSlots).toString)
```
zjuwangg, line 191: This config should be reset to original value if stage still use the default resource profile after some stage hit the rule here.

philo-he: Yes, we should consider this situation. Just updated. Thanks!

philo-he:
@zjuwangg, could you help review this pr again?

zjuwangg:
> @zjuwangg, could you help review this pr again?

LGTM. Thanks

zjuwangg (approved):

ayushi-agarwal:
@PHILO-HE We are seeing performance degradation in q67 and q28 of TPCDS on 1TB dataset after this patch if we enable auto adjust resource profile feature  - In both the queries after expand operator there is a flushable Hash aggregate operator which is not aggregating rows as it used to before this patch which results in large shuffle and processing in later stage, could you help with some pointers as to why it can happen? There was no ApplyResourceProfileExec node in both the queries in the plan.

philo-he:
Assume the regression has been fixed by PR #10389, according to https://github.com/apache/incubator-gluten/pull/10389#issuecomment-3183285805.

PR #3076, opened 2023-09-08
[GLUTEN-3075][VL]FEAT: Support static partition write in velox backend

JkSelf:
## What changes were proposed in this pull request?

This PR only supports static dynamic partition writes. Additionally, we have removed the sort operator before writing, as it is unnecessary for static partition writes.

## How was this patch tested?

existing uts

PR #10183, opened 2025-07-14
[VL] Spark 3.2 / Spark 3.3, V1 write: Dynamic partition write

zhztheplayer:
In https://github.com/apache/incubator-gluten/pull/3076 we had added (3.2 / 3.3) V1 write support for statically partitioned input data.

The PR fills in the relevant APIs to support dynamically partitioned data as well.

The reason we are continuing enhancing the 3.2 / 3.3 V1 write is, the V1 file format API is not only used by legacy Spark, but lake formats either. This PR essentially makes the API `splitBlockByPartitionAndBucket` work as expected so (e.g.,) it could be called by a delta lake / iceberg plugin to support partitioned write without having to send the partition keys / expressions to native.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxParquetWriteForHiveSuite.scala
@@ -196,7 +196,8 @@ class VeloxParquetWriteForHiveSuite
   }
 
   test(
-    "test hive dynamic and static partition write table multiple keys, multiple partitions, single batch to write") {
+    "test hive dynamic and static partition write table multiple keys, multiple partitions, " +
+      "single batch to write") {
     withTable("t") {
       spark.sql(
         "CREATE TABLE t (c int, d long, e long, f int)" +


zhztheplayer:

cpp/velox/jni/VeloxJniWrapper.cc
```diff
@@ -401,45 +403,104 @@ Java_org_apache_gluten_datasource_VeloxDataSourceJniWrapper_splitBlockByPartitio
...
-  auto result = batch->toUnsafeRow(0);
-  auto rowBytes = result.data();
-  auto newBatchHandle = ctx->saveObject(ctx->select(batch, dataColIndicesVec));
+  auto pool = dynamic_cast<VeloxMemoryManager*>(ctx->memoryManager())->getLeafMemoryPool();
+  const auto veloxBatch = VeloxColumnarBatch::from(pool.get(), batch);
+  const auto inputRowVector = veloxBatch->getRowVector();
+  const auto numRows = inputRowVector->size();
+
+  connector::hive::PartitionIdGenerator idGen{
+      asRowType(inputRowVector->type()), partitionColIndicesVec, 128, pool.get(), true};
```
JkSelf, line 438: Can we make the maxPartitions and partitionPathAsLowerCase  be configurable?

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
```diff
@@ -127,6 +127,8 @@ object GlutenWriterColumnarRules {
     if (format.isDefined) {
       spark.sparkContext.setLocalProperty("isNativeApplicable", true.toString)
       spark.sparkContext.setLocalProperty("nativeFormat", format.get)
+      // The flag for static-only write is not used by Velox backend where
+      // the static partition write is already supported.
       spark.sparkContext.setLocalProperty(
         "staticPartitionWriteOnly",
```
JkSelf, line 133: We can remove staticPartitionWriteOnly after implement the dynamic partition write support.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
```diff
@@ -291,15 +287,10 @@ object FileFormatWriter extends Logging {
             empty2NullPlan,
             Some(ConcurrentOutputWriterSpec(maxWriters, () => sortPlan.createSorter())))
         } else {
-          if (staticPartitionWriteOnly && nativeEnabled) {
```
JkSelf, line 294: There is no need to add the sort operator before the static partition write. Can we still remove the sort operator for the static partition write?

cpp/velox/jni/VeloxJniWrapper.cc
zhztheplayer: Will follow up.

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala
zhztheplayer: I think we can open a PR to remove this later on to see if CH backend is still relying on this flag. I am not sure about that atm.

shims/spark32/src/main/scala/org/apache/spark/sql/execution/datasources/FileFormatWriter.scala
zhztheplayer: I debugged on this. And it seems we have `orderingMatched = true` in the case that all the partitions are static. So this code path will not be invoked. Would you help me confirm that?

JkSelf (approved):

PR #10186, opened 2025-07-15
[GLUTEN-9337][VL] Support read Paimon non-PK table

liujiayi771:

Support read Paimon non-pk table (without primary key).

## How was this patch tested?

Add some read Paimon table test cases.


liujiayi771:
@YannByron @zhouyuan @Zouxxyy Could you help review?

gluten-paimon/src-paimon/main/scala/org/apache/gluten/execution/PaimonScanTransformer.scala
```diff
@@ -0,0 +1,209 @@
...
+
+  override def getPartitionSchema: StructType = scan match {
+    case paimonScan: PaimonScan =>
+      val partitionKeys = paimonScan.table.partitionKeys()
+      StructType(scan.readSchema().filter(field => partitionKeys.contains(field.name)))
+    case _ =>
+      throw new GlutenNotSupportException("Only support PaimonScan.")
+  }
+
+  override def getDataSchema: StructType = new StructType()
```
jinchengchenghh, line 86: Could we throw exception here?

gluten-paimon/src-paimon/main/scala/org/apache/gluten/execution/PaimonScanTransformer.scala
```diff
@@ -0,0 +1,209 @@
...
+          case table: FileStoreTable =>
+            if (!table.primaryKeys().isEmpty || coreOptions.deletionVectorsEnabled()) {
+              return ValidationResult.failed("Not support Paimon PK/DV table")
+            }
+          case table =>
+            return ValidationResult.failed(
+              s"Not support Paimon ${table.getClass.getSimpleName} table")
+        }
+      case _ =>
+        throw new GlutenNotSupportException("Only support PaimonScan.")
```
jinchengchenghh: Why not return ValidationResult.failed?

gluten-paimon/src-paimon/main/scala/org/apache/gluten/execution/PaimonScanTransformer.scala
```diff
@@ -0,0 +1,209 @@
...
+      throw new GlutenNotSupportException(
+        s"Only support Parquet/ORC format, " +
+          s"but found $formatStr")
+    }
+  }
+
+  override def doValidateInternal(): ValidationResult = {
+    scan match {
+      case paimonScan: PaimonScan =>
+        paimonScan.table match {
```
jinchengchenghh, line 102: Do you to validate fileFormat too?

liujiayi771: Yes, only parquet and orc are supported.

gluten-paimon/src-paimon/main/scala/org/apache/gluten/execution/PaimonScanTransformer.scala
liujiayi771: No, it is used in BasicScanExecTransformer.doValidateInternal, and IcebergScanTransformer also returns an empty StructType.

zhouyuan:
@liujiayi771 could you please do a rebase? there is a small conflict

docs/get-started/Velox.md
```diff
@@ -359,6 +359,20 @@ Once built successfully, iceberg features will be included in gluten-velox-bundl
 
 Gluten with velox backend supports [Hudi](https://hudi.apache.org/) table. Currently, only reading COW (Copy-On-Write) tables is supported.
 
+## Paimon Support
+
+Gluten with velox backend supports [Paimon](https://paimon.apache.org/) table. Currently, only non-pk table is supported, and the Spark version needs to be >= 3.3.
```
zhztheplayer, line 372: > and the Spark version needs to be >= 3.3.

Will there be a Maven build error for `-P spark-3.2,paimon`? If so we can set an enforcer rule to exclude the combination.

E.g., this one:

https://github.com/apache/incubator-gluten/blob/75c6f27cb2a776edca56c1fabc76262f655447ae/backends-clickhouse/pom.xml#L575-L595

gluten-paimon/src-paimon/main/java/org/apache/gluten/substrait/rel/PaimonLocalFilesNode.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gluten.substrait.rel;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class PaimonLocalFilesNode extends LocalFilesNode {
```
zhztheplayer, line 23: Any particular reason for adding this Substrait node?

gluten-paimon/src-paimon/main/scala/org/apache/gluten/execution/PaimonScanTransformer.scala
```diff
@@ -0,0 +1,210 @@
...
+          table.partitionKeys.asScala.toArray,
+          false
+        )
+      case _ =>
+        throw new GlutenNotSupportException("Only support PaimonScan.")
+    }
+  }
+
+  def supportsBatchScan(scan: Scan): Boolean = {
+    scan.getClass.getName == "org.apache.paimon.spark.PaimonScan"
```
zhztheplayer, line 208: nit: Are we able to do `scan.isInstanceOf` here?

gluten-paimon/src-paimon/main/java/org/apache/gluten/substrait/rel/PaimonLocalFilesNode.java
liujiayi771: No special functionality is currently required, as Paimon does not have any parts needing special handling at the moment. However, in the future, when supporting additional features, specific logic will need to be implemented in this file.

docs/get-started/Velox.md
liujiayi771: The maven-enforcer-plugin makes it easy to enforce that two profiles must be specified together, but it's not straightforward to define mutual exclusion. Have you ever configured mutual exclusion before?

gluten-paimon/src-paimon/main/scala/org/apache/gluten/execution/PaimonScanTransformer.scala
liujiayi771: We cannot use `asInstanceOf` here because this method may be triggered during offload when Paimon is not in use, and the missing Paimon dependency would cause a `NoClassDefFoundError`.

pom.xml
```diff
@@ -59,6 +59,7 @@
     <module>package</module>
     <module>shims</module>
     <module>gluten-ras</module>
+    <module>gluten-paimon</module>
```
philo-he: Only include this module when profile paimon is used? Like `gluten-iceberg` module. I assume we should remove this.

gluten-paimon/src-paimon/test/scala/org/apache/gluten/execution/PaimonSuite.scala
```diff
@@ -0,0 +1,97 @@
...
+    Seq("parquet", "orc").foreach {
+      format =>
+        withTable(s"paimon_${format}_tbl") {
+          sql(s"DROP TABLE IF EXISTS paimon_${format}_tbl")
+          sql(
+            s"CREATE TABLE paimon_${format}_tbl (id INT, name STRING) USING PAIMON " +
+              s"TBLPROPERTIES ('file.format'='$format')")
+          sql(s"INSERT INTO paimon_${format}_tbl VALUES (1, 'Bob'), (2, 'Blue'), (3, 'Mike')")
+          runQueryAndCompare(s"""
+                                |SELECT * from paimon_${format}_tbl;
```
philo-he: Nit: use "FROM" for consistency.

philo-he:
Looks the following error needs to be fixed. See the comment that suggested to only include gluten-paimon module when the profile is activated.
```
The following artifacts could not be resolved: org.apache.paimon:paimon-spark-4.0:jar:1.2.0 (absent)
```

PHILO-HE (commented):
Thanks.

gluten-paimon/src-paimon/main/scala/org/apache/gluten/execution/PaimonScanTransformer.scala
philo-he: This check is in gluten-paimon module where the paimon dependency is included. Then, can we assume the class is available to use?

liujiayi771: @PHILO-HE When -`Ppaimon` is added, the `gluten-paimon` module will be compiled. When executing test cases in `gluten-ut`, due to the service loader mechanism, `VeloxPaimonComponent` will be loaded, triggering the execution of `OffloadPaimonScan.offload`. However, since there is no Paimon dependency in `gluten-ut`, an error will occur indicating that the `PaimonScan` class cannot be found.

PR #10187, opened 2025-07-15
[CORE] Remove the class-overriding on InsertIntoHadoopFsRelationCommand

zhztheplayer:
A follow up to #3460 to avoid overriding too much code of vanilla Spark.
 
Keep in mind that the way we are relying on after the fix, i.e, `sparkSession.sparkContext.getLocalProperty` and `sparkSession.sparkContext.setLocalProperty`, is not fancy either.


zhztheplayer:
@lwz9103 @JkSelf

PR #10190, opened 2025-07-15
[GLUTEN-8070][CORE][VL] Add mechanism for generating configuration files and check configuration content

yikf:
fix https://github.com/apache/incubator-gluten/issues/8070. this pr aims to add mechanism for generating configuration files and check configuration content.


yikf:
@zhouyuan could you please take a look?

docs/velox-configuration.md
```diff
@@ -0,0 +1,81 @@
+<!-- DO NOT MODIFY THIS FILE DIRECTLY, IT IS AUTO-GENERATED BY [org.apache.gluten.config.AllVeloxConfiguration] -->
+
+---
+
+layout: page
+title: Configuration
+nav_order: 3
```
zhouyuan: may need to use different number for the nav_order

docs/Configuration.md
```diff
@@ -1,130 +1,162 @@
...
-| spark.driver.extraClassPath                                   | Gluten Plugin jar file to prepend to the classpath of the driver.                                                                                                                                                                                                                                                                                                                                                                                                                                                                         | /path/to/jar_file                                    |
```
zhouyuan, line 16: It looks like we missed some essential parameters like spark.extraClassPath/spark.plugins/org.apache.spark.shuffle.sort.ColumnarShuffleManager here

zhouyuan (commented):
Cool! We are in the middle of clean up the documents. This patch helped a lot!

docs/Configuration.md
yikf: thanks, these parameters are spark config, i want to create new section to describe the base spark config for gluten.

yikf: and the doc only contain public config, what do you think? will not doc them if the config is internal.

zhouyuan:
@yikf thanks for this improvement. I will add a automated weekly check on this to update the configurations docs.

FelixYBW:
@yikf  Some configurations are defined different default value in scala and c++. Like

https://github.com/apache/incubator-gluten/blob/35f516cf2298c83be723cdd75fa71489a6317e54/gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala#L577-L579

and 

https://github.com/apache/incubator-gluten/blob/d7560eaa9401e55e7be6595f54dafcaf70b59b5e/cpp/velox/config/VeloxConfig.h#L115

Is there good way to handle this? One possible solution is to set invalid value in c++ to make sure the config should always be passed from java or external config.

yikf:
@FelixYBW Thank you for bringing this up.


I have previously considered this proposition. In fact, apart from default values, configuration types (such as bool and int) also collaborate across languages in a conventional way rather than being managed centrally. 

I want to use an intermediate layer to centrally manage and parse configurations, But this cost seems too heavy(not sure).

PR #10193, opened 2025-07-15
[GLUTEN-10192][VL] Fix sort shuffle read segfault in some cases

marin-ma:
Segfault occurs when reading a row that exceed current rowBuffer size

https://github.com/apache/incubator-gluten/pull/10193/files#diff-c1147845fed4ad34faf580505712b04befdd3df0361297f93cb660069bedb7bdR646-R648

### DIFF
--- a/cpp/velox/tests/VeloxShuffleWriterTest.cc
+++ b/cpp/velox/tests/VeloxShuffleWriterTest.cc
@@ -43,6 +43,7 @@ struct ShuffleTestParams {
   int32_t diskWriteBufferSize{0};
   bool useRadixSort{false};
   bool enableDictionary{false};
+  int64_t deserializerBufferSize{0};
 
   std::string toString() const {
     std::ostringstream out;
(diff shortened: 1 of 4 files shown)


cpp/velox/tests/VeloxShuffleWriterTest.cc
```diff
@@ -43,6 +43,7 @@ struct ShuffleTestParams {
   int32_t diskWriteBufferSize{0};
   bool useRadixSort{false};
   bool enableDictionary{false};
+  int64_t deserializerBufferSize{0};
```
jinchengchenghh, line 46: Where is the usage of deserializerBufferSize

marin-ma: It's used on shuffle reader side, to hold the decompressed rows from the inputstream.

jinchengchenghh (approved):

PR #10202, opened 2025-07-16
[GLUTEN-8332][VL] Support explode/posexplode/inline outer

marin-ma:


zhouyuan (commented):
The code looks good, is it convenient to also enable the `GeneratorExpressionSuite` ?
https://github.com/apache/spark/blob/4cc8101cb5e0e52af10a2fa4d74696addd95e478/sql/catalyst/src/test/scala/org/apache/spark/sql/catalyst/expressions/GeneratorExpressionSuite.scala#L26

marin-ma:
@zhouyuan Added UT. Could you help to review again? Thanks!

zhouyuan (approved):
👍 
Thanks!

Issue #10200, opened 2025-07-16
[VL] VeloxRssSortShuffleWriter calculates same stringbuffer multiple times when estimating RowVector size


NEUpanning:
### Backend

VL (Velox)

### Bug description

Currently, VeloxRssSortShuffleWriter calculates RowVector batch size via [RowVector::estimateFlatSize](https://github.com/apache/incubator-gluten/blob/main/cpp/velox/shuffle/VeloxRssSortShuffleWriter.cc#L54), but the string buffer shared by multiple vectors will be calculated multiple times instead of once. As a result, the size could be way smaller than the actual batch size and the buffer sent to Celeborn could have too many small batches leading to poor performance.

I think we can calculate the shared string buffer only once to solve this issue.

### Gluten version

Gluten-1.3

### Spark version

Spark-3.5.x

### Relevant logs

```bash

```

NEUpanning:
@kerwin-zk Hello, could you share your thoughts? Thanks!

PR #10235, opened 2025-07-21
[GLUTEN-10200][VL] Fix estimating row vector size logic of rss shuffle writer

NEUpanning:

(Fixes: \#10200)

## How was this patch tested?

UT

### DIFF
--- a/cpp/velox/tests/VeloxRssSortShuffleWriterTest.cc
+++ b/cpp/velox/tests/VeloxRssSortShuffleWriterTest.cc
@@ -61,7 +61,7 @@ class VeloxRssSortShuffleWriterTest : public VeloxShuffleWriterTestBase, public 
       GLUTEN_ASSIGN_OR_THROW(codec, arrow::util::Codec::Create(writerOptions->compressionType));
     }
     auto partitionWriter = std::make_shared<RssPartitionWriter>(
-        numPartitions, codec, getDefaultMemoryManager(), options, std::move(rssClient));
+        numPartitions, std::move(codec), getDefaultMemoryManager(), options, std::move(rssClient));
     GLUTEN_ASSIGN_OR_THROW(
         auto shuffleWriter,
         VeloxRssSortShuffleWriter::create(


marin-ma:
cc: @kerwin-zk

kerwin-zk:
@NEUpanning  Thanks for your PR! Generally, shared string buffers might occur during copy operations and function calls. Could you conveniently introduce the production scenarios where you have encountered so many shared string buffers?

cpp/velox/shuffle/VeloxRssSortShuffleWriter.h
```diff
@@ -89,6 +95,10 @@ class VeloxRssSortShuffleWriter final : public VeloxShuffleWriter {
 
   void stat() const;
 
+  void calculateBatchesSize(facebook::velox::RowVectorPtr& vector);
```
kerwin-zk: Here the suggestion is to use ```const```

cpp/velox/shuffle/VeloxRssSortShuffleWriter.h
```diff
@@ -89,6 +95,10 @@ class VeloxRssSortShuffleWriter final : public VeloxShuffleWriter {
...
+
+  void deduplicateStrBuffer(facebook::velox::VectorPtr& vector);
```
kerwin-zk: Same as above

cpp/velox/shuffle/VeloxRssSortShuffleWriter.h
```diff
@@ -108,6 +118,10 @@ class VeloxRssSortShuffleWriter final : public VeloxShuffleWriter {
   uint32_t currentInputColumnBytes_ = 0;
 
   RssSortState sortState_{kSortInit};
+
+  // The existing string buffers in the current batches.
+  std::unordered_set<facebook::velox::Buffer*> stringBuffers_;
```
kerwin-zk: It would be better to use  ```folly::F14FastSet<const Buffer*>```

NEUpanning:
> Could you conveniently introduce the production scenarios where you have encountered so many shared string buffers?

@kerwin-zk The physical plan before shuffle did not include any string functions but I noticed one string buffer was shared by over 30 row vectors. I am not sure the cause of that.

```
ColumnarExchange hashpartitioning(user_key#326, 8000), ENSURE_REQUIREMENTS, [user_key#326, collect_date_list#338], [plan_id=1608], [shuffle_writer_type=sort], [OUTPUT] List(user_key:StringType, collect_date_list:StringType)
            +- ^(2) ProjectExecTransformer [hash(user_key#326, 42) AS hash_partition_key#391, collect_date_list#338]
                +- ^(2) FileScanTransformer orc tableA PartitionFilters: [isnotnull(partition_date#339), isnotnull(partition_identity_type#341), isnotnull(partition_sourc...
```

PR #10240, opened 2025-07-22
[GLUTEN-9849][VL] Reenable native `might_contain` evaluation that was disabled in #9850 

zhztheplayer:
Following: https://github.com/apache/incubator-gluten/pull/9850
Velox PR needed: https://github.com/facebookincubator/velox/pull/13955

In #9850 we have disabled native `might_contain` in some cases where the expression is evaluated on Spark driver.

The patch uses static evaluation of `might_contain` for such cases and re-enable the native code path.


zhztheplayer:
@wForget @jinchengchenghh Thanks

backends-velox/src/main/scala/org/apache/gluten/expression/VeloxBloomFilterMightContain.scala
```diff
@@ -26,21 +25,23 @@ import org.apache.spark.sql.catalyst.expressions.{BinaryExpression, Expression}
...
- * FIXME: Remove GlutenTaskOnlyExpression after the VeloxBloomFilter expr is made compatible with
- * spark. See: https://github.com/apache/incubator-gluten/pull/9850#issuecomment-3007448538
  */
 case class VeloxBloomFilterMightContain(
     bloomFilterExpression: Expression,
     valueExpression: Expression)
-  extends BinaryExpression
-  with GlutenTaskOnlyExpression {
+  extends BinaryExpression {
+  import VeloxBloomFilterMightContain._
```
jinchengchenghh, line 44: Do we import here?

zhztheplayer: Yes, for `SerializableDirectByteBuffer`.

jinchengchenghh (approved):

zhztheplayer:
Thanks @wForget @jinchengchenghh!

PR #9770, opened 2025-05-27
[GLUTEN-9769][CELEBORN]  Use the correct shuffle time metrics for celeborn columnar shuffle

wankunde:
## What changes were proposed in this pull request?

Bug fix for velox + celeborn + columnar shuffle.

gluten changes the shuffle writer type of celeborn in https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L555-L557

As a result, we cannot determine whether the shuffle writer in VeloxCelebornColumnarShuffleWriter is sort or not,  can only  use the passed in shuffle time metric .

(Fixes: \#9769)

## How was this patch tested?

manual tests

Issue #9769, opened 2025-05-27
[VL] Fail to get the correct celeborn shuffle time metrics


wankunde:
### Backend

VL (Velox)

### Bug description

# Expected behavior

```ini
  --conf spark.gluten.sql.columnar.shuffle=true \
  --conf spark.shuffle.manager=org.apache.spark.shuffle.gluten.celeborn.CelebornShuffleManager \
  --conf spark.gluten.sql.columnar.shuffle.sort.partitions.threshold=100 \
  --conf spark.celeborn.client.spark.shuffle.writer=hash \
```

When I run a query with gluten + celeborn + columnar shuffle + hash writer , when the 
https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L555-L557
 return true, 
 the `useSortBasedShuffle` in ColumnarShuffleExchangeExec will is true while the isSort in VeloxCelebornColumnarShuffleWriter will be false

# actual behavior

The isSort in VeloxCelebornColumnarShuffleWriter false while the `useSortBasedShuffle` in ColumnarShuffleExchangeExec is true.
They are not matched.

### Spark version

None

### Spark configurations

  --conf spark.gluten.sql.columnar.shuffle=true 
  --conf spark.shuffle.manager=org.apache.spark.shuffle.gluten.celeborn.CelebornShuffleManager 
  --conf spark.gluten.sql.columnar.shuffle.sort.partitions.threshold=100 
  --conf spark.celeborn.client.spark.shuffle.writer=hash 

### Relevant logs

```bash
Lost task 0.0 in stage 2.0 (TID 2) (datanode-4258.host executor 1): java.util.NoSuchElementException: key not found: splitTime
	at scala.collection.immutable.BitmapIndexedMapNode.apply(HashMap.scala:674)
	at scala.collection.immutable.BitmapIndexedMapNode.apply(HashMap.scala:672)
	at scala.collection.immutable.HashMap.apply(HashMap.scala:132)
	at org.apache.spark.shuffle.VeloxCelebornColumnarShuffleWriter.internalWrite(VeloxCelebornColumnarShuffleWriter.scala:104)
	at org.apache.spark.shuffle.CelebornColumnarShuffleWriter.write(CelebornColumnarShuffleWriter.scala:119)
[...]

PR #10241, opened 2025-07-22
[GLUTEN-9769][CELEBORN][1.4]  Use the correct shuffle time metrics for celeborn columnar shuffle (#9770)

wForget:
Backport #9770 to branch-1.4

Bug fix for velox + celeborn + columnar shuffle.

gluten changes the shuffle writer type of celeborn in https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L555-L557

As a result, we cannot determine whether the shuffle writer in VeloxCelebornColumnarShuffleWriter is sort or not,  can only  use the passed in shuffle time metric .

(Fixes: \#9769)

## How was this patch tested?

manual tests

### DIFF
--- a/backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala
+++ b/backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala
@@ -34,7 +34,6 @@ import org.apache.spark.util.SparkResourceUtil
 import org.apache.celeborn.client.ShuffleClient
 import org.apache.celeborn.common.CelebornConf
 import org.apache.gluten.config.GlutenConfig
-import org.apache.gluten.config.ReservedKeys
 
 import java.io.IOException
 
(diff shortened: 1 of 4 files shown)


zhouyuan:
@wForget 
The GHA failures are due to recent pyarrow upgrade. We need to pin it to 20.0.0
https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_backend_x86.yml#L642

Could you please also add this fix in 1.4 branch as well? 
https://github.com/apache/incubator-gluten/blob/branch-1.4/.github/workflows/velox_backend.yml#L638

wForget:
> @wForget The GHA failures are due to recent pyarrow upgrade. We need to pin it to 20.0.0 https://github.com/apache/incubator-gluten/blob/main/.github/workflows/velox_backend_x86.yml#L642
> 
> Could you please also add this fix in 1.4 branch as well? https://github.com/apache/incubator-gluten/blob/branch-1.4/.github/workflows/velox_backend.yml#L638

Thanks, I will fix CI for 1.4

PR #10244, opened 2025-07-22
[GLUTEN-10236][VL] Support both sort and rss_sort shuffle writer for Celeborn

marin-ma:
By default, use rss sort shuffle implementation for Celeborn. User can set `spark.gluten.sql.columnar.shuffle.celeborn.useRssSort=false` to use row-based sort implementation.


zhouyuan:
Cc @kerwin-zk

FelixYBW:
Could you mention this in https://github.com/apache/incubator-gluten/blob/main/docs/get-started/Velox.md#celeborn-support? There are two set of sort based shuffle implemented and supported in celeborn. Use the config to switch.

How about Uniffle? Does it only support sort based shuffle, right?

marin-ma:
> How about Uniffle? Does it only support sort based shuffle, right?

@FelixYBW The supported shuffle writer type for Uniffle is same as the local shuffle. We support both hash and row-based sort. We don't use the rss sort shuffle implementation for Uniffle.

Issue #10247, opened 2025-07-23
[FLINK] Support date_format function


KevinyhZou:
### Description

support date_format

### Gluten version

None

PR #10248, opened 2025-07-23
[GLUTEN-10247][FLINK]Support date_format function

KevinyhZou:

(Fixes: \#10247)

## How was this patch tested?
UT


lgbo-ustc (approved):

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -126,4 +126,14 @@ void testEqual() {
...
     runAndCheck(query4, Arrays.asList("+I[false]", "+I[true]", "+I[false]"));
   }
+
+  @Test
+  void testDateFormat() {
+    List<Row> rows =
+        Arrays.asList(Row.of(1, "2024-12-31 12:12:12"), Row.of(2, "2025-02-28 12:12:12"));
+    createSimpleBoundedValuesTable("dateFormatTbl", "a int, b string", rows);
+    String query =
+        "select a, DATE_FORMAT(cast(b as Timestamp(3)), 'yyyy-MM-dd') from dateFormatTbl";
```
philo-he: In Spark and Velox, date_format respects the configured session timezone. I assume Flink should also respect it.

The test here cannot expose the issue that local session timezone is not respected. The default timezone (utc) is used in cast string to timestamp. Then, the same default timezone is used for date_format. As long as the timezone is consistent, there is no issue exposed. But the actual timezone configured for Flink is not actually respected.

I think we need to pass Flink's session timezone to Velox's config, as Gluten Spark code does.

KevinyhZou: Now I have pass the flink session timezone to velox, and add the ut to test this. please take review again. @PHILO-HE

KevinyhZou: It seems not properly to pass the session timezone config to the velox, as in `StatefulTask#initOperators` will check the memory reversed, and would throw exceptions if it's not 0.

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/config/VeloxQueryConfig.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public static Config getConfig(RuntimeContext context) {
+    if (!(context instanceof StreamingRuntimeContext)) {
+      return Config.empty();
+    }
+    Configuration config = ((StreamingRuntimeContext) context).getJobConfiguration();
+    Map<String, String> configMap = new HashMap<>();
+    if (config.get(ADJUST_TIMESTMP_TO_SESSION_TIMEZONE)) {
+      String localTimeZone = config.get(TableConfigOptions.LOCAL_TIME_ZONE);
+      configMap.put(keyVeloxAdjustTimestampToSessionTimeZone, "true");
```
philo-he: In Spark, we always set this to true to let velox respect configured session timezone. I assume we should also do this for Flink, not depending on the user's setting for `ADJUST_TIMESTMP_TO_SESSION_TIMEZONE`.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -169,4 +172,28 @@ void testDecimal() {
...
+    runAndCheck(
+        query,
+        Arrays.asList(
+            "+I[1, 2024-12-31, 2024-12-31 12:12:12]", "+I[2, 2025-02-28, 2025-02-28 12:12:12]"));
+    Map<String, String> configs =
+        Map.of(
+            VeloxQueryConfig.ADJUST_TIMESTMP_TO_SESSION_TIMEZONE.key(),
+            "true",
+            TableConfigOptions.LOCAL_TIME_ZONE.key(),
+            "America/Los_Angeles");
```
philo-he: In this designed test, the result cannot reflect the timezone is changed. As I mentioned, both cast & date_format are respecting the timezone. As long as the timezone they are using is same, the result aligns with the input date time string, even though `America/Los_Angeles` is not passed to velox but the default timezone is in use. So the test here cannot confirm this new timezone is passed to Velox and taking effect.

I haven't come up with a good way. Do you have any idea?

KevinyhZou: I think we can set some illegal timezone value to velox, but which is correct to flink , e.g : `default` timezone.  And velox would throw a exception with `invalid timezone`, but flink can pass it. and in contrast, we pass a valid timezone  to the test, e.g : `America/Los_Angeles`, can get the correct result. and it will prove that the timezone is passed to velox. @PHILO-HE

philo-he: @KevinyhZou, it may not well test the behavior of date_format function on respecting timezone.

Could you confirm if the suggestion below makes sense?

1. Set timezone to UTC
`CREATE TEMPRARY VIEW my_view AS SELECT CAST(b AS TIMESTAMP) as c1 FROM tbl;`

2. Set timezone to "America/Los_Angeles"
`SELECT date_format(c1) FROM my_view;`
Check the result to see if it matches the expected string (-8 hours offset, compared with the string column b).

3. Set timezone to "Asia/Shanghai"
`SELECT date_format(c1) FROM my_view;`
Check the result to see if it matches the expected string (+8 hours offset, compared with the string column b).

KevinyhZou: It does not work. It always return the same string while time zone set to different(`UTC`, `Asia/Shanghai`, `America/Los_Angeles`).

philo-he: Do you know the reason? If different timestamp is set in Velox, date_format should return different date/time string for the same timestamp input.

KevinyhZou: The flink `create view` is `logical view`, data is not materialized， so you described above it is the same as `select ... date_format(cast(xxx as timestamp), 'yyyy-MM-dd HH:mm:ss')`, and this is different from spark. 

And we can still test this, by insert the result of 1st step to a `table` instead of `view`, and then the 2nd and 3rd step result can be affected by the timezone, but the timestamp with timezone type  is not supported, so the 1st step would cause errors, and I have tried this yesterday.

philo-he: Do you mean the inserting will create TIMESTAMP_LTZ type in the 1st step, not TIMESTAMP type?

KevinyhZou: Yes, similar to that. `cast(xxx as Timestamp(3))` will return a Timestamp vector with timezone by using the arrow's `TimeStampMilliTZVector`, and now we only map the `TimeStampMilliVector` to flink's `timestamp` type, but not do anything about `TimeStampMilliTZVector` and flink's timestamp type with timezone, like `TIMESTAMP_LTZ`. maybe we should open another issue for this?

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/config/VeloxQueryConfig.java
```diff
@@ -0,0 +1,72 @@
...
+    String localTimeZone = config.get(TableConfigOptions.LOCAL_TIME_ZONE);
+    if (TableConfigOptions.LOCAL_TIME_ZONE.defaultValue().equals(localTimeZone)) {
+      configMap.put(keyVeloxSessionTimezone, "UTC");
+    } else {
+      configMap.put(keyVeloxSessionTimezone, localTimeZone);
+    }
+    return Config.create(configMap);
+  }
+
+  private static boolean isConfigUsedForTest(Configuration config) {
```
shuai-xu: Why need to add test related code here?

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
philo-he: @KevinyhZou, if I understand correctly, we should ensure the result type is Flink's `timestamp` type for `cast(xxx as timestamp)`. You can submit another PR to fix the issue.

I investigated Flink just now. It seems Flink's `timestamp` type is corresponding to Spark's `timestamp_ntz` type (no counterpart in Velox), and Flinks' `timestamp_ltz` type is corresponding to Spark/Velox's `timestamp` type. Please confirm this to avoid wrong type mapping.

In Flink, when casting datetime string to `timestamp_ltz` or outputting string represented result for `timestamp_ltz`, timezone is respected. But regarding `timestamp` type, timezone is not respected in such conversions.

Both `timestamp` and `timestamp_ltz` are supported for Flink's date_format function. So if `timestamp` is used, the result is not influenced by session timezone. But if `timestamp_ltz` is used, the result should be adjusted based on session timezone.

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/operators/GlutenSingleInputOperator.java
```diff
@@ -133,7 +135,9 @@ public void processElement(StreamRecord<RowData> element) {
   @Override
   public void close() throws Exception {
     inputQueue.close();
-    task.close();
+    if (task != null) {
```
shuai-xu: why need this?

shuai-xu (changes requested):

KevinyhZou:
depends on pr: https://github.com/bigo-sg/velox/pull/9

KevinyhZou:
I have consider timestamp timezone in `date_format` function ut tests by using flink's `timestamp_ltz` type, please take a review at this, thanks. @PHILO-HE

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/config/VeloxQueryConfig.java
```diff
@@ -0,0 +1,50 @@
...
+  public static Config getConfig(RuntimeContext context) {
+    if (!(context instanceof StreamingRuntimeContext)) {
+      return Config.empty();
+    }
+    Configuration config = ((StreamingRuntimeContext) context).getJobConfiguration();
+    Map<String, String> configMap = new HashMap<>();
+    configMap.put(keyVeloxAdjustTimestampToSessionTimeZone, "true");
+    String localTimeZone = config.get(TableConfigOptions.LOCAL_TIME_ZONE);
+    if (TableConfigOptions.LOCAL_TIME_ZONE.defaultValue().equals(localTimeZone)) {
+      configMap.put(keyVeloxSessionTimezone, "UTC");
```
philo-he, line 46: Could you explain this code?

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -190,4 +196,45 @@ void testDecimal() {
...
   }
+
+  @Test
+  void testDateFormat() {
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+    List<Row> rows =
+        Arrays.asList(
+            Row.of(1, LocalDateTime.parse("2024-12-31 12:12:12", formatter)),
+            Row.of(2, LocalDateTime.parse("2025-02-28 12:12:12", formatter)));
+    createSimpleBoundedValuesTable("dateFormatTbl", "a int, b Timestamp(3)", rows);
```
philo-he: Nit:
Use descriptive table name: timestampTable,

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -190,4 +196,45 @@ void testDecimal() {
...
+    Map<String, String> configs = new HashMap<>();
+    configs.put(TableConfigOptions.LOCAL_TIME_ZONE.key(), "America/Los_Angeles");
+    rows =
+        Arrays.asList(
+            Row.of(
+                1, LocalDateTime.parse("2024-12-31 12:12:12", formatter).toInstant(ZoneOffset.UTC)),
+            Row.of(
+                2,
+                LocalDateTime.parse("2025-02-28 12:12:12", formatter).toInstant(ZoneOffset.UTC)));
+    createSimpleBoundedValuesTable("dateFormatTblLTZ", "a int, b Timestamp_LTZ(3)", rows);
```
philo-he: dateFormatTblLTZ->timestampLtzTable

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -190,4 +196,45 @@ void testDecimal() {
...
+    String query1 =
```
philo-he: Nit: maybe, just use "query = " to overwrite the previous value.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -190,4 +196,45 @@ void testDecimal() {
...
+        Arrays.asList(
+            Row.of(1, LocalDateTime.parse("2024-12-31 12:12:12", formatter)),
+            Row.of(2, LocalDateTime.parse("2025-02-28 12:12:12", formatter)));
+    createSimpleBoundedValuesTable("dateFormatTbl", "a int, b Timestamp(3)", rows);
+    String query =
+        "select a, DATE_FORMAT(b, 'yyyy-MM-dd'), DATE_FORMAT(b, 'yyyy-MM-dd HH:mm:ss') from dateFormatTbl";
+    runAndCheck(
+        query,
+        Arrays.asList(
+            "+I[1, 2024-12-31, 2024-12-31 12:12:12]", "+I[2, 2025-02-28, 2025-02-28 12:12:12]"));
```
philo-he, line 252: I assume the result will not be impacted by local time zone. Could you add a repeat test with a local time zone set firstly?

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/ScalarFunctionsTest.java
```diff
@@ -190,4 +196,45 @@ void testDecimal() {
...
+
+    Map<String, String> configs = new HashMap<>();
+    configs.put(TableConfigOptions.LOCAL_TIME_ZONE.key(), "America/Los_Angeles");
```
philo-he: I note in Flink, the following code can be called to set timezone. It may be better to directly call such API to do the setting before the test. Then, no need to do the setting inside `runAndCheck`. 
```java
 tEnv.getConfig().setLocalTimeZone(ZoneId.of("Asia/Shanghai"));
```

PHILO-HE (commented):
Thank for your updating. The added tests are what I wanted, which ensures timezone is actually respected. I posted only few minor comments.

BTW, ~are there any other code changes required to map Flink's `timestampLtz` type to Velox's `timestamp` type~? Could you verify the query is offloaded to Gluten with this PR? Thanks.

I note the type mapping already exists. It would be better to add some comments to clarify.
https://github.com/apache/incubator-gluten/blob/main/gluten-flink/runtime/src/main/java/org/apache/gluten/util/LogicalTypeConverter.java#L109-L111

KevinyhZou:
The query in the test would be offloaded to Velox backend，The source table would be mapped to `velox`'s `FromElementsSource` by using the `VeloxSourceBuidler`, which had been introduced this feature in pr: https://github.com/apache/incubator-gluten/pull/10362, and the `FromElementsSource` would convert the input timestamp string to velox's timestamp value with respecting to  the timezone value, which was introduced by: https://github.com/bigo-sg/velox/pull/9. @PHILO-HE

PHILO-HE (approved):
Looks good. I note Flink CI timed out and didn't run successfully. Maybe, a separate PR is required to fix it firstly.

philo-he:
> Looks good. I note Flink CI timed out and didn't run successfully. Maybe, a separate PR is required to fix it firstly.

@KevinyhZou, please first confirm if the CI failure is related to the PR.

KevinyhZou:
I think the failure reason is the pr: https://github.com/bigo-sg/velox/pull/9 that not merged. But from the ci log, it seems to be was blocked for 4 hours, and then was cancelled. @PHILO-HE . Please help to reiew:https://github.com/bigo-sg/velox/pull/9 @shuai-xu , and merge it first if it's no problems.

philo-he:
@KevinyhZou, you can temporarily change the CI code to use your Velox branch for testing, then to see if your patch can fix this failure.

Another question.
With the code below that maps Flink Timestamp type to Velox Timestamp type, I am curious how to distinguish timestamp with timezone and timestamp without timezone at Velox side, given that Flink LocalZonedTimestampType is mapped to Velox Timestamp type also.

https://github.com/apache/incubator-gluten/blob/4d2a200c45d6a8849124d6eed4d54f1ae48568fd/gluten-flink/runtime/src/main/java/org/apache/gluten/util/LogicalTypeConverter.java#L67-L69

KevinyhZou:
> @KevinyhZou, you can temporarily change the CI code to use your Velox branch for testing, then to see if your patch can fix this failure.
> 
> Another question. With the code below that maps Flink Timestamp type to Velox Timestamp type, I am curious how to distinguish timestamp with timezone and timestamp without timezone at Velox side, given that Flink LocalZonedTimestampType is mapped to Velox Timestamp type also.
> 
> https://github.com/apache/incubator-gluten/blob/4d2a200c45d6a8849124d6eed4d54f1ae48568fd/gluten-flink/runtime/src/main/java/org/apache/gluten/util/LogicalTypeConverter.java#L67-L69

Timestamp with timezone would be specified with `session_timezone` config, while Timestamp without timezone will not be. @PHILO-HE

KevinyhZou:
It seems something wrong with the ci, I use my own branch, and it still sutcked, it seems to have same problem as described in issue: https://github.com/apache/incubator-gluten/issues/10258. @PHILO-HE

philo-he:
> It seems something wrong with the ci, I use my own branch, and it still sutcked, it seems to have same problem as described in issue: #10258. @PHILO-HE

@KevinyhZou, please add the below command in flink.yml to install a new version of tzdata. We have merged some patch in upstream Velox to remove the dependency on new tzdata, and we've also removed the installation for new tzdata from docker image. I assume an old Velox version is used by Gluten Flink. So before the upgrade, we have to install new tzdata in runtime environment (not only the requirement for CI), as long as timezone is used in Velox.

```
yum install https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/tzdata-2025a-1.el9.noarch.rpm -y
```

philo-he:
> > @KevinyhZou, you can temporarily change the CI code to use your Velox branch for testing, then to see if your patch can fix this failure.
> > Another question. With the code below that maps Flink Timestamp type to Velox Timestamp type, I am curious how to distinguish timestamp with timezone and timestamp without timezone at Velox side, given that Flink LocalZonedTimestampType is mapped to Velox Timestamp type also.
> > https://github.com/apache/incubator-gluten/blob/4d2a200c45d6a8849124d6eed4d54f1ae48568fd/gluten-flink/runtime/src/main/java/org/apache/gluten/util/LogicalTypeConverter.java#L67-L69
> 
> Timestamp with timezone would be specified with `session_timezone` config, while Timestamp without timezone will not be. @PHILO-HE

@KevinyhZou, would you mind clarifying how this works if both Flink's timestamp and timestampLTZ are used in one session, even in one query?

KevinyhZou:
Both `Timestamp` and `TimestampLTZ` field in flink query will be converted to string and then passed to velox's `FromElementSource` to convert to timestamp of velox internal. Different from `Timestamp` field string, the `TimestampLTZ` field string will be end with a ‘utc’ timezone character `Z`, such as `2025-08-29T12:12:12Z`, and it will be converted to timestamp with timezone by using `TimestampConversion#fromTimestampWithTimezoneString` in vlelox, on the contrast, the `Timestamp` field string will be converted to timestamp without timezone in the same way.  And then `TimestampConversion#fromParsedTimestampWithTimeZone` will make different behavior to the different converted timestamps according to whether they have timezone and the timezone that in `session_timezone` config. @PHILO-HE

philo-he:
@KevinyhZou, thanks for the clarification. My understanding is that TimestampLTZ data is internally represented as a UTC timestamp, even though it is associated with a specific local timezone, which is different from Timestamp data.

PHILO-HE (approved):
Looks good. Thanks for your continuous efforts. Please revert the changes for Velox repo in flink.yml.

KevinyhZou:
done

philo-he:
@shuai-xu, do you have any comments?

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/operators/GlutenVectorTwoInputOperator.java
```diff
@@ -94,11 +94,13 @@ public GlutenVectorTwoInputOperator(
   private void initGlutenTask() {
     memoryManager = MemoryManager.create(AllocationListener.NOOP);
     session = Velox4j.newSession(memoryManager);
-    query = new Query(glutenPlan, Config.empty(), ConnectorConfig.empty());
+    query =
```
shuai-xu, line 97: It seems missing the GlutenVectorOneInputOperator

shuai-xu (changes requested):

gluten-flink/runtime/src/main/java/org/apache/gluten/table/runtime/operators/GlutenVectorTwoInputOperator.java
KevinyhZou: fixed

philo-he:
@KevinyhZou, could you check if the CI failure is related?

KevinyhZou:
velox4j has not been updated yet，and we will re-run the ci after it update. @PHILO-HE

Issue #9456, opened 2025-04-28
[VL] Memory Leak possibly TableScan


nimesh1601:
### Backend

VL (Velox)

### Bug description

There seems to be a memory leak from Table Scan
Got this response from Velox team : [Velox Issue](https://github.com/facebookincubator/velox/issues/13168#issuecomment-2835696507)

Sample query
WITH
 
cdl_summary AS (
  SELECT
    user_id,
    CASE
      WHEN <TRACKING_LABEL_COL> = 'store_front' THEN 'storefront'
      ELSE <TRACKING_LABEL_COL>
    END AS source,
     
    SUM(CASE WHEN name IN (<IMPRESSION_EVENTS>)
         AND datestr BETWEEN :START_7D   AND :END_DATE THEN 1 ELSE 0 END) AS impression_count_7d,
    SUM(CASE WHEN name IN (<IMPRESSION_EVENTS>)
         AND datestr BETWEEN :START_14D  AND :END_DATE THEN 1 ELSE 0 END) AS impression_count_14d,
    SUM(CASE WHEN name IN (<IMPRESSION_EVENTS>)
         AND datestr BETWEEN :START_28D  AND :END_DATE THEN 1 ELSE 0 END) AS impression_count_28d,
    SUM(CASE WHEN name IN (<IMPRESSION_EVENTS>)
         AND datestr BETWEEN :START_56D  AND :END_DATE THEN 1 ELSE 0 END) AS impression_count_56d,
    SUM(CASE WHEN name IN (<IMPRESSION_EVENTS>)
         AND datestr BETWEEN :START_112D AND :END_DATE THEN 1 ELSE 0 END) AS impression_count_112d,
     
    SUM(CASE WHEN name IN (<CLICK_EVENTS>)
         AND datestr BETWEEN :START_7D   AND :END_DATE THEN 1 ELSE 0 END) AS click_count_7d,
    SUM(CASE WHEN name IN (<CLICK_EVENTS>)
         AND datestr BETWEEN :START_14D  AND :END_DATE THEN 1 ELSE 0 END) AS click_count_14d,
    SUM(CASE WHEN name IN (<CLICK_EVENTS>)
         AND datestr BETWEEN :START_28D  AND :END_DATE THEN 1 ELSE 0 END) AS click_count_28d,
    SUM(CASE WHEN name IN (<CLICK_EVENTS>)
[...]

PR #10249, opened 2025-07-23
[VL] Increase the frequency of calling tryDestructSafe, and make one final attempt at the end

j7nhai:

(Fixes: \#ISSUE-ID)

## How was this patch tested?

### DIFF
--- a/cpp/velox/memory/VeloxMemoryManager.cc
+++ b/cpp/velox/memory/VeloxMemoryManager.cc
@@ -471,7 +471,6 @@ VeloxMemoryManager::~VeloxMemoryManager() {
   if (!destructed) {
     LOG(ERROR) << "Failed to release Velox memory manager after " << accumulatedWaitMs
                << "ms as there are still outstanding memory resources. ";
-    finalDestructUnsafe();
   }
 #ifdef ENABLE_JEMALLOC_STATS
   malloc_stats_print(NULL, NULL, NULL);


cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -444,6 +444,13 @@ bool VeloxMemoryManager::tryDestructSafe() {
   return true;
 }
 
+void VeloxMemoryManager::finalDestructUnsafe() {
+  heldVeloxPools_.clear();
+  veloxLeafPool_.reset();
+  veloxAggregatePool_.reset();
+  veloxMemoryManager_.reset();
```
zhztheplayer: This destruction order should already be guaranteed as it's the reversed order where they are defined. Are we missing something here?

j7nhai: You are right, they will be destructed in order. However, I did find many cases where the Velox memory manager had already been destructed when dropPool was called. Maybe we need to perform tryDestructSafe one more time at the end. Since the sleep time keeps multiplying by 1.5, the last sleep can be over ten seconds. Doing one more tryDestructSafe at the end can help avoid unnecessary waiting. At the same time, I have increased the frequency of tryDestructSafe calls.

zhztheplayer: Thank you for the explanation. Would like to know how the change is currently tested from your end. Is the issue reproducible locally?

I assume after some certain patches were landed into Velox, there's barely such case as the memory manager needs to wait when being destructed. But if the destruction is triggered by an error thrown from task, waiting may still be needed. Like to know more about your case as well.

j7nhai: I cannot reproduce this problem locally, it is triggered by our actual production tasks. After such a small optimization, the stability can be improved a lot. Extreme tasks also need to increase `spark.gluten.sql.columnar.backend.velox.asyncTimeoutOnTaskStopping`, so I think it is necessary to limit the interval to a maximum of 2s, otherwise the spark task will have to wait for a long time after several attempts.

j7nhai:
I have a question. When the manager exits, it means that task has finished computing, and it also means that all the data in the memory pools can be released, so the memory pools can be freed. Why is it still necessary to wait in this case? @zhztheplayer

zhztheplayer:
@j7nhai 

>  Why is it still necessary to wait in this case?

We used to have some issues causing memory leak even task has exited. For instance, https://github.com/facebookincubator/velox/issues/8198.

But since we have already solved them I think the memory leak should not happen anymore, ideally.

Are you using the newest Gluten build?

j7nhai:
> @j7nhai
> 
> > Why is it still necessary to wait in this case?
> 
> We used to have some issues causing memory leak even task has exited. For instance, [facebookincubator/velox#8198](https://github.com/facebookincubator/velox/issues/8198).
> 
> But since we have already solved them I think the memory leak should not happen anymore, ideally.
> 
> Are you using the newest Gluten build?

I use the release 1.4.0 and when tryDestructSafe fail, it shows that ` veloxMemoryManager_->numPools()` is 4, with the INFO log:

```
Attempt to destruct VeloxMemoryManager failed because there are 4 outstanding memory pools.
```

j7nhai:
> @j7nhai
> 
> > Why is it still necessary to wait in this case?
> 
> We used to have some issues causing memory leak even task has exited. For instance, [facebookincubator/velox#8198](https://github.com/facebookincubator/velox/issues/8198).
> 
> But since we have already solved them I think the memory leak should not happen anymore, ideally.
> 
> Are you using the newest Gluten build?

@zhztheplayer 


Here is the usage leak log:
```
E20250814 17:32:41.672763   215 Memory.cpp:161] pools_.size() != 0 (1 vs 0). There are unexpected alive memory pools allocated by user on memory manager destruction:
Memory Manager[capacity UNLIMITED alignment 64B usedBytes 0B number of pools 4
List of root pools:
__sys_root__ usage 0B reserved 0B peak 0B
    __sys_shared_leaf__0 usage 0B reserved 0B peak 0B
    __sys_tracing__ usage 0B reserved 0B peak 0B
    __sys_caching__ usage 0B reserved 0B peak 0B
    __sys_spilling__ usage 0B reserved 0B peak 0B
root usage 0B reserved 0B peak 8.00MB
    task.Gluten_Stage_2_TID_1959_VTID_12 usage 0B reserved 0B peak 8.00MB
        node.0 usage 0B reserved 0B peak 7.00MB
            op.0.0.0.TableScan usage 0B reserved 0B peak 6.53MB
        refcount 2
```

it looks that the pool of TableScan does not release.

And it also has a warning log in JVM.

```
25/08/14 17:32:41 WARN NativeMemoryManager: NativePlanEvaluator-12 Reservation listener org.apache.gluten.memory.listener.ManagedReservationListener@4fff3055 still reserved non-zero bytes, which may cause memory leak, size: 8.0 MiB.
```

j7nhai:
Hi @zhztheplayer 

By referring to https://github.com/apache/incubator-gluten/issues/9456#issuecomment-2841986145, I set

```
spark.gluten.sql.columnar.backend.velox.IOThreads=0
```

And it run ok now. However, it looks there is memory leak in TableScan (maybe from the dataSource_?). How do I debug?

zhztheplayer:
@j7nhai Usually the IO threads will be closed via `Task::requestCancel` with blocking:

https://github.com/apache/incubator-gluten/blob/d76865ab73f39d074806600e209da6ae8ed10fa3/cpp/velox/compute/WholeStageResultIterator.h#L45-L50

j7nhai:
@zhztheplayer This is actually a bug in Velox, triggered during data prefetching. I fixed it in PR: https://github.com/facebookincubator/velox/pull/14722. Please review it.

PR #9572, opened 2025-05-08
[GLUTEN-9571][VL] Respect parquet configs, parquet.page.size and parquet.compression.codec.zstd.level etc.

WangGuangxin:
## What changes were proposed in this pull request?

Respect parquet writer configs
1. parquet.page.size
2. parquet.compression.codec.zstd.level
3. parquet.enable.dictionary
4. parquet.writer.version

(Fixes: \#9571)

## How was this patch tested?

manually

PR #10259, opened 2025-07-24
[CORE] Gen config also verify config as expected

yikf:
https://github.com/apache/incubator-gluten/pull/9572 added respected parquet configurations, which were not generated by the configuration generation mechanism. surprisingly, the check mechanism passed. this was because the check mechanism incorrectly handled the row count check. this PR fixes this issue.

And this PR temporarily remove the section on the respected parquet configuration description. these configurations should at least be in separate files or be generated automatically.


yikf:
cc @zhouyuan @WangGuangxin @FelixYBW please take a look if you find a time, thanks!

FelixYBW:
The purpose of parquet writer section is to list all the spark configurations which are hornored or not by Gluten. We can refer to the table here: https://spark.apache.org/docs/3.5.0/configuration.html

But since current configuration.md is auto generated for Gluten specific configuration only. We may introduce a new spark-configurations.md file to list the spark's.

FelixYBW (approved):

zhouyuan (approved):
Thanks!

Issue #10170, opened 2025-07-11
[VL] For Try Functions it should not matter whether ansi is on or off


nimesh1601:
### Backend

VL (Velox)

### Bug description

We throw [error](https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L118) on try arithmetic functions if ansi is enabled, but try functions return the same result whether ansi is enabled or not. Am I missing something here ?

### Gluten version

main branch

### Spark version

None

### Spark configurations

NA

### System information

NA

### Relevant logs

```bash
NA
```

nimesh1601:
@PHILO-HE any idea ?

philo-he:
@zhli1142015, do you have any comment?

nimesh1601:
I tried out by removing this check , seems to be working fine. I will raise a PR with the changes, you can suggest in case I have missed anything

PR #10267, opened 2025-07-28
[GLUTEN-10170][VL] Offload `try` arithmetic functions regardless of ANSI configuration

nimesh1601:

Remove ansi check for try eval, as try functions work the same whether ANSI is on or off

(Fixes: \#10170)

## How was this patch tested?

Uts and production queries


nimesh1601:
cc: @zhli1142015 / @PHILO-HE

FelixYBW:
Have you tested Spark UT? If UT passed, let's enable the UT.

nimesh1601:
@FelixYBW, all tests of GlutenTryEvalSuite are already included. No tests were excluded. I checked [VeloxTestSettings.scala](https://github.com/apache/incubator-gluten/blob/main/gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala#L191). 

In the test logs, I can see 
```
2025-07-29T00:27:07.1611958Z - try_divide
2025-07-29T00:27:07.2195540Z 'try_subtract' offload to gluten
2025-07-29T00:27:07.2196030Z - try_subtract
2025-07-29T00:27:07.3096229Z 'try_multiply' offload to gluten
2025-07-29T00:27:07.3098568Z - try_multiply
2025-07-29T00:27:07.3701933Z 'Throw exceptions from children' offload to gluten
2025-07-29T00:27:07.3702425Z - Throw exceptions from children
2025-07-29T00:27:07.3734960Z 00:27:07.373 WARN org.apache.spark.sql.catalyst.expressions.GlutenTryEvalSuite:
```

But while testing for production workload, I also removed the ANSI fallback check in [VeloxRuleApi.scala](https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala#L68)

Current Test failures seems unrelated 
`SPARK-19059: read file based table whose name starts with underscore *** FAILED ***.`

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -115,10 +115,6 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
       right: ExpressionTransformer,
       original: TryEval,
       checkArithmeticExprName: String): ExpressionTransformer = {
-    if (SparkShimLoader.getSparkShims.withAnsiEvalMode(original.child)) {
```
zhli1142015, line 119: I assume that by removing this conditional check, the original try_xx function—which used to fall back—will now be offloaded to Velox. Could you add corresponding tests? Something like the example below, to make sure they actually get offloaded to Velox.
https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/test/scala/org/apache/gluten/functions/MathFunctionsValidateSuite.scala#L302

nimesh1601: @zhli1142015 I think it still won't get offloaded to Velox as we have a fallback rule for ANSI in [VeloxRuleApi.scala](https://github.com/apache/incubator-gluten/blob/main/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala#L68). Even while testing in production, I needed to remove this rule, so testing by query might not work. Should I add a config that skips this fallback, maybe something like spark.gluten.sql.ansi.forceEnabled ? This will also allow other ansi supported operators to work with ansi too. wdyt ?

nimesh1601: @zhli1142015 can you review ?

zhli1142015 (approved):
Thanks.

PR #9396, opened 2025-04-22
[GLUTEN-9392][VL] Support casting integral types to double for array elements

ArnavBalyan:
## What changes were proposed in this pull request?
 - Support array cast for integrals to decimals.

## How was this patch tested?
 - Unit tests added.
 - Existing Spark tests.

PR #9394, opened 2025-04-22
[GLUTEN-9392] [VL] Support casting array element to varchar

ArnavBalyan:
## What changes were proposed in this pull request?
 - Support array cast to varchar. 
 - More cast types support will be added in the upcoming PRs.

## How was this patch tested?
 - Unit tests added
 - Spark existing unit tests

PR #10270, opened 2025-07-29
[VL] Support casting array element from more types to varchar

zml1206:

Support casting array element from integer/long/date to varchar, and remove some duplicate tests.

## How was this patch tested?

UT

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
@@ -1156,6 +1156,25 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
     runQueryAndCompare("select cast('123.0' AS INT)") {
       checkGlutenOperatorMatch[ProjectExecTransformer]
     }
+    // Cast Array as Array[String]
+    runQueryAndCompare("select cast(array(1, null) AS array<string>)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
+    runQueryAndCompare("select cast(array(1L, null) AS array<string>)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
+    runQueryAndCompare("select cast(array(1.1d, null) AS array<string>)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
+    runQueryAndCompare("select cast(array(false, null) AS array<string>)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
+    runQueryAndCompare("select cast(array(date'2024-01-01') AS array<string>)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
+    runQueryAndCompare("select cast(array(timestamp'2024-01-01 12:00:00') AS array<string>)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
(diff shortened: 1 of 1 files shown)


zhouyuan:
Is this pattern a real usage in production? `varchar(n)` seems marked as experimental in current Spark

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenCastSuite.scala
```diff
@@ -43,69 +43,6 @@ class GlutenCastSuite extends CastSuite with GlutenTestsTrait {
   UDTRegistration.register(classOf[IExampleBaseType].getName, classOf[ExampleBaseTypeUDT].getName)
   UDTRegistration.register(classOf[IExampleSubType].getName, classOf[ExampleSubTypeUDT].getName)
 
-  test("cast array element from integer to string") {
```
philo-he, line 46: Could you confirm why these tests were added before? Are they duplicate to the existing Spark tests?

zml1206:
> Is this pattern a real usage in production? `varchar(n)` seems marked as experimental in current Spark

Here varchar is java String. `cast(c1 as Array<String>)` if c1 is Array[Integer/Long/Date].

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenCastSuite.scala
zml1206: Deleted UT and "cast array of integer types to array of double" are duplicates. They are added in #9394 and #9396. They are unrelated to the existing Spark tests.

zml1206: @PHILO-HE Do you have any other comments? I would like to quickly support `cast(array as string)` based on this, velox does not support cast Array to String. I will add rule to replace `cast(array as String)` with `concat('[', array_join(array, ', ', 'null'), ']')` to support offload. Thank you.

philo-he: @zml1206, the patch looks good. Just curious why not adding the direct conversion in Velox, which should perform better than the proposed replacement.

zml1206: I am not familiar with the `cast` code of velox. Velox's merger cycle is also relatively long. The conversion performance loss is not large. This is just a temporary quick support idea.

philo-he: I see. But I still encourage you to try to submit a PR to Velox community. And once the patch is ready, you can add the PR ID in the code below to include the feature for your Gluten.

https://github.com/apache/incubator-gluten/blob/29c5e0050c741ddac560db3e7acd83548e402f32/ep/build-velox/src/get_velox.sh#L28

zml1206: OK, I'll first propose a PR to rewrite cast to quickly support it, and then I'll try to add support to velox when I have time.

zml1206:
Thanks for review @PHILO-HE @zhouyuan,  merge to main.

PR #10285, opened 2025-07-30
[GLUTEN-9366][VL] Support Iceberg functions

jinchengchenghh:


gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -183,6 +208,8 @@ object ExpressionConverter extends SQLConfHelper with Logging {
           replaceWithExpressionTransformer0(i.arguments.head, attributeSeq, expressionsMap),
           i
         )
+      case i @ StaticInvoke(_, _, "invoke", _, _, _, _, _) =>
+        return replaceIcebergStaticInvoke(i, attributeSeq, expressionsMap)
```
zhztheplayer: It looks like more accurate to match on `org.apache.iceberg.spark.functions.*` than on `invoke` to forward to a method as `replaceIcebergStaticInvoke`?

zhztheplayer (approved):
👍 Thanks!

PR #10300, opened 2025-07-31
[VL] Support cast from array to string

zml1206:

Velox does not support cast Array to String. Before velox support, temporarily add a rule to rewrite cast to support it.

## How was this patch tested?

ut.


zml1206:
@PHILO-HE Can you help take a look? Thanks.

backends-velox/src/main/scala/org/apache/gluten/extension/RewriteCastFromArray.scala
```diff
@@ -0,0 +1,64 @@
...
+      SQLConf.get.getConf(SQLConf.LEGACY_COMPLEX_TYPES_TO_STRING)
+    ) {
+      return plan
+    }
+    plan.transformUpWithPruning(_.containsPattern(CAST)) {
+      case p =>
+        p.transformExpressionsUpWithPruning(_.containsPattern(CAST)) {
+          case cast @ Cast(child, StringType, timeZoneId, evalMode)
+              if child.dataType.isInstanceOf[ArrayType] =>
+            child.dataType match {
```
philo-he: Using array's element type to match for brevity.

backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
```diff
@@ -1175,6 +1175,25 @@ abstract class ScalarFunctionsValidateSuite extends FunctionsValidateSuite {
...
+    }
+    runQueryAndCompare("select cast(array(false, null) AS string)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
+    runQueryAndCompare("select cast(array(date'2024-01-01') AS string)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
+    runQueryAndCompare("select cast(array(timestamp'2024-01-01 12:00:00') AS string)") {
+      checkGlutenOperatorMatch[ProjectExecTransformer]
+    }
```
philo-he, line 1209: Please add a test for empty array.

docs/Configuration.md
```diff
@@ -156,6 +156,7 @@ nav_order: 15
...
+| spark.gluten.sql.rewrite.castArrayToString                         | true              | When true, rewrite `cast(array as String)` to `concat('[', array_join(array, ', ', null), ']')` to make velox support offload.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
```
philo-he: ** to allow offloading to Velox.

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1418,6 +1421,14 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val ENABLE_REWRITE_CAST_ARRAY_TO_STRING =
+    buildConf("spark.gluten.sql.rewrite.castArrayToString")
+      .internal()
+      .doc("When true, rewrite `cast(array as String)` to" +
+        " `concat('[', array_join(array, ', ', null), ']')` to make velox support offload.")
```
philo-he: ditto.

philo-he (commented):
@taiyang-li, @exmy, does CH backend need this rule as a workaround support?

If so, this rule can be moved to a common place. Otherwise, keep it in velox module and move new config to VeloxConfig.

zml1206:
> @taiyang-li, @exmy, does CH backend need this rule as a workaround support?
> 
> If so, this rule can be moved to a common place. Otherwise, keep it in velox module and move new config to VeloxConfig.

I think it is not reusable. I moved new config to VeloxConfig.

exmy:
> @taiyang-li, @exmy, does CH backend need this rule as a workaround support?
> 
> If so, this rule can be moved to a common place. Otherwise, keep it in velox module and move new config to VeloxConfig.

CH backend supports cast array to string, so don't need this rule.

backends-velox/src/main/scala/org/apache/gluten/extension/RewriteCastFromArray.scala
```diff
@@ -0,0 +1,62 @@
...
+              case StringType =>
+                val arrayJoin = ArrayJoin(child, Literal(", "), Some(Literal("null")))
+                Concat(Seq(Literal("["), arrayJoin, Literal("]")))
+              case IntegerType | LongType | DoubleType | BooleanType | DateType | TimestampType =>
+                val arrayJoin = ArrayJoin(
+                  Cast(child, ArrayType(StringType), timeZoneId, evalMode),
+                  Literal(", "),
+                  Some(Literal("null")))
+                Concat(Seq(Literal("["), arrayJoin, Literal("]")))
+              case _ => cast
```
philo-he: Seems Velox's array cast support is not limited to these types. Maybe, just add a cast for non-String type and let unsupported case fall back in the later validation phase.

``` scala
val joinChild = child.dataType.asInstanceOf[ArrayType].elementType match {
     case StringType =>
          child
     case _ =>
          Cast(child, ArrayType(StringType), timeZoneId, evalMode)
}
val arrayJoin = ArrayJoin(joinChild , Literal(", "), Some(Literal("null")))
Concat(Seq(Literal("["), arrayJoin, Literal("]")))
```

zml1206: Good catch! Spark Cast from Array to String reused `castToString`.

philo-he: Can we just use the above proposed code for brevity?

philo-he (approved):
Looks good. Thanks.

zml1206:
@PHILO-HE I fixed the spark ut, do you have any other comments? Thanks.

philo-he (approved):
Thanks.

jiangjiangtian:
@zml1206 Sorry to bother you. I am wondering why we need to disable this rule when config `LEGACY_COMPLEX_TYPES_TO_STRING` is true. It seems that the config just controls [casting array and map to string ](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L7378). Is there anything that I miss?

zml1206:
> @zml1206 Sorry to bother you. I am wondering why we need to disable this rule when config `LEGACY_COMPLEX_TYPES_TO_STRING` is true. It seems that the config just controls [casting array and map to string ](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L7378). Is there anything that I miss?

```
LEGACY=false: cast(array(1, null, 2) as string) => [1, null, 2]
LEGACY=true:  cast(array(1, null, 2) as string) => [1,, 2]
```

jiangjiangtian:
> > @zml1206 Sorry to bother you. I am wondering why we need to disable this rule when config `LEGACY_COMPLEX_TYPES_TO_STRING` is true. It seems that the config just controls [casting array and map to string ](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L7378). Is there anything that I miss?
> 
> ```
> LEGACY=false: cast(array(1, null, 2) as string) => [1, null, 2]
> LEGACY=true:  cast(array(1, null, 2) as string) => [1,, 2]
> ```

Oh, I see! Thanks for your reply!

PR #10303, opened 2025-07-31
[VL] Fix gen all config

zml1206:

```
dev/gen_all_config_docs.sh
[INFO] Error stacktraces are turned on.
[INFO] Scanning for projects...
[ERROR] [ERROR] Could not find the selected project in the reactor: :backends-velox @ 
[ERROR] Could not find the selected project in the reactor: :backends-velox -> [Help 1]
org.apache.maven.MavenExecutionException: Could not find the selected project in the reactor: :backends-velox
    at org.apache.maven.graph.DefaultGraphBuilder.trimSelectedProjects (DefaultGraphBuilder.java:148)
```

The backends-velox module is activated via Maven Profile.

## How was this patch tested?

Local test.

### DIFF
--- a/dev/gen_all_config_docs.sh
+++ b/dev/gen_all_config_docs.sh
@@ -18,7 +18,7 @@
 
 GLUTEN_UPDATE="${GLUTEN_UPDATE:-1}" \
 mvn clean test \
-  -pl gluten-substrait -pl backends-velox -am \
+  -pl gluten-substrait -P backends-velox -am \
   -Dtest=none \
   -DfailIfNoTests=false \
   -DwildcardSuites=org.apache.gluten.config.AllGlutenConfiguration,org.apache.gluten.config.AllVeloxConfiguration


zml1206:
cc @zhouyuan @yikf Thanks

zhouyuan (approved):
Thanks!

dev/gen_all_config_docs.sh
```diff
@@ -18,7 +18,7 @@
 
 GLUTEN_UPDATE="${GLUTEN_UPDATE:-1}" \
 mvn clean test \
-  -pl gluten-substrait -pl backends-velox -am \
+  -pl gluten-substrait -Pbackends-velox -am \
```
yikf: should also contain `-pl backends-velox`? since the gen velox config in this module

zhztheplayer: And usually we shouldn't have to specify `-pl gluten-substrait` if we already set `-pl backends-velox -am`. No?

zml1206: Updated, thanks.

zhouyuan: is this the right syntax? 
` -Pbackends-velox -pl backends-velox -am \`

zhouyuan:
@zml1206 thanks! merge into main.

dev/gen_all_config_docs.sh
yikf: @zhztheplayer Right, but at least it's `-Pbackends-velox -pl backends-velox -am \`, as @zhouyuan said.

All in all, it's a small point. Thank you all!

Issue #10317, opened 2025-08-01
[FLINK] Support state for window and agg operator


shuai-xu:
### Description

Flink window operator and agg operator need state to buffer data and the calculate them. We should support state related operation in velox.

### Gluten version

None

PR #10320, opened 2025-08-01
[Gluten-10317][FLINK] Support state related operation

shuai-xu:

This pr update velox to a version, which contains support for state. And make serval operators contains Join, WindowJoin, GlobalAgg, LocalAgg use state for calculation. These operators's bahivor is like flink now, but still not the same.

(Fixes: \#10317)

## How was this patch tested?


.github/workflows/flink.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           source /opt/rh/gcc-toolset-11/enable
           sudo dnf install -y patchelf
           git clone -b gluten-0530 https://github.com/bigo-sg/velox4j.git
-          cd velox4j && git reset --hard 0eb9eef589692dbde953c36ecd2d8f9d3a34a59d
+          cd velox4j && git reset --hard 7bf3bd14624733230a395306275ccc3bed671b6c
```
zjuwangg: how about use tag instead of commit hash to track the version.

shuai-xu: Good idea, we now use a private branch and it does't update with main branch, we can use git tag when after merge with the oap velox.

gluten-flink/planner/src/main/java/org/apache/flink/table/planner/plan/nodes/exec/stream/StreamExecLocalWindowAggregate.java
```diff
@@ -157,30 +172,59 @@ protected Transformation<RowData> translateToPlanInternal(
...
+    // TODO: support more window types.
+    Tuple5<Long, Long, Long, Integer, Integer> windowSpecParams =
+        WindowUtils.extractWindowParameters(windowing);
+    long size = windowSpecParams.f0;
+    long slide = windowSpecParams.f1;
+    long offset = windowSpecParams.f2;
+    int rowtimeIndex = windowSpecParams.f3;
+    int windowType = windowSpecParams.f4;
+    PartitionFunctionSpec sliceAssignerSpec =
+        new WindowPartitionFunctionSpec(inputType, rowtimeIndex, size, slide, offset, windowType);
```
philo-he: I note Flink's window is completely different from Spark SQL's. It would be better to use a better name to easily distinguish in Velox4j and Velox, e.g., `StreamWindowPartitionFunctionSpec` looks better for clarity than `WindowPartitionFunctionSpec`. cc @zhztheplayer

KevinyhZou: It seems the `WindowPartitionFunctionSpec` is newly added in `bigo-sg` velox，and it’s specifically for flink to divide the window into different slices by the input `time` field，and spark window does not need this.  and I think it is better to use some name that can be identified as flink specific. @PHILO-HE @shuai-xu

shuai-xu: I have rename related classes to begin with Stream, please help to review again.

PR #10322, opened 2025-08-01
[VL] Remove RAS rule: RemoveSort

li-boxuan:
`RemoveSort` rule assumes if a Sort operator's parent doesn't require ordering, then it can be removed. This isn't always true, since in Spark there's no "required physical property" concept as in Cascades framework.

[51768](https://github.com/apache/spark/pull/51768) is one example. Although I tend to believe it's a Spark bug, it also reveals the fact that it's dangerous to hold this assumption in Gluten.

Motivation: when RAS is turned on, https://github.com/apache/spark/blob/92df232addc5653504ec788332b5d52f936c5184/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/AggregationQuerySuite.scala#L463-L466 won't pass because Gluten RAS removes the Sort. Of course, we can also say that's just a bad test, but its existence says something about what customers could do.


zhztheplayer (approved):
LGTM

zhztheplayer:

PR #10381, opened 2025-08-07
[GLUTEN-10377][CH] Move specialized delta internal column mapping to CH backend

zhztheplayer:
In preparation for supporting deletion vector in Velox backend, move the specialized column mapping `__delta_internal_is_row_deleted` -> `row_index` from common place to CH backend.


zhztheplayer:
@zzcclp ptal

zzcclp (approved):
LGTM

Issue #10421, opened 2025-08-12
[VL] Avoid fallback in Expand


wecharyu:
### Backend

VL (Velox)

### Bug description

`Expand` operator could fallback by following error:
```bash
== Fallback Summary ==
(6) Expand: Failed to bind reference for _pre_6#44282: Couldn't find _pre_6#44282 in [shop_id#18512L,seller_id#18521L,grass_region#18955,_groupingexpression#33997]; Failed to bind reference for _pre_6#44282: Couldn't find _pre_6#44282 in [shop_id#18512L,seller_id#18521L,grass_region#18955,_groupingexpression#33997]
```

**Root cause:**
`newProjections` is evaluated after `newProject` construction. Due to potential compiler reordering, `expressionMap` may not be populated when `newProject` is created, leading to an empty map and incorrect behavior.
https://github.com/apache/incubator-gluten/blob/5734fe71956916803e61014ce09f333b06e4caef/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala#L258-L273

### Spark version

None

### Relevant logs

```bash

```

PR #10422, opened 2025-08-12
[GLUTEN-10421][CORE] Force evaluation of newProjections to prevent empty expressionMap

wecharyu:

(Fixes: \#10421)

## How was this patch tested?
Add new unit test.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
@@ -265,6 +265,10 @@ object PullOutPreProject extends RewriteSingleNode with PullOutProjectHelper {
               expressionMap,
               replaceBoundReference = false,
               replaceLiteral = false)))
+
+      // Force evaluation so expressionMap gets populated
+      newProjections.foreach(_.length)
+
       val newProject = ProjectExec(
         eliminateProjectList(expand.child.outputSet, expressionMap.values.toSeq),
         expand.child)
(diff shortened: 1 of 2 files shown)


wForget:
FYI: a similar fix: https://github.com/apache/spark/pull/35537

wForget:
`replaceExpressionWithAttribute` seems to be called in multiple places, could you check where else there might be similar issues?

wecharyu:
Other calls on `replaceExpressionWithAttribute` not has this issue, for example there is `toIndexedSeq` for `agg.groupingExpressions`:
https://github.com/apache/incubator-gluten/blob/170590458f9b6125a11cc4f38919a04360697141/gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala#L173-L175

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/PullOutPreProject.scala
```diff
@@ -259,12 +259,13 @@ object PullOutPreProject extends RewriteSingleNode with PullOutProjectHelper {
       val expressionMap = new mutable.HashMap[Expression, NamedExpression]()
       val newProjections =
         expand.projections.map(
```
wForget: `expand.projections` should also use `toIndexedSeq`

wForget:

wForget:
> `replaceExpressionWithAttribute` seems to be called in multiple places, could you check where else there might be similar issues?

@wecharyu I found more similar cases https://github.com/wecharyu/incubator-gluten/pull/1

wecharyu:
> I found more similar cases https://github.com/wecharyu/incubator-gluten/pull/1

I have also checked these cases before and thought it would not be lazy. Anyway it's more safe to add the `toIndexedSeq`.

wForget:
Thanks, merged to main

PR #10435, opened 2025-08-14
[INFRA] Update .gitignore to track vcs.xml and remove duplicates

philo-he:
Reflected by title.

## How was this patch tested?
Not required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,51 +2,24 @@
 *.class
 *.iml
 .DS_Store
-
-# logs
 *.log
 **/core$
-
-# cache files
+hs_err_pid*
 *.cache
+*.war
+*.ear
+.clangd
 
-# MacOS dir files
-.DS_Store
-
-# IDEA config files
+# IDEA config
 .idea/
-*.iml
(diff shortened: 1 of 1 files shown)


philo-he:
@zhouyuan, please take a look.

.gitignore
```diff
@@ -2,52 +2,25 @@
...
-# MacOS dir files
-.DS_Store
-
-# IDEA config files
+# IDEA config
 .idea/
-*.iml
-*.class
-
+!.idea/vcs.xml
```
zhouyuan, line 15: I found this file will be updated if my local IntelliJ is upgraded. If someone updated this file, would it still work for others using old IntelliJ?

zhouyuan (approved):
👍  
Thanks for cleanup on this

zjuwangg (approved):

.gitignore
philo-he: I think in this file the structure and the components are standard. So as long as we keep using standard components, it should work across different Intellij versions.

PR #10443, opened 2025-08-14
[GLUTEN-9392][VL] Support casting complex types

kevinwilfong:

This adds general support for casting complex types.

Previously, casting of complex types was limited to arrays where the elements were casted from/to specific types. This 
change allows casting of arrays/maps/rows as long as all conversions of child types are generally supported in Gluten, 
including support for nested complex types.

From testing, the main blocker to supporting this was the different failure semantics between Spark and Velox. In particular
if a child conversion failed Velox would return NULL for the entire complex value, while Spark would simply make that child
NULL within the larger complex value. This was addressed in https://github.com/facebookincubator/velox/pull/14451

## How was this patch tested?

Spark already has existing tests for casting complex types in org.apache.spark.sql.catalyst.expressions.CastSuite I verified
these do not fallback, unless Gluten doesn't support casting the child types, and those tests pass.

I noticed there weren't many tests in Spark, and many of them involved casting child types Gluten doesn't support, so I
added additional unit tests in org.apache.gluten.expressions.GlutenCastSuite

I ran these tests against Spark versions 3.2-3.5

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/functions/ScalarFunctionsValidateSuite.scala
@@ -20,8 +20,10 @@ import org.apache.gluten.config.GlutenConfig
 import org.apache.gluten.execution.{BatchScanExecTransformer, FilterExecTransformer, ProjectExecTransformer}
 
 import org.apache.spark.{SparkConf, SparkException}
+import org.apache.spark.sql.Row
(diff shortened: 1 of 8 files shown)


zml1206 (commented):
Thank you for your contribution. Could you please improve the `cast` test in `ScalarFunctionsValidateSuite` to ensure that the newly added types are offloded to native?

kevinwilfong:
@zml1206 Thanks for taking a look! Added.

PHILO-HE (approved):
Looks good. Thanks.

PR #10446, opened 2025-08-15
[VL][TEST] test velox PR#14457

JkSelf:

test https://github.com/facebookincubator/velox/pull/14457

## How was this patch tested?

existing test

### DIFF
--- a/ep/build-velox/src/get_velox.sh
+++ b/ep/build-velox/src/get_velox.sh
@@ -16,8 +16,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=2025_08_14
+VELOX_REPO=https://github.com/JkSelf/velox.git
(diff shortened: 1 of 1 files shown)


ep/build-velox/src/get_velox.sh
```diff
@@ -16,8 +16,8 @@
 
 set -exu
 
-VELOX_REPO=https://github.com/oap-project/velox.git
-VELOX_BRANCH=2025_08_14
+VELOX_REPO=https://github.com/JkSelf/velox.git
+VELOX_BRANCH=test-14457
```
philo-he, line 20: @JkSelf, we can just set `UPSTREAM_VELOX_PR_ID` to test an upstream PR.

JkSelf: @PHILO-HE Can we directly set the `UPSTREAM_VELOX_PR_ID` if exist conflicts.

PR #10458, opened 2025-08-15
[GLUTEN-10457][VL] Iceberg supports copy on write

jinchengchenghh:
The only implementation difference with AppendDataExec is the BatchWrite commit function, but it is called by the interface, so we only apply the rule is enough.


zhouyuan (approved):
Thanks.

PR #10477, opened 2025-08-18
[GLUTEN-8953][VL] Fix iceberg codec uncompressed

jinchengchenghh:
Fix https://github.com/oap-project/velox/issues/1039

### DIFF
--- a/gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergAppendDataExec.scala
+++ b/gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergAppendDataExec.scala
@@ -44,7 +44,7 @@ trait IcebergAppendDataExec extends ColumnarAppendDataExec {
       case FileFormat.ORC => config.getOrDefault(ORC_COMPRESSION, ORC_COMPRESSION_DEFAULT)
       case _ => throw new UnsupportedOperationException()
     }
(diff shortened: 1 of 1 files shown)


jinchengchenghh:
@4ertus2 Could you help review? Thanks!

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergAppendDataExec.scala
```diff
@@ -44,7 +44,7 @@ trait IcebergAppendDataExec extends ColumnarAppendDataExec {
       case FileFormat.ORC => config.getOrDefault(ORC_COMPRESSION, ORC_COMPRESSION_DEFAULT)
       case _ => throw new UnsupportedOperationException()
     }
-    if (codec == "UNCOMPRESSED") {
+    if (codec == "UNCOMPRESSED" || codec == "uncompressed") {
```
marin-ma: use `toLowerCase` ?

4ertus2:
> @4ertus2 Could you help review? Thanks!

Unfortunately not. We've hold our work around Gluten/Velox till it has better iceberg support. Especially writing partitioned data.

We'd fixed it in Velox part making "uncompressed" an alias for "none". Your solution looks better. But I cannot test it until we will test partitions.

zhouyuan (approved):

PR #10479, opened 2025-08-18
[GLUTEN-8852][TEST] Support Spark 4.0 in gluten-it

### LINKED ISSUE #10438: Fix gluten-tool build for Spark-4.0
### Description

There are some compilation failures that blocking the use of gluten-tool for TPC-H test.

### Gluten version

None

PHILO-HE:

This PR makes the code under gluten-it compatible with Spark 4.0 and scala-2.13,
and also enables Spark-4.0 TPC-H test in CI.

Fixes #10438.

## How was this patch tested?
CI.


.github/workflows/velox_backend_x86.yml
```diff
@@ -242,22 +244,21 @@ jobs:
...
             $MVN_CMD clean install -P${{ matrix.spark }} -P${{ matrix.java }} -Pbackends-velox -DskipTests
           fi
           cd $GITHUB_WORKSPACE/tools/gluten-it
           $MVN_CMD clean install -P${{ matrix.spark }} -P${{ matrix.java }}
       - name: Run TPC-H / TPC-DS
         run: |
+          echo "JAVA_HOME: $JAVA_HOME"
+          cd $GITHUB_WORKSPACE/tools/gluten-it
+          GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
+            --local --preset=velox --benchmark-type=h --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1
```
zhouyuan, line 256: is there an option to enable partition table test here?

philo-he: Looks we can set this option for achieve that: https://github.com/apache/incubator-gluten/blob/8e2a897143c892979a166d4dff00c42c80d08f72/tools/gluten-it/common/src/main/java/org/apache/gluten/integration/command/DataGenMixin.java#L30

philo-he:
@zhztheplayer, could you take a look? Thanks.

zhztheplayer:
@PHILO-HE  it feels a bit off to me to introduce Gluten‘s shim layer to gluten-it, although I am sure there was a good motivation to do that. I will take a closer look tomorrow.

zhztheplayer (approved):
No idea how you got rid of the shims but thanks for iterating :)

philo-he:
> No idea how you got rid of the shims but thanks for iterating :)

I replaced `new SparkSession` with `SparkSession.builder()` to create the session instance. In Spark 4.0, the builder API is kept in `sql.SparkSession`, and the classic SparkSession can be created with it by default.

In addition, I removed `SparkSession.cleanupAnyExistingSession` which requires using classic SparkSession companion object, then requires shim layer to differentiate the type if we call it. It's redundant to the existing code which already covers the cleanup.

@zhztheplayer, @zhouyuan, thanks for the review.

PR #10503, opened 2025-08-21
[MINOR] Apply spotless plugin's fix for code format in gluten-it

PHILO-HE:
Added spotless plugin check for gluten-it. The build for gluten-it will fail during validation phase if there is code violation.
The code style setting is consistent with the parent project.

## How was this patch tested?
Local test.


philo-he:
@zhztheplayer, could you take a look?

tools/gluten-it/tmp.diff
```diff
@@ -0,0 +1,66 @@
+diff --git a/tools/gluten-it/common/pom.xml b/tools/gluten-it/common/pom.xml
```
zhztheplayer: Is this file intentionally added?

philo-he: @zhztheplayer, definitely not. Thanks so much for catching this. Just updated.

PR #10505, opened 2025-08-21
[GLUTEN-10392][VL] Fix filter fallback in scan-only execution

rui-mo:

Filter used to fallback when reading delta table in the scan-only execution. This PR fixes this issue.

## How was this patch tested?

UT verified.


zhztheplayer:

PR #10507, opened 2025-08-22
[GLUTEN-10506] Include compression time in the shuffle write metric for Uniffle

zuston:

The compression time of shuffle data should be included into the shuffle write metric.

## How was this patch tested?

Haven't


jackylee-ch (approved):
👍

PR #10532, opened 2025-08-26
[TEST][VL] Reinclude "cast string to timestamp" test

philo-he:

This PR reverts https://github.com/apache/incubator-gluten/commit/a3a6b39b28cbcf40926d93dce6ad7c1ed5ca433d.

The root cause of the UT failure is that we removed the installation of the new tzdata after confirming that the old tzdata was compatible with Velox. However, the old tzdata does not work correctly for the following timezones: "Asia/Tehran" and "Iran". These timezones appear to be uncommon. This PR simply excludes these two timezones from the test.

Additionally, to enhance debuggability, this PR introduces a helper method to log error messages when using `ThreadUtils.parmap` for multi-threaded testing.

## How was this patch tested?
The test passed.

### DIFF
--- a/gluten-ut/common/src/test/scala/org/apache/spark/util/DebuggableThreadUtils.scala
+++ b/gluten-ut/common/src/test/scala/org/apache/spark/util/DebuggableThreadUtils.scala
@@ -16,20 +16,21 @@
  */
 package org.apache.spark.util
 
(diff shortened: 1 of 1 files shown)


philo-he:
@zhouyuan, could you review this PR?

PR #10533, opened 2025-08-26
[VL] Gluten-it: Simplify CollectionConverter.scala

zhztheplayer:
1. Rename to JavaCollectionConverter.scala
2. Simplify code


zhztheplayer:
@PHILO-HE  Thanks

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/collections/JavaCollectionConverter.scala
```diff
@@ -14,24 +14,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.gluten.utils
+package org.apache.gluten.integration.collections
 
-import scala.collection.immutable.{Map => IMap}
-import scala.collection.immutable.{Seq => ISeq}
+import scala.collection.JavaConverters._
```
philo-he, line 19: I note `scala.collection.JavaConverters` has been marked ad deprecated in Scala 2.13. Do we need to use the recommended `scala.jdk.CollectionConverters` API for 2.13?

zhztheplayer: No worries, the `scala.collection.JavaConverters` is kept in 2.13 for compatibility with 2.12. `gluten-it` supports both versions so it's a norm to access this API.

PR #10540, opened 2025-08-26
[VL] Separate filesystem configuration initialization

marin-ma:
`getHiveConfig` now parses configurations for all kinds of filesystems. It should not parse configurations for other kinds if it's called from a specific `VeloxParquetDataSource` implementation.

### DIFF
--- a/cpp/velox/utils/ConfigExtractor.cc
+++ b/cpp/velox/utils/ConfigExtractor.cc
@@ -27,6 +27,14 @@
 
 namespace gluten {
 
+namespace {
+
+bool contains(FileSystemType fsType, FileSystemType t) {
+  return (static_cast<uint8_t>(fsType) & static_cast<uint8_t>(t)) != 0;
+}
+
+} // namespace
+
 std::string getConfigValue(
     const std::unordered_map<std::string, std::string>& confMap,
     const std::string& key,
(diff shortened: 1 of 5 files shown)


cpp/velox/utils/ConfigExtractor.cc
```diff
@@ -42,176 +50,184 @@ std::string getConfigValue(
...
-      {S3Config::Keys::kIamRoleSessionName, std::make_pair("iam.role.session.name", "gluten-session")},
-      {S3Config::Keys::kEndpointRegion, std::make_pair("endpoint.region", std::nullopt)},
-  };
-
-  // get Velox S3 config key from Spark Suffix.
-  auto getVeloxKey = [&](std::string_view suffix) {
-    for (const auto& [key, value] : sparkSuffixes) {
-      if (value.first == suffix) {
-        return std::optional<S3Config::Keys>(key);
+  if (contains(fsType, FileSystemType::kS3)) {
```
zhouyuan: Shall we also add `else` clause here to guard some corner case?

cpp/velox/utils/ConfigExtractor.h
```diff
@@ -28,12 +28,21 @@
 
 namespace gluten {
 
+enum class FileSystemType : uint8_t {
+  kHdfs = 1 << 0,
+  kS3 = 1 << 1,
+  kAbfs = 1 << 2,
+  kGcs = 1 << 3,
+  kAll = (1 << 4) - 1
```
zhouyuan: The code seems a bit complicated, how about using 0, 1, 2, 3 and std::num_limits<uint8_t>::max for kAll?

marin-ma:
@jinchengchenghh Could you help to review? Thanks!

zhouyuan (approved):
Thanks!

cpp/velox/utils/ConfigExtractor.h
marin-ma: One bit is used for each fs type and we check if the corresponding bit is set. Seems like 0, 1, 2, 3 can't do the job.

marin-ma: It may not be a trivial code path to only parse two or three fs types. I will update the code

marin-ma:
@zhouyuan Could you help to review again? Thanks!

zhouyuan (approved):
👍

Issue #9671, opened 2025-05-16
[VL] BHJ caused stackoverflow


FelixYBW:
### Backend

VL (Velox)

### Bug description

```
25/05/09 19:10:38 ERROR [broadcast-exchange-0] broadcast.TorrentBroadcast: Store broadcast broadcast_1 fail, remove all pieces of the broadcast
25/05/09 19:10:38 ERROR [broadcast-exchange-0] execution.ColumnarBroadcastExchangeExec: Fatal error in broadcast exchange:
              |  Exception type: java.lang.StackOverflowError
              |  Message: 
              |  Cause: 
              |  Thread: broadcast-exchange-0
              |  Stacktrace:
java.lang.StackOverflowError
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.esotericsoftware.kryo.serializers.ClosureSerializer.write(ClosureSerializer.java:62)
	at com.esotericsoftware.kryo.Kryo.writeClassAndObject(Kryo.java:651)
	at com.esotericsoftware.kryo.serializers.DefaultArraySerializers$ObjectArraySerializer.write(DefaultArraySerializers.java:361)
	at com.esotericsoftware.kryo.serializers.DefaultArraySerializers$ObjectArraySerializer.write(DefaultArraySerializers.java:302)
	at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:575)
	at com.esotericsoftware.kryo.serializers.ObjectField.write(ObjectField.java:79)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.write(FieldSerializer.java:508)
	at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:557)
```

There is a R2C conversion before broadcast.

### Gluten version

main branch

### Spark version

Spark-3.2.x

### Relevant logs

```bash

```

FelixYBW:
@JkSelf do you know where the large memory block is allocated in stack?

JkSelf:
[...]

PR #10541, opened 2025-08-26
[GLUTEN-9671][VL] Fix broadcast exchange stackoverflow due to Kryo serialization

felixloesing:
This pull request introduces a safer and more robust approach for handling Spark's `BroadcastMode` during serialization. The main improvement is the introduction of a new `SafeBroadcastMode` abstraction and related utilities, which help avoid serialization issues that caused a Stackoverflow exception during broadcast exchanges. BroadcastMode was introduced in this [PR](https://github.com/apache/incubator-gluten/pull/8116) that caused the issue we observed. HashedRelationBroadcastMode embeds Catalyst expression trees, which are not safe to Kryo-serialize when running with `spark.kryo.referenceTracking=false` (default internally).

With this change, the broadcast payload now contains only primitives and byte arrays (no Catalyst trees). For bound keys, we serialize just column ordinals (+ null-aware flag) and for computed keys (e.g., upper(col)), we serialize the key expressions once as Java bytes and deserialize only where needed to build projections.

#### Test Plan
Ran internal test set (50 queries) and ran other query specifically checking if
`spark.gluten.velox.offHeapBroadcastBuildRelation.enabled=true;` works.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastModeUtils.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastModeUtils.scala
@@ -16,11 +16,12 @@
  */
 package org.apache.spark.sql.execution
 
+import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.expressions.{Attribute, BoundReference, Expression}
 import org.apache.spark.sql.catalyst.plans.physical.{BroadcastMode, IdentityBroadcastMode}
 import org.apache.spark.sql.execution.joins.HashedRelationBroadcastMode
 
-import java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, IOException, ObjectInputStream, ObjectOutputStream}
 
 /**
  * Provides serialization-safe representations of BroadcastMode to avoid issues with circular
@@ -45,7 +46,7 @@ final case class HashSafeBroadcastMode(ordinals: Array[Int], isNullAware: Boolea
 final case class HashExprSafeBroadcastMode(exprBytes: Array[Byte], isNullAware: Boolean)
   extends SafeBroadcastMode
 
-object BroadcastModeUtils {
+object BroadcastModeUtils extends Logging {
 
   /**
    * Converts a BroadcastMode to its SafeBroadcastMode equivalent. Uses ordinals for simple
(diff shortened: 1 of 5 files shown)


FelixYBW:
@JkSelf It fixed the odd issue #9671. Can you take a look?

Spark set `spark.kryo.referenceTracking=true`. The fix can set it to false which boosts performance.

JkSelf:
@felixloesing Thanks for your fix. Just curious — why does Gluten require this change, while vanilla Spark doesn't?

felixloesing:
> @felixloesing Thanks for your fix. Just curious — why does Gluten require this change, while vanilla Spark doesn't?

@JkSelf This is also described in the [PR](https://github.com/apache/incubator-gluten/pull/8116) that introduced the `BroadcastMode` to `BuildSideRelation` but let me summarize. Vanilla Spark doesn’t have this because it computes the transform inside BroadcastExchangeExec, so there’s no post‑reuse mismatch to reconcile. In Gluten, if exchange re-use, the buildKeys of ColumnarSubqueryBroadcastExec and the buildKeys of BroadcastExchangeExec may not be the same, so it is difficult to construct the projection of the transform. That's why Gluten contains the `BroadcastMode` in the `BuildSideRelation` and that introduces Catalyst trees in the exchange that cannot be serialized in Kryo so it causes a StackOverflow exception.

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastModeUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+  // Helpers for expression serialization (used in HashExprSafeBroadcastMode)
+  private[execution] def serializeExpressions(keys: Seq[Expression]): Array[Byte] = {
+    val bos = new ByteArrayOutputStream()
+    var oos: ObjectOutputStream = null
+    try {
+      oos = new ObjectOutputStream(bos)
+      oos.writeObject(keys)
+      oos.flush()
+      bos.toByteArray
+    } finally {
```
JkSelf, line 106: nit: Can we add a catch clause to log the exception and make debugging easier?

backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastModeUtils.scala
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  private[execution] def deserializeExpressions(bytes: Array[Byte]): Seq[Expression] = {
+    val bis = new ByteArrayInputStream(bytes)
+    var ois: ObjectInputStream = null
+    try {
+      ois = new ObjectInputStream(bis)
+      ois.readObject().asInstanceOf[Seq[Expression]]
+    } finally {
```
JkSelf, line 129: ditto.

JkSelf (commented):
Thanks for your fix. LGTM except two minor comments.

felixloesing:
@JkSelf Thanks for the review! Added a catch clause and logged the exception.

PR #10561, opened 2025-08-27
[FLINK] Add `java-17` profile for Flink build and update project version in flink doc

zjuwangg:

https://github.com/apache/incubator-gluten/pull/10515 has made a great effort to decouple flink from parent gluten.

This PR adds a profile named java-17 to set both `maven.compiler.source/target` to 17 (defaults to 11) if users want to build the code with Java 17. 

This PR update compiler version and update  build jar version to match the master pom version change.
## How was this patch tested?

NA

### DIFF
--- a/gluten-flink/docs/Flink.md
+++ b/gluten-flink/docs/Flink.md
@@ -98,8 +98,8 @@ export FLINK_HOME=
 cd $FLINK_HOME
 mkdir -p gluten_lib
 ln -s $VELOX4J_HOME/target/velox4j-0.1.0-SNAPSHOT.jar $FLINK_HOME/gluten_lib/velox4j-0.1.0-SNAPSHOT.jar
-ln -s $GLUTEN_FLINK_HOME/runtime/target/gluten-flink-runtime-1.5.0-SNAPSHOT.jar $FLINK_HOME/gluten_lib/gluten-flink-runtime-1.5.0.jar
-ln -s $GLUTEN_FLINK_HOME/loader/target/gluten-flink-loader-1.5.0-SNAPSHOT.jar $FLINK_HOME/gluten_lib/gluten-flink-loader-1.5.0.jar
+ln -s $GLUTEN_FLINK_HOME/runtime/target/gluten-flink-runtime-1.6.0-SNAPSHOT.jar $FLINK_HOME/gluten_lib/gluten-flink-runtime-1.6.0.jar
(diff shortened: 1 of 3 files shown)


philo-he:
@zjuwangg, It makes sense to fix this since both Java 11 and 17 are supported. Setting maven.compiler.source/target to 17 would cause compatibility issues when building Gluten Flink with Java 11.

I think we can add a profile named java-17 to set both properties to 17 (defaults to 11) if users want to build the code with Java 17. This approach is different from setting the compiler properties to 11 while building with Java 17, as it enables Java 17 features. As you mentioned, velox4j is compiled with Java 11, but it should work when running on a Java 17 JVM due to Java's backward compatibility. Therefore, I assume we don't need to hardcode the compiler source and target properties to 11.

zjuwangg:
> @zjuwangg, It makes sense to fix this since both Java 11 and 17 are supported. Setting maven.compiler.source/target to 17 would cause compatibility issues when building Gluten Flink with Java 11.
> 
> I think we can add a profile named java-17 to set both properties to 17 (defaults to 11) if users want to build the code with Java 17. This approach is different from setting the compiler properties to 11 while building with Java 17, as it enables Java 17 features. As you mentioned, velox4j is compiled with Java 11, but it should work when running on a Java 17 JVM due to Java's backward compatibility. Therefore, I assume we don't need to hardcode the compiler source and target properties to 11.

@PHILO-HE 
Really make sense to me. Will add a jdk17 compile profile soon.

zjuwangg:
@PHILO-HE Just add a java-17 compile profile.

PHILO-HE (approved):
Looks good. Could you enable this profile in Flink CI job? I note Java 17 is used. Please also update the build section in Flink.md to indicate that users should add `-Pjava-17` when building with Java 17.

zjuwangg:
> Looks good. Could you enable this profile in Flink CI job? I note Java 17 is used. Please also update the build section in Flink.md to indicate that users should add `-Pjava-17` when building with Java 17.

@PHILO-HE 
```
            <activation>
                <jdk>17</jdk>
            </activation>
```

Since we add jdk 17 activation option, java-17 profile will be auto selected when build with java 17. Tested passed in my owne box.

zjuwangg:
> Could you enable this profile in Flink CI job? 

I'll investigate on this and enable this profile in another separate PR.

philo-he:
@zjuwangg, I see. Then, there is no need to explicitly add the option in Flink CI. Thanks.

PHILO-HE (approved):

PR #10567, opened 2025-08-27
[GLUTEN-10566][VL] Add Spark unix_timestamp support with timestamp and format arguments

nimesh1601:

Support unix_timestamp and to_unix_timestamp registered with arguments: (TIMESTAMP VARCHAR) by ignoring format arguments to be in parity with Spark

## How was this patch tested?
Uts

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
@@ -474,13 +474,7 @@ trait SparkPlanExecApi {
       substraitExprName: String,
       timeExp: ExpressionTransformer,
       format: ExpressionTransformer,
-      original: Expression): ExpressionTransformer = {
-    GenericExpressionTransformer(
-      substraitExprName,
-      Seq(timeExp, format),
-      original
-    )
-  }
+      original: Expression): ExpressionTransformer
 
   def genDateDiffTransformer(
       substraitExprName: String,


nimesh1601:
@rui-mo can you review this? test failures are unrelated

nimesh1601:
@rui-mo Seems like even if we passed just one argument in the function, still the signature of the function is retrieved from the original expr, and it will still fallback. Fixing this in gluten might be complex, as we would need to create expression again. Do you think making this change in Velox would make more sense ?

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -279,12 +279,23 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+              replaceWithExpressionTransformer0(t.format, attributeSeq, expressionsMap)
+            )
+        }
         GenericExpressionTransformer(
           substraitExprName,
-          Seq(
-            replaceWithExpressionTransformer0(t.timeExp, attributeSeq, expressionsMap),
-            replaceWithExpressionTransformer0(t.format, attributeSeq, expressionsMap)
-          ),
+          children,
```
rui-mo: A more typical approach is to implement a custom transformer for this function and put the specific logic within it. For example,
`BackendsApiManager.getSparkPlanExecApiInstance.genTruncTimestampTransformer` returns the transformer for TruncTimestamp. I’d prefer addressing this issue in Gluten, since introducing a function with an unused parameter to Velox would be confusing.

nimesh1601: Thanks @rui-mo for the suggestion. Made changes accordingly

nimesh1601:
Test failures are unrelated. @rui-mo can you review?

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -276,26 +276,38 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+            // For other inputs (date, string), use original logic
+            GenericExpressionTransformer(
+              substraitExprName,
+              Seq(
+                timeExpTransformer,
+                replaceWithExpressionTransformer0(format, attributeSeq, expressionsMap)
+              ),
+              toUnixTimestamp
+            )
+        }
```
rui-mo: The specific transformer can be used to handle different logics between Timestamp and String types, and 'ExpressionConverter' could be much simplified to just call the transformer. FYI: I added the modified version in my commit: https://github.com/rui-mo/gazelle-jni/commit/0430595e01cf0b8d59c6ee53248e81a4e2186cf7.

nimesh1601: Thanks @rui-mo !

rui-mo:
@nimesh1601 It appears that the Clickhouse UT fails because of below error. The way to solve it is to implement different transformers for CH and Velox backends, like the `BackendsApiManager.getSparkPlanExecApiInstance.genTruncTimestampTransformer`. In this case, let's keep using the `GenericExpressionTransformer` for CH, and use `ToUnixTimestampTransformer` only for Velox. If you have any questions about how to implement this, please let me know.

> GLUTEN-4085: Fix unix_timestamp/to_unix_timestamp *** FAILED ***
18:18:26    org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 2168.0 failed 1 times, most recent failure: Lost task 0.0 in stage 2168.0 (TID 4566) (gluten-gluten-ci-17239-86s6m-6phlb-32x8g executor driver): org.apache.gluten.exception.GlutenException: Function to_unix_timestamp requires exactly two arguments

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala
```diff
@@ -470,6 +470,18 @@ trait SparkPlanExecApi {
     TruncTimestampTransformer(substraitExprName, format, timestamp, original)
   }
 
+  def genToUnixTimestampTransformer(
+      substraitExprName: String,
+      timeExp: ExpressionTransformer,
+      format: ExpressionTransformer,
+      original: Expression): ExpressionTransformer = {
```
rui-mo: The implementation can be removed from this API, and allow backend to provide specific implementation.

gluten-substrait/src/main/scala/org/apache/gluten/expression/DateTimeExpressionsTransformer.scala
```diff
@@ -102,3 +103,18 @@ object DateTimeExpressionsTransformer {
     scala.reflect.classTag[Second].runtimeClass -> "SECOND"
   )
 }
+
+case class ToUnixTimestampTransformer(
```
rui-mo: Consider moving this transformer to 'backends-velox/src/main/scala/org/apache/gluten/expression/ExpressionTransformer.scala' because it's only used by Velox.

nimesh1601: done

rui-mo:
@nimesh1601 Can you please follow up this comment: https://github.com/apache/incubator-gluten/pull/10567#discussion_r2316099446? It helps avoid duplication.

rui-mo (approved):
Thanks!

PR #10573, opened 2025-08-28
[GLUTEN-10649] Avoid repeated calls to identifiyBatchType

### LINKED ISSUE #10649: Avoid repeated calls to identifiyBatchType
### Description

Currently, Gluten has many places call method `identifiyBatchType`. But `identifiyBatchType` always called repeatedly. For example, `getNativeHandle` calls `identifiyBatchType` repeatedly.
```
  public static long getNativeHandle(String backendName, ColumnarBatch batch) {
    if (isZeroColumnBatch(batch)) {
      final ColumnarBatchJniWrapper jniWrapper =
          ColumnarBatchJniWrapper.create(
              Runtimes.contextInstance(backendName, "ColumnarBatches#getNativeHandle"));
      return jniWrapper.getForEmptySchema(batch.numRows());
    }
    return getIndicatorVector(batch).handle();
  }
```

The implementation of `isZeroColumnBatch` show below.
```
  static boolean isZeroColumnBatch(ColumnarBatch batch) {
    return identifyBatchType(batch) == BatchType.ZERO_COLUMN;
  }
```
We can see `isZeroColumnBatch` calls `identifyBatchType` once.
The implementation of `getIndicatorVector ` show below.

```
  private static IndicatorVector getIndicatorVector(ColumnarBatch input) {
    if (!isLightBatch(input)) {
      throw new UnsupportedOperationException("Input batch is not light batch");
    }
    return (IndicatorVector) input.column(0);
  }
```
`getIndicatorVector ` calls `isLightBatch` here.
The implementation of `isLightBatch ` show below.
```
  static boolean isLightBatch(ColumnarBatch batch) {
    return identifyBatchType(batch) == BatchType.LIGHT;
  }
```
Because `identifyBatchType` has a lot of overhead, I think we should only call it once.

```
  private static BatchType identifyBatchType(ColumnarBatch batch) {
    if (batch.numCols() == 0) {
      return BatchType.ZERO_COLUMN;
    }

    final ColumnVector col0 = batch.column(0);
    if (col0 instanceof IndicatorVector) {
      // it's likely a light batch
      for (int i = 1; i < batch.numCols(); i++) {
        ColumnVector col = batch.column(i);
        if (!(col instanceof PlaceholderVector)) {
          throw new IllegalStateException(
              "Light batch should consist of one indicator vector "
                  + "and (numCols - 1) placeholder vectors");
        }
      }
      return BatchType.LIGHT;
    }

    // it's likely a heavy batch
    for (int i = 0; i < batch.numCols(); i++) {
      ColumnVector col = batch.column(i);
      if (!(col instanceof ArrowWritableColumnVector)) {
        throw new IllegalStateException("Heavy batch should consist of arrow vectors");
      }
    }
    return BatchType.HEAVY;
  }
```

### Gluten version

main branch

beliefer:

This PR proposes to avoid repeated calls to `identifiyBatchType`.
Fixes #10649

This improvement can advance the perf.
Before this improvement, the benchmark of TPCDS SF1000 show below.
|     First     |  Second |    Third.  |     Avg      |  Median  |
|----------|---------|----------|----------|----------|
| 3091.7 s | 3012.2 s | 3100.7 s | 3068.2 s | 3034.5 s |

Before this improvement, the benchmark of TPCDS SF1000 show below.
|     First     |  Second |    Third.  |     Avg      |  Median  |
|----------|---------|----------|----------|----------|
| 3056.8 s | 2990.7 s | 3033.7 s | 3027.067 s | 3026.8 s |

The average perf advances 1.34% and the median perf advances 0.25%.

## How was this patch tested?

GA tests.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
@@ -333,8 +333,9 @@ case class ColumnarArrowEvalPythonExec(
         val inputBatchIter = contextAwareIterator.map {
           inputCb =>
             start_time = System.nanoTime()
-            ColumnarBatches.checkLoaded(inputCb)
-            ColumnarBatches.retain(inputCb)
+            val batchType = ColumnarBatches.identifyBatchType(inputCb)
+            ColumnarBatches.checkLoaded(batchType)
+            ColumnarBatches.retain(inputCb, batchType)
             // 0. cache input for later merge
             inputCbCache += inputCb
             numInputRows += inputCb.numRows
@@ -365,7 +366,8 @@ case class ColumnarArrowEvalPythonExec(
               numOutputBatches += 1
               numOutputRows += numRows
               val batch = new ColumnarBatch(joinedVectors, numRows)
-              ColumnarBatches.checkLoaded(batch)
+              val batchType = ColumnarBatches.identifyBatchType(batch)
+              ColumnarBatches.checkLoaded(batchType)
               procTime += (System.nanoTime() - start_time) / 1000000
               batch
           }
--- a/gluten-arrow/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
+++ b/gluten-arrow/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
@@ -49,13 +49,13 @@ public final class ColumnarBatches {
 
   private ColumnarBatches() {}
 
-  private enum BatchType {
+  public enum BatchType {
     LIGHT,
     HEAVY,
     ZERO_COLUMN
   }
 
-  private static BatchType identifyBatchType(ColumnarBatch batch) {
+  public static BatchType identifyBatchType(ColumnarBatch batch) {
     if (batch.numCols() == 0) {
       return BatchType.ZERO_COLUMN;
     }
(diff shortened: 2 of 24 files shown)


beliefer:
ping @zhztheplayer @PHILO-HE @zhouyuan @zml1206 @FelixYBW

gluten-arrow/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
```diff
@@ -381,29 +397,31 @@ public static void release(ColumnarBatch b) {
...
+  private static IndicatorVector getIndicatorVector(ColumnarBatch input, BatchType batchType) {
+    if (!isLightBatch(batchType)) {
       throw new UnsupportedOperationException("Input batch is not light batch");
     }
     return (IndicatorVector) input.column(0);
   }
 
-  public static long getNativeHandle(String backendName, ColumnarBatch batch) {
-    if (isZeroColumnBatch(batch)) {
+  public static long getNativeHandle(String backendName, ColumnarBatch batch, BatchType batchType) {
```
zml1206, line 432: nit: There are multiple separate calls to getNativeHandle, we can overload it to simplify the call.

zhztheplayer:
Have you seen performance improvement after this change?

As of Spark 3.3 ColumnBatch becomes extensible so we can have a thorough cleanup over the batch type code after dropping 3.2.

beliefer:
@zhztheplayer I added the TPCDS benchmark into PR's description.
Because this is an E-To-E tests, the improvement seems not obvious.
Maybe I should create a benchmark like `IteratorBenchmark` ?

zhztheplayer:
> Maybe I should create a benchmark like IteratorBenchmark ?

Good idea. Would do that in a subsequent PR.

backends-velox/src/main/scala/org/apache/spark/api/python/ColumnarArrowEvalPythonExec.scala
```diff
@@ -333,8 +333,9 @@ case class ColumnarArrowEvalPythonExec(
         val inputBatchIter = contextAwareIterator.map {
           inputCb =>
             start_time = System.nanoTime()
-            ColumnarBatches.checkLoaded(inputCb)
-            ColumnarBatches.retain(inputCb)
+            val batchType = ColumnarBatches.identifyBatchType(inputCb)
+            ColumnarBatches.checkLoaded(batchType)
+            ColumnarBatches.retain(inputCb, batchType)
```
zhztheplayer, line 338: I still wonder exposing the batch types to developers will increase the overall complexity. Would you be okay if we extend the ColumnarBatch class after 3.2 is dropped instead? I assume that would not be a long time to wait.

beliefer: Yeah. It seems we should do the refactoring. I think it's OK.

zhztheplayer: I meant hold this PR off and wait for 3.2 to be dropped

beliefer: Got it. Please hold this one until you finished extend `ColumnarBatch`.

zhztheplayer: Absolutely. Thanks!

beliefer:
Please hold the PR.

PR #10575, opened 2025-08-28
[VL] Support independent Gluten CPP build

kerwin-zk:
       
Currently, when using ```builddeps-veloxbe.sh``` to compile gluten cpp independently, it first fetches the velox code and installs related dependencies, so you can’t compile only gluten cpp even if velox has already been built. After this PR, running ```builddeps-veloxbe.sh <build_option> build_gluten_cpp``` will compile only gluten cpp.

## How was this patch tested?

CI

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -259,45 +259,51 @@ function build_velox_backend {
   build_gluten_cpp
 }
 
-(
-  cd $GLUTEN_DIR/ep/build-velox/src
-  ./get_velox.sh $VELOX_PARAMETER
-)
+function get_velox {
+  (
+    cd $GLUTEN_DIR/ep/build-velox/src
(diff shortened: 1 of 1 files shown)


dev/builddeps-veloxbe.sh
```diff
@@ -259,45 +259,51 @@ function build_velox_backend {
   build_gluten_cpp
 }
 
-(
-  cd $GLUTEN_DIR/ep/build-velox/src
-  ./get_velox.sh $VELOX_PARAMETER
-)
+function get_velox_code {
```
philo-he: prefer a shorter name: `get_velox`.

kerwin-zk: @PHILO-HE done

philo-he:
Will merge this pr today if no other comments.

zhouyuan (approved):
👍

PR #10581, opened 2025-08-29
[GLUTEN-10607][MINOR] Fix: Use  `setSafe` in DateWriter to avoid overflow

jiangjiangtian:
If the Date type is nested within an Array, Map, or Struct, invoking `writeRow` may result in an exception due to DateWriter utilizing the unsafe set interface. The `set` interface does not reallocate the value buffer in cases of overflow. Consequently, this pull request changes to use the safer `setSafe` interface.


zhouyuan (approved):

xinghuayu007:
could you create an issue to repeat the exception?

jiangjiangtian:
> could you create an issue to repeat the exception?

I create an issue here https://github.com/apache/incubator-gluten/issues/10607

PR #10585, opened 2025-08-29
[GLUTEN-10113][VL] Refactor hive connector register to accept session level config

zhouyuan:

This patch adds a refactor on hive connector registration. instead of registering when instancing Velox, we defer it to when creating the wholestage context. This allows us to get the session level configurations, and pass to hive/velox 

## How was this patch tested?
local verified


cpp/velox/utils/ConfigExtractor.cc
```diff
@@ -281,7 +281,7 @@ std::shared_ptr<facebook::velox::config::ConfigBase> getHiveConfig(
   hiveConfMap[facebook::velox::connector::hive::HiveConfig::kParquetUseColumnNames] = "true";
   hiveConfMap[facebook::velox::connector::hive::HiveConfig::kOrcUseColumnNames] = "true";
 
-  return std::make_shared<facebook::velox::config::ConfigBase>(std::move(hiveConfMap));
+  return std::make_shared<facebook::velox::config::ConfigBase>(std::move(hiveConfMap), false/*mutable*/);
```
xinghuayu007: The default value of parameter 'mutable' is false, no need to pass it.

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -102,6 +103,10 @@ WholeStageResultIterator::WholeStageResultIterator(
   fileSystem->mkdir(spillDir);
   task_->setSpillDirectory(spillDir);
 
+  // register the hive connectors
+  std::call_once(
+      gluten::VeloxBackend::get()->regFlag, [&]() { gluten::VeloxBackend::get()->initConnector(veloxCfg_); });
```
marin-ma: Wonder if we could register with a unique connector ID for each `WholeStageResultIterator` and call `velox::connector::unregisterConnector` from the dtor?

The connector ID can use similar naming as the queryID

```
fmt::format(
          "Gluten_Stage_{}_TID_{}_VTID_{}",
          std::to_string(taskInfo_.stageId),
          std::to_string(taskInfo_.taskId),
          std::to_string(taskInfo_.vId))
```

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -294,6 +293,12 @@ void VeloxBackend::initCache() {
 }
 
 void VeloxBackend::initConnector(const std::shared_ptr<velox::config::ConfigBase>& hiveConf) {
+  for (const auto& [key, value] : hiveConf->rawConfigs()) {
+    // always update to use new session level conf
+    backendConf_->set(key, value);
```
marin-ma, line 307: If the `initConnector` is only called from `WholeStageResultIterator`, is it possible to remove the filesystem related configurations from the backend conf?

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -294,15 +293,24 @@ void VeloxBackend::initCache() {
 }
 
 void VeloxBackend::initConnector(const std::shared_ptr<velox::config::ConfigBase>& hiveConf) {
```
jinchengchenghh, line 304: Should we move the function to WholeStageResultIterator?

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -102,6 +103,10 @@ WholeStageResultIterator::WholeStageResultIterator(
   fileSystem->mkdir(spillDir);
   task_->setSpillDirectory(spillDir);
 
+  // register the hive connectors
+  std::call_once(
```
jinchengchenghh: In what case, we may register twice?

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -91,8 +91,8 @@ void veloxRuntimeReleaser(Runtime* runtime) {
 void VeloxBackend::init(
     std::unique_ptr<AllocationListener> listener,
     const std::unordered_map<std::string, std::string>& conf) {
-  backendConf_ =
-      std::make_shared<facebook::velox::config::ConfigBase>(std::unordered_map<std::string, std::string>(conf));
+  backendConf_ = std::make_shared<facebook::velox::config::ConfigBase>(
```
jinchengchenghh, line 97: mutable true make copy the configs when iterating the configs

jinchengchenghh: Could we generate the HiveConfig by veloxCfg_ in WholeStageResultIterator every time?

zhouyuan: No, here `veloxCfg_` contains only session level configurations, it does not cover the configurations defined in spark.conf

cpp/velox/compute/VeloxBackend.cc
zhouyuan: Users may pass/modify the filesystem related configurations at session level, by setting the keys here it will allow us to pick the latest values

cpp/velox/compute/WholeStageResultIterator.cc
zhouyuan: it looks like in Spark case, sometimes two stages will be scheduled to run at the same time.

PR #10587, opened 2025-08-29
[CORE] Merge SubstraitUtil classes

kevinwilfong:

There are two classes called SubstraitUtil, one in gluten/utils and one in gluten/substrait/utils. The latter has one method 
used in one location. This PR refactors them into a single class.

I ran into this when I tried to import both in the same file.

## How was this patch tested?

Existing unit tests, this is just a refactor.


gluten-substrait/src/main/scala/org/apache/gluten/utils/SubstraitUtil.scala
```diff
@@ -107,4 +108,20 @@ object SubstraitUtil {
     val nameList = ConverterUtils.collectAttributeNamesWithExprId(output)
     createNameStructBuilder(typeList, nameList, Collections.emptyList()).build()
   }
+
+  def convertJavaObjectToAny(obj: AnyRef): Any = {
+    if (obj == null) return null
+    var msg: Message = null
+    obj match {
+      case Some(s: String) => msg = StringValue.newBuilder.setValue(s).build
```
zml1206: Why use `Some` here?

kevinwilfong: Sorry, I'm new to Scala, I was having trouble getting this code to compile and that was something I found that worked. Spoke with someone more knowledgeable than me and they pointed out what I was doing wrong. Fixed.

gluten-substrait/src/main/scala/org/apache/gluten/utils/SubstraitUtil.scala
```diff
@@ -107,4 +109,20 @@ object SubstraitUtil {
     val nameList = ConverterUtils.collectAttributeNamesWithExprId(output)
     createNameStructBuilder(typeList, nameList, Collections.emptyList()).build()
   }
+
+  def convertJavaObjectToAny(obj: AnyRef): Any = {
+    if (obj == null) return null
+    var msg: Message = null
```
zml1206: suggestion: 
```
val msg = obj match {
    ***
}
```

gluten-substrait/src/main/scala/org/apache/gluten/utils/SubstraitUtil.scala
```diff
@@ -107,4 +109,20 @@ object SubstraitUtil {
...
+    obj match {
+      case s: String => msg = StringValue.newBuilder.setValue(s).build
+      case i: Integer => msg = Int32Value.newBuilder.setValue(i).build
+      case l: JLong => msg = Int64Value.newBuilder.setValue(l).build
+      case d: JDouble => msg = DoubleValue.newBuilder.setValue(d).build
+      case _ =>
+        // TODO: generate the message according to the object type
+        msg = StringValue.newBuilder.setValue(obj.toString).build
+    }
+    if (msg == null) return null
```
zml1206: I think this is redundant.

gluten-substrait/src/main/scala/org/apache/gluten/utils/SubstraitUtil.scala
kevinwilfong: I had to leave in the `: Message` otherwise I got a compiler error I think because the types in the different cases are different, but otherwise made the change.

gluten-substrait/src/main/scala/org/apache/gluten/utils/SubstraitUtil.scala
kevinwilfong: You're right, I copied that from the original code, but the contract of the build functions is that they don't return null so it's safe to remove it.

kevinwilfong:
ClickHouse CI failed with

org.apache.hadoop.ipc.RemoteException(java.io.IOException): File /testFile.txt._COPYING_ could only be replicated to 0 nodes instead of minReplication (=1).  There are 0 datanode(s) running and no node(s) are excluded in this operation.

I assume this is transient

zml1206:

zml1206 (approved):
LGTM

kevinwilfong:
Thanks @zml1206 !

PR #10600, opened 2025-09-01
[GLUTEN-10599][VL] Fix centos7/8 docker image build

zhouyuan:

This patch fixes the centos7/8 docker image build

## How was this patch tested?

Pass GHA

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -36,7 +36,6 @@ env:
 
 jobs:
   build-vcpkg-centos-7:
-    if: ${{ startsWith(github.repository, 'apache/') }}
     runs-on: ubuntu-latest
 
     steps:
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -208,6 +208,8 @@ fi
 concat_velox_param
 
 function build_arrow {
+  get_velox
+  setup_dependencies
   cd $GLUTEN_DIR/dev
   source ./build_arrow.sh
 }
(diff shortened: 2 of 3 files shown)


.github/workflows/docker_image.yml
```diff
@@ -16,6 +16,9 @@
 name: Build and Push Docker Image
 
 on:
+  pull_request:
```
zhouyuan: will revert these changes later

dev/vcpkg/setup-build-depends.sh
```diff
@@ -99,9 +99,9 @@ install_centos_7() {
     # Requires git >= 2.7.4
     if [[ "$(git --version)" != "git version 2."* ]]; then
         yum -y remove git
-        # Requires 'centos-release-scl' package to be installed.
-        yum -y install rh-git227
-        source /opt/rh/rh-git227/enable
```
philo-he, line 104: We already installed git 2.27.0, which can satisfy the minimum requirement 2.7.4. But according to the CI failure, this new git version is not found in later execution. 

The root cause is the above command doesn't change the environment for other subshell execution like `bash xxx.sh`. Maybe, we can simply add the following command to persist the version changes.

```
echo "source /opt/rh/rh-git227/enable" >> ~/.bashrc
```

This git comes from `centos-release-scl`, which should be more stable, meaning it should be always available for future use.

zhouyuan: @PHILO-HE Thanks. Updated to use rh-git227
 it looks like we need to enable the rh-git227 in the docker file instead in the setup script.

PHILO-HE (approved):
Thanks for the fix.

dev/builddeps-veloxbe.sh
```diff
@@ -208,6 +208,8 @@ fi
 concat_velox_param
 
 function build_arrow {
+  get_velox
+  setup_dependencies
```
philo-he: If we add them here, they will be executed two times in the following complete build at line 306.
```
get_velox
setup_dependencies
build_velox_backend
```

Can we move `get_velox` into `setup_dependencies` and explicitly add `setup_dependencies` in dockerfile?

`./dev/builddeps-veloxbe.sh --run_setup_script=ON setup_dependencies build_arrow`

zhouyuan: let me add a flag to check if we need to run get_velox

Issue #10450, opened 2025-08-15
[VL] Remove internal configurations from gluten doc


zhouyuan:
### Description

Some configurations are internally used for passing to Velox backend. Users are not expected to update on these
```
spark.gluten.memory.offHeap.size.in.bytes
spark.gluten.memory.task.offHeap.size.in.bytes
spark.gluten.memoryOverhead.size.in.bytes
spark.gluten.memory.conservative.task.offHeap.size.in.bytes
```

### Gluten version

None

philo-he:
The removal makes sense. Let's target it for 1.5.0.

PR #10603, opened 2025-09-02
[GLUTEN-10450][VL] Reclassify internal/public configs and remove internal configs from doc

zjuwangg:

Remove internal configurations from gluten doc as described (https://github.com/apache/incubator-gluten/issues/10450)

Fix https://github.com/apache/incubator-gluten/issues/10450

## How was this patch tested?
No need


docs/velox-configuration.md
```diff
@@ -9,77 +9,9 @@ nav_order: 16
...
-| spark.gluten.sql.columnar.backend.velox.bloomFilter.numBits                      | 8388608           | The default number of bits to use for the velox bloom filter: 'spark.bloom_filter.num_bits'                                                                                                                                                                                                                                                                                                                                                           |
-| spark.gluten.sql.columnar.backend.velox.cacheEnabled                             | false             | Enable Velox cache, default off. It's recommended to enablesoft-affinity as well when enable velox cache.                                                                                                                                                                                                                                                                                                                                             |
```
zhouyuan: it looks like too many configurations are removed, I believe all these are still required 
Cc @PHILO-HE

zjuwangg: @zhouyuan I used the `dev/gen_all_config_docs.sh` to re-generate the `configuration.md`. I'will review this config and add back this config item if needed.

zjuwangg: I founded the problem, some config in `VeloxConfig.scala` internal flag are not removed, will address it soon

zjuwangg: Just fixed. Thx for your detailed review @zhouyuan

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1172,68 +1128,54 @@ object GlutenConfig {
...
       .checkValue(
         logLevel => Set("TRACE", "DEBUG", "INFO", "WARN", "ERROR").contains(logLevel),
         "Valid values are 'trace', 'debug', 'info', 'warn' and 'error'.")
       .createWithDefault("DEBUG")
 
   val DEBUG_ENABLED =
-    buildConf("spark.gluten.sql.debug")
-      .internal()
-      .booleanConf
+    buildConf("spark.gluten.sql.debug").booleanConf
```
philo-he: I agree with exposing only public configs in Gluten document. However, we need to establish a convention to guide developers on how to categorize configs.

It seems Spark has no official guide to categorize a config to either internal or public. I note debug-related configs are marked as internal, including debug enabling, debug level setting. And some experimental configs are also marked as internal. I think we can follow such convention.

According to my understanding, public configs are those exposed to end users and they are already stable to use. And internal configs are those for developers or advanced users to use. 

Please add a comment for Gluten's `internal()` method to clarify, and review all Gluten configs to see if `internal` is correctly used or removed.

cc @zhouyuan @FelixYBW

zjuwangg: @PHILO-HE Agreed.
I think we can review `internal()` method and  add `experimental()` method to gluten config. Also we can add `version()` like Spark does to make config more explainable.

zhouyuan: I suggest asking reviewers/maintainers to carefully check whether these configurations should be public when adding them

zjuwangg: @PHILO-HE 
Just add comments to `internal()` method and review the internal() flag according to my understanding.

zjuwangg: Also just open a https://github.com/apache/incubator-gluten/issues/10613 to add `experimental` and `version` method to make config more maintainable.

gluten-core/src/main/scala/org/apache/gluten/config/ConfigBuilder.scala
```diff
@@ -51,6 +51,13 @@ private[gluten] case class ConfigBuilder(key: String) {
     this
   }
 
+  /**
+   * `internal()` method are designed to tag config item as: * internal or experimental
+   * functionality * intended exclusively for developers or advanced users * unstable and not
+   * exposed to external users * allowing for flexibility in development and testing without
+   * compromising the public API's stability
+   * @return
+   */
```
philo-he, line 60: Using `*` to mark items is not friendly to readers in the comment section.
Suggested:

```
/**
 * This method marks a config as internal for any of the following reasons:
 * - Intended exclusively for developers or advanced users
 * - Experimental or unstable, not yet exposed to end users
 * - Allows for flexibility in development and testing without
 *   compromising the public API's stability
 */
```

zjuwangg: Make sense, just addressed

docs/velox-configuration.md
```diff
@@ -75,11 +66,6 @@ nav_order: 16
...
 | spark.gluten.velox.castFromVarcharAddTrimNode                                    | false             | If true, will add a trim node which has the same sementic as vanilla Spark to CAST-from-varchar.Otherwise, do nothing.                                                                                                                                                                                                                                                                                                                                |
-| spark.gluten.velox.fs.s3a.connect.timeout                                        | 200s              | Timeout for AWS s3 connection.                                                                                                                                                                                                                                                                                                                                                                                                                        |
```
zhouyuan, line 82: I'm not sure if this should be marked as internal

zjuwangg: Just addressed.

PHILO-HE (approved):
Looks good. Thanks.

PR #10205, opened 2025-07-17
[DOC][Flink] Update flink build command to skip gpg and spotless check

zjuwangg:
## What changes were proposed in this pull request?
Add flink related velox4j build problem

PR #10604, opened 2025-09-02
[DOC][FLINK] Update flink build command to skip gpg and spotless check

zjuwangg:

Since https://github.com/apache/incubator-gluten/pull/10205/files has provide a runnable command to build velox4j, but it's broken after recent code commit. This PR re-update the build command.

## How was this patch tested?

NA


PHILO-HE (approved):
Thanks.

PR #10606, opened 2025-09-02
[GLUTEN-10605][VL] Rewrite unbounded window to an equivalent aggregate join

zml1206:

For unbounded window, velox needs to load the entire partition's data into memory for calculation, which can easily cause OOM when the partition is too large. This rule rewrites unbounded window to an equivalent aggregate join operation to avoid OOM.
Input query:
```
SELECT *, SUM(c0) OVER (PARTITION BY c1) AS sum FROM t
```
Rewritten query:
```
SELECT t.*, t1.sum FROM t LEFT JOIN (SELECT c1, SUM(c0) AS sum FROM t GROUP BY c1) t1  ON t. c1 <=> t1.c1
```

## How was this patch tested?

UT


zml1206:
@JkSelf What do you think of this feature? It should make up for some of the current shortcomings of Velox Window.

JkSelf (approved):
LGTM. Thanks for your improving.

zml1206:
@JkSelf Thanks for review, merge into main.

PR #10614, opened 2025-09-03
[VL] Example workload for benchmarking Gluten + Delta on TPC datasets

zhztheplayer:
This adds a workload example `tools/workload/tpcds-delta`.

Data generator is using `gluten-it`.

Workload is a trivial TPC-DS query benchmark on Delta tables without the DML commands (delete, merge, insert, etc.).

### DIFF
--- a/tools/workload/tpcds-delta/README.md
+++ b/tools/workload/tpcds-delta/README.md
@@ -1,0 +1,53 @@
+# Test on Velox backend with Delta Lake and TPC-DS workload
+
+This workload example is verified with JDK 8, Spark 3.4.4 and Delta 2.4.0.
+
+## Test dataset
+
+A simple way to generate Delta tables is to use the built-in TPC-DS data generator of `gluten-it`.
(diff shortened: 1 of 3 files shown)


tools/workload/tpcds-delta/README.md
```diff
@@ -0,0 +1,45 @@
...
+This workload example is verified with JDK 8, Spark 3.4.4 and Delta 2.4.0.
+
+## Test dataset
+
+Use bash script `tpcds_datagen_delta.sh` to generate the data. The script relies on a already-built gluten-it
+executable. To build it, following the steps:
+
+```bash
+cd ${GLUTEN_HOME}/tools/gluten-it/
+mvn clean install -D spark-3.4,delta
```
jinchengchenghh: Maybe -P

zhztheplayer: Should be -P. Thanks for catching.

Issue #10107, opened 2025-07-03
[VL] Make ColumnarShuffleExchangeExec extensible for different serializers


zhztheplayer:
An API `#isUseCelebornShuffleManager` in the common module is heavily relied in our code:

https://github.com/apache/incubator-gluten/blob/4017788a1e5d782d844260528bb2a49ccff02e44/gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala#L132-L136

The API is related tp Celeborn RSS, which is plugged in Gluten's framework as an optional dependency.

Having this kind of API in common module weakens the extensibility and portability of the core code.

To enhance, we could make ColumnarShuffleExchangeExec extensible so serializer instances can be inlined into the plan node. So we don't have to call `#isUseCelebornShuffleManager` to find a proper serializer anymore.

zhztheplayer:
Be free to take if anyone wants to.

zjuwangg:
I'd like to take it.

zjuwangg:
@zhztheplayer 
https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/SparkPlanExecApi.scala#L366 
```
 /**
   * Generate ColumnarBatchSerializer for ColumnarShuffleExchangeExec.
   *
   * @return
   */
  def createColumnarBatchSerializer(
      schema: StructType,
      metrics: Map[String, SQLMetric],
      isSort: Boolean): Serializer
```

Since now columnarBatchSerializer depends on schema metrics and isSort flag, I think it's a little complex to extract the serializer  from ColumnarShuffleExchangeExec. Another problem is that there is no proper chance for custom shuffle manager to inject their own ColumnarShuffleExchangeExec with minor refactor.

I try add a new trait `NeedCustomColumnarBatchSerializer` in https://github.com/apache/incubator-gluten/pull/10201/files#diff-a59ec7a2382de83da668099f5733b99beeceeea10085187a6d7861772d19c741 and leverage java spi to implment auto injection like `CelebornShuffleWriterFactory` does. 

Please help have a look whether it's a proper way.

zhztheplayer:
@zjuwangg I'll leave this open until we can do a deep refactor, let me know once you have any idea about that.

zjuwangg:
> [@zjuwangg](https://github.com/zjuwangg) I'll leave this open until we can do a deep refactor, let me know once you have any idea about that.

[...]

PR #10616, opened 2025-09-03
[GLUEN-10107] Deprecate `isUseCelebornShuffleManager` from GlutenConfig

zjuwangg:
Aims to accomplish #10107 by:

* Introduce `NeedCustomShuffleWriterType` to let shuffle manager to decide shuffle writer type so that we can decouple `isUseCelebornShuffleManager` from GlutenConfig.
* Remove unnecessary `isUseCelebornShuffleManager` in CH backend since corresponding code logic has guaranteed config value is always true. 

## How was this patch tested?

Existing test

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -63,8 +63,6 @@ import javax.ws.rs.core.UriBuilder
 
 import java.util.Locale
 
-import scala.collection.Seq
-
 class VeloxSparkPlanExecApi extends SparkPlanExecApi {
 
   /** Transform GetArrayItem to Substrait. */


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -553,30 +553,19 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi {
...
       output: Seq[Attribute]): ShuffleWriterType = {
     val conf = GlutenConfig.get
-    // todo: remove isUseCelebornShuffleManager here
-    if (conf.isUseCelebornShuffleManager) {
-      if (conf.celebornShuffleWriterType == ReservedKeys.GLUTEN_SORT_SHUFFLE_WRITER) {
-        if (conf.useCelebornRssSort) {
-          RssSortShuffleWriterType
-        } else if (partitioning != SinglePartition) {
+    SparkEnv.get.shuffleManager match {
+      case shuffleManager: NeedCustomShuffleWriterType =>
```
marin-ma, line 557: There's a trait `SupportsColumnarShuffle`. We can add and implement method `getColumnarShuffleWriterType` for the supported shuffle managers

zjuwangg: @marin-ma It's a good idea. But since `ColumnarShuffleManager` is in common module, we must hardcode `velox`/`ch` backendName to distinguish different logic like in this poc code  https://github.com/zjuwangg/gluten/commit/5dc4da65c6f5eeb3057d1377e283d2bba17b74e9. 
It go against our design to decouple backend-velox and common module dependency.
Please ping me if there is better way to refactor this or am I understand wrong

zjuwangg: @marin-ma can you help review this comments again?

zjuwangg:
@zhztheplayer Would you like to help have a look of this refactor of config?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
marin-ma: In my opinion, the implementation of sort shuffle writer type for velox+celeborn is a special case. Perhaps we can add a function like `getShuffleWriterTypeForCeleborn` to the BackendsApi to handle this special case, and call it in the `getColumnarShuffleWriterType` in `CelebornShuffleManager`.

zjuwangg: > In my opinion, the implementation of sort shuffle writer type for velox+celeborn is a special case. Perhaps we can add a function like `getShuffleWriterTypeForCeleborn` to the BackendsApi to handle this special case, and call it in the `getColumnarShuffleWriterType` in `CelebornShuffleManager`.

Do you mean in this way https://github.com/apache/incubator-gluten/pull/10616/commits/47f2edc4789a3c19e014b48f30984bca986a2ce7 ?

I still think `getColumnarShuffleWriterType` can not put in the `SupportsColumnarShuffle ` but need a separate trait to mark shuffle manage need to do special process

zjuwangg: Ping  again @marin-ma

zjuwangg: @zhztheplayer or @PHILO-HE  can you help also take a look?

PR #10634, opened 2025-09-05
[GLUTEN-10660][VL] Adding configuration for hash table build

zhouyuan:

Adding configurations for hash map build optimization in https://github.com/facebookincubator/velox/pull/7066

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -597,7 +597,7 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
     configs[velox::core::QueryConfig::kAbandonBuildNoDupHashMinRows] =
         std::to_string(veloxCfg_->get<int32_t>(kAbandonBuildNoDupHashMinRows, 100000));
     configs[velox::core::QueryConfig::kAbandonBuildNoDupHashMinPct] =
-        std::to_string(veloxCfg_->get<int32_t>(kAbandonBuildNoDupHashMinPct, 100));
+        std::to_string(veloxCfg_->get<int32_t>(kAbandonBuildNoDupHashMinPct, 0));
 
     // Disable driver cpu time slicing.
     configs[velox::core::QueryConfig::kDriverCpuTimeSliceLimitMs] = "0";


liujiayi771:
We need to set spark.gluten.velox.abandonbuild.noduphashminpct=100 to verify the behavior when deduplication is enabled.

liujiayi771:
After the testing, should we enable this feature by default?
In addition, the config key may still change in the upstream Velox. It's better to wait until the code review comments are finalized.

zhouyuan:
@liujiayi771 I vote for enabling this by default. The staging patch in our Velox fork is not updated. We will update on that after https://github.com/facebookincubator/velox/pull/7066 merged

FelixYBW:
At least for TPCDS test, I didn't see any performance regression with ` spark.gluten.velox.abandonbuild.noduphashminpct=100 `. I think we can enable it by default.

liujiayi771:
TPCDS has very few queries that contain semi/anti joins, and all of these involve duplicate join keys.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -545,6 +545,20 @@ object VeloxConfig {
       .booleanConf
       .createWithDefault(false)
 
+  val VELOX_HASHMAP_ABANDON_BUILD_DUPHASH_MIN_ROWS =
+    buildConf("spark.gluten.velox.abandonbuild.noduphashminrows")
+      .internal()
+      .doc("Experimental: abandon hashmap build if duplicated rows more than this number.")
```
zjuwangg, line 535: Should add `experimental()` method and re-generate configuration doc if  needed.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -545,6 +545,20 @@ object VeloxConfig {
...
+      .intConf
+      .createWithDefault(100000)
+
+  val VELOX_HASHMAP_ABANDON_BUILD_DUPHASH_MIN_PCT =
+    buildConf("spark.gluten.velox.abandonbuild.noduphashminpct")
+      .internal()
```
zjuwangg: ditto.

zhouyuan:
@liujiayi771 are you suggesting it's better to disable this feature by default?

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -529,6 +529,20 @@ object VeloxConfig {
...
+    buildConf("spark.gluten.velox.abandonbuild.noduphashminrows")
+      .experimental()
+      .doc("Experimental: abandon hashmap build if duplicated rows more than this number.")
+      .intConf
+      .createWithDefault(100000)
+
+  val VELOX_HASHMAP_ABANDON_BUILD_DUPHASH_MIN_PCT =
+    buildConf("spark.gluten.velox.abandonbuild.noduphashminpct")
+      .experimental()
+      .doc("Experimental: abandon hashmap build if duplicated rows are more than this pct.")
```
marin-ma: Perhaps use "percentile" to avoid abbreviations in the doc.

liujiayi771:
> @liujiayi771 are you suggesting it's better to disable this feature by default? 

Yes, for the Q95 build side join keys, the duplication rate is very high, so the optimization is very significant. However, for production jobs without a high duplication rate, there could be a performance regression. It's necessary to adjust the abandon percent based on the specific job conditions.

Issue #10636, opened 2025-09-05
[VL] Find unsupported expressions using native validation in PartialProject


jiangjiangtian:
### Description

Now `PartialProject` can't be applied to the expressions that are unsupported by native backend, so maybe we can use native validation to find all unsupported expressions and apply `PartialProject` to them.

### Gluten version

None

PR #10637, opened 2025-09-05
[GLUTEN-10636][VL]Use backend validation to find all unsupported expression

jiangjiangtian:
Now PartialProject can't be applied to the expressions that are unsupported by native backend, so maybe we can use native validation to find all unsupported expressions and apply PartialProject to them.

This PR uses the following method to find all unsupported expressions: Traverse the expression trees in post-order. For each tree node, if this expression is not offload-able, then we replace it by an `Alias`.
For example, if the expression is `func4(func3(func2(func1())))`, then we first determine whether `func1()` is offload-able. Second, we determine whether `func2()` is offload-able and so on.

After this PR, expressions like `map_from_arrays` that hasn't been supported by velox can be calculated using `PartialProjectExec`.
This PR also adds a config named `spark.gluten.sql.columnar.partial.validation` to control whether to use native validation to find all unsupported expressions.

## How was this patch tested?
unit test.

Related issue: #10636

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
@@ -19,7 +19,6 @@ package org.apache.gluten.execution
 import org.apache.gluten.backendsapi.BackendsApiManager
 import org.apache.gluten.columnarbatch.{ColumnarBatches, VeloxColumnarBatches}
 import org.apache.gluten.config.GlutenConfig
-import org.apache.gluten.exception.GlutenNotSupportException
 import org.apache.gluten.expression.{ArrowProjection, ConverterUtils, ExpressionConverter, ExpressionMappings, ExpressionUtils, TransformerState}
 import org.apache.gluten.extension.columnar.transition.Convention
 import org.apache.gluten.iterator.Iterators
@@ -395,14 +394,18 @@ object ColumnarPartialProjectExec {
         // To prevent nested expressions be validated multiple times, before doing the validation,
         // we replace it by `Alias`.
         val tempAttributes = new ListBuffer[Attribute]()
-        val toValidatedExpression = child.withNewChildren(
-          child.children.zipWithIndex.map(
-            c => {
-              val child = c._1
-              val a = AttributeReference(s"$dummyPrefix${c._2}", child.dataType, child.nullable)()
-              tempAttributes.append(a)
-              a
-            }))
+        val tempChildren = child.children.zipWithIndex.map(
+          c => {
+            val child = c._1
+            val a = AttributeReference(s"$dummyPrefix${c._2}", child.dataType, child.nullable)()
+            tempAttributes.append(a)
+            a
+          })
+        // For CreateNamedStruct, we need to create it with `CreateStruct.apply`.
+        val toValidatedExpression = child match {
+          case CreateNamedStruct(_) => CreateStruct(tempChildren)
+          case _ => child.withNewChildren(tempChildren)
+        }
         if (
           !doNativeValidateExpression(
             toValidatedExpression,
@@ -448,7 +451,7 @@ object ColumnarPartialProjectExec {
             x
           }
         } catch {
-          case _: GlutenNotSupportException | _: UnsupportedOperationException =>
+          case _: Throwable =>
             // If the process of conversion of the expression throws exception, then we need to
             // fallback the whole operator.
             newExpr
@@ -468,7 +471,7 @@ object ColumnarPartialProjectExec {
             traverseUpExpression(p, replacedAlias, childOutput)
           }
         } catch {
-          case _: GlutenNotSupportException | _: UnsupportedOperationException =>
+          case _: Throwable =>
             // If the process of conversion of the expression throws exception, then we need to
             // fallback the whole operator. The unsupported expression may cause the calculation
             // crash. For example, the result data type of the expression is decimal and the scale


jiangjiangtian:
@jinchengchenghh @WangGuangxin @zhztheplayer Do you think it is a reasonable way to extend `PartialProject` to all unsupported expressions. Please take a look, thanks!

zhztheplayer:
I agree we should prefer PartialProject over vanilla Spark project in almost all cases where projects fall back.

jiangjiangtian:
> I agree we should prefer PartialProject over vanilla Spark project in almost all cases where projects fall back.

I think so. In this PR, I use validation to find the expressions that backend doesn't support, do you think it is a reasonable way?

zhztheplayer:
Kindly ping @jinchengchenghh @WangGuangxin

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -50,7 +54,9 @@ import scala.collection.mutable.ListBuffer
  * @param child
  *   child plan
  */
-case class ColumnarPartialProjectExec(projectList: Seq[NamedExpression], child: SparkPlan)(
+case class ColumnarPartialProjectExec(
+    projectList: Seq[Expression],
+    child: SparkPlan,
     replacedAlias: Seq[Alias])
```
jinchengchenghh, line 57: Why you move replacedAlias to argument list?

jiangjiangtian: Because after this PR, `SubQuery` may be replaced. If `replacedAlias` is not put into the first argument list, then [prepareSubqueries](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/SparkPlan.scala#L274-L282) can't find the subquery in `replacedAlias`. So I put it in the first argument list.

jiangjiangtian:
> Thanks for your enhancement. This looks like a bit complicated, could we reuse replaceWithExpressionTransformer to validate?

Do you mean that we shouldn't create expression transformer for one expression for multiple times?

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
jinchengchenghh: The Alias is unique, with this change, the exprId is different, and the alias origins from projectList, the subquery should be found from it.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
```diff
@@ -349,13 +351,115 @@ object ColumnarPartialProjectExec {
     }
   }
 
+  private def doNativeValidateExpression(
+      expr: Expression,
+      replacedAlias: ListBuffer[Alias],
+      childOutput: Seq[Attribute]): Boolean = {
+    val substraitContext = new SubstraitContext
+    val output = childOutput ++ replacedAlias.map(_.toAttribute)
+    val exprTransformer = ExpressionConverter.replaceWithExpressionTransformer(expr, output)
```
jinchengchenghh, line 366: Here may throw exception for function name not mapping

jinchengchenghh:
Could you describe in the PR description how you validate the nested functions, such as func1(func2(func3(func4())))?
You need to find which function is the not supported expression.

And, if it is possible to reuse the node validation?
      val relNode =
        getRelNode(substraitContext, list, child.output, operatorId, null, validation = true)
      // Then, validate the generated plan in native engine.
      doNativeValidation(substraitContext, relNode)

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
jiangjiangtian: Yes, I have wrapped it in `try-catch` clause when I call this function.

jiangjiangtian:
> Could you describe in the PR description how you validate the nested functions, such as func1(func2(func3(func4())))? You need to find which function is the not supported expression.
> 
> And, if it is possible to reuse the node validation? val relNode = getRelNode(substraitContext, list, child.output, operatorId, null, validation = true) // Then, validate the generated plan in native engine. doNativeValidation(substraitContext, relNode)

Thanks! I have update the PR description.

I think we can't reuse the node to validate because the children of the expression may change.

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarPartialProjectExec.scala
jiangjiangtian: > The Alias is unique, with this change, the exprId is different

Sorry, I can't understand this. Do you mean that the `Alias`'s exprId is different from the original exprId?

> the alias origins from projectList, the subquery should be found from it.

I get it. Thanks!

jinchengchenghh:
With that validation strategy, do you mean we will validate func1() 4 times if func4 is not supported? Could you optimize it?

> For example, if the expression is func4(func3(func2(func1()))), then we first determine whether func1() is offload-able. Second, we determine whether func2() is offload-able and so on.

jiangjiangtian:
> With that validation strategy, do you mean we will validate func1() 4 times if func4 is not supported? Could you optimize it?
> 
> > For example, if the expression is func4(func3(func2(func1()))), then we first determine whether func1() is offload-able. Second, we determine whether func2() is offload-able and so on.

Oh yes, I will try to optimize it. Thank you!

FelixYBW:
good enhancement!

jiangjiangtian:
@jinchengchenghh Please take a look again. It seems that the CI failure has nothing to do with this PR. Thanks!

PR #10639, opened 2025-09-05
[VL] Support generating DV-enabled TPC-DS Delta table in gluten-it

zhztheplayer:
Add a new option `data-gen-features` in gluten-it, to optionally append some post operations after the table is generated.

For example, using 

```
--data-source=delta --data-gen-features=dv_enabled,10pc_deleted
```

To generate a DV-enabled Delta table, where 10% of fact table data is deleted.

Another example:

```
--data-source=delta --data-gen-features=10pc_deleted
```

To generate a regular copy-on-write Delta table , where 10% of fact table data is deleted.

The table folder will be named according to the enabled features, for example:

`/tmp/tpcds-generated-1.0-delta-partitioned-dv_enabled-10pc_deleted`

### DIFF
--- a/tools/gluten-it/common/src/main/java/org/apache/gluten/integration/BaseMixin.java
+++ b/tools/gluten-it/common/src/main/java/org/apache/gluten/integration/BaseMixin.java
@@ -29,6 +29,7 @@ import org.apache.log4j.LogManager;
 import org.apache.spark.SparkConf;
 import picocli.CommandLine;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


tools/gluten-it/common/src/main/java/org/apache/gluten/integration/BaseMixin.java
```diff
@@ -91,6 +92,13 @@ public class BaseMixin {
       defaultValue = "false")
   private boolean genPartitionedData;
 
+  @CommandLine.Option(
+      names = {"--data-gen-features"},
+      description =
+          "Set a comma-separated list of data generator features to enable. Example: --queries=dv_enabled,10pc_deleted",
```
jinchengchenghh: --queries=dv_enabled,10pc_deleted typo?

zhztheplayer: Yes, it's a typo.

Issue #10644, opened 2025-09-06
NumberFormatException in WindowFunctionNode.setBound()


mingyi6666:
#10645 
### Bug description
As shown in the example below:  In WindowFunctionNode.setBound()，If frameType == "ROW" and boundType is DoubleType,
the code:
https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/java/org/apache/gluten/substrait/expression/WindowFunctionNode.java#L124
```
Long offset = Long.valueOf(boundType.eval(null).toString());
```

will have a  java.lang.NumberFormatException in Long.valueOf();
### sql:
```
CREATE TEMPORARY VIEW t AS SELECT 1;
SELECT COUNT(*) OVER (PARTITION BY 1 ORDER BY cast(1 as double) DESC RANGE BETWEEN CURRENT ROW AND 1 FOLLOWING) FROM t;
```
### result:
```
java.lang.NumberFormatException: For input string: "1.0"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Long.parseLong(Long.java:589)
	at java.lang.Long.valueOf(Long.java:803)
	at org.apache.gluten.substrait.expression.WindowFunctionNode.setBound(WindowFunctionNode.java:124)
	at org.apache.gluten.substrait.expression.WindowFunctionNode.toProtobuf(WindowFunctionNode.java:181)
	at org.apache.gluten.substrait.rel.WindowRelNode.toProtobuf(WindowRelNode.java:77)
	at org.apache.gluten.substrait.plan.PlanNode.toProtobuf(PlanNode.java:78)
```

### Gluten version

main branch, Gluten-1.4, Gluten-1.3

### Spark version

Spark-3.3.x

### Relevant logs

```bash

```

PR #10645, opened 2025-09-06
[GLUTEN-10644][VL] Fix NumberFormatException in WindowFunctionNode.setBound

mingyi6666:
Fix NumberFormatException in WindowFunctionNode.setBound

fixes: #10644


zml1206:
I think this is not the velox backend, velox does not have this problem, because velox only supports integral type and date type. Can you confirm? @mingyi6666 
https://github.com/apache/incubator-gluten/blob/73a7673fd0a402a23d8cc8b1fa4529538c5ed256/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala#L423-L431

PR #10646, opened 2025-09-06
[GLUTEN-10625][CORE] Add RewriteCreateTableAsSelect for spark34

Zouxxyy:

Add RewriteCreateTableAsSelect for spark34 for DSv2 Column Write

### DIFF
--- a/backends-velox/src-iceberg/test/scala/org/apache/gluten/execution/enhanced/VeloxIcebergSuite.scala
+++ b/backends-velox/src-iceberg/test/scala/org/apache/gluten/execution/enhanced/VeloxIcebergSuite.scala
@@ -19,7 +19,9 @@ package org.apache.gluten.execution.enhanced
 import org.apache.gluten.execution.{IcebergSuite, VeloxIcebergAppendDataExec, VeloxIcebergOverwriteByExpressionExec, VeloxIcebergReplaceDataExec}
 import org.apache.gluten.tags.EnhancedFeaturesTest
 
+import org.apache.spark.sql.Row
 import org.apache.spark.sql.execution.CommandResultExec
+import org.apache.spark.sql.gluten.TestUtils
 
 @EnhancedFeaturesTest
 class VeloxIcebergSuite extends IcebergSuite {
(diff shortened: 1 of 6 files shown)


Zouxxyy:
@jinchengchenghh  I tried to offload the CTAS command, and I found two main reasons that prevented me from doing so:

1. Many classes need to be copied, and it is difficult to validate fallback logic:
This is because we can only construct the v2 `write` after creating the table, which is required for performing the validation. By that point, it's already too late to perform proper fallback.
2. Spark 3.5 alreadly supports columnar v2 write for CTAS after this PR was introduced https://github.com/apache/spark/pull/40734, this PR is essentially a copy of it.

zhouyuan:
@Zouxxyy @jinchengchenghh  could we also enable this in Spark-35 code path?

Zouxxyy:
> @Zouxxyy @jinchengchenghh could we also enable this in Spark-35 code path?

If we run VeloxIcebergSuite on Spark35, it should work too

backends-velox/src-iceberg/test/scala/org/apache/gluten/execution/enhanced/VeloxIcebergSuite.scala
```diff
@@ -190,4 +192,29 @@ class VeloxIcebergSuite extends IcebergSuite {
...
+      spark.sql("insert into table iceberg_tb1 values (1, 1), (2, 2)")
+
+      // CTAS
+      val sqlStr = """
+                     |create table iceberg_tb2 using iceberg
+                     |partitioned by (pt)
+                     |as select * from iceberg_tb1
+                     |""".stripMargin
+
+      TestUtils.checkExecutedPlanContains[VeloxIcebergAppendDataExec](spark, sqlStr)
```
jinchengchenghh, line 212: Could we use assert(
            getExecutedPlan(df).exists(
              plan => plan.find(_.isInstanceOf[VeloxIcebergAppendDataExec]).isDefined))

jinchengchenghh (commented):
Thanks for your contribution!

jinchengchenghh:
VeloxIcebergSuite runs on Spark34 and Spark35, so this feature is supported in both Spark 34 and 35

backends-velox/src-iceberg/test/scala/org/apache/gluten/execution/enhanced/VeloxIcebergSuite.scala
Zouxxyy: I tried it, seemed didn't work. Maybe it's because `AtomicCreateTableAsSelect` constructs another AppendData exec runtime. The `df.queryExecution.executedPlan` just like this

```text
CommandResult <empty>
   +- AtomicCreateTableAsSelect org.apache.iceberg.spark.SparkCatalog@69c263f8, default.iceberg_tb2, [identity(pt)], Project [a#10, pt#11], TableSpec(Map(),Some(iceberg),Map(),None,None,None,false), false

```

shims/spark34/src/main/scala/org/apache/spark/sql/extension/RewriteCreateTableAsSelect.scala
```diff
@@ -0,0 +1,202 @@
...
+  private def makeQualifiedDBObjectPath(location: String): String = {
+    CatalogUtils.makeQualifiedDBObjectPath(
+      session.sharedState.conf.get(WAREHOUSE_PATH),
+      location,
+      session.sharedState.hadoopConf)
+  }
+}
+
+/** Port from Spark 3.5. */
+case class AtomicCreateTableAsSelectExec(
```
zhouyuan, line 93: I see this logic now. so this rewrite is only required for spark-34, spark-35 already provide this support

Zouxxyy:
@jinchengchenghh @zhouyuan Could you have a look again? Thanks.

PR #10647, opened 2025-09-07
[VL] Minor refactoring of get_velox.sh for better readability

Zouxxyy:

Wrapping the key steps into functions will help everyone understand the flow of the script.

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -18,7 +18,7 @@
 ####################################################################################################
 #  The main function of this script is to allow developers to build the environment with one click #
 #  Recommended commands for first-time installation:                                               #
-#  ./dev/buildbundle-veloxbe.sh                                                            #
+#  ./dev/buildbundle-veloxbe.sh                                                                    #
 ####################################################################################################
 set -exu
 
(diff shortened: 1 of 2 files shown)


ep/build-velox/src/get_velox.sh
```diff
@@ -242,8 +230,16 @@ if [[ "$RUN_SETUP_SCRIPT" == "ON" ]]; then
     echo "Unsupported kernel: $OS"
     exit 1
   fi
+}
+
+prepare_velox_source_code
+
+if [[ "$RUN_SETUP_SCRIPT" == "ON" ]]; then
```
Zouxxyy, line 237: Actually, I don't think it's necessary to pass `RUN_SETUP_SCRIPT` here. If we have a brand new velox, we should just apply the setup fixes as we apply other patches or fixes. (Once is enough)

Then the user can repeatedly execute .builddeps-veloxbe.sh setup_dependencies, because setup generally needs to be executed repeatedly in a new environment (my experience)

dev/builddeps-veloxbe.sh
```diff
@@ -305,7 +305,9 @@ commands_to_run=${OTHER_ARGUMENTS:-}
 (
   if [[ "x$commands_to_run" == "x" ]]; then
     get_velox
-    setup_dependencies
+    if [[ "$RUN_SETUP_SCRIPT" == "ON" ]]; then
```
philo-he: Can we also move the check `-z "${GLUTEN_VCPKG_ENABLED:-}"` here?

PHILO-HE (commented):
Please do a rebase. Thanks.

Zouxxyy:
@PHILO-HE  Can you take a look again thanks, the failed case should be irrelevant.

ep/build-velox/src/get_velox.sh
```diff
@@ -62,18 +62,15 @@ for arg in "$@"; do
...
-}
+CURRENT_DIR=$(
+  cd "$(dirname "$BASH_SOURCE")"
+  pwd
+)
+
+if [ "$VELOX_HOME" == "" ]; then
+  VELOX_HOME="$CURRENT_DIR/../build/velox_ep"
+fi
+VELOX_SOURCE_DIR="${VELOX_HOME}"
```
philo-he: It seems not necessary to define `VELOX_SOURCE_DIR`.

ep/build-velox/src/get_velox.sh
```diff
@@ -242,8 +229,16 @@ if [[ "$RUN_SETUP_SCRIPT" == "ON" ]]; then
...
+}
+
+prepare_velox_source_code
+
+if [[ "$RUN_SETUP_SCRIPT" == "ON" ]]; then
+  apply_setup_fixes
 fi
 
-apply_compilation_fixes $CURRENT_DIR $VELOX_SOURCE_DIR
+apply_provided_velox_patch "$VELOX_SOURCE_DIR"
```
philo-he: VELOX_HOME is a basic global variable. I think we can just access it inside the function instead of passing the arg.

ep/build-velox/src/get_velox.sh
```diff
@@ -242,8 +229,16 @@ if [[ "$RUN_SETUP_SCRIPT" == "ON" ]]; then
...
+
+apply_compilation_fixes "$CURRENT_DIR" "$VELOX_SOURCE_DIR"
```
philo-he: Ditto. $CURRENT_DIR is accessible inside the function.

PHILO-HE (commented):
@Zouxxyy, thanks for the update. I posted some minor comments, which is not related to your changes. Please help fix them if it makes sense to do so.

Zouxxyy:
I removed the useless variables, CC @PHILO-HE

PHILO-HE (approved):
Looks good. Thanks.

Issue #10613, opened 2025-09-03
Add version and experimental method to `ConfigBuilder`


zjuwangg:
### Description

Like Spark does, we can add `version` method and add `experimental` flag to make config more maintainable and more readable.

### Gluten version

main branch

PR #10659, opened 2025-09-09
[GLUTEN-10613][VL] Add `experimental` method to ConfigBuilder and document experimental config

zjuwangg:

 Add `experimental` method to ConfigBuilder and add experimental config to doc as in #10613 

## How was this patch tested?
NA


zjuwangg:
@zhouyuan @PHILO-HE please help review this PR of improving config doc.

zhztheplayer (commented):
With the API added, we may be able to implement a warning for the case that experimental configurations are set when Gluten is loaded?

zjuwangg:
> With the API added, we may be able to implement a warning for the case that experimental configurations are set?

yeah, I think so.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -538,6 +538,7 @@ object VeloxConfig {
   val VELOX_BROADCAST_BUILD_RELATION_USE_OFFHEAP =
     buildConf("spark.gluten.velox.offHeapBroadcastBuildRelation.enabled")
       .internal()
+      .experimental()
```
philo-he, line 540: If we always assume experimental configs fall under the internal category, I think it would be better to add an assertion check in the experimental API to ensure they have been marked as internal. Otherwise, I'm concerned that developers might forget to mark experimental configs as internal, potentially making experimental configs public unintentionally.

zjuwangg: @PHILO-HE I have also thought add check here during development. But `experimental` and `internal` behavior should be orthogonality. And if we add check, we must first call `internal` and then call `experimental` which also is confusing.

philo-he: Thanks for your reply. My point is we should establish a clear convention to developers and users. It may sound good to make internal and experimental independent.

We can view internal configs as those intended for developers or advanced users only. They can also be experimental—for example, some unstable tracing configs for debugging can be experimental. On the other hand, not all experimental configs are internal. If they can be exposed to end users for early evaluation even though still under experimental, they shouldn't be marked as internal.

Possible rule
1. Public/internal: Who should see/use it
2. Experimental: How reliable it is (can be marked as either public or internal based on the first item)

Based on this rule, many experimental configs can be marked as public if they can be exposed to end users.

If this makes sense, we need to update the comments for the internal API, where experimental configs are previously viewed as internal always. And we also need to note that some experimental configs may fall under public category and some under internal category. Then, I feel it may be not good to have a dedicated section to document experimental configs, considering possible duplication. Instead, it may be better to just keep two sections in document: public and internal.

Actually, I am not quite sure which convention is clearer to users and developers. It seems always putting experimental in internal category is clearer? Discussion is welcome. Thanks.

zhouyuan: @PHILO-HE @zjuwangg could we move this to a discussion topic for more attention in the community? this is a important update on adding new features, bigger than the purpose of update the documents in this patch.
To me the `experimental` section would be better to be public otherwise it maybe difficult for people to test it

zjuwangg: > Possible rule
>
> Public/internal: Who should see/use it
> Experimental: How reliable it is (can be marked as either public or internal based on the first item)

I second this idea.

> Then, I feel it may be not good to have a dedicated section to document experimental configs, considering possible duplication

Maybe we also can only list public(but not experimental) config and experimental section.

zjuwangg: Just update `internal()` comments and follow following rule to mark config and gen doc.
```
Possible rule

Public/internal: Who should see/use it
Experimental: How reliable it is (can be marked as either public or internal based on the first item)
```

philo-he: @zhouyuan, just created a discussion to bring this to the community's attention: https://github.com/apache/incubator-gluten/discussions/10686

PR #10661, opened 2025-09-09
[VL] Support Velox's preferred_output_batch_bytes config

kevinwilfong:

Velox supports a config preferred_output_batch_bytes which is a soft/best effort limit on the number of bytes consumed in
a batch output by an Operator. It is used alongside max_output_batch_rows and in place of preferred_output_batch_rows, and applies when the number of bytes in rows can be estimated and that estimate exceeds preferred_output_batch_bytes before max_output_batch_rows is reached. This can help the system to adapt to very wide rows without OOM'ing.

In addition to plumbing this config through Spark to Velox, I also modified RowToVeloxColumnarExec to respect preferred_output_batch_bytes. Here I limit the Arrow buffer it allocates to hold preferred_output_batch_bytes + whatever overflow is needed to hold the last row that causes us to exceed the limit (again it's a soft limit).

We are currently using the default in Velox so users shouldn't see any change in behavior there unless they choose to set the config. Users may see smaller batches coming from RowToVeloxColumnarExec if the rows are currently exceeding the limit as we were previously not imposing the limit there.

## How was this patch tested?

Verified locally, we were seeing OOMs coming from RowToVeloxColumnarExec due to very wide rows.

Added a unit test to check that preferred_output_batch_bytes is respected in RowToVeloxColumnarExec.

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
@@ -80,6 +80,8 @@ class VeloxConfig(conf: SQLConf) extends GlutenConfig(conf) {
 
   def enableEnhancedFeatures(): Boolean = ConfigJniWrapper.isEnhancedFeaturesEnabled &&
     getConf(ENABLE_ENHANCED_FEATURES)
+
+  def veloxPreferredBatchBytes: Long = getConf(COLUMNAR_VELOX_PREFERRED_BATCH_BYTES)
 }
 
 object VeloxConfig {
@@ -702,4 +704,10 @@ object VeloxConfig {
       .doc("Enable some features including iceberg native write and other features.")
       .booleanConf
       .createWithDefault(true)
+
+  val COLUMNAR_VELOX_PREFERRED_BATCH_BYTES =
+    buildConf("spark.gluten.sql.columnar.backend.velox.preferredBatchBytes")
+      .internal()
+      .bytesConf(ByteUnit.BYTE)
+      .createWithDefaultString("10MB")
 }
--- a/backends-velox/src/main/scala/org/apache/gluten/datasource/ArrowCSVFileFormat.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/datasource/ArrowCSVFileFormat.scala
@@ -17,6 +17,7 @@
 package org.apache.gluten.datasource
 
 import org.apache.gluten.columnarbatch.ColumnarBatches
+import org.apache.gluten.config.VeloxConfig
 import org.apache.gluten.exception.SchemaMismatchException
 import org.apache.gluten.execution.RowToVeloxColumnarExec
 import org.apache.gluten.iterator.Iterators
@@ -312,7 +313,8 @@ object ArrowCSVFileFormat {
     val veloxBatch = RowToVeloxColumnarExec.toColumnarBatchIterator(
       it,
       schema,
-      batchSize
+      batchSize,
+      VeloxConfig.get.veloxPreferredBatchBytes
     )
     veloxBatch
       .map(v => ColumnarBatches.load(ArrowBufferAllocators.contextInstance(), v))
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/RowToVeloxColumnarExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/RowToVeloxColumnarExec.scala
@@ -18,7 +18,7 @@ package org.apache.gluten.execution
 
 import org.apache.gluten.backendsapi.BackendsApiManager
 import org.apache.gluten.columnarbatch.ColumnarBatches
-import org.apache.gluten.config.GlutenConfig
+import org.apache.gluten.config.{GlutenConfig, VeloxConfig}
 import org.apache.gluten.iterator.Iterators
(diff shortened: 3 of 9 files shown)


kevinwilfong:
The failures in Gluten ClickHouse CI and Velox backend Github Runner (x86) / spark-test-spark35-ras do not look to me to be related to my change.

kevinwilfong:
@zhztheplayer Thanks for taking a look!

> How do we handle the relationship between spark.gluten.sql.columnar.maxBatchSize and the new option?

I was thinking of spark.gluten.sql.columnar.maxBatchSize as analogous to max_output_batch_rows in Velox, since Gluten assigns the latter using the former. In Velox we consider both preferred_output_batch_bytes and max_output_batch_rows when determining the size of a batch and applying whichever limit we hit first. I was thinking to do the same in Gluten with spark.gluten.sql.columnar.maxBatchSize and spark.gluten.sql.columnar.backend.velox.preferredBatchBytes to be consistent. I applied this approach in RowToVeloxColumnar for example.

I think there's value in keeping both. Looking at Velox, preferred_output_batch_bytes is not universally applied, our FilterProject operator does not apply it as there is no buffering. If there is no filter the number of rows in the output batch is the same as in the input batch.  As a toy example, suppose the input batch has rows of a single 4 byte integer, and the FilterProject has a UDF the creates arrays of integers from 1..N where N is the value in the row, here we would need some way to limit the number of rows in a batch independent of their size to prevent OOMs while also not ending up with unreasonably small batches in other parts of the query plan.

> There is also spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInputOuptut.minSize by the way.

I admit I'm still familiarizing myself with the Gluten code base, but looking at VeloxBatchResizer if we could get a reasonably accurate estimate for the size in bytes of the batch coming in, we could apply preferred_output_batch_bytes there as well. Likewise, I think this would enhance spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInputOuptut.minSize rather than replace it.

FelixYBW (approved):

FelixYBW:
@kevinwilfong Have you take a look if all velox operators hornors max_output_batch_rows and preferred_output_batch_bytes? Last year we noted some operators may exceed the max_output_batch_rows. We fixed some of them but not all.

zhztheplayer:
> @zhztheplayer Thanks for taking a look!
> 
> > How do we handle the relationship between spark.gluten.sql.columnar.maxBatchSize and the new option?
> 
> I was thinking of spark.gluten.sql.columnar.maxBatchSize as analogous to max_output_batch_rows in Velox, since Gluten assigns the latter using the former. In Velox we consider both preferred_output_batch_bytes and max_output_batch_rows when determining the size of a batch and applying whichever limit we hit first. I was thinking to do the same in Gluten with spark.gluten.sql.columnar.maxBatchSize and spark.gluten.sql.columnar.backend.velox.preferredBatchBytes to be consistent. I applied this approach in RowToVeloxColumnar for example.
> 
> I think there's value in keeping both. Looking at Velox, preferred_output_batch_bytes is not universally applied, our FilterProject operator does not apply it as there is no buffering. If there is no filter the number of rows in the output batch is the same as in the input batch. As a toy example, suppose the input batch has rows of a single 4 byte integer, and the FilterProject has a UDF the creates arrays of integers from 1..N where N is the value in the row, here we would need some way to limit the number of rows in a batch independent of their size to prevent OOMs while also not ending up with unreasonably small batches in other parts of the query plan.
> 
> > There is also spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInputOuptut.minSize by the way.
> 
> I admit I'm still familiarizing myself with the Gluten code base, but looking at VeloxBatchResizer if we could get a reasonably accurate estimate for the size in bytes of the batch coming in, we could apply preferred_output_batch_bytes there as well. Likewise, I think this would enhance spark.gluten.sql.columnar.backend.velox.resizeBatches.shuffleInputOuptut.minSize rather than replace it.

Somehow I misread the PR :(. Thought you were to add [this option](https://github.com/facebookincubator/velox/blob/71a310307a734e170206163efb504afb7e6d834c/velox/core/QueryConfig.h#L209-L214) as an independent Gluten option. Totally fine if we are justing adding the one for byte size.

Thank you for the inputs by the way.

zhztheplayer:

kevinwilfong:
> Somehow I misread the PR :(. Thought you were to add [this option](https://github.com/facebookincubator/velox/blob/71a310307a734e170206163efb504afb7e6d834c/velox/core/QueryConfig.h#L209-L214) as an independent Gluten option. Totally fine if we are justing adding the one for byte size.
> 
> Thank you for the inputs by the way.

@zhztheplayer  Np! Thanks for the review! And just to confirm, my only intention here was to add the option for byte size :)

kevinwilfong:
> @kevinwilfong Have you take a look if all velox operators hornors max_output_batch_rows and preferred_output_batch_bytes? Last year we noted some operators may exceed the max_output_batch_rows. We fixed some of them but not all.

@FelixYBW I vaguely remember some fixes a while ago, but I was not aware there are still operators with issues. Is there an open Issue in the Velox repo with those operators? That's a bug we probably should be addressing on the Velox side.

FelixYBW:
> @FelixYBW I vaguely remember some fixes a while ago, but I was not aware there are still operators with issues. Is there an open Issue in the Velox repo with those operators? That's a bug we probably should be addressing on the Velox side.

Can't find the issue. hopefully they are all fixed.

PR #6572, opened 2024-07-24
[GLUTEN-5320][VL] Reduce driver memory footprint by postpone the creation and serialization of LocalFilesNode

Yohahaha:
## What changes were proposed in this pull request?

Followup of https://github.com/apache/incubator-gluten/pull/5321

(Fixes: \#5320)

PR #10662, opened 2025-09-09
[CORE] Defer Protobuf serialization of SplitInfos in GlutenPartitions

kevinwilfong:

Today the GlutenPartition objects contain an array of byte arrays which are the Protobuf serialized ReadRel.read_type objects from the SplitInfos. The GlutenPartitions are Java serialized and sent to the Executors responsible for their respective Tasks. Looking through the code it appears we Protobuf serialize the SplitInfos so we can easily pass them across the JNI boundary.

We see the serialized SplitInfos can consume a significant amount of memory in the Driver, this is because as SplitInfo objects their state can share references tot he same objects, but once serialized they share nothing, which explodes their size in memory.

If we Java serialize the SplitInfo objects like the rest of the GlutenPartition state, and Protobuf serialize them as part of the Task, this can significantly save driver memory. The cost is a little additional memory in the Task, the size of the SplitInfo objects for a single GlutenPartition which should be trivial, and a little additional CPU instead of Protobuf serializing in the Driver and Java serializing the array of byte arrays, we Java serialize the array of SplitInfos, and on the Task we pay the additional cost of Java deserializing an array of SplitInfos and Protobuf serializing them, overall the difference is just the additional cost of Java serializing the SplitInfos instead of byte arrays.

## How was this patch tested?

Ran the existing unit tests.


kevinwilfong:
It looks like tests are failing due to https://github.com/apache/incubator-gluten/issues/10671 (not related to this change)

kevinwilfong:
@marin-ma  Thanks for the review!

> Does it mean the driver memory can be decreased with this patch because java serialisation only serialise the same object only once?

I suspect the reason is that Spark Java serializes the GlutenPartitions as needed and does not hold the serialized values in memory for a long time. In Gluten, we're currently Protobuf serializing the SplitInfos when we create the GlutenPartitions, and I see a large number of these GlutenPartitions getting held in the Driver's memory while the query is running, so the serialized SplitInfos all exist together at the same time. If Spark is Java serializing the GlutenPartitions only when a Task is ready to execute, and evicts the serialized value from memory as soon as it's been sent to the Executor, with this change we'll only end up with a relatively small number of serialized values present in the Driver's memory at the same time (proportional to the number of Executors).

Yohahaha:
https://github.com/apache/incubator-gluten/pull/6572

We have another pr to decrease driver memory pressure, just post here to see if we can apply.

Issue #10666, opened 2025-09-10
[VL] optimize sort window to streaming window if child order already satisfied


zjuwangg:
### Description

We can optimize sort window to streaming window if child order already satisfied

### Gluten version

main branch

zjuwangg:
No need since we have removed sort window in https://github.com/apache/incubator-gluten/issues/10731

PR #10667, opened 2025-09-10
[GLUTEN-10666][VL]Optimize sort window to streaming window if child order matchs

zjuwangg:

Ooptimize sort window to streaming window if child order already satisfied #10666

## How was this patch tested?
new added test

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
@@ -528,16 +528,66 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
     withSQLConf(VeloxConfig.COLUMNAR_VELOX_WINDOW_TYPE.key -> "sort") {
       runQueryAndCompare(
         """
-          |select sum(l_partkey + 1) over
-          |   (partition by l_suppkey order by l_orderkey)
-          |    from
-          |    (  select * from lineitem
-          |    sort by l_suppkey, l_orderkey
-          |    )
+          |SELECT  avg(sum_part_key) OVER (
+          |            PARTITION BY
+          |                    l_suppkey
+          |            ORDER BY
+          |                    l_orderkey
+          |        ) AS avg_sum_part_key
+          |FROM    (
+          |            SELECT  l_suppkey,
+          |                    l_orderkey,
+          |                    sum(l_partkey+1) OVER (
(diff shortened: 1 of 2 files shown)


zjuwangg:
cc @WangGuangxin please help have a look

backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
```diff
@@ -524,6 +524,74 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
     }
   }
 
+  test("optimize sort window to streaming window") {
+    withSQLConf(VeloxConfig.COLUMNAR_VELOX_WINDOW_TYPE.key -> "sort") {
```
JkSelf, line 528: @zjuwangg Just curious why you choose to use sort window not straeming window in you env? Because the sort operator already before the window opeartor, and then we can directly use the StreamigWindow in velox side.

I plan to remove the SortWindow type and switch to using the StreamingWindow directly, since Spark already inserts a sort operator before the window operator. This means we no longer need to apply the `EliminateLocalSort `rule to remove the sort, and then use `EnsureRequirements ` to re-insert it when the window type is streaming.
Initially, we provided both window types because we believed the SortWindow could support spilling and help fix some OOM scenarios. However, in Velox, spilling in SortWindow is only supported during the addInput stage, not during getOutput, so it doesn't effectively address most OOM cases. As a result, we developed `PartitionStreamingWindow `and `RowsStreamingWindow `to better handle these situations.

zjuwangg: > Just curious why you choose to use sort window not straeming window in you env? Because the sort operator already before the window opeartor, and then we can directly use the StreamigWindow in velox side.

In fact, in our env we also using streaming window instead of sort window. But we observed such optimization can be made when sort window is used.

zjuwangg: >  plan to remove the SortWindow type and switch to using the StreamingWindow directly, since Spark already inserts a sort operator before the window operator.

Also make sense to me.

JkSelf: @zjuwangg We've merged PR #10734 (https://github.com/apache/incubator-gluten/pull/10734), which removes the sort window functionality. This PR can now be closed. Thanks.

PR #10674, opened 2025-09-11
[GLUTEN-10629][FLINK] Support udf count_char

shuai-xu:

Some minor changes:
1. support count_char udf and extract for q14.
2. small refine for log.
3. support TinyInt and Symbol type.

## How was this patch tested?

This patch was tested by manually running nexmark tests.

### DIFF
--- a/gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/RexCallConverterFactory.java
+++ b/gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/RexCallConverterFactory.java
@@ -99,7 +99,9 @@ public class RexCallConverterFactory {
           Map.entry("IS NOT NULL", Arrays.asList(() -> new DefaultRexCallConverter("is_not_null"))),
           Map.entry(
               "REGEXP_EXTRACT", Arrays.asList(() -> new DefaultRexCallConverter("regexp_extract"))),
-          Map.entry("LOWER", Arrays.asList(() -> new DefaultRexCallConverter("lower"))));
+          Map.entry("LOWER", Arrays.asList(() -> new DefaultRexCallConverter("lower"))),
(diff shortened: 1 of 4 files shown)


gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/RexCallConverterFactory.java
```diff
@@ -99,7 +99,9 @@ public class RexCallConverterFactory {
           Map.entry("IS NOT NULL", Arrays.asList(() -> new DefaultRexCallConverter("is_not_null"))),
           Map.entry(
               "REGEXP_EXTRACT", Arrays.asList(() -> new DefaultRexCallConverter("regexp_extract"))),
-          Map.entry("LOWER", Arrays.asList(() -> new DefaultRexCallConverter("lower"))));
+          Map.entry("LOWER", Arrays.asList(() -> new DefaultRexCallConverter("lower"))),
+          Map.entry("count_char", Arrays.asList(() -> new DefaultRexCallConverter("count_char"))),
```
philo-he, line 103: Do we have a convention that capitalizes the function name as the key like others? i.e., `Map.entry("COUNT_CHAR" `. Though I guess it is case insensitive.

philo-he: Has Velox already supported this function?

shuai-xu: This is a udf created by nexmrk, flink seems to will not convert udf names to upper case. The velox has already contains the implement of this udf.

PHILO-HE (approved):
Looks good. Thanks.

PR #10695, opened 2025-09-12
[CORE] Minor cleanups for Delta support code

zhztheplayer:
1. Use constants over string literals
2. Move rules to `extension` package

### DIFF
--- a/backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
+++ b/backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
@@ -18,8 +18,7 @@ package org.apache.gluten.component
 
 import org.apache.gluten.backendsapi.velox.VeloxBackend
 import org.apache.gluten.config.GlutenConfig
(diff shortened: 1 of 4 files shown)


zhztheplayer:
@jinchengchenghh Thanks.

gluten-delta/src/main/scala/org/apache/gluten/execution/DeltaScanTransformer.scala
```diff
@@ -53,8 +54,8 @@ case class DeltaScanTransformer(
   override protected def doValidateInternal(): ValidationResult = {
     if (
       requiredSchema.fields.exists(
-        _.name == "__delta_internal_is_row_deleted") || requiredSchema.fields.exists(
-        _.name == "__delta_internal_row_index")
+        _.name == DeltaParquetFileFormat.IS_ROW_DELETED_COLUMN_NAME) || requiredSchema.fields
+        .exists(_.name == DeltaParquetFileFormat.ROW_INDEX_COLUMN_NAME)
```
zhztheplayer: Delta 2.3 doesn't have the constants, will revert these two lines.

PR #10696, opened 2025-09-12
[CORE] Fix FallbackByNativeValidation attempts to offload a whole tree than a single node

zhztheplayer:
As title. FallbackByNativeValidation is supposed to handle a single node. We currently pass a tree to the offload attempt which overkills the goal. It's functioning but may impact the performance.

The patch fixes the issue.


zml1206 (approved):
LGTM, thanks.

zhztheplayer:
Thanks @zml1206 .

PR #10697, opened 2025-09-12
[VL] Support mapping columns by position index for ORC and Parquet files

kevinwilfong:

In our data warehouse we support schema evolution by column index rather than by name. E.g. if a Hive table has schema a, b, c but the partition has schema c, a, b we won't reorder the columns from the partition, but read partition column c as column a, partition column a as column b, etc.

This is supported in Velox by setting the configs hive.orc.use-column-names and hive.parquet.use-column-names in the HiveConfig to false for ORC and Parquet files respectively. Currently these are both hard coded to true in Gluten. This change adds configs to Gluten's VeloxConfig spark.gluten.sql.columnar.backend.velox.orcUseColumnNames and spark.gluten.sql.columnar.backend.velox.parquetUseColumnNames and plumbs these to the HiveConfig in Velox.

In addition, we need to pass the full table schema to the HiveTableHandle, as this is how Velox determines the indices of each column. I updated VeloxIteratorApi to set the FileSchema for the LocalFilesNodes it generates if necessary (if the config is enabled for the format of the file), and VeloxPlanConverter/SubstraitToVeloxPlan to propagate this to the HiveTableHandle when present.

Note that I considered just setting it in the ReadRel rather than in each LocalFilesNode. This however introduced the problem that we could no longer read from tables with column types we don't support, even if we don't read those columns, as we still need to propagate them to the HiveTableHandle. Since partition file formats don't always match table file formats, we don't know if we need the schema until we generate the splits, at which point it's too late to update the plan. See https://github.com/apache/incubator-gluten/pull/10569

Please note that vanilla Spark partially supports matching by the position index https://issues.apache.org/jira/browse/SPARK-32864 and can be extended to do so by customizing readers.

## How was this patch tested?

Added tests for ORC and Parquet files where the column names in the table don't match the column names in the file, and verified we could still read them by index when the flags are enabled.


kevinwilfong:
@Yohahaha  I know you had concerns about Driver memory on https://github.com/apache/incubator-gluten/pull/10569 I didn't see any meaningful difference after https://github.com/apache/incubator-gluten/pull/10662 landed

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxIteratorApi.scala
```diff
@@ -49,9 +50,25 @@ import scala.collection.JavaConverters._
 
 class VeloxIteratorApi extends IteratorApi with Logging {
 
+  private def setFileSchemaForLocalFiles(
+      localFilesNode: LocalFilesNode,
+      dataSchema: StructType,
```
Yohahaha: maybe `fileSchema` align with localFilesNode?

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
```diff
@@ -207,4 +207,46 @@ class VeloxScanSuite extends VeloxWholeStageTransformerSuite {
...
+          spark
+            .range(2)
+            .selectExpr("id as a", "cast(id + 10 as string) as b")
+            .write
+            .mode("overwrite")
+            .parquet(path)
+
+          withTable("test") {
+            sql("create table test (c long, d string) using parquet options (path '" + path + "')")
+            val df = sql("select c, d from test")
```
Yohahaha: could you test `select d from test`？

zhztheplayer:
Thanks for the PR and detailed PR description. While I am a bit curious about the alignment with vanilla Spark - IIRC Spark hasn't yet supported index-based schema evolution? Are there any similar enhancements like this for your internal JVM Spark?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxIteratorApi.scala
kevinwilfong: Sure, I chose dataSchema to be consistent with the ScanExecTransformer API, but you're right, by the time it gets down here fileSchema is more appropriate.

kevinwilfong:
> Thanks for the PR and detailed PR description. While I am a bit curious about the alignment with vanilla Spark - IIRC Spark hasn't yet supported index-based schema evolution? Are there any similar enhancements like this for your internal JVM Spark?

Thanks for taking a look @zhztheplayer. Admittedly most of my past experience outside of Velox comes from Hive (many years ago) and Presto both of which supported it (if Spark doesn't support it I'm not sure what the current state of it is in Hive). It looks like you're right, Spark only supports a flavor of it for ORC https://issues.apache.org/jira/browse/SPARK-32864 Internally it looks like we're using our own Java readers/writers. Fortunately with Gluten/Velox we don't need to do that as long as we can configure it correctly.

I was planning to reuse most of the code in this change for supporting Text files (Gluten already supports this with ClickHouse), we need to be able to propagate the data schema to the readers for that. The configs are really the only thing not necessarily aligned with vanilla Spark.

With the changes to propagate the table schema (for Text file), I could see extending VeloxIteratorApi internally and overriding  setFileSchemaForLocalFiles to enable passing the schema for DwrfReadFormat (which admittedly is what I'm mostly interested in).

But the connector configs in C++ don't seem very extensible at the moment. I think I would have to fork ConfigExtractor, VeloxBackend, or WholeStageResultIterator to get the value of the kOrcUseColumnNames config set to what I need.

If you don't want those configs explicitly plumbed, a more generic way I can think of to handle this might be to add a config to Gluten whose value is a comma separated list of configs to extract and add to the connector config.  That way we could plumb the configs through the C++ code without forking the files.

Let me know what you think

zhztheplayer:
Thank you @kevinwilfong. I will take a closer look tomorrow.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
```diff
@@ -207,4 +207,64 @@ class VeloxScanSuite extends VeloxWholeStageTransformerSuite {
...
+
+            df = sql("select d from test")
+            checkAnswer(df, Seq(Row("10"), Row("11")))
+
+            df = sql("select c from test")
+            checkAnswer(df, Seq(Row(0L), Row(1L)))
+
+            df = sql("select d, c from test")
+            checkAnswer(df, Seq(Row("10", 0L), Row("11", 1L)))
+          }
```
rui-mo, line 303: Can we also add tests for selecting more columns than the data columns?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1272,26 +1272,30 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
...
 
-  // Velox requires Filter Pushdown must being enabled.
-  bool filterPushdownEnabled = true;
   auto names = colNameList;
   auto types = veloxTypeList;
-  auto dataColumns = ROW(std::move(names), std::move(types));
+  // The columns we project from the file.
+  auto baseSchema = ROW(std::move(names), std::move(types));
+  // The columns present in the table, if not available default to the baseSchema.
+  auto tableSchema = splitInfo->tableSchema ? splitInfo->tableSchema : baseSchema;
```
rui-mo, line 1337: > we need to pass the full table schema to the HiveTableHandle,

Is table schema set only when mapping by index?

rui-mo (commented):
Thanks.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
kevinwilfong: For now, when I add support for TextFiles it will be set when the file format is text as well.

kevinwilfong: Just for clarity, we could always set the full table schema in the HiveTableHandle and Velox would work (whether or not we're mapping by index or reading from a text file).  The reason I only set it when it's absolutely necessary is because if the table schema contains types that aren't supported in the Substrait, sending the table schema would cause the query to fail (see my reasoning for not putting it in the ReadRel in the description).

rui-mo: > if the table schema contains types that aren't supported in the Substrait, sending the table schema would cause the query to fail

Can we say that setting the table schema can lead to unexpected failure even when matching by index?

kevinwilfong: That's correct, the query would fail when serializing the SplitInfos to Protobuf.  Would docs/velox-configuration.md be the right place to add a warning about this?

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
```diff
@@ -207,4 +207,78 @@ class VeloxScanSuite extends VeloxWholeStageTransformerSuite {
...
+            .selectExpr("id as a", "cast(id + 10 as string) as b")
+            .write
+            .mode("overwrite")
+            .orc(path)
+
+          withTable("test") {
+            sql(s"""create table test (c long, d string, e float) using orc options
+                   |(path '$path')""".stripMargin)
+            var df = sql("select c, d from test")
+            checkAnswer(df, Seq(Row(0L, "10"), Row(1L, "11")))
```
Yohahaha: would you replace with `runQueryAndCompare() {checkGlutenOperatorMatch[xxx]}` ?

Yohahaha (commented):
Add one comments for UT, thank you!

cpp/velox/substrait/SubstraitToVeloxPlan.cc
rui-mo: Yes, I assume it would be helpful to document this configuration. Would you please clarify what kind of table schema is not supported and would cause failure ? Thanks.

kevinwilfong: Got it, will do.

As an alternative I could make the Validator more pessimistic when either of these options are set and always validate the entire table schema, then fallback to Vanilla Spark to avoid the failure. I'm not sure which is preferrable (falling back to Vanilla Spark more often or failing in what I'm assuming will be rare cases).

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
kevinwilfong: @Yohahaha  Unfortunately, I tried and it doesn't work because this compares against Vanilla Spark and Vanilla Spark doesn't support this out of the box (see the discussion with Rui above).

I made some changes to add checkGlutenOperatorMatch, disallow fallback, and refactor it to clean it up a bit, let me know what you think.

Yohahaha: > Unfortunately, I tried and it doesn't work because this compares against Vanilla Spark and Vanilla Spark doesn't support this out of the box (see the discussion with Rui above).

does vanilla spark need any extra configs? 

users always expects gluten's behavior align with spark, if not, I think we need rethinking this feature. @rui-mo

cpp/velox/substrait/SubstraitToVeloxPlan.cc
kevinwilfong: Added the documentation

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxScanSuite.scala
kevinwilfong: See my comment here https://github.com/apache/incubator-gluten/pull/10697#issuecomment-3325592859

I will reuse most of the code in this change to support Text files, these are supported in Vanilla Spark and already supported in Gluten with ClickHouse.

The configs are the only thing that don't completely align with Vanilla Spark.  In Vanilla Spark we support mapping columns by index by adding custom readers that support mapping columns by index instead of by name. This is partially supported in Vanilla Spark with ORC. https://issues.apache.org/jira/browse/SPARK-32864 With Velox we simply do this by setting a config in the connector. The C++ code in Gluten is very rigid and doesn't allow us to adjust connector configs to our needs without forking files. In addition, we need to pass the table schema in the split, and we don't want to always do this because Gluten doesn't support the full Spark type system yet, so like for Text files we only want to enable passing the schema when the config is enabled. For users who don't want this, they can simply not use the configs.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
rui-mo: > I'm not sure which is preferrable (falling back to Vanilla Spark more often or failing in what I'm assuming will be rare cases).

If possible, I’d prefer to fallback rather than fail for unsupported schema.

kevinwilfong: @rui-mo  updated to fallback if either setting is applied and the table has an unsupported column type

rui-mo: Thanks!

cpp/velox/compute/VeloxPlanConverter.cc
```diff
@@ -98,18 +101,40 @@ std::shared_ptr<SplitInfo> parseScanSplitInfo(
         splitInfo->format = dwio::common::FileFormat::UNKNOWN;
         break;
     }
+
+    if (!splitInfo->tableSchema && file.has_schema()) {
```
rui-mo, line 109: nit: Would you please add some comments to clarify when this case occurs?

kevinwilfong:
> Here’s what comes to mind. There are the below three strategies for column mapping:
> 
> 1. Match by position
> 2. Match by field name
> 3. Match by unique permanent ID
> 
> And I suppose Spark only supports (2) and (3) (seeing [facebookincubator/velox#6065 (comment)](https://github.com/facebookincubator/velox/discussions/6065#discussioncomment-6698452)), while Velox supports (1) and (2). Would you please clarify which is supported in this PR?

@rui-mo  This exposes Velox's support for (1)

rui-mo:
@kevinwilfong I wonder if Spark supports matching by the position index, and I assumed it only supports matching by the file ID or column name. Please correct me if it's wrong.

kevinwilfong:
Vanilla Spark partially supports it https://issues.apache.org/jira/browse/SPARK-32864 and can be extended to do so by customizing readers (it's what we do)

rui-mo:
cc: @zhztheplayer If there are no further comments, we can proceed to merge this PR.

zhztheplayer:
@rui-mo Please proceed. Thank you very much for the review.

beliefer:
@kevinwilfong I encountered the issue too. I picked up this PR to my Gluten. But the problem still exists.
We created the table with Hive client.
```
CREATE TABLE default.test_orc_table_hive_gluten
(
    id int,
    name string
)
PARTITIONED BY (dt string)
STORED AS ORC;

insert into test_orc_table_hive_gluten partition(dt='20240728') values (1, 'a'),(2,'b');
```
And query this table with Spark SQL.
```
select * from test_orc_table_hive_gluten where dt = '20240728';
```
The output show below.
```
NULL	NULL	20240728
NULL	NULL	20240728
```
I set these configs show below.
```
set spark.gluten.sql.columnar.backend.velox.orcUseColumnNames=false; 
set spark.gluten.sql.complexType.scan.fallback.enabled=false;
```

But nothing helped.

Did I miss something?

kevinwilfong:
@beliefer That's strange, do you get the same results if you don't set spark.gluten.sql.columnar.backend.velox.orcUseColumnNames to false?

Nothing in your repro changes the schema of your data, so it should work regardless of the value of that config

beliefer:
@kevinwilfong Yes. No matter the value of `spark.gluten.sql.columnar.backend.velox.orcUseColumnNames`, the output is wrong.

kevinwilfong:
In that case, as was mentioned in the Issue that was linked, I think it's not related to this change.

PR #10706, opened 2025-09-15
[VL] Make PartialProject support struct with null fields

jiangjiangtian:
For now, if PartialProject evaluates expressions that accept data whose type is struct and its input has NULL values for some fields, the evaluation will throw out exception like follows:
```
Reason: Operator::getOutput failed for [operator: ValueStream, plan node ID: 0]: Error during calling Java code from native code: java.lang.IllegalStateException: Value at index is null
	at org.apache.arrow.vector.BitVector.get(BitVector.java:252)
	at org.apache.gluten.vectorized.ArrowWritableColumnVector$BooleanAccessor.getBoolean(ArrowWritableColumnVector.java:913)
	at org.apache.gluten.vectorized.ArrowWritableColumnVector.getBoolean(ArrowWritableColumnVector.java:471)
	at org.apache.spark.sql.vectorized.ColumnarRow.getBoolean(ColumnarRow.java:104)
	at org.apache.spark.sql.vectorized.ColumnarRow.get(ColumnarRow.java:162)
	at org.apache.spark.sql.catalyst.expressions.GetStructField.nullSafeEval(complexTypeExtractors.scala:128)
```
The reason is that ColumnarRow in Spark doesn't check whether the field to get is null. This PR fixes this bug.

## How was this patch tested?
Unit tests.

### DIFF
--- a/gluten-arrow/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
+++ b/gluten-arrow/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
@@ -22,6 +22,7 @@ import org.apache.gluten.runtime.Runtimes;
 import org.apache.gluten.utils.ArrowAbiUtil;
 import org.apache.gluten.utils.ArrowUtil;
 import org.apache.gluten.utils.InternalRowUtl;
+import org.apache.gluten.vectorized.ArrowColumnarBatch;
 import org.apache.gluten.vectorized.ArrowWritableColumnVector;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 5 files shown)


WangGuangxin:
In fact, I also create a PR in Spark (https://github.com/apache/spark/pull/52175/files) , but now it is only merged to Spark 4.0. I think it's reasonable to handle in Gluten here if possible

WangGuangxin:
@jiangjiangtian But the compile error seems that it's hard to overwrite `ColumnarRow`

image

jiangjiangtian:
> @jiangjiangtian But the compile error seems that it's hard to overwrite `ColumnarRow`
> 
> image

@WangGuangxin yeah, I find that in Spark-3.2, `ColumnarBatch` is declared as a `final` class. I don't know if there is any way to work around it.

jiangjiangtian:
@WangGuangxin I have fixed the problem, please take a look again.

gluten-arrow/src/main/java/org/apache/gluten/columnarbatch/ColumnarBatches.java
```diff
@@ -171,6 +172,15 @@ public static void checkOffloaded(ColumnarBatch batch) {
     }
   }
 
+  public static ArrowColumnarBatch convertToArrowColumnarBatch(ColumnarBatch sparkColumnarBatch) {
+    int numCols = sparkColumnarBatch.numCols();
+    ArrowWritableColumnVector[] writableColumns = new ArrowWritableColumnVector[numCols];
+    for (int i = 0; i < numCols; i++) {
+      writableColumns[i] = (ArrowWritableColumnVector) sparkColumnarBatch.column(i);
+    }
+    return new ArrowColumnarBatch(writableColumns, sparkColumnarBatch.numRows());
```
WangGuangxin, line 181: do we need to close `sparkColumnarBatch` here?

jiangjiangtian: We needn't, because we don't call method `retain` here. We just transfer the columns in `sparkColumnarBatch` to the result `ArrowColumnarBatch`. Should we call method `retain` here?

WangGuangxin: ok, got it

WangGuangxin (approved):
LGTM

PR #10708, opened 2025-09-15
[GLUTEN-10707] Improve SparkDirectoryUtil by reduce synchronized

### LINKED ISSUE #10707: Improve performance for SparkDirectoryUtil
### Description

Currently, `SparkDirectoryUtil` is designed  as singleton mode.
The implementation of `SparkDirectoryUtil` adopted `synchronized` to keep thread safe.

### Gluten version

main branch

beliefer:

This PR proposes to improve `SparkDirectoryUtil` by reduce `synchronized`
Fixes #10707

## How was this patch tested?

GA tests.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
@@ -67,12 +67,12 @@ class SparkDirectoryUtil private (val roots: Array[String]) extends Logging {
 }
 
 object SparkDirectoryUtil extends Logging {
-  @volatile private var targetRoots: Array[String] = _
+  @volatile private var roots: Array[String] = _
   private lazy val INSTANCE: SparkDirectoryUtil = {
-    if (targetRoots == null) {
+    if (this.roots == null) {
       throw new IllegalStateException("SparkDirectoryUtil not initialized")
     }
-    new SparkDirectoryUtil(targetRoots)
+    new SparkDirectoryUtil(this.roots)
   }
 
   def init(conf: SparkConf): Unit = {
@@ -81,11 +81,11 @@ object SparkDirectoryUtil extends Logging {
   }
 
   private def init(roots: Array[String]): Unit = synchronized {
-    if (targetRoots == null) {
-      targetRoots = roots
-    } else if (targetRoots.toSet != roots.toSet) {
+    if (this.roots == null) {
+      this.roots = roots
+    } else if (this.roots.toSet != roots.toSet) {
       throw new IllegalArgumentException(
-        s"Reinitialize SparkDirectoryUtil with different root dirs: old: ${targetRoots
+        s"Reinitialize SparkDirectoryUtil with different root dirs: old: ${this.roots
             .mkString("Array(", ", ", ")")}, new: ${roots.mkString("Array(", ", ", ")")}"
       )
     }


gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
```diff
@@ -67,29 +68,31 @@ class SparkDirectoryUtil private (val roots: Array[String]) extends Logging {
...
     val roots = Utils.getConfiguredLocalDirs(conf)
     init(roots)
   }
 
-  private def init(roots: Array[String]): Unit = synchronized {
-    if (INSTANCE == null) {
-      INSTANCE = new SparkDirectoryUtil(roots)
+  private def init(roots: Array[String]): Unit = {
+    if (INSTANCE.get() == null) {
+      INSTANCE.compareAndSet(null, new SparkDirectoryUtil(roots))
```
zhztheplayer: The code is workable but the constructor of `SparkDirectoryUtil` may be unnecessarily called in concurrent invocations.

We can use `AtomicBoolean` instead if unable to avoid this.

beliefer: I'm sorry. I didn't understand what you said. Could you tell me more ?

beliefer:

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
zhztheplayer: For example,  thread 1 and thread 2 can both reach line 80 at the same time. Although only one thread will initialize the `INSTANCE`, another thread will still do `new SparkDirectoryUtil(roots)` which is not necessary.

beliefer: Got it.

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
```diff
@@ -67,16 +68,19 @@ class SparkDirectoryUtil private (val roots: Array[String]) extends Logging {
 }
 
 object SparkDirectoryUtil extends Logging {
+  private val state = new AtomicBoolean(false)
```
zhztheplayer: Let's rename it. E.g., `INSTANCE_INITIALIZED`

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
```diff
@@ -67,16 +68,19 @@ class SparkDirectoryUtil private (val roots: Array[String]) extends Logging {
...
   }
 
-  private def init(roots: Array[String]): Unit = synchronized {
+  private def init(roots: Array[String]): Unit = {
     if (INSTANCE == null) {
-      INSTANCE = new SparkDirectoryUtil(roots)
+      if (state.compareAndSet(false, true)) {
+        INSTANCE = new SparkDirectoryUtil(roots)
+      }
       return
```
zhztheplayer: It's not the hot code path, let's remove the `if (INSTANCE == null)` condition.

Just directly call  `compareAndSet`.

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
```diff
@@ -87,7 +91,7 @@ object SparkDirectoryUtil extends Logging {
     }
   }
 
-  def get(): SparkDirectoryUtil = synchronized {
+  def get(): SparkDirectoryUtil = {
     assert(INSTANCE != null, "Default instance of SparkDirectoryUtil was not set yet")
     INSTANCE
   }
```
zhztheplayer: The get operation is not synchronized.

zhztheplayer (changes requested):
The get operation is still not synchronized... I am inclined to keep the current main code. Because this is not the hot code path and is not complicated. Do you agree?

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
beliefer: Because remove `if (INSTANCE == null)` will cause NPE. So I removed `if (INSTANCE.roots.toSet != roots.toSet)`.

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
beliefer: Now, it is guarded with `INSTANCE_INITIALIZED`.

beliefer:
> The get operation is still not synchronized... I am inclined to keep the current main code. Because this is not the hot code path and is not complicated. Do you agree?

Now, the issue has been fixed.

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
```diff
@@ -67,28 +68,24 @@ class SparkDirectoryUtil private (val roots: Array[String]) extends Logging {
...
+      logWarning("SparkDirectoryUtil is already initialized")
     }
   }
 
-  def get(): SparkDirectoryUtil = synchronized {
-    assert(INSTANCE != null, "Default instance of SparkDirectoryUtil was not set yet")
+  def get(): SparkDirectoryUtil = {
+    assert(INSTANCE_INITIALIZED.get(), "Default instance of SparkDirectoryUtil was not set yet")
     INSTANCE
   }
```
zhztheplayer: Hi @beliefer, thank you for keeping iterating the code, but it is still problematic :(

When thread 1 reaches line 81 but hasn't yet set `INSTANCE`, thread 2 can pass line 88 and access `INSTANCE` which may give an unexpected result to caller.

I know it's a corner case, but we should make sure the new code covers what is covered by the old code completely.

beliefer: Thank you for the explanation. I will try other way.

beliefer:
@zhztheplayer Could you help me review this again ?

gluten-core/src/main/scala/org/apache/spark/util/SparkDirectoryUtil.scala
```diff
@@ -67,30 +67,31 @@ class SparkDirectoryUtil private (val roots: Array[String]) extends Logging {
...
 
   private def init(roots: Array[String]): Unit = synchronized {
-    if (INSTANCE == null) {
-      INSTANCE = new SparkDirectoryUtil(roots)
-      return
-    }
-    if (INSTANCE.roots.toSet != roots.toSet) {
+    if (targetRoots == null) {
+      targetRoots = roots
```
zhztheplayer: Perhaps just use the same name and

```scala
this.roots = roots
```

. Just a nit. Both are fine.

beliefer:
@zhztheplayer Thank you!

PR #10714, opened 2025-09-15
[GLUTEN-10713] Eliminate the c2r before iceberg partition write

Zouxxyy:

Currently, for iceberg partitioned write, an AQE node is inserted before the write node, which results in unnecessary c2r and r2c operations being added when we offload the native write.

Before
image

After
image


jinchengchenghh:
https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarV2TableWriteExec.scala#L45-L51 should eliminate the format conversion, but seems it does not work in some condition, could you help take a look? Thanks! @zhztheplayer

Zouxxyy:
> https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarV2TableWriteExec.scala#L45-L51 should eliminate the format conversion, but seems it does not work in some condition, could you help take a look? Thanks! @zhztheplayer

Yes, the `ColumnarV2TableWriteExec` is fine, but the inserted AQE node (by iceberg partitioned write which will produce a sort) is not a batch node, which causes c2r and r2c to be inserted around it. 

This PR just mark the AQE child of v2 columnar write as supportsColumnar = true.

jinchengchenghh:
Could you find why this cannot solve this problem? https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala#L77

Zouxxyy:
> Could you find why this cannot solve this problem? https://github.com/apache/incubator-gluten/blob/main/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/datasources/GlutenWriterColumnarRules.scala#L77

This just only match the v1 write command

jinchengchenghh:
Could we extend it to support V2? This fix iceberg, but other data lake may also need it.

Zouxxyy:
@jinchengchenghh  I checked that `GlutenWriterColumnarRules` is almost entirely v1 logic, so I create a separate rule `V2WritePostRule` for commen v2 write, what do you think?

Zouxxyy:
@jinchengchenghh @zhztheplayer Can you help review this further? Thanks.

jinchengchenghh (approved):
Looks good for me, would you like to also take a look? @zhztheplayer

PR #10729, opened 2025-09-17
[CORE][VL] Upgrade the default Spark version to Spark 3.5

kerwin-zk:

Spark 4.0 is now supported, and it is recommended to upgrade the default compilation version to Spark 3.5.

## How was this patch tested?

CI


Yohahaha (approved):
👍

PR #10735, opened 2025-09-17
[GLUTEN-10632][FLINK] Add nexmark test for q11, q12 and q22

shuai-xu:

Add nexmark tests to cover q11 q12 and q22 as they have been supportted.
Optimize NexmartTest to support query files have multi create sqls.

## How was this patch tested?

This pr is tests.

### DIFF
--- a/gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
+++ b/gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
@@ -124,6 +124,8 @@ public class NexmarkTest {
     assertThat(sqlStatements.length).isGreaterThanOrEqualTo(2);
 
     for (int i = 0; i < sqlStatements.length - 2; i++) {
+      // For some query tests like q12 q13 q14, the first two of the three statements create tables
+      // or views. For others, there are only two statements, with the first one creating a table.
       String createResultTable = sqlStatements[i].trim();
       if (!createResultTable.isEmpty()) {
         TableResult createResult = tEnv.executeSql(createResultTable);


gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/custom/NexmarkTest.java
```diff
@@ -123,17 +123,20 @@ private void executeQuery(StreamTableEnvironment tEnv, String queryFileName)
     String[] sqlStatements = queryContent.split(";");
     assertThat(sqlStatements.length).isGreaterThanOrEqualTo(2);
 
-    String createResultTable = sqlStatements[0].trim();
-    if (!createResultTable.isEmpty()) {
-      TableResult createResult = tEnv.executeSql(createResultTable);
-      assertFalse(createResult.getJobClient().isPresent());
+    for (int i = 0; i < sqlStatements.length - 2; i++) {
```
philo-he, line 126: Could you explain why we need a loop here?

shuai-xu: Such as q12, there are multi create statement in it contains `create table` and `create function`.

philo-he: Could you add a comment above the code?

```
// For some query tests like q12, the first two of the three statements create tables
// or views. For others, there are only two statements, with the first one creating a table.
```

shuai-xu: done, I will merge it after the ci

PR #10740, opened 2025-09-17
[VL] Delta deletion vector read support for Spark 3.4 + Delta 2.4

zhztheplayer:
This patch adds deletion vector read support for Spark 3.4 + Delta 2.4.

(WIP)


zhztheplayer:
@zzcclp  I see the following error in CH CI:

```
iceberg/src/main/java/org/apache/gluten/substrait/rel/IcebergLocalFilesNode.java:[22,25] error: cannot access DeleteFile
[2025-09-18T11:30:05.371Z]   bad class file: /root/.m2/repository/org/apache/iceberg/iceberg-spark-runtime-3.5_2.13/1.8.0/iceberg-spark-runtime-3.5_2.13-1.8.0.jar(org/apache/iceberg/DeleteFile.class)
[2025-09-18T11:30:05.371Z]     class file has wrong version 55.0, should be 52.0
[2025-09-18T11:30:05.371Z]     Please remove or make sure it appears in the correct subdirectory of the classpath.
```

Is the JDK used still 17?

PR #10744, opened 2025-09-17
[MINOR] Fix pom.xml code format

liujiayi771:

Some of the pom.xml files in the project had inconsistent indentation widths, which looked odd. This PR fixed them and also made some minor style adjustments.


liujiayi771:
cc @zhouyuan @PHILO-HE.

liujiayi771:
@PHILO-HE I will add spotless check to other pom.xml

zhztheplayer:
> Thanks for the fix. Can we enable some format checker to catch formatting issues automatically? The gluten-it module has enabled spotless plugin for checking pom files. Maybe, we can just add such setting for other modules. It is acceptable to do this in another PR. However, we need to ensure the current fix aligns with spotless plugin.
> 
> https://github.com/apache/incubator-gluten/blob/0a2ad390875cdb13746e8977eaa1550145b9ad03/tools/gluten-it/pom.xml#L188-L198

I second this. I haven't noticed the checker wasn't enabled for main Gluten code.

liujiayi771:
@zhztheplayer We need to enable the pom check for spotless plugin in the root pom.xml. I can added it in a separate PR. Currently, only some indentation issues have been fixed.

However, spotless's reorganization feature will cause significant changes across all pom.xml files.

zhztheplayer:
Thank you @liujiayi771, I see https://github.com/apache/incubator-gluten/pull/10755 was opened.

PR #10747, opened 2025-09-18
[VL][CI] Add a format checker for scala code to enable early failure reporting

PHILO-HE:

Previously, code format violations were reported by the CI only after completing the time-consuming native code build. We should perform these preliminary checks earlier to quickly surface issues, potentially saving CI resources.


philo-he:
@zhouyuan, could you take a look? This is part of preliminary checks.

.github/workflows/velox_backend_enhanced.yml
```diff
@@ -13,12 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: Velox backend Github Runner (Enhanced Features)
+name: Velox Backend (Enhanced)
```
zhztheplayer, line 16: nit, once may not figure out the meaning of `Enhanced` at first glance.

dev/format-scala-code.sh
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# If a new profile is introduced for new modules, please add it here to ensure
+# the new modules are covered.
+PROFILES="-Pbackends-velox -Pceleborn,uniffle -Piceberg,delta,hudi,paimon \
+          -Pspark-3.2,spark-3.3,spark-3.4,spark-3.5,spark-4.0 -Pspark-ut"
```
zhztheplayer, line 21: If this is not for a CI task that is only for VL backend, let's add CH's code as well.

.github/workflows/velox_backend_enhanced.yml
philo-he: @zhztheplayer, thanks for the comment. I feel this simple name is descriptive enough, especially since it's intended for developers only. Not a strong preference. I can revert to the full name if you think it's more appropriate. :)

dev/format-scala-code.sh
philo-he: Thanks for the comment. I note all scripts under dev/ are related to velox backend. Perhaps it’s better to continue excluding the CH backend in this script.

philo-he:
@zhztheplayer, could you take a look again? The CI failure has been fixed by this patch, but that job always uses upstream gluten for verification due to some known issues of checkout plugin on Centos-7.

.github/workflows/velox_backend_enhanced.yml
zhztheplayer: No, just keep it if you have strong reason. Never tended to be nitpicking :)

PR #10751, opened 2025-09-18
[VL] Gluten-it: Fix inverted expected vs actual in result mismatch er…

zhztheplayer:
When calling Spark test API `SparkUtils.compareAnswers`, the expected result should be passed by 2nd parameter, not the 1st one.


zhztheplayer:
Thanks @PHILO-HE .

PR #10755, opened 2025-09-18
[MINOR] Enable pom spotless format check

liujiayi771:

Enable spotless pom format checker to catch formatting issues automatically.


lgbo-ustc:
sortPom seems cannot work with a soft-linked directory
```
[ERROR] Step 'sortPom' found problem in 'pom.xml':
Could not save sorted pom file: /tmp/pom8977291333658807347.xml
```
`/tmp` is a link to another directory

PR #10763, opened 2025-09-21
[VL] Make VeloxResizeBatchesExec inherit from ColumnarToColumnarExec to simplify the code

Zouxxyy:

Make VeloxResizeBatchesExec inherit from ColumnarToColumnarExec to simplify the code


Zouxxyy:
CC @zhztheplayer Thanks

zhztheplayer (commented):
Thanks for the PR.

Would you like to help check the UI sanity, especially the metrics? Could post a screenshot here if possible? Thanks!

Zouxxyy:
@zhztheplayer Sure, here is `SELECT max(l_orderkey) FROM lineitem`

one metric description changed
"time to append / split batches" -> "time to convert batches"

image

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition/package.scala
```diff
@@ -65,11 +65,11 @@ package object transition {
     }
   }
 
-  // Extractor for Gluten's C2C
+  // Extractor for Gluten's C2C with different convention
   object ColumnarToColumnarLike {
     def unapply(plan: SparkPlan): Option[SparkPlan] = {
       plan match {
-        case c2c: ColumnarToColumnarTransition =>
+        case c2c: ColumnarToColumnarTransition if !c2c.isSameConvention =>
```
zhztheplayer, line 72: @Zouxxyy 

Sorry I missed out this change. Would you explain the purpose of it? Thanks.

Zouxxyy: `ColumnarToColumnarLike` is used by the Coster to calculate C2C costs. Previously, `VeloxResizeBatchesExec` didn't inherit from `ColumnarToColumnarTransition`, so it wasn't included in the cost calculation. This change is to maintain the previous behavior.

Besides, in my testing, without this, `ELECT max(l_orderkey) FROM lineitem` will not generate a `VeloxResizeBatchesExec` node because it is eliminated after the cost calculation.

zhztheplayer: Currently `VeloxResizeBatchesExec` is added by rule https://github.com/apache/incubator-gluten/blob/cd2c0cca9b9478a050bfbb90f15e75f99e7adcd2/backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala#L31-L59 without costers involved. Am I missing something? How is it eliminated after being added by the rule?

Zouxxyy: @zhztheplayer , Not sure if my understanding is correct, in `InsertTransitions` we removeForNode `ColumnarToColumnarLike`, and then `fillWithTransitions`

```scala
case class InsertTransitions(convReq: ConventionReq) extends Rule[SparkPlan] {
  private val convFunc = ConventionFunc.create()

    // Remove all transitions at first.
    val removed = RemoveTransitions.apply(plan)
    val filled = fillWithTransitions(removed)
    val out = Transitions.enforceReq(filled, convReq)
    out
  }
```

but in VeloxBatchType we have no the Transitions, so after we remove node, we loss VeloxResizeBatchesExec
```scala
object VeloxBatchType extends Convention.BatchType {
  override protected def registerTransitions(): Unit = {
    fromRow(Convention.RowType.VanillaRowType, RowToVeloxColumnarExec.apply)
    toRow(Convention.RowType.VanillaRowType, VeloxColumnarToRowExec.apply)
    fromBatch(ArrowBatchTypes.ArrowNativeBatchType, ArrowColumnarToVeloxColumnarExec.apply)
    toBatch(ArrowBatchTypes.ArrowNativeBatchType, Transition.empty)
  }
}
```

Zouxxyy: It seems that the current c2c is automatically added, and resize is manually added, so we should not remove it. and  whether velox resize should also be included in the cost

zhztheplayer:
Merging, thank you @Zouxxyy. I felt there is opportunity to refactor the code to make it less impact the transition planner, but haven't got chance to experiment. I will revisit later.

PR #10769, opened 2025-09-22
[GLUTEN-10628][FLINK] Add decimal type support to `greaterthan` function

shuai-xu:
Make greater than and less than support decimal type to pass q14

## How was this patch tested?

This patch was tested with nexmark q14 unittest.


PHILO-HE (approved):
Looks good. Thanks.

PR #10774, opened 2025-09-22
[VL] Promote SharedLibraryLoader to SPI

yaooqinn:

This PR promotes SharedLibraryLoader to SPI. Enable developers to create a custom SharedLibraryLoader tailored to their specific environments.

Besides the existing `loadLib` function, a new method `accepts` is defined for `SharedLibraryLoader`s to identify whether it satisfies the given options or os release information.
 
## How was this patch tested?

new tests


backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoaderUtils.scala
```diff
@@ -0,0 +1,94 @@
...
+        val props = new Properties()
+        val in = new FileInputStream("/etc/os-release")
+        props.load(in)
+        (props.getProperty("NAME"), props.getProperty("VERSION"))
+    }
+
+    val loaders = ServiceLoader
+      .load(classOf[SharedLibraryLoader])
+      .asScala
+      .filter(loader => loader.accepts(osName, osVersion))
```
zhztheplayer, line 66: What happens when more than one loaders are returned?

zhztheplayer (commented):
Thanks. May I know the motivation of adding the SPI? Do you place a user-customized shared loader implementation?

yaooqinn:
> Do you place a user-customized shared loader implementation?

Yes, and then I don't need to modify the gluten main code base

backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoaderUtils.scala
yaooqinn: First in, first load && Best effort to load one if successful.

zhztheplayer: Do we already have this use case (multiple loaders returned)? Otherwise let's simply throw an error because the loading priority is not guaranteed?

yaooqinn: This is new. If we leave this conflict error for users, how do they suppose to fix it

zhztheplayer: For example, if user adds their own loader implementation of CentOS 9, does the code ensure their implementation is chosen over the built-in `SharedLibraryLoaderCentos9`?

Can we add a simple priority API to `SharedLibraryLoader`, so the loading process is more deterministic?

yaooqinn: > For example, if user adds their own loader implementation of CentOS 9, does the code ensure their implementation is chosen over the built-in SharedLibraryLoaderCentos9?

The new SharedLibraryLoaderCentos9 can define its own `accepts` logic with osName or osVersion to distinguish the built-in one. Then, configs `loadLibOS/loadLibOSVersion` can help the caller side to decide which one to pick.

> Can we add a simple priority API to SharedLibraryLoader, so the loading process is more deterministic?

I tried, but the priority API didn't mitigate the issue here since the conflict loaders can also have identical priority.

Maybe we shall just throw errors when conflicting

zhztheplayer: Making sense to me then. Thanks for the explanation.

yaooqinn:
Thank you, @zhztheplayer, for the detailed review and suggestions.

zhztheplayer:
No problem. Thank you for the contribution!

PR #10775, opened 2025-09-22
[GLUTEN-10772][CH] Fix crash caused by not releasing `QueryContext` resources properly.

lgbo-ustc:

Fix crash caused by not releasing `QueryContext` resources properly.

This could happen when a query is canceled by YARN, causing the release action for resources in QueryContext::query_map_to not be called as expected."
## How was this patch tested?

test manually


zhanglistar (approved):
LGTM

Issue #10781, opened 2025-09-23
[VL] Add months_between support


zml1206:
### Description

https://github.com/facebookincubator/velox/pull/14909

### Gluten version

None

PR #10782, opened 2025-09-23
[GLUTEN-10781][VL] Add months_between support

zml1206:

Fixes: #10781 

## How was this patch tested?

CI


gluten-ut/spark32/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenDateExpressionsSuite.scala
```diff
@@ -476,4 +476,83 @@ class GlutenDateExpressionsSuite extends DateExpressionsSuite with GlutenTestsTr
       }
     }
   }
+
+  testGluten("months_between") {
```
zhouyuan, line 480: it looks like this test is added for all 4 spark versions, also there is a new test for velox backend, is this intended?

zml1206: This is a rewritten spark unit test, and the other is a test to ensure offload.

zml1206:
Thanks for review. @zhouyuan @PHILO-HE

PR #10783, opened 2025-09-23
[VL] tools/workload/tpch: Delete temporary files generated by dbgen

xinghuayu007:

When using TPCH shell script to generate test data, the script got stuck when execute it again.
The tool dbGen will generate a lots of temporary files and not be deleted after the first running. When running the tool again, it asks whether to overwrite the temporary files, which make the script stuck.

The patch adds a function to delete these temporary files before running the dbGen tool.


FelixYBW:
Thank you for the fix.

PR #10785, opened 2025-09-23
[VL] tools/workload/tpch: Close sql files to avoid resources leak

xinghuayu007:

SQL files is not close after running, which maybe cause resources leak.
This patch also prints the results in lines for statistics convenience.

### DIFF
--- a/tools/workload/tpch/run_tpch/tpch_parquet.scala
+++ b/tools/workload/tpch/run_tpch/tpch_parquet.scala
@@ -16,7 +16,6 @@
  */
 import org.apache.spark.sql.execution.debug._
 import scala.io.Source
-import scala.collection.mutable
 import java.io.File
 import java.util.Arrays
 import sys.process._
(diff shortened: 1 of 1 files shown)


tools/workload/tpch/run_tpch/tpch_parquet.scala
```diff
@@ -77,11 +70,24 @@ val sorted = fileLists.sortBy {
...
 
 // Main program to run TPC-H testing
+val resultMap = mutable.Map[String, Double]()
 for (t <- sorted) {
-  println(t)
-  val fileContents = Source.fromFile(t).getLines.filter(!_.startsWith("--")).mkString(" ")
-  println(fileContents)
-  time{spark.sql(fileContents).collectAsList()}
-  //spark.sql(fileContents).explain
+  val fileContents = {
```
jinchengchenghh, line 81: Please don't change the function time, just close the file

xinghuayu007: Done. I mean to print the final results for statistics without any middle results. Maybe the middle results is important for  other situations.

PR #10787, opened 2025-09-23
[VL] Add datasource add-split and read metrics

rui-mo:

Add Velox datasource add-split and read metrics in Gluten.

## How was this patch tested?

UT.


FelixYBW:
data source add split time counts the time to add a split to table scan in main thread. The split may or may not be prefetched.

data source read time counts the time to read a page from parquet in table scan in main thread. The page may or maynot be prefetched. it count the decompress time as well. It doesn't count the time for lazy read. Lazy read is counted in following operators or when we do flatten.

Issue #10671, opened 2025-09-10
[VL] Mismatched exception for timestamp_seconds


rui-mo:
### Backend

VL (Velox)

### Bug description

https://github.com/apache/incubator-gluten/actions/runs/17615930994/job/50049264079?pr=9195

Perhaps we need to override the sql file.

### Spark version

None

### Relevant logs

```bash

```

PR #10789, opened 2025-09-23
[GLUTEN-10671][VL] Enable datetime-legacy.sql

rui-mo:

Align the error message with Velox's so the test can be enabled.
Fixes https://github.com/apache/incubator-gluten/issues/10671.

## How was this patch tested?

UT


zhouyuan (approved):
👍

PR #10802, opened 2025-09-25
[VL] Delta: Add Delta Lake write unit test for Spark 3.5 + Delta 3.3

zhztheplayer:
Delta doesn't release its testing Jars so we copy-paste some of their test code here as a quick solution to test Gluten with Delta.

In long-term, we may consider releasing a Delta test jar ourselves, or adopting download-and-patching for the sources to avoid maintaining Delta's test code in our code base.

In this PR, we only introduce one single test suite `DeleteSQLSuite` but it will cover the main Delta write functionality. We will add more in the future.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxColumnarCacheSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxColumnarCacheSuite.scala
@@ -152,11 +152,6 @@ class VeloxColumnarCacheSuite extends VeloxWholeStageTransformerSuite with Adapt
   }
 
   test("Support transform count(1) with table cache") {
-    // scalastyle:off println
-    println(
-      s"spark.gluten.sql.columnar.tableCache=" +
-        spark.conf.get("spark.gluten.sql.columnar.tableCache", "missing"))
-    // scalastyle:on println
     val cached = spark.table("lineitem").cache()
     try {
       val df = spark.sql("SELECT COUNT(*) FROM lineitem")


gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -963,7 +963,7 @@ object GlutenConfig {
       .createWithDefault(100000)
 
   val COLUMNAR_TABLE_CACHE_ENABLED =
-    buildConf("spark.gluten.sql.columnar.tableCache")
+    buildStaticConf("spark.gluten.sql.columnar.tableCache")
```
zhztheplayer, line 966: I suspect this option is somehow leaked among different test suites so caused some strange errors. Let's correct it as static to align with its life scope and doing this will also help diagnosing the issue.

zhouyuan (approved):

backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/DeleteSQLSuite.scala
```diff
@@ -0,0 +1,154 @@
...
+  override protected def executeDelete(target: String, where: String = null): Unit = {
+    val whereClause = Option(where).map(c => s"WHERE $c").getOrElse("")
+    sql(s"DELETE FROM $target $whereClause")
+  }
+
+  override def excluded: Seq[String] = super.excluded ++
+    Seq(
+      // FIXME: Excluded by Gluten as results are mismatch.
+      "test delete on temp view - nontrivial projection - SQL TempView",
+      "test delete on temp view - nontrivial projection - Dataset TempView"
```
zhztheplayer, line 41: TODO: Fix these cases.

backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/DeleteSQLSuite.scala
```diff
@@ -0,0 +1,154 @@
...
+      "data and partition columns - Partition=true Skipping=false",
+      "data and partition columns - Partition=false Skipping=false",
+      // The scan schema contains additional row index filter columns.
+      "nested schema pruning on data condition",
+      // The number of records is not recomputed when using DVs
+      "delete throws error if number of records increases",
+      "delete logs error if number of records are missing in stats",
+      // FIXME: Excluded by Gluten as results are mismatch.
+      "test delete on temp view - nontrivial projection - SQL TempView",
+      "test delete on temp view - nontrivial projection - Dataset TempView"
```
zhztheplayer, line 136: TODO: Fix these cases.

backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/DeleteSQLSuite.scala
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+  }
+}
+
+// FIXME: Enable the test.
+//  Skipping as function input_file_name doesn't get correctly resolved.
+@Ignore
+class DeleteSQLNameColumnMappingSuite extends DeleteSQLSuite
+  with DeltaColumnMappingEnableNameMode {
```
zhztheplayer, line 102: TODO: Fix this suite.

PR #10815, opened 2025-09-28
[GLUTEN-10797][CH] Reject all JNI calls after native destroy is invoked.

lgbo-ustc:

Reject all JNI calls after native destroy is invoked,  to prevent access to invalid resources.
## How was this patch tested?

test manually


zzcclp (approved):
LGTM

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -27,6 +27,8 @@
 #include <Common/ThreadStatus.h>
 #include <Common/formatReadable.h>
 #include <Common/logger_useful.h>
+#include <Parser/LocalExecutor.h>
+#include <iostream>
```
zhanglistar: Should not use iostream

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -71,7 +74,16 @@ void QueryContext::resetGlobal()
 
 void QueryContext::reset()
 {
+    for (auto it = active_executors_.begin(); it != active_executors_.end(); ++it)
+    {
+        auto * executor = *it;
+        executor->cancel();
+        std::cerr << "Not closed LocalExecutor:\n" << executor->dumpPipeline() << std::endl;
```
zhanglistar: Use log instead of std::cerrr if need.

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -71,7 +74,16 @@ void QueryContext::resetGlobal()
 
 void QueryContext::reset()
 {
+    for (auto it = active_executors_.begin(); it != active_executors_.end(); ++it)
+    {
+        auto * executor = *it;
+        executor->cancel();
```
zhanglistar: Should we check the result status? maybe fail or something else.

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -71,7 +74,16 @@ void QueryContext::resetGlobal()
...
+        std::cerr << "Not closed LocalExecutor:\n" << executor->dumpPipeline() << std::endl;
+        delete executor;
```
zhanglistar: Useless, since `active_executors_.clear();` will call dtor.

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -200,14 +216,18 @@ void QueryContext::finalizeQuery(int64_t id)
 size_t currentThreadGroupMemoryUsage()
 {
     if (!CurrentThread::getGroup())
-        throw DB::Exception(ErrorCodes::LOGICAL_ERROR, "Thread group not found, please call initializeQuery first.");
+    {
```
zhanglistar, line 217: Add a log to indicate the exception?

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -200,14 +216,18 @@ void QueryContext::finalizeQuery(int64_t id)
...
+        return 0;
+    }
     return CurrentThread::getGroup()->memory_tracker.get();
 }
 
 double currentThreadGroupMemoryUsageRatio()
 {
     if (!CurrentThread::getGroup())
-        throw DB::Exception(ErrorCodes::LOGICAL_ERROR, "Thread group not found, please call initializeQuery first.");
+    {
```
zhanglistar, line 227: The same as above.

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -71,7 +74,16 @@ void QueryContext::resetGlobal()
 
 void QueryContext::reset()
 {
+    for (auto it = active_executors_.begin(); it != active_executors_.end(); ++it)
```
zhanglistar: Here we manipulate the `active_executors_`, lock must be added.

cpp-ch/local-engine/Common/QueryContext.cpp
lgbo-ustc: Cannot use exception here, the jvm resource may become invalid

zhanglistar: > Cannot use exception here, the jvm resource may become invalid

I mean we should log this exception, not really throw exception.

cpp-ch/local-engine/Common/QueryContext.cpp
lgbo-ustc: the result type of cancel is void

cpp-ch/local-engine/Common/QueryContext.cpp
lgbo-ustc: these are raw pointers, not shared_ptr

cpp-ch/local-engine/jni/jni_common.h
```diff
@@ -46,6 +47,106 @@ jstring charTojstring(JNIEnv * env, const char * pat);
 
 jbyteArray stringTojbyteArray(JNIEnv * env, const std::string & str);
 
+// A watcher to monitor JNIEnv status: it's active or not
+class JniEnvStatusWatcher
```
zhanglistar, line 51: Better add 
```
JniEnvStatusWatcher() = default;
    ~JniEnvStatusWatcher() = default;  
``` 
and set to private to avoid compiler autogenerate. Same as above other singleton classes.

cpp-ch/local-engine/jni/jni_common.h
```diff
@@ -46,6 +47,106 @@ jstring charTojstring(JNIEnv * env, const char * pat);
...
+{
+public:
+    static JniEnvStatusWatcher & instance();
+    void active() { is_active.store(true, std::memory_order_release); }
```
zhanglistar: Better to use `setActive` name as it does.

cpp-ch/local-engine/jni/jni_common.h
```diff
@@ -46,6 +47,106 @@ jstring charTojstring(JNIEnv * env, const char * pat);
...
+    void inactive() { is_active.store(false, std::memory_order_release); }
```
zhanglistar: Better use `setInactive` name as it does.

cpp-ch/local-engine/Common/QueryContext.cpp
```diff
@@ -16,6 +16,7 @@
  */
 #include "QueryContext.h"
 
+#include <cstdint>
```
zhanglistar: Useless？

lgbo-ustc: introduced by autocomplete

zhanglistar:
Good job and hope we can solve the early stop causing crash on CH BE.

PR #10817, opened 2025-09-29
[VL]Implement KnownSizeEstimation for buildSideRelation

zjuwangg:

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarBuildSideRelation.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarBuildSideRelation.scala
@@ -33,6 +33,7 @@ import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.utils.SparkArrowUtil
 import org.apache.spark.sql.vectorized.ColumnarBatch
 import org.apache.spark.task.TaskResources
+import org.apache.spark.util.KnownSizeEstimation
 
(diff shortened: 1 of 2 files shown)


backends-velox/src/main/scala/org/apache/spark/sql/execution/unsafe/UnsafeColumnarBuildSideRelation.scala
```diff
@@ -366,4 +367,12 @@ case class UnsafeColumnarBuildSideRelation(
     }
     iterator.toArray
   }
+
+  override def estimatedSize: Long = {
+    if (null != batches) {
```
zhztheplayer: nit: Can we just use `batches != null`? It's more consistent with our current coding style.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarBuildSideRelation.scala
```diff
@@ -236,4 +238,11 @@ case class ColumnarBuildSideRelation(
     }
     iterator.toArray
   }
+  override def estimatedSize: Long = {
+    if (null != batches) {
```
zhztheplayer: Curious when will the batches be null?

zjuwangg: It shouldn't happen. Just add here to avoid potential npe.

PR #10821, opened 2025-09-29
[GLUTEN-8851][VL] Fix the TakeOrderedAndProjectExecTransformer does not respect cudf tag

jinchengchenghh:
A follow up of https://github.com/apache/incubator-gluten/pull/10754

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/TakeOrderedAndProjectExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/TakeOrderedAndProjectExecTransformer.scala
@@ -143,7 +143,10 @@ case class TakeOrderedAndProjectExecTransformer(
       } else {
         val limitStagePlan =
           WholeStageTransformer(limitBeforeShuffle)(transformStageCounter.incrementAndGet())
-        limitStagePlan.copyTagsFrom(child)
+        val cudfTag = child.getTagValue(CudfTag.CudfTag)
+        if (cudfTag.isDefined) {
+          limitStagePlan.setTagValue(CudfTag.CudfTag, cudfTag.get)
+        }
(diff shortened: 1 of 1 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/execution/TakeOrderedAndProjectExecTransformer.scala
```diff
@@ -143,6 +143,7 @@ case class TakeOrderedAndProjectExecTransformer(
       } else {
         val limitStagePlan =
           WholeStageTransformer(limitBeforeShuffle)(transformStageCounter.incrementAndGet())
+        limitStagePlan.copyTagsFrom(child)
```
zhztheplayer: Doing this may bring some subtle shortcomings because `limitStagePlan` is not logically equal to `child` (they don't output the same data from user's view). Thus, the new code may work for cuDF but may technically break other tags.

Can we invoke the code in `CudfNodeValidationRule` here to specifically add the cuDF tags?

jinchengchenghh: This change is in function doExecuteColumnar, so we cannot change it in Plan level.

jinchengchenghh: Do I change it to only apply for Cudf tag?

zhztheplayer: > Do I change it to only apply for Cudf tag?

For sure. And it's reasonable to keep calling `copyTagsFrom` on the `WholeStageTransformer` but let's only add cuDF tag for `limitStagePlan`, if it works for you.

PR #10822, opened 2025-09-29
[VL] Delta: Fix input file expressions not getting pushed to native scan when column mapping is applicable

zhztheplayer:
Issue is because of the incorrect application order of `pushDownInputFileExprRule` and `columnMappingRule`, sometimes the input file expressions are not pushed to Velox scan, causing the input file expression columns to be returned with empty values incorrectly by Gluten.

The PR fixes the issue by adjusting the application order of `pushDownInputFileExprRule` and `columnMappingRule`.

This fixes the ignored Delta test suite `DeleteSQLNameColumnMappingSuite`.


zhztheplayer:
cc @dcoliversun @zhli1142015 @YannByron

zhztheplayer:
Thanks for reviewing @dcoliversun

Issue #9896, opened 2025-06-06
[CORE][VL] V2 columnar write


zhztheplayer:
### Description

The topic is to enable DSV2 write in Gluten, also in preparation for Iceberg support.

This will involve:

1. Have a V2 columnar write API
2. Test: Add columnar version of `InMemoryTableCatalog` (e.g., `ColumnarInMemoryTableCatalog`) that fits into V2 columnar write API, but not necessarily do columnar processing internally (could reuse `InMemoryTableCatalog`'s row-based storage after the data passes through the columnar API) 
3. Enable `DataFrameWriterV2Suite` 

### Gluten version

None

PR #10823, opened 2025-09-30
[GLUTEN-10648][VL] Support Iceberg overwrite partitions dynamic

Zouxxyy:

Support Iceberg overwrite partitions dynamic

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -551,15 +551,11 @@ object VeloxBackendSettings extends BackendSettingsApi {
 
   override def enableEnhancedFeatures(): Boolean = VeloxConfig.get.enableEnhancedFeatures()
 
-  override def supportAppendDataExec(): Boolean =
-    GlutenConfig.get.enableAppendData && enableEnhancedFeatures()
+  override def supportAppendDataExec(): Boolean = enableEnhancedFeatures()
 
-  override def supportReplaceDataExec(): Boolean =
-    GlutenConfig.get.enableReplaceData && enableEnhancedFeatures()
+  override def supportReplaceDataExec(): Boolean = enableEnhancedFeatures()
 
-  override def supportOverwriteByExpression(): Boolean =
-    GlutenConfig.get.enableOverwriteByExpression && enableEnhancedFeatures()
+  override def supportOverwriteByExpression(): Boolean = enableEnhancedFeatures()
 
-  override def supportOverwritePartitionsDynamic(): Boolean =
-    GlutenConfig.get.enableOverwritePartitionsDynamic && enableEnhancedFeatures()
+  override def supportOverwritePartitionsDynamic(): Boolean = enableEnhancedFeatures()
 }


jinchengchenghh:
It is public, username  gluten password hN2xX3uQ4m

jinchengchenghh:
This is an unstable test, I will create a PR to fix it.
```
[2025-09-30T11:51:08.252Z] - Gluten - SPARK-35650: Coalesce number of partitions by AEQ *** FAILED ***
[2025-09-30T11:51:08.252Z]   2 did not equal 1 (ClickHouseAdaptiveQueryExecSuite.scala:84)
```

jinchengchenghh:
The line number is different with source ClickHouseAdaptiveQueryExecSuite.scala:84

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -559,4 +559,7 @@ object VeloxBackendSettings extends BackendSettingsApi {
 
   override def supportOverwriteByExpression(): Boolean =
     GlutenConfig.get.enableOverwriteByExpression && enableEnhancedFeatures()
+
+  override def supportOverwritePartitionsDynamic(): Boolean =
```
jinchengchenghh: Here is the backend setting config, just return enableEnhancedFeatures(), and this change can trigger the CI again.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
```diff
@@ -161,4 +161,6 @@ trait BackendSettingsApi {
   def supportReplaceDataExec(): Boolean = false
 
   def supportOverwriteByExpression(): Boolean = false
+
+  def supportOverwritePartitionsDynamic(): Boolean = false
```
zhztheplayer, line 165: Unnecessarily related to this PR, but we do need to have the new V2 columnar write operators covered in tests individually without having to enable the Iceberg writer, as they were design to be general. Vanilla Spark uses an in-memory catalog for testing the row-based V2 write operators. We may want to introduce something similar just for testing. https://github.com/apache/incubator-gluten/issues/9896

Zouxxyy: Sounds good, v2 write is a general capability that can be used in all other lake formats. I know a bit about DSv2, and happy to help if needed.

zhztheplayer: Thanks! Feel free to open issues and PRs.

zhztheplayer: > Spark uses an in-memory catalog for testing the row-based V2 write operators. We may want to introduce something similar just for testing.

@Zouxxyy Just recalled that other contributors might already work on something similar, let me confirm first to avoid duplicated work. :) I don't have their GitHub ID or Email at this moment but I will try to get them into the public discussion.

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -864,6 +867,12 @@ object GlutenConfig {
       .booleanConf
       .createWithDefault(true)
 
+  val COLUMNAR_OVERWRIET_PARTITIONS_DYNAMIC_ENABLED =
+    buildConf("spark.gluten.sql.columnar.overwriteOverwritePartitionsDynamic")
```
zhztheplayer, line 871: Hi @Zouxxyy,

Should this be `spark.gluten.sql.columnar.overwritePartitionsDynamic`?

Zouxxyy: Sorry, my mistake, I might have copied the wrong content.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
zhztheplayer: I've confirmed, they are not working on the test topic. So feel free to take if wanted. We may have public discussions about the further matters later on.

Zouxxyy: Thanks, I'd like to, testing is the foundation, which will make the integration more directional and reliable.

PR #10824, opened 2025-09-30
[GLUTEN-10833][VL] Reduce memcpy in ColumnToRow

### LINKED ISSUE #10833: [VL] Improve columnar to row memory allocation
### Description

There's no need to do re-allocate as we always fill `0` for the memory allocated
https://github.com/apache/incubator-gluten/blob/main/cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc#L60-L67

### Gluten version

None

zhouyuan:

We always fill the buffer with `0` so it's no necessary to keep the old content when doing allocation in C2R

Fixes #10833

## How was this patch tested?
pass GHA

### DIFF
--- a/cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
+++ b/cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
@@ -58,8 +58,11 @@ void VeloxColumnarToRowConverter::refreshStates(facebook::velox::RowVectorPtr ro
   }
 
   if (nullptr == veloxBuffers_ || veloxBuffers_->capacity() < totalMemorySize) {
-    veloxBuffers_ = velox::AlignedBuffer::allocate<uint8_t>(totalMemorySize, veloxPool_.get(), 0);
+    veloxBuffers_ = velox::AlignedBuffer::allocate<uint8_t>(totalMemorySize, veloxPool_.get());
   }
+
+  bufferAddress_ = veloxBuffers_->asMutable<uint8_t>();
+  memset(bufferAddress_, 0, sizeof(int8_t) * totalMemorySize);
 }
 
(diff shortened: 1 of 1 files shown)


cpp/velox/operators/serializer/VeloxColumnarToRowConverter.cc
```diff
@@ -60,7 +60,8 @@ void VeloxColumnarToRowConverter::refreshStates(facebook::velox::RowVectorPtr ro
   if (nullptr == veloxBuffers_) {
     veloxBuffers_ = velox::AlignedBuffer::allocate<uint8_t>(totalMemorySize, veloxPool_.get());
   } else if (veloxBuffers_->capacity() < totalMemorySize) {
-    velox::AlignedBuffer::reallocate<uint8_t>(&veloxBuffers_, totalMemorySize);
+    veloxBuffers_.reset();
```
jinchengchenghh: Do we need reset?

zhouyuan: this will ensure the buffer allocated previously released

JkSelf: @zhouyuan The resource will be automatically released when the smart pointer is assigned a new value, as the assignment operator handles the reference counting internally. We can simplify the code to: 

```
if (veloxBuffers_ == nullptr || veloxBuffers_->capacity() < totalMemorySize) {
    veloxBuffers_ = velox::AlignedBuffer::allocate<uint8_t>(totalMemorySize, veloxPool_.get());
}
```

zhouyuan: @JkSelf @jinchengchenghh done

PR #10801, opened 2025-09-25
[VL] Delta: Native write support for Delta 3.3.1 / Spark 3.5

zhztheplayer:
The code is out of PoC from https://github.com/apache/incubator-gluten/pull/10216.

The PR adds native Delta write support by offloading native Parquet writer to Velox.

The PR only adds support for Spark 3.5 / Delta 3.3.

View https://github.com/apache/incubator-gluten/issues/10215 for a TODO item list.

PR #10835, opened 2025-10-02
[CORE] Refactor configuration API to allow explicit `getAllEntries` calls on specific configuration registry objects

zhztheplayer:
This is to propose a new API to support the calls like:

```scala
val allGlutenCoreEntries = GlutenCoreConfig.allEntries
val allGlutenEntries = GlutenConfig.allEntries
val allVeloxEntries = VeloxConfig.allEntries
```

The story behind this is, I was trying to add a new `VeloxDeltaConfig` object in PR https://github.com/apache/incubator-gluten/pull/10801, but `ConfigEntry.getAllEntries` returns uncertain results when switching on / off `-P delta` Maven profile, so CI test that does golden checks for the documented configuration options fails. So we should add a reliable API that returns whatever we need no matter how we trigger the test using Maven.

For example, after the PR change, `VeloxConfig.allEntries` will return exactly the config options that are registered in `VeloxConfig` only.


zhztheplayer:
@zjuwangg @yikf Would you help review this? Thanks.

zhouyuan (approved):
👍

zhztheplayer:
Thank you @zhouyuan

PR #10838, opened 2025-10-03
[CORE] Refactor BasicScanExecTransformer to make it accept Seq[Seq[InputPartition]] as InputPartitions

Zouxxyy:


Zouxxyy:

zhztheplayer:
Thanks.

Should we also refactor GlutenWholeStageColumnarRDD? Does soft-affinity still work correctly after the change?

Zouxxyy:
> Should we also refactor GlutenWholeStageColumnarRDD? 

For GlutenWholeStageColumnarRDD, I think these modifications below are sufficient. Do you think it needs to be modified in this PR as well? (Besieds. NativeFileScanColumnarRDD doesn't seem to be used. Should I delete it?)

```scala
case class FirstZippedPartitionsPartition(
    index: Int,
    inputPartition: Partition,
    inputColumnarRDDPartitions: Seq[Partition] = Seq.empty)
  extends Partition

class GlutenWholeStageColumnarRDD(
    @transient sc: SparkContext,
    @transient private val inputPartitions: Seq[Partition],
```

> Does soft-affinity still work correctly after the change?

do you mean getPreferredLocations? I haven't modified its logic, so it still applies

e.g. in GlutenWholeStageColumnarRDD, It will do the correct cast.
```
  override def getPreferredLocations(split: Partition): Seq[String] = {
    castNativePartition(split)._1.preferredLocations()
  }
```

Or just like `FileScanRDD` in apache spark
```scala
  override protected def getPreferredLocations(split: RDDPartition): Seq[String] = {
    split.asInstanceOf[FilePartition].preferredLocations()
  }
```

zhztheplayer:
> For GlutenWholeStageColumnarRDD, I think these modifications below are sufficient. Do you think it needs to be modified in this PR as well? 

Yes and I saw you're working on this. Thanks.

> Besieds. NativeFileScanColumnarRDD doesn't seem to be used. Should I delete it?

For sure. Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -298,27 +298,13 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
...
-      leafTransformer => {
-        if (isKeyGroupPartition) {
-          leafTransformer.asInstanceOf[BatchScanExecTransformerBase].getPartitionsWithIndex
-        } else {
-          Seq(leafTransformer.getPartitions)
-        }
-      })
-
-    val allSplitInfos = getSplitInfosFromPartitions(isKeyGroupPartition, leafTransformers)
+    val allSplitInfos = leafTransformers.map(_.getSplitInfos).transpose
```
zhztheplayer, line 327: Do we need to preserve a similar comment like [this](https://github.com/apache/incubator-gluten/blob/84f040f5fd57e3940bfba17713dd55ba47287224/gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala#L373-L408)? Which will help user understand the `.transpose` usage here.

Zouxxyy:
@zhztheplayer I rebased and resolved the conflict, can you have a look, thanks

PR #10846, opened 2025-10-07
[CORE] Optimize the verboseString of ColumnarShuffleExchangeExec

Zouxxyy:

```
before:
ColumnarExchange SinglePartition, ENSURE_REQUIREMENTS, [plan_id=744], [shuffle_writer_type=hash], [OUTPUT] List(sum:DoubleType, count:LongType), [OUTPUT] List(sum:DoubleType, count:LongType)


after:
ColumnarExchange SinglePartition, ENSURE_REQUIREMENTS, [plan_id=710], [shuffle_writer_type=hash], [output=[sum#133: double, count#134: bigint]]
```

The reason why the OUTPUT appeared twice previously is that verboseString internally calls simpleString. Moreover, simpleString generally does not include output information, so this PR simply removes it.


marin-ma:
Can we preserve the data type?

marin-ma (approved):
LGTM. Thanks!

PR #10850, opened 2025-10-08
[GLUTEN-10774][VL][FOLLOWUP] Strip quotes for osName and osVersion

liujiayi771:
Follow-up PR for #10774

The content in `/etc/os-release` is probably in the format of key="value". Previously, when extracting with regular expressions, the double quotes were removed. Currently, using `Properties.load` does not remove the double quotes, causing some `startsWith` logic to return false.

## How was this patch tested?
N/A

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoaderUtils.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoaderUtils.scala
@@ -35,6 +35,14 @@ object SharedLibraryLoaderUtils {
     osName.startsWith("Mac OS X") || osName.startsWith("macOS")
   }
 
+  private def stripQuotes(s: String): String = {
+    if (s == null) {
+      null
+    } else {
+      s.stripPrefix("\"").stripSuffix("\"")
+    }
+  }
+
   def load(conf: SparkConf, jni: JniLibLoader): Unit = {
     val shouldLoad = conf.get(GLUTEN_LOAD_LIB_FROM_JAR)
     if (!shouldLoad) {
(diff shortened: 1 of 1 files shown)


liujiayi771:
@yaooqinn @zhztheplayer Could you help to review? The issue is that our OS `/etc/os-release` file contains values wrapped in double quotes (e.g., VERSION_ID="3 (OpenAnolis Edition)"). Currently, this line in Gluten's code returns false.

https://github.com/apache/incubator-gluten/blob/f057958545480e3aa9be80c867a4d4455bbed33f/backends-velox/src/main/scala/org/apache/gluten/spi/SharedLibraryLoaderCentos8.scala#L28

```
NAME="Alibaba Cloud Linux"
VERSION="3 (OpenAnolis Edition)"
```

zhztheplayer:
Similar issue is repeatable from my end with Ubuntu 24.04 (although we don't have a loader for it). CC @yaooqinn for more insights.

If we are making the change maybe we also want to strengthen the `accepts` implementation code in the loader implementations by turning `osName.contains("Ubuntu")` to `osName.startsWith("Ubuntu")` / `osName.equals("Ubuntu")` or so.

liujiayi771:
I believe that checking if osName `contains` some string is sufficient for matching. However, for osVersion, we need more precise matching. Currently, `startsWith` seems appropriate since there might be minor version numbers (e.g., CentOS 8.x or ), or cases like alinux that may include additional parenthetical information (e.g., 3 (OpenAnolis Edition)).

We don't need to use equals for exact matching. In many cases, the system can still run normally on these operating systems even without an exact loader match.

yaooqinn:
It might be better to handle the returned values read from the system file on the caller side. Quotes might not be the only special character to deal with.

liujiayi771:
@yaooqinn We can first remove the double quotes as a quick fix to maintain consistency with previous results. I think the vast majority of operating systems will only have either `key=value` or `key="value"` formats.

https://github.com/apache/incubator-gluten/blob/50dd117dadb9fe3e9bba5c5002168e1294b2e2bb/backends-velox/src/main/scala/org/apache/gluten/utils/SharedLibraryLoader.scala#L68

PR #8967, opened 2025-03-11
[GLUTEN-8966][VL] Propagate HashAggregate's ignoreNullKeys when possible

WangGuangxin:
## What changes were proposed in this pull request?

For aggregation followed by an inner join on the grouping keys, it's safe to mark the `ignoreNullKeys` flag to true to avoid unnecessary aggregation on null keys.

(Fixes: \#8966)

## How was this patch tested?

UT and manually

PR #10811, opened 2025-09-26
[GLUTEN-10810][VL] Change the default value of propagateIgnoreNullKeys to false

beliefer:
## What changes are proposed in this pull request?

This PR proposes to change the default value of propagateIgnoreNullKeys to false
Fixes #10810

## How was this patch tested?

GA tests.

PR #10852, opened 2025-10-09
Revert "[GLUTEN-8966][VL] Propagate HashAggregate's ignoreNullKeys when possible"

zml1206:

Reverts #8967.

Spark optimize rule `InferFiltersFromConstraints` will extract filter `IsNotNull`, rule `PushDownPredicates` will push down `IsNotNull` through aggregate. So the data processed by aggregate is already filtered, adding `ignoreNullKeys` will incur additional computational cost for velox.


zml1206:
@zhztheplayer @WangGuangxin cc @jackylee-ch @jinchengchenghh

jackylee-ch:
Thanks for clarifying. I just tested the [test case](https://github.com/apache/incubator-gluten/pull/8967/files#diff-742ce9027f0d774d324125a130d9993bad09d4ea5eb6d35d791b39bb57138c77R1200) added in #8967 on vanilla Spark — `isNotNull` is added after the file scan. So we shouldn’t worry about `ignoreNullKeys` right now.

jackylee-ch:
@WangGuangxin Are there any other cases that shows we require the `ignoreNullKeys`? If not, we might revert this PR.

zhztheplayer:
@beliefer @zml1206 I noticed this PR and https://github.com/apache/incubator-gluten/pull/10811 are for similar purposes. Maybe it's a coincidence?

zml1206:
> @beliefer @zml1206 I noticed this PR and #10811 are for similar purposes. Maybe it's a coincidence?


Yes, the purpose is the same, but I think it is more appropriate to remove it. Even if there are other cases, it would be more appropriate to optimize it in the spark optimizer.

beliefer:
@zhztheplayer @zml1206 Whether reverting or modifying, I think it's okay.

WangGuangxin:
@beliefer @zml1206 Get. I'm ok to revert it

zml1206:
Thanks for review @jackylee-ch @zhztheplayer @WangGuangxin @beliefer , Merge into main.

PR #10855, opened 2025-10-09
[GLUTEN-10574] Include copyright for Hudi and Paimon in NOTICE-binary

philo-he:
Required for release. I will backport this patch to 1.5.0.

Reference:
https://github.com/apache/hudi/blob/ab23ac6218d5ac3d361e151eed5140be16eb99e5/NOTICE#L1-L2.
https://github.com/apache/paimon/blob/60454a1b190513f4be42554f0f6acdf16bb9e611/NOTICE#L1-L2.

## How was this patch tested?
N/A

### DIFF
--- a/NOTICE-binary
+++ b/NOTICE-binary
@@ -53,6 +53,16 @@ Copyright (2021) The Delta Lake Project Authors.
 
 ---------------------------------------------------------
 
+Apache Hudi
+Copyright 2019-2020 The Apache Software Foundation.
+
+---------------------------------------------------------
+
+Apache Paimon
+Copyright 2023-2025 The Apache Software Foundation.
+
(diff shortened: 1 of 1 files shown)


philo-he:
@weiting-chen, could you take a look?

NOTICE-binary
```diff
@@ -53,6 +53,16 @@ Copyright (2021) The Delta Lake Project Authors.
 
 ---------------------------------------------------------
 
+Apache Hudi
+Copyright 2019-2020 The Apache Software Foundation.
```
weiting-chen, line 57: It looks like Apache Hudi hasn't updated its copyright years for a while.
It's ok in this case as we should reproduce the upstream NOTICE content.

PR #10858, opened 2025-10-09
[GLUTEN-10574] Correct Gluten versions in documents and jupyter notebooks

philo-he:
Several files contain specific Gluten versions. For documentation, we can use a generic statement to avoid referencing a Gluten JAR with a specific version number. Additionally, some unnecessary example files can be removed.

This patch will be backported to 1.5.0, requiring a few lines to be updated.

## How was this patch tested?
N/A


philo-he:
@zhztheplayer, could you take a look?

PR #10865, opened 2025-10-11
[VL] Check VELOX_HOME variable in build_arrow to allow custom Velox path

liujiayi771:
The build_arrow function should check the value of `VELOX_HOME` to determine whether it needs to execute `get_velox && setup_dependencies`.

Currently, if the user provides a custom `VELOX_HOME` path (pointing to an external Velox installation), the script still unconditionally runs `get_velox && setup_dependencies`, which is redundant and inefficient.

The default value of `VELOX_HOME` is `$GLUTEN_DIR/ep/build-velox/build/velox_ep`

## How was this patch tested?
N/A

### DIFF
--- a/dev/builddeps-veloxbe.sh
+++ b/dev/builddeps-veloxbe.sh
@@ -206,7 +206,7 @@ fi
 concat_velox_param
 
 function build_arrow {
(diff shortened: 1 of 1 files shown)


liujiayi771:
@PHILO-HE, @zhouyuan Could you help to review?

PR #10870, opened 2025-10-12
[VL] Update user-facing metric name for `wallNanos` in batch scan

Zouxxyy:

Velox scan will do scan and filter, update batch scan wallNanos to time of batch scan and filter just like what HiveScan and FileSourceScan do


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
```diff
@@ -95,7 +95,9 @@ class VeloxMetricsApi extends MetricsApi with Logging {
       "numOutputRows" -> SQLMetrics.createMetric(sparkContext, "number of output rows"),
       "outputVectors" -> SQLMetrics.createMetric(sparkContext, "number of output vectors"),
       "outputBytes" -> SQLMetrics.createSizeMetric(sparkContext, "number of output bytes"),
-      "wallNanos" -> SQLMetrics.createNanoTimingMetric(sparkContext, "time of batch scan"),
+      "wallNanos" -> SQLMetrics.createNanoTimingMetric(
+        sparkContext,
+        "time of batch scan and filter"),
```
zml1206: How about unifying them with other scans?
"time of batch scan" -> "time of scan and filter"
"scan time" -> "time of scan"

PR #10872, opened 2025-10-13
[GLUTEN-10574] Add missing license headers

philo-he:

With Apache Rat, we found some missing license headers, which should be fixed before the release.
This patch will be backported to branch-1.5.

FYI.
`java -jar apache-rat-0.16.1.jar -d /path/to/gluten/code -E excludes.txt`, where excludes.txt contains

```
.*\.md
.*\.patch
.*\.txt
.*\.log
.*\.sql
.*\.out
.*\.proto
.*\.csv
.*\.parquet

```

## How was this patch tested?
N/A

### DIFF
--- a/backends-clickhouse/pom.xml
+++ b/backends-clickhouse/pom.xml
@@ -15,7 +15,6 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-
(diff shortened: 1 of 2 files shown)


philo-he:
@weiting-chen, could you take a look?

wForget:
Should we add `apache-rat-plugin` maven plugin for gltuen?

philo-he:
> Should we add `apache-rat-plugin` maven plugin for gltuen?

@wForget, thanks for your comment. We already have a CI job to check the license. It seems not all types of files are checked. We may just extend this job with Apache Rat check command used for better coverage. Maybe this centralized checking is better than introducing the maven plugin since the latter only covers maven project code.

Issue #10874, opened 2025-10-13
[INFRA] Maturity RE50: In documentation, add a section for the release process


zhztheplayer:
https://community.apache.org/apache-way/apache-project-maturity-model.html

We'd add a section demonstrating the release process of Apache Gluten.

PR #10883, opened 2025-10-14
[GLUTEN-10874][INFRA] Maturity RE50: In documentation, add a section for the release process

zhztheplayer:
#10874

The PR adds a documentation to demonstrate how to generate a complete release for a specific Gluten version locally.

The steps should be repeatable as long as the environment  meets the prerequisites.

### DIFF
--- a/docs/developers/HowToRelease.md
+++ b/docs/developers/HowToRelease.md
@@ -2,7 +2,7 @@
 layout: page
 title: How to Release
 nav_order: 17
-parent: Getting-Started
+parent: Developer Overview
 ---
 
 # Create Release Source and Binaries


zhztheplayer:
@PHILO-HE @zhouyuan

dev/release/package-release.sh
```diff
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Note: Manually create $GLUTEN_HOME/release/ and place the release JARs inside.
+# Note: The script looks up the release JARs inside $GLUTEN_HOME/package/target/.
```
philo-he: I have created one GHA job to build Gluten, and then upload the JARs. The job is auto triggered when tag like v1.5.0-rc0 is created in the repo.
https://github.com/apache/incubator-gluten/actions/runs/18470382803/job/52622515460

So $GLUTEN_HOME/release/ is used to manually place the JARs. The JARs can be obtained from either the above GHA job (recommended) or manual build.

docs/developers/HowToRelease.md
philo-he, line 1: Can we move this doc to `dev/release/` together with the release scripts? I think this doc is primarily for release managers' reference. So it seems not necessary to expose it to broader readers by putting it here?

PHILO-HE (commented):
@zhztheplayer, thanks for your work. Please check if my comments make sense.

weiting-chen:
Please also copy the below file into incubator-gluten's doc, I noted we only put this md in the incubator-gluten-site repository and missed to add it in the main documentation.
The release process not only include how to compile, but also include releasing to Apache repository, voting, and more.
https://github.com/apache/incubator-gluten-site/blob/main/_docs/v1.3.0/developers/HowToRelease.md

dev/release/package-release.sh
zhztheplayer: Hi @PHILO-HE,

> I have created one GHA job to build Gluten, and then upload the JARs.

I believe there are two separate things we need:

1. Guide for user to create a release by themself
2. Guide for release manager to publish a release

For users, they don't usually have the immediate permission to run the GitHub Actions jobs. So we need provide a way for them to create the release artifacts from the source code.

I can leave the `$GLUTEN_HOME/release/` solution unchanged but add an extra copy-paste step in the documentation. Does that make sense?

zhztheplayer:
> The release process not only include how to compile, but also include releasing to Apache repository, voting, and more.

@weiting-chen, I understand this. This doc is initialized for new users to reproduce the artifact sets by themselves. I'll see how to organize the user and release manager guide later on. 

For release managers, they are usually committers, so could conveniently invoke the GitHub Actions jobs to create the assets. For users, they have to build from source. It's the reason that we may finally need separate guides here.

zhztheplayer:
> I noted we only put this md in the incubator-gluten-site repository and missed to add it in the main documentation.

We need a synchronizer between the two repos. I will look into it.

dev/release/package-release.sh
philo-he: @zhztheplayer, thank you for the reply. I am not sure if users really need to create an official-like release by themselves. In official release, we build Gluten for all supported Spark versions with all feature modules enabled, even though generally some of them are not enabled together by users (e.g., Celeborn and Uniffle). However, most Gluten users only focus on one specific Spark version, and some of the the feature modules, which means they just need a clear build guide, not a release guide. Users like cloud service providers may require building releases for multi-Spark-version support. But they likely maintain their own processes and scripts to customize releases for forked repositories, which I assume differs significantly from our official release process.

If my understanding is correct, maybe we can focus only on the official release process, primarily for release managers' reference?

zhztheplayer: Thanks for the insights. Now I think I am convinced we should go with a release manager guide first. 

RE50 itself was seen a bit ambiguous and you are right it might better to read it as "creating new releases" rather than "reproducing old releases".

docs/developers/HowToRelease.md
zhztheplayer: I think we'd better publish it to the website. This could let more people know how we handle the releases, even if they are not developers or release managers.

dev/release/package-release.sh
philo-he, line 1: @zhztheplayer, could you please include the following changes?  I have applied these changes when publishing the recent 1.5.0 RC1. It's for meeting the release requirement (The approved release candidate will be directly converted to a formal release. So "rc*" shouldn't be used in naming packaged folders).

```diff
diff --git a/dev/release/package-release.sh b/dev/release/package-release.sh
index ec18101ae..d60a14e78 100755
--- a/dev/release/package-release.sh
+++ b/dev/release/package-release.sh
@@ -53,11 +53,14 @@ for v in $SPARK_VERSIONS; do
 done

 SRC_ZIP="${TAG}.zip"
-SRC_DIR="incubator-gluten-${TAG_VERSION}"
+SRC_DIR="incubator-gluten-${RELEASE_VERSION}"

 echo "Packaging source code..."
 wget https://github.com/apache/incubator-gluten/archive/refs/tags/${SRC_ZIP}
 unzip -q ${SRC_ZIP}
+
+# Rename folder to remove "rc*" for formal release.
+mv incubator-gluten-${TAG_VERSION} ${SRC_DIR}
 tar -czf apache-gluten-${RELEASE_VERSION}-incubating-src.tar.gz ${SRC_DIR}
 rm -r ${SRC_ZIP} ${SRC_DIR}
```

docs/developers/HowToRelease.md
philo-he: Sounds good to me. Thanks.

docs/developers/HowToRelease.md
```diff
@@ -0,0 +1,76 @@
+---
+layout: page
+title: How to Release
+nav_order: 17
+parent: Getting-Started
```
philo-he: Please update accordingly. Thanks.

zhztheplayer:
Here is the update of the PR:

1. I've left `package-release.sh` nearly intact and changed the `HowToRelease.md` to do an extra copy from `package/target` to `release`
2. I've moved the release publication guide from `incubator-gluten-site`'s 1.3 branch to this documentation

@PHILO-HE and feel free to help update `HowToRelease.md` about using the CI job to obtain release binaries in later PRs if that works for you. I think I can help either. This PR should first close RE50 then.

philo-he:
> Here is the update of the PR:
> 
> 1. I've left `package-release.sh` nearly intact and changed the `HowToRelease.md` to do an extra copy from `package/target` to `release`
> 2. I've moved the release publication guide from `incubator-gluten-site`'s 1.3 branch to this documentation
> 
> @PHILO-HE and feel free to help update `HowToRelease.md` about using the CI job to obtain release binaries in later PRs if that works for you. I think I can help either. This PR should first close RE50 then.

@zhztheplayer, thanks for your work. I will update the doc in separate PRs for better clarity according to the recent release process.

docs/developers/HowToRelease.md
philo-he: @zhztheplayer, could you change the `nav_order` and `parent`? I assume they are applicable for the previous doc location.

zhztheplayer: It seems there are quite some issues with the current nav orders. 

I am working on a fix which can be merged prior to this.

zhztheplayer: > It seems there are quite some issues with the current nav orders.

Unfortunately, after a quick check, I think I need more time for fixing this.

I've update the PR to make the new page correctly show up, so we can merge this first.

zhztheplayer: > nav_order: 17

`nav_order: 17` was intentional to make it be the last so safe to display. I will find the doc a better order in the next PR.

Issue #10876, opened 2025-10-13
[INFRA] Maturity QU30: In documentation, add a section for security and a link to the ASF security documentation


zhztheplayer:
https://community.apache.org/apache-way/apache-project-maturity-model.html

We'd add a simple section about security in our documentation. This could include a link to https://www.apache.org/security/.

zhztheplayer:
Section added in https://github.com/apache/incubator-gluten-site/pull/53.

PR #10885, opened 2025-10-14
[GLUTEN-10876][INFRA] Maturity QU30: In documentation, add a section for security and a link to the ASF security documentation

zhztheplayer:
#10876

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ The [Qualification Tool](./tools/qualification-tool/README.md) is a utility to a
 
 ## 9. Security
 
-Gluten aims to provide secure software. If you discover a vulnerability, please report it promptly to [the project's PPMC](private@gluten.apache.org) or [the ASF security team](https://www.apache.org/security/). We appreciate your effort to help keep the project secure.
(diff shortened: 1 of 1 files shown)


zhztheplayer:
@zhouyuan @weiting-chen @yaooqinn

yaooqinn:
Is https://github.com/apache/incubator-gluten?tab=security-ov-file OK?

zhztheplayer:
> Is https://github.com/apache/incubator-gluten?tab=security-ov-file OK?

Ahh, thanks. Looks sufficient? 

I had also updated https://gluten.apache.org/#6-security so the section in GitHub readme is also not a must I guess.

yaooqinn:
According to the diffs made in this PR, the auto-detected [security](https://github.com/apache/incubator-gluten?tab=security-ov-file) is comparatively sufficient.

W.r.t long-term CVE maintenance, the best practice to satisfy QU30 requirements is to create a SECURITY.md in the root or docs folder. It helps clarify the security features involved in authz/authn, network encryption, etc., in advance. For those reported but triaged to non-CVE ones, it's also a good place to update further document-like resolutions.

zhztheplayer:
Thanks @yaooqinn.

I think we can close QU30 now. We may need to discuss the demand of having the `SECURITY.md` in the code base later on. The dependency set of Gluten is large so I guess we will need something like that anyway.

yaooqinn:
Thank you @zhztheplayer

Issue #8851, opened 2025-02-28
[Core] Add GPU support in Gluten


FelixYBW:
### Description

GPU shows performance boost in some operators. 
Velox community is working with libcudf team to add GPU support to Velox. We need to add GPU support to Gluten

zhouyuan:
Velox PR branch used for test. CI passed.

zhouyuan:
`Orderby` is supported now

zhouyuan:
Q95 passed, plan validation done

jinchengchenghh:
Q95 passed in single thread, result mismatch for multiple threads because memory issue, spark-rapids uses Semaphore to control if this task can execute now, the config `spark.rapids.sql.concurrentGpuTasks` default value is 2.
So I assume we could only access gpu memory by one thread, but GPU memory RMM supports concurrently access the memory,  the reason why memory corrupted is not clear.
```
# Example: Parallel data processing
stream1 = stream_pool.get_stream()
stream2 = stream_pool.get_stream()

# Process data asynchronously
df1 = cudf.read_csv("data1.csv", stream=stream1)
df2 = cudf.read_csv("data2.csv", stream=stream2)
```
```
The RAPIDS Accelerator can further limit the number of tasks that are actively sharing the GPU. It does this using a semaphore. When metrics or documentation refers to the GPU semaphore it is referring to this. This restriction is useful for avoiding GPU out of memory errors while still allowing full concurrency for the portions of the job that are not executing on the GPU. Care is taken to try and avoid doing I/O or other CPU operations while the GPU semaphore is held. But in the case of a join two batches are required for processing, and it is not always possible to avoid this case.
```
https://docs.nvidia.com/spark-rapids/user-guide/23.10/tuning-guide.html#number-of-concurrent-tasks-per-gpu

zhouyuan:
TPCH SF1 passed

zhouyuan:
GPU document updated 
patches for rapids-ai/velox merged
will upstream to meta/velox

FelixYBW:
The plan is to offload computation intensive stage to GPU cluster. Checklist:

- [ ] Start hybird cluster (some node is CPU only, some node has GPU support, may have different executor.cores for CPU nodes and GPU node. 1 executor per GPU node. 1 or more executor per CPU node)
- [ ] set stage's profile (task.cpus=executor.cores, executor.gpu=1)
- [ ] schedule the GPU stage to GPU node

philo-he:
[...]

PR #10889, opened 2025-10-14
[GLUTEN-8851][VL] cuDF: Validate the plan before execution

jinchengchenghh:
Initializing task in Task constructor has been merged, so we can get the operator information from task, if validate successfully, set the whole stage transformer cudf tag to true to enable session config kCudfEnabled to true, and set the children operator CUDF tag to true to update the name in Spark UI with "Cudf".

Related issue: #8851

### DIFF
--- a/cpp/velox/cudf/CudfPlanValidator.cc
+++ b/cpp/velox/cudf/CudfPlanValidator.cc
@@ -34,7 +34,7 @@ namespace gluten {
 namespace {
 
 bool isCudfOperator(const exec::Operator* op) {
-  return isAnyOf<velox::cudf_velox::NvtxHelper>(op);
+  return dynamic_cast<const velox::cudf_velox::NvtxHelper*>(op) != nullptr;
 }
 
 }


backends-velox/src/main/scala/org/apache/gluten/extension/CudfNodeValidationRule.scala
```diff
@@ -37,7 +38,22 @@ case class CudfNodeValidationRule(glutenConf: GlutenConfig) extends Rule[SparkPl
...
+                case _: LeafTransformSupport =>
+                case t: TransformSupport =>
+                  t.setTagValue(CudfTag.CudfTag, true)
+                case _ =>
+              }
+              transformer.setTagValue(CudfTag.CudfTag, true)
+            }
+          } else {
+            transformer.setTagValue(CudfTag.CudfTag, !hasLeaf)
+          }
```
zhztheplayer, line 56: A dumb question, after this PR, what happens if user sets `spark.gluten.sql.columnar.backend.velox.cudf.enableValidation=false` but the plan is actually invalid for cuDF?

jinchengchenghh: Will not check if the pan is fully offloaded to GPU, may introduce extra cudf to velox format conversion. This is not a suggested behavior.

PR #10893, opened 2025-10-15
[CORE] Rename doc `release.md` to `releases.md` to avoid confusing

zhztheplayer:
`release` is at the same time a verb so it might lead user to think of it a releasing guide.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: Gluten Release
+title: Gluten Releases
 nav_order: 11
 ---
 


yaooqinn:
With ASF's hat on, this file itself contains inappropriate content for the term `release`.

We shall only use https://gluten.apache.org/downloads/

PHILO-HE (approved):
Make sense.

philo-he:
@yaooqinn, did you mean we should remove this doc? I'm not familiar with the Apache rule.

yaooqinn:
Yes, we shall remove it. @PHILO-HE.

zhztheplayer:
Making sense to me. Let's open another PR to remove it.

What I am thinking about is the nightly releases - is there a good way to let user know about them?

yaooqinn:
We can also add nightly previews in https://gluten.apache.org/downloads/

yaooqinn:
Or README.md

zhztheplayer:
Okay. If having them in https://gluten.apache.org/downloads/ is acceptable, I'd add a section there.

Issue #10915, opened 2025-10-20
[VL] `dynamicoffheapsizing` feature broken


zhouyuan:
### Backend

VL (Velox)

### Bug description

With dynamic offheap sizing enabled, Gluten is still looking for offheap size key.  It seems a regression on recent code changes. 
```
25/10/20 07:56:00 WARN MetricsSystem: Stopping a MetricsSystem that is not running
25/10/20 07:56:00 ERROR Main: Failed to initialize Spark session.
java.util.NoSuchElementException: spark.memory.offHeap.size
	at org.apache.spark.SparkConf.$anonfun$get$1(SparkConf.scala:245)
	at scala.Option.getOrElse(Option.scala:189)
	at org.apache.spark.SparkConf.get(SparkConf.scala:245)
	at org.apache.spark.SparkConf.$anonfun$getSizeAsBytes$1(SparkConf.scala:309)
	at scala.runtime.java8.JFunction0$mcJ$sp.apply(JFunction0$mcJ$sp.java:23)
	at org.apache.spark.SparkConf.catchIllegalValue(SparkConf.scala:482)
	at org.apache.spark.SparkConf.getSizeAsBytes(SparkConf.scala:309)
	at org.apache.gluten.backendsapi.velox.VeloxListenerApi.onDriverStart(VeloxListenerApi.scala:97)
	at org.apache.gluten.backendsapi.SubstraitBackend.onDriverStart(SubstraitBackend.scala:56)
	at org.apache.gluten.backendsapi.SubstraitBackend.onDriverStart$(SubstraitBackend.scala:42)
	at org.apache.gluten.backendsapi.velox.VeloxBackend.onDriverStart(VeloxBackend.scala:56)
	at org.apache.gluten.GlutenDriverPlugin.$anonfun$init$2(GlutenPlugin.scala:67)
	at org.apache.gluten.GlutenDriverPlugin.$anonfun$init$2$adapted(GlutenPlugin.scala:67)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at org.apache.gluten.GlutenDriverPlugin.init(GlutenPlugin.scala:67)
	at org.apache.spark.internal.plugin.DriverPluginContainer.$anonfun$driverPlugins$1(PluginContainer.scala:53)
[...]

PR #10916, opened 2025-10-20
[GLUTEN-10915][VL] Fix dynamic offheap sizing feature by setting default offheap size

zhouyuan:

This patch set the default offheap size(0) for dynamic offheap sizing feature. As Gluten will look for these settings internally

## How was this patch tested?

new UT + pass GHA

Fixes: #10915

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
@@ -118,8 +118,20 @@ private object GlutenDriverPlugin extends Logging {
             // 1GB default
             1024 * 1024 * 1024
           }
+
+        if (conf.contains(GlutenCoreConfig.SPARK_OFFHEAP_ENABLED_KEY)) {
+          logWarning(
+            s"Dynamic off-heap sizing is enabled. Ignoring user-defined " +
+              s"'${GlutenCoreConfig.SPARK_OFFHEAP_SIZE_KEY}' setting.")
+        }
+        if (conf.contains(GlutenCoreConfig.SPARK_OFFHEAP_SIZE_KEY)) {
+          logWarning(
+            s"Dynamic off-heap sizing is enabled. Ignoring user-defined " +
+              s"'${GlutenCoreConfig.SPARK_OFFHEAP_SIZE_KEY}' setting.")
+        }
         conf.set(GlutenCoreConfig.SPARK_OFFHEAP_SIZE_KEY, "0")
         conf.set(GlutenCoreConfig.SPARK_OFFHEAP_ENABLED_KEY, "false")
+
         ((onHeapSize - (300 * 1024 * 1024)) *
           conf.getDouble(GlutenCoreConfig.DYNAMIC_OFFHEAP_SIZING_MEMORY_FRACTION.key, 0.6d)).toLong
       } else {
--- a/docs/developers/VeloxDynamicSizingOffheap.md
+++ b/docs/developers/VeloxDynamicSizingOffheap.md
@@ -8,7 +8,7 @@ parent: Developer Overview
 ## Dynamic Off-heap Sizing
 Gluten requires setting both on-heap and off-heap memory sizes, which initializes different memory layouts. Improper configuration of these settings can lead to lower performance. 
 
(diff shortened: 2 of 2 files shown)


backends-velox/src/test/scala/org/apache/gluten/execution/DynamicOffHeapSizingSuite.scala
```diff
@@ -70,4 +70,38 @@ class DynamicOffHeapSizingSuite extends VeloxWholeStageTransformerSuite {
       runAndCompare(query)
     }
   }
+
+  test("Dynamic off-heap sizing with setting offheap") {
+    withSQLConf(
+      GlutenCoreConfig.SPARK_OFFHEAP_ENABLED_KEY -> "false") {
+      if (DynamicOffHeapSizingMemoryTarget.isJava9OrLater()) {
```
wangyum:
```
Error:  Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.27.2:check (spotless-check) on project backends-velox: The following files had format violations:
Error:      src/test/scala/org/apache/gluten/execution/DynamicOffHeapSizingSuite.scala
Error:          @@ -72,8 +72,7 @@
Error:           ??}
Error:           
Error:           ??test("Dynamic?off-heap?sizing?with?setting?offheap")?{
Error:          -????withSQLConf(
Error:          -??????GlutenCoreConfig.SPARK_OFFHEAP_ENABLED_KEY?->?"false")?{
Error:          +????withSQLConf(GlutenCoreConfig.SPARK_OFFHEAP_ENABLED_KEY?->?"false")?{
Error:           ??????if?(DynamicOffHeapSizingMemoryTarget.isJava9OrLater())?{
Error:           ????????val?query?=
Error:           ??????????"""
Error:  Run 'mvn spotless:apply' to fix these violations.
```

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
```diff
@@ -118,6 +118,7 @@ private object GlutenDriverPlugin extends Logging {
             // 1GB default
             1024 * 1024 * 1024
           }
+        conf.set(GlutenCoreConfig.SPARK_OFFHEAP_SIZE_KEY, "0")
```
zhouyuan, line 132: @zhli1142015 this config is required during gluten startup, previously if this is set in spark.conf then gluten will fail to start

zhli1142015 (approved):

gluten-core/src/main/scala/org/apache/gluten/GlutenPlugin.scala
wForget: Should we allow both `DYNAMIC_OFFHEAP_SIZING_ENABLED` and `SPARK_OFFHEAP_ENABLED_KEY` to be enabled at the same time?

zhli1142015: When DYNAMIC_OFFHEAP_SIZING_ENABLED is enabled, the memory used by Velox is tracked under the on-heap category by the Spark memory manager, so Ithink there is no need to enable off-heap memory.

wForget: > When DYNAMIC_OFFHEAP_SIZING_ENABLED is enabled, the memory used by Velox is tracked under the on-heap category by the Spark memory manager, so Ithink there is no need to enable off-heap memory.

Thanks, that makes sense to me. @zhouyuan Can we point this out in the documentation?

zhouyuan: @wForget for now if dynamic offheap sizing feature is enabled, the spark off-heap settings will be override(ignored). We have already documented a bit on this behavior in below doc.  
https://github.com/apache/incubator-gluten/blob/main/docs/developers/VeloxDynamicSizingOffheap.md#dynamic-off-heap-sizing
I should add WARN log here if off-heap related settings are detected, so users may pay more attention on this.

wForget (approved):
Thanks @zhouyuan

zhouyuan:
@zhli1142015 @wForget Thanks.

Issue #10920, opened 2025-10-21
[VL] Performance regression on uniffle hash shuffle reader


wForget:
### Description

I've noticed a hash shuffle read performance regression when running gluten 1.5.0 tpcds (compared to gluten 1.2.0).

Gluten 1.2.0:

<img width="1887" height="93" alt="Image" src="https://github.com/user-attachments/assets/6a0c5f60-aacd-4d6b-81f5-5364349d5e9b" />

Gluten 1.5.0:

<img width="1915" height="105" alt="Image" src="https://github.com/user-attachments/assets/171734c7-4646-4a60-acd5-21833c41eb9f" />

<img width="1687" height="497" alt="Image" src="https://github.com/user-attachments/assets/fab8c8f1-4cd8-4a57-b7d0-1019de7c51ce" />

### Gluten version

None

wForget:
I found it's related to https://github.com/apache/incubator-gluten/pull/7897, after removing the input stream buffer the performance recovered.

<img width="1552" height="308" alt="Image" src="https://github.com/user-attachments/assets/8ac586e7-b9bb-49dd-b4e4-fd968b0f5626" />

wForget:
I suspect that calling memory copy of large memory using JNI may lead to bad performance, and I will create some test cases to prove it.

zuston:
The buffered read seems unnecessary for Uniffle, as a single read batch in Uniffle is already quite large — 14 MB by default (see #7897 for reference).

wForget:
By the way, I also tried #10499 and it had no noticeable benefit in this case.

wForget:
> The buffered read seems unnecessary for Uniffle, as a single read batch in Uniffle is already quite large — 14 MB by default (see [#7897](https://github.com/apache/incubator-gluten/pull/7897) for reference).

Thanks for the info, I submitted a PR #10922 to allow disabling the shuffle reader buffer.

PR #10922, opened 2025-10-22
[GLUTEN-10920][VL] Allow disabling hash/sort shuffle reader buffer

wForget:

Add buffer to the shuffle read input stream only if `readerBufferSize` is greater than 0.

## How was this patch tested?

Manually testing internal test case:

add `spark.gluten.sql.columnar.shuffle.readerBufferSize=1MB;` conf:
image

add `spark.gluten.sql.columnar.shuffle.readerBufferSize=0;` conf:

image

Related issue: #10920

### DIFF
--- a/cpp/velox/shuffle/VeloxShuffleReader.cc
+++ b/cpp/velox/shuffle/VeloxShuffleReader.cc
@@ -501,10 +501,14 @@ void VeloxHashShuffleReaderDeserializer::loadNextStream() {
     return;
   }
 
-  GLUTEN_ASSIGN_OR_THROW(
-      in_,
-      arrow::io::BufferedInputStream::Create(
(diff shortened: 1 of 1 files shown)


FelixYBW:
@marin-ma why it's onheap copy? shouldn't reducer use netty to load data into direct memory?

image

FelixYBW:
@wForget did you set spark.shuffle.io.preferDirectBufs=False?

zuston:
> @wForget did you set spark.shuffle.io.preferDirectBufs=False?

The issue occurs when using Uniffle, rather than the vanilla shuffle.

wForget:
> @marin-ma why it's onheap copy? shouldn't reducer use netty to load data into direct memory?

I filed #10923 for this issue

wForget:
Thanks @zuston @FelixYBW for the review, merged to main

PR #10930, opened 2025-10-23
[VL] Fix setup scripts with multiple commands to run

marin-ma:


dev/builddeps-veloxbe.sh
```diff
@@ -298,17 +298,17 @@ function setup_dependencies {
...
-  if [[ "x$commands_to_run" == "x" ]]; then
+  if [[ ${#commands_to_run[@]} -eq 0 ]]; then
     get_velox
     if [ -z "${GLUTEN_VCPKG_ENABLED:-}" ] && [ $RUN_SETUP_SCRIPT == "ON" ]; then
       setup_dependencies
     fi
     build_velox_backend
   else
-    echo "Commands to run: $commands_to_run"
-    for cmd in "$commands_to_run"; do
```
philo-he, line 311: It seems we can also fix it by removing the quote mark as follows:
`for cmd in $commands_to_run; do`

They are both good to me.

Issue #10933, opened 2025-10-23
[VL] Support GPU shuffle reader


jinchengchenghh:
### Description

1. Move the lock from WholeStageResultIterator constructor to shuffle reader, the threads can prepare the first batch in advance, now it is a small batch, for GPU, it will produce a big batch with bytes 1GB
2. Implement the decompression, resize batch in cpu, convert the batch to gpu when gpu is available
3. Save the batch to cache and consider the server memory to fetch the number of batches.
4. Polling threads to fetch the batch from several CPU threads and feed into GPU
5. The first batch is synchronous to fetch, may be a less size batch and processed by GPU to let GPU operate always.
6. The priority pool to fetch data from which threads(more batches, more priority)
7. The probe side batch is bigger that build side batch, and processed after build is finished, so the build threads will have high priority than probe when the memory is not enough.

### Gluten version

None

jinchengchenghh:
In CPU Gluten, we uses VeloxResizesBatchesExec to concat the batch after shuffle reader, this can help concat from different stream readers VS concat in ShuffleReader.
But in GPU gluten, if we split to cudf::table and concatenateTables, most of the concatenateTables operation is CPU work, because we add the GPU lock to restrict GPU threads before generating cudf::table, so we should do the concat in CPU side.

The lock is used for limited GPU memory. So most time the number of GPU threads should be less than CPU threads.

For example, the EC2 instance g4dn.2xlarge, CPU memory is 30G while GPU memory is 15G.

Future work:
- Operator CudfColumnarExchangeExec, save all the buffers to std::vector, wrap it to new class BufferColumnarBatch. For example, the flat Vector contains {BufferPtr nulls, BufferPtr values}, not it is {std::vector<BufferPtr> nulls, std::vector<BufferPtr> values}
- Operator VeloxBuffersBatchExec, The first thread uses GPU decompression while the other threads still uses CPU decompression, because the other threads cannot execute in GPU immediately, they need to wait for GPU execution.
- The other threads fetch the BufferColumnarBatch and concat to big batch continuously util there is not enough memory, save the Batches to std::vector<ColumnarBatch> and wait the downstream operator to fetch.
- After decompression, merge the buffer especially for the validity buffer which requires ByteForBits, different copy, and also different for the offset buffer, needs to recompute the offset.
- Lock and construct cudf::table

Question:

Is the first thread GPU decompression make senses? Looks like complex and may not benefit because the concat buffer is in single GPU thread in these cases.

PR #10934, opened 2025-10-23
[GLUTEN-10933][VL] cuDF: Move lock to ShuffleReader

jinchengchenghh:
The lock in WholeStageResultIterator restrict the cpu thread to produce batch, move the lock to here can let threads produce first batch with 1 GB in advance. Maybe the threads should prepare more data and let the GPU consume, this depends on the GPU operator time.

Need to restrict the total stage can offload to GPU, otherwise, after fallback, the lock cannot make effect on GPU execution.

Related issue: #10933

### DIFF
--- a/cpp/velox/cudf/GpuLock.cc
+++ b/cpp/velox/cudf/GpuLock.cc
@@ -38,35 +38,35 @@ GpuLockState& getGpuLockState() {
 
 void lockGpu() {
     std::thread::id tid = std::this_thread::get_id();
-    std::unique_lock<std::mutex> lock(gGpuMutex);
-    if (gGpuOwner == tid) {
+    std::unique_lock<std::mutex> lock(getGpuLockState().gGpuMutex);
+    if (getGpuLockState().gGpuOwner == tid) {
         // Reentrant call from the same thread — do nothing
         return;
     }
 
 
     // Wait until the GPU lock becomes available
-    gGpuCv.wait(lock, [] {
-        return !gGpuOwner.has_value();
+    getGpuLockState().gGpuCv.wait(lock, [] {
+        return !getGpuLockState().gGpuOwner.has_value();
     });
 
     // Acquire ownership
-    gGpuOwner = tid;
+    getGpuLockState().gGpuOwner = tid;
 }
 
 void unlockGpu() {
     std::thread::id tid = std::this_thread::get_id();
-    std::unique_lock<std::mutex> lock(gGpuMutex);
-    if (!gGpuOwner.has_value() || gGpuOwner != tid) {
+    std::unique_lock<std::mutex> lock(getGpuLockState().gGpuMutex);
+    if (!getGpuLockState().gGpuOwner.has_value() || getGpuLockState().gGpuOwner != tid) {
         throw std::runtime_error("unlockGpu() called by non-owner thread!");
     }
 
     // Release ownership
-    gGpuOwner = std::nullopt;
+    getGpuLockState().gGpuOwner = std::nullopt;
 
     // Notify one waiting thread
     lock.unlock();
-    gGpuCv.notify_one();
+    getGpuLockState().gGpuCv.notify_one();
 }
 
 


cpp/velox/cudf/GpuLock.cc
```diff
@@ -0,0 +1,64 @@
...
+#include <mutex>
+#include <condition_variable>
+#include <optional>
+#include <stdexcept>
+
+namespace gluten {
+
+static std::mutex gGpuMutex;
+static std::condition_variable gGpuCv;
+static std::optional<std::thread::id> gGpuOwner;
```
zhztheplayer:
```suggestion
namespace {
struct GpuLockState {
  std::mutex gGpuMutex;
  std::condition_variable gGpuCv;
  std::optional<std::thread::id> gGpuOwner;
};

GpuLockState& getGpuLockState() {
  static GpuLockState gGpuLockState;
  return gGpuLockState;
}
}
```

I would suggest the above change because I used to encounter some SIOF issues with similar global static variable code in the past. Construct-on-First-Use can definitely prevent the risk of wrong initialization order.

jinchengchenghh:
1. Refactored.
2. We will only offload the stage which can totally run in GPU.
4. Yes, this will cause the lock does not take effect on GPU, now we only check the WholeStageTransformer operator, in my design, the total stage should can offload to GPU, I will add restriction in following PR, do not contain any format conversion in the stage plan.

Thanks for your advice! @zhztheplayer

cpp/velox/shuffle/GpuShuffleReader.cc
```diff
@@ -0,0 +1,393 @@
...
+  std::vector<BufferPtr> veloxBuffers;
+  veloxBuffers.reserve(arrowBuffers.size());
+  for (auto& buffer : arrowBuffers) {
+    veloxBuffers.push_back(convertToVeloxBuffer(std::move(buffer)));
+  }
+  auto rowVector = deserialize(type, numRows, veloxBuffers, dictionaryFields, dictionaries, pool);
+  return std::make_shared<VeloxColumnarBatch>(std::move(rowVector));
+}
+
+} // namespace
```
zhztheplayer, line 299: Hi @jinchengchenghh, any idea on reducing the code duplication of this part?

Should we just put the GPU reader implementation to VeloxShuffleReader.h / VeloxShuffleReader.cc ? cc @marin-ma

jinchengchenghh: These all will be removed after support Buffer to cudf::table, a draft version is https://github.com/jinchengchenghh/gluten/blob/shuffle/cpp/velox/shuffle/GpuShuffleReader.cc

zhztheplayer: Thank you for the explanation.

backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
```diff
@@ -454,7 +454,8 @@ class CHSparkPlanExecApi extends SparkPlanExecApi with Logging {
   override def createColumnarBatchSerializer(
       schema: StructType,
       metrics: Map[String, SQLMetric],
-      shuffleWriterType: ShuffleWriterType): Serializer = {
+      shuffleWriterType: ShuffleWriterType,
+      enableCudf: Boolean): Serializer = {
```
marin-ma, line 458: Just noticed this change. Is it possible to just add one or more new `ShuffleWriterType` for cuff so we can avoid adding new parameters?

jinchengchenghh: Shuffle writer is same, only the shuffle reader is different. Each ShuffleWriterType matches to its own GpuShuffleReader

marin-ma: The shuffle writer type is also used on the reader side to determine which deserialiser to use. If the deserialisation are different, then would be better to have different shuffle writer type like `gpu_hash`, `gpu_sort`, etc.

jinchengchenghh: Make sense, and I notice the shuffle writer type name exists in cpp and scala code, should we use protobuf for them?

marin-ma: Sure. Would be nice to have a better way to align them.

PR #10942, opened 2025-10-24
[VL] Validate runtime compatibility using reflection before registering Gluten components

philo-he:

In Gluten, components like Iceberg depend on an Iceberg JAR (e.g., iceberg-spark-runtime-3.4_2.12), which is in "provided" scope. If Iceberg is enabled at build time or a Gluten release JAR (with all feature modules enabled) is used, the absence of corresponding classes at runtime can result in a `ClassNotFoundException`.

By validating runtime compatibility through a core class check, we can:

- Safely reference classes in the component module without relying on reflection which brings perf overhead.
- Prevent `ClassNotFoundException` issues during development of the related modules.
- Improve the efficiency for Spark plan transformation by avoiding the execution of inapplicable rules injected by runtime-not-compatible components.

Related discussion: https://github.com/apache/incubator-gluten/pull/10900#discussion_r2447034497

## How was this patch tested?
Local test and CI verification.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/component/package.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/component/package.scala
@@ -28,11 +28,13 @@ package object component extends Logging {
       return
     }
 
-    // Load all components in classpath.
-    val all = Discovery.discoverAll().filter(_.isRuntimeCompatible)
-
-    // Register all components.
-    all.foreach(_.ensureRegistered())
+    // Discover all components available in the classpath.
+    val all = Discovery.discoverAll()
+    val (compatibleComponents, incompatibleComponents) = all.partition(_.isRuntimeCompatible)
+    incompatibleComponents.foreach(
+      c => logWarning(s"Excluding runtime-incompatible component: ${c.name}"))
+    // Register all runtime-compatible components.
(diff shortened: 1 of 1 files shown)


backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
```diff
@@ -25,12 +25,24 @@ import org.apache.gluten.extension.columnar.validator.Validators
...
     Component.BuildInfo("VeloxDelta", "N/A", "N/A", "N/A")
   override def dependencies(): Seq[Class[_ <: Component]] = classOf[VeloxBackend] :: Nil
+
+  override def isRuntimeCompatible: Boolean = {
+    try {
+      SparkReflectionUtil.classForName("io.delta.sql.DeltaSparkSessionExtension")
+      true
+    } catch {
+      case _: ClassNotFoundException =>
+        false
```
wForget, line 42: Should we fail fast or add some warn logs?

philo-he: @wForget, I've added some warning message for components that are filtered out in the check. Thanks.

Issue #10950, opened 2025-10-27
[VL] 'spark.hadoop.fs.s3a.connection.timeout' by velox requires time unit


zml1206:
### Backend

VL (Velox)

### Bug description

```
Caused by: org.apache.gluten.exception.GlutenException: Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: Invalid duration '200000'
Retriable: False
Context: Split [Hive: s3a://***/_delta_log/00000000000000002170.checkpoint.parquet 0 - 26719] Task Gluten_Stage_3_TID_27_VTID_23
Function: toDuration
File: /opt/gluten/ep/build-velox/build/velox_ep/velox/common/config/Config.cpp
Line: 88
Stack trace:
# 0  _ZN8facebook5velox7process10StackTraceC1Ei
# 1  _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2  _ZN8facebook5velox6detail14veloxCheckFailINS0_14VeloxUserErrorERKSsEEvRKNS1_18VeloxCheckFailArgsET0_
# 3  _ZN8facebook5velox6config10toDurationERKSs
# 4  _ZN8facebook5velox11filesystems12S3FileSystem4ImplC1ERKNS1_8S3ConfigE
# 5  _ZN8facebook5velox11filesystems12S3FileSystemC2ESt17basic_string_viewIcSt11char_traitsIcEESt10shared_ptrIKNS0_6config10ConfigBaseEE
# 6  _ZN8facebook5velox11filesystems19fileSystemGeneratorESt10shared_ptrIKNS0_6config10ConfigBaseEESt17basic_string_viewIcSt11char_traitsIcEE
# 7  _ZNSt17_Function_handlerIFSt10shared_ptrIN8facebook5velox11filesystems10FileSystemEES0_IKNS2_6config10ConfigBaseEESt17basic_string_viewIcSt11char_traitsIcEEEPSE_E9_M_invokeERKSt9_Any_dataOS9_OSD_
```

### Spark version

None

### Relevant logs

```bash

```

xumanbu:
However, the configuration spark.hadoop.fs.s3a.connection.establish.timeout has the same issue. I'll commit a follow up pr.

PR #10951, opened 2025-10-27
[GLUTEN-10950][VL] Add time unit to 'spark.hadoop.fs.s3a.connection.timeout'

zml1206:
Fixes: #10950

## How was this patch tested?

Local test

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxTransformerApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxTransformerApi.scala
@@ -71,9 +71,8 @@ class VeloxTransformerApi extends TransformerApi with Logging {
       nativeConfMap: JMap[String, String],
       backendPrefix: String): Unit = {
     // 'spark.hadoop.fs.s3a.connection.timeout' by velox requires time unit, hadoop-aws versions
-    // before 3.4 do not have time units.
-    val s3sConnectionTimeout =
-      nativeConfMap.getOrDefault("spark.hadoop.fs.s3a.connection.timeout", "200000")
+    // before 3.4 do not have time unit.
+    val s3sConnectionTimeout = nativeConfMap.get("spark.hadoop.fs.s3a.connection.timeout")
     if (NumberUtils.isCreatable(s3sConnectionTimeout)) {
       nativeConfMap.put("spark.hadoop.fs.s3a.connection.timeout", s"${s3sConnectionTimeout}ms")
     }


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxTransformerApi.scala
```diff
@@ -69,7 +69,10 @@ class VeloxTransformerApi extends TransformerApi with Logging {
   override def postProcessNativeConfig(
       nativeConfMap: JMap[String, String],
       backendPrefix: String): Unit = {
-    // TODO: IMPLEMENT SPECIAL PROCESS FOR VELOX BACKEND
+    // 'spark.hadoop.fs.s3a.connection.timeout' by velox requires time unit.
+    nativeConfMap.put(
+      "spark.hadoop.fs.s3a.connection.timeout",
+      s"${nativeConfMap.getOrDefault("spark.hadoop.fs.s3a.connection.timeout", "200000")}ms")
```
wForget: Is the value of `spark.hadoop.fs.s3a.connection.timeout` always without time unit?

wForget: I saw from [hadoop-aws documentation ](https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/performance.html)that its default value is with unit.

Image

zml1206: https://github.com/apache/hadoop/blob/84e8b89ee2ebe6923691205b9e171badde7a495c/hadoop-common-project/hadoop-common/src/main/resources/core-default.xml#L1636-L1640
https://github.com/apache/hadoop/blob/1be78238728da9266a4f88195058f08fd012bf9c/hadoop-common-project/hadoop-common/src/main/resources/core-default.xml#L1441-L1445
There is a version difference here, I will make it compatible, thanks.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxTransformerApi.scala
```diff
@@ -69,7 +70,13 @@ class VeloxTransformerApi extends TransformerApi with Logging {
   override def postProcessNativeConfig(
       nativeConfMap: JMap[String, String],
       backendPrefix: String): Unit = {
-    // TODO: IMPLEMENT SPECIAL PROCESS FOR VELOX BACKEND
+    // 'spark.hadoop.fs.s3a.connection.timeout' by velox requires time unit, hadoop-aws versions
+    // before 3.4 do not have time units.
+    val s3sConnectionTimeout =
+      nativeConfMap.getOrDefault("spark.hadoop.fs.s3a.connection.timeout", "200000")
```
wForget: Do not set the default value, keep the original default behavior.

wForget:
If this is the configuration of s3 client, would it be better to make it compatible on the velox side?

zml1206:
> If this is the configuration of s3 client, would it be better to make it compatible on the velox side?

https://github.com/facebookincubator/velox/issues/12753#issuecomment-2783643175

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxTransformerApi.scala
zml1206: I think it is OK to add or not add the default value, because hadoop-aws has a default value.

FelixYBW:
We need to do all spark's configuration conversion in Gluten since Velox is a general library.

wForget:
> We need to do all spark's configuration conversion in Gluten since Velox is a general library.

`fs.s3a.xxx` looks like a s3 client configuration rather than spark specific configuration. But I accept converting it in gluten.

zml1206:
Thanks @wForget @FelixYBW , merge into main.

xumanbu:
@wForget @zml1206  However, the configuration `spark.hadoop.fs.s3a.connection.establish.timeout` has the same issue. It would be better if Velox provided broader compatibility with S3-related Spark/Hadoop configuration keys.

FelixYBW:
These configurations all have issues. I'd think it's something wrong in the prebuild spark4.0.1 jar because it needs to set without unit explicitly. Gluten/velox needs to set with unit which is default behavior.

It's better to fix in Velox to support the config without unit.

```
    "        self.conf.set('spark.hadoop.fs.s3a.threads.keepalivetime','60')",
    "            .set('spark.hadoop.fs.s3a.connection.establish.timeout','30')",
    "            .set('spark.hadoop.fs.s3a.connection.timeout','200')",
    "            .set('spark.hadoop.fs.s3a.multipart.purge.age','24')"
```

zml1206:
> It's better to fix in Velox to support the config without unit.


@FelixYBW Velox disagrees to support. https://github.com/facebookincubator/velox/issues/12753#issuecomment-2783643175

PR #10952, opened 2025-10-27
[VL] Add dlerror message for library load failures

liujiayi771:

Print dlerror() output when dlsym() fails to load function in shared library.

## How was this patch tested?

N/A

### DIFF
--- a/cpp/velox/udf/UdfLoader.cc
+++ b/cpp/velox/udf/UdfLoader.cc
@@ -36,12 +36,13 @@ void* loadSymFromLibrary(
     const std::string& libPath,
     const std::string& func,
     bool throwIfNotFound = true) {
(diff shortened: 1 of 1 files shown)


cpp/velox/udf/UdfLoader.cc
```diff
@@ -36,9 +36,12 @@ void* loadSymFromLibrary(
     const std::string& libPath,
     const std::string& func,
     bool throwIfNotFound = true) {
+  dlerror();
```
wForget, line 40: Is this a redundant call?

liujiayi771: This will clear the dlerror before.

wForget: > This will clear the dlerror before.

Got it, thanks.

PR #10958, opened 2025-10-28
[VL][MINOR] Replace NULL with nullptr

beliefer:

This PR proposes to replace `NULL` with `nullptr`.
In C++, `nullptr` should always be used instead of `NULL` because:

- Provide better type security
- Avoid overloading and resolving ambiguity
- Better performance in template programming
- The code intent is clearer
- It is a standard practice in modern C++

Only consider using `NULL` when maintaining very old code repositories or interacting with C interfaces that only accept `NULL`.

## How was this patch tested?

GA tests.

### DIFF
--- a/docs/developers/CppCodingStyle.md
+++ b/docs/developers/CppCodingStyle.md
@@ -125,6 +125,8 @@ cmake-format --first-comment-is-literal True --in-place cpp/velox/CMakeLists.txt
 ## Constant
 
 * Prefer const variables to using preprocessor (`#define`) to define constant values.
+* Always use nullptr if you need a constant that represents a null pointer (T* for some T);
+    use 0 otherwise for a zero value.
 
 ## Macro
 


wForget:
The original code is referenced from https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Heap-Profiling. If `NULL` and `nullptr` are equivalent, this change looks good to me. @PHILO-HE Could you please take a look?

beliefer:
@wForget Thanks ! I think https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Heap-Profiling just give us an use case.

FelixYBW:
Can we add it to the https://github.com/apache/incubator-gluten/blob/main/dev/format-cpp-code.sh?

FelixYBW:
> @FelixYBW, it seems clang-format doesn't support the check for using nullptr. 

I mean to add a sed like `sed -i 's/NULL/nullptr/g'`

philo-he:
@FelixYBW, I feel that using sed command may cause unintentional replacement issue.

I just investigated Velox. It uses clang-tidy to check code style based on a configuration file, where we can set to replace NULL with nullptr, see https://github.com/facebookincubator/velox/blob/8a5a520fba06ab1fc10f124f3eb000bab886cd5d/.clang-tidy#L102.

If this sounds good to you, I can help enable clang-tidy in another PR. It can also cover some other code styles checks.

FelixYBW:
> If this sounds good to you, I can help enable clang-tidy in another PR. It can also cover some other code styles checks.

Thank you, @PHILO-HE good idea.

FelixYBW (approved):

beliefer:
@PHILO-HE @FelixYBW @wForget  Thank you all

Issue #10961, opened 2025-10-28
[CH] Invalid result for `full join` without join condition


lgbo-ustc:
### Backend

CH (ClickHouse)

### Bug description

[Expected behavior] and [actual behavior].

For following query
```sql
select a,b from (
    select id as a from range(1)
) as t1
full outer join (
    select * from ( select id as b from range(2) ) t2 where b > 10
) as t3
```

`vanilla`'s result is 
```
0, null
```

but `gluten`'s result is empty. 

This is a conner case, the `full join` is tranformed into `cross join`, `cross join` returns empty if any table is empty.

### Spark version

None

### Relevant logs

```bash

```

PR #10967, opened 2025-10-29
[GLUTEN-10961][CH] Support joins without any join clause

lgbo-ustc:

Support joins without any join clauses. For example
```sql
select a, b from (select id as a from range(1) ) full join (select id as b from range(2));
```

## How was this patch tested?

Tested by UTs

Related issue: #10961


lgbo-ustc:
depents on https://github.com/Kyligence/ClickHouse/pull/520

Issue #10969, opened 2025-10-29
Flink unit tests maybe get stuck when some failures happened in velox


xinghuayu007:
### Backend

VL (Velox)

### Bug description

Function 'runAndCheck' in gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/common/GlutenStreamingTestBase.java waits for the result in file 'taskmanager.out'. If some failures happend in velox, the file keeps empty, the function waits for the result always the time.

How to reproduce:
1. register not equal function to gluten for data type integer (gluten-flink/planner/src/main/java/org/apache/gluten/rexnode/functions/RexCallConverterFactory.java), but it is not supported by velox.

          Map.entry(
              "<>",
              Arrays.asList(
                  () -> new BasicArithmeticOperatorRexCallConverter("notequalto"),
                  () -> new StringCompareRexCallConverter("notequalto"),
                  () -> new StringNumberCompareRexCallConverter("notequalto"))),

2. run an unit test like the following:
 @Test
  void testNEqual() {
    List<Row> rows =
        Arrays.asList(Row.of(1, 1L, "2", "1"), Row.of(2, 2L, "2", "2"), Row.of(3, 3L, "2", "1"));
    createSimpleBoundedValuesTable("tblLess", "a int, b bigint, c string, d string", rows);
    String query = "select a <> 2 as x from tblLess where a > 0";
    runAndCheck(query, Arrays.asList("+I[true]", "+I[false]", "+I[true]"));
  }

3. the unit test gets stuck.

### Spark version

None

### Relevant logs

```bash

```

PR #10970, opened 2025-10-29
[GLUTEN-10969][FLINK] Set timeout for waiting for task output

xinghuayu007:

Flink unit tests wait for the result file in circle, but if some failures happened in velox, the result file keeps empty.
This PR set timeout for waiting for the result file.

## How was this patch tested?

Related issue: #10969

### DIFF
--- a/gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/common/GlutenStreamingTestBase.java
+++ b/gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/common/GlutenStreamingTestBase.java
@@ -138,7 +138,11 @@ public class GlutenStreamingTestBase extends StreamingTestBase {
             Thread.sleep(10);
           }
           long fileSize = -1L;
+          startTime = System.currentTimeMillis();
           while (printResultFile.length() > fileSize) {
+            if (System.currentTimeMillis() - startTime > timeoutMS) {
+              break;
+            }
             fileSize = printResultFile.length();
             Thread.sleep(3000);
           }


gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/common/GlutenStreamingTestBase.java
```diff
@@ -129,7 +130,11 @@ protected void runAndCheck(String query, List<String> expected) {
       JobClient jobClient = tableResult.getJobClient().get();
       if (deleteResultFile) {
         try {
+          long timeout = System.currentTimeMillis() + timeoutMS;
           while (!printResultFile.exists()) {
+            if (System.currentTimeMillis() > timeoutMS) {
+              break;
+            }
```
wForget:
```suggestion
          long startTime = System.currentTimeMillis();
          while (!printResultFile.exists()) {
            if (System.currentTimeMillis() - startTime > timeoutMS) {
              break;
            }
```

xinghuayu007: done.Thanks for your comments.

gluten-flink/ut/src/test/java/org/apache/gluten/table/runtime/stream/common/GlutenStreamingTestBase.java
wForget, line 142: Does this while loop also need a timeout?

xinghuayu007: It is better to add a timeout. In my case, if the result file is empty, it would not get stuck. But if the result outputs every slowly, it is better to set a timeout.

wForget:
Thanks @xinghuayu007 , merged to main

PR #10973, opened 2025-10-29
[VL] Base Delta read support for Spark 4.0 + Delta 4.0

zhztheplayer:
The PR adds base support of Delta Lake read for Spark 4.0 + Delta 4.0.


zhztheplayer:
I've tested the change locally with ANSI disabled. We don't have the unit testing CI set up for Spark 4.0, so the change is also not verified by CI.

Moving forward one of the highest priorities for Spark 4.0 support could be adding a basic Spark 4.0 test CI.

Issue #10980, opened 2025-10-30
[CH] Unix_timestamp diff with valina spark


KevinyhZou:
### Backend

CH (ClickHouse)

### Bug description

`select unix_timestamp('2021-10-01', 'yyyy-M-d')`

gluten result: NULL
spark: 2021-10-01

### Spark version

None

### Relevant logs

```bash

```

PR #10981, opened 2025-10-30
[GLUTEN-10980][CH]Fix unix_timestamp diff when format is not `yyyy-MM-dd` or `yyyy-MM-dd HH:mm:ss`

KevinyhZou:

Fix diff of unix_timestamp when date format is not `yyyy-MM-dd` or `yyyy-MM-dd HH:mm:ss`

## How was this patch tested?

UT

Related issue: #10980

### DIFF
--- a/cpp-ch/local-engine/Parser/scalar_function_parser/getTimestamp.h
+++ b/cpp-ch/local-engine/Parser/scalar_function_parser/getTimestamp.h
@@ -23,6 +23,7 @@
 #include <Parser/FunctionParser.h>
 #include <boost/algorithm/string/case_conv.hpp>
 #include <Common/CHUtil.h>
+#include <regex>
 
 namespace DB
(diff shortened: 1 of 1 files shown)


cpp-ch/local-engine/Parser/scalar_function_parser/getTimestamp.h
```diff
@@ -104,5 +104,48 @@ class FunctionParserGetTimestamp : public FunctionParser
             return convertNodeTypeIfNeeded(substrait_func, result_node, actions_dag);
         }
     }
+
+private:
```
lgbo-ustc, line 108: Checking by regex should be better

KevinyhZou: fixed

Issue #9785, opened 2025-05-28
[VL] Pass additional configuration options programmatically from Java to C++


zhztheplayer:
Enhance this API https://github.com/apache/incubator-gluten/blob/2feb5f936ccb81b11b053f27cbfbf653e1a3c742/gluten-arrow/src/main/java/org/apache/gluten/vectorized/NativePlanEvaluator.java#L46-L47 by adding a parameter that allows additional configuration options being passed to the native Velox runtime.

By doing this we could avoid adding JNI parameters every time we need some customization ([example](https://github.com/apache/incubator-gluten/pull/9634/files#diff-4f3b38d38be472d3634702bf87c6ac1a990997eafe4207f67ce48f7b5f550bb1R401)) for native execution.

For example, JNI signature

```cpp
Java_org_apache_gluten_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithIterator( // NOLINT
    JNIEnv* env,
    jobject wrapper,
    jbyteArray planArr,
    jobjectArray splitInfosArr,
    jobjectArray iterArr,
    jint stageId,
    jint partitionId,
    jlong taskId,
    jboolean enableDumping,
    jstring spillDir,
    jboolean enableCudf)
```

may be simplified to

```cpp
Java_org_apache_gluten_vectorized_PlanEvaluatorJniWrapper_nativeCreateKernelWithIterator( // NOLINT
    JNIEnv* env,
    jobject wrapper,
    jbyteArray planArr,
    jobjectArray splitInfosArr,
    jobjectArray iterArr)
```

once the feature is implemented and used.

PR #10984, opened 2025-10-30
[GLUTEN-10933][VL] Introduce GPU ShuffleWriterType kGpuHashShuffle

jinchengchenghh:
Related issue: #10933

### DIFF
--- a/backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarBatchSerializer.scala
+++ b/backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarBatchSerializer.scala
@@ -103,8 +103,7 @@ private class CelebornColumnarBatchSerializerInstance(
         batchSize,
         readerBufferSize,
         deserializerBufferSize,
-        shuffleWriterType.name,
-        false
+        shuffleWriterType.name
       )
     // Close shuffle reader instance as lately as the end of task processing,
     // since the native reader could hold a reference to memory pool that


marin-ma (approved):
LGTM. Just one more question: how do we currently test` gpu_hash` shuffle?

jinchengchenghh:
Now it is tested on my local environment, because it relies on GPU, I test TPCDS Q95 for previous PR, this is just a refactor, I don't test it. I will add cpp test for GPU shuffle reader.

zhztheplayer:
BTW We have the same parameter `enableCudf` for Velox plan execution (See `NativePlanEvaluator.java`). I was hoping we could finally pass these parameters from Java to C++ using configuration options. https://github.com/apache/incubator-gluten/issues/9785

jinchengchenghh:
Yes, use pair parameter name `enableCudf` would help understand the logic. What do you think about that? @marin-ma 
After resolve https://github.com/apache/incubator-gluten/issues/9785, the logic could be simple.

marin-ma:
@jinchengchenghh I think for shuffle, we still can stick with adding different writer type. Again the shuffle writer type is a match for different shuffle reader deserialisation, and using different writer type it's easy to distinguish them from avoiding introducing params like "isRssShuffle", "isGpuShuffle", etc.

Moreover, `ShuffleWriterType.name` is the display string on Spark UI. We can easily tell whether the shuffle is gpu shuffle with this change.

jinchengchenghh:
I find I also  need to update shuffle writer, but shuffleWriterType does not send to native, do you have any suggestion? @marin-ma 
```
          nativeShuffleWriter = if (isSort) {
            shuffleWriterJniWrapper.createSortShuffleWriter(
              numPartitions,
              dep.nativePartitioning.getShortName,
              GlutenShuffleUtils.getStartPartitionId(
                dep.nativePartitioning,
                taskContext.partitionId),
              conf.get(SHUFFLE_DISK_WRITE_BUFFER_SIZE).toInt,
              conf.get(SHUFFLE_SORT_INIT_BUFFER_SIZE).toInt,
              conf.get(SHUFFLE_SORT_USE_RADIXSORT),
              partitionWriterHandle
            )
          } else {
            shuffleWriterJniWrapper.createHashShuffleWriter(
              numPartitions,
              dep.nativePartitioning.getShortName,
              GlutenShuffleUtils.getStartPartitionId(
                dep.nativePartitioning,
                taskContext.partitionId),
              nativeBufferSize,
              reallocThreshold,
              partitionWriterHandle
            )
          }
```

marin-ma:
@jinchengchenghh Do you have the code for the changes in shuffle writer? I don't see any modifications to the existing shuffle writer.

jinchengchenghh:
In the followup PR, Velox bool is bit, but cudf is BOOL8, it is byte

marin-ma:
@jinchengchenghh Let's discuss and address that in your next pr. Based on the discussion offline seems like there are many discrepancies between cudf and the current hash shuffle writer implementation.

Issue #10966, opened 2025-10-29
[VL] Support codegen for ArrowProjection


wForget:
### Description

During upgrade from gluten 1.2 to 1.5, I found that PartialProject with complex expressions could cause performance regressions. Supporting codegen for ArrowProjection should improve performance.

gluten 1.2:

<img width="574" height="828" alt="Image" src="https://github.com/user-attachments/assets/8bee9c38-c286-4c64-b3c0-afc60342781a" />

gluten 1.5:

<img width="569" height="829" alt="Image" src="https://github.com/user-attachments/assets/5c2019ef-d7a4-442c-b7ef-36900be09c41" />

### Gluten version

Gluten-1.5

PR #10986, opened 2025-10-31
[GLUTEN-10966][VL][FOLLOWUP] Fix UnsupportedOperationException when the result type of expressions in GenerateArrowProjection is array

jiangjiangtian:
When running CI in this [PR](https://github.com/apache/incubator-gluten/pull/10637), there is an exception as follows:
```
Function:operator(), Expression:  Operator::getOutput failed for [operator: ValueStream, plan node ID: 0]: Error during calling Java code from native code: java.lang.UnsupportedOperationException
	at org.apache.gluten.vectorized.ArrowWritableColumnVector$ArrowVectorAccessor.getInts(ArrowWritableColumnVector.java:857)
	at org.apache.gluten.vectorized.ArrowWritableColumnVector.getInts(ArrowWritableColumnVector.java:605)
	at org.apache.spark.sql.vectorized.ColumnarArray.toIntArray(ColumnarArray.java:102)
	at org.apache.spark.sql.vectorized.ColumnarArray.copy(ColumnarArray.java:79)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificArrowProjection.apply(Unknown Source)
	at org.apache.gluten.execution.ColumnarPartialProjectExec.$anonfun$getProjectedBatchArrow$3(ColumnarPartialProjectExec.scala:241)
	at org.apache.gluten.execution.ColumnarPartialProjectExec.$anonfun$getProjectedBatchArrow$3$adapted(ColumnarPartialProjectExec.scala:239)
	at scala.collection.immutable.Range.foreach(Range.scala:158)
	at org.apache.gluten.execution.ColumnarPartialProjectExec.org$apache$gluten$execution$ColumnarPartialProjectExec$$getProjectedBatchArrow(ColumnarPartialProjectExec.scala:239)
	at org.apache.gluten.execution.ColumnarPartialProjectExec$$anon$1.next(ColumnarPartialProjectExec.scala:180)
	at org.apache.gluten.execution.ColumnarPartialProjectExec$$anon$1.next(ColumnarPartialProjectExec.scala:164)
   ...
```
The reason is that the generated code will call `copy` on the results and finally call `getxxxs` on `ArrowWritableColumnVector`. But `ArrowWritableColumnVector` doesn't implement these methods.
So this PR follows `InterpretedArrowProjection`, in the generated class, we use an `Object` array called `values` to store the results of expressions and use these `values` to create `GenericInternalRow`.

## How was this patch tested?
Unit tests.

Related issue: #10966

### DIFF
--- a/gluten-arrow/src/main/scala/org/apache/gluten/expression/GenerateArrowProjection.scala
+++ b/gluten-arrow/src/main/scala/org/apache/gluten/expression/GenerateArrowProjection.scala
@@ -88,8 +88,14 @@ object GenerateArrowProjection extends CodeGenerator[Seq[Expression], ArrowProje
             FalseLiteral)
         }
         // update value into intermediate
-        val update = CodeGenerator
-          .updateColumn("intermediate", e.dataType, i, ExprCode(isNull, value), e.nullable)
+        val update =
+          s"""
+             |if (!$isNull) {
+             |  values[$i] = $value;
+             |} else {
+             |  values[$i] = null;
+             |}
+             |""".stripMargin
         (code, update)
     }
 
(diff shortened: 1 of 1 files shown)


wForget (approved):
Thanks @jiangjiangtian

Issue #10989, opened 2025-10-31
[VL] Full outer join without join condition in BroadcastNestedLoopJoinExecTransformer failed


zml1206:
### Backend

VL (Velox)

### Bug description

```
java.lang.IllegalArgumentException: BroadcastNestedLoopJoin should not take FullOuter as the JoinType with building right side
```

### Spark version

None

### Relevant logs

```bash

```

PR #10990, opened 2025-10-31
[GLUTEN-10989][VL] Fix outputPartitioning of BroadcastNestedLoopJoinExecTransformer

zml1206:
Velox supports full out join without join condition , and we should not add exceptions to the `outputPartitioning` because `doValidate` should ensure safety.
Fixes: #10989

## How was this patch tested?

UT


zml1206:

jinchengchenghh:
As this discussion, https://github.com/apache/incubator-gluten/pull/8965#issuecomment-3357318380, we will remove this support

zml1206:
> #8965 (comment)


Thanks @jinchengchenghh. Let me confirm first. If it needs to be removed, I will resubmit a PR to remove it.

PR #10991, opened 2025-10-31
[GLUTEN-10988][VL] Do not resize batches for sort-based/rss-sort shuffle

### LINKED ISSUE #10988: [VL] Makes `resizeBatches.shuffleOutput` work only for hash shuffle
### Description

The sort-based shuffle reader already respects the batch size configuration, so we don't need to support `resizeBatches.shuffleOutput` for it.

https://github.com/apache/incubator-gluten/blob/2f7b138f24f16a249cea57217e50962d3b1a8ee4/cpp/velox/shuffle/VeloxShuffleReader.cc#L602

### Gluten version

None

wForget:

The sort-based shuffle reader already respects the batch size configuration, so we don't need to support `resizeBatches.shuffleOutput` for it.

sort:
https://github.com/apache/incubator-gluten/blob/2f7b138f24f16a249cea57217e50962d3b1a8ee4/cpp/velox/shuffle/VeloxShuffleReader.cc#L602
rss-sort:
https://github.com/apache/incubator-gluten/blob/2f7b138f24f16a249cea57217e50962d3b1a8ee4/cpp/velox/shuffle/VeloxShuffleReader.cc#L766

closes #10988

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
@@ -31,12 +31,14 @@ import scala.collection.JavaConverters._
 
 trait ShuffleWriterType {
   val name: String
-  val requiresResizingShuffleInput: Boolean = true
-  val requiresResizingShuffleOutput: Boolean = true
+  val requiresResizingShuffleInput: Boolean
+  val requiresResizingShuffleOutput: Boolean
 }
 
 case object HashShuffleWriterType extends ShuffleWriterType {
   override val name: String = ReservedKeys.GLUTEN_HASH_SHUFFLE_WRITER
+  override val requiresResizingShuffleInput: Boolean = true
+  override val requiresResizingShuffleOutput: Boolean = true
 }
 
 case object SortShuffleWriterType extends ShuffleWriterType {
@@ -53,6 +55,8 @@ case object RssSortShuffleWriterType extends ShuffleWriterType {
 
 case object GpuHashShuffleWriterType extends ShuffleWriterType {
   override val name: String = ReservedKeys.GLUTEN_GPU_HASH_SHUFFLE_WRITER
+  override val requiresResizingShuffleInput: Boolean = true
+  override val requiresResizingShuffleOutput: Boolean = true
 }
 
 /*


marin-ma:
Could you only remove this for the specific shuffle writer type `SortShuffleWriterType` and `RssSortShuffleWriterType`? There's going to be more shuffle writer type such as gpu shuffle writer in the future.

wForget:
> Could you only remove this for the specific shuffle writer type `SortShuffleWriterType` and `RssSortShuffleWriterType`? There's going to be more shuffle writer type such as gpu shuffle writer in the future.

@marin-ma Thank you. I have only removed support for `SortShuffleWriterType/RssSortShuffleWriterType`. Could you please take another look?

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -30,30 +31,50 @@ import org.apache.spark.sql.execution.adaptive.{AQEShuffleReadExec, ShuffleQuery
...
         VeloxResizeBatchesExec(s, range.min, range.max)
     }
   }
+
+  private object ColumnarResizeableShuffleExchangeExec {
+    def unapply(plan: SparkPlan): Option[ColumnarShuffleExchangeExec] = {
+      plan match {
+        // sort-based/rss-sort shuffle has already resized the batch size in shuffle read,
+        // so no need to resize again
+        case shuffle: ColumnarShuffleExchangeExec
```
marin-ma: nit: Perhaps a more readable way for this logic

```
case shuffle: ColumnarShuffleExchangeExec
            if shuffle.shuffleWriterType != SortShuffleWriterType
              && shuffle.shuffleWriterType != RssSortShuffleWriterType => Some(shuffle)
case _ => None
```

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -30,30 +31,50 @@ import org.apache.spark.sql.execution.adaptive.{AQEShuffleReadExec, ShuffleQuery
...
+    if (!resizeBatchesShuffleInputEnabled && !resizeBatchesShuffleOutputEnabled) {
+      return plan
+    }
+
     val range = VeloxConfig.get.veloxResizeBatchesShuffleInputOutputRange
     plan.transformUp {
-      case shuffle: ColumnarShuffleExchangeExec
-          if shuffle.shuffleWriterType == HashShuffleWriterType &&
-            VeloxConfig.get.veloxResizeBatchesShuffleInput =>
+      case ColumnarResizeableShuffleExchangeExec(shuffle) if resizeBatchesShuffleInputEnabled =>
```
marin-ma: It seems this change assumes a specific shuffle writer type requires both shuffle input and output require resizing at the same time, but I don't think that's necessarily true. We may assume for a shuffle type it may need input resizing but not output resizing, and vice versa. (Although for the existing types they are the same).

Maybe we can generalize this by adding these flags to the ShuffleWriterType trait:
```
trait ShuffleWriterType {
  val name: String
  val requiresResizingShuffleInput: Boolean
  val requiresResizingShuffleOutput: Boolean
}
```

What do you think?

wForget: Thanks, sounds good. I’ll make that change.

wForget: @marin-ma Thank you for your suggestion, I have updated it. Also, do we need to change the default value of `COLUMNAR_VELOX_RESIZE_BATCHES_SHUFFLE_OUTPUT` to true?
https://github.com/apache/incubator-gluten/blob/95f271d7f683deb9d61fb019806d6d221d0fc6c9/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala#L279-L286

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -31,6 +31,8 @@ import scala.collection.JavaConverters._
 
 trait ShuffleWriterType {
   val name: String
+  val requiresResizingShuffleInput: Boolean = true
+  val requiresResizingShuffleOutput: Boolean = true
```
marin-ma: Please move these into `HashShuffleWriterType`

wForget: Thanks, updated

Issue #10926, opened 2025-10-22
[VL] Prepare for GCC-13 upgrade


zhouyuan:
### Description

Velox community is preparing to adopt Gcc-13 
https://github.com/facebookincubator/velox/discussions/15236
For now Velox backend is using Centos-7  + GCC-11 for CI. May need some efforts to upgrade on the toolchain if it happens.  

### Gluten version

None

PR #10993, opened 2025-10-31
[GLUTEN-10926][VL] Add GCC-13 image for CI

zhouyuan:
Velox community is trying to upgrade to GCC-13 which has better support of C++-20
https://github.com/facebookincubator/velox/discussions/15236
This patch adds the Gcc-13 image for CI. Will switch to it after Velox upgrade is done

fixes: #10926 

## How was this patch tested?
pass GHA

Related issue: #10926

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -170,7 +170,7 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: ${{ env.DOCKERHUB_REPO }}
-          tags: vcpkg-centos-8
+          tags: vcpkg-centos-8-gcc13
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
@@ -199,7 +199,7 @@ jobs:
       - name: Upload digest
         uses: actions/upload-artifact@v4
         with:
-          name: digests-vcpkg-centos-8-${{ matrix.os }}
+          name: digests-vcpkg-centos-8-gcc13-${{ matrix.os }}
           path: ${{ runner.temp }}/digests/*
           if-no-files-found: error
           retention-days: 1
@@ -388,6 +388,7 @@ jobs:
         digests: [ vcpkg-centos-8, vcpkg-centos-8-gcc13, vcpkg-centos-9, centos-8-jdk8, centos-8-jdk11, centos-8-jdk17, centos-9-jdk8, centos-9-jdk11, centos-9-jdk17 ]
     needs:
       - build-vcpkg-centos-8
+      - build-vcpkg-centos-8-gcc13
       - build-vcpkg-centos-9
       - build-centos-8
       - build-centos-9


dev/vcpkg/triplets/x64-linux-avx.cmake
```diff
@@ -8,3 +8,7 @@ set(VCPKG_CMAKE_SYSTEM_NAME Linux)
 set(VCPKG_C_FLAGS "-mavx2 -mfma -mavx -mf16c -mlzcnt -mbmi2")
 set(VCPKG_CXX_FLAGS "-mavx2 -mfma -mavx -mf16c -mlzcnt -std=c++17 -mbmi2")
 set(VCPKG_LINKER_FLAGS "-static-libstdc++ -static-libgcc")
+
+if("${PORT}" STREQUAL "snappy")
+    list(APPEND VCPKG_CMAKE_CONFIGURE_OPTIONS -DSNAPPY_HAVE_BMI2=ON)
+endif()
```
philo-he: nit: please leave an empty at the end.

dev/docker/Dockerfile.centos8-gcc13
philo-he, line 1: Should we name it `Dockerfile.centos8-gcc13` for consistency?

dev/docker/Dockerfile.centos7-gcc13
philo-he, line 1: ditto

philo-he:
I just had another thought. Please consider if it makes sense. :)

Can we replace CentOS 7 with CentOS 8 in our CI verification for PRs? Using CentOS 8 could reduce our maintenance efforts for the most critical CI jobs. This change wouldn't mean dropping CentOS 7 support. Instead, we could implement a weekly job to monitor CentOS 7 compatibility.

zhouyuan:
@PHILO-HE 
>Can we replace CentOS 7 with CentOS 8

I assume it's for the native lib part. Using Centos7 as the base building system so that Gluten only rely on Glibc-2.17, which should not introduce any issues for production env with legacy systems. However on Centos8, the glibc version is 2.28 which may introduce some issue for old systems

philo-he:
> @PHILO-HE
> 
> > Can we replace CentOS 7 with CentOS 8
> 
> I assume it's for the native lib part. Using Centos7 as the base building system so that Gluten only rely on Glibc-2.17, which should not introduce any issues for production env with legacy systems. However on Centos8, the glibc version is 2.28 which may introduce some issue for old systems

@zhouyuan With Centos 8 used to build Gluten native libs, I think we can still transfer the built binary to Ubuntu 20.04/22.04, Centos 9 for test verification, excluding Centos 7. For Centos 7, we only need a weekly job to track build and runtime compatibility.

We've already encountered issues with CentOS 7, such as GitHub checkout incompatibility. I am afraid that unnecessarily insisting on using Centos 7 in primary CI pipelines will bring more burdens to us.

While GCC 13 installation is technically possible on CentOS 7, it may introduce runtime compatibility problems due to the old system libraries. Given these challenges, maybe at some time point, we'll have to drop CentOS 7 support. Moving it to an independent weekly job would allow us to smoothly end the support someday without impacting our primary CI pipeline. If you agree, I can start this replacement work in the coming weeks.

I'm totally OK to first merge this PR. Maybe in the medium term, we still need Centos 7 GCC 13 for building release, etc.

zhouyuan:
@PHILO-HE 

>I am afraid that unnecessarily insisting on using Centos 7 in primary CI pipelines will bring more burdens to us.

I think the key question is do we still need to keep the compatibility for old Glibc or not. Using a weekly job does not fix this key issue. As far as i know there are some real production deployments still based on CentOS 7(they may included local security patches) in China. That's the main reason to keep using Centos-7


>While GCC 13 installation is technically possible on CentOS 7, it may introduce runtime compatibility problems due to the old system libraries

Most of the dependencies are statically linked, the only dependency left is for the glibc 
```
ldd libvelox.so
	linux-vdso.so.1 =>  (0x00007ffd30343000)
	libgluten.so => /opt/gluten/cpp/build/releases/libgluten.so (0x00007f55b8d9b000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f55b8b93000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f55b8977000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f55b8773000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f55b8471000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f55b80a3000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f55c1e15000)
```

PR #10999, opened 2025-11-03
[CORE] Follow-up: switch to direct class references guarded by component's runtime-compatibility validation

philo-he:

With runtime-compatibility validation introduced by https://github.com/apache/incubator-gluten/pull/10942 for component, we can (relatively) safely use class references in components modules. This PR removes the use of reflection or class string comparison, and also improves some reflection code which is still needed.

### DIFF
--- a/backends-clickhouse/src-delta/main/scala/org/apache/gluten/component/CHDeltaComponent.scala
+++ b/backends-clickhouse/src-delta/main/scala/org/apache/gluten/component/CHDeltaComponent.scala
@@ -29,12 +29,17 @@ import org.apache.gluten.sql.shims.DeltaShimLoader
 import org.apache.spark.SparkContext
 import org.apache.spark.api.plugin.PluginContext
 import org.apache.spark.sql.execution.{FilterExec, ProjectExec}
+import org.apache.spark.util.SparkReflectionUtil
 
 class CHDeltaComponent extends Component {
   override def name(): String = "ch-delta"
   override def buildInfo(): Component.BuildInfo =
     Component.BuildInfo("CHDelta", "N/A", "N/A", "N/A")
   override def dependencies(): Seq[Class[_ <: Component]] = classOf[CHBackend] :: Nil
+
+  override def isRuntimeCompatible: Boolean = {
+    SparkReflectionUtil.isClassPresent("io.delta.sql.DeltaSparkSessionExtension")
+  }
 
   override def onDriverStart(sc: SparkContext, pc: PluginContext): Unit =
     DeltaShimLoader.getDeltaShims.onDriverStart(sc, pc)
--- a/backends-clickhouse/src-iceberg/main/scala/org/apache/gluten/component/CHIcebergComponent.scala
+++ b/backends-clickhouse/src-iceberg/main/scala/org/apache/gluten/component/CHIcebergComponent.scala
@@ -14,18 +14,25 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.gluten.component
 
 import org.apache.gluten.backendsapi.clickhouse.CHBackend
 import org.apache.gluten.extension.OffloadIcebergScan
 import org.apache.gluten.extension.injector.Injector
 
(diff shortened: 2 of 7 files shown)


philo-he:
@Zouxxyy, @liujiayi771, could you please take a look?

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergScanTransformer.scala
```diff
@@ -253,7 +253,7 @@ object IcebergScanTransformer {
   }
 
   def supportsBatchScan(scan: Scan): Boolean = {
-    scan.getClass.getName == "org.apache.iceberg.spark.source.SparkBatchQueryScan"
+    scan.getClass == GlutenIcebergSourceUtil.getClassOfSparkBatchQueryScan
```
Zouxxyy, line 256: What about using isInstanceOf too

philo-he: @Zouxxyy Thanks for the review.

The logic remains identical to the previous implementation by explicitly excluding inherited relationships. Since `SparkBatchQueryScan` has no subclass, I think this check for strict equality looks better for us.

zhztheplayer:
> With runtime-compatibility validation introduced by https://github.com/apache/incubator-gluten/pull/10942 for component, we can safely use class references in components modules.

The change of this PR looks good, but I am trying to understand why #10942 is required for this PR. As far as I understand, no matter a component is valid or not, error will be thrown when we are trying to access a class that is not present in class path by reference. No?

philo-he:
> > With runtime-compatibility validation introduced by #10942 for component, we can safely use class references in components modules.
> 
> The change of this PR looks good, but I am trying to understand why #10942 is required for this PR. As far as I understand, no matter a component is valid or not, error will be thrown when we are trying to access a class that is not present in class path by reference. No?

@zhztheplayer, thanks for the review. Here are some historical discussions:
https://github.com/apache/incubator-gluten/pull/10900#discussion_r2446868970
https://github.com/apache/incubator-gluten/pull/10186#discussion_r2280763905

In the component validation, we checked the presence of a core class in the classpath, assuming that other classes in the corresponding JAR would also be available. While there are some rare edge cases (which likely requires fixes) that violates this, I think we can ignore them here.

zhztheplayer:
> In the component validation, we checked the presence of a core class in the classpath, assuming that other classes in the corresponding JAR would also be available. While there are some rare edge cases (which likely requires fixes) that violates this, I think we can ignore them here.

Thank you for the explanation. I get the reason now.

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergScanTransformer.scala
zhztheplayer: Note: Equality check on class: Java `equals`or Scala `==` depends on class loader. If the class was loaded by a custom class loader in the 3rd plugin, this quality may be tested false negative. Having said that, this is a theoretical risk.

philo-he:

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergScanTransformer.scala
liujiayi771: The CH CI failure might be due to the equality comparison here.

philo-he: The failure comes from the ClassNotFoundException for SparkBatchQueryScan.

I just noted that CH backend module has its own definition for iceberg component. Like velox backend, we also need to add validation check to determine if the component should be registered. Then, we can relatively safely use these iceberg class references. Just updated the code.

philo-he: I think we can keep using class object equality check. I will follow up this if any related failure is reported. Thanks.

PR #11004, opened 2025-11-03
[VL] Upgrade vcpkg builtin-baseline

philo-he:

Upgrade vcpkg baseline to easily use newer libraries and build tools. The latest tag 2025.10.17 requires a higher ninja version which is not compatible with Centos 7 glibc. So tag 2025.09.17 is set in this PR.

## How was this patch tested?
CI.

### DIFF
--- a/dev/vcpkg/README.md
+++ b/dev/vcpkg/README.md
@@ -4,7 +4,7 @@
 
 Currently, the `builtin-baseline` set in `vcpkg.json` is the commit hash for the `2025.09.17` tag of vcpkg.
 The versions of all dependency libraries are determined by their respective ports at this vcpkg version,
-except for those overridden in `vcpkg.json` and `vcpkg-configuration.json`.
+except for those overridden in `vcpkg.json`, `vcpkg-configuration.json`, and overlay ports.
 
 ## Build in docker
 
@@ -25,7 +25,7 @@ Please install build depends on your system to compile all libraries:
 sudo $GLUTEN_REPO/dev/vcpkg/setup-build-depends.sh
 ```
 
-GCC-11 is the minimum required compiler. It needs to be enabled beforehand. Take Centos-7/8 as example:
+GCC 11 is the minimum required compiler. It needs to be enabled beforehand. Take Centos 7/8 as example:
 
 ``` sh
 # CentOS 8


zhouyuan:
@PHILO-HE as Gluten already defined overlays, directly bumping is not working, note initially we have already included some compiler fixes for Gcc-11
https://github.com/apache/incubator-gluten/compare/main...zhouyuan:gluten:main_docker_fix_1027

philo-he:
> @PHILO-HE as Gluten already defined overlays, directly bumping is not working, note initially we have already included some compiler fixes for Gcc-11 [main...zhouyuan:gluten:main_docker_fix_1027](https://github.com/apache/incubator-gluten/compare/main...zhouyuan:gluten:main_docker_fix_1027)

@zhouyuan, thanks for the review. I assume these overlays are prioritized to be used. With baseline upgraded, I expect we can directly use upstream ports for some new versions of dependency libraries, which can ease the maintenance.

zhouyuan:
@PHILO-HE I think we will also need to 

- bump gcc-13 / 14 image to fully support with C++ 20 (depends on https://github.com/apache/incubator-gluten/pull/11064)

- bump the SW version to align with velox(https://github.com/facebookincubator/velox/blob/main/scripts/setup-versions.sh)

philo-he:
> @PHILO-HE I think we will also need to
> 
> * bump gcc-13 / 14 image to fully support with C++ 20 (depends on [[GLUTEN-10926][VL] followup to fix docker image conflict #11064](https://github.com/apache/incubator-gluten/pull/11064))
> * bump the SW version to align with velox(https://github.com/facebookincubator/velox/blob/main/scripts/setup-versions.sh)

@zhouyuan, it seems they can be fixed in separate PRs? If so, we can first merge this one once all failures are fixed. I note some libs are not consistent with Velox's setup versions for a long time.

dev/vcpkg/init.sh
```diff
@@ -68,13 +68,10 @@ SCRIPT_ROOT="$(realpath "$(dirname "$0")")"
 cd "$SCRIPT_ROOT"
 
 if [ ! -d "$VCPKG_ROOT" ] || [ -z "$(ls "$VCPKG_ROOT")" ]; then
-    git clone https://github.com/microsoft/vcpkg.git --branch 2023.10.19 "$VCPKG_ROOT"
+    git clone https://github.com/microsoft/vcpkg.git --branch 2025.09.17 "$VCPKG_ROOT"
```
zhouyuan, line 73: `2025.09.17`
it seems not match the PR desc here

dev/vcpkg/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
-  "builtin-baseline": "a7b6122f6b6504d16d96117336a0562693579933",
+  "builtin-baseline": "4334d8b4c8916018600212ab4dd4bbdc343065d1",
```
zhouyuan, line 3: could we also add a note for this commit:  `branch 2025.09.17`

zhouyuan (approved):
👍

dev/vcpkg/vcpkg.json
philo-he: It seems adding comments is not recommended for JSON file. So I just added an Overview section in `dev/vcpkg/README.md` to clarify this. Thanks.

philo-he:
I just tested this PR with GCC 13 docker. Those dependency libraries can be built successfully with GCC 13. But when building velox, some errors related to folly are reported to complain the use of uninitialized variable, which should be caused by stricter check of GCC 13. I think we can fix them by upgrading folly in another PR.

PR #11006, opened 2025-11-03
[CORE] Add a test case where dependencies of a component are not registered

zhztheplayer:
https://github.com/apache/incubator-gluten/pull/10942 introduced a method to exclude incompatible components at runtime. This PR adds an anti test case matching the current code, awaiting future improvements.

The PR also includes some code cleanup for ComponentSuite.

### DIFF
--- a/gluten-core/src/test/scala/org/apache/gluten/component/ComponentSuite.scala
+++ b/gluten-core/src/test/scala/org/apache/gluten/component/ComponentSuite.scala
@@ -68,15 +68,12 @@ class ComponentSuite extends AnyFunSuite with BeforeAndAfterAll {
     val c = new DummyComponent("C") {}
 
     c.dependsOn(a)
-
     c.ensureRegistered()
-
     assertThrows[IllegalArgumentException] {
       Component.sorted()
     }
 
     a.ensureRegistered()
-
     assert(Component.sorted().filter(Seq(a, c).contains(_)) === Seq(a, c))
   }
 }


gluten-core/src/test/scala/org/apache/gluten/component/ComponentSuite.scala
```diff
@@ -49,53 +55,53 @@ class ComponentSuite extends AnyFunSuite with BeforeAndAfterAll {
...
-}
 
-object ComponentSuite {
-  private class DummyBackendA extends Backend {
-    override def name(): String = "dummy-backend-a"
-    override def buildInfo(): Component.BuildInfo =
-      Component.BuildInfo("DUMMY_BACKEND_A", "N/A", "N/A", "N/A")
-    override def injectRules(injector: Injector): Unit = {}
-  }
+  test("Dependencies not registered") {
```
zhztheplayer, line 66: @PHILO-HE, this test case shows an error is thrown when we only exclude a component, but still register its dependents at runtime.

Moving forward we'd exclude all dependents as well when a component is excluded.

This is not really a corner case but something practical - for example, if `VeloxDeltaComponent` is excluded, `VeloxDelta33WriteComponent` has to be excluded as well. Otherwise, an exception will be thrown during component registration.

zhztheplayer:
@PHILO-HE @wForget Would you please review? Thanks!

gluten-core/src/test/scala/org/apache/gluten/component/ComponentSuite.scala
philo-he: @zhztheplayer, yes, for this case, the dependent components should also be excluded.

zhztheplayer:
@PHILO-HE Thanks for reviewing.

gluten-core/src/test/scala/org/apache/gluten/component/ComponentSuite.scala
zhztheplayer: Yes. I can take a look after this PR is merged.

PR #11008, opened 2025-11-03
[CORE] ComponentSuite: Add a test case against dependency cycles

zhztheplayer:
As the test suite is refactored in https://github.com/apache/incubator-gluten/pull/11006, we can now easily add new unit test cases in ComponentSuite. 

This PR adds another test case to guard against Component discovery to avoid cyclic dependency graph.


zhztheplayer:
Ping @PHILO-HE, thanks.

PR #11015, opened 2025-11-04
[CH] Include iceberg module in spotless plugin and apply formatting fixes

philo-he:
The backends-clickhouse module has its own setting for "includes" for scala code, which overrides the setting in parent pom. The code under "src-icerberg" should be included.

## How was this patch tested?
Local test.


philo-he:
@ lgbo-ustc, could you take a look?

PR #11019, opened 2025-11-04
[MINOR] Add Gluten-1.5 and spark-4.0.x option to bug issue template

wForget:

Add Gluten-1.5 and spark-4.0.x option to bug issue template

## How was this patch tested?

minor change

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -50,6 +50,7 @@ body:
       description: Please provide the gluten version in your environment.
       options:
         - main branch
+        - Gluten-1.5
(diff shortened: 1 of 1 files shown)


yaooqinn (approved):

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -50,6 +50,7 @@ body:
       description: Please provide the gluten version in your environment.
       options:
         - main branch
+        - Gluten-1.5
```
philo-he, line 53: The proposed changes make sense.

I am wondering if we need to establish an EOL policy to allow marking some old release branches EOL. Then, those branches can be removed from this list.

PR #11022, opened 2025-11-04
[VL] Fix to use unified docker image name in CI

zhouyuan:

This patch unifies the docker image name in CI. 
In this way we can add version tag for the docker images in different branch

## How was this patch tested?
pass GHA

### DIFF
--- a/.github/workflows/velox_backend_enhanced.yml
+++ b/.github/workflows/velox_backend_enhanced.yml
@@ -44,12 +44,24 @@ env:
   WGET_CMD: 'wget -nv'
   SETUP: 'bash .github/workflows/util/setup-helper.sh'
   CCACHE_DIR: "${{ github.workspace }}/.ccache"
+  CI_DOCKER_IMAGE_PREFIX_STATIC: 'apache/gluten:vcpkg-centos'
+  CI_DOCKER_IMAGE_PREFIX_DYNAMIC: 'apache/gluten:centos'
 
 concurrency:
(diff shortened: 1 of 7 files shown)


.github/workflows/velox_backend_enhanced.yml
```diff
@@ -95,9 +106,9 @@ jobs:
           if-no-files-found: error
 
   spark-test-spark34:
-    needs: build-native-lib-centos-7
+    needs: [build-native-lib-centos-7, setup-image-prefix-for-test]
     runs-on: ubuntu-22.04
-    container: apache/gluten:centos-8-jdk8
+    container: ${{ needs.setup-image-prefix-for-test.outputs.image_tag_prefix_dynamic }}-8-jdk8
```
philo-he, line 111: @zhouyuan Could you explain the benefit of introducing `setup-image-prefix-for-test`?

I feel it may be good enough for us by declaring environment variables at the beginning of each workflow, like
```
DOCKER_IMAGE_FOR_BUILD: 'apache/gluten:vcpkg-centos-7'
DOCKER_IMAGE_FOR_TEST: 'apache/gluten:centos-8'
```
Then, we can use these env arg to specify the docker image for running jobs. We can also flexibly specify some other public images (e.g. ubuntu:20.04) in the job section.

For branch development, we can directly change these environment variables to use images specific to a branch.

Please let me know if I missed anything.

zhouyuan: @PHILO-HE 
This is a workaround of limitation of GHA - it does not accept env in container image filed
https://github.com/orgs/community/discussions/26324

PR #11025, opened 2025-11-04
[VL] TEST for scanonly enable by default

zhouyuan:


FelixYBW:
Good catch.

We may test with dynamic memory enabled together. jdk8 is an issue

Issue #11027, opened 2025-11-05
Use clang-tidy to check cpp files


xinghuayu007:
### Description

Use clang-tidy to check cpp files to make code robot.
1. Install clang-tidy tool
Download the tool from: https://github.com/llvm/llvm-project/releases/download/llvmorg-18.1.8/clang+llvm-18.1.8-x86_64-linux-gnu-ubuntu-18.04.tar.xz
3. Compile the code
bash dev/builddeps-veloxbe.sh
4. run clang-tidy check script
python3 cpp/run-clang-tidy.py --fix fix --build-dir cpp/build/ --project-root /app/incubator-gluten

### Gluten version

None

PR #11028, opened 2025-11-05
[GLUTEN-11027] Use clang-tidy to check cpp files

xinghuayu007:

Use clang-tidy to check cpp files to make code robot.

## How was this patch tested?

Related issue: #11027


philo-he:
Assume with this script, the portal to trigger the check is here: https://github.com/apache/incubator-gluten/blob/056d0e32034c367764d58a514a73fc9126b78862/dev/check.py#L170

The installation for clang-tidy seems necessary. In addition, if running the entire check requires too much time, we need to let it only check the CPP files changed by a PR.

PR #11032, opened 2025-11-05
[GLUTEN-9785][VL] Pass additional configuration options programmatically from Java to C++

jinchengchenghh:
Related issue: #9785

### DIFF
--- a/gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtimes.scala
+++ b/gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtimes.scala
@@ -22,17 +22,7 @@ import java.util
 
 object Runtimes {
 
-  /** Get or create the runtime which bound with Spark TaskContext. */
-  def contextInstance(backendName: String, name: String): Runtime = {
-    if (!TaskResources.inSparkTask()) {
-      throw new IllegalStateException("This method must be called in a Spark task.")
-    }
-    TaskResources.addResourceIfNotRegistered(
-      s"$backendName:$name",
-      () => Runtime(backendName, name))
-  }
-
-  def newInstance(
+  def contextInstance(
       backendName: String,
       name: String,
       extraConf: util.Map[String, String]): Runtime = {
@@ -42,6 +32,11 @@ object Runtimes {
     TaskResources.addResourceIfNotRegistered(
       s"$backendName:$name:$extraConf",
       () => Runtime(backendName, name, extraConf))
+  }
+
+  /** Get or create the runtime which bound with Spark TaskContext. */
+  def contextInstance(backendName: String, name: String): Runtime = {
+    contextInstance(backendName, name, new util.HashMap[String, String]())
   }
 
 }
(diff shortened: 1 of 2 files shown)


gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtimes.scala
```diff
@@ -18,15 +18,25 @@ package org.apache.gluten.runtime
 
 import org.apache.spark.task.TaskResources
 
+import java.util
+
 object Runtimes {
 
   /** Get or create the runtime which bound with Spark TaskContext. */
-  def contextInstance(backendName: String, name: String): Runtime = {
+  def contextInstance(
```
zhztheplayer: It turns out that we might have to add a new API, something  like the following and use it for creating runtime instances with extra configuration, because these configurations we pass down to native might be different in different invocations.

The original `contextInstance` finds existing instance in the cache so is fast. But if the operations are done less frequently, like when we create plan evaluator instances, probably it's tolerable to create a new instance every single time.

```scala
  def contextInstance(backendName: String, name: String, extraConf: Map[String, String]): Runtime = {
    if (!TaskResources.inSparkTask()) {
      throw new IllegalStateException("This method must be called in a Spark task.")
    }
    TaskResources.addResource(s"$backendName:$name:$extraConf", Runtime(backendName, name, extraConf))
  }
```

gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtimes.scala
```diff
@@ -40,7 +40,7 @@ object Runtimes {
       throw new IllegalStateException("This method must be called in a Spark task.")
     }
     TaskResources.addResourceIfNotRegistered(
-      s"$backendName:$name",
+      s"$backendName:$name:${extraConf.hashCode()}",
```
zhztheplayer: Thank you for updating. I got what's happening here. 

The code will work in most cases but there's a small chance that two different `extraConf` instance procude the same hashCode which is unexpected.

To address this, maybe we can either always create new instance with a unique ID in name (with `addResource`), or to pass `${extraConf}` for getting an instance from cache (with `addResourceIfNotRegistered`).

jinchengchenghh: Now there is only enableCudf 1 config, so I update to ${extraConf}. If there is more config then, we can try other solution. Total string will be slow.

gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtimes.scala
```diff
@@ -29,4 +31,17 @@ object Runtimes {
       s"$backendName:$name",
       () => Runtime(backendName, name))
   }
+
+  def newInstance(
```
zhztheplayer: If `addResourceIfNotRegistered` is called then the mothod name should be `contextInstance` for accuracy. Then we can make the other method `contextInstance(backendName: String, name: String)` a shortcut to this one, or merge the two methods.

gluten-arrow/src/main/scala/org/apache/gluten/runtime/Runtimes.scala
jinchengchenghh: Do we need to optimize the RunTime to stage level instead of Task level, get the stageId by val stageId = TaskContext.get().stageId(), and reuse the Runtime?

PR #11033, opened 2025-11-06
feat: move the output schema from Substrait's RelRoot into an extension

EpsilonPrime:
Moved the output schema used to solve the ClickHouse output schema nullability mismatch into an extension instead of being a core feature of Substrait.  This helps align the Gluten fork of Substrait with that of the core project.


EpsilonPrime:
Will solve this by removing output_schema instead.

PR #11041, opened 2025-11-06
[GLUTEN-11106][VL] Spark 3.5 / Delta 3.3: Add DeltaSuite / UpdateSQLSuite test suits

zhztheplayer:
As title. Import more test suites from Delta 3.3 to Gluten's code base.

Moving forward, we may need a way to leverage Delta's unit tests with less code change to Gluten. Maybe by downloading the test code during compilation time, or by release a test jar by ourselves.

Related issue: #11106

### DIFF
--- a/backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/UpdateSuiteBase.scala
+++ b/backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/UpdateSuiteBase.scala
@@ -857,14 +857,16 @@ abstract class UpdateSuiteBase
       set: String,
       where: String,
       expectException: Boolean,
-      customErrorRegex: Option[String] = None) {
+      customErrorRegex: Option[String] = None): Unit = {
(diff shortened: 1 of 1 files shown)


jinchengchenghh:
My PR also failed with  cpp-test-udf-test (pull_request), and same failure

zhztheplayer:
> My PR also failed with cpp-test-udf-test (pull_request), and same failure

Yes, that error seemed unrelated.

zhztheplayer:
> The directory name may be noticed, for iceberg, it is src-iceberg34, 34 means the Spark version34

I found `src-icebreg-spark34`. Am I getting something wrong here?

jinchengchenghh:
> > The directory name may be noticed, for iceberg, it is src-iceberg34, 34 means the Spark version34
> 
> I found `src-icebreg-spark34`. Am I getting something wrong here?
OK, I might have remembered the folder name wrong.

zhztheplayer:
@jinchengchenghh I see. Thanks for reviewing.

PR #11059, opened 2025-11-10
[VL] Remove buffering of sorted partitions in RSS writer to prevent OOM

boneanxs:
https://github.com/apache/incubator-gluten/pull/10244 removed `RssPartitionWriterOutputStream` which could buffer the whole partition in memory which could cause oom.

This pr directly push data to rssClient when sortEvict calls, since rssClient itself will also buffer data before sending to the remote.

### DIFF
--- a/gluten-celeborn/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
+++ b/gluten-celeborn/src/main/java/org/apache/spark/shuffle/gluten/celeborn/CelebornShuffleManager.java
@@ -134,6 +134,9 @@ public class CelebornShuffleManager
       rowBasedConf.set(SPARK_CELEBORN_COMPRESSION_CODEC_KEY, celebornDefaultCodec);
       rowBasedCelebornConf.set(CELEBORN_COMPRESSION_CODEC_KEY, celebornDefaultCodec);
     }
+
+    // Disable celeborn compression
+    celebornConf.set(CELEBORN_COMPRESSION_CODEC_KEY, "none");
   }
 
   private boolean isDriver() {


boneanxs:
@marin-ma @kerwin-zk hey, could you please help review this? Thanks!

cpp/core/shuffle/rss/RssPartitionWriter.cc
```diff
@@ -70,36 +60,25 @@ arrow::Status RssPartitionWriter::hashEvict(
...
-    lastEvictedPartitionId_ = partitionId;
-  }
-
   rawPartitionLengths_[partitionId] += inMemoryPayload->rawSize();
-  if (compressedOs_ != nullptr) {
-    RETURN_NOT_OK(inMemoryPayload->serialize(compressedOs_.get()));
+  ARROW_ASSIGN_OR_RAISE(
+      auto rssOs, arrow::io::BufferOutputStream::Create(options_->pushBufferMaxSize, arrow::default_memory_pool()));
+  if (codec_ != nullptr) {
+    ARROW_ASSIGN_OR_RAISE(
```
boneanxs, line 67: Do we need to compress data here again for rss shuffle? Looks there's a compression already inside shuffleClient: https://github.com/apache/celeborn/blob/5e4d80bb1e764b80f5d3462bb8ffb9061efc63b4/client/src/main/java/org/apache/celeborn/client/ShuffleClientImpl.java#L1052

@kerwin-zk

wForget: FYI: We have disabled compression of rss client(Uniffle) and only retained compression of the gluten shuffle.

boneanxs: make sense, celeborn also don't need to compress again IMO, let me remove it as well.

marin-ma:
@boneanxs Thanks for identifying this issue. I wonder if this change may affect the shuffled data size. The original design was aimed at generating a smaller compressed output by flushing the compressed data with more buffered input.

boneanxs:
> @boneanxs Thanks for identifying this issue. I wonder if this change may affect the shuffled data size. The original design was aimed at generating a smaller compressed output by flushing the compressed data with more buffered input.

@marin-ma Thanks for pointing that out. I tested it and found that it still produces less shuffle data than rss_sort. I can add more tests to measure how much the data volume increases when compared to buffering the entire partition. However, even if the volume does increase, I think we still can’t buffer the whole partition for large ones

wForget:
> which could buffer the whole partition in memory which could cause oom.

Doesn't rss shuffle writer trigger a spill?

boneanxs:
> > which could buffer the whole partition in memory which could cause oom.
> 
> Doesn't rss shuffle writer trigger a spill?

Spill still can't evict buffers, which will call sortEvict and then buffered in arrow bufferOutputStream

wForget:
> > > which could buffer the whole partition in memory which could cause oom.
> > 
> > 
> > Doesn't rss shuffle writer trigger a spill?
> 
> Spill still can't evict buffers, which will call sortEvict and then buffered in arrow bufferOutputStream

That's right, thank you for your explanation.

marin-ma:
@boneanxs Thanks for following.

> I tested it and found that it still produces less shuffle data than rss_sort.

Could you also compare with vanilla spark + celeborn?

> I can add more tests to measure how much the data volume increases when compared to buffering the entire partition.

It would be nice if there are more performance results to be shared.

boneanxs:
> @boneanxs Thanks for following.
> 
> > I tested it and found that it still produces less shuffle data than rss_sort.
> 
> Could you also compare with vanilla spark + celeborn?
> 
> > I can add more tests to measure how much the data volume increases when compared to buffering the entire partition.
> 
> It would be nice if there are more performance results to be shared.


@marin-ma I ran a test query on TPC-DS (3TB) using the following SQL:

```sql
SELECT
  ss.ss_customer_sk,
  ss.ss_item_sk,
  ss.ss_ticket_number,
  ss.ss_store_sk,
  ss.ss_promo_sk,
  ss.ss_sold_date_sk,
  c.c_customer_id,
  c.c_first_name,
  c.c_last_name,
  SUM(ss.ss_net_paid) AS total_paid
FROM 
(SELECT /*+ REPARTITION(50)*/ * from store_sales) ss
LEFT JOIN customer c
  ON ss.ss_customer_sk = c.c_customer_sk
GROUP BY
  ss.ss_customer_sk,
  ss.ss_item_sk,
  ss.ss_ticket_number,
  ss.ss_store_sk,
  ss.ss_promo_sk,
  ss.ss_sold_date_sk,
  c.c_customer_id,
  c.c_first_name,
  c.c_last_name
limit 100;
```

When comparing the first shuffle stage, I didn’t observe any performance regression with this patch applied.

1. Vanilla Spark
Screenshot 2025-11-14 at 10 35 33

2. Buffering all partitions + 512m offheap
Screenshot 2025-11-14 at 10 37 34

3. With this patch + 512m offheap
Screenshot 2025-11-14 at 10 38 38

4. With this patch + 256m off-heap (to force more spill)
Screenshot 2025-11-14 at 10 41 57

By counting the number of `VeloxCelebornColumnarShuffleWriter: Gluten shuffle writer: Trying to push` for the same partition, I can confirm there're more spill happens when reducing the memory.

cpp/core/shuffle/rss/RssPartitionWriter.cc
zuston: > make sense, celeborn also don't need to compress again IMO, let me remove it as well.

Maybe disabling compression on the Celeborn side by config(if having this option) is sufficient, and there is no need to implicitly disable it in the Gluten codebase.

marin-ma (approved):
LGTM. Thanks for confirming the performance results!

wForget (approved):

cpp/core/shuffle/rss/RssPartitionWriter.cc
kerwin-zk: @zuston This has already been implemented in the new version of Celeborn, and I'll adapt it accordingly later.

PR #11064, opened 2025-11-10
[GLUTEN-10926][VL] followup to fix docker image conflict

zhouyuan:

after #10926 the docker image for vcpkg-centos-8 has a conlict with vcpkg-centos-8-gcc13 when merging digests. 
this patch renames the merge pattern so there is not conflict.

## How was this patch tested?

Related issue: #10926

### DIFF
--- a/dev/vcpkg/CONTRIBUTING.md
+++ b/dev/vcpkg/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Please init vcpkg env first:
 
 Vcpkg already maintains a lot of libraries.
 You can find them by vcpkg cli.
-(NOTE: Please always use cli because [packages on vcpkg.io](https://vcpkg.io/en/packages.html) is outdate).
+(NOTE: Please always use cli because [packages on vcpkg.io](https://vcpkg.io/en/packages.html) is outdated).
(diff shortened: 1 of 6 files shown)


.github/workflows/docker_image.yml
```diff
@@ -385,10 +385,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        digests: [ vcpkg-centos-8, vcpkg-centos-8-gcc13, vcpkg-centos-9, centos-8-jdk8, centos-8-jdk11, centos-8-jdk17, centos-9-jdk8, centos-9-jdk11, centos-9-jdk17 ]
+        digests: [ vcpkg-centos-8, vcpkg-centos-8_gcc13, vcpkg-centos-9, centos-8-jdk8, centos-8-jdk11, centos-8-jdk17, centos-9-jdk8, centos-9-jdk11, centos-9-jdk17 ]
```
philo-he: How about creating a separate merge job for gcc13 dockers? Then, no impact due to the name pattern conflict.

I feel it may be better to keep the naming consistency with hyphen as separator.

PR #11067, opened 2025-11-11
[MINOR][VL] Fix unused variable when GPU was not enabled

wForget:

Fix unused variable when GPU was not enabled

## How was this patch tested?

minor fix

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -135,9 +135,10 @@ WholeStageResultIterator::WholeStageResultIterator(
     const auto& format = scanInfo->format;
     const auto& partitionColumns = scanInfo->partitionColumns;
     const auto& metadataColumns = scanInfo->metadataColumns;
+#ifdef GLUTEN_ENABLE_GPU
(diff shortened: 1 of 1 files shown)


cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -135,9 +135,10 @@ WholeStageResultIterator::WholeStageResultIterator(
     const auto& format = scanInfo->format;
     const auto& partitionColumns = scanInfo->partitionColumns;
     const auto& metadataColumns = scanInfo->metadataColumns;
+#ifdef GLUTEN_ENABLE_GPU
     // Under the pre-condition that all the split infos has same partition column and format.
```
jinchengchenghh, line 139: It is used in this line https://github.com/apache/incubator-gluten/pull/11067/files#diff-013353c1aa2ebe3f4c5c014b936dd52b2f91bef0a51421c1a901a57cac361f17L173

wForget: But it is only used when GLUTEN_ENABLE_GPU is defined.

jinchengchenghh (approved):

PR #11072, opened 2025-11-12
[VL] Make VeloxResizeBatchesExec not inherit ColumnarToColumnarTransition

zhztheplayer:
The PR's an improvement following #10763. 

It decouples `ColumnarToColumnarExec` from `ColumnarToColumnarTransition` to make them both can be solely implemented. If a class inherits from `ColumnarToColumnarExec` but not from `ColumnarToColumnarTransition` at the same time, it will not be considered as a transition, so it's not going to be removed by RemoveTransitions rule.

After the refactor, `VeloxResizeBatchesExec` will not inherit `ColumnarToColumnarTransition` and it will have nothing to do with the transition planner. So developers don't have to consider the transition logic when working with it or other similar unary columnar operators.


zhztheplayer:
cc @jinchengchenghh @Zouxxyy

zhztheplayer:
Thanks @Zouxxyy @jinchengchenghh

PR #11074, opened 2025-11-13
[CORE] Use Substrait timestamp_tz for Spark TimestampType to preserve timezone-aware semantics

liujiayi771:

Spark’s TimestampType is timezone-aware: it internally stores timestamps in UTC (by converting input values to UTC based on the session time zone or just read UTC timestamp from parquet file) and represents an absolute point in time. This semantics aligns with Substrait’s timestamp_tz type, which also denotes a timezone-aware timestamp that can be unambiguously mapped to a moment on the timeline.

To maintain semantic consistency between Spark and Substrait, this PR maps Spark’s TimestampType to Substrait’s timestamp_tz.

https://substrait.io/types/type_classes

This approach is consistent with other projects—for example, Apache Iceberg also maps Spark’s TimestampType to TIMESTAMP WITH TIME ZONE and Spark’s TimestampNTZ to TIMESTAMP WITHOUT TIME ZONE.

Note: For future support of Spark’s TimestampNTZType (timezone-naive timestamps), we should map it to Substrait’s timestamp type instead.

## How was this patch tested?

The existing tests already cover this change.

### DIFF
--- a/cpp-ch/local-engine/Builder/SerializedPlanBuilder.cpp
+++ b/cpp-ch/local-engine/Builder/SerializedPlanBuilder.cpp
@@ -107,7 +107,7 @@ SchemaPtr SerializedSchemaBuilder::build()
         else if (type == "Timestamp")
         {
             auto * t = type_struct->mutable_types()->Add();
-            t->mutable_timestamp()->set_nullability(
+            t->mutable_timestamp_tz()->set_nullability(
                 this->nullability_map[name] ? substrait::Type_Nullability_NULLABILITY_NULLABLE
                                             : substrait::Type_Nullability_NULLABILITY_REQUIRED);
         }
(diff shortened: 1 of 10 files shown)


liujiayi771:
@rui-mo @zhztheplayer Could you please help review this? @taiyang-li @lgbo-ustc @zzcclp could you confirm whether the changes here have any impact on the ClickHouse backend?

FelixYBW:
Spark3.5 added TIMESTAMP_NTZ data type. @rui-mo does velox support it? have we enabled the UT for for?

liujiayi771:
> Spark3.5 added TIMESTAMP_NTZ data type. @rui-mo does velox support it? have we enabled the UT for for?

This PR only modified the Substrait mapping, reserving the corresponding mapping type for future support of timestamp_ntz, velox doesn't support it yet.

FelixYBW:
We need a through clean up of timestamp and timezone support in Gluten sometime later.

taiyang-li:
> @rui-mo @zhztheplayer Could you please help review this? @taiyang-li @lgbo-ustc @zzcclp could you confirm whether the changes here have any impact on the ClickHouse backend?

CH backend changes are ok to me. Sorry for the late reply.

PR #11076, opened 2025-11-13
[CORE] Remove pushed down filter in  FilterExecTransformer

Zouxxyy:

Before we use  `getRemainingCondition` to caculate the real fitler condiction in FilterExecTransformer, in this pr, just remove pushed down filter in `FilterExecTransformer` to reflect the node's actual behavior.

### DIFF
--- a/gluten-ut/spark34/src/test/scala/org/apache/spark/sql/extension/TestFileSourceScanExecTransformer.scala
+++ b/gluten-ut/spark34/src/test/scala/org/apache/spark/sql/extension/TestFileSourceScanExecTransformer.scala
@@ -61,7 +61,6 @@ case class TestFileSourceScanExecTransformer(
 
   override val nodeNamePrefix: String = "TestFile"
 
-  override def withNewOutput(newOutput: Seq[Attribute]): BasicScanExecTransformer = {
+  override def withNewOutput(newOutput: Seq[Attribute]): BasicScanExecTransformer =
     copy(output = newOutput)
-  }
 }
--- a/gluten-ut/spark40/src/test/scala/org/apache/spark/sql/extension/TestFileSourceScanExecTransformer.scala
+++ b/gluten-ut/spark40/src/test/scala/org/apache/spark/sql/extension/TestFileSourceScanExecTransformer.scala
@@ -62,5 +62,4 @@ case class TestFileSourceScanExecTransformer(
 
   override def withNewOutput(newOutput: Seq[Attribute]): BasicScanExecTransformer =
     copy(output = newOutput)
-  }
 }


gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/RemovePushedDownFilter.scala
```diff
@@ -0,0 +1,49 @@
...
+        // For fallback scan, we need to keep original filter.
+        case _ =>
+          Seq.empty[Expression]
+      }
+      if (scanFilters.isEmpty) {
+        filter
+      } else {
+        val remainingFilters =
+          FilterHandler.getRemainingFilters(scanFilters, splitConjunctivePredicates(filter.cond))
+        val newCond = remainingFilters.reduceLeftOption(And).orNull
```
Zouxxyy: Perhaps we could remove the filterNode if newCond is null. If guys agree, I can include this change in this PR CC @zhztheplayer

wForget:
Do `pushedDownFilters` always work for all datasource scans? I'm worried that some datasource scan implementations might ignore some unsupported pushedDownFilters.

Zouxxyy:
> Do `pushedDownFilters` always work for all datasource scans? I'm worried that some datasource scan implementations might ignore some unsupported pushedDownFilters.

The logic should be the same as before, because we were already using getRemainingCondition to generate the Filter Transform previously. I just made the execution plan more accurately reflect its actual behavior. For example, users might notice there's a filter node, but in reality, it doesn't do anything.

wForget:
> > Do `pushedDownFilters` always work for all datasource scans? I'm worried that some datasource scan implementations might ignore some unsupported pushedDownFilters.
> 
> The logic should be the same as before, because we were already using getRemainingCondition to generate the Filter Transform previously. I just made the execution plan more accurately reflect its actual behavior. For example, users might notice there's a filter node, but in reality, it doesn't do anything.

Got it, thank you for your explanation.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/RemovePushedDownFilter.scala
zhztheplayer: > Perhaps we could remove the filterNode if newCond is null.

I previously attempted something similar, but ended up facing some attribute mismatch errors from Catalyst. 

I didn't look into that issue too much, so feel free to try from your end if you're interested.

Zouxxyy: > facing some attribute mismatch errors from Catalyst.

@zhztheplayer I checked the logic of FilterExec and confirmed that its output indeed changes—it equals the child's output plus some NOT NULL attributes added by the condition. Therefore, I made the following modifications:

- Extracted a helper method `buildNewOutput(output, conds)` from FilterExec.
- If pushedFilter nonEmpty:
  - Set ScanNode's newOutput = buildNewOutput(output, pushedFilter).
  - Set FilterNode's newCond = cond -- pushedFilter. (If newCond == null, remove the filter node.)

I believe this better reflects Node's correct behavior, although perhaps all golden files should be updated.

PR #11083, opened 2025-11-14
[VL] Update shuffle write metrics per batch

Yohahaha:

In our prod env, some stages write out hundreds of terabytes of shuffle data, but current UI does not shown it.

### DIFF
--- a/backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala
+++ b/backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala
@@ -85,8 +85,8 @@ class VeloxCelebornColumnarShuffleWriter[K, V](
         dep.metrics("shuffleWallTime").add(System.nanoTime() - startTime)
         dep.metrics("numInputRows").add(cb.numRows)
         dep.metrics("inputBatches").add(1)
-        // This metric is important, AQE use it to decide if EliminateLimit
         writeMetrics.incBytesWritten(bytesWritten)
+        // This metric is important, AQE use it to decide if EliminateLimit
         writeMetrics.incRecordsWritten(cb.numRows())
       }
     }
--- a/backends-velox/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
@@ -222,8 +222,8 @@ class ColumnarShuffleWriter[K, V](
         dep.metrics("shuffleWallTime").add(System.nanoTime() - startTime)
         dep.metrics("numInputRows").add(rows)
         dep.metrics("inputBatches").add(1)
-        // This metric is important, AQE use it to decide if EliminateLimit
         writeMetrics.incBytesWritten(bytesWritten)
+        // This metric is important, AQE use it to decide if EliminateLimit
         writeMetrics.incRecordsWritten(rows)
       }
       cb.close()


Yohahaha:
@kerwin-zk @marin-ma please help review this pr, thank you!

cpp/core/shuffle/PartitionWriter.h
```diff
@@ -48,7 +48,12 @@ class PartitionWriter : public Reclaimable {
 
   ~PartitionWriter() override = default;
 
-  virtual arrow::Status stop(ShuffleWriterMetrics* metrics) = 0;
+  void acceptMetrics(ShuffleWriterMetrics* metrics) {
```
marin-ma: Why do we need this change?

backends-velox/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -203,7 +205,7 @@ class ColumnarShuffleWriter[K, V](
         val columnarBatchHandle =
           ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, cb)
         val startTime = System.nanoTime()
-        shuffleWriterJniWrapper.write(
+        val curBytesWritten = shuffleWriterJniWrapper.write(
```
marin-ma: https://github.com/apache/incubator-gluten/blob/main/cpp/core/jni/JniWrapper.cc#L1000-L1002

Looks like the returned value from each call is the size of the input batch, but looks like here it's treated as an accumulated value. I wonder if this should be 

```
val bytesWritten = shuffleWriterJniWrapper.write(...)
writeMetrics.incBytesWritten(bytesWritten)

```

Yohahaha: shuffle write bytes does not equals to batch size, so before logic was wrong.

> I wonder if this should be

it depends on where we count incremental bytes, currently jni call returns total written bytes but spark only expose incremental interface.

cpp/core/shuffle/PartitionWriter.h
Yohahaha: use metrics member to count written bytes per batch.

backends-velox/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
marin-ma:
```
JNIEXPORT jlong JNICALL Java_org_apache_gluten_vectorized_ShuffleWriterJniWrapper_write( // NOLINT
    JNIEnv* env,
    jobject wrapper,
    jlong shuffleWriterHandle,
    jint numRows,
    jlong batchHandle,
    jlong memLimit) {
  JNI_METHOD_START
  auto shuffleWriter = ObjectStore::retrieve<ShuffleWriter>(shuffleWriterHandle);
  if (!shuffleWriter) {
    std::string errorMessage = "Invalid shuffle writer handle " + std::to_string(shuffleWriterHandle);
    throw GlutenException(errorMessage);
  }

  // The column batch maybe VeloxColumnBatch or ArrowCStructColumnarBatch(FallbackRangeShuffleWriter)
  auto batch = ObjectStore::retrieve<ColumnarBatch>(batchHandle);
  auto numBytes = batch->numBytes();
  arrowAssertOkOrThrow(shuffleWriter->write(batch, memLimit), "Native write: shuffle writer failed");
  return numBytes;
  JNI_METHOD_END(kInvalidObjectHandle)
}
```
Seems like it only returns one input batch size. Am I misunderstanding something?

Yohahaha: after this pr, `Java_org_apache_gluten_vectorized_ShuffleWriterJniWrapper_write` will return the bytes we written to shuffle service during this call which was compressed and accurate. 

`batch->numBytes` is not accurate right? and we did not use it in java side before, so this pr wants to use the correct return value to enhance observability.

marin-ma: @Yohahaha I don't see any modification to the `Java_org_apache_gluten_vectorized_ShuffleWriterJniWrapper_write` in this patch.

> `batch->numBytes` is not accurate right?

I'm also not aware of that. Could you elaborate in which cases it could be inaccurate? If the value is wrong we should fix this api as well.

Yohahaha: `Java_org_apache_gluten_vectorized_ShuffleWriterJniWrapper_write` I mean the result of this method was not used before, after this pr, its results can be used to update shuffle written bytes per batch

Yohahaha:
@marin-ma thank you for the detailed review, it's my mistake for lost some codes when rebasing.

marin-ma:
@Yohahaha Thanks for the clarification and updates. Can we simply add an API for the PartitionWriter to count the total  written bytes, and calculate then return the delta of the bytes written in the native ShuffleWriter?

Yohahaha:
> @Yohahaha Thanks for the clarification and updates. Can we simply add an API for the PartitionWriter to count the total written bytes, and calculate then return the delta of the bytes written in the native ShuffleWriter?

thank you for the suggestion, I will refactor it later.

Yohahaha:
@marin-ma please help review again, thank you!

backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala
```diff
@@ -77,7 +77,7 @@ class VeloxCelebornColumnarShuffleWriter[K, V](
         val columnarBatchHandle =
           ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, cb)
         val startTime = System.nanoTime()
-        shuffleWriterJniWrapper.write(
+        val bytesWritten = shuffleWriterJniWrapper.write(
```
marin-ma, line 80: I think `VeloxUniffleColumnarShuffleWriter` also needs to be updated. It's using the same native partition writer as celeborn.

cpp/core/shuffle/ShuffleWriter.cc
```diff
@@ -61,8 +61,8 @@ int32_t ShuffleWriter::numPartitions() const {
   return numPartitions_;
 }
 
-int64_t ShuffleWriter::totalBytesWritten() const {
-  return metrics_.totalBytesWritten;
+int64_t ShuffleWriter::bytesWritten() const {
```
marin-ma, line 64: Can we simply modify the signature of`ShuffleWriter::write` and  `ShuffleWriter::stop` to return the bytes evicted as `Result<int64_t>`? And seems like this value is not updated and used for local shuffle write, please add some explanation that only rss shuffle write returns the evicted bytes.

backends-velox/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -223,6 +223,7 @@ class ColumnarShuffleWriter[K, V](
         dep.metrics("numInputRows").add(rows)
         dep.metrics("inputBatches").add(1)
         // This metric is important, AQE use it to decide if EliminateLimit
+        writeMetrics.incBytesWritten(bytesWritten)
```
marin-ma, line 225: Please move the comment above this line to below.

cpp/core/shuffle/ShuffleWriter.cc
Yohahaha: > Can we simply modify the signature ofShuffleWriter::write and ShuffleWriter::stop to return the bytes evicted as Result<int64_t>

The write api can be easily modified, but the stop api is not, stop api collect and return whole shuffle metrics of task, I prefer do refactor work for api change in another pr and make current pr as simple as possible.

backends-velox/src-celeborn/main/scala/org/apache/spark/shuffle/VeloxCelebornColumnarShuffleWriter.scala
Yohahaha: Uniffle does not call incBytesWritten and the logic of metrics collecting differs with local and celeborn, it's better to let related developer to change it.

Yohahaha:
@marin-ma any more comments?

marin-ma:
@Yohahaha Thanks

PR #11084, opened 2025-11-14
[MINOR] Replace match plan with startsWith by case match

beliefer:

This PR proposes to replace match plan with `startsWith` by Scala case match.

## How was this patch tested?

GA tests.


beliefer:
@FelixYBW merged into branch main. Thank you!

Issue #11091, opened 2025-11-14
[VL] weekly test job failed


zhouyuan:
### Backend

VL (Velox)

### Bug description

https://github.com/apache/incubator-gluten/actions/runs/19213688366

on centos 7: should skip install conda
on ubuntu-20.04: not enough disk space
on openeuler: 

```
[ 94%] Linking CXX executable velox_plan_conversion_test
/usr/bin/ld: /__w/incubator-gluten/incubator-gluten/dev/../ep/build-velox/build/velox_ep/_build/release/lib/libvelox.a(DateTimeFunctionsRegistration.cpp.o): undefined reference to symbol '_ZN4absl12lts_2023080213base_internal12SpinLockWaitEPSt6atomicIjEiPKNS1_22SpinLockWaitTransitionENS1_14SchedulingModeE'
/usr/bin/ld: /usr/lib64/libabsl_spinlock_wait.so.2308.0.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

### Spark version

None

### Relevant logs

```bash

```

PR #11092, opened 2025-11-14
[GLUTEN-11091][VL] Fix weekly build jobs on CentOS/Ubuntu platform

zhouyuan:

This patch fixes the failed weekly job on Centos 7/8/9

## How was this patch tested?
pass GHA

Related issue: #11091

### DIFF
--- a/.github/workflows/velox_weekly.yml
+++ b/.github/workflows/velox_weekly.yml
@@ -117,7 +117,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ "ubuntu:22.04" ] # add "ubuntu:20.04" if needed, disk space is not enough in GHA
+        os: [ "ubuntu:22.04" ] # skip "ubuntu:20.04" due to not enough disk space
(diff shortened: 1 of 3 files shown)


ep/build-velox/src/setup-centos7.sh
```diff
@@ -224,7 +224,7 @@ function install_velox_deps {
   run_and_time install_folly
   run_and_time install_protobuf
   run_and_time install_gtest
-  run_and_time install_conda
+  # run_and_time install_conda
```
zhztheplayer: Is this intentional?

FelixYBW: Does Gluten build need conda? Is it for pyvelox?
If we don't need we may remove it from all of our setup script. we also need to clean up our setup script.

zhouyuan:
The failures are not related

ep/build-velox/src/setup-centos7.sh
zhouyuan: removed

Issue #11103, opened 2025-11-17
[VL] Flaky Celeborn tests


zhouyuan:
### Backend

VL (Velox)

### Bug description
https://github.com/apache/incubator-gluten/actions/runs/19399179485/job/55503865154?pr=11095

```
There are failed queries.

Query q1 failed by error: org.apache.spark.SparkException: Job aborted due to stage failure: Task 2 in stage 28.0 failed 1 times, most recent failure: Lost task 2.0 in stage 28.0 (TID 34) (ea04227c6a83 executor driver): org.apache.gluten.exception.GlutenException: Error during calling Java code from native code: org.apache.celeborn.common.exception.CelebornIOException: Register shuffle failed for shuffle 1, reason: RESERVE_SLOTS_FAILED
	at org.apache.celeborn.client.ShuffleClientImpl.registerShuffleInternal(ShuffleClientImpl.java:746)
	at org.apache.celeborn.client.ShuffleClientImpl.registerShuffle(ShuffleClientImpl.java:547)
	at org.apache.celeborn.client.ShuffleClientImpl.lambda$getPartitionLocation$4(ShuffleClientImpl.java:609)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
	at org.apache.celeborn.common.util.JavaUtils$ConcurrentHashMapForJDK8.computeIfAbsent(JavaUtils.java:492)
	at org.apache.celeborn.client.ShuffleClientImpl.getPartitionLocation(ShuffleClientImpl.java:605)
	at org.apache.celeborn.client.ShuffleClientImpl.pushOrMergeData(ShuffleClientImpl.java:970)
	at org.apache.celeborn.client.ShuffleClientImpl.mergeData(ShuffleClientImpl.java:1362)
	at org.apache.spark.shuffle.CelebornPartitionPusher.pushPartitionData(CelebornPartitionPusher.scala:61)
	at org.apache.gluten.vectorized.ShuffleWriterJniWrapper.stop(Native Method)
	at org.apache.spark.shuffle.VeloxCelebornColumnarShuffleWriter.internalWrite(VeloxCelebornColumnarShuffleWriter.scala:100)
	at org.apache.spark.shuffle.CelebornColumnarShuffleWriter.write(CelebornColumnarShuffleWriter.scala:113)
	at org.apache.spark.shuffle.ShuffleWriteProcessor.write(ShuffleWriteProcessor.scala:59)
[...]

PR #11105, opened 2025-11-17
[GLUTEN-11103][VL] Optimize Celeborn tests

kerwin-zk:
#11103 

## How was this patch tested?
CI

### DIFF
--- a/.github/workflows/velox_backend_x86.yml
+++ b/.github/workflows/velox_backend_x86.yml
@@ -555,7 +555,8 @@ jobs:
       fail-fast: false
       matrix:
         spark: [ "spark-3.2" ]
-        celeborn: [ "celeborn-0.6.1", "celeborn-0.5.4", "celeborn-0.4.3"]
+        celeborn: [ "celeborn-0.6.1", "celeborn-0.5.4"]
+        writer: [ "sort", "hash"]
     runs-on: ubuntu-22.04
     container: apache/gluten:centos-8-jdk8
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/velox_backend_x86.yml
```diff
@@ -555,7 +555,8 @@ jobs:
       fail-fast: false
       matrix:
         spark: [ "spark-3.2" ]
-        celeborn: [ "celeborn-0.6.1", "celeborn-0.5.4", "celeborn-0.4.3"]
+        celeborn: [ "celeborn-0.6.1", "celeborn-0.5.4"]
```
philo-he, line 558: @kerwin-zk, will 0.4.3 be dropped permanently? If so, could you clean up some related code in a separate PR?

PHILO-HE (approved):
Thanks for the fix.

.github/workflows/velox_backend_x86.yml
kerwin-zk: @PHILO-HE  The current main version of Celeborn in use is 0.5, so version 0.4 no longer needs to be tested. I'll submit a new PR to clean this up.

zhouyuan:
@kerwin-zk thanks for the fix. just want to confirm, is this due to mixing the sort/hash tests in one instance?

kerwin-zk:
> @kerwin-zk thanks for the fix. just want to confirm, is this due to mixing the sort/hash tests in one instance?

@zhouyuan Currently, log analysis indicates that the primary cause is insufficient disk space, which led to Celeborn worker failures.

PR #11118, opened 2025-11-19
[VL] Suppress "uninitialized" warnings treated as errors in GCC 13 builds

philo-he:

- ~GCC 13 will fail the compilation due to the unrecognized flag `-Wno-unknown-warning-option` is used. This flag is for Clang to allow the use of Clang unrecognized flag `-Wno-error=stringop-overflow`. We should set the latter by checking if compiler accepts it. And then, the former can be removed.~
- GCC 13 has stricter check for using uninitialized variables. Currently, these errors reported by compiler are related to some folly headers. We set two flags to suppress such warnings treated as errors.
- Due to the above uninitialized variable errors, the compiler also reported the error below, which is easily misleading: 
`cc1plus: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics`. 
This `-Wno-unknown-warning-option` flag is not belonging to GCC. But when facing such unrecognized flags in `-Wno-xxx` pattern, the GCC compiler normally ignore them. Only when there are real errors, the compiler will report error about this unrecognized flag in case the real warning or error is somehow related to the use of an unknown flag.

## How was this patch tested?
CI.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -100,6 +100,7 @@ set(KNOWN_WARNINGS
        -Wno-error=unused-variable \
        -Wno-error=uninitialized \
        -Wno-error=maybe-uninitialized \
+       -Wno-unknown-warning-option \
        -Wno-strict-aliasing \
        -Wno-ignored-qualifiers \
        -Wno-deprecated-declarations \
--- a/ep/build-velox/src/build-velox.sh
+++ b/ep/build-velox/src/build-velox.sh
@@ -98,16 +98,11 @@ for arg in "$@"; do
 done
 
 function compile {
-  CXX_FLAGS='-Wno-error=cpp -Wno-missing-field-initializers -Wno-error=uninitialized'
-  # Set compiler-specific flags.
-  local compiler="${CXX:-c++}"
-  local compiler_specific_flags="-Wno-error=stringop-overflow -Wno-unknown-warning-option"
-  for flag in $compiler_specific_flags; do
-    # Validate if a flag is accepted by the compiler.
-    if $compiler $flag -E - </dev/null >/dev/null 2>&1; then
-      CXX_FLAGS="$CXX_FLAGS $flag"
-    fi
-  done
+  # -Wno-unknown-warning-option is a Clang-originated flag. GCC ignores unrecognized -Wno- flags to
+  # maintain compatibility, but it prints a diagnostic note about the unknown flag if a true warning
+  # or error occurs.
+  CXX_FLAGS='-Wno-error=stringop-overflow -Wno-error=cpp -Wno-missing-field-initializers \
+    -Wno-error=uninitialized -Wno-unknown-warning-option'
 
   COMPILE_OPTION="-DCMAKE_CXX_FLAGS=\"$CXX_FLAGS\" -DVELOX_ENABLE_PARQUET=ON -DVELOX_BUILD_TESTING=OFF \
       -DVELOX_MONO_LIBRARY=ON -DVELOX_BUILD_RUNNER=OFF -DVELOX_SIMDJSON_SKIPUTF8VALIDATION=ON \


philo-he:
@zml1206, could you take a look?

philo-he:
@zml1206, thanks for the feedback. It seems `-Werror=stringop-overflow` was set by dependency lib's cmake build. Then, we have to keep using the `-Wno-unknown-warning-option` flag. Could you try again with the updated patch?

zml1206:
Clang doesn't recognize `uninitialized`, and it seems that `-Wno-unknown-warning-option` also needs to be added to the gluten cpp build. @PHILO-HE 
```
[  3%] Built target velox_jni_proto
[  3%] Built target jni_proto
[  6%] Building CXX object core/CMakeFiles/gluten.dir/proto/substrait/extended_expression.pb.cc.o
[  6%] Building CXX object core/CMakeFiles/gluten.dir/proto/substrait/function.pb.cc.o
[  6%] Building CXX object core/CMakeFiles/gluten.dir/proto/substrait/algebra.pb.cc.o
[  8%] Building CXX object core/CMakeFiles/gluten.dir/proto/substrait/capabilities.pb.cc.o
[  8%] Building CXX object core/CMakeFiles/gluten.dir/proto/substrait/parameterized_types.pb.cc.o
[  9%] Building CXX object core/CMakeFiles/gluten.dir/proto/substrait/extensions/extensions.pb.cc.o
error: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option]
make[2]: *** [core/CMakeFiles/gluten.dir/proto/substrait/capabilities.pb.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
error: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option]
make[2]: *** [core/CMakeFiles/gluten.dir/proto/substrait/extensions/extensions.pb.cc.o] Error 1
error: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option]
make[2]: *** [core/CMakeFiles/gluten.dir/proto/substrait/parameterized_types.pb.cc.o] Error 1
error: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option]
make[2]: *** [core/CMakeFiles/gluten.dir/proto/substrait/function.pb.cc.o] Error 1
error: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option]
make[2]: *** [core/CMakeFiles/gluten.dir/proto/substrait/extended_expression.pb.cc.o] Error 1
error: unknown warning option '-Werror=maybe-uninitialized'; did you mean '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option]
make[2]: *** [core/CMakeFiles/gluten.dir/proto/substrait/algebra.pb.cc.o] Error 1
make[1]: *** [core/CMakeFiles/gluten.dir/all] Error 2
make: *** [all] Error 2
```

cpp/CMakeLists.txt
```diff
@@ -100,6 +100,7 @@ set(KNOWN_WARNINGS
        -Wno-error=unused-variable \
        -Wno-error=uninitialized \
        -Wno-error=maybe-uninitialized \
+       -Wno-unknown-warning-option \
```
zml1206, line 103: Will adding it directly like this cause any problems with GCC 13?

philo-he:
@zml1206, yes, I just added it in cpp/CMakeLists.txt. I also reverted some changes according to the new finding. See the explanations in the PR description. Thanks.

cpp/CMakeLists.txt
philo-he: No, I have verified it with GCC 13 in my local environment.

zml1206: So `-Wno-unknown-warning-option` doesn't affect GCC 13, it just doesn't work, right?

philo-he: Yes. Only when there are some other errors during the build, GCC reported that this flag is not recognized, which is printed to help developer diagnose issues. In normal situation, GCC just ignores it for maintaining compatibility.

zml1206: Thanks for your explanation. I'll test in Mac again.

zml1206 (approved):
Successfully built on Mac.

PR #11121, opened 2025-11-19
[VL] Clean up vcpkg overlay ports

philo-he:

1. Some of overlay ports were added because the previous baseline doesn't contain the required library versions. As we have upgraded the baseline, some of these ports can be removed.
2. Some dependencies like gsasl are used for libhdfs3.so (native C++ support), not for libhdfs.so (JNI based API support) currently in use. We can remove them.
3. Upgrade simdjson to 4.1.0, aligning with Velox's requirement when finding the package.

## How was this patch tested?
CI.


philo-he:

PR #11122, opened 2025-11-19
[INFRA][VL] Fix build failures for docker images

philo-he:

```log
 1381.8 /opt/gluten/dev/vcpkg/triplets/x64-linux-avx.cmake: info: loaded overlay triplet from here
#6 1381.8 Downloading https://github.com/protocolbuffers/protobuf/archive/v3.21.8.tar.gz -> protocolbuffers-protobuf-v3.21.8.tar.gz
#6 1515.5 error: curl: (7) Failed to connect to github.com port 443: Connection timed out
#6 1515.5 note: If you are using a proxy, please ensure your proxy settings are correct.
#6 1515.5 Possible causes are:
#6 1515.5 1. You are actually using an HTTP proxy, but setting HTTPS_PROXY variable to `https//address:port`.
#6 1515.5 This is not correct, because `https://` prefix claims the proxy is an HTTPS proxy, while your proxy (v2ray, shadowsocksr, etc...) is an HTTP proxy.
#6 1515.5 Try setting `http://address:port` to both HTTP_PROXY and HTTPS_PROXY instead.
#6 1515.5 2. If you are using Windows, vcpkg will automatically use your Windows IE Proxy Settings set by your proxy software. See: https://github.com/microsoft/vcpkg-tool/pull/77
#6 1515.5 The value set by your proxy might be wrong, or have same `https://` prefix issue.
#6 1515.5 3. Your proxy's remote server is out of service.
#6 1515.5 If you believe this is not a temporary download server failure and vcpkg needs to be changed to download this file from a different location, please submit an issue to https://github.com/Microsoft/vcpkg/issues
#6 1515.5 CMake Error at scripts/cmake/vcpkg_download_distfile.cmake:136 (message):
#6 1515.5   Download failed, halting portfile.        
```

### DIFF
--- a/dev/docker/Dockerfile.centos8-static-build
+++ b/dev/docker/Dockerfile.centos8-static-build
@@ -44,6 +44,7 @@ RUN set -ex; \
     fi; \
     source /opt/rh/gcc-toolset-11/enable; \
     cd /opt/gluten; \
+    export SSL_VERIFY=false \
     bash ./dev/builddeps-veloxbe.sh --enable_vcpkg=ON --build_tests=ON --enable_s3=ON --enable_gcs=ON \
                                     --enable_hdfs=ON --enable_abfs=ON  build_arrow; \
     mvn dependency:go-offline -Pbackends-velox -Piceberg -Pdelta -Pspark-3.5 -DskipTests; \


dev/docker/Dockerfile.centos8-static-build
```diff
@@ -44,6 +44,7 @@ RUN set -ex; \
     fi; \
     source /opt/rh/gcc-toolset-11/enable; \
     cd /opt/gluten; \
+    export SSL_VERIFY=false \
```
philo-he, line 47: Not security, just a quick fix.

PR #11123, opened 2025-11-19
[VL] RSS client should push complete rows

wecharyu:

https://github.com/apache/incubator-gluten/pull/11059 will push data to rss once there are available inMemoryPayload, but for large rows the data may be split into multiple payloads and push multi times.

If there is another worker push data to the same partition, the split row are broken, so we should ensure only the complete rows can be push to rss.

## How was this patch tested?


wecharyu:
cc: @boneanxs

marin-ma:
We should avoid future changes to break this logic again. Could you add unit test for this fix?

kerwin-zk:
@wecharyu @boneanxs Please add a unit test.

wecharyu:
@marin-ma @kerwin-zk Pls help take a look.

PR #11127, opened 2025-11-19
[GLUTEN-11133][VL] Refactor batch serialization API to defer the buffer copy from C++ code to Java code

zhztheplayer:
In preparation for moving the memory allocation triggered by [this code](https://github.com/apache/incubator-gluten/blob/aadc68b7cdfd0de5aff4638bcd16a802aece9901/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L673-L677) from heap to off-heap.

The patch defers the off-heap-to-on-heap buffer copy during columnar batch serialization from [here](https://github.com/apache/incubator-gluten/blob/0f8eb0fc694e453e9556b4d826eacd43a45f70a1/cpp/core/jni/JniWrapper.cc#L1156-L1161) to [here](https://github.com/zhztheplayer/gluten/blob/8565de81f678aaece50658591b471556502d959e/backends-velox/src/main/scala/org/apache/spark/sql/execution/BroadcastUtils.scala#L186).

In the next PR(s), we'll refactor the Java / Scala code to remove the copy and store the off-heap binary data directly when `spark.gluten.velox.offHeapBroadcastBuildRelation.enabled=true`.

Related issue: #11133

### DIFF
--- a/cpp/core/operators/serializer/ColumnarBatchSerializer.h
+++ b/cpp/core/operators/serializer/ColumnarBatchSerializer.h
@@ -29,9 +29,9 @@ class ColumnarBatchSerializer {
 
   virtual ~ColumnarBatchSerializer() = default;
 
-  virtual void addForSerialization(const std::shared_ptr<ColumnarBatch>& batch) = 0;
+  virtual void append(const std::shared_ptr<ColumnarBatch>& batch) = 0;
 
-  virtual int64_t serializedSize() = 0;
+  virtual int64_t maxSerializedSize() = 0;
 
   virtual void serializeTo(uint8_t* address, int64_t size) = 0;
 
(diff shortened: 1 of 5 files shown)


zhztheplayer:
cc @zjuwangg @jinchengchenghh Thanks.

cpp/core/operators/serializer/ColumnarBatchSerializer.h
```diff
@@ -29,8 +29,11 @@ class ColumnarBatchSerializer {
 
   virtual ~ColumnarBatchSerializer() = default;
 
-  virtual std::shared_ptr<arrow::Buffer> serializeColumnarBatches(
-      const std::vector<std::shared_ptr<ColumnarBatch>>& batches) = 0;
+  virtual void addForSerialization(const std::shared_ptr<ColumnarBatch>& batch) = 0;
```
jinchengchenghh: This function calls Velox function append, so I would prefer function name `append`

cpp/core/operators/serializer/ColumnarBatchSerializer.h
```diff
@@ -29,8 +29,11 @@ class ColumnarBatchSerializer {
...
+
+  virtual int64_t serializedSize() = 0;
```
jinchengchenghh: same maxSerializedSize

zhztheplayer:
@jinchengchenghh Let me know if you have further comments. Thanks.

zhztheplayer:
Thanks for reviewing @zjuwangg @jinchengchenghh

Issue #10660, opened 2025-09-09
[VL] add configurations for hash map build optimizations


zhouyuan:
### Description

https://github.com/facebookincubator/velox/pull/7066 

### Gluten version

None

PR #11134, opened 2025-11-20
[GLUTEN-10660][VL] Fix passing Velox session configurations on hash build optimizations

zhouyuan:
some important configurations on Velox hashmap dedup are not passed to Velox backend

## How was this patch tested?
locally verified

Related issue: #10660

### DIFF
--- a/.github/workflows/velox_backend_x86.yml
+++ b/.github/workflows/velox_backend_x86.yml
@@ -561,7 +561,7 @@ jobs:
         celeborn: [ "celeborn-0.6.1", "celeborn-0.5.4"]
         writer: ["sort", "hash"]
     runs-on: ubuntu-22.04
-    container: apache/gluten:centos-8-jdk8
+    container: centos:8
     steps:
       - uses: actions/checkout@v2
       - name: Download All Native Artifacts
@@ -574,6 +574,17 @@ jobs:
         with:
           name: arrow-jars-centos-7-${{github.sha}}
           path: /root/.m2/repository/org/apache/arrow/
+      - name: Update mirror list
+        run: |
+          sed -i -e "s|mirrorlist=|#mirrorlist=|g" /etc/yum.repos.d/CentOS-* || true
+          sed -i -e "s|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g" /etc/yum.repos.d/CentOS-* || true
+      - name: Setup java and maven
+        run: |
+          yum update -y && yum install -y java-1.8.0-openjdk-devel wget
+          $SETUP install_maven
+      - name: Set environment variables
+        run: |
+          echo "JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk" >> $GITHUB_ENV
       - name: Build for Spark ${{ matrix.spark }}
         run: |
           cd $GITHUB_WORKSPACE/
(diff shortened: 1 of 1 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -551,6 +551,11 @@ object GlutenConfig extends ConfigRegistry {
           nativeConfMap
             .put(SQLConf.LEGACY_TIME_PARSER_POLICY.key, v.toUpperCase(Locale.ROOT)))
 
+    // put in all gluten velox configs
+    conf
+      .filter(_._1.startsWith(s"spark.gluten.$backendName"))
+      .foreach(entry => nativeConfMap.put(entry._1, entry._2))
+
     // Backend's dynamic session conf only.
     val confPrefix = prefixOf(backendName)
```
wForget, line 577: Could you change this method name from `prefixOf` to `sessionPrefixOf`?

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -617,7 +617,7 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
     configs[velox::core::QueryConfig::kAbandonBuildNoDupHashMinRows] =
         std::to_string(veloxCfg_->get<int32_t>(kAbandonBuildNoDupHashMinRows, 100000));
     configs[velox::core::QueryConfig::kAbandonBuildNoDupHashMinPct] =
-        std::to_string(veloxCfg_->get<int32_t>(kAbandonBuildNoDupHashMinPct, 0));
+        std::to_string(static_cast<int32_t>(veloxCfg_->get<double>(kAbandonBuildNoDupHashMinPct, 0)));
```
wForget: Why not just change the type of original configuration?

https://github.com/apache/incubator-gluten/blob/3ff9254d9cf72e1e0fa58942a4b3d2fd8dc7cb54/backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala#L549

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
zhouyuan: done

cpp/velox/compute/WholeStageResultIterator.cc
zhouyuan: it's mostly due to the config name has a `percentile` so we change that to double based config when enable this feature. I have modified to use int based and update the comments

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
zhouyuan: The code should be work for the hash build dedup feature. However for now Gluten actually does not have a clear boundary on session level vs. static config, I guess we may need some refactor on this part in future work.

beliefer (commented):
It looks good to me if the session configurations used by Velox.

beliefer:
The fail test seems not related to this PR. It is failed in other PR too.

PHILO-HE (approved):
Thanks. Assume the CI failure is not related.

zhouyuan:
@wForget @beliefer @PHILO-HE thanks for the review.

Issue #11088, opened 2025-11-14
[VL] Track on Spark-4.0 failed unit tests


zhouyuan:
Please leave a comment and mark on the TODO tests in this issue if you want to contribute

### Backend

VL (Velox)

### Bug description

After #10725 the general framework of unit tests are merged. We still have failed unit tests left here: 

### Velox backend
| Sutie | Owner /  PR |
|-----------------|-----------------|
| NativeBenchmarkPlanGenerator.scala |✅https://github.com/apache/incubator-gluten/pull/11144 @marin-ma |
| MiscOperatorSuite |✅https://github.com/apache/incubator-gluten/pull/11147 @marin-ma |
| ArrowEvalPythonExecSuite | ✅https://github.com/apache/incubator-gluten/pull/11288 @zhztheplayer |
| ArithmeticAnsiValidateSuite |✅https://github.com/apache/incubator-gluten/pull/11150 @marin-ma |
| JsonFunctionsValidateSuite | ✅https://github.com/apache/incubator-gluten/pull/11294 @baibaichen |
| DateFunctionsValidateSuite | ✅https://github.com/apache/incubator-gluten/pull/11171 @zml1206 |
| ScalarFunctionsValidateSuite | ✅ https://github.com/apache/incubator-gluten/pull/11170 https://github.com/apache/incubator-gluten/pull/11180 @marin-ma |

### Gluten UT
| Sutie | Owner /  PR |
|-----------------|-----------------|
| GlutenSQLQueryTestSuite | https://github.com/apache/incubator-gluten/pull/11136 @zhouyuan |
| GlutenKeyGroupedPartitioningSuite |  ✅ https://github.com/apache/incubator-gluten/pull/11184 @jinchengchenghh |
| GlutenQueryExecutionErrorsSuite |  ✅ https://github.com/apache/incubator-gluten/pull/11185 @marin-ma |
| GlutenCastSuite | ✅ https://github.com/apache/incubator-gluten/pull/11181 @marin-ma |
| GlutenTryCastSuite | ✅ https://github.com/apache/incubator-gluten/pull/11181 @marin-ma |
| GlutenJsonFunctionsSuite| ✅https://github.com/apache/incubator-gluten/pull/11276 @baibaichen |
| GlutenCSVv1Suite | https://github.com/apache/incubator-gluten/pull/11190 @jinchengchenghh|
[...]

PR #11136, opened 2025-11-20
[GLUTEN-11088][VL] Adding SQLQueryTestSuite for Spark-4.0

zhouyuan:
There's a refactor on SQLQueryTestSuite in Spark-4.0, which pulls out some traits and added more sql tests.
This patch ports the SQLQueryTestSuite for Gluten.

```
  // There is a limitation in Spark-4.0 on ANSI gap and exception message gap.
  // we disabled these sql tests temporarily.
  // limit.sql, postgreSQL/limit.sql, subquery/in-subquery/in-limit.sql,
  // subquery/in-subquery/in-null-semantics.sql
  // try_arithmetic.sql, try_element_at.sql, typeCoercion/native/stringCastAndExpressions.sql,
  // window.sql
```

## How was this patch tested?
Pass GHA

Related issue: #11088

### DIFF
--- a/gluten-ut/spark40/src/test/resources/backends-velox/sql-tests/results/nonansi/string-functions.sql.out
+++ b/gluten-ut/spark40/src/test/resources/backends-velox/sql-tests/results/nonansi/string-functions.sql.out
@@ -1821,7 +1821,7 @@ struct<to_binary(12345, hex):binary>
 -- !query
 select to_binary('GG')
 -- !query schema
-struct<>
+struct<to_binary(GG):binary>
 -- !query output
 org.apache.spark.SparkIllegalArgumentException
 {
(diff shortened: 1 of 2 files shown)


gluten-ut/spark40/src/test/resources/backends-velox/sql-tests/inputs/binary.sql
```diff
@@ -0,0 +1,8 @@
+--SET spark.sql.binaryOutputStyle=UTF-8
+
+SELECT X'';
```
jinchengchenghh, line 3: Could we not copy so many files from Spark, can we use spark.test.home?

zhouyuan:
The SQL query test is specially handled: we need to modify the sql outputs to align with spark for some cases.
In the same time there are also updates from spark4.0 so we need to merge the modifications

________________________________

Issue #10892, opened 2025-10-15
[VL] VeloxBatchResizer may cause OOM when there exists type of variable length


jiangjiangtian:
### Backend

VL (Velox)

### Bug description

There is a job failure because of oom in our test and its memory usage statistics is as follows:

<img width="2778" height="1164" alt="Image" src="https://github.com/user-attachments/assets/93ed4fc0-080a-496b-bde4-a8b601346482" />

We can see that `VeloxBatchResizer` uses most of the memory. I find that most of the columns of the table is variable length, like `VARCHAR, ARRAR, MAP` etc.

I find there is a method called `maybeReserve` in velox's [MemoryPool](https://github.com/facebookincubator/velox/blob/main/velox/common/memory/MemoryPool.cpp#L800). Maybe we can call this method in `VeloxBatchResizer` before calling the `append` method to try to allocate memory. If the process fails, then we don't append the batch. In this way, we need to do some code modification to make the memory allocation not to throw `OutOfMemoryException`.

### Spark version

None

### Relevant logs

```bash

```

jiangjiangtian:
Besides, I find that the OOM is caused by the memory allocation for shuffle write.

zhztheplayer:
I am curious why the resizer occupies this much of memory. How complex the batch schema is?

zhztheplayer:
Maybe we need to introduce the estimate size in bytes as another factor for batch resizing.

jiangjiangtian:
> I am curious why the resizer occupies this much of memory. How complex the batch schema is?

I try to output some statistics about one of the row vector and its children vectors:
```
estimated byte size 1998751236, numRows: 904
encoding: FLAT, isLazy: 0, type: INTEGER
child 0 estimated byte size 4000
encoding: LAZY, isLazy: 1, type: VARCHAR
isLoaded: 1, DICTIONARY
child 1 estimated byte size 57856
encoding: LAZY, isLazy: 1, type: VARCHAR
isLoaded: 1, DICTIONARY
child 2 estimated byte size 28928
encoding: DICTIONARY, isLazy: 0, type: VARCHAR
child 3 estimated byte size 45858
encoding: DICTIONARY, isLazy: 0, type: VARCHAR
child 4 estimated byte size 105160
encoding: DICTIONARY, isLazy: 0, type: VARCHAR
child 5 estimated byte size 31520
encoding: LAZY, isLazy: 1, type: VARCHAR
isLoaded: 1, DICTIONARY
[...]

PR #11140, opened 2025-11-21
[GLUTEN-10892][VL] Use `veloxPreferredBatchBytes` to control the max size of memory of batches combined

jiangjiangtian:
Now the operator `VeloxBatchResizer` can use as many memory as possible, which can cause OOM. So this PR use the existing config `spark.gluten.sql.columnar.backend.velox.preferredBatchBytes` to control it.
Closes https://github.com/apache/incubator-gluten/issues/10892

## How was this patch tested?
Inner ETL jobs.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
@@ -73,6 +73,7 @@ abstract class VeloxTPCHSuite extends VeloxTPCHTableSupport {
       // for unexpected blank
       .replaceAll("Scan parquet ", "Scan parquet")
       // Spark QueryStageExec will take it's id as argument, replace it with X
+      .replaceAll("Arguments: [0-9]+, [0-9]+, [0-9]+", "Arguments: X, X, X")
       .replaceAll("Arguments: [0-9]+, [0-9]+", "Arguments: X, X")
       .replaceAll("Arguments: [0-9]+", "Arguments: X")
       // mask PullOutPostProject and PullOutPreProject id
--- a/cpp/velox/utils/VeloxBatchResizer.cc
+++ b/cpp/velox/utils/VeloxBatchResizer.cc
@@ -77,27 +77,16 @@ std::shared_ptr<ColumnarBatch> VeloxBatchResizer::next() {
     return nullptr;
   }
 
-  if (cb->numRows() < minOutputBatchSize_) {
+  uint64_t numBytes = cb->numBytes();
+  if (cb->numRows() < minOutputBatchSize_ && numBytes <= preferredBatchBytes_) {
     auto vb = VeloxColumnarBatch::from(pool_, cb);
     auto rv = vb->getRowVector();
-    auto vector = std::static_pointer_cast<facebook::velox::BaseVector>(rv);
-    uint64_t numBytes = cb->numBytes();
-    if (numBytes > preferredBatchBytes_) {
-      // Input batch is too large. Just return it as is.
-      return cb;
-    }
     auto buffer = facebook::velox::RowVector::createEmpty(rv->type(), pool_);
     buffer->append(rv.get());
 
-    // Call reset manully to potentially release memory
-    vector.reset();
-    rv.reset();
-    vb.reset();
-    cb.reset();
     for (cb = in_->next(); cb != nullptr; cb = in_->next()) {
       vb = VeloxColumnarBatch::from(pool_, cb);
       rv = vb->getRowVector();
-      vector = std::static_pointer_cast<facebook::velox::BaseVector>(rv);
       uint64_t addedBytes = cb->numBytes();
       if (buffer->size() + rv->size() > maxOutputBatchSize_ ||
           numBytes + addedBytes > static_cast<uint64_t>(preferredBatchBytes_)) {
@@ -111,7 +100,7 @@ std::shared_ptr<ColumnarBatch> VeloxBatchResizer::next() {
         // Buffer is full.
         break;
       }
-      vector.reset();
+      // Call reset manully to potentially release memory
       rv.reset();
       vb.reset();
       cb.reset();
--- a/cpp/velox/memory/VeloxColumnarBatch.cc
+++ b/cpp/velox/memory/VeloxColumnarBatch.cc
@@ -79,7 +79,7 @@ std::shared_ptr<ArrowArray> VeloxColumnarBatch::exportArrowArray() {
 }
 
 int64_t VeloxColumnarBatch::numBytes() {
-  ensureFlattened();
+  BaseVector::loadedVectorShared(rowVector_);
   return rowVector_->estimateFlatSize();
 }
 


wForget:
Could you add a new configuration for `memoryThreshold` and allow disabling it? I'm concerned that estimating batch size might cause performance overhead.

jiangjiangtian:
> Could you add a new configuration for `memoryThreshold` and allow disabling it? I'm concerned that estimating batch size might cause performance overhead.

Sure. Thanks for your advice.
Actually, I have tested 300 jobs and the overall execution time decrease by about 1.1%.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -680,6 +682,15 @@ object VeloxConfig extends ConfigRegistry {
       .bytesConf(ByteUnit.BYTE)
       .createWithDefaultString("10MB")
 
+  val ENABLE_LIMIT_BATCH_RESIZER = {
```
jinchengchenghh: This can be a default behavior, please remove this config, all the operator should respect `preferredBatchBytes`

backends-velox/src/test/resources/tpch-approved-plan/v1-bhj-ras/spark32/1.txt
```diff
@@ -60,7 +60,7 @@ Arguments: false
 
 (7) VeloxResizeBatches
 Input [18]: [hash_partition_key#X, l_returnflag#X, l_linestatus#X, sum#X, isEmpty#X, sum#X, isEmpty#X, sum#X, isEmpty#X, sum#X, isEmpty#X, sum#X, count#X, sum#X, count#X, sum#X, count#X, count#X]
-Arguments: X, X
+Arguments: X, X, 10485760
```
jinchengchenghh: Please change it to X, we don't need to update it if we change the default value in the future

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -40,19 +40,20 @@ case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
...
 
     val range = VeloxConfig.get.veloxResizeBatchesShuffleInputOutputRange
+    val memoryThreshold = VeloxConfig.get.veloxPreferredBatchBytes
     plan.transformUp {
       case shuffle: ColumnarShuffleExchangeExec
           if resizeBatchesShuffleInputEnabled &&
             shuffle.shuffleWriterType.requiresResizingShuffleInput =>
         val appendBatches =
-          VeloxResizeBatchesExec(shuffle.child, range.min, range.max)
+          VeloxResizeBatchesExec(shuffle.child, range.min, range.max, memoryThreshold)
```
jinchengchenghh: Please use the original config name preferredBatchBytes instead of a new name

cpp/velox/utils/VeloxBatchResizer.cc
```diff
@@ -78,22 +80,43 @@ std::shared_ptr<ColumnarBatch> VeloxBatchResizer::next() {
   if (cb->numRows() < minOutputBatchSize_) {
```
jinchengchenghh: Please use cb->numBytes() to estimate the batch size, and update the numBytes(), get the numBytes does not need to flat the vector, just load the vector

cpp/velox/utils/VeloxBatchResizer.cc
```diff
@@ -78,22 +80,41 @@ std::shared_ptr<ColumnarBatch> VeloxBatchResizer::next() {
...
     auto vb = VeloxColumnarBatch::from(pool_, cb);
     auto rv = vb->getRowVector();
+    auto vector = std::static_pointer_cast<facebook::velox::BaseVector>(rv);
```
jinchengchenghh: I don't see the vector usage

cpp/velox/utils/VeloxBatchResizer.cc
```diff
@@ -78,22 +80,41 @@ std::shared_ptr<ColumnarBatch> VeloxBatchResizer::next() {
...
+    }
     auto buffer = facebook::velox::RowVector::createEmpty(rv->type(), pool_);
     buffer->append(rv.get());
 
-    for (auto nextCb = in_->next(); nextCb != nullptr; nextCb = in_->next()) {
-      auto nextVb = VeloxColumnarBatch::from(pool_, nextCb);
-      auto nextRv = nextVb->getRowVector();
-      if (buffer->size() + nextRv->size() > maxOutputBatchSize_) {
+    // Call reset manully to potentially release memory
+    vector.reset();
```
jinchengchenghh: L114 is enough, you reuse the variables in following for loop or release in the end

cpp/velox/utils/VeloxBatchResizer.cc
```diff
@@ -78,22 +80,41 @@ std::shared_ptr<ColumnarBatch> VeloxBatchResizer::next() {
   if (cb->numRows() < minOutputBatchSize_) {
     auto vb = VeloxColumnarBatch::from(pool_, cb);
     auto rv = vb->getRowVector();
+    auto vector = std::static_pointer_cast<facebook::velox::BaseVector>(rv);
+    uint64_t numBytes = cb->numBytes();
```
jinchengchenghh: Update here to only load the lazy vector, we don't need to flat dictionary vector   https://github.com/apache/incubator-gluten/blob/7d491d5b373f77a3aa9a1bad14c34acb8987cfb5/cpp/velox/memory/VeloxColumnarBatch.cc#L82

jiangjiangtian: I get it. Thanks!

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
```diff
@@ -73,6 +73,7 @@ abstract class VeloxTPCHSuite extends VeloxTPCHTableSupport {
       // for unexpected blank
       .replaceAll("Scan parquet ", "Scan parquet")
       // Spark QueryStageExec will take it's id as argument, replace it with X
+      .replaceAll("Arguments: [0-9]+, [0-9]+, [0-9]+", "Arguments: X, X")
```
jinchengchenghh: X, X, X

cpp/velox/utils/VeloxBatchResizer.cc
jinchengchenghh: And also remove following rv batch.reset

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxTPCHSuite.scala
jiangjiangtian: Thanks

cpp/velox/utils/VeloxBatchResizer.cc
jiangjiangtian: I remove the four lines of code here.

cpp/velox/utils/VeloxBatchResizer.cc
```diff
@@ -78,22 +80,34 @@ std::shared_ptr<ColumnarBatch> VeloxBatchResizer::next() {
   if (cb->numRows() < minOutputBatchSize_) {
     auto vb = VeloxColumnarBatch::from(pool_, cb);
     auto rv = vb->getRowVector();
+    uint64_t numBytes = cb->numBytes();
```
jinchengchenghh: Move the check with L80 `if (cb->numRows() < minOutputBatchSize_ && numBytes <= preferredBatchBytes_)`

jiangjiangtian: Done. Thanks

jinchengchenghh:
Thanks!

PR #11143, opened 2025-11-21
[GLUTEN-10926][VL] Update Spark mirror to speed up downloading in CI docker image

zhouyuan:

The mirror of "https://www.apache.org/dyn/closer.lua" is very slow and the docker image build timeout
This patch adds a new mirror to download Spark binary/source 

## How was this patch tested?
locally verified

Related issue: #10926

### DIFF
--- a/.github/workflows/docker_image.yml
+++ b/.github/workflows/docker_image.yml
@@ -154,6 +154,14 @@ jobs:
           digest="${{ steps.build.outputs.digest }}"
           touch "${{ runner.temp }}/digests/${digest#sha256:}"
 
+      - name: Upload digest
+        uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


FelixYBW:
can we push the base image with spark and .m2 installed to github hub?

zhouyuan:
> can we push the base image with spark and .m2 installed to github hub?

We should have installed the basic dependencies in centos8/9 dynamic images already. 
https://github.com/apache/incubator-gluten/blob/main/dev/docker/Dockerfile.centos8-dynamic-build#L61
For the TPC tests we are using the clean images from docker hub(different version of centos/ubuntu), we should be able to make a `cache` for the m2 folder and reuse it. Will follow up on this.

PR #11144, opened 2025-11-21
[GLUTEN-11088][VL] Fix NativeBenchmarkPlanGenerator

marin-ma:
Related issue: #11088


backends-velox/src/test/scala/org/apache/gluten/benchmarks/NativeBenchmarkPlanGenerator.scala
```diff
@@ -72,7 +72,7 @@ class NativeBenchmarkPlanGenerator extends VeloxWholeStageTransformerSuite {
   }
 
   // TODO: fix on spark-4.0
```
zhouyuan: Nit: remove TODO

PR #11147, opened 2025-11-21
[GLUTEN-11088][VL] Fix MiscOperatorSuite

marin-ma:
Related issue: #11088


backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
```diff
@@ -177,10 +176,10 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
   }
 
   // TODO: fix on spark-4.0
-  testWithMaxSparkVersion("and pushdown", "3.5") {
+  test("and pushdown") {
     val df = runQueryAndCompare(
       "select l_orderkey from lineitem where l_orderkey > 2 " +
-        "and l_orderkey = 1") { _ => }
+        "and l_orderkey < 2") { _ => }
```
marin-ma, line 182: This change is due to the optimizer change in spark 4.0. The filter condition in the original SQL is optimzed out.

Issue #11133, opened 2025-11-20
[VL] JVM heap OOM in BHJ caused by on-heap columnar batch serialization


zhztheplayer:
Found the following OOM:

> ``` **/**/** **:**:** INFO [dag-scheduler-event-loop] scheduler.DAGScheduler: ResultStage 0 (collect at VeloxSparkPlanExecApi.scala:665) failed in 31.198 s due to Job aborted due to stage failure: Task 0 in stage 0.0 failed 4 times, most recent failure: Lost task 0.3 in stage 0.0 (TID 33) (*.*.*.* executor 8): java.lang.OutOfMemoryError: Java heap space ```

To solve this, we'd optimize against [this code](https://github.com/apache/incubator-gluten/blob/aadc68b7cdfd0de5aff4638bcd16a802aece9901/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala#L673-L677) to mitigate the heap pressure.

PR #11148, opened 2025-11-21
[GLUTEN-11133][VL] Reduce on-heap memory allocation and memory copy when off-heap BHJ is enabled

zhztheplayer:
This targets to fix https://github.com/apache/incubator-gluten/issues/11133 by moving most on-heap allocations on the BHJ code path to off-heap when `spark.gluten.velox.offHeapBroadcastBuildRelation.enabled=true`.

# Peak On-Heap Memory Usage Benchmark

## Results

Benchmark ran with TPC-DS Q64 / SF100 / -Xmx5G / 8 core Apple silicon.

### Before (Off-Heap BHJ Off)

Peak heap memory usage is ~3.5GiB.

Screenshot 2025-11-26 at 16 28 54

### Before (Off-Heap BHJ On)

Peak heap memory usage is ~3GiB.

Screenshot 2025-11-26 at 16 29 04

### After (Off-Heap BHJ On)

Peak heap memory usage is ~2GiB.

Screenshot 2025-11-26 at 16 29 13

## Observation

In the benchmark, the PR reduces the peak heap memory usage by ~30% (from 3GiB to 2GiB).

### DIFF
--- a/gluten-arrow/src/main/java/org/apache/spark/sql/execution/unsafe/JniUnsafeByteBuffer.java
+++ b/gluten-arrow/src/main/java/org/apache/spark/sql/execution/unsafe/JniUnsafeByteBuffer.java
@@ -29,42 +29,41 @@ import org.apache.spark.unsafe.Platform;
 public class JniUnsafeByteBuffer {
   private ArrowBuf buffer;
   private long size;
-  private boolean closed = false;
+  private boolean released = false;
 
   private JniUnsafeByteBuffer(ArrowBuf buffer, long size) {
     this.buffer = buffer;
     this.size = size;
   }
 
+  // Invoked by C++ code via JNI.
   public static JniUnsafeByteBuffer allocate(long size) {
     final ArrowBuf arrowBuf = ArrowBufferAllocators.globalInstance().buffer(size);
     return new JniUnsafeByteBuffer(arrowBuf, size);
   }
 
+  // Invoked by C++ code via JNI.
   public long address() {
     ensureOpen();
     return buffer.memoryAddress();
   }
 
+  // Invoked by C++ code via JNI.
   public long size() {
     ensureOpen();
     return size;
   }
 
   private synchronized void ensureOpen() {
-    if (closed) {
-      throw new IllegalStateException("Already closed");
+    if (released) {
+      throw new IllegalStateException("Already released");
     }
   }
 
   private synchronized void release() {
     ensureOpen();
     buffer.close();
-  }
-
-  private synchronized void close() {
-    ensureOpen();
-    closed = true;
+    released = true;
     buffer = null;
     size = 0;
   }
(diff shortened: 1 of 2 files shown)


zhztheplayer:
cc @zjuwangg @jinchengchenghh, please review the code change, thanks.

I am benchmarking the memory footprint and will update here.

gluten-arrow/src/main/java/org/apache/spark/sql/execution/unsafe/JniUnsafeByteBuffer.java
```diff
@@ -0,0 +1,91 @@
...
+      throw new IllegalStateException("Already closed");
+    }
+  }
+
+  private synchronized void release() {
+    ensureOpen();
+    buffer.close();
+  }
+
+  private synchronized void close() {
```
jinchengchenghh: Please add a comment that the buffer might not released here. Looks a bit hard to understand here, but looks like we don't have a better way, maybe don't call close in toUnsafeByteArray

gluten-arrow/src/main/java/org/apache/spark/sql/execution/unsafe/UnsafeByteArray.java
```diff
@@ -0,0 +1,165 @@
...
+public class UnsafeByteArray implements Externalizable, KryoSerializable {
+  private ArrowBuf buffer;
+  private long size;
+
+  UnsafeByteArray(ArrowBuf buffer, long size) {
+    this.buffer = buffer;
+    this.size = size;
+  }
+
+  public UnsafeByteArray() {}
```
jinchengchenghh, line 43: Do we need this constructor?

zhztheplayer: Yes, it's for kryo serialization.

gluten-arrow/src/main/java/org/apache/spark/sql/execution/unsafe/JniUnsafeByteBuffer.java
zhztheplayer: Yes. I am thinking whether it's easier to understand if using Arrow's reference management instead. Let me give it a try in this PR.

zhztheplayer: @jinchengchenghh Any thoughts about the new way: https://github.com/apache/incubator-gluten/pull/11148/commits/c1accf5bc8826c35db5bcd1b065c9fb76c47cc9a, thanks.

jinchengchenghh: Looks good

zhztheplayer:
I've updated the benchmark results in the PR description.

zhztheplayer:
Thanks @jinchengchenghh for the review.

PR #11153, opened 2025-11-22
[CORE] Refactor push down filter to scan

Zouxxyy:

Before when perform pushing filter to scan we change `dataFilters` in FileSourceScanExecTransformer or just `setPushDownFilters` in `BatchScanExecTransformerBase`, we'd better not modify the existing variables.

Instead, Introduce these functions in `BasicScanExecTransformer`

```scala
  /** Returns the filters that already exists in scan. */
  def scanFilters: Seq[Expression]

  /**
   * Returns the filters that pushed by
   * [[org.apache.gluten.extension.columnar.PushDownFilterToScan]].
   */
  def pushDownFilters: Option[Seq[Expression]]

  /** Copy the scan with filters that pushed by filterNode. */
  def withNewPushdownFilters(filters: Seq[Expression]): BasicScanExecTransformer
```

And then combine them to filterExprs

```scala
  /** Returns the filters that can be pushed down to native file scan */
  final def filterExprs(): Seq[Expression] = {
     ...
     // combine scanFilters + pushDownFilters and do validation
     ...
  }
```

Besides strengthened the filter validation in `combine scanFilters + pushDownFilters`: if there are unsupported filters present in the scan node but not in the filter node, the scan must fall back (even though I haven't constructed an example of this scenario, it serves as a safeguard).

Refactor `PushDownFilterToScan`

```scala
object PushDownFilterToScan extends Rule[SparkPlan] with PredicateHelper {
  override def apply(plan: SparkPlan): SparkPlan = plan.transformUp {
    case filter: FilterExecTransformerBase =>
      filter.child match {
        case scan: BasicScanExecTransformer
            if BackendsApiManager.getSparkPlanExecApiInstance.supportPushDownFilterToScan(
              scan) && scan.pushDownFilters.isEmpty =>
          val newScan = scan.withNewPushdownFilters(splitConjunctivePredicates(filter.cond))
          if (newScan.doValidate().ok()) {
            filter.withNewChildren(Seq(newScan))
          } else {
            filter
          }
        case _ => filter
      }
  }
}
```

### DIFF
--- a/gluten-kafka/src/main/scala/org/apache/gluten/execution/MicroBatchScanExecTransformer.scala
+++ b/gluten-kafka/src/main/scala/org/apache/gluten/execution/MicroBatchScanExecTransformer.scala
@@ -66,7 +66,7 @@ case class MicroBatchScanExecTransformer(
 
   override def scanFilters: Seq[Expression] = Seq.empty
 
-  override def supportPushDownFilters = false
+  override def supportPushDownFilters: Boolean = false
 
   def pushDownFilters: Option[Seq[Expression]] = None
 
--- a/gluten-substrait/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
@@ -65,7 +65,7 @@ case class HiveTableScanExecTransformer(
 
   override def scanFilters: Seq[Expression] = Seq.empty
 
-  override def supportPushDownFilters = false
+  override def supportPushDownFilters: Boolean = false
 
   override def pushDownFilters: Option[Seq[Expression]] = None
 


backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
```diff
@@ -849,10 +849,7 @@ class CHSparkPlanExecApi extends SparkPlanExecApi with Logging {
     }
   }
 
-  /** Clickhouse Backend only supports part of filters for parquet. */
-  override def postProcessPushDownFilter(
-      extraFilters: Seq[Expression],
-      sparkExecNode: LeafExecNode): Seq[Expression] = {
+  override def supportPushDownFilterToScan(sparkExecNode: LeafExecNode): Boolean = {
```
Zouxxyy, line 852: Here I'm just retaining the previous logic:
1. Do not push down filters when using Parquet scan.
2. Add CK's additional `isSupportPushDownFilter` condition.

Zouxxyy:
CC @liujiayi771, @jinchengchenghh  for a look too

gluten-kafka/src/main/scala/org/apache/gluten/execution/MicroBatchScanExecTransformer.scala
```diff
@@ -64,7 +64,13 @@ case class MicroBatchScanExecTransformer(
   @transient override lazy val inputPartitionsShim: Seq[InputPartition] =
     stream.planInputPartitions(start, end)
 
-  override def filterExprs(): Seq[Expression] = Seq.empty
+  override def scanFilters: Seq[Expression] = Seq.empty
+
+  def pushDownFilters: Option[Seq[Expression]] = Some(Seq.empty)
+
+  def withNewPushdownFilters(filters: Seq[Expression]): BatchScanExecTransformerBase = {
+    throw new UnsupportedOperationException()
```
liujiayi771: Pls add some exception msg.

gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
```diff
@@ -198,6 +188,15 @@ abstract class BatchScanExecTransformerBase(
...
 
+  override def equals(other: Any): Boolean = other match {
+    case other: BatchScanExecTransformerBase =>
+      this.pushDownFilters == other.pushDownFilters && super.equals(other)
+    case _ =>
+      false
+  }
+
+  override def hashCode(): Int = Objects.hashCode(batch, runtimeFilters, pushDownFilters)
+
```
liujiayi771, line 200: What is the purpose of this here?

gluten-substrait/src/main/scala/org/apache/spark/sql/hive/HiveTableScanExecTransformer.scala
```diff
@@ -63,7 +63,12 @@ case class HiveTableScanExecTransformer(
     hiveQlTable.getOutputFormatClass,
     hiveQlTable.getMetadata)
 
-  override def filterExprs(): Seq[Expression] = Seq.empty
+  override def scanFilters: Seq[Expression] = Seq.empty
+
+  override def pushDownFilters: Option[Seq[Expression]] = Some(Seq.empty)
+
+  override def withNewPushdownFilters(filters: Seq[Expression]): BasicScanExecTransformer =
+    throw new UnsupportedOperationException()
```
liujiayi771: ditto

gluten-ut/spark32/src/test/scala/org/apache/spark/sql/execution/datasources/orc/GlutenOrcV2SchemaPruningSuite.scala
```diff
@@ -30,7 +30,8 @@ class GlutenOrcV2SchemaPruningSuite extends OrcV2SchemaPruningSuite with GlutenS
     val fileSourceScanSchemata =
       collect(df.queryExecution.executedPlan) {
         case BatchScanExec(_, scan: OrcScan, _) => scan.readDataSchema
-        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _) => scan.readDataSchema
```
liujiayi771, line 33: I think using this approach is better, as it's less likely to cause compatibility issues:
```scala
case b: BatchScanExecTransformer if b.scan.isInstanceOf[OrcScan] =>
```

gluten-ut/spark33/src/test/scala/org/apache/spark/sql/execution/datasources/orc/GlutenOrcV2SchemaPruningSuite.scala
```diff
@@ -35,7 +35,8 @@ class GlutenOrcV2SchemaPruningSuite extends OrcV2SchemaPruningSuite with GlutenS
     val fileSourceScanSchemata =
       collect(df.queryExecution.executedPlan) {
         case BatchScanExec(_, scan: OrcScan, _, _) => scan.readDataSchema
-        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _) => scan.readDataSchema
+        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _, _) =>
```
liujiayi771: ditto

gluten-ut/spark34/src/test/scala/org/apache/spark/sql/execution/datasources/orc/GlutenOrcV2SchemaPruningSuite.scala
```diff
@@ -35,7 +35,8 @@ class GlutenOrcV2SchemaPruningSuite extends OrcV2SchemaPruningSuite with GlutenS
     val fileSourceScanSchemata =
       collect(df.queryExecution.executedPlan) {
         case BatchScanExec(_, scan: OrcScan, _, _, _, _, _, _, _) => scan.readDataSchema
-        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _) => scan.readDataSchema
+        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _, _) =>
```
liujiayi771: ditto

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/execution/datasources/orc/GlutenOrcV2SchemaPruningSuite.scala
```diff
@@ -35,7 +35,8 @@ class GlutenOrcV2SchemaPruningSuite extends OrcV2SchemaPruningSuite with GlutenS
     val fileSourceScanSchemata =
       collect(df.queryExecution.executedPlan) {
         case BatchScanExec(_, scan: OrcScan, _, _, _, _) => scan.readDataSchema
-        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _) => scan.readDataSchema
+        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _, _) =>
```
liujiayi771: ditto

gluten-ut/spark40/src/test/scala/org/apache/spark/sql/execution/datasources/orc/GlutenOrcV2SchemaPruningSuite.scala
```diff
@@ -35,7 +35,8 @@ class GlutenOrcV2SchemaPruningSuite extends OrcV2SchemaPruningSuite with GlutenS
     val fileSourceScanSchemata =
       collect(df.queryExecution.executedPlan) {
         case BatchScanExec(_, scan: OrcScan, _, _, _, _) => scan.readDataSchema
-        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _) => scan.readDataSchema
+        case BatchScanExecTransformer(_, scan: OrcScan, _, _, _, _, _, _, _, _) =>
```
liujiayi771: ditto

gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
Zouxxyy: This is because even though BatchScanExec is a case class, spark still overwrite the equals and hashCode

See in gluten's `AbstractBatchScanExec` or spark's  `BatchScanExec`

```scala
  // TODO: unify the equal/hashCode implementation for all data source v2 query plans.
  override def equals(other: Any): Boolean = other match {
    case other: AbstractBatchScanExec =>
      this.batch != null && this.batch == other.batch &&
      this.runtimeFilters == other.runtimeFilters &&
      this.commonPartitionValues == other.commonPartitionValues &&
      this.replicatePartitions == other.replicatePartitions &&
      this.applyPartialClustering == other.applyPartialClustering
    case _ =>
      false
  }

  override def hashCode(): Int = Objects.hashCode(batch, runtimeFilters)
```

Here add a new val `pushDownFilters `, so we need to overwrite them too.

liujiayi771 (approved):

gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
```diff
@@ -102,30 +109,13 @@ abstract class BatchScanExecTransformerBase(
...
-        expr =>
-          ExpressionConverter.canReplaceWithExpressionTransformer(
-            ExpressionConverter.replaceAttributeReference(expr),
-            output)
-      }
+  override def scanFilters: Seq[Expression] = scan match {
+    case fileScan: FileScan => fileScan.dataFilters
     case _ =>
-      logInfo(s"${scan.getClass.toString} does not support push down filters")
+      // todo: support DSv2 scan
```
jinchengchenghh: Please keep this comment, DSv2 scan is also FileScan, InMeroryTableScan may trigger this

jinchengchenghh:

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/PushDownFilterToScan.scala
```diff
@@ -31,35 +31,15 @@ object PushDownFilterToScan extends Rule[SparkPlan] with PredicateHelper {
...
       filter.child match {
-        case fileScan: FileSourceScanExecTransformer =>
-          val pushDownFilters =
-            BackendsApiManager.getSparkPlanExecApiInstance.postProcessPushDownFilter(
-              splitConjunctivePredicates(filter.cond),
-              fileScan)
-          val newScan = fileScan.copy(dataFilters = pushDownFilters)
+        case scan: BasicScanExecTransformer
+            if BackendsApiManager.getSparkPlanExecApiInstance.supportPushDownFilterToScan(
+              scan) && scan.pushDownFilters.isEmpty =>
```
zml1206: Is `scan.pushDownFilters.isEmpty` redundant?

gluten-substrait/src/main/scala/org/apache/gluten/execution/BatchScanExecTransformer.scala
Zouxxyy: Thanks for the review, I've made some minor adjustments.

I'd like to comment here is that we could support other DSv2 scans in the future. For example, if IcebergScan or PaimonScan retains Spark's Catalyst Filters and provides interface to expose them, then we could either match more case or just inherit this method.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/PushDownFilterToScan.scala
Zouxxyy: > Is `scan.pushDownFilters.isEmpty` redundant?

`pushDownFilters` is an Option variable, and populating it once is sufficient.

gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
```diff
@@ -37,7 +37,40 @@ trait BasicScanExecTransformer extends LeafTransformSupport with BaseDataSource
...
+  def scanFilters: Seq[Expression]
+
+  /**
+   * Returns the filters that pushed by
+   * [[org.apache.gluten.extension.columnar.PushDownFilterToScan]].
+   */
+  def pushDownFilters: Option[Seq[Expression]]
+
+  /** Copy the scan with filters that pushed by filterNode. */
+  def withNewPushdownFilters(filters: Seq[Expression]): BasicScanExecTransformer
```
zml1206: Suggestion:
  ```
  val pushDownFilters: Option[Seq[Expression]]

  /** Copy the scan with filters that pushed by filterNode. */
  def withNewPushdownFilters(filters: Seq[Expression]): BasicScanExecTransformer = {
    this.copy(pushDownFilters = Some(filters))
  }
  ```
Reduce the override of `withNewPushdownFilters`.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/PushDownFilterToScan.scala
zml1206: The same plan will only be traversed once?

Zouxxyy: Thanks, updated. It seems more appropriate to let Spark handle the planning itself.

Previously, I explicitly set `Some(Seq.empty)` to indicate that a scan (e.g., `HiveTableScanExecTransformer` and `MicroBatchScanExecTransformer`) does not support filter pushdown. 

Now that this has been removed, I've introduced a `supportPushDownFilters` flag to explicitly indicate whether filter pushdown is supported by the Scan.

gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
Zouxxyy: We can only call copy on final case classes, it's not available in abstract class.

Zouxxyy:
The CK's test pipline appears to be failing to trigger correctly, it was successful previously.

PR #11181, opened 2025-11-25
[GLUTEN-11088][VL] Add config in GlutenCastSuite to cast to char/varchar

marin-ma:
Need to explicitly set `spark.sql.preserveCharVarcharTypeInfo=true` for gluten's test framework. We need to add this into the Gluten test because it overrides the checkEvaluation that invokes Spark's RowEncoder, which requires this configuration to be set. But in Vanilla spark, the checkEvaluation method doesn't invoke RowEncoder.

Related issue: #11088

### DIFF
--- a/gluten-ut/spark40/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
+++ b/gluten-ut/spark40/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
@@ -716,14 +716,14 @@ class VeloxTestSettings extends BackendTestSettings {
     // Extra ColumnarToRow is needed to transform vanilla columnar data to gluten columnar data.
     .exclude("SPARK-37369: Avoid redundant ColumnarToRow transition on InMemoryTableScan")
   // TODO: fix in Spark-4.0
-  enableSuite[GlutenFileSourceCharVarcharTestSuite]
-    .exclude("length check for input string values: nested in array")
-    .exclude("length check for input string values: nested in array")
-    .exclude("length check for input string values: nested in map key")
-    .exclude("length check for input string values: nested in map value")
-    .exclude("length check for input string values: nested in both map key and value")
-    .exclude("length check for input string values: nested in array of struct")
-    .exclude("length check for input string values: nested in array of array")
+  // enableSuite[GlutenFileSourceCharVarcharTestSuite]
+  //   .exclude("length check for input string values: nested in array")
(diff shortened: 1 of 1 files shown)


gluten-ut/spark40/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenCastSuite.scala
```diff
@@ -288,4 +288,26 @@ class GlutenCastSuite extends CastWithAnsiOffSuite with GlutenTestsTrait {
     checkEvaluation(cast(d, TimestampType), new Timestamp(c.getTimeInMillis))
   }
 
+  testGluten("Casting to char/varchar") {
+    withSQLConf(SQLConf.PRESERVE_CHAR_VARCHAR_TYPE_INFO.key -> "true") {
```
jinchengchenghh: Do we not support PRESERVE_CHAR_VARCHAR_TYPE_INFO false? What's the failure? If we always need to set PRESERVE_CHAR_VARCHAR_TYPE_INFO to true, may set in the plugin as default value

marin-ma: If false, then exception will be thrown in Spark. We need to add this into the Gluten test because it overrides the `checkEvaluation` that invokes Spark's RowEncoder, which requires this configuration to be set. But in Vanilla spark, the `checkEvaluation` method doesn't invoke RowEncoder.

jinchengchenghh: Thanks for you explaination, could you add it to PR description?

jinchengchenghh: What if we set the config to all the tests in this suite? Looks like if add the new test, we still needs this config

jinchengchenghh: Then we can set it in beforeClass

marin-ma: Make sense. I will make this change.

marin-ma:
@jinchengchenghh Do you have further comments? cc: @zhouyuan

Issue #11052, opened 2025-11-07
[CH] ARM CI is not stable


zhouyuan:
### Backend

CH (ClickHouse)

### Bug description

```
[2025-11-05T11:32:57.997Z] [ERROR] COMPILATION ERROR : 
[2025-11-05T11:32:57.997Z] [INFO] -------------------------------------------------------------
[2025-11-05T11:32:57.997Z] [ERROR] /home/jenkins/agent/workspace/gluten/gluten-ci-arm/ut-stage-5/gluten-iceberg/src/main/java/org/apache/gluten/substrait/rel/IcebergLocalFilesNode.java:[22,25] error: cannot access DeleteFile
[2025-11-05T11:32:57.997Z]   bad class file: /root/.m2/repository/org/apache/iceberg/iceberg-spark-runtime-3.5_2.13/1.8.0/iceberg-spark-runtime-3.5_2.13-1.8.0.jar(org/apache/iceberg/DeleteFile.class)
[2025-11-05T11:32:57.997Z]     class file has wrong version 55.0, should be 52.0
[2025-11-05T11:32:57.997Z]     Please remove or make sure it appears in the correct subdirectory of the classpath.
[2025-11-05T11:32:57.997Z] [INFO] 1 error
```

### Spark version

None

### Relevant logs

```bash

```

philo-he:
Similar to:
- https://github.com/apache/incubator-gluten/issues/10918

PR #11187, opened 2025-11-25
[GLUTEN-11052][CH] Disable ARM CI temporary

zhouyuan:
ARM CI in CH is not stable. This patch disable it temporary 

## How was this patch tested?
pass CH CI

fixes: #11052


zhouyuan:
@zzcclp could you please help to take a look on this? We recently ran into the unstable CI issue on CH + ARM

PR #11195, opened 2025-11-26
[GLUTEN-11088] Fix GlutenDataFrameFunctionsSuite in Spark-4.0

zml1206:
https://github.com/apache/spark/blob/29434ea766b0fc3c3bf6eaadb43a8f931133649e/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala#L2928-L2937
Vanilla spark throw SparkRuntimeException, gluten throw SparkException.

Related issue: #11088

### DIFF
--- a/gluten-ut/spark40/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
+++ b/gluten-ut/spark40/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
@@ -773,7 +773,7 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude("aggregate function - array for non-primitive type")
     // Rewrite this test because Velox sorts rows by key for primitive data types, which disrupts the original row sequence.
     .exclude("map_zip_with function - map of primitive types")
-    // // Vanilla spark throw SparkRuntimeException, gluten throw SparkException.
+    // Vanilla spark throw SparkRuntimeException, gluten throw SparkException.
     .exclude("map_concat function")
     .exclude("transform keys function - primitive data types")
   enableSuite[GlutenDataFrameHintSuite]


gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenDataFrameFunctionsSuite.scala
```diff
@@ -49,4 +52,230 @@ class GlutenDataFrameFunctionsSuite extends DataFrameFunctionsSuite with GlutenS
...
+      (null, Map[Int, Int](3 -> 300, 4 -> 400))
+    ).toDF("map1", "map2")
+
+    val expected1a = Seq(
+      Row(Map(1 -> 100, 2 -> 200, 3 -> 300, 4 -> 400)),
+      Row(Map(1 -> 400, 2 -> 200, 3 -> 300)),
+      Row(null)
+    )
+
+    intercept[SparkException](df1.selectExpr("map_concat(map1, map2)").collect())
```
zhouyuan, line 69: I think there are similar tests found when i work on the first patch to enable unit tests. Is it too heavy to modify the `map_concat `function to throw the right exception?

zml1206: I haven't found anything similar yet. Would the impact be greater if GlutenException were changed to SparkRuntimeException?

jinchengchenghh: Can GlutenException inherit SparkRuntimeException? Now it inherits RuntimeException, I'm not sure if it is more feasible.

zml1206: If it inherits SparkRuntimeException, it throws GlutenException, not SparkRuntimeException.

PR #11202, opened 2025-11-26
[GLUTEN-11088][VL] Fix Spark4.0 varchar type check suites

jinchengchenghh:
Use one test to make sure the native code is really called, override function assertLengthCheckFailure to assert result correct.

Related issue: #11088

### DIFF
--- a/gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenCharVarcharTestSuite.scala
+++ b/gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenCharVarcharTestSuite.scala
@@ -16,33 +16,39 @@
  */
 package org.apache.spark.sql
 
-import org.apache.spark.SparkException
+import org.apache.spark.{SparkException, SparkRuntimeException, SparkThrowable}
+
+trait GlutenCharVarcharTestSuite extends CharVarcharTestSuite with GlutenSQLTestsTrait {
+  protected val ERROR_MESSAGE =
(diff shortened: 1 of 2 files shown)


gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenCharVarcharTestSuite.scala
```diff
@@ -51,73 +60,11 @@ class GlutenFileSourceCharVarcharTestSuite
         val e = intercept[SparkException] {
           sql("INSERT INTO t VALUES (array('a', '123456'))")
         }
-        assert(e.getMessage.contains(ERROR_MESSAGE))
-    }
-  }
-
-  testGluten("length check for input string values: nested in map key") {
```
zhouyuan, line 58: It seems these tests are still valid in original Spark tests

jinchengchenghh: Yes, I override assertLengthCheckFailure to test the test result as PR description

jinchengchenghh: Spark 4.0 does a refactor, move the test result check to a public function, so we can override this function

jinchengchenghh:
Do you have further comments? @zhouyuan

Issue #8782, opened 2025-02-19
[VL] upstream OAP/Velox commits to upstream


FelixYBW:
### Description

| oap-velox commit | upstream PR(s) | progress | comments |
|-------------------|-----------------|----------|------------|
| https://github.com/oap-project/velox/commit/1a24a70b446f569f9c6f24b0a6b7bb65fb9c9435 | https://github.com/facebookincubator/velox/pull/5962 | PR under review | |
| https://github.com/oap-project/velox/commit/757f14b345451c6fa7a781be8311bea438c2d79c | | **Blocked** | PR is not created. Blocked by decimal avg signature issue. Relevant PR: https://github.com/facebookincubator/velox/pull/11999. Also see discussion: https://github.com/facebookincubator/velox/discussions/9048#discussioncomment-10622862 |
| https://github.com/oap-project/velox/commit/72dd385994bdcfd51a00ae2ac562f813c3b8b0f4 | | **Blocked** | PR is not created. This commit has conflict with https://github.com/facebookincubator/velox/pull/11999 . Waiting for the decimal avg signature issue to be resolved first. |
| ~https://github.com/oap-project/velox/commit/fae08281d5a9db217feb74e4b0f07147e4bf8301~ | ~https://github.com/facebookincubator/velox/pull/11957~ | ~Merged~ | ~After this PR, Gluten will need to adapt to the new configuration.~ |
| ~https://github.com/oap-project/velox/commit/bf0b43de67f7353a76fc93ceb6f6dc8d102f16a6~ | ~https://github.com/facebookincubator/velox/pull/11949~ | ~Merged~ | ~(Timestamp Int64 is supported in https://github.com/facebookincubator/velox/pull/11530) Gluten PR https://github.com/apache/incubator-gluten/pull/8317 needs to be merged along with the daily rebase.~ |
| https://github.com/oap-project/velox/commit/6452363571543549059b2297a1dcda7867e97265 | | Will not upstream | To pass security check. |
| https://github.com/oap-project/velox/commit/c9e876e1d97f9f60956be0327a7e0e30ffa9793c | https://github.com/facebookincubator/velox/pull/11431</br>https://github.com/facebookincubator/velox/pull/11067 | PR under review | |
[...]

PR #11210, opened 2025-11-27
[GLUTEN-8782][VL] Add correct suffix to the name of merge_extract companion function

rui-mo:

Add correct suffix to the name of merge_extract companion function, so we no longer need the change made to Velox https://github.com/IBM/velox/commit/b8da5d0d33fbf177e79fe5df9e7e43809d562a21.

## How was this patch tested?

Verified with existing unit tests.

Related issue: #8782


cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -231,15 +267,29 @@ core::AggregationNode::Step SubstraitToVeloxPlanConverter::toAggregationFunction
...
       suffix = "_partial";
       break;
-    case core::AggregationNode::Step::kFinal:
-      suffix = "_merge_extract";
-      break;
+    case core::AggregationNode::Step::kFinal: {
+      auto functionName = baseName + "_merge_extract";
+      auto signatures = exec::getAggregateFunctionSignatures(functionName);
+      if (signatures.has_value() && signatures.value().size() > 0) {
+        // The merge_extract function is registered without suffix.
```
xinghuayu007, line 281: nit: Put the comment on line 280.

xinghuayu007:
+1

cpp/velox/substrait/SubstraitToVeloxPlan.cc
rui-mo: This comment is true only when the condition is met, so I assume keeping it here is more appropriate. Thanks.

ep/build-velox/src/get-velox.sh
```diff
@@ -18,8 +18,8 @@ set -exu
 
 CURRENT_DIR=$(cd "$(dirname "$BASH_SOURCE")"; pwd)
 VELOX_REPO=https://github.com/IBM/velox.git
-VELOX_BRANCH=dft-2025_11_26
-VELOX_ENHANCED_BRANCH=ibm-2025_11_26
+VELOX_BRANCH=dft-2025_11_26_fix
+VELOX_ENHANCED_BRANCH=ibm-2025_11_26_fix
```
zhouyuan, line 22: Cc: @FelixYBW 
removed one commit from original OAP patches

PR #2324, opened 2023-07-13
[VL][DOC] Refine build guide for velox backend

tomscut:
## What changes were proposed in this pull request?

Remove the extra command `make -j` because it is already in the script `compile.sh`. In addition, updated the version of spark.

## How was this patch tested?

(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #11212, opened 2025-11-27
[GLUTEN-11088][CORE] Fix Spark 4.0 GlutenDynamicPartitionPruningV1SuiteAEOn

jinchengchenghh:
Support FileSourceScanExecTransformer stream.

FileSourceScanExecTransformer is not equal because dataFilters is not equal, `(cast(x#688 as double) = ReusedSubquery Subquery subquery#683, [id=#2324])` and  `(cast(x#688 as double) = Subquery subquery#683, [id=#2324])` should be same, must use canonicalized expression to deduplicate.

Fix dynamic pruning FileSourceScanExec getPartitions because add new class ScanFileListing and other refactor for inputRDD

Related issue: #11088

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicPhysicalOperatorTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicPhysicalOperatorTransformer.scala
@@ -313,7 +313,8 @@ object FilterHandler extends PredicateHelper {
   }
 
   def subtractFilters(left: Seq[Expression], right: Seq[Expression]): Seq[Expression] = {
-    (left.toSet -- right.toSet).toSeq
+    val scanSet = right.map(_.canonicalized).toSet
+    left.filter(f => !scanSet.contains(f.canonicalized))
   }
 
   def combineFilters(left: Seq[Expression], right: Seq[Expression]): Seq[Expression] = {
--- a/gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenDynamicPartitionPruningSuite.scala
+++ b/gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenDynamicPartitionPruningSuite.scala
@@ -30,7 +30,6 @@ import org.apache.spark.sql.execution.exchange.{BroadcastExchangeLike, ReusedExc
 import org.apache.spark.sql.execution.joins.BroadcastHashJoinExec
 import org.apache.spark.sql.execution.metric.SQLMetric
 import org.apache.spark.sql.execution.streaming.{MemoryStream, StreamingQueryWrapper}
-import org.apache.spark.sql.functions.col
 import org.apache.spark.sql.internal.SQLConf
 
 abstract class GlutenDynamicPartitionPruningSuiteBase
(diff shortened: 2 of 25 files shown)


jinchengchenghh:
Run Gluten ClickHouse CI on x86

gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicPhysicalOperatorTransformer.scala
```diff
@@ -313,7 +313,8 @@ object FilterHandler extends PredicateHelper {
   }
 
   def subtractFilters(left: Seq[Expression], right: Seq[Expression]): Seq[Expression] = {
-    (left.toSet -- right.toSet).toSeq
+    val scanSet = left.map(_.canonicalized).toSet
```
jinchengchenghh: Hi, I find it causes ReuseExchange issue, I find you add combineFilters, do you have any suggestions for that, I think it may meet the same issue @Zouxxyy

Zouxxyy: Oh, I checked the commit history, and this PR (https://github.com/apache/incubator-gluten/pull/6296) changed ExpressionSet to a regular Set. The original ExpressionSet compared expressions based on both determinism and canonical representation, whereas a regular Set only compares expressions directly (using their default equals/hashCode).

It seems what we actually need is comparison based on canonicalized forms. Perhaps we could modify the code like this:

```scala
def subtractFilters(left: Seq[Expression], right: Seq[Expression]): Seq[Expression] = {
  val rightCanonicalSet = right.map(_.canonicalized).toSet
  left.filterNot(expr => rightCanonicalSet.contains(expr.canonicalized))
}

def combineFilters(left: Seq[Expression], right: Seq[Expression]): Seq[Expression] = {
  val seen = scala.collection.mutable.Set[Expression]()
  val result = scala.collection.mutable.ListBuffer[Expression]()

  def tryAdd(expr: Expression): Unit = {
    val canon = expr.canonicalized
    if (!seen.contains(canon)) {
      seen += canon
      result += expr
    }
  }

  (left ++ right).foreach(tryAdd)
  result.toList
}

```

jinchengchenghh: Maybe here we cannot use canonicalized expression, otherwise, will throw subquery is not finished exception. with following code
```
val scanSet = left.map(_.canonicalized).toSet
scanSet.toSeq ++ right.filter(f => !scanSet.contains(f.canonicalized))
```
Keep current code now

gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenDynamicPartitionPruningSuite.scala
```diff
@@ -577,64 +572,6 @@ class GlutenDynamicPartitionPruningV1SuiteAEOff
...
-        val subqueryIds = plan.collectWithSubqueries { case s: SubqueryExec => s.id }
-        val reusedSubqueryIds = plan.collectWithSubqueries {
-          case rs: ReusedSubqueryExec => rs.child.id
-        }
-
-        // By default Gluten pushes more filters than vanilla Spark.
-        //
-        // See also org.apache.gluten.execution.FilterHandler#applyFilterPushdownToScan
-        // See also DynamicPartitionPruningSuite.scala:1362
-        assert(subqueryIds.size == 3, "Whole plan subquery reusing not working correctly")
```
jinchengchenghh, line 630: Maybe the previous version has extra subquery, the filter should not affect ReusedSubquery, Gluten with Spark4.0 result is same with jvm Spark

PR #11213, opened 2025-11-28
[VL] Fast fail or resend data to uniffle ASSP

zuston:

Uniffle’s partition reassignment mechanism enables fast switching to alternative shuffle servers, improving performance and stability. However, in the previous Uniffle integration with Gluten, this mechanism did not function properly due to a missing trigger, which could cause significant partition backpressure without activating multi shuffle-server load balancing.

This PR also enables fast fail for the shuffle writer in case any data is lost.

## How was this patch tested?

Existing tests are sufficient


zuston:
Could you help take a look? @zhouyuan

zhouyuan (approved):

Issue #11224, opened 2025-11-28
[VL] FileSourceScan check codec


jinchengchenghh:
### Backend

VL (Velox)

### Bug description

GlutenParquetCodecSuite write and read - file source parquet - codec: lz4_raw, lz4_raw is not supported, now we only check codec in the write, but not check in the read

### Spark version

None

### Relevant logs

```bash

```

PR #11225, opened 2025-11-28
[GLUTEN-11224][VL] FileSourceScan check codec

jinchengchenghh:
Check the codec in read
Enable some tests in spark-ut
Create the issue for some tests which do not fail by previous reason.
Related issue: #11224

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/utils/ParquetEncryptionDetectionSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/utils/ParquetEncryptionDetectionSuite.scala
@@ -16,10 +16,13 @@
  */
 package org.apache.gluten.utils
 
-import org.apache.spark.sql.{GlutenQueryTest, SparkSession}
+import org.apache.gluten.config.GlutenConfig
+
+import org.apache.spark.SparkConf
 import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
 import org.apache.spark.sql.execution.datasources.parquet.ParquetOptions
 import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.test.SharedSparkSession
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
@@ -51,7 +54,7 @@ import scala.collection.JavaConverters._
  *   - Ensures the file is still detected as encrypted despite the plaintext footer.
  */
 
-class ParquetEncryptionDetectionSuite extends GlutenQueryTest {
+class ParquetEncryptionDetectionSuite extends SharedSparkSession {
 
   private val masterKey =
     Base64.getEncoder.encodeToString("0123456789012345".getBytes(StandardCharsets.UTF_8))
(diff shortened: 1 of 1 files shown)


backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
```diff
@@ -119,18 +134,35 @@ object ParquetMetadataUtils {
...
+        ParquetFooterReader.readFooter(conf, fileStatus, ParquetMetadataConverter.NO_FILTER)
       } catch {
         case _: RuntimeException =>
           // Ignored as it's could be a "Not a Parquet file" exception.
-          return false
+          return None
+      }
+    val validationChecks = Seq(
+      validateCodec(footer),
+      isTimezoneFoundInMetadata(footer, parquetOptions)
```
zhztheplayer, line 149: Actually `isParquetFileEncrypted` is also part of the metadata check. Can we move union all the 3 checks together?

zhztheplayer (commented):
Let's wait for https://github.com/apache/incubator-gluten/pull/11233 so we can track the performance history more conveniently. Thanks!

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
jinchengchenghh: Yes, it has a config to control, we could union all the 3 checks and config, please union the config in https://github.com/apache/incubator-gluten/pull/11233, I will union the check

zhztheplayer: > Yes, it has a config to control, we could union all the 3 checks and config, please union the config in #11233, I will union the check

Yes, let's use `spark.gluten.sql.fallbackUnexpectedMetadataParquet` for newly added checks. Though it's yet to be discussed whether to remove the old option for encryption validation.

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
```diff
@@ -94,24 +109,17 @@ object ParquetMetadataUtils {
       parquetOptions: ParquetOptions,
       fileLimit: Int
   ): Option[String] = {
-    val isEncryptionValidationEnabled = GlutenConfig.get.parquetEncryptionValidationEnabled
```
jinchengchenghh, line 97: Next PR will remove parquetEncryptionValidationEnabled and make the fileLimit to be correct

zhztheplayer: Will it be okay to you if we remove the config option in an individual PR? So it's easier to call the original author for a review, or to revert if something unexpected happens, etc.

zhztheplayer: Because I am not sure whether users are turning on the option `spark.gluten.sql.fallbackEncryptedParquet` so immediately changing the option key can possibly break their use cases?

cc @ArnavBalyan

jinchengchenghh: Sure, added back.

ArnavBalyan: Thanks @zhztheplayer and @jinchengchenghh, we upgraded the parquet version and it should be safe to remove it, if no other users have a dependency on it

PR #11227, opened 2025-11-29
[VL] Extract common createHiveConnectorSessionConfig method and use it for Iceberg write

Zouxxyy:

Currently, the Hive connector's conf is composed of sessionConf + conf.

The HiveConnectorSessionConfig currently used for Iceberg writes is defined separately and misses many configurations, such as `partition_path_as_lower_case`=false. This causes the case of partition values in the written file paths to be lost.

## How was this patch tested?

Add a case "iceberg insert partition table with uppercase partition name"


Zouxxyy:
CC @jinchengchenghh @marin-ma for a look, thanks

PR #11229, opened 2025-12-01
[VL] Support pass unmapped dynamic and static configs to Velox 

Yohahaha:

Avoid re-compile and publish whole project when we need overwrite some configs in prod env.


Yohahaha:
@Zouxxyy @liujiayi771

kerwin-zk (approved):

cpp/velox/utils/ConfigExtractor.cc
```diff
@@ -237,6 +237,8 @@ std::shared_ptr<facebook::velox::config::ConfigBase> createHiveConnectorSessionC
       conf->get<bool>(kParquetUseColumnNames, true) ? "true" : "false";
   configs[facebook::velox::connector::hive::HiveConfig::kOrcUseColumnNamesSession] =
       conf->get<bool>(kOrcUseColumnNames, true) ? "true" : "false";
+
+  overwriteVeloxConf(conf.get(), configs, kDynamicBackendConfPrefix);
```
zhouyuan, line 241: Would this introduce some redundant config set?

Yohahaha: only works for unmapped configs, mapped configs use spark defined name.

PR #11233, opened 2025-12-01
[VL] Disable Parquet metadata validation by default due to performance regression

zhztheplayer:
Performance regressions is [found](https://github.com/apache/incubator-gluten/pull/11117#issuecomment-3587911637) in PR https://github.com/apache/incubator-gluten/pull/11117. We should disable the validation by default temporarily.


gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1569,6 +1576,14 @@ object GlutenConfig extends ConfigRegistry {
       .checkValue(_ > 0, s"must be positive.")
       .createWithDefault(10)
 
+  val ENCRYPTED_PARQUET_FALLBACK_ENABLED =
```
jinchengchenghh, line 1579: Could we use one config for all the detect?

jinchengchenghh:
Please run the gen-config shell script

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
zhztheplayer: I think it's feasible but we'd open a new PR to remove. I suspect there're still users relying on the old option.

zhztheplayer:
cc @Yohahaha

Yohahaha:
```
def validateMetadata(): Option[String] = {
      if (format != ParquetReadFormat || rootPaths.isEmpty) {
        // Only Parquet is needed for metadata validation so far.
        return None
      }
      val fileLimit = GlutenConfig.get.parquetMetadataFallbackFileLimit
        .max(GlutenConfig.get.parquetEncryptionValidationFileLimit)
      val parquetOptions = new ParquetOptions(CaseInsensitiveMap(properties), SQLConf.get)
      val parquetMetadataValidationResult =
        ParquetMetadataUtils.validateMetadata(
          format,
          rootPaths,
          hadoopConf,
          parquetOptions,
          fileLimit)
      parquetMetadataValidationResult.map(
        reason => s"Detected unsupported metadata in parquet files: $reason")
    }
```

The issue still exists, the rootPaths maybe huge in large partition table, too many list call happens..

FelixYBW:
I noted performance regression in our perf test. driver spends about 12s before each queries.

FelixYBW:
@Yohahaha would you submit a quick fix? Or I can revert the two PRs tomorrow. Revert has conflict now.

It's too bad our perf track is disabled recently.

@zhztheplayer @jinchengchenghh can we cache the infor for each table after first time sample? not only this parquet metadata validation but all other samples like check codec.

also, we can save the info to delta/iceberg.

FelixYBW:
Sorry, just noted your fix. #11307

Issue #11169, opened 2025-11-24
[VL] Enhancements to move more on-heap memory allocation to off-heap


zhztheplayer:
### Description

This will be a series of efforts to move more on-heap memory allocation to off-heap against both Spark and Gluten's code, to reduce the risk of JVM heap OOMs, and to allow users to be able to configure a relatively smaller JVM heap size when Gluten is used.

Gluten: 

- https://github.com/apache/incubator-gluten/issues/11133 

Spark:

Tracked by https://issues.apache.org/jira/browse/SPARK-54089.

### Gluten version

None

PR #11236, opened 2025-12-01
[GLUTEN-11169][VL] Trigger GC before throwing OOM to ensure unused off-heap broadcasted relations are correctly released

zhztheplayer:
Off-heap relations used in broadcast hash join rely on JVM GC to release their off-heap memory blocks.

We should trigger a GC before throwing any off-heap OOM to make sure we have checked and released all the unused off-heap broadcast relations.

The PR also adds a retry mechanism before OOM is thrown, up to 9 numbers of retries or totally 1000ms, depending on which one comes first.

Related issue: #11169


zhztheplayer:
cc @zjuwangg @JkSelf @FelixYBW

zhztheplayer:
Thanks @zjuwangg.

Issue #11251, opened 2025-12-04
[CORE] Incorrect wholestage id in WholeStageTransformerExec


marin-ma:
### Backend

VL (Velox)

### Bug description

In vanilla spark, the stage id of the `WholeStageCodegenExec` always starts with 1 in each SQL query
<img width="858" height="908" alt="Image" src="https://github.com/user-attachments/assets/3ca3dff2-18c9-4581-bc89-5bb3b47df842" />

However in Gluten, the stage id is generated from a global counter, which accumulates across different queries. This number can grow very large in a long running spark session.
<img width="749" height="919" alt="Image" src="https://github.com/user-attachments/assets/1835f4af-79b7-47e6-b42c-8c4f430f4269" />

### Spark version

None

### Relevant logs

```bash

```

PR #11252, opened 2025-12-04
[GLUTEN-11251] Fix incorrect whole stage id in WholeStageTransformerExec

marin-ma:
Fix the transformer stage id contiguously increasing across different sql queries.

When AQE is off, the fix can be directly applied.

When AQE is on, vanillas spark set the rule `CollapseCodegenStages` to be stateful:
https://github.com/apache/spark/blob/branch-4.0/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L153-L156

However in Gluten, when AQE is on, the columnar rules are applied upon each individual query stages, and there's no stateful context shared across query stages that is visible to the columnar rules. 

Seems like there's no way to use a stateful counter to generate incremental ids across different query stages. This pr adds a new rule to update the stage id after the physical plan is generated.

Related issue: #11251

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
@@ -139,6 +139,7 @@ object VeloxRuleApi {
       .getExtendedColumnarPostRules()
       .foreach(each => injector.injectPost(c => each(c.session)))
     injector.injectPost(c => ColumnarCollapseTransformStages(new GlutenConfig(c.sqlConf)))
+    injector.injectPost(_ => GenerateTransformStageId())
     injector.injectPost(c => CudfNodeValidationRule(new GlutenConfig(c.sqlConf)))
 
     injector.injectPost(c => GlutenNoopWriterRule(c.session))
@@ -151,7 +152,6 @@ object VeloxRuleApi {
       c => GlutenAutoAdjustStageResourceProfile(new GlutenConfig(c.sqlConf), c.session))
     injector.injectFinal(c => GlutenFallbackReporter(new GlutenConfig(c.sqlConf), c.session))
     injector.injectFinal(_ => RemoveFallbackTagRule())
-    injector.injectFinal(_ => GenerateTransformStageId())
   }
 
   /**
@@ -241,6 +241,7 @@ object VeloxRuleApi {
       .getExtendedColumnarPostRules()
       .foreach(each => injector.injectPostTransform(c => each(c.session)))
     injector.injectPostTransform(c => ColumnarCollapseTransformStages(new GlutenConfig(c.sqlConf)))
+    injector.injectPostTransform(_ => GenerateTransformStageId())
     injector.injectPostTransform(c => CudfNodeValidationRule(new GlutenConfig(c.sqlConf)))
     injector.injectPostTransform(c => GlutenNoopWriterRule(c.session))
     injector.injectPostTransform(c => RemoveGlutenTableCacheColumnarToRow(c.session))
@@ -251,6 +252,5 @@ object VeloxRuleApi {
     injector.injectPostTransform(
       c => GlutenFallbackReporter(new GlutenConfig(c.sqlConf), c.session))
     injector.injectPostTransform(_ => RemoveFallbackTagRule())
-    injector.injectPostTransform(_ => GenerateTransformStageId())
   }
 }


marin-ma:
@zhztheplayer @zhouyuan Could you help to review? Thanks!

gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -155,7 +155,7 @@ trait UnaryTransformSupport extends TransformSupport with UnaryExecNode {
 }
 
 case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = false)(
-    val transformStageId: Int
+    var transformStageId: Int
```
zhztheplayer, line 158: Why changing to `var`?

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/RegenerateTransformStageId.scala
```diff
@@ -0,0 +1,78 @@
...
+
+import scala.collection.mutable
+
+/**
+ * Regenerate `transformStageId` for `WholeStageTransformerExec` when AQE is on. QueryStages are
+ * invisible to each other during execution. During runtime, the transformStageId of each QueryStage
+ * starts from 1. This rule updates the whole plan tree with incremental and unique transform stage
+ * id before the final execution.
+ */
+case class RegenerateTransformStageId() extends Rule[SparkPlan] with AdaptiveSparkPlanHelper {
```
zhztheplayer: Have you compared with vanilla Spark's query planner? I assume Spark doesn't need such rule for the correctness of the stage IDs in AQE?

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/ColumnarCollapseTransformStages.scala
```diff
@@ -141,7 +141,7 @@ case class InputIteratorTransformer(child: SparkPlan) extends UnaryTransformSupp
  */
 case class ColumnarCollapseTransformStages(
     glutenConf: GlutenConfig,
-    transformStageCounter: AtomicInteger = ColumnarCollapseTransformStages.transformStageCounter)
+    transformStageCounter: AtomicInteger = new AtomicInteger(0))
```
zhztheplayer: 👍

gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
marin-ma: When AQE is on, `transformStageId` always starts at 0 within each query stage. In this case `transformStageId` is not globally unique within the whole plan tree. After all query stages have been executed, the `RegenerateTransformStageId` rule traverses the whole plan tree and updates this value incrementally to a unique id.

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/RegenerateTransformStageId.scala
marin-ma: Spark maintains a global counter in rule `CollapseCodegenStages` inside of the `AdaptiveSparkPlanExec` to make sure it generates unique id across query stages. https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/AdaptiveSparkPlanExec.scala#L147-L150


But for Gluten we only can use a new counter for each query stage.

zhztheplayer: Understood. Though I feel the code of ID generation gets much harder to reason about after the PR's change. 

Do you think it's feasible to change `ColumnarCollapseTransformStages` to leave all stage IDs unassigned (e.g., -1), then rely on `RegenerateTransformStageId` to assign all IDs, both for AQE and non-AQE cases? So it's clear that only `RegenerateTransformStageId` manages these IDs.

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/RegenerateTransformStageId.scala
```diff
@@ -0,0 +1,78 @@
...
+  private val transformStageCounter: AtomicInteger = new AtomicInteger(0)
+
+  private val wholeStageTransformerCache =
+    new mutable.HashSet[WholeStageTransformer]()
```
zhztheplayer: `HashSet[WholeStageTransformer]()` is expensive. Can we use `IdentityHashMap` or so?

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/RegenerateTransformStageId.scala
marin-ma: > Do you think it's feasible to change ColumnarCollapseTransformStages to leave all stage IDs unassigned (e.g., -1), then rely on RegenerateTransformStageId to assign all IDs, both for AQE and non-AQE cases? So it's clear that only RegenerateTransformStageId manages these IDs.

Yes make sense.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
```diff
@@ -250,5 +251,6 @@ object VeloxRuleApi {
     injector.injectPostTransform(
       c => GlutenFallbackReporter(new GlutenConfig(c.sqlConf), c.session))
     injector.injectPostTransform(_ => RemoveFallbackTagRule())
+    injector.injectPostTransform(_ => GenerateTransformStageId())
```
zhztheplayer: Can we move the rule right after ColumnarCollapseTransformStages? To make it clear that `GenerateTransformStageId` doesn't depend on other rules.

zhztheplayer (approved):
Thanks. Only a minor left from my end.

PR #11257, opened 2025-12-05
[VL] Enable `SIMDJSON_SKIPUTF8VALIDATION` for simdjson build in setup script

philo-he:
This build option is required to get one Gluten UT passed. It has been enabled in vcpkg build, but not in setup script. If the lib installed by setup script is used in Velox build, that Gluten UT will fail.

With the following Velox patch merged, we can just enable the build option by setting the environment variable .
https://github.com/facebookincubator/velox/commit/868f27959c7ba2243dc446a48401bb7c87d586a4

## How was this patch tested?
Local verification.


philo-he:
@baibaichen, could you take a look?

zhouyuan (approved):

PR #11264, opened 2025-12-07
[VL] Correct the conversion from `AggregateRel.groupings` to `AggregationNode.groupingKeys`

Zouxxyy:
Currently, Velox's `AggregationNode` only supports a single grouping keys list,

```cpp
AggregationNode(
    const PlanNodeId& id,
    Step step,
    const std::vector<FieldAccessTypedExprPtr>& groupingKeys,
    const std::vector<FieldAccessTypedExprPtr>& preGroupedKeys,
    const std::vector<std::string>& aggregateNames,
    const std::vector<Aggregate>& aggregates,
    bool ignoreNullKeys,
    PlanNodePtr source);
```

whereas the groupings field in `AggregateRel` is actually a list of grouping keys list.

```
message AggregateRel {
  ...
  repeated Grouping groupings = 3;
  message Grouping {
    repeated Expression grouping_expressions = 1;
  }
 ...
}
```

Therefore, the correct conversion is:

- Map `AggregateRel.groupings[0]` to `AggregationNode`'s `groupingKeys`,
- And throw an exception if the number of groupings is greater than 1.

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -463,7 +463,7 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
   std::vector<core::FieldAccessTypedExprPtr> veloxGroupingExprs;
 
   // Get the grouping expressions.
-  VELOX_FAIL(
+  VELOX_CHECK(
     aggRel.groupings().size() <= 1, "At most one grouping is supported, but got {}.", aggRel.groupings().size());
   if (aggRel.groupings().size() == 1) {
     for (const auto& groupingExpr : aggRel.groupings()[0].grouping_expressions()) {


Zouxxyy:
CC @rui-mo @PHILO-HE for a look, thanks

FelixYBW:
Original design is to put all groupings into velox group keys. 
Is there any cases we may generate more than 1 AggregateRel.groupings? Will it cause issue on original design?

Zouxxyy:
> Original design is to put all groupings into velox group keys. Is there any cases we may generate more than 1 AggregateRel.groupings? Will it cause issue on original design?

Currently, `AggregateRel.groupings` is actually a `List<List<keys>>`, representing multiple grouping sets. However, Velox's AggregationNode only supports a single grouping set.

In the current implementation of AggregateRelNode.toProtobuf, only groupings[0] = groupingKeys is populated.

I discovered this issue because we have developed an AggregationNode that supports multiple grouping sets. 

In this PR, no proto changes are made—only defensive code is added. In fact, you’ll notice that the following two nested loops look quite odd:

```c++
for (const auto& grouping : aggRel.groupings()) {
  for (const auto& groupingExpr : grouping.grouping_expressions()) {
    // Velox's groupings are limited to be Field.
    veloxGroupingExprs.emplace_back(exprConverter_->toVeloxExpr(groupingExpr.selection(), inputType));
  }
}
```

FelixYBW:
So then root cause is that Velox's AggregationNode can't use a single grouping set to simulate multiple groups as below

```
[groupings[0].key[0],groupings[0].key[1],groupings[1].key[0],groupings[1].key[1]]
```

we have to implement multiple groups support in AggregationNode. 

Right?

Zouxxyy:
> So then root cause is that Velox's AggregationNode can't use a single grouping set to simulate multiple groups as below
> 
> ```
> [groupings[0].key[0],groupings[0].key[1],groupings[1].key[0],groupings[1].key[1]]
> ```
> 
> we have to implement multiple groups support in AggregationNode.
> 
> Right?

Yes, it only accept

```
[groupings[0].key[0],groupings[0].key[1],groupings[0].key[2],groupings[0].key[3]]
```

BTY, in the case of multiple groups, the aggregates functions also need to be a list of list.

FelixYBW (approved):

PR #11275, opened 2025-12-10
[GLUTEN-11302][VL] Fix gpu build by bumping to cuda-13.1

zhouyuan:
Fix GPU build by
- switch to use `gcc-14`
- bumping to `cuda-toolkit-13.1`

The new `cuda-toolkit-13.1` requires larger disk spaces, so this patch also modified GHA to clean up the disk space firstly

## How was this patch tested?
pass GHA

fixes: #11302

### DIFF
--- a/dev/docker/cudf/Dockerfile
+++ b/dev/docker/cudf/Dockerfile
@@ -28,7 +28,10 @@ ENV CUDA_ARCHITECTURES=70
 
 
 WORKDIR /opt/gluten
-RUN rm -rf /opt/rh/gcc-toolset-12 && cp -r /opt/rh/gcc-toolset-14 /opt/rh/gcc-toolset-12; \
+RUN rm -rf /opt/rh/gcc-toolset-12 && ln -s /opt/rh/gcc-toolset-14 /opt/rh/gcc-toolset-12; \
+    dnf remove -y cuda-toolkit-12* && dnf install -y cuda-toolkit-13-1; \
+    dnf autoremove -y && dnf clean all; \
+    source /opt/rh/gcc-toolset-12/enable; \
     bash ./dev/buildbundle-veloxbe.sh --run_setup_script=OFF --build_arrow=ON --spark_version=3.4 --build_tests=ON --build_benchmarks=ON --enable_gpu=ON && rm -rf /opt/gluten
 
 # You can try the data in folder backends-velox/src/test/resources/tpch-data-parquet


cpp/velox/CMakeLists.txt
```diff
@@ -439,6 +439,7 @@ if(ENABLE_GPU)
             ${VELOX_BUILD_PATH}/_deps/nvtx3-src/c/include
             ${VELOX_BUILD_PATH}/_deps/nvcomp_proprietary_binary-src/include
             ${VELOX_BUILD_PATH}/_deps/rapids_logger-src/include
+            /usr/local/cuda/include/cccl
```
bdice, line 442: If possible, we should try to fix this by calling `find_package` for cudf. It should set up all these include paths. I'll try to repro this locally with @karthikeyann and make a suggestion.

We don't want to require a specific CUDA version just to get a particular CCCL version -- those don't always move in lockstep and sometimes RAPIDS requires CCCL versions that have been publicly released but are not yet shipped in a CUDA toolkit.

zhouyuan: @bdice thanks for the inputs. Yes, I think we ran into version mismatch for rapids rmm and cuda. Initially in Gluten we prepared a docker env with cuda-12.8 pre-installed, everything works and the CMake piece was targeting for the old env. However with the recent `cudf-25.12` change, It does not compile, hence I'm experimenting on how to fix this. In my local env, i will need to bump to use `cuda-13.1` otherwise there will be issues on some header definition. I also tried `cuda-12.9` and `cuda-13.0` - does not work
```
/__w/incubator-gluten/incubator-gluten/dev/../ep/build-velox/build/velox_ep/_build/release/_deps/rmm-src/cpp/include/rmm/detail/cuda_memory_resource.hpp:23:49: error: 'synchronous_resource_with' is not a member of 'cuda::mr'
   23 | inline constexpr bool resource_with = cuda::mr::synchronous_resource_with<Resource, Properties...>;
```

ep/build-velox/src/build-velox.sh
```diff
@@ -134,8 +134,8 @@ function compile {
   if [ $ENABLE_GPU == "ON" ]; then
     # the cuda default options are for Centos9 image from Meta
     echo "enable GPU support."
-    COMPILE_OPTION="$COMPILE_OPTION -DVELOX_ENABLE_GPU=ON -DVELOX_ENABLE_CUDF=ON -DCMAKE_CUDA_ARCHITECTURES=70 \
-        -DCMAKE_CUDA_COMPILER=/usr/local/cuda-12.8/bin/nvcc"
+    COMPILE_OPTION="$COMPILE_OPTION -DVELOX_ENABLE_GPU=ON -DVELOX_ENABLE_CUDF=ON -DCMAKE_CUDA_ARCHITECTURES=75 \
```
zhouyuan, line 137: for cuda-13.1 it supports 75 at minimal

rui-mo (approved):
Thanks!

cpp/velox/CMakeLists.txt
bdice: I filed https://github.com/apache/incubator-gluten/pull/11407 as a follow-up!

PR #11292, opened 2025-12-14
[CORE] Replace Component BuildInfo with String Map

Zouxxyy:

Not all Components come from a specific compilation branch—some may originate from just a JAR dependency. For example, for PaimonComponent, we might only need to know the version of the Paimon library it includes. In the future, we may add `paimon_version=xxx`

Currently, under the changes in this PR, before
```
Gluten components:
==============================================================
Components: Velox, VeloxPaimon, VeloxIceberg
Component Velox Branch: imb/velox_20251129
Component Velox Revision: 77c239c7e9b17df7f998e5022b0b005e2b35a08b
Component Velox Revision Time: 2025-11-29 11:39:25 +0800
Component VeloxPaimon Branch: N/A
Component VeloxPaimon Revision: N/A
Component VeloxPaimon Revision Time: N/A
Component VeloxIceberg Branch: N/A
Component VeloxIceberg Revision: N/A
Component VeloxIceberg Revision Time: N/A
==============================================================
```

after
```
Gluten components:
==============================================================
Component velox
  velox_branch = imb/velox_20251129
  velox_revision = 77c239c7e9b17df7f998e5022b0b005e2b35a08b
  velox_revisionTime = 2025-11-29 11:39:25 +0800
Component velox-paimon
Component velox-iceberg
==============================================================
```

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHBackend.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHBackend.scala
@@ -55,7 +55,7 @@ class CHBackend extends SubstraitBackend {
   override def info(): Map[String, String] = {
     Map(
       "ch_branch" -> CH_BRANCH,
-      "ch_revision" -> CH_COMMIT,
+      "ch_revision" -> CH_COMMIT
     )
   }
   override def iteratorApi(): IteratorApi = new CHIteratorApi


Zouxxyy:
CC @zhztheplayer for a look, thanks

zhztheplayer:
I was also thinking about a way to show these custom properties on Gluten UI, related code is [here](https://github.com/apache/incubator-gluten/blob/main/gluten-ui/src/main/scala/org/apache/spark/sql/execution/ui/GlutenAllExecutionsPage.scala).

`BuildInfo` and `GlutenBuildInfoEvent` are so similar in their purposes so it's better to unify them together then display all the relevant properties on UI.

Zouxxyy:
> I was also thinking about a way to show these custom properties on Gluten UI, related code is [here](https://github.com/apache/incubator-gluten/blob/main/gluten-ui/src/main/scala/org/apache/spark/sql/execution/ui/GlutenAllExecutionsPage.scala).
> 
> `BuildInfo` and `GlutenBuildInfoEvent` are so similar in their purposes so it's better to unify them together then display all the relevant properties on UI.

Seems we are currently printing the information from GlutenBuildInfoEvent in the Gluten UI. Meanwhile, some of the configurations within it are used by Components, and this information is only logged in the logs.  Perhaps we could organize it like this, for example: "Build Information", "Components", and "Queries".

zhztheplayer:
Yes. `GlutenBuildInfoEvent` was used for UI representation but it's incomplete regarding these additional component KV values. 

> Perhaps we could organize it like this, for example: "Build Information", "Components", and "Queries".

Sounds great to me.

PR #11295, opened 2025-12-14
[CORE] Get correct fallback reason on nodes without logicalLink

liujiayi771:

This PR adds a fallback mechanism to attempt retrieving the fallback reason from a SparkPlan node's tags when its `logicalLink` is unavailable. Some SparkPlan nodes, such as the `SortExec` generated by Spark's `EnsureRequirements` rule, inherently lack a `logicalLink`. Furthermore, some plan conversion rules in Gluten might omit calling `copyTagsFrom`, leading to the loss of the original `logicalLink`.

In such cases, if the node falls back, the fallback reason might be stored within the physical node's tags. This change attempts to retrieve the reason from there as a best-effort approach.

## How was this patch tested?
Add a test case when window sort fallback. Before this change, the fallback reason was a generic "Gluten does not touch it or does not support it". After the change, it can display a more precise reason, such as "[FallbackByUserOptions] Validation failed on node Sort", which is retrieved from the physical node's tags.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
@@ -17,12 +17,17 @@
 package org.apache.gluten.execution
 
 import org.apache.gluten.config.{GlutenConfig, VeloxConfig}
+import org.apache.gluten.events.GlutenPlanFallbackEvent
 
 import org.apache.spark.SparkConf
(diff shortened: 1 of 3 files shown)


liujiayi771:
Before:
iShot_2025-12-14_18 34 39

After:
iShot_2025-12-14_18 06 29

liujiayi771:
cc @zhztheplayer @ulysses-you

PR #11304, opened 2025-12-16
[GLUTEN-11316][VL] Enable JDK-21 for Spark-4.0

zhouyuan:

JDK-21 is supported since Spark-4.0. This patch adds:
-  the new java-21 profile
-  tpc tests
- necessary dependencies update to support java-21

## How was this patch tested?
new tpc tests w/ java- 21, pass old GHA

Related issue: #11316

### DIFF
--- a/tools/gluten-it/pom.xml
+++ b/tools/gluten-it/pom.xml
@@ -321,7 +321,7 @@
       <id>spark-4.0</id>
       <properties>
         <spark.version>4.0.1</spark.version>
-        <scala.version>2.13.8</scala.version>
+        <scala.version>2.13.16</scala.version>
         <scala.binary.version>2.13</scala.binary.version>
         <delta.package.name>delta-spark</delta.package.name>
         <delta.version>4.0.0</delta.version>


pom.xml
```diff
@@ -1055,6 +1056,18 @@
...
+    <profile>
+      <id>java-21</id>
+      <activation>
+        <jdk>21</jdk>
+      </activation>
+      <properties>
+        <java.version>21</java.version>
+        <caffeine.version>3.1.8</caffeine.version>
+        <spotless.version>2.44.2</spotless.version>
+        <googlejavaformat.version>1.17.0</googlejavaformat.version>
```
zhouyuan, line 1068: need to bump to support java-21

backends-velox/src-iceberg-spark34/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
```diff
@@ -1994,11 +1994,13 @@ public void testFilesTablePartitionId() {
         .mode("append")
         .save(loadLocation(tableIdentifier));
 
+    // spotless:off
```
zhouyuan, line 1997: new google java format introduced conflicts check with old version

gluten-core/src/main/java/org/apache/gluten/memory/MemoryUsageRecorder.java
```diff
@@ -17,9 +17,11 @@
 package org.apache.gluten.memory;
 
 public interface MemoryUsageRecorder extends MemoryUsageStatsBuilder {
+  // spotless:off
```
zhouyuan, line 20: new google java format introduced conflicts check with old version

tools/gluten-it/pom.xml
```diff
@@ -262,6 +263,17 @@
...
     </profile>
+    <profile>
+      <id>java-21</id>
+      <activation>
+        <jdk>21</jdk>
+      </activation>
+      <properties>
+        <java.version>21</java.version>
+        <spotless.version>2.44.2</spotless.version>
+        <googlejavaformat.version>1.17.0</googlejavaformat.version>
```
zhouyuan, line 274: to support java-21

baibaichen:
LGTM

backends-velox/src-iceberg-spark34/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
jinchengchenghh: The spark40 changes the google format, but this folder only compiles on Spark3.4, why this file is changed?

zhouyuan: It seems there is a wrong depdency issue, I'm testing with below cmd and it compiles the iceberg-test also. 
`mvn clean install -Pbackends-velox -Pspark-4.0 -Pjava-21 -Pscala-2.13 -DskipTests -Dmaven.source.skip`

zhouyuan:
@jinchengchenghh @baibaichen thanks.

backends-velox/src-iceberg-spark34/test/java/org/apache/iceberg/spark/source/TestIcebergSourceTablesBase.java
zhouyuan: @jinchengchenghh 

In spotless check it's using regex to match all the test code:
https://github.com/apache/incubator-gluten/blob/main/pom.xml#L717

so even if we dont enable spark-34 profile it will also do the style check always. in this patch we introduced a new format in java-21 profile so it introduced some conflicts. I fired one issue to track on this:
https://github.com/apache/incubator-gluten/issues/11321

PR #11326, opened 2025-12-23
[GLUTEN-10933][VL]fix: Fix the CudfVector childreSize is not correct

jinchengchenghh:
After this PR https://github.com/facebookincubator/velox/pull/15629, the CudfVector childrenSize is not correct, if correct it, will cause many unit tests fail as this discussion https://github.com/facebookincubator/velox/pull/15629/files#r2572687025, use a work around in Gluten to fix it.

Related issue: #10933

### DIFF
--- a/cpp/velox/memory/VeloxColumnarBatch.h
+++ b/cpp/velox/memory/VeloxColumnarBatch.h
@@ -29,6 +29,13 @@ class VeloxColumnarBatch final : public ColumnarBatch {
   VeloxColumnarBatch(facebook::velox::RowVectorPtr rowVector)
       : ColumnarBatch(rowVector->childrenSize(), rowVector->size()), rowVector_(rowVector) {}
 
+#ifdef GLUTEN_ENABLE_GPU
+  // The batch may be CudfVector, it's childrenSize is not correct, many tests failed if correcting the value
+  // https://github.com/facebookincubator/velox/pull/15629#discussion_r2581655771
+  VeloxColumnarBatch(facebook::velox::RowVectorPtr rowVector, int32_t numColumns)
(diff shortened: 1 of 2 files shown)


cpp/velox/memory/VeloxColumnarBatch.h
```diff
@@ -29,6 +29,13 @@ class VeloxColumnarBatch final : public ColumnarBatch {
   VeloxColumnarBatch(facebook::velox::RowVectorPtr rowVector)
       : ColumnarBatch(rowVector->childrenSize(), rowVector->size()), rowVector_(rowVector) {}
 
+#ifdef GLUTEN_ENABLE_GPU
+  // The batch may be CudfVector, it's childrenSize is not correct, many tests failed if correcting the value
+  // https://github.com/facebookincubator/velox/pull/15629#discussion_r2581655771
+  VeloxColumnarBatch(facebook::velox::RowVectorPtr rowVector, int32_t numColumns)
+      : ColumnarBatch(numColumns, rowVector->size()), rowVector_(rowVector) {}
+#endif
```
zhztheplayer, line 37: Compared to adding a new ctor, what if we just change the old ctor to call `rowVector->rowType()->size()`?

By the way. Why `childrenSize()` is inaccurate for cuDF row vector?

jinchengchenghh: Because the childrenSize() always returns 0, RowVector contains flatVector, but CudfVector contains cudf::table(contains cudf::column)

jinchengchenghh: As this discussion https://github.com/facebookincubator/velox/pull/15629/files#r2572687025

Issue #11330, opened 2025-12-25
[VL] IllegalStateException thrown when accessing null values of map in partial project


jiangjiangtian:
### Backend

VL (Velox)

### Bug description

The exception is as follows:
```
Reason: Operator::getOutput failed for [operator: ValueStream, plan node ID: 0]: Error during calling Java code from native code: java.lang.IllegalStateException: Value at index is null
	at org.apache.gluten.shaded.org.apache.arrow.vector.BigIntVector.get(BigIntVector.java:105)
	at org.apache.gluten.vectorized.ArrowWritableColumnVector$LongAccessor.getLong(ArrowWritableColumnVector.java:989)
	at org.apache.gluten.vectorized.ArrowWritableColumnVector.getLong(ArrowWritableColumnVector.java:639)
	at org.apache.spark.sql.vectorized.ColumnarArray.getLong(ColumnarArray.java:133)
	at org.apache.spark.sql.catalyst.expressions.SpecializedGettersReader.read(SpecializedGettersReader.java:49)
	at org.apache.spark.sql.vectorized.ColumnarArray.get(ColumnarArray.java:180)
	at org.apache.spark.sql.catalyst.util.MapData.foreach(MapData.scala:38)
	at org.apache.spark.sql.hive.HiveInspectors.$anonfun$wrapperFor$49(HiveInspectors.scala:433)
	at org.apache.spark.sql.hive.HiveInspectors.$anonfun$withNullSafe$1(HiveInspectors.scala:262)
	at org.apache.spark.sql.hive.HiveSimpleUDFEvaluator.setArg(hiveUDFEvaluators.scala:92)
	at org.apache.spark.sql.hive.HiveSimpleUDF.$anonfun$eval$1(hiveUDFs.scala:67)
	at org.apache.spark.sql.hive.HiveSimpleUDF.$anonfun$eval$1$adapted(hiveUDFs.scala:66)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at org.apache.spark.sql.hive.HiveSimpleUDF.eval(hiveUDFs.scala:66)
	at org.apache.spark.sql.catalyst.expressions.Alias.eval(namedExpressions.scala:158)
[...]

PR #11331, opened 2025-12-25
[GLUTEN-11330][VL] Make PartialProject support array and map with null values

jiangjiangtian:
This PR introduces a new class named `ArrowColumnarArray`. Its implementation is copied from Spark-4.0, except that the `handleNull` parameter is set to true when we call `SpecializedGettersReader.read` in `get`, which means that when trying to access a value of array, we will check whether the value to get is null first. So we can avoid throwing exception when we try to access a null value of array.
Besides, this PR introduces another new class named `ArrowColumnarMap`. This class defines two fields of type `ArrowColumnarArray` to represent keys and values, separately. With this class, we can also avoid throwing exception when we try to access a null value of map.

## How was this patch tested?
unit tests.

Related issue: #11330

### DIFF
--- a/gluten-arrow/src/main/java/org/apache/gluten/vectorized/ArrowColumnarArray.java
+++ b/gluten-arrow/src/main/java/org/apache/gluten/vectorized/ArrowColumnarArray.java
@@ -16,208 +16,21 @@
  */
 package org.apache.gluten.vectorized;
 
-import org.apache.spark.sql.catalyst.expressions.SpecializedGettersReader;
-import org.apache.spark.sql.catalyst.expressions.UnsafeArrayData;
-import org.apache.spark.sql.catalyst.util.ArrayData;
-import org.apache.spark.sql.catalyst.util.GenericArrayData;
-import org.apache.spark.sql.types.*;
+import org.apache.spark.sql.execution.vectorized.ColumnarArrayShim;
 import org.apache.spark.sql.vectorized.ColumnVector;
-import org.apache.spark.sql.vectorized.ColumnarArray;
-import org.apache.spark.sql.vectorized.ColumnarMap;
-import org.apache.spark.sql.vectorized.ColumnarRow;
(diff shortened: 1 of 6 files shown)


gluten-arrow/src/main/java/org/apache/gluten/vectorized/ArrowColumnarArray.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gluten.vectorized;
+
+import org.apache.spark.sql.execution.vectorized.ColumnarArrayShim;
+import org.apache.spark.sql.vectorized.ColumnVector;
+
+/**
+ * Because `get` method in `ColumnarArray` don't check whether the data to get is null and arrow
```
jinchengchenghh, line 23: Is this the Spark shortage or design? What's the Spark usage for ColumnarArray with null value?

jiangjiangtian: I think it is Spark shortage. If there exists null values, ColumnarArray will get the value(this might be a default value or previously set value) because call `get` on `ColumnarArray` will eventually call `getXXX` on `ColumnVector` and `getXXX` will not check if it is null value, either.

jinchengchenghh: Could you also raise an issue in Spark?

jinchengchenghh (approved):

gluten-arrow/src/main/java/org/apache/gluten/vectorized/ArrowColumnarArray.java
jiangjiangtian: OK, I will. Thanks.

PR #11358, opened 2026-01-05
[VL] Do not rewriteAggBufferAttributes for Complete agg mode

Zouxxyy:

rewriteAggBufferAttributes is only needed by PartialMerge and Final mode

## How was this patch tested?

test by the existing cases

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
@@ -284,7 +284,7 @@ abstract class HashAggregateExecTransformer(
       val aggFunc = aggregateExpression.aggregateFunction
       val functionInputAttributes = aggFunc.inputAggBufferAttributes
       aggFunc match {
-        case _ if aggregateExpression.mode == Partial =>
+        case _ if aggregateExpression.mode == Partial || aggregateExpression.mode == Complete =>
           val childNodes = aggFunc.children
             .map(
               ExpressionConverter
@@ -303,7 +303,7 @@ abstract class HashAggregateExecTransformer(
           // The process of handling the inconsistency in column types and order between
           // Spark and Velox is exactly the opposite of applyExtractStruct.
           aggregateExpression.mode match {
-            case PartialMerge | Final | Complete =>
+            case PartialMerge | Final =>
               val newInputAttributes = new ArrayBuffer[Attribute]()
               val childNodes = new JArrayList[ExpressionNode]()
               val (sparkOrders, sparkTypes) =


rui-mo (commented):
Could you please clarify if this change fixes an existing problem? Thanks.

Zouxxyy:
@rui-mo During our implementation of hybrid AGG (where a single AGG contains both partial-merge and complete modes), we discovered this issue:
Inside `rewriteAggBufferAttributes`, the code compares `aggBufferAttrs` with `InputAttributes`, which is actually incorrect in Complete mode. However, since Spark + Gluten currently does not support having both partial-merge and complete functions simultaneously, the current code path will not trigger this scenario, and it's difficult to construct a unit test (UT) for it. Nevertheless, I think we can add this defensive code.

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
rui-mo, line 268: Why would this be called for the Complete mode, as `rowConstructNeeded()` returns false for it?

Zouxxyy: > Why would this be called for the Complete mode, as `rowConstructNeeded()` returns false for it?

As mentioned earlier, we implemented a hybrid AGG that contains both PartialMerge (or Final) and complete aggregation functions, and `rowConstructNeeded` is triggered as long as a PartialMerge (or Final) function is present.

It won't be triggered for the now code, but I think we can add defensive code to handle this scenario in future implementations .

rui-mo: > rowConstructNeeded is triggered as long as a PartialMerge (or Final) function is present

Are you referring to the modified code on your side, or the current Gluten code?

> As mentioned earlier, we implemented a hybrid AGG that contains both PartialMerge (or Final) and complete aggregation functions

Is there any plan to upstream this change to Gluten, or will it remain local to your side?

Zouxxyy: > Are you referring to the modified code on your side, or the current Gluten code?

The current Gluten code, see this
```scala
  private def rowConstructNeeded(): Boolean = aggregateExpressions.exists {
    case AggregateExpression(aggFunc, PartialMerge | Final, _, _, _) =>
      aggFunc.inputAggBufferAttributes.size > 1
    case _ => false
  }
```
> Is there any plan to upstream this change to Gluten, or will it remain local to your side?

Oh, maybe it can be launched when it becomes stable!

rui-mo (approved):

backends-velox/src/main/scala/org/apache/gluten/execution/HashAggregateExecTransformer.scala
```diff
@@ -284,7 +284,7 @@ abstract class HashAggregateExecTransformer(
       val aggFunc = aggregateExpression.aggregateFunction
       val functionInputAttributes = aggFunc.inputAggBufferAttributes
       aggFunc match {
-        case _ if aggregateExpression.mode == Partial =>
+        case _ if aggregateExpression.mode == Partial || aggregateExpression.mode == Complete =>
```
rui-mo, line 287: I understand that mixing Complete mode with other modes isn’t supported, but is there a way to add a unit test for pure Complete mode to ensure this works?

Zouxxyy: We already have pure complete mode case, see https://github.com/apache/incubator-gluten/pull/7752, and it can run correctly

PR #11362, opened 2026-01-05
[VL] Fix memory leak caused by hard references on keys of TreeMemoryConsumers#FACTORIES

zhztheplayer:
https://github.com/apache/incubator-gluten/pull/4145 was not correctly carried over during the refactor in PR https://github.com/apache/incubator-gluten/pull/8254.

This is to fix the potential memory leak following the solution in #4145.


zhztheplayer:
Thanks for reviewing @jinchengchenghh

Issue #9619, opened 2025-05-14
[VL] Adding integration tests for HDFS


ashahba:
### Description

This is sub-issue for #9184 

### Gluten version

main branch

infvg:
Raised a PR: https://github.com/apache/incubator-gluten/pull/11373

Issue #9618, opened 2025-05-14
[VL] Adding integration tests for S3


ashahba:
### Backend

VL (Velox)

### Bug description

This is a sub-issue for: #9184 

### Spark version

None

### Relevant logs

```bash

```

philo-he:
@mhbuehler created a draft patch: https://github.com/mhbuehler/incubator-gluten/pull/1/files. Thank you, @mhbuehler, for your work.

We can continue this work, maybe based on this patch. cc @zhouyuan

zhouyuan:
@mhbuehler @PHILO-HE thanks for improving on this. Will refine and make a pull request on this.

Mariamalmesfer:
take

PR #11373, opened 2026-01-06
[GLUTEN-9619][VL] Add HDFS integration tests to gluten-it

infvg:
Added HDFS integration tests to gluten-it & added them to the velox github workflow

## How was this patch tested?
I ran a hdfs server locally & tested the new command. I added the tests to the github workflow.

Resolves: 
https://github.com/apache/incubator-gluten/issues/9619

### DIFF
--- a/.github/workflows/velox_backend_x86.yml
+++ b/.github/workflows/velox_backend_x86.yml
@@ -181,6 +181,11 @@ jobs:
             apt remove openjdk-11* -y
           fi
           ls -l /root/.m2/repository/org/apache/arrow/arrow-dataset/15.0.0-gluten/
+      - name: Install Hadoop & Setup HDFS
+        if: matrix.os == 'ubuntu:22.04' && matrix.spark == 'spark-3.5' && matrix.java == 'java-8'
+        run: |
+          export JAVA_HOME=/usr/lib/jvm/${{ matrix.java }}-openjdk-amd64
+          bash .github/workflows/util/setup-helper.sh install_hadoop setup_hdfs
       - name: Build and run TPC-H / TPC-DS
         run: |
           cd $GITHUB_WORKSPACE/
@@ -200,6 +205,13 @@ jobs:
             --local --preset=velox --benchmark-type=h --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1 \
           && GLUTEN_IT_JVM_ARGS=-Xmx6G sbin/gluten-it.sh queries-compare \
             --local --preset=velox --benchmark-type=ds --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1
+          if [ "${{ matrix.os }}" = "ubuntu:22.04" ] && \
+            [ "${{ matrix.spark }}" = "spark-3.5" ] && \
+            [ "${{ matrix.java }}" = "java-8" ]; then
+            GLUTEN_IT_JVM_ARGS=-Xmx6G sbin/gluten-it.sh queries-compare \
+            --local --preset=velox --benchmark-type=h --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1 \
+            --queries=q1 --data-dir="hdfs://localhost:9000/test"
+          fi
 
   tpc-test-centos8:
     needs: build-native-lib-centos-7
--- a/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/QueryRunner.scala
+++ b/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/QueryRunner.scala
@@ -22,14 +22,16 @@ import org.apache.spark.sql.{RunResult, SparkQueryRunner, SparkSession}
 
 import com.google.common.base.Preconditions
 import org.apache.commons.lang3.exception.ExceptionUtils
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
 
 import java.io.File
+import java.net.URI
 
 class QueryRunner(val source: String, val dataPath: String) {
   import QueryRunner._
-
   Preconditions.checkState(
-    new File(dataPath).exists(),
+    fileExists(dataPath),
     s"Data not found at $dataPath, try using command `<gluten-it> data-gen-only <options>` to generate it first.",
     Array(): _*)
 
@@ -62,6 +64,13 @@ class QueryRunner(val source: String, val dataPath: String) {
         println(s"Error running query ${query.id}. Error: ${ExceptionUtils.getStackTrace(e)}")
         Failure(query.id, e)
     }
+  }
+
+  private def fileExists(datapath: String): Boolean = {
+    if (datapath.startsWith("hdfs:")) {
+      val uri = URI.create(datapath)
+      FileSystem.get(uri, new Configuration()).exists(new Path(uri.getPath))
+    } else new File(datapath).exists()
   }
 }
 
--- a/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/TableCreator.scala
+++ b/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/TableCreator.scala
@@ -18,9 +18,13 @@ package org.apache.gluten.integration
 
 import org.apache.spark.sql.{AnalysisException, SparkSession}
 
-import java.io.File
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
+
+import java.net.URI
 
 import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
 
 trait TableCreator {
   def create(spark: SparkSession, source: String, dataPath: String): Unit
@@ -33,44 +37,54 @@ object TableCreator {
 
   private object DiscoverSchema extends TableCreator {
     override def create(spark: SparkSession, source: String, dataPath: String): Unit = {
-      val files = new File(dataPath).listFiles()
-      val tableNames = files.map(_.getName)
+      val uri = URI.create(dataPath)
+      val fs = FileSystem.get(uri, new Configuration())
+
+      val basePath = new Path(dataPath)
+      val statuses = fs.listStatus(basePath)
+
+      val tableDirs = statuses.filter(_.isDirectory).map(_.getPath)
+
+      val tableNames = ArrayBuffer[String]()
+
       val existedTableNames = mutable.ArrayBuffer[String]()
       val createdTableNames = mutable.ArrayBuffer[String]()
       val recoveredPartitionTableNames = mutable.ArrayBuffer[String]()
 
-      if (tableNames.isEmpty) {
-        return
+      tableDirs.foreach {
+        tablePath =>
+          val tableName = tablePath.getName
+          tableNames += tableName
       }
 
       println("Creating catalog tables: " + tableNames.mkString(", "))
 
-      files.foreach(
-        file => {
-          val tableName = file.getName
+      tableDirs.foreach {
+        tablePath =>
+          val tableName = tablePath.getName
           if (spark.catalog.tableExists(tableName)) {
             existedTableNames += tableName
           } else {
-            spark.catalog.createTable(tableName, file.getAbsolutePath, source)
+            spark.catalog.createTable(tableName, tablePath.toString, source)
             createdTableNames += tableName
             try {
               spark.catalog.recoverPartitions(tableName)
               recoveredPartitionTableNames += tableName
             } catch {
               case _: AnalysisException =>
-              // Swallows analysis exceptions.
             }
           }
-        })
+      }
 
+      if (tableNames.isEmpty) {
+        return
+      }
       if (existedTableNames.nonEmpty) {
         println("Tables already exists: " + existedTableNames.mkString(", "))
       }
-
       if (createdTableNames.nonEmpty) {
         println("Tables created: " + createdTableNames.mkString(", "))
       }
-
       if (recoveredPartitionTableNames.nonEmpty) {
         println("Recovered partition tables: " + recoveredPartitionTableNames.mkString(", "))
       }
--- a/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/ds/TpcdsSuite.scala
+++ b/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/ds/TpcdsSuite.scala
@@ -81,6 +81,9 @@ class TpcdsSuite(
       "non_partitioned"
     }
     val featureFlags = dataGenFeatures.map(feature => s"-$feature").mkString("")
+    if (dataDir.startsWith("hdfs://")) {
+      return s"$dataDir/$TPCDS_WRITE_RELATIVE_PATH-$dataScale-$dataSource-$partitionedFlag$featureFlags"
+    }
     new File(dataDir).toPath
       .resolve(s"$TPCDS_WRITE_RELATIVE_PATH-$dataScale-$dataSource-$partitionedFlag$featureFlags")
       .toFile
--- a/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/h/TpchSuite.scala
+++ b/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/h/TpchSuite.scala
@@ -76,6 +76,9 @@ class TpchSuite(
 
   override private[integration] def dataWritePath(): String = {
     val featureFlags = dataGenFeatures.map(feature => s"-$feature").mkString("")
+    if (dataDir.startsWith("hdfs://")) {
+      return s"$dataDir/$TPCH_WRITE_RELATIVE_PATH-$dataScale-$dataSource$featureFlags"
+    }
     new File(dataDir).toPath
       .resolve(s"$TPCH_WRITE_RELATIVE_PATH-$dataScale-$dataSource$featureFlags")
       .toFile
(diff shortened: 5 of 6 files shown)


.github/workflows/velox_backend_arm.yml
```diff
@@ -120,21 +120,75 @@ jobs:
...
-      - name: Run TPC-H / TPC-DS
+      - name: Run TPC-H / TPC-DS / HDFS
         run: |
           echo "JAVA_HOME: $JAVA_HOME"
           cd $GITHUB_WORKSPACE/tools/gluten-it
           GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
             --local --preset=velox --benchmark-type=h --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1 \
           && GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
             --local --preset=velox --benchmark-type=ds --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1
+          && GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
```
zhouyuan: This will introduce bigger test matrix size, better to use a single standalone job

.github/workflows/velox_backend_arm.yml
```diff
@@ -120,21 +120,75 @@ jobs:
       - name: Set environment variables
         run: |
           echo "JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk" >> $GITHUB_ENV
+      - name: Install Hadoop
+        run: |
+          yum install -y tar gzip curl which hostname
+          HADOOP_VERSION=3.3.6
+          curl -L -o /tmp/hadoop.tar.gz https://archive.apache.org/dist/hadoop/common/hadoop-${HADOOP_VERSION}/hadoop-${HADOOP_VERSION}.tar.gz
```
zhouyuan: better to move these to a new helper script

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/hdfs/HdfsTableCreator.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileSystem, Path}
+
+import java.net.URI
+
+import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
+
+class HdfsTableCreator extends TableCreator {
```
marin-ma: This can be `object HdfsTableCreator`

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/hdfs/HdfsSuite.scala
```diff
@@ -0,0 +1,109 @@
...
+    hsUiPort,
+    disableAqe,
+    disableBhj,
+    disableWscg,
+    shufflePartitions,
+    scanPartitions,
+    decimalAsDouble,
+    baselineMetricMapper,
+    testMetricMapper
+  ) {
```
zhztheplayer: Thank you for working on this @infvg, this has been needed for a long time.

Compared to adding a new `HdfsSuite`, is it possible to directly extend the current `TpchSuite` to add HDFS support? We can follow the same pattern for `TpcdsSuite` in the future.

philo-he:
@infvg, thanks for your work. This PR will enable full TPC-H test with data on HDFS. As our purpose is only validating the functionality of reading data from HDFS, I am wondering if we can make the tests simpler and less redundant.

There is a drafted PR for S3 integration test. See https://github.com/apache/incubator-gluten/issues/9618#issuecomment-3411763360 if its test design makes sense.

infvg:
@zhztheplayer @PHILO-HE I've changed it from using the TPCH queries to just one custom query & I generated a small dataset instead. Should make it much faster & simpler, please take a look

zhztheplayer:
@infvg Thanks. My thoughts were to see whether we can directly add HDFS support to the existing suites - TpcdsSuite, TpchSuite, etc. Can you also check that?

philo-he:
> @infvg Thanks. My thoughts were to see whether we can directly add HDFS support to the existing suites - TpcdsSuite, TpchSuite, etc. Can you also check that?

@zhztheplayer, do you mean we can directly replace local file system data by HDFS data for benchmark tests? Or keep the both with most code shared?

zhztheplayer:
> > @infvg Thanks. My thoughts were to see whether we can directly add HDFS support to the existing suites - TpcdsSuite, TpchSuite, etc. Can you also check that?
> 
> @zhztheplayer, do you mean we can directly replace local file system data by HDFS data for benchmark tests? Or keep the both with most code shared?

I thought it would not be necessary to add e.g, `HdfsSuite`, if `TpchSuite` could be able to read / write data from HDFS. cc @infvg

infvg:
@zhztheplayer technically yes, we can run the Tpch suite with hdfs URL with this (if we make it use HdfsTableCreator) but if we only need to test reading/writing we don't really have to run all tpch queries

.github/workflows/velox_backend_arm.yml
```diff
@@ -195,4 +195,4 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: ${{ github.job }}-test-log
-          path: "**/target/*.log"
+          path: "**/target/*.log"
```
rui-mo: nit: Perhaps keep the empty line at the end of file.

zhztheplayer:
> if we only need to test reading/writing we don't really have to run all tpch queries

When using gluten-it, it's a common practice to specify argument `--queries=q1` if only Q1 is to be tested. Does that meet the requirement here? @infvg

.github/workflows/velox_backend_arm.yml
infvg: Fixed

infvg:
@zhztheplayer ah I see, 
I've reworked it so that it uses tpch/tpcds but with hdfs as an optional datapath. Now we can do for example: 
`gluten-it.sh --benchmark-type=h --queries=q1 --data-dir=hdfs://127.0.0.1:9000/test`

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/hdfs/HdfsTableCreator.scala
```diff
@@ -0,0 +1,71 @@
...
+    val basePath = new Path(dataPath)
+    val statuses = fs.listStatus(basePath)
+
+    val tableDirs = statuses.filter(_.isDirectory).map(_.getPath)
+
+    val tableNames = ArrayBuffer[String]()
+
+    val existedTableNames = mutable.ArrayBuffer[String]()
+    val createdTableNames = mutable.ArrayBuffer[String]()
+    val recoveredPartitionTableNames = mutable.ArrayBuffer[String]()
```
zhztheplayer: @infvg Thanks for iterating. The approach looks much better to me.

Now, can we unify `HdfsTableCreator` and the local implementation `DiscoverSchema`? The HDFS APIs you are calling in `HdfsTableCreator` should work for local files as well.

.github/workflows/velox_backend_x86.yml
```diff
@@ -1601,3 +1601,207 @@ jobs:
...
+          
+          cat > "$HADOOP_CONF_DIR/hadoop-env.sh" <<EOF
+          export JAVA_HOME=${JAVA_HOME}
+          EOF
+          
+          mkdir -p /tmp/hdfs/nn /tmp/hdfs/dn /tmp/hadoop
+          
+          "$HADOOP_HOME/bin/hdfs" namenode -format -force -nonInteractive
+          "$HADOOP_HOME/bin/hdfs" --daemon start namenode
+          "$HADOOP_HOME/bin/hdfs" --daemon start datanode
```
philo-he: I suggest moving these HDFS setup commands into setup_helper.sh located under workflows/util. This would keep the workflow file clean and make the setup logic easier to maintain.

.github/workflows/velox_backend_x86.yml
```diff
@@ -1601,3 +1601,207 @@ jobs:
...
             **/gluten-ut/**/hs_err_*.log
             **/gluten-ut/**/core.*
+  hdfs-test-ubuntu:
+    needs: build-native-lib-centos-7
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ "ubuntu:20.04", "ubuntu:22.04" ]
+        spark: [ "spark-3.2", "spark-3.3", "spark-3.4", "spark-3.5", "spark-4.0", "spark-4.1"  ]
+        java: [ "java-8", "java-11", "java-17", "java-21" ]
```
philo-he: It looks excessive to run the HDFS test with the combinations of the matrix.

Another thought. Instead of a new job, we may enable the HDFS test within the existing `tpc-test-ubuntu` job. We can target only a single matrix combination. For example: if os is ubuntu:22.04, spark is 3.5, and java is 17, we can trigger the HDFS preparation script, then run the full H/DS test with HDFS. This should be a sufficient coverage and can reduce CI resource cost.

PHILO-HE (commented):
Only two comments related to the test job. Thanks.

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/hdfs/HdfsTableCreator.scala
infvg: Replaced the DiscoverySchema create function & deleted HdfsTableCreator

.github/workflows/velox_backend_x86.yml
infvg: I replaced this with another github action to install HDFS

.github/workflows/velox_backend_x86.yml
infvg: Removed hdfs-test & included it in tpc for spark-3.5, ubuntu22.04 & java 17

.github/workflows/velox_backend_x86.yml
```diff
@@ -181,6 +181,11 @@ jobs:
             apt remove openjdk-11* -y
           fi
           ls -l /root/.m2/repository/org/apache/arrow/arrow-dataset/15.0.0-gluten/
+      - name: Setup Apache Hadoop/HDFS
+        if: matrix.os == 'ubuntu:22.04' && matrix.spark == 'spark-3.5' && matrix.java == 'java-17'
+        uses: beyondstorage/setup-hdfs@master
```
zhztheplayer: Let's avoid `@master`. Consider sepecifying a SHA instead.

.github/workflows/velox_backend_x86.yml
```diff
@@ -93,7 +93,7 @@ jobs:
       - uses: actions/upload-artifact@v4
         with:
           name: velox-native-lib-centos-7-${{github.sha}}
-          path: ./cpp/build/releases/
+          path: ./cpp/build/
```
zhztheplayer, line 96: Would you mind explaining this change a bit?

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/QueryRunner.scala
```diff
@@ -63,6 +65,13 @@ class QueryRunner(val source: String, val dataPath: String) {
         Failure(query.id, e)
     }
   }
+
+  private def fileExists(datapath: String): Boolean = {
+    if (datapath.startsWith("hdfs:")) {
+      val uri = URI.create(datapath)
+      FileSystem.get(uri, new Configuration()).exists(new Path(uri.getPath))
+    } else new File(datapath).exists()
```
zhztheplayer, line 73: It's fine to keep 2 branches here, but I would suggest doing a refactor toward a unified approach in the next PR. Let me know your thoughts on this @infvg

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/ds/TpcdsSuite.scala
```diff
@@ -81,6 +81,9 @@ class TpcdsSuite(
       "non_partitioned"
     }
     val featureFlags = dataGenFeatures.map(feature => s"-$feature").mkString("")
+    if (dataDir.startsWith("hdfs://")) {
+      return s"$dataDir/$TPCDS_WRITE_RELATIVE_PATH-$dataScale-$dataSource-$partitionedFlag$featureFlags"
+    }
```
zhztheplayer, line 86: ditto

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/h/TpchSuite.scala
```diff
@@ -76,6 +76,9 @@ class TpchSuite(
 
   override private[integration] def dataWritePath(): String = {
     val featureFlags = dataGenFeatures.map(feature => s"-$feature").mkString("")
+    if (dataDir.startsWith("hdfs://")) {
+      return s"$dataDir/$TPCH_WRITE_RELATIVE_PATH-$dataScale-$dataSource$featureFlags"
+    }
```
zhztheplayer, line 81: ditto

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/TableCreator.scala
```diff
@@ -33,47 +37,44 @@ object TableCreator {
...
-      if (existedTableNames.nonEmpty) {
-        println("Tables already exists: " + existedTableNames.mkString(", "))
-      }
-
-      if (createdTableNames.nonEmpty) {
-        println("Tables created: " + createdTableNames.mkString(", "))
       }
-
-      if (recoveredPartitionTableNames.nonEmpty) {
+      println("Tables: " + tableNames.mkString(", "))
```
zhztheplayer: Let's keep the original logging order. The `Creating catalog tables` log was intentionally paced above the catalog actions so any error will not prevent the logs from being displayed to user.

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/TableCreator.scala
```diff
@@ -33,47 +37,44 @@ object TableCreator {
...
+      if (tableNames.isEmpty) return
+      if (existedTableNames.nonEmpty)
+        println("Tables already exists: " + existedTableNames.mkString(", "))
+      if (createdTableNames.nonEmpty) println("Tables created: " + createdTableNames.mkString(", "))
+      if (recoveredPartitionTableNames.nonEmpty)
         println("Recovered partition tables: " + recoveredPartitionTableNames.mkString(", "))
```
zhztheplayer: In Gluten project, we don't usually omit braces for the if statement unless it's a functional style one. Would you recover the old style? Thanks.

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/QueryRunner.scala
infvg: I can look for a better way to do this, especially since eventually we are gonna add S3 too

.github/workflows/velox_backend_x86.yml
infvg: Rebase bug, fixed.

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/TableCreator.scala
infvg: Fixed

tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/TableCreator.scala
infvg: Returned this functionality

.github/workflows/velox_backend_x86.yml
infvg: Fixed

zhztheplayer:
@infvg I didn't see the newly added CI job triggered. Would you investigate or rebase? Thanks.

zhztheplayer:
@infvg The CI job is still not running, please take a look at https://github.com/apache/incubator-gluten/actions/runs/21170154644.

infvg:
@zhztheplayer ah I see, the hadoop action is not allowed let me find an alternative

infvg:
@zhztheplayer could you rerun the tests? I added a new HDFS setup step

.github/workflows/velox_backend_x86.yml
```diff
@@ -181,6 +181,64 @@ jobs:
...
+          
+          ssh-keygen -t rsa -P "" -f ~/.ssh/id_rsa
+          cat ~/.ssh/id_rsa.pub >> ~/.ssh/authorized_keys
+          chmod 600 ~/.ssh/authorized_keys
+          ssh -o StrictHostKeyChecking=no localhost true
+          
+          "$HADOOP_HOME/bin/hdfs" namenode -format -force
+          "$HADOOP_HOME/sbin/start-dfs.sh"
+          
+          "$HADOOP_HOME/bin/hdfs" dfs -mkdir -p /tmp
```
philo-he: @infvg, could you create a script for HDFS preparation?
https://github.com/apache/incubator-gluten/pull/11373#discussion_r2689058008

infvg: Sorry yes, moved.

PHILO-HE (approved):
Looks good. Thanks.

infvg:
@PHILO-HE Looks like the tests passed, could we merge?

philo-he:
Thanks for the update. Assume no other comments from the community (if any, please feel free to post for follow-up fix).

Issue #11372, opened 2026-01-06
[VL][Spark-4.0] Add support for LeftSingle join for correlated scalar subqueries


Surbhi-Vijay:
### Description

Spark has introduced new join type `LeftSingle`  which triggers changes in query execution in certain scenarios.

Single join is a left outer join that checks that there is at most 1 build row for every probe row.

Refer:
[[SPARK-49653](https://issues.apache.org/jira/browse/SPARK-49653)][SQL] Single join for correlated scalar subqueries
commit: https://github.com/apache/spark/commit/3c81f076ab9c72514cfc8372edd16e6da7c151d6

### Gluten version

None

PR #11387, opened 2026-01-09
[CORE][VL] Handle exception in Schema validation for unsupported join types

Surbhi-Vijay:
Schema validation fails during validation phase for unsupported join types. When an unsupported join type is encountered, it throws `IllegalArgumentException `. This patch catches the exception and adds validation failure msg. This helps in falling back unsupported join type to Spark and query successfully completes.

This issue was encountered with Spark-4.0 where new join type `LeftSingle` has been introduced.
See: https://github.com/apache/incubator-gluten/issues/11372

## How was this patch tested?
Tested locally by running spark testcase with native session enabled.

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
@@ -2194,7 +2194,7 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
     }
   }
 
-  test("Left single join should not result into exception") {
+  testWithMinSparkVersion("Left single join should not result into exception", "4.0") {
     withSQLConf(SQLConf.ANSI_ENABLED.key -> "false") {
       spark.sql("create temp view x (x1, x2) as values (1, 1), (2, 2);")
       spark.sql("create temp view y (y1, y2) as values (2, 0), (3, -1);")


Surbhi-Vijay:
image

gluten-substrait/src/main/scala/org/apache/gluten/execution/ValidatablePlan.scala
```diff
@@ -85,6 +85,12 @@ trait ValidatablePlan extends GlutenPlan with LogLevelUtil {
               s" consider enabling the spark.sql.legacy.allowHashOnMapType " +
               s"setting to resolve this issue."
           ValidationResult.failed(message)
+        case e: IllegalArgumentException =>
+          // SchemaValidation throws IllegalArgumentException in Join validation
+          // when join type is unsupported. For example,
+          // LeftSingle join in BroadcastHashJoinExecTransformer.
+          ValidationResult.failed(
+            s"Failed to retrieve schema for ${this.nodeName}, due to: ${e.getMessage}")
       }
```
Yohahaha, line 94: Should we match Exception to avoid more special case? maybe IllegalArgumentException will be throw in somewhere else.

Surbhi-Vijay: Do you mean matching the  exception message or changing the exception itself?

Surbhi-Vijay: I think it should be okay to check `IllegalArgumentException`. We are catching this exception during validation phase and even if the validation is failing at some other place than joins, it should fallback to Spark.

Issue #11383, opened 2026-01-08
[VL] Adding configurations for Velox hash join bloom filter feature


zhouyuan:
### Description

Velox recently add the bloom filter when doing hash probe:
https://github.com/facebookincubator/velox/pull/15752 

```
  /// Whether hash probe can generate any dynamic filter (including Bloom
  /// filter) and push down to upstream operators.
  static constexpr const char* kHashProbeDynamicFilterPushdownEnabled =
      "hash_probe_dynamic_filter_pushdown_enabled";

  /// The maximum byte size of Bloom filter that can be generated from hash
  /// probe.  When set to 0, no Bloom filter will be generated.  To achieve
  /// optimal performance, this should not be too larger than the CPU cache size
  /// on the host.
  static constexpr const char* kHashProbeBloomFilterPushdownMaxSize =
      "hash_probe_bloom_filter_pushdown_max_size";

```
Gluten should add these two necessary configurations for Velox backend

### Gluten version

None

ItsZetty:
@zhouyuan 
Hi, if this issue is still open, may I work on it?

Yohahaha:
we already support specify the value with `spark.gluten.sql.columnar.backend.velox.hash_probe_dynamic_filter_pushdown_enabled=true`

FelixYBW:
> we already support specify the value with `spark.gluten.sql.columnar.backend.velox.hash_probe_dynamic_filter_pushdown_enabled=true`

correct, it's `spark.gluten.sql.columnar.backend.velox.hashProbe.dynamicFilterPushdown.enabled`, enabled by default

PR #11392, opened 2026-01-12
[GLUTEN-11383][VL] Allow bloom filter pushdown in hash probe

infvg:
Added ``spark.gluten.sql.columnar.backend.velox.hash_probe_bloom_filter_pushdown_max_size`` and ``spark.gluten.sql.columnar.backend.velox.hash_probe_dynamic_filter_pushdown_enabled`` as config options for velox

Resolves: https://github.com/apache/incubator-gluten/issues/11383


backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -447,6 +447,18 @@ object VeloxConfig extends ConfigRegistry {
       .longConf
       .createWithDefault(4194304L)
 
+  val COLUMNAR_VELOX_HASH_PROBE_DYNAMIC_FILTER_PUSHDOWN_ENABLED =
+    buildConf("spark.gluten.sql.columnar.backend.velox.hash_probe_bloom_filter_pushdown_max_size")
+      .doc("The maximum byte size of Bloom filter that can be generated from hash probe. When set to 0, no Bloom filter will be generated. To achieve optimal performance, this should not be too larger than the CPU cache size on the host.")
+      .intConf
+      .createWithDefault(0)
```
marin-ma, line 461: The configuration type and default value are incorrect.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -447,6 +447,18 @@ object VeloxConfig extends ConfigRegistry {
...
+
+  val COLUMNAR_VELOX_HASH_PROBE_BLOOM_FILTER_PUSHDOWN_MAX_SIZE =
+    buildConf("spark.gluten.sql.columnar.backend.velox.hash_probe_dynamic_filter_pushdown_enabled")
+      .doc("Whether hash probe can generate any dynamic filter (including Bloom filter) and push down to upstream operators.")
+      .booleanConf
+      .createWithDefault(true)
```
marin-ma, line 469: ditto

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
marin-ma: Seems like the variable name and the configuration don't match.

```
val COLUMNAR_VELOX_HASH_PROBE_DYNAMIC_FILTER_PUSHDOWN_ENABLED =
    buildConf("spark.gluten.sql.columnar.backend.velox.hash_probe_bloom_filter_pushdown_max_size")
```

infvg: Fixed

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
infvg: Fixed

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -447,6 +451,21 @@ object VeloxConfig extends ConfigRegistry {
       .longConf
       .createWithDefault(4194304L)
 
+  val COLUMNAR_VELOX_HASH_PROBE_BLOOM_FILTER_PUSHDOWN_MAX_SIZE =
+    buildConf("spark.gluten.sql.columnar.backend.velox.hash_probe_bloom_filter_pushdown_max_size")
```
philo-he: Should we follow the camel case naming convention for consistency instead of using underscore case?

infvg: The other variables in this scope are all underscore though

philo-he: @infvg, let me clarify. I meant the use of `hash_probe_bloom_filter_pushdown_max_size` in the config string. Should we convert it to camel case for consistency?

marin-ma: +1 Let's follow the camel case naming for the newly added configurations. If the configuration name is too long, you can break it into multiple parts. e.g. `.hashProbe.bloomFilterPushDownMaxSize`

For the variable name, please shorten it to `HASH_PROBE_BLOOM_FILTER_PUSHDOWN_MAX_SIZE`

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -85,6 +85,10 @@ class VeloxConfig(conf: SQLConf) extends GlutenConfig(conf) {
   def orcUseColumnNames: Boolean = getConf(ORC_USE_COLUMN_NAMES)
 
   def parquetUseColumnNames: Boolean = getConf(PARQUET_USE_COLUMN_NAMES)
+
+  def veloxHashProbeBloomFilterPushdownMaxSize: Long = getConf(COLUMNAR_VELOX_HASH_PROBE_BLOOM_FILTER_PUSHDOWN_MAX_SIZE)
```
marin-ma: `veloxHashProbeBloomFilterPushdownMaxSize` -> `hashProbeBloomFilterPushdownMaxSize`

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
infvg: Ah I see, I changed it. I set it to `hashProbe.bloomFilterPushdown.maxSize` since it feels more natural since we are setting the max size. Please let me know what you think

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
infvg: Changed

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxHashJoinSuite.scala
```diff
@@ -307,4 +307,75 @@ class VeloxHashJoinSuite extends VeloxWholeStageTransformerSuite {
...
+          df =>
+            val join = find(df.queryExecution.executedPlan) {
+              case _: ShuffledHashJoinExecTransformer => true
+              case _ => false
+            }
+            assert(join.isDefined)
+            val metrics = join.get.metrics
+
+            assert(metrics.contains("hashProbeDynamicFiltersProduced"))
+            assert(metrics("hashProbeDynamicFiltersProduced").value > 0)
```
rui-mo: I investigated the cause of the test failure and identified the following issues:
1) The test uses a broadcast hash join, which causes it to fail at L335. We can set the configuration `spark.sql.autoBroadcastJoinThreshold` to -1 to force the use of a shuffled hash join instead.
2) After applying this change, the metrics assertion still fails because `metrics("hashProbeDynamicFiltersProduced").value` is zero. This happens because, in this scenario, the Velox plan consists of `ValuesStream` operators as the children of the HashJoin, which is a typical join plan in Gluten when shuffle exists, but does not support dynamic filter pushdown. Dynamic filters only take effect when the `HashProbe` child is a `TableScan`. I'm not sure whether such a case can be constructed in a Gluten unit test.

```
-- Project[4][expressions: (n4_2:BIGINT, "n3_2")] -> n4_2:BIGINT
  -- Project[3][expressions: (n3_2:BIGINT, "n0_0"), (n3_3:BIGINT, "n1_0")] -> n3_2:BIGINT, n3_3:BIGINT
    -- HashJoin[2][INNER n0_0=n1_0] -> n0_0:BIGINT, n1_0:BIGINT
      -- ValueStream[0][] -> n0_0:BIGINT
      -- ValueStream[1][] -> n1_0:BIGINT
```

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxHashJoinSuite.scala
```diff
@@ -307,4 +307,75 @@ class VeloxHashJoinSuite extends VeloxWholeStageTransformerSuite {
...
+
+            assert(metrics.contains("hashProbeReplacedWithDynamicFilterRows"))
+            assert(metrics("hashProbeReplacedWithDynamicFilterRows").value > 0)
+        }
+      }
+    }
```
rui-mo: @infvg I made a few changes to the test above and confirmed that bloom filter pushdown is working by printing the following metrics in Velox.

```
bloomFilter->blocksByteSize(): 144704
numFiltersProduced: 1
```

A good next step would be to pass the `bloomFilter->blocksByteSize()` metric from Velox to Gluten (see `WholeStageResultIterator::collectMetrics()`), and then verify in this test that both `blocksByteSize` and `numFiltersProduced` are greater than 0.

This is the test with my modifications: 
```
withSQLConf(
      VeloxConfig.HASH_PROBE_DYNAMIC_FILTER_PUSHDOWN_ENABLED.key -> "true",
      VeloxConfig.HASH_PROBE_BLOOM_FILTER_PUSHDOWN_MAX_SIZE.key -> "1048576"
    ) {
      withTable("probe_table", "build_table") {
        spark.sql("""
        CREATE TABLE probe_table USING PARQUET
        AS SELECT id as a FROM range(110001)
      """)

        spark.sql("""
        CREATE TABLE build_table USING PARQUET
        AS SELECT id * 1000 as b FROM range(220002)
      """)

        runQueryAndCompare(
          "SELECT a FROM probe_table JOIN build_table ON a = b"
        ) {
          df =>
            val join = find(df.queryExecution.executedPlan) {
              case _: BroadcastHashJoinExecTransformer => true
              case _ => false
            }
            assert(join.isDefined)
            val metrics = join.get.metrics
            
             // TODO: assert the relevant metrics.
        }
      }
    }
```

infvg: Added

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -447,6 +452,22 @@ object VeloxConfig extends ConfigRegistry {
       .longConf
       .createWithDefault(4194304L)
 
+  val HASH_PROBE_BLOOM_FILTER_PUSHDOWN_MAX_SIZE =
+    buildConf("spark.gluten.sql.columnar.backend.velox.hashProbe.bloomFilterPushdown.maxSize")
+      .doc("The maximum byte size of Bloom filter that can be generated from hash probe. When " +
+        "set to 0, no Bloom filter will be generated. To achieve optimal performance, this should" +
+        " not be too larger than the CPU cache size on the host.")
+      .bytesConf(ByteUnit.BYTE)
+      .createWithDefault(0)
```
rui-mo, line 461: Do we need to disable the bloom filter pushdown by default?

infvg: Velox has the default set to true & 0:
https://github.com/facebookincubator/velox/blob/10bdc0688f892dcda83cfbcf723f656ba5e1e6b4/velox/core/QueryConfig.h#L1304-L1308

backends-velox/src/main/scala/org/apache/gluten/metrics/HashAggregateMetricsUpdater.scala
```diff
@@ -82,6 +83,7 @@ class HashAggregateMetricsUpdaterImpl(val metrics: Map[String, SQLMetric])
     aggSpilledFiles += aggMetrics.spilledFiles
     flushRowCount += aggMetrics.flushRowCount
     loadedToValueHook += aggMetrics.loadedToValueHook
+    bloomFilterBlocksByteSize += aggMetrics.bloomFilterBlocksByteSize
```
rui-mo: It seems like the bloom filter pushdown in aggregate hasn’t been confirmed to actually take effect yet, right?

infvg: Yep true - removed

rui-mo (approved):

PR #11429, opened 2026-01-16
[VL] Restrict `-Wno-class-memaccess` flag to GNU compiler

liujiayi771:

Only GNU GCC supports this flag; using Clang to compile will result in an error.

## How was this patch tested?

N/A

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -147,8 +147,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   add_compile_definitions(_GNU_SOURCE)
 endif()
 
-if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin" AND "${CMAKE_CXX_COMPILER_ID}"
-                                                    STREQUAL "GNU")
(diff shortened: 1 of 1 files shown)


cpp/CMakeLists.txt
```diff
@@ -147,7 +147,8 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   add_compile_definitions(_GNU_SOURCE)
 endif()
 
-if(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin" AND "${CMAKE_CXX_COMPILER_ID}"
```
philo-he: Thanks for the fix. Since `-Wno-class-memaccess` is a GNU-specific flag (and likely independent of the OS), can we remove the first condition and only check the compiler ID?

liujiayi771: @PHILO-HE Yes, I’ve also considered removing this part before. The "Darwin" check here was likely added because macOS defaults to Clang, which caused some compilation failures.

PR #11332, opened 2025-12-25
Use build/mvn wrapper in velox_backend_x86 workflow

yaooqinn:
## What changes are proposed in this pull request?

This PR updates the `.github/workflows/velox_backend_x86.yml` workflow to use the `build/mvn` wrapper script instead of directly calling `mvn`.

**Changes:**
- Updated `MVN_CMD` environment variable from `mvn -ntp` to `build/mvn -ntp`
- Removed Maven installation steps from various jobs (ubuntu, centos8, centos7 test jobs)
- Removed the `SETUP` environment variable and related helper script calls
- Added `build/mvn` to the workflow trigger paths to ensure workflow runs when the wrapper is modified
- Cleaned up Java setup steps by removing Maven installation dependencies

**Benefits:**
- Ensures consistent Maven version across all CI jobs
- Simplifies job setup by eliminating manual Maven installation
- Aligns with the project's adoption of the Maven wrapper (introduced in commit a35a3bda1)
- Reduces maintenance overhead for CI configuration

## How was this patch tested?

- The workflow changes follow the same pattern used in other workflows that already use `build/mvn`
- The Maven wrapper script (`build/mvn`) handles Maven installation automatically
- CI jobs will validate the changes when this PR is opened

PR #11431, opened 2026-01-16
[GLUTEN-11432][VL] Bump Maven from 3.9.2 to 3.9.12

zhouyuan:
maven 3.9.2 is a little outdated and it is often slow on downloading

## How was this patch tested?
pass GHA

Related issue: #11432

### DIFF
--- a/.github/workflows/util/setup-helper.sh
+++ b/.github/workflows/util/setup-helper.sh
@@ -18,7 +18,7 @@ set -e
 
 function install_maven {
   (
-    local maven_version="3.9.2"
+    local maven_version="3.9.12"
     local local_binary="apache-maven-${maven_version}-bin.tar.gz"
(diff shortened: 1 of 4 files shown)


baibaichen:
@yaooqinn is it right? IIUC, I shoudl only change `pom.xml` after https://github.com/apache/incubator-gluten/issues/11332

yaooqinn:
@baibaichen, I haven't updated all the cases to use maven.version in parent pom since my plate is kind of full recently. This PR only partially works for the cases that is not covered

zhouyuan:
@baibaichen @yaooqinn 
Hi, thanks for notifying, this patch is trying to fix the docker build issue (https://github.com/apache/incubator-gluten/issues/11432). I will update the system level maven as followup.

PR #11441, opened 2026-01-19
[GLUTEN-10933][VL] BroadcastExchange outputs cudf::table

jinchengchenghh:
A follow up of https://github.com/apache/incubator-gluten/pull/11090, broadcast table is usually a small table, so we don't serialize to cudf format in the first step, just convert it to cudf table, this might be expensive for string data type.
May optimize to serialize to cudf format and deserialize to cudf table as hash shuffle reader, copy the buffer to deserialize to cudf::table.

Test in local with TPCDS Q95.

Related issue: #10933

### DIFF
--- a/cpp/velox/compute/VeloxRuntime.cc
+++ b/cpp/velox/compute/VeloxRuntime.cc
@@ -315,9 +315,8 @@ std::unique_ptr<ColumnarBatchSerializer> VeloxRuntime::createColumnarBatchSerial
   if (veloxCfg_->get<bool>(kCudfEnabled, kCudfEnabledDefault)) {
     return std::make_unique<VeloxGpuColumnarBatchSerializer>(arrowPool, veloxPool, cSchema);
   }
-#else
-  return std::make_unique<VeloxColumnarBatchSerializer>(arrowPool, veloxPool, cSchema);
 #endif
+  return std::make_unique<VeloxColumnarBatchSerializer>(arrowPool, veloxPool, cSchema);
 }
 
 void VeloxRuntime::enableDumping() {


cpp/velox/memory/GpuBufferColumnarBatch.cc
```diff
@@ -123,11 +128,21 @@ std::shared_ptr<GpuBufferColumnarBatch> GpuBufferColumnarBatch::compose(
...
+      if (batch->bufferAt(bufferIdx) == nullptr || batch->bufferAt(bufferIdx)->size() == 0) {
+        if (returnBuffers[bufferIdx] != nullptr) {
+          auto* dst = returnBuffers[bufferIdx]->mutable_data();
+          arrow::bit_util::SetBitsTo(dst, rowNumber, batch->numRows(), true);
+        }
       } else {
+        // Need to allocate null buffer.
+        if (returnBuffers[bufferIdx] == nullptr) {
+          std::shared_ptr<arrow::Buffer> buffer;
+          GLUTEN_ASSIGN_OR_THROW(buffer, arrow::AllocateResizableBuffer(bufferSizes[bufferIdx], pool));
```
marin-ma, line 140: Should it be `arrow::bit_util::BytesForBits(bufferSizes[bufferIdx])`？

jinchengchenghh: The bufferSizes is computed here, https://github.com/apache/incubator-gluten/pull/11441/changes#diff-ce18661fb719aacbaf297406b7d1dc8406164158b10644ce68b15cf51d94ea12R102, so it is bytes

Issue #11400, opened 2026-01-13
[VL] Track on Spark-4.1.x failed unit tests


baibaichen:
### Description

After xx the general framework of unit tests are merged. We still have failed unit tests left here:

| Description | Owner | Category | Cause | Affected Files |
|-------------|-------|----------|-------|----------------|
| [Improve `extractShuffleIds` to find `AdaptiveSparkPlanExec` anywhere in plan tree](https://github.com/apache/spark/pull/53620) 4.1.1 | ✔️ @baibaichen  | SQL | [#52157](https://github.com/apache/spark/pull/52157) | `gluten-ut/spark41/.../VeloxTestSettings.scala` |
| Fix `SPARK-47939: Explain should work with parameterized queries` | ✔️ @marin-ma  | TEST | https://github.com/apache/incubator-gluten/pull/11252 |  `gluten-ut/spark41/.../VeloxTestSettings.scala` |
| Support Checksum in Column ShuffleWriters | 🚫 | CORE | [#50230](https://github.com/apache/spark/pull/50230) | gluten-ut/spark41/.../velox/VeloxTestSettings.scala<br>gluten-ut/spark41/.../GlutenMapStatusEndToEndSuite.scala<br>**Excluded tests:**<br>- GlutenMapStatusEndToEndSuite (entire suite) |
| Support `spark.sql.unionOutputPartitioning=true`  | @baibaichen | SQL | [#51623](https://github.com/apache/spark/pull/51623) | `.github/workflows/velox_backend_x86.yml`, `gluten-ut/spark41/.../VeloxTestSettings.scala`, `tools/gluten-it/common/.../Suite.scala`<br>**Excluded tests:**<br>- GlutenBroadcastExchangeSuite.SPARK-52962<br>- GlutenDataFrameSetOperationsSuite.SPARK-52921* |
| Fixes a Spark Parquet read bug where missing struct fields caused the entire struct to be read as NULL. |❓️ | PARQUET | [SPARK-53535](https://issues.apache.org/jira/browse/SPARK-53535) | `gluten-ut/spark41/.../VeloxTestSettings.scala`<br>**Excluded tests:**<br>- SPARK-53535*<br>- vectorized reader: missing all struct fields* |
[...]

PR #11455, opened 2026-01-20
[GLUTEN-11400][CORE]  Implement partitioning-aware union for ColumnarUnionExec

baibaichen:
for #11400

This PR adapts Gluten to support Spark 4.1's new partitioning-aware union feature introduced in [SPARK-52921](https://github.com/apache/spark/pull/51623) (commit `cf3a34e1`).

**Key changes:**
1. **ColumnarUnionExec enhancement**: Modified `ColumnarUnionExec` to preserve `outputPartitioning` from the original `UnionExec`. When children operators have identical partitioning, it now uses `SQLPartitioningAwareUnionRDD` instead of `SparkContext.union()` to maintain correct data distribution.

2. **Shim layer for version compatibility**: Added `SparkContextUtils.createPartitioningAwareUnionRDD()` method in the shim layer - implemented for Spark 4.1, throwing `UnsupportedOperationException` for earlier versions (3.3/3.4/3.5/4.0).

3. **Metrics fix**: Added `resetMetrics()` override in `BroadcastHashJoinExecTransformer` to match the related fix in Spark PR #51673, preventing metrics loss after broadcast materialization.

4. **ColumnarRangeExec refactoring**: Simplified to accept `RangeExec` directly and properly preserve `outputOrdering` and `outputPartitioning`.

Spark 4.1's `UnionExec` now intelligently preserves child partitioning when all children have identical partitioning (via `spark.sql.unionOutputPartitioning=true`, which is **enabled by default**).

**This causes incorrect query results in Gluten without this fix:**
- When `spark.sql.unionOutputPartitioning=true`, Spark's optimizer knows the data distribution after union, so it omits the Exchange (shuffle) operator in downstream operations.
- However, Gluten's `ColumnarUnionExec` did not follow the partitioning information from its children, causing data to be out of order.
- This leads to **incorrect results** because downstream operators (e.g., aggregations, joins) expect data to be partitioned in a specific way, but `ColumnarUnionExec` was not preserving that contract.

### Known Limitation

**RAS does not support `spark.sql.unionOutputPartitioning=true`.** See #11468.

`GroupLeafExec` in RAS always returns `UnknownPartitioning` for `outputPartitioning`, which breaks the partitioning contract. As a workaround, RAS tests are configured with `spark.sql.unionOutputPartitioning=false`.

No. This is a bug fix that ensures correct query results when running on Spark 4.1 with the default configuration.

**Note:** Users who enable RAS should set `spark.sql.unionOutputPartitioning=false` to avoid potential incorrect results.

- Added tests in `GlutenDataFrameSetOperationsSuite` for union partitioning scenarios
- Added tests in `GlutenBroadcastExchangeSuite` for broadcast metrics reset behavior
- Updated CI configuration to properly test with `spark.sql.unionOutputPartitioning` enabled
- RAS tests are configured with `spark.sql.unionOutputPartitioning=false` due to the known limitation

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHRangeExecTransformer.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHRangeExecTransformer.scala
@@ -27,6 +27,7 @@ import org.apache.gluten.substrait.rel.LocalFilesNode.ReadFileFormat
 
 import org.apache.spark.Partition
 import org.apache.spark.sql.catalyst.plans.logical.{Range => LogicalRange}
+import org.apache.spark.sql.execution.{ColumnarRangeBaseExec, SparkPlan}
 import org.apache.spark.sql.execution.datasources.clickhouse.ExtensionTableBuilder
 import org.apache.spark.sql.execution.metric.{SQLMetric, SQLMetrics}
 


gluten-core/src/main/scala/org/apache/gluten/extension/columnar/enumerated/planner/metadata/GlutenMetadataModel.scala
```diff
@@ -33,10 +34,12 @@ object GlutenMetadataModel extends Logging {
       case other =>
         GlutenMetadata(
           Schema(other.output),
-          other.logicalLink.map(LogicalLink(_)).getOrElse(LogicalLink.notFound))
+          other.logicalLink.map(LogicalLink(_)).getOrElse(LogicalLink.notFound),
+          other.outputPartitioning)
```
zhztheplayer: This change may not break anything, but partitioning is not usually considered a logical property in terms of a Cascades-style query optimizer like RAS, so it's essentially not a correct change.

BTW, I was considering completely removing RAS as 1. no active users 2. no one else in the community could maintain the code forward with enough search-based optimizer framework knowledge. More contexts should reveal afterwards.

gluten-substrait/src/main/scala/org/apache/gluten/execution/RangeExecBaseTransformer.scala
```diff
@@ -33,7 +34,9 @@ abstract class ColumnarRangeBaseExec(
     numSlices: Int,
     numElements: BigInt,
     outputAttributes: Seq[Attribute],
-    child: Seq[SparkPlan])
+    child: Seq[SparkPlan],
+    override val outputOrdering: Seq[SortOrder],
+    override val outputPartitioning: Partitioning)
```
zhztheplayer: Why range has children? Can we remove?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/enumerated/planner/metadata/GlutenMetadataModel.scala
baibaichen: `GroupLeafExec` should override `outputPartitioning` to return the correct partitioning.

I agree with removing RAS. If we go ahead with that, I can revert this fix and

1. skip the test
2. file an issue to track this.

gluten-substrait/src/main/scala/org/apache/gluten/execution/RangeExecBaseTransformer.scala
baibaichen: let's try to remove the children

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/enumerated/planner/metadata/GlutenMetadataModel.scala
baibaichen: @zhztheplayer https://github.com/apache/incubator-gluten/issues/11468

zhztheplayer:
@baibaichen CI is still failing.

Issue #11390, opened 2026-01-10
[BUG][VL] Inconsistent C++ Compiler Flags for aarch64 Causing xsimd initialization errors


xumanbu:
### Backend

VL (Velox)

### Environment
- ** Architecture: aarch64 ( aws graviton2 neoverse-n1)
- ** OS: Centos9
- ** Gluten Version: 1.5.x
- ** Build Type: Default build (without explicit CPU_TARGET override)

### Problem Description
When building Gluten on aarch64 architecture with default settings, there is an inconsistency in the C++ compiler flags between Gluten's build script and Velox's build script. This inconsistency causes runtime errors during xsimd initialization due to CPU feature detection mismatch.
```
Current thread (0x0000ffffac02d800):  JavaThread "main" [_thread_in_native, id=52, stack(0x0000ffffb0c38000,0x0000ffffb0e38000)]

Stack: [0x0000ffffb0c38000,0x0000ffffb0e38000],  sp=0x0000ffffb0e31df0,  free space=2023k
Native frames: (J=compiled Java code, A=aot compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libvelox.so+0x2e86470]  facebook::velox::simd::initializeSimdUtil()+0x50
C  [ld-linux-aarch64.so.1+0x5be4]  call_init+0xd4
C  [ld-linux-aarch64.so.1+0x5cec]  _dl_init+0x7c
C  [ld-linux-aarch64.so.1+0x2110]  _dl_catch_exception+0xe0
C  [ld-linux-aarch64.so.1+0xc020]  dl_open_worker+0xe0
C  [ld-linux-aarch64.so.1+0x2094]  _dl_catch_exception+0x64
C  [ld-linux-aarch64.so.1+0xc45c]  _dl_open+0x98
C  [libc.so.6+0x7d194]  dlopen_doit+0x64
C  [ld-linux-aarch64.so.1+0x2094]  _dl_catch_exception+0x64
C  [ld-linux-aarch64.so.1+0x21dc]  _dl_catch_error+0x2c
C  [libc.so.6+0x7cc18]  _dlerror_run+0x88
C  [libc.so.6+0x7d270]  dlopen+0x90
V  [libjvm.so+0xa9c898]  os::Linux::dlopen_helper(char const*, char*, int)+0x28
V  [libjvm.so+0xa9cbc4]  os::dll_load(char const*, char*, int)+0x74
[...]

PR #11457, opened 2026-01-21
[GLUTEN-11390][VL] Align aarch64 compiler flags with Velox to fix xsimd initialization errors

xumanbu:

fix: https://github.com/apache/incubator-gluten/issues/11390

Follow Velox's ARM CPU detection logic to ensure consistent compiler flags, between Gluten and Velox, preventing xsimd initialization issues.See: ep/build-velox/build/velox_ep/scripts/setup-helper-functions.sh:142-180

## How was this patch tested?
build test

### DIFF
--- a/dev/build-helper-functions.sh
+++ b/dev/build-helper-functions.sh
@@ -75,8 +75,8 @@ function get_cxx_flags {
 
     "aarch64")
       # Follow Velox's ARM CPU detection logic to ensure consistent compiler flags
-      # between Gluten and Velox, preventing xsimd initialization issues.
-      # See: ep/build-velox/build/velox_ep/scripts/setup-helper-functions.sh:142-180
+      # between Gluten and Velox, preventing xsimd initialization issues, see GLUTEN-11390.
+      # Reference: function get_cxx_flags in Velox's setup-helper-functions.sh.
 
       # Read Arm MIDR_EL1 register to detect Arm cpu.
       # https://developer.arm.com/documentation/100616/0301/register-descriptions/aarch64-system-registers/midr-el1--main-id-register--el1


xumanbu:
@PHILO-HE PTAL

dev/build-helper-functions.sh
```diff
@@ -74,7 +74,52 @@ function get_cxx_flags {
     ;;
 
     "aarch64")
-      echo -n "-mcpu=neoverse-n1 -std=c++20 $ADDITIONAL_FLAGS"
+      # Follow Velox's ARM CPU detection logic to ensure consistent compiler flags
+      # between Gluten and Velox, preventing xsimd initialization issues.
```
philo-he: Suggest adding the issue to make it clear to developers.
`preventing xsimd initialization issues, see GLUTEN-11390.`

dev/build-helper-functions.sh
```diff
@@ -74,7 +74,52 @@ function get_cxx_flags {
...
+      # See: ep/build-velox/build/velox_ep/scripts/setup-helper-functions.sh:142-180
```
philo-he: Nit: The line numbers can be invalid very likely during Velox's iteration. Suggest the following comment. 

`Reference: function get_cxx_flags in Velox's setup-helper-functions.sh.`

PHILO-HE (commented):
Two minor comments. Thanks.

zhouyuan:
`dev/build-helper-functions.sh` seems duplicated with Velox - could we use the one in Velox repo?

dev/build-helper-functions.sh
xumanbu: indeed, I'll modify it.

xumanbu:
> `dev/build-helper-functions.sh` seems duplicated with Velox - could we use the one in Velox repo?

@zhouyuan  gluten have specific functions like setup_macos/setup_linux that aren't in Velox's version, and  velox's get_cxx_flags returns only base flags (e.g., -mcpu=neoverse-n1) without these additional flags.
they don't look exactly the same, I think it is better to remain independent at now.

dev/build-helper-functions.sh
xumanbu: updated

Issue #11451, opened 2026-01-20
[CH][BUG]  Invalid result from joining an empty right table


lgbo-ustc:
### Backend

CH (ClickHouse)

### Bug description

[Expected behavior] and [actual behavior].

We have a query as following
```sql
select * from t1 left join (select * from t2 where day='2026-01-01') 
```
The partition `2026-01-01` in `t2`  exists, but has no data.  `t1` has one row. The query result has one row for `vanilla`, but is empty for `gluten(ch)`.

The physical plan for gluten is
```
+- == Final Plan ==
   CHNativeColumnarToRow (13)
   +- ^ ProjectExecTransformer (11)
      +- ^ CHBroadcastNestedLoopJoinExecTransformer LeftOuter BuildLeft (10)
         :- ^ InputIteratorTransformer (6)
         :  +- BroadcastQueryStage (4), Statistics(sizeInBytes=27.0 B, rowCount=1)
         :     +- ColumnarBroadcastExchange (3)
         :        +- ^ NativeScan hive t1(1)
         +- ^ ProjectExecTransformer (9)
            +- ^ FilterExecTransformer (8)
               +- ^ NativeScan hive t2(7)
```

the physical plan for vanilla is
```

+- == Final Plan ==
   * Project (2)
   +- Scan hive t1(1)
```

### Spark version

None

### Relevant logs

```bash

```

PR #11466, opened 2026-01-22
[GLUTEN-11451][CH] Fix CH BNLJ left outer build side validation

zhanglistar:
- Re-enable base validation for CH BroadcastNestedLoopJoin to reject invalid build sides (e.g., LeftOuter + BuildLeft).
- Add a regression test for left join with an existing but empty partition on the build side.
Fix [#11451](https://github.com/apache/incubator-gluten/issues/11451).

## Why is this change needed?
Gluten CH was accepting LeftOuter + BuildLeft in BNLJ, which violates Spark join semantics. When the right side is empty, the streamed side becomes empty and drops left rows, producing incorrect results compared to vanilla Spark.

## How was this patch tested?
- Added unit test in `GlutenClickHouseJoinSuite`.

### DIFF
--- a/.github/workflows/velox_backend_x86.yml
+++ b/.github/workflows/velox_backend_x86.yml
@@ -286,12 +286,16 @@ jobs:
         run: |
           echo "JAVA_HOME: $JAVA_HOME"
           cd $GITHUB_WORKSPACE/tools/gluten-it
+          SPARK41_CONF=""
+          if [ "${{ matrix.spark }}" = "spark-4.1" ]; then
+            SPARK41_CONF="--extra-conf=spark.sql.unionOutputPartitioning=false"
+          fi
           GLUTEN_IT_JVM_ARGS=-Xmx5G sbin/gluten-it.sh queries-compare \
             --local --preset=velox --benchmark-type=h --error-on-memleak --off-heap-size=10g -s=1.0 --threads=16 --iterations=1 \
-            --extra-conf=spark.gluten.ras.enabled=true \
+            --extra-conf=spark.gluten.ras.enabled=true $SPARK41_CONF \
           && GLUTEN_IT_JVM_ARGS=-Xmx6G sbin/gluten-it.sh queries-compare \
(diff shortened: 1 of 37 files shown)


backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHBroadcastNestedLoopJoinExecTransformer.scala
```diff
@@ -111,15 +111,24 @@ case class CHBroadcastNestedLoopJoinExecTransformer(
...
   override def validateJoinTypeAndBuildSide(): ValidationResult = {
+    (joinType, buildSide) match {
+      case (LeftOuter, BuildLeft) | (RightOuter, BuildRight) =>
+        return ValidationResult.failed(s"$joinType join is not supported with $buildSide")
+      case _ =>
+    }
+
     joinType match {
-      case _: InnerLike =>
+      case _: InnerLike | FullOuter | LeftOuter | RightOuter =>
```
lgbo-ustc: Is cross join not supported?

backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHBroadcastNestedLoopJoinExecTransformer.scala
```diff
@@ -111,15 +111,24 @@ case class CHBroadcastNestedLoopJoinExecTransformer(
   }
 
   override def validateJoinTypeAndBuildSide(): ValidationResult = {
+    (joinType, buildSide) match {
```
zhanglistar: fallback if empty, changed.

zhanglistar: fallback if empty, changed.

backends-clickhouse/src/main/scala/org/apache/gluten/execution/CHBroadcastNestedLoopJoinExecTransformer.scala
zhanglistar: Should not modify this, restored.

PR #11469, opened 2026-01-22
[VL] Adding cache for m2 repo

zhouyuan:
This patch adds the cache for m2 repo after TPC test

## How was this patch tested?
pass GHA

### DIFF
--- a/.github/workflows/velox_backend_x86.yml
+++ b/.github/workflows/velox_backend_x86.yml
@@ -109,7 +109,7 @@ jobs:
         os: [ "ubuntu:20.04", "ubuntu:22.04" ]
         spark: [ "spark-3.3", "spark-3.4", "spark-3.5", "spark-4.0", "spark-4.1"  ]
         java: [ "java-8", "java-11", "java-17", "java-21" ]
-        # Spark supports JDK17 since 3.3.
+        # Spark supports JDK17 since 3.3, jdk21 since 4.0
         exclude:
           - spark: spark-3.3
             java: java-21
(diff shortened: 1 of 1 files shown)


zhouyuan:
@baibaichen The Java caching can work in this way, however it's too big for now and may trigger evict on the CPP cache
image

FelixYBW:
we may create an image including cache daily, then pull the image and copy .w2 from it. Is it feasible?

baibaichen:
How is the cache generated? Is there a shared disk that we mount for each test?

FelixYBW:
> How is the cache generated? Is there a shared disk that we mount for each test?

https://github.com/apache/incubator-gluten/actions/caches. You can't mount it. You can copy it. The whole cache size is limited

baibaichen:
> > How is the cache generated? Is there a shared disk that we mount for each test?
> 
> https://github.com/apache/incubator-gluten/actions/caches. You can't mount it. You can copy it. The whole cache size is limited

I see. I agree with proposal of @FelixYBW. We can set up a daily or weekly job to generate the .m2 cache.

**Note**: Since Java Jar files are independent of the OS and Java version, we only need to maintain a single, unified cache for .m2.

To populate this cache comprehensively, we can execute the following commands. Theoretically, this sequence should download all necessary dependencies for our various build profiles into the cache:

```shell
build/mvn -P java-17,spark-4.0,scala-2.13,backends-velox,hadoop-3.3,spark-ut -Piceberg,iceberg-test,delta,paimon clean test-compile
build/mvn -P java-17,spark-3.5,backends-velox,hadoop-3.3,spark-ut -Piceberg,iceberg-test,delta,paimon clean test-compile
build/mvn -P java-17,spark-3.4,backends-velox,hadoop-3.3,spark-ut -Piceberg,iceberg-test,delta,paimon clean test-compile
build/mvn -P java-17,spark-3.3,backends-velox,hadoop-3.3,spark-ut -Piceberg,delta,paimon clean test-compile
build/mvn -P java-17,spark-4.1,scala-2.13,backends-velox,hadoop-3.3,spark-ut clean test-compile
```

Regarding `.ccache`, do we currently update it only when upgrading Velox?

zhouyuan:
fired one issue to track

PR #11484, opened 2026-01-24
[DEPRECATED][CORE] Rewrite array_contains(map_keys(m), k) to map_contains_key(m, k)

taiyang-li:
Motivation:
- Reduce runtime overhead for queries using array_contains(map_keys(m), k) and avoid constructing keys array.
- Fix a WholeStageCodegen failure when LogicalPlan-level rewrite produces map_contains_key in row-based ProjectExec.

Implementation:
- Remove the LogicalPlan optimizer rule ArrayContainsMapKeysRewriteRule.
- Add a SparkPlan-level rule RewriteArrayContainsMapKeys under gluten-substrait.
- The rule only rewrites expressions inside ProjectExecTransformer and FilterExecTransformerBase (columnar project/filter nodes).
- Row-based ProjectExec keeps using array_contains(map_keys(m), k), which is codegen-compatible in Spark.

Injection:
- ClickHouse backend: inject RewriteArrayContainsMapKeys as a post-transform rule in CHRuleApi (legacy columnar pipeline).
- Velox backend: inject RewriteArrayContainsMapKeys as a post-transform rule in VeloxRuleApi for both legacy and RAS pipelines.

This keeps the semantics of array_contains(map_keys(m), k) and map_contains_key(m, k) aligned while ensuring that map_contains_key only appears on the columnar path and no longer breaks row-path codegen.


taiyang-li:
map_contains_key not supported in velox backend. Since in vanilla spark map_contains_key is replaced with array_contains(map_keys), we don't need to do replacement in opposite direction.

PR #11490, opened 2026-01-26
[INFRA] Minor follow-up for PR #11482, remove SPARK_HOME setting from cudf docker build

philo-he:

The SPARK_HOME was set with nonexistent directory. It seems not really used.

Follow-up for #11482.


philo-he:
@zhouyuan, this is a minor follow-up, please take a look.

dev/docker/cudf/Dockerfile
zhouyuan: @PHILO-HE spark-344 will be installed here

PR #11492, opened 2026-01-26
[INFRA] Add whether it is an AI-generated tag in the PR and issue template

zml1206:

As title

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -27,6 +27,7 @@ For UI-related changes, attach screenshots to demonstrate the updates.
 -->
 
 ## Was this patch authored or co-authored using generative AI tooling?
+
 <!--
 If generative AI tooling has been used in the process of authoring this patch, please include the
 phrase: 'Generated-by: ' followed by the name of the tool and its version.


zml1206:
@zhztheplayer Thanks

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -25,3 +25,12 @@ Describe how the changes were tested, if applicable.
...
 For UI-related changes, attach screenshots to demonstrate the updates.
 -->
+
+## Was this patch authored or co-authored using generative AI tooling?
+
+<!--
+If generative AI tooling has been used in the process of authoring this patch, please include the
+phrase: 'Generated-by: ' followed by the name of the tool and its version.
+If no, write 'No'.
+Please refer to the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) for details.
```
philo-he, line 35: I assume most patches are human-authored. If so, can we just highlight this in tips section at the top? To me, including it as guideline seems better rather than a mandatory field to fill.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -38,7 +38,7 @@ body:
     attributes:
       label: Bug description
       description: Please describe the issue and the expected behavior.
-      value: "[Expected behavior] and [actual behavior]."
+      value: "[Expected behavior] and [actual behavior]. If this issue was written with the assistance of AI, please note that at the end of the description."
```
philo-he, line 41: "this issue was written with",  does it mean the issue description? Or actually mean the issue was found by AI?

I feel that the code authorship matters according to ASF guideline. But here, if it means issue description drafted or polished by AI tool, it seems unnecessary to include such info to make people aware of that?

zml1206: It means the issue description. I also agree that it's not necessary here. cc @zhztheplayer

.github/PULL_REQUEST_TEMPLATE
zml1206: Thanks for your suggestion. AI-assisted programming will become increasingly common. I'm worried that the tips at the top are easily overlooked.

.github/ISSUE_TEMPLATE/bug.yml
zhztheplayer: > But here, if it means issue description drafted or polished by AI tool, it seems unnecessary to include such info to make people aware of that?

If I understand it correctly, say an issue is only polished by AI, the author will still be human. In that case, we'd recommend to note that in issue description, subject to whether the author thinks it's needed. Apparently, using AI to fix typos and using AI to propose a solution are completely different things, so we can barely draw a clear dividing line here.

Issue #11485, opened 2026-01-26
[VL] Race condition in VeloxMemoryManager::getOrCreateArrowMemoryPool with weak_ptr


baibaichen:
# Backend

VL (Velox)

# Bug description

There is a race condition in `VeloxMemoryManager::getOrCreateArrowMemoryPool` due to the use of `weak_ptr` and destructor callback not being in the same critical section.

Here are failed [test log](https://github.com/apache/incubator-gluten/actions/runs/21333277170/job/61401448463?pr=11475)

```
2026-01-25T13:46:34.3053380Z 
2026-01-25T13:46:34.3053873Z 13:46:34.302 ERROR org.apache.spark.executor.Executor: Exception in task 3.0 in stage 313.0 (TID 409)
2026-01-25T13:46:34.3055309Z org.apache.spark.SparkException: [TASK_WRITE_FAILED] Task failed while writing rows to file:/tmp/spark-0f0102eb-9389-43da-bdcf-3da35ed702d9.
2026-01-25T13:46:34.3056950Z 	at org.apache.spark.sql.errors.QueryExecutionErrors$.taskFailedWhileWritingRowsError(QueryExecutionErrors.scala:775)
2026-01-25T13:46:34.3058533Z 	at org.apache.spark.sql.delta.files.GlutenDeltaFileFormatWriter$.executeTask(GlutenDeltaFileFormatWriter.scala:495)
2026-01-25T13:46:34.3064636Z 	at org.apache.spark.sql.delta.files.GlutenDeltaFileFormatWriter$.$anonfun$executeWrite$4(GlutenDeltaFileFormatWriter.scala:313)
2026-01-25T13:46:34.3065956Z 	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:93)
2026-01-25T13:46:34.3066853Z 	at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:166)
2026-01-25T13:46:34.3067594Z 	at org.apache.spark.scheduler.Task.run(Task.scala:141)
2026-01-25T13:46:34.3068369Z 	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$4(Executor.scala:620)
[...]

PR #11493, opened 2026-01-26
[GLUTEN-11485][VL] Fix the race condition in ArrowMemoryPool

marin-ma:
Related issue: #11485

### DIFF
--- a/cpp/core/memory/ArrowMemoryPool.cc
+++ b/cpp/core/memory/ArrowMemoryPool.cc
@@ -20,12 +20,6 @@
 
 namespace gluten {
 
-ArrowMemoryPool::~ArrowMemoryPool() {
-  if (releaser_ != nullptr) {
-    releaser_(this);
-  }
-}
-
 arrow::Status ArrowMemoryPool::Allocate(int64_t size, int64_t alignment, uint8_t** out) {
   if (!allocator_->allocateAligned(alignment, size, reinterpret_cast<void**>(out))) {
     return arrow::Status::Invalid("WrappedMemoryPool: Error allocating " + std::to_string(size) + " bytes");
--- a/cpp/core/memory/ArrowMemoryPool.h
+++ b/cpp/core/memory/ArrowMemoryPool.h
@@ -33,7 +33,7 @@ class ArrowMemoryPool final : public arrow::MemoryPool {
   explicit ArrowMemoryPool(AllocationListener* listener)
       : allocator_(std::make_unique<ListenableMemoryAllocator>(defaultMemoryAllocator().get(), listener)) {}
 
-  ~ArrowMemoryPool() override;
+  ~ArrowMemoryPool() override = default;
 
   ArrowMemoryPool(const ArrowMemoryPool&) = delete;
 


cpp/velox/memory/VeloxMemoryManager.cc
```diff
@@ -327,22 +330,17 @@ int64_t shrinkVeloxMemoryPool(velox::memory::MemoryManager* mm, velox::memory::M
...
-    return pool;
+    if (auto pool = it->second.lock()) {
+      return pool;
+    }
+    arrowPools_.erase(name);
   }
-  auto pool = std::make_shared<ArrowMemoryPool>(
-      blockListener_.get(), [this, name](arrow::MemoryPool* pool) { this->dropMemoryPool(name); });
+
+  auto pool = std::make_shared<ArrowMemoryPool>(blockListener_.get());
```
zhztheplayer, line 339: If we no longer remove the entries, will there be any memory leak risk caused by the expanding entry list in `arrowPools_`?

baibaichen: It looks like not an issue as long as the number of keys is not too large.

marin-ma: @baibaichen mentioned in the issue that we can use a background thread to clean up, but I wonder if that might be too heavy for the current use of the arrow pool, since only shuffle and parquet write create named arrow memory pool.

If necessary, perhaps looping over the weak_ptr each time `getOrCreateArrowMemoryPool` is called and remove the expired ones. @zhztheplayer What do you think?

zhztheplayer: > It looks like not an issue as long as the number of keys is not too large.

I thought it might be up to the usage. Hi @marin-ma, do you think we can skip cleaning up the Arrow pools? In that case, we can simply store shared pointer as the class member.

> If necessary, perhaps looping over the weak_ptr each time getOrCreateArrowMemoryPool is called and remove the expired ones.

Otherwise, this might be helpful, but we should also figure out whether it will cause any instability regarding performance.

baibaichen: If there aren’t many entries in `arrowPools_`, I think it’s fine to keep them via `shared_ptr`. A `MemoryPool` object itself shouldn’t take much memory anyway.

marin-ma: Using `weak_ptr` can help to clarify the ownership of the arrow pool. For example if the arrow pool is used by shuffle A, it should be destroyed together with the shuffle writer A itself. And then the next shuffle B will create a new pool. In this way we can also track the memory allocation status for each shuffle separately.

zhztheplayer: @marin-ma I meant, is it feasible to use `std::unordered_map<std::string, std::shared_ptr<ArrowMemoryPool>> arrowPools_`? Only by changing the map value type from `std::weak_ptr` to `std::shared_ptr`.

zhztheplayer: Because in the latest PR's code, the expired map entries won't get removed from map until being replaced by new value of the same key. Is this an intentional change?

zhztheplayer (approved):
I think this is a valid fix, was just curious about the reason we are still using `weak_ptr` as the map value, since the expired entries won't be removed from the map anyway.

cpp/velox/memory/VeloxMemoryManager.cc
marin-ma: > Because in the latest PR's code, the expired map entries won't get removed from map until being replaced by new value of the same key. Is this an intentional change?

Yes. In this way it can create new arrow pools for each shuffle writer instance. If changing to `std::unordered_map<std::string, std::shared_ptr<ArrowMemoryPool>> arrowPools_`, the new shuffle writers will reuse the ones created by the previous shuffle.

marin-ma:
@baibaichen Do you have any more comments? Thanks!

baibaichen:
> @baibaichen Do you have any more comments? Thanks!

No. Let's merge to see whether the issue is fixed or not.

PR #11500, opened 2026-01-27
[VL] Fix object leak on `TreeMemoryConsumers#FACTORIES`

zhztheplayer:
The following code came across problematic:

```scala
  private static final ReferenceMap<TaskMemoryManager, Factory> FACTORIES =
      new ReferenceMap<>(
          AbstractReferenceMap.ReferenceStrength.WEAK, AbstractReferenceMap.ReferenceStrength.WEAK);
```

Because value is `WEAK`, the whole entry may be removed from map before the Task is ended because the value might become weak-reachable before that, which causes unexpected large number of creations of `TreeMemoryConsumers#Factory` instances in the same Spark task.

Strength of `key/value` used to be `WEAK/STRONG` but was also leading to object leak because of cycle references from values to keys.

The PR refactors the code to completely remove the usage of `ReferenceMap`. The factory lifecycle will be managed by `TaskResources` via Task completion listener which is far more reliable than the weak reference approach.


zhztheplayer:

PR #11506, opened 2026-01-28
[MINOR] Add missing profiles in `format-scala-code.sh`

liujiayi771:

The scala-format-check CI uses `dev/format-scala-code.sh` to run the spotless check, but it misses the backends-clickhouse and kafka profiles.


PHILO-HE (approved):
Thanks for the fix.

PR #11515, opened 2026-01-29
[CORE] Use build/mvn wrapper for scheduled jobs and in Dockerfiles

yaooqinn:

This PR migrates GitHub Actions workflows and Dockerfiles to use the `./build/mvn` Maven wrapper instead of manually installing Maven via `setup-helper.sh`. The wrapper automatically downloads Maven 3.9.12 when needed, simplifying the CI setup and reducing maintenance overhead.

**Changes:**
- Delete `.github/workflows/util/setup-helper.sh` (no longer needed)
- Update workflows to use `./build/mvn`:
  - `build_bundle_package.yml`
  - `velox_nightly.yml` (6 jobs)
  - `velox_weekly.yml`
  - `scala_code_format.yml` (remove unused SETUP env var)
- Update Dockerfiles to use `./build/mvn`:
  - `dev/docker/Dockerfile.centos7-static-build`
  - `dev/docker/Dockerfile.centos7-gcc13-static-build`
  - `dev/docker/Dockerfile.centos8-static-build`
  - `dev/docker/Dockerfile.centos8-gcc13-static-build`
  - `dev/docker/Dockerfile.centos8-dynamic-build`
  - `dev/docker/Dockerfile.centos9-static-build`
  - `dev/docker/Dockerfile.centos9-dynamic-build`
  - `dev/docker/cudf/Dockerfile`
- Remove `wget` package installation where it was only used for Maven download
- Remove `PATH=/usr/lib/maven/bin` environment variables

## How was this patch tested?

- Local Docker build tests on CentOS 8 and CentOS 9 Stream confirmed that `./build/mvn` correctly auto-downloads Maven 3.9.12 and executes successfully
- Verified no remaining references to `setup-helper.sh` or `install_maven`
- Code review confirmed all mvn → ./build/mvn migrations are correct

Generated-by: GitHub Copilot (Claude Opus 4.5)

### DIFF
--- a/.github/workflows/util/install-resources.sh
+++ b/.github/workflows/util/install-resources.sh
@@ -16,114 +16,12 @@
 
 # Download Spark resources, required by some Spark UTs. The resource path should be set
 # for spark.test.home in mvn test.
+#
+# This file can be:
+# 1. Executed directly: ./install-resources.sh <spark-version> [install-dir]
+# 2. Sourced to use functions: source install-resources.sh; install_hadoop; setup_hdfs
 
 set -e
-
-# Installs Spark binary and source releases with:
-# 1 - spark version
-# 2 - hadoop version
-# 3 - scala version
-function install_spark() {
-  local spark_version="$1"
-  local hadoop_version="$2"
-  local scala_version="$3"
-  local spark_version_short=$(echo "${spark_version}" | cut -d '.' -f 1,2 | tr -d '.')
-  local scala_suffix=$([ "${scala_version}" == '2.13' ] && echo '-scala-2.13' || echo '')
-  local scala_suffix_short=$([ "${scala_version}" == '2.13' ] && echo '-scala2.13' || echo '')
-  local mirror_host='https://www.apache.org/dyn/closer.lua/'
-  local mirror_host2='https://mirror.lyrahosting.com/apache/' # Fallback mirror due to closer.lua slowness
-  local url_query='?action=download'
-  local checksum_suffix='sha512'
-  local url_path="spark/spark-${spark_version}/"
-  local local_binary="spark-${spark_version}-bin-hadoop${hadoop_version}${scala_suffix_short}.tgz"
-  local local_binary_checksum="${local_binary}.${checksum_suffix}"
-  local local_source="spark-${spark_version}.tgz"
-  local local_source_checksum="${local_source}.${checksum_suffix}"
-  local remote_binary="${mirror_host2}${url_path}${local_binary}${url_query}"
-  local remote_binary_checksum="${mirror_host}${url_path}${local_binary_checksum}${url_query}"
-  local remote_source="${mirror_host2}${url_path}${local_source}${url_query}"
-  local remote_source_checksum="${mirror_host}${url_path}${local_source_checksum}${url_query}"
-  local wget_opts="--no-verbose --no-check-certificate"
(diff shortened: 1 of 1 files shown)


dev/docker/Dockerfile.centos8-static-build
```diff
@@ -47,6 +46,6 @@ RUN set -ex; \
     export SSL_VERIFY=false; \
     bash ./dev/builddeps-veloxbe.sh --enable_vcpkg=ON --build_tests=ON --enable_s3=ON --enable_gcs=ON \
                                     --enable_hdfs=ON --enable_abfs=ON  build_arrow; \
-    mvn dependency:go-offline -Pbackends-velox -Piceberg -Pdelta -Pspark-3.5 -DskipTests; \
+    ./build/mvn dependency:go-offline -Pbackends-velox -Piceberg -Pdelta -Pspark-3.5 -DskipTests; \
```
zhouyuan, line 49: The folder for gluten will be deleted in the next step, so the downloaded maven is also removed?

yaooqinn: yes, deleted if there is no system-wide mvn found

zhouyuan: @yaooqinn 
For now it will download maven and extract to `${GLUTEN_HOME}/build`
https://github.com/apache/incubator-gluten/blob/main/build/mvn#L24

In the next step of `docker build` to reduce the docker image size we will remove the whole folder of `${GLUTEN_HOME}`
hence the maven is not actually installed in the docker image, is this correct?

yaooqinn: yes, download on demand if absent while calling `build/mvn`

yaooqinn: @zhouyuan, do we have any concern about this? From my perspective, this maven pull isn't heavy and can reduce the potential incompatibility caused by maven upgrade

philo-he: @yaooqinn, since `build/mvn` fully accepts using the installed maven in the environment, I feel it may be better to keep maven installed in the docker's system path to save CI resources, though it's a small cost.

If we are concerned about the incompatibility, maybe, we can add some check logic in build/mvn to only allow some versions to use. cc @zhouyuan

yaooqinn: > If we are concerned about the incompatibility, maybe, we can add some check logic in build/mvn to only allow some versions to use.

An old maven version caused quite a bit of unexpected issues during my lifespan spark contributing. I'd like to make the subsequent build steps use the up-to-date pom-defined maven version consistently

zhouyuan: @yaooqinn it's not conflict, would you please also install to /usr/local in the docker(copy to build/mvn to /usr/local)? otherwise the maven will be downloaded in each CI job

yaooqinn: Hi @zhouyuan in terms of networking, pulling a bigger docker image is similar to download a mvn tarball. In terms of consistency, it's much more difficult to find the issues both for human and AI-tooling. 

In addition, we can add action@cache optimization for `.mvn` later

yaooqinn:
Let me merge this first for further improvements

Issue #11402, opened 2026-01-13
[VL][BUG] Incorrect decimal casting from floating point value


Surbhi-Vijay:
### Backend

VL (Velox)

### Bug description

Decimal casting from floating point value is resulting in incorrect value.

` cast('100.12' as decimal(5, 1))`
Spark answer: `100.1`
Gluten answer: `10.0`

Same issue is happening with 

Repro Testcase:
```
  test("Incorrect decimal casting") {
    withSQLConf(SQLConf.ANSI_ENABLED.key -> "false") {
      withTable("dynparttest2") {
        Seq[(Integer, Integer)](
          (1, 1),
          (1, 3),
          (2, 3),
          (3, 3),
          (4, null),
          (5, null)
        ).toDF("key", "value").createOrReplaceTempView("src")

        // decimal
        sql("create table dynparttest2 (value int) partitioned by (pdec decimal(5, 1))")
        sql(
          """
            |insert into table dynparttest2 partition(pdec)
            | select count(*), cast('100.12' as decimal(5, 1)) as pdec from src
          """.stripMargin)
        checkAnswer(
          sql("select * from dynparttest2"),
          Seq(Row(6, new java.math.BigDecimal("100.1"))))
      }
    }
  }
```

Test logs
```
Results do not match for query:
Timezone: sun.util.calendar.ZoneInfo[id="America/Los_Angeles",offset=-28800000,dstSavings=3600000,useDaylight=true,transitions=185,lastRule=java.util.SimpleTimeZone[id=America/Los_Angeles,offset=-28800000,dstSavings=3600000,useDaylight=true,startYear=0,startMode=3,startMonth=2,startDay=8,startDayOfWeek=1,startTime=7200000,startTimeMode=0,endMode=3,endMonth=10,endDay=1,endDayOfWeek=1,endTime=7200000,endTimeMode=0]]
Timezone Env: 

== Parsed Logical Plan ==
'Project [*]
+- 'UnresolvedRelation [dynparttest2], [], false

== Analyzed Logical Plan ==
value: int, pdec: decimal(5,1)
Project [value#80, pdec#81]
+- SubqueryAlias spark_catalog.default.dynparttest2
   +- Relation spark_catalog.default.dynparttest2[value#80,pdec#81] parquet

== Optimized Logical Plan ==
Relation spark_catalog.default.dynparttest2[value#80,pdec#81] parquet

== Physical Plan ==
VeloxColumnarToRow
[...]

PR #11518, opened 2026-01-29
[GLUTEN-11402][VL] fix reading decimal partition key

zhouyuan:

decimal partition key is not supported yet. This patch make it to use fallback code path

## How was this patch tested?
new UT, pass GHA

Related issue: #11402

### DIFF
--- a/gluten-ut/test/src/test/scala/org/apache/gluten/sql/SQLQuerySuite.scala
+++ b/gluten-ut/test/src/test/scala/org/apache/gluten/sql/SQLQuerySuite.scala
@@ -43,7 +43,7 @@ class SQLQuerySuite extends WholeStageTransformerSuite {
     conf
   }
 
-  test("Incorrect decimal casting for partition read") {
+  testWithMinSparkVersion("Incorrect decimal casting for partition read", "4.0") {
     withSQLConf(SQLConf.ANSI_ENABLED.key -> "false") {
       withTable("dynparttest2") {
         val data =


gluten-substrait/src/main/scala/org/apache/gluten/execution/FileSourceScanExecTransformer.scala
```diff
@@ -166,6 +166,10 @@ abstract class FileSourceScanExecTransformerBase(
       return ValidationResult.failed(s"Unsupported columns scan in native.")
     }
 
+    if (getPartitionSchema.fields.exists(_.dataType.isInstanceOf[DecimalType])) {
```
liujiayi771: Is this limitation exclusive to `FileSourceScanExecTrasnformer`? Should we apply it to `BasicScanExecTransformer`?

zhouyuan:
related: https://github.com/facebookincubator/velox/pull/12910

gluten-substrait/src/main/scala/org/apache/gluten/execution/FileSourceScanExecTransformer.scala
zhouyuan: @liujiayi771 Thanks. I added in the BatchScanExec also. It seems BasicScanExec does not have the necessary members

baibaichen:

gluten-ut/spark40/src/test/scala/org/apache/spark/sql/GlutenDynamicPartitionPruningSuite.scala
```diff
@@ -77,6 +77,31 @@ abstract class GlutenDynamicPartitionPruningSuiteBase
     }
   }
 
+  testGluten("Incorrect decimal casting for partition read") {
```
Surbhi-Vijay: I think this testcase should be added in some other file. I had derived this testcase from "dynamic partition value test" in https://github.com/apache/spark/blob/master/sql/hive/src/test/scala/org/apache/spark/sql/hive/execution/SQLQuerySuite.scala.

It can be added to `GlutenSQLQueryTestSuite`.

Surbhi-Vijay (approved):

liujiayi771:
@zhouyuan Do you mean getPartitionSchema by 'necessary members'? I think it should be accessible within BasicScanExecTransformer.

baibaichen (approved):

PR #11481, opened 2026-01-23
[VL] Update CI Python to 3.10 for Spark 4.1 and enable ArrowEvalPythonExecSuite tests

ReemaAlzaid:
This pull request upgrades the Python version from 3.9 to 3.10 in the CI workflow and enables previously disabled Arrow UDF tests for Spark 4.1.

## What changes were proposed in this pull request?

1. **CI Workflow Update**: Updated `spark-test-spark41` job in `.github/workflows/velox_backend_x86.yml` to use Python 3.10 instead of Python 3.9
2. **Test Suite Update**: Enabled ArrowEvalPythonExecSuite tests for Spark 4.1 by updating `testWithMaxSparkVersion` from "4.0" to "4.1"

## Why are the changes needed?

Fixes #11400 - Spark 4.1 requires Python 3.10 as the minimum version. The previous Python 3.9 setup was causing ArrowEvalPythonExecSuite tests to fail on Spark 4.1.

## How was this patch tested?

The changes will be tested through the CI pipeline with the updated Python 3.10 environment for Spark 4.1 tests.

Issue #11511, opened 2026-01-28
[VL] CI: Support running Spark UT in centos9 docker image


marin-ma:
### Description

Currently in [.github/workflows/velox_backend_x86.yml](https://github.com/apache/incubator-gluten/blob/c61b2c0965901d9e2e0459a7a2b35be557cf1073/.github/workflows/velox_backend_x86.yml), the java unit tests are running on centos8 docker image. We should add new job to run the tests on centos9 and migrate these jobs to centos 9 in the future. 

### Gluten version

None

marin-ma:
cc: @zhouyuan

ReemaAlzaid:
take

ReemaAlzaid:
Hey @marin-ma, I made a PR covering some test cases as an initial beginning to test it out, also should we add Spark 4.1 to the CentOS 9 docker images?

zhouyuan:
closing as it's done

PR #11519, opened 2026-01-29
[VL][CI] Migrate Spark 4.1 tests to CentOS 9

ReemaAlzaid:

This PR adds CentOS 9 support for running Spark unit tests while maintaining existing CentOS 8 jobs for gradual migration. 
Related Issue: #11511 
Fixes: https://github.com/apache/incubator-gluten/pull/11481

## How was this patch tested?

Will be tested by CI pipeline execution

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/BackendSettingsApi.scala
@@ -19,6 +19,7 @@ package org.apache.gluten.backendsapi
 import org.apache.gluten.config.GlutenConfig
 import org.apache.gluten.execution.ValidationResult
 import org.apache.gluten.extension.columnar.transition.Convention
+import org.apache.gluten.sql.shims.SparkShimLoader
 import org.apache.gluten.substrait.rel.LocalFilesNode
 import org.apache.gluten.substrait.rel.LocalFilesNode.ReadFileFormat
 
@@ -91,6 +92,8 @@ trait BackendSettingsApi {
 
   def supportHashBuildJoinTypeOnRight: JoinType => Boolean = {
     case _: InnerLike | LeftOuter | FullOuter | LeftSemi | LeftAnti | _: ExistenceJoin => true
+    // LeftSingle is a Spark 4.0+ join type with same semantics as LeftOuter for build side.
+    case leftSingle if SparkShimLoader.getSparkShims.isLeftSingleJoinType(leftSingle) => true
     case _ => false
   }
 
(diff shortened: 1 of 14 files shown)


philo-he:
@ReemaAlzaid, thanks for the PR. I'm concerned that this may introduce too much test redundancy, considering these Spark tests are being verified on Centos 8. If the purpose is enabling some Spark tests with python 3.11, can we install this python version by compiling from source to keep using Centos 8? cc @zhouyuan

If we indeed want to migrate from Centos 8 to 9 for Spark tests, maybe we can directly do the replacement. There seems to be no need to keep the both categories. In addition, we may re-use the existing build native lib job, i.e., for tests on Centos 9, download the native binary built on Centos 7. I assume there is no compatibility issue.

.github/workflows/velox_backend_x86.yml
```diff
@@ -1510,3 +1510,326 @@ jobs:
             **/target/*.log
             **/gluten-ut/**/hs_err_*.log
             **/gluten-ut/**/core.*
+
+  build-native-lib-centos-9:
```
zhouyuan: we should download the native libs from `build-native-lib-centos-9`

.github/workflows/velox_backend_x86.yml
```diff
@@ -1510,3 +1510,326 @@ jobs:
...
+          name: velox-native-lib-centos-9-${{github.sha}}
+          path: ./cpp/build/releases/
+          if-no-files-found: error
+      - uses: actions/upload-artifact@v4
+        with:
+          name: arrow-jars-centos-9-${{github.sha}}
+          path: .m2/repository/org/apache/arrow/
+          if-no-files-found: error
+
+  spark-test-spark41-centos9:
```
zhouyuan: let's move the `spark41` tests to centos 9 firstly

.github/workflows/velox_backend_x86.yml
zhouyuan: Typo: `build-native-lib-centos-7` is what i meant for

.github/workflows/velox_backend_x86.yml
```diff
@@ -1467,7 +1467,7 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       SPARK_TESTING: true
-    container: apache/gluten:centos-8-jdk17
+    container: apache/gluten:centos-7-jdk17
```
zhouyuan:
```suggestion
    container: apache/gluten:centos-9-jdk17
```

zhouyuan:
@ReemaAlzaid since it may take some time to get the patches ready in velox, could you please check if we could exclude the new timezone(America/Coyhaique) in those tests? here's one example:
https://github.com/apache/incubator-gluten/blob/main/gluten-ut/spark34/src/test/scala/org/apache/spark/sql/catalyst/expressions/GlutenTryCastSuite.scala#L40

baibaichen:
@ReemaAlzaid would you pelase create new pr for this https://github.com/apache/incubator-gluten/pull/11481/changes#diff-d34bc9515f676ce2605c33b00515f94ff1d6b0f6eb1be32670f9783d79c85968

PR #11523, opened 2026-01-30
Fix window to aggregate conversion with ordering expression validation

lgbo-ustc:

- Add hasOrderExpression method to validate that ordering expressions are not in partition spec
- Include additional test case for window function with row_number over multiple partition columns
- Prevents incorrect window to aggregate conversion when order expressions don't match partition spec

## How was this patch tested?

tested by UTs

No


lgbo-ustc:

Issue #11514, opened 2026-01-28
[VL] Add task barrier support for NativePlanEvaluator to make Velox task reusable


zhztheplayer:
### Description

Currently ColumnarBatchOutIterator returned by `NativePlanEvaluator#createKernelWithBatchIterator` is disposable. It could only serve one input stream and output the result data one single time.

We'd add support for Velox's task barrier to make it reusable. 

Fixing this will significantly improve performance of the [ongoing native Delta statistics writer PR](https://github.com/apache/incubator-gluten/issues/11169) where we are creating a new Velox task for every single written partition, which can become super slow when there are thousands of partitions to write. 

Reference, Velox task barrier:

https://facebookincubator.github.io/velox/develop/task-barrier.html

### Gluten version

Main

PR #11419, opened 2026-01-14
[GLUTEN-10215][VL] Delta write: Native statistics tracker to eliminate C2R overhead

zhztheplayer:
## Description

Currently, there is a C2R converter to get all rows from a columnar batch being written to calculate and gather Delta file statistics. This is inefficient given we expected all operations related to write can be offloaded to native.

The patch adds a native job statistics tracker for Delta write to eliminate this C2R overhead. This tracker is backed by an asynchronous **barrier-enabled** Velox aggregation task (for more information about Velox task barrier, see [this](https://facebookincubator.github.io/velox/develop/task-barrier.html)), where all the data to write is globally aggregated into statistics rows, each of which is for one single written Parquet file.

Existing tests under directory `backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/test` can cover the change.

Depends on #11405

Related issue: #10215

Also fixes #11514

## Performance

The PR is benchmarked by writing TPC-DS SF10 tables in Delta format. Resource is 8 cores + 20 GiB RAM. Typical size of each written file is ~10 MiB.

|Vanilla | Before | After|
|-------|--------|-----|
|467.8s| 478.4s| 290s|

Before this PR, Gluten's speedup on Delta write is observed at **-2.22%**.
After this PR, Gluten's speedup on Delta write is observed at **61.31%**.

Detailed benchmark results per table are as follows:

perf

PR #11527, opened 2026-01-30
[GLUTEN-11514][VL] Refactor plan execution by adding `addIteratorSplits` and `noMoreSplits` methods to the plan execution API

zhztheplayer:
Split plan execution via `NativePlanEvaluator` / `ColumnarBatchOutIterator` into 3 APIs:

1. `NativePlanEvaluator#create`
2. `ColumnarBatchOutIterator#addIteratorSplits`
3. `ColumnarBatchOutIterator#noMoreSplits`

This allows iterator-based splits to be added later than when the task is started. Useful when `NativePlanEvaluator` is used individually, out of the transformer code routine.

This is a prerequisite for https://github.com/apache/incubator-gluten/issues/11514 and https://github.com/apache/incubator-gluten/pull/11419.

Tested by existing tests and https://github.com/apache/incubator-gluten/pull/11419

### DIFF
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -358,6 +358,7 @@ void WholeStageResultIterator::constructPartitionColumns(
 }
 
 void WholeStageResultIterator::addIteratorSplits(const std::vector<std::shared_ptr<ResultIterator>>& inputIterators) {
+  GLUTEN_CHECK(!allSplitsAdded, "Method addIteratorSplits should not be called since all splits has been added to the Velox task.");
   // Create IteratorConnectorSplit for each iterator
   for (size_t i = 0; i < streamIds_.size() && i < inputIterators.size(); ++i) {
     if (inputIterators[i] == nullptr) {


cpp/velox/operators/plannodes/CudfVectorStream.h
```diff
@@ -18,21 +18,90 @@
...
+#include "velox/exec/Task.h"
 #include "velox/experimental/cudf/exec/CudfOperator.h"
 #include "velox/experimental/cudf/exec/Utilities.h"
 #include "velox/experimental/cudf/exec/VeloxCudfInterop.h"
 #include "velox/experimental/cudf/vector/CudfVector.h"
 
 namespace gluten {
 
-class CudfVectorStream : public RowVectorStream {
+class CudfVectorStreamBase {
```
zhztheplayer, line 33: cc @jinchengchenghh 

`RowVectorStream`'s code is being inlined into `CudfVectorStreamBase` here.

jinchengchenghh: ok

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -355,17 +357,39 @@ void WholeStageResultIterator::constructPartitionColumns(
...
+      continue;
+    }
+    auto connectorSplit = std::make_shared<IteratorConnectorSplit>(
+        kIteratorConnectorId, inputIterators[i]);
+    exec::Split split(folly::copy(connectorSplit), -1);
+    task_->addSplit(streamIds_[i], std::move(split));
+  }
+}
+
+void WholeStageResultIterator::noMoreSplits() {
```
zhouyuan, line 374: Would this work if calling from different threads? e.g. thread A called addIteratorSplits(), but thread B called on noMoreSplits()

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -355,17 +357,39 @@ void WholeStageResultIterator::constructPartitionColumns(
   }
 }
 
-void WholeStageResultIterator::tryAddSplitsToTask() {
-  if (noMoreSplits_) {
+void WholeStageResultIterator::addIteratorSplits(const std::vector<std::shared_ptr<ResultIterator>>& inputIterators) {
+  // Create IteratorConnectorSplit for each iterator
```
zhouyuan, line 362: should we also guard for `allSplitsAdded = true`?

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: It's not guaranteed thread-safe. In the Delta PR, the plan is still created and executed in one single thread so no requirement for thread safety on this API now.

cpp/velox/compute/WholeStageResultIterator.cc
zhztheplayer: Yes, good idea.

zhztheplayer:
@zhouyuan 

> Thanks! I understand the basic logic now - just one question on the concurrent execution threading model

The change is not supposed to support a concurrent execution model - it's still single-threaded execution. The goal was to add the barrier API, so we can reuse the Velox task in the same thread but for newer inputs.

Let me know any remaining issues. cc @rui-mo

PR #11528, opened 2026-01-30
[VL] Adding nightly package for enhanced features

zhouyuan:
Only covered JDK17 here to reduce the package size in apache nightly

This patch adds nightly build for enhanced repo
also fixed the spark-41 packaging due to iceberg release is not ready

## How was this patch tested?
pass GHA

### DIFF
--- a/.github/workflows/velox_nightly.yml
+++ b/.github/workflows/velox_nightly.yml
@@ -485,6 +485,9 @@ jobs:
           cd $GITHUB_WORKSPACE/ && \
           ./build/mvn clean install -Pspark-3.4 -Pjava-17 -Pbackends-velox -Pceleborn -Puniffle -Piceberg -Phudi -Pdelta -Ppaimon -DskipTests -Dmaven.source.skip
           ./build/mvn clean install -Pspark-3.5 -Pjava-17 -Pbackends-velox -Pceleborn -Puniffle -Piceberg -Phudi -Pdelta -Ppaimon -DskipTests -Dmaven.source.skip
+          ./build/mvn clean install -Pspark-4.0 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pceleborn -Puniffle -Piceberg -Phudi -Pdelta -Ppaimon -DskipTests -Dmaven.source.skip
+          # iceberg support for spark-4.1 with java-17 is not avaiable, skip it first
+          ./build/mvn clean install -Pspark-4.1 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pceleborn -Puniffle -Phudi -Pdelta -Ppaimon -DskipTests -Dmaven.source.skip
       - name: Upload bundle package
         uses: actions/upload-artifact@v4
         with:


.github/workflows/velox_nightly.yml
```diff
@@ -435,3 +438,89 @@ jobs:
...
+          yum update -y && yum install -y java-17-openjdk-devel wget
+          $SETUP install_maven
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+      - name: Build package for Spark
+        run: |
+          cd $GITHUB_WORKSPACE/ && \
+          export MAVEN_HOME=/usr/lib/maven && \
+          export PATH=${PATH}:${MAVEN_HOME}/bin && \
```
philo-he: `$SETUP install_maven` will set environment variable, so it seems the above two lines are not required.

https://github.com/apache/incubator-gluten/blob/02c776f50d77ec965152b2033388bec5f27dbdd3/.github/workflows/util/setup-helper.sh#L31

.github/workflows/velox_nightly.yml
```diff
@@ -435,3 +438,89 @@ jobs:
...
+  # Build velox with enhanced features
+  build-native-lib-enhanced:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Get Ccache
+        uses: actions/cache/restore@v4
+        with:
+          path: '${{ env.CCACHE_DIR }}'
+          key: ccache-centos7-release-default-${{github.sha}}
```
philo-he: This will download the ccache from regular native lib build, not the enhanced lib build. Are they quite similar? If so, it may be good to share the ccache. Suggest adding one comment to clarify ccache for enhanced lib may be required in the future.

.github/workflows/velox_nightly.yml
```diff
@@ -435,3 +438,89 @@ jobs:
...
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+      - name: Build package for Spark
+        run: |
+          cd $GITHUB_WORKSPACE/ && \
+          export MAVEN_HOME=/usr/lib/maven && \
+          export PATH=${PATH}:${MAVEN_HOME}/bin && \
+          mvn clean install -Pspark-3.4 -Pjava-17 -Pbackends-velox -Pceleborn -Puniffle -Piceberg -Phudi -Pdelta -Ppaimon -DskipTests -Dmaven.source.skip
+          mvn clean install -Pspark-3.5 -Pjava-17 -Pbackends-velox -Pceleborn -Puniffle -Piceberg -Phudi -Pdelta -Ppaimon -DskipTests -Dmaven.source.skip
+          mvn clean install -Pspark-4.0 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pceleborn -Puniffle -Piceberg -Phudi -Pdelta -Ppaimon -DskipTests -Dmaven.source.skip
```
philo-he: Maybe, we should use `build/mvn` as Kent's WIP PR suggested?

.github/workflows/velox_nightly.yml
zhouyuan: not necessary with latest code, removed

.github/workflows/velox_nightly.yml
zhouyuan: it can also speedup with normal branch. fix to download the enhanced lib

.github/workflows/velox_nightly.yml
zhouyuan: fixed

zhouyuan:
@PHILO-HE thanks for the review. The job contains some redundant code now, we could refine it with matrix and different parameters.

PR #11538, opened 2026-02-02
[CORE] Copy tags when new local sort node is added by EnsureLocalSortRequirements rule

Surbhi-Vijay:
`EnsureLocalSortRequirements` rule in Gluten adds local sort node in certain scenarios. The newly added node was missing tags from original child node.

## How was this patch tested?
Locally, using UTs.
Internally, we add a rule to add some tags when streaming nodes are present. These tags were getting missed and testcases were failing. With this change testcases are passing.

No


Surbhi-Vijay:
cc: @baibaichen @weiting-chen

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/EnsureLocalSortRequirements.scala
```diff
@@ -38,6 +38,7 @@ object EnsureLocalSortRequirements extends Rule[SparkPlan] {
       requiredOrdering: Seq[SortOrder]): SparkPlan = {
     // FIXME: HeuristicTransform is costly. Re-applying it may cause performance issues.
     val newChild = SortExec(requiredOrdering, global = false, child = originalChild)
+    newChild.copyTagsFrom(originalChild)
```
zhouyuan, line 41: @Surbhi-Vijay would you please also add the UT you mentioned to guard on this?

PR #11545, opened 2026-02-02
[MINOR] Add GCS Maven Central mirror for faster dependency downloads

yaooqinn:

Add Google Cloud Storage Maven Central mirror as the primary repository for faster dependency downloads during builds.

### Changes:
- Add `gcs-maven-central-mirror` repository pointing to `https://maven-central.storage-download.googleapis.com/maven2/`
- Keep standard Maven Central as fallback
- Apply to both `<repositories>` and `<pluginRepositories>`

The GCS Maven Central mirror is geographically distributed and typically provides faster download speeds than the default Maven Central. This is the same pattern used by Apache Spark.

No. This only affects build-time dependency resolution.

## How was this patch tested?

- `./build/mvn validate` passes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -483,6 +483,57 @@
     </dependencies>
   </dependencyManagement>
 
+  <!-- Use Google Cloud Storage mirror for faster Maven Central downloads -->
+  <repositories>
+    <repository>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>gcs-maven-central-mirror</id>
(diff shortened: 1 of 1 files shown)


FelixYBW (approved):

FelixYBW:
@baibaichen do we still need .m2 cache with this patch?

philo-he:
Thanks for the PR. I'm not sure if gcs maven central mirror is stable in China. @yaooqinn, have you tried without proxies?

yaooqinn:
Spark has this for years, and I don't have any extra settings in both my personal and work envs

yaooqinn:
> do we still need .m2 cache with this patch?

Cache remains an optimization upon this

PR #11561, opened 2026-02-04
[MINOR] Respect Spark bloom filter config

jinchengchenghh:
Respect Spark four configs, if not exists, Velox will use its default config value.
```
spark.sql.optimizer.runtime.bloomFilter.expectedNumItems
spark.sql.optimizer.runtime.bloomFilter.maxNumItems
spark.sql.optimizer.runtime.bloomFilter.numBits
spark.sql.optimizer.runtime.bloomFilter.maxNumBits
```


FelixYBW:
We may remove the gluten config if there are existed ones for Spark.

FelixYBW:
how about config:
```
spark.sql.optimizer.runtime.bloomFilter.applicationSideScanSizeThreshold
spark.sql.optimizer.runtime.bloomFilter.creationSideThreshold
spark.sql.optimizer.runtime.bloomFilter.maxNumItems
```

jinchengchenghh:
spark.sql.optimizer.runtime.bloomFilter.applicationSideScanSizeThreshold
spark.sql.optimizer.runtime.bloomFilter.creationSideThreshold
These above two configs decide if insert bloom filter for the join, which take effect in Spark java code.

spark.sql.optimizer.runtime.bloomFilter.maxNumItems
velox does not have corresponding config, I will check if we can respect the config in Velox.

jinchengchenghh:
After spark3.4 add this PR, the num bits algorithm is changed, https://github.com/apache/spark/pull/37697/, and it consider the max num items config, I will update Velox function to respect this behavior

jinchengchenghh:
Before that PR, max num items is only used in checkInputDataTypes(Spark analysis) to check the config expectedItems is reasonable

FelixYBW:
> spark.sql.optimizer.runtime.bloomFilter.applicationSideScanSizeThreshold spark.sql.optimizer.runtime.bloomFilter.creationSideThreshold These above two configs decide if insert bloom filter for the join, which take effect in Spark java code.

So they are transparent to Gluten, right?

PR #146, opened 2022-05-07
[Gluten-145] Fix library loading

rui-mo:
## What changes were proposed in this pull request?

Fixed the library loading issue when the release directory is not accessible.

Fixed https://github.com/oap-project/gluten/issues/145.

## How was this patch tested?

Verified in a cluster test.

PR #11566, opened 2026-02-05
[VL] Support function ToPrettyString

jinchengchenghh:
The expression is available since Spark 3.5. Prior versions use CAST for binary output.
Spark 4.0 introduces the optional config spark.sql.binaryOutputStyle.
This PR supports only the default behavior (HEX_DISCRETE), which is used when the config is unset.

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/utils/CHExpressionUtil.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/utils/CHExpressionUtil.scala
@@ -174,6 +174,7 @@ object CHExpressionUtil {
 
   final val CH_BLACKLIST_SCALAR_FUNCTION: Map[String, FunctionValidator] = Map(
     SPLIT_PART -> DefaultValidator(),
+    TO_PRETTY_STRING -> DefaultValidator(),
     TO_UNIX_TIMESTAMP -> UnixTimeStampValidator(),
     UNIX_TIMESTAMP -> UnixTimeStampValidator(),
     SEQUENCE -> SequenceValidator(),


jinchengchenghh:
The version spark 3.5 works well, but spark4.0 and Spark 4.1 introduces  `BINARY_OUTPUT_STYLE`, we only support the default HEX_DISCRETE as spark 3.5, so I will update the test after spark 4.0
```
val BINARY_OUTPUT_STYLE = buildConf("spark.sql.binaryOutputStyle")
    .doc("The output style used display binary data. Valid values are 'UTF-8', " +
      "'BASIC', 'BASE64', 'HEX', and 'HEX_DISCRETE'.")
    .version("4.0.0")
    .enumConf(BinaryOutputStyle)
    .createOptional
```

jinchengchenghh:
Could you help approve this PR? Thanks! @rui-mo

rui-mo (approved):
Thanks.

ArnavBalyan:
Thanks

jinchengchenghh:
I try to add a unit test
```
  test("to_pretty_string") {
    val df = spark.sql("SELECT isnull(1)")
    df.show()
    print("plan is " + df.queryExecution.executedPlan.toString())
    checkGlutenPlan[ProjectExecTransformer](df)
  }
```
+-----------+
|(1 IS NULL)|
+-----------+
|      false|
+-----------+

plan is VeloxColumnarToRow
+- ^(1) ProjectExecTransformer [isnull(1) AS (1 IS NULL)#146]
   +- ^(1) InputIteratorTransformer[]
      +- RowToVeloxColumnar
         +- *(1) Scan OneRowRelation[]

But the plan does not contain the to_pretty_string, so it is hard to verify if it is really offload to native unless we can check the native plan.

But ToPrettyString is really offloaded since the unit test getRows::binary failed with this PR

Issue #11550, opened 2026-02-03
Spark 4.x: Tracking disabled test suites


baibaichen:
Please leave a comment and mark on the TODO tests in this issue if you want to contribute

## Backend
VL(Velox)

## Bug Description

These suites were commented out in https://github.com/apache/incubator-gluten/pull/11512 with TODO markers indicating they need to be fixed and re-enabled.

## Context

The test suites below have been disabled due to various failures. Each table shows the status for both Spark 4.0 and 4.1:
- 🔴 = Suite is commented out (disabled)
- 🟢 = Suite is enabled

The failure count (when available) is shown in the status column.

---

### org.apache.spark.sql

| Suite | Spark 4.0 | Spark 4.1 | Owner | Comments |
|-------|-----------|-----------|-------|----------|
| `GlutenDataFrameSubquerySuite` | 🟢 | 🟢 | @baibaichen | Fixed in #11727 |
| `GlutenExplainSuite` | 🔴 (1 failures) | 🔴 (1 failures) | @Surbhi-Vijay  |  |
| `GlutenLogQuerySuite` | 🔴 (2 failures) | 🔴 (2 failures) | @kapilks  |  |
| `GlutenSetCommandSuite` | 🔴 (1 failures) | 🔴 (1 failures) | @Surbhi-Vijay |  |
| `GlutenSingleLevelAggregateHashMapSuite` | 🔴 (1 failures) | 🔴 (1 failures) | @kapilks |  |
| `GlutenSparkSessionJobTaggingAndCancellationSuite` | 🔴 (1 failures) | 🔴 (1 failures) | |  |
| `GlutenTwoLevelAggregateHashMapSuite` | 🔴 (1 failures) | 🔴 (1 failures) | @kapilks |  |
| `GlutenTwoLevelAggregateHashMapWithVectorizedMapSuite` | 🔴 (1 failures) | 🔴 (1 failures) | @kapilks |  |
| `GlutenVariantEndToEndSuite` | 🟢 | 🟢 | @baibaichen | Fixed in #11726 |
| `GlutenVariantShreddingSuite` | 🟢 | 🟢 | @baibaichen | Fixed in #11726 |
| `GlutenXmlFunctionsSuite` | 🟢 | 🟢 | @baibaichen | Fixed in #11725 |

### catalyst.expressions

| Suite | Spark 4.0 | Spark 4.1 | Owner | Comments |
|-------|-----------|-----------|-------|----------|
[...]

PR #11571, opened 2026-02-05
[GLUTEN-11550][UT][Spark4.x] Remove GlutenSingleLevelAggregateHashMapSuite, GlutenTwoLevelAggregateHashMapSuite, GlutenTwoLevelAggregateHashMapWithVectorizedMapSuite

kapilks:
Remove following UTs: 
GlutenSingleLevelAggregateHashMapSuite
GlutenTwoLevelAggregateHashMapSuite
GlutenTwoLevelAggregateHashMapWithVectorizedMapSuite

Above suites runs UTs similar to `GlutenDataFrameAggregateSuite`. Only difference being that wholestage codegen is enabled or not and multi level aggregation map. These are specific to Spark implementation only.
So essentially same as GlutenDataFrameAggregateSuite. So I think these suites would be redundant.

Updated GlutenDataFrameAggregateSuite to add overridden UT for some disables cases.

## How was this patch tested?

UTs.

No

Related issue: #11550


kapilks:
@baibaichen Let me know if we still want to retain these suites?

baibaichen:
> @baibaichen Let me know if we still want to retain these suites?

LGTM to remove them.

Can we keep a small doc listing the excluded/removed suites (and why), so it’s easier to track and revisit later via AI?

kapilks:
@baibaichen Please review, added a doc.

PR #11573, opened 2026-02-06
[MINOR] Remove scan filter extraction logs from `BatchScanExecTransformerBase`

Zouxxyy:

Only Spark's built-in FileScan provides dataFilters; other DSV2 sources don't expose scan filters. The info is unnecessarily logged in large volume (e.g., with Paimon or Iceberg).

No


liujiayi771:
We have observed that when using Paimon, the console is flooded with a large number of such logs.

PR #11586, opened 2026-02-07
[VL] Enhance VeloxIceberg documentation with SQL options

FelixYBW:
Add Iceberg options support status


FelixYBW:
@jinchengchenghh Do we support the write options? I didn't seach any of them in Gluten source code

PR #11606, opened 2026-02-11
[VL] Adding configurations on max write file size

zhouyuan:
Adding config for max write file size in Velox

## How was this patch tested?
pass GHA
Velox UT


backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -274,6 +274,16 @@ object VeloxConfig extends ConfigRegistry {
       .checkValue(_ > 0, "must be a positive number")
       .createWithDefault(10000)
 
+  val MAX_TARGET_FILE_SIZE_SESSION =
+    buildConf("spark.gluten.sql.columnar.backend.velox.maxTargetFileSizeSession")
```
FelixYBW: what does Session mean here?

docs/velox-configuration.md
```diff
@@ -48,6 +48,7 @@ nav_order: 16
...
 | spark.gluten.sql.columnar.backend.velox.maxSpillRunRows                          | 3M                | The maximum row size of a single spill run                                                                                                                                                                                                                                                                                                                                                                                                            |
+| spark.gluten.sql.columnar.backend.velox.maxTargetFileSizeSession                 | 0b                | The target file size for each output file when writing data. 0 means no limit on target file size, and the actual file size will be determined by other factors such as max partition number and shuffle batch size.                                                                                                                                                                                                                                  |
```
FelixYBW: Does it map to iceberg's `write.target-file-size-bytes`? and honor `spark.sql.iceberg.advisory-partition-size`? If so let's honor this config in Gluten as well.

If it only take effect on iceberg, we may just reuse iceberg's config instead of a new config.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
jinchengchenghh: Please remove Session suffix, this is Velox code config type suffix, not the config itself

docs/velox-configuration.md
zhouyuan: As of today Velox is using `kMaxTargetFileSize` to control the parquet write file size, so it will impact on all parquet write. In current iceberg write code path, this is parameter is picked to control the parquet size in each partition.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
zhouyuan: done

docs/velox-configuration.md
zhouyuan: In iceberg java the logic for partition control is: 
https://github.com/apache/iceberg/blob/main/spark/v3.5/spark/src/main/java/org/apache/iceberg/spark/SparkWriteConf.java#L695-L702

```
        .option(SparkWriteOptions.ADVISORY_PARTITION_SIZE)
        .sessionConf(SparkSQLProperties.ADVISORY_PARTITION_SIZE)
        .tableProperty(TableProperties.SPARK_WRITE_ADVISORY_PARTITION_SIZE_BYTES)
        .defaultValue(defaultValue)
```
write_options >  session > table property

jinchengchenghh: This config for Iceberg should get from `SparkWrite` as codec, https://github.com/apache/iceberg/blob/main/spark/v4.1/spark/src/main/java/org/apache/iceberg/spark/source/SparkWrite.java#L138

zhouyuan: this config also work for hive parquet write, not only for iceberg

jinchengchenghh: Yes, iceberg should draft a new PR to respect its target file size.

PR #11619, opened 2026-02-15
[CORE] Use ArrayList instead of LinkedList in Spillers

acvictor:

This PR replaces LinkedList with ArrayList in Spillers.AppendableSpillerList. The list is only used for two operations: add() (append) and sequential for-each iteration in `spill()`. ArrayList is more efficient for both: append is amortized O(1) without allocating a Node object per element, and iteration benefits from sequential memory layout (CPU cache locality). LinkedList is faster for frequent mid-list insertions/removals, which this class never does.  

## How was this patch tested?

Existing unit tests.

No


acvictor:
@zhztheplayer can you please review this PR? Thanks!

PR #11620, opened 2026-02-16
[MINOR][VL] Fix throw-new anti-pattern in FileReaderIterator.cc

acvictor:
This PR replaces `throw new GlutenException(...)` with `throw GlutenException(...)` in `FileReaderIterator::getInputIteratorFromFileReader`.  `throw new` will allocate the exception object on the heap, but standard catch blocks catch by reference (`catch (const std::exception& e)`), so the pointer will never freed causing a memory leak on the exception path.

## How was this patch tested?
Existing tests

No


acvictor:
@marin-ma can you please review this PR? Thanks in advance!

PR #11623, opened 2026-02-17
[MINOR][VL] Use resize() instead of reserve() in getEmitInfo

acvictor:

This PR replaces `reserve()` with `resize()` for `projectNames` and `expressions` vectors in `getEmitInfo()`. `std::vector::reserve()` allocates capacity but does not change `size()`. The subsequent loop assigns elements via `operator[]` which accesses indices beyond `size()` and this is undefined behaviour.

## How was this patch tested?

Existing tests.

No


acvictor:
@rui-mo can you please review this minor change? Thanks in advance!

PR #11628, opened 2026-02-19
[VL] Reduce triple map lookup to single find in dictionary serialization

acvictor:
This PR replaces three separate map look ups on the same key with a single `find()` and iterator reuse in `ArrowShuffleDictionaryWriter::serialize()`

## How was this patch tested?
Existing UTs

No

### DIFF
--- a/cpp/velox/shuffle/ArrowShuffleDictionaryWriter.cc
+++ b/cpp/velox/shuffle/ArrowShuffleDictionaryWriter.cc
@@ -404,14 +404,14 @@ arrow::Status ArrowShuffleDictionaryWriter::serialize(arrow::io::OutputStream* o
   ARROW_RETURN_NOT_OK(out->Write(bitMap.data(), bitMapSize));
 
   for (auto fieldIdx : dictionaryFields_) {
+    auto it = dictionaries_.find(fieldIdx);
     GLUTEN_DCHECK(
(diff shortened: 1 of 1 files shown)


acvictor:
@marin-ma can you please review this PR? Thanks in advance!

cpp/velox/shuffle/ArrowShuffleDictionaryWriter.cc
```diff
@@ -404,14 +404,14 @@ arrow::Status ArrowShuffleDictionaryWriter::serialize(arrow::io::OutputStream* o
   ARROW_RETURN_NOT_OK(out->Write(bitMap.data(), bitMapSize));
 
   for (auto fieldIdx : dictionaryFields_) {
+    auto it = dictionaries_.find(fieldIdx);
     GLUTEN_DCHECK(
-        dictionaries_.find(fieldIdx) != dictionaries_.end(),
+        it != dictionaries_.end(),
```
jinchengchenghh, line 409: GLUTEN_DCHECK means the code only executes on debug compilation mode, so we don't need to add complexity for it. Only if the code logic has something wrong, the check mai fail.

marin-ma: @jinchengchenghh This change makes sense as `dictionaries_.erase(fieldIdx)` also performs another lookup.

PR #11636, opened 2026-02-20
[VL] Allocate VarChar sub-messages on protobuf arena in VeloxToSubstraitExpr

acvictor:
This PR replaces `new` allocation of `VarChar` sub-messages with `Arena::CreateMessage` to match the arena-allocated parent in `VeloxToSubstraitExpr.cc`.  When a protobuf message is arena-allocated, its destructor never runs so children attached via `set_allocated_*` that were heap-allocated with `new` are never deleted. Every other sub-message in this file already uses
   `Arena::CreateMessage` correctly and `VarChar` was the only exception.

## How was this patch tested?

Existing UTs

No


acvictor:
@rui-mo can you please review this PR? Thanks in advance!

rui-mo (approved):

acvictor:
@rui-mo thanks for the review and approvals! Please let me know if there’s anything else needed from my side.

PR #11643, opened 2026-02-23
[VL] Diable spark join keys rewrite

JkSelf:

When offloading Hash Join to Velox, Gluten currently cannot fully leverage the performance benefits of the Hash Join Key Rewriting feature. In fact, enabling this feature can lead to performance degradation.

During our optimization of BHJ—aimed at building the hash table only once per executor—we discovered that Dynamic Filter Push-down DFP is enabled when Key Rewriting is enabled. This occurs because the rewriting process introduces CAST(int AS long) expressions to both the Build and Probe sides within the Project operator. Since Velox only supports filter push-down on columns without complex expressions, these added casts block the optimization, resulting in a performance degradation.

To address this, we have disabled the Key Rewriting feature specifically within the HashJoin#canRewriteAsLongType method.

## How was this patch tested?

Existing unit tests

No

### DIFF
--- a/gluten-ut/spark33/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
+++ b/gluten-ut/spark33/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
@@ -533,6 +533,8 @@ class VeloxTestSettings extends BackendTestSettings {
     .excludeByPrefix("SPARK-35675")
 
   enableSuite[GlutenBroadcastJoinSuite]
+    // Disable hash join key rewrite feature in Gluten
+    .exclude("join key rewritten")
     .exclude("Shouldn't change broadcast join buildSide if user clearly specified")
     .exclude("Shouldn't bias towards build right if user didn't specify")
     .exclude("SPARK-23192: broadcast hint should be retained after using the cached data")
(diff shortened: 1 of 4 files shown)


backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
```diff
@@ -501,7 +501,9 @@ object VeloxBackendSettings extends BackendSettingsApi {
     (conf.isUseGlutenShuffleManager || conf.shuffleManagerSupportsColumnarShuffle)
   }
 
-  override def enableJoinKeysRewrite(): Boolean = false
+  override def enableJoinKeysRewrite(): Boolean = {
+    GlutenConfig.get.hashJoinKeysRewrite
```
jinchengchenghh, line 505: Looks like the previous value is false?

shims/spark33/src/main/scala/org/apache/spark/sql/execution/joins/HashJoin.scala
```diff
@@ -0,0 +1,781 @@
...
+ */
+private[joins] case class HashedRelationInfo(
+    relationTerm: String,
+    keyIsUnique: Boolean,
+    isEmpty: Boolean)
+
+trait HashJoin extends JoinCodegenSupport {
+  def buildSide: BuildSide
+
+  override def simpleStringWithNodeId(): String = {
```
jinchengchenghh, line 52: Why do you need to copy the file?

PR #11648, opened 2026-02-24
[VL] Adding context executor

zhouyuan:

This patch adds the context executor for Velox backend. 
By default it's disabled

## How was this patch tested?
pass GHA


FelixYBW:
Does the patch enable multiple thread execution in Velox? if so let's mark it as experimental

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -213,10 +213,18 @@ WholeStageResultIterator::WholeStageResultIterator(
 }
 
 std::shared_ptr<velox::core::QueryCtx> WholeStageResultIterator::createNewVeloxQueryCtx() {
+  int cpuThreads = veloxCfg_->get<int32_t>(kVeloxCpuExecutorThreads, kVeloxCpuExecutorThreadsDefault);
+  folly::Executor* executor = nullptr;
+  if (cpuThreads > 0) {
+    auto ctxExecutor = std::make_unique<folly::CPUThreadPoolExecutor>(cpuThreads);
```
jinchengchenghh, line 219: Should reserve the ctxExecutor, otherwise it is released after the if block

zhouyuan:
not necessary now

PR #11651, opened 2026-02-25
[GLUTEN-4889][VL] feat: Support approx_percentile aggregate function

Yizhou-Yang:
### What

Add Velox `approx_percentile` support for Spark.

### Why

Velox uses **KLL sketch** while Spark uses **GK algorithm** — their intermediate data formats are incompatible (KLL: 9-field StructType vs GK: single BinaryType buffer). This means fallback between Velox and Spark requires separate handling.

### How

- **`VeloxApproximatePercentile`**: A `DeclarativeAggregate` with 9 `aggBufferAttributes` matching Velox's KLL sketch layout.
- **Spark-side KLL implementation** (`KllSketchHelper`/`KllSketchAdd`/`KllSketchMerge`/`KllSketchEval`): Simplified KLL operations for fallback, binary-compatible with Velox's C++ accumulator.
- **`ApproxPercentileRewriteRule`**: Rewrites Spark's `ApproximatePercentile` to the Velox-compatible version.
- **All 4 fallback modes supported**: Full offload, partial fallback, final fallback, full fallback.

### Key decisions

- Accuracy stored as `IntegerType` (Spark's original value); Velox computes `epsilon = 1.0/accuracy` internally.
- KLL chosen over GK for Spark-side fallback to maintain intermediate data compatibility with Velox.

### Velox dependency

[facebookincubator/velox#16320](https://github.com/facebookincubator/velox/pull/16320)

---

Related issue: #4889

----------

### Testing

Velox uses the **KLL sketch** algorithm for `approx_percentile`, while Spark uses the **GK (Greenwald-Khanna)** algorithm. Both are approximate and produce results within error bounds, but they may select **different concrete values** at percentile boundaries. For example, for integers `1..1000`, the exact 25th percentile is `250.25` — GK returns `250` while KLL may return `251`. This difference cannot be eliminated by increasing precision.

### Changes Overview

```mermaid
graph TD
    subgraph "Root Cause"
        RC["Velox KLL sketch ≠ Spark GK algorithm<br/>Different approximate values at percentile boundaries"]
    end

    subgraph "VeloxTestSettings.scala — Excludes (4 suites)"
        E1["GlutenApproximatePercentileQuerySuite<br/><b>8 tests excluded</b>"]
        E2["GlutenDataFrameAggregateSuite<br/><b>1 test excluded</b>"]
        E3["GlutenDataFramePivotSuite<br/><b>1 test excluded</b>"]
        E4["GlutenDataFrameSuite<br/><b>1 test excluded</b>"]
    end

    subgraph "GlutenApproximatePercentileQuerySuite.scala — Overrides"
        O["<b>8 tests rewritten</b> with tolerance-based assertions<br/>(kllTolerance = 2)"]
    end

    RC --> E1 & E2 & E3 & E4
    E1 -- "excluded then re-added<br/>via testGluten" --> O
```

#### 2. `GlutenDataFrameAggregateSuite` — 1 test excluded
Velox KLL sketch produces different results from Spark's GK algorithm, especially with low accuracy (`accuracy=1`) |

#### 3. `GlutenDataFramePivotSuite` — 1 test excluded
different approximate values from Spark's GK algorithm in pivot context |

#### 4. `GlutenDataFrameSuite` — 1 test excluded
`DataFrame.summary()` uses `approx_percentile` internally; Velox KLL sketch produces different percentile values from Spark's GK algorithm |

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
@@ -56,7 +56,7 @@ const char* extractFileName(const char* file) {
 const std::unordered_set<std::string> kRegexFunctions =
     {"regexp_extract", "regexp_extract_all", "regexp_replace", "rlike", "split"};
 
-const std::unordered_set<std::string> kBlackList = {"split_part", "sequence", "approx_percentile", "map_from_arrays"};
+const std::unordered_set<std::string> kBlackList = {"split_part", "sequence", "map_from_arrays"};
 } // namespace
 
 bool SubstraitToVeloxPlanValidator::parseVeloxType(
@@ -1306,6 +1306,7 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::AggregateRel& ag
       "bloom_filter_agg",
       "var_samp",
       "var_pop",
+      "approx_percentile",
       "bit_and",
       "bit_or",
       "bit_xor",
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHSparkPlanExecApi.scala
@@ -41,7 +41,7 @@ import org.apache.spark.shuffle.utils.CHShuffleUtil
 import org.apache.spark.sql.catalyst.catalog.BucketSpec
 import org.apache.spark.sql.catalyst.catalog.CatalogTypes.TablePartitionSpec
 import org.apache.spark.sql.catalyst.expressions._
-import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, BloomFilterAggregate, CollectList, CollectSet}
+import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, ApproximatePercentile, BloomFilterAggregate, CollectList, CollectSet}
 import org.apache.spark.sql.catalyst.optimizer.BuildSide
 import org.apache.spark.sql.catalyst.plans.JoinType
 import org.apache.spark.sql.catalyst.plans.physical.{BroadcastMode, HashPartitioning, Partitioning, RangePartitioning}
@@ -596,6 +596,7 @@ class CHSparkPlanExecApi extends SparkPlanExecApi with Logging {
   /** Define backend specfic expression mappings. */
   override def extraExpressionMappings: Seq[Sig] = {
     List(
+      Sig[ApproximatePercentile](ExpressionNames.APPROX_PERCENTILE),
       Sig[CollectList](ExpressionNames.COLLECT_LIST),
       Sig[CollectSet](ExpressionNames.COLLECT_SET),
       Sig[MonotonicallyIncreasingID](MONOTONICALLY_INCREASING_ID),
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxRuleApi.scala
@@ -59,6 +59,7 @@ object VeloxRuleApi {
     injector.injectOptimizerRule(RewriteUnboundedWindow.apply)
     injector.injectOptimizerRule(PushAggregateThroughJoinBatch.apply)
     injector.injectPlannerStrategy(ImplementJoinAggregate.apply)
+    injector.injectOptimizerRule(ApproxPercentileRewriteRule.apply)
 
     if (!BackendsApiManager.getSettings.enableJoinKeysRewrite()) {
       injector.injectPlannerStrategy(_ => org.apache.gluten.extension.GlutenJoinKeysCapture())
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -21,7 +21,7 @@ import org.apache.gluten.config.{GlutenConfig, HashShuffleWriterType, ReservedKe
 import org.apache.gluten.exception.{GlutenExceptionUtil, GlutenNotSupportException}
 import org.apache.gluten.execution._
 import org.apache.gluten.expression._
-import org.apache.gluten.expression.aggregate.{HLLAdapter, VeloxBloomFilterAggregate, VeloxCollectList, VeloxCollectSet}
+import org.apache.gluten.expression.aggregate.{HLLAdapter, VeloxApproximatePercentile, VeloxBloomFilterAggregate, VeloxCollectList, VeloxCollectSet}
 import org.apache.gluten.extension.JoinKeysTag
 import org.apache.gluten.extension.columnar.FallbackTags
 import org.apache.gluten.shuffle.NeedCustomColumnarBatchSerializer
@@ -1128,6 +1128,7 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi with Logging {
       Sig[CollectList](ExpressionNames.COLLECT_LIST),
       Sig[VeloxCollectSet](ExpressionNames.COLLECT_SET),
       Sig[CollectSet](ExpressionNames.COLLECT_SET),
+      Sig[VeloxApproximatePercentile](ExpressionNames.APPROX_PERCENTILE),
       Sig[VeloxBloomFilterMightContain](ExpressionNames.MIGHT_CONTAIN),
       Sig[VeloxBloomFilterAggregate](ExpressionNames.BLOOM_FILTER_AGG),
       Sig[MapFilter](ExpressionNames.MAP_FILTER),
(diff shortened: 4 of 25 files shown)


jinchengchenghh:
Please update get-velox.sh to test your PR, then you can verify if both can work well, you may update this line https://github.com/apache/incubator-gluten/blob/5d3f7145cd7fc258aa10b434ea4ec651bd82c764/ep/build-velox/src/get-velox.sh#L28

jinchengchenghh:
Do we need the config? Usually we offload the function to native by default

Yizhou-Yang:
> Please update get-velox.sh to test your PR, then you can verify if both can work well, you may update this line
> 
> https://github.com/apache/incubator-gluten/blob/5d3f7145cd7fc258aa10b434ea4ec651bd82c764/ep/build-velox/src/get-velox.sh#L28

added the 16320 and removed the config

jinchengchenghh:
Please update the PR description to describe the KLL Sketch is different so that we handle fallback separately.

Yizhou-Yang:
> Please update the PR description to describe the KLL Sketch is different so that we handle fallback separately.

done~

Yizhou-Yang:
@jinchengchenghh
hi! The gluten side needs to be reviewed as well for the entire pr to work...

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -60,7 +60,7 @@ const std::unordered_set<std::string> kRegexFunctions = {
     "split"};
 
 const std::unordered_set<std::string> kBlackList =
-    {"split_part", "sequence", "approx_percentile", "map_from_arrays"};
+    {"split_part", "sequence", "map_from_arrays"};
```
jinchengchenghh: I find also the map_from_arrays in black list, another topic, would you like to implement it in Gluten?

jinchengchenghh:
Please resolve CH CI https://github.com/apache/incubator-gluten/blob/main/docs/get-started/ClickHouse.md#new-ci-system

jinchengchenghh (commented):
Most is good, after we resolve CH CI, we can merge it.

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
Yizhou-Yang: sure, I will look into it. @jinchengchenghh 
https://github.com/facebookincubator/velox/pull/13183
I think that the velox part is compatible with spark as of 2025/05, it's just for some reason the velox fix is not adopted in gluten.  Right now I think the only gluten fix needed is to remove the kblacklist and test it... 

I'll do a comparison with spark, post the results and add appropriate ut.

Yizhou-Yang:

ep/build-velox/src/get-velox.sh
```diff
@@ -25,7 +25,7 @@ RUN_SETUP_SCRIPT=ON
 ENABLE_ENHANCED_FEATURES=OFF
 
 # Developer use only for testing Velox PR.
-UPSTREAM_VELOX_PR_ID=""
+UPSTREAM_VELOX_PR_ID="16320"
```
jinchengchenghh: Update here

Yizhou-Yang:
There are test failures in x86 environment non-slow modes. 

edit:
The problem isn't that simple. it is related to gk vs kll. I might need to exclude/rewrite tests.
I will try to keep the original tests, but approximate algorithm does not give a definitive result sometimes.
I need more time on this

jiangjiangtian:
@Yizhou-Yang Thanks for your implementation! Recently I am cherrying pick your PR and I find that the KLL implementation in Gluten has only one level and discards items in odd position when compacting. I am wondering if this implementation can meet the `accuracy` requirement. Will the relative error rate be too high?

Yizhou-Yang:
> @Yizhou-Yang Thanks for your implementation! Recently I am cherrying pick your PR and I find that the KLL implementation in Gluten has only one level and discards items in odd position when compacting. I am wondering if this implementation can meet the `accuracy` requirement. Will the relative error rate be too high?

yes, this is part of the reason for the ut failures. The gluten version of kll is written rather hastily and was primarily concerned with simplicity and fallback compatibiltity with velox, and now it seems to have caused problems including but not limited to ut failures and lots of off-by-one issues. I am editing this implementation and doing a line-by-line comparison with the velox one to ensure the gluten side quality, and enhanced unit tests needs to be performed. Although after all this, there will still be some existing uts that might need to be excluded.

Yizhou-Yang:
> @Yizhou-Yang Thanks for your implementation! Recently I am cherrying pick your PR and I find that the KLL implementation in Gluten has only one level and discards items in odd position when compacting. I am wondering if this implementation can meet the `accuracy` requirement. Will the relative error rate be too high?

@jiangjiangtian Hi, the code has been fixed. Here's what changed:

**Changelog:**
1. Rewrote KLL sketch with proper multi-level compaction — the old implementation only had one level and discarded odd-position items, which is essentially random downsampling with no error bound guarantee. Now it uses the standard KLL algorithm: items across multiple levels, level-0 inserts, sort-and-halve compaction promoting to higher levels, with each level-*i* item representing 2^*i* original values.
2. Merge correctly combines multi-level sketches and re-compacts, instead of simple concatenation + truncation.

**Why some tests are disabled:**

KLL and Spark's native GK (Greenwald-Khanna) are fundamentally different algorithms. Both satisfy the `1/accuracy` error bound, but they produce **different concrete values** at the same percentile boundary. For example, given integers 1–1000, the exact 25th percentile is 250.25 — GK returns 250, KLL may return 251. Both are correct within the error bound, but a strict equality assertion like `assert(result == 250)` will fail. 

For those tests I added TestGluten in the approxpercentile suite. The off-by one problem can't be simply solved by adding more layers in partial result, I tried to not disable any test when developing, but the current version is the best I can do for now. 

The 4 disabled tests (`approx_percentile`, `summary`, `SPARK-35480`, `SPARK-32908`) all use exact-match assertions against GK's specific output. Rather than modifying upstream Spark tests, we excluded them and added a dedicated `GlutenApproximatePercentileQuerySuite` with tolerance-based assertions that validate correctness for both algorithms.

There are some other tests in collect_list that assumes approx_percentile will fallback, changed to percentile.

I tested only spark35/spark35smj and their slow versions, hopefully it also passes spark33 etc...

Yizhou-Yang:
> > @Yizhou-Yang Thanks for your implementation! Recently I am cherrying pick your PR and I find that the KLL implementation in Gluten has only one level and discards items in odd position when compacting. I am wondering if this implementation can meet the `accuracy` requirement. Will the relative error rate be too high?
> 
> 
> **Changelog:**
> 
> 1. Rewrote KLL sketch with proper multi-level compaction — the old implementation only had one level and discarded odd-position items, which is essentially random downsampling with no error bound guarantee. Now it uses the standard KLL algorithm: items across multiple levels, level-0 inserts, sort-and-halve compaction promoting to higher levels, with each level-_i_ item representing 2^_i_ original values.
> 2. Merge correctly combines multi-level sketches and re-compacts, instead of simple concatenation + truncation.
> 
> **Why some tests are disabled:**
> 
> KLL and Spark's native GK (Greenwald-Khanna) are fundamentally different algorithms. Both satisfy the `1/accuracy` error bound, but they produce **different concrete values** at the same percentile boundary. For example, given integers 1–1000, the exact 25th percentile is 250.25 — GK returns 250, KLL may return 251. Both are correct within the error bound, but a strict equality assertion like `assert(result == 250)` will fail.
> 
> For those tests I added TestGluten in the approxpercentile suite. The off-by one problem can't be simply solved by adding more layers in partial result, I tried to not disable any test when developing, but the current version is the best I can do for now.
> 
> The 4 disabled tests (`approx_percentile`, `summary`, `SPARK-35480`, `SPARK-32908`) all use exact-match assertions against GK's specific output. Rather than modifying upstream Spark tests, we excluded them and added a dedicated `GlutenApproximatePercentileQuerySuite` with tolerance-based assertions that validate correctness for both algorithms.
> 
> There are some other tests in collect_list that assumes approx_percentile will fallback, changed to percentile.
> 
> I tested only spark35/spark35smj and their slow versions, hopefully it also passes spark33 etc...

PTAL again... @jinchengchenghh  @zhztheplayer

jinchengchenghh:
Sorry for missing this PR, looks good to me, do you have further comments? @zhztheplayer

zhztheplayer (approved):
👍

jinchengchenghh:
Could you help resolve the conflict? Then we can merge it, thanks! @Yizhou-Yang

Yizhou-Yang:
> Could you help resolve the conflict? Then we can merge it, thanks! @Yizhou-Yang

ready @jinchengchenghh

Yizhou-Yang:
@jinchengchenghh 

sorry was not ready... didn't actually run local tests. 
I reran all the local tests so hopefully it works now.

Clipboard_Screenshot_1779107274

Yizhou-Yang:

Yizhou-Yang:
hopefully I fixed spark 40/41 related cases:
Clipboard_Screenshot_1779245317

PTAL again... @jinchengchenghh

Yizhou-Yang:

Yizhou-Yang:
reran spark 40/spark41 total 1124 cases and fixed issues in new ut. hopefully this is the last time... sorry about rerunning pipeline multiple times... @jinchengchenghh

Yizhou-Yang:
bugfix: accuracy can be 0, leading to kFromAccuracy inifite, and instantly oom. this is the reason for spark 41 failure.
No other test failure is noticed.

Yizhou-Yang:
retest this please... @jinchengchenghh

Issue #11605, opened 2026-02-11
[VL] Push Velox dynamic filters to shuffle reader


zhztheplayer:
As Velox's dynamic filter API is extensible, it's worth investigating whether we can push the filters down to Gluten's Velox shuffle reader to reduce the amount of data materialized from shuffle.

acvictor:
I think it is feasible by making Gluten's `ValueStream` operator in RowVectorStream.h override `canAddDynamicFilter()`  to return true, `addDynamicFilterLocked()` to receive filters, then apply them in `getOutput()` using `Filter::testX()` + Velox zero-copy `wrap()`. I think the changes can be contained to RowVectorStream.h. I can try out a PoC!

zhztheplayer:
@acvictor thanks a lot for helping out here.

> I think it is feasible by making Gluten's ValueStream operator in RowVectorStream.h

I think we don't have to expect much performance gain by filtering from ValueStream solely, since the filter operation is not guaranteed to be cheaper than hash probing. We might benefit from better cache locality as filters can be smaller than hash tables, but I am not sure how much. Worth a try anyway because it's a prerequisite for deeper pruning, e.g., from shuffle reader, since we need to obtain the pushed filters from Gluten side.

zhztheplayer:
https://github.com/apache/incubator-gluten/pull/11657

acvictor:
> [@acvictor](https://github.com/acvictor) thanks a lot for helping out here.
> 
> > I think it is feasible by making Gluten's ValueStream operator in RowVectorStream.h
> 
> I think we don't have to expect much performance gain by filtering from ValueStream solely, since the filter operation is not guaranteed to be cheaper than hash probing. We might benefit from better cache locality as filters can be smaller than hash tables, but I am not sure how much. Worth a try anyway because it's a prerequisite for deeper pruning, e.g., from shuffle reader, since we need to obtain the pushed filters from Gluten side.

I see what you mean. Should this filter push down be under a feature flag?

zhztheplayer:
@acvictor yes. If the benchmark doesn't show much benefit, we might consider using a flag to disable it by default.

zhztheplayer:
For the shuffle-reader pruning, I think we haven't had page-level metadata in shuffle blocks to support that yet. So maybe we need to start from writing metadata from the shuffle writer side - perhaps @marin-ma can help validate the idea later.

acvictor:
[...]

PR #11657, opened 2026-02-26
[GLUTEN-11605][VL] Push dynamic filters down to ValueStream

acvictor:
The PR implements dynamic filter pushdown in `ValueStreamDataSource` so that filters generated by `HashProbe` can propagate upstream to the shuffle reader, reducing the amount of data materialized from shuffle. The change is under a feature flag `spark.gluten.sql.columnar.backend.velox.valueStream.dynamicFilter.enabled` that is disabled by default. 

Dynamic filter pushdown is disabled for non-deterministic expressions safety (SPARK-10316) which is when the join's probe side contains non-deterministic expressions like rand(), as the filter would change which rows reach the scan, shifting the randomly generated sequence and producing incorrect results. 

## How was this patch tested?
Added the following UTs
1. noFilterPassesAllRows - verifies that all rows pass through when no dynamic filter is injected.
2. ` filterBigintRange` - reads first batch unfiltered, injects a `BigintRange(8, 10)` filter and verifies the second batch only returns rows matching `id >= 8`.
3. filterEliminatesEntireBatch - injects a filter `id >= 100` that eliminates the first batch's value range entirely, verifying the data source skips it and returns the matching second batch.
4. filterWithNulls - uses a nullable column with nulls and verifies that null rows are filtered out when `nullAllowed=false`, returning only non-null values that pass the range filter.
5. canAddDynamicFilter - verifies that `canAddDynamicFilter()` returns true on the `TableScan` operator backed by the `ValueStreamConnector`.

No

Related issue: #11605

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
@@ -78,9 +78,9 @@ class VeloxMetricsApi extends MetricsApi with Logging {
         "valueStreamDynamicFiltersAccepted" -> SQLMetrics.createMetric(
           sparkContext,
           "number of dynamic filters accepted by value stream"),
-        "valueStreamDynamicFilteredRows" -> SQLMetrics.createMetric(
+        "valueStreamDynamicFilterInputRows" -> SQLMetrics.createMetric(
           sparkContext,
-          "number of rows filtered by value stream dynamic filter")
+          "number of input rows")
       )
     } else {
       Map.empty[String, SQLMetric]
--- a/cpp/velox/compute/WholeStageResultIterator.cc
+++ b/cpp/velox/compute/WholeStageResultIterator.cc
@@ -43,7 +43,7 @@ namespace {
 const std::string kDynamicFiltersProduced = "dynamicFiltersProduced";
 const std::string kDynamicFiltersAccepted = "dynamicFiltersAccepted";
 const std::string kReplacedWithDynamicFilterRows = "replacedWithDynamicFilterRows";
-const std::string kDynamicFilteredRows = "dynamicFilteredRows";
+const std::string kDynamicFilterInputRows = "dynamicFilterInputRows";
 const std::string kFlushRowCount = "flushRowCount";
 const std::string kLoadedToValueHook = "loadedToValueHook";
 const std::string kBloomFilterBlocksByteSize = "bloomFilterSize";
@@ -493,8 +493,8 @@ void WholeStageResultIterator::collectMetrics() {
           runtimeMetric("sum", second->customStats, kDynamicFiltersAccepted);
       metrics_->get(Metrics::kNumReplacedWithDynamicFilterRows)[metricIndex] =
           runtimeMetric("sum", second->customStats, kReplacedWithDynamicFilterRows);
-      metrics_->get(Metrics::kNumDynamicFilteredRows)[metricIndex] =
-          runtimeMetric("sum", second->customStats, kDynamicFilteredRows);
+      metrics_->get(Metrics::kNumDynamicFilterInputRows)[metricIndex] =
+          runtimeMetric("sum", second->customStats, kDynamicFilterInputRows);
       metrics_->get(Metrics::kFlushRowCount)[metricIndex] = runtimeMetric("sum", second->customStats, kFlushRowCount);
       metrics_->get(Metrics::kLoadedToValueHook)[metricIndex] =
           runtimeMetric("sum", second->customStats, kLoadedToValueHook);
--- a/cpp/velox/operators/plannodes/RowVectorStream.cc
+++ b/cpp/velox/operators/plannodes/RowVectorStream.cc
@@ -198,7 +198,7 @@ facebook::velox::RowVectorPtr ValueStreamDataSource::applyDynamicFilters(
     }
     applyFilterOnColumn(filter, input->childAt(channel), rows);
     if (!rows.hasSelections()) {
-      dynamicFilteredRows_ += numRows;
+      dynamicFilterInputRows_ += numRows;
       return nullptr;
     }
   }
@@ -208,7 +208,7 @@ facebook::velox::RowVectorPtr ValueStreamDataSource::applyDynamicFilters(
     return input;
   }
 
-  dynamicFilteredRows_ += numRows - passedCount;
+  dynamicFilterInputRows_ += numRows;
 
   BufferPtr indices = allocateIndices(passedCount, pool_);
   auto* rawIndices = indices->asMutable<vector_size_t>();
--- a/backends-velox/src/main/java/org/apache/gluten/metrics/Metrics.java
+++ b/backends-velox/src/main/java/org/apache/gluten/metrics/Metrics.java
@@ -40,7 +40,7 @@ public class Metrics implements IMetrics {
   public long[] numDynamicFiltersProduced;
   public long[] numDynamicFiltersAccepted;
   public long[] numReplacedWithDynamicFilterRows;
-  public long[] numDynamicFilteredRows;
+  public long[] numDynamicFilterInputRows;
   public long[] flushRowCount;
   public long[] loadedToValueHook;
   public long[] bloomFilterBlocksByteSize;
@@ -91,7 +91,7 @@ public class Metrics implements IMetrics {
       long[] numDynamicFiltersProduced,
       long[] numDynamicFiltersAccepted,
       long[] numReplacedWithDynamicFilterRows,
-      long[] numDynamicFilteredRows,
+      long[] numDynamicFilterInputRows,
       long[] flushRowCount,
       long[] loadedToValueHook,
       long[] bloomFilterBlocksByteSize,
@@ -136,7 +136,7 @@ public class Metrics implements IMetrics {
     this.numDynamicFiltersProduced = numDynamicFiltersProduced;
     this.numDynamicFiltersAccepted = numDynamicFiltersAccepted;
     this.numReplacedWithDynamicFilterRows = numReplacedWithDynamicFilterRows;
-    this.numDynamicFilteredRows = numDynamicFilteredRows;
+    this.numDynamicFilterInputRows = numDynamicFilterInputRows;
     this.flushRowCount = flushRowCount;
     this.loadedToValueHook = loadedToValueHook;
     this.bloomFilterBlocksByteSize = bloomFilterBlocksByteSize;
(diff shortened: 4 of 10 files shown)


acvictor:
@zhztheplayer can you please review?

cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -652,6 +652,8 @@ std::unordered_map<std::string, std::string> WholeStageResultIterator::getQueryC
         std::to_string(veloxCfg_->get<bool>(kHashProbeDynamicFilterPushdownEnabled, true));
     configs[velox::core::QueryConfig::kHashProbeBloomFilterPushdownMaxSize] =
         std::to_string(veloxCfg_->get<uint64_t>(kHashProbeBloomFilterPushdownMaxSize, 0));
+    configs["value_stream_dynamic_filter_enabled"] =
```
zhztheplayer: `configs` was to store Velox configurations. Do we have other choices?

zhztheplayer:
I benchmarked the feature twice. Both showed clear improvements on TPC-DS Q94. Guessing pre-filter leveraged better cache locality against large hash tables. More verifications needed though.


1.

```
| q94 | 19.11 s | 8.76 s | -10.34 s | -54.13% | 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 |
```

2.

```
| q94 | 19.21 s | 9.82 s | -9.38 s | -48.85% | 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 |
```

cpp/velox/operators/plannodes/RowVectorStream.cc
```diff
@@ -166,7 +169,113 @@ std::optional<facebook::velox::RowVectorPtr> ValueStreamDataSource::next(
   completedRows_ += rowVector->size();
   completedBytes_ += rowVector->estimateFlatSize();
 
+  // Apply dynamic filters if any have been pushed down.
+  if (!dynamicFilters_.empty()) {
+    rowVector = applyDynamicFilters(rowVector);
```
zhztheplayer, line 174: Would you help check whether we can add some SQL metrics to track the filter operation from Spark UI?

acvictor:
> Would you enable the feature temporarily in the PR to see whether all unit tests can pass? Thanks.

There is one test failure
```
2026-02-28T15:42:58.9487030Z - SPARK-10316: respect non-deterministic expressions in PhysicalOperation *** FAILED ***
2026-02-28T15:42:58.9487854Z   0.665148093186119 did not equal 0.0 +- 0.001 (DataFrameSuite.scala:1832)
``` 

I think non-deterministic expressions should not be pushed

cpp/velox/operators/plannodes/RowVectorStream.cc
acvictor: Done - added tracking for number of dynamic filters accepted by value stream and number of rows filtered by value stream

cpp/velox/compute/WholeStageResultIterator.cc
acvictor: Added a flag similar for CUDF

acvictor:
> > Would you enable the feature temporarily in the PR to see whether all unit tests can pass? Thanks.
> 
> There is one test failure
> 
> ```
> 2026-02-28T15:42:58.9487030Z - SPARK-10316: respect non-deterministic expressions in PhysicalOperation *** FAILED ***
> 2026-02-28T15:42:58.9487854Z   0.665148093186119 did not equal 0.0 +- 0.001 (DataFrameSuite.scala:1832)
> ```
> 
> I think non-deterministic expressions should not be pushed

Fixed this by ensuring that non-deterministic expressions are not pushed down.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/IteratorApi.scala
```diff
@@ -85,6 +85,7 @@ trait IteratorApi {
       updateNativeMetrics: IMetrics => Unit,
       partitionIndex: Int,
       materializeInput: Boolean = false,
-      enableCudf: Boolean = false): Iterator[ColumnarBatch]
+      enableCudf: Boolean = false,
+      disableValueStreamDynamicFilter: Boolean = false): Iterator[ColumnarBatch]
```
zhztheplayer: Would it be a bit clearer to reword the parameter as `supports...`?

gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
```diff
@@ -257,7 +258,34 @@ case class WholeStageTransformer(child: SparkPlan, materializeInput: Boolean = f
...
+      hasNonDeterministicExprInJoinProbe(child))
+  }
+
+  /**
+   * Checks whether any HashJoin's probe (streamed) side contains non-deterministic expressions.
+   * When true, ValueStream dynamic filter pushdown must be disabled because the dynamic filter
+   * would be applied at the scan level (below the non-deterministic Project), changing how many
+   * times the non-deterministic expression is evaluated and thus altering its output sequence. See
+   * SPARK-10316.
+   */
```
zhztheplayer, line 274: > ValueStream dynamic filter pushdown must be disabled because the dynamic filter would be applied at the scan level

If a ValueStream dynamic filter is disabled, we won't have a scan to accept the dynamic filter - so am I missing something here?

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -468,6 +471,14 @@ object VeloxConfig extends ConfigRegistry {
       .booleanConf
       .createWithDefault(true)
 
+  val VALUE_STREAM_DYNAMIC_FILTER_ENABLED =
+    buildConf("spark.gluten.sql.columnar.backend.velox.valueStream.dynamicFilter.enabled")
+      .doc(
+        "Whether to apply dynamic filters pushed down from hash probe in the ValueStream" +
+          " (shuffle reader) operator to filter rows before they reach the hash join.")
+      .booleanConf
+      .createWithDefault(true)
```
zhztheplayer: Should it be false?

cpp/velox/config/VeloxConfig.h
```diff
@@ -78,6 +78,10 @@ const std::string kHashProbeDynamicFilterPushdownEnabled =
 const std::string kHashProbeBloomFilterPushdownMaxSize =
     "spark.gluten.sql.columnar.backend.velox.hashProbe.bloomFilterPushdown.maxSize";
 
+const std::string kValueStreamDynamicFilterEnabled =
+    "spark.gluten.sql.columnar.backend.velox.valueStream.dynamicFilter.enabled";
+const bool kValueStreamDynamicFilterEnabledDefault = true;
```
zhztheplayer: Should it be false?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
```diff
@@ -628,6 +628,12 @@ class VeloxMetricsApi extends MetricsApi with Logging {
       "hashProbeDynamicFiltersProduced" -> SQLMetrics.createMetric(
         sparkContext,
         "number of hash probe dynamic filters produced"),
+      "valueStreamDynamicFiltersAccepted" -> SQLMetrics.createMetric(
+        sparkContext,
+        "number of dynamic filters accepted by value stream"),
+      "valueStreamDynamicFilteredRows" -> SQLMetrics.createMetric(
+        sparkContext,
+        "number of rows filtered by value stream dynamic filter"),
```
zhztheplayer: Thanks for adding the metrics. Can you share a screenshot of Spark UI?

cpp/velox/config/VeloxConfig.h
acvictor: Will update

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
acvictor: Will update

gluten-substrait/src/main/scala/org/apache/gluten/execution/WholeStageTransformer.scala
acvictor: I have changed the comment "the dynamic filter would be applied at the scan level" to "if left enabled, the dynamic filter would filter rows at the ValueStream" to make it clear that the comment describes what would go wrong without the disable, not the disabled state itself.

gluten-substrait/src/main/scala/org/apache/gluten/backendsapi/IteratorApi.scala
acvictor: Updated to `supportsValueStreamDynamicFilter`

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
acvictor:
```
 # Create a probe table (large) and a build table (selective) so the filter has rows to prune
 spark.sql("CREATE TABLE vs_probe_table USING PARQUET AS SELECT id AS a FROM range(1000000)")
 spark.sql("CREATE TABLE vs_build_table USING PARQUET AS SELECT id * 1000 AS b FROM range(1000)")
 
 # Push a dynamic filter
 spark.sql("SELECT a FROM vs_probe_table JOIN vs_build_table ON a = b").collect()
```

image

zhztheplayer: Thanks. Would it be possible to move metric `number of dynamic filters accepted` /  `number of rows filtered by value stream dynamic filter:` to `InputIteratorTransformer`? Which will align with scan filter pushdown.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
acvictor: Done.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
acvictor: image

Done

cpp/velox/config/VeloxConfig.h
acvictor: Done

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxMetricsApi.scala
zhztheplayer: Thanks. Is it possible to add `number of input rows` instead of `number of rows filtered by value stream dynamic filter`? To make it more consistent with scan.

acvictor: Done!
image

acvictor: @zhztheplayer does this look good to you?

PR #11660, opened 2026-02-26
[Minor][VL]feat: Relax the restrict since velox-cudf supports reading part parquet file

jinchengchenghh:
After this PR https://github.com/facebookincubator/velox/pull/14968, velox has supported read start and length of a parquet file


wForget (approved):
Thanks @jinchengchenghh

PR #11661, opened 2026-02-26
[MINOR][VL] Add iceberg test for spark40 enhanced features

jinchengchenghh:


.github/workflows/velox_backend_enhanced.yml
```diff
@@ -255,3 +255,60 @@ jobs:
...
+    env:
+      SPARK_TESTING: true
+    container: apache/gluten:centos-8-jdk17
+    steps:
+      - uses: actions/checkout@v4
+      - name: Download All Artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: velox-native-lib-enhanced-centos-7-${{github.sha}}
+          path: ./cpp/build/
```
wForget:
```suggestion
          path: ./cpp/build/releases
```

jinchengchenghh: Other CI also uses this path, we upload to this path before, so we should download from this path.
I also agree this library should be in cpp/build/releases to align with local path, maybe in a separate PR for all the CI.
```
      - uses: actions/upload-artifact@v4
        with:
          name: velox-native-lib-centos-7-${{github.sha}}
          path: ./cpp/build/
          if-no-files-found: error
```

jinchengchenghh: Thanks! I missed that, maybe we could also update other CI in velox_backend_x86.xml

PR #11666, opened 2026-02-27
[BUILD] Upgrade Scala 2.12 version to 2.12.18 to support JDK 21

liujiayi771:
Gluten should upgrade the Scala 2.12 version to 2.12.18 to support JDK 21 and align with the Scala version used in Spark 3.5.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <caffeine.version>2.9.3</caffeine.version>
     <scala.binary.version>2.12</scala.binary.version>
-    <scala.version>2.12.15</scala.version>
+    <scala.version>2.12.18</scala.version>
     <scala-xml.version>2.1.0</scala-xml.version>
     <javax.servlet.version>3.1.0</javax.servlet.version>
     <spark.major.version>3</spark.major.version>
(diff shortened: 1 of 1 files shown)


liujiayi771:
image

liujiayi771:
@PHILO-HE @baibaichen Could you help to review?

philo-he:
@liujiayi771, thanks for your PR. Since the Scala compatibility is essentially related to JDK version, can we set Scala version in the JDK profile configuration?

https://github.com/apache/incubator-gluten/blob/f22e21fe414efe2302a1733782a214f6ae1949b4/pom.xml#L1128-L1139

liujiayi771:
@PHILO-HE The incompatibility between Scala and JDK is just one example of this issue. What I actually want to resolve is to align Gluten’s Scala 2.12 version with Spark.

Assuming the Scala 2.12.x version that Spark 3.x depends on is incompatible with a certain JDK version, it would be meaningless for Gluten to use a compatible Scala 2.12 build. Since Gluten always relies on Spark, we need to stay consistent with Spark.

Beyond JDK 21, there may be other compatibility issues. It is always the right choice for us to align with Spark.

philo-he:
@liujiayi771, thank you for the clarification. According to the table you provided, it seems higher Scala version is compatible with lower Java versions. So I am wondering if we can directly upgrade Scala 2.12 to 2.12.18 for all Spark versions, regardless of the Java versions used for building Spark.

liujiayi771:
@PHILO-HE Okay, this minimizes changes and resolves the JDK 21 compatibility issue.

PR #11676, opened 2026-03-02
[GLUTEN-11677] Support get cpp stack using gdb from Spark UI

WangGuangxin:

In Vanilla Spark WebUI, it is easy to get the current java jstack, which is very helpful to diagnostic the hang/slow problems. But when it comes to Spark on Gluten, the jstack is meaningless since all heavy jobs are under JNI call.

This PR adds the same ability to capture native C++ stack traces from executors directly from the Gluten SQL tab in the Spark UI. It introduces an executors overview with a per‑executor "C++ Stack" action and a dedicated "Gluten C++ Stack" status page that streams the collected stack output.

image

image

There are two modes, sync and async. The async mode is mainly because gdb data collection typically takes a long time, exceeding the yarn proxy timeout in our environment.

## How was this patch tested?

Manually

Generated-by: TRAE with Doubao-Seed-Code

Related issue: #11677

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/spark/rpc/GlutenExecutorEndpoint.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/spark/rpc/GlutenExecutorEndpoint.scala
@@ -21,7 +21,8 @@ import org.apache.gluten.execution.{CHBroadcastBuildSideCache, CHNativeCacheMana
 import org.apache.spark.{SparkConf, SparkEnv}
 import org.apache.spark.internal.{config, Logging}
 import org.apache.spark.rpc.GlutenRpcMessages._
-import org.apache.spark.util.ThreadUtils
+import org.apache.spark.rpc.RpcCallContext
+import org.apache.spark.util.{ExecutorManager, ThreadUtils, Utils}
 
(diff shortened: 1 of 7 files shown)


backends-clickhouse/src/main/scala/org/apache/spark/rpc/GlutenExecutorEndpoint.scala
```diff
@@ -100,6 +230,81 @@ class GlutenExecutorEndpoint(val executorId: String, val conf: SparkConf)
...
+    try {
+      val out = Utils.executeAndGetOutput(
+        Seq("bash", "-c", s"command -v $cmd >/dev/null 2>&1 && echo yes || echo no"))
+      out != null && out.trim == "yes"
+    } catch { case _: Throwable => false }
+  }
+
+  private def ensureGdbInstalled(): Unit = {
+    if (!has("gdb")) {
+      val sudo = if (has("sudo")) "sudo " else ""
```
zhouyuan, line 275: In some env it may not allow to install other software

zhouyuan (commented):
a great feature for debug executor hang! It's better to have a config to switch on/off for this feature?

Issue #11658, opened 2026-02-26
Restore scala.recompile.mode default to 'all' and introduce fast-build profile


baibaichen:
## Problem

PR #11560 changed the default `scala.recompile.mode` from `all` to `incremental`.
This causes build failures in clean/CI environments because Zinc's incremental
compiler does not reliably track transitive dependencies across Spark's module
structure, producing errors like:

> Symbol 'type org.apache.spark.sql.internal.SqlApiConf' is missing from the classpath.

## Solution

1. Restore `scala.recompile.mode` default to `all` (safe for CI).
2. Rename the `bloop` profile (added in #11645) to `fast-build` and add
   `scala.recompile.mode=incremental` inside it, so bloop setup and
   `run-scala-test.sh` (which use `-Pfast-build`) benefit from faster incremental
   compilation in contexts where Zinc's analysis is warm.
3. Simplify `run-scala-test.sh`: replace 5 redundant `-D` skip flags with `-Pfast-build`.
4. Also includes pathing JAR support in `run-scala-test.sh` to avoid command-line length limits.

PR #11681, opened 2026-03-02
[GLUTEN-11658][VL] Add missing direct dependency for each module

zhouyuan:

In my local env I find there are some direct dependencies are missing. This patch added these dependencies into pom.xml

## How was this patch tested?
pass GHA

Related issue: #11658

### DIFF
--- a/shims/common/pom.xml
+++ b/shims/common/pom.xml
@@ -46,12 +46,7 @@
       <groupId>org.apache.parquet</groupId>
       <artifactId>parquet-hadoop</artifactId>
       <version>1.15.2</version>
-      <exclusions>
-        <exclusion>
-          <groupId>io.airlift</groupId>
-          <artifactId>aircompressor</artifactId>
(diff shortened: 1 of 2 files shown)


shims/common/pom.xml
```diff
@@ -37,6 +37,22 @@
...
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-catalyst_${scala.binary.version}</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-hadoop</artifactId>
```
philo-he, line 47: Maybe, we can set it in "provided" scope to avoid version conflict, since a different version is available under Spark jars/ for runtime use.

zhouyuan: done

Issue #11379, opened 2026-01-08
Remove remaining Spark 3.2-specific compatibility code


QCLyu:
### Description

Some Spark 3.2-specific compatibility code remains in the common codebase:
#gluten-core/src/main/scala/org/apache/spark/util/SparkVersionUtil.scala
Defines lteSpark32 (used by other files)
#gluten-core/src/main/scala/org/apache/spark/util/SparkTaskUtil.scala
Lines 67-82: Spark 3.2-specific constructor path for TaskContext
#gluten-core/src/main/scala/org/apache/spark/util/SparkPlanUtil.scala
Lines 26-28: Spark 3.2-specific path for supportsRowBased
#gluten-core/src/main/scala/org/apache/spark/sql/execution/adaptive/GlutenCostEvaluator.scala
Lines 33-35: Spark 3.2-specific CostEvaluator instantiation
#gluten-core/src/main/scala/org/apache/gluten/extension/columnar/transition/Convention.scala
Lines 174-184: Spark 3.2-specific row type handling
#backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
Lines 911-919: Test that checks for Spark 3.2

These items are compatibility shims in shared code that can be cleaned up separately.
Related Issue: #8960 

### Gluten version

None

QCLyu:
take

zhztheplayer:
Thanks for working on the 3.2 deprecation. Will https://github.com/apache/incubator-gluten/pull/11495 also disable 3.2 CI?

QCLyu:
@zhztheplayer In the quoted issue,  there is no Spark 3.2 CI in the current setup.
In .github/workflows/velox_backend_x86.yml the Spark matrix is [ "spark-3.3", "spark-3.4", "spark-3.5", "spark-4.0", "spark-4.1" ]. So there is no spark-3.2 job to disable. The pom.xml has Spark profiles for 3.3, 3.4, 3.5, 4.0, and 4.1, and no spark-3.2 profile. cc @PHILO-HE in case I got confused.

zhztheplayer:
@QCLyu I found https://github.com/apache/incubator-gluten/issues/11351 which I didn't notice earlier. I have no issues with further cleanups and please go ahead.

QCLyu:
Update: I was able to fix local build errors. Resuming the work to deprecate the rest of Spark 3.2 code.
FYI, after running into multiple rabbit holes, the solution to my local build was simple:
[...]

PR #11687, opened 2026-03-03
[GLUTEN-11379][CORE] Clean up Spark shims APIs following Spark 3.2 deprecation

philo-he:

Since Spark 3.2 has been dropped, we need to clean up those shims APIs which were introduced to fix Spark code differences between Spark 3.2 and later versions. Then, the implementation for those APIs can be moved to the caller side.

```
getDistribution
convertPartitionTransforms
getTextScan
bloomFilterExpressionMappings
newBloomFilterAggregate
newMightContain
replaceBloomFilterAggregate
replaceMightContain
getShuffleReaderParam
getPartitionId
supportDuplicateReadingTracking
getFileSizeAndModificationTime
dateTimestampFormatInReadIsDefaultValue
genDecimalRoundExpressionOutput 
```

## How was this patch tested?
Local build.

No.

Related issue: #11379

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/ArrowConvertorRule.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/ArrowConvertorRule.scala
@@ -19,14 +19,13 @@ package org.apache.gluten.extension
 import org.apache.gluten.backendsapi.BackendsApiManager
 import org.apache.gluten.datasource.ArrowCSVFileFormat
 import org.apache.gluten.datasource.v2.ArrowCSVTable
-import org.apache.gluten.sql.shims.SparkShimLoader
 
 import org.apache.spark.annotation.Experimental
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.csv.CSVOptions
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.rules.Rule
-import org.apache.spark.sql.catalyst.util.PermissiveMode
+import org.apache.spark.sql.catalyst.util.{CaseInsensitiveMap, PermissiveMode}
 import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation}
 import org.apache.spark.sql.execution.datasources.csv.CSVFileFormat
 import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation
@@ -102,6 +101,7 @@ case class ArrowConvertorRule(session: SparkSession) extends Rule[LogicalPlan] {
   }
 
   private def checkCsvOptions(csvOptions: CSVOptions, timeZone: String): Boolean = {
+    val default = new CSVOptions(CaseInsensitiveMap(Map()), csvOptions.columnPruning, timeZone)
     csvOptions.headerFlag && !csvOptions.multiLine &&
     csvOptions.delimiter.length == 1 &&
     csvOptions.quote == '\"' &&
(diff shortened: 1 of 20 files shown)


philo-he:
@QCLyu, this is another PR to clean up some code after Spark 3.2 deprecation. Could you take a look? cc @zhouyuan

backends-velox/src/main/scala/org/apache/gluten/extension/ArrowConvertorRule.scala
```diff
@@ -112,7 +112,9 @@ case class ArrowConvertorRule(session: SparkSession) extends Rule[LogicalPlan] {
     csvOptions.nullValue == "" &&
     csvOptions.emptyValueInRead == "" && csvOptions.comment == '\u0000' &&
     csvOptions.columnPruning &&
-    SparkShimLoader.getSparkShims.dateTimestampFormatInReadIsDefaultValue(csvOptions, timeZone)
+    csvOptions.dateFormatInRead == default.dateFormatInRead &&
+    csvOptions.timestampFormatInRead == default.timestampFormatInRead &&
+    csvOptions.timestampNTZFormatInRead == default.timestampNTZFormatInRead
```
QCLyu, line 117: We need to confirm timestampNTZFormatInRead exists in Spark 3.3's CSVOptions

shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
```diff
@@ -278,32 +212,11 @@ trait SparkShims {
...
   def withAnsiEvalMode(expr: Expression): Boolean = false
 
-  def dateTimestampFormatInReadIsDefaultValue(csvOptions: CSVOptions, timeZone: String): Boolean
-
   def createParquetFilters(
       conf: SQLConf,
       schema: MessageType,
       caseSensitive: Option[Boolean] = None): ParquetFilters
 
-  def genDecimalRoundExpressionOutput(decimalType: DecimalType, toScale: Int): DecimalType = {
```
QCLyu, line 288: We need to ensure the base trait's default implementation handles all Spark versions correctly, or that the method is also removed from the SparkPlanExecApi trait

gluten-substrait/src/main/scala/org/apache/spark/shuffle/GlutenShuffleUtils.scala
```diff
@@ -18,10 +18,9 @@ package org.apache.spark.shuffle
 
 import org.apache.gluten.backendsapi.BackendsApiManager
 import org.apache.gluten.config.GlutenConfig
-import org.apache.gluten.sql.shims.SparkShimLoader
 import org.apache.gluten.vectorized.NativePartitioning
 
-import org.apache.spark.{SparkConf, TaskContext}
+import org.apache.spark.{ShuffleUtils, SparkConf, TaskContext}
```
QCLyu, line 23: This ShuffleUtils class should be in the shims (it exists in shims/spark34/ etc.). We need to confirm it's available for Spark 3.3 as well

shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
philo-he: The method here is identical with the one in SparkPlanExecApi, so I think this one is not required to be called for overriding the one in SparkPlanExecApi. Let's just remove this one. Thanks.

gluten-substrait/src/main/scala/org/apache/spark/shuffle/GlutenShuffleUtils.scala
philo-he: Yes, it also exists in shims/spark33.

backends-velox/src/main/scala/org/apache/gluten/extension/ArrowConvertorRule.scala
philo-he: Just confirmed in source code. It exists. And the compilation also help ensures this. Thanks.

philo-he:
@QCLyu, thanks for the review. @zhouyuan, do you have any comment?

zhouyuan (approved):
👍  Thanks. The shim layer looks cleaner now.

shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
```diff
@@ -85,34 +76,19 @@ object SparkShimDescriptor {
...
 
   def scalarExpressionMappings: Seq[Sig]
 
   def aggregateExpressionMappings: Seq[Sig]
 
   def runtimeReplaceableExpressionMappings: Seq[Sig]
 
-  def convertPartitionTransforms(partitions: Seq[Transform]): (Seq[String], Option[BucketSpec])
-
   def generateFileScanRDD(
```
zhouyuan, line 86: We may better add a note on when/why these shim APIs are introduced for future changes

PR #11697, opened 2026-03-03
[GLUTEN-9456][VL] Cancel load earlier in `GlutenDirectBufferedInput` destructor

rui-mo:

While waiting for a load, a planned load may start unnecessarily. This PR cancels the load as early as possible to minimize wait time.

## How was this patch tested?

Verified on CI.

No

Related issue: #9456

### DIFF
--- a/cpp/velox/memory/GlutenDirectBufferedInput.h
+++ b/cpp/velox/memory/GlutenDirectBufferedInput.h
@@ -48,6 +48,14 @@ class GlutenDirectBufferedInput : public facebook::velox::dwio::common::DirectBu
 
   ~GlutenDirectBufferedInput() override {
     requests_.clear();
+    // Cancel all the planned loads as soon as possible to avoid unnecessary IO.
+    for (auto& load : coalescedLoads_) {
+      if (load->state() == facebook::velox::cache::CoalescedLoad::State::kPlanned) {
+        load->cancel();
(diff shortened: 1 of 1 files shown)


cpp/velox/memory/GlutenDirectBufferedInput.h
```diff
@@ -48,6 +48,11 @@ class GlutenDirectBufferedInput : public facebook::velox::dwio::common::DirectBu
 
   ~GlutenDirectBufferedInput() override {
     requests_.clear();
+    for (auto& load : coalescedLoads_) {
```
philo-he, line 52: Can we unify the two loops into one?

rui-mo: I intentionally moved this forward to cancel the load as early as possible, trying to reduce wait time. Thanks.

rui-mo:
@PHILO-HE Thanks. I added some comments to clarify.

cpp/velox/memory/GlutenDirectBufferedInput.h
FelixYBW: The goal is that we should cancel the unused IO as soon as possible. 

In old logic, a I/O thread may still be scheduled when we are waiting for current I/O.

PR #11704, opened 2026-03-05
[INFRA] Update release scripts for 4.0

zhztheplayer:
We need to change the release script to properly include Spark 4.0 support.

### DIFF
--- a/dev/release/build-release.sh
+++ b/dev/release/build-release.sh
@@ -67,5 +67,5 @@ done
 
 for spark_version in 4.0
 do
-  ${GLUTEN_HOME}/build/mvn clean install -Pjava-17 -Pscala-2.13 -Pbackends-velox -Pspark-${spark_version} -DskipTests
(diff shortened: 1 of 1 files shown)


dev/release/build-release.sh
```diff
@@ -64,3 +64,8 @@ do
   ${GLUTEN_HOME}/build/mvn clean install -Pjava-17 -Pbackends-velox -Pspark-${spark_version} -Pceleborn,uniffle \
                     -Piceberg,delta,hudi,paimon -DskipTests
 done
+
+for spark_version in 4.0
+do
+  ${GLUTEN_HOME}/build/mvn clean install -Pjava-17 -Pscala-2.13 -Pbackends-velox -Pspark-${spark_version} -DskipTests
```
zhouyuan: Could we also enable the connectors `-P icberg,hudi,delta,paimon` here?

zhztheplayer: As discussed, I added `-P iceberg` here, will add others once they are guaranteed to be ready for release.

PR #11705, opened 2026-03-05
[VL] Support ANSI mode decimal Add/Subtract with checked overflow

n0r0shi:
## Summary
  - Routes decimal `Add` and `Subtract` to Velox's `checked_add` and `checked_subtract` when ANSI mode is enabled (`nullOnOverflow = false` in `CheckOverflow`), so overflow throws instead of returning null.
  - Adds TRY mode support for decimal `Add`/`Subtract` using the `try(checked_add/checked_subtract)` pattern, consistent with how integer TRY arithmetic is already handled.
  - Previously, TRY mode for decimal arithmetic relied on a `tryCast` in `CheckOverflowTransformer` to detect overflow. This change uses the consistent `try(checked_*)` pattern instead.

  ## Dependencies
  - Depends on facebookincubator/velox#16302 (merged to upstream), which adds `checked_add` and `checked_subtract` support for decimal types.
  - The commit is on IBM/velox `oss-main` but not yet on the `dft-` branch Gluten pins to. The ANSI overflow tests will pass after the next Velox version bump in `get-velox.sh`.

  ## Test plan
  - [x] Normal decimal add/subtract succeeds and matches Spark results
  - [x] Overflow on decimal add/subtract throws in ANSI mode
  - [x] `try_add`/`try_subtract` with decimals returns null on overflow
  - [ ] Full validation after Velox version bump picks up facebookincubator/velox#16302

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxLiteralSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/VeloxLiteralSuite.scala
@@ -139,4 +139,5 @@ class VeloxLiteralSuite extends VeloxWholeStageTransformerSuite {
     validateFallbackResult("SELECT struct(cast(null as struct<a: string>))")
     validateFallbackResult("SELECT array(struct(1, 'a'), null)")
   }
+
 }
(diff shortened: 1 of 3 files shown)


n0r0shi:
Superseded by #12192 after rebase on current main.

Issue #11708, opened 2026-03-05
[VL] Translate bloom filter might_contain to a subfield filter


zhztheplayer:
We need to translate `might_contain` to a subfield filter in `SparkExprToSubfieldFilterParser` for probably more efficient filtering.

acvictor:
@zhztheplayer I have tried this here #11711

PR #11711, opened 2026-03-06
[GLUTEN-11708][VL] Translate might_contain as a subfield filter for scan-level bloom filter pushdown

acvictor:
This PR adds support for pushing `might_contain(bloomFilter, value`) down into Velox's subfield filter system via` SparkExprToSubfieldFilterParser`. Previously, `might_contain` was evaluated as a post-scan expression. With this change, the bloom filter check can be applied at the storage scan level allowing entire row groups to be skipped before data is fully decoded.

Velox has two incompatible bloom filter implementations:
   - BloomFilter: used by `bloom_filter_agg` / `might_contain` (groups-of-64-bits, 4 hash functions)
   - SplitBlockBloomFilter: used by the existing `BigintValuesUsingBloomFilter` filter class (SIMD split-block)

Since these are not interchangeable, a new `SparkBloomFilter` filter class is introduced that wraps the serialized `BloomFilter<>` data and implements `testInt64()` using `BloomFilter<>::mayContain()` with `folly::hasher<int64_t>(`) which is the same code path used by the JNI `mightContainLongOnSerializedBloom`.

## How was this patch tested?
Added new test suite covering basic filtering, null bloom filter, negation, non-column value, range test, and clone behavior.

No

Related issue https://github.com/apache/incubator-gluten/issues/11708


acvictor:
@zhztheplayer can you please review this PR?

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
```diff
@@ -16,11 +16,64 @@
...
+    return nullptr;
+  }
+  RowVector input(evaluator->pool(), ROW({}, {}), nullptr, 1, std::vector<VectorPtr>{});
+  SelectivityVector rows(1);
+  VectorPtr result;
+  try {
+    evaluator->evaluate(exprSet.get(), rows, input, result);
+  } catch (const VeloxUserError&) {
+    return nullptr;
+  }
```
zhztheplayer, line 44: Why error is swallowed?

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
```diff
@@ -16,11 +16,64 @@
...
+  return result;
+}
+
+/// Filter backed by Velox's BloomFilter<> serialized data from bloom_filter_agg.
+class SparkBloomFilter final : public common::Filter {
```
zhztheplayer: Let's use `SparkMightContain` or so to align with Spark's function name.

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
```diff
@@ -16,11 +16,64 @@
...
+ public:
+  SparkBloomFilter(std::vector<char> serializedData, bool nullAllowed)
+      : Filter(true, nullAllowed, common::FilterKind::kBigintValuesUsingBloomFilter),
+        serializedData_(std::move(serializedData)) {}
+
+  bool testInt64(int64_t value) const final {
+    return BloomFilter<>::mayContain(serializedData_.data(), folly::hasher<int64_t>()(value));
```
zhztheplayer: We can create a bloom filter object as class member, since `bloomFilter.mayContain` is faster than `BloomFilter<>::mayContain`.

zhztheplayer (commented):
Thanks.

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
acvictor: Errors are intentionally swallowed because a non-evaluable expression simply means the filter cannot be pushed down. So if the filter cannot be pushed down, `leafCallToSubfieldFilter` returns `std::nullopt`, and Velox will evaluate `might_contain` as a regular post-scan expression which is the same behavior as before. Does this flow make sense? I have updated the comment as well.

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
acvictor: Done.

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
acvictor: Done.

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
zhztheplayer: In which case it cannot be pushed? Can you paste the error here?

zhztheplayer (commented):
Thanks. Code LGTM, will take some time to test

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
acvictor: I don't actually have an example, but I have used the same pattern as `toConstant()` helper in Velox `ExprToSubfieldFilter.cpp` which is used by `makeEqualFilter`, `makeLessThanFilter`, and other filter factories https://github.com/facebookincubator/velox/blob/main/velox/expression/ExprToSubfieldFilter.cpp#L27

cpp/velox/operators/functions/SparkExprToSubfieldFilterParser.cc
```diff
@@ -16,11 +16,69 @@
...
+    return nullptr;
+  }
+  return result;
+}
+
+/// Subfield filter backed by Velox's BloomFilter from bloom_filter_agg / might_contain.
+class SparkMightContain final : public common::Filter {
+ public:
+  SparkMightContain(const char* serializedData, bool nullAllowed)
+      : Filter(true, nullAllowed, common::FilterKind::kBigintValuesUsingBloomFilter) {
```
weiting-chen, line 52: SparkMightContain is reusing an enum value for a class(FilterKind) that doesn't inherit from the original(BigintValuesUsingBloomFilter) may cause undefined behavior (invalid static_cast → crash).
Check if this is a case require to be considering.

acvictor: I think you're right - it might be better to add a new FilterKind to Velox so no code ever tries to static_cast a SparkMightContain to BigintValuesUsingBloomFilter?

weiting-chen: Yes, by creating a new FilterKind in Velox can help to solve this issue.

acvictor: @weiting-chen this was reviewed here https://github.com/facebookincubator/velox/pull/16940 and the feedback was to use one of `kBigintValuesUsingHashTable`, `kBigintValuesUsingBitmask`, `kBigintValuesUsingBloomFilter` wherever needed

PR #11718, opened 2026-03-08
[CORE] Refactor task error logging into dedicated utility class

pratham76:

Extract task error logging from TaskResources into new `TaskErrorLogger` utility, resolving a TODO [here](https://github.com/apache/incubator-gluten/blob/096545f03c4d8aa550902b13d2775a7ae2816599/gluten-core/src/main/scala/org/apache/spark/task/TaskResources.scala#L212). Improves separation of concerns while maintaining identical functionality for filtering speculative execution errors.

## How was this patch tested?

Existing UTs

No

### DIFF
--- a/gluten-core/src/main/scala/org/apache/spark/task/TaskResources.scala
+++ b/gluten-core/src/main/scala/org/apache/spark/task/TaskResources.scala
@@ -204,6 +204,8 @@ object TaskResources extends TaskListener with Logging {
       }
       val registry = new TaskResourceRegistry
       RESOURCE_REGISTRIES.put(tc, registry)
+      // TODO: Propose upstream Spark changes for resilient error logging when
+      // CompletionListener crashes. Using TaskErrorLogger as workaround
       tc.addTaskFailureListener(
         // in case of crashing in task completion listener, errors may be swallowed
         new TaskFailureListener {


pratham76:
@zhztheplayer Could you please review?

pratham76:
Hi @zhztheplayer, Thanks for the clarification, I do get that the proper solution for the above mentioned TODO would be make spark's error logging more resilient when CompletionListener crashes.

In this PR, I have tried to refactor the logging logic into separate utility class, to handle both scenarios of task failures and task recompute/retries, just to make the code a bit cleaner. I'm happy to explore the upstream Spark changes for this TODO (which I have updated here to give better context). For now, could we proceed with this refactoring as an incremental improvement?

zhztheplayer:
> Hi @zhztheplayer, Thanks for the clarification, I do get that the proper solution for the above mentioned TODO would be make spark's error logging more resilient when CompletionListener crashes.
> 
> In this PR, I have tried to refactor the logging logic into separate utility class, to handle both scenarios of task failures and task recompute/retries, just to make the code a bit cleaner. I'm happy to explore the upstream Spark changes for this TODO (which I have updated here to give better context). For now, could we proceed with this refactoring as an incremental improvement?

Sounds reasonable. Thanks for the patch.

Issue #10618, opened 2025-09-03
[VL] Add details to the "time of input iterator" metric


marin-ma:
### Description

For the "time of input iterator" metric in the `InputIteratorTransformer`, the representation of this time depends on its previous operator.

Below are the 3 different cases that I observe:

1. **When the previous operator is shuffle:**
The time is primarily the total shuffle read time, including fetch wait time and native reader processing time (such as decompression and deserialization).
2. **When the previous operator is broadcast:**
The time is nearly zero because the broadcast is already executed before the pipeline starts.
3. **For other cases (e.g., ColumnarUnion or fallback operators within the same Spark stage as the previous Velox pipelines):**
Since wallTimeNanos in Velox is measured by the driver’s getOutput, the time of the previous pipelines is included in the getOutput call from the ValueStreamNode. In this case, the time of input iterators represents the total time counted from the beginning of the current stage.

The discrepancy in behavior across different cases for this operator is not documented and may cause confusion for users. It would be better to document this and highlight it in the metrics description.

### Gluten version

None

marin-ma:
cc: @FelixYBW

Issue #8063, opened 2024-11-27
inputIteratorTransformer exhibits unexpectedly high time cost


zsmj2017:
### Backend

VL (Velox)

### Bug description

Currently, while executing vectorized etls, we unexpectedly found that inputIteratorTransformer is consuming a significant amount of CPU time. 

The Spark UI is as follows:

<img width="409" alt="image" src="https://github.com/user-attachments/assets/f7e8fe31-efa5-48fc-a4a7-3e5c4fd540b1">
<img width="419" alt="image" src="https://github.com/user-attachments/assets/944ec844-2fa1-4edf-9dd6-b814864f1158">
<img width="452" alt="image" src="https://github.com/user-attachments/assets/1120adc6-9900-423a-82cc-8d2771c0e9c1">

The image shows the process from Table Scan to inputIteratorTransformer.
inputIteratorTransformer includes the total execution time of all previous operators, but from the CPU time perspective, it doesn’t seem to be the case.

Spark version: spark-3.0
Gluten version: branch-1.2

### Spark version

None

PR #11724, opened 2026-03-09
[GLUTEN-8063][VL] Export cpuNanos metric for InputIteratorTransformer

n0r0shi:
## Summary
  Surfaces Velox's `cpuWallTiming.cpuNanos` as a separate Spark SQL metric (`cpu time`) on `InputIteratorTransformer`. Currently only `wallNanos` and `cpuCount` are exposed — `cpuNanos` is available from Velox but not propagated to Gluten. This change threads it through: C++ → JNI → Java → Scala SQLMetric. The metric automatically appears in Spark UI and event logs.

  ## Motivation
  Related to #10618. `InputIteratorTransformer` sits at the boundary between Spark (shuffle/broadcast) and Velox. Its wall time includes I/O waits, not just CPU work. Exposing `cpuNanos` separately lets users distinguish compute-bound vs I/O-bound bottlenecks. For most other operators, `cpuNanos ≈ wallNanos` — this boundary operator is where they diverge the most.

Velox already tracks `cpuNanos` on every operator — exposing it as-is is the simplest approach and better than not surfacing CPU time at all.

  ## Open questions
  - Any concerns with the metric name / label (`cpu time`)?
  - Should any other operators also expose `cpuNanos`? (e.g., operators that spill to disk)

  Tests will be added once the approach is confirmed.

Related issue: #8063

### DIFF
--- a/backends-velox/src/main/java/org/apache/gluten/metrics/Metrics.java
+++ b/backends-velox/src/main/java/org/apache/gluten/metrics/Metrics.java
@@ -29,6 +29,7 @@ public class Metrics implements IMetrics {
   public long[] outputBytes;
   public long[] cpuCount;
   public long[] wallNanos;
+  public long[] cpuNanos;
   public long[] scanTime;
   public long[] peakMemoryBytes;
   public long[] numMemoryAllocations;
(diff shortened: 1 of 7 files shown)


FelixYBW:
Looks good. Can you fix conflict?

PR #11723, opened 2026-03-09
[GLUTEN-11550][UT] Enable GlutenParquetVariantShreddingSuite for Spark 4.1

baibaichen:
## What changes are proposed in this pull request?

Enable `GlutenParquetVariantShreddingSuite` for Spark 4.1 by adding fallback logic for Parquet variant logical type annotations.

Spark 4.1 introduced Parquet variant logical type annotations (`PARQUET_ANNOTATE_VARIANT_LOGICAL_TYPE`, `PARQUET_IGNORE_VARIANT_ANNOTATION`). When reading a variant-annotated Parquet file with a non-variant schema, Spark's `ParquetSchemaConverter` validates the annotation and throws an error. Velox native reader does not check variant annotations, so the scan must fall back to vanilla Spark for correct behavior.

Changes:
- Add `shouldFallbackForParquetVariantAnnotation` shim method in `SparkShims` (default: `false`)
- Implement variant annotation detection in `Spark41Shims`: checks `PARQUET_IGNORE_VARIANT_ANNOTATION` config and recursively scans Parquet schema for `VariantLogicalTypeAnnotation`
- Add `validateVariantAnnotation` in `ParquetMetadataUtils` (not gated by `parquetMetadataValidationEnabled` since this is a correctness issue)
- Call `validateVariantAnnotation` from `VeloxBackend.validateMetadata()`
- Enable `GlutenParquetVariantShreddingSuite` in spark41 `VeloxTestSettings`

## How was this patch tested?

- spark41 `GlutenParquetVariantShreddingSuite`: **7/7 passed** (including the previously failing `"variant logical type annotation - ignore variant annotation"`)
- spark40 `GlutenParquetVariantShreddingSuite`: **5/5 passed** (no regression)

## Was this patch authored or co-authored using generative AI tooling?

Generated-by: GitHub Copilot (Claude)

Related issue: #11550

PR #11726, opened 2026-03-09
[GLUTEN-11550][VL][UT] Enable Variant test suites

baibaichen:

Enable `GlutenVariantEndToEndSuite`, `GlutenVariantShreddingSuite`, and `GlutenParquetVariantShreddingSuite` for both spark40 and spark41.

Changes:

1. **VeloxValidatorApi.scala**: Detect variant shredded structs (produced by Spark's `PushVariantIntoScan`) by checking for `__VARIANT_METADATA_KEY` metadata on struct fields. Triggers fallback to Spark's native Parquet reader since Velox cannot read variant shredding encoding.

2. **ParquetMetadataUtils.scala**: Add variant annotation check in `isUnsupportedMetadata`, gated by `parquetMetadataValidationEnabled`. Reads the same footer, no extra I/O.

3. **Spark41Shims.scala**: Add `shouldFallbackForParquetVariantAnnotation` to detect Parquet variant logical type annotations.

4. **GlutenParquetVariantShreddingSuite (spark41)**: Set `parquetMetadataValidationEnabled=true` to enable variant annotation fallback detection.

5. **pom.xml**: Add `-Dfile.encoding=UTF-8` to test JVM args. On JDK 17 with `LANG=C` (CI containers centos-8/9), the default charset is US-ASCII causing garbled output for multi-byte characters. JDK 18+ defaults to UTF-8 via [JEP 400](https://openjdk.org/jeps/400). See [VariantUtil.java#L508](https://github.com/apache/spark/blob/v4.0.1/common/variant/src/main/java/org/apache/spark/types/variant/VariantUtil.java#L508).

## How was this patch tested?

- spark40: `GlutenVariantEndToEndSuite` 14✅, `GlutenVariantShreddingSuite` 8✅, `GlutenParquetVariantShreddingSuite` 5✅
- spark41: `GlutenVariantEndToEndSuite` 14✅, `GlutenVariantShreddingSuite` 8✅, `GlutenParquetVariantShreddingSuite` 7✅

Generated-by: GitHub Copilot CLI

Related issue: #11550
Note: This PR subsumes #11723 (GlutenParquetVariantShreddingSuite).

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
@@ -121,6 +121,15 @@ object VeloxValidatorApi {
       case map: MapType =>
         validateSchema(map.keyType).orElse(validateSchema(map.valueType))
       case struct: StructType =>
+        // Detect variant shredded struct produced by Spark's PushVariantIntoScan.
+        // These structs have all fields annotated with __VARIANT_METADATA_KEY metadata.
+        // Velox cannot read the variant shredding encoding in Parquet files.
+        if (
+          struct.fields.nonEmpty &&
+          struct.fields.forall(_.metadata.contains("__VARIANT_METADATA_KEY"))
+        ) {
+          return Some(s"Variant shredded struct is not supported: $struct")
+        }
         struct.foreach {
           field =>
             val reason = validateSchema(field.dataType)
--- a/backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
@@ -148,6 +148,12 @@ object ParquetMetadataUtils extends Logging {
       isTimezoneFoundInMetadata(footer, parquetOptions)
     )
 
+    // Variant annotation check: Velox native reader does not check variant annotations,
+    // so fallback to vanilla Spark when detected.
+    if (SparkShimLoader.getSparkShims.shouldFallbackForParquetVariantAnnotation(footer)) {
+      return Some("Variant annotation detected in Parquet file.")
+    }
+
     for (check <- validationChecks) {
       if (check.isDefined) {
         return check
--- a/shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
+++ b/shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
@@ -237,6 +237,8 @@ trait SparkShims {
 
   def isParquetFileEncrypted(footer: ParquetMetadata): Boolean
 
+  def shouldFallbackForParquetVariantAnnotation(footer: ParquetMetadata): Boolean = false
+
   def getOtherConstantMetadataColumnValues(file: PartitionedFile): JMap[String, Object] =
     Map.empty[String, Any].asJava.asInstanceOf[JMap[String, Object]]
 
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxBackend.scala
@@ -209,10 +209,9 @@ object VeloxBackendSettings extends BackendSettingsApi {
       }
       val fileLimit = GlutenConfig.get.parquetMetadataFallbackFileLimit
       val parquetOptions = new ParquetOptions(CaseInsensitiveMap(properties), SQLConf.get)
-      val parquetMetadataValidationResult =
-        ParquetMetadataUtils.validateMetadata(rootPaths, hadoopConf, parquetOptions, fileLimit)
-      parquetMetadataValidationResult.map(
-        reason => s"Detected unsupported metadata in parquet files: $reason")
+      ParquetMetadataUtils
+        .validateMetadata(rootPaths, hadoopConf, parquetOptions, fileLimit)
+        .map(reason => s"Detected unsupported metadata in parquet files: $reason")
     }
 
     def validateDataSchema(): Option[String] = {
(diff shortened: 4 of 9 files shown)


baibaichen:
@jinchengchenghh @zhztheplayer please review

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
```diff
@@ -89,77 +74,70 @@ object ParquetMetadataUtils extends Logging {
...
+      footerOrError: Either[Exception, ParquetMetadata],
+      parquetOptions: ParquetOptions,
+      metadataValidationEnabled: Boolean): Option[String] = {
+    footerOrError match {
+      case Left(e)
+          if metadataValidationEnabled &&
+            ExceptionUtils.hasCause(e, classOf[ParquetCryptoRuntimeException]) =>
+        Some("Encrypted Parquet footer detected.")
+      case Left(_: RuntimeException) =>
+        // Ignored as it's could be a "Not a Parquet file" exception.
```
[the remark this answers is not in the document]
baibaichen: same as previous version

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
```diff
@@ -24,53 +24,38 @@ import org.apache.spark.sql.execution.datasources.DataSourceUtils
...
   ): Option[String] = {
-    if (!GlutenConfig.get.parquetMetadataValidationEnabled) {
-      None
+    val enabled = GlutenConfig.get.parquetMetadataValidationEnabled
+    if (enabled || SparkShimLoader.getSparkShims.needsVariantAnnotationCheck) {
+      parquetFooters(rootPaths, hadoopConf, fileLimit)
+        .map(isUnsupportedMetadata(_, parquetOptions, enabled))
+        .find(_.isDefined)
+        .flatten
     } else {
```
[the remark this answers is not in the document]
baibaichen: we need do such update, otherwise UT will be faield

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
```diff
@@ -121,6 +121,15 @@ object VeloxValidatorApi {
       case map: MapType =>
         validateSchema(map.keyType).orElse(validateSchema(map.valueType))
       case struct: StructType =>
+        // Detect variant shredded struct produced by Spark's PushVariantIntoScan.
+        // These structs have all fields annotated with __VARIANT_METADATA_KEY metadata.
+        // Velox cannot read the variant shredding encoding in Parquet files.
+        if (
+          struct.fields.nonEmpty &&
```
jinchengchenghh, line 128: Do we need to add nonEmpty check?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
```diff
@@ -121,6 +121,15 @@ object VeloxValidatorApi {
...
+          struct.fields.forall(_.metadata.contains("__VARIANT_METADATA_KEY"))
+        ) {
+          return Some(s"Variant shredded struct is not supported: $struct")
```
jinchengchenghh, line 131: Might the struct $struct too heavy?

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
```diff
@@ -24,53 +24,38 @@ import org.apache.spark.sql.execution.datasources.DataSourceUtils
...
   def validateMetadata(
       rootPaths: Seq[String],
       hadoopConf: Configuration,
       parquetOptions: ParquetOptions,
       fileLimit: Int
   ): Option[String] = {
-    if (!GlutenConfig.get.parquetMetadataValidationEnabled) {
-      None
+    val enabled = GlutenConfig.get.parquetMetadataValidationEnabled
+    if (enabled || SparkShimLoader.getSparkShims.needsVariantAnnotationCheck) {
```
jinchengchenghh: Can we reuse parquetMetadataValidationEnabled? the default value is false

shims/common/src/main/scala/org/apache/gluten/sql/shims/SparkShims.scala
```diff
@@ -237,6 +237,10 @@ trait SparkShims {
 
   def isParquetFileEncrypted(footer: ParquetMetadata): Boolean
 
+  def shouldFallbackForParquetVariantAnnotation(footer: ParquetMetadata): Boolean = false
```
jinchengchenghh, line 240: We would we not fallback?

baibaichen: Default `false` because Spark 3.x and 4.0 don't have Parquet variant logical type annotations — only Spark 4.1 introduced `VariantLogicalTypeAnnotation`. `Spark41Shims` overrides this to actually check the Parquet schema. For older Spark versions, there's nothing to detect, so no fallback needed.

backends-velox/src/main/scala/org/apache/gluten/utils/ParquetMetadataUtils.scala
baibaichen: Updated. Variant annotation check is now gated by `parquetMetadataValidationEnabled` (same as other checks). For Spark 4.1 `GlutenParquetVariantShreddingSuite`, we set `parquetMetadataValidationEnabled=true` via `sparkConf` in the test suite.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
baibaichen: This is consistent with the existing pattern at line 144: `s"Schema / data type not supported: $schema"` which also prints the full type. Since this is a validation failure message (logged once per scan fallback, not per row), the overhead is negligible.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
baibaichen: Yes, `nonEmpty` is needed because `forall` returns `true` for an empty collection. Without it, an empty struct would be incorrectly detected as variant shredded.

PR #11743, opened 2026-03-11
[DNR] Temp test

infvg:


backends-velox/src-iceberg/main/scala/org/apache/gluten/connector/write/IcebergDataWriteFactory.scala
```diff
@@ -117,4 +120,14 @@ case class IcebergDataWriteFactory(
...
   }
+
+  private def filterMetadataColumns(schema: StructType): StructType = {
+    val filteredFields = schema.fields.filterNot {
+      field =>
+        {
+          SparkShimLoader.getSparkShims.isRowIndexMetadataColumn(field.name);
+        }
+    }
+    StructType(filteredFields)
```
zhouyuan: also need to filter out those columns in the native code
https://github.com/apache/gluten/blob/main/cpp/velox/compute/iceberg/IcebergWriter.cc#L215

backends-velox/src-iceberg/main/scala/org/apache/gluten/connector/write/IcebergDataWriteFactory.scala
```diff
@@ -117,4 +120,14 @@ case class IcebergDataWriteFactory(
     cSchema.close()
     (writer, jniWrapper)
   }
+
+  private def filterMetadataColumns(schema: StructType): StructType = {
+    val filteredFields = schema.fields.filterNot {
+      field =>
+        {
+          SparkShimLoader.getSparkShims.isRowIndexMetadataColumn(field.name);
```
zhouyuan: it looks like this method is for delta only

PR #11754, opened 2026-03-13
[VL] Close ColumnarBatch in ColumnarCollectLimitExec for skipped batches

acvictor:
This PR closes `ColumnarBatch` for skipped batches in `ColumnarCollectLimitExec`.

`ColumnarCollectLimitExec.fetchNextBatch()` leaks memory for skipped batches - when a batch falls entirely within the offset range, it is consumed from the input iterator but never closed.   

The `ColumnarBatch` object is small but it holds a JNI handle to a native Velox vector. Since ColumnarBatch has no finalizer, the native memory is only freed by an explicit close() call. 

## How was this patch tested?
Existing UTs

No

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
@@ -83,9 +83,7 @@ case class ColumnarCollectLimitExec(
                 ColumnarBatches.retain(batch)
                 batch
               } else {
-                val sliced = VeloxColumnarBatches.slice(batch, startIndex, needed)
-                batch.close()
-                sliced
+                VeloxColumnarBatches.slice(batch, startIndex, needed)
               }
 
             rowsToCollect -= needed


acvictor:
@ArnavBalyan can you please review this fix? Thanks!

backends-velox/src/main/scala/org/apache/gluten/execution/ColumnarCollectLimitExec.scala
```diff
@@ -82,7 +83,9 @@ case class ColumnarCollectLimitExec(
                 ColumnarBatches.retain(batch)
                 batch
               } else {
-                VeloxColumnarBatches.slice(batch, startIndex, needed)
+                val sliced = VeloxColumnarBatches.slice(batch, startIndex, needed)
```
jinchengchenghh: The `sliced` calls velox slice API, reuse buffers of `batch`, so we cannot close it.

acvictor: I missed this, thanks!

PR #11756, opened 2026-03-13
[GLUTEN-11578][CORE] Remove RAS

### LINKED ISSUE #11578: [VL] Remove RAS
It turns out the RAS optimizer is causing more maintenance burden than what it helped. We'd remove it from Gluten and I'll maintain it as a small individual project forward.

Related discussion. https://github.com/apache/incubator-gluten/pull/11455#discussion_r2713844340

zhztheplayer:
Fixes #11578

Breaking change for common usage:

Option `spark.gluten.ras.costModel` is renamed to `spark.gluten.costModel`

### DIFF
--- a/gluten-kafka/src/main/scala/org/apache/gluten/execution/OffloadKafkaScan.scala
+++ b/gluten-kafka/src/main/scala/org/apache/gluten/execution/OffloadKafkaScan.scala
@@ -23,7 +23,7 @@ import org.apache.gluten.extension.columnar.validator.Validators
 import org.apache.gluten.extension.injector.Injector
 
 import org.apache.spark.sql.execution.SparkPlan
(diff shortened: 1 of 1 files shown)


jinchengchenghh:
Could we extend the Spark cost model for ColumnarBatch?

zhztheplayer:
> Could we extend the Spark cost model for ColumnarBatch?

@jinchengchenghh would you elaborate a bit, which Spark cost model did you mean?

jinchengchenghh:
Now the Spark cost model is based on Row operator model, can we make it extensible for other Columnar operator, then we may only override some functions to make the cost model work in Gluten.

zhztheplayer:
> Now the Spark cost model is based on Row operator model, can we make it extensible for other Columnar operator, then we may only override some functions to make the cost model work in Gluten.

Does this correspond to some ongoing effort for another cost-based optimization? Would you like to share more context about that?

PR #11757, opened 2026-03-13
[CORE] Refactor ExpressionConverter to use explicit Option return instead of Option(null)

acvictor:
This PR replaces `Option { expr match { ... case _ => null } }` with explicit `Some(...)`/`None` returns in `tryTransformWithoutExpressionMapping` to make the intent clearer and avoid the subtle dependency on Option.apply's null-handling behavior.

## How was this patch tested?
Existing UTs.

No

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHRuleApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHRuleApi.scala
@@ -164,22 +164,6 @@ object CHRuleApi {
   }
 
   /**
-   * Registers Gluten's columnar rules. These rules will be executed only when RAS (relational
-   * algebra selector) is enabled by spark.gluten.ras.enabled=true.
-   *
-   * These rules are covered by CI test job spark-test-spark35-ras.
-   */
-  private def injectRas(injector: RasInjector): Unit = {
-    // CH backend doesn't work with RAS at the moment. Inject a rule that aborts any
-    // execution calls.
-    injector.injectPreTransform(
-      _ =>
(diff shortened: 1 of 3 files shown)


acvictor:
@baibaichen can you please review this PR? Thanks!

acvictor:
> This is not a bug fix, it's a code optimization. `Option(null)` is equivalent to None, `Some(null)` is not equivalent, `Option` is used here. Please change the title and description.

Done, thanks for the review!

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -284,21 +284,21 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+      case scalaUDF: ScalaUDF =>
+        Some(replaceScalaUDFWithExpressionTransformer(scalaUDF, attributeSeq, expressionsMap))
+      case _ if HiveUDFTransformer.isHiveUDF(expr) =>
+        Some(
+          BackendsApiManager.getSparkPlanExecApiInstance.genHiveUDFTransformer(expr, attributeSeq))
+      case staticInvoke: StaticInvoke =>
+        Some(
+          replaceStaticInvokeWithExpressionTransformer(staticInvoke, attributeSeq, expressionsMap))
+      case invoke: Invoke =>
+        Some(replaceInvokeWithExpressionTransformer(invoke, attributeSeq, expressionsMap))
```
zml1206: Please use `Option` instead of `Some` to prevent the existence of `Some(null)`, although it should not exist.

acvictor: Done

PR #11769, opened 2026-03-16
[GLUTEN-11605][VL] Write per-block column statistics in shuffle writer

acvictor:
This PR adds per-block column statistics (min/max/hasNull) to the shuffle writer pipeline as a prerequisite for block-level pruning using dynamic filters at the shuffle reader. When `spark.gluten.sql.columnar.backend.velox.valueStream.dynamicFilter.enabled` is true, the shuffle writer computes per-column min/max statistics from raw Arrow buffers during `evictBuffers()` and serializes them as a `kStatisticsPayload` block before each non-dictionary payload in the output file. This mirrors how parquet row group statistics enable predicate pushdown.

## How was this patch tested?
Added new tests and also ran the CI with config set to true.

No

Related issue: #11605

### DIFF
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -945,7 +945,8 @@ Java_org_apache_gluten_vectorized_LocalPartitionWriterJniWrapper_createPartition
     jint shuffleFileBufferSize,
     jstring dataFileJstr,
     jstring localDirsJstr,
-    jboolean enableDictionary) {
+    jboolean enableDictionary,
+    jboolean enableBlockStatistics) {
   JNI_METHOD_START
 
   const auto ctx = getRuntime(env, wrapper);
@@ -961,14 +962,7 @@ Java_org_apache_gluten_vectorized_LocalPartitionWriterJniWrapper_createPartition
       mergeThreshold,
       numSubDirs,
       enableDictionary);
-
-  // Reuse the dynamic filter config to also enable block statistics collection,
-  // since stats are only useful when dynamic filter pushdown is active.
-  const auto& confMap = ctx->getConfMap();
-  auto it = confMap.find("spark.gluten.sql.columnar.backend.velox.valueStream.dynamicFilter.enabled");
-  if (it != confMap.end() && it->second == "true") {
-    partitionWriterOptions->blockStatisticsEnabled = true;
-  }
+  partitionWriterOptions->blockStatisticsEnabled = enableBlockStatistics;
 
   auto partitionWriter = std::make_shared<LocalPartitionWriter>(
       numPartitions,
--- a/cpp/core/jni/JniWrapper.cc
+++ b/cpp/core/jni/JniWrapper.cc
@@ -962,6 +962,14 @@ Java_org_apache_gluten_vectorized_LocalPartitionWriterJniWrapper_createPartition
       numSubDirs,
       enableDictionary);
 
+  // Reuse the dynamic filter config to also enable block statistics collection,
+  // since stats are only useful when dynamic filter pushdown is active.
+  const auto& confMap = ctx->getConfMap();
+  auto it = confMap.find("spark.gluten.sql.columnar.backend.velox.valueStream.dynamicFilter.enabled");
+  if (it != confMap.end() && it->second == "true") {
+    partitionWriterOptions->blockStatisticsEnabled = true;
+  }
+
   auto partitionWriter = std::make_shared<LocalPartitionWriter>(
       numPartitions,
       createCompressionCodec(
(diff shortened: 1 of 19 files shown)


acvictor:
@marin-ma @zhztheplayer this is ready for review. I will push one more commit to again disable by default.

cpp/core/shuffle/BlockStatistics.cc
```diff
@@ -0,0 +1,316 @@
...
+}
+
+// Returns true if the column has any null rows.
+bool hasAnyNull(const std::shared_ptr<arrow::Buffer>& validityBuffer, uint32_t numRows) {
+  if (!validityBuffer || numRows == 0) {
+    return false;
+  }
+  // Check each bit — return early on first null found.
+  for (uint32_t i = 0; i < numRows; ++i) {
+    if (!arrow::bit_util::GetBit(validityBuffer->data(), i)) {
```
marin-ma: Perhaps check each bytes by comparing with 0xff can be faster

cpp/core/jni/JniWrapper.cc
```diff
@@ -962,6 +962,14 @@ Java_org_apache_gluten_vectorized_LocalPartitionWriterJniWrapper_createPartition
       numSubDirs,
       enableDictionary);
 
+  // Reuse the dynamic filter config to also enable block statistics collection,
+  // since stats are only useful when dynamic filter pushdown is active.
+  const auto& confMap = ctx->getConfMap();
+  auto it = confMap.find("spark.gluten.sql.columnar.backend.velox.valueStream.dynamicFilter.enabled");
```
marin-ma: Other option values are passing through function args. Can you add a new arg `enableBlockStatistics`?

cpp/core/shuffle/Payload.cc
```diff
@@ -413,12 +437,27 @@ arrow::Result<std::unique_ptr<InMemoryPayload>> InMemoryPayload::merge(
...
   }
-  return std::make_unique<InMemoryPayload>(mergedRows, isValidityBuffer, source->schema(), std::move(merged));
+  auto result = std::make_unique<InMemoryPayload>(mergedRows, isValidityBuffer, source->schema(), std::move(merged));
+  // Merge block statistics if both payloads have them.
+  if (source->hasBlockStats() && append->hasBlockStats()) {
+    auto mergedStats = *source->blockStats_;
+    mergedStats.merge(*append->blockStats_);
+    result->setBlockStats(std::move(mergedStats));
+  } else if (source->hasBlockStats()) {
+    result->setBlockStats(*source->blockStats_);
```
marin-ma: If only source or append has blockStats, should we either discard it or compute for the missing side and merge them?

marin-ma (commented):
@acvictor Thanks for adding this feature! Please check my comments below.

As adding the statistics can grow the shuffle data size, have you tested the overall growth of shuffled data size for tpch/tpcds benchmarks?

cpp/core/shuffle/BlockStatistics.cc
acvictor: Done, thanks!

cpp/core/jni/JniWrapper.cc
acvictor: Done.

cpp/core/shuffle/Payload.cc
acvictor: Done - stats are only kept when both payloads have them and if one side is missing stats are discarded

cpp/core/shuffle/BlockStatistics.cc
```diff
@@ -0,0 +1,326 @@
...
+  const auto* values = reinterpret_cast<const T*>(valueBuffer->data());
+  bool foundAny = false;
+  T minVal{};
+  T maxVal{};
+
+  for (uint32_t i = 0; i < numRows; ++i) {
+    if (!isRowValid(validityBuffer, i)) {
+      continue;
+    }
+    T val = values[i];
```
weiting-chen, line 92: Check if require to consider NaN cases,
NaN may be silently skipped by the comparisons (neither updates min nor max) and making minVal = maxVal = NaN.
See if require to add a NaN check:
if constexpr (std::is_floating_point_v<T>) {
   if (std::isnan(val)) continue; // Skip NaN values
 }

cpp/core/shuffle/Payload.cc
```diff
@@ -277,6 +277,30 @@ arrow::Status BlockPayload::serialize(arrow::io::OutputStream* outputStream) {
...
+        if (buffer && buffer->size() > 0) {
+          size += buffer->size();
+        }
+      }
+      return size;
+    }
+    case Type::kCompressed: {
+      int64_t size = sizeof(Type) + sizeof(uint32_t) + sizeof(uint32_t); // type + numRows + numBuffers
+      if (!buffers_.empty() && buffers_[0]) {
+        size += buffers_[0]->size();
```
weiting-chen, line 295: Missing buffer size field?
size += sizeof(int64_t) + buffers_[0]->size(); // buffer size field + data

acvictor:
> @acvictor Thanks for adding this feature! Please check my comments below.
> 
> As adding the statistics can grow the shuffle data size, have you tested the overall growth of shuffled data size for tpch/tpcds benchmarks?

Will revert on this soon

PR #11775, opened 2026-03-17
[GLUTEN-11774][VL] Use runtime stats to choose hash build side

wecharyu:
When AQE is enabled, we can use `QueryStageExec`'s runtime stats to choose the smaller side as optimal build side.

Fix https://github.com/apache/gluten/issues/11774.

## How was this patch tested?
Test locally.

No


wecharyu:
This patch could estimate right plan: 
```bash
[pool-4-thread-1] OffloadJoin: Estimated join output: rows=250, bytes=23962
```

image

rui-mo:
@wecharyu Apologies for the late response. Do we still need this PR?

wecharyu:
@rui-mo I think this patch is unnecessary, since it estimates the child output stats in the same way as Spark CBO.

rui-mo:
@wecharyu Got it. Thanks.

PR #11776, opened 2026-03-17
[VL] Added iceberg write configs

infvg:
Adds support for: `write.target-file-size-bytes`, `write.parquet.page-size-bytes`
Resolves part of https://github.com/apache/gluten/issues/11703


rui-mo (commented):
Looks good to me. I’ll also loop in @jinchengchenghh for a final pass.

backends-velox/src/main/scala/org/apache/gluten/config/VeloxConfig.scala
```diff
@@ -755,4 +755,16 @@ object VeloxConfig extends ConfigRegistry {
       .doc("Maps table field names to file field names using names, not indices for Parquet files.")
       .booleanConf
       .createWithDefault(true)
+
+  val PARQUET_PAGE_SIZE_BYTES =
```
jinchengchenghh, line 781: Please remove the config, the config should not exposed to users, we should respect iceberg corresponding config, in this case, please update the config map as https://github.com/apache/gluten/blob/main/backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala#L37

Please find the codec path, and set to Velox config.

infvg: Changed the implementation. Now we create an IcebergProperties object so that we can move as many properties in the future down to the Velox layer. It should make it easier to pass stuff down & it doesn't expose anything to the user.

backends-velox/src/main/resources/org/apache/gluten/proto/IcebergProperties.proto
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+syntax = "proto3";
+
+package gluten;
+
+option java_package = "org.apache.gluten.proto";
+option java_multiple_files = true;
+
+message IcebergProperties {
```
jinchengchenghh: We already have ConfigMap

docs/get-started/VeloxIceberg.md
```diff
@@ -193,54 +193,54 @@ extracted from https://iceberg.apache.org/docs/latest/configuration/
...
-| write.data.path | table location + /data | Base location for data files |  |
-| write.metadata.path | table location + /metadata | Base location for metadata files |  |
-| write.delete.mode | copy-on-write | Mode used for delete commands: copy-on-write or merge-on-read (v2 and above) |  |
-| write.delete.isolation-level | serializable | Isolation level for delete commands: serializable or snapshot |  |
-| write.update.mode | copy-on-write | Mode used for update commands: copy-on-write or merge-on-read (v2 and above) |  |
-| write.update.isolation-level | serializable | Isolation level for update commands: serializable or snapshot |  |
-| write.merge.mode | copy-on-write | Mode used for merge commands: copy-on-write or merge-on-read (v2 and above) |  |
-| write.merge.isolation-level | serializable | Isolation level for merge commands: serializable or snapshot |  |
-| write.delete.granularity | partition | Controls the granularity of generated delete files: partition or file |  |
+| --- | --- | --- | - |
```
jinchengchenghh: The document format looks not correct, do you use https://github.com/apache/gluten/blob/main/dev/gen-all-config-docs.sh to generate the configs?

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergWriteExec.scala
```diff
@@ -117,4 +131,8 @@ trait IcebergWriteExec extends ColumnarV2TableWriteExec {
     ValidationResult.succeeded
   }
 
+  private def normalizeCapacityString(value: String): String = {
+    val trimmed = value.trim
```
jinchengchenghh, line 133: Do you require the normalize?, what if user sets `5MB`?

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergWriteExec.scala
```diff
@@ -49,6 +49,20 @@ trait IcebergWriteExec extends ColumnarV2TableWriteExec {
...
+    val config = IcebergWriteUtil.getWriteProperty(write)
+    config.getOrDefault(
+      WRITE_TARGET_FILE_SIZE_BYTES,
+      WRITE_TARGET_FILE_SIZE_BYTES_DEFAULT.toString);
+  }
+
+  protected def getParquetPageSizeBytes: String = {
+    val config = IcebergWriteUtil.getWriteProperty(write)
+    config.getOrDefault(
+      normalizeCapacityString(PARQUET_PAGE_SIZE_BYTES),
```
jinchengchenghh: If  this helps?
def maxBroadcastTableSize: Long =
    JavaUtils.byteStringAsBytes(conf.getConfString(SPARK_MAX_BROADCAST_TABLE_SIZE, "8GB"))

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergWriteExec.scala
infvg: The value in TableProperties is bytes as a number. If the value is 5MB it will be unchanged. But if the value is just a number eg 123 it is in bytes so we add B

jinchengchenghh: I remember we always transfer the original number to native side, that means, without `MB`, `B`

infvg: For the other config it's fine, but this one goes through the velox `toCapacity` function which requires a unit

backends-velox/src/main/resources/org/apache/gluten/proto/IcebergProperties.proto
infvg: Switched implementation to use ConfigMap

docs/get-started/VeloxIceberg.md
infvg: Yes it removed the space I think because the config support I added is the first one in this table so it changed the spacing to match

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergWriteExec.scala
infvg: I think this is a different config, I mapped `hive.parquet.writer.page_size` to Iceberg's `write.parquet.page-size-bytes`

docs/get-started/VeloxIceberg.md
jinchengchenghh: Hi, @FelixYBW Do you know if https://github.com/apache/gluten/blob/main/dev/gen-all-config-docs.sh can generate the VeloxIceberg.md? I'm not sure if the change is right.

backends-velox/src-iceberg/main/scala/org/apache/gluten/connector/write/IcebergDataWriteFactory.scala
```diff
@@ -113,7 +122,8 @@ case class IcebergDataWriteFactory(
         taskId,
         operationId,
         partitionSpec.toByteArray,
-        field.toByteArray)
+        field.toByteArray
```
jinchengchenghh: Unrelated change

cpp/velox/utils/ConfigExtractor.cc
```diff
@@ -244,6 +244,11 @@ std::shared_ptr<facebook::velox::config::ConfigBase> createHiveConnectorSessionC
   configs[facebook::velox::connector::hive::HiveConfig::kOrcUseColumnNamesSession] =
       conf->get<bool>(kOrcUseColumnNames, true) ? "true" : "false";
 
+  configs[facebook::velox::connector::hive::HiveConfig::kMaxTargetFileSize] =
```
jinchengchenghh: We already have this config https://github.com/apache/gluten/pull/11776/changes#diff-de468e0497c9eb8ce0cce6b72646a18df94e3d25e17cdfb9e6acaa1985a1eb62L236

infvg: Removed

backends-velox/src-iceberg/main/scala/org/apache/gluten/connector/write/IcebergDataWriteFactory.scala
infvg: Removed

cpp/velox/config/VeloxConfig.h
```diff
@@ -185,6 +185,9 @@ const uint32_t kGlogVerboseLevelMaximum = 99;
 const std::string kGlogSeverityLevel = "spark.gluten.sql.columnar.backend.velox.glogSeverityLevel";
 const uint32_t kGlogSeverityLevelDefault = 1;
 
+// Iceberg write configs
+const std::string kWriteParquetPageSizeBytes = "spark.gluten.sql.columnar.backend.velox.parquet.page-size-bytes";
```
jinchengchenghh: Please align with other config format pageSizeBytes

backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala
```diff
@@ -39,6 +42,23 @@ abstract class AbstractIcebergWriteExec extends IcebergWriteExec {
     val filteredSchema = StructType(
       schema.fields.filter(field => writeFieldNames.contains(field.name))
     )
+
+    val icebergProperties = new util.HashMap[String, String]
+    Map(
```
jinchengchenghh: Please add it to VeloxConfig and reference it, after this change, hive can  also set the config

backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala
```diff
@@ -39,6 +42,23 @@ abstract class AbstractIcebergWriteExec extends IcebergWriteExec {
...
+      "spark.gluten.sql.columnar.backend.velox.parquet.page-size-bytes" -> getParquetPageSizeBytes
+    ).++(
+      if (
+        SQLConf.get
+          .getConfString("spark.gluten.sql.columnar.backend.velox.maxTargetFileSize", null) != null
```
jinchengchenghh: Please don't use the plain string, use the VeloxConfig.xxx.key

gluten-iceberg/src/main/scala/org/apache/gluten/execution/IcebergWriteExec.scala
jinchengchenghh: Then you only need to check if the end character is number

cpp/velox/config/VeloxConfig.h
infvg: Fixed

backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala
infvg: Changed

backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala
infvg: Fixed

docs/get-started/VeloxIceberg.md
jinchengchenghh: I don't think this is the reason. Please only add your change, maybe your local code editor does the change

docs/get-started/VeloxIceberg.md
jinchengchenghh: This config should also be marked as supported

jinchengchenghh (approved):
Thanks for your work and your iterating.

PR #11777, opened 2026-03-17
[VL] Reuse byte[] buffers in shuffle read and broadcast serialization paths

acvictor:
This PR makes changes to reuse temporary byte[] buffers instead of allocating new ones on every call.
- OnHeapJniByteInputStream.read(): Cache the buffer as a field; only reallocate when a larger size is requested
- UnsafeByteArray: Share a single lazily-initialized transient byte[] field across all methods

## How was this patch tested?
Existing UTs

No


acvictor:
@zhztheplayer can you please review this change? Thanks!

PR #11778, opened 2026-03-18
[SETUP] Use `#!/usr/bin/env bash` instead of `#!/bin/bash` in scripts

Zouxxyy:

Replace all `#!/bin/bash` shebangs with `#!/usr/bin/env bash` in project-owned shell scripts.
Third-party dependencies and documentation code blocks are left unchanged.

Some operating systems (e.g. NixOS) do not have a Bash executable at `/bin/bash`, so using
`env` to locate bash improves cross-platform compatibility.

For example, running `dev/run-scala-test.sh` on macOS may fail when Bash is installed via
Homebrew at `/opt/homebrew/bin/bash` rather than `/bin/bash`, and the hardcoded `#!/bin/bash`
shebang picks up the outdated system Bash (3.2) instead of the user-installed version. Apple
ships Bash 3.2 and cannot upgrade it due to the GPLv3 license of Bash 4+, so macOS users who
need modern Bash features must rely on a separately installed copy, which `#!/usr/bin/env bash`
will correctly resolve.

This aligns with the same change adopted by GitHub Actions runner
(see https://github.com/actions/runner/pull/314) and is also consistent with what Apache Spark
does in its codebase.

Additionally, a shebang check is added to `.github/workflows/util/check.sh` so that future PRs
introducing `#!/bin/bash` in new or modified `.sh` files will be caught by the License Header
Check CI workflow.

## How was this patch tested?

No functional change. Verified all project-owned `.sh` files now use `#!/usr/bin/env bash`
and confirmed scripts execute correctly on macOS with Homebrew Bash. The new CI shebang check
was tested locally against both good and bad shebangs.

Generated-by: Kiro (Claude Opus 4.6)

### DIFF
--- a/.github/workflows/util/check.sh
+++ b/.github/workflows/util/check.sh
@@ -15,6 +15,9 @@
 # limitations under the License.
 
 export BASE_COMMIT=$1
+
+RESULT=0
+
 ./.github/workflows/util/check.py header branch
 if [ $? -ne 0 ]; then
   ./.github/workflows/util/check.py header branch --fix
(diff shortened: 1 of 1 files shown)


Zouxxyy:
Test on my own mac, CC @zhouyuan @baibaichen @liuneng1994 for a looks, thanks.

zhouyuan (approved):
👍

PR #11792, opened 2026-03-19
[VL] Gluten-it: Rename column `c_last_review_date` to `c_last_review_date_sk`

zhztheplayer:
To comply with TPC-DS v2.7's schema specification:

https://www.tpc.org/tpc_documents_current_versions/pdf/tpc-ds_v2.7.0.pdf

The wrong name is causing inconsistency in some internal tests on Q30.

It's needed to add `--data-gen=always` the first time when a test is executed to regenerate the tables within new schema after this change.

### DIFF
--- a/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/ds/TpcdsTableLayout.scala
+++ b/tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/ds/TpcdsTableLayout.scala
@@ -359,7 +359,7 @@ object TpcdsTableLayout {
         StructField("c_birth_country", StringType),
         StructField("c_login", StringType),
         StructField("c_email_address", StringType),
-        StructField("c_last_review_date_sk", StringType)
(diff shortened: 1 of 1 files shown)


tools/gluten-it/common/src/main/scala/org/apache/gluten/integration/ds/TpcdsTableLayout.scala
```diff
@@ -359,7 +359,7 @@ object TpcdsTableLayout {
         StructField("c_birth_country", StringType),
         StructField("c_login", StringType),
         StructField("c_email_address", StringType),
-        StructField("c_last_review_date", StringType)
+        StructField("c_last_review_date_sk", StringType)
```
zhouyuan: IntType?

zhztheplayer: Yes, looks like it need to be long type to match d_date_sk

PR #11795, opened 2026-03-19
[CORE] Fix multi-key DPP support in ColumnarSubqueryBroadcastExec

yaooqinn:

Fix the `BuildSideRelation` path in `ColumnarSubqueryBroadcastExec` to handle multiple filtering keys instead of only using the first key (`indices(0)`).

### Problem

The existing code had a `TODO(): fixme` at line 92 that only used `indices(0)`, silently dropping extra keys when Spark 4.0 generates multi-key DPP subqueries (SPARK-46946):

```scala
val index = indices(0) // TODO(): fixme
```

This caused multi-key Dynamic Partition Pruning to be silently ineffective — only the first key was used for partition filtering.

### Fix

- For single-key DPP: behavior unchanged (direct key projection)
- For multi-key DPP: all keys are projected via `CreateStruct`, matching the `HashedRelation` path's multi-key support

This resolves potential DPP loss in queries like TPC-DS q23a/q23b with multi-column partition join keys.

## How was this patch tested?

Existing CI tests. The multi-key DPP path is exercised by Spark 4.0's DPP test suites.

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/ColumnarSubqueryBroadcastExec.scala
+++ b/gluten-substrait/src/main/scala/org/apache/spark/sql/execution/ColumnarSubqueryBroadcastExec.scala
@@ -89,14 +89,35 @@ case class ColumnarSubqueryBroadcastExec(
             val relation = child.executeBroadcast[Any]().value
             relation match {
               case b: BuildSideRelation =>
-                val index = indices(0) // TODO(): fixme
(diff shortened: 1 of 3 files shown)


yaooqinn:
Closing — after deeper investigation, Spark enforces `broadcastKeyIndices.size == 1` in `DynamicPruningSubquery` (even in 4.0 master). Multi-key DPP doesn't exist yet. The `TODO(): fixme` at `indices(0)` is aspirational, not a current bug. The fix targets dead code.

PR #11802, opened 2026-03-20
[GLUTEN-11766][VL] Implement the executeCollect() method in ColumnarCollectLimitExec

JkSelf:
Fix https://github.com/apache/gluten/issues/11766

## How was this patch tested?

Manual test.

Before this PR:

image

After this PR

image

No

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarToRowExecBase.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarToRowExecBase.scala
@@ -60,13 +60,4 @@ abstract class ColumnarToRowExecBase(child: SparkPlan)
   override def doExecute(): RDD[InternalRow] = {
     doExecuteInternal()
   }
-
-  override def executeCollect(): Array[InternalRow] = {
-    child match {
-      case l: ColumnarCollectLimitBaseExec =>
-        l.executeCollect()
-      case _ =>
-        super.executeCollect()
-    }
-  }
 }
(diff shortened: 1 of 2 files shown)


jinchengchenghh (approved):
Thanks!

gluten-substrait/src/main/scala/org/apache/gluten/execution/ColumnarToRowExecBase.scala
```diff
@@ -60,4 +60,13 @@ abstract class ColumnarToRowExecBase(child: SparkPlan)
   override def doExecute(): RDD[InternalRow] = {
     doExecuteInternal()
   }
+
+  override def executeCollect(): Array[InternalRow] = {
+    child match {
+      case l: ColumnarCollectLimitBaseExec =>
+        l.executeCollect()
```
jackylee-ch: Do we need to support this for ch backend?

JkSelf: @jackylee-ch No. Move this changes into VeloxColumnarToRowExec.

JkSelf:
> code looks good, could you please also add a test on this?

@zhouyuan https://github.com/apache/gluten/blob/main/gluten-ut/spark35/src/test/scala/org/apache/spark/sql/execution/GlutenSQLCollectLimitExecSuite.scala can cover this change.

JkSelf:
> Could you also support for ColumnarCollectTailExec? Thanks.

Yes. Will support ColumnarCollectTailExec after fixing https://github.com/apache/gluten/issues/11819?reload=1

PR #11806, opened 2026-03-22
[VL] Enhance VeloxS3 documentation with config details

FelixYBW:
List the aws s3 configuration details


FelixYBW:
aws_cpp_sdk features are much less than java version. We may need to use libhdfs similar way to call the java APIs.

@JkSelf @zhouyuan

PR #11807, opened 2026-03-22
[VL] Add executor pool config

FelixYBW:
Depends on https://github.com/facebookincubator/velox/pull/16879

respect spark config `spark.hadoop.fs.s3a.executor.capacity`


FelixYBW:
close this since we currently use sync GetObject function call.

We need to investigate if S3CrtClient can help here

Issue #11828, opened 2026-03-25
[VL] GPU code shouldn't be running on CPU node when cudf is enabled


marin-ma:
### Backend

VL (Velox)

### Bug description

Currently, Gluten uses the configuration `spark.gluten.sql.columnar.cudf=true` to enable GPU support. When running on a hybrid cluster, CPU nodes will fail if this configuration is set. We should disable the GPU code by checking whether the node has the CUDA runtime and available devices.

Another issue that needs to be resolved is that we currently pass the cudf configurations `spark.gluten.sql.columnar.cudf` and `spark.gluten.sql.columnar.backend.velox.cudf.enableTableScan` to the session configuration at runtime for the WholeStageResultIterator. However, these are immutable configurations. We should remove them from the runtime session configuration and instead set their values from the backend configuration.

### Spark version

None

### Relevant logs

```bash

```

PR #11830, opened 2026-03-25
[GLUTEN-11828][VL] Add cuda device detection

marin-ma:
Skip Cudf initialisation if no cuda device found.

Related issue: #11828

### DIFF
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -121,7 +121,7 @@ void veloxThreadManagerReleaser(ThreadManager* threadManager) {
   delete threadManager;
 }
 
-bool hasCudaRuntimeAndDevice() {
+bool hasCudaDevice() {
 #ifdef GLUTEN_ENABLE_GPU
   int count = 0;
   cudaError_t err = cudaGetDeviceCount(&count);
@@ -205,7 +205,7 @@ void VeloxBackend::init(
 
 #ifdef GLUTEN_ENABLE_GPU
   if (backendConf_->get<bool>(kCudfEnabled, kCudfEnabledDefault)) {
-    if (hasCudaRuntimeAndDevice()) {
+    if (hasCudaDevice()) {
       configureGpuTaskConcurrency(backendConf_->get<uint32_t>(kCudfConcurrentGpuTasks, kCudfConcurrentGpuTasksDefault));
       std::unordered_map<std::string, std::string> options = {
           {velox::cudf_velox::CudfConfig::kCudfEnabled, "true"},
@@ -223,7 +223,7 @@ void VeloxBackend::init(
       velox::cudf_velox::registerSparkFunctions("");
       velox::cudf_velox::registerSparkAggregateFunctions("");
     } else {
-      LOG(WARNING) << "No Cuda runtime and device found. Skip Cudf initialization.";
+      LOG(WARNING) << "No Cuda device found. Skip Cudf initialization.";
     }
   }
 #endif
--- a/cpp/velox/compute/VeloxBackend.cc
+++ b/cpp/velox/compute/VeloxBackend.cc
@@ -120,6 +120,16 @@ ThreadManager* veloxThreadManagerFactory(const std::string& kind, std::unique_pt
 void veloxThreadManagerReleaser(ThreadManager* threadManager) {
   delete threadManager;
 }
+
+bool hasCudaRuntimeAndDevice() {
+#ifdef GLUTEN_ENABLE_GPU
+  int count = 0;
+  cudaError_t err = cudaGetDeviceCount(&count);
+  return err == cudaSuccess && count > 0;
+#else
+  return false;
+#endif
+}
 } // namespace
 
 void VeloxBackend::init(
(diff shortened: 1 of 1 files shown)


marin-ma:
Verified using gpu build + spark.gluten.sql.columnar.cudf=true on cpu node. @jinchengchenghh Could you help to review? Thanks!

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -81,13 +92,20 @@ void veloxMemoryManagerReleaser(MemoryManager* memoryManager) {
...
-    const std::unordered_map<std::string, std::string>& sessionConf) {
-  auto* vmm = dynamic_cast<VeloxMemoryManager*>(memoryManager);
-  GLUTEN_CHECK(vmm != nullptr, "Not a Velox memory manager");
-  return new VeloxRuntime(kind, vmm, sessionConf);
+Runtime::Factory createVeloxRuntimeFactory(std::unordered_map<std::string, std::string> immutableConf) {
+  return [immutableConf = std::move(immutableConf)](
+             const std::string& kind,
+             MemoryManager* memoryManager,
+             const std::unordered_map<std::string, std::string>& sessionConf) -> Runtime* {
+    auto* vmm = dynamic_cast<VeloxMemoryManager*>(memoryManager);
```
jinchengchenghh: The configuration is not the static config, session level config can make user more flexible to decide which query is running on GPU, so I prefer these configs still be session config, the session level config can overwrite the static config

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -73,6 +73,17 @@ using namespace facebook;
 namespace gluten {
 
 namespace {
+
+bool hasCudaRuntimeAndDevice() {
+#ifdef GLUTEN_ENABLE_GPU
+  int count = 0;
+  cudaError_t err = cudaGetDeviceCount(&count);
```
jinchengchenghh: Looks like this code would execute on CPU node, but is the function executes by header, if not, the cuda library does not exist in CPU node, I'm not sure if it can run successfully. If you very it can run well on CPU node without CUDA environment, we may need to add a comment on it.

And the common way is to check if nvidia-smi command exists, if exists, we can check further.

cpp/velox/compute/VeloxBackend.cc
marin-ma: The cudf initialisation is only executed once in `VeloxBackend::init`. If the cudf related configurations are set at runtime by user, the initialisation code will be not be executed.

jinchengchenghh: At this case, we need to set the config in java side to static to forbid user setting.

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -105,7 +123,15 @@ void VeloxBackend::init(
 
   // Register factories.
   MemoryManager::registerFactory(kVeloxBackendKind, veloxMemoryManagerFactory, veloxMemoryManagerReleaser);
-  Runtime::registerFactory(kVeloxBackendKind, veloxRuntimeFactory, veloxRuntimeReleaser);
+
+  // Set immutable configurations from backend conf.
+  const bool enableCudf = backendConf_->get<bool>(kCudfEnabled, kCudfEnabledDefault) && hasCudaRuntimeAndDevice();
+  const bool enableCudfTableScan =
+      enableCudf && backendConf_->get<bool>(kCudfEnableTableScan, kCudfEnableTableScanDefault);
+  std::unordered_map<std::string, std::string> immutableConf = {
```
jinchengchenghh: We add it here, if detects it is cpu node, we should not set the enable cudf to true, could we add a jni call to detect it?
```
    val extraConfMap = mutable.Map(GlutenConfig.COLUMNAR_CUDF_ENABLED.key -> enableCudf.toString)
    if (!supportsValueStreamDynamicFilter) {
      extraConfMap(VeloxConfig.VALUE_STREAM_DYNAMIC_FILTER_ENABLED.key) = "false"
    }
    val extraConf = extraConfMap.asJava
    val transKernel = NativePlanEvaluator.create(BackendsApiManager.getBackendName, extraConf)
```

cpp/velox/compute/VeloxBackend.cc
marin-ma: If we execute the GPU build on a cpu node without CUDA Runtime installed, the process will fail early when loading libvelox.so and reporting the cuda library is missing.

cpp/velox/compute/VeloxBackend.cc
marin-ma: In the current implementation the detection is only executed in backend initialisation. If we set it here then the detection will be executed for every pipeline, which seems redundant.

cpp/velox/compute/VeloxBackend.cc
jinchengchenghh: This is a new problem, we should not require user to install CUDA in CPU node, the build pipeline may also need to be updated

cpp/velox/compute/VeloxBackend.cc
jinchengchenghh: Make senses!

jinchengchenghh (approved):

cpp/velox/compute/VeloxBackend.cc
jinchengchenghh: Please create an issue to track this, thanks!

jinchengchenghh: added https://github.com/apache/gluten/issues/11844

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -120,6 +120,16 @@ ThreadManager* veloxThreadManagerFactory(const std::string& kind, std::unique_pt
...
+
+bool hasCudaRuntimeAndDevice() {
+#ifdef GLUTEN_ENABLE_GPU
+  int count = 0;
+  cudaError_t err = cudaGetDeviceCount(&count);
+  return err == cudaSuccess && count > 0;
+#else
+  return false;
+#endif
+}
```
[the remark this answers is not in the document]
marin-ma: Renamed to `hasCudaDevice`

PR #11831, opened 2026-03-26
[VL] Pass row count to CudfVector in GPU batch serializer

acvictor:
The `deserialize(uint8_t* data, int32_t size)` method receives `size` as the length of the serialized byte array (set via `env->GetArrayLength(data)` in `JniWrapper.cc`) and this byte count was forwarded directly as the `vector_size_t size` parameter to `CudfVector`, This PR replaces `size` with `vb->numRows()`, which is the actual row count from the deserialized Velox batch.

## How was this patch tested?
Existing UTs.

No.


acvictor:
@jinchengchenghh can you please review this PR? Thank you!

PR #11832, opened 2026-03-26
[VL] Fix native `Union` result name

lifulong:

Fix native union result use column type name as column name, which lead to same data type column has same data, but is not right result. eg all string columns has same data value as the first string column

```const auto name = outRowType->childAt(colIdx)->name();```
result name is column type name
```const auto name = outRowType->nameOf(colIdx);```
result name is column name

with conf spark.gluten.sql.native.union=true and use sql like 
```
with deduplicated_data as (
  select col1, col2, col3, col4, col6
  from (
    select
      u.col1,
      u.col2,
      u.col3,
      u.col4,
      u.col6,
      row_number() over (partition by u.col2 order by u.col5 desc) as rn
    from (
      select col1, col2, col3, col4, 98 as col5, '' as col6 from union_src_a
      union all
      select col1, col2, col3, col4, 100 as col5, '' as col6 from union_src_b
    ) u
  ) t
  where t.rn = 1
)
select col1, col2, col3, col4
from deduplicated_data
where col1 != 'valueC'
union all
select col1, col2, col3, col4
from deduplicated_data
where col1 = 'valueC'
``` can reproduce the error

## How was this patch tested?

<!--
Describe how the changes were tested, if applicable.
Include new tests to validate the functionality, if necessary.
For UI-related changes, attach screenshots to demonstrate the updates.
-->
test at our produce env, and add unit test
yes, co-authored by cursor
<!--
If generative AI tooling has been used in the process of authoring this patch, please include the
phrase: 'Generated-by: ' followed by the name of the tool and its version.
If no, write 'No'.
Please refer to the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) for details.
-->

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/MiscOperatorSuite.scala
@@ -582,6 +582,55 @@ class MiscOperatorSuite extends VeloxWholeStageTransformerSuite with AdaptiveSpa
     }
   }
 
+  test("native union_all with two level union keeps distinct output columns") {
+    withTempView("union_src_a", "union_src_b", "union_src_c") {
+      Seq(
+        ("valueA", "value1", "value11", "value111"),
+        ("valueA", "value2", "value22", "value222")
+      ).toDF("col1", "col2", "col3", "col4")
+        .createOrReplaceTempView("union_src_a")
+      Seq(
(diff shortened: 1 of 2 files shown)


lifulong:
the error is clear, reproduce error result need complex union sql, maybe no need add unittest, i am not sure

lifulong:
@zhouyuan can you help review this pr?

PR #11834, opened 2026-03-26
[GLUTEN-11888] [VL] Remove the synchronized lock in VeloxBroadcastBuildSideCache

JkSelf:
VeloxBroadcastBuildSideCache is managed by Caffe, which is already thread-safe. So we don't need to add manual synchronization.

## How was this patch tested?
Existing tests

no

Related issue: #11888


jinchengchenghh:
Could you describe the existing synchronized mechanism in PR description?

PR #11835, opened 2026-03-26
[VL] Refine docker image build

zhouyuan:
This patch refines the docker file for gpu image.
- switch to use centos9
- install cuda-toolkit-13.1
- gcc-14 is used

The image size is reduced from ~27GB to ~13GB

The patch also fixed the gcc-12 installation on centos8

The patch also installed spark-41 to image

## How was this patch tested?
verified local docker build


philo-he (approved):
Thanks.

PR #11836, opened 2026-03-26
[MINOR][VL] Add cudf build docker file with JDK 17 and cuda 12.9

jinchengchenghh:
Build the docker image `apache/gluten:centos-9-jdk17-cuda12.9-cudf`, installed the cudf related libraries in system library path.


dev/docker/cudf/Dockerfile.centos-9-jdk17-cuda12.9-cudf
philo-he, line 1: Can we name it `Dockerfile.centos9-cuda12.9-cudf` which is compact and seems to be more readable?

.github/workflows/docker_image.yml
```diff
@@ -111,6 +111,37 @@ jobs:
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          build-args: JAVA_VERSION=17
+          file: dev/docker/cudf/Dockerfile.centos9-cudf-cuda-12-9
+          push: true
+          tags: ${{ env.DOCKERHUB_REPO }}:centos-9-jdk17-cudf-cuda-12-9
```
philo-he: Docker name suggestion: centos-9-jdk17-cuda12.9-cudf

.github/workflows/docker_image.yml
```diff
@@ -111,6 +111,37 @@ jobs:
           file: dev/docker/cudf/Dockerfile
           push: true
           tags: ${{ env.DOCKERHUB_REPO }}:centos-9-jdk8-cudf
+          
+  build-centos-9-jdk17-cudf-cuda-12-9:
```
philo-he: `build-centos-9-jdk17-cuda12.9-cudf:`

.github/workflows/docker_image.yml
jinchengchenghh: The docker name aligns with centos-9-jdk8-cudf, do you think I should also update the former?

jinchengchenghh: But for compatibility, I don't support to change the old name, then if the new image name should align with the former pattern? Looking for your suggestiion.

philo-he: @jinchengchenghh, maybe, we can just follow the suggested name pattern for new docker images, not changing the old ones. We don’t need to force consistency. Since the old images may become legacy, we can drop them in the future or defer a full update until we can guarantee compatibility. Just a minor suggestion. Thanks.

.github/workflows/docker_image.yml
jinchengchenghh: The name is invalid, so I rename to build-centos-9-jdk17-cuda12_9-cudf
```
The identifier 'build-centos-9-jdk17-cuda12.9-cudf' is invalid. IDs may only contain alphanumeric characters, '_', and '-'. IDs must start with a letter or '_' and and must be less than 100 characters.
```

philo-he (approved):
Thanks.

PR #11837, opened 2026-03-27
feat(velox): Support RESPECT NULLS for collect_list/collect_set

yaooqinn:

Add `ignoreNulls` parameter to `VeloxCollectList`/`VeloxCollectSet` to support Spark's RESPECT NULLS syntax (SPARK-55256). When `ignoreNulls=false`, null elements are included in the collected array.

### Changes:
- **`VeloxCollectSet`/`VeloxCollectList`**: Accept `ignoreNulls` parameter (default `true` for backward compatibility).
- **`CollectRewriteRule`**: Propagates `ignoreNulls` from Spark's `CollectList`/`CollectSet` using reflection, making it backward-compatible with Spark versions that don't have `ignoreNulls`.

### Related:
- Velox PR: https://github.com/facebookincubator/velox/pull/16933 (collect_list RESPECT NULLS)
- Velox PR: https://github.com/facebookincubator/velox/pull/16416 (collect_set RESPECT NULLS, merged)
- Spark JIRA: SPARK-55256

## How was this patch tested?
CI validation. The default behavior (`ignoreNulls=true`) is unchanged, ensuring backward compatibility.

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -284,14 +284,30 @@ std::string SubstraitToVeloxPlanConverter::toAggregationFunctionName(
         // The merge_extract function is registered without suffix.
         return functionName;
       }
-      // The merge_extract function must be registered with suffix based on result type.
-      functionName += ("_" + companionFunctionSuffix(resultType));
-      signatures = exec::getAggregateFunctionSignatures(functionName);
-      VELOX_CHECK(
-          signatures.has_value() && signatures.value().size() > 0,
+      // The merge_extract function must be registered with suffix based on
+      // result type. First try exact concrete type suffix.
+      auto suffixedName =
+          functionName + "_" + companionFunctionSuffix(resultType);
+      signatures = exec::getAggregateFunctionSignatures(suffixedName);
+      if (signatures.has_value() && signatures.value().size() > 0) {
+        return suffixedName;
+      }
+      // When companion functions are registered with generic type variables
+      // (e.g., "collect_set_merge_extract_array_T"), look up companion
+      // function names from the aggregate function registry.
+      auto companionSigs = exec::getCompanionFunctionSignatures(baseName);
+      if (companionSigs.has_value()) {
+        for (const auto& entry : companionSigs->mergeExtract) {
+          auto entrySigs =
+              exec::getAggregateFunctionSignatures(entry.functionName);
+          if (entrySigs.has_value() && entrySigs.value().size() > 0) {
+            return entry.functionName;
+          }
+        }
+      }
+      VELOX_FAIL(
           "Cannot find function signature for {} in final aggregation step.",
-          functionName);
-      return functionName;
+          suffixedName);
     }
     case core::AggregationNode::Step::kIntermediate:
       suffix = "_merge";
--- a/backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/expression/aggregate/VeloxCollect.scala
@@ -21,13 +21,13 @@ import org.apache.spark.sql.catalyst.expressions.aggregate.DeclarativeAggregate
 import org.apache.spark.sql.catalyst.trees.UnaryLike
 import org.apache.spark.sql.types.{ArrayType, DataType}
 
-abstract class VeloxCollect(child: Expression)
+abstract class VeloxCollect(child: Expression, val ignoreNulls: Boolean)
   extends DeclarativeAggregate
   with UnaryLike[Expression] {
 
   protected lazy val buffer: AttributeReference = AttributeReference("buffer", dataType)()
 
-  override def dataType: DataType = ArrayType(child.dataType, false)
+  override def dataType: DataType = ArrayType(child.dataType, !ignoreNulls)
 
   override def nullable: Boolean = false
 
@@ -35,12 +35,17 @@ abstract class VeloxCollect(child: Expression)
 
   override lazy val initialValues: Seq[Expression] = Seq(Literal.create(Array(), dataType))
 
-  override lazy val updateExpressions: Seq[Expression] = Seq(
-    If(
-      IsNull(child),
-      buffer,
-      Concat(Seq(buffer, CreateArray(Seq(child), useStringTypeWhenEmpty = false))))
-  )
+  override lazy val updateExpressions: Seq[Expression] = {
+    val append = if (ignoreNulls) {
+      If(
+        IsNull(child),
+        buffer,
+        Concat(Seq(buffer, CreateArray(Seq(child), useStringTypeWhenEmpty = false))))
+    } else {
+      Concat(Seq(buffer, CreateArray(Seq(child), useStringTypeWhenEmpty = false)))
+    }
+    Seq(append)
+  }
 
   override lazy val mergeExpressions: Seq[Expression] = Seq(
     Concat(Seq(buffer.left, buffer.right))
@@ -49,7 +54,8 @@ abstract class VeloxCollect(child: Expression)
   override def defaultResult: Option[Literal] = Option(Literal.create(Array(), dataType))
 }
 
-case class VeloxCollectSet(child: Expression) extends VeloxCollect(child) {
+case class VeloxCollectSet(child: Expression, override val ignoreNulls: Boolean = true)
+  extends VeloxCollect(child, ignoreNulls) {
 
   override lazy val evaluateExpression: Expression =
     ArrayDistinct(buffer)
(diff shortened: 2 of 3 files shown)


zhouyuan:
@yaooqinn thanks, could you please also verify on a new veox branch with patches for collect_list/set - in the current branch the collect_set commit is actually reverted https://github.com/IBM/velox/commits/dft-2026_03_24/

yaooqinn:
Attempted local verification with Velox patches (collect_set from upstream main + collect_list from PR #16933). The Velox native lib (libvelox.a) builds successfully with both patches. However, full Gluten build cannot complete locally due to pre-existing Scala version mismatch in cached Maven artifacts (unrelated to this PR).

The VeloxCollect changes are purely Scala/JVM-side (DeclarativeAggregate), so they're independent of the Velox native collect_set/collect_list functions. CI will provide the definitive verification.

@zhouyuan The native Velox patches for collect_set/collect_list add `Aggregate::setConstantInputs()` for the boolean ignoreNulls parameter, but since VeloxCollect uses Scala-based array operations, it doesn't need those native changes. The Gluten PR works with any Velox version.

zhouyuan:
@yaooqinn not really
the patch (https://github.com/facebookincubator/velox/pull/16416/changes ) changed the signature of collect_set functions, which requires a modification on substrait layer

I just made a test with your patch + velox changes, the spark unit tests are failing

https://github.com/apache/gluten/actions/runs/23652189825/job/68903134168?pr=11712

```
E20260327 15:24:19.552174  1769 Exceptions.h:87] Line: /work/cpp/velox/substrait/SubstraitToVeloxPlan.cc:290, Function:toAggregationFunctionName, Expression: signatures.has_value() && signatures.value().size() > 0 Cannot find function signature for collect_set_merge_extract_array_row_VARCHAR_BIGINT_BIGINT_endrow in final aggregation step., Source: RUNTIME, ErrorCode: INVALID_STATE
15:24:19.555 WARN org.apache.spark.sql.execution.GlutenFallbackReporter: Validation failed for plan: SortAggregate[QueryId=405], due to: 
 - Native validation failed: 
   |- Validation failed due to exception caught at file:SubstraitToVeloxPlanValidator.cc line:1450 function:validate, thrown from file:SubstraitToVeloxPlan.cc line:290 function:toAggregationFunctionName, reason:Cannot find function signature for collect_set_merge_extract_array_row_VARCHAR_BIGINT_BIGINT_endrow in final aggregation step.
```

yaooqinn:
## Root Cause Analysis

Thanks @zhouyuan for catching this! I traced the issue:

### What changed

**Before PR #16416**: `collect_set` had **1 signature** → `hasSameIntermediateTypesAcrossSignatures()` returns `false` → companion function registered as `collect_set_merge_extract` (no suffix).

**After PR #16416**: `collect_set` has **2 signatures** (1-arg + 2-arg with boolean), both with `intermediateType("array(T)")` → `hasSameIntermediateTypesAcrossSignatures()` returns `true` → companion function registered as `collect_set_merge_extract_array_T` (**with suffix**).

### The mismatch

Gluten's `SubstraitToVeloxPlan.cc::toAggregationFunctionName()` (line 280-294):
1. First tries `collect_set_merge_extract` → **not found** (was registered with suffix now)
2. Falls through, constructs suffix from concrete result type: `collect_set_merge_extract_array_row_VARCHAR_BIGINT_BIGINT_endrow`
3. Looks up that name → **not found** (registered as generic `collect_set_merge_extract_array_T`)
4. **Throws**: `Cannot find function signature`

### Fix options

1. **In Velox**: Ensure `hasSameIntermediateTypesAcrossSignatures` returns `false` by keeping companion functions registered without suffix. Could use `ignoreDuplicates` or separate registration for the 2-arg signature.
2. **In Gluten Substrait C++**: Update `toAggregationFunctionName` to try the generic (no-suffix) companion name via `getAggregateFunctionSignatures(baseName + "_merge_extract")` first with type resolution, similar to how Velox's own planner resolves generic companion functions.

I'll fix this in the Velox PR #16416 (or a follow-up) since the root cause is the signature change creating a suffix-based companion registration.

rui-mo:
@yaooqinn I assume we should adapt to the signature change in Gluten (see `toAggregationFunctionName` which intends to resolve the name incompatibility caused by `hasSameIntermediateTypesAcrossSignatures`) rather than making additional changes in Velox.

yaooqinn:
Pushed fix: Updated `SubstraitToVeloxPlan.cc::toAggregationFunctionName()` to handle generic-typed companion functions.

**Change**: After failing the concrete type suffix lookup (e.g., `collect_set_merge_extract_array_row_VARCHAR_...`), now falls back to `getCompanionFunctionSignatures(baseName)` API to discover the actual companion function names (e.g., `collect_set_merge_extract_array_T`).

The file compiles cleanly against the existing Velox EP. @zhouyuan could you verify with your test setup?

yaooqinn:
## Verified ✅

Tested locally against `IBM/velox` `dft-2026_03_24` branch (which reverts PR #16416). All collect_set/collect_list tests pass:

```
Tests run: 16, Failures: 0, Errors: 0, Skipped: 0
- test collect_set ✅
- test collect_set/collect_list with null ✅
- test fallback collect_set/collect_list with null (Offload, FallbackPartial, FallbackFinal, FallbackAll) ✅
- collect_list null inputs ✅
- test collect_list with ordering ✅
BUILD SUCCESS
```

The C++ Substrait fix (`toAggregationFunctionName` companion function fallback) compiles cleanly. It will handle both cases:
1. **Current IBM/velox** (without PR #16416): companion registered without suffix → found at first try
2. **Future with PR #16416**: companion registered with generic suffix → found via `getCompanionFunctionSignatures()` fallback

@zhouyuan

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -284,14 +284,30 @@ std::string SubstraitToVeloxPlanConverter::toAggregationFunctionName(
...
+      // When companion functions are registered with generic type variables
+      // (e.g., "collect_set_merge_extract_array_T"), look up companion
+      // function names from the aggregate function registry.
+      auto companionSigs = exec::getCompanionFunctionSignatures(baseName);
+      if (companionSigs.has_value()) {
+        for (const auto& entry : companionSigs->mergeExtract) {
+          auto entrySigs =
+              exec::getAggregateFunctionSignatures(entry.functionName);
+          if (entrySigs.has_value() && entrySigs.value().size() > 0) {
+            return entry.functionName;
```
rui-mo, line 304: When two merge-extract functions share the same intermediate type, this approach cannot guarantee that the correct function will be returned. We need to fix the `companionFunctionSuffix`, which needs to be fully compatible with Velox’s `std::string toSuffixString(const TypeSignature& type)` function. Currently, we are encountering fallback issue because Gluten appears to generate a function name that differs from the one Velox uses. It looks like Gluten is generating `collect_set_merge_extract_array_row_VARCHAR_BIGINT_BIGINT_endrow` while Velox uses `collect_set_merge_extract_array_T`.

rui-mo (commented):
Also consider adding test in https://github.com/apache/gluten/blob/main/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala to ensure the two functions can be offloaded to native.

zhouyuan:
@yaooqinn thanks for the efforts, looks like dont not work:
https://github.com/apache/gluten/actions/runs/23664885562/job/68946268175?pr=11712

The IBM/velox dft-2026_03_24 actually reverted the collect_set patch, could you please have a try with this branch:
```
VELOX_REPO=https://github.com/zhouyuan/velox.git
VELOX_BRANCH=dft-2026_03_24
```

yaooqinn:
## Verified with `zhouyuan/velox` `dft-2026_03_24` ✅

Found and fixed a **bug in Velox PR #16416**: `ignoreNulls_` default was `false` (RESPECT NULLS) instead of `true` (IGNORE NULLS). When native collect_set uses the 1-arg signature (no boolean constant), `setConstantInputs` doesn't set the flag, so it defaults to including nulls — breaking backward compatibility.

**Fix**: `bool ignoreNulls_{false}` → `bool ignoreNulls_{true}` in `CollectSetAggregate.cpp:218`

After this fix, all 16 tests pass:
```
- test collect_set ✅
- test collect_set/collect_list with null ✅
- test fallback collect_set/collect_list with null (4 variants) ✅
- collect_list null inputs ✅
- test collect_list with ordering ✅
BUILD SUCCESS (01:47 min)
```

I'll send a follow-up fix to Velox upstream for this default value bug.

@zhouyuan

zhouyuan:
@yaooqinn @rui-mo to make the rebase work I have included this patch in https://github.com/apache/gluten/pull/11891
CI is passing now, please followup if any further patches required

yaooqinn:
Closing this — it's been sitting since April with merge conflicts, and I'm not in a position to drive it forward right now. Happy to reopen later if needed, or someone else is welcome to pick up the work.

Thanks @rui-mo for the earlier review.

PR #11838, opened 2026-03-27
[MINOR] Expose Gluten and component build information to SparkConf

jiangjiangtian:
Currently, Gluten build information is not injected into the Spark configuration, which causes difficulties in certain troubleshooting and tracking scenarios. For example:
1. To find out which specific Git revision a job is running on, we have to grep the driver's syslog. Since driver logs are deleted much earlier than Spark History logs, this information becomes unavailable once the syslog is purged.
2. We have a tracking table that stores all configs of our running jobs. Without the build metadata in the config, it's hard to tell when a job starts running on a newer version of Gluten.
Therefore, this PR sets the Gluten build information as part of the Spark configurations. With this change, we can simply check the build revision by looking at the Spark UI:
image


philo-he (approved):
Looks good. Thanks.

Issue #11849, opened 2026-03-30
[CH] var_samp returns NaN instead of NULL when effective row count < 2


exmy:
### Backend

CH (ClickHouse)

### Bug description

select var_samp(id) from range(1);
     select covar_samp(id, id) from range(1);

    Expected NULL

    Actual NaN

### Spark version

None

### Relevant logs

```bash

```

PR #11850, opened 2026-03-30
[GLUTEN-11849][CH] Fix diff for var_samp returns NaN instead of NULL when effective row count < 2

exmy:
Fix diff for var_samp returns NaN instead of NULL when effective row count < 2

## How was this patch tested?
Add ut

co-authored with copilot cli

Related issue: #11849


zzcclp:

PR #11853, opened 2026-03-30
[CORE][UI] Fix fallback info for V2 writes and align plan with Spark SQL tab

Zouxxyy:

Fix Gluten UI not displaying fallback info for V2 writes (e.g., Iceberg INSERT with sort order), and align the plan display with Spark SQL tab (consistent operator IDs, complete plan tree).

**Root causes:**
1. Plan traversal skips `CommandResultExec` and `V2CommandExec`, missing V2 write fallback reasons.
2. In AQE mode, per-stage events have partial plans with inconsistent operator IDs; last stage overwrites earlier ones.

**Fix:** Add `CommandResultExec` recursion in `GlutenExplainUtils`/`GlutenImplicits`/`GlutenFallbackReporter`. Add `GlutenQueryExecutionListener` that generates complete fallback data from the fully materialized `QueryExecution` after execution ends, posted as event for both live UI and History Server.

We also considered a listener-side merge approach (accumulating per-stage events), but it cannot produce consistent operator IDs or a complete plan tree.

## How was this patch tested?

Added test in `VeloxIcebergSuite`. Verified on History Server.

before

image

after

image

Generated-by: Kiro (Claude Opus 4.6), co-authored with Zouxxyy through iterative design, code review, and testing.

### DIFF
--- a/gluten-ut/spark33/src/test/scala/org/apache/spark/sql/gluten/GlutenFallbackSuite.scala
+++ b/gluten-ut/spark33/src/test/scala/org/apache/spark/sql/gluten/GlutenFallbackSuite.scala
@@ -233,7 +233,7 @@ class GlutenFallbackSuite extends GlutenSQLTestsTrait with AdaptiveSparkPlanHelp
             events.count(
               _.fallbackNodeToReason.values.toSet.exists(_.contains(
                 "Could not find a valid substrait mapping name for max"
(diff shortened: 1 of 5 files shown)


Zouxxyy:
CC @zhztheplayer @JkSelf @jinchengchenghh for a look, thanks

PR #11855, opened 2026-03-31
[VL] Fix StdMemoryAllocator::allocateZeroFilled byte accounting

LuciferYang:

Fix incorrect byte accounting in `StdMemoryAllocator::allocateZeroFilled`. The method calls `std::calloc(nmemb, size)` which allocates `nmemb * size` bytes, but `bytes_` was only incremented by `size` instead of `nmemb * size`.

This is consistent with `ListenableMemoryAllocator::allocateZeroFilled`, which already correctly tracks `size * nmemb`.

## How was this patch tested?

Added a unit test `StdMemoryAllocator.allocateZeroFilledAccounting` that verifies `allocateZeroFilled(10, 64, &buf)` correctly tracks 640 bytes.

Generated-by: Claude Code

### DIFF
--- a/cpp/core/memory/MemoryAllocator.cc
+++ b/cpp/core/memory/MemoryAllocator.cc
@@ -18,6 +18,8 @@
 #include "MemoryAllocator.h"
 #include "utils/Macros.h"
 
+#include <limits>
+
 namespace gluten {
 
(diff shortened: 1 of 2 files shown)


LuciferYang:
Thank you @jackylee-ch and @zhztheplayer

PR #11868, opened 2026-04-02
[VL] Enable enhanced tests for spark 4.0 & fix failures

infvg:
This PR enables enhanced test for spark 4.0 & resolves sql queries in iceberg due to the new metadata columns.

### DIFF
--- a/backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala
+++ b/backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala
@@ -24,14 +24,23 @@ import org.apache.spark.sql.types.StructType
 import org.apache.iceberg.spark.source.IcebergWriteUtil
 import org.apache.iceberg.types.TypeUtil
 
+import scala.collection.JavaConverters._
+
 abstract class AbstractIcebergWriteExec extends IcebergWriteExec {
 
   // the writer factory works for both batch and streaming
   private def createIcebergDataWriteFactory(schema: StructType): IcebergDataWriteFactory = {
     val writeSchema = IcebergWriteUtil.getWriteSchema(write)
     val nestedField = TypeUtil.visit(writeSchema, new IcebergNestedFieldVisitor)
+    // Filter out metadata columns from the Spark output schema and reorder to match Iceberg schema
+    // Spark 4.0 may include metadata columns in the output schema during UPDATE operations,
+    // but these should not be written to the Iceberg table
+    val writeFieldNames = writeSchema.columns().asScala.map(_.name()).toSet
+    val filteredSchema = StructType(
+      schema.fields.filter(field => writeFieldNames.contains(field.name))
+    )
     IcebergDataWriteFactory(
-      schema,
+      filteredSchema,
       getFileFormat(IcebergWriteUtil.getFileFormat(write)),
       IcebergWriteUtil.getDirectory(write),
       getCodec,
--- a/cpp/velox/compute/iceberg/IcebergWriter.cc
+++ b/cpp/velox/compute/iceberg/IcebergWriter.cc
@@ -154,10 +154,10 @@ std::shared_ptr<IcebergInsertTableHandle> createIcebergInsertTableHandle(
               nestedField.children[i]));
     }
   }
-  
+
   auto fileNameGenerator = std::make_shared<const GlutenIcebergFileNameGenerator>(
       partitionId, taskId, operationId, fileFormat);
-  
+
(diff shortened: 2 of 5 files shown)


cpp/velox/compute/iceberg/IcebergWriter.cc
```diff
@@ -212,7 +221,47 @@ IcebergWriter::IcebergWriter(
 }
 
 void IcebergWriter::write(const VeloxColumnarBatch& batch) {
-  dataSink_->appendData(batch.getRowVector());
+  auto inputRowVector = batch.getRowVector();
+  auto inputRowType = asRowType(inputRowVector->type());
+  
+  // Filter columns to match the expected schema (rowType_)
```
jinchengchenghh: The metadata columns should append in the end or beginning of the schema, and the number of the columns should be fixed value, so could we simplify the logic?

jinchengchenghh: And the metadata column name is the specific name, we only need to match the pattern to decide it is metadata column, could you show the example schema to let us understand this issue.

infvg: I think we can simplify it by using fieldIds. Field ID > Integer.MAX - 200 are reserved for metadata columns: 
https://iceberg.apache.org/spec/#reserved-field-ids

infvg: We can just splice the columns and remove any additional columns that appear at the end so we don't have to add any loops

jinchengchenghh: Yes, that's what I want

backends-velox/src-iceberg/main/scala/org/apache/gluten/execution/AbstractIcebergWriteExec.scala
```diff
@@ -24,14 +24,23 @@ import org.apache.spark.sql.types.StructType
...
 abstract class AbstractIcebergWriteExec extends IcebergWriteExec {
 
   // the writer factory works for both batch and streaming
   private def createIcebergDataWriteFactory(schema: StructType): IcebergDataWriteFactory = {
     val writeSchema = IcebergWriteUtil.getWriteSchema(write)
     val nestedField = TypeUtil.visit(writeSchema, new IcebergNestedFieldVisitor)
+    // Filter out metadata columns from the Spark output schema and reorder to match Iceberg schema
+    // Spark 4.0 may include metadata columns in the output schema during UPDATE operations,
+    // but these should not be written to the Iceberg table
+    val schemaFieldMap = schema.fields.map(f => f.name -> f).toMap
```
jinchengchenghh: You could use Intellij to debug here, see the writeSchema and schema: StructType difference, also use slice to take some of the columns

zhouyuan:
@infvg Thanks for the fix. Please fix the CI.

cpp/velox/compute/iceberg/IcebergWriter.cc
```diff
@@ -212,7 +212,16 @@ IcebergWriter::IcebergWriter(
 }
 
 void IcebergWriter::write(const VeloxColumnarBatch& batch) {
-  dataSink_->appendData(batch.getRowVector());
+  auto inputRowVector = batch.getRowVector();
+  
+  auto outputRowVector = std::make_shared<RowVector>(
```
jinchengchenghh: Why you need it? Just set the rowType_?

jinchengchenghh (approved):

cpp/velox/compute/iceberg/IcebergWriter.cc
```diff
@@ -212,7 +212,28 @@ IcebergWriter::IcebergWriter(
...
 void IcebergWriter::write(const VeloxColumnarBatch& batch) {
-  dataSink_->appendData(batch.getRowVector());
+  auto inputRowVector = batch.getRowVector();
+  auto inputRowType = asRowType(inputRowVector->type());
+
+  if (inputRowType->size() != rowType_->size()) {
+    std::vector<VectorPtr> dataColumns;
+    dataColumns.reserve(rowType_->size());
+
+    for (size_t i = 0; i < rowType_->size(); ++i) {
```
jinchengchenghh: Use like this to simplify the logic std::vector<int> sub(v.begin() + 1, v.begin() + 4);

infvg: Done

cpp/velox/compute/iceberg/IcebergWriter.cc
infvg: The first column is a metadata column __row_operation and the other metadata columns are appended at the end, so I sliced it from 1, i columns to extract the regular columns.

PR #11869, opened 2026-04-02
[GLUTEN-11862][VL] Work around GMT session timezone validation failure on macOS

### LINKED ISSUE #11862: [VL] Native validation fails when Spark session timezone is GMT
### Backend

VL (Velox)

### Bug description

Expected behavior:

Spark 3.5 queries on Gluten `main` should be able to use the Velox backend even when the effective Spark session timezone is `GMT`. Gluten should either pass a native-compatible timezone string to Velox or normalize the Spark value before native validation.

Actual behavior:

Gluten forwards the effective Spark session timezone into native query config. When that value is `GMT`, native validation fails before execution starts:

```text
Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: session 'session_timezone' set with invalid value 'GMT'
Expression: tz::getTimeZoneID(*tz, false) != -1
Function: validateConfig
File: .../ep/build-velox/build/velox_ep/velox/core/QueryConfig.cpp
Line: 44
```

As a result, the query never reaches the actual Delta/Iceberg/Velox execution path. Native offload is rejected at validation time because of the session timezone value alone.

This was reproduced on a clean detached worktree without the local timezone normalization patch, after a fresh native rebuild using Gluten's pinned Velox dependency. The failure is not Delta-only; it also reproduces in a non-Delta Velox suite, which indicates this is a general Spark-to-native integration bug in the Velox backend path rather than a Delta-specific semantic issue.

Reproduction summary:

1. Create a clean detached worktree at `f350a440dbfa3a3412f0aed48a89fdac0b16ad48`.
2. Rebuild native dependencies from the clean checkout:

   ```bash
   ./dev/builddeps-veloxbe.sh --run_setup_script=OFF --build_arrow=OFF --build_tests=ON --spark_version=3.5
   ```

3. Run the backend tests from the same clean checkout:

   ```bash
   ./build/mvn \
     -s /path/to/dev/maven-public-settings.xml \
     -Dmaven.repo.local=/path/to/.run-scala-test-cache/m2 \
     -pl backends-velox \
     -am \
     -Pbackends-velox,delta,iceberg,spark-3.5 \
     -Dtest=OptimizedWritesSuite \
     -Dsurefire.failIfNoSpecifiedTests=false \
     test
   ```

Observed failures include:

- `org.apache.spark.sql.delta.DeltaColumnDefaultsInsertSuite`
- `org.apache.gluten.execution.VeloxDeltaSuite`
- `org.apache.gluten.execution.VeloxTPCHMiscSuite`

Example clean repro reports:

- `backends-velox/target/surefire-reports/TEST-org.apache.spark.sql.delta.DeltaColumnDefaultsInsertSuite.xml`
- `backends-velox/target/surefire-reports/TEST-org.apache.gluten.execution.VeloxDeltaSuite.xml`
- `backends-velox/target/surefire-reports/TEST-org.apache.gluten.execution.VeloxTPCHMiscSuite.xml`

Why this matters:

- It blocks native validation before query execution.
- It prevents meaningful verification of Delta/Iceberg behavior because suites fail on config setup first.
- It hides deeper runtime issues behind a config gate.
- It affects the current clean `main` build path.

Why this appears to be a Gluten-side issue:

- Spark 3.5 can accept and surface `GMT`-family timezone IDs as an effective session timezone.
- Gluten forwards that session timezone into native config.
- The current clean Gluten product path fails before execution on that value.
- The practical compatibility boundary between Spark session config and native Velox config is in Gluten.

Candidate fix:

Normalize Spark `GMT` / `GMT+/-offset` session timezone values to native-safe `UTC` forms at the Gluten/native boundary before native validation.

### Gluten version

main branch

### Spark version

Spark-3.5.x

### Spark configurations

No explicit `spark.sql.session.timeZone` override was set in the reproducer.

The effective session timezone seen by native validation was `GMT`.

### System information

````
./dev/info.sh: line 47: lscpu: command not found

Gluten Version: 1.7.0-SNAPSHOT
Commit: f350a440dbfa3a3412f0aed48a89fdac0b16ad48
CMake Version: 4.2.3
System: Darwin-25.3.0
Arch: arm64
CPU Name: 
C++ Compiler: /usr/bin/c++
C++ Compiler Version: 17.0.0.17000013
17.0.0.17000013
C Compiler: /usr/bin/cc
C Compiler Version: 17.0.0.17000013
17.0.0.17000013
CMake Prefix Path: /Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr;/opt/homebrew;/usr/local;/usr;/;/opt/homebrew;/usr/local;/usr/X11R6;/usr/pkg;/opt;/sw;/opt/local


### Relevant logs

```bash
Exception: VeloxUserError
Error Source: USER
Error Code: INVALID_ARGUMENT
Reason: session 'session_timezone' set with invalid value 'GMT'
Retriable: False
Expression: tz::getTimeZoneID(*tz, false) != -1
Function: validateConfig
File: /Users/malinjawi/Documents/GitHub/GlutenVelox/incubator-gluten-verify-clean/ep/build-velox/build/velox_ep/velox/core/QueryConfig.cpp
Line: 44
```

malinjawi:

Fixes #11862.

This PR works around a macOS-specific native validation failure in the Velox backend when the effective Spark session timezone is `GMT`.

Based on the discussion in #11862, the underlying behavior is likely related to upstream Folly/Velox timezone handling on macOS rather than a general Spark semantics problem. This patch keeps the workaround on the Gluten side so Spark session timezone values do not block native validation and execution in the Velox path.

The changes in this PR are:

- normalize `GMT` and `GMT+/-offset` session timezone values to `UTC`-equivalent forms before passing them into native Velox query config
- use the runtime session timezone in `SubstraitToVeloxPlanValidator` instead of hardcoding `GMT`
- apply the same normalization when setting the Velox writer timezone config
- add a regression test covering execution with `spark.sql.session.timeZone=GMT`

## How was this patch tested?

- reproduced the failure on a clean unpatched checkout on macOS after a fresh native rebuild
- verified the failure is:
  - `session 'session_timezone' set with invalid value 'GMT'`
- rebuilt native Gluten/Velox with this patch
- ran the new targeted regression test in `MiscOperatorSuite`:
  - `cast date to timestamp with GMT session timezone`
- the targeted regression test passes with this patch

Generated-by: IBM BOB

### DIFF
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHConfig.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHConfig.scala
@@ -102,6 +102,28 @@ object CHConfig extends ConfigRegistry {
       .doc("Enable local cache for CH backend.")
       .booleanConf
       .createWithDefault(false)
+
+  val EXTENDED_COLUMNAR_TRANSFORM_RULES =
+    buildConf("spark.gluten.sql.columnar.extended.columnar.transform.rules")
+      .internal()
+      .withAlternative("spark.gluten.sql.columnar.extended.columnar.pre.rules")
+      .doc("A comma-separated list of classes for the extended columnar transform rules.")
+      .stringConf
+      .createWithDefaultString("")
+
+  val EXTENDED_COLUMNAR_POST_RULES =
+    buildConf("spark.gluten.sql.columnar.extended.columnar.post.rules")
+      .internal()
+      .doc("A comma-separated list of classes for the extended columnar post rules.")
+      .stringConf
+      .createWithDefaultString("")
+
+  val EXTENDED_EXPRESSION_TRAN_CONF =
+    buildConf("spark.gluten.sql.columnar.extended.expressions.transformer")
+      .internal()
+      .doc("A class for the extended expressions transformer.")
+      .stringConf
+      .createWithDefaultString("")
 }
 
 class CHConfig(conf: SQLConf) extends GlutenConfig(conf) {
@@ -128,6 +150,14 @@ class CHConfig(conf: SQLConf) extends GlutenConfig(conf) {
     getConf(ENABLE_CH_REWRITE_DATE_CONVERSION)
 
   def enableGlutenLocalFileCache: Boolean = getConf(ENABLE_GLUTEN_LOCAL_FILE_CACHE)
+
+  // A comma-separated list of classes for the extended columnar pre rules
+  def extendedColumnarTransformRules: String = getConf(EXTENDED_COLUMNAR_TRANSFORM_RULES)
+
+  // A comma-separated list of classes for the extended columnar post rules
+  def extendedColumnarPostRules: String = getConf(EXTENDED_COLUMNAR_POST_RULES)
+
+  def extendedExpressionTransformer: String = getConf(EXTENDED_EXPRESSION_TRAN_CONF)
 }
 
 object GlutenObjectStorageConfig {
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHListenerApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHListenerApi.scala
@@ -49,7 +49,7 @@ class CHListenerApi extends ListenerApi with Logging {
     initialize(pc.conf, isDriver = true)
 
     val expressionExtensionTransformer = ExpressionUtil.extendedExpressionTransformer(
-      pc.conf.get(GlutenConfig.EXTENDED_EXPRESSION_TRAN_CONF)
+      pc.conf.get(CHConfig.EXTENDED_EXPRESSION_TRAN_CONF)
     )
     if (expressionExtensionTransformer != null) {
       ExpressionExtensionTrait.registerExpressionExtension(expressionExtensionTransformer)
--- a/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHRuleApi.scala
+++ b/backends-clickhouse/src/main/scala/org/apache/gluten/backendsapi/clickhouse/CHRuleApi.scala
@@ -130,7 +130,7 @@ object CHRuleApi {
       c =>
         intercept(
           SparkPlanRules.extendedColumnarRule(
-            new GlutenConfig(c.sqlConf).extendedColumnarTransformRules)(c.session)))
+            new CHConfig(c.sqlConf).extendedColumnarTransformRules)(c.session)))
     injector.injectPostTransform(_ => CollectLimitTransformerRule())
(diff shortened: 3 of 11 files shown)


cpp/velox/compute/WholeStageResultIterator.cc
```diff
@@ -358,14 +357,15 @@ void WholeStageResultIterator::constructPartitionColumns(
 }
 
 void WholeStageResultIterator::addIteratorSplits(const std::vector<std::shared_ptr<ResultIterator>>& inputIterators) {
-  GLUTEN_CHECK(!allSplitsAdded_, "Method addIteratorSplits should not be called since all splits has been added to the Velox task.");
+  GLUTEN_CHECK(
```
jinchengchenghh: Please remove unrelated changes

cpp/velox/substrait/SubstraitToVeloxPlanValidator.h
```diff
@@ -29,9 +30,15 @@ namespace gluten {
...
  public:
-  SubstraitToVeloxPlanValidator(memory::MemoryPool* pool) {
+  explicit SubstraitToVeloxPlanValidator(
+      memory::MemoryPool* pool,
+      const std::unordered_map<std::string, std::string>& confMap = {}) {
+    const auto it = confMap.find(kSessionTimezone);
+    const auto sessionTimezone =
+        normalizeSessionTimezone(it == confMap.end() ? std::string_view("UTC") : std::string_view(it->second));
     std::unordered_map<std::string, std::string> configs{
-        {velox::core::QueryConfig::kSparkPartitionId, "0"}, {velox::core::QueryConfig::kSessionTimezone, "GMT"}};
```
jinchengchenghh, line 34: update the GMT to other value is enough

malinjawi: Thanks, addressed. I updated `kSessionTimezone` in to `UTC`.

cpp/velox/compute/WholeStageResultIterator.cc
malinjawi: Hey @jinchengchenghh removed dist change

zhouyuan:
related: https://github.com/facebookincubator/velox/pull/17010

cpp/core/config/GlutenConfig.cc
```diff
@@ -49,6 +49,17 @@ parseConfMap(JNIEnv* env, const uint8_t* planData, const int32_t planDataLength)
   return sparkConfs;
 }
 
+std::string normalizeSessionTimezone(std::string_view sessionTimezone) {
```
rui-mo, line 52: Is it more appropriate to move the normalization step to Scala before sending data to native, so that the timezone received by C++ is always correct and we don’t need to apply normalization in every caller of the session timezone?

PR #11882, opened 2026-04-07
[VL] Rework option `spark.gluten.sql.columnar.backend.velox.asyncTimeoutOnTaskStopping`

zhztheplayer:
1. Ensures Velox async threads can be joined after Task ends;
2. Removes the current looped-waiting code since no longer needed. Used new blocking-join mechanism instead.
3. `spark.gluten.sql.columnar.backend.velox.asyncTimeoutOnTaskStopping` is still used as the joining timeout [here](https://github.com/apache/gluten/pull/11882/changes#diff-db1ac7042ef035db368cd4a771440c1db5ec453e13032a0f38d099132822f2deR118-R121).


zhztheplayer:
@zhouyuan @jinchengchenghh @rui-mo

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -190,8 +183,13 @@ void VeloxBackend::init(
...
 #endif
 
+  const auto spillThreadNum = backendConf_->get<uint32_t>(kSpillThreadNum, kSpillThreadNumDefaultValue);
+  if (spillThreadNum > 0) {
+    spillExecutor_ = std::make_unique<folly::CPUThreadPoolExecutor>(spillThreadNum);
+  }
+
   initJolFilesystem();
-  initConnector(hiveConf);
+  initConnector(hiveConnectorConfig_);
```
jinchengchenghh: Do we need to pass the argument?

cpp/velox/compute/VeloxBackend.cc
```diff
@@ -313,6 +311,7 @@ void VeloxBackend::initCache() {
 }
 
 void VeloxBackend::initConnector(const std::shared_ptr<velox::config::ConfigBase>& hiveConf) {
+  (void)hiveConf;
```
jinchengchenghh: why this line?

zhztheplayer: Just to mark out an unused parameter.

cpp/velox/compute/VeloxBackend.cc
zhztheplayer: No. `initConnector` is now misleading, and let me untangle this code a bit.

jinchengchenghh (approved):

kuwylsr:
@zhztheplayer 
Hi zhztheplayer, I’d like to consult you on an issue regarding memory pool leaks in async I/O during Velox scan prefetching.

I came across this community PR: https://github.com/apache/gluten/pull/11882. It appears that at the Gluten framework level, the PR already ensures async I/O threads are waited on (via reference counting and timeout) until they finish execution and are properly destructed.

However, in our current implementation, we are still blocking indefinitely on loads that remain in the loadingstate. Given the framework-level support introduced in this PR, is it still necessary to block indefinitely here?

`~GlutenDirectBufferedInput() override {
    requests_.clear();
    for (auto& load : coalescedLoads_) {
      if (load->state() == facebook::velox::cache::CoalescedLoad::State::kLoading) {
        folly::SemiFuture<bool> waitFuture(false);
        if (!load->loadOrFuture(&waitFuture)) {
          auto& exec = folly::QueuedImmediateExecutor::instance();
          std::move(waitFuture).via(&exec).wait();
        }
      }
      load->cancel();
    }
    coalescedLoads_.clear();
  }`

zhztheplayer:
@kuwylsr Yes, in theory, the code you referenced is now not needed since #11882 ensured proper thread exiting order.

PR #11893, opened 2026-04-08
[DNM] Adding Upstream Spark

zhouyuan:
This patch adds a `spark-main` profile which builds Gluten against Spark nightly builds
TPCH/DS tests added

With this patch Gluten should be able to catch-up with Spark release, e.g., if Spark cut the branch for release Gluten could also move the spark-main shim layer to the target release shim easily 

## How was this patch tested?
pass GHA


philo-he:
@zhouyuan, thanks for proposing this patch.
Perhaps we should move the Spark main tests to a nightly CI job. Since the Spark snapshot is under active development, keeping these tests in our regular CI may frequently cause stability issues and build failures that possibly blocking PR merging. Thanks.

baibaichen:
@zhouyuan would you please follow https://github.com/apache/gluten/discussions/11352?

zhouyuan:
@philo-he @baibaichen Thanks for the inputs. I have monitored this for almost 2 weeks and there are two API changes from Spark in this timeframe which requires modifying the whole shim layers, moving the tests to nightly is a better option.

PR #11894, opened 2026-04-09
[VL] shuffle: TypeAwareCompress(tac) for column-wise data compression like (U)INT64

guowangy:

Introduces **TypeAwareCompress (TAC)** — a column-wise compression layer for shuffle that selects
an algorithm based on each buffer's data type, applied per-buffer alongside the existing LZ4/ZSTD
codec path.

For `INT64`/`UINT64` columns the values are often clustered in a small range, making
Frame-of-Reference + Bit-Packing (FFOR) significantly more effective than generic byte-level
compression. TAC exploits this by encoding 8-byte integer buffers with a 4-lane FFOR codec before
the standard codec sees them.

Here is the performance data on TPCH/TPCDS:
|        |Total Latency|Shuffle Write Size|
|--------|-------------|------------------|
|TPCH-6T |-15%         |-32%              |
|TPCDS-6T|-6%          |-14%              |

### New files

| Path | Purpose |
|------|---------|
| `cpp/core/utils/tac/ffor.hpp` | Header-only 4-lane FFOR codec for `uint64_t` |
| `cpp/core/utils/tac/FForCodec.{h,cc}` | Arrow-Result wrapper around `ffor.hpp` |
| `cpp/core/utils/tac/TypeAwareCompressCodec.{h,cc}` | Type dispatch; self-describing wire format (codec ID + element width embedded in header, so decompression needs no type hint) |
| `cpp/velox/shuffle/VeloxTypeAwareCompress.h` | Maps Velox `TypeKind` → `TacDataType` (`BIGINT` → `kUInt64`) |

### Shuffle integration

- `Payload.cc/h`: `BlockPayload::fromBuffers` accepts an optional `bufferTypes` vector. Per-buffer:
  if `TypeAwareCompressCodec::support(type)` is true, use TAC; otherwise fall back to LZ4/ZSTD.
  A new wire marker `kTypeAwareBuffer = -3` is added; decompression in `readCompressedBuffer` is
  self-describing. If TAC compressed size ≥ original, falls back to `kUncompressedBuffer`.
- `Options.h`: adds `enableTypeAwareCompress` (default `false`) to `LocalPartitionWriterOptions`.
- `VeloxHashShuffleWriter`: populates `bufferTypes` from the schema when TAC is enabled.
- `GlutenConfig.scala`: new config `spark.gluten.sql.columnar.shuffle.typeAwareCompress.enabled` (default `false`).
- `ColumnarShuffleWriter` / `LocalPartitionWriterJniWrapper`: forward the new option to native.

Disabled by default — no behaviour changes for existing deployments.

## How was this patch tested?

`cpp/core/tests/FForCodecTest.cc` covers:
- Round-trip correctness for random, all-zero, monotonic, and near-max value patterns
- `maxCompressedLength` boundary checks
- Invalid input size rejection

`cpp/velox/tests/VeloxShuffleWriterTest.cc`: extended to exercise the TAC path end-to-end through
`VeloxHashShuffleWriter`.

Co-authored-by: Claude Sonnet 4.6


FelixYBW:
We have enabled this in Gazelle. @marin-ma do you still remember why we didn't introduce it in Gluten?

marin-ma:
@FelixYBW We used the compression and arrow ipc payload API in gazelle, and added the FastPFor compression for integer column types and it's also used as the default compression method for integers in gazelle https://github.com/fast-pack/FastPFOR

Because adding a new compression algorithm for the Arrow API requires extra patches, we removed it mainly due to maintenance concerns. Using the default LZ4 algorithm did not result in significant performance regression.

guowangy:
@marin-ma We have also investigated FastPFor library, but it only supports int32 type. This version is aimed to optimize int64 data type, which can archive better compression ratio and speed than int32 types.

FelixYBW:
Thank you for your PR. TPCDS perf looks good.  SF10T perf boosts by 5%. shuffle size decrease by 7%. most of the queries benefit. compress and decompress time contrinutes most of the perf improvement.

Did you implemented the FFor codec or refered from some other github project? 

@marin-ma can you take a look of the PR? Let's merge it once done.

guowangy:
@FelixYBW As in `ffor.hpp` shows, we take [FastLanes](https://www.vldb.org/pvldb/vol16/p2132-afroozeh.pdf) as reference ([src](https://github.com/cwida/FastLanes/tree/dev/src/alp/src)), with below modification:
* Simplify the code using C++ template
* Using 4 lanes instead of 8 lanes to support shorter data block
* More flexible to handle the tailing data that not fit in a whole block
* Add the storage format and make it support generic int64 array compress/uncompress

FelixYBW:
> @FelixYBW As in `ffor.hpp` shows, we take [FastLanes](https://www.vldb.org/pvldb/vol16/p2132-afroozeh.pdf) as reference ([src](https://github.com/cwida/FastLanes/tree/dev/src/alp/src)), with below modification:
> 
> * Simplify the code using C++ template
> * Using 4 lanes instead of 8 lanes to support shorter data block
> * More flexible to handle the tailing data that not fit in a whole block
> * Add the storage format and make it support generic int64 array compress/uncompress

It's MIT license ...

**Option A: Header Preservation (Recommended)**
```
Keep the original MIT license header at the top of the specific files you moved, and add a note below it:

Copyright (c) [Year] [Original Author]

Licensed under the MIT License... (original text)

---

Modifications Copyright (c) [Year] [Your Name/Company], licensed under the Apache License, Version 2.0.
```

guowangy:
@FelixYBW 
> It's MIT license ...
> 
> **Option A: Header Preservation (Recommended)**

`ffor.hpp` license header changed at [01d7388](https://github.com/apache/gluten/pull/11894/commits/01d73889b03e3b93101b8ed3da503d82cb96c1e0).

cpp/velox/shuffle/VeloxHashShuffleWriter.h
```diff
@@ -336,6 +336,7 @@ class VeloxHashShuffleWriter : public VeloxShuffleWriter {
 
   bool hasComplexType_ = false;
   std::vector<bool> isValidityBuffer_;
+  std::vector<int8_t> bufferTypes_;
```
marin-ma: nit: bufferTypes_ -> tacBufferTypes_

marin-ma:
> @marin-ma We have also investigated FastPFor library, but it only supports int32 type. This version is aimed to optimize int64 data type, which can archive better compression ratio and speed than int32 types.

Since the compression algos are for different data types, does it mean there's no conflict if we also add FastPFor for int32 compression?

guowangy:
> > @marin-ma We have also investigated FastPFor library, but it only supports int32 type. This version is aimed to optimize int64 data type, which can archive better compression ratio and speed than int32 types.
> 
> Since the compression algos are for different data types, does it mean there's no conflict if we also add FastPFor for int32 compression?

Yes.

marin-ma (approved):
LGTM. Thanks!

FelixYBW:
> `ffor.hpp` license header changed at [01d7388](https://github.com/apache/gluten/pull/11894/commits/01d73889b03e3b93101b8ed3da503d82cb96c1e0).

@weiting-chen @zhouyuan Can you take a look if license modification is OK.

looks not: [License Header Check / check (pull_request)](https://github.com/apache/gluten/actions/runs/24446162383/job/71535076603?pr=11894)
License Header Check / check (pull_request)Failing after 23s

FelixYBW (approved):

FelixYBW:
Interesting, due to the shuffle datasize is changed, some querys' query plan is changed. Some of the query difference comes from query plan changes.

cpp/core/utils/tac/ffor.hpp
```diff
@@ -0,0 +1,493 @@
+/*
```
zhouyuan, line 18: please also add a apache header above this line

zhouyuan:
@guowangy would you please also fix the C++ format issue?

guowangy:
> @guowangy would you please also fix the C++ format issue?

Done

PR #11749, opened 2026-03-13
[CORE] Fix AssertNotNull mapping

zml1206:
## What changes are proposed in this pull request?

Fix CH CI.

## How was this patch tested?

CI.

## Was this patch authored or co-authored using generative AI tooling?

No.

PR #11898, opened 2026-04-09
[GLUTEN-11749][VL] Throw exception if cannot get the pre-built hash table from ObjectStore correctly

JkSelf:
When enabling `HashTableBuildOncePerExecutor`, we need use the pre-built hash table. If not, it should throw exception directly.

Related issue: #11749

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -399,24 +399,32 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
   } else if (
       sJoin.has_advanced_extension() &&
       SubstraitParser::configSetInOptimization(sJoin.advanced_extension(), "isBHJ=")) {
+    bool hashTableBuildOncePerExecutorEnabled =
+        veloxCfg_->get<bool>(kHashTableBuildOncePerExecutor, kHashTableBuildOncePerExecutorDefault);
+
     std::string hashTableId = sJoin.hashtableid();
 
     std::shared_ptr<core::OpaqueHashTable> opaqueSharedHashTable = nullptr;
     bool joinHasNullKeys = false;
 
-    try {
-      auto hashTableBuilder = ObjectStore::retrieve<gluten::HashTableBuilder>(getJoin(hashTableId));
-      joinHasNullKeys = hashTableBuilder->joinHasNullKeys();
-      auto originalShared = hashTableBuilder->hashTable();
-      opaqueSharedHashTable = std::shared_ptr<core::OpaqueHashTable>(
-          originalShared, reinterpret_cast<core::OpaqueHashTable*>(originalShared.get()));
+    if (hashTableBuildOncePerExecutorEnabled) {
+      std::cout << "the hashTableBuildOncePerExecutorEnabled is set" << "\n";
+      try {
+        auto hashTableBuilder = ObjectStore::retrieve<gluten::HashTableBuilder>(getJoin(hashTableId));
+        joinHasNullKeys = hashTableBuilder->joinHasNullKeys();
+        auto originalShared = hashTableBuilder->hashTable();
+        opaqueSharedHashTable = std::shared_ptr<core::OpaqueHashTable>(
+            originalShared, reinterpret_cast<core::OpaqueHashTable*>(originalShared.get()));
 
-      LOG(INFO) << "Successfully retrieved and aliased HashTable for reuse. ID: " << hashTableId;
-    } catch (const std::exception& e) {
-      LOG(WARNING)
-          << "Error retrieving HashTable from ObjectStore: " << e.what()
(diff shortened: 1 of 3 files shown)


cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -399,24 +399,27 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
...
-      opaqueSharedHashTable = std::shared_ptr<core::OpaqueHashTable>(
-          originalShared, reinterpret_cast<core::OpaqueHashTable*>(originalShared.get()));
-
-      LOG(INFO) << "Successfully retrieved and aliased HashTable for reuse. ID: " << hashTableId;
-    } catch (const std::exception& e) {
-      LOG(WARNING)
-          << "Error retrieving HashTable from ObjectStore: " << e.what()
-          << ". Falling back to building new table. To ensure correct results, please verify that spark.gluten.velox.buildHashTableOncePerExecutor.enabled is set to false.";
-      opaqueSharedHashTable = nullptr;
+    if (isHashTableBuildOncePerExecutor) {
```
jinchengchenghh: Could you read the config from the member veloxCfg_ of SubstraitToVeloxPlan?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -399,24 +399,27 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
...
+      try {
+        auto hashTableBuilder = ObjectStore::retrieve<gluten::HashTableBuilder>(getJoin(hashTableId));
+        joinHasNullKeys = hashTableBuilder->joinHasNullKeys();
+        auto originalShared = hashTableBuilder->hashTable();
+        opaqueSharedHashTable = std::shared_ptr<core::OpaqueHashTable>(
+            originalShared, reinterpret_cast<core::OpaqueHashTable*>(originalShared.get()));
+
+        LOG(INFO) << "Successfully retrieved and aliased HashTable for reuse. ID: " << hashTableId;
+      } catch (const std::exception& e) {
+        LOG(ERROR)
```
jinchengchenghh: LOG(ERROR) does not throw exception, if the config is false, please interrupt here. throw e. Could you not use try catch for normal execution path?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
JkSelf: Yes. Updated.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
JkSelf: When this configuration is enabled, the hash table must be retrieved from the ObjectStore. If retrieval fails, an exception should be thrown on the Gluten side rather than letting it propagate to the Velox side, consistent with the approach in [PR #11749](https://github.com/apache/gluten/pull/11749). If configuration is disabled, we already set opaqueSharedHashTable to null [here](https://github.com/apache/gluten/pull/11898/changes#diff-08ec1ecb10e48cd471188daf2152228da4b00a81c16acaf2e87b040fec16f8c5R407) and no need further process.

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -399,24 +399,32 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
...
-          originalShared, reinterpret_cast<core::OpaqueHashTable*>(originalShared.get()));
-
-      LOG(INFO) << "Successfully retrieved and aliased HashTable for reuse. ID: " << hashTableId;
-    } catch (const std::exception& e) {
-      LOG(WARNING)
-          << "Error retrieving HashTable from ObjectStore: " << e.what()
-          << ". Falling back to building new table. To ensure correct results, please verify that spark.gluten.velox.buildHashTableOncePerExecutor.enabled is set to false.";
-      opaqueSharedHashTable = nullptr;
+    if (hashTableBuildOncePerExecutorEnabled) {
+      std::cout << "the hashTableBuildOncePerExecutorEnabled is set" << "\n";
```
jinchengchenghh: Do you need the log?

cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -399,24 +399,32 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::toVeloxPlan(const ::substrait::
...
+
+        LOG(INFO) << "Successfully retrieved and aliased HashTable for reuse. ID: " << hashTableId;
+      } catch (const std::exception& e) {
+        throw GlutenException(
+            "Error retrieving HashTable from ObjectStore: " + std::string(e.what()) +
+            " You can set spark.gluten.velox.buildHashTableOncePerExecutor.enabled"
+            " to false as workaround.");
+      }
+    } else {
+      std::cout << "the hashTableBuildOncePerExecutorEnabled is false" << "\n";
```
jinchengchenghh: LOG(INFO)

PR #11899, opened 2026-04-09
[CORE] optimize driver heap memory by lazily store metadata columns and fix mismatch result when input_file_name is only in filter condition

taiyang-li:

When scanning tens of millions of files, the driver may run out of heap memory during split construction.

Currently we always generate per-file metadata maps (InputFileName / InputFileBlockStart / InputFileBlockLength) for each split, even if the query does not reference these metadata columns. This causes a large number of `HashMap` instances to be kept in memory and may trigger OOM in the planner.

## How was this patch tested?

New ut.

Generated-by: trae

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
@@ -112,6 +112,18 @@ trait BasicScanExecTransformer extends LeafTransformSupport with BaseDataSource 
       case _ => Seq(partition)
     }
 
+    val metadataFromSpark = getMetadataColumns().map(_.name)
+
+    val inputFileRelatedMetadataKeys = Seq(
+      InputFileName().prettyName,
+      InputFileBlockStart().prettyName,
+      InputFileBlockLength().prettyName)
+
+    val neededInputFileRelatedMetadataKeys =
+      inputFileRelatedMetadataKeys.filter(k => output.exists(_.name == k))
+
+    val metadataColumnNames = (metadataFromSpark ++ neededInputFileRelatedMetadataKeys).distinct
+
     BackendsApiManager.getIteratorApiInstance
       .genSplitInfo(
         partition.index,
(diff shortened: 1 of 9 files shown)


taiyang-li:
@baibaichen  @rui-mo  could you help review this pr, thanks!

gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
```diff
@@ -112,14 +112,26 @@ trait BasicScanExecTransformer extends LeafTransformSupport with BaseDataSource
...
 
+    val metadataFromSpark = getMetadataColumns().map(_.name)
+
+    val inputFileRelatedMetadataKeys = Seq(
+      InputFileName().prettyName,
+      InputFileBlockStart().prettyName,
+      InputFileBlockLength().prettyName)
+
+    val neededInputFileRelatedMetadataKeys =
+      inputFileRelatedMetadataKeys.filter(k => output.exists(_.name == k))
```
rui-mo, line 123: Thanks for the fix. If the output does not include a specific column, but the scan filter references it, do we still need to keep the metadata?

taiyang-li: @rui-mo thanks for you reply. I'll add a test to verify it.

zzcclp:
any process about this pr @taiyang-li

taiyang-li:
@zzcclp I've been a bit busy lately and don't have the energy to read this PR for now. I'll check it out when I have idle time

gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
taiyang-li: @rui-mo I found gluten main doesn't support metadata columns like `input_file_name()` only in filter condition, e.g. `select id from table where input_file_name like '%parquet%'`. It will produce wrong result.  I fixed it in current pr. 

After that, medata columns would be pushed down to scan. The case you described before won't happen. Besides, I added ut to make sure it it ok.

zzcclp (approved):
LGTM

gluten-substrait/src/main/scala/org/apache/gluten/execution/BasicScanExecTransformer.scala
rui-mo: Thanks @taiyang-li.

PR #11905, opened 2026-04-09
[GLUTEN-11885][VL] Respect custom VELOX_HOME in build info generation

officialasishkumar:

Fixes #11885.

`dev/gluten-build-info.sh` was always reading Velox revision metadata from the default `ep/build-velox/build/velox_ep` checkout, even when the build used a different Velox source tree via `--velox_home`.

This PR updates the build-info path resolution so the Velox revision metadata is taken from the actual configured Velox checkout:

- honor an explicit backend path passed to `gluten-build-info.sh`
- fall back to the exported `VELOX_HOME` environment variable when present
- fall back to the generated `cpp/build/CMakeCache.txt` `VELOX_HOME` entry before using the default path
- quote the resolved backend path when collecting git metadata
- export `VELOX_HOME` from `dev/builddeps-veloxbe.sh` so the one-shot Velox build flow propagates the configured path into Maven's build-info generation step
- parse `--velox_repo`, `--velox_branch`, and `--velox_home` as scalars so `VELOX_HOME` can be exported to child processes correctly

## How was this patch tested?

- `bash -n dev/gluten-build-info.sh`
- `bash -n dev/builddeps-veloxbe.sh`
- created a temporary standalone git repo at `/tmp/fake-velox`
- verified `VELOX_HOME=/tmp/fake-velox bash dev/gluten-build-info.sh ... --backend velox` writes the fake repo commit into `gluten-build-info.properties`
- verified `source ./dev/builddeps-veloxbe.sh --velox_home=/tmp/fake-velox true` exports `VELOX_HOME` for child processes

No


FelixYBW:
Thank you for the fix

dev/builddeps-veloxbe.sh
```diff
@@ -120,15 +120,15 @@ do
         shift # Remove argument name from processing
         ;;
         --velox_repo=*)
-        VELOX_REPO=("${arg#*=}")
+        VELOX_REPO="${arg#*=}"
```
philo-he, line 123: Do we also need such modification for some other args?

philo-he (commented):
Looks good overall. Thanks.

dev/builddeps-veloxbe.sh
officialasishkumar: I checked the current script again, and I do not think we need the same change for the other args in this PR. The scalar handling here matters for `--velox_repo`, `--velox_branch`, and especially `--velox_home` because they are reused when building `VELOX_PARAMETER` and when propagating `VELOX_HOME` to child processes. The other flags are only consumed locally in this script today. If there is another arg that also needs to cross that boundary, I can handle it in a follow-up.

philo-he: @officialasishkumar, thanks for the check. Since this functionality is not well covered by CI, could you verify it in local build with a custom Velox home used? You may simply check a local resource file generated by   gluten-build-info.sh.

dev/gluten-build-info.sh
```diff
@@ -49,6 +50,34 @@ function echo_clickhouse_revision_info() {
...
+    echo "$BACKEND_HOME"
+    return
+  fi
+
+  if [ -n "${VELOX_HOME:-}" ]; then
+    echo "$VELOX_HOME"
+    return
+  fi
+
+  CACHED_VELOX_HOME=$(read_cmake_cache_path VELOX_HOME)
```
philo-he: Is this detection always reliable? Can we just use this for getting velox home?

philo-he (commented):
I tried with these changes and specified a custom path for --velox_home, but after the build, it appears that the velox revision in the generated build info doesn't align with the revision of my custom velox home, according to `gluten-core/target/generated-resources/gluten-build-info.properties`. Could you help verify again? Thank you.

dev/gluten-build-info.sh
officialasishkumar: Fixed in ee37515a0. The script now reads `VELOX_HOME` from `CMakeCache.txt` regardless of CMake's cache value type, so `VELOX_HOME:UNINITIALIZED=...` from `-DVELOX_HOME=...` is honored. I verified it with a local fake Velox checkout and confirmed `gluten-build-info.properties` uses the custom checkout revision.

philo-he (approved):
LGTM. My local test also verified that it works. Thank you!

PR #11931, opened 2026-04-13
[VL] Find the aggregate function name by matching result type

rui-mo:

Find the aggregate function name by matching the result type, rather than returning an arbitrary aggregate function name blindly.
Follow-up for: https://github.com/apache/gluten/pull/11891.

## How was this patch tested?

With existing UTs.

No

### DIFF
--- a/cpp/velox/substrait/SubstraitToVeloxPlan.cc
+++ b/cpp/velox/substrait/SubstraitToVeloxPlan.cc
@@ -165,17 +165,36 @@ RowTypePtr getJoinOutputType(
   VELOX_FAIL("Output should include left or right columns.");
 }
 
+// The base names can be referred from 'velox/expression/FunctionSignature.cpp'.
+std::string typeBaseName(const TypePtr& type) {
+  if (type->isDecimal()) {
+    return "decimal";
+  }
+  if (type->isPrimitiveType()) {
+    return type->toString();
+  }
+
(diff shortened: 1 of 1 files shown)


cpp/velox/substrait/SubstraitToVeloxPlan.cc
```diff
@@ -1383,11 +1422,7 @@ core::PlanNodePtr SubstraitToVeloxPlanConverter::constructValueStreamNode(
   }
 
   // Create TableScanNode
-  auto tableScanNode = std::make_shared<core::TableScanNode>(
-      nodeId,
-      outputType,
-      tableHandle,
-      assignments);
+  auto tableScanNode = std::make_shared<core::TableScanNode>(nodeId, outputType, tableHandle, assignments);
```
rui-mo, line 1421: @zhouyuan The clang-format check in CI doesn’t seem to work correctly, as it didn’t report any formatting issues when the code was unformatted.
Also, I’m using clang-format version 15.0.7 for formatting, but the results appear inconsistent.
Do you have any input? Thanks.

zhouyuan: It may due to `cc` files are missed here: 
https://github.com/apache/gluten/blob/main/dev/check.py#L27

zhouyuan (approved):

PR #11947, opened 2026-04-15
[VL][UT] Enable two tests previously failing due to Velox Expand operator bug

philo-he:

Enable these two tests in `GlutenDeltaBasedUpdateAsDeleteAndInsertTableSuite`. It covers Spark's delta-based row update, which uses a row operation flag to mark deletes and inserts. An Expand operator duplicates each row into two — one for delete and one for insert. When updating a table with the WriteDelta operator, delete records are filtered out based on the row operation flag.

For the delete flag marked row, NULL literal will be used to replace the deprecated data, passing to Expand operators. Previously, the operator incorrectly handles literals of complex types, finally producing NULLs instead of the expected values. This bug was fixed by Velox PR some months ago: https://github.com/facebookincubator/velox/pull/15684.

## How was this patch tested?
Local test passed.

No.


philo-he:
@jinchengchenghh, could you take a look?

JkSelf (approved):
LGTM.

jinchengchenghh (approved):
Thanks for figuring out!

PR #11948, opened 2026-04-16
[VL][DELTA] Reconcile struct field names in CaseWhen/If for Velox SWITCH compatibility

acvictor:
Delta column mapping remaps struct field names to physical UUIDs (col-<uuid>), but MERGE plan CaseWhen branches can mix physical and logical field names. Velox SWITCH requires exact type matching including field names, causing InvariantViolationException on MERGE with column-mapped struct columns:

```
 InvariantViolationException: VeloxRuntimeError
 Error Code: INVALID_STATE
 Reason: Else clause of a SWITCH statement must have the same type as 'then' clauses.
   Expected ROW<"col-e49d92e3-0b53-46a2-ba29-c928d381433d":INTEGER>,
   but got ROW<foo:INTEGER>
```
    
This PR adds StructTypeReconciler that detects struct field name mismatches across conditional expression branches and wraps mismatched branches in ROW casts to normalize field names. Handles nested structs, arrays of structs, and maps.

## How was this patch tested?
Added UT

No


dcoliversun (approved):
LGTM. Thanks for great work!
If others don't have comments, I will merge this PR later.

PR #11953, opened 2026-04-17
[VL] Fix shallow CreateNamedStruct detection in CollapseProjectExecTransformer

acvictor:
`containsNamedStructAlias` in `CollapseProjectExecTransformer` guards against collapsing project nodes when `CreateNamedStruct` is present, since Velox generates a special obj output that breaks bind reference resolution after collapse. The existing pattern match only detects `CreateNamedStruct` as the direct child of `Alias` and misses `CreateNamedStruct` nested inside any wrapper expression like `If`, `CaseWhen` allowing unsafe collapse that causes Velox bind reference failures at runtime. This PR uses `expression.exists` to walk the full subtree.

## How was this patch tested?
Existing UT

No


liujiayi771:
@acvictor Thanks for the fix. Could you add a test case for that error case, if possible?

jinchengchenghh (approved):
Thanks for your fix!

acvictor:
> @acvictor Thanks for the fix. Could you add a test case for that error case, if possible?

Added UTs for all Spark versions.

Issue #11969, opened 2026-04-21
[VL] Spark 4.1: Compile error with spark 4.1


Yao-MR:
### Backend

VL (Velox)

### Bug description

compile command:

./dev/buildbundle-veloxbe.sh  --spark_version=4.1

### Gluten version

master branch

### Spark version

4.1

### System information

env:
  os: ubuntu
  java: 17
  scala: 2.13
  spark: 4.1

### Relevant logs

```bash
[INFO] --- scala:4.9.5:add-source (scala-compile-first) @ gluten-core ---
[INFO] Add Source directory: /data1/community/incubator-gluten/gluten-core/src/main/scala
[INFO] Add Test Source directory: /data1/community/incubator-gluten/gluten-core/src/test/scala
[INFO] 
[INFO] --- scala:4.9.5:compile (scala-compile-first) @ gluten-core ---
[INFO] /data1/community/incubator-gluten/gluten-core/src/main/java:-1: info: compiling
[INFO] /data1/community/incubator-gluten/gluten-core/target/generated-sources/protobuf/java:-1: info: compiling
[INFO] /data1/community/incubator-gluten/gluten-core/src/main/scala:-1: info: compiling
[INFO] Compiling 116 source files to /data1/community/incubator-gluten/gluten-core/target/scala-2.13/classes at 1776748484794
[INFO] compiler plugin: BasicArtifact(org.wartremover,wartremover_2.13,3.2.0,null)
[ERROR] scalac error: '1.8' is not a valid choice for '-release'
[INFO]   scalac -help  gives more information
[ERROR] scalac error: bad option: '-release'
[INFO]   scalac -help  gives more information
[ERROR] exception compilation error occurred!!!
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
    at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:355)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:253)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:236)
    at scala_maven_executions.JavaMainCallerByFork.run (JavaMainCallerByFork.java:95)
    at scala_maven.ScalaCompilerSupport.compile (ScalaCompilerSupport.java:173)
    at scala_maven.ScalaCompilerSupport.doExecute (ScalaCompilerSupport.java:86)
[...]

PR #11971, opened 2026-04-21
[GLUTEN-11969] Enable Java 17 release flag for Spark 4+ build

Yao-MR:

Adding -Dmaven.compiler.release=17 in the Spark 4+ command explicitly forces the release target to 17, making the Java/Scala compilation pipeline consistent.
After this override, the build no longer falls back to 1.8-style release values,

For Spark 4.x, the build should target Java 17.
But the project’s default properties still originate from Java 8 in some paths, so Scala/Maven can end up with a release argument equivalent to 1.8, which is incompatible with the Spark 4 + Scala 2.13 + JDK17 toolchain.
So the build effectively mixed:
Expected: Java 17
Leaked/default value: 1.8 release semantics

## How was this patch tested?

UT

CHANGE BEFORE

./dev/buildbundle-veloxbe.sh --spark_version=4.1 failed!

CHANGE AFTER

./dev/buildbundle-veloxbe.sh --spark_version=4.1 succeed!

NO

Related issue: #11969

### DIFF
--- a/dev/buildbundle-veloxbe.sh
+++ b/dev/buildbundle-veloxbe.sh
@@ -27,7 +27,10 @@ function build_for_spark {
   major_version=$(echo $spark_version | cut -d'.' -f1)
 
   if [ "$major_version" -ge 4 ]; then
-    ${MVN_CMD} clean install -Pbackends-velox -Pspark-$spark_version -Pjava-17 -Pscala-2.13 -DskipTests
+    # Force Java 17 release target to override any `maven.compiler.source/target=1.8`
+    # that may be injected by a user's ~/.m2/settings.xml (e.g. an active jdk-8 profile),
+    # which would otherwise cause scalac to fail with: "'1.8' is not a valid choice for '-release'".
+    ${MVN_CMD} clean install -Pbackends-velox -Pspark-$spark_version -Pjava-17 -Pscala-2.13 -DskipTests -Dmaven.compiler.release=17
   else
     ${MVN_CMD} clean install -Pbackends-velox -Pspark-$spark_version -DskipTests
   fi


philo-he:
@Yao-MR, thanks for the PR.

Can we add this setting in pom.xml next to the following line for simplicity?
`<maven.compiler.release>${java.version}</maven.compiler.release>`

https://github.com/apache/gluten/blob/1ca995c17a64f5e38086dd8d599088177973882a/pom.xml#L63

Yao-MR:
> @Yao-MR, thanks for the PR.
> 
> Can we add this setting in pom.xml next to the following line for simplicity? `<maven.compiler.release>${java.version}</maven.compiler.release>`
> 
> https://github.com/apache/gluten/blob/1ca995c17a64f5e38086dd8d599088177973882a/pom.xml#L63



> @Yao-MR, thanks for the PR.
> 
> Can we add this setting in pom.xml next to the following line for simplicity? `<maven.compiler.release>${java.version}</maven.compiler.release>`
> 
> https://github.com/apache/gluten/blob/1ca995c17a64f5e38086dd8d599088177973882a/pom.xml#L63

@philo-he 

An alternative would be to set <maven.compiler.release>17</maven.compiler.release> inside the java-17 profile in the root pom.xml. That works for top-level invocations, but profile-level properties are still overridable by command-line -D and by settings.xml activeByDefault profiles on child modules, depending on activation order. Pinning the value at the script's mvn invocation is the most deterministic layer — it's where the "bundle build" contract is actually defined, and it matches how the script already pins -Pjava-17 -Pscala-2.13 for the same reason.

philo-he (approved):
Looks good. Thanks.

PR #11974, opened 2026-04-21
[VL] Remove unused libprotobuf.so.32 from shared library loaders and build script

liujiayi771:
## Summary

`libvelox.so` no longer depends on `libprotobuf.so.32` (verified via `ldd`). This PR removes the unnecessary loading and copying of this library:

- Remove `loader.loadAndCreateLink("libprotobuf.so.32", "libprotobuf.so")` from all 8 OS-specific `SharedLibraryLoader` implementations (CentOS 7/8/9, Ubuntu 20.04/22.04, Debian 11/12, openEuler 24.03)
- Remove `libprotobuf.so.32` from `cp` commands in `dev/build-thirdparty.sh` for CentOS 7, Debian 11/12, and openEuler 24.03
- Add missing `libgeos.so.3.10.7` to `process_setup_centos_8` in `dev/build-thirdparty.sh` to align with its `SharedLibraryLoaderCentos8`

## Test plan

- [x] Verify `ldd libvelox.so` does not show `libprotobuf.so.32` dependency
- [x] Build and run on CentOS 8 to confirm no missing library errors


philo-he (approved):
Looks good. Thank you!
It appears that the static protobuf lib is linked, so we don't need to consider the shared lib in the thirdparty build script and shared lib loading.

PR #11983, opened 2026-04-24
[VL] Move glog initialization before first LOG() call in  VeloxBackend::init

acvictor:

`LOG(INFO)` was called before `google::InitGoogleLogging()`, which is undefined behavior. This PR moves glog setup to immediately after `backendConf_` is available, ensuring all subsequent `LOG()` calls are safe.

## How was this patch tested?

Exisitng UTs

No


Yohahaha (approved):
LGTM! but I dont really realize what UB will happened, have you encounter such case?

acvictor:
> LGTM! but I dont really realize what UB will happened, have you encounter such case?

I got one crash while printing configs during testing that I could not reproduce. This was the only explanation I could see.

Issue #11635, opened 2026-02-20
ColumnarPartialProject not take effect when the parent is also ProjectExec


wecharyu:
### Description

https://github.com/apache/incubator-gluten/pull/9026 adds `ColumnarPartialProject` when the parent is also columnar, but it should be added when the parent supports partial fallback. Here is a simple example with nested `ProjectExec`:
```sql
select plus_one(col1) as col2, l_partkey from (
  select plus_one(l_orderkey) as col1, l_partkey from lineitem
)
```
```bash
== Physical Plan ==
*(1) Project [if (isnull(col1#73L)) null else plus_one(knownnotnull(col1#73L)) AS col2#74L, l_partkey#1L]
+- *(1) Project [if (isnull(l_orderkey#0L)) null else plus_one(knownnotnull(l_orderkey#0L)) AS col1#73L, l_partkey#1L]
   +- VeloxColumnarToRow
      +- ^(1) BatchScanExecTransformer parquet file:/root/workspace/gluten-community/backends-velox/target/scala-2.13/test-classes/tpch-data-parquet/lineitem[l_orderkey#0L, l_partkey#1L] ParquetScan DataFilters: [], Format: parquet, Location: InMemoryFileIndex(1 paths)[file:/root/workspace/gluten-community/backends-velox/target/scala-2.13..., PartitionFilters: [], PushedAggregation: [], PushedFilters: [], PushedGroupBy: [], ReadSchema: struct<l_orderkey:bigint,l_partkey:bigint> RuntimeFilters: [] NativeFilters: []
```

There is a full C2R fallback even if the two `ProjectExec` supports partial projection.

### Gluten version

main branch

jinchengchenghh:
Do you know why the Spark CollapseProject rule cannot take effect? Usually, there should be only one Project.

wForget:
> Do you know why the Spark CollapseProject rule cannot take effect? Usually, there should be only one Project.

CollapseProject rule does not always work, see: https://issues.apache.org/jira/browse/SPARK-36718

PR #11988, opened 2026-04-26
[GLUTEN-11635][FOLLOW-UP] Respect partial fallback configs when checking node support

wecharyu:
Minor follow up for https://github.com/apache/gluten/pull/11637.

## How was this patch tested?

Add UT.

No.

Related issue: #11635

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/PartialFallback.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/PartialFallback.scala
@@ -16,6 +16,8 @@
  */
 package org.apache.gluten.extension
 
+import org.apache.gluten.config.GlutenConfig
+
 import org.apache.spark.sql.catalyst.rules.{Rule, RuleExecutor}
 import org.apache.spark.sql.execution.{GenerateExec, ProjectExec, SparkPlan}
 import org.apache.spark.sql.internal.SQLConf
@@ -37,8 +39,9 @@ case class PartialFallbackRules() extends Rule[SparkPlan] {
 }
 
 object PartialFallback {
-  def supportPartialFallback(plan: SparkPlan): Boolean = {
-    plan.isInstanceOf[ProjectExec] ||
-    plan.isInstanceOf[GenerateExec]
+  def supportPartialFallback(plan: SparkPlan): Boolean = plan match {
+    case _: ProjectExec => GlutenConfig.get.enableColumnarPartialProject
+    case _: GenerateExec => GlutenConfig.get.enableColumnarPartialGenerate
+    case _ => false
   }
 }
--- a/backends-velox/src/test/scala/org/apache/spark/sql/execution/GlutenHiveUDFSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/spark/sql/execution/GlutenHiveUDFSuite.scala
@@ -162,14 +162,28 @@ class GlutenHiveUDFSuite extends GlutenQueryComparisonTest with SQLTestUtils {
       val plusOne = udf((x: Long) => x + 1)
       spark.udf.register("plus_one", plusOne)
       sql(s"CREATE TEMPORARY FUNCTION noInputUDTF AS '${classOf[NoInputUDTF].getName}'")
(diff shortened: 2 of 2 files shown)


zml1206 (approved):
lgtm

backends-velox/src/main/scala/org/apache/gluten/extension/PartialFallback.scala
```diff
@@ -37,8 +39,9 @@ case class PartialFallbackRules() extends Rule[SparkPlan] {
 }
 
 object PartialFallback {
-  def supportPartialFallback(plan: SparkPlan): Boolean = {
-    plan.isInstanceOf[ProjectExec] ||
-    plan.isInstanceOf[GenerateExec]
+  def supportPartialFallback(plan: SparkPlan): Boolean = plan match {
+    case _: ProjectExec => GlutenConfig.get.enableColumnarPartialProject
```
philo-he, line 43: I noticed the following code already performs this check before PartialFallback.supportPartialFallback is called. Is this a duplicate?


https://github.com/apache/gluten/blob/f0a3df894481d0db844bb7943b59b063e08b26a1/backends-velox/src/main/scala/org/apache/gluten/extension/PartialProjectRule.scala#L28

backends-velox/src/main/scala/org/apache/gluten/extension/PartialFallback.scala
```diff
@@ -37,8 +39,9 @@ case class PartialFallbackRules() extends Rule[SparkPlan] {
...
+    case _: GenerateExec => GlutenConfig.get.enableColumnarPartialGenerate
```
philo-he, line 44: Ditto.

philo-he (commented):
One comment. Please check if it makes sense. Thank you.

backends-velox/src/main/scala/org/apache/gluten/extension/PartialFallback.scala
zml1206: I think this is to handle situations where the parent and child are different; one is ProjectExec, and the other is GenerateExec.

philo-he: I meant the check for config. Am I missing something?

zml1206: The rule-level check only controls whether that specific rule runs. `supportPartialFallback` is used when checking whether a parent SparkPlan can be used as a partial-fallback boundary by another rule.

In nested cases, `PartialGenerateRule` may inspect a `ProjectExec` parent, and `PartialProjectRule` may inspect a `GenerateExec` parent. Without checking the corresponding config in `supportPartialFallback`, a disabled partial fallback type can still be considered a supported boundary indirectly.

philo-he: I see. It's cross referenced. Thanks.

backends-velox/src/test/scala/org/apache/spark/sql/execution/GlutenHiveUDFSuite.scala
```diff
@@ -162,14 +162,28 @@ class GlutenHiveUDFSuite extends GlutenQueryComparisonTest with SQLTestUtils {
...
+                |)""".stripMargin,
+              noFallBack = enablePartialProject
+            ) {
+              df =>
+                val executedPlan = getExecutedPlan(df)
+                assert(
+                  executedPlan.exists(_.isInstanceOf[ColumnarPartialProjectExec]) ==
+                    enablePartialProject)
+                assert(
+                  executedPlan.exists(_.isInstanceOf[ColumnarPartialGenerateExec]) ==
```
philo-he: Can we also test the case that ENABLE_COLUMNAR_PARTIAL_GENERATE is disabled?

wecharyu: Done.

philo-he (approved):
Thanks.

PR #11992, opened 2026-04-27
[VL] Support case-insensitive celeborn shuffle manager detection

kerwin-zk:

Support case-insensitive celeborn shuffle manager detection

## How was this patch tested?

CI


zml1206 (approved):
lgtm

PR #11996, opened 2026-04-27
[CORE] Upgrade Paimon to 1.4.1

Zouxxyy:

- Upgrade default Paimon version from 1.2.0 to 1.4.1
- Remove hardcoded Paimon 1.3.0 from Spark 4.0 and Spark 4.1 profiles to inherit from the default
- Fix Paimon artifact name to include the Scala binary version suffix (`_${scala.binary.version}`) as required by Paimon 1.3+

## How was this patch tested?

Existing tests.

No.

### DIFF
--- a/.github/workflows/velox_backend_x86.yml
+++ b/.github/workflows/velox_backend_x86.yml
@@ -1269,7 +1269,7 @@ jobs:
           export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
           export PATH=$JAVA_HOME/bin:$PATH
           java -version
-          $MVN_CMD clean test -Pspark-4.0 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pdelta \
+          $MVN_CMD clean test -Pspark-4.0 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pdelta -Ppaimon \
           -Pspark-ut -DargLine="-Dspark.test.home=/opt/shims/spark40/spark_home/" \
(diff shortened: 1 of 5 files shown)


philo-he:
Thanks for the PR. Is this Paimon version fully compatible with Spark 3.3, 3.4, 3.5? Could you find any official doc for reference?

We have a release script to enable Paimon for these old Spark releases. I am wondering if we can still build Gluten release like this after the bump. Not sue if the Gluten CI has a sufficient validation.

https://github.com/apache/gluten/blob/main/dev/release/build-release.sh

Zouxxyy:
Paimon has been carefully maintaining backward compatibility with Spark versions starting from 3.2. According to the [official documentation](https://paimon.apache.org/docs/1.4/spark/quick-start/), Paimon 1.4.1 supports Spark 3.2, 3.3, 3.4, 3.5, and 4.0. The 3.x series is built with JDK 8 by default, while 4.x uses JDK 17.

Regarding CI coverage: Spark 3.3/3.4/3.5 UT jobs all include `-Ppaimon` in the Velox backend workflow. Spark 4.0 did not have Paimon enabled in CI, so I've added `-Ppaimon` to both the `spark-test-spark40` and `spark-test-spark40-slow` jobs.

For the release script (`dev/release/build-release.sh`), Spark 3.3/3.4/3.5 all build with `-Ppaimon`. I've also added `-Ppaimon` for Spark 4.0 in the release script to keep it consistent.

philo-he (approved):
Looks good. Thank you. cc @zhouyuan

zhouyuan (approved):
👍

PR #12000, opened 2026-04-28
[GLUTEN-11980][CORE] For decimal-key joins, if one side falls back to Spark, force fallback the other side

### LINKED ISSUE #11980: Join operator lost data due to decimal join key.
### Backend

VL (Velox)

### Bug description

There are two tables: A and B. 

```
// The main schema of A
CREATE TABLE A (
  invoiceid DECIMAL(20,0),
  time TIMESTAMP,
  update_time TIMESTAMP,
  dt STRING,
  hour STRING)
USING orc
PARTITIONED BY (dt, hour)
LOCATION 'hdfs://test/A'
TBLPROPERTIES (
  'transient_lastDdlTime'='1718612895')
```

```
// The main schema of B.
CREATE TABLE B (
  `invoiceid` DECIMAL(20,0),
  `status` STRING,
  `update_time` TIMESTAMP,
  `dt` STRING,
  `hour` STRING)
USING orc
PARTITIONED BY (dt, hour)
LOCATION 'hdfs://test/B'
TBLPROPERTIES (
  'transient_lastDdlTime'='1718612913')
```

// The orc file schema in hdfs://test/A

`Type: struct<invoiceid:decimal(38,18), time:timestamp, update_time:timestamp>`

// The orc file schema in hdfs://test/B

`Type: struct<invoiceid:decimal(38,18), status:string, update_time:timestamp>`

```
// The query info.
SELECT count(*)
FROM (
   SELECT c.invoiceid AS invoice_id,
       c.orderid AS order_id,
       c.time AS order_time,
       c.update_time AS update_time
FROM
    (SELECT *
     FROM
         (SELECT *,
                 ROW_NUMBER() OVER (PARTITION BY invoiceid
                                    ORDER BY update_time DESC) AS rn
          FROM A
          WHERE dt >= '20260421') b
     WHERE rn = 1) c
JOIN
    (SELECT *
     FROM B
     WHERE dt >= '20260421'
         AND status = 7) d ON c.invoiceid = d.invoiceid
);
```

The root cause
```
Scan operator of A is Spark Scan hive due to the time and update_time are TIMESTAMP.
Scan operator of B is Gluten NativeScan hive.

Left side: Scan hive (vanilla) → DECIMAL (20,2) row → RowToVeloxColumnar → DECIMAL(20,2) columnar                                                                         
Right: NativeScan hive (Velox) → may be widen when reading ORC internally DECIMAL(38,2) columnar                                                                                   
                                                                                                                                                                      
→ Inconsistent join key types on both sides → Key matching failed in Velox hash join → 0 line output
                                                                                                                                                                      
The problem with RowToVeloxColumnarEXE itself:
When converting RowToVeloxColumnar, the Velox Vector is constructed strictly according to the DECIMAL (20,2) declared in Spark schema, without actively expanding the precision. When NativeScan reads Hive ORC, Velox will internally decode it
Type regularization, for example, DECIMAL (20,2) may be mapped to an internal precision of 128 bit representation DECIMAL (38,2) or Velox ShortDecimal/LongDecimal.
                                                                                                                                                                      
This results in a situation where the plan level type is consistent (both are DECIMAL (20,2)), but the type tags of the actual columnar vectors are inconsistent, causing the key comparison logic of Velox HashJoin to go to the type mismatch branch → join
Output 0 rows.
```

### Gluten version

Gluten-1.5

### Spark version

Spark-3.5.x

### Spark configurations

_No response_

### System information

_No response_

### Relevant logs

```bash

```

beliefer:

This PR proposes to fix a bug. Join operator lost data due to decimal join key if one side is vanilla scan and the other side is native scan.

For each join, if one side's scan (`FileSourceScanExec` or `HiveTableScanExec`) cannot be offloaded to the native engine, but the other side. If the join key of this join exists decmal type, the row could not be matched well due to the  decimal precision within Spark is different from Velox.

Then we will see many rows cannot be matched and even zero rows could be matched.

Why do Spark and Gluten have different precision processing for decimal types?
Spark prioritizes accuracy consistency and would rather sacrifice some performance for this purpose; Gluten pursues ultimate performance, which may introduce subtle differences in accuracy in certain calculations.
There are a lot of issues: https://github.com/apache/gluten/issues/4652

Fixes #11980

## How was this patch tested?

Manual tests in our production environment.
UT.

'No'.
I just use AI to generate the comments.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
@@ -20,7 +20,10 @@ import org.apache.gluten.extension.columnar.FallbackTags
 import org.apache.gluten.extension.columnar.validator.Validator
 
 import org.apache.spark.sql.catalyst.rules.Rule
-import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.{FileSourceScanExec, SparkPlan}
+import org.apache.spark.sql.execution.adaptive.{AdaptiveSparkPlanExec, BroadcastQueryStageExec, ShuffleQueryStageExec}
+import org.apache.spark.sql.execution.joins.{BroadcastHashJoinExec, ShuffledHashJoinExec, SortMergeJoinExec}
+import org.apache.spark.sql.types.DecimalType
 
 // Add fallback tags when validator returns negative outcome.
 case class AddFallbackTags(validator: Validator) extends Rule[SparkPlan] {
@@ -29,6 +32,14 @@ case class AddFallbackTags(validator: Validator) extends Rule[SparkPlan] {
       case p if FallbackTags.maybeOffloadable(p) => addFallbackTag(p)
       case _ =>
     }
+
+    plan.foreach {
+      case adaptive: AdaptiveSparkPlanExec =>
+        validateJoin(adaptive.initialPlan)
+      case other =>
+        validateJoin(other)
+    }
+
     plan
   }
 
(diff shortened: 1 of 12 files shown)


gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
```diff
@@ -40,4 +46,131 @@ case class AddFallbackTags(validator: Validator) extends Rule[SparkPlan] {
...
+   * For each such join, delegates to [[setFallbackTagForOtherSide]] to ensure that if one side's
+   * scan ([[FileSourceScanExec]] or `HiveTableScanExec`) cannot be offloaded to the native engine,
+   * the other side is also forced to fall back. This prevents a decimal-value mismatch that would
+   * produce incorrect (typically empty) join results when one side applies Spark's precision
+   * coercion and the other side reads raw native values.
+   *
+   * AdaptiveSparkPlanExec is handled by descending into its `initialPlan`; all other non-join nodes
+   * are handled recursively through their children.
+   */
+  private def validateJoin(plan: SparkPlan): Unit =
```
jinchengchenghh, line 63: Could we move this to a separate rule?

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/hive/execution/GlutenHiveSQLQuerySuite.scala
```diff
@@ -149,4 +151,162 @@ class GlutenHiveSQLQuerySuite extends GlutenHiveSQLQuerySuiteBase {
       }
     }
   }
+
+  testGluten(
```
jinchengchenghh, line 155: Why do we need this test here? Looks like FallbackSuite has covered the test. And please update the PR description, this problem because one side join is native while another side fallback, so we need to fallback both side, it does not related to decimal, right?

beliefer: No. It is related to decimal join key. We do not need fallback the other side while one side falling back if the join key have good equality(e.g. string, int) .

The test `FallbackSuite` cannot cover all the code path, so we need to add these test cases.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
beliefer: I created another rule `AddFallbackTagsForJoin` like `AddFallbackTags` and change the code of `WithRewrites` like:

```
case class WithRewrites(
      validator: Validator,
      rewriteRules: Seq[RewriteSingleNode],
      offloadRules: Seq[OffloadSingleNode])
    extends Rule[SparkPlan] {
    private val validate = AddFallbackTags(validator)
    private val validateJoin = AddFallbackTagsForJoin(validator)
    private val rewrite = RewriteSparkPlanRulesManager(validate, validateJoin, rewriteRules)
    private val offload = LegacyOffload(offloadRules)

      Seq(rewrite, validate, validateJoin, offload).foldLeft(plan) {
        case (plan, stage) =>
          stage(plan)
      }
    }
  }
```

Then we must change the constructor of `RewriteSparkPlanRulesManager` to accept `AddFallbackTagsForJoin` and change the logic of `RewriteSparkPlanRulesManager`.
I think it's not worth to do this, so I merged the code into `AddFallbackTags`.

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/hive/execution/GlutenHiveSQLQuerySuite.scala
jinchengchenghh: Could you describe more why decimal is different and need to handle differently in PR description?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
```diff
@@ -40,4 +46,131 @@ case class AddFallbackTags(validator: Validator) extends Rule[SparkPlan] {
...
+        validateJoin(a.initialPlan)
+      case _ => plan.children.foreach(validateJoin(_))
+    }
+
+  /**
+   * Enforces symmetric scan fallback for the two sides of a decimal-key join.
+   *
+   * When the join key is a decimal type, a native (Velox) scan and a vanilla Spark scan
+   * ([[FileSourceScanExec]] or `HiveTableScanExec`) may produce different representations of the
+   * same decimal value: the native reader may surface raw uncoerced int128_t values while the
```
jinchengchenghh, line 84: Can we update the native side to support this case?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
```diff
@@ -40,4 +46,131 @@ case class AddFallbackTags(validator: Validator) extends Rule[SparkPlan] {
...
+   * fallback tag to the native scan on the other side, so that both sides end up using the same
+   * read path.
+   *
+   * @param leftChild
+   *   the left subtree of the join
+   * @param rightChild
+   *   the right subtree of the join
+   */
+  private def setFallbackTagForOtherSide(leftChild: SparkPlan, rightChild: SparkPlan): Unit = {
+    val leftHasFallbackScan = hasFallbackScan(leftChild)
```
jinchengchenghh, line 98: Not only the scan fallback cause this issue, after filter, it may also occur?

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
jinchengchenghh: If one side fallbacks, this side should insert ColumnarToRow, why this representation issue?

gluten-ut/spark35/src/test/scala/org/apache/spark/sql/hive/execution/GlutenHiveSQLQuerySuite.scala
beliefer: Updated.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
beliefer: The implementation principle of decimal determines their different accuracies, and it is a difficult problem for me to solve now.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
beliefer: No matter what, we should keep the two side are offloaded at the same time or not.

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/heuristic/AddFallbackTags.scala
jinchengchenghh: We should find the root cause and fix the result mismatch issue in native side, other one in community may fix this issue, please keep this issue open now.

beliefer: The root cause is the decimal accuracy is different from Java side and native side. The implementation mechanism for decimal in Velox is significantly different from that in Java, and I think it will be difficult to make them consistent in the short term.

beliefer: @jinchengchenghh I added more information into https://github.com/apache/gluten/issues/11980.

beliefer:
Because https://github.com/facebookincubator/velox/pull/17463 merged, I will close this one.

PR #12004, opened 2026-04-28
[GLUTEN-11920][CORE] Keep columnar children for dual-mode parents with row output

Yao-MR:

For dual-mode parents that support both columnar and row-based execution, Gluten previously let the child convention follow the parent's output convention by default. When the parent was required to produce rows, this could incorrectly push a row requirement down to its children even though the parent could still consume batch input and satisfy the row output requirement by itself.

This PR changes dual-mode parents to keep a batch child convention under a row output requirement, which aligns transition insertion with Spark behavior and avoids unnecessary row transitions.

The change is applied in both the plan-side child convention derivation and the transition convention derivation so the behavior stays consistent.

This PR also:
- adds transition tests to verify that a dual-mode parent keeps a columnar child for row output;
- adds coverage to verify that a row child is converted to batch when required by a dual-mode parent; and
- re-enables `GlutenColumnarRulesSuite` on Spark 4.1.

## How was this patch tested?

- Added transition tests for dual-mode parent convention handling under row output requirements.
- Re-enabled `GlutenColumnarRulesSuite` on Spark 4.1 to cover the fixed behavior.

issue: https://github.com/apache/gluten/issues/11920

fix in spark: https://github.com/apache/spark/pull/50239


zhztheplayer (approved):

zml1206 (approved):
lgtm

PR #12005, opened 2026-04-28
[GLUTEN-10072][UT] Enable hive execution tests

wecharyu:
Enable unit tests in Spark `sql/hive/src/test/scala/org/apache/spark/sql/hive/execution`:
- Extract `spark-hive` test resources to `spark.test.home` in script `install-spark-resources.sh`
- Use `GlutenHiveResourcePathSupport` to load test resources from `spark.test.home` 
- Use `GlutenTestHiveTables` to register hive qtest tables

https://github.com/apache/gluten/issues/10072

## How was this patch tested?
Pass all new tests.

Codex gpt-5.5


pom.xml
```diff
@@ -1168,6 +1168,7 @@
         <delta.version>2.3.0</delta.version>
         <delta.binary.version>23</delta.binary.version>
         <antlr4.version>4.8</antlr4.version>
+        <hadoop.version>3.3.2</hadoop.version>
```
wecharyu, line 1188: The Hadoop version of Spark 3.3.1 is `3.3.2`, we need explicitly declare it here. Otherwise there are version conflicts:
-  hadoop-common-2.7.4 (default version)
-  hadoop-client-api/runtime-3.3.4 (from spark-hive test)

That mix is unsafe. The old Hadoop 2.7.4 `FsUrlStreamHandlerFactory` can be loaded together with Hadoop 3.3.4 HTTP/HTTPS filesystem classes. As a result, HTTPS URLs used by Hive test
  jar resolution and Ivy ADD JAR resolution may be incorrectly handled as Hadoop filesystem URLs, which can recurse through `HttpsFileSystem.open()` and fail with StackOverflowError.

wecharyu:
@zhouyuan @rui-mo could you help review this patch when you have time? Thanks!

.github/workflows/velox_backend_enhanced.yml
```diff
@@ -215,6 +215,9 @@ jobs:
   spark-test-spark35-slow:
     needs: build-native-lib-centos-7
     runs-on: ubuntu-22.04
+    env: #TODO remove after image update
```
rui-mo, line 218: Could you provide a bit more detail in the PR description on how we should follow up? Thanks.

rui-mo (approved):
Thanks for this work! The change looks good to me. Would you mind sharing some progress details on enabling the Hive execution tests? For example, how many suites were added in this PR, and how many still remain to be added?

wecharyu:
> Would you mind sharing some progress details on enabling the Hive execution tests? 

This PR add 25 hive execution tests in Spark `sql/hive/src/test/scala/org/apache/spark/sql/hive/execution`. There are some other hive tests could be added:
- `sql/hive/src/test/scala/org/apache/spark/sql/hive/orc`
- `sql/hive/src/test/scala/org/apache/spark/sql/hive`

Other tests are related to hive client and DDL operations, which is not necessary test in gluten.

.github/workflows/velox_backend_enhanced.yml
wecharyu: This is an existing pain point in gluten image update:
- Docker image is pushed to apache/gluten repo when merge a related commit to main branch
- This PR that needs update the image is still test on existing 'old' image

Thus we need add scripts in test action before the PR is merged, and remove such actions when the PR is merged and image is updated.

I'm considering to let the such PR that would update the image build a snapshot image and test on the snapshot image, in this case we do not need add additional scripts on test action and do not need a follow-up remove operation.

rui-mo:
@wecharyu Thanks. I’m okay with landing this PR. Before merging, could you check whether the CH workflow failure is related?

wecharyu:
@rui-mo CI is green now.

PR #12006, opened 2026-04-28
[VL] Fix nightly upload action

zhouyuan:
bump the action version so it's allowed in apache repos

## How was this patch tested?
pass GHA

### DIFF
--- a/.github/workflows/velox_nightly.yml
+++ b/.github/workflows/velox_nightly.yml
@@ -353,7 +353,7 @@ jobs:
           name: nightly-gluten-velox-bundle-package-jdk8-x86-${{ steps.date.outputs.date }}
           path: package/
       - name: rsync to apache nightly
-        uses: burnett01/rsync-deployments@0dc935cdecc5f5e571865e60d2a6cdc673704823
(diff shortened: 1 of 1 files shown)


.github/workflows/velox_nightly.yml
```diff
@@ -496,7 +498,7 @@ jobs:
           path: package/target/gluten-velox-bundle-*.jar
           retention-days: 7
       - name: rsync to apache nightly
-        uses: burnett01/rsync-deployments@0dc935cdecc5f5e571865e60d2a6cdc673704823
+        uses: burnett01/rsync-deployments@7659d600d8bdd035bb5cdfba1d4bd0dd4a307ca6 # 8.0.2
```
zml1206: 8.0.3？

zml1206: Why was only this one changed, while the other three remained unchanged? Is there a specific reason for this?

zhouyuan: @zml1206 thanks, missed some of the commits - I have amended in the new commit

zhouyuan: aligned here: https://github.com/apache/flink/blob/master/.github/workflows/docs.yml#L85

zml1206 (approved):
lgtm

Issue #8304, opened 2024-12-23
[CH] Optimize get_json_object nested function calls


KevinyhZou:
### Description

In our online sqls, have some nested `get_json_object` function calls, like `get_json_object(get_json_object(d, '$.a'), '$.b')`, and some of them can be optimized to `get_json_object(d, '$.a.b')`, which can improve the performance

![image](https://github.com/user-attachments/assets/bfdb5aa9-044b-48c8-a78c-b8d41bf13e5d)

![image](https://github.com/user-attachments/assets/dfc0e3f4-7dde-4feb-9447-8b45181a66bd)

KevinyhZou:
some nested get_json_object call case can not optimized:  `select get_json_object(get_json_object('{"a":"{\\\"x\\\":5}"}', '$.a'), '$.x')` can not be converted to `select get_json_object('{"a":"{\\\"x\\\":5}"}', '$.a.x');`

KevinyhZou:
Yes, it is a corner case. In most cases, it can be converted. @PHILO-HE

PHILO-HE:
@KevinyhZou, I feel we can merge the patch for both CH backend and velox backend if it's in good shape. If you are afraid of any similar issue, it's ok to keep a config for users to control the use of this optimization. Thanks!

PR #12007, opened 2026-04-29
[GLUTEN-8304][CORE] Fix CollapseGetJsonObjectExpressionRule incorrectly merging paths across non-GJO expressions

liujiayi771:

Fix a bug in `CollapseGetJsonObjectExpressionRule` where the rule incorrectly merges JSON paths when an inner GJO is separated from an outer GJO by a non-GJO expression.

### Problem

The `CollapseGetJsonObjectExpressionRule` collapses nested `get_json_object` (GJO) calls into one, e.g.:
```sql
get_json_object(get_json_object(col, '$.a'), '$.b')
-- optimized to:
get_json_object(col, '$.a.b')
```

However, when a non-GJO expression (such as `substring`, `concat`) sits between two GJO calls, the rule incorrectly propagated the outer accumulated path into the non-GJO expression's children. Since non-GJO children are recursed with the accumulated `path` parameter, any inner GJO found within will have its path incorrectly merged with the outer path. For example:

```sql
-- original:
get_json_object(substring(get_json_object(txt, '$.a'), 1), '$.y')

-- was incorrectly transformed to:
get_json_object(substring(get_json_object(txt, '$.a.y'), 1), '$.y')
```

The inner GJO's path was corrupted from `$.a` to `$.a.y` (the outer path `$.y` leaked through `substring` and merged into the inner GJO's path), while the outer GJO wrapper remained unchanged. This produces wrong results because `substring` transforms the JSON string — the outer `$.y` lookup should operate on substring's output, not be merged into the inner path.

### Fix

Stop propagating the accumulated `path` into children of non-GJO expressions. In the `case _` branch, recurse with empty path so each subtree starts fresh optimization independently. Inner consecutive GJOs within a subtree are still correctly collapsed. For example:

```sql
get_json_object(substring(get_json_object(get_json_object(col, '$.a'), '$.z'), 1), '$.m')
-- correctly optimized to:
get_json_object(substring(get_json_object(col, '$.a.z'), 1), '$.m')
-- inner GJOs collapsed to '$.a.z', outer '$.m' stays separate
```

## How was this patch tested?

Added comprehensive test coverage for both Velox and ClickHouse backends:
- **GLUTEN-8304 collapse optimization tests** (Velox backend): ported from CH backend with adaptation for `optimizedPlan` (Velox uses `injectOptimizerRule` vs CH's `injectResolutionRule`)
- **Bracket notation tests** using `$['a']` quoted syntax (Velox backend)
- **Non-GJO boundary tests** with `substring`/`concat` verifying paths are NOT merged across non-GJO expressions (both backends)
- **Mixed nesting tests** `GJO(substring(GJO(GJO(col, '$.a'), '$.z'), 1), '$.m')` verifying inner GJOs collapse to `$.a.z` while outer path stays separate (both backends)

Related issue: #8304

### DIFF
--- a/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala
+++ b/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala
@@ -25,7 +25,7 @@ import org.apache.spark.sql.{DataFrame, GlutenTestUtils, Row}
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.aggregate._
 import org.apache.spark.sql.catalyst.optimizer.{ConstantFolding, NullPropagation}
-import org.apache.spark.sql.catalyst.plans.logical.{Aggregate, Filter, LogicalPlan, Project}
+import org.apache.spark.sql.catalyst.plans.logical.Aggregate
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.datasources.v2.clickhouse.ClickHouseConfig
 import org.apache.spark.sql.internal.SQLConf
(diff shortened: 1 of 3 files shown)


liujiayi771:
@KevinyhZou @philo-he Could you help to review?

philo-he (commented):
Looks good. One minor comment. Thank you.

Could you move the newly added test, along with the existing test below, to a new test suite (e.g., `CollapseGetJsonObjectSuite.scala`)?

"GLUTEN-8304: Optimize nested get_json_object"

The `GlutenFunctionValidateSuite.scala‎` should only hold basic tests for SQL functions to validate they are really offloaded.

liujiayi771:
@philo-he Thanks for the suggestion! I've moved the collapse tests to a shared test suite at `gluten-substrait/src/test/scala/org/apache/gluten/extension/columnar/CollapseGetJsonObjectExpressionRuleSuite.scala`.
Instead of using `runQueryAndCompare` (which is backend-specific), the shared suite directly calls `CollapseGetJsonObjectExpressionRule(spark).apply(analyzed)` to verify the plan transformation logic. This way both ClickHouse and Velox share the same set of test cases without needing a real backend, and the duplicate tests in GlutenFunctionValidateSuite (CH) and JsonFunctionsValidateSuite (Velox) have been removed.

philo-he (approved):
LGTM. Thank you!

Issue #12008, opened 2026-04-29
[VL] ExpandExecTransformer should align projection types with Expand output attributes


jianzhenwu:
### Backend

VL (Velox)

### Backend

VL (Velox)

### Bug description

`ExpandExecTransformer` currently converts Spark `ExpandExec.projections` to Substrait `ExpandRel` mostly according to each projection expression's own type. However, Spark's `ExpandExec.output` attributes are the schema contract for each Expand output column.

In some multiple-distinct aggregate plans, especially with decimal expressions, `CASE WHEN`, `round(avg(decimal))`, or duplicated grouping keys, a projection expression's Spark/Gluten transformer type can differ from the corresponding `ExpandExec.output` attribute type. This may produce an ExpandRel whose per-column projection types are inconsistent with the output schema.

Expected behavior:

Gluten should make sure every Expand projection column is type-compatible with the corresponding `ExpandExec.output` attribute before Substrait conversion. If the backend cannot support the required alignment in native Expand, validation should fail cleanly and fallback, instead of producing an invalid native plan or failing later.

Actual behavior:

Plans can hit native planning/validation failures or fallback reasons such as:

```text
Reason: The projections type does not match across different rows in the same column. Got: DECIMAL(27, 10), DECIMAL(28, 10)
Retriable: False
Expression: projections_[j][i]->type()->equivalent(*type)
Function: ExpandNode
/home/jenkins/agent/workspace/di-spark/Gluten/CI_spark_gluten/ep/build-velox/build/velox_ep/velox/core/PlanNode.cpp
Line: 511
Stack trace:
# 0 _ZN8facebook5velox7process10StackTraceC1Ei
# 1 _ZN8facebook5velox14VeloxExceptionC1EPKcmS3_St17basic_string_viewIcSt11char_traitsIcEES7_S7_S7_bNS1_4TypeES7_
# 2 _ZN8facebook5velox6detail14veloxCheckFailINS0_14VeloxUserErrorERKSsEEvRKNS1_18VeloxCheckFailArgsET0_
# 3 _ZN8facebook5velox4core10ExpandNodeC1ESsSt6vectorIS3_ISt10shared_ptrIKNS1_10ITypedExprEESaIS7_EESaIS9_EES3_ISsSaISsEES4_IKNS1_8PlanNodeEE
# 4 _ZN6gluten29SubstraitToVeloxPlanConverter11toVeloxPlanERKN9substrait9ExpandRelE
[...]

PR #12009, opened 2026-04-29
[GLUTEN-12008][VL] Align Expand projection types with output

jianzhenwu:
### What changes are proposed in this pull request?

This PR fixes Expand projection/output type alignment for Velox.

The main changes are:

- Align `ExpandExec` projection expressions with the corresponding `ExpandExec.output` attribute type before native conversion.
- Push non-literal type alignment casts into `PullOutPreProject`, so native `ExpandRel` still only contains fields or literals, which is required by Velox.
- Align null literals inside `ExpandExecTransformer` directly to the output type.
- Add validation-time diagnostics for remaining projection/output type mismatches and return `ValidationResult.failed(...)` instead of generating an invalid native Expand plan.
- Keep a hard guard in `doTransform` to prevent inconsistent Expand projections from reaching Velox.
- Add Velox regression tests for decimal expressions with multiple distinct aggregates.

- Passed:

```bash
git diff --check
```
Passed:

```
mvn -Pbackends-velox -pl backends-velox -am \
  -DskipTests -DskipUTs \
  -Dscalastyle.skip=true \
  -Dcheckstyle.skip=true \
  -Dspotless.check.skip=true \
  -Dlicense.skip=true \
  clean test-compile
```
Added regression coverage in VeloxExpandSuite for:

- round(avg(decimal)) with multiple distinct aggregates
- decimal CASE WHEN sum with multiple distinct aggregates

Yes, this patch was co-authored using generative AI tooling.

Related issue: #12008


jinchengchenghh:
Why this problem occurs? Does the native side produces unexpected result?

jianzhenwu:
> Why this problem occurs? Does the native side produces unexpected result?

@jinchengchenghh Thank you for your reply. Sorry for the late reply. I think this is a bug in Gluten. Here is the SQL that reproduces the exception. [issue 12008](https://github.com/apache/gluten/issues/12008)

jianzhenwu:
hi @JkSelf pls help review.

backends-velox/src/test/scala/org/apache/spark/sql/execution/VeloxExpandSuite.scala
```diff
@@ -77,4 +88,128 @@ class VeloxExpandSuite extends VeloxWholeStageTransformerSuite {
       }
     }
   }
+
+  test("Expand with round(avg(decimal)) and multiple distinct aggregates") {
```
JkSelf, line 92: @jianzhenwu Thanks for your fixing. 

I tried reproducing the issue in my local environment using the SQL you provided (https://github.com/apache/gluten/issues/12008#issue-4348749924 and the two tests here) and the latest main branch(82644d3ca3f2d93b18b4cfb682656ff3b0111668) with Spark 3.5, but I was unable to reproduce it.

In Spark, the projection expressions are passed directly from ExpandExec and should already be aligned with the output schema https://github.com/apache/spark/blob/c26a127ba33137f36d55bf95cac71471e2a1704f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/basicLogicalOperators.scala#L1398-L1407. Could you provide more details on your environment or help investigate why this occurs on your side?

Thanks for your help!

jianzhenwu: I encountered this problem using Spark 3.2. I believe it's also possible to reproduce the problem using Spark 3.3. I've tried using AI to explain the issue.

Spark 3.3 can reproduce the issue because its physical ExpandExec contains this decimal expression shape:

```
CAST((case_dd_decimal26 + case_ccb_decimal26) AS DECIMAL(27,10)) + case_fsv_decimal27
```
Spark declares the Expand output column as:
```
DECIMAL(27,10)
```
The null rows in the same Expand column are also:
```
CAST(NULL AS DECIMAL(27,10))
```
But when Velox compiles the non-null decimal arithmetic row, it infers:
```
DECIMAL(28,10)
```
So native ExpandNode sees mixed types in the same output column:
```
row 0: DECIMAL(28,10)
row 1: DECIMAL(27,10)
```
Then Velox fails with:
```
The projections type does not match across different rows in the same column.
Got: DECIMAL(27, 10), DECIMAL(28, 10)
```
Spark 3.5 does not reproduce it because the generated ExpandExec expression is different:
```
(case_dd_decimal25 + case_ccb_decimal25) + case_fsv_decimal25
```
It does not insert the intermediate:
```
CAST(... AS DECIMAL(27,10))
```
that Spark 3.3 has. With this Spark 3.5 plan shape, Velox’s inferred type stays compatible with Spark’s Expand output type, so all projection rows in the Expand column remain consistent.

So the difference is not the SQL result type. Both Spark versions declare the Expand output as DECIMAL(27,10). The difference is the internal decimal expression tree Spark generates before Gluten/Velox conversion. Spark 3.3’s tree causes Velox to widen one projection row to DECIMAL(28,10); Spark 3.5’s tree does not.

jianzhenwu: hi @JkSelf Do you think this fix is ​​correct to address the issue in the Spark32 scenario?

JkSelf: @jianzhenwu Yes, this issue persists in Spark 3.2 and 3.3. Would it be possible to introduce a new rule to align the output types for spark 32 and 33? This would make the logic much more clear.

jianzhenwu: @JkSelf  Agreed. Added a dedicated rule for Spark 3.2/3.3 to align the output types would make the logic much clearer and easier to maintain.

jianzhenwu:
hi @JkSelf would you mind reviewing this PR when you have some time? I'd appreciate your feedback. Thanks!

JkSelf:
@jianzhenwu Thanks for your fix.


1. The scope of this fix seems to be the Expand type alignment issue on the Velox backend. Given that both the problem statement and the added validation focus on Velox, do we really need to apply `AlignExpandOutputTypes` to the CH backend as well?
2. The current fix feels more complicated than necessary. This PR adds a rewrite rule to align projection types with Expand output types, but then also performs another mismatch check inside `ExpandExecTransformer`. For the Velox path, `ExpandExecTransformer` is created from the rewritten `ExpandExec`, so if the rewrite already guarantees type alignment, the second check looks redundant.
3. The core of the fix seems quite simple: check whether each projection type matches the corresponding output type, and if not, convert it to the expected type. However, the current implementation introduces several helper layers and special-case checks, which makes the code harder to follow and the intent less clear. I wonder if we can simplify this into a more direct implementation with better readability as below.
```
for each projectionRow in projections:
  for each column i:
    val expr = projectionRow(i)
    val targetType = output(i).dataType

    if expr is null literal:
      projectionRow(i) = typedNull(targetType)
    else if expr.dataType != targetType:
      projectionRow(i) = Cast(expr, targetType)
```

jianzhenwu:
@JkSelf Thank you for your reply. I will use a unit test to verify whether CH needs to add this rule.

jianzhenwu:
hi @JkSelf  I verified through unit tests that CH does not need to apply AlignExpandOutputTypes. Then I removed it. Could you please review this and see if any other changes are needed?

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/AlignExpandOutputTypes.scala
```diff
@@ -0,0 +1,84 @@
...
+        val newExpand = expand.copy(projections = alignedProjections)
+        newExpand.copyTagsFrom(expand)
+        newExpand
+      }
+    case _ => plan
+  }
+}
+
+private[gluten] object ExpandOutputTypeAlignment {
+  val isSpark32Or33: Boolean =
```
JkSelf: @jianzhenwu Spark 3.2 support has already been removed, so we only need to handle Spark 3.3 (via SparkVersionUtil.eqSpark33 ). Could we move the version check into this spot: https://github.com/apache/gluten/pull/12009/changes#diff-86ffecde82890df19c91011a1faad3b896b0069025e896d4ce241981378a8213R100 ? If we do that, we should be able to delete the helper methods here: https://github.com/apache/gluten/pull/12009/changes#diff-412628303ba53650091d2847abb93192c3682fe217e3a8578fbe75e79ea8a800R33-R35 . Also, since only ExpandExec can be rewritable in this context, we don’t need to add an extra plan.isInstanceOf[ExpandExec] check inside isRewritable .

JkSelf (commented):
LGTM except one small comment.

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/AlignExpandOutputTypes.scala
jianzhenwu: @JkSelf Thank you for your suggestion. I have updated the code. Could you please review it again?

gluten-substrait/src/main/scala/org/apache/gluten/extension/columnar/rewrite/AlignExpandOutputTypes.scala
```diff
@@ -0,0 +1,82 @@
...
+object AlignExpandOutputTypes extends RewriteSingleNode {
+  override def isRewritable(plan: SparkPlan): Boolean = {
+    plan match {
+      case _: ExpandExec => true
+      case _ => false
+    }
+  }
+
+  override def rewrite(plan: SparkPlan): SparkPlan = plan match {
+    case expand: ExpandExec if isRewritable(expand) =>
```
JkSelf: Can we remove the `isRewritable` method? The `case expand: ExpandExec` pattern match already guarantees the plan is an `ExpandExec`, making the additional `isRewritable` check redundant

jianzhenwu: Thanks, updated.

JkSelf (approved):
LGTM. Thanks for your work.

Issue #10215, opened 2025-07-18
[VL] Delta Lake write


zhztheplayer:
To add write support for Delta Lake in Velox backend.

Work items:

1. Add native write support for Spark 3.5 + Delta 3.3
  https://github.com/apache/incubator-gluten/pull/10802
  https://github.com/apache/incubator-gluten/pull/10801
3. (TODO) Add native write support for Spark 4.0 + Delta 4.0 
4. (PR pending) Native statistics tracker to avoid C2R overhead
  https://github.com/apache/incubator-gluten/pull/11419
5. (TODO, optional) Add native write support for lower versions (Spark 3.4 + Delta 2.4)
6. Offload [`DeltaOptimizedWriterExec`](https://github.com/delta-io/delta/blob/4b2beb096017d813475f692270bc20c113c0d974/spark/src/main/scala/org/apache/spark/sql/delta/perf/DeltaOptimizedWriterExec.scala) for optimize options
  https://github.com/apache/incubator-gluten/pull/11461

---

PoC at #10216.

Gaps so far:

1. Constraint expression needs to offload (C2R / R2C added otherwise).
2. The overwritten Delta classes in Gluten should be avoided (`DeltaParquetFileFormat` for example).
3. Copied test code from Delta, if there's a way to avoid the practice.

FelixYBW:
@zhztheplayer can you update the gap? we already removed fallback of https://github.com/apache/gluten/issues/11471 and  https://github.com/apache/gluten/issues/11503

sezruby:
**Follow-up / backlog: Parquet page index (ColumnIndex/OffsetIndex) not written by the Velox backend**

Flagging a Velox-writer gap that affects native Delta (and plain Parquet) writes. Velox's Parquet writer wraps Arrow (`velox/dwio/parquet/writer/Writer.cpp` → `getArrowParquetWriterOptions`) and never calls Arrow's `enable_write_page_index()`; `ParquetWriterOptions` exposes no option for it. Arrow defaults page-index writing to off, so files written through the Velox backend carry no page index — unlike vanilla Spark/parquet-mr, which writes it by default (SPARK-26345).

**Impact:** Delta/Parquet files produced by native write can't be page-pruned by readers relying on the page index — including Velox's own reader-side page-level pruning effort (facebookincubator/velox#14195, facebookincubator/velox#18079).

**Proposed split:**
[...]

PR #12016, opened 2026-04-30
[GLUTEN-10215][VL] Delta write: Fix native partitioned layout accounting

malinjawi:

This patch fixes native Delta partitioned write layout accounting in the Velox backend.

The change:
- Writes each native partition stripe as its own accounting unit.
- Enforces `maxRecordsPerFile` within native partition stripes by slicing columnar batches when needed.
- Updates `recordsInFile` by the actual written chunk row count instead of the original input batch row count.
- Adds Delta 4.0 coverage for optimized partitioned native writes with stats enabled.

The same writer accounting fix is applied to Delta 3.3 and Delta 4.0 sources.

The partition-column split-output contract has been moved out of this PR and into a follow-up branch: `malinjawi:feature/delta-native-write-partition-column-output`.

The existing native partitioned writer split incoming Velox batches by partition, but it accounted file layout at the original batch level. That can make partitioned optimized writes violate `maxRecordsPerFile` when a single native partition stripe is larger than the file limit.

No public API change. This improves correctness/layout behavior for native Delta writes.

Built locally and ran:

```bash
JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home \
./dev/run-scala-test.sh --force \
  -Pjava-17,spark-4.0,scala-2.13,backends-velox,hadoop-3.3,spark-ut,delta \
  -pl backends-velox \
  -s org.apache.spark.sql.delta.DeltaNativeWriteSuite \
  -t "native delta optimized partitioned write should collect stats and honor file layout"
```

Result: 1 test passed, 0 failures.

Also ran Spark 3.5 `backends-velox` test compilation and `git diff --check`.

Related issue: #10215


zhztheplayer:
@malinjawi 

> Enforces maxRecordsPerFile within native partition stripes by slicing columnar batches when needed.

> Preserves partition columns in split output only when Delta's write contract includes partition columns in dataColumns.

Thanks. Can we separate the above 2 parts into 2 PRs?

malinjawi:
Thanks @zhztheplayer! Split done.

I force-pushed this PR down to the native partition-stripe layout accounting fix only, and rebased it on current `main`, so the DeltaNativeWriteSuite conflict is gone. The partition-column split-output contract is moved to draft follow-up PR #12069.

This PR now keeps only:
- native partition stripe accounting
- `maxRecordsPerFile` slicing within native partition stripes
- the optimized partitioned write layout/stat coverage

PR #12027, opened 2026-05-03
[CORE][UI] Skip GlutenPlanFallbackEvent posting when gluten is disabled

liujiayi771:
## Summary

When `spark.gluten.enabled=false` (e.g. the vanilla Spark baseline run inside `runQueryAndCompare`), `GlutenQueryExecutionListener` still posted a `GlutenPlanFallbackEvent`. Because no Gluten rules ran, all nodes appeared as fallback with the generic reason *"Gluten does not touch it or does not support it"*, polluting the event stream.

This was an oversight in https://github.com/apache/gluten/pull/11853, which introduced the fallback event posting but did not account for the case where Gluten is disabled entirely via session config. When Gluten is off, every node is trivially "fallback" — posting an event is misleading and noisy.

**Fix:** Early-return before `collectQueryExecutionFallbackSummary` when the session's `spark.gluten.enabled` config is `false`. The config is read directly from `qe.sparkSession.sessionState.conf` via `getConfString` to avoid thread-local `SQLConf` timing issues in the async listener thread.

## Test plan

- [x] Added regression test `FallbackSuite`: verifies no `GlutenPlanFallbackEvent` is emitted when a query runs with `spark.gluten.enabled=false`

### DIFF
--- a/backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/execution/FallbackSuite.scala
@@ -387,4 +387,37 @@ class FallbackSuite extends VeloxWholeStageTransformerSuite with AdaptiveSparkPl
       spark.sparkContext.removeSparkListener(listener)
     }
   }
+
+  test("no fallback event emitted for vanilla Spark execution with gluten disabled") {
+    // Regression test: before the fix, GlutenQueryExecutionListener would post a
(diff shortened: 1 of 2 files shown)


liujiayi771:
cc @FelixYBW @Zouxxyy. Could you help to review?

FelixYBW:
make sense to me. Is failed CI related?

liujiayi771:
> make sense to me. Is failed CI related?

Yes, I will fix them later.

liujiayi771:

philo-he (approved):
LGTM. Thanks. cc @FelixYBW

Issue #1433, opened 2023-04-20
[VL] Support timestamp on Velox backend


rui-mo:
Phase 1:
- [x] Timestamp data conversions, including r2c, c2r etc.
- [x] Timestamp split
- [x] Timestamp Scan (TODO: support timestamp int96 type)
https://github.com/oap-project/velox/pull/205

Phase 2:
- [ ] Timestamp functions
- ToUnixTimestamp
- UnixTimestamp
- GetTimestamp
- Hour
- Minute
- Second
- SecondWithFraction
- SecondsToTimestamp
- MillisToTimestamp
- MicrosToTimestamp
- Cast from timestamp
- Cast to timestamp
- CurrentTimestamp
- Now
- LocalTimestamp
- CurrentBatchTimestamp
- CastTimestampNTZToLong
- FromUnixTime
- TimeAdd
- DatetimeSub
- FromUTCTimestamp
- ToUTCTimestamp
- TimestampAddYMInterval
- ParseToTimestamp
- TruncTimestamp
- MakeTimestamp
- SubtractTimestamps
- SubtractTimestamps
- TimestampAdd
- TimestampDiff

- [ ] Timezone
- ConvertTimezone

Phase 3: 

- [ ] Enable operators including Filter/Project/Sort/Agg/Hash-partition Shuffle.

FelixYBW:
@rui-mo do you have an update of this thread? couple of functions are already supported I think.

rui-mo:
Updated the status on timestamp cast. Thanks.

zhouyuan:
Cc: @Mariamalmesfer

Mariamalmesfer:
did some  analysis on **SecondsToTimestamp**, and it looks like it may already be supported end-to-end.
@rui-mo can please confirm this ?

rui-mo:
> did some analysis on **SecondsToTimestamp**, and it looks like it may already be supported end-to-end. [@rui-mo](https://github.com/rui-mo) can please confirm this ?

@Mariamalmesfer I updated the list, thanks.

Mariamalmesfer:
@rui-mo I checked TruncTimestamp (date_trunc) and it looks supported.

There is already a test in DateFunctionsValidateSuite:
https://github.com/apache/incubator-gluten/blob/e3ba96191200878a4da8ed587b8d51562dc080c4/backends-velox/src/test/scala/org/apache/gluten/functions/DateFunctionsValidateSuite.scala#L131

[...]

PR #12029, opened 2026-05-03
[GLUTEN-1433][VL][FOLLOWUP] Fix TimestampNTZ schema validation never matching due to Scala object simpleName suffix

liujiayi771:
## Summary

`isPrimitiveType` in `VeloxValidatorApi` uses `dt.getClass.getSimpleName == "TimestampNTZType"` to check for TimestampNTZ type. However, `TimestampNTZType` is a Scala `case object`, so `getClass.getSimpleName` returns `"TimestampNTZType$"` (with a trailing `$`), causing the check to always fail. This means `validateSchema` rejects TimestampNTZ even when `enableTimestampNtzValidation` is set to `false`.

**Fix:** Use `dt.catalogString == "timestamp_ntz"` instead, which is reliable regardless of how the type is defined in Scala.

**Test:** Added a test in `VeloxParquetDataTypeValidationSuite` to verify that `validateSchema` correctly accepts/rejects TimestampNTZ based on the `enableTimestampNtzValidation` config.

Related issue: #1433

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxValidatorApi.scala
@@ -111,12 +111,7 @@ object VeloxValidatorApi {
           StringType | BinaryType | _: DecimalType | DateType | TimestampType |
           YearMonthIntervalType.DEFAULT | NullType =>
         true
-      case dt
-          if !enableTimestampNtzValidation &&
-            dt.getClass.getSimpleName == "TimestampNTZType" =>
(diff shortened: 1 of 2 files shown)


liujiayi771:
cc @rui-mo @Mariamalmesfer. Could you help to review?

Mariamalmesfer (approved):
LGTM
@rui-mo can you take a look?

zml1206 (approved):
lgtm

rui-mo (approved):
Thanks @liujiayi771. This is a known issue, and I made a fix in our dev branch: https://github.com/apache/gluten/tree/ts_ntz_dev.

Issue #12025, opened 2026-05-03
[VL][Delta] Track native Delta writer improvements and optimization support


malinjawi:
## Goal

Track Delta-only work to improve native Gluten/Velox support for Delta writer and table optimization paths.

This issue is organized by feature-sized work areas. Each top-level task should map to one reviewable PR or a small stack of tightly related PRs.

## Related Work

- [x] #11419 Delta write native statistics tracker to eliminate C2R overhead
- [ ] #12016 Delta native partitioned layout accounting
- [ ] #12024 Delta OPTIMIZE compaction command transaction offload

## Feature Tracks

- [ ] **OPTIMIZE Compaction**

  Native support for Delta `OPTIMIZE` compaction/bin-packing command paths.

  Scope:

  - offload `OPTIMIZE` command transactions through GlutenOptimisticTransaction
  - cover path-based and table-name `OPTIMIZE` forms
  - cover `OPTIMIZE ... WHERE` partition predicates
  - keep OPTIMIZE read/shuffle/write native where supported
  - validate returned OPTIMIZE metrics and file statistics
  - benchmark compaction on small-file Delta tables

  Related PR:

  - #12024 Delta OPTIMIZE compaction command transaction offload

  Expected coverage:

  - path-based `OPTIMIZE`
  - table-name `OPTIMIZE`
  - `OPTIMIZE ... WHERE partition_predicate`
  - native-write-disabled fallback
  - data correctness before and after compaction
  - Delta log add/remove-file metadata correctness

- [ ] **Optimized Write**

  Native support and correctness hardening for Delta optimized write paths.

  Scope:

  - verify native behavior when `delta.autoOptimize.optimizeWrite` is enabled
  - verify native behavior when `spark.databricks.delta.optimizeWrite.enabled` is enabled
  - verify DataFrameWriter option `optimizeWrite` behavior
  - cover non-partitioned optimized writes
  - cover partitioned optimized writes
  - validate output file sizing and partition layout metadata
  - reduce unnecessary columnar-to-row transitions in write, stats, and commit paths

  Related PRs:

  - #11419 native Delta write statistics tracker
  - #12016 native partitioned layout accounting

  Expected coverage:

  - non-partitioned append and overwrite
  - partitioned append and overwrite
  - optimized-write table property, SQL conf, and writer option
  - partition values in add-file metadata
  - min/max/nullCount stats in add-file metadata
  - native and fallback plan assertions

- [ ] **OPTIMIZE ZORDER**

[...]

PR #12030, opened 2026-05-03
[GLUTEN-10215][VL] Delta write: Native top-level NOT NULL checks

malinjawi:
What changes were proposed in this pull request?
This patch keeps simple Delta write invariants native in the Velox backend.

The change:

- Adds `GlutenDeltaInvariantChecker` for Delta 3.3 and Delta 4.0 sources.
- Validates top-level `NOT NULL` constraints directly on Velox columnar batches before write.
- Avoids `DeltaInvariantCheckerExec` and `ColumnarToRow` for supported NOT NULL-only writes.
- Keeps unsupported constraints conservative: `CHECK` constraints and nested NOT NULL constraints still use Delta's row checker.
- Adds a Velox JNI helper to detect nulls in selected columns, using cached `nullCount` when available.
- Adds Spark 3.5 and Spark 4.0 tests for native NOT NULL checks, fallback behavior, and violation reporting.

This is independent from #12024 and follows the Delta write C2R reduction work in #11419 and #12016.

Why are the changes needed?
Delta writes with table invariants currently go through `DeltaInvariantCheckerExec`, which is row-based. Even simple top-level `NOT NULL` constraints can introduce a C2R transition in an otherwise native Delta write path.

This patch handles the safe common case natively while preserving Delta's existing fallback behavior for unsupported constraints.

Does this PR introduce any user-facing change?
No public API change. Delta constraint behavior is preserved; supported NOT NULL-only writes can stay native.

How was this patch tested?
Built locally and ran:

- Spark 3.5 `DeltaNativeWriteInvariantSuite`: passed
- Spark 4.0 `DeltaNativeWriteSuite`: passed
- Spark 4.0 focused NOT NULL tests: passed
- C++ Velox build: passed
- JNI symbol verified in `libvelox.dylib`
- Spark 3.5 and Spark 4.0 spotless/checkstyle/scalastyle: passed
- `git diff --check`: passed

Covered cases:

- native Delta write checks top-level NOT NULL without `DeltaInvariantCheckerExec`
- native NOT NULL path avoids `ColumnarToRow`
- CHECK constraints keep `DeltaInvariantCheckerExec`
- NOT NULL violations report `InvariantViolationException`

I also ran a targeted local benchmark for append workloads comparing the native top-level NOT NULL path with an equivalent unsupported CHECK fallback path.

Workload: 2,000,000 rows, 14 columns, 3 append iterations.

| path | avg ms |
| --- | ---: |
| native top-level NOT NULL checker | 1713 |
| row CHECK fallback path | 1664 |

Excluding the first warm-up append:

| path | avg ms |
| --- | ---: |
| native top-level NOT NULL checker | 1605 |
| row CHECK fallback path | 1647 |

The benchmark is effectively neutral because Delta write setup, Parquet output, and commit/log work dominate this microbenchmark. This PR is mainly native write correctness/posture work: it removes a row invariant operator and C2R transition from a common constrained Delta write path.

Related issue: #10215

Tracked by #12025

### DIFF
--- a/backends-velox/src-delta33/main/scala/org/apache/spark/sql/delta/constraints/GlutenDeltaInvariantChecker.scala
+++ b/backends-velox/src-delta33/main/scala/org/apache/spark/sql/delta/constraints/GlutenDeltaInvariantChecker.scala
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.delta.constraints
+
+import org.apache.gluten.columnarbatch.VeloxColumnarBatches
+import org.apache.gluten.execution.{PlaceholderRow, TerminalRow}
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.delta.constraints.Constraints.NotNull
+import org.apache.spark.sql.delta.schema.{DeltaInvariantViolationException, SchemaUtils}
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+/**
+ * Native write-time invariant checker for constraints that can be validated without converting
+ * Velox batches back to Spark rows.
+ */
+private[delta] case class GlutenDeltaInvariantChecker private (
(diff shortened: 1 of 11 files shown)


backends-velox/src-delta33/main/scala/org/apache/spark/sql/delta/constraints/GlutenDeltaInvariantChecker.scala
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.spark.sql.delta.schema.{DeltaInvariantViolationException, SchemaUtils}
+import org.apache.spark.sql.vectorized.ColumnarBatch
+
+/**
+ * Native write-time invariant checker for constraints that can be validated without converting
+ * Velox batches back to Spark rows.
+ */
+private[delta] case class GlutenDeltaInvariantChecker private (
+    notNullConstraints: Seq[(Int, NotNull)])
+  extends Serializable {
```
zhztheplayer, line 34: Compared to adding a runtime check, can we offload https://github.com/delta-io/delta/blob/ff715542a7b5b3aa6f0809b7a2c9b1760566ef60/spark/src/main/scala/org/apache/spark/sql/delta/constraints/CheckDeltaInvariant.scala#L35 to Velox instead?

zhztheplayer (commented):
Just a initial design question atm.

felipepessoto:
@malinjawi the PR has been auto-closed, isn't it the plan to merge to merge it anymore?

Thanks.

malinjawi:
> @malinjawi the PR has been auto-closed, isn't it the plan to merge to merge it anymore?
> 
> Thanks.

Hey @felipepessoto! You are right this PR should be revived. I'll raise a new one and tag you once its up.

Issue #12034, opened 2026-05-04
Fix str_to_map MAP_KEY_DEDUP_POLICY comparison for Spark 4.1 compatibility


surendralilhore:
### Backend

VL (Velox)

### Bug description

Starting from Spark 4.1, [SQLConf.get.getConf(SQLConf.MAP_KEY_DEDUP_POLICY)](https://github.com/apache/spark/blob/3c1a8423d8d330facbfd25b55c80fc01c32f357c/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L5277) returns an enum value instead of a String. The existing comparison against the `.toString` of the `EXCEPTION` variant would always evaluate to not-equal (comparing enum != String), incorrectly rejecting `str_to_map` offloading even when the policy is set to `EXCEPTION`.

Add `.toString` on the left side so both operands are compared as Strings, restoring correct behavior across all Spark versions.

### Gluten version

main branch

### Spark version

spark-4.0.x

### Relevant logs

```bash

```

PR #12035, opened 2026-05-04
[GLUTEN-12034][VL] Fix str_to_map MAP_KEY_DEDUP_POLICY comparison for Spark 4.1 compatibility

surendralilhore:
Starting from Spark 4.1, [SQLConf.get.getConf(SQLConf.MAP_KEY_DEDUP_POLICY)](https://github.com/apache/spark/blob/3c1a8423d8d330facbfd25b55c80fc01c32f357c/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L5277) returns an enum value instead of a String. The existing comparison against the `.toString` of the `EXCEPTION` variant would always evaluate to not-equal (comparing enum != String), incorrectly rejecting `str_to_map` offloading even when the policy is set to `EXCEPTION`.

Add `.toString` on the left side so both operands are compared as Strings, restoring correct behavior across all Spark versions.

Fix the code but to make code compatible with Spark 4.1

## How was this patch tested?

Added new UT test to verify the fix.

GitHub Copilot CLI (powered by Claude Opus 4.6)

Related issue: #12034


liuneng1994 (approved):
LGTM

PR #12038, opened 2026-05-05
[VL] Fix json_tuple rewrite producing incompatible JSON path in fallback scenarios

Zouxxyy:

When `json_tuple` is rewritten to `get_json_object` calls by `PullOutGenerateProjectHelper.pullOutPreProject`, the generated JSON path used bare bracket notation `$[key]` (without quotes). This works fine in Velox/simdjson, but when the expression falls back to Spark JVM execution (e.g., `get_json_object` is blacklisted or validation fails), Spark's `JsonPathParser` rejects `$[key]` and returns NULL.

This PR changes the generated path format from `$[key]` to `$['key']` (single-quoted bracket notation), which is accepted by both Velox/simdjson and Spark JVM's JsonPathParser.

**Root cause:**
- Velox's `JsonPathNormalizer` normalizes `$['key']` → `$[key]` internally for simdjson, so both forms work.
- Spark JVM only accepts `$.name` or `$['name']`, and rejects `$[name]` (bare brackets without quotes), returning NULL directly.

## How was this patch tested?

Added unit tests in `MiscOperatorSuite` covering:
- Basic single key extraction with fallback
- Dot-containing field names (e.g., `a.b`) — the core scenario for bracket notation
- Multiple keys extraction
- Non-existent keys returning null
- Mix of existing and non-existing keys
- NULL JSON input handling

Yes

Generated-by: Qoder


Zouxxyy:
The failure ci should be unrelated, CC @lyy-pineapple @philo-he for a look thanks

philo-he:
Thank you for the PR.
`MiscOperatorSuite.scala` generally holds some basic tests for expressions/operators. I would recommend to move the proposed test to a new test suite, named like `JsonTuplePathRewriteSuite.scala`. Thanks.

philo-he (approved):
LGTM. Thank you!

PR #12001, opened 2026-04-28
[VL][Delta] Add roaring bitmap facilities for Delta DV support

malinjawi:
## What changes are proposed in this pull request?

This PR is the first step in the split [Delta deletion-vector (DV) stack](https://github.com/apache/gluten/pull/11963).

It introduces the native roaring bitmap facilities that later PRs use for Delta DV payload handling, without adding Delta scan integration yet.

Main changes:
- add `FindRoaring.cmake` to discover CRoaring from an existing Velox build, pkg-config, or `FetchContent`
- wire `roaring` into the native Velox build
- add `gluten::delta::RoaringBitmapArray`, a small 64-bit roaring bitmap wrapper for Delta DV payloads
- use the same payload shape expected by the later native DV reader:
  - 4-byte magic number
  - CRoaring portable `Roaring64Map` serialization
- add focused native unit coverage for:
  - serialization / deserialization round-trip
  - invalid magic-number rejection

This PR is intentionally infrastructure-only:
- no Delta scan offload behavior changes yet
- no JVM-side Delta scan integration yet

Those pieces will be added in follow-up stacked PRs.

issue #11901.

## How was this patch tested?

Added focused native test coverage in:
- `cpp/velox/compute/delta/tests/RoaringBitmapArrayTest.cpp`

Covered cases:
- round-trip serialization / deserialization of 64-bit bitmap entries
- rejection of malformed payloads with an invalid magic number

## Was this patch authored or co-authored using generative AI tooling?

Generated-by: IBM BOB

PR #12040, opened 2026-05-05
[VL] Delta: Add native Delta DV reader support

malinjawi:

This PR is the second step in the split Delta deletion-vector (DV) stack, following #12001.

It adds the native Velox-side Delta DV reader layer that consumes the roaring bitmap payload facilities introduced by #12001, without adding the JVM-side Delta scan metadata handoff yet.

Main changes:

- add a native Delta connector and data source backed by the Hive connector/data source infrastructure
- register a scoped Delta connector alongside the existing scoped Hive connector for each Velox runtime
- add Delta split metadata types for:
  - deletion-vector descriptors
  - protocol metadata
  - file statistics used for DV validation
  - serialized split payload buffer views
- add `DeltaDeletionVectorReader` to load materialized Delta DV payloads using `RoaringBitmapArray`
- add `DeltaSplitReader` to validate DV protocol/statistics metadata and apply row-index filtering semantics
- add focused native unit coverage for connector setup, split metadata, and deletion-vector reader behavior

This PR is intentionally native-reader only:

- no JVM-side Delta scan metadata handoff yet
- no end-to-end Delta scan offload behavior change yet

Those pieces will be added in follow-up split PRs.

issue #11901.

How was this patch tested?

Added focused native test coverage in:

- `cpp/velox/compute/delta/tests/DeltaConnectorTest.cpp`
- `cpp/velox/compute/delta/tests/DeltaSplitTest.cpp`
- `cpp/velox/compute/delta/tests/DeltaDeletionVectorReaderTest.cpp`

Covered cases:

- Delta connector configuration and connector properties
- split-carried deletion-vector descriptors and logical row-count accounting
- loading materialized DV payloads from `RoaringBitmapArray`
- row deletion checks and keep/drop filter decisions
- empty payload handling and invalid payload rejection
- protocol/statistics validation for DV-bearing splits

Validation run:

- fork preview CI against `malinjawi/incubator-gluten:main` on the combined PR2 branch: all checks passed after rerunning two infra-flaky jobs
- local `git diff --check upstream/main...HEAD`
- local clang-format pass with `/opt/homebrew/opt/llvm@15/bin/clang-format` over changed C++ files

Generated-by: IBM BOB


cpp/velox/compute/delta/DeltaSplit.h
```diff
@@ -0,0 +1,151 @@
...
+  bool supportsDeletionVectors() const {
+    if (minReaderVersion < 3 || minWriterVersion < 7) {
+      return false;
+    }
+    if (!readerFeatures.has_value()) {
+      return false;
+    }
+    return std::find(readerFeatures->begin(), readerFeatures->end(), "deletionVectors") != readerFeatures->end();
+  }
+};
```
zhztheplayer: In which case do we need this struct and the validation logics? Won't we only use native DV features when the Java side feature exists?

cpp/velox/compute/delta/tests/DeltaConnectorTest.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "compute/delta/DeltaConnector.h"
+#include "velox/connectors/Connector.h"
+#include "velox/connectors/hive/HiveConfig.h"
+
+namespace gluten::delta {
+
+namespace {
+
+class DeltaConnectorTest : public ::testing::Test {
```
zhztheplayer, line 51: Can we also add some test cases that are more E2E? E.g., read a file with a passed-in DV, then verify whether the unwanted rows are filtered out?

cpp/velox/compute/delta/DeltaDeletionVectorReader.h
```diff
@@ -0,0 +1,106 @@
...
+///   }
+/// @endcode
+class DeltaDeletionVectorReader {
+ public:
+  DeltaDeletionVectorReader() = default;
+
+  /// Loads a deletion vector from an already decoded serialized Delta payload.
+  void loadSerializedDeletionVector(
+      std::string_view serializedPayload,
+      std::optional<uint64_t> expectedCardinality = std::nullopt);
```
zhztheplayer, line 74: Can we add comments explaining `expectedCardinality`?

zhztheplayer (commented):
Thanks.

cpp/velox/compute/delta/DeltaSplit.h
malinjawi: Good point @zhztheplayer, this was carried over from the earlier [delta-dv-read-foundation](https://github.com/apache/gluten/pull/11900) branch, where native had a broader/future-scoped DV descriptor path.

For this split PR, native only consumes JVM-materialized DV payloads. The Java-side Delta integration is responsible for gating DV support before handing the payload to native, so this protocol struct and validation path are not needed in the current scope.

I removed `DeltaProtocolInfo` and the related validation logic from this PR. Native now keeps only the checks it actually needs here: materialized payload presence, optional cardinality validation, and file statistics consistency. 

I will take a note of that and maybe we can reintroduce protocol validation later if/when native directly resolves Delta DV descriptors.

cpp/velox/compute/delta/tests/DeltaConnectorTest.cpp
malinjawi: Good catch @zhztheplayer 

Added a native test for this:
https://github.com/apache/gluten/pull/12040/changes#diff-c7e9378ffacac4c918121006a2b4606297a5a321fa34ab432e82ef9748aead4bR148-R167

The test writes a small DWRF file, scans it through the Delta connector with a materialized DV payload passed in the split, and verifies that rows whose file positions are in the DV are filtered out from the scan result.

cpp/velox/compute/delta/DeltaDeletionVectorReader.h
malinjawi: Thanks @zhztheplayer!

Updated the comment to explain `expectedCardinality` more directly:

https://github.com/apache/gluten/pull/12040/changes/BASE..e2aeaed2de9e03de82f61968faa83c51c5bf7c7e#diff-31100294214221e6cd8528fb9ae69506a6a3d6e45e4f24b126146b8db3d1863cR67-R76

cpp/velox/compute/delta/tests/DeltaConnectorTest.cpp
```diff
@@ -0,0 +1,176 @@
...
+                        .outputType(rowType)
+                        .endTableScan()
+                        .planNode();
+
+  const auto payload = createSerializedPayload({2, 5, 8});
+  const auto split = makeDeltaSplit(file->getPath(), payload, 3);
+  const auto expected = makeRowVector({"id"}, {makeFlatVector<int64_t>({0, 1, 3, 4, 6, 7, 9})});
+
+  facebook::velox::exec::test::AssertQueryBuilder(plan).split(split).assertResults(expected);
+}
```
zhztheplayer, line 170: Can we cover both `if-contained` and `if-not-contained` in the test?

zhztheplayer (commented):
Thanks for iterating @malinjawi. Just one comment about test remaining.

cpp/velox/compute/delta/tests/DeltaConnectorTest.cpp
malinjawi: Thanks @zhztheplayer the added test if-contained and if-not-contained.

zhztheplayer (approved):

philo-he:
@malinjawi, I noticed that a few files added in this PR carry the `Copyright (c) Facebook, Inc. and its affiliates.` header. Does this source code originate from Velox itself, or perhaps from the Velox fork branch but not actually belonging to the Velox project?

malinjawi:
> @malinjawi, I noticed that a few files added in this PR carry the `Copyright (c) Facebook, Inc. and its affiliates.` header. Does this source code originate from Velox itself, or perhaps from the Velox fork branch but not actually belonging to the Velox project?

Thanks @philo-he  for catching this.

It's your second guess: the code doesn't belong to the Velox project. I originally wrote it on a Velox fork branch under `velox/connectors/hive/delta/`, where Velox's file template carries the Copyright (c) Facebook, Inc. and its affiliates. header. When I ported it into Gluten the ASF header was prepended on top instead of replacing that block, so both ended up stacked. It was never merged into `facebookincubator/velox`, so the Meta copyright doesn't apply here.

I'll send a minor PR removing the stray block from the 15 affected files under `cpp/velox/compute/delta/ `(13 from this PR, plus RoaringBitmapArray.{h,cpp} from #12001), leaving the ASF header only — which is what the Delta files added in the later PRs already use.

PR #12046, opened 2026-05-06
[GLUTEN-12044][CH] Fix build on arm

zhanglistar:
Fix issue https://github.com/apache/gluten/issues/12044.
Fix an ARM64/NEON build failure in `LocalDigitsToAsciiDigitForDate`. The previous implementation called `simd8_u8::load(...).is_ascii()`, but on ARM64 the simdjson `simd8<uint8_t>::load` returns the underlying vector type, so the member call is invalid. 

This PR replaces the ASCII check with a portable SIMD predicate using `any_bits_set_anywhere(simd8_u8(0x80))` to detect any byte with the high bit set, preserving the original intent while compiling across SIMD backends.

## How was this patch tested?

- Rebuilt the failing object/target in the ARM64 cross-compilation environment and confirmed it compiles successfully.
- (Optional) Ran a full build to ensure no regressions.

Generated-by: Cursor (GPT-5.2)

### DIFF
--- a/cpp-ch/local-engine/Functions/LocalDigitsToAsciiDigitForDate.cpp
+++ b/cpp-ch/local-engine/Functions/LocalDigitsToAsciiDigitForDate.cpp
@@ -219,7 +219,8 @@ private:
         size_t i = 0;
         for (; i + kBlockSize <= size; i += kBlockSize)
         {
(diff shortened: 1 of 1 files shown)


zzcclp (approved):
LGTM, build with the arm platform passed.

zzcclp:
image

PR #12048, opened 2026-05-07
[GLUTEN-11918][VL] Fall back Cast when per-expression timezone differs from session timezone

Yao-MR:

When CastTransformer passes expressions through Substrait, it does not carry
per-expression timezone information. Gluten/Velox only uses the session-level
timezone for cast operations. This causes incorrect results when the
per-expression timezone differs from the session timezone and the cast involves
TimestampType (e.g., timestamp to string formatting in ToPrettyString).

This patch adds a timezone consistency check in ExpressionConverter before
creating CastTransformer. If the per-expression timezone differs from the
session timezone and the cast involves TimestampType, a GlutenNotSupportException
is thrown to fall back to Spark native execution.

## How was this patch tested?

Enabled GlutenToPrettyStringSuite for both Spark 4.0 and Spark 4.1 which was
previously disabled due to 1 test failure caused by this timezone issue.

ISSUE: https://github.com/apache/gluten/issues/11918

### DIFF
--- a/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
@@ -932,8 +932,8 @@ object ExpressionConverter extends SQLConfHelper with Logging {
   }
 
   /**
-   * Recursively checks whether the given data type is, or contains, a TimestampType,
-   * including nested array/map/struct/UDT element types.
+   * Recursively checks whether the given data type is, or contains, a TimestampType, including
+   * nested array/map/struct/UDT element types.
    */
   private def involvesTimestampType(dataType: DataType): Boolean = dataType match {
     case TimestampType => true
--- a/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
+++ b/gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
@@ -470,13 +470,13 @@ object ExpressionConverter extends SQLConfHelper with Logging {
         // Gluten uses session-level timezone for cast. If the per-expression timezone
         // differs from session timezone and the cast involves timestamp type, we must
         // fall back to Spark native execution to ensure correctness.
+        // Note: Spark Cast applies zoneId recursively to array/map/struct elements,
+        // so we must check nested types as well.
         c.timeZoneId.foreach {
           tz =>
             val sessionTz = SQLConf.get.sessionLocalTimeZone
             if (tz != sessionTz) {
-              val involvesTimestamp = c.child.dataType == TimestampType ||
-                c.dataType == TimestampType
-              if (involvesTimestamp) {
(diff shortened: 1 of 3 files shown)


gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -467,6 +467,22 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
       case c: Cast =>
+        // Gluten uses session-level timezone for cast. If the per-expression timezone
+        // differs from session timezone and the cast involves timestamp type, we must
+        // fall back to Spark native execution to ensure correctness.
+        c.timeZoneId.foreach {
+          tz =>
+            val sessionTz = SQLConf.get.sessionLocalTimeZone
+            if (tz != sessionTz) {
+              val involvesTimestamp = c.child.dataType == TimestampType ||
+                c.dataType == TimestampType
```
zml1206: Thanks. This only checks the top-level child/output type, so casts like ArrayType(TimestampType) -> ArrayType(StringType), MapType(TimestampType, ...), or structs containing timestamp still go to Velox when the Cast has a per-expression timezone different from the session timezone. Spark Cast applies the same zoneId recursively to array/map/struct elements, so these cases can still return the session-timezone result instead of the expression-timezone result.

Yao-MR: thanks for review the key point, add other complex types such as ArrayType, MapType,StructType,UserDefinedType to recursive detect the timestamp

gluten-substrait/src/main/scala/org/apache/gluten/expression/ExpressionConverter.scala
```diff
@@ -467,6 +467,34 @@ object ExpressionConverter extends SQLConfHelper with Logging {
...
+              def containsTimestamp(dataType: DataType): Boolean = dataType match {
+                case TimestampType => true
+                case a: ArrayType => containsTimestamp(a.elementType)
+                case m: MapType =>
+                  containsTimestamp(m.keyType) || containsTimestamp(m.valueType)
+                case s: StructType => s.exists(f => containsTimestamp(f.dataType))
+                case udt: UserDefinedType[_] =>
+                  containsTimestamp(udt.sqlType)
+                case _ => false
+              }
```
zml1206: Could we move `containsTimestamp` out of the `Cast` branch as a private helper method?
The recursive type check is independent from the local `Cast` state, and keeping it as a small private helper would make the `Cast` branch easier to read. It also avoids redefining the local function on every visit and makes the logic easier to reuse if another timezone-sensitive expression needs the same nested timestamp check later.

Yao-MR: sure, and extract the recursive timestamp-type check as a private helper
  `involvesTimestampType`,no functional change; pure readability refactor.

zml1206 (approved):

PR #12049, opened 2026-05-07
[CORE] Deduplicate fallback reason when merging Appendable tags

liujiayi771:
## Summary

When AQE re-runs columnar rules across stages, `GlutenFallbackReporter` repeatedly calls
`FallbackTags.add` on the same shared `logicalLink` with the same reason. The previous merge
logic in `FallbackTags.add` unconditionally concatenated `Appendable` reasons, producing strings
like `"r; r; r; ..."` that grow with every AQE iteration. This is especially noticeable when users
manually fall back specific node types and the same fixed reason is added many times.

The relevant comment at `GlutenFallbackReporter.scala:66-71` already explains the trigger:

> With in next round stage in AQE, the physical plan would be a new instance that can not preserve
> the tag, so we need to set the fallback reason to logical plan ... If a logical plan mapping to
> several physical plan, we add all reason into that logical plan to make sure we do not lose any
> fallback reason.

This PR makes `FallbackTag.Appendable` store fallback reasons in a `HashSet[String]` and merge
`Appendable` tags with set union. `reason()` joins the set only when the reason text is reported,
so duplicate reasons are removed precisely without relying on substring checks.

The reason order is not significant here; the important behavior is that the reported text contains
each distinct fallback reason once.

## Drive-by: stabilize FallbackSuite

While testing, the existing test `no fallback event emitted for vanilla Spark execution with gluten
disabled` (added in #12027) turned out to be flaky. Spark's `LiveListenerBus` dispatches events
asynchronously, so `GlutenPlanFallbackEvent`s posted by the previous test in the suite can still be
delivered to the listener registered by the next test, contaminating its `events` buffer. The dedup
change subtly shifted CPU/scheduling timing and exposed it.

Fix is bundled in this PR since it's directly triggered by these changes:

- Added `GlutenSuiteUtils.withFallbackEventListener` that drains the bus before registering the
  listener and removes it afterwards.
- All three event-listener tests in `FallbackSuite`, which previously duplicated the same listener
  boilerplate, now go through this helper.

## Test plan

- [x] Added `FallbackTagSuite` in `gluten-core` covering:
  - repeated `add` of the same reason does not grow the reason string
  - distinct `Appendable` reasons are preserved
  - reasons with substring overlap are still treated as distinct reasons
- [x] Verified the substring-overlap test fails with the previous `contains`-based implementation
  and passes after switching `Appendable` to `HashSet`-based dedup.
- [x] `mvn -pl gluten-core -DwildcardSuites=org.apache.gluten.extension.columnar.FallbackTagSuite test`
- [x] `FallbackSuite` continues to pass with the new helper; the previously flaky test is now stable
  across reruns.

### DIFF
--- a/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/FallbackTag.scala
+++ b/gluten-core/src/main/scala/org/apache/gluten/extension/columnar/FallbackTag.scala
@@ -22,6 +22,8 @@ import org.apache.spark.sql.execution.SparkPlan
 
 import org.apache.commons.lang3.exception.ExceptionUtils
 
+import scala.collection.immutable.HashSet
+
 sealed trait FallbackTag {
   val stacktrace: Option[String] =
     if (FallbackTags.DEBUG) {
@@ -33,8 +35,16 @@ sealed trait FallbackTag {
 
 object FallbackTag {
 
-  /** A tag that stores one reason text of fall back. */
-  case class Appendable(override val reason: String) extends FallbackTag
+  /** A tag that stores distinct reason texts of fall back. */
+  case class Appendable private (private val reasonSet: HashSet[String]) extends FallbackTag {
+    override def reason(): String = reasonSet.mkString("; ")
+
+    def append(other: Appendable): Appendable = Appendable(reasonSet ++ other.reasonSet)
+  }
+
+  object Appendable {
+    def apply(reason: String): Appendable = new Appendable(HashSet(reason))
+  }
 
   /**
    * A tag that stores reason text of fall back. Other reasons will be discarded when this tag is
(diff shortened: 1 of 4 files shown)


liujiayi771:
before:
image

after:
image

gluten-core/src/main/scala/org/apache/gluten/extension/columnar/FallbackTag.scala
```diff
@@ -93,7 +93,13 @@ object FallbackTags {
...
           exclusive
         case (l: FallbackTag.Appendable, r: FallbackTag.Appendable) =>
-          FallbackTag.Appendable(s"${l.reason}; ${r.reason}")
+          if (l.reason == r.reason || l.reason.contains(r.reason)) {
+            l
+          } else if (r.reason.contains(l.reason)) {
+            r
+          } else {
+            FallbackTag.Appendable(s"${l.reason}; ${r.reason}")
+          }
```
zhztheplayer: How about storing a hash set in `FallbackTag.Appendable`? Could be faster and preciser.

liujiayi771: @zhztheplayer Good idea, using a hash set is cleaner and more precise. Updated, thanks!

philo-he (approved):
LGTM. Thanks.

zhztheplayer (approved):

Issue #10134, opened 2025-07-07
[VL] Add ANSI mode support


PHILO-HE:
### Description

In this issue, we will list and track all tasks for ANSI mode support.

There are two Spark configurations directly related to ANSI usage.

* `spark.sql.ansi.enabled` (default is true since Spark 4.0)
* `spark.sql.storeAssignmentPolicy` (default is ANSI since Spark 3.0)

### Tasks

- [x] https://github.com/apache/incubator-gluten/issues/10387
- [ ] cast string to boolean
https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L701

- [ ] Cast decimal to string
// In ANSI mode, Spark always use plain string representation on casting Decimal values
// as strings. Otherwise, the casting is using `BigDecimal.toString` which may use scientific
// notation if an exponent is needed.
https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L678

- [ ] cast string to timestamp
https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L733

- [ ] cast String to timestampNTZ
https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L775

- [ ] cast float/double to timestamp
https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L758
https://github.com/apache/spark/blob/e221b56be7b6d9e48e107fc4d1cf0c15f02700f8/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/Cast.scala#L765

- [ ] cast string to date
[...]

PR #12051, opened 2026-05-07
[GLUTEN-10134][VL] Preserve store assignment cast modes (`STORE_ASSIGNMENT_POLICY` defaults to ANSI)

ReemaAlzaid:

This PR preserves Spark cast eval mode when Gluten translates casts to Velox, so store assignment casts keep ANSI behavior even when `spark.sql.ansi.enabled=false`.
#10134
- Preserves Spark cast mode in Gluten Substrait.
- Adds Velox ANSI/legacy cast handling.
- Keeps `TRY_CAST` as Velox try-cast.
- Tests default `STORE_ASSIGNMENT_POLICY=ANSI`.


ReemaAlzaid:
Could you please review this? @rui-mo

rui-mo (commented):
Do we have any background to support this policy, and would it make sense to add some tests covering the configurable behaviors under different policy settings?

cpp/velox/operators/functions/SparkCastModeSpecialForms.cc
```diff
@@ -0,0 +1,115 @@
...
+using namespace facebook::velox;
+using facebook::velox::functions::sparksql::SparkCastExpr;
+using facebook::velox::functions::sparksql::SparkCastHooks;
+
+bool isIntegralType(const TypePtr& type) {
+  return type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+      type == BIGINT();
+}
+
+bool isAnsiSupported(const TypePtr& fromType, const TypePtr& toType) {
```
philo-he: There's also a check on the Velox side. Will we maintain the ANSI support check only here?

gluten-ut/spark40/src/test/scala/org/apache/spark/sql/sources/GlutenInsertSuite.scala
```diff
@@ -124,6 +124,25 @@ class GlutenInsertSuite
     }
   }
 
+  testGluten("storeAssignmentPolicy default ANSI is independent from ANSI mode") {
```
philo-he: This doesn't appear to be a variant of a Spark test. Can we move it to a new suite under `backends-velox/`?

gluten-ut/spark40/src/test/scala/org/apache/spark/sql/sources/GlutenInsertSuite.scala
```diff
@@ -124,6 +124,25 @@ class GlutenInsertSuite
...
+    withTable("store_assignment_ansi") {
+      withSQLConf(SQLConf.ANSI_ENABLED.key -> "false") {
+        assert(SQLConf.get.storeAssignmentPolicy == SQLConf.StoreAssignmentPolicy.ANSI)
+
+        spark.sql("CREATE TABLE store_assignment_ansi (c INT) USING PARQUET")
+        intercept[Exception] {
+          spark.sql("INSERT INTO store_assignment_ansi SELECT '2147483648'").collect()
```
philo-he: Can we add a check for the exception message to confirm the expected exception is thrown?

ReemaAlzaid:
> Do we have any background to support this policy, and would it make sense to add some tests covering the configurable behaviors under different policy settings?

My understanding is that Spark applies `spark.sql.storeAssignmentPolicy` during analysis, and Gluten should just preserve the cast mode Spark selected when translating the resulting casts. Otherwise INSERT/store assignment behavior may differ from Spark. I’ll add tests for the configurable policy cases.

cpp/velox/operators/functions/SparkCastModeSpecialForms.cc
ReemaAlzaid: This is intended to mirror the existing Velox ANSI support side check not to replace it, since that check is private today, I added this local helper for the expression level ANSI/legacy cast mode. I’ll add a comment to make sure we keep it aligned with Velox

cpp/velox/operators/functions/SparkCastModeSpecialForms.cc
```diff
@@ -0,0 +1,118 @@
...
+    const core::QueryConfig& config) {
+  return std::make_shared<SparkCastExpr>(
+      type,
+      std::move(input),
+      trackCpuUsage,
+      isTryCast,
+      std::make_shared<SparkCastHooks>(config, allowOverflow));
+}
+
+class SparkAnsiCastCallToSpecialForm : public exec::CastCallToSpecialForm {
```
philo-he: I am still wondering if we can move or keep these code on Velox side, and just call the register API in Gluten C++ code. Could you please clarify?

cpp/velox/operators/functions/SparkCastModeSpecialForms.cc
```diff
@@ -0,0 +1,118 @@
...
+
+bool isIntegralType(const TypePtr& type) {
+  return type == TINYINT() || type == SMALLINT() || type == INTEGER() ||
+      type == BIGINT();
+}
+
+// Keep this in sync with Velox's SparkCastCallToSpecialForm::isAnsiSupported.
+// Velox's helper is private today; this local copy is needed for expression-level
+// ANSI and legacy cast modes.
+bool isAnsiSupported(const TypePtr& fromType, const TypePtr& toType) {
```
philo-he: If we really need it, to reduce code maintenance effort, could we make it public in Velox to allow calling it here?

philo-he (commented):
Thanks for the PR. Two comments, possibly due to missing context on my side.

cpp/velox/operators/functions/SparkCastModeSpecialForms.cc
ReemaAlzaid: Yes. Existing Velox Spark `cast` depends on session `ansiEnabled`, but Gluten needs Substrait’s per expression cast mode. This patch adds thin Gluten registration while reusing Velox cast code. I agree we can move it to Velox and call the registration API from Gluten

cpp/velox/operators/functions/SparkCastModeSpecialForms.cc
ReemaAlzaid: Created the pr here https://github.com/facebookincubator/velox/pull/17687

philo-he:
@ReemaAlzaid, thanks for the update. Could you check if the CI failures are related?

ReemaAlzaid:
The pr got merged https://github.com/facebookincubator/velox/pull/17687 @philo-he

philo-he:
> The pr got merged [facebookincubator/velox#17687](https://github.com/facebookincubator/velox/pull/17687) @philo-he

@ReemaAlzaid, thanks for the update. I noticed the CI failed. Is it because Gluten's referenced Velox branch doesn't yet include that Velox commit? If so, you could reference the PR ID in the following code so Gluten picks up that Velox patch for the CI tests.
https://github.com/apache/gluten/blob/cd994505b08c6ae37ab6813a56c0139b226f4e66/ep/build-velox/src/get-velox.sh#L28

ReemaAlzaid:
Hey @philo-he,  the velox commit is added now and has been merged I just forgot to update the ansi config key class here [Fix Spark ANSI config key in cast tests](https://github.com/apache/gluten/pull/12051/commits/570e3c2e51c9892fa573bb842b8e157c200b427a). So the CI should run fine now

philo-he:
@ReemaAlzaid, thanks for the update. It seems that CI failure is related. Could you check further? Thank you!

```
- storeAssignmentPolicy preserves configured cast modes *** FAILED ***
  executedPlan.exists(((plan: org.apache.spark.sql.execution.SparkPlan) => tag.runtimeClass.isInstance(plan))) was false Expect ProjectExecTransformer exists in executedPlan:
   CommandResult <empty>
     +- Execute InsertIntoHadoopFsRelationCommand file:/__w/gluten/gluten/backends-velox/spark-warehouse/org.apache.gluten.execution.VeloxInsertSuite/store_assignment_legacy, false, Parquet, [path=file:/__w/gluten/gluten/backends-velox/spark-warehouse/org.apache.gluten.execution.VeloxInsertSuite/store_assignment_legacy], Append, `spark_catalog`.`default`.`store_assignment_legacy`, org.apache.spark.sql.execution.datasources.InMemoryFileIndex(file:/__w/gluten/gluten/backends-velox/spark-warehouse/org.apache.gluten.execution.VeloxInsertSuite/store_assignment_legacy), [c]
        +- WriteFiles
           +- *(1) Project [cast(c#242653 as int) AS c#242655]
              +- *(1) ColumnarToRow
                 +- FileScan parquet spark_catalog.default.store_assignment_legacy_src[c#242653] Batched: true, DataFilters: [], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:/__w/gluten/gluten/backends-velox/spark-warehouse/org.apache.glut..., PartitionFilters: [], PushedFilters: [], ReadSchema: struct<c:string> (GlutenQueryTest.scala:427)
```

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxInsertSuite.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  private def insertIntoFrom(target: String, source: String) =
+    spark.sql(s"INSERT INTO $target SELECT c FROM $source")
+
+  private def assertUnsafeCastAnalysisException(
+      fromType: String,
+      toType: String)(f: => Unit): Unit = {
+    val exception = intercept[AnalysisException](f)
+    // Older Spark versions report the types in lower case, e.g. "string to int",
+    // while newer ones quote them in upper case, e.g. "STRING" to "INT".
```
philo-he: Nit:
Could you leave some details about since which Spark versions  upper case is used? Then, we can remove the case conversion when those old Spark versions are deprecated.

You may clarify like this "Since Spark xxx, the types in the exception message are in upper case, e.g. ...."

philo-he:
@ReemaAlzaid, thanks for the iterating. Could you check if the clickhouse backend CI failure is related? You can use this account to login: https://github.com/apache/gluten/blob/82da039b7bc5ba2344f132404c77fe0b06b48bee/docs/get-started/ClickHouse.md?plain=1#L651

If not related, you can re-trigger that CI by posting a comment to this PR page: `Run Gluten Clickhouse CI on x86`.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxInsertSuite.scala
ReemaAlzaid: > Nit: Could you leave some details about since which Spark versions upper case is used? Then, we can remove the case conversion when those old Spark versions are deprecated.
> 
> You may clarify like this "Since Spark xxx, the types in the exception message are in upper case, e.g. ...."

Sure I have added the comment thanks

ReemaAlzaid:
Hey @philo-he thanks for the review! I have addressed the comments and reran the CI

cpp/velox/operators/functions/RegistrationAllFunctions.cc
```diff
@@ -84,6 +85,7 @@ void registerFunctionOverwrite() {
 
 void registerAllFunctions() {
   velox::functions::sparksql::registerFunctions("");
+  velox::functions::sparksql::registerSparkCastModeSpecialForms();
```
philo-he, line 88: `registerFunctions` will call `registerSpecialFormGeneralFunctions`, then call `registerSparkCastModeSpecialForms`. If so, this line should be redundant.

philo-he (approved):
@ReemaAlzaid, thanks for the continued efforts on this.

One last minor comment — please create a separate PR to address it if it makes sense. I'll merge this PR first.

felipepessoto:
@ReemaAlzaid, @philo-he, does this fix #12779? It seems so, the tests are not failing anymore.

ReemaAlzaid:
Hey @felipepessoto 
Yes, this fixes #12779 verified on the latest Delta UT run on main: all 4 failing tests now pass.

Those tests use a LEGACY cast with `spark.sql.ansi.enabled=true`. Before this PR, Gluten picked the cast mode from the session ANSI flag, so the cast threw on overflow instead of returning null like Spark. It also fixed 4 older baseline entries

felipepessoto:
Also fixes #12784

felipepessoto:
@ReemaAlzaid @philo-he could you help with this small follow-up to make the Delta CI green again? #12815

This PR also fixed 4 tests that were in the known issue list and we need to remove.

### Now-passing (global) (4)

```
org.apache.spark.sql.delta.ImplicitMergeCastingSuite#MERGE overflow in WHEN MATCHED THEN UPDATE SET t.value = s.value sourceType: BIGINT, targetType: DECIMAL(7,2) followAnsiEnabled: false, ansiEnabled: true, storeAssignmentPolicy: LEGACY
org.apache.spark.sql.delta.ImplicitMergeCastingSuite#MERGE overflow in WHEN MATCHED THEN UPDATE SET t.value = s.value sourceType: STRING, targetType: INT followAnsiEnabled: false, ansiEnabled: true, storeAssignmentPolicy: LEGACY
org.apache.spark.sql.delta.ImplicitStreamingMergeCastingSuite#Streaming MERGE overflow sourceType: BIGINT, targetType: DECIMAL(7,2) followAnsiEnabled: false, ansiEnabled: true, storeAssignmentPolicy: LEGACY
org.apache.spark.sql.delta.ImplicitStreamingMergeCastingSuite#Streaming MERGE overflow sourceType: STRING, targetType: INT followAnsiEnabled: false, ansiEnabled: true, storeAssignmentPolicy: LEGACY
```

https://github.com/apache/gluten/actions/runs/32101727188

cpp/velox/operators/functions/RegistrationAllFunctions.cc
ReemaAlzaid: https://github.com/apache/gluten/pull/12816

Issue #11921, opened 2026-04-11
[VL] Spark 4.1: Support NullType Parquet read/write (SPARK-54220)


baibaichen:
## Backend
VL (Velox)

**Gluten version**: main branch

## Description

Spark 4.1 added support for reading and writing NullType columns in Parquet (SPARK-54220). Gluten/Velox does not handle NullType in Parquet format.

Spark 4.1 only.

**Parent issue**: #11910 (`[VL] Spark 4.x: Tracking new feature support`)

### Impact

| Suite | Exclude | spark40 | spark41 |
|-------|---------|:-------:|:-------:|
| GlutenParquetIOSuite | SPARK-54220 | 🟢 | 🔴 |

Note: GlutenParquetIOSuite also has struct field excludes tracked under a separate #11550 sub-issue.

### References

- Apache Spark JIRA: [SPARK-54220](https://issues.apache.org/jira/browse/SPARK-54220)

PR #12056, opened 2026-05-08
[GLUTEN-11921] Enable Parquet read/write test for NullType

wecharyu:
Enable parquet NullType read/write test.

With `excludeByPrefix("SPARK-54220")` removed, "SPARK-54220: NullType" test is put back for testing. And "SPARK-54220: vectorized reader: missing all struct fields, struct with NullType only" test is replaced by a variant with a check removed since it is only applicable to vanilla Spark.

This PR depends on https://github.com/facebookincubator/velox/pull/17391.

## How was this patch tested?
Pass UT in `GlutenParquetIOSuite`.

No.

Related issue: #11921

### DIFF
--- a/gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
+++ b/gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
@@ -22,9 +22,6 @@ import org.apache.spark.sql.types.{IntegerType, StringType, StructType}
 
 /** A test suite that tests basic Parquet I/O. */
 class GlutenParquetIOSuite extends ParquetIOSuite with GlutenSQLTestsBaseTrait {
-  override def testNameBlackList: Seq[String] =
-    Seq("SPARK-54220: vectorized reader: missing all struct fields, struct with NullType only")
-
   override protected def testFile(fileName: String): String = {
     getWorkspaceFilePath("sql", "core", "src", "test", "resources").toString + "/" + fileName
   }
--- a/gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
+++ b/gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
@@ -17,14 +17,51 @@
 package org.apache.spark.sql.execution.datasources.parquet
 
 import org.apache.spark.sql._
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.{IntegerType, StringType, StructType}
 
 /** A test suite that tests basic Parquet I/O. */
 class GlutenParquetIOSuite extends ParquetIOSuite with GlutenSQLTestsBaseTrait {
+  override def testNameBlackList: Seq[String] =
+    Seq("SPARK-54220: vectorized reader: missing all struct fields, struct with NullType only")
+
   override protected def testFile(fileName: String): String = {
     getWorkspaceFilePath("sql", "core", "src", "test", "resources").toString + "/" + fileName
   }
 
   override protected def readResourceParquetFile(name: String): DataFrame = {
     spark.read.parquet(testFile(name))
+  }
+
+  testGluten(
+    "SPARK-54220: vectorized reader: missing all struct fields, struct with NullType only") {
+    val data = Seq(
+      Tuple1((null, null)),
+      Tuple1((null, null)),
+      Tuple1(null)
+    )
+    val readSchema = new StructType().add(
+      "_1",
+      new StructType()
+        .add("_3", IntegerType, nullable = true)
+        .add("_4", StringType, nullable = true),
+      nullable = true)
+    val expectedAnswer = Row(Row(null, null)) :: Row(Row(null, null)) :: Row(null) :: Nil
+
+    withParquetFile(data) {
+      file =>
+        for (offheapEnabled <- Seq(true, false)) {
+          withSQLConf(
+            SQLConf.PARQUET_VECTORIZED_READER_NESTED_COLUMN_ENABLED.key -> "true",
+            SQLConf.LEGACY_PARQUET_RETURN_NULL_STRUCT_IF_ALL_FIELDS_MISSING.key -> "false",
+            SQLConf.COLUMN_VECTOR_OFFHEAP_ENABLED.key -> offheapEnabled.toString
+          ) {
+            withAllParquetReaders {
+              val df = spark.read.schema(readSchema).parquet(file)
+              checkAnswer(df, expectedAnswer)
+            }
+          }
+        }
+    }
   }
 }
--- a/cpp/core/config/GlutenConfig.h
+++ b/cpp/core/config/GlutenConfig.h
@@ -76,6 +76,9 @@ const std::string kParquetWriterVersion = "parquet.writer.version";
 
 const std::string kParquetCompressionCodec = "spark.sql.parquet.compression.codec";
 
+const std::string kLegacyParquetReturnNullStructIfAllFieldsMissing =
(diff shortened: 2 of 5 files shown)


gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
```diff
@@ -17,14 +17,51 @@
 package org.apache.spark.sql.execution.datasources.parquet
 
 import org.apache.spark.sql._
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.types.{IntegerType, StringType, StructType}
 
 /** A test suite that tests basic Parquet I/O. */
 class GlutenParquetIOSuite extends ParquetIOSuite with GlutenSQLTestsBaseTrait {
+  override def testNameBlackList: Seq[String] =
```
philo-he: Can we keep this test excluded in VeloxTestSettings.scala? That way, it won't affect other backends' test enabling controls.

philo-he: `testGluten("SPARK-54220: xxx` will add a prefix into the test name. I assume `excludeByPrefix("SPARK-54220")` will not exclude this test variant.

philo-he (commented):
Thanks for the PR.

gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
wecharyu: I'd prefer exclude it here because this test failure is not Velox-specific, it's caused by Gluten instead of backends.

Spark test collect leaf scan node `FileSourceScanExecTransformer` and run `executeColumnar()` in `VerifyNoAdditionalScanOutputExec`: https://github.com/apache/gluten/blob/356fda012d859f693fc613e6ac3585b7661bdbc5/shims/spark41/src/main/scala/org/apache/spark/sql/execution/AbstractFileSourceScanExec.scala#L139-L142

The inputRDD is `RDD[UnsafeRow]` for `parquetVectorizedReaderEnabled=false` test path, so it would always failed by:
```bash
java.lang.ClassCastException: class org.apache.spark.sql.catalyst.expressions.UnsafeRow cannot be cast to class org.apache.spark.sql.vectorized.ColumnarBatch
```

philo-he (commented):
Thanks for the clarification. I updated the PR description according to my understanding. 

Please rebase the code and check if the CI reported failures are related. Thank you!

wecharyu:
The failure is expected, we need first support it in Velox, PR is https://github.com/facebookincubator/velox/pull/17391

philo-he:
> The failure is expected, we need first support it in Velox, PR is [facebookincubator/velox#17391](https://github.com/facebookincubator/velox/pull/17391)

@wecharyu, thank you for the clarification. If you need to validate the Velox PR, you can push a temporary commit with the PR ID added in the code below.
https://github.com/apache/gluten/blob/97632d8435a8df1279478629122fdffee82bbeab/ep/build-velox/src/get-velox.sh#L28

wecharyu:
@philo-he Could you help reopen this PR? The Velox support is landed. Thanks!

wecharyu:
Gentle ping @philo-he @zhouyuan, pls help reopen it.

zhouyuan:
@wecharyu just opened this patch, would you please do a rebase?

gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
jackylee-ch: +1 for `excludeByPrefix` in `VeloxTestSettings.scala`. Keeping all skipped/excluded tests centralized makes them easier to track and revisit. Otherwise, such exclusions can easily be forgotten when the context is only known to a few contributors.

wecharyu: @jackylee-ch Thanks for the suggestion. This is not a Velox-backend-specific issue, and the test is not excluded; it is adapted for Gluten. Keeping the exclusion in `VeloxTestSettings.scala` would make it appear that this test is disabled.

philo-he (approved):
LGTM. Please check if the CI failure is related. Thanks.

gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
jackylee-ch: `excludeByPrefix` does not match the `Gluten - ` prefix added by `testGluten`; consequently, the `.excludeByPrefix("SPARK-54220")` setting in `VeloxTestSettings` effectively allows your variant to keep running—a point @philo-he raised back in May.

More importantly, the `testNameBlackList` mechanism is difficult to track: it takes precedence over all backend-specific settings, meaning no backend can re-enable the test, and the skipped tests won't be reflected in #11550—easily leading to genuine issues being obscured within the test suite files. 

It would be more appropriate to consolidate skip configurations applicable to all backends into `GlutenTestSettings`; I will file an issue to see if we can improve this.

wecharyu:
@philo-he Added `spark.sql.legacy.parquet.returnNullStructIfAllFieldsMissing` config support to solve the failed test, pls take a look again.

philo-he:
@wecharyu, thank you for the update. Could you check if this comment makes sense? https://github.com/apache/gluten/pull/12056#discussion_r3889036234

gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
wecharyu: This PR aims to enable all "SPARK-54220" prefix tests, even if it makes some adoption for gluten, so I think we do not need track it as exclusion test because the test is really not excluded. cc @philo-he

jackylee-ch: Oh, got it. Then we can just change it with `.exclude("SPARK-54220: vectorized reader: missing all struct fields, struct with NullType only")`. Can we use this?

jackylee-ch:
@wecharyu Just a little more comments, please check it again. Thanks for your work~

gluten-ut/spark41/src/test/scala/org/apache/spark/sql/execution/datasources/parquet/GlutenParquetIOSuite.scala
wecharyu: If exclude it in `VeloxTestSettings.scala`, we will also need exclude it in `ClickHouseTestSettings.scala` once the clickhouse ci enables gluten-ut for Spark 4.1,   because the original Spark test is not compatible with gluten as discussed in https://github.com/apache/gluten/pull/12056#discussion_r3252527294.

If the duplicated excludes in both backends are acceptable, I'm happy to change this exclude to `VeloxTestSettings.scala`.

jackylee-ch: The duplicated exclusions are acceptable for now. Backend-specific exclusions are reasonable, since different backends may have different implementations or optimization coverage.

wecharyu: Addressed.

jackylee-ch (approved):
Thanks for your work.

philo-he (approved):
Thanks.

PR #12061, opened 2026-05-09
[GLUTEN-11980][CORE][TESTS] Add test cases for decimal-key file type and read type different

beliefer:

This PR proposes to add test cases for decimal-key joins if either side of join exists native scan fallback to vanilla.
Before https://github.com/facebookincubator/velox/commit/472d3196bb506f33e0aa1573d9fb445fe60fe676, there is a bug that join key contains decimal type if either side of join exists native scan fallback to vanilla. The bug is caused by decimal precision/scale is different from read side and write side.
After https://github.com/facebookincubator/velox/commit/472d3196bb506f33e0aa1573d9fb445fe60fe676, the bug is fixed.

## How was this patch tested?

Manual tests in our production environment.
UT.

'No'.
I just use AI to generate the comments.

Related issue: #11980


beliefer:
@jinchengchenghh Now, this PR passed. Shall we push this PR to add these tests ?

jinchengchenghh (commented):
Thanks！

jinchengchenghh (approved):

beliefer:
@jinchengchenghh Thank you!

Issue #11915, opened 2026-04-11
[VL] Spark 4.1: Support checksum-based shuffle writers (SPARK-53322)


baibaichen:
## Backend
VL (Velox)

## Bug description

Spark 4.1 introduced checksum-based shuffle writers (SPARK-53322/SPARK-54439, sub-tasks of SPARK-51166 Umbrella). Gluten's shuffle writer implementation does not produce the expected checksums, causing test failures.

Spark 4.1 only.

Parent issue: #11550

### Impact

| Suite | Status | spark40 | spark41 |
|-------|--------|:-------:|:-------:|
| GlutenMapStatusEndToEndSuite | Entire suite TODO | 🟢 | 🔴 |
| GlutenKeyGroupedPartitioningSuite | `.excludeByPrefix("SPARK-53322")` | 🟢 | 🔴 |
| GlutenKeyGroupedPartitioningSuite | `.excludeByPrefix("SPARK-54439")` | 🟢 | 🔴 |

### References

- Apache Spark JIRA: [SPARK-53322](https://issues.apache.org/jira/browse/SPARK-53322)
- Apache Spark JIRA: [SPARK-54439](https://issues.apache.org/jira/browse/SPARK-54439)
- Apache Spark JIRA: [SPARK-51166](https://issues.apache.org/jira/browse/SPARK-51166) (Umbrella)

PR #12067, opened 2026-05-11
[GLUTEN-11915][VL] Support RowBasedChecksum in ColumnarShuffleWriter (SPARK-51756)

jaylisde:
## Summary

Spark 4.1 introduced RowBasedChecksum ([SPARK-51756](https://issues.apache.org/jira/browse/SPARK-51756)) for detecting non-deterministic stage retries. When `spark.sql.shuffle.orderIndependentChecksum.enabled` or `spark.sql.shuffle.checksum.mismatchFullRetry.enabled` is true, the shuffle writer must compute an order-independent per-row checksum and pass it via `MapStatus.checksumValue` to the driver for comparing across task attempts.

**Problem:** Gluten's `ColumnarShuffleWriter` always returns `checksumValue = 0`, causing the driver to skip non-deterministic retry detection. If a task retry produces different output (e.g., due to round-robin partitioning), downstream consumers may silently read inconsistent data without triggering a full stage retry.

**Fix:** Implement native C++ row-based checksum computation in `VeloxHashShuffleWriter`. For each row in `doSplit()`, serialize via `UnsafeRowFast` and compute XXH64 hash. Aggregate per-partition using XOR+SUM (order-independent). Return checksum array via JNI to Scala layer, which passes the aggregated value to `MapStatus.checksumValue`.

## Changes

- `VeloxHashShuffleWriter.cc`: Added `computeRowBasedChecksums()` using UnsafeRowFast + XXH64 with per-partition XOR+SUM aggregation.
- `Options.h`, `ShuffleWriter.h/cc`: Added `rowBasedChecksumEnabled` option and `rowBasedChecksums()` accessor.
- `JniWrapper.cc`: Accept boolean config param, return checksum array.
- `GlutenSplitResult.java`, `ShuffleWriterJniWrapper.java`: Added `rowBasedChecksums` field and param.
- `ColumnarShuffleWriter.scala`: Read SQLConf (OR logic), pass to native, use for MapStatus.
- `GlutenMapStatusUtil.scala` (shims/spark33-41): Cross-version MapStatus compatibility.
- `RowBasedChecksumTest.cc`: C++ unit test for order-independence, null handling, determinism.
- `GlutenRowBasedChecksumSuite.scala`: End-to-end Scala test for checksum determinism and data-change detection.

## Test

- C++ unit test: 4/4 pass (order-independence, data-change detection, null handling, deterministic)
- `GlutenRowBasedChecksumSuite`: 2/2 pass (deterministic, detects data change)

Partially addresses #11915.


wForget:
> Spark 4.1 introduced shuffle checksum end-to-end verification (SPARK-53322), requiring MapStatus.checksumValue to be non-zero and .checksum files to contain valid per-partition checksums.

Do you mean https://issues.apache.org/jira/browse/SPARK-51756?

jaylisde:
Thanks @wForget for catching that. The correct reference should be [SPARK-54663](https://issues.apache.org/jira/browse/SPARK-54663). I'll update the PR title and description.

wForget:
> Thanks @wForget for catching that. The correct reference should be [SPARK-54663](https://issues.apache.org/jira/browse/SPARK-54663). I'll update the PR title and description.

SPARK-54663 proposes row-based checksum, but the current implementation is based on data file.

jaylisde:
Thanks @wForget. Updated to proper RowBasedChecksum (SPARK-51756) with native per-row XXH64 + order-independent aggregation. File-based checksum will be a follow-up PR.

philo-he:
@marin-ma, could you take a look when you get a chance?

backends-velox/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -60,6 +60,14 @@ class ColumnarShuffleWriter[K, V](
 
   private val blockManager = SparkEnv.get.blockManager
 
+  private val rowBasedChecksumEnabled: Boolean = {
+    try {
+      GlutenMapStatusUtil.isRowBasedChecksumEnabled
+    } catch {
+      case _: NoSuchMethodError | _: NoSuchFieldError => false
```
wForget: We have created shim of `GlutenMapStatusUtil.isRowBasedChecksumEnabled` method for each spark version. Do we still need to catch NoSuchMethodError/NoSuchFieldError exceptions?

jaylisde: The try-catch was a leftover from before introducing the shim. Removed.

wForget:
Thanks @jaylisde , overall looks good to me, could you fix the conflicts? cc @philo-he @marin-ma could you please review the native code?

jaylisde:
Thanks @wForget for the review. Fixed the conflicts.

gluten-ut/spark41/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
```diff
@@ -1027,7 +1027,7 @@ class VeloxTestSettings extends BackendTestSettings {
     .exclude("SPARK-49386: test SortMergeJoin (with spill by size threshold)")
   enableSuite[GlutenMathFunctionsSuite]
   // TODO: fix on Spark-4.1 see https://github.com/apache/spark/pull/50230
```
philo-he: Should we remove this comment?

gluten-ut/spark41/src/test/scala/org/apache/spark/sql/GlutenMapStatusEndToEndSuite.scala
```diff
@@ -28,5 +29,58 @@ class GlutenMapStatusEndToEndSuite extends MapStatusEndToEndSuite with GlutenTes
...
+  import org.apache.spark.MapOutputTrackerMaster
+
+  private def getLatestShuffleChecksumValues(): Array[Long] = {
+    val tracker = _spark.sparkContext.env.mapOutputTracker
+      .asInstanceOf[MapOutputTrackerMaster]
+    val latestShuffleId = tracker.shuffleStatuses.keys.max
+    tracker.shuffleStatuses(latestShuffleId).mapStatuses.map(_.checksumValue)
+  }
+
+  test("Gluten row-based checksum is deterministic") {
```
philo-he: Test suites like GlutenMapStatusEndToEndSuite extend Spark's own test classes to run Spark tests with Gluten enabled. For the newly added tests here, if they aren't variants of existing tests in the parent Spark test class, it may be better to move them to a dedicated test suite. Could you do a refactor?

philo-he (commented):
Looks good overall. Only some minor comments. Please check if they make sense. Thanks.

gluten-ut/spark41/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
jaylisde: Done

gluten-ut/spark41/src/test/scala/org/apache/spark/sql/GlutenMapStatusEndToEndSuite.scala
jaylisde: Done - moved the two Gluten-specific tests into a new GlutenRowBasedChecksumSuite under `org.apache.spark.sql.gluten`. GlutenMapStatusEndToEndSuite is restored to a thin wrapper that only runs upstream Spark's MapStatusEndToEndSuite tests under Gluten. The new suite extends GlutenSQLTestsTrait directly and configures the required `SQLConfs / GLUTEN_ANSI_FALLBACK_ENABLED` via sparkConf. Thanks for the suggestion.

jaylisde:
> Looks good overall. Only some minor comments. Please check if they make sense. Thanks.

Thanks @philo-he for reviewing.

philo-he (approved):
LGTM. Thanks!

marin-ma:
Apologies for missing this. @wForget @philo-he Thanks for the review. I will take a look today. Thanks!

marin-ma:
Hi @jaylisde I understand the checksum computation was introduced to resolve the data consistency issue with task retries, and we do lack this support for spark 4.0 in Gluten. But since Gluten's hash-based shuffle handles the data in columnar format, I wonder if it's possible to implement this based on each column buffer? The conversion from columnar to row for the checksum computation looks expensive. Have you measured the performance impact from the checksum computation?

wForget:
> Hi @jaylisde I understand the checksum computation was introduced to resolve the data consistency issue with task retries, and we do lack this support for spark 4.0 in Gluten. But since Gluten's hash-based shuffle handles the data in columnar format, I wonder if it's possible to implement this based on each column buffer?

The column buffer of shuffle write data is unstable; when upstream recalculates, using it to calculate checksun cannot guarantee that the same value will be generated.

jaylisde:
Thanks @marin-ma. Column-buffer hashing isn't retry-deterministic as @wForget mentioned - Gluten splices per-partition column buffers across batches, so identical rows across retries can produce different buffer bytes. `UnsafeRow`  is the data-semantics layer Spark's upstream `RowBasedChecksum` already uses, so Gluten and vanilla Spark checksums stay comparable.

On perf: opt-in (default false), only paid when debugging non-deterministic stages.

PR #12070, opened 2026-05-11
[VL] Fix parallel BHJ build after #11882 

zhztheplayer:
Before #11882, BHJ parallel build was using the io executor. #11882 changed the behavior accidentally. This patch fixes the issue.


JkSelf (approved):
Thanks for your fix. LGTM.

PR #12073, opened 2026-05-11
[VELOX] Support push filter for table cache

jackylee-ch:

  This PR adds batch-level statistics collection and filter pushdown for Velox columnar cached batches,
   enabling Spark to skip cached batches that cannot satisfy query predicates (similar to Parquet
  row-group pruning but for in-memory cache).

  Architecture:

  - C++ (BatchStatsCollector): During columnar batch serialization, computes per-column min/max bounds,
   null counts, row counts, and byte sizes. Stats are appended to the serialized payload in a compact
  binary wire format.
  - Scala (ColumnarCachedBatchSerializer): Decodes the wire-format stats and integrates with Spark's
  SimpleMetricsCachedBatch filter evaluation to skip batches whose bounds prove the predicate cannot
  match.

  Supported bound types: Boolean, Byte, Short, Int, Long, Float, Double, Date, Timestamp, String, and
  Decimal(precision<=18).

  Key design decisions:
  - Wire format v1: per-column tag(1B) + hasBounds(1B) + bounds(variable) + nullCount(4B) +
  rowCount(4B) + sizeInBytes(8B)
  - Tautological bounds (type extremes) for unknown/absent bounds to avoid the 3VL null-skip
  correctness bug (where null bounds cause Spark's predicate to evaluate to null → coerced to false →
  batch incorrectly skipped)
  - Float/Double with NaN degrade to pass-through (no finite tautological pair exists due to NaN
  ordering)
  - String bounds capped at 64 KiB to prevent metadata bloat
  - Tag/dataType compatibility validation to reject corrupt payloads gracefully
  - Backward compatible: unknown tags fall through to pass-through filtering
  - Controlled by config spark.gluten.sql.columnar.tableCacheFilterEnabled (default: true)

 ## How was this patch tested?

  1. Unit tests (ColumnarCachedBatchSerializerSuite): 38 tests covering wire format round-trip for all
  supported types, NaN poisoning, inverted bounds rejection, tag/schema mismatch detection,
  tautological bounds fallback, truncated payload handling, negative counter rejection, and Decimal
  bounds (with/without bounds, precision>18 rejection).
  2. E2E tests (VeloxColumnarCacheSuite): Integration tests that cache real data, run filter queries,
  and validate correctness via checkAnswer for Int, String, Timestamp, and Decimal predicates including
   >, BETWEEN, =, and IS NULL.

##  Was this patch authored or co-authored using generative AI tooling?

  Generated-by: Claude Code (Claude Opus 4.7)

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
@@ -29,19 +29,22 @@ import org.apache.gluten.vectorized.ColumnarBatchSerializerJniWrapper
 import org.apache.spark.internal.Logging
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression}
-import org.apache.spark.sql.columnar.{CachedBatch, CachedBatchSerializer}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression, GenericInternalRow}
+import org.apache.spark.sql.columnar.{CachedBatch, SimpleMetricsCachedBatch, SimpleMetricsCachedBatchSerializer}
 import org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.types.{StructField, StructType}
+import org.apache.spark.sql.types._
 import org.apache.spark.sql.utils.SparkArrowUtil
 import org.apache.spark.sql.vectorized.ColumnarBatch
 import org.apache.spark.storage.StorageLevel
+import org.apache.spark.unsafe.types.UTF8String
 
-import com.esotericsoftware.kryo.{Kryo, Serializer => KryoSerializer}
+import com.esotericsoftware.kryo.{Kryo, KryoException, Serializer => KryoSerializer}
 import com.esotericsoftware.kryo.DefaultSerializer
 import com.esotericsoftware.kryo.io.{Input, Output}
 import org.apache.arrow.c.ArrowSchema
+
+import java.nio.{ByteBuffer, ByteOrder}
 
 /**
  * TODO: fix on Spark-4.1 - Documentation
@@ -50,22 +53,119 @@ import org.apache.arrow.c.ArrowSchema
  * {{{
  *   spark.kryo.classesToRegister=org.apache.spark.sql.execution.CachedColumnarBatch
  * }}}
+ *
+ * `sizeInBytes` semantics: this is the wire size of the Presto-encoded payload (plus the optional
+ * stats payload when present), not the sum of per-column `sizeInBytes` stats. This diverges from
+ * `SimpleMetricsCachedBatch`'s default of `Long.MaxValue`. We override with the wire size because
+ * `InMemoryRelation.computeStats` / CBO consumes it as the on-disk/in-memory cache footprint for
+ * cost modeling, which is the Presto payload size here -- not the uncompressed column byte total.
+ * Per-column `sizeInBytes` is available via `stats` for finer-grained accounting.
  */
 @DefaultSerializer(classOf[CachedColumnarBatchKryoSerializer])
 case class CachedColumnarBatch(
     override val numRows: Int,
     override val sizeInBytes: Long,
-    bytes: Array[Byte])
-  extends CachedBatch {}
+    bytes: Array[Byte],
+    override val stats: InternalRow)
+  extends SimpleMetricsCachedBatch {}
 
-class CachedColumnarBatchKryoSerializer extends KryoSerializer[CachedColumnarBatch] {
+object CachedColumnarBatch {
+  // Backward-compatible constructor for call sites that don't (yet) produce stats.
+  // Defaults to null, which will cause `ColumnarCachedBatchSerializer.buildFilter` to
+  // fall back to pass-through for the containing partition.
+  def apply(numRows: Int, sizeInBytes: Long, bytes: Array[Byte]): CachedColumnarBatch =
+    CachedColumnarBatch(numRows, sizeInBytes, bytes, stats = null)
+}
+
+/**
+ * Kryo serializer for [[CachedColumnarBatch]] supporting two wire formats.
+ *
+ *   - v0 (legacy): `[numRows:int32][sizeInBytes:int64][bytesLen+1:int32][bytes]`
+ *   - v1 (with stats): `[magic:int32=0xC0DEC0DE][version:int8=1]` followed by v0 payload, then
+ *     `[statsMarker:int8][numFields:int32][perField...]` where statsMarker is 0 for null stats or 1
+ *     when stats row is present.
+ *
+ * Rolling-upgrade safety: a new executor writes v0 when `stats == null` (default, filter-pushdown
+ * disabled, rolling-upgrade kill switch set, or legacy serializer path) and v1 only when a real
+ * stats row is attached. Pre-filter-pushdown Gluten binaries have NO `numRows >= 0` guard on their
+ * v0 reader -- they read the v1 magic int directly as `numRows`, which is negative, and proceed to
+ * compute a garbage `length` from the misaligned follow-up bytes, likely allocating a huge `byte[]`
+ * and crashing or OOM-ing. Spark does not rolling-upgrade a SparkContext's executor binaries
+ * mid-application, so this is primarily a concern for caches persisted with MEMORY_AND_DISK(_2)
+ * that outlive a version change. If your deployment shares caches across mixed binaries, set
+ * `spark.gluten.sql.columnar.tableCache.stats.wire.v1.enabled=false` on the writer to force v0
+ * emission regardless of filter pushdown until the cluster is uniform.
+ *
+ * Forward compatibility: v1 always starts with a negative magic int; v0 can never begin with a
+ * negative int because `numRows` is non-negative. When reading, if the first int4 != magic, treat
+ * the stream as v0 and set `stats = null`.
+ */
+class CachedColumnarBatchKryoSerializer
+  extends KryoSerializer[CachedColumnarBatch]
+  with Logging {
+  import CachedColumnarBatchKryoSerializer._
+
   override def write(kryo: Kryo, output: Output, batch: CachedColumnarBatch): Unit = {
+    // Stats must be a GenericInternalRow for the Kryo writer to enumerate fields
+    // (Spark's InternalRow abstract API has no type-erased `get(i)`). Other
+    // implementations (`UnsafeRow`, projection-wrapped rows) cannot be written
+    // without column schema, which is unavailable at this layer. Rather than
+    // throw mid-stream -- which would leave a half-written block of garbage
+    // after the MAGIC / VERSION bytes had already been committed -- gracefully
+    // fall back to v0 (no-stats) format when the invariant is violated.
+    // decodeStats (our only producer today) always returns GenericInternalRow.
+    val writeV1 = batch.stats match {
+      case null => false
+      case _: GenericInternalRow => true
+      case other =>
+        warnOnce(
+          "stats-non-generic",
+          s"CachedColumnarBatch stats is ${other.getClass.getName}, expected " +
+            "GenericInternalRow; falling back to v0 (no-stats) cache block."
+        )
+        false
+    }
+    if (!writeV1) {
+      // Emit the legacy v0 header so that a rolling upgrade with an older
+      // Gluten executor still on the pre-filter-pushdown binary can read
+      // caches produced by a new executor whenever stats are absent. The v1
+      // magic is negative, which would trip the non-negative `numRows` guard
+      // on the old reader path.
+      writeV0Payload(output, batch)
+    } else {
+      output.writeInt(MAGIC)
+      output.writeByte(VERSION_V1)
+      writeV1Payload(output, batch)
+    }
+  }
+
+  private def writeV0Payload(output: Output, batch: CachedColumnarBatch): Unit = {
+    // Defensive write-side invariants: `numRows` is non-negative and cannot
+    // collide with the v1 MAGIC value. Spark semantics already guarantee
+    // `numRows >= 0`, but an upstream bug passing `Int.MinValue` or a
+    // MAGIC-valued counter would silently produce a stream that the reader
+    // would misinterpret as v1. Failing at write time keeps such a bug from
+    // poisoning the cache store.
+    require(
+      batch.numRows >= 0,
+      s"CachedColumnarBatch numRows must be non-negative, got ${batch.numRows}")
+    require(
+      batch.numRows != MAGIC,
+      s"CachedColumnarBatch numRows collides with v1 MAGIC ($MAGIC); refusing to write")
     output.writeInt(batch.numRows)
     output.writeLong(batch.sizeInBytes)
     require(
       batch.bytes != null,
       "The object 'CachedColumnarBatch.bytes' is invalid or malformed to " +
         s"serialize using ${this.getClass.getName}")
+    // Symmetric writer-side cap: the reader enforces `[0, MAX_BATCH_LEN]` on the
+    // length field, and `bytes.length + 1` would overflow a signed int at 2 GiB.
+    // Failing at cache-fill time (where the real payload is sitting right here)
+    // gives a clearer diagnostic than a corrupted-looking deserialize failure.
+    require(
+      batch.bytes.length <= MAX_BATCH_LEN,
+      s"CachedColumnarBatch payload length ${batch.bytes.length} exceeds " +
+        s"serializable cap $MAX_BATCH_LEN")
     output.writeInt(batch.bytes.length + 1) // +1 to distinguish Kryo.NULL
     output.writeBytes(batch.bytes)
   }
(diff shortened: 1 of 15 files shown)


backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -74,30 +174,470 @@ class CachedColumnarBatchKryoSerializer extends KryoSerializer[CachedColumnarBat
...
+    require(
+      n >= 0 && n <= MAX_STATS_ARR_LEN,
+      s"CachedColumnarBatch stats field count $n is out of range [0, $MAX_STATS_ARR_LEN]")
+    val values = stats match {
+      case g: GenericInternalRow => g.values
+      case other =>
+        throw new UnsupportedOperationException(
+          s"CachedColumnarBatch stats must be GenericInternalRow, got ${other.getClass}")
+    }
+    // H8: Encode stats body into a sized buffer so we can write a length prefix.
```
LuciferYang: Comments cover non-obvious rationale well, but density slows the diff:

- `CachedColumnarBatchKryoSerializer.write`: ~30 lines code, ~50 lines comments
- `tautologicalBoundsFor` FloatType case: 17 lines of NaN rationale
- The NaN-3VL-skip explanation repeats 5+ times in `ColumnarCachedBatchSerializer.scala`
- Internal review-cycle tags (`R2-H4`, `R2-H11`, `R3-H1`, `H8`, etc.) leak into production
  code

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -296,10 +864,799 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+        // ordering per collation, not byte-wise. When we cannot fabricate a
+        // safe bound for a StringType column, escalate to a null stats row
+        // for the entire batch so `buildFilter` falls into its `smb.stats ==
+        // null => pass through` branch. Per-column null sentinels would re-
+        // introduce the 3VL-skip bug above.
+        val (lower, upper) = if (hasBounds) {
+          val (lo, hi) = readBounds(buf, typeTag, dataType)
+          if (lo == null || hi == null) {
+            val opt = tautologicalBoundsFor(dataType)
+            if (opt.isEmpty) return null else opt.get
```
LuciferYang, line 1116: When any column's `tautologicalBoundsFor` returns None (NaN-degraded Float/Double, collated
StringType, unknown DataType), `decodeStats` demotes the entire batch's stats to null. This
nullifies valid min/max bounds on the OTHER columns in the same batch — e.g. a 12-column ML
feature table where only one column carries NaN: the remaining 11 columns lose pruning
entirely on that batch.

The unknown-DataType path (YearMonthIntervalType etc.) can switch to per-column demotion:
fill the offending column's `lower`/`upper` slots with type-extreme sentinels (long min/max)
and set `nullCount=1` plus the real `rowCount` so `IsNull`/`IsNotNull` predicates remain
useful while bound-comparison predicates degrade to no-ops on that column only. The other
columns keep their valid bounds. **Prerequisite**: extend `tautologicalBoundsFor` to
recognise these types — the current `case _ => None` (`:1250-1255`) is a deliberate
conservative choice ("safer than fabricating bounds we can't prove correct"); per-column
demotion flips this trade-off from "conservative across the board" to "non-conservative
once a type is recognised".

Caveat: Float/Double NaN truly has no finite tautological pair (Spark orders NaN above +Inf),
and collated StringType has no byte-wise max either (the code comment at `:1104-1107` admits
this — "0xFF*256 is not the max", and Spark 4.0+ collation-aware ordering breaks any byte-
wise sentinel). Both must keep the all-or-nothing demotion or the three-valued-logic skip
bug returns (a NULL bound coerces the predicate to NULL, which is then coerced to false,
wrongly skipping the batch). Only the unknown-DataType case can be safely narrowed.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -296,10 +864,799 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+        // on both 3.x (singleton, always binary) and 4.x (case class with
+        // collationName == "UTF8_BINARY"), and "string collate xxx" for
+        // any other collation. This gives us a shim-safe predicate that
+        // compiles unchanged across Spark 3.3..4.1.
+        dataType.isInstanceOf[StringType] && dataType.catalogString == "string"
+      case StatsTypeTag.DATE => dataType == DateType
+      case StatsTypeTag.TIMESTAMP =>
+        // Match the compat pattern used in Validators.containsNTZ so this
+        // compiles across Spark 3.3 (no TimestampNTZType) through 3.5+.
+        dataType == TimestampType || dataType.catalogString == "timestamp_ntz"
```
LuciferYang, line 1418: `isTagCompatibleWithDataType` accepts both `TimestampType` and `TimestampNTZType` for the
TIMESTAMP wire tag. Both are int64 underneath, but their semantics are NOT interchangeable:

- `TimestampType`: epoch micros (UTC)
- `TimestampNTZType`: local micros (no tz)

If a cache block is written under a `TimestampType` schema (C++ converts Velox `Timestamp`
to UTC micros) but later read against a `TimestampNTZType` schema, the predicate literal is
parsed differently on each side — both end up as `long`, but the values now disagree by
exactly the local-tz offset. Filter pushdown silently mis-prunes against tz-shifted bounds.

**Pre-merge**: Tighten to TimestampType-only; let TimestampNTZType fall into `case _ => false`
and pass through.

**Follow-up issue**: Split the C++ `typeTagFor` into `kTimestamp` / `kTimestampNtz` so
cross-version schema drift becomes an explicit incompat rather than silent wrong answers.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -296,10 +864,799 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+  // Upper bound on the per-string lower/upper bound length read from the
+  // stats payload. Must match the C++ writer's `kStringBoundsCap` in
+  // `cpp/velox/operators/serializer/BatchStatsCollector.cc` (currently 64
+  // KiB). Any larger value on the wire is either a future writer that
+  // intentionally bumped the cap (in which case both sides move together)
+  // or a corrupt cache -- reject to prevent memory-DoS from a bogus length
+  // prefix. Prior release mismatched the reader at 1 MiB vs. the writer at
+  // 64 KiB; tightening the reader side aligns the contract and closes the
+  // over-allocation window.
+  private val MAX_STATS_STRING_LEN: Int = 64 * 1024
```
LuciferYang, line 981: The 64-KiB string cap is hand-replicated on both sides:
- `kStringBoundsCap = 64 * 1024` in `BatchStatsCollector.cc:296`
- `MAX_STATS_STRING_LEN = 64 * 1024` in `ColumnarCachedBatchSerializer.scala:981`

Unlike `StatsTypeTag` (which has cross-side `static_assert` plus a Scala unit-test pinning),
this constant has no automated drift guard. If a future change bumps the C++ cap to 1 MiB
without updating Scala, the reader rejects every string in the 64 KiB–1 MiB range
(`len > MAX_STATS_STRING_LEN` → null stats → pass-through) — a silent pruning regression
with green `checkAnswer` tests.

Two options, either acceptable:
1. Add `static_assert(kStringBoundsCap == 64 * 1024, ...)` on C++ AND a unit test pinning
   `MAX_STATS_STRING_LEN == 64 * 1024` on Scala — makes the value a build-break contract.
2. Plumb the cap through `GlutenConfig` so there is only one source of truth.

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1024,6 +1023,37 @@ object GlutenConfig extends ConfigRegistry {
...
+          "InMemoryTableScanExec can skip cached batches whose statistics don't match the " +
+          "scan predicate (equivalent to Spark's SimpleMetricsCachedBatchSerializer). " +
+          "Disabled by default to match pre-feature behavior; flip to true to opt in. Has no " +
+          "effect when spark.gluten.sql.columnar.tableCache is false, and is additionally " +
+          "suppressed at the writer when " +
+          "spark.gluten.sql.columnar.tableCache.stats.wire.v1.enabled=false (rolling-upgrade " +
+          "kill switch); in that case reads continue to decode both v0 and v1 transparently, " +
+          "but writes emit v0 only and no stats are produced for new cache blocks.")
+      .booleanConf
+      .createWithDefault(false)
```
LuciferYang, line 1041: PR description states "default: true" but the code is `createWithDefault(false)`, and the
config key itself differs (PR description: `spark.gluten.sql.columnar.tableCacheFilterEnabled`;
code: `spark.gluten.sql.columnar.tableCache.filterPushdown.enabled`).

Keep the code default (false is safer for a v1 feature: opt-in lets production clusters get
observability before the behavior change lands), and update the PR description and release
notes to match.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -29,43 +29,143 @@ import org.apache.gluten.vectorized.ColumnarBatchSerializerJniWrapper
...
-  extends CachedBatch {}
+    bytes: Array[Byte],
+    override val stats: InternalRow)
+  extends SimpleMetricsCachedBatch {}
+
+object CachedColumnarBatch {
+  // Backward-compatible constructor for call sites that don't (yet) produce stats.
+  // Defaults to null, which will cause `ColumnarCachedBatchSerializer.buildFilter` to
+  // fall back to pass-through for the containing partition.
+  def apply(numRows: Int, sizeInBytes: Long, bytes: Array[Byte]): CachedColumnarBatch =
```
LuciferYang, line 76: The 3-arg `CachedColumnarBatch.apply(numRows, sizeInBytes, bytes)` companion is documented
as a backward-compat shortcut, but the PR itself has no call sites (everything explicitly
passes 4 args with `stats = null` or a real row).

Risks:
1. Future code can call the 3-arg variant without thinking and inherit `stats = null` as the
   implicit default, making the null-stats path invisible to `grep "CachedColumnarBatch("`.
2. If this is for third-party backend compat, document that explicitly; otherwise it's an
   undocumented escape hatch.

Either:
- Annotate `@deprecated("Use the 4-arg form with explicit stats", since="X.Y")` and add
  scaladoc pointing to the legacy-fallback as the only legitimate caller.
- Or delete it entirely and force every call site to pass `stats = null` explicitly. This
  makes the null-stats path grep-able and code-review-able.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxColumnarCacheSuite.scala
```diff
@@ -244,4 +244,132 @@ class VeloxColumnarCacheSuite extends VeloxWholeStageTransformerSuite with Adapt
...
+  }
+
+  test("Filter pushdown: selective predicate returns zero rows without error") {
+    // H11 guard: the earlier filter-pushdown tests verify that cached queries
+    // return the RIGHT rows; this test additionally verifies the end-to-end
+    // path on a highly selective predicate (literal far outside the column's
+    // range) executes cleanly and returns the expected empty result set.
+    //
+    // NOTE: earlier revisions of this test asserted on
+    // `InMemoryTableScanExec.numCachedBatchesSkipped` to prove that pruning
```
LuciferYang, line 326: The test explicitly notes that `InMemoryTableScanExec.numCachedBatchesSkipped` does not exist.
Once this feature ships, there is NO observability into:

- Whether pushdown actually fires for a given query
- False-positive skips (incorrect pruning) — no alarm channel
- Silent pass-through regressions (schema drift / NaN poisoning lighting up across the cluster)

Concrete implementation:
1. In `ColumnarCachedBatchSerializer.buildFilter`, count batches via `SQLMetric` long counters
   — `numCachedBatchesSkipped` (rejected by parent filter) and `numCachedBatchesPassThrough`
   (stats=null bypassed) — exposed on a Gluten-owned metric channel (avoids upstream
   `InMemoryTableScanExec` ABI risk).
2. Add a `VeloxColumnarCacheSuite` assertion: under a highly selective predicate,
   `numCachedBatchesSkipped > 0`, proving pushdown actually fires.

gluten-arrow/src/main/java/org/apache/spark/sql/execution/unsafe/JniUnsafeByteBuffer.java
```diff
@@ -60,7 +85,25 @@ private synchronized void ensureOpen() {
...
+   * #ensureOpen()}. JNI callers MUST {@code env->ExceptionCheck() / env->ExceptionClear()}
+   * immediately after calling this method so the double-free exception does not mask the stashed
+   * primary failure.
+   *
+   * <p>NOTE: This is intentionally package-private rather than public because the only legitimate
+   * callers are JNI error-recovery paths and the two public {@code toByteArray}/{@code
+   * toUnsafeByteArray} methods on this class. Broader external use would risk use-after-free by
+   * racing with those getters.
+   */
+  synchronized void release() {
```
LuciferYang, line 106: `release()` was promoted from `private` to package-private with a doc comment stating "only
legitimate callers are JNI error-recovery paths and the two public toByteArray methods". A
subtlety:

- JNI can call ANY method via `env->CallVoidMethod(buf, releaseId)`; Java access modifiers do
  not gate JNI access — the JVM/JNI spec does no access check on `CallVoidMethod`.
- `package-private` only restricts in-package Java callers, which is orthogonal to the JNI
  access path.
- The doc-comment implies access is restricted to "JNI + toByteArray", which is misleading —
  any in-package Java class can call `buf.release()` directly.

Two options:
1. Tighten the doc: drop the "access is restricted" implication; replace with "exposes
   `release` for in-package Java callers; JNI uses jmethodID lookup, registered in
   `JNI_OnLoad`".
2. Restructure: revert `release` to `private`, add a separate `releaseFromJni()`
   package-private wrapper called only from the JNI guard. Aligns the access modifier with
   the actual contract.

Option 2 is cleaner; option 1 is the minimum doc fix.

cpp/velox/operators/serializer/BatchStatsCollector.cc
```diff
@@ -0,0 +1,710 @@
...
+// bounds would produce wrong filter results.
+StatsTypeTag typeTagFor(const TypePtr& type) {
+  if (type->isDate()) {
+    return StatsTypeTag::kDate;
+  }
+  if (type->isShortDecimal()) {
+    return StatsTypeTag::kDecimal;
+  }
+  if (type->isLongDecimal()) {
+    return StatsTypeTag::kUnsupported;
```
LuciferYang, line 118: Long decimal (precision 19-38) returns `kUnsupported` from `typeTagFor`. The reader gracefully
degrades — `tautologicalBoundsFor(DecimalType(p>18))` produces all-nines tautological bounds
plus real nullCount/rowCount, so `IsNull`/`IsNotNull` predicates still prune; but bound
predicates (`>`, `BETWEEN`, `=`) become trivially true on long-decimal columns and lose
pruning.

Not a wrong-answer bug, just a UX gap:
1. A user with a bound predicate on a long-decimal column enables pushdown and sees no
   speedup on that predicate — no log explains why.
2. Tests cover the reader-side rejection (precision>18) but not the writer-side silent
   `kUnsupported` E2E.
3. The PR description mentions "Decimal(precision<=18)" but it's easy for operators to miss.

Pick one or more:
1. Schema-level INFO log in `BatchStatsCollector::ensureInitialized` when any long-decimal
   column is present in the schema.
2. Update the GlutenConfig `.doc(...)` to call out the precision-18 ceiling explicitly.
3. Follow-up PR to support long decimal in the wire format (16-byte unscaled).

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxColumnarCacheSuite.scala
```diff
@@ -244,4 +244,132 @@ class VeloxColumnarCacheSuite extends VeloxWholeStageTransformerSuite with Adapt
       }
     }
   }
+
+  test("Filter pushdown: cached scan returns correct rows for numeric and string predicates") {
```
LuciferYang, line 248: The 38 unit + 5 E2E tests cover wire format and the happy path, but the review surface has
gaps:

1. **Cache eviction × pushdown race**: A partition LRU-evicted mid-prune — does the
   re-cache path produce fresh stats?
2. **MEMORY_AND_DISK_2 (replication)**: Wire-format consistency across replication paths.
3. **Schema drift mid-partition**: The C++ `schemaDriftPoisoned_` latched path — is there a
   unit test exercising the Scala side reading the resulting null-stats end-to-end? Trigger:
   feed two batches into `VeloxColumnarBatchSerializer::append` where the second has a
   different column count/type from the first.

Not blocking.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -296,10 +864,799 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+    // gate or the wire-v1 flag is flipped off at query time, we MUST NOT
+    // evaluate the parent filter even when individual cached batches carry
+    // v1 stats rows written under a prior configuration. Doing so would
+    // re-enable exactly the code path the operator is trying to disable.
+    // The writer-side check in `convertColumnarBatchToCachedBatch` only
+    // suppresses NEW stats emission; without this reader-side check, any
+    // batch already in the cache from before the kill-switch flip would
+    // still drive pruning decisions. Fall through to pass-through.
+    if (
+      !glutenConf.getConf(GlutenConfig.COLUMNAR_TABLE_CACHE_FILTER_PUSHDOWN_ENABLED) ||
```
LuciferYang, line 912: When `tableCache.filterPushdown.enabled` or `wire.v1.enabled` is flipped off, `buildFilter`
returns the pass-through closure. Correct behavior, but operationally invisible:

- No log entry confirming "I saw a v1 cache block but the kill-switch is set, so ignoring"
- No metric counting bypass events (related to Item 8)

For incident response — when an operator flips a kill-switch and expects pruning to stop —
they have no signal that the kill-switch actually fired vs. silently no-op'd. Add a single
`warnOnce("kill-switch-bypass", "...")`-style log on first entry per JVM, modeled after the
existing `sampledWarn` helper in the same file (defined at `:1001`/`:1009`, used throughout
`decodeStats`). Operator can grep the log to verify after the flip.

gluten-substrait/src/main/scala/org/apache/gluten/config/GlutenConfig.scala
```diff
@@ -1024,6 +1023,37 @@ object GlutenConfig extends ConfigRegistry {
...
+
+  val COLUMNAR_TABLE_CACHE_STATS_WIRE_V1_ENABLED =
+    buildConf("spark.gluten.sql.columnar.tableCache.stats.wire.v1.enabled")
+      .internal()
+      .doc(
```
LuciferYang, line 1046: The current doc only mentions rolling upgrades but misses the reader-side effect:

1. Writer: forces v0 wire format
2. Reader Kryo decoding: always double-decodes v0/v1 (independent of this flag)
3. Reader `buildFilter` (`:911-916`): `!filterPushdown.enabled || !wire.v1.enabled` returns
   pass-through, including for pre-cached v1 blocks

So flipping `wire.v1.enabled=false` alone is enough to stop reader-side pruning (OR, not
AND). **No code change required**, only the `.doc(...)` expansion below; collapsing the two
flags into a single kill switch can be a follow-up.

Suggest expanding the `.doc(...)`:

> "Rolling-upgrade kill switch for the stats-carrying v1 wire format. Affects both writer
> AND reader. Writer side: when false, executors emit v0 (no stats) regardless of
> `filterPushdown.enabled`. Reader side: when false, `buildFilter` returns pass-through
> for ALL cached batches (including v1 blocks already on disk), identical to flipping
> `filterPushdown.enabled` off. Either flag set to false fully disables pruning — they are
> OR'ed in the reader-side kill-switch (`ColumnarCachedBatchSerializer.scala:911-916`).
> Use during a rolling upgrade with mixed Gluten binaries; flip back to true once the
> cluster is uniform."

cpp/velox/operators/serializer/BatchStatsCollector.cc
```diff
@@ -0,0 +1,710 @@
...
+  const auto size = out.size();
+  out.resize(size + sizeof(T));
+  std::memcpy(out.data() + size, &value, sizeof(T));
+}
+
+// Overwrite `sizeof(T)` bytes starting at `offset` with the little-endian
+// representation of `value`. Used by the string path to backfill a length
+// header after the payload is written.
+template <typename T>
+void writeLEAt(std::vector<uint8_t>& out, size_t offset, T value) {
```
LuciferYang, line 52: `writeLEAt` seems dead code. The doc-comment claims "used by the string path to backfill a
length header after the payload is written", but the actual string branch in `toBytes()`
writes length first, then payload — it never calls `writeLEAt`. `git grep writeLEAt` finds
only this definition.

The comment doesn't match the implementation; future readers chasing the doc will look for
a backfill path that doesn't exist. Delete the function and its `static_assert`. If a future
change wants the backfill shape (reserve a slot, write payload, then patch length),
reintroduce it together with a `reserveLE` helper at that time.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -74,30 +174,470 @@ class CachedColumnarBatchKryoSerializer extends KryoSerializer[CachedColumnarBat
...
+    } else {
+      logDebug(msg)
+    }
   }
 }
 
+object CachedColumnarBatchKryoSerializer {
+  // 0xC0DEC0DE as a signed int is -1059192130 (negative). `numRows` in Spark is non-negative,
+  // so any v0 stream starts with a non-negative int and can never collide with the magic.
+  private[execution] val MAGIC: Int = 0xc0dec0de
```
LuciferYang, line 566: The decimal equivalent in the comment is wrong. `0xC0DEC0DE` as signed int32 is
`-1059143458`, not `-1059192130` (verify: `python3 -c 'print(0xC0DEC0DE - 2**32)'`).

`grep -rn 1059192130 backends-velox/` finds two copies that need the same fix:
- `ColumnarCachedBatchSerializer.scala:564` (production comment)
- `ColumnarCachedBatchSerializerSuite.scala:108` (test comment)

Functionally harmless (the code uses the hex literal), but any docs or tools that grep on
this exact decimal value will mismatch. Either fix both occurrences to `-1059143458` or drop
the specific decimal and keep just the qualitative "negative" claim.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -296,10 +864,799 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+        // under UTF8_LCASE ordering -- Spark would silently drop a batch
+        // that contains a matching row under the query predicate.
+        // We therefore only accept binary (UTF8_BINARY) collation; non-
+        // binary collations fall through to tag-incompat → pass-through
+        // for this batch. `catalogString` returns "string" for UTF8_BINARY
+        // on both 3.x (singleton, always binary) and 4.x (case class with
+        // collationName == "UTF8_BINARY"), and "string collate xxx" for
+        // any other collation. This gives us a shim-safe predicate that
+        // compiles unchanged across Spark 3.3..4.1.
+        dataType.isInstanceOf[StringType] && dataType.catalogString == "string"
```
LuciferYang, line 1413:
```
test("StringType.catalogString stability for UTF8_BINARY default") {
  assert(StringType.catalogString == "string")
}
```

It is recommended to add a simple test case to confirm that this behavior remains consistent across different Spark versions.

cpp/velox/operators/serializer/BatchStatsCollector.cc
```diff
@@ -0,0 +1,710 @@
...
+  columns_.resize(numColumns);
+  columnTypes_.resize(numColumns);
+  for (size_t i = 0; i < numColumns; ++i) {
+    const auto& childType = type.childAt(i);
+    columnTypes_[i] = childType;
+    columns_[i].tag = typeTagFor(childType);
+  }
+}
+
+void BatchStatsCollector::update(const RowVectorPtr& vector) {
```
LuciferYang, line 484: update has no try/catch. The caller (VeloxColumnarBatchSerializer::append) already
handles exceptions: for std::exception / unknown exceptions it calls
statsCollector_.reset(), destroying the unique_ptr (the next append sees nullptr and
skips stats — cache write itself never crashes); for std::bad_alloc it rethrows so OOM
propagates as a cluster-wide fatal error.
Note: toBytes() at BatchStatsCollector.cc:641 returns an empty vector when
schemaDriftPoisoned_ is true, identical to the wire-level outcome of caller-reset followed
by subsequent skip-on-nullptr. Along the std::exception path, both produce a 0-byte stats
payload that the Scala side treats as pass-through.
Adding an in-update try/catch is therefore a hygiene-only change: it localizes exception
handling and avoids the unique_ptr destroy/recreate cost. If you do add it, latch
schemaDriftPoisoned_=true plus a single LOG(WARNING) so the path matches the existing
schema-drift handler.

jackylee-ch:
this would be closed and keep on in https://github.com/apache/gluten/pull/12092

PR #12077, opened 2026-05-12
[GLUTEN-8629][VL] Add RDDScanExec support to Velox backend

minni31:
## CONTEXT

`RDDScanExec` is Spark's physical plan node used when creating DataFrames from in-memory RDDs
via `sparkSession.createDataset(rdd)` or `createDataFrame(rdd, schema)`. Currently, when Gluten
encounters this node with the Velox backend, it falls back to vanilla Spark's row-based execution.
The ClickHouse backend already supports this via `CHRDDScanTransformer`.

## WHAT

This PR adds native Velox execution support for `RDDScanExec` by implementing a
`VeloxRDDScanTransformer`. The transformer converts `RDD[InternalRow]` into Velox columnar
batches using the existing `RowToVeloxColumnarExec` JNI infrastructure, so no new native code
is needed.

Key design decisions:

- **Reuses existing infrastructure**: The transformer delegates to `RowToVeloxColumnarExec`
  for the actual row-to-columnar conversion, keeping the implementation lean and consistent
  with how Velox already handles row-based input.
- **Schema validation**: Delegates to `VeloxValidatorApi.validateSchema` for recursive type
  validation. Supports all Velox-compatible types including complex types (Array, Map, Struct)
  which are handled via the `UnsafeRowFast::deserialize` path in the native converter. Rejects
  only truly unsupported types with clean fallback to vanilla Spark.
- **Leaf node correctness**: `withNewChildrenInternal` returns `copy(...)` consistent with
  CHRDDScanTransformer.
- **Follows existing patterns**: Mirrors the structure of `CHRDDScanTransformer` in the
  ClickHouse backend.

## Changes

- **`VeloxRDDScanTransformer.scala`** (new) — Columnar execution node wrapping
  `RowToVeloxColumnarExec` for native row-to-columnar conversion.
- **`VeloxSparkPlanExecApi.scala`** (modified) — Overrides `isSupportRDDScanExec` and
  `getRDDScanTransform` to wire up the new transformer.
- **`VeloxRDDScanSuite.scala`** (new) — 7 unit tests covering plan replacement, type coverage,
  aggregation, empty RDD, null values, idempotent reads, and all primitive types.

## Test Results

All **7 unit tests** passed on the internal CI pipeline (build 218528457):

| Test Name | Status |
|-----------|--------|
| basic RDDScanExec is replaced by VeloxRDDScanTransformer | ✅ |
| RDDScan with string and numeric types | ✅ |
| RDDScan with aggregation downstream | ✅ |
| RDDScan with empty RDD | ✅ |
| RDDScan preserves data correctness with multiple re-reads | ✅ |
| RDDScan with null values | ✅ |
| RDDScan with all supported primitive types | ✅ |

Related issue: #8629


backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,97 @@
...
+ * conversion (same JNI path as RowToVeloxColumnarExec).
+ */
+case class VeloxRDDScanTransformer(
+    outputAttributes: Seq[Attribute],
+    rdd: RDD[InternalRow],
+    name: String,
+    override val outputPartitioning: Partitioning,
+    override val outputOrdering: Seq[SortOrder]
+) extends RDDScanTransformer(outputAttributes, outputPartitioning, outputOrdering) {
+
```
weiting-chen: **PR description contradicts validation logic for complex types**

**Problem:** The PR description states *"rejects complex types (ARRAY, MAP, STRUCT)"* but `doValidateInternal()` explicitly **accepts** these types. The code is correct — Velox does support complex types via `UnsafeRowFast::deserialize`. The PR description should be updated to avoid misleading reviewers.

**Evidence:**
```scala
case _: org.apache.spark.sql.types.ArrayType =>
case _: org.apache.spark.sql.types.MapType =>
case _: org.apache.spark.sql.types.StructType =>
```
These cases fall through to `ValidationResult.succeeded`, meaning complex types are accepted.

**Suggested Fix:** Update the PR description to remove the claim that complex types are rejected, e.g.:
> *Supports all Velox-compatible types including complex types (Array, Map, Struct). Rejects only truly unsupported types (e.g., CalendarIntervalType) with clean fallback to vanilla Spark.*

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * Converts an RDD[InternalRow] into columnar batches using Velox's native row-to-columnar
+ * conversion (same JNI path as RowToVeloxColumnarExec).
+ */
+case class VeloxRDDScanTransformer(
+    outputAttributes: Seq[Attribute],
+    rdd: RDD[InternalRow],
+    name: String,
+    override val outputPartitioning: Partitioning,
+    override val outputOrdering: Seq[SortOrder]
```
weiting-chen: **Validation does not recurse into complex type element types**

**Problem:** The type allowlist checks top-level types only. An `ArrayType(UnsupportedType)` or `MapType(StringType, UnsupportedType)` would pass validation but could fail at native execution time. The CH backend avoids this by delegating to `ConverterUtils.getTypeNode()` which recursively validates.

**Evidence:**
```scala
case _: org.apache.spark.sql.types.ArrayType =>   // passes any ArrayType, no element check
case _: org.apache.spark.sql.types.MapType =>      // passes any MapType, no key/value check
case _: org.apache.spark.sql.types.StructType =>   // passes any StructType, no field check
```

**Suggested Fix:**
```scala
case a: org.apache.spark.sql.types.ArrayType =>
  validateType(a.elementType)
case m: org.apache.spark.sql.types.MapType =>
  validateType(m.keyType)
  validateType(m.valueType)
case s: org.apache.spark.sql.types.StructType =>
  s.fields.foreach(f => validateType(f.dataType))
```
Alternatively, delegate to `VeloxValidatorApi` for centralized type validation.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,97 @@
...
+        case _: org.apache.spark.sql.types.DateType =>
+        case _: org.apache.spark.sql.types.BinaryType =>
+        case _: org.apache.spark.sql.types.DecimalType =>
+        case _: org.apache.spark.sql.types.ArrayType =>
+        case _: org.apache.spark.sql.types.MapType =>
+        case _: org.apache.spark.sql.types.StructType =>
+        case org.apache.spark.sql.types.YearMonthIntervalType.DEFAULT =>
+        case _: org.apache.spark.sql.types.NullType =>
+        case dt
+            if !VeloxConfig.get.enableTimestampNtzValidation &&
```
weiting-chen: **`withNewChildrenInternal` returns `this` instead of `copy()`**

**Problem:** Returning `this` from a case class's `withNewChildrenInternal` breaks Spark's convention that tree transformations produce structurally new nodes. The CH backend returns `copy(...)` for the equivalent transformer. While this is functionally safe for a leaf node today, it's inconsistent with the project pattern.

**Evidence:**
```scala
// Velox (this PR):
override protected def withNewChildrenInternal(newChildren: IndexedSeq[SparkPlan]): SparkPlan =
  this

// CH backend (CHRDDScanTransformer):
override protected def withNewChildrenInternal(newChildren: IndexedSeq[SparkPlan]): SparkPlan =
  copy(outputAttributes, rdd, name, outputPartitioning, outputOrdering)
```

**Suggested Fix:**
```scala
override protected def withNewChildrenInternal(newChildren: IndexedSeq[SparkPlan]): SparkPlan =
  copy()
```

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,97 @@
...
+        case _: org.apache.spark.sql.types.ByteType =>
+        case _: org.apache.spark.sql.types.ShortType =>
+        case _: org.apache.spark.sql.types.IntegerType =>
+        case _: org.apache.spark.sql.types.LongType =>
+        case _: org.apache.spark.sql.types.FloatType =>
+        case _: org.apache.spark.sql.types.DoubleType =>
+        case _: org.apache.spark.sql.types.StringType =>
+        case _: org.apache.spark.sql.types.TimestampType =>
+        case _: org.apache.spark.sql.types.DateType =>
+        case _: org.apache.spark.sql.types.BinaryType =>
```
weiting-chen: **No SQLMetrics propagation — Spark UI won't show conversion metrics**

**Problem:** The 4-param overload of `toColumnarBatchIterator` creates throwaway `SQLMetric` instances not attached to this plan's metrics map. As a result, `numInputRows`, `numOutputBatches`, and `convertTime` won't appear in the Spark UI for this operator, making production debugging harder.

**Evidence:**
```scala
// Current (4-param overload creates throwaway metrics):
RowToVeloxColumnarExec.toColumnarBatchIterator(iter, localSchema, batchSize, batchBytes)
```

**Suggested Fix:** Define plan-level metrics and use the 7-param overload:
```scala
override lazy val metrics = Map(
  "numInputRows" -> SQLMetrics.createMetric(sparkContext, "number of input rows"),
  "numOutputBatches" -> SQLMetrics.createMetric(sparkContext, "number of output batches"),
  "convertTime" -> SQLMetrics.createTimingMetric(sparkContext, "time to convert"))

override def doExecuteColumnar(): RDD[ColumnarBatch] = {
  val numInputRows = longMetric("numInputRows")
  val numOutputBatches = longMetric("numOutputBatches")
  val convertTime = longMetric("convertTime")
  val localSchema = this.schema
  val batchSize = GlutenConfig.get.maxBatchSize
  val batchBytes = VeloxConfig.get.veloxPreferredBatchBytes
  rdd.mapPartitions { iter =>
    RowToVeloxColumnarExec.toColumnarBatchIterator(
      iter, localSchema, numInputRows, numOutputBatches, convertTime, batchSize, batchBytes)
  }
}
```

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRDDScanSuite.scala
```diff
@@ -0,0 +1,189 @@
...
+
+    val node = LogicalRDD.fromDataset(
+      rdd = data.queryExecution.toRdd,
+      originDataset = data,
+      isStreaming = false)
+    val df = ClassicDataset.ofRows(spark, node).toDF()
+
+    checkAnswer(df, expectedAnswer)
+  }
+}
```
weiting-chen: **Missing test coverage for complex types and unsupported-type fallback**

**Problem:** The 7 tests cover primitives, nulls, empty RDD, and aggregation — but two important scenarios are untested:
1. **Complex types** (ArrayType, MapType, StructType) — validation explicitly accepts them, but no test exercises the full row-to-columnar JNI path with nested data.
2. **Unsupported type fallback** — no test verifies that a truly unsupported type (e.g., `CalendarIntervalType`) triggers graceful fallback to vanilla Spark instead of a runtime crash.

**Suggested Fix:** Add at least these two tests:
```scala
test("RDDScan with array type") {
  val rdd = spark.sparkContext.parallelize(Seq(Row(Seq(1, 2, 3)), Row(Seq(4, 5))))
  val schema = StructType(Seq(StructField("arr", ArrayType(IntegerType))))
  val data = spark.createDataFrame(rdd, schema)
  val expectedAnswer = data.collect()
  val node = LogicalRDD.fromDataset(
    rdd = data.queryExecution.toRdd, originDataset = data, isStreaming = false)
  val df = ClassicDataset.ofRows(spark, node).toDF()
  checkAnswer(df, expectedAnswer)
}

test("RDDScan falls back for unsupported types") {
  // Create RDD with CalendarIntervalType or another unsupported type
  // Verify plan does NOT contain VeloxRDDScanTransformer (i.e., fallback occurred)
}
```

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: Good catch — updated the PR description. It now correctly states that complex types (Array, Map, Struct) are supported via the UnsafeRowFast::deserialize path, and only truly unsupported types trigger fallback

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: Thanks, this is a great point. Replaced the manual allowlist with VeloxValidatorApi.validateSchema which handles recursive validation for complex type elements and also catches variant shredded structs. This keeps validation logic centralized

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: Fixed — now uses copy(outputAttributes, rdd, name, outputPartitioning, outputOrdering) consistent with CHRDDScanTransformer

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: Added numInputRows, numOutputBatches, and convertTime as plan-level metrics and switched to the 7-param toColumnarBatchIterator overload. These will now show up in the Spark UI

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRDDScanSuite.scala
minni31: Added 4 new tests: array type, map type, struct type, and unsupported-type fallback (DayTimeIntervalType → verifies VeloxRDDScanTransformer is absent from plan). Total coverage is now 11 tests.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,92 @@
...
+    for (field <- schema.fields) {
+      val reason = VeloxValidatorApi.validateSchema(field.dataType)
+      if (reason.isDefined) {
+        return ValidationResult.failed(reason.get)
+      }
+    }
+    ValidationResult.succeeded
+  }
+
+  override def doExecuteColumnar(): RDD[ColumnarBatch] = {
```
yaooqinn: `RowToVeloxColumnarExec.toColumnarBatchIterator` does `UnsafeProjection.apply(row)`, which throws on a `BatchCarrierRow` since `PlaceholderRow`'s getters all throw `UnsupportedOperationException`. This can show up via `df.checkpoint()` or user code that does `df.queryExecution.toRdd` and re-wraps with `LogicalRDD.fromDataset`, when the upstream Gluten plan ends in `VeloxColumnarToCarrierRowExec`. `CHRDDScanTransformer.scala` L101-104 detects this and unwraps via `findNextTerminalRow.batch()`. Either mirror that, or fail fast with a clear error for carrier rows and add a checkpoint round-trip test to document the current behavior.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,92 @@
...
+          localSchema,
+          numInputRows,
+          numOutputBatches,
+          convertTime,
+          batchSize,
+          batchBytes)
+    }
+  }
+
+  override protected def withNewChildrenInternal(newChildren: IndexedSeq[SparkPlan]): SparkPlan =
```
yaooqinn: Leaf node — could `assert(newChildren.isEmpty)` here.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,92 @@
...
+    copy(outputAttributes, rdd, name, outputPartitioning, outputOrdering)
+}
+
+object VeloxRDDScanTransformer {
+  def replace(plan: org.apache.spark.sql.execution.RDDScanExec): RDDScanTransformer =
```
yaooqinn: CH uses `UnknownPartitioning(0)`; we pass `plan.outputPartitioning` through. If the original `RDDScanExec` declares e.g. `HashPartitioning`, downstream Velox ops might skip a shuffle based on a hint we never verified survives the row→columnar conversion. Worth either justifying with a comment or aligning with CH.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRDDScanSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+    val data = spark.sql(query)
+    val expectedAnswer = data.collect()
+
+    val node = LogicalRDD.fromDataset(
+      rdd = data.queryExecution.toRdd,
+      originDataset = data,
+      isStreaming = false)
+    val df = ClassicDataset.ofRows(spark, node).toDF()
+
+    checkAnswer(df, expectedAnswer)
```
yaooqinn: This test — and the following `empty RDD` / `multiple re-reads` / `null values` / array / map / struct ones — only does `checkAnswer`. Without a `collectFirst { case _: VeloxRDDScanTransformer => true }` assertion they'd silently pass even if the rewriter stopped offloading (vanilla Spark also gets the right answer). Tests 1 and 2 already assert plan shape; please add the same here.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: Great catch — this is a real bug. If the upstream RDD was produced by a Gluten plan ending in VeloxColumnarToCarrierRowExec (e.g., via df.checkpoint()), the rows would be BatchCarrierRow instances and UnsafeProjection.apply() would throw. Fixed by peeking at the first row and branching: carrier rows are unwrapped directly via BatchCarrierRow.unwrap(), skipping row-to-columnar conversion entirely. This mirrors the CH pattern.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: Added assert(newChildren.isEmpty, "VeloxRDDScanTransformer is a leaf node").

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: Valid concern. Row-to-columnar conversion doesn't change data distribution — it converts row format within each partition, preserving the partition layout. This is consistent with RowToVeloxColumnarExec which also carries through the child's outputPartitioning. Added an inline comment explaining the rationale and the difference from CH's approach

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRDDScanSuite.scala
minni31: You're right — tests 3–11 would silently pass even if offloading stopped working. Added collect { case _: VeloxRDDScanTransformer => true } assertions to all 8 tests that were missing them. The unsupported-type fallback test already asserts the absence of the transformer, so that one was fine as-is.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
```diff
@@ -0,0 +1,111 @@
...
+  )
+
+  override protected def doValidateInternal(): ValidationResult = {
+    for (field <- schema.fields) {
+      val reason = VeloxValidatorApi.validateSchema(field.dataType)
+      if (reason.isDefined) {
+        return ValidationResult.failed(reason.get)
+      }
+    }
+    ValidationResult.succeeded
```
weiting-chen: **Metrics gap in BatchCarrierRow unwrap path**

**Problem:** When the RDD contains `BatchCarrierRow` instances (e.g., from `df.checkpoint()` on a Gluten plan), the code unwraps columnar batches directly without updating `numInputRows`, `numOutputBatches`, or `convertTime`. Spark UI will show zeros for this operator when processing checkpointed data, making performance debugging difficult.

**Evidence:**
```scala
case _: BatchCarrierRow =>
  // No metrics updated here
  (Iterator.single(first) ++ iter).flatMap(row => BatchCarrierRow.unwrap(row))
```

**Suggested Fix:**
```scala
case _: BatchCarrierRow =>
  (Iterator.single(first) ++ iter).flatMap { row =>
    BatchCarrierRow.unwrap(row).map { batch =>
      numOutputBatches += 1
      numInputRows += batch.numRows()
      batch
    }
  }
```

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRDDScanSuite.scala
```diff
@@ -0,0 +1,235 @@
...
+    val result = asRDDScanDF(data)
+
+    // Should still produce correct results via fallback to vanilla Spark
+    checkAnswer(result, expectedAnswer)
+    val cnt = collect(result.queryExecution.executedPlan) {
+      case _: VeloxRDDScanTransformer => true
+    }
+    assert(cnt.isEmpty, "Expected fallback - VeloxRDDScanTransformer should NOT be in plan")
+  }
+}
```
weiting-chen: **Missing test for BatchCarrierRow unwrap path**

**Problem:** The new `BatchCarrierRow` detection logic in `doExecuteColumnar` is production code added to handle checkpointed Gluten DataFrames, but no test exercises this specific branch. If the unwrap logic regresses, the existing tests won't catch it since they all go through the standard InternalRow conversion path.

**Suggested Fix:** Add a test that forces the BatchCarrierRow path:
```scala
test("RDDScan handles BatchCarrierRow from checkpoint") {
  spark.sparkContext.setCheckpointDir(tempPath)
  val df = spark.range(100).selectExpr("id", "id * 2 as value")
  val checkpointed = df.localCheckpoint()
  val result = asRDDScanDF(checkpointed)
  checkAnswer(result, df.collect())
  val cnt = collect(result.queryExecution.executedPlan) {
    case _: VeloxRDDScanTransformer => true
  }
  assert(cnt.nonEmpty, "Expected VeloxRDDScanTransformer in plan")
}
```

weiting-chen:
**Minor: PR title format**

Per CONTRIBUTING.md, the title should include the issue ID: `[GLUTEN-8629][VL] Add RDDScanExec support to Velox backend` instead of the current `[VL] Add RDDScanExec support to Velox backend`.

backends-velox/src/main/scala/org/apache/gluten/execution/VeloxRDDScanTransformer.scala
minni31: updated the BatchCarrierRow unwrap path to increment numOutputBatches and numInputRows per batch, so Spark UI now shows correct metrics for  checkpointed data. convertTime is intentionally omitted since no row-to-columnar conversion happens in this path.

yaooqinn (approved):

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxRDDScanSuite.scala
minni31: Added a localCheckpoint() round-trip test that exercises the BatchCarrierRow detection and unwrap logic. It verifies both result correctness and that  VeloxRDDScanTransformer is present in the plan.

PR #12083, opened 2026-05-13
[VL] Fix SCOPED_TIMER macro destroying timer immediately

luis4a0:

The `SCOPED_TIMER` macro in `cpp/velox/utils/Common.h` wraps its
`DeltaCpuWallTimer` in a `do { ... } while (0)` block. The
`DeltaCpuWallTimer` is constructed inside the do-block, so it is also
**destroyed at the closing brace of the do-block — i.e. immediately
after construction**. The lambda fires recording essentially zero ns,
and the surrounding scope (which the call sites intend to time) is
not measured at all.

As a consequence, every existing `SCOPED_TIMER` call in the Velox
shuffle writers (`VeloxHashShuffleWriter.cc`,
`VeloxRssSortShuffleWriter.cc`) has been silently recording near-zero
values into `cpuWallTimingList_`. The per-bucket time breakdown
(Compute, FlattenRV, SplitRV, EvictPartition, Stop, …) that
`stat()` would log under `VELOX_SHUFFLE_WRITER_LOG_FLAG=1` is
unusable for performance analysis as a result.

This PR replaces the do-while wrapper with two uniquely-named local
variables that survive until the end of the **enclosing** scope —
which is what every call site already expects from a SCOPED-style
RAII timer.

```cpp
#define SCOPED_TIMER(timing)                                                                         \
  auto GLUTEN_SCOPED_TIMER_CONCAT(_glutenScopedTimerTarget_, __LINE__) = &(timing);                  \
  facebook::velox::DeltaCpuWallTimer GLUTEN_SCOPED_TIMER_CONCAT(_glutenScopedTimer_, __LINE__) {     \
    [_glutenScopedTimerTargetPtr = GLUTEN_SCOPED_TIMER_CONCAT(_glutenScopedTimerTarget_, __LINE__)]( \
        const facebook::velox::CpuWallTiming& delta) { _glutenScopedTimerTargetPtr->add(delta); }    \
  }
```

A few implementation notes:

- The unique-naming pattern (using `__LINE__`) is needed so that
  multiple `SCOPED_TIMER` calls in the same scope don't shadow each
  other.
- The extra local for the timing target (a pointer) is needed because
  the `timing` macro argument may be an arbitrary expression
  (e.g. `cpuWallTimingList_[CpuWallTimingCompute]`) that is not a
  valid lambda capture name. Capturing a pointer to it sidesteps that.

## How was this patch tested?

This PR adds a new test file `cpp/velox/tests/ScopedTimerTest.cc`
(second commit) with three GTest cases:

- `accumulatesElapsedTime` — single `SCOPED_TIMER` over a 50 ms sleep,
  expects `wallNanos > 25 ms`.
- `multipleTimersInSameScope` — two `SCOPED_TIMER` calls in the same
  scope (validates the `__LINE__`-based unique naming).
- `acceptsArrayElementAsTarget` — target is `vec[1]`, validates the
  pointer-indirection on non-trivial timing-target expressions.

I additionally validated the test logic with a standalone harness
that compiles the **same** test source against both versions of the
macro (gated by `-DUSE_BUGGY_SCOPED_TIMER`). Concretely, the in-tree
test file would have caught the bug at PR time — every case fails
loudly under the old macro and passes under the new one:

| Test                             | Fixed (this PR) | Buggy (original) |
|----------------------------------|----------------:|-----------------:|
| `accumulatesElapsedTime`         | wall ≈ 50 ms ✅ | wall ≈ 200 ns ❌ |
| `multipleTimersInSameScope` (t1) | wall ≈ 20 ms ✅ | wall ≈ 170 ns ❌ |
| `multipleTimersInSameScope` (t2) | wall ≈ 20 ms ✅ | wall ≈ 40 ns  ❌ |
| `acceptsArrayElementAsTarget`    | wall ≈ 20 ms ✅ | wall ≈ 50 ns  ❌ |

The buggy version records wall time off by a million-fold (50 ns
instead of 50 000 000 ns), exactly matching the predicted "timer
destructs immediately after construction" behavior. The fixed version
records wall time correctly within scheduler jitter of the sleep.

I also verified the macro change compiles cleanly across all 18 call
sites of `SCOPED_TIMER` in
`cpp/velox/shuffle/Velox{Hash,RssSort}ShuffleWriter.cc`.

End-to-end runtime validation in a real Spark/Gluten run (running
TPC-DS with `VELOX_SHUFFLE_WRITER_LOG_FLAG=1` and confirming the
per-bucket `cpuWallTimingList_` numbers are now non-zero) was not
possible in my development environment because a system-wide
gluten-velox JAR took precedence over the locally-built OSS bundle on
the JNI symbol resolution path, shadowing the patched `libvelox.so`.
The CI-built bundle will not have that issue.

Marked as draft for one more review pass before flipping to ready.

### DIFF
--- a/cpp/velox/utils/Common.h
+++ b/cpp/velox/utils/Common.h
@@ -21,6 +21,7 @@
 #include <memory>
 #include <string>
 
+#include "utils/Macros.h"
 #include "velox/common/base/SimdUtil.h"
 #include "velox/common/time/CpuWallTimer.h"
 
@@ -41,21 +42,20 @@ static inline void fastCopy(void* dst, const void* src, size_t n) {
 //
 // Implementation note: a `do { … } while (0)` wrapper would destroy
 // the DeltaCpuWallTimer immediately at the end of the do-block,
-// recording essentially zero. Use uniquely-named local variables at
-// the call site so the timer survives until the enclosing scope
-// closes (which is what every existing call site already expects).
+// recording essentially zero. Instead we declare uniquely-named
+// locals (via GLUTEN_CONCAT(..., __LINE__)) so the timer survives
+// until the enclosing scope closes — which is what every call site
+// already expects.
 //
 // Note: we capture the timing target by pointer in a unique local
 // variable rather than directly in the lambda, because `timing` may
 // be an arbitrary expression (e.g. `array_[idx]`) that is not a valid
 // lambda capture name.
-#define GLUTEN_SCOPED_TIMER_CONCAT_INNER(a, b) a##b
-#define GLUTEN_SCOPED_TIMER_CONCAT(a, b) GLUTEN_SCOPED_TIMER_CONCAT_INNER(a, b)
-#define SCOPED_TIMER(timing)                                                                         \
-  auto GLUTEN_SCOPED_TIMER_CONCAT(_glutenScopedTimerTarget_, __LINE__) = &(timing);                  \
-  facebook::velox::DeltaCpuWallTimer GLUTEN_SCOPED_TIMER_CONCAT(_glutenScopedTimer_, __LINE__) {     \
-    [_glutenScopedTimerTargetPtr = GLUTEN_SCOPED_TIMER_CONCAT(_glutenScopedTimerTarget_, __LINE__)]( \
-        const facebook::velox::CpuWallTiming& delta) { _glutenScopedTimerTargetPtr->add(delta); }    \
+#define SCOPED_TIMER(timing)                                                                      \
+  auto GLUTEN_CONCAT(_glutenScopedTimerTarget_, __LINE__) = &(timing);                            \
+  facebook::velox::DeltaCpuWallTimer GLUTEN_CONCAT(_glutenScopedTimer_, __LINE__) {               \
+    [_glutenScopedTimerTargetPtr = GLUTEN_CONCAT(_glutenScopedTimerTarget_, __LINE__)](           \
+        const facebook::velox::CpuWallTiming& delta) { _glutenScopedTimerTargetPtr->add(delta); } \
   }
 
 } // namespace gluten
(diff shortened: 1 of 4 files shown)


baibaichen:
@marin-ma pelease review it

cpp/velox/utils/Common.h
```diff
@@ -35,11 +35,27 @@ static inline void fastCopy(void* dst, const void* src, size_t n) {
...
+// recording essentially zero. Use uniquely-named local variables at
+// the call site so the timer survives until the enclosing scope
+// closes (which is what every existing call site already expects).
+//
+// Note: we capture the timing target by pointer in a unique local
+// variable rather than directly in the lambda, because `timing` may
+// be an arbitrary expression (e.g. `array_[idx]`) that is not a valid
+// lambda capture name.
+#define GLUTEN_SCOPED_TIMER_CONCAT_INNER(a, b) a##b
+#define GLUTEN_SCOPED_TIMER_CONCAT(a, b) GLUTEN_SCOPED_TIMER_CONCAT_INNER(a, b)
```
marin-ma: This is a general macro. Can you use `GLUTEN_CONCAT` in `core/utils/Macros.h`

luis4a0: Done in 771d0e0. Promoted `GLUTEN_CONCAT` in `core/utils/Macros.h` to a two-level macro — the existing form was single-level (`GLUTEN_CONCAT(foo, __LINE__)` would have expanded to literal `foo__LINE__` since `##` suppresses argument expansion), but it had zero callers in the cpp/ tree, so the upgrade is backward-compatible. Style mirrors the existing `GLUTEN_STRINGIFY` / `GLUTEN_TOSTRING` pair just above it. Per-file helpers removed from `Common.h`, which now `#include`s `utils/Macros.h` and uses `GLUTEN_CONCAT` directly. `ScopedTimerTest` continues to pass with identical timing profile. Thanks for the catch.

marin-ma:
@luis4a0 LGTM. I see this PR is marked as draft. Is it ready for review now?

luis4a0:
> @luis4a0 LGTM. I see this PR is marked as draft. Is it ready for review now?

@marin-ma Yes, I marked it as ready for review. I should do it earlier, sorry.

marin-ma (approved):

PR #12087, opened 2026-05-13
[VL] Simplify the logic in AppendBatchResizeForShuffleInputAndOutput

marin-ma:
The current logic in `AppendBatchResizeForShuffleInputAndOutput` is a bit convoluted and unclear. Refactor it by separating the logic into `addResizeBatchesForShuffleInput` and `addResizeBatchesForShuffleOutput`.

For `addResizeBatchesForShuffleInput`, simply wrap and replace the child of exchange.
For `addResizeBatchesForShuffleOutput`, when AQE is enabled, extract and wrap the exchange inside `ShuffleQueryStageExec`, or `AQEShuffleReadExec`

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
@@ -28,11 +28,13 @@ import org.apache.spark.sql.execution.exchange.ReusedExchangeExec
  * Try to append [[VeloxResizeBatchesExec]] for shuffle input and output to make the batch sizes in
  * good shape.
  */
-case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
+case class AppendBatchResizeForShuffleInputAndOutput(isAdaptiveContext: Boolean)
+  extends Rule[SparkPlan] {
   override def apply(plan: SparkPlan): SparkPlan = {
     if (VeloxConfig.get.enableColumnarCudf) {
       return plan
     }
+
     val resizeBatchesShuffleInputEnabled = VeloxConfig.get.veloxResizeBatchesShuffleInput
     val resizeBatchesShuffleOutputEnabled = VeloxConfig.get.veloxResizeBatchesShuffleOutput
     if (!resizeBatchesShuffleInputEnabled && !resizeBatchesShuffleOutputEnabled) {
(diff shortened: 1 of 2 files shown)


marin-ma:
@wForget Could you help to review? Thanks.

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -22,17 +22,18 @@ import org.apache.gluten.execution.VeloxResizeBatchesExec
...
 import org.apache.spark.sql.execution.{ColumnarShuffleExchangeExec, SparkPlan}
 import org.apache.spark.sql.execution.adaptive.{AQEShuffleReadExec, ShuffleQueryStageExec}
-import org.apache.spark.sql.execution.exchange.ReusedExchangeExec
 
 /**
  * Try to append [[VeloxResizeBatchesExec]] for shuffle input and output to make the batch sizes in
  * good shape.
  */
-case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
+case class AppendBatchResizeForShuffleInputAndOutput(isAdaptiveContext: Boolean)
```
wForget, line 30: Where is `isAdaptiveContext` used?

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -41,65 +42,59 @@ case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
...
-        VeloxResizeBatchesExec(s, range.min, range.max, preferredBatchBytes)
+    }
+  }
+
+  private def addResizeBatchesForShuffleOutput(
+      plan: SparkPlan,
+      min: Int,
+      max: Int,
+      preferredBatchBytes: Long): SparkPlan = {
+    plan match {
```
wForget, line 80: Is it effective to directly match the root node of this plan?

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -41,65 +42,59 @@ case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
...
+      case a @ AQEShuffleReadExec(s @ ShuffleQueryStageExec(_, _, _), _)
+          if requiresResizingShuffleOutput(s) =>
+        VeloxResizeBatchesExec(a, min, max, preferredBatchBytes)
+      case other =>
+        other.withNewChildren(other.children.map(
+          p => addResizeBatchesForShuffleOutput(p, min, max, preferredBatchBytes)))
+    }
+  }
+
+  private def requiresResizingShuffleOutput(s: ShuffleQueryStageExec): Boolean = {
```
wForget, line 91: Why were scenarios for ReusedExchangeExec removed?

```
ShuffleQueryStageExec(
    _,
    ReusedExchangeExec(_, shuffle: ColumnarShuffleExchangeExec),
    _),
```

marin-ma: Both Exchange and ReusedExchange will be handled by `s.shuffle` in the line below

https://github.com/apache/spark/blob/26d44f3922b423738760610a51c4daf6f4f58f8d/sql/core/src/main/scala/org/apache/spark/sql/execution/adaptive/QueryStageExec.scala#L203-L208

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -41,65 +42,59 @@ case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
...
+      min: Int,
+      max: Int,
+      preferredBatchBytes: Long): SparkPlan = {
+    plan match {
+      case s: ShuffleQueryStageExec if requiresResizingShuffleOutput(s) =>
+        VeloxResizeBatchesExec(s, min, max, preferredBatchBytes)
+      case a @ AQEShuffleReadExec(s @ ShuffleQueryStageExec(_, _, _), _)
+          if requiresResizingShuffleOutput(s) =>
+        VeloxResizeBatchesExec(a, min, max, preferredBatchBytes)
+      case other =>
```
marin-ma, line 86: @wForget The normal entry point for the root node is `case other`. It traverses the whole plan until a AQE query stage leaf node (plain `ShuffleQueryStageExec`, or `AQEShuffleReadExec` wrapped `AQEShuffleReadExec`) is found.

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
marin-ma: In the original logic, the "resize batches for shuffle output"  only applies to aqe stage nodes, so `addResizeBatchesForShuffleOutput` is a no-op when AQE is off.

Thanks for catching and this should be added [here](https://github.com/apache/gluten/pull/12087/changes#diff-94c14147622eeb96705fe5fc473a0d6876edb5d18cb4f412e168134f44feec6cR52) to indicate the rule doesn't apply when AQE is off.

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
wForget: Sorry, I missed this recursive call.

backends-velox/src/main/scala/org/apache/gluten/extension/AppendBatchResizeForShuffleInputAndOutput.scala
```diff
@@ -41,65 +42,59 @@ case class AppendBatchResizeForShuffleInputAndOutput() extends Rule[SparkPlan] {
...
+        other.withNewChildren(other.children.map(
```
wForget: Perhaps it could be changed to:
```
other.mapChildren(addResizeBatchesForShuffleOutput(_, min, max, preferredBatchBytes))
```

marin-ma: Thanks. Updated.

wForget (approved):
Thanks @marin-ma

PR #12088, opened 2026-05-13
[VL] minor: Use isNullAt to check whether a value is null in VeloxHashShuffleWriter

wankunde:

According to https://github.com/facebookincubator/velox/blob/main/velox/vector/BaseVector.h#L241-L250
we should not use `rawNulls` directly,  so use `isNullAt()`  method to check whether a value is null in VeloxHashShuffleWriter

## How was this patch tested?

Minor change, local test

No


marin-ma:
The reason we don't use `isNullAt()` is that it's a virtual function. Since it's on the hotspot code path, we want to avoid virtual function calls as much as possible.

cc: @FelixYBW

PR #12090, opened 2026-05-13
[VL] Add concurrency control to GpuLock

marin-ma:
Currently, GPULock allows only one Spark task to acquire GPU resources at a time, which effectively makes GPU offloading run in single-threaded mode within a Spark executor. However, when testing the TPCDS queries, we observed that GPU utilization and memory usage can be very low (< 20%), meaning GPU resources are underutilized.

To improve GPU utilization, this PR adds support for running multiple GPU tasks concurrently through the configuration `spark.gluten.sql.columnar.backend.velox.cudf.concurrentGpuTasks`. The default value is 1.


zhouyuan (approved):

marin-ma:
Verified locally `velox_gpu_lock_test` has passed.

PR #12092, opened 2026-05-14
[VL] Add min/max partition stats to columnar InMemoryRelation cache for partition pruning

yaooqinn:

This PR fills the long-standing TODO in `ColumnarCachedBatchSerializer.buildFilter`
by adding native-side min/max partition statistics to the columnar `InMemoryRelation`
cache, enabling partition-level pruning when filtering over a cached `DataFrame`.

The Velox backend now computes per-batch min/max/null-count stats during
`serializeWithStats` (cpp side, Arrow C-Data export-free) and embeds them
in the cached batch envelope. On the read path, the JVM serializer extends
`SimpleMetricsCachedBatchSerializer` and routes batches with stats through
the inherited `buildFilter` for partition pruning; batches without stats
(legacy v1 binary, or the SQLConf gate disabled) lazy-pass-through unchanged.

**Type matrix covered (vanilla Spark `ColumnStats.scala` parity):**
- Integer family: TINYINT / SMALLINT / INT / BIGINT
- Date / Timestamp / TimestampNTZ
- YearMonth / DayTime intervals
- Decimal (short P<19 / long P>=19 / HUGEINT)
- String (JVM marshal; cpp VARCHAR scan deferred — see notes)
- Boolean
- Float / Double (NaN-poison guarded)

VARCHAR cpp-side scan is intentionally deferred (`supported=0` from cpp,
no JVM crash path) due to Velox StringView lifetime considerations across
RowVector boundaries — tracked for a follow-up PR.

Partition stats let `InMemoryRelation` skip whole cached partitions that
cannot satisfy a filter predicate, materially speeding up repeated point
lookups and selective filters over a cached DataFrame — a common pattern
in iterative SQL workloads (notebooks, ML feature engineering, BI tools
with cached intermediate results). Vanilla Spark has had this since
SPARK-32274 (3.1, ~5 years); Gluten's columnar cache previously
extended `CachedBatchSerializer` directly, leaving the `buildFilter` TODO
empty and skipping pruning entirely.

### Does this PR introduce _any_ user-facing change?

Yes — a new SQLConf:

| Key | Default | Meaning |
|---|---|---|
| `spark.gluten.sql.columnar.tableCache.partitionStats.enabled` | `false` | Enable native partition stats computation + read-side pruning |

Default-off ship for the first release; flip to default-on after community
feedback on the included benchmark + a release of soak time. Gate is
double-checked on both the write path (skip stats compute) and read path
(via the inherited `buildFilter`).

**Unit / integration tests:**
- 19 cpp gtest cases (Velox `velox_operators_test`) — per-type min/max scan
  + framing + carry-overflow + NaN poison guard + truncate semantics
- 8 JVM Scala suites, 37 cases — Kryo wire format / stats blob marshal /
  buildFilter wrapper / per-arm prune semantics (BIGINT / Date / Timestamp /
  String / NaN sentinel / numCols cap)
- 1 Velox e2e suite (`ColumnarCachedBatchE2ESuite`) — cache + equality
  filter end-to-end, plan shape verification, `numOutputRows` prune
  evidence

**Benchmark** (`backends-velox/benchmarks/ColumnarTableCachePartitionStatsBenchmark-results.txt`):

100M-row range partitioned by `c2` into 32 in-memory partitions,
`groupBy + sum/count/avg` follow-up to make pruned-partition savings
visible. Driver: `local[1]`, AMD EPYC 7763.

| Case | partitionStats off | partitionStats on | Speedup |
|---|---|---|---|
| cache build (write path) | 126,425 ms | 131,431 ms | **1.0x** (no measurable overhead) |
| filter+agg, high selectivity (`c2 < 1000`, ~0.001%) | 4,431 ms | 1,744 ms | **2.5x** |
| filter+agg, low selectivity (`c2 < 50000000`, ~50%) | 5,332 ms | 3,392 ms | **1.6x** |
| filter+agg, point lookup (`c2 = 50000000`, 1 row) | 4,343 ms | 1,686 ms | **2.6x** |

Write path shows no measurable overhead — stats computation runs in cpp
inside the existing batch scan loop.

### Notes on commit history

This PR retains the atomic per-slice commit history (~58 commits) rather
than squashing, to make per-feature `git bisect` and incremental review
practical. Each commit is self-contained: a feature slice, its test, and
any fixes amended into the slice.


gluten-arrow/src/main/java/org/apache/gluten/vectorized/ColumnarBatchSerializerJniWrapper.java
```diff
@@ -48,4 +54,27 @@ public long rtHandle() {
...
   public native void close(long serializerHandle);
+
+  // Capability check for the serializeWithStats JNI extension. A new Gluten jar paired with an
+  // older libgluten.so may lack this symbol; the cache write path consults this helper and
+  // falls back to the legacy serialize() when false. Probing is lazy and one-shot: reflection
+  // on the declared native method proves the JVM side is wired (the cpp symbol is verified at
+  // first real invocation; callers must catch UnsatisfiedLinkError there too).
+  private static volatile Boolean supportsStatsExtCached = null;
+
+  public static boolean supportsStatsExt() {
```
weiting-chen: **`supportsStatsExt()` reflects on Java declaration, not native symbol**

**Problem:** `getDeclaredMethod("serializeWithStats", long.class)` checks the Java method declaration, which will always exist in this class regardless of whether the loaded `libgluten.so` contains the corresponding JNI symbol. If a newer jar is paired with an older native library, enabling `partitionStats.enabled=true` will throw `UnsatisfiedLinkError` on first invocation instead of gracefully falling back.

**Evidence:**
```java
public static boolean supportsStatsExt() {
    Boolean cached = supportsStatsExtCached;
    if (cached != null) return cached;
    boolean result;
    try {
      ColumnarBatchSerializerJniWrapper.class.getDeclaredMethod(
          "serializeWithStats", long.class); // always succeeds
      result = true;
    } catch (NoSuchMethodException e) {
      result = false; // never reached
    }
    // ...
}
```

**Suggested Fix:** Wrap the actual `serializeWithStats()` callsite in `ColumnarCachedBatchSerializer.scala` with a try/catch for `UnsatisfiedLinkError` (one-shot cached), or probe native linkage here by invoking the method with an invalid handle and distinguishing `GlutenException` (linked) from `UnsatisfiedLinkError` (not linked).

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
```diff
@@ -95,4 +98,405 @@ std::shared_ptr<ColumnarBatch> VeloxColumnarBatchSerializer::deserialize(uint8_t
...
+  framed.push_back(0xFE);
+  framed.push_back(0xCA);
+  framed.push_back(0x53);
+  framed.push_back(0x02);
+  framed.push_back(static_cast<uint8_t>(statsLen & 0xFF));
+  framed.push_back(static_cast<uint8_t>((statsLen >> 8) & 0xFF));
+  framed.push_back(static_cast<uint8_t>((statsLen >> 16) & 0xFF));
+  framed.push_back(static_cast<uint8_t>((statsLen >> 24) & 0xFF));
+  framed.insert(framed.end(), statsBlob.begin(), statsBlob.end());
+  const uint32_t bytesLen32 = static_cast<uint32_t>(bytesLen);
```
weiting-chen, line 558: **`uint32_t` truncation on serialized batch length — missing bounds check**

**Problem:** `maxSerializedSize()` returns `int64_t` but is silently cast to `uint32_t` for the framed wire format. If a single batch exceeds 4GB (pathological but possible with very wide schemas or huge string columns), the length wraps around silently, producing a corrupt frame that the JVM will mis-parse.

**Evidence:**
```cpp
const int64_t bytesLen = maxSerializedSize();
std::vector<uint8_t> bytesBlob(bytesLen);
// ...
const uint32_t bytesLen32 = static_cast<uint32_t>(bytesLen);  // silent truncation
```

**Suggested Fix:**
```cpp
GLUTEN_CHECK(
    bytesLen >= 0 && bytesLen <= std::numeric_limits<uint32_t>::max(),
    "Serialized batch size (" + std::to_string(bytesLen) +
    ") exceeds u32 framing limit");
const uint32_t bytesLen32 = static_cast<uint32_t>(bytesLen);
```

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -81,43 +133,443 @@ class CachedColumnarBatchKryoSerializer extends KryoSerializer[CachedColumnarBat
       length != Kryo.NULL,
       "The object 'CachedColumnarBatch.bytes' is invalid or malformed to " +
         s"deserialize using ${this.getClass.getName}")
-    val bytes = new Array[Byte](length - 1) // -1 to restore
+    val bytes = new Array[Byte](length - 1)
     input.readBytes(bytes)
-    CachedColumnarBatch(numRows, sizeInBytes, bytes)
+    val hasStats = input.readBoolean()
```
weiting-chen: **Kryo `read()` not backward-compatible with pre-stats cached batches**

**Problem:** `input.readBoolean()` is called unconditionally after the `bytes` payload. Previously serialized `CachedColumnarBatch` instances (e.g., from `DISK_ONLY` or `MEMORY_AND_DISK` storage levels surviving a rolling upgrade) lack the trailing `hasStats`/`hasSchema` booleans, causing `KryoException` on deserialization.

**Evidence:**
```scala
val bytes = new Array[Byte](length - 1)
input.readBytes(bytes)
val hasStats = input.readBoolean()  // throws if stream ends here (old format)
```

**Suggested Fix:** Guard with available-byte check to gracefully handle old format:
```scala
val hasStats = if (input.available() > 0) input.readBoolean() else false
val stats: InternalRow = if (hasStats) { /* ... */ } else null
val hasSchema = if (input.available() > 0) input.readBoolean() else false
```

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -213,14 +665,40 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+            val handle =
+              ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, batch)
+            // Route through serializeWithStats when the JNI extension is available AND the
+            // partition-stats conf is enabled. Capability is cached after first probe. When
+            // unavailable (older libgluten.so without the symbol, or conf left off) we fall back
+            // to the original serialize() path and emit stats=null; the buildFilter wrapper
+            // directs such batches through without pruning.
+            val partitionStatsEnabled =
+              GlutenConfig.get.getConf(GlutenConfig.COLUMNAR_TABLE_CACHE_PARTITION_STATS_ENABLED)
+            if (partitionStatsEnabled && ColumnarBatchSerializerJniWrapper.supportsStatsExt()) {
```
weiting-chen: **Config-gate negative test gap**

**Problem:** All E2E tests in `ColumnarCachedBatchE2ESuite` set `partitionStats.enabled=true`. No integration test verifies that when the config is `false` (the production default), stats are NOT computed and pruning does NOT occur. A bug in the gating logic could silently activate stats for all users.

**Evidence:** The gate check is here:
```scala
if (partitionStatsEnabled && ColumnarBatchSerializerJniWrapper.supportsStatsExt()) {
  val framed = jni.serializeWithStats(handle)
```
But this path is never exercised with `partitionStatsEnabled=false` in E2E tests.

**Suggested Fix:** Add a test to `ColumnarCachedBatchE2ESuite`:
```scala
test("config disabled: no stats in cached batch, no pruning") {
  withSQLConf(
    GlutenConfig.COLUMNAR_TABLE_CACHE_PARTITION_STATS_ENABLED.key -> "false"
  ) {
    val df = spark.range(1000).selectExpr("id as c1").repartition(4).cache()
    try {
      df.count() // materialize
      val filtered = df.filter("c1 = 500")
      checkAnswer(filtered, Row(500))
      // Verify full scan (no partition pruning)
    } finally df.unpersist()
  }
}
```

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -213,14 +665,40 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
-              .serialize(ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, batch))
-            val bytes = unsafeBuffer.toByteArray
-            CachedColumnarBatch(batch.numRows(), bytes.length, bytes)
+            val jni = ColumnarBatchSerializerJniWrapper.create(
+              Runtimes.contextInstance(
+                BackendsApiManager.getBackendName,
+                "ColumnarCachedBatchSerializer#serialize"))
+            val handle =
+              ColumnarBatches.getNativeHandle(BackendsApiManager.getBackendName, batch)
+            // Route through serializeWithStats when the JNI extension is available AND the
```
weiting-chen: **Config is write-path only — disabling after cache build still prunes**

**Problem:** The config `partitionStats.enabled` only gates the write path (stats computation during `serialize`). On the read path, `buildFilter` inherits from `SimpleMetricsCachedBatchSerializer` and will apply pruning whenever stats are present in the cached batch — regardless of the config value. If a user enables the config, caches data, then disables it, pruning continues until the cache is invalidated.

This is arguably correct behavior (stats already embedded should be used), but it may surprise users who expect the config to be a full kill-switch.

**Investigation Needed:** Consider either:
1. **Document this** — add a note to the config description: "Controls stats computation on the write path. Cached batches that already contain stats will continue to use them for pruning."
2. **Gate read path too** — check the config in the `buildFilter` override and return `Seq.empty` when disabled.

weiting-chen (commented):
Thanks for this excellent PR — the partition stats design is clean and well-tested. A few suggestions for hardening (none are merge-blocking).

liuneng1994 (commented):
Thanks for the substantial work and the thorough TDD trail. Reviewed focusing on Spark behavior parity and test coverage. Three findings can return wrong results / crash and aren't covered by tests; flagging before any future default-on flip.

## 🔴 1. Non-binary collation strings can be **wrongly pruned (silent wrong results)**

`CachedColumnarBatchKryoSerializer.isDispatchable` accepts any `StringType` (the comment even says "any collation"):

```scala
case _: StringType => true // truncated to 256B; see encodeStringBounds (any collation)
```

But both the cpp side (`scanMinMax<StringView>` → `memcmp`) and JVM `encodeStringBounds` (byte-wise truncate + `+1` carry) compute min/max in **unsigned byte order**. For non-binary collations (`UTF8_LCASE`, `UNICODE_CI`, …) Spark's equality / ordering is not byte order, so a partition whose true min/max under the collation overlaps the predicate range may have a byte-order min/max that does not — the partition gets pruned and matching rows disappear.

Suggested fix: gate on the binary collation only, e.g.
```scala
case s: StringType if s.collationId == CollationFactory.UTF8_BINARY_COLLATION_ID => true
```
and add an e2e test under a non-binary collation.

## 🔴 2. CONSTANT/DICTIONARY-encoded columns + `IsNull` can prune wrongly

```cpp
if (child == nullptr || !child->isFlatEncoding()) { continue; }   // computeStats
```

For dict/const-encoded children, `nullCount` stays at `0` and we then emit `nullCount=0, count=numRows, supported=0`. The vanilla `SimpleMetricsCachedBatchSerializer.buildFilter` evaluates `IsNull(col)` purely from `nullCount` (it does not require `supported`/lower/upper) — so it concludes "no nulls in this partition" and prunes it. If the column is in fact a CONSTANT-null vector (very common after outer joins or projection of literal null), `IsNull(col)` will incorrectly return zero rows.

Suggested fix: at minimum decode `nullCount` for non-flat encodings (e.g. via `DecodedVector` or `child->mayHaveNulls()` + per-row check), or skip writing the `nullCount` slot so the parent treats it as unknown. New test needed: cache a frame containing an all-null column on the non-flat path, then filter with `col.isNull` and assert the count equals N (not 0).

## 🟠 3. `supportsStatsExt()` does not actually probe the cpp symbol

```java
public static boolean supportsStatsExt() {
  ...
  ColumnarBatchSerializerJniWrapper.class.getDeclaredMethod("serializeWithStats", long.class);
  ...
}
```

This reflects on the JVM class's own declared method, which is always true once the PR lands. The comment acknowledges "the cpp symbol is verified at first real invocation; callers must catch UnsatisfiedLinkError there too" — but `convertColumnarBatchToCachedBatch` does not catch. So a new gluten jar paired with an older `libgluten.so` will hard-crash with `UnsatisfiedLinkError` instead of falling back to the legacy `serialize()` path that the comment promises.

Suggested fix: try/catch `UnsatisfiedLinkError` once around an inert probe call (or do `dlsym` natively at lib-load), cache the result, and only then dispatch to `serializeWithStats`.

---

## 🟡 Smaller correctness / parity notes

- **NaN poison preserves partial `nullCount`.** `scanMinMax` returns false on the first NaN; the `nullCnt` accumulated up to that point is still written into stats. It happens to be conservative for `IsNull/IsNotNull` (always under-counts), but it is a behavior divergence from `ColumnStats.gatherNullStats` and is fragile to future predicate-eval changes — recommend zeroing nullCount when the column is NaN-poisoned.
- **TimestampNTZ.** JVM dispatch lumps `TimestampType | TimestampNTZType` into the 8B Long arm; cpp only inspects `TypeKind::TIMESTAMP`. Worth verifying Velox's NTZ TypeKind in Gluten's mapping and adding an NTZ e2e — currently no NTZ test.
- **`buildFilter` lazy-split calls `parent(index, runIt)` per contiguous run.** Each invocation may compile a fresh codegen filter; in mixed (stats / no-stats) iterators this can balloon. Consider memoizing or pre-warming the parent closure.

---

## Test coverage gaps to add before default-on

The unit / wire tests are thorough, but the following e2e gaps should be closed before flipping the default to `true`:

1. **Non-binary collation string column** + equality / range filter (covers #1).
2. **CONSTANT/dict-encoded all-null column** + `col.isNull` filter (covers #2).
3. **Stale-symbol fallback**: simulate `serializeWithStats` missing on the cpp side and assert graceful fallback to the legacy path (covers #3).
4. **TimestampNTZ** end-to-end equality.
5. **Cross-config**: cache built with stats=on, read after toggling stats=off (and vice versa); also mixed v1-binary + v2-stats batches in the same partition.
6. **Disk spill via Kryo**: existing Kryo tests round-trip in-memory only; cover the `DiskStore` spill + read-back path with stats present.
7. **Mixed `supported=1` and `supported=0` columns in one batch** — wire alignment is only implicitly tested.
8. **Decimal with `scale > 0`** and **long-Decimal (precision > 18)** end-to-end.
9. **YearMonth / DayTime interval** end-to-end (only marshal unit tests today).
10. **Schema mismatch**: stats column count vs `schema.length` mismatch — currently relies on caller invariant with no `require`.

---

## Non-blocking polish

- The production code carries a lot of TDD-process noise (`PA-3.2 RED`, `PA-6.A`, `refs: todos/features/…`). These are valuable in the commit messages but are likely to confuse future maintainers when they appear in `// comments` of merged code — consider rewriting them as forward-looking design comments at squash / merge.
- `serializeStats(stats, schema)` accepts `schema == null` and falls back to BIGINT; the write path always passes a non-null schema, but no `require` enforces it. A future caller could regress silently — recommend `require(schema != null, …)` once V1-only paths are gone.
- `count` / `nullCount` are `uint32` on the wire — fine in practice, but a comment about the >2³¹-row assumption (or widening to `uint64`) would help.
- cpp `framedSerializeWithStats` writes `pushU64(0)` for per-column `sizeInBytes`; a one-line comment that this is intentional because `CachedColumnarBatch.sizeInBytes` is overridden at the batch level would prevent future "fix me" PRs.

---

Overall the design is clean and the TDD discipline is great; default-off ship is the right call. With #1–#3 fixed and the e2e gaps closed, this should be solid for a default-on flip.

jackylee-ch:
@yaooqinn it looks similar to the pr(#12073 ) I submitted before, both of which implement filter pushdown for the cache, with only slight differences in implementation details.

jackylee-ch:
it  seems the current pr also implement the CachedBatch skipping with pushed filter not just the partition, right?

jackylee-ch:
image
here is the main different between this two prs.

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
```diff
@@ -95,4 +98,405 @@ std::shared_ptr<ColumnarBatch> VeloxColumnarBatchSerializer::deserialize(uint8_t
...
+//
+// Floating-point edge cases that DO NOT poison the column:
+// - +/-Infinity: ordered (-Inf < x < +Inf for finite x); participate in min/max normally.
+// - +0 and -0: IEEE 754 declares them equal under <, ==; min/max bound is correct either way.
+// - subnormal (denormal) values: ordered like normal floats; no special handling needed.
+template <typename T>
+bool scanMinMax(const facebook::velox::FlatVector<T>* flat, T& tLo, T& tHi, int64_t& nullCnt, bool& seen) {
+  const auto size = flat->size();
+  const uint64_t* nulls = flat->rawNulls();
+  const T* values = flat->rawValues();
```
zhli1142015, line 120: This helper is also instantiated by the BOOLEAN case below, but `FlatVector<bool>::rawValues()` is unsupported in Velox and throws. With partition stats enabled, materializing a cached `BooleanType` column can therefore fail instead of just producing min/max stats. Could the bool path use a bool-specific scan via `valueAt(i)`/`isNullAt(i)` (or the underlying bits) and add a C++ or E2E regression test that caches a Boolean column with stats enabled?

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
```diff
@@ -95,4 +98,405 @@ std::shared_ptr<ColumnarBatch> VeloxColumnarBatchSerializer::deserialize(uint8_t
...
+  for (vector_size_t i = 0; i < size; ++i) {
+    if (nulls != nullptr && bits::isBitNull(nulls, i)) {
+      ++nullCnt;
+      continue;
+    }
+    T v = values[i];
+    if constexpr (std::is_floating_point_v<T>) {
+      if (std::isnan(v)) {
+        return false;
```
zhli1142015: Returning on the first NaN leaves `nullCnt` as only the nulls seen before that NaN. Even when bounds are disabled, the framed stats still emit `nullCount`, and Spark cache pruning uses `IsNull` as `statsFor(a).nullCount > 0`. A partition like `[NaN, null]` would be serialized with `nullCount = 0`, so `col IS NULL` could prune away matching rows. Please keep scanning after marking min/max unsupported (or compute nullCount separately) and add a Float/Double NaN+null `IS NULL` regression test.

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
```diff
@@ -95,4 +98,405 @@ std::shared_ptr<ColumnarBatch> VeloxColumnarBatchSerializer::deserialize(uint8_t
...
+} // namespace
+
+std::vector<ColumnStats> VeloxColumnarBatchSerializer::computeStats(RowVectorPtr rowVector) {
+  std::vector<ColumnStats> result;
+  const auto numCols = rowVector->childrenSize();
+  result.resize(numCols);
+  for (column_index_t col = 0; col < numCols; ++col) {
+    auto& stats = result[col];
+    auto child = rowVector->childAt(col);
+    if (child == nullptr || !child->isFlatEncoding()) {
```
zhli1142015: Skipping non-flat children leaves `ColumnStats::nullCount` at the default 0, but unsupported columns still serialize that null count with `emitSupported = 0`. The JVM side can still use `nullCount` for `IsNull` pruning even when min/max bounds are unsupported, so a dictionary/constant/complex column containing nulls may be advertised as having no nulls and be incorrectly pruned by `IS NULL`. Since `framedSerializeWithStats` passes `getRowVector()` directly into `computeStats`, these encodings can still reach this path. Could we compute nullCount independently of flat/min-max support (or flatten/decode before stats) and cover non-flat/unsupported-type null stats?

zhli1142015 (commented):
I reviewed the partition-stats changes and left a few correctness comments.

gluten-arrow/src/main/java/org/apache/gluten/vectorized/ColumnarBatchSerializerJniWrapper.java
yaooqinn: Addressed in `9145670865` (callsite-fallback). Thank you for the advice!

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
yaooqinn: Addressed in `94d7207af1` (u32-framing-check). Thank you for the advice!

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
yaooqinn: Addressed in `491070bf34` (kryo-bounds-and-v1-back-compat). Thank you for the advice!

gluten-arrow/src/main/java/org/apache/gluten/vectorized/ColumnarBatchSerializerJniWrapper.java
```diff
@@ -48,4 +54,27 @@ public long rtHandle() {
...
+    }
+    boolean result;
+    try {
+      ColumnarBatchSerializerJniWrapper.class.getDeclaredMethod("serializeWithStats", long.class);
+      result = true;
+    } catch (NoSuchMethodException e) {
+      result = false;
+    }
+    supportsStatsExtCached = result;
+    return result;
```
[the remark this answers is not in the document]
yaooqinn: Addressed in `9145670865` (callsite-fallback). Thank you for the advice!

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -213,14 +665,40 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+            } else {
+              val unsafeBuffer = jni.serialize(handle)
+              val bytes = unsafeBuffer.toByteArray
+              CachedColumnarBatch(
+                batch.numRows(),
+                bytes.length,
+                bytes,
+                stats = null,
+                schema = null)
+            }
```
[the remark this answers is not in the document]
yaooqinn: Addressed in `9145670865` (callsite-fallback). Thank you for the advice!

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -81,43 +133,443 @@ class CachedColumnarBatchKryoSerializer extends KryoSerializer[CachedColumnarBat
...
+
+  private def readOptionalSchema(input: Input): StructType = {
+    if (!input.readBoolean()) {
+      null
+    } else {
+      val schemaLen = input.readInt()
+      val schemaBytes = new Array[Byte](schemaLen)
+      input.readBytes(schemaBytes)
+      DataType.fromJson(new String(schemaBytes, UTF_8)).asInstanceOf[StructType]
+    }
```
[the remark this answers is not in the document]
yaooqinn: Addressed in `491070bf34` (kryo-bounds-and-v1-back-compat). Thank you for the advice!

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
```diff
@@ -95,4 +98,405 @@ std::shared_ptr<ColumnarBatch> VeloxColumnarBatchSerializer::deserialize(uint8_t
...
+  framed.push_back(static_cast<uint8_t>((statsLen >> 8) & 0xFF));
+  framed.push_back(static_cast<uint8_t>((statsLen >> 16) & 0xFF));
+  framed.push_back(static_cast<uint8_t>((statsLen >> 24) & 0xFF));
+  framed.insert(framed.end(), statsBlob.begin(), statsBlob.end());
+  const uint32_t bytesLen32 = static_cast<uint32_t>(bytesLen);
+  framed.push_back(static_cast<uint8_t>(bytesLen32 & 0xFF));
+  framed.push_back(static_cast<uint8_t>((bytesLen32 >> 8) & 0xFF));
+  framed.push_back(static_cast<uint8_t>((bytesLen32 >> 16) & 0xFF));
+  framed.push_back(static_cast<uint8_t>((bytesLen32 >> 24) & 0xFF));
+  framed.insert(framed.end(), bytesBlob.begin(), bytesBlob.end());
```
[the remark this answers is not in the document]
yaooqinn: Addressed in `94d7207af1` (u32-framing-check). Thank you for the advice!

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
yaooqinn: Addressed in 1bf1524ff5 (Batch 2 — cpp stats correctness: bool/NaN/non-flat null counting). Added scanBoolMinMax helper for the BOOLEAN path, NaN scan now continues to accrue real nullCount (early-return removed), and non-flat encoding gets a dedicated isNullAt loop. New gtest case verifies the BOOLEAN scan fix; RED-validated by stashing the production change. Thank you for the careful review!

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
yaooqinn: Addressed in 1bf1524ff5 (Batch 2 — cpp stats correctness: bool/NaN/non-flat null counting). Added scanBoolMinMax helper for the BOOLEAN path, NaN scan now continues to accrue real nullCount (early-return removed), and non-flat encoding gets a dedicated isNullAt loop. New gtest case verifies the NaN nullCount fix; RED-validated by stashing the production change. Thank you for the careful review!

cpp/velox/operators/serializer/VeloxColumnarBatchSerializer.cc
yaooqinn: Addressed in 1bf1524ff5 (Batch 2 — cpp stats correctness: bool/NaN/non-flat null counting). Added scanBoolMinMax helper for the BOOLEAN path, NaN scan now continues to accrue real nullCount (early-return removed), and non-flat encoding gets a dedicated isNullAt loop. New gtest case verifies the non-flat nullCount fix; RED-validated by stashing the production change. Thank you for the careful review!

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
yaooqinn: Addressed in 3f9c923644. Added a new E2E test `partitionStats.enabled=false: legacy serialize() path correctness preserved` to `ColumnarCachedBatchE2ESuite` that wraps `cacheRange()` in a `withSQLConf` block setting the production default `false`, materializes the cache, and verifies the equality filter result. This exercises the disabled branch of the gate end-to-end so any regression that silently activates stats or breaks the stats=null read path surfaces here. Suite passes 12/12 locally (11 pre-existing + 1 new). Thanks for catching the gap.

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
yaooqinn: Thanks -- this is intentional and won't be addressed in this PR. The config is scoped to the write path by design (see design D-A4 trade-off in the PR description): stats already embedded in cached batches should be honored, otherwise toggling the config mid-session would leave behind un-prunable cached data with no way to drop the stats blob. The existing  is the user-facing way to drop cached stats. Behavior is correct (using available stats is never wrong), and no current user could rely on the read-path being gated since the feature is new. Adding a read-path kill switch is tracked as a future possibility (covered by the same on/off SQLConf if demand emerges).

backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
```diff
@@ -213,14 +665,40 @@ class ColumnarCachedBatchSerializer extends CachedBatchSerializer with Logging {
...
+              val framed = jni.serializeWithStats(handle)
+              // Carry the per-batch StructType so the Kryo (spill / disk cache) read path can
+              // dispatch by dataType.
+              val structSchema = StructType(
+                schema.map(
+                  a =>
+                    StructField(a.name, a.dataType, a.nullable)))
+              val (stats, bytesBlob) =
+                CachedColumnarBatchKryoSerializer.parseFramedBytes(framed, structSchema)
+              CachedColumnarBatch(batch.numRows(), bytesBlob.length, bytesBlob, stats, structSchema)
```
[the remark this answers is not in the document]
yaooqinn: Addressed in 3f9c923644. Hoisted the `StructType(schema.map(StructField(...)))` allocation out of the per-batch `Iterator.next()` body to a single iterator-scoped val. Schema is constant for the lifetime of a partition iterator so re-building per batch was wasted GC, especially for the many-small-batch case as you noted. Thank you!

liuneng1994 (approved):

zhli1142015 (approved):

PR #12105, opened 2026-05-18
[VL] Isolate /usr/local from native builds on macOS

jackylee-ch:

On Homebrew-based macOS, `/usr/local` accumulates leftover headers,
libraries, and CMake packages from prior Velox/Gluten builds (often
root-owned after past `sudo make install` runs from `cmake_install`
in the bootstrap). When the native build includes `/usr/local/include`
and `/usr/local/lib/cmake` in its CMake search paths it picks up stale
or mismatched versions of re2, glog, fmt, etc., shadowing Velox's
`deps-install` copies and producing confusing link or compile errors.

Concrete failure modes that show up before this fix:

- Fresh build picks up `/usr/local/lib/cmake/glog` from a prior
  `cmake --install` and links a stale glog into `libvelox.dylib`
  alongside the bundled glog from `deps-install`.
- `/usr/local/include` shadows Cellar headers (e.g. fmt) and yields
  `static_assert` failures because Velox's bundled fmt advanced past
  what `/usr/local/include/fmt` provides.
- Link errors against an old re2 left over from an aborted Velox
  bootstrap.

This PR forces CMake to ignore `/usr/local` when building native deps
on macOS, in three places that drive the third-party / Velox / Gluten
CMake configures:

| File / Function | Change |
|-----------------|--------|
| `dev/build-helper-functions.sh::cmake_install` | adds `-DCMAKE_NO_SYSTEM_FROM_IMPORTED=ON`, `-DCMAKE_IGNORE_PREFIX_PATH=/usr/local`, `-DCMAKE_IGNORE_PATH`, `-DCMAKE_SYSTEM_IGNORE_PATH` for the third-party CMake invocations driven by the Arrow/Velox bootstrap |
| `ep/build-velox/src/build-velox.sh::compile` | same flags when configuring Velox itself, plus suppresses two Apple-Clang-only warnings that fire under `-Werror` on macOS but are benign noise: `-Wno-inconsistent-missing-override` and `-Wno-macro-redefined` |
| `dev/builddeps-veloxbe.sh::build_gluten_cpp` | same isolation flags plus the same warning suppressions for the Gluten CMake configure |

All blocks are gated on `[[ "$(uname)" == "Darwin" ]]` / `if [ $OS ==
'Darwin' ]`. Linux build/link semantics are unchanged.

Without this isolation, the macOS native build is non-deterministic
depending on what's left over in `/usr/local`; with it, the build only
reads from the well-defined Homebrew Cellar prefix (`/opt/homebrew`)
and Velox's `deps-install`.

## How was this patch tested?

- macOS 14 arm64 with Apple Clang 17 and a `/usr/local` populated
  with stale glog/re2/fmt installs from a prior session: with these
  flags applied, the build no longer references `/usr/local/include`
  or `/usr/local/lib` in any compile or link command (verified via
  `_build/release/build.ninja`), and produces clean `libvelox.dylib`
  + `libgluten.dylib` that pass the Gluten CPP ctest matrix.
- Spark 3.5 + Velox backend Java JNI canary suites
  (VeloxBloomFilterTest, ColumnarBatchTest, VeloxListenerApiTest,
  OnHeapFileSystemTest, ArrowColumnVectorTest) all pass on the
  resulting build.
- Linux x86_64 Ubuntu 22.04 build green; all isolation blocks are
  Darwin-gated. Local Ubuntu build verified clean.

### DIFF
--- a/dev/build-helper-functions.sh
+++ b/dev/build-helper-functions.sh
@@ -178,10 +178,14 @@ function cmake_install {
 
   local MACOS_ISOLATION_FLAGS=""
   if [[ "$(uname)" == "Darwin" ]]; then
-    MACOS_ISOLATION_FLAGS="-DCMAKE_NO_SYSTEM_FROM_IMPORTED=ON \
-      -DCMAKE_IGNORE_PREFIX_PATH=/usr/local \
-      -DCMAKE_IGNORE_PATH=/usr/local;/usr/local/include;/usr/local/lib;/usr/local/lib/cmake \
-      -DCMAKE_SYSTEM_IGNORE_PATH=/usr/local;/usr/local/include;/usr/local/lib;/usr/local/lib/cmake"
+    if [[ "${INSTALL_PREFIX:-}" == "/usr/local" || "${INSTALL_PREFIX:-}" == /usr/local/* ]]; then
+      echo "INFO: INSTALL_PREFIX=${INSTALL_PREFIX} is under /usr/local; keeping /usr/local visible to CMake." >&2
+    else
+      MACOS_ISOLATION_FLAGS="-DCMAKE_NO_SYSTEM_FROM_IMPORTED=ON \
+        -DCMAKE_IGNORE_PREFIX_PATH=/usr/local \
+        -DCMAKE_IGNORE_PATH=/usr/local;/usr/local/include;/usr/local/lib;/usr/local/lib/cmake \
+        -DCMAKE_SYSTEM_IGNORE_PATH=/usr/local;/usr/local/include;/usr/local/lib;/usr/local/lib/cmake"
+    fi
   fi
 
   # CMAKE_POSITION_INDEPENDENT_CODE is required so that Velox can be built into dynamic libraries \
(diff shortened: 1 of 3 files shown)


jackylee-ch:
@zhouyuan @weiting-chen @philo-he  I have meet the build problem with system libraries, I'm not sure if the community need this, any idea?

philo-he:
@jackylee-ch, thanks for the PR. Two questions. Please check if they make sense. Thank you!

1) Should we move the Velox compilation option setting to the Velox code? I'd assume this issue also needs the fix when building Velox standalone, not just from the Gluten build.
2) Is it possible for users to set INSTALL_PREFIX to /usr/local when running setup-macos.sh? If that could cause issues, perhaps we should add a check to warn (or disallow) users from doing so.

jackylee-ch:
@philo-he good point, i would check for this.

jackylee-ch:
@philo-he PTAL. I've tried to analyze the build script and extract the install configuration to create environment isolation, not only for system directories like /usr/local, but also for the Arrow environment and conda, etc. However, it seems like a huge project, and I'm still trying to streamline it. My current PR only focuses on adjusting the system directories for macOS.

philo-he (approved):
Looks good. Thanks.

PR #12107, opened 2026-05-18
[VL] Add per-batch input-encoding counter to VeloxHashShuffleWriter

luis4a0:

This PR is companion to #12083 (`SCOPED_TIMER` fix). With the per-stage timers in `cpuWallTimingList_` now recording real numbers, the natural complement is a counter that tells the reader what shape the writer actually sees; i.e. for each `write(cb, ...)` call, how many of the input children arrive as FLAT / DICTIONARY / CONSTANT / LAZY / other encoding (the values of `facebook::velox::VectorEncoding::Simple`).

A new `inputEncodingCounts_` member (`std::array<int64_t, 5>`) is maintained on `VeloxHashShuffleWriter`. The counts are taken *before* any `getFlattenedRowVector()` invocation in `write(cb, ...)`, so they reflect the encoding the writer actually receives -- not the post-flatten shape (which is always FLAT by construction). Output is gated through the existing `VELOX_SHUFFLE_WRITER_LOG_FLAG` log channel and uses the same line shape as the existing `cpuWallTimingList_` lines,v so behavior is unchanged when the flag is off (default).

This information is useful for at least three things:

- Verifying whether a workload would benefit from writer-side dictionary/constant pass-through optimizations *before* writing the pass-through code (does the writer ever actually see those encodings?). Recent OSS PR #9727 (writer-side dictionary rebuild) and several proposed constant pass-through patches all share the same assumption that the writer sees DICT/CONST inputs; this counter makes that assumption directly checkable on any workload.
- Spotting unexpected upstream operator changes that collapse/preserve encodings differently across Velox versions (e.g., a change in HashAggregate that starts emitting CONSTANT children when it didn't before).
- Correlating a wall-time spike in a particular stage with the encoding mix that drove it.

The counter is always-on (one branch+one increment per input child, swamped by the surrounding shuffle work). Scope is deliberately narrow: only the 5-bucket encoding mix at the hash-shuffle writer entry, surfaced via the existing log channel. A per-type breakdown (BIGINT/DECIMAL/VARCHAR/...) and exposure through `ShuffleWriterMetrics` to the JVM SparkListener are both plausible follow-ups but kept out of this PR.

Diff: +277 / -0 over 4 files (3 in `cpp/velox/`, 1 new test).

## How was this patch tested?

This PR adds `cpp/velox/tests/VeloxHashShuffleWriterInputEncodingTest.cc` (second commit) with three gtest cases:

- `allFlat`: two batches with 3 flat children each; asserts `FLAT` bucket accumulates to 6 and the other 4 buckets stay 0.
- `mixedFlatDictConst`: one batch with FLAT + FLAT + DICTIONARY + CONSTANT children; asserts FLAT=2, DICTIONARY=1, CONSTANT=1, and the remaining 2 buckets stay 0.
- `lazy`: wraps a child in a `LazyVector`; asserts the `LAZY` bucket increments (the encoding is captured BEFORE any `loadedVector()` call inside the writer).

The new test is a standalone `add_velox_test(...)` entry rather than a new `TEST_P` case in `VeloxShuffleWriterTest.cc`, so it isolates the counter behavior from the parameterised round-trip matrix, which keeps the failure mode obvious if a future change quietly breaks one of the buckets.

### DIFF
--- a/cpp/velox/shuffle/VeloxHashShuffleWriter.cc
+++ b/cpp/velox/shuffle/VeloxHashShuffleWriter.cc
@@ -1320,16 +1320,22 @@ void VeloxHashShuffleWriter::accumulateInputEncodingCounts(const ColumnarBatch& 
   // Only velox-typed batches expose per-child encoding; foreign batches
   // (e.g. arrow round-trips coming from non-velox sources) will be flattened
   // by `VeloxColumnarBatch::from` later and we'd undercount, so just skip
-  // them here rather than reporting a misleading "all flat" mix.
+  // them here rather than reporting a misleading "all flat" mix. The skip
+  // counter is exposed via `inputEncodingSkippedBatches()` and printed in
+  // `stat()` so a reader can tell whether a low encoding-bucket total means
+  // "writer saw few children" or "writer saw many but most were not velox".
   if (cb.getType() != "velox") {
+    ++inputEncodingSkippedBatches_;
     return;
   }
   const auto* veloxBatch = dynamic_cast<const VeloxColumnarBatch*>(&cb);
   if (veloxBatch == nullptr) {
+    ++inputEncodingSkippedBatches_;
     return;
   }
   const auto& rowVector = veloxBatch->getRowVector();
   if (rowVector == nullptr) {
+    ++inputEncodingSkippedBatches_;
     return;
   }
   for (const auto& child : rowVector->children()) {
@@ -1350,7 +1356,20 @@ void VeloxHashShuffleWriter::accumulateInputEncodingCounts(const ColumnarBatch& 
       case facebook::velox::VectorEncoding::Simple::LAZY:
         ++inputEncodingCounts_[kInputEncodingLazy];
         break;
+      case facebook::velox::VectorEncoding::Simple::ROW:
+      case facebook::velox::VectorEncoding::Simple::MAP:
+      case facebook::velox::VectorEncoding::Simple::FLAT_MAP:
+      case facebook::velox::VectorEncoding::Simple::ARRAY:
+        // Struct / map / array column types — first-class in Spark workloads
+        // and exercised by the sibling shuffle writer tests
+        // (`makeArrayVector` / `makeMapVector`). Kept distinct from `Other`
+        // so a reader interpreting the log doesn't conflate a struct column
+        // with a rare encoding.
+        ++inputEncodingCounts_[kInputEncodingComplex];
+        break;
       default:
+        // BIASED, SEQUENCE, FUNCTION, and any future additions to
+        // `VectorEncoding::Simple`.
         ++inputEncodingCounts_[kInputEncodingOther];
         break;
     }
@@ -1384,6 +1403,10 @@ void VeloxHashShuffleWriter::stat() const {
         oss << "(" << (100.0 * static_cast<double>(v) / static_cast<double>(total)) << "%)";
       }
     }
+    // Non-velox `write()` calls contribute 0 to the buckets above; expose the
+    // skip count so the denominator (total = sum of buckets) is comparable to
+    // the `count` field in the `cpuWallTimingList_` lines emitted above.
+    oss << " SkippedNonVeloxBatches=" << inputEncodingSkippedBatches_;
     LOG(INFO) << oss.str();
   }
 #endif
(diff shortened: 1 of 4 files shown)


cpp/velox/shuffle/VeloxHashShuffleWriter.cc
```diff
@@ -1315,6 +1316,47 @@ uint64_t VeloxHashShuffleWriter::valueBufferSizeForFixedWidthArray(uint32_t fixe
...
+        ++inputEncodingCounts_[kInputEncodingDictionary];
+        break;
+      case facebook::velox::VectorEncoding::Simple::CONSTANT:
+        ++inputEncodingCounts_[kInputEncodingConstant];
+        break;
+      case facebook::velox::VectorEncoding::Simple::LAZY:
+        ++inputEncodingCounts_[kInputEncodingLazy];
+        break;
+      default:
+        ++inputEncodingCounts_[kInputEncodingOther];
```
yaooqinn, line 1373: `facebook::velox::VectorEncoding::Simple` (velox/vector/VectorEncoding.h:29-41) has 11 values; the switch handles 4 and the default funnels the remaining 7 (BIASED, SEQUENCE, ROW, MAP, FLAT_MAP, ARRAY, FUNCTION) into "Other". ROW / MAP / ARRAY are first-class struct / map / array column types — the sibling `VeloxShuffleWriterTest.cc:488,492` exercises exactly those via `makeArrayVector` / `makeMapVector`, so the shuffle writer is expected to see them in normal Spark workloads. A reader interpreting "Other=80%" in the log will assume rare/unknown encodings, not "I have a struct column". Would you consider renaming `Other` to `OtherOrComplex` and/or splitting out a `kInputEncodingComplex` bucket for ROW / MAP / FLAT_MAP / ARRAY?

cpp/velox/shuffle/VeloxHashShuffleWriter.cc
```diff
@@ -1328,6 +1370,22 @@ void VeloxHashShuffleWriter::stat() const {
...
+    oss << "Velox shuffle writer stat:InputEncoding";
+    int64_t total = 0;
+    for (auto v : inputEncodingCounts_) {
+      total += v;
+    }
+    for (int b = 0; b < kInputEncodingNum; ++b) {
+      auto v = inputEncodingCounts_[b];
+      oss << " " << inputEncodingName(static_cast<InputEncodingBucket>(b)) << "=" << v;
+      if (total > 0) {
+        oss << "(" << (100.0 * static_cast<double>(v) / static_cast<double>(total)) << "%)";
```
yaooqinn, line 1403: Since `accumulateInputEncodingCounts` early-returns on non-velox batches (line 1322), `total` here covers only velox-typed children, while the sibling `cpuWallTimingList_` block above prints raw counts over all `write()` calls. Worth printing the skip count next to the percentages so the two log lines have a comparable denominator?

cpp/velox/shuffle/VeloxHashShuffleWriter.cc
luis4a0: Good catch, fixed in a208a78. Added a new `kInputEncodingComplex` bucket so ROW / MAP / FLAT_MAP / ARRAY no longer get conflated with the rare-encoding catch-all. `kInputEncodingOther` now only covers BIASED / SEQUENCE / FUNCTION (and any future additions to `VectorEncoding::Simple`). New `complex` gtest case exercises ARRAY + MAP children landing in the new bucket; existing cases also assert `kInputEncodingComplex == 0` so the boundary is checked.

cpp/velox/shuffle/VeloxHashShuffleWriter.cc
luis4a0: Right, fixed in a208a78. Added an `inputEncodingSkippedBatches_` counter that's incremented on every early-return path in `accumulateInputEncodingCounts` and printed at the end of the InputEncoding log line as `SkippedNonVeloxBatches=N`. The encoding-bucket total + the skip count now equal the `count` field on the surrounding `cpuWallTimingList_` lines, so the two log blocks have comparable denominators.

yaooqinn (approved):
LGTM, thank you @luis4a0!

marin-ma (approved):

Issue #11912, opened 2026-04-11
[VL] Spark 4.x: JNI and Velox exception handling loses Spark error condition and exception type


baibaichen:
## Backend
VL (Velox)

## Bug description

Gluten's exception handling has two layers of information loss that cause ~49 test excludes and 2 disabled suites:

**Sub-problem 1: JNI → Java exception rewriting**
Java-side exceptions are replaced with `GlutenException` when crossing the JNI boundary, losing the original exception type and error condition code. Related: #11825.

**Sub-problem 2: Velox C++ → Spark exception mapping**
Velox errors go through the `JNI_METHOD_END` macro and are uniformly thrown as `GlutenException`. Spark expects specific subtypes like `SparkRuntimeException` with error condition codes (e.g., `NOT_NULL_ASSERT_VIOLATION`).

Parent issue: #11550

### Impact

~49 excludes + 2 TODO suites:

| Category | Excludes | Affected suites |
|----------|:--------:|-----------------|
| assert_not_null (NOT_NULL_ASSERT_VIOLATION) | 14 | RuntimeNullChecksV2Writes(7), InsertSuite(1), DataSourceV2SQL(1), Delta/GroupBased(5) |
| merge cardinality (prefix excludes) | 24 (18 real + 6 collateral) | DeltaBasedMerge(x2), GroupBasedMerge |
| raise_error / assert_true | 3 | ColumnExpressionSuite(2), CachedTableSuite(1) |
| SparkRuntimeException → SparkException | 2 | DataFrameFunctionsSuite |
| Exception class/message mismatch | 4 | JsonExpressionsSuite(1), SQLQuerySuite(3) |
| TODO suites | 2 | GlutenGroupBasedUpdateTableSuite(1), GlutenSimpleSQLViewSuite(1) |

### Cross-reference

GlutenSimpleSQLViewSuite has 2 failures: 1 belongs to this issue (error condition), 1 belongs to the decimal precision issue (SPARK-53968, spark41-only).

PR #12110, opened 2026-05-18
[GLUTEN-11917][VL] Respect allowPrecisionLoss from expression context in Spark 4.1

### LINKED ISSUE #11917: [VL] Spark 4.1: Velox decimal arithmetic does not respect allowPrecisionLoss context (SPARK-53968)
## Backend
VL (Velox)

## Bug description

`DECIMAL(38,18) + DECIMAL(38,18)` produces a result with incorrect scale (off by 1 digit, i.e., 10x error). Spark 4.1 introduced `NumericEvalContext` (SPARK-53968) which freezes `allowPrecisionLoss=true` in expression state. Velox does not correctly receive this context, resulting in decimal scale mismatch.

This is **not** ANSI-related.

Spark 4.1 only.

Parent issue: #11550

### Impact

| Suite | Failures | spark40 | spark41 |
|-------|:--------:|:-------:|:-------:|
| GlutenSimpleSQLViewSuite | 1 | 🟢 | 🔴 |

Note: GlutenSimpleSQLViewSuite has 2 failures total — the other failure (error condition) is tracked under #11912.

### References

- Apache Spark JIRA: [SPARK-53968](https://issues.apache.org/jira/browse/SPARK-53968)

nookcreed:

  In Spark 4.1, [SPARK-53968](https://issues.apache.org/jira/browse/SPARK-53968) introduced `NumericEvalContext` which captures `allowPrecisionLoss` **at analysis time** and embeds it in arithmetic expressions (`Add`,
  `Subtract`, `Multiply`, `Divide`) via an `evalContext` field.

  Previously, Gluten read `SQLConf.get.decimalOperationsAllowPrecisionLoss` at plan-conversion time. These can diverge — for example, when querying a persistent view whose expression was analyzed under a different session
  config. This causes a decimal scale mismatch (e.g. `DECIMAL(38,18) + DECIMAL(38,18)` producing scale 18 instead of 17 — a 10× error).

  **Fix:** Add a shim method `decimalAllowPrecisionLoss(expr: BinaryArithmetic): Boolean`:
  - **`Spark41Shims`**: reads `expr.evalContext.allowDecimalPrecisionLoss` directly from the expression
  - **All other Spark versions**: fall back to `SQLConf.get` (preserving existing behavior)

  `DecimalArithmeticUtil.getResultType` and `VeloxSparkPlanExecApi.getDecimalArithmeticExprName` now use the shim so the correct result type and Velox native function name are selected.

  ## How was this patch tested?

Root cause scenario

  The bug triggers when allowPrecisionLoss captured in an expression's NumericEvalContext at analysis time differs from the current SQLConf at plan-conversion time. The canonical case is querying a persistent view whose
  decimal arithmetic was analyzed under one config:

  -- View analyzed with allowPrecisionLoss=true (default)
  -- DECIMAL(38,18) + DECIMAL(38,18) → Spark computes result type as DECIMAL(38,17)
  CREATE VIEW v AS SELECT col1 + col2 AS result FROM t;

  -- Query run in a session with allowPrecisionLoss=false
  SET spark.sql.decimalOperations.allowPrecisionLoss=false;
  -- Old code: reads SQLConf → uses false → computes DECIMAL(38,18), calls add_deny_precision_loss
  -- New code: reads expr.evalContext → uses true  → computes DECIMAL(38,17), calls add
  SELECT * FROM v;
  -- Scale mismatch: 18 vs 17 = 10× error

  Existing test (no regression)

  MathFunctionsValidateSuite#decimal arithmetic (backends-velox) already covers allowPrecisionLoss=true/false toggled within a single session via withSQLConf. This test continues to pass because expression creation and plan
  conversion both happen inside the same config scope.

  ./dev/run-scala-test.sh \
    -Pjava-17,spark-4.1,scala-2.13,backends-velox,hadoop-3.3,spark-ut \
    -pl backends-velox \
    -s org.apache.gluten.functions.MathFunctionsValidateSuite \
    -t "decimal arithmetic"

  Fixed test: GlutenSQLQuerySuite

  The exclusion on line 1047 of VeloxTestSettings.scala (spark41) should be removed and the test should now pass:

  ./dev/run-scala-test.sh \
    -Pjava-17,spark-4.1,scala-2.13,backends-velox,hadoop-3.3,spark-ut \
    -pl gluten-ut/spark41 \
    -s org.apache.spark.sql.GlutenSQLQuerySuite \
    -t "should be able to resolve a persistent view"

  GlutenSimpleSQLViewSuite (partial — 1 of 2 failures)

  GlutenSimpleSQLViewSuite is currently fully disabled (// TODO: 4.x enableSuite on line 701). This PR resolves 1 of its 2 failures; the remaining failure is tracked in #11912. The suite can be re-enabled with an exclusion for
   the #11912 test once that test name is identified from a CI run.

  CI

  Trigger Velox CI on the PR to run the full spark-ut suite against Spark 4.1.

  ---

  Closes #11917

### DIFF
--- a/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
+++ b/backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
@@ -162,8 +162,8 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi with Logging {
     }
   }
 
-  override def getDecimalArithmeticExprName(exprName: String): String =
-    if (!SQLConf.get.decimalOperationsAllowPrecisionLoss) { exprName + "_deny_precision_loss" }
+  override def getDecimalArithmeticExprName(exprName: String, allowPrecisionLoss: Boolean): String =
+    if (!allowPrecisionLoss) { exprName + "_deny_precision_loss" }
     else { exprName }
 
   /** Transform map_entries to Substrait. */
--- a/backends-velox/src/test/scala/org/apache/gluten/functions/MathFunctionsValidateSuite.scala
+++ b/backends-velox/src/test/scala/org/apache/gluten/functions/MathFunctionsValidateSuite.scala
@@ -413,4 +413,30 @@ abstract class MathFunctionsValidateSuite extends FunctionsValidateSuite {
       }
     }
   }
+
+  test("decimal arithmetic respects allowPrecisionLoss captured at view analysis time") {
+    // Regression test for GLUTEN-11917: in Spark 4.1, arithmetic expressions embed
+    // allowPrecisionLoss in their evalContext at analysis time. Gluten must read from
+    // the expression rather than SQLConf.get, which can differ when querying a view
+    // analyzed under a different session config.
+    withTempView("t", "v") {
+      sql("""
+            |SELECT
+            |CAST('1234567890123456789012345.12345678901' AS DECIMAL(38,11)) AS a,
+            |CAST('1234567890123456789012345.02345678901' AS DECIMAL(38,11)) AS b""".stripMargin)
+        .createOrReplaceTempView("t")
+
+      // Analyze arithmetic with allowPrecisionLoss=false and cache it in the view's plan.
+      withSQLConf("spark.sql.decimalOperations.allowPrecisionLoss" -> "false") {
+        sql("CREATE OR REPLACE TEMP VIEW v AS SELECT a - b, a + b, a * b, a / b FROM t")
+      }
+
+      // Query under the opposite setting -- Gluten must use the captured context, not SQLConf.
+      withSQLConf("spark.sql.decimalOperations.allowPrecisionLoss" -> "true") {
+        runQueryAndCompare("SELECT * FROM v") {
+          checkGlutenPlan[ProjectExecTransformer]
+        }
+      }
+    }
+  }
 }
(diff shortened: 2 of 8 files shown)


zhouyuan:
could we enable the disabled suite in spark41 with this change?
https://github.com/apache/gluten/blob/b8c5711d672c40ef3cc0717646074fa1bdeac46e/gluten-ut/spark41/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala#L705

nookcreed:
> could we enable the disabled suite in spark41 with this change?
> 
> https://github.com/apache/gluten/blob/b8c5711d672c40ef3cc0717646074fa1bdeac46e/gluten-ut/spark41/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala#L705

@zhouyuan The suite comment originally said "2 failures" for Spark 4.1 vs "1 failure" for Spark 4.0. The key insight is that those two numbers indicate two distinct root causes:

  - Failure A — present in both Spark 4.0 and 4.1, so it predates SPARK-53968 and is unrelated to allowPrecisionLoss/evalContext. It's whatever causes the suite to stay disabled in 4.0 today.
  - Failure B (4.1-only) — introduced by SPARK-53968, which embedded allowPrecisionLoss in evalContext at analysis time. This is exactly what GLUTEN-11917 fixes.

  This PR resolves Failure B. Failure A remains, which is why enabling the suite unconditionally caused CI to fail. The updated comment (// 1 failure remains after GLUTEN-11917 fix) documents this so the
  next contributor knows where things stand — they just need to identify and exclude Failure A, then the suite can be enabled.

nookcreed:
@zhouyuan please review

zhouyuan (approved):

nookcreed:
@zhouyuan could you please merge? thanks

zhouyuan:
Cc @baibaichen

PR #12112, opened 2026-05-19
[VL] Fix silent stats prune for non-binary collation StringType in cache

yaooqinn:

Gate non-binary-collation `StringType` columns in the Velox cache path to `supported=0` (writer side) AND strip any AND-conjunct that references such a column from the reader-side `buildFilter` predicate vector (before delegating to `super.buildFilter`). Writer / wire format unchanged.

New shim API `SparkShims.isBinaryCollationString` — default `true` for Spark 3.x shims (no collation concept), overridden on Spark 4.0 / 4.1 to check `collationId == UTF8_BINARY_COLLATION_ID`.

On Spark 4.x with a non-binary collation, Velox's `scanMinMax<StringView>` does an unsigned byte-order compare while Spark's filter compare is collation-aware (`PhysicalStringType.ordering = CollationFactory.fetchCollation(id).comparator`). The two disagree, so stats-based pruning can silently drop matching rows.

Repro:
```scala
spark.sql("CREATE TABLE t(s STRING COLLATE UTF8_LCASE) USING parquet")
spark.sql("INSERT INTO t VALUES 'abc', 'XYZ'")
spark.sql("CACHE TABLE t")
spark.sql("SELECT * FROM t WHERE s = 'ABC'").show()
// Before: 0 rows (wrong). After: 1 row.
```

Vanilla Spark's `StringColumnStats` is collation-aware, so this is Gluten-specific.

### Reader-side approach (rev 3.2)

Earlier revisions filled a `0xFF * 256B` sentinel upper bound on the deserialize side to keep vanilla `buildFilter` from pruning. As pointed out by @zhli1142015, that sentinel is not a universal upper bound under non-binary collation orderings, so it is not safe.

Rev 3.2 drops the sentinel and instead wraps `SimpleMetricsCachedBatchSerializer.buildFilter` with a `splitConjunctivePredicates`-based predicate-strip layer (`stripUnsupportedConjuncts`):

- For each input predicate, split into AND-conjuncts.
- Drop every conjunct whose `references` contain any attribute that was demoted to `supported=0` (i.e. a non-binary collation StringType in `cachedAttributes`).
- Rebuild surviving conjuncts with `And.reduce`; bypass entirely if nothing references a demoted column.
- `Or` sub-trees stay intact (one losing-stats disjunct already loses the whole Or anyway, so it's conservative).

Empty filtered predicates degrade gracefully: vanilla `SimpleMetricsCachedBatchSerializer` reduces `partitionFilters` with `.reduceOption(And).getOrElse(Literal(true))`, so the partition filter becomes pass-through (verified against `spark-sql_2.13-4.0.1-sources` `CachedBatchSerializer.scala`).

A real collation-aware bound (matching vanilla `StringColumnStats.semanticCompare`) would require teaching the cpp `scanMinMax` path about collations, likely via ICU sort keys — tracked as a Phase-2 follow-up.

### Does this PR introduce _any_ user-facing change?

Yes — correctness fix. No new config.

- New `ColumnarCachedBatchBuildFilterPruneSuite` W1–W8 (wrapper behavior + anti-regression bypass).
- New `ColumnarCachedBatchE2ESuite` cases for UTF8_LCASE + UNICODE_CI predicate over cached batch.
- Existing suites: `ColumnarCacheShipBlockerMarshalSuite`, `ColumnarCachedBatchStatsBlobSuite`, `ColumnarCachedBatchIntFamilyMarshalSuite`.
- `mvn clean install` + suites verified on:
  - `-Pspark-4.0 -Pscala-2.13` — 42/42 PASS
  - `-Pspark-4.1 -Pscala-2.13` — 42/42 PASS
  - `-Pspark-3.5 -Pscala-2.12` — 32 PASS + 10 cleanly canceled (W1–W8 + 2 E2E collation cases guarded by `assume(isCollationAware)`, since `CollationFactory` does not exist on 3.5)

Generated-by: Claude Opus 4.7

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
@@ -23,6 +23,7 @@ import org.apache.gluten.execution.{RowToVeloxColumnarExec, VeloxColumnarToRowEx
 import org.apache.gluten.iterator.Iterators
 import org.apache.gluten.memory.arrow.alloc.ArrowBufferAllocators
 import org.apache.gluten.runtime.Runtimes
+import org.apache.gluten.sql.shims.SparkShimLoader
 import org.apache.gluten.utils.ArrowAbiUtil
 import org.apache.gluten.vectorized.ColumnarBatchSerializerJniWrapper
 
@@ -31,8 +32,10 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.internal.config.Kryo.KRYO_SERIALIZER_MAX_BUFFER_SIZE
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression, GenericInternalRow}
-import org.apache.spark.sql.columnar.{CachedBatch, SimpleMetricsCachedBatch, SimpleMetricsCachedBatchSerializer}
+import org.apache.spark.sql.catalyst.expressions.{And, Attribute, Expression, ExprId}
+import org.apache.spark.sql.catalyst.expressions.{GenericInternalRow, PredicateHelper}
+import org.apache.spark.sql.columnar.{CachedBatch, SimpleMetricsCachedBatch}
+import org.apache.spark.sql.columnar.SimpleMetricsCachedBatchSerializer
 import org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
@@ -243,7 +246,12 @@ object CachedColumnarBatchKryoSerializer {
       case FloatType => true // 4B IEEE 754; NaN guard in cpp scanMinMax
       case DoubleType => true // 8B IEEE 754; NaN guard in cpp scanMinMax
       case BooleanType => true
-      case _: StringType => true // truncated to 256B; see encodeStringBounds (any collation)
+      case s: StringType if SparkShimLoader.getSparkShims.isBinaryCollationString(s) => true
+      // Non-binary collation: cpp scanMinMax byte-order disagrees with Spark's
+      // collation-aware String ordering at runtime (PhysicalStringType.ordering
+      // dispatches to CollationFactory.fetchCollation(id).comparator). Demote to
+      // supported=0; the buildFilter wrapper strips any AND-conjunct that
+      // references such columns to guarantee pass-through.
       case _ => false
     }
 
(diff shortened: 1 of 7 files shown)


zhli1142015:
Comment 1 — Test coverage gap

  The new sentinel buildFilter tests currently use AttributeReference("c", StringType)(), which is the default binary-collation StringType. That does not exercise  the actual non-binary collation path this PR is trying to fix. SimpleMetricsCachedBatchSerializer.buildFilter builds lower/upper-bound attributes from the cached  attribute data type, so for the real failure case comparisons are collation-aware, not binary byte-order comparisons.

  Could you add test coverage with an explicit non-binary collation StringType (for example UTF8_BINARY_LCASE/UNICODE, depending on the supported Spark profile) so  EqualTo, In, StartsWith, and range predicates are evaluated through the same collation-aware comparison path as production?

  Comment 2 — Sentinel upper bound correctness under non-binary collation

  The sentinel upper bound is currently 256 bytes of 0xff. That is a safe “max” value for binary byte ordering, but for non-binary collations Spark comparisons are  collation-aware. Those bytes are invalid UTF-8 and will not necessarily compare as a universal maximum value under every Spark collation. If a literal can compare  greater than this sentinel under a non-binary collation, the parent buildFilter may still prune the batch incorrectly.

  A safer design would be to make the Velox serializer’s buildFilter wrapper ignore/minimize min/max pruning for non-binary-collation string attributes, rather than  fabricating synthetic min/max bounds. That would preserve correctness by passing through batches whenever the predicate depends on unsupported string ordering  stats.

yaooqinn:
Thanks @zhli1142015 for the thorough review — both points are valid and addressed in rev 3.2 (`48c8bc749d`).

**C1 (test coverage)**: The earlier sentinel-suite only exercised the deserialize hot path. Added 8 wrapper-behavior tests (`ColumnarCachedBatchBuildFilterPruneSuite` W1–W8) plus end-to-end UTF8_LCASE / UNICODE_CI cases in `ColumnarCachedBatchE2ESuite`. W8 is an explicit anti-regression that bypasses the wrapper to prove a UTF8_LCASE bound check would have wrongly pruned the batch.

**C2 (sentinel correctness)**: You are right — `0xFF * 256B` is not a universal upper bound under non-binary collations. `PhysicalStringType.ordering = CollationFactory.fetchCollation(id).comparator` (`PhysicalDataType.scala:334`) means `<=` on collation-aware StringType is governed by the collation’s comparator, not by raw byte order, so any fixed sentinel can falsify the bound check.

Rev 3.2 abandons the sentinel approach entirely. Reader-side mechanism is now a `splitConjunctivePredicates`-based wrapper around `SimpleMetricsCachedBatchSerializer.buildFilter`: any AND-conjunct referencing a non-binary collation StringType attribute is dropped before partition-stats evaluation; binary attributes still prune. Conjuncts that cannot be split (Or, etc.) referencing such attributes conservatively keep the batch. Writer/wire-format unchanged.

Verified locally:
- `mvn clean install` + suites on `-Pspark-3.5/scala-2.12`, `-Pspark-4.0/scala-2.13`, `-Pspark-4.1/scala-2.13`
- spark-4.0 / 4.1: 42/42 PASS
- spark-3.5: 32 PASS + 10 cleanly canceled (W1–W8 + 2 E2E collation cases guarded by `assume(isCollationAware)`, since CollationFactory does not exist on 3.5)

PR description updated to reflect rev 3.2. PTAL.

zhli1142015 (approved):
LGTM, thanks

PR #12114, opened 2026-05-19
[CORE] Add customMetrics extension point to ShuffleWriterMetrics for backend-specific shuffle stats

luis4a0:

`ShuffleWriterMetrics` (in `cpp/core/shuffle/Options.h`) currently has a hand-rolled list of 9 scalar fields, two of which (`avgDictionaryFields`, `dictionarySize`) are Velox-specific. Every time someone wants to expose another backend-specific shuffle stat, the pattern is: new scalar field on the struct → new constructor parameter on `GlutenSplitResult` → new getter → new line in `ColumnarShuffleWriter.scala`. That's four files per metric, and it forces cross-backend coordination (the ClickHouse / GPU / RSS backends would each need similar plumbing for their own stats).

This PR adds a generic `std::unordered_map<std::string, int64_t> customMetrics` field to `ShuffleWriterMetrics` that any shuffle writer can populate with backend-specific stats, plumbed through the existing JNI `stop()` path as two parallel arrays (keys + values) into `GlutenSplitResult`. The JVM side reassembles them lazily into an unmodifiable `Map<String, Long>` on first `getCustomMetrics()` access, with a defensive length-match check that fails loudly on producer-side bugs.

Convention for keys: `<Backend>.<Family>.<Stat>`. Spark-side surfacing (registration as SQLMetrics) happens per-key in the backend's MetricsApi.

### Files touched

- `cpp/core/shuffle/Options.h` — new `customMetrics` field with key-naming convention doc
- `cpp/core/shuffle/ShuffleWriter.{h,cc}` — `customMetrics()` const accessor
- `cpp/core/jni/JniWrapper.cc` — marshal map as two parallel arrays at the existing `stop()` site
- `gluten-arrow/src/main/java/org/apache/gluten/vectorized/GlutenSplitResult.java` — new constructor params, lazy `getCustomMetrics()` accessor with caching + defensive length-match check
- `gluten-arrow/src/test/scala/org/apache/gluten/vectorized/GlutenSplitResultSuite.scala` — JVM-side unit tests for the lazy reassembly logic

## How was this patch tested?

- `ninja gluten velox`: clean
- `mvn -Pbackends-velox,spark-3.5 -pl gluten-arrow compile`: BUILD SUCCESS
- `mvn -Pbackends-velox,spark-3.5 -pl gluten-arrow scalatest:test -DwildcardSuites='org.apache.gluten.vectorized.GlutenSplitResultSuite'`: 8/8 passed
- `dev/git-clang-format --binary clang-format-15 --style=file`: clean
- `dev/format-scala-code.sh`: clean
- `dev/check.py format main`: Ok on all C++ files touched

Generated-by: GitHub Copilot CLI (Claude Opus 4.7 1M context)

### DIFF
--- a/gluten-arrow/src/main/java/org/apache/gluten/vectorized/GlutenSplitResult.java
+++ b/gluten-arrow/src/main/java/org/apache/gluten/vectorized/GlutenSplitResult.java
@@ -143,8 +143,8 @@ public class GlutenSplitResult {
   /**
    * Backend-specific shuffle writer metrics, keyed by `<Backend>.<Family>.<Stat>`. The map
    * preserves the iteration order JNI marshalled, but callers should treat the map as unordered.
-   * Returns an empty map if the native side did not populate any custom metrics (e.g. older Gluten
-   * libs, or backends that don't yet emit any). The returned map is unmodifiable.
+   * Returns an empty map if the native side did not populate any custom metrics. The returned map
+   * is unmodifiable.
    */
   public Map<String, Long> getCustomMetrics() {
     Map<String, Long> cached = customMetricsCache;
@@ -158,6 +158,19 @@ public class GlutenSplitResult {
       if (customMetricsKeys == null || customMetricsKeys.length == 0) {
         customMetricsCache = Collections.emptyMap();
       } else {
+        // Defensive check: if a future native-side producer ever ships
+        // mismatched arrays, fail loudly here (before the cache field is
+        // assigned to a partial result). Without this, the AIOOBE inside the
+        // loop would leave `customMetricsCache` null and every subsequent
+        // `getCustomMetrics()` call would re-enter the synchronized block
+        // and re-throw — a confusing failure mode.
+        if (customMetricsValues == null || customMetricsValues.length != customMetricsKeys.length) {
+          throw new IllegalStateException(
+              "customMetricsKeys / customMetricsValues length mismatch: "
+                  + customMetricsKeys.length
+                  + " vs "
+                  + (customMetricsValues == null ? "null" : customMetricsValues.length));
+        }
         LinkedHashMap<String, Long> map = new LinkedHashMap<>(customMetricsKeys.length);
         for (int i = 0; i < customMetricsKeys.length; i++) {
           map.put(customMetricsKeys[i], customMetricsValues[i]);
--- a/gluten-arrow/src/main/java/org/apache/gluten/vectorized/GlutenSplitResult.java
+++ b/gluten-arrow/src/main/java/org/apache/gluten/vectorized/GlutenSplitResult.java
@@ -16,6 +16,10 @@
  */
 package org.apache.gluten.vectorized;
 
+import java.util.Collections;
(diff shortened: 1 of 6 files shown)


backends-velox/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -258,6 +258,18 @@ class ColumnarShuffleWriter[K, V](
...
+    // Backend-specific custom metrics (see GlutenSplitResult.getCustomMetrics).
+    // Only entries whose key is pre-registered in `dep.metrics` are recorded;
+    // unknown keys are silently dropped so that the native side can ship new
+    // metrics ahead of the Spark-side `VeloxMetricsApi` registration without
+    // breaking older builds.
+    splitResult.getCustomMetrics.forEach {
+      (key, value) =>
+        val m = dep.metrics.get(key)
+        if (m.isDefined) {
+          m.get.add(value)
```
yaooqinn: `splitResult.getCustomMetrics()` is consumed here but not in `VeloxCelebornColumnarShuffleWriter.scala:101+` or `VeloxUniffleColumnarShuffleWriter.java:232`. RSS users won't see new `Velox.InputEncoding.*` metrics. Suggest a shared `applyCustomMetrics(splitResult, dep)` helper called from all three writers.

gluten-arrow/src/main/java/org/apache/gluten/vectorized/GlutenSplitResult.java
```diff
@@ -122,4 +139,32 @@ public double getAvgDictionaryFields() {
...
+  public Map<String, Long> getCustomMetrics() {
+    Map<String, Long> cached = customMetricsCache;
+    if (cached != null) {
+      return cached;
+    }
+    synchronized (this) {
+      if (customMetricsCache != null) {
+        return customMetricsCache;
+      }
+      if (customMetricsKeys == null || customMetricsKeys.length == 0) {
```
yaooqinn, line 158: `getCustomMetrics()` trusts `customMetricsValues.length == customMetricsKeys.length`. If a future producer ships mismatched arrays, the AIOOBE inside the synchronized block leaves the volatile cache field unassigned — subsequent calls re-enter and AIOOBE again. Suggest a `Preconditions.checkArgument(...)` at the top + one suite case.

luis4a0: Excellent point. Fixed in 722fa45. Added an explicit length-match check at the top of the synchronized block that throws `IllegalStateException` mentioning both lengths (and the case where values is null while keys is non-empty), so the producer-side bug is unambiguous and the cache field never gets assigned to a partial map. Two new `GlutenSplitResultSuite` cases:

- `fails loudly on mismatched key/value array lengths` — also asserts the second call still throws (the cache field stays null on the failure path)
- `fails loudly when values array is null but keys is non-empty`

Both pass locally; full suite is now 8/8.

PR #12129, opened 2026-05-22
[Velox][Build][MIRROR] Skip sudo on macOS when staging Velox Arrow patches

jackylee-ch:

`apply_compilation_fixes` in `ep/build-velox/src/get-velox.sh` always
calls `sudo cp` when staging the two Arrow patch files into
`$VELOX_HOME/CMake/resolve_dependency_modules/arrow/`. On macOS the
target directory is user-owned, so the `sudo` triggers an interactive
password prompt on every build with no functional benefit.

Introduce a `SUDO_CMD` variable that resolves to `sudo` on Linux and
an empty string elsewhere; behavior on Linux is unchanged.

## How was this patch tested?

- macOS 14 arm64: `dev/builddeps-veloxbe.sh` no longer prompts for a
  password when re-running `apply_compilation_fixes`. The two patch
  files land in `$VELOX_HOME/CMake/resolve_dependency_modules/arrow/`
  with the user's own ownership.
- Linux x86_64: `SUDO_CMD=sudo`, identical to the previous behavior.

### DIFF
--- a/ep/build-velox/src/get-velox.sh
+++ b/ep/build-velox/src/get-velox.sh
@@ -141,8 +141,12 @@ function apply_provided_velox_patch {
 }
 
 function apply_compilation_fixes {
-  sudo cp ${CURRENT_DIR}/modify_arrow.patch ${VELOX_HOME}/CMake/resolve_dependency_modules/arrow/
-  sudo cp ${CURRENT_DIR}/modify_arrow_dataset_scan_option.patch ${VELOX_HOME}/CMake/resolve_dependency_modules/arrow/
+  local SUDO_CMD=""
+  if [ "$OS" == "Linux" ]; then
+    SUDO_CMD="sudo"
+  fi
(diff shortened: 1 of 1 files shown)


jackylee-ch:
cc @zhouyuan

philo-he (approved):
Thanks.

zhouyuan (commented):
for me i'm setting sudo password-less on mac

jackylee-ch:
@zhouyuan Yeap. I previously also used `sudo` to execute the commands and then installed them to a local macOS path. Recently, I tried disabling sudo permissions and found that compilation still worked normally without affecting the system environment. For many developers, minimizing the impact on the system environment is essential, especially for macOS.

However, the current changes do not affect compilation, they simply change `cp` to non-sudo execution, so the impact is minimal.

zhouyuan (approved):

PR #12131, opened 2026-05-24
[VL][Delta] Add JVM Delta DV scan handoff

malinjawi:

This PR is the next step in the split Delta deletion-vector (DV) stack, following #12040.

It adds the JVM-side Delta DV scan metadata handoff that consumes the native Velox Delta DV reader layer introduced by #12040, without adding DELETE/UPDATE/MERGE DV DML support yet.

Main changes:

- derive Delta DV payloads from per-file Delta scan metadata while building Velox split info
- materialize Delta DV descriptors on the JVM side through Delta's `StoredBitmap` / `RoaringBitmapArrayFormat.Portable` path
- pass materialized DV payloads from JVM split planning into native Velox scan execution through external split payload buffers
- add Delta 3.3 and Delta 4.0 metadata shims for normalizing DV split metadata
- add split payload plumbing through the Velox iterator, JNI, plan evaluator, runtime, and Substrait-to-Velox conversion path
- strip Spark's injected DV predicate and internal DV columns only after native DV scan handoff is available
- preserve Delta row-index metadata when the plan still needs it for DML-like metrics/row-index consumers
- preserve Delta column-mapping behavior for native Delta scans
- add fallback/reporting visibility for unsupported Delta DV scan paths
- keep unsafe or unsupported paths on Spark rather than silently offloading them

This PR is intentionally JVM scan handoff only:

- no DELETE/UPDATE/MERGE DV DML command implementation yet
- no native DV bitmap construction for DML yet
- no Delta 2.4 / Spark 3.4 DV scan handoff
- no native handoff for `spark.databricks.delta.deletionVectors.useMetadataRowIndex=false` yet

The no-metadata-row-index path currently falls back because it relies on Spark's row-index filtering contract, especially for DML-created DVs. Native support for that path should be added only after it can prove the same correctness contract.

Those pieces will be added in follow-up split PRs.

issue #11901.

How was this patch tested?

Added focused JVM/Velox Delta scan coverage in:

- `backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/DeltaDeletionVectorHandoffSuite.scala`
- `backends-velox/src-delta33/test/scala/org/apache/spark/sql/delta/DeltaSuite.scala`
- `backends-velox/src-delta40/test/scala/org/apache/spark/sql/delta/DeltaDeletionVectorHandoffSuite.scala`
- `gluten-delta/src/test/scala/org/apache/gluten/execution/DeltaSuite.scala`

Covered cases:

- native scan handoff for Delta tables with deletion vectors
- materialized DV payload extraction from Delta scan metadata
- safe stripping of Spark DV predicates after native payload handoff
- partitioned Delta tables with DVs
- multiple DV-bearing files
- prepared Delta scans / stats-skipping paths
- Delta column mapping with DVs
- Spark 4 Delta DV scan behavior
- fallback correctness when metadata row index is disabled
- fallback/reporting visibility for unsupported DV scan paths

Validation run:

- local `./dev/format-scala-code.sh check`
- local `./dev/run-scala-test.sh --force -Pjava-17,spark-3.5,backends-velox,hadoop-3.3,spark-ut,delta -pl backends-velox -s org.apache.gluten.execution.VeloxDeltaSuite` (`19/19` passed)
- local `./dev/run-scala-test.sh --force -Pjava-17,spark-4.0,scala-2.13,backends-velox,hadoop-3.3,spark-ut,delta -pl backends-velox -s org.apache.gluten.execution.VeloxDeltaSuite` (`19/19` passed)
- local `./build/mvn -pl gluten-delta -am -Pbackends-velox -Pspark-3.4 -Pdelta -DskipTests compile`
- local `./build/mvn -pl gluten-delta,backends-velox -am -Pbackends-velox -Pspark-3.5 -Pdelta -DskipTests test-compile`
- local `./build/mvn -pl gluten-delta,backends-velox -am -Pbackends-velox -Pspark-4.0 -Pscala-2.13 -Pdelta -DskipTests test-compile`
- local `git diff --check upstream/main..HEAD`

After addressing review feedback to follow the #10740 split-time handoff shape more closely:

- local `./dev/format-scala-code.sh check`
- local `./build/mvn -pl gluten-delta,backends-velox -am -Pbackends-velox -Pspark-3.5 -Pdelta -DskipTests test-compile`
- local `./build/mvn -pl gluten-delta,backends-velox -am -Pbackends-velox -Pspark-4.0 -Pscala-2.13 -Pdelta -DskipTests test-compile`
- local `git diff --check`

A focused local handoff suite run was attempted, but this machine does not have `darwin/aarch64/libgluten.dylib`, so the suite aborted during SparkContext startup before executing tests. CI should cover the runtime path.

Generated-by: IBM BOB


gluten-delta/src/main/scala/org/apache/gluten/execution/DeltaScanTransformer.scala
```diff
@@ -55,16 +67,30 @@ case class DeltaScanTransformer(
 
   override lazy val fileFormat: ReadFileFormat = ReadFileFormat.ParquetReadFormat
 
+  private lazy val deltaDeletionVectorRegistration
+      : DeltaScanTransformer.DeletionVectorRegistration =
+    DeltaScanTransformer.registerDeletionVectorsFromFileFormat(relation)
```
zhztheplayer: Can we follow the approach in the previous attempt https://github.com/apache/gluten/pull/10740 to pass DVs to native? Would code be simpler that way?

zhztheplayer (commented):
Thanks @malinjawi.

gluten-delta/src/main/scala/org/apache/gluten/execution/DeltaScanTransformer.scala
malinjawi: Thanks for pointing that out @zhztheplayer

 I updated this in d06ec27bb to follow the same split-time handoff shape from #10740 more closely.

The PR now removes the driver-side `DeltaDeletionVectorRegistry` and the relation-wide lazy DV registration from `DeltaScanTransformer`. DV payloads are materialized from each `PartitionedFile`'s Delta metadata when building split info, then passed to native through the external split payload buffers added in this PR.

I kept the #12040-style external payload channel instead of embedding the serialized bitmap directly into Substrait like #10740 did, because #12040 introduced the native Delta split descriptor path that consumes payload buffers separately. This keeps the code simpler while avoiding large binary DV payloads in the Substrait plan.

backends-velox/src-delta33/main/scala/org/apache/spark/sql/delta/stats/PrepareDeltaScan.scala
```diff
@@ -0,0 +1,58 @@
...
+package org.apache.spark.sql.delta.stats
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, V2WriteCommand}
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.delta.{DeltaTable, OptimisticTransaction, PreprocessTableWithDVs}
+import org.apache.spark.sql.delta.sources.DeltaSQLConf
+
+/** Shadow Delta's PrepareDeltaScan to inject backend-specific DV preprocessing. */
+class PrepareDeltaScan(protected val spark: SparkSession)
```
zhztheplayer: Would you explain a bit what this rule is doing and why it's needed?

gluten-arrow/src/main/java/org/apache/gluten/vectorized/ColumnarBatchOutIterator.java
```diff
@@ -31,11 +31,17 @@ public class ColumnarBatchOutIterator extends ClosableIterator<ColumnarBatch>
     implements RuntimeAware {
   private final Runtime runtime;
   private final long iterHandle;
+  private final Object retainedReference;
 
   public ColumnarBatchOutIterator(Runtime runtime, long iterHandle) {
+    this(runtime, iterHandle, null);
+  }
+
+  public ColumnarBatchOutIterator(Runtime runtime, long iterHandle, Object retainedReference) {
```
zhztheplayer: What is `retainedReference` for?

cpp/velox/substrait/SubstraitToVeloxPlan.h
```diff
@@ -50,6 +51,9 @@ struct SplitInfo {
   /// The metadata columns associated with partitioned table.
   std::vector<std::unordered_map<std::string, std::string>> metadataColumns;
 
+  /// Optional externally provided deletion vector payloads aligned with metadataColumns.
+  std::vector<std::optional<SplitPayloadBufferView>> deletionVectorPayloads;
+
```
zhztheplayer: Can we add `DeltaSplitInfo : SplitInfo`?

cpp/velox/compute/VeloxPlanConverter.cc
```diff
@@ -75,6 +111,25 @@ std::shared_ptr<SplitInfo> parseScanSplitInfo(
     for (const auto& metadataColumn : file.metadata_columns()) {
       metadataColumnMap[metadataColumn.key()] = metadataColumn.value();
     }
+    for (const auto& otherMetadataColumn : file.other_const_metadata_columns()) {
+      if (auto unpackedValue = unpackMetadataValue(otherMetadataColumn.value())) {
```
zhztheplayer, line 158: Can we extend the substrait proto, to avoid placing the essential info in `otherMetadataColumn`?

backends-velox/src-delta33/main/scala/org/apache/spark/sql/delta/stats/PrepareDeltaScan.scala
malinjawi: Thanks @zhztheplayer. This is a Delta 3.3 compatibility shim around Delta's PrepareDeltaScan.

It keeps Delta's normal scan preparation first, including stats skipping, metadata-query optimization, and transaction read tracking. After that, Gluten runs DV preprocessing so the scan exposes Delta's internal DV metadata/row-deleted column.

Gluten needs that metadata to materialize the per-file DV payload for the native split. Later, once the native split has the payload, Gluten strips Spark's synthetic DV predicate/internal columns so Velox applies the DV filter natively and we do not filter twice.

I can expand the comment to make this flow clear.

gluten-arrow/src/main/java/org/apache/gluten/vectorized/ColumnarBatchOutIterator.java
malinjawi: Yea that was unclear name so I renamed this to `retainedSplitPayloadBuffers` and typed it as ByteBuffer[][]. 

It keeps the Java-owned direct buffers reachable while Velox holds native views into them, so the payload memory cannot be GC’d during scan execution.

cpp/velox/substrait/SubstraitToVeloxPlan.h
malinjawi: Addressed. I added `DeltaSplitInfo` on the Velox side and moved the Delta DV split state there instead of keeping it on the generic split metadata path.

cpp/velox/compute/VeloxPlanConverter.cc
malinjawi: Got it. I extended the Substrait read options with typed Delta DV fields and stopped carrying this essential Delta metadata through `other_const_metadata_columns`.

gluten-substrait/src/main/resources/substrait/proto/substrait/algebra.proto
```diff
@@ -197,6 +197,18 @@ message ReadRel {
...
+        enum RowIndexFilterType {
+          KEEP_ALL = 0;
+          IF_CONTAINED = 1;
+          IF_NOT_CONTAINED = 2;
+        }
+        RowIndexFilterType row_index_filter_type = 1;
+        bool has_deletion_vector = 2;
+        uint64 deletion_vector_cardinality = 3;
+        uint32 deletion_vector_payload_index = 4;
+      }
```
zhztheplayer, line 210: In this PR, can we embed the serialized DV in the read options?

And we can improve the perf with benchmark numbers in further development iterations.

backends-velox/src-delta33/main/scala/org/apache/spark/sql/delta/stats/PrepareDeltaScan.scala
zhztheplayer: Thanks for the explanation. Is it replacing the same class from Delta? If so we'd avoid doing this in production. Have you checked the rule in the [PoC](https://github.com/apache/gluten/pull/10740)?

zhztheplayer (commented):
Thanks for iterating.

gluten-substrait/src/main/resources/substrait/proto/substrait/algebra.proto
malinjawi: Done @zhztheplayer

 I updated this PR to embed the serialized DV bytes directly in `DeltaReadOptions` as `serialized_deletion_vector`.

This removes the separate payload index / side-band payload registry path. The JVM now writes the materialized DV payload into the Substrait read options, and Velox copies it into `DeltaSplitInfo`-owned storage before creating the DV descriptor view.

I will keep deeper payload-size/perf work for the follow-up as you suggested thanks.

backends-velox/src-delta33/main/scala/org/apache/spark/sql/delta/stats/PrepareDeltaScan.scala
malinjawi: Yes, the earlier version did replace/shadow Delta’s `PrepareDeltaScan`, and I agree we should avoid that in production.

I removed that approach. The current version no longer defines `org.apache.spark.sql.delta.stats.PrepareDeltaScan`; it uses a Gluten-scoped rule under `org.apache.gluten.extension` instead.

I also checked the [PoC](https://github.com/apache/gluten/pull/10740) direction. The intent is the same: expose the DV metadata needed for native handoff without replacing Delta classes. After the Spark 3.5 CI feedback, I narrowed the rule further so it only rewrites eligible DV scans and leaves unpinned `TahoeLogFileIndex` / subquery-shaped plans untouched.

zhztheplayer (commented):
Hi @malinjawi, would you help split the PR?

I recommend to first raise a PR to add a utility to extract DV info from Delta scan to essential DV info, something like [this API](https://github.com/apache/gluten/pull/10740/changes#diff-79019208b21aca99fd6e93c9ab3554c6c54f82aad5111e100cb21886ba1b7c7eR29) in the PoC. I feel that part needs to be reviewed first.

malinjawi:
> Hi @malinjawi, would you help split the PR?
> 
> I recommend to first raise a PR to add a utility to extract DV info from Delta scan to essential DV info, something like [this API](https://github.com/apache/gluten/pull/10740/changes#diff-79019208b21aca99fd6e93c9ab3554c6c54f82aad5111e100cb21886ba1b7c7eR29) in the PoC. I feel that part needs to be reviewed first.

Thanks @zhztheplayer, that makes sense.

I split the original PR into smaller pieces as suggested:

- #12197 adds the focused Delta DV scan-info extraction utility first.
  - It extracts the essential per-file DV scan info from `PartitionedFile` metadata.
  - It covers Delta 3.3 and Delta 4.0.
  - It is utility-only: no Substrait changes, no native/C++ changes, and no scan offload behavior changes.

- #12198 is the follow-up handoff PR stacked after #12197.
  - It consumes the utility from #12197.
  - It adds the JVM/Substrait/Velox scan handoff path.

I’ll keep #12198 as the follow-up and would appreciate review on #12197 first so we can settle the utility/API shape before continuing with the full handoff.

malinjawi:
Now that #12197 has landed, the active split PR for the JVM Delta DV scan handoff is #12198.

#12198 is rebased on current main, has the accepted scan-info utility as its base, and is now marked ready for review. This broader PR should be treated as superseded by the split stack unless reviewers prefer a different path.

PR #12142, opened 2026-05-26
[GLUTEN-12143][VL] Route bitmap_construct_agg to native Velox execution

minni31:
## Route `bitmap_construct_agg` to Native Velox Execution

This PR enables native Velox execution for the `bitmap_construct_agg` aggregate function, avoiding fallback to vanilla Spark.

### Changes

1. **Expression Registration** (`ExpressionNames.scala`): Added `BITMAP_CONSTRUCT_AGG` constant.
2. **Shim Registration** (`Spark35Shims.scala`, `Spark40Shims.scala`, `Spark41Shims.scala`): Register `Sig[BitmapConstructAgg]` mapping to the substrait function name.
3. **Native Validator** (`SubstraitToVeloxPlanValidator.cc`): Added `"bitmap_construct_agg"` to the `supportedAggFuncs` set so the native validator accepts this function.
4. **Plan-shape assertion test** (`GlutenBitmapExpressionsQuerySuite.scala` for spark35/40/41): Verifies the query plan contains `HashAggregateExecBaseTransformer` (native execution) rather than falling back.

### Test Results

| Spark Version | Tests | Result |
|---------------|-------|--------|
| Spark 3.5 | 11/11 | ✅ Pass |
| Spark 4.0 | 11/11 | ✅ Pass |
| Spark 4.1 | 13/13 | ✅ Pass |

All tests include the new plan-shape assertion test confirming native Velox execution.

Related issue: #12143


weiting-chen (approved):

shims/spark35/src/main/scala/org/apache/gluten/sql/shims/spark35/Spark35Shims.scala
```diff
@@ -80,7 +80,8 @@ class Spark35Shims extends SparkShims {
       Sig[RegrSlope](ExpressionNames.REGR_SLOPE),
       Sig[RegrIntercept](ExpressionNames.REGR_INTERCEPT),
       Sig[RegrSXY](ExpressionNames.REGR_SXY),
-      Sig[RegrReplacement](ExpressionNames.REGR_REPLACEMENT)
+      Sig[RegrReplacement](ExpressionNames.REGR_REPLACEMENT),
+      Sig[BitmapConstructAgg](ExpressionNames.BITMAP_CONSTRUCT_AGG)
```
weiting-chen, line 84: **Consider adding plan-shape assertion in tests**

**Problem:** `GlutenBitmapExpressionsQuerySuite` inherits upstream Spark tests that only use `checkAnswer()` — they verify correctness but do not assert that native Velox execution is actually active. A regression that silently disables offloading would go undetected since vanilla Spark also produces correct results.

**Evidence:**
```scala
class GlutenBitmapExpressionsQuerySuite
  extends BitmapExpressionsQuerySuite
  with GlutenSQLTestsTrait {}
// All upstream tests only call checkAnswer(df, expected)
```

**Suggested Fix:** Add at least one Gluten-specific test asserting the native transformer is present:
```scala
test("bitmap_construct_agg routes to native") {
  val df = spark.sql("SELECT bitmap_construct_agg(col) FROM values (1L), (2L) AS t(col)")
  val plan = df.queryExecution.executedPlan
  assert(collect(plan) { case h: HashAggregateExecTransformer => h }.nonEmpty,
    "Expected native HashAggregateExecTransformer in plan")
}
```

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
```diff
@@ -1323,7 +1323,8 @@ bool SubstraitToVeloxPlanValidator::validate(const ::substrait::AggregateRel& ag
       "regr_slope",
       "regr_intercept",
       "regr_sxy",
-      "regr_replacement"};
+      "regr_replacement",
+      "bitmap_construct_agg"};
```
weiting-chen, line 1327: **Follow-up opportunity: register bitmap_or_agg (and bitmap_and_agg for Spark 4.1)**

**Problem:** `bitmap_or_agg` was introduced in the same Spark 3.5 commit as `bitmap_construct_agg`, and `bitmap_and_agg` in Spark 4.1. If native Velox implementations exist for these, registering them together would avoid unnecessary columnar-to-row transitions when users combine bitmap functions in the same query stage.

**Investigation Needed:** Confirm whether `bitmap_or_agg` and `bitmap_and_agg` have native Velox implementations (look for registration in `cpp/velox/udf/` or the Velox aggregate function registry). If yes, consider adding them in a follow-up PR:
```cpp
      "bitmap_construct_agg",
      "bitmap_or_agg",
      "bitmap_and_agg"};
```

weiting-chen (commented):
Thanks for the clean implementation — the three-layer plumbing (constant + shim registration + C++ validator) follows the established pattern correctly.

Two suggestions for consideration (non-blocking):

cpp/velox/substrait/SubstraitToVeloxPlanValidator.cc
minni31: I've Velox PR for bitmap_or_agg in review phase. Will add follow up PR for Gluten once Velox PR merges. Thanks!

shims/spark35/src/main/scala/org/apache/gluten/sql/shims/spark35/Spark35Shims.scala
minni31: Added in all spark version. Thanks @weiting-chen

yaooqinn:
+1. Three-layer plumbing (`Sig[BitmapConstructAgg]` on all 3 shims + `supportedAggFuncs` entry + plan-shape assertion) follows the same shape as the existing `BloomFilterAgg` registration in this file, which is the right precedent here.

Grounded the plan-shape test guard: `bitmap_bit_position` is registered in `gen-function-support-docs.py` (BitmapBitPosition -> bitmap_bit_position), so the inner expression won't silently fallback and mask a `bitmap_construct_agg` regression. Good defensive test shape.

yaooqinn:
Following up on my earlier +1 — looked deeper after the ClickHouse CI failures and I think the registration is at the wrong layer.

**CH CI is PR-tied (not infra flake).** 5 tests fail in `GlutenBitmapExpressionsQuerySuite` (CH backend) with:

```
GlutenException: Unknown aggregate function bitmap_construct_agg
  at AggregateFunctionParserFactory::get (cpp-ch/local-engine/Parser/AggregateFunctionParser.cpp:255)
```

Root cause: `Sig[BitmapConstructAgg]` is registered in `shims/spark35,40,41/SparkXShims.aggregateExpressionMappings`, which is **all-backend** — so the CH planner also accepts `bitmap_construct_agg` and pushes it down to CH native, where the parser isn't registered, hence the failure.

**Sibling pattern — `BloomFilterAgg`:**

- `backends-velox/.../VeloxSparkPlanExecApi.scala:1120` — `Sig[VeloxBloomFilterAggregate](ExpressionNames.BLOOM_FILTER_AGG)` is registered in `extraExpressionMappings`, which is **Velox-only**. `BLOOM_FILTER_AGG` does **not** appear in any `SparkXShims.aggregateExpressionMappings`.

Suggest moving `Sig[BitmapConstructAgg]` from the three `SparkXShims` files to `VeloxSparkPlanExecApi.extraExpressionMappings`. `BITMAP_CONSTRUCT_AGG` constant in `shims/common/ExpressionNames.scala` can stay (it's just a string), or move it with the Sig — either is fine.

That should keep CH on the vanilla fallback path and let Velox continue to route natively.

(Sorry for the late catch — my +1 cited `bitmap_bit_position` registration as a fallback guard, but that doesn't protect against the outer `bitmap_construct_agg` Sig being visible to non-Velox backends. The shim-vs-backend layering is the actual gate.)

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
```diff
@@ -1134,7 +1134,13 @@ class VeloxSparkPlanExecApi extends SparkPlanExecApi with Logging {
       Sig[AssertNotNull](ExpressionNames.ASSERT_NOT_NULL),
       // For test purpose.
       Sig[VeloxDummyExpression](VeloxDummyExpression.VELOX_DUMMY_EXPRESSION)
-    )
+    ) ++ scala.util.Try(
```
baibaichen: This reflection-based registration is a workaround, not the right long-term shape — let's move it to the shims.

baibaichen:
The CH CI failure (`bitmap_construct_agg routes to native *** FAILED *** — List() was empty Expected native HashAggregateExecBaseTransformer in plan`) needs to be fixed by adding a blacklist entry on the CH side, following the same pattern used for `regr_slope` / `regr_intercept` / `regr_sxy` / `regr_replacement`: https://github.com/apache/gluten/blob/main/backends-clickhouse/src/main/scala/org/apache/gluten/utils/CHExpressionUtil.scala#L200

minni31:
> Following up on my earlier +1 — looked deeper after the ClickHouse CI failures and I think the registration is at the wrong layer.
> 
> **CH CI is PR-tied (not infra flake).** 5 tests fail in `GlutenBitmapExpressionsQuerySuite` (CH backend) with:
> 
> ```
> GlutenException: Unknown aggregate function bitmap_construct_agg
>   at AggregateFunctionParserFactory::get (cpp-ch/local-engine/Parser/AggregateFunctionParser.cpp:255)
> ```
> 
> Root cause: `Sig[BitmapConstructAgg]` is registered in `shims/spark35,40,41/SparkXShims.aggregateExpressionMappings`, which is **all-backend** — so the CH planner also accepts `bitmap_construct_agg` and pushes it down to CH native, where the parser isn't registered, hence the failure.
> 
> **Sibling pattern — `BloomFilterAgg`:**
> 
> * `backends-velox/.../VeloxSparkPlanExecApi.scala:1120` — `Sig[VeloxBloomFilterAggregate](ExpressionNames.BLOOM_FILTER_AGG)` is registered in `extraExpressionMappings`, which is **Velox-only**. `BLOOM_FILTER_AGG` does **not** appear in any `SparkXShims.aggregateExpressionMappings`.
> 
> Suggest moving `Sig[BitmapConstructAgg]` from the three `SparkXShims` files to `VeloxSparkPlanExecApi.extraExpressionMappings`. `BITMAP_CONSTRUCT_AGG` constant in `shims/common/ExpressionNames.scala` can stay (it's just a string), or move it with the Sig — either is fine.
> 
> That should keep CH on the vanilla fallback path and let Velox continue to route natively.
> 
> (Sorry for the late catch — my +1 cited `bitmap_bit_position` registration as a fallback guard, but that doesn't protect against the outer `bitmap_construct_agg` Sig being visible to non-Velox backends. The shim-vs-backend layering is the actual gate.)

I discussed with @baibaichen and he suggested to put this function in blacklist at CH backend and put it in SHIM layers only for Velox. Updated the code. Thanks @yaooqinn for the review.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxSparkPlanExecApi.scala
minni31: Done. Moved to Shim and put it in blacklist on CH backend.

minni31:
> The CH CI failure (`bitmap_construct_agg routes to native *** FAILED *** — List() was empty Expected native HashAggregateExecBaseTransformer in plan`) needs to be fixed by adding a blacklist entry on the CH side, following the same pattern used for `regr_slope` / `regr_intercept` / `regr_sxy` / `regr_replacement`: https://github.com/apache/gluten/blob/main/backends-clickhouse/src/main/scala/org/apache/gluten/utils/CHExpressionUtil.scala#L200

Done, addressed. Please help with review @baibaichen

baibaichen:
LGTM

baibaichen (approved):

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/clickhouse/ClickHouseTestSettings.scala
```diff
@@ -90,6 +90,8 @@ class ClickHouseTestSettings extends BackendTestSettings {
     // Exception.
     .exclude("column pruning - non-readable file")
   enableSuite[GlutenBitmapExpressionsQuerySuite]
+    // bitmap_construct_agg is not supported natively in CH backend.
+    .excludeCH("bitmap_construct_agg routes to native")
```
baibaichen, line 94: remove?

gluten-ut/spark41/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
```diff
@@ -293,6 +293,10 @@ class VeloxTestSettings extends BackendTestSettings {
       "INCONSISTENT_BEHAVIOR_CROSS_VERSION: compatibility with Spark 2.4/3.2 in reading/writing dates")
     // Doesn't support unhex with failOnError=true.
     .exclude("CONVERSION_INVALID_INPUT: to_binary conversion function hex")
+    // bitmap_construct_agg offloaded to Velox throws GlutenException instead of
+    // SparkArrayIndexOutOfBoundsException.
+    .exclude("INVALID_BITMAP_POSITION: position out of bounds")
```
baibaichen, line 298: Better to add an issue to track this, we need translate expcetion at jvm

gluten-ut/spark35/src/test/scala/org/apache/gluten/utils/clickhouse/ClickHouseTestSettings.scala
minni31: This one checks the native plan path, specifically added to test the native plan. Since for CH backend it runs on spark, this test fail in CH backend.

gluten-ut/spark41/src/test/scala/org/apache/gluten/utils/velox/VeloxTestSettings.scala
minni31: Added issue here: https://github.com/apache/gluten/issues/12238

PR #12147, opened 2026-05-26
[VL] Fix false-EOF probe in ColumnarCachedBatchSerializer.read trailing markers

yaooqinn:

`ColumnarCachedBatchSerializer.read` used `input.available() > 0` as
an EOF probe before reading the trailing `hasStats` / `hasSchema`
boolean markers. This PR replaces the probe with a direct `readBoolean()`
inside a narrow `try/catch` on `KryoException` whose message starts
with `"Buffer underflow"`, which is the actual signal Kryo emits at
true end-of-input.

The V1 wire back-compat behavior (absent trailing booleans read as
`null`, not throw) is preserved — and is locked by the existing test
`ColumnarCachedBatchKryoSuite#"V1 wire (no trailing hasStats/hasSchema booleans) reads as stats=null/schema=null"`.

`Kryo Input.available()` returns
`(limit - position) + underlyingStream.available()`. The JDK
`InputStream.available()` contract permits any implementation to
return `0` even when more data follows — `BufferedInputStream` over
shuffle-spill / network chunk boundaries routinely does so. When the
Kryo buffer is drained AND the underlying stream reports `0` at the
trailing-boolean byte position, the probe falsely concludes EOF,
skips `hasStats`, and the next `readClassAndObject` interprets the
stats payload (which contains the schema JSON) as a class name.

Production stack we observed (truncated):

```
ClassNotFoundException: {"type":"struct","fields":[...
  at DefaultClassResolver.readName(DefaultClassResolver.java:154)
  at DefaultClassResolver.readClass(DefaultClassResolver.java:133)
  at Kryo.readClass(Kryo.java:693)
  at Kryo.readClassAndObject(Kryo.java:804)
  at ColumnarCachedBatchSerializer$$anon$2.hasNext(...)
```

The length-bound `require(... <= maxLen ...)` guard from
`491070bf34` is preserved — it remains useful against
`NegativeArraySizeException` / oversized allocation and is
orthogonal to the EOF-probe fix.

## Why catch `KryoException` by message prefix?

Kryo 4.x (the version on Gluten's classpath via `kryo-shaded`) has
no dedicated `KryoBufferUnderflowException` subclass — that was added
in Kryo 5.x. The message-prefix check is the narrowest filter we can
apply on 4.x without swallowing real corruption (e.g. a
`ClassNotFoundException` wrapped during `readClassAndObject`, which
does **not** start with `"Buffer underflow"`).

No (bug fix in a build that has not shipped to any tagged release).

## How was this patch tested?

New `ColumnarCachedBatchKryoBoundaryProbeBugSuite`:

- `"multi-batch deserialize survives boundary-aligned trailing-boolean probe"`
  — deterministic repro of the production stack using a
  1-byte-per-read `InputStream` that returns `available() == 0` at
  the trailing-boolean position. Fails on the pre-patch code, passes
  after the fix.

Existing `ColumnarCachedBatchKryoSuite#"V1 wire ..."` continues to
pass, locking the V1 silent-null back-compat contract.

Verified locally on `-Pspark-4.1 -Pscala-2.13`; affected suites
(`Framed`, `BuildFilter`, `StatsBlob`, `ShipBlockerMarshal`,
`IntFamilyMarshal`) all green (23/23). CI 60/60 green on
`da460551e7`.

---

**Was this patch authored or co-authored using generative AI tooling?**
No

### DIFF
--- a/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
+++ b/backends-velox/src/main/scala/org/apache/spark/sql/execution/ColumnarCachedBatchSerializer.scala
@@ -44,7 +44,7 @@ import org.apache.spark.sql.vectorized.ColumnarBatch
 import org.apache.spark.storage.StorageLevel
 import org.apache.spark.unsafe.types.UTF8String
 
-import com.esotericsoftware.kryo.{Kryo, Serializer => KryoSerializer}
+import com.esotericsoftware.kryo.{Kryo, KryoException, Serializer => KryoSerializer}
 import com.esotericsoftware.kryo.DefaultSerializer
 import com.esotericsoftware.kryo.io.{Input, Output}
 import org.apache.arrow.c.ArrowSchema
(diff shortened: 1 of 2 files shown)


liuneng1994 (commented):
The underlying diagnosis is correct — `Kryo Input.available()` is not a reliable EOF probe over chunk-boundary `BufferedInputStream`s, and the repro test is convincing. Two real concerns plus a doc/code mismatch worth fixing before merge.

## 🟠 PR description contradicts the actual code

The description claims:
> **Spark cached blocks live and die within a single application instance.** … the V1 wire era (commit `97632d8435`) was never released to any tag … so the guard protects nothing.
> A truncated stream now surfaces `KryoException` rather than silently returning null …
> A new test `…BoundaryProbeBugSuite#"truncated wire ... fails fast, not silently"` locks this contract.

But the implemented code does the opposite: it explicitly **preserves** V1 back-compat via a `try/catch` on `KryoException`-with-`"Buffer underflow"`-prefix, and the comments even cite `ColumnarCachedBatchKryoSuite#"V1 wire ..."` as a locked contract. The "truncated wire … fails fast" test is not in the diff — the only new test is the multi-batch boundary probe.

Pick a story and align: either truly remove V1 compat (drop the `try/catch`, add the fail-fast test, keep the test name promised in the description), or keep the compat and update the description so reviewers/git-log readers aren't misled about behavior.

## 🟠 Matching `KryoException` by message-prefix `"Buffer underflow"` is fragile

```scala
private def isBufferUnderflow(e: KryoException): Boolean = {
  val msg = e.getMessage
  msg != null && msg.startsWith("Buffer underflow")
}
```

Kryo's exception message text is not part of any stability contract; an upstream rewording or i18n would silently flip this branch and we'd start re-throwing real EOF (or, worse, swallowing real corruption that happens to share the prefix). Two safer options:

1. If on Kryo 5.x, catch `com.esotericsoftware.kryo.KryoBufferUnderflowException` (the dedicated subclass) instead of the parent + message match.
2. If staying on Kryo 4.x, side-step the question entirely by writing a 1-byte sentinel `0xFF` "end-of-record" tag after `hasSchema` on the write path, and treating its absence (only on legacy V1) as the back-compat trigger. That eliminates the EOF-detection requirement.

## 🟡 Silent under-flow → silent stats drop

With the current `try/catch`, a genuinely truncated/corrupt V2 stream gets `hasStats = false` instead of an exception. The batch then flows through `buildFilter`'s stats=null pass-through and is read unpruned — **no observable error, but unprunable until the cache is rebuilt**. At minimum, log a `warn` once when this branch fires under V2 (e.g. when the rest of the record proves a V2 producer — but you have no marker yet, which is the other reason a sentinel/version byte helps).

## ✅ Nits

- The repro test is good and deterministic; nice fixture design. Consider also adding an actual `BufferedInputStream` over a `ChunkedByteBuffer`-style source as a second case, since that's the production stack quoted in the description and exercises the real path rather than only the `LyingOneByteStream` synthetic.
- `readOptionalSchema` now does its own `try/catch`; if write path always emits both `hasStats` and `hasSchema` markers, the `hasSchema` underflow case is unreachable on any non-V1 stream. Adding a comment that explicitly says "only reachable on V1" (and pointing at the producer site that proves it) would help future readers.

liuneng1994 (commented):
## PR description contradicts the code

Cross-referencing the existing locked test in `ColumnarCachedBatchKryoSuite.scala`:

```scala
test("V1 wire (no trailing hasStats/hasSchema booleans) reads as stats=null/schema=null") {
  ...
  assert(read.stats === null, "absent trailing hasStats must read as null, not throw")
  assert(read.schema === null, "absent trailing hasSchema must read as null, not throw")
}
```

This contract **requires** V1 compat to be preserved — and the new `try/catch (KryoException if isBufferUnderflow) => false` is exactly what satisfies it. But the PR description says the opposite:

| description claims | code / tests actually |
|---|---|
| "V1 wire era … never released … guard protects nothing" | try/catch preserves V1 compat; locked by existing `"V1 wire …"` test |
| "A truncated stream now surfaces KryoException rather than silently returning null" | truncated stream is caught and returns `null` stats — same silent behavior |
| Cites `BoundaryProbeBugSuite#"truncated wire … fails fast, not silently"` | new file only contains 1 test (`"multi-batch deserialize survives …"`) — the cited test doesn't exist |

Suggested fix: keep the code as-is (preserving V1 compat is correct), and trim the description down to honestly what the patch does — "fix the false-EOF probe in `Input.available()` so the trailing booleans are read across chunk boundaries; V1 compat is now achieved via `try/catch` on `KryoException` instead". Drop the "Spark cached blocks live and die within a single application instance" paragraph and the cite to a non-existent test.

liuneng1994 (approved):

PR #12152, opened 2026-05-27
[VL] Skip escape argument when offloading `Like` expressions without backslashes

zhulipeng:

When offloading `Like` to Velox, omit the escape literal argument and emit
the 2-arg form (`like(input, pattern)`) when both:

1. the pattern is a constant `Literal`, and
2. the escape char is Spark's default `\` AND the pattern does not contain `\`.

Otherwise we still emit the 3-arg form (`like(input, pattern, escape)`) as before.

Spark's `Like` node always carries an `escapeChar` (defaulting to `\`) even when
the SQL did not specify `ESCAPE`. We previously always sent the 3-arg form to
Velox, which forces Velox's `makeLike` (`Re2Functions.cpp`) to take the
escape-aware path: `parsePattern` runs an extra unescape pass, and
`determinePatternKind` runs with `escapeChar.has_value() == true`, even when
no actual escaping is needed.

When the pattern literal contains no `\`, the 2-arg and 3-arg forms are
semantically identical, so we can safely send the cheaper 2-arg form. 

## How was this patch tested?

- Existing `Like` query coverage in `VeloxStringFunctionsSuite` (`like` /
    `rlike` / `ilike`) — query results unchanged.
- TPC-H Q13 end-to-end run at 6 TB scale — see Performance section.
- `./dev/format-scala-code.sh` clean.

 ## Performance

**TPC-H Q13 @ 6 TB scale: >6% end-to-end latency reduction.**

Q13's `l_comment NOT LIKE '%special%requests%'` filter scans every lineitem row.
With the 3-arg form, the constant-pattern fast paths in `determinePatternKind`
are bypassed and Velox falls back to the generic `LikeWithRe2` path, which
hot-loops in `re2::DFA::InlinedSearchLoop` — CPU profiling shows
`InlinedSearchLoop` accounts for **>8% of total cycles** on Q13.

Sending the 2-arg form lets `makeLike` recognize this as the
`kSubstrings` shape and dispatch to the dedicated `OptimizedLike<kSubstrings>`
kernel, eliminating the RE2 DFA cost. No regression observed on other queries.

Reviewed-by: Claude claude-opus-4-7


zhulipeng:
@philo-he I am new to the Gluten community and not clear about the maintainer/reviewer of this module. Could you kindly provide assistance or arrange relevant staff to review this patch?

philo-he (approved):
LGTM. Thanks. Will merge if no further comments from the community.

zhulipeng:
The failed CI checks should have no relationship with this patch.

PR #12170, opened 2026-05-28
[GLUTEN-12169][CH] Fix str_to_map nullable input null-map access

### LINKED ISSUE #12169: [CH] undefined behavior, or a crash in `SparkFunctionStrToMap`
### Backend

CH (ClickHouse)

### Bug description

 `cpp-ch/local-engine/Functions/SparkFunctionStrToMap.cpp:241-244` reads one past the nullable null map.

`executeImpl` iterates rows with `i`, but checks `(*null_map)[n]`, where `n` is `offsets.size()`. For any non-empty `Nullable(String)` input, this is one past the end of the null map. The result can be a wrong null decision, undefined behavior, or a crash. The null check should use the current row index `i`.

### Gluten version

_No response_

### Spark version

None

### Spark configurations

_No response_

### System information

_No response_

### Relevant logs

```bash

```

lgbo-ustc:

SparkFunctionStrToMap iterates rows with index i, but the nullable string path checked null_map[n], where n is offsets.size(). For any non-empty nullable input this reads one element past the null map and can produce an incorrect null decision, undefined behavior, or a crash.

Use the current row index when reading the nullable null map. Add a ClickHouse backend regression test that evaluates str_to_map over a nullable string expression containing both null and non-null rows, then compares the result with vanilla Spark.

closed #12169

## How was this patch tested?

UT

### DIFF
--- a/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala
+++ b/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala
@@ -489,6 +489,20 @@ class GlutenFunctionValidateSuite extends GlutenClickHouseWholeStageTransformerS
     runQueryAndCompare(sql1)(checkGlutenPlan[ProjectExecTransformer])
   }
 
+  test("test str2map with nullable string input") {
+    val sql =
+      """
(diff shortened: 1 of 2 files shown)


zzcclp (approved):
LGTM

PR #12173, opened 2026-05-28
[GLUTEN-12172][CH] Fix group limit first array result offset

### LINKED ISSUE #12172: [CH] Fix group limit first array result offset
### Backend

CH (ClickHouse)

### Bug description

`cpp-ch/local-engine/AggregateFunctions/GroupLimitFunctions.cpp:138-145` reads `ColumnArray` offsets from an empty result column.

`insertResultInto` calls `result_array_offsets.back()` before pushing the first offset. A newly created aggregate result column has no offsets yet, so the first output row can access an empty container. The previous offset should be treated as `0` when `result_array_offsets.empty()`.

### Gluten version

_No response_

### Spark version

None

### Spark configurations

_No response_

### System information

_No response_

### Relevant logs

```bash

```

lgbo-ustc:

RowNumGroupArraySorted writes aggregate results into a newly created ColumnArray. For the first output row the array offsets vector can be empty, but insertResultInto read result_array_offsets.back() before appending the first offset. That is undefined behavior and can crash when aggregate top-k writes its first result.

Treat an empty offsets vector as having previous offset 0 before appending the next cumulative offset. Add a ClickHouse backend regression test that forces row_number top-k through the aggregate group limit path and validates the first array result row against vanilla Spark.

closed #12172

## How was this patch tested?

UTs


zzcclp (approved):
LGTM

PR #12175, opened 2026-05-28
[GLUTEN-12174][CH] Fix flatten nullable inner array row handling

### LINKED ISSUE #12174: [CH] Fix flatten nullable inner array row handling
### Backend

CH (ClickHouse)

### Bug description

For `Array(Nullable(Array(T)))`, the code scans all nested elements and returns `input_rows_count` default nullable rows as soon as it sees one null inner array. This makes unrelated rows null. Example: `[[[1], null], [[2]]]` should only make the first row null under Spark-style flatten semantics, but the current implementation returns both rows as null.

### Gluten version

_No response_

### Spark version

None

### Spark configurations

_No response_

### System information

_No response_

### Relevant logs

```bash

```

lgbo-ustc:

SparkArrayFlatten handled Array(Nullable(Array(T))) by scanning all nested inner arrays and returning a fully-null result column as soon as any inner array was null. That made unrelated rows null, even though Spark flatten semantics only null the outer row that contains a null inner array.

Build a result null map per outer row, mark only rows containing null inner arrays as null, and keep non-null rows using the flattened array offsets. Add a ClickHouse backend regression test where the first row contains a null inner array and the second row remains non-null.

closed #12174 

## How was this patch tested?

UTs

### DIFF
--- a/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala
+++ b/backends-clickhouse/src/test/scala/org/apache/gluten/execution/GlutenFunctionValidateSuite.scala
@@ -668,6 +668,22 @@ class GlutenFunctionValidateSuite extends GlutenClickHouseWholeStageTransformerS
     }
   }
 
+  test("test flatten with nullable inner arrays") {
+    val sql =
+      """
+        |select id, flatten(arr)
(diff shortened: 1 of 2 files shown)


zzcclp (approved):
LGTM

PR #12176, opened 2026-05-28
[VL] Compute real nullCount for flat-encoded unsupported types in cached-batch partition stats

yaooqinn:

`VeloxColumnarBatchSerializer::computeStats` did not accrue `nullCount` in the
`default:` arm of the flat-encoding TypeKind switch. For flat-encoded
unsupported TypeKinds (`VARBINARY`, `OPAQUE`, `UNKNOWN`) with V2 partition
stats enabled, `stats.nullCount` was 0 regardless of real null count, so
Spark `IsNull` predicate pruning silently dropped null-bearing partitions.

Mirror the non-flat path's `countNullsAny` call. Wire format unchanged
(the `nullCount` slot has always been emitted, just zero for the affected
types). No `SQLConf` change.

## How was this patch tested?

New e2e test in `ColumnarCachedBatchE2ESuite`. Pre-fix: `0 did not equal 1`.
Post-fix: pass.

## (Optional) Was this patch authored or co-authored using generative AI tooling?

Generated-by: Claude claude-opus-4.7

### DIFF
--- a/backends-velox/src/test/scala/org/apache/spark/sql/execution/ColumnarCachedBatchE2ESuite.scala
+++ b/backends-velox/src/test/scala/org/apache/spark/sql/execution/ColumnarCachedBatchE2ESuite.scala
@@ -488,4 +488,25 @@ class ColumnarCachedBatchE2ESuite
       }
     }
   }
+
+  test("V2 stats: IsNull on VARBINARY null-bearing partition is not pruned") {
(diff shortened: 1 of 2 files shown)


liuneng1994 (approved):

yaooqinn:
Thanks for the review @liuneng1994!

PR #12178, opened 2026-05-28
[GLUTEN-8232][VL] Followup to update doc for VCPKG dynamic openssl feature

zhouyuan:
follow up on #11444 to update doc

Related issue: #8232


yaooqinn (approved):

docs/get-started/VeloxDynamicOpenSSL.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+## How It Works
+
+When `VCPKG_DYNAMIC_OPENSSL=ON` is set:
+
+1. **vcpkg Configuration**: The `dynamic-openssl` feature is enabled in `vcpkg.json`, which includes OpenSSL with FIPS support
+2. **Triplet Override**: Custom vcpkg triplets (`x64-linux-avx.cmake` and `arm64-linux-neon.cmake`) detect the environment variable and switch OpenSSL from static to dynamic linkage
+3. **Build Process**: The `init.sh` script adds the `--x-feature=dynamic-openssl` flag to vcpkg install
+4. **Library Packaging**: OpenSSL shared libraries are excluded from static linking but must be available at runtime
```
philo-he, line 43: Should we add the following explicit clarification for users? 

* At runtime, LD_LIBRARY_PATH should point to the OS-provided OpenSSL package, which includes libssl.so, libcrypto.so, and the FIPS-certified fips.so, so that these libraries are loaded instead of the ones built at compile time by vcpkg.

* Users should ensure compatibility between the OpenSSL libraries (libssl.so and libcrypto.so) used at link time and those available at runtime. We recommend using the same major version to avoid potential compatibility issues.

philo-he (approved):
Looks good overall. One minor comment. Thanks.

docs/get-started/VeloxDynamicOpenSSL.md
FelixYBW: @zhouyuan can you also add the openssl version in vcpkg here?

zhouyuan: @philo-he @FelixYBW Thanks, I add more notes to callout the version and usage

PR #12177, opened 2026-05-28
[MINOR] Upgrade Spark 4.1 to 4.1.2

yaooqinn:
### What changes were proposed in this pull request?

Bumps the `spark-4.1` profile from **4.1.1** to **4.1.2** (Spark patch release).

Changes:
- `pom.xml` / `tools/gluten-it/pom.xml`: `<spark.version>`
- `.github/workflows/util/install-spark-resources.sh`: tarball download URL
- `.github/workflows/velox_backend_x86.yml`: 2 step names
- `docs/get-started/{build-guide,getting-started}.md`: supported version list
- `shims/spark41/.../MemoryStream.scala`: follow [SPARK-55337](https://issues.apache.org/jira/browse/SPARK-55337), which reverted `MemoryStream.apply` implicit param from `SparkSession` back to `SQLContext` for binary compat. The shim no longer needs the `.sparkSession` conversion.

### Why are the changes needed?

Pick up the upstream Spark 4.1.2 patch release (86 bug fixes including the binary-compat fix above).

### Does this PR introduce _any_ user-facing change?

No (patch version bump).

### How was this patch tested?

Locally:
```
./build/mvn clean install -pl shims/spark41 -am -Pspark-4.1 -Pscala-2.13 -Pjava-17 -DskipTests
```
Relying on CI for the full Spark 4.1 test matrix.

Generated-by: claude-opus-4.7

PR #12180, opened 2026-05-29
[MINOR] Upgrade Spark 4.0 to 4.0.2

yaooqinn:

Bumps the `spark-4.0` profile from 4.0.1 to 4.0.2 (patch release).

Sibling of #12177 (Spark 4.1 → 4.1.2), kept as a separate PR per one-concern-per-PR.

### Files touched (6 / +7 −7)
- `pom.xml` / `tools/gluten-it/pom.xml` — `spark.version`
- `.github/workflows/util/install-spark-resources.sh` — download version
- `.github/workflows/velox_backend_x86.yml` — step names
- `docs/get-started/{build-guide,getting-started}.md` — supported versions

`delta.version` (also `4.0.1`, but for Delta Lake) is intentionally **not** touched.

### Why no shim code change
4.0.2 is a maintenance release. Unlike 4.1.2 (#12177), it does **not** revert any binary signatures the spark40 shim depends on — SPARK-55337 (MemoryStream binary-compat reversion) only lives on the 4.1.x branch.

### Upstream fixes worth watching CI for

| JIRA | Why it might matter to Gluten |
|---|---|
| SPARK-54439 | SPJ KeyGroupedPartitioning + join key size mismatch — may shift plan-stability files |
| SPARK-53434 | ColumnarRow#get should check isNullAt — touches columnar row path |
| SPARK-54917 | ORC bumped to 2.1.4 — Velox uses its own ORC reader; only affects Spark vectorized fallback |
| SPARK-54753 | persist/unpersist memory leak — observability for cache users |

No Gluten-side code change is needed for any of the above.

Relying on the existing Spark 4.0 CI matrix (gluten-ut/spark40 + velox_backend_x86 4.0 jobs).

Yes

Generated-by: claude-opus-4.7

### DIFF
--- a/.github/workflows/util/install-spark-resources.sh
+++ b/.github/workflows/util/install-spark-resources.sh
@@ -118,7 +118,7 @@ if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
   4.0)
       # Spark-4.0, scala 2.12 // using 2.12 as a hack as 4.0 does not have 2.13 suffix
       cd ${INSTALL_DIR} && \
-      install_spark "4.0.1" "3" "2.12"
+      install_spark "4.0.2" "3" "2.12"
       mv /opt/shims/spark40/spark_home/assembly/target/scala-2.12 /opt/shims/spark40/spark_home/assembly/target/scala-2.13
       ;;
   4.1)
--- a/.github/workflows/velox_backend_x86.yml
+++ b/.github/workflows/velox_backend_x86.yml
@@ -1307,7 +1307,7 @@ jobs:
           pip3 install setuptools==77.0.3 && \
           pip3 install pyspark==3.5.5 cython && \
           pip3 install pandas==2.2.3 pyarrow==20.0.0
-      - name: Prepare Spark Resources for Spark 4.0.1 #TODO remove after image update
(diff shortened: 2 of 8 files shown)


zhouyuan:
@yaooqinn the KeyGroupedPartitioningSuite is re-written in Gluten tests: 
https://github.com/apache/gluten/blob/main/gluten-ut/spark40/src/test/scala/org/apache/spark/sql/connector/GlutenKeyGroupedPartitioningSuite.scala#L1039

yaooqinn:
Thanks @zhouyuan, good catch. Done — instead of excluding `SPARK-54439`, I ported the two new tests as `testGluten` overrides in `GlutenKeyGroupedPartitioningSuite` (same pattern as the existing `SPARK-41471` ones), so they reuse the columnar-aware `collectShuffles` helper and we keep coverage of the correctness fix. Dropped the `.excludeByPrefix("SPARK-54439")` line.

Verified locally on the Velox backend (Spark 4.0.2 / Scala 2.13): both new tests pass (`shuffles.size == 1` and `checkAnswer`), and the set of pre-existing suite failures is unchanged.

zhouyuan (approved):

PR #12188, opened 2026-05-29
[GLUTEN-12187][VL] Port AttachDistributedSequenceExec to Velox backend

### LINKED ISSUE #12187: [VL] Port AttachDistributedSequenceExec to Velox backend
### Description

Spark's `AttachDistributedSequenceExec` prepends a contiguous, globally increasing `Long` id column to its child output. It is used by pandas-on-Spark's `distributed-sequence` default index and by `DataFrame.zipWithIndex`. Today Gluten falls back to vanilla Spark for this operator, which forces a columnar → row transition that dominates runtime for wide / nested-typed inputs.

This issue tracks porting the operator to the Velox backend so that it runs end-to-end on columnar batches.

baibaichen:

Closes #12187.

Adds a Velox implementation of Spark's `AttachDistributedSequenceExec` (prepends a contiguous `Long` id column to child output). Used by pandas-on-Spark distributed-sequence index and `DataFrame.zipWithIndex`.

### How is this implemented?

**Plan-level**

- New abstract base `ColumnarAttachDistributedSequenceBaseExec` in `gluten-substrait/` with factory `from(plan)` delegating to the backend API.
- New offload rule case + validator gate in `OffloadSingleNodeRules` / `Validators`.
- New backend hook `genColumnarAttachDistributedSequenceExec` on `SparkPlanExecApi`. Velox override returns the columnar impl; the CH override throws `GlutenNotSupportException` until that backend is ported.
- Config: `spark.gluten.sql.columnar.attachDistributedSequence` (default `true`) lets users disable the offload.

**Velox runtime (`ColumnarAttachDistributedSequenceExec`)**

For >1 partition:

1. Materialize the child output once via Gluten's existing `ColumnarCachedBatchSerializer`, persisted at `MEMORY_AND_DISK_SER`. The cache blob is Velox-native serialization (`CachedColumnarBatch`) — kryo-friendly and typically much more compact than unsafe-row SER.
2. **Count pass**: read `CachedColumnarBatch.numRows` of partitions `[0, numPartitions - 1)` — no native deserialization.
3. **Assign pass**: `convertCachedBatchToColumnarBatch` → Velox-native batch → `ColumnarBatches.load` (zero-copy Arrow C-Data ABI handoff) → prepend one `ArrowWritableColumnVector` with the id column.

Single-partition queries skip caching entirely (`startOffset = 0`).

**Memory hygiene**

- The base class exposes a `doColumnarCleanup()` hook called from `cleanupResources()`. The Velox impl uses it to `unpersist` the cached RDD when the query finishes, so BlockManager does not hold the serialized batches beyond the operator's lifetime.
- The persisted RDD is cached behind a `synchronized` accessor so repeated `doExecuteColumnar()` calls share a single `persist()` handle.
- `assignIds` wraps the per-batch build in a `try/catch` that closes the freshly-loaded heavy batch on failure, so a mid-build OOM (e.g. while allocating the id vector) cannot leak Arrow buffers.

### Known overhead: cache write/read crosses the heap boundary

Per batch, the cache path does two full data copies:

- **Write**: JNI `serialize` copies the off-heap Velox batch into an on-heap `Array[Byte]` (`CachedColumnarBatch.bytes`).
- **Read**: JNI `deserialize` copies the bytes back into a fresh off-heap Velox batch.

This is the price for `zipWithIndex`'s two-pass semantics (count + assign) without re-executing the child plan. The Velox↔Arrow hand-offs elsewhere in the operator are zero-copy ABI transfers and not relevant to this cost.

#### Alternative considered

We considered the row-mode pipeline `Velox → C2R → RDD[InternalRow] cached → R2C`. For Gluten that costs a full C2R/R2C transition on every row, and the unsafe-row serialized cache is typically 2–5× larger than Velox-native serialization for wide / nested data. The columnar path keeps everything columnar and pays serialization only once.

New `VeloxAttachDistributedSequenceExecSuite` in `backends-velox`.

Yes — a new config:

- `spark.gluten.sql.columnar.attachDistributedSequence` (default `true`).

When enabled, `df.zipWithIndex` and pandas-on-Spark `distributed-sequence` index materialize the id column columnarly on Velox instead of falling back.


baibaichen:
@zhztheplayer would you please review this PR?

gluten-substrait/src/main/scala/org/apache/spark/sql/execution/ColumnarAttachDistributedSequenceBaseExec.scala
```diff
@@ -0,0 +1,85 @@
...
+    val indexColumn = s"Index: $sequenceAttr"
+    s"$nodeName$truncatedOutputString $indexColumn"
+  }
+
+  /**
+   * Hook for backend implementations to release any resources (e.g. cached RDDs) that were
+   * materialized during execution. Called from [[cleanupResources]] after children have been
+   * cleaned up. The default implementation is a no-op.
+   */
+  protected def doColumnarCleanup(): Unit = {}
```
philo-he, line 65: Do we still need this? It seems to be useless.

backends-velox/src/test/scala/org/apache/gluten/execution/VeloxAttachDistributedSequenceExecSuite.scala
```diff
@@ -0,0 +1,127 @@
...
+class VeloxAttachDistributedSequenceExecSuite extends VeloxWholeStageTransformerSuite {
+
+  override protected val resourcePath: String = "/tpch-data-parquet"
+  override protected val fileFormat: String = "parquet"
+
+  override def sparkConf: SparkConf = {
+    super.sparkConf
+      .set("spark.sql.shuffle.partitions", "3")
+      .set("spark.default.parallelism", "3")
+      .set(SQLConf.ANSI_ENABLED.key, "false")
```
philo-he, line 39: Can we also set Spark master "local[3]" explicitly?

philo-he: It seems that gluten columnar shuffle is not set. Do we need to set it in the base class WholeStageTransformerSuite?

philo-he (commented):
Looks good overall. Could you check whether any of Spark's own related test suites are enabled in Gluten? Thanks.

PR #12197, opened 2026-05-31
[VL][Delta] Add DV scan info extraction utility

malinjawi:

This PR is the next split from the Delta deletion-vector (DV) scan stack, following the native reader support already merged in #12040 and before the full JVM scan handoff work from #12131.

It adds a focused Scala utility layer that extracts the essential DV scan information from Spark/Delta `PartitionedFile` metadata without changing scan offload behavior yet.

Main changes:

- add `DeltaDeletionVectorScanInfo` for Delta 3.3 and Delta 4.0 source sets
- extract per-file DV scan info from `PartitionedFile` metadata:
  - row-index filter type
  - deletion-vector descriptor and cardinality
  - serialized DV bitmap payload bytes
  - normalized non-DV metadata columns
- keep the utility independent from Substrait, Velox native split conversion, and scan offload behavior
- add focused Delta 3.3 and Delta 4.0 tests for DV extraction, keep-all/no-DV extraction, and invalid partial DV metadata

This PR is intentionally utility-only:

- no Substrait proto changes
- no native/C++ changes
- no Delta scan rule replacement
- no end-to-end scan offload behavior change yet

Those pieces stay in follow-up PRs after this API is reviewed.

How was this patch tested?

Validation run:

- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./build/mvn test-compile -pl backends-velox -am -Pjava-17,spark-3.5,backends-velox,hadoop-3.3,spark-ut,delta -DskipTests`
- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./build/mvn test-compile -pl backends-velox -am -Pjava-17,spark-4.0,scala-2.13,backends-velox,hadoop-3.3,spark-ut,delta -DskipTests`
- `git diff --check`

Also attempted the focused suite with `dev/run-scala-test.sh`, but the local runner failed during classpath compilation before executing the suite while switching profiles locally. The module-level Spark 3.5 and Spark 4.0 test-compile checks above pass.

Generated-by: IBM BOB

### DIFF
--- a/backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
+++ b/backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
@@ -41,8 +41,8 @@ object DeltaDeletionVectorScanInfo {
   import RowIndexFilterType._
 
   final case class DeletionVectorInfo(
-      rowIndexFilterType: RowIndexFilterType,
       hasDeletionVector: Boolean,
+      rowIndexFilterType: RowIndexFilterType,
       cardinality: Long,
       serializedDeletionVector: Array[Byte])
 
@@ -89,13 +89,13 @@ object DeltaDeletionVectorScanInfo {
 
     (descriptorValue, filterTypeValue) match {
       case (None, None) =>
-        DeletionVectorInfo(KEEP_ALL, false, 0L, Array.emptyByteArray)
+        DeletionVectorInfo(false, KEEP_ALL, 0L, Array.emptyByteArray)
       case (Some(encodedDescriptor), Some(filterType)) =>
         val descriptor = parseDescriptor(encodedDescriptor.toString)
         val serializedPayload = serializePayload(spark, partitionColumnCount, file, descriptor)
         DeletionVectorInfo(
-          parseRowIndexFilterType(filterType.toString),
           true,
+          parseRowIndexFilterType(filterType.toString),
           descriptor.cardinality,
           serializedPayload)
       case _ =>
--- a/backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
+++ b/backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
@@ -1,0 +1,208 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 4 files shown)


backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
```diff
@@ -0,0 +1,208 @@
...
+    type RowIndexFilterType = Value
+    val KEEP_ALL, IF_CONTAINED, IF_NOT_CONTAINED = Value
+  }
+
+  import RowIndexFilterType._
+
+  final case class DeletionVectorInfo(
+      rowIndexFilterType: RowIndexFilterType,
+      descriptor: Option[DeletionVectorDescriptor],
+      serializedDeletionVector: Array[Byte]) {
```
zhztheplayer: As we are passing a serialized deletion vector to C++, why do we need to preserve the original `DeletionVectorDescriptor` alongside?

zhztheplayer (commented):
Thanks for the work.

backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
```diff
@@ -0,0 +1,207 @@
...
+    val KEEP_ALL, IF_CONTAINED, IF_NOT_CONTAINED = Value
+  }
+
+  import RowIndexFilterType._
+
+  final case class DeletionVectorInfo(
+      rowIndexFilterType: RowIndexFilterType,
+      hasDeletionVector: Boolean,
+      cardinality: Long,
+      serializedDeletionVector: Array[Byte])
```
zhztheplayer, line 47: nit: can we reorder to place `hasDeletionVector` as the first field?

zhztheplayer (approved):
+1 % a minor nit.

backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
malinjawi: Good point @zhztheplayer , I removed `DeletionVectorDescriptor` from the exported scan info and only keep it locally while materializing the serialized DV payload.

backends-velox/src-delta33/main/scala/org/apache/gluten/delta/DeltaDeletionVectorScanInfo.scala
malinjawi: Thanks @zhztheplayer! Will reorder `hasDeletionVector` first in both Delta 3.3 and Delta 4.0.

PR #12198, opened 2026-05-31
[VL][Delta] Add JVM Delta DV scan handoff

malinjawi:

This PR is the next split in the Delta deletion-vector scan stack after #12197, which has now landed.

It adds the JVM/Substrait/Velox handoff that consumes the essential Delta DV scan info extracted by #12197, materializes serialized DV payloads on the JVM side, and passes them to native scan execution.

Main changes:

- add a Delta DV preprocessing rule for the Velox Delta component without replacing Delta's `PrepareDeltaScan`
- reuse `DeltaDeletionVectorScanInfo` from #12197 to extract per-file DV metadata and serialized DV bytes from Delta-prepared scan files
- add Delta local files Substrait nodes/builders carrying `DeltaReadOptions`
- embed the serialized DV payload in `DeltaReadOptions`, instead of passing essential DV data through generic metadata columns
- add a native `DeltaSplitInfo` path for Delta-specific split metadata
- wire the handoff through `VeloxIteratorApi`, `VeloxPlanConverter`, `WholeStageResultIterator`, and `SubstraitToVeloxPlan`
- strip Spark's synthetic DV predicate/internal columns only after the native split has the payload, so Velox applies the DV filter natively and avoids double filtering
- add Spark 3.5 and Spark 4.0 focused handoff coverage

This PR is intentionally handoff-only:

- #12197 reviewed and landed the DV scan info extraction utility
- performance/benchmark iteration remains a follow-up after the correctness handoff shape is reviewed
- DELETE/UPDATE/MERGE DV DML support remains in later split work

Issue: #11901

How was this patch tested?

Validation after rebasing on current `upstream/main` with #12197 included:

- `git diff --check upstream/main..HEAD`
- container `./dev/format-scala-code.sh check`
- container `clang-format --dry-run -Werror` on touched C++ files
- container `./build/mvn test-compile -pl backends-velox -am -Pjava-17,spark-3.5,backends-velox,hadoop-3.3,spark-ut,delta -DskipTests`
- container `./build/mvn test-compile -pl backends-velox -am -Pjava-17,spark-4.0,scala-2.13,backends-velox,hadoop-3.3,spark-ut,delta -DskipTests`

Generated-by: IBM BOB

### DIFF
--- a/gluten-delta/src/main/scala/org/apache/gluten/execution/DeltaScanTransformer.scala
+++ b/gluten-delta/src/main/scala/org/apache/gluten/execution/DeltaScanTransformer.scala
@@ -20,9 +20,10 @@ import org.apache.gluten.sql.shims.SparkShimLoader
 import org.apache.gluten.substrait.rel.LocalFilesNode.ReadFileFormat
 
 import org.apache.spark.sql.catalyst.TableIdentifier
-import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression}
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference, Expression}
 import org.apache.spark.sql.catalyst.plans.QueryPlan
 import org.apache.spark.sql.connector.read.streaming.SparkDataStream
+import org.apache.spark.sql.delta.{DeltaParquetFileFormat, NoMapping}
 import org.apache.spark.sql.execution.FileSourceScanExec
 import org.apache.spark.sql.execution.datasources.HadoopFsRelation
 import org.apache.spark.sql.types.StructType
@@ -54,6 +55,34 @@ case class DeltaScanTransformer(
   ) {
 
   override lazy val fileFormat: ReadFileFormat = ReadFileFormat.ParquetReadFormat
+
+  // For Delta column-mapping tables, `dataFilters` on the scan node are LOGICAL-named so Delta's
+  // file index (`PreparedDeltaFileIndex.matchingFiles`, `Snapshot.filesForScan`) can do partition
+  // pruning and stats-based file skipping -- both resolve filter attrs against logical schemas.
+  //
+  // The native (Velox) side, however, must see PHYSICAL names: `output` and `dataSchema` are
+  // physical (so the parquet reader finds the right column), and `BasicScanExecTransformer`
+  // matches `scanFilters` against `pushDownFilters` (built from a `Filter` that references the
+  // physical-named scan output) by `AttributeReference.equals`, which compares names. Without
+  // this override, the logical-named `scanFilters` and physical-named `pushDownFilters` would
+  // never match, causing duplicate filter evaluation in the substrait plan.
+  //
+  // Translate by exprId match against `output` rather than by re-running Delta's column-mapping
+  // helpers; exprIds are stable across the post-transform rewrite and don't require a second
+  // metadata lookup.
+  //
+  // See `DeltaPostTransformRules.transformColumnMappingPlan` for the full picture of which
+  // fields stay logical vs. become physical, and the longer-term cleanup direction (do all
+  // physical translation at substrait emission time so this override and the alias-back
+  // ProjectExec both go away).
+  override def scanFilters: Seq[Expression] = relation.fileFormat match {
+    case d: DeltaParquetFileFormat if d.columnMappingMode != NoMapping =>
+      val physicalByExprId = output.collect { case ar: AttributeReference => ar.exprId -> ar }.toMap
+      dataFilters.map(_.transformDown {
+        case ar: AttributeReference => physicalByExprId.getOrElse(ar.exprId, ar)
+      })
+    case _ => dataFilters
+  }
 
   override protected def doValidateInternal(): ValidationResult = {
     if (
--- a/gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaPostTransformRules.scala
+++ b/gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaPostTransformRules.scala
@@ -164,20 +164,55 @@ object DeltaPostTransformRules {
   }
 
   /**
-   * This method is only used for Delta ColumnMapping FileFormat(e.g. nameMapping and idMapping)
-   * transform the metadata of Delta into Parquet's, each plan should only be transformed once.
+   * Used for Delta ColumnMapping FileFormat (nameMapping and idMapping). Each plan is transformed
+   * at most once; the first run is tagged so re-runs are no-ops.
+   *
+   * Background: with column mapping, Delta files are written with PHYSICAL column names while
+   * Delta's metadata (partition schema, column stats) keeps LOGICAL names. Vanilla Spark + Delta
+   * resolves this asymmetry inside `DeltaParquetFileFormat.buildReaderWithPartitionValues`:
+   * everything on the scan node stays logical, and physical translation happens just-in-time when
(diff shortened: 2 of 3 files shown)


malinjawi:
Hey @zhztheplayer @zhouyuan this PR is ready for review. This should close the DV scan migration form the initial POC work.

backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
```diff
@@ -46,4 +53,22 @@ class VeloxDeltaComponent extends Component {
...
+  private def deltaDvPreprocessRule(spark: SparkSession): Rule[LogicalPlan] = {
+    if (!SparkReflectionUtil.isClassPresent(deltaDvPreprocessRuleClassName)) {
+      return VeloxDeltaComponent.IdentityRule
+    }
+
+    Class
+      .forName(deltaDvPreprocessRuleClassName)
+      .getConstructor(classOf[SparkSession])
+      .newInstance(spark)
+      .asInstanceOf[Rule[LogicalPlan]]
```
zhztheplayer: Is there any particular reason for using reflection here? Given it's against Gluten's own class?

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxIteratorApi.scala
```diff
@@ -179,6 +214,93 @@ class VeloxIteratorApi extends IteratorApi with Logging {
...
+      .asInstanceOf[Array[Byte]]
+
+    (
+      metadata,
+      new DeltaFileReadOptions(
+        toDeltaRowIndexFilterType(rowIndexFilterType),
+        hasDeletionVector,
+        cardinality,
+        serializedDeletionVector))
+  }
```
zhztheplayer: If reflection is inevitable, can we encapsulate the reflection code in a utility class?

zhztheplayer (commented):
Thanks. I have some initial broad questions about the use of reflections.

backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
malinjawi: Thanks @zhztheplayer you are right since this rule is Gluten-owned reflection was not needed here.

I updated this path to construct `PreprocessDeltaScanWithDeletionVectors` directly. To keep the direct construction compiling across the existing Spark/Delta profile matrix, I also added no-op compatibility definitions for Delta 2.3/2.4. The actual DV preprocessing behavior remains in the Delta 3.3/4.x profiles.

backends-velox/src/main/scala/org/apache/gluten/backendsapi/velox/VeloxIteratorApi.scala
malinjawi: @zhztheplayer  I moved this reflection out of `VeloxIteratorApi` into `DeltaSplitMetadataExtractor`.

Some reflection I think is still needed here because `VeloxIteratorApi` is in the common Velox source set, while `DeltaDeletionVectorScanInfo` is only compiled in Delta-specific source profiles. The extractor now isolates that bridge and caches the class/method lookups, so the iterator path only calls the utility and does not carry the reflection block inline.

Let me know if you know a workaround?

backends-velox/src-delta40/main/scala/org/apache/gluten/extension/PreprocessDeltaScanWithDeletionVectors.scala
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.gluten.extension
+
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.rules.Rule
+
+/**
+ * Delta 4 prepares DV scan metadata before Gluten offload, so this compatibility rule does not need
+ * to add backend-visible DV columns.
+ */
```
zhztheplayer: @malinjawi does this mean we need to add new columns to table scan for Delta 3.3? if yes, can you help me understand the design here?

malinjawi: > @malinjawi does this mean we need to add new columns to table scan for Delta 3.3? if yes, can you help me understand the design here?

@zhztheplayer 

No, we do NOT add new columns to the table scan output for Delta 3.3. The key insight is we're passing DV data as **file-level metadata**, not as scan output columns.
 The design works differently"

**What Delta 3.3's `PreprocessTableWithDVs` does:**
- Prepares Delta's internal scan metadata structures
- Makes DV information accessible via Delta's file APIs
- Does NOT inject `__delta_internal_*` columns into the scan output

**What our code does:**
- Extracts DV metadata from Delta's prepared file structures
- Passes it to Velox as **per-file metadata** through `DeltaLocalFilesNode` → Substrait protobuf
- Velox receives: `serializedDeletionVector`, `cardinality`, `rowIndexFilterType` for each file
- No columns added to scan output - tests verify the plan contains no `__delta_internal_*` columns

**Delta 4.0 difference:**
- Delta 4.0 prepares this metadata earlier in its own pipeline
- We just extract and forward it the same way
- The no-op rule exists because Delta 4.0 doesn't need the explicit `preprocessTablesWithDVs()` call

zhztheplayer: @malinjawi thanks for explanation. I may be misunderstanding but `PreprocessTableWithDVs` is shadowing a Delta class, we should be extremely carefully when doing so. What's the main difference for this class, compared to the one from vanilla Delta?

malinjawi: @zhztheplayer Good catch on the shadowing concern.

 The shadowing `PreprocessTableWithDVs` is necessary because:

1. **Mixed format handling**: Delta's optimizer creates scans with `DeltaParquetFileFormat`, but Gluten's write path produces tables with `GlutenDeltaParquetFileFormat`. The DV preprocessing rule must handle BOTH formats.

2. **Package-private API access**: We need direct access to Delta's package-private APIs (`TahoeFileIndex`, `deletionVectorsReadable()`, `DeltaSQLConf`, etc.) without reflection overhead. Being in `org.apache.spark.sql.delta` package provides this.

We could eliminate shadowing by making `GlutenDeltaParquetFileFormat` extend `DeltaParquetFileFormat` instead of `GlutenParquetFileFormat`, then override only write-specific methods. This would let Delta's original `PreprocessTableWithDVs` handle both formats naturally.

But the tradoeff would be this breaks the clean format hierarchy separation:
- Current: `GlutenParquetFileFormat` → `GlutenDeltaParquetFileFormat` (Gluten hierarchy)
- Alternative: `DeltaParquetFileFormat` → `GlutenDeltaParquetFileFormat` (mixed hierarchy)

Right now I would say the code maintains format hierarchy separation, avoids reflection for package-private API access, and type-safe handling of both formats in DV preprocessing. But this Requires shadowing Delta's trait (maintenance burden if Delta changes internals)

Would you prefer we explore the alternative approach, or is the current shadowing acceptable given the current state?

PR #12214, opened 2026-06-01
[VL][Delta] Add native Delta bitmap aggregation support

malinjawi:

Deleting rows from a Delta table with deletion vectors means building (and merging) a roaring bitmap of deleted row indices. Today that happens on the JVM. This adds the native Velox building block for it, so a later PR can run DELETE's bitmap construction natively.

It's just the primitive — nothing is wired into DELETE yet:

- `RoaringBitmapArray` handles Delta's Portable-format DV payloads, with bounded deserialization (CRoaring portable sizing before `readSafe`)
- a native `bitmapaggregator` aggregate for row-index aggregation, registered and wired through Gluten's expression/substrait planning
- native tests for serialize/deserialize + aggregate behavior, and a `delta_bitmap_benchmark`

Explicitly out of scope (follow-up PRs): DELETE command routing, DML row-index scan planning, and enabling this aggregate as the default DELETE path.

## How was this patch tested?

Rebased onto current `main` (after the DV scan handoff merged in #12269) and re-checked the JVM wiring with JDK 17 (`mvn compile -pl gluten-substrait -am`, `spotless:check`, `git diff --check`). The native build/tests run in CI — that's the authoritative lane here, since local C++ validation is limited by the Velox build tree.

Native evidence from the standalone run (C++ is unchanged by the rebase):

- `RoaringBitmapArrayTest` passes; round-trips a payload through both native and a Delta 3.3.2 JVM helper (values `1`, `7`, `1 << 33`; cardinality `3`)
- `delta_bitmap_benchmark`: 1M build+serialize ~8–10 ms (~105–132M rows/s), merge of 64 partials ~1.1 ms, deserialize+probe ~487 µs

Generated-by: IBM BOB


gluten-substrait/src/main/scala/org/apache/gluten/expression/AggregateFunctionsBuilder.scala
```diff
@@ -62,4 +66,14 @@ object AggregateFunctionsBuilder {
...
+
+  private def isPortableDeltaBitmapAggregator(aggregateFunc: AggregateFunction): Boolean = {
+    aggregateFunc.prettyName == ExpressionNames.BITMAP_AGGREGATOR &&
+    Try {
+      aggregateFunc.getClass
+        .getMethod("serializationFormatString")
+        .invoke(aggregateFunc)
+        .asInstanceOf[String]
+    }.toOption.contains("Portable")
+  }
```
zhztheplayer: Would you explain a bit more about the method? Why reflection is needed, etc.

cpp/velox/operators/functions/delta/DeltaBitmapAggregator.h
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <string>
+
+namespace gluten {
+
+void registerDeltaBitmapAggregator(
```
zhztheplayer, line 24: Can we add a new source folder `operators/functions/delta`, for all Delta functions?

zhztheplayer (commented):
LGTM overall. Thanks.

gluten-substrait/src/main/scala/org/apache/gluten/expression/AggregateFunctionsBuilder.scala
malinjawi: Yeah, dug into this and the reflection wasn't actually buying us anything, so I dropped it it's just a prettyName match now (cb2f28772).

The Portable check was me being paranoid. Delta only ever builds this aggregator in one spot (DMLWithDeletionVectorsHelper) and always with the Portable format, so it never actually filtered anything. And if a weird non-Portable payload ever did show up, the native deserialize already throws on the magic number so the real check is on the C++ side anyway. Bonus: gluten-substrait no longer needs to know anything about delta-spark.

cpp/velox/operators/functions/delta/DeltaBitmapAggregator.h
malinjawi: Done, moved them into operators/functions/delta/ and fixed up the include + test + CMake (cb2f28772). Future Delta functions can land there too.

cpp/velox/operators/functions/delta/DeltaBitmapAggregator.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  exec::registerAggregateFunction(
+      prefix + "bitmapaggregator",
+      std::move(signatures),
+      [](core::AggregationNode::Step step,
+         const std::vector<TypePtr>& argTypes,
+         const TypePtr& resultType,
+         const core::QueryConfig& /* config */) -> std::unique_ptr<exec::Aggregate> {
+        VELOX_CHECK_EQ(argTypes.size(), 1, "bitmapaggregator takes one argument");
+        VELOX_CHECK_EQ(argTypes[0]->kind(), exec::isRawInput(step) ? TypeKind::BIGINT : TypeKind::VARBINARY);
```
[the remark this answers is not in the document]
malinjawi: fixed in 302653952

zhztheplayer (approved):
👍

felipepessoto:
Contributes to #11901

PR #12215, opened 2026-06-01
[VL][Delta] Add an opt-out for the native Delta DV DML row-index scan

malinjawi:

Delta's DELETE/UPDATE/MERGE target scan reads file paths and row indexes so the command can
build updated deletion-vector bitmaps. That scan already offloads to Velox today, and there
is currently no way to keep just it on Spark.

This adds `spark.gluten.sql.columnar.backend.velox.delta.enableNativeDmlRowIndexScan`
(default `true`, so behaviour is unchanged) as a scoped escape hatch. With it off, only the
DV-writing DML target scan falls back; every other Delta scan keeps offloading.

The check lives in one place, `OffloadDeltaScan`, and is decided from the scan alone: Delta
builds every DML target relation over a `TahoeBatchFileIndex` carrying the command name,
which survives AQE stage splits and arbitrary join placement, and only DV-writing DML reads
a row-index column from that relation. DML that rewrites whole files does not, so it stays
eligible for native execution.

Also included:

- `DeltaScanUtils` — pulls the shared `isDeltaScan` structural check out of
  `OffloadDeltaScan` so the offload rules can share it.
- `DeltaPostTransformRules` — `referencesDeletionVectorRowIndex` now matches all three
  row-index column spellings (`__delta_internal_row_index`,
  `DeltaParquetFileFormat.ROW_INDEX_COLUMN_NAME`, `_tmp_metadata_row_index`) instead of
  only the first.
- `DeltaDeletionVectorHandoffSuite` (Spark 3.3/3.5 and 4.0) grows from 1 test to 4, covering
  the native path: DELETE in both metadata-row-index modes, repeated DELETE over an existing
  DV (leading rows first, so a scan that renumbered row indexes after applying the DV would
  fail the checks), MERGE with broadcast and shuffle joins, and the fallback config.
- Docs for the new config.

A switch for the whole DV scan feature will come in a separate PR, as agreed in review.

Issue: #11901

## How was this patch tested?

With JDK 17: `test-compile` for Spark 3.3, 3.4, 3.5, 4.0 and 4.1 with the Velox and Delta
profiles, Scala formatting, license headers, `git diff --check`. The handoff suites run in
CI and check both plan shape and results (row content and DV cardinality).

Generated-by: IBM BOB


malinjawi:
cc: @felipepessoto

backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
```diff
@@ -40,9 +40,22 @@ class VeloxDeltaComponent extends Component {
...
+    // For native DELETE/UPDATE/MERGE, the DML target row-index scan is deliberately kept on Spark
+    // until native row-index execution is proven; tag those scans here so the post-transform rules
+    // can keep the small subtree off the native path.
+    legacy.injectPreTransform(_ => DeltaDeletionVectorDmlUtils.tagDmlRowIndexScans)
     legacy.injectTransform {
       c =>
-        val offload = Seq(OffloadDeltaScan(), OffloadDeltaProject(), OffloadDeltaFilter())
+        val offload = Seq(
+          OffloadDeltaScan(
+            enableNativeDeltaWriteKey = VeloxDeltaConfig.ENABLE_NATIVE_WRITE.key,
```
zhztheplayer: Is there any particular reason we should check `VeloxDeltaConfig.ENABLE_NATIVE_WRITE.key` here?

gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaPostTransformRules.scala
```diff
@@ -90,6 +99,25 @@ object DeltaPostTransformRules {
         child.copy(output = p.output)
     }
 
+  /**
+   * Native DELETE/UPDATE/MERGE DV support can deliberately keep the target row-index scan in Spark.
+   * Keep only the contiguous scan-adjacent filter/project chain in Spark, avoiding an isolated
+   * native island without propagating fallback through joins, exchanges, or aggregations. The
+   * BitmapAggregator and the rest of the DML plan remain eligible for native execution.
+   */
+  val keepDmlRowIndexFallbackSubtreeOnSpark: Rule[SparkPlan] = (plan: SparkPlan) =>
```
felipepessoto: Do we need the same `if (!plan.exists(_.isInstanceOf[DeltaScanTransformer]))` that `deltaSpecificRules` uses?

backends-velox/src-delta/main/scala/org/apache/gluten/config/VeloxDeltaConfig.scala
```diff
@@ -22,6 +22,8 @@ class VeloxDeltaConfig(conf: SQLConf) extends GlutenCoreConfig(conf) {
   import VeloxDeltaConfig._
 
   def enableNativeWrite: Boolean = getConf(ENABLE_NATIVE_WRITE)
+
+  def enableNativeDmlRowIndexScan: Boolean = getConf(ENABLE_NATIVE_DML_ROW_INDEX_SCAN)
```
felipepessoto, line 26: Is it the convention to have a function for all configs even if not used?

felipepessoto (approved):
LGTM overall, just left a few non-blocking comments.
Also, PR description says the native target scan needs both `enableNativeWrite` and `enableNativeDmlRowIndexScan` , but after decoupling (per @zhztheplayer's comment) the code now gates only on enableNativeDmlRowIndexScan . Could you update the description?

backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
malinjawi: @zhztheplayer  Nope this is needless. The scan gating now only reads `ENABLE_NATIVE_DML_ROW_INDEX_SCAN;` enableNativeWrite stays write-only.

backends-velox/src-delta/main/scala/org/apache/gluten/config/VeloxDeltaConfig.scala
malinjawi: You're right @felipepessoto  it's not the convention since accessors are only added when some code actually calls one (like `enableNativeWrite`, used by `OffloadDeltaCommand`). 

`OffloadDeltaScan` reads the config by its key string instead, because gluten-delta can't depend on `VeloxDeltaConfig` directly; so I removed the unused accessor.

gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaPostTransformRules.scala
malinjawi: @felipepessoto I think That check would disable this rule exactly when it's needed as it fires when the DML target scan did not offload, so the plan has a `FileSourceScanExec` with a fallback tag and possibly no `DeltaScanTransformer` at all. 
So I added the equivalent short-circuit in the latest commit instead such that the rule exits early unless the plan contains a tagged DML row-index fallback scan for the same condition it rewrites on.

felipepessoto (approved):

gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaDeletionVectorDmlUtils.scala
```diff
@@ -0,0 +1,166 @@
...
+    plan
+  }
+
+  def copyDmlRowIndexScanTag(from: SparkPlan, to: SparkPlan): Unit = {
+    if (from.getTagValue(DmlRowIndexScanTag).contains(true)) {
+      to.setTagValue(DmlRowIndexScanTag, true)
+    }
+  }
+
+  def isDeltaScan(scan: FileSourceScanExec): Boolean = {
```
zhztheplayer: Any code duplication with the same APIs in OffloadDeltaScan? Should we reorganize them?

malinjawi: @zhztheplayer 

No duplicated implementation `OffloadDeltaScan.isDeltaScan` was a one-line delegate left behind when this PR moved its old private helpers into the utils object so the DML shape check could reuse them. 

But agreed the organization was off for generic scan detection didn't belong in a DV-DML utils object. Reorganized now so the shared checks now live in `DeltaScanUtils`, `OffloadDeltaScan` calls it directly, and `DeltaDeletionVectorDmlUtils` keeps only the DML tagging and shape logic.

felipepessoto:
@malinjawi could you cherry-pick this? https://github.com/apache/gluten/pull/12601/changes/06e772eaf6f9ab83ade65c2f9a37d89fa77adb46

It removes the test from flaky list

malinjawi:
> @malinjawi could you cherry-pick this? [06e772e](https://github.com/apache/gluten/commit/06e772eaf6f9ab83ade65c2f9a37d89fa77adb46)
> 
> It removes the test from flaky list


Thanks @felipepessoto  cherry-picked now.

gluten-delta/src/main/scala/org/apache/gluten/extension/OffloadDeltaScan.scala
```diff
@@ -20,15 +20,16 @@ import org.apache.gluten.execution.DeltaScanTransformer
...
 import org.apache.spark.sql.delta.SnapshotDescriptor
 import org.apache.spark.sql.delta.commands.DeletionVectorUtils.deletionVectorsReadable
 import org.apache.spark.sql.delta.files.TahoeFileIndex
 import org.apache.spark.sql.delta.stats.PreparedDeltaFileIndex
 import org.apache.spark.sql.execution.{FileSourceScanExec, SparkPlan}
 import org.apache.spark.util.SparkVersionUtil
 
-case class OffloadDeltaScan() extends OffloadSingleNode {
+case class OffloadDeltaScan(
+    enableNativeDeletionVectorDmlRowIndexScanKey: String)
```
zhztheplayer: nit: Can we just pass a boolean flag?

zhztheplayer:
@malinjawi 

Thanks for the PR. I went through https://github.com/apache/gluten/issues/12377 and was trying to understand why row ID 9223372036854775807 was written / read. Have you guys got any clue? cc @felipepessoto

felipepessoto:
@zhztheplayer I couldn't reproduce it locally — including the real `*MergeIntoExtendedSyntax*DVs*`
suites from Delta v4.2.0 on Spark 4.1 (200 tests, incl. `...DVsPredPushOnSuite`). The native
`_metadata.row_index` values always matched vanilla exactly.

That said, I don't think `9223372036854775807` was ever written or read as a row ID. Delta's own
`RoaringBitmapArray.add` enforces the identical bound (`MAX_REPRESENTABLE_VALUE =
9223372030412324864`), so vanilla Delta would abort the same way if that value ever reached the
aggregator. And the sibling signature we quarantined for the same root cause is
`Delta bitmap row index cannot be negative: -6254810385378525259` — arbitrary, not a sentinel.

So it looks like garbage reaching the aggregator's input column, not a real row index. Since it
only shows up under the full Delta suite, I opened #12783 to attach the offending column
(encoding, nulls, value distribution) to the error, which should tell us whether it's a real
row-index column with one corrupt slot or the wrong column entirely.

felipepessoto:
@zhztheplayer I tried to repro the issue several times and I couldn't. Maybe something else have fixed it.

@malinjawi the current CI failures are not related to your change:
https://github.com/apache/gluten/issues/12779
https://github.com/apache/gluten/pull/12780

gluten-delta/src/main/scala/org/apache/gluten/extension/OffloadDeltaScan.scala
malinjawi: Done in b55362e58 now a boolean. The rule is built per query from ColumnarRuleCall's live conf, so session-level toggling still works.

malinjawi:
> @zhztheplayer I couldn't reproduce it locally — including the real `*MergeIntoExtendedSyntax*DVs*` suites from Delta v4.2.0 on Spark 4.1 (200 tests, incl. `...DVsPredPushOnSuite`). The native `_metadata.row_index` values always matched vanilla exactly.
> 
> That said, I don't think `9223372036854775807` was ever written or read as a row ID. Delta's own `RoaringBitmapArray.add` enforces the identical bound (`MAX_REPRESENTABLE_VALUE = 9223372030412324864`), so vanilla Delta would abort the same way if that value ever reached the aggregator. And the sibling signature we quarantined for the same root cause is `Delta bitmap row index cannot be negative: -6254810385378525259` — arbitrary, not a sentinel.
> 
> So it looks like garbage reaching the aggregator's input column, not a real row index. Since it only shows up under the full Delta suite, I opened #12783 to attach the offending column (encoding, nulls, value distribution) to the error, which should tell us whether it's a real row-index column with one corrupt slot or the wrong column entirely.

Agree with @felipepessoto everything we found also says garbage in the row-index column, not a real row ID. One upstream data point: [velox#9943](https://github.com/facebookincubator/velox/issues/9943) documents the row-index vector getting only outputRows().size() slots written, with the tail serving recycled buffer contents as valid non-null BIGINTs  same huge positive/negative value shapes as our two signatures. Not yet confirmed live in our pinned Velox, but the mechanism and the plan-shape-dependent intermittency line up. Should pair well with the diagnostics in [#12783](https://github.com/apache/incubator-gluten/issues/12783).

malinjawi:
> @zhztheplayer I tried to repro the issue several times and I couldn't. Maybe something else have fixed it.
> 
> @malinjawi the current CI failures are not related to your change: #12779 #12780

Thanks @felipepessoto I'd added the same 4 entries on this branch (f8b4b5a4c) to unblock CI before seeing [#12779](https://github.com/apache/gluten/issues/12779)/[#12780](https://github.com/apache/gluten/issues/12780). Same lines, so they'll merge cleanly whichever lands first; [#12779](https://github.com/apache/gluten/issues/12779) is clearly the right attribution rather than my commit message's flakiness guess.

felipepessoto:
This is the most recent CI failure I found, from Aug 14:

https://github.com/apache/gluten/actions/runs/31773203848/job/94686948558

- Job  delta-spark-test (7) , run on commit  cd22fb88f  (Velox  dft-2026_08_12 )
- Suite  MergeIntoSQLSQLPathBasedDVsPredPushOffSuite , test "basic case - column pruning, isPartitioned: true"
- Delta bitmap row index cannot be negative: -2569406872788402175 
- Raised at  DeltaBitmapAggregator.cc:44 , at  06:15:01Z

felipepessoto:
Answering the question about row ID `9223372036854775807`: it was never written. It is not a row index at all, and the defect is upstream of the deletion-vector aggregator.

**The chain, and how each step was checked.**

The scan emits a batch whose row-index child is empty. With Velox's operator output validation turned on:

```
Output validation failed for [operator: TableScan, plan node ID: 0]:
  Child vector has size 0 less than parent and parent has no nulls 1.
```

Downstream, that empty child gets wrapped in a dictionary of size 1, so the aggregator's input is a dictionary addressing row 0 of a base that has no rows:

```
Delta bitmap aggregator input column:
  type=BIGINT encoding=DICTIONARY size=1 selected=1/1 decodedBase=FLAT
  baseSize=0 valuesBytes=0 valuesCapacityRows=0
  badValues=[(row=0 baseIndex=0 value=-5163232936757035008
              hex=0xb858830000000000 neighbours=<empty values buffer>)]
```

Reading it reads past the end of a zero-length buffer, so the value is whatever heap memory is there. That is why it differs on every occurrence. `BaseVector::wrapInDictionary` does not validate indexes against the base's size, so the malformed vector is built silently and only misbehaves when read.

**This now reproduces deterministically**, which it did not before:

```
spark/testOnly *MergeIntoExtendedSyntaxSQLPathBasedDVsPredPushOnSuite \
  -- -z "only update - isPartitioned: true"
```

with `debug.validate_output_from_operators=true` in the Velox query config. It fails on every run (~65 s, single fork). The same test on the non-DV suite (`MergeIntoExtendedSyntaxSQLPathBasedSuite`) passes with zero validation failures, so this is specific to the path that requests `_metadata.row_index`, not general scan noise.

**Why it looked so rare.** The bounds check only rejects negatives and values above `kMaxRepresentableValue`. Linux user-space pointers are around 1.4e14 -- positive, and well inside Delta's range -- so they pass. Instrumenting the malformed read instead of the value, a single test pass showed 176 out-of-bounds reads: 144 returned pointer-magnitude values, 27 returned `0`, 18 returned small values, and all were accepted. The abort we have been chasing is the rare tail.

That means a garbage row index is usually added to the deletion vector rather than raising. Pointer-magnitude values address rows that do not exist; a `0` marks row 0 of that file deleted. The suites do pass at baseline, so something is evidently limiting the impact and I have not worked out what -- but I would not assume this is only a flaky crash.

The fix does not belong in the aggregator; it is the first reader of a column that was already malformed when the scan produced it. #12783 has the diagnostic that produced the output above.

zhztheplayer:
@malinjawi @felipepessoto Thank you for the contexts and investigation. 

If there is something else who caused the abnormal row ID, should we fix the bug first and hold this PR off? cc @malinjawi

felipepessoto:
Following up on which stage builds the dictionary over an empty base: it is not Delta, and not the aggregator. It is an upstream Velox bug in the scan.

`SelectiveStructColumnReaderBase::next()` has a branch for structs with **no child readers** — a scan that projects only columns which are not read from the file. There it clears `outputRows_` and repopulates it only when there is a deletion, then sizes the result vector to `numValues` but sizes the synthesized fields to `outputRows()`:

```cpp
  outputRows_.clear();
  if (hasDeletion_) {
    fillOutputRowsFromMutation(numValues);
    numValues = outputRows_.size();
  }
  ...
  resultRowVector->unsafeResize(numValues);                                  // parent <- numValues
  setRowNumberField(currentRowNumber_, outputRows(), ..., childField);       // child  <- outputRows().size()
```

and

```cpp
  bool useOutputRows() const { return scanSpec_->hasFilter() || hasDeletion(); }
```

So a scan with a **filter but no deletion** gets the empty `outputRows_`, and `_metadata.row_index` comes back with zero rows while the RowVector containing it reports N rows. `BaseVector::wrapInDictionary` does not validate indexes against base size, so that empty child is then wrapped in a dictionary of size N, and the aggregator reads past the end of a zero-length buffer — which is where the garbage row IDs come from.

The Delta DV write path hits all three conditions at once: it projects only `_metadata.row_index`, `file_path` and a partition key, and predicate pushdown puts a filter on the partition key. That gives a clean discriminator — same test, same DV setup, only pushdown differs:

| variant | `TableScan[0]` output-validation failures | result |
|---|---|---|
| `...DVsPredPushOnSuite` | 9 | FAILED |
| `...DVsPredPushOffSuite` | 0 | passed |

Upstream:

* Issue: https://github.com/facebookincubator/velox/issues/18535
* Fix: https://github.com/facebookincubator/velox/pull/18536

With the fix applied to the Velox revision we pin, the deterministic reproduction goes from 9 validation failures and a failed test to 0 and a pass, and the full Delta DV suite run with `debug.validate_output_from_operators` enabled reports 0 validation failures with results identical to the existing baseline.

Two side notes that may be useful:

1. This explains why bumping or reverting Velox never changed the CI outcome. `SelectiveStructColumnReader.cpp` and `SelectiveColumnReader.h` are unchanged between `dft-2026_08_05` and `dft-2026_08_14`; the only commit touching them in that range is an unrelated DWRF flat-map fix. The defect was never in the code being bumped.
2. We do not have to wait for a Velox bump to unblock CI — `ep/build-velox/src/get-velox.sh` already supports `UPSTREAM_VELOX_PR_ID`, which applies a patch directly from a `facebookincubator/velox` PR.

malinjawi:
> @malinjawi @felipepessoto Thank you for the contexts and investigation.
> 
> If there is something else who caused the abnormal row ID, should we fix the bug first and hold this PR off? cc @malinjawi

@zhztheplayer With @felipepessoto  [velox#18536](https://github.com/facebookincubator/velox/pull/18536) the root cause is now found and fixed at the right layer, so I don't think holding this PR buys anything as the guard is what keeps the affected path off by default until that fix is actually in our pin and proven.

I would suggest we merge this (safe default + gate coverage), pull [velox#18536](https://github.com/facebookincubator/velox/pull/18536) in the next bump, then a small follow-up flips `enableNativeDmlRowIndexScan` default once the DV suites pass with output validation on keeping the config as an escape hatch. 

 Huge thanks @felipepessoto for the fix.

zhztheplayer:
@malinjawi Thanks. 

I am more or less inclined to hold this off unless such issue still exist after the fixes are landed. This PR can be a bit costly to disable a particular type of scan.

It would be simpler to add an option to enable / disable the entire DV scan feature if you think it's feasible. I assume we are now missing that option in the codebase.

felipepessoto:
> > @zhztheplayer I tried to repro the issue several times and I couldn't. Maybe something else have fixed it.
> > @malinjawi the current CI failures are not related to your change: #12779 #12780
> 
> Thanks @felipepessoto I'd added the same 4 entries on this branch ([f8b4b5a](https://github.com/apache/gluten/commit/f8b4b5a4c791f61850f76aec8f26ea0524ae2c29)) to unblock CI before seeing [#12779](https://github.com/apache/gluten/issues/12779)/[#12780](https://github.com/apache/gluten/issues/12780). Same lines, so they'll merge cleanly whichever lands first; [#12779](https://github.com/apache/gluten/issues/12779) is clearly the right attribution rather than my commit message's flakiness guess.

@malinjawi the tests are not failing anymore, fixed by #12051. Now we have a different problem, that PR also fixed other 4 tests that we need to remove known issues list: https://github.com/apache/gluten/pull/12815

felipepessoto:
https://github.com/facebookincubator/velox/pull/18536 is merged

Does anyone know in which version of daily Velox it will be included? 2026-08-20?
The Daily update PRs mention an IBM fork, so I'm not sure if there is a delay

felipepessoto:
https://github.com/apache/gluten/pull/12846 with updated Velox is merged

malinjawi:
> #12846 with updated Velox is merged

@felipepessoto thanks for tracking all this down  the scan reader root cause, https://github.com/facebookincubator/velox/pull/18536, and the #12815 / #12051 attribution. Saved me a lot of guessing.

@zhztheplayer https://github.com/facebookincubator/velox/pull/18536 is in main now via [#12846](https://github.com/apache/incubator-gluten/issues/12846), so we can just test this instead of arguing it. I haven't run the DV suites with the flag on yet so I can't say it's clean. Let me flip the default on a scratch commit and run them with Velox output validation on.

If nothing reproduces I'll delete the scan-shape detection and this PR becomes just the config + gate coverage, default on. You're right that disabling one scan type is a lot of machinery if the bug is gone.

On the feature-level switch — agreed it's missing and it's the better knob either way. I'd do it as a follow-up so this PR stays one thing, but if you want it here I'll put it here and drop the scan-specific guard in the same change.

felipepessoto:
@malinjawi FYI, I have tested after the daily velox update and debug.validate_output_from_operators enabled here: https://github.com/apache/gluten/pull/12783#issuecomment-5377218836

It doesn't show the CIs anymore, I guess this was the run: https://github.com/apache/gluten/actions/runs/32538314861

zhztheplayer:
Hi @malinjawi @felipepessoto, thanks for keeping troubleshooting and for iterating the patch. 

> If nothing reproduces I'll delete the scan-shape detection and this PR becomes just the config + gate coverage, default on. You're right that disabling one scan type is a lot of machinery if the bug is gone.

Yes. Regarding the change itself, what concerned me a little bit is the new tag introduced. Generally we need to minimize the number of cross-rule tags to limit the complexity of the query planner. So I would suggest to check whether the conditioned-fallback code can be put in one single place rather than tagging, if this PR needs to be continued further.

> On the feature-level switch — agreed it's missing and it's the better knob either way. I'd do it as a follow-up so this PR stays one thing, but if you want it here I'll put it here and drop the scan-specific guard in the same change.

Yes, we can open another PR for that. Thanks.

malinjawi:
> Hi @malinjawi @felipepessoto, thanks for keeping troubleshooting and for iterating the patch.
> 
> > If nothing reproduces I'll delete the scan-shape detection and this PR becomes just the config + gate coverage, default on. You're right that disabling one scan type is a lot of machinery if the bug is gone.
> 
> Yes. Regarding the change itself, what concerned me a little bit is the new tag introduced. Generally we need to minimize the number of cross-rule tags to limit the complexity of the query planner. So I would suggest to check whether the conditioned-fallback code can be put in one single place rather than tagging, if this PR needs to be continued further.
> 
> > On the feature-level switch — agreed it's missing and it's the better knob either way. I'd do it as a follow-up so this PR stays one thing, but if you want it here I'll put it here and drop the scan-specific guard in the same change.
> 
> Yes, we can open another PR for that. Thanks.



@zhztheplayer done, this revision follows that exactly. The tag, the tagging rule and the post-transform subtree rule are gone (DeltaDeletionVectorDmlUtils deleted). The fallback check now lives in one place, OffloadDeltaScan, decided from the scan alone: the DML target's TahoeBatchFileIndex carries the command name and survives AQE stage splits, and only DV-writing DML reads a row-index column from it. enableNativeDmlRowIndexScan now defaults to true, and disabling it keeps only that one scan on Spark. On the feature-level DV switch: agreed, I'll open a separate PR after this lands.


> @malinjawi FYI, I have tested after the daily velox update and debug.validate_output_from_operators enabled here: [#12783 (comment)](https://github.com/apache/gluten/pull/12783#issuecomment-5377218836)
> 
> It doesn't show the CIs anymore, I guess this was the run: https://github.com/apache/gluten/actions/runs/32538314861

@felipepessoto thanks for the verification. For the record, run 32538314861 is red only on FAIL_ON_FIXED bookkeeping (two decimal type-widening tests that now pass after the Velox update) — 0 output-validation failures, 0 regressions, 0 quarantined DV errors across the suite. Since main has no guard, that run exercised the native DML target scan directly, which is what backs flipping the default here.

One test note: the repeated-DELETE case now deletes leading rows first, so a scan that renumbered row indexes after applying an existing DV (the [#12377](https://github.com/apache/gluten/issues/12377) failure mode) would fail the cardinality and result checks instead of passing by luck. PTAL.

zhztheplayer (approved):

felipepessoto:
@malinjawi the title and description seems like it is describing the changes made to the PR since it was created. But for one looking the before and after the PR it might be misleading. Should it be describing the current changes only?

malinjawi:
> @malinjawi the title and description seems like it is describing the changes made to the PR since it was created. But for one looking the before and after the PR it might be misleading. Should it be describing the current changes only?

@felipepessoto Yes, it should describe the current changes only. Fixed the new title and body describe before and after the PR.

gluten-delta/src/main/scala/org/apache/gluten/extension/OffloadDeltaScan.scala
```diff
@@ -20,47 +20,80 @@ import org.apache.gluten.execution.DeltaScanTransformer
...
+        struct.fields.flatMap(field => field.name +: nestedFieldNames(field.dataType)).toSeq
+      case _ => Seq.empty
+    }
+
+    val outputColumnNames =
+      scan.output.flatMap(attribute => attribute.name +: nestedFieldNames(attribute.dataType))
+    val requiredColumnNames = scan.requiredSchema.fields.flatMap {
+      field => field.name +: nestedFieldNames(field.dataType)
+    }
+    (outputColumnNames ++ requiredColumnNames).exists(rowIndexColumnNames.contains)
```
felipepessoto: Non-blocking: `scanReadsRowIndexColumn` recursively flattens every struct and treats any field named `row_index` as Delta metadata. Could we narrow that match to the actual `_metadata.row_index` field? A regular Delta table may have a top-level or nested user column with that name, and non-DV DELETE/UPDATE paths still use a `TahoeBatchFileIndex` with action type `delete`/`update`. With this config disabled, those ordinary scans would therefore fall back to Spark even though the config promises that other scans are unaffected.

For example, the check could retain exact matching for generated top-level columns while inspecting `row_index` only under `_metadata`:

```scala
private val generatedRowIndexColumnNames = Set(
  "__delta_internal_row_index",
  DeltaParquetFileFormat.ROW_INDEX_COLUMN_NAME,
  parquetTemporaryRowIndexColumnName)

private def isRowIndexColumn(name: String, dataType: DataType): Boolean = {
  generatedRowIndexColumnNames.contains(name) ||
  (name == "_metadata" && (dataType match {
    case struct: StructType => struct.fieldNames.contains("row_index")
    case _ => false
  }))
}

private def scanReadsRowIndexColumn(scan: FileSourceScanExec): Boolean = {
  val outputFields = scan.output.iterator.map(attr => attr.name -> attr.dataType)
  val requiredFields = scan.requiredSchema.fields.iterator.map(field => field.name -> field.dataType)
  (outputFields ++ requiredFields).exists { case (name, dataType) =>
    isRowIndexColumn(name, dataType)
  }
}
```

A non-DV DML regression test using a user-defined `row_index` column would also lock in the intended scope.

felipepessoto (approved):

gluten-delta/src/main/scala/org/apache/gluten/extension/OffloadDeltaScan.scala
malinjawi: Yeah your right @felipepessoto  It's actually worse than nested structs a plain top-level column named `row_index` is enough. `DeleteCommand` sets the action type to `delete` before the `shouldWriteDVs` branch, so it says nothing about whether DVs are involved, and the bare `row_index` matched the user column straight away.

Also turns out narrowing to `_metadata.row_index` costs nothing, since that's the only mode the rule is live in. With `useMetadataRowIndex=false` the scan already falls back through `shouldFallbackDeletionVectorScanWithoutMetadataRowIndex`  the suite asserts that. And `addRowIndexIfMissing` requires `METADATA_NAME`, so the index is never top-level anyway.

Small bonus: the set had a duplicate. `ROW_INDEX_COLUMN_NAME` is `"__delta_internal_row_index"`, same string as the line above it.

Went with your suggestion, using `FileFormat.METADATA_NAME`. Added the non-DV DELETE test over a user `row_index` column too.

zhouyuan:
Cc @marin-ma

Issue #12195, opened 2026-05-30
[VL][DELTA] Support reading Delta Lake Change Data Feed (CDF) without falling back to vanilla Spark


felipepessoto:
## Description

Gluten currently does not offload reads of Delta tables' **Change Data Feed** (`spark.read.format("delta").option("readChangeFeed", "true")...` or the `table_changes()` SQL function). These queries run entirely on vanilla Spark instead of the Velox backend.

## Why it falls back today

A normal Delta scan is a `FileSourceScanExec` whose `relation.fileFormat` is a `DeltaParquetFileFormat`. Gluten's `OffloadDeltaScan` only matches that exact case and rewrites it into a `DeltaScanTransformer`:

```scala
case scan: FileSourceScanExec
    if scan.relation.fileFormat.getClass == classOf[DeltaParquetFileFormat] =>
  DeltaScanTransformer(scan)
```

CDF reads do **not** produce that plan. Delta builds them through `CDCReader.DeltaCDFRelation`, a generic `BaseRelation` whose `buildScan` returns RDD[Row]

Because the resulting plan is not a `FileSourceScanExec` over `DeltaParquetFileFormat`, `OffloadDeltaScan` never matches it, so the entire query (scan + projections building the metadata columns) stays on vanilla Spark.

## Proposed work

- Recognize the CDF scan path (`DeltaCDFRelation` / the CDC file indexes) and offload the underlying parquet reads to Velox.
- Materialize the synthesized `_change_type` / `_commit_version` / `_commit_timestamp` columns (literals + projections) so they can be produced natively rather than forcing a fallback.
- Add `gluten-ut` coverage for batch CDF reads (`readChangeFeed` and `table_changes()`), including add/remove/cdc-file combinations and column mapping.

### Gluten version

main branch

malinjawi:
Thanks @felipepessoto for raising this issue!

I took a closer look at this, and I think this is a valid Gluten/Velox Delta gap.

The fallback seems to happen before we reach the normal Delta scan offload path. Delta CDF reads are exposed to Spark as `CDCReader.DeltaCDFRelation`, so Gluten does not initially see the usual Delta file scan shape that `OffloadDeltaScan` handles.

[...]

PR #11924, opened 2026-04-12
[GLUTEN-10215][VL] Delta 4 native write support for Spark 4.0

malinjawi:
## What changes are proposed in this pull request?

This PR adds native Delta Lake write support for Spark 4.0 + Delta 4.0 in the Velox backend.

Main changes:
- Add Delta 4 write offload integration in Delta write operators and command offload path.
- Wire Delta-specific parquet file format and optimistic transaction write path for native execution.
- Add native write statistics tracking plumbing used by Delta write.
- Add Delta optimized writer offload integration for the Delta 4 path.

Scope note:
- This PR is focused on native write support only.
- Delta 4 test-suite expansion is split into separate PR(s) for reviewability.

Part of #10215.

## How was this patch tested?

- Compile/build validation with Spark 4.0 + Delta 4 profile:

```bash

./build/mvn -pl backends-velox -am -Pbackends-velox,delta,spark-4.0,scala-2.13,java-17 -DskipTests -Dcheckstyle.skip=true -Dspotless.skip=true -Dscalastyle.skip=true compile

PR #12218, opened 2026-06-01
[VL][DELTA] Support Delta CDF scan offload

malinjawi:

Addresses #12195.

Delta CDF reads enter Spark as `CDCReader.DeltaCDFRelation`, so they do not initially have the normal `FileSourceScanExec` + `DeltaParquetFileFormat` shape that Gluten's existing Delta scan offload rule recognizes.

This PR adds a Gluten Delta planner strategy, wired from the Velox Delta component, that recognizes `DeltaCDFRelation`, expands it through Delta's own CDF batch planning path, and rewrites the original projection/filter attributes onto the expanded logical plan. After that, the existing Delta scan offload path can plan the underlying CDF file scans as `DeltaScanTransformer`.

The change is intentionally scoped to Gluten's Delta/Spark planning layer rather than Velox C++:

- Add `DeltaCDFScanStrategy` for `table_changes(...)` and DataFrame `readChangeFeed` scans.
- Add Delta-version helper shims for Delta 2.3, 2.4, 3.3, and 4.x API differences.
- Register the planner strategy from `VeloxDeltaComponent`.
- Add Delta regression coverage for insert/update/delete CDF rows, filter/projection handling, bounded version reads, DataFrame `readChangeFeed`, column mapping, and a `startingVersion = 0` case.

One planner/test nuance: Delta CDF expansion can keep a Spark-side `ExistingRDD` branch for synthesized change rows, including on the tested update/delete CDF paths. The regression suite therefore compares results against vanilla Spark and asserts that the expanded CDF file scans are transformed to `DeltaScanTransformer`, rather than requiring the entire expanded CDF union to be globally fallback-free.

## How was this patch tested?

Local checks used `JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home`.

- `git diff --check`
- `./dev/format-scala-code.sh check`
- `./build/mvn -pl gluten-delta -am -Pspark-3.5 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests test-compile`
- `./build/mvn -pl gluten-delta -am -Pspark-4.0 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests test-compile`
- `./build/mvn -pl gluten-delta -am -Pspark-4.1 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests test-compile`

Earlier cross-version compile checks also passed before the test-expectation-only update:

- `./build/mvn -pl gluten-delta -am -Pspark-3.3 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests -Dcheckstyle.skip=true -Dscalastyle.skip=true -Dspotless.check.skip=true test-compile`
- `./build/mvn -pl gluten-delta -am -Pspark-3.4 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests -Dcheckstyle.skip=true -Dscalastyle.skip=true -Dspotless.check.skip=true test-compile`
- `./build/mvn -pl gluten-delta -am -Pspark-3.5 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests -Dcheckstyle.skip=true -Dscalastyle.skip=true -Dspotless.check.skip=true test-compile`
- `./build/mvn -pl gluten-delta -am -Pspark-4.0 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests -Dcheckstyle.skip=true -Dscalastyle.skip=true -Dspotless.check.skip=true test-compile`
- `./build/mvn -pl gluten-delta -am -Pspark-4.1 -Pscala-2.13 -Pjava-17 -Pbackends-velox -Pdelta -DskipTests -Dcheckstyle.skip=true -Dscalastyle.skip=true -Dspotless.check.skip=true test-compile`

Full native Velox runtime and benchmarking are left to CI / a native Gluten-optimized environment; this local checkout does not have `cpp/build/releases/libgluten.so` and the Velox external project build available.

Generated-by: IBM BOB


felipepessoto:
Could you add a DV-enabled CDF regression test to confirm the behavior? Something like:
 
```
  CREATE TABLE t (id INT, name STRING) USING delta
  TBLPROPERTIES ('delta.enableChangeDataFeed'='true', 'delta.enableDeletionVectors'='true');
  INSERT INTO t VALUES (1,'v1'),(2,'v2'),(3,'v3');
  DELETE FROM t WHERE id = 2;
  SELECT id, name, _change_type FROM table_changes('t', 0);
```
 
 compared against vanilla Spark (checkAnswer)

gluten-delta/src-delta40/main/scala/org/apache/gluten/extension/DeltaCDFRelationHelper.scala
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.gluten.extension
+
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.apache.spark.sql.delta.BatchCDFSchemaEndVersion
+import org.apache.spark.sql.delta.commands.cdc.CDCReader
+
+object DeltaCDFRelationHelper {
```
felipepessoto, line 23: Do you know what happens in Delta 4.1/4.2/4.3? Would it take the code from 4.0 folder?

malinjawi:
> Could you add a DV-enabled CDF regression test to confirm the behavior? Something like:
> 
> ```
>   CREATE TABLE t (id INT, name STRING) USING delta
>   TBLPROPERTIES ('delta.enableChangeDataFeed'='true', 'delta.enableDeletionVectors'='true');
>   INSERT INTO t VALUES (1,'v1'),(2,'v2'),(3,'v3');
>   DELETE FROM t WHERE id = 2;
>   SELECT id, name, _change_type FROM table_changes('t', 0);
> ```
> 
> compared against vanilla Spark (checkAnswer)

Thanks @felipepessoto  good point! I agree this is worth covering explicitly.

I’ll add a focused DV-enabled CDF regression with `delta.enableChangeDataFeed=true` and `delta.enableDeletionVectors=true`, comparing `table_changes(...)` against vanilla Spark via the existing `runAndCompare` / `checkAnswer` path.

One important nuance: this PR is not intended to claim full native DV support. DV-backed Delta scan execution is still a separate active area in Gluten/Velox and may continue to fall back where native DV handling is not available. The goal of the regression here is to make sure enabling DV does not break the CDF planning path or result correctness.

gluten-delta/src-delta40/main/scala/org/apache/gluten/extension/DeltaCDFRelationHelper.scala
malinjawi: Yes @felipepessoto  I checked the current Maven profile/source wiring:

- `spark-4.0` sets `delta.version=4.0.1` and `delta.binary.version=40`
- `spark-4.1` sets `delta.version=4.1.0` and `delta.binary.version=40`
- the Delta profile adds sources from `src-delta${delta.binary.version}/main/scala`

So with the current profile setup, both Spark 4.0 / Delta 4.0.x and Spark 4.1 / Delta 4.1.x use `gluten-delta/src-delta40/...`.

For future Delta 4.2 / 4.3 support, it depends on how those profiles are introduced. If they keep `delta.binary.version=40`, they will continue to use the same `src-delta40` helper. If Delta changes the relevant CDF APIs and Gluten introduces a new binary bucket, then we should add a new version-specific helper.

This also relates to the earlier folder-activation discussion in #11924: supporting family-level folders like `src-spark4` / `src-delta4` would make this cleaner and avoid copying code across Spark/Delta 4.x profiles when the APIs stay compatible. I think that belongs in a separate Maven/source-layout refactor rather than in this CDF PR.

gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaCDFScanStrategy.scala
```diff
@@ -0,0 +1,112 @@
...
+  // for a commit are only the rows the DV newly marks. Re-planning the analyzed CDF batch plan here
+  // loses that reconciliation on some Delta versions (e.g. Delta 2.4 / Spark 3.4), where the remove
+  // side would then surface every row of a logically-removed file as a `delete` change row instead
+  // of just the DV-marked ones. Decline to intercept such reads and let Spark serve them through
+  // Delta's own DeltaCDFRelation scan, which applies DVs correctly. Offloading loses nothing here:
+  // the native CDF scan path does not support DV reconciliation either (see the matching guard in
+  // DeltaScanTransformer).
+  private def touchesDeletionVectors(cdfRelation: CDCReader.DeltaCDFRelation): Boolean =
+    cdfRelation.snapshotWithSchemaMode.snapshot.metadata.configuration
+      .get("delta.enableDeletionVectors")
```
felipepessoto: This table property says if new writes can write DV. But I think you want to find if table has any DV

malinjawi: Good catch; you were right that the table property only controls future writes. This is now addressed in two layers:

- `3b462034a` makes `DeltaCDFScanStrategy` inspect `AddFile`/`RemoveFile` actions only in the requested CDF version range.
- Spark 3.4 CI then exposed an older physical fallback that still treated a DV-capable `TahoeFileIndex` as if every CDF range contained a DV. `049e608a3` now special-cases `CdcAddFileIndex` and `TahoeRemoveFileIndex` there as well, using their exact files/actions before the generic table-level fallback.

The regression covers both directions: a DV-enabled table's insert-only `0..1` range must offload, while a range containing a DV-backed delete must fall back even after `delta.enableDeletionVectors` is set back to `false`, with the CDF rows checked for correctness. Formatting and test compilation pass for Spark 3.3, 3.4, 3.5, 4.0, and 4.1.

backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
```diff
@@ -35,6 +35,8 @@ class VeloxDeltaComponent extends Component {
   }
 
   override def injectRules(injector: Injector): Unit = {
+    injector.spark.injectPlannerStrategy(DeltaCDFScanStrategy(_))
```
felipepessoto: In general, do we create individual flags for each feature we offload? If so, could we have one for CDF offload?

felipepessoto:
@malinjawi I have run the Delta CI with your changes, there are only a few failures related to filter push down, I guess it is not a blocker, but could you please double check?

https://github.com/apache/gluten/actions/runs/29772714951/job/88458293204?pr=12582
https://github.com/apache/gluten/actions/runs/29772714951/job/88458293208?pr=12582
https://github.com/apache/gluten/actions/runs/29772714951/job/88458293260?pr=12582
https://github.com/apache/gluten/actions/runs/29772714951/job/88458293225?pr=12582

felipepessoto (approved):

malinjawi:
Thanks @felipepessoto, I double-checked the combined Delta CI. The 13 reported regressions all had the same root cause: `DeltaCDFScanStrategy` expands the CDF relation during physical planning, after Spark's normal optimizer pass, so the reconstructed filter stayed above the CDF union and never reached the Delta file scans. Row correctness was unaffected, but this was a real predicate-pushdown regression rather than a baseline issue.

I fixed it in d740b0da2 by optimizing the resolved replacement plan before handing it back to physical planning. I also added a regression assertion that verifies the `id` predicate is present in `DeltaScanTransformer.dataFilters`. The change passes Spotless and clean test-compiles for Spark 3.3, 3.4, 3.5, 4.0, and 4.1.

Could you please rerun the combined Delta CI against the updated head? It should specifically clear the filter-pushdown failures from run 29772714951.

backends-velox/src-delta/main/scala/org/apache/gluten/component/VeloxDeltaComponent.scala
malinjawi: Yes, good suggestion. I added a dedicated experimental guard in 65178209a:

`spark.gluten.sql.columnar.backend.velox.delta.enableChangeDataFeedScan`

It defaults to `true`, is checked against the active session config on each planning pass, and setting it to `false` makes `DeltaCDFScanStrategy` decline the relation so Delta uses its vanilla CDF path. I also added a regression test for the disabled path and documented the setting.

malinjawi:
Update: the standard x86 Spark 3.3 job caught a separate output-attribute regression in the first pushdown fix. Optimizing the whole reconstructed plan allowed Spark 3.3 to remove the aliases that preserve the original CDF relation expression IDs, so a parent sort later tried to bind a stale `name` attribute.

I fixed that in bf2ac70f5 by optimizing only the filtered CDF child, then applying the output-restoring `Project` after the optimizer. This keeps predicate pushdown while preserving the expression IDs expected by already-planned parent operators. Spotless and Delta test-compiles pass for Spark 3.3 and Spark 4.1, and a fresh standard CI run is now queued on this head.

The combined Delta run 29772714951 is still the earlier run against the old combined branch. @felipepessoto, could you please rerun it against bf2ac70f5 once the branch is refreshed?

felipepessoto:
@malinjawi I synced your 3 new commits and started CI here: https://github.com/apache/gluten/pull/12582

felipepessoto:
@malinjawi I can see the Pushed Filters, but there is a difference, the expected filter contains a `*`, do you know what this mean?

```
...PushedFilters: [IsNotNull(`id with space`), LessThan(`id with space`,5)], ReadSchema: struct<col-215ff29d-abc0-48ea-92fc-0c09b73e9f40:bigint,col-c6bb9720-4fc4-4181-b423-bf2050fbb9dc:b... NativeFilters: [isnotnull(col-c6bb9720-4fc4-4181-b423-bf2050fbb9dc#1159077L),(col-c6bb9720-4fc4-4181-b423-bf2050fbb9dc#1159077L < 5)]
[info]   )" did not contain "PushedFilters: [*IsNotNull(id with space), *LessThan(id with space,5)]" (DeltaCDCColumnMappingSuite.scala:671)
```

I asked AI to explain, but I'm not familiar with this and I can't confirm, I hope you know more:

> What the  *  means
> 
> In a scan's  PushedFilters  metadata, a  *  prefix marks a filter the data source guarantees to fully evaluate itself, so Spark drops its own redundant  Filter  above the scan. It's emitted only by  RowDataSourceScanExec  ( DataSourceScanExec.scala:168 ):
> 
> if (handledFilters.contains(filter)) s"*$filter" else s"$filter"
> 
>  handledFilters  = pushed filters minus the source's  unhandledFilters(...) . A filter without  *  is offered to the source but Spark still re-applies it (best-effort, for skipping only).
> 
> Why the offloaded plan has no  * 
> 
> It's a different scan node, not a "less pushed-down" one.
> 
> Before the PR —  readChangeFeed  produces  DeltaCDFRelation , which is  BaseRelation with CatalystScan  and declares  unhandledFilters = Array.empty  ( CDCReaderBase.scala:109 ). Spark plans it as a  RowDataSourceScanExec ; since nothing is unhandled, every filter is "handled" and gets a  * :
> 
> PushedFilters: [*IsNotNull(id), *LessThan(id,5)]   ✅ test matches
> 
> After the PR —  DeltaCDFScanStrategy  expands that relation into the underlying Parquet plan, which Gluten offloads to  DeltaScanTransformer  (a  FileSourceScanLike ). Its metadata comes from  FileSourceScanLike.metadata  ( DataSourceScanExec.scala:444 ), which never marks filters:
> 
> "PushedFilters" -> seqToString(pushedFiltersForDisplay)   // no "*"
> 
> PushedFilters: [IsNotNull(id), LessThan(id,5)] ... NativeFilters: [...]   ❌ no "*"
> 
> So the exact substring  [*IsNotNull(id), *LessThan(id,5)]  disappears and the assertion fails.
> 
> The key point: the filter is still pushed down — it's in  DeltaScanTransformer.dataFilters  (your own test asserts exactly that). File-source scans (Parquet, hence Velox) never emit  * , because Parquet pushdown is best-effort row-group skipping and Spark always keeps a post-scan filter for correctness. The  *  is a  CatalystScan / PrunedFilteredScan  "I own this filter" marker that a file scan structurally cannot reproduce.


https://github.com/apache/gluten/actions/runs/30104513088/job/89521902732?pr=12582
https://github.com/apache/gluten/actions/runs/30104513088/job/89521902757?pr=12582
https://github.com/apache/gluten/actions/runs/30104513088/job/89521902748?pr=12582
https://github.com/apache/gluten/actions/runs/30104513088/job/89521902791?pr=12582

malinjawi:
> @malinjawi I can see the Pushed Filters, but there is a difference, the expected filter contains a `*`, do you know what this mean?
> 
> ```
> ...PushedFilters: [IsNotNull(`id with space`), LessThan(`id with space`,5)], ReadSchema: struct<col-215ff29d-abc0-48ea-92fc-0c09b73e9f40:bigint,col-c6bb9720-4fc4-4181-b423-bf2050fbb9dc:b... NativeFilters: [isnotnull(col-c6bb9720-4fc4-4181-b423-bf2050fbb9dc#1159077L),(col-c6bb9720-4fc4-4181-b423-bf2050fbb9dc#1159077L < 5)]
> [info]   )" did not contain "PushedFilters: [*IsNotNull(id with space), *LessThan(id with space,5)]" (DeltaCDCColumnMappingSuite.scala:671)
> ```
> 
> I asked AI to explain, but I'm not familiar with this and I can't confirm, I hope you know more:
> 
> > What the  *  means
> > In a scan's  PushedFilters  metadata, a  *  prefix marks a filter the data source guarantees to fully evaluate itself, so Spark drops its own redundant  Filter  above the scan. It's emitted only by  RowDataSourceScanExec  ( DataSourceScanExec.scala:168 ):
> > if (handledFilters.contains(filter)) s"*$filter" else s"$filter"
> >  handledFilters  = pushed filters minus the source's  unhandledFilters(...) . A filter without  *  is offered to the source but Spark still re-applies it (best-effort, for skipping only).
> > Why the offloaded plan has no  * 
> > It's a different scan node, not a "less pushed-down" one.
> > Before the PR —  readChangeFeed  produces  DeltaCDFRelation , which is  BaseRelation with CatalystScan  and declares  unhandledFilters = Array.empty  ( CDCReaderBase.scala:109 ). Spark plans it as a  RowDataSourceScanExec ; since nothing is unhandled, every filter is "handled" and gets a  * :
> > PushedFilters: [*IsNotNull(id), *LessThan(id,5)]   ✅ test matches
> > After the PR —  DeltaCDFScanStrategy  expands that relation into the underlying Parquet plan, which Gluten offloads to  DeltaScanTransformer  (a  FileSourceScanLike ). Its metadata comes from  FileSourceScanLike.metadata  ( DataSourceScanExec.scala:444 ), which never marks filters:
> > "PushedFilters" -> seqToString(pushedFiltersForDisplay)   // no "*"
> > PushedFilters: [IsNotNull(id), LessThan(id,5)] ... NativeFilters: [...]   ❌ no "*"
> > So the exact substring  [*IsNotNull(id), *LessThan(id,5)]  disappears and the assertion fails.
> > The key point: the filter is still pushed down — it's in  DeltaScanTransformer.dataFilters  (your own test asserts exactly that). File-source scans (Parquet, hence Velox) never emit  * , because Parquet pushdown is best-effort row-group skipping and Spark always keeps a post-scan filter for correctness. The  *  is a  CatalystScan / PrunedFilteredScan  "I own this filter" marker that a file scan structurally cannot reproduce.
> 
> https://github.com/apache/gluten/actions/runs/30104513088/job/89521902732?pr=12582 https://github.com/apache/gluten/actions/runs/30104513088/job/89521902757?pr=12582 https://github.com/apache/gluten/actions/runs/30104513088/job/89521902748?pr=12582 https://github.com/apache/gluten/actions/runs/30104513088/job/89521902791?pr=12582

Thanks @felipepessoto. I confirmed the explanation is correct: * marks a filter reported as fully handled. The predicates are still present in DataFilters, PushedFilters, and NativeFilters, so the CDF pushdown and results are correct.
I also saw the shared * rendering change in #12582. Since it affects all file scans, I think it should be reviewed separately from this CDF PR.

Could you sync the latest head into #12582 and rerun the Delta CI once more?

felipepessoto:
@malinjawi new run started

malinjawi:
Thanks @felipepessoto. I checked the completed run on #12582. It includes the latest #12218 head and reports tests with zero new regressions.

cc: 
@zhztheplayer @zhouyuan  Could you have a look?

gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaCDFScanStrategy.scala
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.expressions.{Alias, Attribute, AttributeReference, Expression, NamedExpression}
+import org.apache.spark.sql.catalyst.planning.PhysicalOperation
+import org.apache.spark.sql.catalyst.plans.logical.{Filter, LocalRelation, LogicalPlan, Project}
+import org.apache.spark.sql.delta.actions.{AddFile, RemoveFile}
+import org.apache.spark.sql.delta.commands.cdc.CDCReader
+import org.apache.spark.sql.execution.{SparkPlan, SparkStrategy}
+import org.apache.spark.sql.execution.datasources.LogicalRelation
+
+case class DeltaCDFScanStrategy(spark: SparkSession, offloadEnabled: () => Boolean)
```
zhztheplayer: Why a strategy is needed instead of a pre-transform columnar rule?

malinjawi: Good question @zhztheplayer  I used a planner strategy because from what I see Gluten’s pre-transform rules run after Spark has already converted `DeltaCDFRelation` into a `RowDataSourceScanExec`, where the underlying file scans are hidden inside the RDD. Expanding it during planning keeps those scans visible for the normal Gluten offload rules. 

If there’s an existing pre-transform pattern that handles this replanning cleanly, I’m happy to look into it. Do you have another implementation or did I miss something?

felipepessoto:
The CI pipeline is approved. Hopefully it will be merged soon so we wont need manual merges across PRs to validate it.

gluten-delta/src/main/scala/org/apache/gluten/extension/DeltaCDFScanStrategy.scala
malinjawi: @zhztheplayer I looked into this deeper.

I Switched to a logical rule in`DeltaCDFScanRule`, injected at post-hoc
resolution. Running before the optimizer means only an output-restoring projection is
needed; Spark pushes the predicates into the expanded scans itself. That drops the
`PhysicalOperation` matching, the filter/project rewrites and the out-of-band
`optimizer.execute`.

Let me know if this is a better design.

zhztheplayer (approved):

felipepessoto:
@malinjawi @zhztheplayer if you don't mind, I suggest holding the merge until we have the Delta CI pipeline, unless it takes too long to be merged.

felipepessoto:
@malinjawi could you make some change to retrigger CI, it will now include Delta CI

felipepessoto:
Thanks @malinjawi, the remaining seems the same you investigated before: https://github.com/apache/gluten/pull/12218#issuecomment-5103016490

I opened an issue: https://github.com/apache/gluten/issues/12753

Could you add these lines:

.github/workflows/util/delta-spark-ut/known-failures.txt

```
org.apache.spark.sql.delta.DeltaCDCIdColumnMappingSuite#filters with special characters in name should be pushed down - column mapping id mode
org.apache.spark.sql.delta.DeltaCDCNameColumnMappingSuite#filters with special characters in name should be pushed down - column mapping name mode
org.apache.spark.sql.delta.DeltaCDCSQLIdColumnMappingSuite#filters with special characters in name should be pushed down - column mapping id mode
org.apache.spark.sql.delta.DeltaCDCSQLNameColumnMappingSuite#filters with special characters in name should be pushed down - column mapping name mode
org.apache.spark.sql.delta.DeltaCDCSQLSuite#select individual column should push down filters
org.apache.spark.sql.delta.DeltaCDCSQLWithCatalogOwnedBatch100Suite#select individual column should push down filters
org.apache.spark.sql.delta.DeltaCDCSQLWithCatalogOwnedBatch1Suite#select individual column should push down filters
org.apache.spark.sql.delta.DeltaCDCSQLWithCatalogOwnedBatch2Suite#select individual column should push down filters
org.apache.spark.sql.delta.DeltaCDCScalaSuite#filters should be pushed down
org.apache.spark.sql.delta.DeltaCDCScalaWithCatalogOwnedBatch100Suite#filters should be pushed down
org.apache.spark.sql.delta.DeltaCDCScalaWithCatalogOwnedBatch1Suite#filters should be pushed down
org.apache.spark.sql.delta.DeltaCDCScalaWithCatalogOwnedBatch2Suite#filters should be pushed down
org.apache.spark.sql.delta.DeltaCDCScalaWithDeletionVectorsSuite#filters should be pushed down
```

After the

`org.apache.spark.sql.delta.DeltaArbitraryColumnNameSuite#create table`

zhztheplayer (approved):

PR #12221, opened 2026-06-02
[VL] Fixed iceberg table properties config logic

infvg:
The previous logic overrode SQLConf property with TableProperty, while not setting any value when SQLConf not set. Now it will prioritize SQLConf, then TableProperty


zhouyuan (approved):

zhouyuan:
This aligns the java iceberg behavior

PR #12222, opened 2026-06-02
[GLUTEN-10215][VL] Remove streaming aggregation workaround in Delta stats writer

zhztheplayer:
Velox now supports hash aggregation task barrier. We can now safely remove the previous dummy-keyed streaming aggregation workaround in Delta stats writer.

Related to https://github.com/apache/gluten/pull/11419.

Related issue: #10215


zhouyuan (approved):

zhztheplayer:
Overall benchmark (tpc-ds write) result is stable, except `store_returns` gets 15% faster than it was before.

